PR #1193, opened 2022-11-28
Adding product tags to most of the documentation

reshmakh:


codyebberson (approved):
Nice.

One tricky part is that most of the files in `docs/api/fhir/resources` are autogenerated.  We've known that we need some way to "merge" rather than "overwrite", so this PR will force that issue.  But I'm good with it 👍

reshmakh:
Abandoning this one, as I'm going to use another approach here.

Issue #992, opened 2022-09-19
Wishlist: Terminology Service support


goelalex:
***This is a wishlist item*** 

Ability to pull down and validate terminology from an external service like VSAC (vsac.nlm.nih.gov/) or something else and check that the codes that have been received are valid. Similar to the FHIR validation use case: https://github.com/medplum/medplum/issues/389 (I recognize it should be covered already). E.g. receiving a resource and checking that the LOINC, SNOMED, or RXnorm code is valid from the sender. 

Even better would be the ability to embed a standard code in a newly created resource e.g. a doctor is creating a medication and we can add an RXNorm code to it immediately.

reshmakh:
[Terminology reference](https://build.fhir.org/ig/HL7/fhir-mCODE-ig/terminology.html)

codyebberson:
Implementation discussion: https://github.com/medplum/medplum/discussions/1411

codyebberson:
Medplum terminology is live

See:
* https://www.medplum.com/docs/terminology
* https://www.medplum.com/docs/terminology/medplum-terminology-services
* https://www.medplum.com/docs/contributing/terminology-architecture

PR #1213, opened 2022-12-04
Support terminologies in ValueSet expand operation

codyebberson:
Before:
* We did not actually create `ValueSet` and `CodeSystem` resources
* We precomputed everything for the `ValueSet/$expand` operation
* It was not possible to use the `$expand` operation for custom resources
* It was not possible to follow a `ValueSet.url` to an external terminology such as SNOMED, LOINC, etc

After:
* We create `ValueSet` and `CodeSystem` resources
* We evaluate the mapping of `ValueSet` to `system` (s) at runtime
* We re-index `ValueSet` and `CodeSystem` resources codes, using the same pattern as `Address`, `ContactPoint`, `HumanName`, etc
* This enables `$expand` for custom resources
* This enables `$expand` for external terminologies


codyebberson:
Big step toward https://github.com/medplum/medplum/issues/992

PR #1214, opened 2022-12-05
UMLS data generator utility

codyebberson:
Checkpoint for a utility that generates bite sized files out of the 30 GB UMLS data set.

There is separate future work to ingest these files by the server. The terminologies are too big to be stored in the naive default "JSON string in a TEXT column" method.

Options for consideration:
* Completely separate super admin endpoint to ingest the files directly - path of least resistance, allows for fine grained control over behavior
* Accept huge ValueSet resources, but strip out all `"concept"` properties - more semantically correct, allows for end users to submit their own large terminologies

Reference:  HAPI FHIR docs on importing terminologies: https://hapifhir.io/hapi-fhir/docs/tools/hapi_fhir_cli.html#upload-terminology

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ packages/definitions/dist/**/*.js.map
 packages/definitions/dist/**/*.d.ts
 packages/examples/dist/
 packages/generator/dist/
+packages/generator/output/
 packages/graphiql/dist/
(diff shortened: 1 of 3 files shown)


reshmakh:
Related to #992

reshmakh (approved):

PR #1229, opened 2022-12-08
Fixing search params in ResourceInput

### LINKED ISSUE #1226: ResourceInput react component uriEncodes all query parameters twice
## To demonstrate

- go to https://app.medplum.com/Patient/:id/edit (for some patient id)
- From there, add a link to another patient and type "test patient" into the patient search bar
- Check the dev console and notice how the query sent to the api is `https://api.medplum.com/fhir/R4/Patient?name=test%2520patient&_count=10` 
- Note that the query should be `https://api.medplum.com/fhir/R4/Patient?name=test%20patient&_count=10` 

## Relevance

This happens for all queries and it is most important because it makes it impossible to search for emails (the @ symbol is uri encoded making the search not match)

## To Fix

Change this line

https://github.com/medplum/medplum/blob/f8c64b633e6ac20b16e8a08832751c00f72b22de/packages/react/src/ResourceInput/ResourceInput.tsx#L53

```diff
-       [searchCode]: encodeURIComponent(input),
+       [searchCode]: input,
```

### LINKED ISSUE #1228: ResourceInput for users should match `email:contains` instead of email
## Relavance

Currently, the resource input for user searches users by `User?email=<search input>`. This returns without error, but it will only return users whose email EXACTLY matches the search input. This goes against users' expectations because the search results are there to provide auto-complete functionality.

## To Demonstrate

- Spin up a local medplum stack
- Navigate to `http://localhost:3000/Project/:id/edit` (for some project id) 
- Try typing in the project owner field to search for a user by email
- Notice how no users appear unless the email exactly matches the user's email (well, currently even that doesn't work due to #1226)

## Fix

There is a quick fix for this that will allow search to find any users whose email contains the search string by changing this line:

https://github.com/medplum/medplum/blob/f8c64b633e6ac20b16e8a08832751c00f72b22de/packages/react/src/ResourceInput/ResourceInput.tsx#L24

```diff
-   User: 'email',
+   User: 'email:contains',
```

msdrigg:
This pr performs two improvements

- Allow searching for users by partial email (Fixes #1228)
- Don't double-encode all search params as uri string (Fixes #1226)

I can split these up into two pr's if you would like. The changes were so small I didn't think it necessary.

### DIFF
--- a/packages/react/src/ResourceInput/ResourceInput.tsx
+++ b/packages/react/src/ResourceInput/ResourceInput.tsx
@@ -21,7 +21,7 @@ const SEARCH_CODES: Record<string, string> = {
   Observation: 'code',
   RequestGroup: '_id',
   ActivityDefinition: 'name',
-  User: 'email',
+  User: 'email:contains',
 };
 
 export interface ResourceInputProps<T extends Resource = Resource> {
@@ -50,7 +50,7 @@ export function ResourceInput<T extends Resource = Resource>(props: ResourceInpu
     setLoading(true);
     const searchCode = SEARCH_CODES[props.resourceType] || 'name';
     const searchParams = new URLSearchParams({
-      [searchCode]: encodeURIComponent(input),
+      [searchCode]: input,
       _count: '10',
     });
 


codyebberson (approved):

codyebberson:
Thank you @msdrigg !

PR #1232, opened 2022-12-09
CLI with self-hosted medplum

natemerrill:
This change makes it possible to use the CLI with self-hosted Medplum by adding another environment variable. Didn't think it was a big enough change for a feature request, but if there's more to this than I thought, feel free to reject and I can open an issue. Thanks!


codyebberson (approved):

codyebberson:
Thanks @natemerrill !

PR #1322, opened 2022-12-11
**WARNING: SECURITY** [infra] Lets ECS reach ECR

msdrigg:
**THIS PR HAS SECURITY IMPLICATIONS**

**PLEASE THOROUGHLY REVIEW TO DECIDE IF THIS IS SOMETHING YOU WANT IN YOUR INFRA**

This PR allows the ECS container and Fargate agents to access private Amazon ECR repositorys. This change allows the medplum ecs server `TaskDefinition` to point a private ecr repository in the account that hosts the ECS task. Currently if you try to do this, the ECS task creation will fail with an error saying that ECS could not connect to the ECR repository.

- This commit changes the main CDK backend construct to add an IAM Role with the `AmazonECSTaskExecutionRolePolicy` managed policy
- See this guide here for my motivation for adding this policy https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html
- This policy provides the following permissions to the task creation agents

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "ecr:GetAuthorizationToken",
        "ecr:BatchCheckLayerAvailability",
        "ecr:GetDownloadUrlForLayer",
        "ecr:BatchGetImage",
        "logs:CreateLogStream",
        "logs:PutLogEvents"
      ],
      "Resource": "*"
    }
  ]
}
```

By default, CDK only provides permissions to create log streams and put log events.

## Motivation

- Allow people self-hosting Medplum on aws infra to use a private aws ecr image as their server image (currently this is not possible)

## Drawbacks

- Wider security surface area
- If this hasn't been a problem for hosted medplum infra so far, it isn't an issue for you

### DIFF
--- a/packages/infra/src/backend.ts
+++ b/packages/infra/src/backend.ts
@@ -213,10 +213,16 @@ export class BackEnd extends Construct {
       },
     });
 
+    const taskDefinitionRole = new iam.Role(this, 'TaskDefinitionRole', {
+      assumedBy: new iam.ServicePrincipal('ecs-tasks.amazonaws.com'),
+      managedPolicies: [iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AmazonECSTaskExecutionRolePolicy')],
+    });
+
     // Task Definitions
     const taskDefinition = new ecs.FargateTaskDefinition(this, 'TaskDefinition', {
       memoryLimitMiB: config.serverMemory,
       cpu: config.serverCpu,
+      executionRole: taskDefinitionRole,
       taskRole: taskRole,
     });
 


codyebberson:
Thanks @msdrigg - I'll review soon.  I agree with the feature in principle.  One idea to limit impact is to wrap this in a config setting (i.e., something like "ecrEnabled"), such that this is not executed by default.

msdrigg:
Great @codyebberson. Since the infra package already needs to parse the `cdk-config.json` file. It even would be possible to parse the config to see if `config.serverImage` matches `ACCOUNT.dkr.ecr.REGION.amazonaws.com/*` and only create the resource if that is the case. 

I am happy to make the change and publish an update if you say the word

msdrigg:
But then again, explicit is generally better than implicit especially when it comes to security so I don't know if that is a beneficial feature -- you just let me know what you think is best and I'll update the pr

codyebberson:
> Great @codyebberson. Since the infra package already needs to parse the `cdk-config.json` file. It even would be possible to parse the config to see if `config.serverImage` matches `ACCOUNT.dkr.ecr.REGION.amazonaws.com/*` and only create the role if that is the case.
> 
> That way users wouldn't have any additional complexity but the security would be higher.
> 
> I am happy to make the change and publish an update if you say the word

I really like that idea!  Let's do it.  Thanks @msdrigg !

msdrigg:
@codyebberson Ok, so I found what I think is a perfect solution.

So I did some research and apparently aws cdk has an explicit `EcrImage` resource that is a subclass of `ContainerImage`. If we use `EcrImage`, CDK will automatically create fine grained permissions to allow the fargate agent to access the specific ecr image.

This means that all we need to do is test if the `serverImage` is a ecr image and then create the `EcrImage` container explicitly.

I verified for myself that `npx cdk diff` should show no difference between this PR and your existing setup (but of course I can't be sure what your setup looks like exactly, so please test for yourself).

I also verified that this works for self-hosted with a private ecr repository and it creates the following permission diff

```
IAM Statement Changes
┌───┬─────────────────────────────────────────────────────────────┬────────┬─────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────┬───────────┐
│   │ Resource                                                    │ Effect │ Action                                                      │ Principal                                                   │ Condition │
├───┼─────────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────┼───────────┤
│ + │ *                                                           │ Allow  │ ecr:GetAuthorizationToken                                   │ AWS:${BackEnd/TaskDefinition/ExecutionRole}                 │           │
├───┼─────────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────┼───────────┤
│ + │ arn:aws:ecr:us-east-1:ACCOUNT:repository/medplum-           │ Allow  │ ecr:BatchCheckLayerAvailability                             │ AWS:${BackEnd/TaskDefinition/ExecutionRole}                 │           │
│   │ server                                                      │        │ ecr:BatchGetImage                                           │                                                             │           │
│   │                                                             │        │ ecr:GetDownloadUrlForLayer                                  │                                                             │           │
└───┴─────────────────────────────────────────────────────────────┴────────┴─────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────┴───────────┘
```

(And some other diffs referencing the `EcrImage` resource

codyebberson (approved):

codyebberson:
Thank you @msdrigg !

PR #1337, opened 2022-12-15
Added UI elements to initiate Smart App Launch

codyebberson:


reshmakh (approved):

reshmakh:
Adding some breadcrumbs here for documentation purposes.  

Now the /Encounter/<id>/apps page and the /Patient/<id>/apps tab on app.medplum.com will have the appropriate link to initiate the SmartApp Launch.

Filed #1339 to add documentation

packages/app/src/resource/AppsPage.tsx
```diff
@@ -24,18 +31,57 @@ export function AppsPage(props: AppsPageProps): JSX.Element {
...
+      })
+      .catch((err) => showNotification({ color: 'red', message: normalizeErrorString(err) }));
+  }
+
   return (
     <Document>
       {questionnaires.map((questionnaire) => (
         <div key={questionnaire.id}>
-          <h3>
+          <Title order={3}>
```
rahul1, line 64: Unrelated to this PR, but it would be better for this link to read from `questionnaire.title` rather than `questionnaire.name`. 

`questionnaire.name` is supposed to be a more computer-friendly identifier, but the `title` is the human readable one

codyebberson: I updated it to prefer `questionnaire.title`, but still fallback to `questionnaire.name`, because it looks like there are a bunch of questionnaires with names but not titles.

PR #1338, opened 2022-12-15
Fixes #1250 - client credentials refresh

### LINKED ISSUE #1250: MedplumClient should automatically refresh when using client ID/secret
Setup MedplumClient:

```
    const medplum = new MedplumClient({ fetch });
    await medplum.startClientLogin(MY_CLIENT_ID, MY_CLIENT_SECRET);
```

If it runs for more than an hour, eventually it stops working.  With client ID and secret, it should automatically refresh.

codyebberson:
Consider:

```ts
    const medplum = new MedplumClient({ fetch });
    await medplum.startClientLogin(MY_CLIENT_ID, MY_CLIENT_SECRET);
```

Before: After running for more than an hour, it stops working.

After:  It automatically "refreshes" with client ID and secret.  Behind the scenes, it uses `client_credentials` flow, not `refresh_token`.

Also, `client_credentials` no longer includes refresh tokens.  This was incorrect, and is explicitly forbidden in the spec:
* https://www.rfc-editor.org/rfc/rfc6749#section-4.4.3
* https://community.auth0.com/t/how-to-get-refresh-token-with-client-credentials/7028
* https://stackoverflow.com/questions/43340580/oauth-client-credential-flow-refresh-tokens

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -2,6 +2,7 @@ import { Bundle, Patient, SearchParameter, StructureDefinition } from '@medplum/
 import { webcrypto } from 'crypto';
 import PdfPrinter from 'pdfmake';
 import type { CustomTableLayout, TDocumentDefinitions, TFontDictionary } from 'pdfmake/interfaces';
+import { URLSearchParams } from 'url';
 import { TextEncoder } from 'util';
 import { MedplumClient, NewPatientRequest, NewProjectRequest, NewUserRequest } from './client';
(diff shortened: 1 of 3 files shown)


reshmakh:
This is a great one - does one more test case make sense here for the automatic refresh case?

codyebberson:
> This is a great one - does one more test case make sense here for the automatic refresh case?

I extended the existing client credentials test to invoke the refresh flow.

PR #1350, opened 2022-12-17
Client assigned IDs - part 1

codyebberson:
See https://github.com/medplum/medplum/discussions/1175


reshmakh (approved):

codyebberson:
This PR is too far gone.  Not only is it woefully out of date, we also changed our proposed strategy to using V5 UUIDs which is fundamentally different.

leite08:
On the comment about using UUID v5, have you considered v7? It addresses the issue with bad performance of random UUIDs for indexing. It's very close to be active/published and there are already at least a couple implementations in Typescript out there: https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/07/

PR #1359, opened 2022-12-18
Auto-batch proof-of-concept

codyebberson:
Q: What is a "batch request"?

A: FHIR includes "batch" support which allows you to make multiple FHIR API calls in a single HTTP request.  This can be a performance improvement for both the client and the server.

Q: What's an example of a "batch request"?

A: Imagine visiting the a Patient page.  To render the page, we need to load:

1. The resource
2. The resource history
3. Potentially the resource type schema StructureDefinition
4. Related resources such as Communication, ServiceRequest, DiagnosticReport, etc

The Medplum app packages all of those together into a batch request.

Q: What is "auto batching"?

Before, if a developer wanted to make a batch request, they had to create the batch request manually.  The semantics are subtly different between "make a request" and "make a request as part of a batch", so this had some cognitive overhead.  Also, you had to organize all of the requests in a single place, which made it difficult to successfully batch requests in separate React components.

With auto batching, all of that happens seamlessly behind the scenes.  MedplumClient adds a small delay (i.e., ~20 ms) to every FHIR request.  At the end of the delay, it groups together all of the requests into a batch request.  After executing the batch, MedplumClient returns the results as if the requests were made independently.

Q: Is auto batching required?

A: No, not at all.  It is off by default.  It is opt in.

Q: Are there other benefits?

A: I'm so glad that you asked.

Auto batching also works better for resource caching.  Consider the previous example of visiting a Patient page.  Consider that 1+ of those requests is already in the cache.  Before, with manual batching, we would not take advantage of preexisting cache values.  With auto batching, MedplumClient only makes the requests that it needs to make.

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1,10 +1,11 @@
-import { Bundle, Patient, SearchParameter, StructureDefinition } from '@medplum/fhirtypes';
+import { Bundle, BundleEntry, Patient, SearchParameter, StructureDefinition } from '@medplum/fhirtypes';
 import { webcrypto } from 'crypto';
 import PdfPrinter from 'pdfmake';
 import type { CustomTableLayout, TDocumentDefinitions, TFontDictionary } from 'pdfmake/interfaces';
 import { URLSearchParams } from 'url';
 import { TextEncoder } from 'util';
 import { MedplumClient, NewPatientRequest, NewProjectRequest, NewUserRequest } from './client';
+import { getStatus, notFound } from './outcomes';
 import { ProfileResource, stringify } from './utils';
 
 const defaultOptions = {
@@ -59,7 +60,16 @@ let tokenExpired = false;
 
 function mockFetch(url: string, options: any): Promise<any> {
   const { method } = options;
+  const response = mockHandler(method, url, options);
+  return Promise.resolve({
+    ok: response.status === 200 || response.status === undefined,
+    status: response.status,
+    blob: () => Promise.resolve(response),
+    json: () => Promise.resolve(response),
+  });
+}
 
+function mockHandler(method: string, url: string, options: any): any {
   let result: any;
 
   if (method === 'POST' && url.endsWith('auth/login')) {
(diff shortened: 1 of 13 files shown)


packages/core/src/client.ts
```diff
@@ -1859,6 +1904,53 @@ export class MedplumClient extends EventTarget {
     return obj;
   }
 
+  /**
+   * Executes a batch of requests that were automatically batched together.
+   */
+  async #executeAutoBatch(): Promise<void> {
+    // Get the current queue
+    const entries = [...this.#autoBatchQueue];
```
rahul1, line 1912: Do we have to do any locking of the queue here?

packages/core/src/client.ts
```diff
@@ -113,6 +114,15 @@ export interface MedplumClientOptions {
...
   cacheTime?: number;
 
+  /**
+   * The length of time in milliseconds to delay requests for auto batching.
+   *
+   * Auto batching attempts to group multiple requests together into a single batch request.
+   *
+   * Default value is 0, which disables auto batching.
+   */
+  autoBatchTime?: number;
```
rahul1, line 124: One small request - can we get a getter/setter for autoBatchTime? As a component developer, I might want to enable autobatching for a specific component, and then disable again without creating a new client. 

But i'm not sure if that breaks with concurrent rendering

rahul1 (approved):
LGTM, with one potential small change

packages/core/src/client.test.ts
```diff
@@ -1041,6 +1067,27 @@ test('graphql variables', async () => {
   expect(body.variables).toBeDefined();
 });
 
+test('Auto batch single request', async () => {
+  const medplum = new MedplumClient({ ...defaultOptions, autoBatchTime: 100 });
+  const patient = await medplum.readResource('Patient', '123');
+  expect(patient).toBeDefined();
+});
+
+test('Auto batch multiple requests', async () => {
```
rahul1, line 1076: Would be good to add a few  tests for non-`readResource` cases (e.g readHistory, readReference)

PR #1373, opened 2022-12-23
Only set meta.account on resource creation

rahul1:
Previously, the `meta.account` property could be overwritten on every resource update. This caused some problems with a customer, where a customer service update of a resource overwrote the account of a patient, hiding visibility from downstream clients. 

This PR implements what is probably the more correct default behavior. The `meta.account` property is only updated when a resource is created, and then persists throughout updates

### DIFF
--- a/packages/server/src/fhir/accesspolicy.test.ts
+++ b/packages/server/src/fhir/accesspolicy.test.ts
@@ -429,6 +429,68 @@ describe('AccessPolicy', () => {
     }
   });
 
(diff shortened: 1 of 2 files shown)


codyebberson (commented):
Looks good from a technical perspective.  I haven't thought through all operational consequences.

I.e., are there cases where (1) create a ServiceRequest without subject, and (2) set subject later?  I think account would typically be set by access policy in step 1?

codyebberson (approved):
Discussed offline 👍

Issue #1381, opened 2023-01-08
Enable Okta authentication via OpenID


codyebberson:
The basics of "auth with Okta" are straightforward.  The challenge here will be weaving it into the current auth flow, which is quite complex.

In a perfect world, we could add Okta simply as another auth provider at the beginning of the flow:

![image](https://user-images.githubusercontent.com/749094/211216216-1dccd7c3-b23c-4b0f-a793-fee2eb1d5fd1.png)

Unfortunately, Okta partitions auth endpoints by customer, so there is no global auth endpoint.  Instead, you need to know the domain by customer.

Options:
1. Global rule by email address
  a. For example, first ask for email.  If the email address ends with "@example.com" then redirect to the Okta auth page configured for "example.com".  This is what Box does.
  b. This is probably the simplest and least intrusive option.
  c. Important distinction between "global user" and "project user".  By default, we configure most users as "global users", meaning that they can be members of multiple projects.  For open patient registration, we use "project user", which means that the user account is isolated to the project.  We should only apply this global Okta rule to global users, and not project users.
2. Ask for project at top of sign-in
  a. For example, ask for a project ID or a project alias as the first field in the sign-in page. This is what the AWS console does.
  b. This would simplify things for Medplum engineering, but it's a pretty bad experience for users, because now they need to remember another piece of information.  Even if we make it easy to use friendly alias names like "foo", it will be confusing and will probably increase support tickets.

reshmakh:
This is a hard one - as there aren't really any good options that you need to know a lot alot about a user in order to choose the right authentication method - but we are authenticating them because we don't know who they are.  

Both options are viable, but there are drawbacks to each.  Our multi-project UI is another good experience that would be hampered by this as well.  

It does seem like Option 1 is preferable to Option 2 as we it is least intrusive, and represents the spirit of what customers want.  

We need to improve the error message for developers who are using the register and sign in react components to ensure that they are passing in project Ids to streamline the login for end users.  

Question: this will force Okta login for customers with linked emails if they are developing on localhost right?  Will Okta allow that?

codyebberson:
> Question: this will force Okta login for customers with linked emails if they are developing on localhost right? Will Okta allow that?

Oo.  Good catch.  Not only localhost, but all separate front-ends.

rahul1:
Just chiming in here, it does seem like option 1 is preferrable if we can make it work. We have a lot of name collisions for Project names, so creating some kind of alias system for them would be difficult at a practical level as well.

[...]

PR #1419, opened 2023-01-19
Updating auth documentation Okta auth related to #1381

reshmakh:


codyebberson (changes requested):
Looks good!  2 minor style issues

packages/docs/docs/auth/index.md
```diff
@@ -12,6 +12,22 @@ Authentication (_are you who you say you are?_) and authorization (_what can you
...
+| ----------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------ |
+| Credentials | In the Credentials phase of login the authentication provider is determined (Medplum auth, Google auth, Okta, etc.), the credentials are collected and user authenticated with the auth provider of choice. If Okta is enabled for a domain, all emails at that domain will go to Okta for authentication. | `auth/login` <br /> `auth/external` <br /> |
+| Profile     | In the Profile phase, if the user is a member of multiple projects, one must be selected to proceed                                                                                                                                                                                                        | `auth/profile`<br />                       |
```
reshmakh, line 22: @codyebberson I like this third column that outlines the endpoints, I think it makes it more concrete - but wanted your feedback.

packages/docs/docs/auth/okta-auth.md
```diff
@@ -0,0 +1,86 @@
...
+- Client ID
+- Client Secret
+
+Okta uses separate domains per organization. You will need your organization's Okta server. This is called the Okta "baseUrl". You can find it in the top-left menu of the Okta admin panel by clicking on your name when logged in.
+
+![Okta Find URL](/img/auth/okta-find-url.png)
+
+The "baseUrl" could look something like this `dev-71569655.okta.com` or could be a named subdomain like `mydomain.okta.com`. Once you have obtained it, construct the Authorize, Token and UserInfo URLs as follows:
+
+Authorize URL = ${baseUrl}/v1/authorize (e.g., "https://dev-71569655.okta.com/oauth2/v1/authorize")
```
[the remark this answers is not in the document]
reshmakh: good catch!

packages/docs/docs/auth/index.md
codyebberson: Not sure what the goal is of the section?  The "/auth/" endpoints are pseudo-internal endpoints at the moment.  I.e., they're an implementation detail, and not intended for customers to use directly.  No expectations of backwards compatibility, etc.

That is in contrast to the "/oauth/" endpoints, which are well defined with the OpenID/OAuth spec, which we are committed to following.

If necessary, then we could change that, and decide to make them official public API's.  If that's the case, then we would need more details.  I.e., list out each of the endpoints with inputs and outputs for each.

I also don't think this belongs on the top-level Auth page.  It is deep in the weeds, and (in my opinion) this page should be more about big picture of auth.

With regards to the actual content...

Suggestions:

- Split "Users can belong to multiple Medplum projects, and the service supports multiple types of authentication. " into two sentences.
- Let's move the diagram above the table?
- "There are three major stages in the login flow." - there are now 4 major stages if you include "Domain"
- `auth/google` could be added to the "Endpoints" column
- Other endpoints?
  - `auth/method` to determine auth methods for an email address
  - `auth/me` for details about current login
  - `auth/mfa` for 2FA input
  - `auth/newpatient` for new patient registration
  - `auth/newuser` for new user registration
  - `auth/newproject` for new project registration
  - etc

Overall.... I think we need to be clear and consistent about the intentions of the page.

reshmakh: This is an [explanation](https://diataxis.fr/explanation/) style page, and so I'm taking cues off of that.  The endpoints, though an implementation detail, give people a framework to understand what's going on and introduce them to concepts that they will need to understand like "Project", "Profile" which are Medplum specific.  

This is challenging, and it is rare to see good documentation on this front, but I think that a peek under the hood here will help in terms of clarity.  Personally, browsing the auth endpoints was the most helpful for me in trying to get oriented.

Will add the "Domain" step and draft of additional endpoints.

codyebberson: Your call.  My 2 cents is to remove the endpoints, because it adds cognitive overhead, and we do not provide any additional supporting information.

Issue #1323, opened 2022-12-11
ONC G10 Dry Run - documentation and data preparation


reshmakh:
[Test script - internal only](https://docs.google.com/document/d/1f7yPjeeQ7CMrVZOggzwsXezyr4gS22GsU4kemGs0zDo/edit?usp=sharing)

Need to document the following:
1. Documentation on how to set up [Inferno](https://github.com/inferno-framework/inferno-reference-server) and run it off of app.medplum.com
2. Confirm that all test tool scenarios have been pre-tested
3. Document which version of testing will be accomplished in session
US Core Version:☐  US Core 3.1.1/USCDI v1☐  US Core 4.0.0/USCDI v1☐  US Core 5.0.1/USCDI v2SMART App Launch Version:☐  SMART App Launch 1.0.0☐  SMART App Launch 2.0.0Bulk Data Version:☐  Bulk Data 1.0.1☐  Bulk Data 2.0.0
4. Document the publicly accessible URL for access
5. Provide link to API documentation
6. Link to Terms of Use 
7. Machine readable base URL for listing. The Service Base URL must be a machine readable FHIR resources endpoint that will be used for the ONC Lantern Project.

Related: #1291, #1140

codyebberson:
## Setup a Medplum project for testing

1. Setup your `Practitioner` resources with address, email, phone number, and profile picture
2. Invite a `Patient` user with address, email, phone number, and profile picture (you will need to sign in as this Patient)
3. Setup 2 `ClientApplication` resources with "Redirect URI" set to "https://inferno.healthit.gov/suites/custom/smart/redirect"

## Setup [ONC Inferno](https://inferno.healthit.gov/suites/g10_certification) project

1. Use "US Core 5.0.1 / USCDI v2"
2. Use "SMART App Launch 2.0.0"
3. Use "Bulk Data 2.0.0"

## Running tests

### 1. Standalone Patient App

1. Select "1. Standalone Patient App"
2. Click "Run Tests"
  a. FHIR Endpoint:  https://api.medplum.com/fhir/R4
  b. Standalone Client ID:  Client ID created above
  c. Standalone Client Secret: Client secret created above
  d. Standalone Scope: Use defaults, but you may need to add "patient/ServiceRequest.rs"
  e. Click "Submit"
3. Click on "Follow this link to authorize..."
4. Sign in as the Patient you created above
  a. Click "Remember me"
  b. Click "Sign in"
  c. Accept all scopes and click "Set scope"

[...]

PR #1428, opened 2023-01-21
Fixes from checklist review related to #1323

reshmakh:


codyebberson (approved):

packages/docs/docs/auth/smart-scopes.md
```diff
@@ -0,0 +1,31 @@
+# SMART Scopes for Access Control
+
+SMART on FHIR’s authorization scheme uses OAuth scopes to communicate (and negotiate) access requirements. There is relevant [background information on HL7.org](http://hl7.org/fhir/smart-app-launch/STU2/scopes-and-launch-context.html).
```
rahul1, line 3: I think backing up a few steps and giving some more framing would help make this document more accessible. To me, the biggest thing that's missing is information about "when do I need to use this?". Some questions that would be helpful to answer:

- What is the overall purpose of SMART on FHIR scopes? Might be able to copy some verbiage from [here](http://hl7.org/fhir/smart-app-launch/STU2/app-launch.html)
- What kind of apps will _need_ to use this? Some examples: 
  -  Any 1st party app that Medplum customer builds?
  - An app that launches from inside the Medplum App?
  - An app that I (medplum customer) am building to launch from a 3rd party EPIC instance? 
  - Allowing 3rd party developers to build apps on top of my (medplum customer's) data?

packages/docs/docs/auth/smart-scopes.md
```diff
@@ -0,0 +1,31 @@
...
+
+Medplum supports the `SMART App Launch 2.0.0` [standard](https://hl7.org/fhir/smart-app-launch/).
+
+Historically, apps developed with SMART Scopes were designed to be links, embedded in a traditional EHR. Once those links were clicked on an app would launch and authenticate the user, and request a certain FHIR resources, or grant rights like `offline_access`. Commonly, EHRs support "launching" from either a patient history screen or an encounter documentation screen. So when you see "launch" language in the documentation, hopefully that gives some context on what that means.
```
rahul1, line 7: I think this paragraph might be too confusing. 
What I found confusing was that this paragraph starts off saying "Historically, SMART scopes were designed for X", which implies that this is _not_ what SMART scopes are used for nowadays. But it doesn't seem to clearly address what use cases SMART scopes _are_ used for nowadays

After addressing my comment above, this probably becomes unnecessary.

packages/docs/docs/auth/smart-scopes.md
rahul1, line 3: Maybe link the words "Oauth scopes" to some basic supporting info about what Oauth scopes are? I found the Auth0 docs pretty well written: https://auth0.com/docs/get-started/apis/scopes

packages/docs/docs/auth/smart-scopes.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Supported Scopes
+
+Medplum supports the following scopes.
+
+- `patient/\*.rs`
```
codyebberson, line 13: I'd remove the backslash before the asterisk here.  It should just be `patient/*.rs`

packages/docs/docs/auth/smart-scopes.md
codyebberson: Yeah... This page is the best explanation: https://hl7.org/fhir/smart-app-launch/scopes-and-launch-context.html

We should either reference that page directly, or mimic the flow.

PR #1435, opened 2023-01-25
FHIR Router

codyebberson:
Implements https://github.com/medplum/medplum/discussions/1218


packages/core/src/search/parse.ts
```diff
@@ -142,11 +129,12 @@ class SearchParser implements SearchRequest {
         break;
 
       default: {
-        const param = getSearchParameter(this.resourceType, code);
-        if (!param) {
-          throw badRequest('Unknown search parameter: ' + code);
+        const param = globalSchema.types[this.resourceType]?.searchParams?.[code];
```
reshmakh, line 132: Interesting.  No op needed but I see how this works now.

packages/fhir-router/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Provides a FHIR URL router. Standard CRUD operations map to a standard `FhirRepository` interface. FHIR operations registered separately.
+
+## About Medplum
+
+Medplum is a healthcare platform that helps you quickly develop high-quality compliant applications. Medplum includes a FHIR server, React component library, and developer app.
+
+## License
+
+Apache 2.0. Copyright &copy; Medplum 2022
```
reshmakh, line 11: 2023

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -0,0 +1,134 @@
...
+    this.router.add('GET', ':resourceType', search);
+    this.router.add('POST', ':resourceType/_search', searchByPost);
+    this.router.add('POST', ':resourceType', createResource);
+    this.router.add('GET', ':resourceType/:id', readResourceById);
+    this.router.add('GET', ':resourceType/:id/_history', readHistory);
+    this.router.add('GET', ':resourceType/:id/_history/:vid', readVersion);
+    this.router.add('PUT', ':resourceType/:id', updateResource);
+    this.router.add('DELETE', ':resourceType/:id', deleteResource);
+    this.router.add('PATCH', ':resourceType/:id', patchResource);
+    this.router.add('POST', '$graphql', graphqlHandler);
```
reshmakh, line 122: Is bulk affected?

packages/fhir-router/README.md
codyebberson: good catch

packages/fhir-router/src/fhirrouter.ts
codyebberson: Yes, very much so.  See line 112.

packages/core/src/search/parse.ts
codyebberson: This change is worth a discussion.

Before, this URL parsing code was in the server project, where you could assume that the full FHIR schema was loaded, including all FHIR search parameters.  That meant you could validate search parameters as you were parsing the URL.

After, this URL parsing code is in core, where it might run in the server, or it might run in the browser.  So we're more tolerant of missing search parameters.

At the time that the search is actually executed, we we still validate search parameters, and will throw on unrecognized search param.

reshmakh (approved):

PR #1447, opened 2023-01-27
Enable custom AWS lambda function names

codyebberson:
Note that we will not enable this on Medplum's hosted offering.  This should only ever be enabled in self-hosted environments.

### DIFF
--- a/packages/server/src/fhir/operations/execute.ts
+++ b/packages/server/src/fhir/operations/execute.ts
@@ -1,5 +1,5 @@
 import { InvokeCommand, LambdaClient } from '@aws-sdk/client-lambda';
-import { badRequest, createReference, Hl7Message, resolveId } from '@medplum/core';
+import { badRequest, createReference, getIdentifier, Hl7Message, resolveId } from '@medplum/core';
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/operations/execute.ts
```diff
@@ -172,6 +172,25 @@ async function runInLambda(request: BotExecutionRequest): Promise<BotExecutionRe
...
+/**
+ * Returns the AWS Lambda function name for the given bot.
+ * By default, the function name is based on the bot ID.
+ * If the bot has a custom function, and the server allows it, then that is used instead.
+ * @param bot The Bot resource.
+ * @returns The AWS Lambda function name.
+ */
+export function getLambdaFunctionName(bot: Bot): string {
+  if (getConfig().botCustomFunctionsEnabled) {
+    const customFunction = bot.identifier?.find((id) => id.system === 'https://medplum.com/custom-function');
```
rahul1, line 184: I think you can just use our `getIdentifier` helper here: 
https://www.medplum.com/docs/sdk/modules#getidentifier

rahul1 (approved):

rahul1:
One minor change request, but LGTM

Issue #1417, opened 2023-01-19
Auth0 support in MedplumClient


rahul1:
We have customers who would like to use Auth0 as their primary identity provider for a Medplum project. We'll eventually need to support this in the `SignInForm` react component as well, but the first step is to get this working in the client, since the customer would like to embed this into their own signup UI

rahul1:
First step would be to get the redirects all working with the Auth0 login page

PR #1452, opened 2023-01-29
Updated core base-schema.json

codyebberson:
The original work item was to extract the "IdentityProvider" backbone element from "DomainConfiguration" into a separate element definition, so we can use it elsewhere.

However, it turns out the script to generate "base-schema.json" was never committed, so we needed to do that from scratch.

While doing that, I added a long overdue mini optimization of omitting "min" and "max" when they are default values of `0` and `"1"` respectively.


rahul1 (approved):

reshmakh:
This is related to #1417

PR #1461, opened 2023-02-01
ClientApplication external auth

codyebberson:
Fixes https://github.com/medplum/medplum/issues/1417

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -352,6 +352,27 @@ describe('Client', () => {
     expect(window.location.assign).toBeCalled();
   });
 
+  test('Sign in with external auth', async () => {
+    global.window = Object.create(window);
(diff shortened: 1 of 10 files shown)


rahul1 (approved):

reshmakh:
This looks very sensible and cool.  The only question I have is that should we check whether this is coming from a registered domain?

reshmakh:
Perhaps require that it be launched from registered domain or launch URI (or localhost)?

codyebberson:
> Perhaps require that it be launched from registered domain or launch URI (or localhost)?

Yeah... there are 3 TODO's in the current commit that are all about validation.  This commit demonstrates end to end working example.  I need to add tests and safeguards.

reshmakh:
This needs to be documented as part of #1467

PR #1473, opened 2023-02-05
Fixes #1332 - Multi-platform Docker images

### LINKED ISSUE #1332: Multi-arch docker image
@opiation @julienblin

codyebberson:
Enables "multi arch" / "multi platform" Docker images

After, new architectures will be available on Docker Hub:

image

This only downside is that this is slower than a single architecture (adds about 5 min to deploy).

Currently building "linux/amd64, linux/arm64, linux/arm/v7".  We could also build "linux/ppc64le" and/or "linux/s390x", as Node supports them, but that seems unnecessary and speculative.  Easy to change later too.

See:

- https://docs.docker.com/build/building/multi-platform/
- https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

Other drive-by fixes:
- Removed ECR deploys, which we have not used for over 2 months
- Excluded ".cjs.map" and ".mjs.map" sourcemap files from Docker images (we already excluded ".js.map")
- Use single command for Docker "build and push"
- Additional Docker tag for "version" (in addition to "latest" and git sha)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,7 @@ jobs:
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
             ${{ runner.os }}-
+      - uses: docker/setup-buildx-action@v2
       - name: Build Project
         run: ./scripts/build.sh
         env:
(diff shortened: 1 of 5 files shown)


reshmakh:
If I understand - we are moving production to pull directly from DockerHub?

codyebberson:
> If I understand - we are moving production to pull directly from DockerHub?

We made that change in Aug/Sept 2022: https://github.com/medplum/medplum/pull/916

We have continued to push Docker images to our private ECR for backup.  Everything has been stable for a while, so cutting the backup.  Easy to bring back if we need it.

reshmakh (approved):

Issue #1463, opened 2023-02-02
Disable scroll wheel on quantity input


rahul1:

rahul1:
Fixed by #1477

PR #1477, opened 2023-02-06
Disable scroll wheel on quantity input

rahul1:
One of our lab customers gave us feedback that using the scroll wheel over a quantity input sometimes causes them to input incorrect values. This PR adds a prop called `disableWheel` to disable scroll wheel action on the numeric part of the QuantityInput, and enables the prop on the `QuestionnaireForm` component

### DIFF
--- a/packages/react/src/QuantityInput/QuantityInput.test.tsx
+++ b/packages/react/src/QuantityInput/QuantityInput.test.tsx
@@ -64,11 +64,11 @@ describe('QuantityInput', () => {
     expect(lastValue).toMatchObject({ comparator: '<', value: 123, unit: 'mg' });
   });
 
-  test('Set value with scroll', async () => {
+  test('Set value with wheel', async () => {
     render(<QuantityInput name="a" defaultValue={{ value: 2.3, unit: 'ng' }} />);
 
     const valueInput = screen.getByPlaceholderText('Value');
-    fireEvent.scroll(valueInput, {
+    fireEvent.wheel(valueInput, {
       deltaY: 100,
     });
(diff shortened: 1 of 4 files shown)


rahul1:
fixes #1463

packages/react/src/QuantityInput/QuantityInput.tsx
```diff
@@ -32,20 +33,24 @@ export function QuantityInput(props: QuantityInputProps): JSX.Element {
           })
         }
       />
-      <TextInput
+      <NumberInput
         id={props.name}
         name={props.name}
         data-testid={props.name + '-value'}
         type="number"
-        step="any"
```
codyebberson, line 40: Removing `step="any"` makes it integer only.  Not sure what the correct Mantine convention is here.

packages/react/src/QuantityInput/QuantityInput.test.tsx
```diff
@@ -63,4 +63,18 @@ describe('QuantityInput', () => {
     expect(lastValue).toBeDefined();
     expect(lastValue).toMatchObject({ comparator: '<', value: 123, unit: 'mg' });
   });
+
+  test('Set value with scroll', async () => {
+    render(<QuantityInput name="a" defaultValue={{ value: 2.3, unit: 'ng' }} />);
+
+    const valueInput = screen.getByPlaceholderText('Value');
+    fireEvent.scroll(valueInput, {
```
codyebberson, line 71: This isn't firing the "wheel" event.  Maybe `fireEvent.wheel`?

codyebberson (approved):

PR #1482, opened 2023-02-07
Fixed stale state bug in useResource

rahul1:
**The Problem**
The `resource` state variable is initially set based on the `value` prop sent to the `useResource` hook. However, this is only set on the _first_ render of the component calling `useResource`. If the component renders again with a different value for `value`, the `resource` state will not be updated. 

This was manifesting in a number of places through the Medplum App, where clicking on links would update the url in the address bar, but would not force a re-render of components.

**Solution**
An additional call to useEffect to reset the state in response to the prop. 
ref: https://stackoverflow.com/questions/54865764/react-usestate-does-not-reload-state-from-props

### DIFF
--- a/packages/react/src/useResource/useResource.ts
+++ b/packages/react/src/useResource/useResource.ts
@@ -24,6 +24,10 @@ export function useResource<T extends Resource>(value: Reference<T> | T | undefi
   const [resource, setResource] = useState<T | undefined>(getInitialResource(medplum, value));
 
   useEffect(() => {
+    setResource(getInitialResource(medplum, value));
+  }, [medplum, value]);
+
+  useEffect(() => {
     let subscribed = true;
 
     if (!resource && value && 'reference' in value && value.reference) {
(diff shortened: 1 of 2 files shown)


packages/react/src/useResource/useResource.ts
```diff
@@ -23,6 +23,10 @@ export function useResource<T extends Resource>(value: Reference<T> | T | undefi
   const medplum = useMedplum();
   const [resource, setResource] = useState<T | undefined>(getInitialResource(medplum, value));
 
+  useEffect(() => {
```
codyebberson, line 26: IIRC, the original implementation used this exact pattern, which resulted in unnecessary rerenders.  Removing `useEffect` may have been too far.

I agree that stale data is a worse problem than unnecessary renders, so I'm ok with this.

I wonder if there's a check we could add inside the effect to get the best of both worlds.

For example:

```ts
const initialResource = getInitialResource(medplum, value);
if (initialResource?.meta?.version !== resource?.meta?.version) {
  setResource(initialResource);
}

rahul1: I think that might break in cases where the version might the same, but the full object is different. 

One case is modifying a resource on the client-side in editing workflows. I just added a test for this:

Another case would be when first `value` is a partial resource from GraphQL, and then later is a full resource from `readResource()`.

codyebberson (approved):

PR #1493, opened 2023-02-09
Removed fast-json-patch dependency from @medplum/mock

rahul1:
Replaced with this library: https://github.com/chbrown/rfc6902

The nice thing about this package is that it's small, easy to understand, and MIT license. If perf becomes an issue, we can probably fork and optimize.

I was able to test in a customer repo, and it was compatible with vitetest !


codyebberson (approved):
Nice find!  I like it.

Future task, we should consider using `rfc6902` in the server as well.

PR #1495, opened 2023-02-09
Display notes on DiagnosticReportDisplay

rahul1:


packages/react/src/NoteDisplay/NoteDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { classes } = useStyles();
+  if (!value) {
+    return null;
+  }
+
+  if (!Array.isArray(value)) {
+    value = [value];
+  }
+  return (
+    <Stack justify={'flex-start'} spacing="xs">
```
codyebberson, line 25: Minor nit: for constant strings, you can use normal HTML attribute syntax

```tsx
<Stack justify="flex-start" spacing="xs">
```

packages/react/src/NoteDisplay/NoteDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+import React from 'react';
+
+const useStyles = createStyles((theme) => ({
+  noteBody: { fontSize: theme.fontSizes.sm },
+  noteCite: { fontSize: theme.fontSizes.xs, marginBlockStart: 3 },
+  noteRoot: { padding: 5 },
+}));
+
+export interface NoteDisplayProps {
+  value?: Annotation | Annotation[];
```
codyebberson, line 12: I believe that every use of `<NoteDisplay>` uses `Annotation[]` array, never single `Annotation`.  Any objections to simplifying and just taking `Annotation[]`?

packages/react/src/NoteDisplay/NoteDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function NoteDisplay({ value }: NoteDisplayProps): JSX.Element | null {
+  const { classes } = useStyles();
+  if (!value) {
+    return null;
+  }
+
+  if (!Array.isArray(value)) {
```
codyebberson, line 21: Then you could remove this check.

codyebberson (changes requested):
Overall looks good.

Could you look at the comment on `<NoteDisplay>` about changing `value` to only use `Annotation[]` array?  Everything else optional.

packages/react/src/NoteDisplay/NoteDisplay.tsx
rahul1: I actually originally had it like this, but then was worried that the name `Note` (singular) would imply that non-arrays would be ok. 

But `NotesDisplay` seemed like a weird name. 

I'll change it to avoid the perf hit of the array check

codyebberson (approved):

PR #1497, opened 2023-02-09
Fixes #1491 - Next.js and SSR issues

### LINKED ISSUE #1491: MedplumProvider errors in NextJS
Create a new nextjs app
Copied MedplumProvider from @medplum/react package
```
@medplum/react  package didn’t work with nextjs.
Had to install @mantine/core @tabler/icons @mantine/notifications manually to make build and got error from tabler/icons package
error - file:///Users/mimu/dev/examples/medplum-test/node_modules/@medplum/react/dist/esm/AttachmentArrayInput/AttachmentArrayInput.mjs:2
import { IconCircleMinus, IconCloudUpload } from '@tabler/icons';
         ^^^^^^^^^^^^^^^
SyntaxError: Named export 'IconCircleMinus' not found. The requested module '@tabler/icons' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@tabler/icons';
const { IconCircleMinus, IconCloudUpload } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:437:15) {
  page: '/'
}

```

Wrap App with MedplumProvider
Got this error
```
error - ReferenceError: window is not defined
    at new MedplumClient (file:///Users/mimu/dev/examples/medplum-test/node_modules/@medplum/core/dist/esm/client.mjs:104:78)
    at eval (webpack-internal:///./pages/_app.tsx:18:17)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  page: '/'
}

```

codyebberson:
Standalone Next.js demo app that will be published separate.

Issues:
* `@tabler/icons`
  * It appears CJS/ESM exports were incompatible with Next.js bundler without bundling all icons
  * See: https://github.com/tabler/tabler-icons/issues/471
  * So instead, we're baking in the ~30 icons that we use
  * This does increase the bundle size of `@medplum/react`, but only if using those components
  * Otherwise tree shaking should make it a no-op
  * `@tabler/icons` is no longer required nor a peer dependency
* `react-router-dom`
  * Before, we assumed that `react-router-dom` was the routing engine, and we used `useNavigate()` to follow links
  * Now, you can pass in a custom `navigate()` method to `<MedplumProvider>`
  * By default, it uses `window.location.href = x;`, which is simple, but that won't take advantage of whatever routing 
  * `react-router-dom` is no longer required nor a peer dependencyframework you are using
  * Theoretically, this should also work with other frameworks like Remix Run and Tanstack Router
* `window` access
  * Basically all access to `window` needs to be inside a `useEffect()` or handled separately
  * For example, determining the current URL with `window.location.host`
  * Or base-64 encoding with `window.btoa`

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -782,11 +782,6 @@ export class MedplumClient extends EventTarget {
    * @param loginParams Optional login parameters.
    */
   async signInWithRedirect(loginParams?: Partial<BaseLoginRequest>): Promise<ProfileResource | void> {
-    const window = getWindow();
-    if (!window) {
-      // This method is only available in the browser.
-      return;
-    }
     const urlParams = new URLSearchParams(window.location.search);
     const code = urlParams.get('code');
     if (!code) {
@@ -803,11 +798,6 @@ export class MedplumClient extends EventTarget {
    * @category Authentication
    */
   signOutWithRedirect(): void {
-    const window = getWindow();
-    if (!window) {
-      // This method is only available in the browser.
-      return;
-    }
     window.location.assign(this.#logoutUrl);
   }
 
@@ -824,11 +814,6 @@ export class MedplumClient extends EventTarget {
     redirectUri: string,
     baseLogin: BaseLoginRequest
   ): Promise<void> {
-    const window = getWindow();
-    if (!window) {
-      // This method is only available in the browser.
-      return;
-    }
     const loginRequest = await this.ensureCodeChallenge(baseLogin);
     const url = new URL(authorizeUrl);
     url.searchParams.set('response_type', 'code');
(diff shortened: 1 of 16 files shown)


packages/core/src/client.ts
```diff
@@ -2227,12 +2227,36 @@ export class MedplumClient extends EventTarget {
...
+  return window.fetch.bind(window);
+}
+
 /**
  * Returns the base URL for the current page.
  * @category HTTP
  */
 function getBaseUrl(): string {
-  return window.location.protocol + '//' + window.location.host + '/';
+  const window = getWindow();
```
rahul1, line 2258: I know this isn't actually the point of this PR, but why doesn't this return `this.#baseURL`. The naming is confusing about whether the "BaseURL" is the baseURL for the FHIR server, or the baseURL of the current app

codyebberson: Yeah... that's a fair critique.... this is ancient code.  It's particularly bad because there are two `getBaseUrl` methods in the same file that do very different things.  I'll rename.

rahul1 (changes requested):
Excited for this one!

rahul1 (approved):

PR #1499, opened 2023-02-09
Display Observation notes by default in the ResourceTimeline component

rahul1:


codyebberson (commented):
@rahul1  Should we just make it default enabled for `<DiagnosticReportDisplay>` ?

codyebberson (approved):

PR #1512, opened 2023-02-12
Resolve redis caching and access policies

codyebberson:
Context:
1. We use Redis caching for "read by ID" operations when available
2. AccessPolicy resources define filters/constraints on what a user can read and write
3. Before, the AccessPolicy read filters were only implemented as SQL clauses
4. That meant that we didn't have a clean way to enforce AccessPolicy constraints when reading from Redis
5. So the solution was simply "don't use Redis if the user has an AccessPolicy"
6. That left a lot of low hanging fruit for cache optimization

This PR fixes all of that.  We actually did have most of the machinery to say "does this resource satisfy the current user's access policy".  Just a matter of tying it all together.


reshmakh (approved):
Tightly scoped, like it.

reshmakh (approved):

PR #1518, opened 2023-02-13
Resource page perf improvements

codyebberson:
Fixes a variety of minor performance issues in `<ResourcePage>` in the app project.

The biggest one is the classic React waterfall problem.  Before, `<ResourcePage>` requested the primary resource at the top level, and then passed it as a prop to the active tab.  And then other requests could get piled on with other nested components.  

image

After, the pattern is changed.  `resource` is not passed as a prop.  Each tab component now uses `{resourceType}` and `{id}` from the URL params, which are available immediately.

The great thing about this is that `MedplumClient` "auto batching" combines all of these requests -- from multiple different React components -- and gracefully combines them into a single FHIR batch request.

image

The next big change is cleaning up some performance debt in `useResource()`.  There were a few problems, notably that the return value of `useResource()` was managed by React `useState()`.  Now the value comes from `medplum.getCachedReference` which will always return the most recent value.  It still uses `useState()` to trigger re-renders on changes.  It still uses `useEffect()` to initiate network requests to fetch the value.

Last but not least, this PR includes a much needed refactoring overhaul of the `<ResourcePage>` tab routing.  Before, it used a mishmash of `if` and `switch` statements to determine the tab contents.  Now it uses `react-router-dom` [nested routes](https://reactrouter.com/en/main/start/overview#nested-routes) for a more maintainable system.

### DIFF
--- a/packages/app/src/AppRoutes.tsx
+++ b/packages/app/src/AppRoutes.tsx
@@ -25,8 +25,24 @@ import { MfaPage } from './MfaPage';
 import { OAuthPage } from './OAuthPage';
 import { RegisterPage } from './RegisterPage';
 import { ResetPasswordPage } from './ResetPasswordPage';
+import { ApplyPage } from './resource/ApplyPage';
+import { AppsPage } from './resource/AppsPage';
+import { BlamePage } from './resource/BlamePage';
+import { BotEditor } from './resource/BotEditor';
+import { BuilderPage } from './resource/BuilderPage';
+import { ChecklistPage } from './resource/ChecklistPage';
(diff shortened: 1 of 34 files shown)


reshmakh (approved):

reshmakh (approved):

rahul1:
> After, the pattern is changed. resource is not passed as a prop. Each tab component now uses {resourceType} and {id} from the URL params, which are available immediately.

Looks good. ^ Would this performance optimization still work with autobatching turned off? My understanding is that on the first page load, all the tabs would still make their own requests, but after the first one returns, it would be read from the cache. Is that right?

PR #1520, opened 2023-02-14
Parameterized access policies

codyebberson:
Resolves https://github.com/medplum/medplum/discussions/1453

### DIFF
--- a/packages/definitions/dist/fhir/r4/profiles-medplum.json
+++ b/packages/definitions/dist/fhir/r4/profiles-medplum.json
@@ -749,7 +749,7 @@
           {
             "id" : "ProjectMembership.accessPolicy",
             "path" : "ProjectMembership.accessPolicy",
-            "definition" : "The access policy for the user within the project memebership. If the \"access\" field is specified, then the \"accessPolicy\" field is ignored.",
+            "definition" : "The access policy for the user within the project memebership.",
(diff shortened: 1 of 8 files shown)


reshmakh (approved):

packages/definitions/dist/fhir/r4/profiles-medplum.json
```diff
@@ -749,14 +749,72 @@
           {
             "id" : "ProjectMembership.accessPolicy",
             "path" : "ProjectMembership.accessPolicy",
-            "definition" : "The access policy for the user within the project memebership.",
+            "definition" : "The access policy for the user within the project memebership. If the \"access\" field is specified, then the \"accessPolicy\" field is ignored.",
```
rahul1, line 752: Should we mark this as "deprecated", to nudge users towards the new system?

rahul1 (approved):

packages/definitions/dist/fhir/r4/profiles-medplum.json
codyebberson: @rahul1 

Good question.  I don't know...  It might still be valuable as a simple default?

It has never been exposed via API, so it would not be a breaking change to remove it entirely.

I don't have strong opinions on that right now.

PR #1521, opened 2023-02-14
Use @aws-sdk/cloudfront-sign

### LINKED ISSUE #1414: Use @aws-sdk/cloudfront-sign
For a long time, the AWS JS SDK v3 did not include the ability to generate a presigned URL (see: https://github.com/aws/aws-sdk-js-v3/issues/1822)

So Medplum included a small custom implementation: https://github.com/medplum/medplum/blob/main/packages/server/src/fhir/signer.ts

There is now an official library: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_cloudfront_signer.html

We should replace the custom implementation with the library.

aurambaj:
Replaces Medplum implementation with AWS SDK, now it's been made available (https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_cloudfront_signer.html)

There is a small difference in the character escaping. Medplum used to escape `~` but AWS SDK does not.

Fix #1414


codyebberson (approved):

codyebberson:
Thanks @aurambaj !

PR #1523, opened 2023-02-14
Docs on Unit Testing Bots

rahul1:


packages/examples/src/bots/finalize-report.ts
```diff
@@ -0,0 +1,25 @@
+import { BotEvent, MedplumClient } from '@medplum/core';
+import { DiagnosticReport } from '@medplum/fhirtypes';
+
+export async function handler(medplum: MedplumClient, event: BotEvent): Promise<any> {
+  // Cast the Bot input as a Diagnostic report
+  const report = event.input as DiagnosticReport;
+  if (report.resourceType !== 'DiagnosticReport') {
```
codyebberson, line 7: For the purposes of examples, I'd probably cut this?

packages/examples/src/bots/finalize-report.ts
```diff
@@ -0,0 +1,25 @@
...
+    report.status = 'final';
+    await medplum.updateResource(report);
+  }
+
+  if (report.result) {
+    for (const observationRef of report.result) {
+      const observation = await medplum.readReference(observationRef);
+      if (observation.status !== 'final') {
+        observation.status = 'final';
+        await medplum.updateResource(observation);
```
codyebberson, line 21: Low pri:  This could be refactored to use auto batching.  The trick would be to not use `await`, and store all of the promises in an array.  Then use `Promise.allSettled` after the loop.

rahul1: Yeah that makes sense, but I think it introduces more cognitive load to someone just learning. Probably worth separate examples showing how to exploit auto-batching

packages/examples/package.json
```diff
@@ -16,6 +16,8 @@
   },
   "devDependencies": {
     "@medplum/fhirtypes": "*",
-    "@medplum/core": "*"
+    "@medplum/core": "*",
+    "@medplum/mock": "*",
+    "vitest": "0.22.0"
```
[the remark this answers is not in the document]
rahul1: no objection in principle, but except the corresponding test in medplum-demo-bots is written using vitest

packages/examples/src/bots/finalize-report.ts
codyebberson: 👍

codyebberson (approved):

PR #1526, opened 2023-02-15
Add db index on AuditEvent.entity

codyebberson:


reshmakh (approved):
This is a heavyweight one right?

Issue #1530, opened 2023-02-15
Create Users without emails


rahul1:
Some of our customers have users that authenticate primarily via phone or other mechanism, and don't store the user's email in their authorization service. This situation occurs when using external auth providers.  We should find a way to allow the creation of a `User` object that doesn't explicitly rely on email as the unique key.

codyebberson:
Fixed in #1535 + #1541 + #1551

PR #1532, opened 2023-02-15
Add option to use 'sub' from external auth

codyebberson:


rahul1:
Fixes #1530

packages/server/src/auth/external.ts
```diff
@@ -49,7 +49,7 @@ export const externalCallbackHandler = async (req: Request, res: Response): Prom
   }
 
   const userInfo = await verifyCode(idp, code);
-  const email = userInfo.email as string;
+  const email = (idp.useSubject ? userInfo.sub : userInfo.email) as string;
```
rahul1, line 52: nit: rename this variable since it won't always be an email

codyebberson:
Replaced with https://github.com/medplum/medplum/pull/1535

PR #1548, opened 2023-02-17
update default docker image in init script

cstoltze:
Thanks for your work on medplum! The experience of setting up a self-hosted server on AWS was even smoother than I hoped.

I did spend a bit of time trying to figure out why fargate was failing to start. It turns out that the configuration I was using specified a docker image that doesn't exist on docker hub. Modifying the default value in the init script would have saved me some time and confusion.

This PR modifies the image name to match the one I'm using.


codyebberson (approved):

codyebberson:
@cstoltze Eek!  Good catch!  Thank you!

codyebberson (approved):

PR #1583, opened 2023-02-23
Bugfix: useResource makes incorrect fetch when called with resource with id

rahul1:
Before: 
calling `useResource({resourceType: 'ServiceRequest', id: '123'})` would always fetch from the server, even if the resource had been modified on the client.

After: 
If calling with a resource that has been modified client-side, re-render if `deepEquals` fails

### DIFF
--- a/packages/react/src/useResource/useResource.test.tsx
+++ b/packages/react/src/useResource/useResource.test.tsx
@@ -150,7 +150,7 @@ describe('useResource', () => {
 
   test('Responds to value edit after cache', async () => {
     function TestComponentWrapper(): JSX.Element {
-      // Call useResource with a reference to fille the cache
+      // Call useResource with a reference to fill the cache
       useResource({ reference: 'ServiceRequest/123' });
       const [resource, setResource] = useState<ServiceRequest>({
         resourceType: 'ServiceRequest',
--- a/packages/react/src/useResource/useResource.test.tsx
+++ b/packages/react/src/useResource/useResource.test.tsx
@@ -119,6 +119,7 @@ describe('useResource', () => {
       const [resource, setResource] = useState<ServiceRequest>({
         resourceType: 'ServiceRequest',
         status: 'draft',
+        id: '123',
       });
       return (
         <>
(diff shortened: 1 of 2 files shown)


codyebberson (changes requested):
Let's discuss offline.  I think there's a philosophical disconnect in the purpose of `useResource()`.  The behavior here doesn't align with my expectations.  My expectations might be wrong, but we should figure that out first.

codyebberson:
Summarizing my thoughts:

>  would always fetch from the server, even if the resource had been modified on the client

At this point, your component is managing its own state, and should use `React.useState`, not depend on `useResource`.

We could review the developer experience of `useResource` + `useState` -- for example, in cases where you want `useState` to start the the most fresh data, but you want to be able to manipulate the data.

Consider the setup code in `<ResourceForm>` (https://github.com/medplum/medplum/blob/main/packages/react/src/ResourceForm/ResourceForm.tsx#L17-L28):

```tsx
  const medplum = useMedplum();
  const defaultValue = useResource(props.defaultValue);
  const [schema, setSchema] = useState<IndexedStructureDefinition | undefined>();
  const [value, setValue] = useState<Resource | undefined>();

  useEffect(() => {
    if (defaultValue) {
      setValue(JSON.parse(JSON.stringify(defaultValue)));
      medplum.requestSchema(defaultValue.resourceType).then(setSchema).catch(console.log);
    }
  }, [medplum, defaultValue]);
```

rahul1:
The problem arises when you have a parent component that calls a child that uses `useResource`. Even if the parent is managing it's own state with `useState` and passes the modified resource to a child component, it will ignore what was passed in and read from cache instead. 

This breaks the parent/child abstraction. The implementer of the parent needs to:
1. Know that the child component uses `useResource`
2. Know that `useResource` will read from cache if the `Resource.id` property is set
3. Delete the `Resource.id` property from the resource in memory, before passing it as a prop to child. 

That requires too much knowledge about the internals of both the child and `useResource`

codyebberson:
> Even if the parent is managing it's own state with `useState` and passes the modified resource to a child component, it will ignore what was passed in and read from cache instead.

`useResource` and the `MedplumClient` cache are for "live" data.  In this example, if the parent is managing state and wants to share state with the child, it should pass the modified resource as a prop.

`useResource` is for "I want the most recent version of this resource".

rahul1:
Closing in favor of #1622

PR #1584, opened 2023-02-23
Typeguards for `isResource` and `isReference`

rahul1:
I saw a common pattern in our code where we check if `resourceType` or `reference` is in an object, and then casts it. 

This PR takes advantage of Typescript's user-defined type guards to centralize that check in a utility function and eliminate the cast

### DIFF
--- a/packages/core/src/fhirpath/functions.ts
+++ b/packages/core/src/fhirpath/functions.ts
@@ -1543,7 +1543,7 @@ export const functions: Record<string, FhirPathFunction> = {
       if (typeof value === 'number') {
         return { type: PropertyType.BackboneElement, value: { namespace: 'System', name: 'Integer' } };
       }
-      if (value && isResource(value)) {
+      if (isResource(value)) {
         return {
           type: PropertyType.BackboneElement,
           value: { namespace: 'FHIR', name: value.resourceType },
--- a/packages/core/src/fhirpath/functions.ts
+++ b/packages/core/src/fhirpath/functions.ts
@@ -1,6 +1,6 @@
-import { Reference, Resource } from '@medplum/fhirtypes';
+import { Reference } from '@medplum/fhirtypes';
 import { Atom } from '../fhirlexer';
-import { PropertyType, TypedValue } from '../types';
+import { isResource, PropertyType, TypedValue } from '../types';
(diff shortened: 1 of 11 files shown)


packages/core/src/fhirpath/functions.ts
```diff
@@ -1543,10 +1543,10 @@ export const functions: Record<string, FhirPathFunction> = {
       if (typeof value === 'number') {
         return { type: PropertyType.BackboneElement, value: { namespace: 'System', name: 'Integer' } };
       }
-      if (value && typeof value === 'object' && 'resourceType' in value) {
+      if (value && isResource(value)) {
```
codyebberson, line 1546: We should be able to remove the `value &&` check

packages/server/src/openapi.ts
```diff
@@ -515,5 +516,5 @@ function buildPatchPath(): any {
 
 function isResourceType(definition: JSONSchema4): boolean {
   const props = definition?.properties;
-  return !!(props && 'resourceType' in props && 'id' in props && 'meta' in props);
+  return !!(props && isResource(props) && 'id' in props && 'meta' in props);
```
codyebberson, line 519: Remove the `props &&` check

codyebberson (commented):
Nice, good pattern matching.  Couple minor suggestions, otherwise this is great

packages/core/src/types.ts
```diff
@@ -423,6 +424,26 @@ export function getElementDefinition(typeName: string, propertyName: string): El
   return property;
 }
 
+/**
+ * Typeguard to validate that an object is a FHIR resource
+ * @param value The object to check
+ * @returns True if the input is of type 'object' and contains property 'resourceType'
+ */
+export function isResource<T extends Resource = Resource>(value: T | Reference<T> | string | undefined): value is T {
```
codyebberson, line 432: We may want to consider adding `null` to the function signature, in which case we would also need to add a null check in the function body because `typeof null === 'object'` 😡

You could go even further and change the function signature to `value: unknown`, which would accept anything.  That would be nice, because then you could throw anything at it, and it will tell you if it's a resource or not.  That might break the generic `<T>` return type though.

codyebberson (approved):

PR #1602, opened 2023-02-28
Project admin resource access

codyebberson:
This enables limited access to `Project` and `ProjectMembership` resources to project admin users.

### DIFF
--- a/packages/server/src/fhir/accesspolicy.ts
+++ b/packages/server/src/fhir/accesspolicy.ts
@@ -145,8 +145,8 @@ function applyProjectAdminAccessPolicy(
   if (accessPolicy) {
     // If there is an existing access policy
     // Remove any references to project admin resource types
-    accessPolicy.resource = accessPolicy.resource?.filter((r) =>
-      projectAdminResourceTypes.includes(r.resourceType as string)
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/accesspolicy.ts
```diff
@@ -113,3 +118,59 @@ async function buildAccessPolicyResources(
...
+  if (!membership.admin && !accessPolicy) {
+    // Not a project admin and no access policy, so return default access.
+    return undefined;
+  }
+
+  if (accessPolicy) {
+    // If there is an existing access policy
+    // Remove any references to project admin resource types
+    accessPolicy.resource = accessPolicy.resource?.filter(
+      (r) => r.resourceType !== 'Project' && r.resourceType !== 'ProjectMembership'
```
rahul1, line 149: It would be safer to export the constant `projectAdminResourceTypes` from repo and and refer to them here, in case we expand the contents of `projectAdminResourceTypes` in the future

rahul1 (approved):

packages/server/src/fhir/accesspolicy.ts
codyebberson: 👍

PR #1623, opened 2023-03-04
Parse _filter search parameter expressions

codyebberson:
Prerequisite for https://github.com/medplum/medplum/issues/1619

This only includes "parsing", it does not include execution.

Limitations:
* Does not support search parameter chaining (i.e., `Observation: subject.name co "pet"`)
* Does not support search parameter indexing (i.e., `Observation: related[type eq "has-component"].target pr true`)


codyebberson:
Replaced by https://github.com/medplum/medplum/pull/1625

PR #1656, opened 2023-03-15
Adding some documentation on invite and reparenting external auth art…

### LINKED ISSUE #1612: Update documentation to include patient invite
Smart app launch links won't work without a User resource, that needs to be reflected in the documentation https://www.medplum.com/docs/integration/smart-app-launch

reshmakh:
…icles.

Fixes #1612


codyebberson (commented):
My preference would be not to rename those files -- This is going to break a bunch of links, and it doesn't not strike me as 10x better than before.

The content changes all look good to me.

codyebberson (approved):

PR #1667, opened 2023-03-16
Updating terms to include use of UMLS

reshmakh:


packages/docs/src/pages/terms.md
```diff
@@ -146,4 +146,31 @@ You also acknowledge and understand that, with respect to any dispute with the R
...
+
+#### UCUM License Notice
```
codyebberson, line 157: Should this be "####" or "###"?

reshmakh: This should be #### - because it is logically grouped under LOINC.

codyebberson: Hm, ok.  We may want to keep it top level?  We use UCUM for other non-LOINC cases as well.  We may want to treat it as it's own top level thing.

codyebberson (approved):

PR #1707, opened 2023-03-23
Enabled eslint eqeqeq rule

codyebberson:


tonylee80 (approved):
Thanks for adding the === linting rule. I'll add that to my Vs code eslinter as well.

rahul1 (approved):

PR #1712, opened 2023-03-24
Fix total count for search by Reference.identifier

rahul1:
When the Medplum App makes a search call, it includes the `_total=accurate` search parameter to get the total count of resources. 

There was a bug in the code block handling the `reference:identifer` that would mutate the `SearchRequest` object. 

```
if (filter.operator === FhirOperator.IDENTIFIER) {
    param = deriveIdentifierSearchParameter(param);
    filter.code = param.code as string,
    filter.operator = FhirOperator.EQUALS,
 }
```

This only becomes a problem when `getTotalCount()` is invoked in the same API call, because changing `filter.code` to `<param>:identifer` causes this to be treated as an invalid search parameter

### DIFF
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -2348,6 +2348,15 @@ describe('FHIR Repo', () => {
     expect(bundle2.entry?.length).toEqual(2);
     expect(bundleContains(bundle2, c1)).toBeTruthy();
     expect(bundleContains(bundle2, c2)).toBeTruthy();
+
+    // Search with count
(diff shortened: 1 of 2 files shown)


codyebberson (commented):
Oh, nice, good find.

I have mixed feelings about the `Readonly<>`.  I agree that it's technically correct.  My hesitation is that it's inconsistent, and it sends a weird counter signal, unless we use it everywhere.

My preference would be to remove the `Readonly<>` wrappers for now, and revisit with a more holistic strategy.

The core fix is great, excellent find 👍

codyebberson (approved):

PR #1716, opened 2023-03-24
Create bot from CLI

jamestouri:
Adding ability to create a bot from the medplum cli. The expected logic will be 

`npx medplum <bot-name> <project-id> <file-source)`
 
The test coverage is a bit tough and my thought:
- Thinking about creating a `helper` folder for the CLI to make the functions exportable that way we can test them directly. This can allows us to test the helper methods in the cli/index.ts file.  However I wonder how much that'll help.

Any other suggestions on testing this further will be fantastic

### DIFF
--- a/packages/cli/src/index.test.ts
+++ b/packages/cli/src/index.test.ts
@@ -269,25 +269,7 @@ describe('CLI', () => {
   });
 
   test('Create bot success', async () => {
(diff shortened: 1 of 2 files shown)


codyebberson (approved):
Nice! 👍

PR #1729, opened 2023-03-28
Added searchResources and searchOne helpers to FhirRepository

codyebberson:
This is a 100% pure refactoring / cleanup PR.

By default, `search()` calls return a FHIR `Bundle`.  `Bundle` is a wrapper type.  The meat of the results is inside `Bundle.entry`.  `Bundle` also includes a bunch of metadata (total count, next/previous links, etc).  However, in the common case, you just want the `entry` values.

In `MedplumClient` we have 2 extremely handy convenience methods: `searchResources` and `searchOne`.  Where the normal `search()` method returns a FHIR `Bundle`, `searchResources` returns a simple `Resource[]` array and `searchOne` returns a single `Resource`.

In the server, we did not have those convenience methods, and so we ended up doing a lot of `bundle.entry?.map(...)` calls.

This PR finally gets around to adding the convenience methods, and using it in a bunch of places.  I don't think I found all of them, so there might be some ongoing opportunities to replace `search() + map()` with `searchResources()`.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -987,7 +987,7 @@ export class MedplumClient extends EventTarget {
    * @param resourceType The FHIR resource type.
    * @param query Optional FHIR search query or structured query object. Can be any valid input to the URLSearchParams() constructor.
    * @param options Optional fetch options.
-   * @returns Promise to the search result bundle.
+   * @returns Promise to the first search result.
    */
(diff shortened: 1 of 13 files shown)


jamestouri (approved):

rahul1 (approved):
Look good!

PR #1732, opened 2023-03-28
check if google site id and project id exist

tonylee80:


packages/server/src/auth/google.test.ts
```diff
@@ -227,6 +227,84 @@ describe('Google Auth', () => {
...
+      ...project,
+      site: [
+        {
+          name: 'Test Site',
+          domain: ['example.com'],
+          googleClientId,
+        },
+      ],
+    });
+    console.log(project);
```
codyebberson, line 266: Remove `console.log`

codyebberson (changes requested):
Nice!  Overall, looks great.  One `console.log` in tests that should be removed, otherwise this is good to go.

Thanks @tonylee80 !

codyebberson (approved):
lgtm 👍  Don't worry about this Sonar duplication warning.  IMHO, that check is too aggressive, and it would be awkward to merge login.ts and google.ts like it wants.

tonylee80:
> lgtm 👍 Don't worry about this Sonar duplication warning. IMHO, that check is too aggressive, and it would be awkward to merge login.ts and google.ts like it wants.

@codyebberson I was able to resolve the duplication warning, could you take a look at the quick refactor to move some of the duplicate code to `packages/server/src/auth/utils.ts`? I'm more than happy to revert the changes if it doesn't make sense to have shared logic between login.ts and google.ts

codyebberson (changes requested):

PR #1794, opened 2023-04-06
Fixes #1793 - error handling in autobatch single request case

### LINKED ISSUE #1793: readResource promise never resolves when autoBatchTime set and resource does not exist
When autoBatchTime is set, the promise returned by readResource never resolves when trying to read a nonexistent resource. If I don't use autoBatchTime, it works fine and the `catch` and `finally` blocks handle the error like normal.

codyebberson:
Similar bug was fixed in https://github.com/medplum/medplum/pull/1507

This bug is in the special case of a single request in the batch.

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -12,8 +12,8 @@ import PdfPrinter from 'pdfmake';
 import type { CustomTableLayout, TDocumentDefinitions, TFontDictionary } from 'pdfmake/interfaces';
 import { TextEncoder } from 'util';
 import { InviteBody, MedplumClient, NewPatientRequest, NewProjectRequest, NewUserRequest } from './client';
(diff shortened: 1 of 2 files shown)


jamestouri (approved):

packages/core/src/client.test.ts
```diff
@@ -57,189 +43,23 @@ const schemaResponse = {
...
-function mockFetch(url: string, options: any): Promise<any> {
-  const { method } = options;
-  const response = mockHandler(method, url, options);
-  response.status = response.status || 200;
-  return Promise.resolve({
-    ok: response.status === 200 || response.status === undefined,
-    status: response.status,
-    blob: () => Promise.resolve(response),
-    json: () => Promise.resolve(response),
+function mockFetch(
```
tonylee80, line 46: Nice!

tonylee80 (approved):

Issue #1752, opened 2023-03-31
Download as urn uuid bundle


codyebberson:
On the Search page, add an option to download all results as a FHIR `Bundle` with all `id` and `reference` values rewritten as `urn:uuid:` syntax.

* Combine multiple search results into one bundle
* Create `urn:uuid:` URLs for all resources
* Replace all references with the `urn:uuid:` references

```js

for (const resource of input) {
  const meta = resource.meta;
  delete meta.versionId;
  delete meta.lastUpdated;

  if (Object.keys(meta).length === 0) {
    delete resource.meta;
  }
}

console.log(
  JSON.stringify(
    {
      resourceType: 'Bundle',
      type: 'transaction',
      entry: input.map((resource) => ({
        fullUrl: 'urn:uuid:' + resource.id,
        request: { method: 'POST', url: resource.resourceType },
        resource,
      })),
    },
    replacer,
    2
  )
);

function replacer(key, value) {
  // Filtering out properties
  if (key === 'reference' && typeof value === "string" && value.includes('/')) {
    // Input: "Patient/{id}"
    // Output: "urn:uuid:{id}"
    return 'urn:uuid:' + value.split('/')[1];
  }
  return value;
}
```

PR #1796, opened 2023-04-07
Download/Export as a Transaction Bundle

jamestouri:
For #1752 

https://user-images.githubusercontent.com/6913745/230534274-793a13a6-3c4c-456e-be88-00fb079d7170.mov

Sample file of the output for the FHIR Bundle download

Screenshot 2023-04-06 at 6 46 11 PM


packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -80,7 +81,8 @@ export interface SearchControlProps {
   onClick?: (e: SearchClickEvent) => void;
   onAuxClick?: (e: SearchClickEvent) => void;
   onNew?: () => void;
-  onExport?: () => void;
+  onExportCSV?: () => void;
+  onExportFHIRBundle?: () => void;
```
codyebberson, line 85: Can we rename this to `onExportTransactionBundle`?

Here's my thinking:
1. We rarely (never?) use "fhir" in variable or function names.  It's often implicit.
2. `Bundle` is well established throughout the code base, so this should be self explanatory.
3. In future work (or maybe you could add it to this PR if it is easy), I think we should also include `Export Bundle` / `onExportBundle` which is the raw FHIR Bundle we receive from the API server.

To elaborate on that:

- "Export Bundle"
  - Exports the search Bundle exactly as is
  - `Bundle.type` = `searchset`
  - Keep original ID's
  - Keep original `meta`
  - Keep original `fullUrl`
- "Export Transaction Bundle"
  - What you implemented in this PR
  - `Bundle.type` = `transaction`
  - Remove original ID`s
  - Remove original meta
  - Add `urn:uuid:` to `fullUrl`

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -92,6 +94,7 @@ interface SearchControlState {
   fieldEditorVisible: boolean;
   filterEditorVisible: boolean;
   filterDialogVisible: boolean;
+  exportEditorVisible: boolean;
```
codyebberson, line 97: Can we rename this to `exportDialog` / `exportDialogVisible` ?

We use "Editor" for complex dialogs when the user is editing something, and "Dialog" for simple dialogs.

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -526,6 +530,17 @@ export function SearchControl(props: SearchControlProps): JSX.Element {
           });
         }}
       />
+      <SearchExportDisplay
```
codyebberson, line 533: Rename to `SearchExportDialog`

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -80,7 +81,8 @@ export interface SearchControlProps {
   onClick?: (e: SearchClickEvent) => void;
   onAuxClick?: (e: SearchClickEvent) => void;
   onNew?: () => void;
-  onExport?: () => void;
+  onExportCSV?: () => void;
```
codyebberson, line 84: Can we rename this to `onExportCsv`?  By convention, we always use `camelCase`, even with abbreviations like `Id`, etc.

packages/react/src/SearchExportDisplay/SearchExportDisplay.test.tsx
```diff
@@ -0,0 +1,56 @@
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+import { ExportButton, SearchExportDisplay } from './SearchExportDisplay';
+
+describe('SearchExportDisplay', () => {
```
codyebberson, line 5: Rename file to `SearchExportDialog.test.tsx`

codyebberson (changes requested):
Hey @jamestouri  - nice, this is a great start.

I left a bunch of nits on code conventions.

I have one bigger design question -- maybe we should discuss offline with the team -- about whether the Export dialog belongs in the `react` package or in the `app` package.  I don't have a good sense of whether this is something other customers would want to bake into their applications or not.

The existing `onExport` prop on `<SearchControl>` was kinda hastily jammed in -- it was not well designed.

By renaming it, technically this is an API breaking change, so we have that to consider.

Let's discuss w/ @reshmakh  and @rahul1  to see if they have thoughts.

jamestouri:
> I have one bigger design question -- maybe we should discuss offline with the team -- about whether the Export dialog belongs in the `react` package or in the `app` package. I don't have a good sense of whether this is something other customers would want to bake into their applications or not.

Thinking about this again, I'm more for moving the Export dialog to the `app` package as a component there, given the customizability for it. If we keep onExport as the only prop in Search control, it's cleaner and potentially a better look.

packages/app/src/utils.ts
```diff
@@ -76,3 +76,75 @@ export function sendCommand(frame: HTMLIFrameElement, command: any): Promise<any
...
   return resolveId(medplum.getActiveLogin()?.project) as string;
 }
+
+/**
+ * This function reads the contents of the a JSON object, manipulates the object
+ * to add UUIDs and other modifications, and then exports the
+ * modified JSON object to a file.
+ * @param blob The Blob object that we'll receive from the search query
+ */
+export function getFHIRBundle(entry: any): void {
```
rahul1, line 86: I would suggest some tweaks to these utility functions, to make them more broadly useful. 
1. This function to should take in a `Bundle` and output a `Bundle`. Have the stringifying / exporting happen in the calling function
2. rename this to `convertToTransactionBundle`, to clarify that the main purpose of this Bundle is to convert one type of Bundle to another
3. Move the bundle conversion functions to `@medplum/core`, that way they can also be used by the CLI and other tools

rahul1 (changes requested):

rahul1:
> I have one bigger design question -- maybe we should discuss offline with the team -- about whether the Export dialog belongs in the react package or in the app package. I don't have a good sense of whether this is something other customers would want to bake into their applications or not.

I actually do think that export to CSV is a common "nice to have" for our customers, especially for ops teams who want ot export to excel for some light analytics. I would advocate for keeping it as part of the core component. 

Use case for the transaction bundle is less clear, but feels weird to keep one in app and one in `react`

rahul1:
> I have one bigger design question -- maybe we should discuss offline with the team -- about whether the Export dialog belongs in the react package or in the app package. I don't have a good sense of whether this is something other customers would want to bake into their applications or not.

I actually do think that export to CSV is a common "nice to have" for our customers, especially for ops teams who want ot export to excel for some light analytics. I would advocate for keeping it as part of the core component. 

Use case for the transaction bundle is less clear, but feels weird to keep one in app and one in `react`

reshmakh:
Agree with @rahul1 - export as a functionality is super useful for customers.  Both on medplum-app and in their custom apps.

packages/react/src/SearchExportDisplay/SearchExportDialog.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import { Meta } from '@storybook/react';
```
codyebberson, line 1: Rename the folder to `SearchExportDialog` to match the component name.

codyebberson (commented):
Hey @jamestouri  - this is looking pretty good.  Just a handful of minor cleanup items then good to go 👍

rahul1 (approved):

codyebberson (commented):
Thanks @jamestouri  -- I just tested it out on localhost.  Things looking good.  One blocker is that it only uses the current page rather than requesting a full set.  For example, I have ~1100 patients, this only exports the current 20 being displayed.  @reshmakh  is the primary customer for this feature, so you may want to sync with her, but I believe the expectation is that this would export all (or up to 1,000 or 10,000 results -- some preconfigured number).

In that case, on click, you'll have to make another request to the server for the same search, except updating `count` and setting `offset=0`.  Rewrite the response using your existing utilities.  You can then remove `setLoadResponse`.

rahul1:
@codyebberson wouldn't we run into the server side cap of 1000 regardless? Unless we paginate the search inside the export. 

It would be more clear / safe if we asked the user to specify the count as part of the search URL, and then export what is displayed. Exporting 10000 results should probably happen using the BulkExport API

codyebberson:
@rahul1 -- You're right, that we should encourage people to use the Bulk Data API.  Arguably "Export CSV", "Export Bundle", and "Export Transaction Bundle" should all use something similar to the Bulk Data API, and start asynchronous background jobs, and should not have any arbitrary caps.

My comment is due to the confusing behavior as currently implemented.  In the current implementation, the two export buttons right next to each other use different export logic.  There is nothing in the UI that would indicate that.

Backing up... This feature is underspecified.  Let's take offline.

packages/app/src/HomePage.tsx
```diff
@@ -218,3 +226,18 @@ function saveLastSearch(search: SearchRequest): void {
 function canCreate(resourceType: string): boolean {
   return resourceType !== 'Bot' && resourceType !== 'ClientApplication';
 }
+
+async function getTransactionBundle(search: SearchRequest, medplum: MedplumClient): Promise<Bundle> {
+  const transactionBundleSearch: SearchRequest = {
+    resourceType: search.resourceType,
+    count: 1000,
```
jamestouri, line 233: Limit to 1k records for transaction bundle

codyebberson (approved):

PR #1801, opened 2023-04-08
Fixes #1730 Project delete operations have arbitrary caps Part 1

### LINKED ISSUE #1730: Project clone and Project delete operations have arbitrary caps
The initial implementation of `Project/ID/$clone` and `Project/ID/$expunge` have arbitrary resource caps.

That is ok-ish for `Project/ID/$clone`.  It should be documented, and perhaps include a warning in the response.

That is not ok for `Project/ID/$expunge`, and should be fixed.  That will require some changes:

1. Move the operation to a background / async job.  It is currently synchronous with the HTTP request.
2. More reliable pagination crawler.
3. Lots of low hanging fruit opportunities to bulk delete more efficiently (currently doing 1-by-1 to use existing `expungeResource` functionality -- it could easily be `expungeAllResources(resourceType)` with a little SQL muscle).

tonylee80:
- have delete operation be asynchronous
- "crawl" through all resources by resource type without arbitrary limit
- batch sql calls using IN operator of resource ids
- batch delete cachekeys

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -863,6 +863,21 @@ export class Repository extends BaseRepository implements FhirRepository {
   }
 
   /**
+   * Permanently deletes the specified resources and all of its history.
+   * This is only available to the system and super admin accounts.
+   * @param resourceType The FHIR resource type.
+   * @param ids The resource IDs.
+   */
+  async expungeResources(resourceType: string, ids: string[]): Promise<void> {
+    if (!this.#isSuperAdmin()) {
+      throw new OperationOutcomeError(forbidden);
+    }
+    await new DeleteQuery(resourceType).where('id', Operator.IN, ids).execute(getClient());
+    await new DeleteQuery(resourceType + '_History').where('id', Operator.IN, ids).execute(getClient());
+    await deleteCacheEntries(resourceType, ids);
+  }
+
+  /**
    * Purges resources of the specified type that were last updated before the specified date.
    * This is only available to the system and super admin accounts.
    * @param resourceType The FHIR resource type.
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -2300,6 +2315,19 @@ async function deleteCacheEntry(resourceType: string, id: string): Promise<void>
...
+ * Deletes cache entries from Redis.
+ * @param resourceType The resource type.
+ * @param ids The resource IDs.
+ */
+async function deleteCacheEntries(resourceType: string, ids: string[]): Promise<void> {
+  const cacheKeys = ids.map((id) => {
+    return getCacheKey(resourceType, id);
+  });
+
+  await getRedis().del(cacheKeys);
```
tonylee80, line 2328: https://redis.io/commands/del/

codyebberson (approved):
Nice, thanks @tonylee80 

One quick question for my own curiosity:  did you try it locally with >1000 resources per resource type?  I'm mostly curious about performance when we run `DELETE FROM "x" WHERE "id" IN ( [1000 UUID values] )`.  It looks like that should be fine for Postgres -- queries can be huge, and can have up to 32k values in `IN` blocks.  So this is pure personal curiosity.

tonylee80:
> Nice, thanks @tonylee80
> 
> One quick question for my own curiosity: did you try it locally with >1000 resources per resource type? I'm mostly curious about performance when we run `DELETE FROM "x" WHERE "id" IN ( [1000 UUID values] )`. It looks like that should be fine for Postgres -- queries can be huge, and can have up to 32k values in `IN` blocks. So this is pure personal curiosity.

I haven't tested this with >1000 resources. I am curious as well! I'll give it a try locally to see how postgres performs.

tonylee80:
> Nice, thanks @tonylee80
> 
> One quick question for my own curiosity: did you try it locally with >1000 resources per resource type? I'm mostly curious about performance when we run `DELETE FROM "x" WHERE "id" IN ( [1000 UUID values] )`. It looks like that should be fine for Postgres -- queries can be huge, and can have up to 32k values in `IN` blocks. So this is pure personal curiosity.

I was able to test this locally where a Project had easily over 10000 resources where one resource type (Patient) with greater than 18k rows. The expunge endpoint was able to remove all the project resources without noticeable issues. I did see the CPU jumping up to around 55% on my docker instances.

PR #1804, opened 2023-04-10
Token Exchange Documentation and API Cleanup

rahul1:


packages/core/src/client.ts
```diff
@@ -855,6 +857,37 @@ export class MedplumClient extends EventTarget {
     window.location.assign(this.getExternalAuthRedirectUri(authorizeUrl, clientId, redirectUri, loginRequest));
   }
 
+  /**
+   * @param token The access token that was generated by the external identity provider.
+   * @param clientId The ID of the `ClientApplication` in your Medplum project that will be making the exchange request.
+   * @param projectId The ID of your Medplum project.
+   * @category Authentication
+   */
+  async exchangeExternalAccessToken(token: string): Promise<ProfileResource> {
```
codyebberson, line 866: 1. JSDoc params out of date with function params
2. We may want to allow `clientId` as a param?  Maybe an optional param that can override `this.#clientId`?  We've used that pattern in other external auth methods.

codyebberson (approved):

PR #1807, opened 2023-04-11
Updating terms to address UMLS licensing

reshmakh:


codyebberson (approved):
I'm not qualified to evaluate the content.  The only thing I noticed is inconsistent use of periods in titles.

https://en.wiktionary.org/wiki/bikeshedding 🙃

codyebberson (approved):

PR #1808, opened 2023-04-11
Paginated Search Documentation

rahul1:


packages/core/src/client.test.ts
```diff
@@ -1142,6 +1150,116 @@ describe('Client', () => {
...
+      for await (const page of client.searchResourcePages('Patient', '_count=1')) {
+        expect(page).toHaveLength(1);
+        expect(page[0].resourceType).toBe('Patient');
+        numPages += 1;
+      }
+
+      expect(numPages).toBe(3);
+    });
+
+    // test('Search resources ReadablePromise', async () => {
```
codyebberson, line 1234: Cleanup

packages/core/src/client.ts
```diff
@@ -1064,6 +1065,54 @@ export class MedplumClient extends EventTarget {
     return promise;
   }
 
+  /**
+   *
```
codyebberson, line 1069: Delete blank line (JSDoc has special treatment of first line)

packages/core/src/client.ts
```diff
@@ -1064,6 +1065,54 @@ export class MedplumClient extends EventTarget {
...
+    resourceType: K,
+    query?: QueryTypes,
+    options: RequestInit = {}
+  ): AsyncGenerator<ExtractResource<K>[]> {
+    let url: URL | undefined = this.fhirSearchUrl(resourceType, query);
+
+    while (url) {
+      const cacheKey = (url.toString() + '-searchResourcesPages') as string;
+      let bundle: Bundle | undefined = await this.#getCacheEntry(cacheKey, options)?.value;
+      console.log('Foo', bundle);
```
codyebberson, line 1100: Delete

packages/core/src/client.ts
```diff
@@ -1064,6 +1065,54 @@ export class MedplumClient extends EventTarget {
...
+   */
+  async *searchResourcePages<K extends ResourceType>(
+    resourceType: K,
+    query?: QueryTypes,
+    options: RequestInit = {}
+  ): AsyncGenerator<ExtractResource<K>[]> {
+    let url: URL | undefined = this.fhirSearchUrl(resourceType, query);
+
+    while (url) {
+      const cacheKey = (url.toString() + '-searchResourcesPages') as string;
```
codyebberson, line 1098: Hm, is there a reason we can't leverage the existing logic inside `search`?  We can take advantage of the fact that `URLSearchParams` is a valid `QueryTypes` option.

```ts
while (query) {
  // Use normal search
  const bundle = await this.search(resourceType, query);

  if (bundle.entry) {
    yield bundle.entry.map((e) => e.resource as ExtractResource<K>);
  }

  const nextLink = bundle?.link?.find((link) => link.relation === 'next');
  if (nextLink) {
    query = new URL(nextLink).searchParams;
  } else {
    query = undefined;
  }
}
```

packages/core/src/client.ts
```diff
@@ -1064,6 +1065,54 @@ export class MedplumClient extends EventTarget {
...
+   * }
+   * ```
+   *
+   * @category Search
+   * @param resourceType The FHIR resource type.
+   * @param query Optional FHIR search query or structured query object. Can be any valid input to the URLSearchParams() constructor.
+   * @param options Optional fetch options.
+   * @returns An async generator, where each result is an array of resources for each page.
+   */
+  async *searchResourcePages<K extends ResourceType>(
```
codyebberson, line 1090: I'm 95% ok with this name.  I'm not aware of "async generator" naming patterns, so we don't have much to go on.  `searchResourcesPaginated`?  `paginatedSearchResources`?  `searchResourcesGenerator`?

codyebberson (approved):

PR #1813, opened 2023-04-12
Perf improvements for partial string matching

codyebberson:
Consider simple "Patient search by name" such as `Patient?name=eve`

Before, that used `name LIKE '%eve%'`.  When patient count gets high, performance degrades badly, and the query could take 50-100 ms.

After, that uses Postgres `tsvector` queries.  Even with millions of patients, the query is single digit milliseconds.


packages/core/src/fhirpath/atoms.ts
```diff
@@ -55,13 +55,10 @@ export class SymbolAtom implements Atom {
     if (this.name === '$this') {
       return context;
     }
-    return context
-      .map((e) => this.#evalValue(e))
-      .flat()
-      .filter((e) => e?.value !== undefined) as TypedValue[];
+    return context.flatMap((e) => this.evalValue(e)).filter((e) => e?.value !== undefined) as TypedValue[];
```
codyebberson, line 58: While researching perf issues, the Node.js profiler revealed that this function is on our CPU hot path.  The reason is that we end up doing about a gajillion `evalFhirPath()` calls when evaluating all search parameters.

packages/core/src/fhirpath/atoms.ts
```diff
@@ -55,13 +55,10 @@ export class SymbolAtom implements Atom {
...
   }
 
-  #evalValue(typedValue: TypedValue): TypedValue[] | TypedValue | undefined {
+  private evalValue(typedValue: TypedValue): TypedValue[] | TypedValue | undefined {
```
codyebberson, line 61: Another interesting revelation from the Node.js profiler is that using the built-in JavaScript `#` for private has significantly worse performance than using TypeScript's `private`. 

TypeScript's `private` only works at compile time.  There is no runtime cost at all.

JavaScript's `#` is transpiled to a polyfill using `WeakMap` and a bunch of other complexity.  I'm tempted to say we should remove all `#` private methods and go back to TypeScript `private`.

packages/server/src/fhir/lookups/address.ts
```diff
@@ -53,15 +53,21 @@ export class AddressTable extends LookupTable<Address> {
    *   address            | address
    *   address-city       | city
    *   address-country    | country
-   *   address-postalcode | postalcode
+   *   address-postalcode | postalCode
```
codyebberson, line 56: This bug has apparently been here forever.  Currently a search for `Patient?address-postalcode` triggers a 500.

rahul1 (approved):

packages/server/src/fhir/repo.ts
```diff
@@ -947,14 +947,18 @@ export class Repository extends BaseRepository implements FhirRepository {
...
-      .groupBy({ tableName: resourceType, columnName: 'id' });
+      .column({ tableName: resourceType, columnName: 'content' });
 
     this.#addDeletedFilter(builder);
     this.#addSecurityFilters(builder, resourceType);
     this.#addSearchFilters(builder, searchRequest);
     this.#addSortRules(builder, searchRequest);
 
+    if (builder.joins.length > 0) {
+      // builder.groupBy({ tableName: resourceType, columnName: 'id' });
```
tonylee80, line 958: remove comment

tonylee80 (approved):

Issue #1749, opened 2023-03-30
Configurable retry policies for webhooks and cron jobs


codyebberson:
Currently the retry policy for webhooks (and soon with cron jobs) is predefined and not configurable.

Different customers often want different policies (initial delay, how many retries, which HTTP status codes map to success/failure).

We should add some mechanism to configure the retry policies.

jamestouri:
@codyebberson is there a list of the different policies we should make customizable? We can start with a small list of attributes to configure or we can make everything customizable from the get go.

codyebberson:
Possible config settings:

*P1: Maximum retry count*

* The goal is to let users control how long to keep retrying
* Current global configuration is 18 retries (~72 hours): https://github.com/medplum/medplum/blob/main/packages/server/src/workers/subscription.ts#L55-L64
* That was done for a customer request, although it's abnormally aggressive
* Probably a simple integer greater than or equal to zero
* The trick will be to to `catch` here: https://github.com/medplum/medplum/blob/main/packages/server/src/workers/subscription.ts#L284-L291
* And then check if the `job.attemptsMade >=` the custom setting

*P2: Success/Failure status codes*

* The goal is to let users control what defines "success"
* Everybody agrees that "200" and "201" are "success" status codes
* Everybody agrees that "429" and "500" should be retries
* There is more disagreement about "400", "401", "404"
* Our current logic is `success = status >= 200 && status < 400`: https://github.com/medplum/medplum/blob/main/packages/server/src/workers/subscription.ts#L322
* Rather than trying to find a global solution, it would be nice to let users specify
* Simple option is to accept a comma separated list of HTTP status codes for success (i.e., "200,201")
* Maybe could specify ranges?  That gets complicated fast though (i.e., "200-399,404")

In a perfect world, we could allow customers to customize "initial delay time" and "exponential backoff factor", but those appear to be global settings in BullMQ `QueueOptions`, so I don't think it's worth the effort to preemptively build.

In both cases, the config settings should be extensions on the FHIR `Subscription` resource.

PR #1816, opened 2023-04-12
Subscription extensions for custom retry and status codes 

jamestouri:
https://github.com/medplum/medplum/issues/1749

### DIFF
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -294,7 +294,7 @@ export async function execSubscriptionJob(job: Job<SubscriptionJobData>): Promis
     }
   } catch (err) {
     const maxJobAttempts =
-      getExtension(subscription, '"http://medplum.com/fhir/StructureDefinition/subscription-max-attempts')
+      getExtension(subscription, 'http://medplum.com/fhir/StructureDefinition/subscription-max-attempts')
         ?.valueInteger ?? MAX_JOB_ATTEMPTS;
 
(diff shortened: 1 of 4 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -55,7 +57,7 @@ export function initSubscriptionWorker(config: MedplumRedisConfig): void {
   queue = new Queue<SubscriptionJobData>(queueName, {
     ...defaultOptions,
     defaultJobOptions: {
-      attempts: 18, // 1 second * 2^18 = 73 hours
+      attempts: MAX_JOB_ATTEMPTS, // 1 second * 2^18 = 73 hours
```
jamestouri, line 60: This will put a limit on the value that the subscription resource can have for max retries as well

packages/server/src/workers/utils.ts
```diff
@@ -106,3 +107,47 @@ export async function createAuditEvent(
...
 }
+
+export function isJobSuccessful(subscription: Subscription, status: number): boolean {
+  const successCodes = getExtension(
+    subscription,
+    'https://medplum.com/fhir/StructureDefinition/subscription-success-codes'
+  );
+
+  if (!successCodes?.valueString) {
+    return status >= 200 && status < 400;
```
codyebberson, line 118: It looks like this line is duplicated in a few places in this function.  Can we break it out as a helper function?

codyebberson (approved):
Overall, this looks great.  One minor suggestion, otherwise good to go 👍

PR #1821, opened 2023-04-13
Config settings to disable new user registration

codyebberson:


reshmakh (approved):

packages/app/src/SignInPage.tsx
```diff
@@ -11,7 +11,7 @@ export function SignInPage(): JSX.Element {
     <SignInForm
       onSuccess={() => navigate(searchParams.get('next') || '/')}
       onForgotPassword={() => navigate('/resetpassword')}
-      onRegister={() => navigate('/register')}
+      onRegister={process.env.REGISTER_ENABLED === 'false' ? undefined : () => navigate('/register')}
```
tonylee80, line 14: Just to confirm, this could be disabled for the app by adding `REGISTER_ENABLED=false` to an optional `.env`? If so, should we add this to https://github.com/medplum/medplum/blob/main/packages/app/.env.defaults?

codyebberson: Correct, I didn't want to change the default behavior and trigger a breaking version change.

Adding to defaults is a good idea.  I'll do that.  Thanks @tonylee80

tonylee80:
Another question, what's the difference between this config vs REGISTER_DISABLED https://github.com/medplum/medplum/blob/cd4d819320f9cc8e0bafe574694d784e04b8381c/packages/app/src/RegisterPage.tsx#L17?

PR #1823, opened 2023-04-13
Changed JS hash private fields to TS private

codyebberson:
During previous performance exploration (https://github.com/medplum/medplum/pull/1813) we found that using the JavaScript hash `#` private modifier caused measurable performance penalties when used in inner loops.

Because we use a relatively conservative JS version target, the `#` modifier was transpiled to a `WeakMap` with a whole bunch of additional boilerplate code.

In contrast, the TypeScript `private` modifier is only used at compile-time, and does not require transpiling, and does not have any runtime penalty.

There should be no functional changes in this PR.  Everything was done with a handful of search-and-replace calls.

----

### Summary

No summary available (the operation was too large to complete, the maximum model work limit was reached)

### Walkthrough
No walkthrough available (the operation was too large to complete, the maximum model work limit was reached)

### DIFF
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -205,13 +205,13 @@ export abstract class BaseRepository {
 }
 
 export class MemoryRepository extends BaseRepository implements FhirRepository {
-  private readonly resources: Record<string, Record<string, Resource>>;
-  private readonly history: Record<string, Record<string, Resource[]>>;
+  private readonly resources: Map<string, Map<string, Resource>>;
+  private readonly history: Map<string, Map<string, Resource[]>>;
 
   constructor() {
     super();
-    this.resources = {};
-    this.history = {};
+    this.resources = new Map();
+    this.history = new Map();
   }
 
   async createResource<T extends Resource>(resource: T): Promise<T> {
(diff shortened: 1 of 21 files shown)


reshmakh (approved):

packages/fhir-router/src/repo.ts
```diff
@@ -305,15 +310,18 @@ export class MemoryRepository extends BaseRepository implements FhirRepository {
     return {
       resourceType: 'Bundle',
       type: 'history',
-      entry: ((this.#history?.[resourceType]?.[id] ?? []) as T[])
+      entry: ((this.history.get(resourceType)?.get(id) || []) as T[])
```
tonylee80, line 313: Just curious, was there a reason why there was a change from using `??` to `||`. It seems like it shouldn't affect the behavior and should be fine as is.

packages/fhir-router/src/repo.ts
```diff
@@ -322,8 +330,13 @@ export class MemoryRepository extends BaseRepository implements FhirRepository {
 
   async search<T extends Resource>(searchRequest: SearchRequest): Promise<Bundle<T>> {
     const { resourceType } = searchRequest;
-    const resources = this.#resources[resourceType] ?? {};
-    const result = Object.values(resources).filter((resource) => matchesSearchRequest(resource, searchRequest));
+    const resources = this.resources.get(resourceType) || new Map();
```
tonylee80, line 333: similar question as above `??` -> `||`

tonylee80 (approved):

packages/fhir-router/src/repo.ts
codyebberson: Good question.  `??` is technically more correct, but it's not natively supported by our target JavaScript version, so it gets transpiled to more complex code.

See: https://www.typescriptlang.org/play?ts=5.0.4#code/GYVwdgxgLglg9mABAEziARgGwKYEUTYDOsCAsgIYBOA1gBQxgAOIUAXIsZQwOaIA+icMmzAG2ZAEpEAbwBQiRBASE4OAHSY43ekxaIA-PsQByYcHIhMUYxIDcsgL6zZoSCSSoMOAAoxG2HWY2DiguMF4BIRExSRl5RWVVbA0tQL0+AVMRCysbewcgA

image

Input:

```ts
function doubleQuestionMark(input: string | undefined) {
  console.log(input ?? 'default');
}

function doublePipe(input: string | undefined) {
  console.log(input || 'default');
}
```

Output:

```js
"use strict";
function doubleQuestionMark(input) {
    console.log(input !== null && input !== void 0 ? input : 'default');
}
function doublePipe(input) {
    console.log(input || 'default');
}
```

In some cases, like this sample code when `input` is a `string`, then the behavior is subtly different when `input` is empty string `''`.

In the PR above, the value is always an `Object`, so `||` and `??` have identical behavior.

By using `||` our bundle size is slightly smaller and the runtime is slightly faster.  This is probably a silly micro optimization.

PR #1843, opened 2023-04-17
Clarify use of Quantities in RangeDisplay component

### LINKED ISSUE #1679: `RangeDisplay` shows a double comparator
image

mattwiller:
Fixes #1679 

The FHIR [`Range`](http://hl7.org/fhir/R4/datatypes.html#Range) data type uses `SimpleQuantity` variants for the range endpoints, which forbids use of the `comparator` field.  References to using Quantity values with a comparator in a Range have been removed.

### DIFF
--- a/packages/core/src/format.test.ts
+++ b/packages/core/src/format.test.ts
@@ -309,6 +309,7 @@ test('Format timing', () => {
 });
 
 test('Format Range', () => {
+  expect(formatRange(undefined)).toBe('');
   expect(formatRange({})).toBe('');
   expect(formatRange({ low: {}, high: {} })).toBe('');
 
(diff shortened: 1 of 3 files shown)


jamestouri (approved):

rahul1:
This is a good clarification! But I think we should leave the underlying issue open, as we have a customer who's seeing this on their dashboard - we may need to modify the react component or the format function to ignore any comparators passed in (or clean up how the data is being input).

rahul1 (approved):

codyebberson:
We could filter out `comparator` inside `formatRange`:

Change https://github.com/medplum/medplum/blob/main/packages/core/src/schema.ts

```ts
  const low = range?.low && { ...range.low };
  const high = range?.high && { ...range.high };
```

To:

```ts
  const low = range?.low && { ...range.low, comparator: undefined };
  const high = range?.high && { ...range.high, comparator: undefined };
```

mattwiller:
@rahul1 Thanks for pointing that out; I'll test out the suggestion from @codyebberson to resolve the issue a bit more cleanly

tonylee80 (approved):
Looks good! 

image

codyebberson (approved):

PR #1845, opened 2023-04-18
chore: Use crypto.randomUUID in @medplum/fhir-router

jdjkelly:
Was poking around in the repository and saw this custom UUID generator. I'm 80% confident that this contribution makes sense but I could have missed something admittedly.

* crypto.randomUUID was released in Node v14.17.0, making the existing comment outdated
* crypto.randomUUID is already used in @medplum/server
* crypto.randomUUID [is now available in a secure context on all modern browsers ](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)(meaning its use in @medplum/mock is fine so long as it's being loaded on `localhost` or `127.0.0.1`)

### DIFF
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -12,6 +12,7 @@ import {
 } from '@medplum/core';
 import { Bundle, BundleEntry, Reference, Resource } from '@medplum/fhirtypes';
 import { applyPatch, Operation } from 'rfc6902';
+import { randomUUID as generateId } from 'crypto';
 
 /**
  * The FhirRepository interface defines the methods that are required to implement a FHIR repository.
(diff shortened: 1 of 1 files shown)


codyebberson:
Hey @jdjkelly  - thanks for the contribution.

One consideration for this is that `fhir-router` is used both client side and server side.  Server side is obvious.  Client side is less obvious, it's used in Medplum [Storybook](https://storybook.medplum.com/) for mock API access.

The `generateId` implementation is only used in the `MemoryRepository`, which is an in-memory implementation of a FHIR data store, again, only used for mock API access and unit tests.

Rather than `from 'crypto'`, we could probably use [`window.crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), although I thought I remembered cross-browser issues with that.  I haven't looked into that in a while. 

As a separate issue - I'm not sure if our tests will catch the fact that `generateId` needs to run client side.... 🍿

codyebberson:
I researched a few options for this, but they all had a certain level of jankiness.

Node.js 20 provides an implementation of the standard [Web Crypto API](https://www.w3.org/TR/WebCryptoAPI/).  When we upgrade to Node.js 20, this will have a very easy solution.  For now, I think we're going to leave this as-is.

Thanks again for the contribution.  Feel free to DM if you'd like to discuss.

https://nodejs.org/api/webcrypto.html#web-crypto-api

PR #1849, opened 2023-04-18
Fixes #1829: Expunge test async error

### LINKED ISSUE #1829: Expunge test async error
image

I think the issue is in `expunge.test.ts` test `Expunge project compartment` - the request continues past the end of the test, and the database has been closed.

May need to add some kind of `wait` mechanisms.

tonylee80:
Before:
image

After:
image

### Summary
🕒🗑️🧪

Added a delay in the expunge test case to avoid a race condition. This ensures that the test checks the database state after the expunge operation is done.

> _`expunge` is queued_
> _test must wait for deletion_
> _winter of patience_

### Walkthrough
* Add a delay to the expunge test case to avoid race condition ([link](https://github.com/medplum/medplum/pull/1849/files?diff=unified&w=0#diff-f750e7e7d34f2965a0d321fc6790e2da27f664ce73dd506e56be3a3592401982R94))

### DIFF
--- a/packages/server/src/fhir/operations/expunge.test.ts
+++ b/packages/server/src/fhir/operations/expunge.test.ts
@@ -5,7 +5,7 @@ import request from 'supertest';
 import { initApp, shutdownApp } from '../../app';
 import { loadTestConfig } from '../../config';
 import { getClient } from '../../database';
-import { createTestProject, initTestAuth, waitFor } from '../../test.setup';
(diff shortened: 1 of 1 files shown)


codyebberson (approved):
Discussed offline.  The "right" solution here is the "AsyncJob" or "BackgroundJob" resource that can track background job state.  Then we could implement polling on the endpoint until the job is done.  This is ok for now to clear up the console output.

Issue #1820, opened 2023-04-13
Help command for the CLI 


jamestouri:
As we are adding more functionality to the CLI we should add the --help/-h flag that shows all of the possible commands

PR #1855, opened 2023-04-19
Integrate Commander js to CLI tool 

jamestouri:
This creates a top down change to integrate commander js to all of the commands, as well as new ones. 

This introduces a bot command hierarchy that leaves room for scaling more in a cleaner way

Commander js automatically adds a helper flag 
https://github.com/medplum/medplum/issues/1820


codyebberson (commented):
Awesome.  One super minor nit in `package.json`, otherwise this looks good to go

codyebberson (approved):

Issue #1872, opened 2023-04-20
Medplum CLI Project List 


jamestouri:
Be able to view projects from the CLI. 

Command will be `medplum project list`

PR #1874, opened 2023-04-20
`Project` command in CLI

jamestouri:
First command is list to see the different projects 

ran locally `node dist/cjs/index.cjs project list` and outputted 

```

Candid <> Medplum Demo (Project/7b01521f-4d9e-41c9-8d0e-d3514a84fffe)

Super Admin (Project/ae4eb181-c468-4d20-8f04-4c6b988790dc)

My-first-project (Project/749bdcee-c77f-4fe8-8c3d-8abc44268215)
```

https://github.com/medplum/medplum/issues/1872

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17755,7 +17755,7 @@
     "node_modules/concat-map": {
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
-      "integrity": "sha512-/Srv4dswyQNBfohGpz9o6Yb3Gz3SrUDqBH5rTuhGR7ahtlbYKnVxw2bCFMRljaA7EXHaXZ8wsHdodFvbkhKmqg==",
+      "integrity": "sha512-/Srv4dswyQNBfohGpz9o6Yb3Gz3SrUDqBH5rTuhGR7ahtlbYKnVxw2bCFMRljaA7EXHaXZ8wsHdodFvbkhKmqg=="
     },
     "node_modules/concat-stream": {
       "version": "2.0.0",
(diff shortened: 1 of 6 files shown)


packages/cli/src/projects.ts
```diff
@@ -0,0 +1,28 @@
+import { Command } from 'commander';
+import { medplum } from '.';
+import { MedplumClient } from '@medplum/core';
+// import { prettyPrint } from './utils';
```
tonylee80, line 4: extra comment

tonylee80 (approved):
Looks good! I love how we can easily add new commands.

packages/cli/src/projects.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+function projectList(medplum: MedplumClient): void {
+  const logins = medplum.getLogins();
+
+  const projects = logins
+    .map((login) => {
+      login.project;
+      const keyValue = Object.entries(login.project);
+      const values = keyValue.map(([key, value]) => `${key}: ${value}`);
```
codyebberson, line 22: This makes assumptions about the contents of `getLogins()` which might not always be true.

Rather than showing all values, let's be explicit about showing just the project details.

For example:

```ts
.map((login) => `${login.project.display} (${login.project.reference})`)
```

codyebberson (commented):
Nice!  One cleanup tweak, otherwise lgtm 👍

package-lock.json
```diff
@@ -16396,6 +16396,7 @@
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
+      "dev": true,
```
tonylee80, line 16399: Just to confirm, is this change intentional?

packages/mock/src/client.ts
```diff
@@ -142,6 +142,13 @@ export class MockClient extends MedplumClient {
     return super.getActiveLogin();
   }
 
+  getLogins(): LoginState[] {
+    if (this.activeLoginOverride !== undefined) {
```
tonylee80, line 146: What's the uses for this change for the CLI update?

package-lock.json
codyebberson: Yeah, we should only change `package-lock.json` if one of the `package.json` files changed.  Let's revert this.

packages/mock/src/client.ts
codyebberson: This is an oddity of `MockClient`.  By default, `MockClient` is logged in a "Dr. Alice Smith", which is a default `Practitioner` resource that you can use in tests.  `activeLoginOverride` is a `MockClient` feature to modify the current mocked login state.

package-lock.json
jamestouri: I ran npm i in the server package to fix a type error in outcomes.ts

```
function getValidationErrorExpression(error: ValidationError): string[] | undefined {
  // ValidationError can be AlternativeValidationError | GroupedAlternativeValidationError | UnknownFieldsError | FieldValidationError
  if (error.type === 'field') {
    return [error.path];
  }
  return undefined;
}
```
The error.type and error.path kept producing errors.

tonylee80 (approved):

PR #1875, opened 2023-04-20
Implement system level export operation

mattwiller:
Adding the [system level `/$export`](https://build.fhir.org/ig/HL7/bulk-data/export.html#endpoint---system-level-export) operation, which should include all resources within the Project that the user has access to.

### DIFF
--- a/packages/server/src/fhir/operations/export.ts
+++ b/packages/server/src/fhir/operations/export.ts
@@ -2,9 +2,8 @@ import { Request, Response } from 'express';
 import { Repository, protectedResourceTypes, publicResourceTypes, systemRepo } from '../repo';
 import { BulkDataExport, Project, ResourceType } from '@medplum/fhirtypes';
 import { BulkExporter } from './utils/bulkexporter';
-import { getReferenceString, getResourceTypes } from '@medplum/core';
+import { getReferenceString, getResourceTypes, accepted } from '@medplum/core';
 import { getConfig } from '../../config';
-import { randomUUID } from 'crypto';
 
 /**
  * Handles a bulk export request.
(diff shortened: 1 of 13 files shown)


packages/server/src/fhir/storage.ts
```diff
@@ -57,18 +57,22 @@ class FileSystemStorage implements BinaryStorage {
...
     const dir = this.getDir(binary);
     if (!existsSync(dir)) {
       mkdirSync(dir);
     }
     const writeStream = createWriteStream(this.getPath(binary), { flags: 'w' });
-    stream.pipe(writeStream);
     return new Promise((resolve, reject) => {
-      writeStream.on('close', resolve);
-      writeStream.on('error', reject);
+      pipeline(input, writeStream, (err) => {
```
codyebberson, line 69: Out of curiosity, what is the reason behind this?

packages/server/src/fhir/operations/export.ts
```diff
@@ -0,0 +1,113 @@
...
+            text: 'Accepted',
+          },
+        },
+      ],
+    });
+}
+
+async function exportResourceType(exporter: BulkExporter, project: Project, resourceType: ResourceType): Promise<void> {
+  const repo = exporter.repo;
+  while (true) {
```
codyebberson, line 82: Sonar will complain about `while (true)`.  You can do some simple refactoring with `let hasMore = true;`

codyebberson (approved):
Overall looks great 👍 

We should make sure you get the Prettier and ESLint VS Code plugins, looks like some formatting issues that ESLint will complain about.

I added a question about stream/pipelines.  I trust that there was a good reason for that, I just don't know what it is.  Looking forward to learning.

packages/server/src/fhir/storage.ts
mattwiller: Nothing major, but I preferred this style for two reasons:
1. It co-locates the pipe operation and the handling of ultimate completion or error neatly
2. It abstracted away the details of which events were necessary to listen to, which I worry about with streams because I don't know them super well

codyebberson:
Looks like Sonar doesn't like that these blocks are duplicated:

```ts
  // Update the BulkDataExport
  await systemRepo.updateResource<BulkDataExport>({
    ...bulkDataExport,
    meta: {
      project: (res.locals.project as Project).id,
    },
    status: 'completed',
    transactionTime: new Date().toISOString(),
    output: Object.entries(exporter.writers).map(([resourceType, writer]) => ({
      type: resourceType as ResourceType,
      url: getReferenceString(writer.binary),
    })),
  });

  // Send the response
  res
    .set('Content-Location', `${baseUrl}fhir/R4/bulkdata/export/${bulkDataExport.id}`)
    .status(202)
    .json({
      resourceType: 'OperationOutcome',
      id: randomUUID(),
      issue: [
        {
          severity: 'information',
          code: 'informational',
          details: {
            text: 'Accepted',
          },
        },
      ],
    });
```

We could fix by moving some into the `BulkExporter` class as utilties (i.e., set `status` to `completed` on finished inside the class).  

We could create a new constant `accepted` `OperationOutcome` here along with other constant `OperationOutcome` resources: https://github.com/medplum/medplum/blob/main/packages/core/src/outcomes.ts

The `OperationOutcome` 

We can also file those as a follow-up task.

Issue #1858, opened 2023-04-19
Storybook not rendering JSX


rahul1:
![Image](https://user-images.githubusercontent.com/1259236/233152495-764958f3-cf91-4cb8-859a-f82405d1f2bf.png)

codyebberson:
@rahul1 @mattwiller What is the actual bug here?  I don't repro:

https://storybook.medplum.com/?path=/docs/medplum-descriptionlist--docs

<img width="957" alt="image" src="https://user-images.githubusercontent.com/749094/233698324-0eaa3f3e-dcc0-4d8a-a99b-9c3dc2703727.png">

codyebberson:
Fixed upstream: https://github.com/medplum/medplum/pull/1882#issuecomment-1518190607

PR #1882, opened 2023-04-21
Fix rendering of source JSX for Storybook components

mattwiller:
Adding the `storysource` addon to correctly render the source code for Storybook components as JSX.  This currently appears to add a bit of extra cruft to the code, but I was unable to configure the addon not to do that.

Screenshot 2023-04-21 at 10 04 49 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "root",
-  "version": "2.0.15",
+  "version": "2.0.16",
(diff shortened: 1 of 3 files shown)


codyebberson:
@rahul1 @mattwiller  - I don't think I follow what we're trying to solve here.  The JSX source already renders, yeah?  See comment here: https://github.com/medplum/medplum/issues/1858#issuecomment-1518133388

PR #1889, opened 2023-04-21
Add _since and _type parameters to system level export

mattwiller:
Added support for the `_since` and `_type` parameters on the [system level bulk export][export] endpoint.

Also, cleaned up the implementation of the `/$export` operation endpoint by extracting code duplicated with the group export endpoint into utility methods on the `BulkExporter` utility class.

[export]: https://build.fhir.org/ig/HL7/bulk-data/OperationDefinition-export.html


codyebberson (approved):
Nice 👍 

Thanks for following up on the code duplication warnings too 🙏

PR #1890, opened 2023-04-22
Add example code for downloading bulk export data to disk

mattwiller:


packages/examples/src/api/fhir/operations/bulk-fhir.py
```diff
@@ -73,8 +71,20 @@
...
+def download_export_to_file(export_record, access_token):
+  # Request the NDJSON export data
+  conn.request(
+    'GET', export_record.url, None, {
+      'Authorization': 'Bearer ' + access_token,
+    })
+  export_data = conn.getresponse()
+
+  # Append NDJSON data to file on disk
+  with open(export_record.type + '.ndjson', 'a') as f:
```
codyebberson, line 83: Super minor, but 'a' vs 'w'?

codyebberson (approved):
👍

packages/examples/src/api/fhir/operations/bulk-fhir.py
mattwiller: 'a' for append, so if there are hypothetically multiple export records for the same resource type they won't overwrite one another

PR #1901, opened 2023-04-23
Draft AWS CLI commands

codyebberson:
Beginning of AWS infra commands as described here: https://github.com/medplum/medplum/discussions/1839

This includes working versions of `aws list`, `aws describe <tag>`, and `aws update-server <tag>`

For example:

```bash
$ medplum aws list

Medplum Tag:     staging
Stack Name:      MedplumStagingStack
Stack ID:        arn:aws:cloudformation:us-east-1:647991932601:stack/MedplumStagingStack/185989a0-7a56-11ec-ab09-12d7316e9f69
Status:          UPDATE_COMPLETE
ECS Cluster:     MedplumStagingStack-BackEndCluster6B6DC4A8-ftLHnzwn7iAy
ECS Service:     MedplumStagingStack-BackEndFargateServiceD3B260C0-2Vhn3insgOUd
App Bucket:      app.staging.medplum.com
Storage Bucket:  medplum-staging-storage

Medplum Tag:     prod
Stack Name:      MedplumStack
Stack ID:        arn:aws:cloudformation:us-east-1:647991932601:stack/MedplumStack/a620de40-bd74-11eb-ad09-0eec813a27fb
Status:          UPDATE_COMPLETE
ECS Cluster:     MedplumStack-BackEndCluster6B6DC4A8-b7rLxsX1zQdL
ECS Service:     MedplumStack-BackEndFargateServiceD3B260C0-S35OIZcOSp1P
App Bucket:      app.medplum.com
Storage Bucket:  medplum-storage

$ medplum aws update-server staging

Service "MedplumStagingStack-BackEndFargateServiceD3B260C0-2Vhn3insgOUd" updated successfully.
```

Related future work: 

 - `init` command - move over the `init` command in `@medplum/cdk`, might also be out of scope for this PR
 - `deploy-app` command - that might be out of scope for this PR, because it requires a number of other supporting changes
 - ability to list available versions
 - ability to upgrade to a specific version

### DIFF
--- a/packages/cli/src/aws.ts
+++ b/packages/cli/src/aws.ts
@@ -1,0 +1,160 @@
+import {
+  CloudFormationClient,
+  DescribeStackResourcesCommand,
+  DescribeStacksCommand,
+  ListStacksCommand,
+  Stack,
+  StackResource,
+  StackSummary,
+} from '@aws-sdk/client-cloudformation';
+import { ECSClient, UpdateServiceCommand } from '@aws-sdk/client-ecs';
+import { Command } from 'commander';
+
+interface MedplumStackDetails {
+  stack: Stack;
+  tag: string;
+  ecsCluster?: StackResource;
+  ecsService?: StackResource;
+  appBucket?: StackResource;
+  storageBucket?: StackResource;
+}
+
+export const aws = new Command('aws').description('Commands to manage AWS resources');
+const client = new CloudFormationClient({});
+const tagKey = 'medplum:environment';
+
+aws
+  .command('list')
+  .description('List Medplum AWS CloudFormation stacks')
+  .action(async () => {
+    const stackSummaries = await listStacks();
+    for (const stackSummary of stackSummaries) {
+      const stackName = stackSummary.StackName;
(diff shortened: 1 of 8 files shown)


packages/cli/src/aws.ts
```diff
@@ -0,0 +1,160 @@
...
+      const details = await getStackDetails(stackName);
+      if (!details) {
+        continue;
+      }
+      printStackDetails(details);
+      console.log('');
+    }
+  });
+
+aws
```
jamestouri, line 42: Another thought could be having the arg `tag` as optional (I believe in commander optional arguments use [] e.g '[tag]'?) And we can see the details of the stack this way too. Up to you

jamestouri (approved):
Very Very nice 👍, love where this is going!

packages/cli/src/aws.ts
codyebberson: Yeah... I considered that, given that the `list` and `describe` commands are currently so similar.  My thinking is that `describe` will evolve over time to show more details, vs `list` will be more of an abbreviated summary.

PR #1914, opened 2023-04-25
Correct search control icons for fields and filters buttons

tonylee80:
Before:
image

After:
image


codyebberson (approved):
D'oh!  🤦   Nice catch

Issue #1913, opened 2023-04-25
Use correct version in `medplum --version`


codyebberson:
As I was responding to https://github.com/medplum/medplum/issues/1909 I noticed that `medplum --version` displays a separate version number from the `@medplum/core` version.  We should synchronize those. 

You can get the version like this:

```ts
import { MEDPLUM_VERSION } from '@medplum/core';
```

PR #1915, opened 2023-04-25
Synchronize medplum version

jamestouri:
https://github.com/medplum/medplum/issues/1913


mattwiller:
We may want to version these independently, since the interfaces could change at different times from one another.  For example, one week's release might include breaking changes to the CLI, but not to the server — those should probably result in different version numbers in my mind

packages/cli/src/index.ts
```diff
@@ -20,7 +21,7 @@ export async function main(medplumClient: MedplumClient, argv: string[]): Promis
   }
   try {
     const index = new Command('medplum').description('Command to access Medplum CLI');
-    index.version('0.1.0');
+    index.version(MEDPLUM_VERSION ?? '');
```
tonylee80, line 24: Would `MEDPLUM_VERSION` be undefined or null?

jamestouri: It's based on the environment variables

codyebberson: Hm, this is a good catch.  Arguably we should fix this in `packages/core/src/client.ts` so that `MEDPLUM_VERSION` is always a string.

codyebberson (approved):

packages/cli/src/index.ts
```diff
@@ -5,6 +5,7 @@ import { login, whoami } from './auth';
 import { aws } from './aws';
 import { bot, createBotDeprecate, deployBotDeprecate, saveBotDeprecate } from './bots';
 import { project } from './projects';
+import { MEDPLUM_VERSION } from '@medplum/core';
```
tonylee80, line 8: There is a sonar warning regarding multiple imports of '@medplum/code`.

https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=1915&id=medplum_medplum&open=AYe5bOm_LR-8VPjwL99h

codyebberson: Ah, this is a good catch.  The VS Code "Organize Imports" feature is great.  I recommend setting it up to run it on save (see https://stackoverflow.com/questions/49919726/how-to-invoke-the-organize-imports-typescript-feature-in-visual-studio) or run compulsively manually. 

image

PR #1917, opened 2023-04-25
Storybook SearchControl: clicking filter button causes exception

tonylee80:
Before:
image

After:
image


packages/react/src/SearchFilterEditor/SearchFilterEditor.tsx
```diff
@@ -40,7 +40,7 @@ export function SearchFilterEditor(props: SearchFilterEditorProps): JSX.Element
   }
 
   const resourceType = props.search.resourceType;
-  const searchParams = globalSchema.types[resourceType].searchParams as Record<string, SearchParameter>;
+  const searchParams = (globalSchema.types[resourceType].searchParams as Record<string, SearchParameter>) ?? {};
```
tonylee80, line 43: default to object if there is no globalSchema searchParams for a resourceType in storybook.

mattwiller (approved):

Issue #1658, opened 2023-03-15
Support multiple JWKS matches in client assertion auth


codyebberson:
Medplum supports "client assertion" auth for clients configured with a JWKS URL.

To verify the JWT against the JWKS, we use the `jwtVerify` method from the `jose` library.

The default implementation of `jwtVerify` uses relatively strict logic, such that there must be only one eligible key after filtering on `kid`,  `alg`, `use`, and `key_ops`.

Apparently other SMART-on-FHIR servers, such as Epic, do not use such strict logic, and allow for multiple matching keys.

The `jose` library documents how to support this: https://github.com/panva/jose/blob/main/docs/functions/jwks_local.createLocalJWKSet.md#readme

```js
const options = {
  issuer: 'urn:example:issuer',
  audience: 'urn:example:audience',
}
const { payload, protectedHeader } = await jose
  .jwtVerify(jwt, JWKS, options)
  .catch(async (error) => {
    if (error?.code === 'ERR_JWKS_MULTIPLE_MATCHING_KEYS') {
      for await (const publicKey of error) {
        try {
          return await jose.jwtVerify(jwt, publicKey, options)
        } catch (innerError) {
          if (innerError?.code === 'ERR_JWS_SIGNATURE_VERIFICATION_FAILED') {
            continue
          }
          throw innerError
        }
      }
      throw new jose.errors.JWSSignatureVerificationFailed()
    }

    throw error
  })
console.log(protectedHeader)
console.log(payload)
```

That logic would need to be integrated into the Medplum client assertion logic here: https://github.com/medplum/medplum/blob/main/packages/server/src/oauth/token.ts#L307-L366

PR #1921, opened 2023-04-26
Support multiple JWKS matches in client assertion auth

jamestouri:
https://github.com/medplum/medplum/issues/1658

### DIFF
--- a/packages/server/src/oauth/token.test.ts
+++ b/packages/server/src/oauth/token.test.ts
@@ -2,7 +2,7 @@ import { OAuthGrantType, OAuthTokenType, createReference, parseJWTPayload, parse
 import { AccessPolicy, ClientApplication, Login, Project, SmartAppLaunch } from '@medplum/fhirtypes';
 import { randomUUID } from 'crypto';
 import express from 'express';
-import { SignJWT, generateKeyPair } from 'jose';
+import { SignJWT, generateKeyPair, jwtVerify } from 'jose';
 import fetch from 'node-fetch';
 import request from 'supertest';
 import { createClient } from '../admin/client';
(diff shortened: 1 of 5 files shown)


packages/server/src/oauth/token.ts
```diff
@@ -462,7 +462,25 @@ async function parseClientAssertion(clientAssertiontype: string, clientAssertion
 
   try {
     await jwtVerify(clientAssertion, JWKS, verifyOptions);
-  } catch (err) {
+  } catch (error: any) {
+    // There are some edge cases where there are multiple matching JWKS
+    // and we need to iterate throught the JWKSMultipleMatchingKeys error
+    // and return the first verified match
+    if (error?.code === 'ERR_JWKS_MULTIPLE_MATCHING_KEYS') {
```
jamestouri, line 469: https://github.com/panva/jose/blob/main/docs/functions/jwks_remote.createRemoteJWKSet.md

This shows the logic for multiple matching keys

packages/server/src/oauth/token.test.ts
```diff
@@ -18,12 +18,27 @@ import { hashCode } from './token';
 jest.mock('jose', () => {
   const core = jest.requireActual('@medplum/core');
   const original = jest.requireActual('jose');
+  let count = 0;
```
jamestouri, line 21: Test is based on counting the number of iterations to determine coverage of the different errors and invoking the jwt verify method

codyebberson (approved):
Nice, this is great.  I'm sure it was a pain to get those mocks right.  Nice job.

Sonar thinks the method is too complicated.  You could probably move lines 470-482 into a helper function to make Sonar happy.  Your call if you want to do that now, or in a separate PR.

tonylee80 (approved):
Great work on solving the jwt jest tests!

PR #1930, opened 2023-04-27
Implement iterate modifier for _(rev)include search params

### LINKED ISSUE #1767: Implement FHIR search `_iterate`
Some of our customers could benefit from this in places where they are currently using graphQL. Some specific examples from the field: 

- Querying `RelatedPersons` and `Patients` related to a specific patient
- Querying `ActivityDefinitions` and `ObservationDefinitions` associated with a `PlanDefinition`

Need to implement some safeguards to maintain server performance

### LINKED ISSUE #1929: _include on a search which returns 0 resources throws an error
Issue:
Adding the search param `_include` to a FHIR search which returns 0 resources results in a 500 status code being returned by the server.

How to reproduce:
- Call a FHIR endpoint with search params which result in 0 resources being returned
- Add the param _include to the previous search query

Expectation:
- Successful http status code
- 0 resources

Result:
- a 500 http status code

mattwiller:
Adding support for the `:iterate` modifier on `_include` and `_revinclude` search parameters.  This enables searches to recursively include resources by following `reference` search parameters.  For example, consider a graph of resources like the following:

```mermaid
flowchart TD
    Observation -->| patient | Patient
    Patient -->| link | P2[Patient 2]
    Patient -->| link | RelatedPerson
    P2 -->| link | P3[Patient 3]
```
The following search query from the [FHIR specification](https://www.hl7.org/fhir/search.html#modifieriterate) would return all five of these resources together in one Bundle:
```http
GET /Observation?code=http://snomed.info/sct|3738000&_include=Observation:patient&_include:iterate=Patient:link
```

The `:iterate` modifier causes the related `_include` or `_revinclude` to apply iteratively against each level of included resources until no more results are found.  There are limits on the recursion depth and total result set size to prevent excessive server load.

Fixes #1767
Fixes #1929


packages/core/src/search/search.ts
```diff
@@ -199,11 +210,27 @@ function parseKeyValue(searchRequest: SearchRequest, key: string, value: string)
       break;
 
     case '_include':
-      searchRequest.include = value;
+      const includeTarget = parseIncludeTarget(value); //eslint-disable-line no-case-declarations
```
codyebberson, line 213: If you want to avoid the eslint override, you can also use a `{}` block

```ts
case '_include': {
    // ...
  }
  break;
```

packages/core/src/search/search.ts
```diff
@@ -337,6 +364,34 @@ function parseModifier(modifier: string): Operator {
...
+
+  parts.forEach((p) => {
+    if (p === '*') {
+      throw new Error(`'*' is not supported as a value for search inclusion parameters`);
+    }
+  });
+
+  if (parts.length === 1) {
+    // Full wildcard, not currently supported
+    throw new Error(`Invalid include value '${input}': must be of the form ResourceType:search-parameter`);
```
codyebberson, line 378: For errors, we try to use `OperationOutcomeError`

For example:

```ts
throw new OperationOutcomeError(badRequest('Invalid ...'));
```

This provides some extra structure:
* On the server, it can set the appropriate HTTP status codes from the `OperationOutcome`
* On the client side, a bunch of the UI elements use `OperationOutcome.issue.expression` to show the error next to the appropriate UI element

I'm not 100% sure what will happen here, but I suspect it has risk of HTTP 500?  Depends on context and if we're `try/catch`ing upstream.

packages/server/src/fhir/repo.test.ts
```diff
@@ -2229,6 +2246,365 @@ describe('FHIR Repo', () => {
...
+    └┬────────┬────────┬────────┬─────┬┘            
+    ┌▽──────┐┌▽──────┐┌▽──────┐┌▽───┐┌▽────────────┐
+    │linked2││linked1││related││org1││practitioner1│
+    └┬──────┘└┬──────┘└───────┘└────┘└─────────────┘
+     │┌───────▽───────┐
+     ││linked3        │
+     │└┬────────────┬─┘                
+    ┌▽─▽──────────┐┌▽─────┐                         
+    │practitioner2││org2 *│                         
+    └─────────────┘└──────┘
```
codyebberson, line 2269: Fancy!

packages/server/src/fhir/repo.ts
```diff
@@ -972,12 +973,43 @@ export class Repository extends BaseRepository implements FhirRepository {
         } as BundleEntry)
     );
 
-    if (searchRequest.include) {
-      entries.push(...(await this.getSearchIncludeEntries(searchRequest.include, resources)));
-    }
+    if (searchRequest.include || searchRequest.revInclude) {
```
codyebberson, line 976: I imagine that Sonar will complain about code complexity here.  This method has had a lot of churn, and a lot of refactoring to keep complexity down.

codyebberson (commented):
Nice, this looks great.  And ASCII art! 🕺 

See comments on `OperationOutcomeError` + helper function in `getSearchEntries`.

After that, lgtm, ship it 👍

codyebberson (approved):

rahul1:
Looks great! Can't wait to try it out. Just a friendly remind to follow up with some docs :)

PR #1932, opened 2023-04-27
Project Invite Command

jamestouri:
https://github.com/medplum/medplum/issues/1923

Ran `node dist/cjs/index.cjs project invite --admin james touri james@medplum.com` and it showed up on the `/admin/users` page

Ran `node dist/cjs/index.cjs project invite --admin --send-email james touri james@medplum.com` and I received an email for the invite


packages/cli/src/projects.test.ts
```diff
@@ -0,0 +1,252 @@
...
+  writeFileSync: jest.fn(),
+  promises: {
+    readFile: jest.fn(async () => '{}'),
+  },
+}));
+
+let medplum: MedplumClient;
+
+describe('CLI Project', () => {
+  const env = process.env;
```
codyebberson, line 23: I think you can delete all of the `process.env` saving and restoring, because you're not modifying env variables anywhere in the file.

packages/cli/src/projects.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  beforeEach(() => {
+    jest.resetModules();
+    jest.clearAllMocks();
+    process.env = { ...env };
```
codyebberson, line 28: Delete

packages/cli/src/projects.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    medplum = new MockClient({ storage: new FileSystemStorage() });
+    console.log = jest.fn();
+    console.error = jest.fn();
+    process.exit = jest.fn() as never;
+  });
+
+  afterEach(() => {
+    process.env = env;
```
codyebberson, line 36: Delete

packages/cli/src/projects.test.ts
```diff
@@ -0,0 +1,252 @@
+import { MedplumClient } from '@medplum/core';
```
codyebberson, line 1: Nice, I like breaking this out into separate file.

Any objections to using `project.ts` and `project.test.ts`?  We can establish the convention that the filename should === the command category.

packages/cli/src/projects.ts
```diff
@@ -50,3 +73,31 @@ async function switchProject(medplum: MedplumClient, projectId: string): Promise
...
+  const body = {
+    firstName,
+    lastName,
+    resourceType,
+    email,
+    sendEmail,
+    admin,
+  };
+  try {
+    await medplum.post('admin/projects/' + projectId + '/invite', body);
```
codyebberson, line 95: You should be able to now use `medplum.invite()`, which would provide some type safety.

It might make sense to align this function signature with that signature as well.  That's more of a style choice though, your call.

codyebberson (changes requested):

packages/cli/src/projects.test.ts
jamestouri: Yeah good cal, I'll change that

tonylee80 (approved):

codyebberson (approved):

PR #1934, opened 2023-04-28
Fixes #1933 - GraphQL Connection API

### LINKED ISSUE #1933: FHIR GraphQL Connection API
# FHIR GraphQL Connection API

The connection api for graphql is outlined here: https://hl7.org/fhir/graphql.html#searching

It is an alternative api to query lists of resources that provides some additional metadata about the resource list.

Specifically, it allows querying for `ResourceConnection`s of the form

```graphql
{ 
   ConditionConnection (clinical_status: active, patient: example) { 
     count offset pagesize
     edges {
       mode, score, resource { id, active }
     }
     first previous next last
   } 
}
```

The equivalent query with the `ResourceList` api would look like this

```graphql
{ 
   ConditionList(clinical_status: relapse, patient: example) { 
      id, active
   } 
}
``` 

Simple queries are more difficult to express with the `Connection` api, but it offers metadata about the query that could not be expressed in the `List` api. 

My specific use case is wanting to know how many items are in the list without having to query for every item (there may be very many items). I don't see a way to do this with the existing api. 

## Challenges with implementing this new feature

- Implementing the cursor parameter in a reasonable way
  - Fhir spec doesn't specify how to implement cursors
  - I could see cursors being implemented as a combination of an offset and a count for ease of implementation
  - Cursors could be a reference to the last datatype int the list, but I don't know of an easy api to query postgres for `next x items matching these filters after this item`  
- Ongoing maintenance 
  - Fhir spec doesn't specify how to deal with this either :(

## Alternatives Considered
- Just query for every resource 
  - This is likely what I will do in the meantime because my queries are currently not too large
- Perform the query with the fhir api instead of the graphql api
  - The standard api doesn't offer the query nesting that the graphql api does offer

_Originally posted by @msdrigg in https://github.com/medplum/medplum/discussions/1231_

codyebberson:


packages/docs/docs/search/graphql.mdx
```diff
@@ -222,9 +222,46 @@ Another common use is to filter an `extension` array by `url`:
 
 See the "[List Navigation](https://hl7.org/fhir/r4/graphql.html#list)" section of the FHIR GraphQL specification for more information.
 
+## Connection API
+
+Using the normal "List" search (i.e., "PatientList") is the most common way to search for resources. However, the FHIR GraphQL specification also supports the [Connection API](https://hl7.org/fhir/graphql.html#searching), which is a more complex way to search for resources.
+
+The most immediate advantage to the Connection API is better support for pagination and retrieving total counts. The Conenction API also includes more features from FHIR Bundle such as `mode` and `score`.
```
rahul1, line 229: > Conenction
nit: typo

tonylee80 (approved):

packages/docs/docs/search/graphql.mdx
rahul1, line 229: > better support for pagination 
Would you mind adding a bit about this in your current example? I read the spec but was a bit confused about how to use it end-to-end.  A working example would be helpful

rahul1: Or remove "better support for pagination" until we have the cursor implemented

PR #1940, opened 2023-04-28
Document search inclusion parameters

mattwiller:


codyebberson:
Snyk warnings: https://app.snyk.io/org/medplum/project/299880bd-1b52-470c-9418-0df5c447e1f1/pr-check/268f19f5-78f8-4a46-a2aa-6beffb01a466

Can be fixed by specifying version overrides:
- Upgrade trim to version 0.0.3 or higher.
- Upgrade got to version 11.8.5, 12.1.0 or higher.

packages/examples/src/search/includes.ts
```diff
@@ -0,0 +1,11 @@
+import { MedplumClient } from '@medplum/core';
+
+const medplum = new MedplumClient();
+
+// start-block search
```
rahul1, line 5: Two asks on this: 

1. Could you generate an example that uses `_revInclude` as well?
2. Could you generate an example response for each, here ([example](https://github.com/medplum/medplum/blob/main/packages/examples/src/search/graphql.ts#L48-L67)). ChatGPT is pretty good for this, so you don't have to generate fake data yourself

rahul1 (changes requested):
This is fantastic - thank you for taking the time to put this together. 

Made a few requests, mostly cosmetic

rahul1:
This is almost done! My last request is to unify the code examples, diagrams and explanations by inlining the 'Example Usage' examples directly into your explanations. 

Right now you have a lot of great explanations up top, but readers need to map those to the "Example Usage" section. 

My ask is: 

- For each example where you show a REST command (`GET /fhir/R4/Observation...`):
  - inline  the equivalent example using the `MedplumClient`. put this in a Typescript tab as we do [here](https://medplum-www-git-documentincludeiterate-medplum.vercel.app/docs/search/basic-search) 
  - inline example response (as you do in "example usage"), hidden inside a summary. 
  - Also modify all the ids/references in your response to match your diagrams (e.g. Patient/1) instead of `'Patient/e8585ae6-921f-415f-984e-dc5695de0e36`

rahul1 (approved):

PR #1943, opened 2023-04-29
UI suggestions

codyebberson:
- Move "Add bookmark" link to bottom
- Make button smaller
- Make icon smaller
- Use "subtle" variant rather than "white" to get hover styles
- Use `<NativeSelect>` rather than `<Select>` (smaller, faster)

### DIFF
--- a/packages/react/src/AppShell/Navbar.tsx
+++ b/packages/react/src/AppShell/Navbar.tsx
@@ -1,6 +1,6 @@
 import { Button, createStyles, Navbar as MantineNavbar, Space, Text } from '@mantine/core';
 import { IconPlus } from '@tabler/icons-react';
-import React, { useRef, useState } from 'react';
+import React, { useState } from 'react';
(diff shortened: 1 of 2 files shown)


packages/react/src/AppShell/Navbar.tsx
```diff
@@ -76,18 +76,10 @@ export interface NavbarProps {
...
-
 export function Navbar(props: NavbarProps): JSX.Element {
   const { classes } = useStyles();
   const navigate = useMedplumNavigate();
-  const [state, setState] = useState<NavBarState>({
-    bookmarkDialogVisible: false,
-  });
-  const stateRef = useRef<NavBarState>(state);
-  stateRef.current = state;
+  const [bookmarkDialogVisible, setBookmarkDialogVisible] = useState(false);
```
tonylee80, line 82: Thanks for adding this!

tonylee80 (approved):

PR #1944, opened 2023-04-30
Refactored AWS CLI commands into separate files

codyebberson:
No functional changes, pure code cleanup.

Before: All AWS CLI commands were in a single file (`aws.ts`).

After: Each AWS CLI command is in its own separate file (`describe.ts`, `list.ts`, `update-app.ts`, `update-server.ts`).

This is prep work for moving the CDK `init.ts` into the CLI project.  The `aws.ts` file was already getting unwieldy, so this PR is cleaning house to get ready for the party.

Net lines increased, because I also added more tests.  Also, each command has its own test files, and there's some boilerplate to setup AWS mocks.


tonylee80 (approved):

packages/cli/src/aws/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+aws.command('list').description('List Medplum AWS CloudFormation stacks').action(listStacksCommand);
+
+aws
+  .command('describe')
+  .description('Describe a Medplum AWS CloudFormation stack by tag')
+  .argument('<tag>')
+  .action(describeStacksCommand);
+
+aws.command('update-server').description('Update the server image').argument('<tag>').action(updateServerCommand);
```
tonylee80, line 17: Nice cleanup of all the commands!

Issue #1937, opened 2023-04-28
Improvements to precomputed graphql schema


codyebberson:
Medplum disables GraphQL "introspection" queries in production due to DoS risks.  The FHIR GraphQL schema is massive (~47 MB), which makes it an easy target for malicious users to quickly generate server load.

Instead, we build a static copy of the introspection query response, and host it in S3 + CloudFront.  If you sign into https://graphiql.medplum.com/ you will notice that the introspection query loads this static file.

Here is our "fetcher" override in the graphiql project: https://github.com/medplum/medplum/blob/main/packages/graphiql/src/index.tsx#L63-L68

```ts
function fetcher(params: FetcherParams): Promise<SyncExecutionResult> {
  if (params.operationName === 'IntrospectionQuery') {
    return fetch('/schema/schema-v3.json').then((res) => res.json());
  }
  return medplum.graphql(params.query, params.operationName, params.variables);
}
```

That all works "fine", but there are a number of opportunities for improvement:
1. "Introspection" queries are disabled in code.  Ideally it would be a server config setting. 
2. The prebuilt schema URL is hard coded.  Ideally it would be a config setting in the `graphiql` project.
3. The prebuilt schema is too big for CloudFront's automatic compression feature, so we send the 47 MB file uncompressed.  This is an embarrassingly compressible file, it should be ~2 MB gzipped.  To use gzip, we would need to manually compress the file, and upload it with the appropriate `Content-Encoding` header.
4. Re-building the schema is currently a manual process.  Ideally it would be automated.
    a. Enable introspection queries in the server
    b. Disable the graphiql fetcher override
    c. Find the introspection query in chrome dev tools
    d. Save the response as a file
    e. Upload to S3 (ideally gzipped)

codyebberson:
I manually uploaded `schema-v5.json` with the following settings:

<img width="610" alt="image" src="https://github.com/medplum/medplum/assets/749094/f6b48863-8292-4380-8e57-97e3b536d597">

Which works: https://graphiql.medplum.com/schema/schema-v5.json

And is only 2mb vs 48mb 🎉

codyebberson:
If possible, we should enable CORS on the schema file as well: 
https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-cloudfront-supports-cors-security-custom-http-response-headers/

That way we can use it from localhost.

PR #1946, opened 2023-04-30
Add unit test coverage for fhir router graphql

tonylee80:
### Summary
🧪🔗🚀

Added a test case for the GraphQL connection API without the count field. This tests the performance and correctness of the custom FHIR pagination and filtering extension in `packages/fhir-router/src/graphql.test.ts`.

> _`connection` test_
> _without the costly `count` field_
> _autumn leaves fall fast_

### Walkthrough
* Add a test case for the GraphQL connection API without the count field ([link](https://github.com/medplum/medplum/pull/1946/files?diff=unified&w=0#diff-a0896bc706df6ce472112e4f8181a84141bdeb5fab7ec9fa85896ef053be94bbR979-R1011)). This test verifies that the connection API can handle queries that do not request the total count of matching resources, which can be expensive to compute. The connection API is a custom extension of the FHIR standard that allows pagination and filtering of FHIR resources using GraphQL syntax. The test file is `packages/fhir-router/src/graphql.test.ts`.

### DIFF
--- a/packages/fhir-router/src/graphql.test.ts
+++ b/packages/fhir-router/src/graphql.test.ts
@@ -976,4 +976,37 @@ describe('GraphQL', () => {
       edges: [{ resource: { name: [{ given: ['Alice'] }] } }],
     });
   });
+
+  test('Connection API without count field', async () => {
+    const request: FhirRequest = {
+      method: 'POST',
+      pathname: '/fhir/R4/$graphql',
+      query: {},
+      params: {},
+      body: {
+        query: `
+      {
+        PatientConnection(name: "Smith") {
+          offset pageSize
+          edges {
+            mode, score, resource { id name { given } }
+          }
+          first previous next last
+        }
+      }
+    `,
+      },
+    };
+
+    const res = await graphqlHandler(request, repo);
+    expect(res[0]).toMatchObject(allOk);
+
+    const data = (res?.[1] as any).data;
+    expect(data.PatientConnection).toBeDefined();
+    expect(data.PatientConnection).toMatchObject({
+      offset: 0,
+      pageSize: 20,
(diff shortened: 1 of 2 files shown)


packages/fhir-router/src/graphql.ts
```diff
@@ -157,16 +157,13 @@ export async function graphqlHandler(req: FhirRequest, repo: FhirRepository): Pr
 
   const dataLoader = new DataLoader<Reference, Resource>((keys) => repo.readReferences(keys));
 
-  let result: any = introspection && introspectionResults.get(query);
```
tonylee80, line 160: due to added guard above at https://github.com/medplum/medplum/pull/1946/files#diff-edf7ac6d441b8b49c1649bdc9c7da39a333b108c8494d4324eaaf5d19a29eff5L154, `introspection` will always be false here.

packages/fhir-router/src/graphql.test.ts
```diff
@@ -976,4 +976,37 @@ describe('GraphQL', () => {
       edges: [{ resource: { name: [{ given: ['Alice'] }] } }],
     });
   });
+
+  test('Connection API without count field', async () => {
```
tonylee80, line 980: adds coverage for https://sonarcloud.io/code?id=medplum_medplum&pullRequest=1934&selected=medplum_medplum%3Apackages%2Ffhir-router%2Fsrc%2Fgraphql.ts&line=590

packages/fhir-router/src/graphql.ts
tonylee80: added the code back, just noticed #1937 where introspection support would be managed through a server config.

codyebberson (approved):
🎉

PR #1973, opened 2023-05-02
Fix error when :iterate tries to search past leaf resources

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -2327,6 +2327,11 @@ async function getCacheEntry<T extends Resource>(resourceType: string, id: strin
  * @returns Array of cache entries or undefined.
  */
 async function getCacheEntries(references: Reference[]): Promise<(CacheEntry<Resource> | undefined)[]> {
+  const referenceKeys = references.map((r) => r.reference as string);
+  if (referenceKeys.length === 0) {
```
codyebberson, line 2331: Looks like we're currently double `map`ping.  I'd either check `if (references.length === 0)` or reuse `referenceKeys` online 2335.

mattwiller: Ooops, the edit got lost in some branch shuffling — thanks for catching that!

tonylee80 (approved):

Issue #1981, opened 2023-05-04
Gracefully handle 404s for non-existent route


rahul1:
Community Request

When the `MedplumClient` makes a request to a non-existing path (e.g. `https://this.is.madeup/fhir/R4`), a 404 is returned. However, [we still attempt to parse the response as JSON](https://github.com/medplum/medplum/blob/0dfa5f4abb09719d371081b05e5227a8475b8c58/packages/core/src/client.ts#L2211) and show a JSON parse error. A clearer error message would help our users

Example code: 
https://codesandbox.io/p/sandbox/gracious-sound-f3t3zh?file=%2Fsrc%2Findex.ts

PR #1985, opened 2023-05-04
Community Request: Gracefully Handle 404s 

jamestouri:
https://github.com/medplum/medplum/issues/1981

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -79,6 +79,7 @@ describe('Client', () => {
 
   beforeEach(() => {
     localStorage.clear();
+    console.error = jest.fn();
     Object.defineProperty(globalThis, 'Buffer', { get: () => originalBuffer });
     Object.defineProperty(globalThis, 'window', { get: () => originalWindow });
   });
@@ -1263,6 +1264,19 @@ describe('Client', () => {
     const result = await client.search('Patient');
     expect(result).toBeDefined();
     expect(client.getCachedReference(createReference(result.entry?.[0]?.resource as Patient))).toBeDefined();
+  });
+
+  test('Search and return 404', async () => {
+    const fetch = mockFetch(404, {
+      resourceType: 'Bundle',
+      entry: [{ resource: { resourceType: 'Patient', id: '123' } }],
+    });
+    const client = new MedplumClient({ fetch });
+    try {
+      await client.search('Patient');
+    } catch (err) {
+      expect(console.error).toBeCalledWith(`404, couldn't retrieve response from https://api.medplum.com/fhir/R4/Patient`);
+    }
   });
 
   describe('Paginated Search ', () => {
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1266,18 +1266,12 @@ describe('Client', () => {
   });
 
   test('Search and return 404', async () => {
-    console.error = jest.fn();
-    const fetch = mockFetch(404, {
-      resourceType: 'Bundle',
-      entry: [{ resource: { resourceType: 'Patient', id: '123' } }],
-    });
+    const fetch = mockFetch(404, () => 'string_representation');
     const client = new MedplumClient({ fetch });
     try {
       await client.search('Patient');
     } catch (err) {
-      expect(console.error).toBeCalledWith(
-        `404, couldn't retrieve response from https://api.medplum.com/fhir/R4/Patient`
-      );
+      expect((err as OperationOutcomeError).outcome).toMatchObject(notFound);
     }
   });
 
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -79,7 +79,6 @@ describe('Client', () => {
 
   beforeEach(() => {
     localStorage.clear();
-    console.error = jest.fn();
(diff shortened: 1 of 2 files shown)


packages/core/src/client.ts
```diff
@@ -2206,6 +2206,11 @@ export class MedplumClient extends EventTarget {
       return undefined as unknown as T;
     }
 
+    if (response.status === 404) {
+      console.error(`404, couldn't retrieve response from ${url}`);
```
mattwiller, line 2210: Other error messages are prepending with e.g. "Error: "; consider aligning with that pattern.  Maybe something like `Error: URL ${url} not found`

packages/core/src/client.ts
```diff
@@ -2206,6 +2206,11 @@ export class MedplumClient extends EventTarget {
...
+      throw new OperationOutcomeError(normalizeOperationOutcome(notFound));
```
mattwiller, line 2211: Does this error being thrown also result in any output?

packages/core/src/client.ts
```diff
@@ -2206,6 +2206,11 @@ export class MedplumClient extends EventTarget {
       return undefined as unknown as T;
     }
 
+    if (response.status === 404) {
```
mattwiller, line 2209: It looks like we might be kind of playing whack-a-mole with error cases here; would it be better to handle anything `response.status` that ISN'T 200/201/202 by outputting an error message?

packages/core/src/client.ts
jamestouri: Good call! I'll add Error:

packages/core/src/client.ts
jamestouri: I could be wrong but I believe the 404 output will result in a notFound

rahul1: @mattwiller to your point, it seems like we do handle all responses >= 400 on line 2222.

@codyebberson   - do you remember why we handle 400's after the `.json()` parse?

packages/core/src/client.ts
rahul1: @mattwiller to your point, it seems like we do handle all responses >= 400 on line 2222.

@codyebberson   - do you remember why we handle 400's after the `.json()` parse?

packages/core/src/client.ts
rahul1: Since we're already returning an `OperationOutcomeError`, do we even need this logging statement?

packages/core/src/client.test.ts
```diff
@@ -1265,6 +1265,22 @@ describe('Client', () => {
     expect(client.getCachedReference(createReference(result.entry?.[0]?.resource as Patient))).toBeDefined();
   });
 
+  test('Search and return 404', async () => {
+    console.error = jest.fn();
+    const fetch = mockFetch(404, {
```
rahul1, line 1270: This isn't testing for the correct thing. In the example the user gave, the response does not have a json representation. Better to do a `mockFetch` that returns a string. 

For example, the following from the original ticket
```
const resp = await fetch('https://api.medplum.com/nonExistingPath/Patient?_count=1')
await resp.text()
```

returns: 
```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /nonExistingPath/Patient</pre>
</body>
</html>
````

File: packages/core/src/client.test.ts [hunk 5]
```diff
@@ -1265,6 +1265,22 @@ describe('Client', () => {
...
+      resourceType: 'Bundle',
+      entry: [{ resource: { resourceType: 'Patient', id: '123' } }],
+    });
+    const client = new MedplumClient({ fetch });
+    try {
+      await client.search('Patient');
```
rahul1, line 1276: More precise to check that the `OperationOutcomeError` is thrown:
https://jestjs.io/docs/tutorial-async

tonylee80 (approved):

packages/core/src/client.ts
codyebberson: Yeah, I don't think this is right.

We parse the body because it often includes more details about the error.  This *should* have been handled by the check below:

```ts
    if (response.status >= 400) {
      throw new OperationOutcomeError(normalizeOperationOutcome(obj));
    }
```

I don't know why that code path didn't solve this particular case?

codyebberson:
@jamestouri - Can you review?  I don't think this is right.

We parse the body because it often includes more details about the error.  This *should* have been handled by the check below:

```ts
    if (response.status >= 400) {
      throw new OperationOutcomeError(normalizeOperationOutcome(obj));
    }
```

I don't know why that code path didn't solve this particular case?

jamestouri:
@codyebberson here is the PR for fix https://github.com/medplum/medplum/pull/1994

packages/core/src/client.ts
Goncharo: @codyebberson

FYI that error handling code is problematic depending on the FHIR server you're using.

For example with HAPI, their error responses are sometimes XML if `response.status >= 400`, in which case normalizing the operation outcome throws an error - since this always expects a JSON response. Then, we're left with a JSON parsing error, and lose the original context of the server response.

We'll raise a separate issue for this, and/or push a PR for the client to fix this.

PR #1992, opened 2023-05-08
Code cleanup for canonical references in search includes

mattwiller:


codyebberson (approved):

codyebberson:
Thanks @mattwiller this is great.

Through this PR I learned that we've been doing `canonical` wrong as well, using it more as an equivalent of `Reference.reference`.

PR #1999, opened 2023-05-09
Example of generating the datetime db migration

codyebberson:
@mattwiller  - Here is an example of how to use the generator project to generate a migration.  As you can see, there is still some manual work involved.

Obviously the ideal state would be that the migration generator builds a full model of "before" and "after" states, and generates a diff automatically.  That is not far off, but would take some work to complete.


packages/generator/src/migrate.ts
```diff
@@ -151,6 +140,19 @@ function buildSearchColumns(resourceType: string): string[] {
     const columnName = details.columnName;
     const newColumnType = getColumnType(details);
     result.push(`"${columnName}" ${newColumnType}`);
+
+    if (searchParam.type === 'date') {
+      const propertyType = details.elementDefinition?.type?.[0].code;
```
codyebberson, line 145: @mattwiller This line is the same as what's used in `getSearchParamDetails`: https://github.com/medplum/medplum/blob/main/packages/core/src/search/details.ts#L82

That's probably wrong, because there are probably `SearchParameter` resources with the `[x]` choice of type that can refer to both `date` and `dateTime`.  In those cases, we should probably also use the `datetime` code path.

With that in mind....  It might be more straightforward to treat all of these as `datetime`.

Issue #1810, opened 2023-04-11
Add HL7 Helper Methods to SDK


rahul1:
Below are two bots for our integration between Ro and their 3rd party lab. There are a lot of helper functions in here that make the code longer than it needs to be. Most of these are parsing/formatting utilities for the HL7 format that could live inside our [core repo](https://github.com/medplum/medplum/blob/main/packages/core/src/hl7.ts). The `uploadMedia` function would also be good to move into the [medplum client](https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts)

https://github.com/medplum/medplum-kit-bots/blob/main/src/read-from-adl.ts
https://github.com/medplum/medplum-kit-bots/blob/main/src/send-to-adl.ts

codyebberson:
@rahul1  - Can you take a quick pass at listing which functions you think would be good candidates for `@medplum/core`?  Then we can assign to engineering to move them over.

PR #2004, opened 2023-05-09
#1810 Add HL7 Helper Methods to SDK

jamestouri:
https://github.com/medplum/medplum/issues/1810

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -3,15 +3,20 @@
 
 import {
   AccessPolicy,
+  Attachment,
   Binary,
   Bundle,
   BundleEntry,
   BundleLink,
+  CodeableConcept,
   Communication,
   Device,
+  DiagnosticReport,
   Encounter,
   ExtractResource,
   Identifier,
+  Media,
+  Observation,
   OperationOutcome,
   Patient,
   Project,
@@ -21,6 +26,7 @@ import {
   Resource,
   ResourceType,
   SearchParameter,
+  ServiceRequest,
   StructureDefinition,
   UserConfiguration,
   ValueSet,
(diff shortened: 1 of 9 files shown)


packages/core/src/utils.ts
```diff
@@ -741,3 +743,76 @@ function toPreciseInteger(a: number, precision?: number): number {
...
+  return n.toString().padStart(2, '0');
+}
+
+/**
+ * Finds the first observation in the input array that matches the specified code and system.
+ * @param observations - The array of observations to search.
+ * @param code - The code to search for.
+ * @param system - The system to search for.
+ * @returns The first observation in the input array that matches the specified code and system, or undefined if no such observation is found.
+ */
```
rahul1, line 788: I think this is a good function to have, but it should be generalized to more resource types than just `Observations`. 
- Any resourceType that has a `code` property of type `CodeableConcept` or `CodableConcept[]`
- Allow `code` param to be a string or `CodeableConcept`.

packages/core/src/client.ts
```diff
@@ -2119,6 +2122,56 @@ export class MedplumClient extends EventTarget {
...
 
+  /**
+   * Create or update an Observation instance based on the provided code and system.
+   *
+   * @param observation provided Observation
+   * @param existingObservations current Observations
+   * @param system URL to request
+   * @returns Promise of the Observation updated or created
+   */
+  createOrUpdateObservation(
```
rahul1, line 2133: I think this is too specific for the client. Remove for now

rahul1 (changes requested):

packages/core/src/utils.ts
```diff
@@ -741,3 +749,22 @@ function toPreciseInteger(a: number, precision?: number): number {
...
 }
+
+/**
+ * Finds the first resource in the input array that matches the specified code and system.
+ * @param resources - The array of resources to search.
+ * @param code - The code to search for.
+ * @param system - The system to search for.
+ * @returns The first resource in the input array that matches the specified code and system, or undefined if no such resource is found.
+ */
+export function findByCode(
```
rahul1, line 760: suggested rename: `findResourceByCode`

packages/core/src/utils.ts
```diff
@@ -32,6 +32,14 @@ export interface InviteResult {
   membership: ProjectMembership;
 }
 
+interface Code {
+  code?: CodeableConcept | string;
```
rahul1, line 36: remove `string`

rahul1 (approved):

PR #2014, opened 2023-05-11
Cancel in-progress CI jobs when new commit is pushed

mattwiller:


.github/workflows/build.yml
```diff
@@ -1,4 +1,9 @@
 name: Build
+# Limit a single job to run at a time for a given branch/PR to save resources and speed up CI
+# see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
+concurrency:
+  group: ci-${{ github.ref_name == 'main' && github.sha || github.ref }}
```
codyebberson, line 5: My YAML string logic is poor... Is this the right way to parse this?

```
if (github.ref_name == 'main') {
  group = 'ci-' + github.sha
} else {
  group = 'ci-' + github.ref
}
```

?

TIL about `ref` vs `ref_name`: https://docs.github.com/en/actions/learn-github-actions/contexts

mattwiller: Yep, that's how it should be parsed: it's taking advantage of the short-circuiting boolean operators to mimic a ternary statement like `github.ref_name == 'main' ? github.sha : github.ref`

codyebberson (approved):

PR #2018, opened 2023-05-11
Fixes #2001 - better model for appshell navbar active link

### LINKED ISSUE #2001: AppShell navbar highlights multiple menu items
Possible fixes:
* Use "last clicked" model, and track which menu item was most recently clicked
* Use "scoring" model to determine which menu item is most appropriate

Edge cases to consider:
* We currently ignore all query search parameters, to support adding filters or moving pages on the search page
* Users may have multiple menu items for the same resource type with preconfigured filters (status=active vs status=completed)

codyebberson:
Fixes the double-highlight bug when there are overlapping links:

image


rahul1 (approved):

tonylee80 (approved):
Looks great!

image

PR #2022, opened 2023-05-11
Fixes Flaky export test

tonylee80:


jamestouri (approved):

packages/server/src/fhir/operations/export.test.ts
```diff
@@ -300,7 +300,7 @@ describe('System export', () => {
 
     // Output format is "ndjson", new line delimited JSON
     const resourceJSON = dataRes.text.trim().split('\n');
-    expect(resourceJSON).toHaveLength(3);
+    expect(resourceJSON.length).toBeGreaterThan(1);
```
tonylee80, line 303: The failed build is showing that the test is getting duplicated resources. It's likely caused by race conditions introduced when creating Observation resources in the test inside `await waitFor`. Updating the test to assert that at least more than 1 Observation resources is created.

codyebberson (approved):

PR #2023, opened 2023-05-11
Fixes #2019 - filled in _filter operators

### LINKED ISSUE #2019: Map  `_filter`-specific operators
FHIR defines an advanced search parameter called `_filter`, that allows for more complicated queries. It's primary value is to be able to specify a mixture of conjunctions (ANDs) and disjunctions (ORs)

Currently, we assume that a the search modifier or search prefix from basic REST search is the same as used in the `_filter` syntax. (e.g. eq, sa, ...). 

However, some modifiers have a special syntax in the `_filter` parameter. We need to map these to our internal `Operator` enum to make sure they're handled correctly

Operation | Definition
-- | --
co | An item in the set contains this value
sw | An item in the set starts with this value
ew | An item in the set ends with this value
ap | A value in the set is approximately the same as this value.Note that the recommended value for the approximation is 10% of the stated value (or for a date, 10% of the gap between now and the date), but systems may choose other values where appropriate
pr | The set is empty or not (value is false or true)
po | True if a (implied) date period in the set overlaps with the implied period in the value
ss | True if the value subsumes a concept in the set
sb | True if the value is subsumed by a concept in the set
ni | True if none of the concepts are in the nominated value set by URI, either a relative, literal or logical vs
re | True if one of the references in set points to the given URL

codyebberson:


packages/core/src/filter/parse.ts
```diff
@@ -1,8 +1,63 @@
...
+  // sb - True if the value is subsumed by a concept in the set
+  sb: undefined,
+  // in - True if one of the concepts is in the nominated value set by URI, either a relative, literal or logical vs
+  in: Operator.IN,
+  // ni - True if none of the concepts are in the nominated value set by URI, either a relative, literal or logical vs
+  ni: Operator.NOT_IN,
+  // re - True if one of the references in set points to the given URL
+  re: undefined,
+  // identifier - True if the identifier is in the identifier set (Medplum extension)
+  identifier: Operator.IDENTIFIER,
```
mattwiller, line 50: This kind of extension might be tricky to dig out later if we need to, but seems like a fine path for now

mattwiller (approved):

packages/core/src/filter/parse.ts
codyebberson: Yeah :(

Issue #2010, opened 2023-05-10
Refresh `Timeline` component after "App" submission


rahul1:
Right now, after a change to resource data, the timeline component will show data from the client-side cache until page refresh, potentially showing stale data. 

This ticket is to update the caching behavior in the Medplum App to update this cache with the most recent data when the user changes tabs to view the Timeline Component

codyebberson:
As discussed offline:
* Need to add a new `invalidateAll()` method to `MedplumClient`
  * Similar to `invalidateUrl()`: https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts#L608-L616
  * Except it flushes everything
* Add a new optional query string parameter to `FormPage.tsx` for "clear cache on success" (name?)
* In the link from `AppsPage.tsx` (https://github.com/medplum/medplum/blob/main/packages/app/src/resource/AppsPage.tsx#L77), add clearCache=true or whatever
* At the end of the `handleSubmit` method (https://github.com/medplum/medplum/blob/main/packages/app/src/FormPage.tsx#L145-L164), if the query string param is set, then clear the cache using the new `invalidateAll()`

PR #2044, opened 2023-05-16
Update cache when navigating to Timeline

jamestouri:
https://github.com/medplum/medplum/issues/2010

https://github.com/medplum/medplum/assets/6913745/2f4ea5c1-5349-4b1a-845c-0450ec4332f8

### DIFF
--- a/packages/app/src/FormPage.test.tsx
+++ b/packages/app/src/FormPage.test.tsx
@@ -44,7 +44,7 @@ describe('FormPage', () => {
   });
 
   test('Patient subject', async () => {
-    await setup('/forms/123?subject=Patient/123');
+    await setup('/forms/123?subject=Patient/123&clearCache=true');
     await waitFor(() => screen.getByText('First question'));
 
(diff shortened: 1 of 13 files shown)


codyebberson (commented):
Hm, interesting.

This works.  It's a lot though, and it adds a burden to future Timeline components and pages.  I wonder if there's a more systematic way to do this.

A potentially brute force way would be to flush the cache here: https://github.com/medplum/medplum/blob/main/packages/app/src/resource/ResourcePage.tsx#L100 (in `handleTabChange`)

I'm also not sure that the issue title is correct... do we actually want to refresh on every tab change?  Or would it be better after other various events (editing a resource, adding a comment, etc)?  Or should the timeline have an automatic refresh on a timer?

Basically I think we should noodle on this one a little bit more.

codyebberson:
I added more thoughts here: https://github.com/medplum/medplum/issues/2010#issuecomment-1550326041

codyebberson (approved):
👍

PR #2055, opened 2023-05-17
Fixes #2054 - Add db indexes to all search params

### LINKED ISSUE #2054: Add database indexes for all on-table search parameters

codyebberson:


mattwiller:
I recently read [this article](https://sirupsen.com/index-merges), which touches on some performance gotchas that we might run into with this strategy: mainly, that index merges are possibly not as well-supported by Postgres as they are by MySQL.  The article was using a `COUNT(*)` query though, which is known to be slower in Postgres, so I'm unsure how much of their findings to attribute to that vs. the query planner

codyebberson:
@mattwiller That is interesting, thanks for sharing.  It is an unfortunate downside of the flexible FHIR search model that it would be difficult to preempt all of the composite indexes.

1. We should continue to track common query patterns, and strategically add composite indexes
2. Considering that most queries are for `compartment` + one or more search params, it's possible that the better model would be that most of these are composite searches for `("compartment", "<search param>")`. :/

codyebberson:
Update: I'm running benchmarks to determine whether this will meaningfully improve performance.

codyebberson:
Query: 

```sql
SELECT 
  "Communication"."id", 
  "Communication"."content" 
FROM 
  "Communication" 
WHERE 
  (
    "Communication"."deleted" = $1 
    AND (
      "Communication"."compartments" IS NOT NULL 
      AND "Communication"."compartments" && ARRAY[$2] :: UUID[]
    )
  ) 
ORDER BY 
  "Communication"."received" DESC 
LIMIT ?
OFFSET ?
```

Before index: 25-35ms

After index: 1-3ms

Unfortunately the `COUNT` query is still slow :(

mattwiller:
That's a great result! 🎉

tonylee80 (approved):

Issue #2047, opened 2023-05-16
App should detect if cannot reach API network


codyebberson:
Use case is publicly hosted "App", trying to reach the API server in private VPC / requiring VPN.

Currently, it would be a network timeout, and might log errors to console.log().  We should make sure there's a user-friendly error message that explains that the API server is not available.

codyebberson:
The trick will be to detect no network connectivity.  That's probably a special error from `fetch`.

When detected, we should show an error with the Mantine Notification system: https://mantine.dev/others/notifications/

We can use `autoClose: false` to keep the notification open.

PR #2067, opened 2023-05-18
Notification in UI for unreachable network 

jamestouri:
https://github.com/medplum/medplum/issues/2047

Screenshot 2023-05-18 at 10 22 12 AM

https://github.com/medplum/medplum/assets/6913745/b0997f53-3fd2-4001-b0a8-745e8fba1f19


packages/core/src/client.ts
```diff
@@ -2259,11 +2259,17 @@ export class MedplumClient extends EventTarget {
...
-      if (response.status < 500) {
-        return response;
+      try {
+        response = (await this.fetch(url, options)) as Response;
+        if (response.status < 500) {
+          return response;
+        }
+        await new Promise((resolve) => setTimeout(resolve, retryDelay));
+      } catch (err: any) {
+        if (err.message === 'Failed to fetch' && retry === 1) {
```
jamestouri, line 2269: This is so that it shows the notification 1 time rather than multiple

packages/core/src/client.ts
```diff
@@ -2628,6 +2633,16 @@ export class MedplumClient extends EventTarget {
       // Silently ignore if this environment does not support storage events
     }
   }
+
+  private retryCatch(retryNumber: number, maxRetries: number, err: Error): void {
+    // This is for the 1st retry to avoid multiple notifications
+    if (retryNumber === 1) {
```
jamestouri, line 2639: This is to show just 1 notification despite the retry

packages/core/src/client.ts
```diff
@@ -2267,11 +2268,15 @@ export class MedplumClient extends EventTarget {
     const retryDelay = 200;
     let response: Response | undefined = undefined;
     for (let retry = 0; retry < maxRetries; retry++) {
-      response = (await this.fetch(url, options)) as Response;
-      if (response.status < 500) {
-        return response;
+      try {
+        response = (await this.fetch(url, options)) as Response;
+        if (response.status < 500) {
```
jamestouri, line 2273: Unless if it's a 500 error, the response will still return, thus the only way for an error message is a network error, which is important because every browser had a different Error message and was difficult to key in the right conditional if we wanted to be more specific

packages/core/src/client.ts
```diff
@@ -2267,11 +2268,15 @@ export class MedplumClient extends EventTarget {
...
+          return response;
+        }
+        await new Promise((resolve) => setTimeout(resolve, retryDelay));
+      } catch (err: any) {
+        this.retryCatch(retry, maxRetries, err);
       }
-      await new Promise((resolve) => setTimeout(resolve, retryDelay));
```
codyebberson, line 2274: This line effectively sleeps for `retryDelay` milliseconds.  I think It belongs here, outside of the `try/catch` rather than inside.

As currently written, the code will not sleep on error.

codyebberson (commented):
See comment on the `retryDelay` being outside of `try/catch`

Otherwise lgtm 👍

packages/core/src/client.test.ts
```diff
@@ -1667,6 +1667,21 @@ describe('Client', () => {
...
 
+  test('Dispatch on bad connection', async () => {
+    const fetch = jest.fn(async () => {
+      throw { message: 'Failed to fetch' };
+    });
+    const mockDispatchEvent = jest.fn();
+    const client = new MedplumClient({ fetch });
+    client.dispatchEvent = mockDispatchEvent;
+    try {
+      await client.readResource('Patient', '123');
```
mattwiller, line 1678: To ensure that this test completely verifies the behavior you expect, you should put a `fail()` assertion on the next line after this one.  That way, if this line ever DOESN'T throw and enter the `catch` block, the test won't silently pass

packages/core/src/client.ts
mattwiller: It looks like `retryNumber` is actually zero the first time this would be called on failure; should that be the condition instead?

packages/core/src/client.ts
jamestouri: @codyebberson We can discuss this offline if needed. I was trying to add the `retryDelay` and it broke a lot of tests. After doing some more investigation, I don't think the this line ever gets called, even before I added the `try/catch`, so I effectively removed it completely. Let me know your thoughts

package-lock.json
```diff
@@ -16510,6 +16510,7 @@
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
+      "dev": true,
```
tonylee80, line 16513: just to confirm, are the `package.json` changes related to this PR?

jamestouri: No, removing that right now

packages/core/src/client.ts
jamestouri: I'm still investigating this, but if I put the condition as 0, it doesn't dispatch, but it dispatches at 1 and 2.

packages/app/src/resource/ResourcePage.test.tsx
```diff
@@ -76,13 +76,8 @@ describe('ResourcePage', () => {
       fireEvent.click(screen.getByText('Delete'));
     });
 
-    try {
```
jamestouri, line 79: This test passed when I added     
```
await act(async () => {
      await jest.runAllTimersAsync();
    });
```

However, I get a large, noisy warning about the mantine notifications not being wrapped in act(). There was no place to put it due to the fact that the warning will show before the mantine notification shows and then there would be an error if I put it above. I believe that this test is mainly to test the UI of the delete button rather than seeing if the delete button actually deletes (since our other tests solve that) so I removed it, but open to any other input.

packages/core/src/client.ts
```diff
@@ -2692,6 +2697,16 @@ export class MedplumClient extends EventTarget {
       // Silently ignore if this environment does not support storage events
     }
   }
+
+  private retryCatch(retryNumber: number, maxRetries: number, err: Error): void {
+    // This is for the 1st retry to avoid multiple notifications
+    if (retryNumber === 1) {
+      this.dispatchEvent({ type: 'offline' });
+    }
+    if (retryNumber === maxRetries - 1) {
```
tonylee80, line 2706: possibly update logic to check `retryNumber >= maxRetries - 1` and a comment that the exception only thrown on the last retry attempt.

tonylee80 (approved):
Minor comment, but looks good to me!

PR #2075, opened 2023-05-19
Use postgres explain hack for fast 'estimate' count

codyebberson:
By default, FHIR search does not include the `total` count.  In the query string params, you can pass in `_total` with value of `none`, `estimate`, or `accurate`.

Before, we treated `estimate` and `accurate` as the same, both using a Postgres `COUNT` query.  Unfortunately, counts are slow in Postgres.

In this PR, we use the Postgres "Count Estimate" hack of inspecting the output of an `EXPLAIN` query.  The technique is described here: https://wiki.postgresql.org/wiki/Count_estimate  It is dramatically faster (typically ~1ms response time vs 20ms+ for an accurate count).

On that page, see the note about `ANALYZE` and "autovacuum".  AWS Aurora Postgres has autovacuuming enabled by default with reasonable (?) defaults: https://aws.amazon.com/blogs/database/understanding-autovacuum-in-amazon-rds-for-postgresql-environments/

### DIFF
--- a/packages/react/src/SearchControl/SearchControl.tsx
+++ b/packages/react/src/SearchControl/SearchControl.tsx
@@ -31,6 +31,7 @@ import {
 import React, { useEffect, useRef, useState } from 'react';
 import { Container } from '../Container/Container';
 import { useMedplum } from '../MedplumProvider/MedplumProvider';
+import { SearchExportDialog } from '../SearchExportDialog/SearchExportDialog';
 import { SearchFieldEditor } from '../SearchFieldEditor/SearchFieldEditor';
 import { SearchFilterEditor } from '../SearchFilterEditor/SearchFilterEditor';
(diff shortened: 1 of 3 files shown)


tonylee80 (approved):

mattwiller:
The defaults are probably reasonable in general, but may not give us tight enough estimates without tuning.  This query should give us all the parameters we could consider tuning and their current values:
```sql
select category, name,setting,unit,source,min_val,max_val from pg_settings where category = 'Autovacuum' ;
```

If we really want the estimates to be reasonably accurate, we may need to use settings like this to force period autovacuum runs:
```
autovacuum_analyze_scale_factor = 0
autovacuum_analyze_threshold = 10000
```
This should make Postgres reanalyze the tables to compute statistics about every 10k row writes, no matter how big the table gets.  Do we have logs around autovacuum currently that we could use to diagnose any issues with it?

PR #2087, opened 2023-05-21
set pathname as default key for ErrorBoundary

Aquilaafuadajo:


Aquilaafuadajo:
Hi @codyebberson, please help take a look at this

codyebberson:
Hi @Aquilaafuadajo  - thanks for submitting this PR!

Unfortunately using `window.location.pathname` in this way does not trigger an update in React.

I pulled your branch and tested it on localhost.  The way that I tested it was adding a `throw new Error('test')` to a random page.  Going to the page causes the error, and brings up the `<ErrorBoundary>` display.  But it does not go away if I click on a link or the back button.

I played with some ideas -- it turns out that this one is deceptively tricky, because it depends on your routing library.  The Medplum `app` uses `react-router-dom`, but that is not a required dependency of `@medplum/react`, so we cannot use it directly.

Instead, what I think you can do is fix it in `packages/react/src/AppShell/AppShell.tsx`.  We pass in props of `pathname` and `searchParams` for similar reasons.  We can use those props here:

```diff
--- a/packages/react/src/AppShell/AppShell.tsx
+++ b/packages/react/src/AppShell/AppShell.tsx
@@ -61,7 +61,7 @@ export function AppShell(props: AppShellProps): JSX.Element {
         ) : undefined
       }
     >
-      <ErrorBoundary>
+      <ErrorBoundary key={`${props.pathname}?${props.searchParams?.toString()}`}>
         <Suspense fallback={<Loading />}>{props.children}</Suspense>
       </ErrorBoundary>
     </MantineAppShell>
```

Can you give that a try?  Please feel free to reach out if you have any questions.

Aquilaafuadajo:
Hi @codyebberson, Thanks for the pointers. I've made the changes just as you suggested and also tested to confirm that it works as intended.

codyebberson (approved):

codyebberson:
Thanks @Aquilaafuadajo !

Issue #2003, opened 2023-05-09
Update Medplum hello world to use router for tab components


codyebberson:
The goal is to standardize on using `react-router-dom` `<Outlet>` component for tab content rather than inline tab logic.  That makes it more extensible, and avoids the overloaded mega pages.

PR #2096, opened 2023-05-22
Add Outlet components to ResourcePage in the examples/medplum-hello-world

jamestouri:
https://github.com/medplum/medplum/issues/2003

https://github.com/medplum/medplum/assets/6913745/d8580179-9c66-429d-88d8-ba64b1afd00f

### DIFF
--- a/examples/medplum-hello-world/src/components/PatientHistory.tsx
+++ b/examples/medplum-hello-world/src/components/PatientHistory.tsx
@@ -6,6 +6,7 @@ export function PatientHistory(): JSX.Element {
   const { patient } = usePatientInfo().data;
   return (
     <Document>
+      <h3>Patient History</h3>
       <ResourceHistoryTable resourceType="Patient" id={patient.id} />
     </Document>
   );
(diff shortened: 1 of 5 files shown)


examples/medplum-hello-world/src/components/PatientHistory.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { usePatientInfo } from '../pages/PatientPage';
+import { Document, ResourceHistoryTable } from '@medplum/react';
+
+export function PatientHistory(): JSX.Element {
+  const { patient } = usePatientInfo().data;
+  return (
+    <Document>
+      <h3>Patient History</h3>
```
jamestouri, line 9: There would be a lot of added complexity with adding an underline over the tab due to the constant re-rendering. I figure it'll be simpler to add a title to each of the new components

codyebberson (commented):
Thanks @jamestouri 

In the video, it looks like there is a very noticeable flicker when changing tabs.  I think that's related to the `createContext`.  I'll send notes offline about how we can avoid that.

codyebberson (approved):

PR #2098, opened 2023-05-23
Hello world suggestions

codyebberson:


examples/medplum-hello-world/src/components/PatientHistory.tsx
```diff
@@ -1,13 +1,16 @@
...
-  const { patient } = usePatientInfo().data;
+  const { id } = useParams();
   return (
     <Document>
-      <h3>Patient History</h3>
-      <ResourceHistoryTable resourceType="Patient" id={patient.id} />
+      <Title order={3} mb="xl">
+        Patient History
+      </Title>
+      <ResourceHistoryTable resourceType="Patient" id={id} />
```
codyebberson, line 13: The `HistoryTable` components intentionally allow to pass in just `resourceType` and `id` so that you don't need to load the resource.  That helps reduce the page flicker.

examples/medplum-hello-world/src/components/PatientOverview.tsx
```diff
@@ -12,63 +12,13 @@ import { formatDate } from '@medplum/core';
...
-                {o.code?.text} ({formatDate(o.meta?.lastUpdated)})
-              </a>
-            </li>
-          ))}
-        </ul>
-      </div>
+      <Title order={3} mb="xl">
+        Patient Overview
+      </Title>
+      <ResourceTable value={{ reference: `Patient/${id}` }} />
```
codyebberson, line 21: This might be controversial, but I think we should avoid GraphQL for the simple "Hello World" demo.

The `<ResourceTable>` component is a quick and easy way to show a resource.

Like many Medplum components, `<ResourceTable>` can accept either a `Resource` or a `Reference`.  The advantage of passing in a `Reference` is that you don't need to preload the resource.  So, in this case, we can just pass in the `id`.

examples/medplum-hello-world/src/pages/PatientPage.tsx
```diff
@@ -92,23 +24,14 @@ export function PatientPage(): JSX.Element {
...
-  }
-  return context;
-}
-
 function TabLink(props: { value: string }): JSX.Element {
-  const capitalizedFirstLetterValue = props.value.charAt(0).toUpperCase() + props.value.slice(1);
   return (
     <Link to={props.value} style={{ textDecoration: 'none' }}>
-      <Tabs.Tab value={props.value}>{capitalizedFirstLetterValue}</Tabs.Tab>
+      <Tabs.Tab value={props.value}>{capitalize(props.value)}</Tabs.Tab>
```
codyebberson, line 34: `@medplum/core` includes a `capitalize()` helper.

reshmakh (approved):

jamestouri (approved):

PR #2099, opened 2023-05-23
Add foomedical to examples

jamestouri:
https://github.com/medplum/medplum/issues/2073


examples/foomedical/.github/workflows/build.yml
```diff
@@ -0,0 +1,21 @@
+name: Build
```
codyebberson, line 1: Do we still want these`.github/workflows` files?

examples/foomedical/sonar-project.properties
```diff
@@ -0,0 +1,13 @@
+sonar.organization=medplum
```
codyebberson, line 1: Not sure if we want this file

codyebberson (approved):
lgtm 👍 

Low pri, but we should figure out our intentions with Github Actions, Sonar, and Coveralls -- whether we want to run those both in the monorepo AND in the separate repo, or if we think that just the monorepo is good enough.

examples/foomedical/.github/workflows/build.yml
jamestouri: Likely not, we should have it build in the main repo, otherwise we'd need to make changes in the listening repo which we are trying to avoid.

PR #2103, opened 2023-05-24
Adding g10 blog content

reshmakh:


codyebberson (approved):
Nice, love it

Missed opportunity to include the best g10 song of all time https://soundcloud.com/medplum/medplum-g-10

Issue #2056, opened 2023-05-17
Refactor validation logic to accept StructureDefinition parameter


mattwiller:
The [validation logic in `@medplum/server`](https://github.com/medplum/medplum/blob/fbbbc390274a5e4e7ec01635d4f8a7679e837f46/packages/core/src/schema.ts#L175) currently depends on the hardcoded [`globalSchema` in `@medplum/core`](https://github.com/medplum/medplum/blob/fbbbc390274a5e4e7ec01635d4f8a7679e837f46/packages/core/src/types.ts#L470), meaning that it only has access to the base FHIR specification and cannot validate resources using arbitrary profiles.  This logic should be refactored to accept a `StructureDefinition` resource as a parameter, to enable easily validating against custom resources and constraining profiles, e.g.:
```ts
function validateResource(resource: Resource, structDef: StructureDefinition): void { /* ... */ }  // throws if resource is invalid
```

The server validation code uses a preprocessed form of the `StructureDefinition` that we call [`IndexedStructureDefinition`](https://github.com/medplum/medplum/blob/fbbbc390274a5e4e7ec01635d4f8a7679e837f46/packages/core/src/types.ts#L87-L133).  This preprocessing currently happens at server startup, but rough benchmarking showed that parsing a single `StructureDefinition` take on the order of just 50 µs — which should be efficient enough to perform on the fly.  The validation code should be able to take a raw `StructureDefinition` resource, process it into the indexed form, and then call the existing validation logic.

Additionally, it may make implementing future validation functionality like slicing easier if there is a utility function that can validate a single value from inside a resource against a list of `ElementDefinition` items related to the field it came from.  Slicing breaks the assumption that each resource field has only one `ElementDefinition` that applies to it: sliced fields or subfields will generally have N+1 associated `ElementDefinition` items in the `StructureDefinition.snapshot`, where N is the number of slices on the field.

For the base resource type specification files, they could either continue to be preprocessed at server startup, with a special case for handling them in their preprocessed form, or we could remove the preprocessing step and instead just load the raw `StructureDefinition` data form the base spec into memory, so it can easily be fed into the function as if it were any other profile.

PR #2119, opened 2023-05-24
Create new parser for StructureDefinitions into schema representation

mattwiller:
Creates a new internal FHIR type schema representation to use when validating profiles, with methods to parse a `StructureDefinition` into the new structure.  This will eventually replace the existing type schema as we build out profile validation.  This PR only includes the parser; subsequent PRs will create a parallel resource validation code path using the new schema (to complete work on #2056), so we can easily transition to the new logic over time.

The new `InternalTypeSchema` is intended to eventually provide a complete internal abstraction for FHIR types, including all necessary information from the `StructureDefinition` to represent the type in programmatic contexts.  The new parsing logic aims to build a complete representation in a single pass over the `StructureDefinition.snapshot.element` array, and includes new data elements related to profile validation and slicing.


packages/core/src/typesnew.ts
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Internal representation of a non-primitive FHIR type, suitable for use in resource validation
+ */
+export interface InternalTypeSchema {
+  name: ResourceType;
+  fields: Record<string, FieldValidator>;
+  constraints: Constraint[];
+}
+
+export interface FieldValidator {
```
codyebberson, line 14: If we're introducing new types, it might be a nice opportunity to be more disciplined on aligning with FHIR naming, so for example calling this `ElementValidator`?

codyebberson (approved):
See comment on `FieldValidator` vs `ElementValidator` - feel free to disregard, it's a minor nit.

For consideration, we may want a separate subdirectory for the new schema validation work.  That would give you a clearly fenced off area to build out the experimental new types and utilities.

For consideration, we may want to add `@experimental` JSDoc annotations.  These are currently not exported in the root `index.ts`, so no way to access them for now, so not a big deal.

Overall, looks good, great start 👍

PR #2131, opened 2023-05-26
Introspection query enabled setting in server config

jamestouri:
https://github.com/medplum/medplum/issues/1937

This PR is the 1st and 2nd step

### DIFF
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -18,6 +18,10 @@ export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
 
 export type FhirRouteHandler = (req: FhirRequest, repo: FhirRepository, router: FhirRouter) => Promise<FhirResponse>;
 
+interface FhirOptions {
+  introspectionEnabled?: boolean;
+}
+
 // Execute batch
 async function batch(req: FhirRequest, repo: FhirRepository, router: FhirRouter): Promise<FhirResponse> {
   const bundle = req.body as Resource;
@@ -107,7 +111,7 @@ async function patchResource(req: FhirRequest, repo: FhirRepository): Promise<Fh
 
 export class FhirRouter {
   readonly router = new Router<FhirRouteHandler>();
-  readonly options: Record<string, any>;
+  readonly options: FhirOptions;
 
   constructor(options = {}) {
     this.options = options;
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -12,7 +12,6 @@ export type FhirRequest = {
   body: any;
   params: Record<string, string>;
   query: Record<string, string>;
-  options?: Record<string, any>;
 };
 
 export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
@@ -108,8 +107,11 @@ async function patchResource(req: FhirRequest, repo: FhirRepository): Promise<Fh
 
 export class FhirRouter {
   readonly router = new Router<FhirRouteHandler>();
+  readonly options: Record<string, any>;
 
-  constructor() {
+  constructor(options = {}) {
+    this.options = options;
+
     this.router.add('POST', '', batch);
     this.router.add('GET', ':resourceType', search);
     this.router.add('POST', ':resourceType/_search', searchByPost);
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -12,7 +12,7 @@ export type FhirRequest = {
   body: any;
   params: Record<string, string>;
   query: Record<string, string>;
-  // options:
+  options?: Record<string, any>;
 };
 
 export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -12,6 +12,7 @@ export type FhirRequest = {
   body: any;
   params: Record<string, string>;
   query: Record<string, string>;
+  // options:
 };
 
 export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
--- a/packages/fhir-router/src/graphql.ts
+++ b/packages/fhir-router/src/graphql.ts
@@ -62,7 +62,7 @@ import {
   validate,
   ValidationContext,
 } from 'graphql';
-import { FhirRequest, FhirResponse } from './fhirrouter';
+import { FhirRequest, FhirResponse, FhirRouter } from './fhirrouter';
 import { FhirRepository } from './repo';
 
 const typeCache: Record<string, GraphQLOutputType | undefined> = {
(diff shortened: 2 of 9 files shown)


packages/graphiql/src/config.ts
```diff
@@ -0,0 +1,11 @@
+export interface MedplumGraphiQLConfig {
+  introspectionUrl: string;
+}
+
+const config: MedplumGraphiQLConfig = {
+  introspectionUrl: process.env.MEDPLUM_INTROSPECTION_URL || '/schema/schema-v4.json',
```
codyebberson, line 6: Prefer to leave the default value `undefined`.  That is a change in behavior, but the current default behavior is broken anyway (because it depends on `schema-v4.json` which isn't in the repo).

packages/graphiql/src/index.tsx
```diff
@@ -10,6 +10,7 @@ import { createRoot } from 'react-dom/client';
 import 'regenerator-runtime/runtime.js';
 
 import 'graphiql/graphiql.css';
+import { getConfig } from './config';
```
codyebberson, line 13: Can you move this import up to the other JS imports?

We group JS imports and CSS imports separately, using a blank line to separate.

packages/server/medplum.config.json
```diff
@@ -21,6 +21,7 @@
   "adminClientId": "2a4b77f2-4d4e-43c6-9b01-330eb5ca772f",
   "maxJsonSize": "1mb",
   "maxUploadSize": "10mb",
+  "introspectionEnabled": false,
```
codyebberson, line 24: We can delete this line, as it's an optional value, and the default is false.

packages/server/src/config.ts
```diff
@@ -201,6 +202,7 @@ function addDefaults(config: MedplumServerConfig): MedplumServerConfig {
   config.awsRegion = config.awsRegion || DEFAULT_AWS_REGION;
   config.botLambdaLayerName = config.botLambdaLayerName || 'medplum-bot-layer';
   config.bcryptHashSalt = config.bcryptHashSalt || 10;
+  config.introspectionEnabled = config.introspectionEnabled ?? false;
```
codyebberson, line 205: We can delete this line.

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -12,6 +12,7 @@ export type FhirRequest = {
   body: any;
   params: Record<string, string>;
   query: Record<string, string>;
+  options?: Record<string, any>;
```
codyebberson, line 15: Can we move `options` to the `FhirRouter` constructor?  That way we don't have to pass it in on every request.

In the future, we could explore having per-request options, but this one makes more sense as a one-time setting.

packages/fhir-router/src/graphql.ts
```diff
@@ -151,7 +151,8 @@ export async function graphqlHandler(req: FhirRequest, repo: FhirRepository): Pr
   }
 
   const introspection = isIntrospectionQuery(query);
-  if (introspection) {
+  const introspectionEnabled = !!req.options?.introspectionEnabled;
+  if (introspection && !introspectionEnabled) {
```
codyebberson, line 155: It'd probably be cleaner to just inline the check to avoid triple negation

`if (introspection && !req.options?.introspectionEnabled) {`

packages/graphiql/src/index.tsx
```diff
@@ -62,7 +63,10 @@ const theme: MantineThemeOverride = {
 
 function fetcher(params: FetcherParams): Promise<SyncExecutionResult> {
   if (params.operationName === 'IntrospectionQuery') {
-    return fetch('/schema/schema-v4.json').then((res) => res.json());
+    const config = getConfig().introspectionUrl;
```
codyebberson, line 66: 👍

packages/server/src/fhir/routes.ts
```diff
@@ -23,8 +23,10 @@ import { sendOutcome } from './outcomes';
 import { Repository } from './repo';
 import { rewriteAttachments, RewriteMode } from './rewrite';
 import { smartConfigurationHandler, smartStylingHandler } from './smart';
+import { getConfig } from '../config';
 
 export const fhirRouter = Router();
+const router = new FhirRouter();
```
codyebberson, line 29: If we move `options` to the constructor, then we can pass in `introspectionEnabled` here.

```ts
const router = new FhirRouter({ introspectionEnabled: getConfig().introspectionEnabled });
```

packages/server/src/fhir/routes.ts
```diff
@@ -155,15 +157,16 @@ protectedRoutes.post(
...
+    const introspectionEnabled = getConfig().introspectionEnabled;
 
+    const repo = res.locals.repo as Repository;
     const request: FhirRequest = {
       method: req.method as HttpMethod,
       pathname: req.originalUrl.replace('/fhir/R4', '').split('?').shift() as string,
       params: req.params,
       query: req.query as Record<string, string>,
       body: req.body,
+      options: { introspectionEnabled: !!introspectionEnabled },
```
codyebberson, line 169: Then we can delete this line.

codyebberson (changes requested):
Nice, thanks @jamestouri 

Lots of little nits, overall structure looks good 👍

packages/graphiql/src/config.ts
codyebberson: That does mean that we'll need to add an env var to Github Secrets and pass it in in `.github/workflows/build.yml`

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -107,8 +107,11 @@ async function patchResource(req: FhirRequest, repo: FhirRepository): Promise<Fh
 
 export class FhirRouter {
   readonly router = new Router<FhirRouteHandler>();
+  readonly options: Record<string, any>;
```
codyebberson, line 110: Let's create an interface with defined options.

packages/server/src/fhir/routes.ts
```diff
@@ -23,9 +23,19 @@ import { sendOutcome } from './outcomes';
 import { Repository } from './repo';
 import { rewriteAttachments, RewriteMode } from './rewrite';
 import { smartConfigurationHandler, smartStylingHandler } from './smart';
+import { getConfig } from '../config';
 
 export const fhirRouter = Router();
 
+// If the config file is not present, assume introspection is disabled
+let config;
```
codyebberson, line 31: Let's discuss this offline to brainstorm ways to avoid this.

packages/server/src/fhir/routes.ts
```diff
@@ -155,9 +158,8 @@ protectedRoutes.post(
 protectedRoutes.use(
   '*',
   asyncWrap(async (req: Request, res: Response) => {
-    const router = new FhirRouter();
+    router = new FhirRouter({ introspectionEnabled: getConfig().introspectionEnabled });
```
codyebberson, line 161: We should either:
1. Go back to `const`, or
2. Do proper lazy load

The lazy initialization is intended to only create a `new FhirRouter` once.  I'm ok to move that to a separate commit though.

Lazy initialization would look like this:

```ts
if (!router) {
  router = new FhirRouter({ introspectionEnabled: getConfig().introspectionEnabled });
}
```

packages/server/src/fhir/routes.ts
```diff
@@ -23,9 +23,12 @@ import { sendOutcome } from './outcomes';
 import { Repository } from './repo';
 import { rewriteAttachments, RewriteMode } from './rewrite';
 import { smartConfigurationHandler, smartStylingHandler } from './smart';
+import { getConfig } from '../config';
 
 export const fhirRouter = Router();
 
+let router: FhirRouter;
```
codyebberson, line 30: If we're going to make this a global for the file, we should give it a more descriptive name (i.e., `internalFhirRouter` or something like that).  Especially when it's sitting right next to `fhirRouter`.

codyebberson (approved):

PR #2142, opened 2023-05-30
Fixes #2108 - support large file bulk import in CLI

### LINKED ISSUE #2108: support large file bulk import in CLI
Related Issue: #2024

tonylee80:
During importing there might be a size limits on API requests. CLI bulk import command now batches 25 resources per requests. There is an option to override the default value where the batch requests are too large for the server to handle.

### Summary
📄🔢🚀

Added a new option to the bulk import command to control the batch size and improved the code structure. The option `--numResourcesPerRequest` lets the user specify how many resources to import per request. The code was refactored into two functions: `importFile` and `sendBatchEntries` to handle reading the input file and sending the batch requests.

> _We are the masters of the batch_
> _We import resources with a switch_
> _`--numResourcesPerRequest` is our power_
> _We send the entries and make them cower_

### Walkthrough
*  Add option `--numResourcesPerRequest` to bulk import command ([link](https://github.com/medplum/medplum/pull/2142/files?diff=unified&w=0#diff-aff1e5d2e9becbf0e02038967adcd4314c1c6588ecca9d825de98d5110a42cafL36-R84))
* Add test case for bulk import command with `--numResourcesPerRequest` set to 2 ([link](https://github.com/medplum/medplum/pull/2142/files?diff=unified&w=0#diff-ba0eec9b49430d3db287c8fc6a316004d00df70f9396382fef7d8a1c513824eeL196-R227))
   * This test uses a mock input file with three patient resources ([link](https://github.com/medplum/medplum/pull/2142/files?diff=unified&w=0#diff-ba0eec9b49430d3db287c8fc6a316004d00df70f9396382fef7d8a1c513824eeL4-R8))

### DIFF
--- a/packages/cli/src/bulk.test.ts
+++ b/packages/cli/src/bulk.test.ts
@@ -1,7 +1,11 @@
 import { MedplumClient, created } from '@medplum/core';
 import { main } from '.';
 
-const testLineOutput = [`{"resourceType":"Patient", "id":"1111111"}`, `{"resourceType":"Patient", "id":"2222222"}`];
+const testLineOutput = [
+  `{"resourceType":"Patient", "id":"1111111"}`,
+  `{"resourceType":"Patient", "id":"2222222"}`,
+  `{"resourceType":"Patient", "id":"3333333"}`,
+];
 jest.mock('child_process');
 jest.mock('http');
 jest.mock('readline', () => ({
@@ -199,7 +203,7 @@ describe('CLI Bulk Commands', () => {
 
     test('success with option numResourcesPerRequest', async () => {
       medplum = new MedplumClient({ fetch });
-      await main(medplum, ['node', 'index.js', 'bulk', 'import', 'Patient.json', '--numResourcesPerRequest', '1']);
+      await main(medplum, ['node', 'index.js', 'bulk', 'import', 'Patient.json', '--numResourcesPerRequest', '2']);
 
       testLineOutput.forEach((line) => {
         const resource = JSON.parse(line);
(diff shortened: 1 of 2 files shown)


packages/cli/src/bulk.ts
```diff
@@ -33,29 +33,52 @@ bulk
 bulk
   .command('import')
   .argument('<filename>', 'File Name')
-  .action(async (fileName) => {
+  .option(
+    '--numResourcesPerRequest <numResourcesPerRequest>',
+    'optional number of resources to import per batch request. Defaults to 25.'
```
jamestouri, line 38: I believe the 3rd argument of option can is the default value and you can set '25' after the description. More on that here 

https://github.com/tj/commander.js/#other-option-types-negatable-boolean-and-booleanvalue

jamestouri (approved):
Looks great! Good stuff on testing this too. Added a comment to just get your thoughts

packages/cli/src/bulk.ts
tonylee80: This is great! It simplifies the logic to check for `undefined`

packages/cli/src/bulk.test.ts
```diff
@@ -193,10 +197,33 @@ describe('CLI Bulk Commands', () => {
...
+                  method: 'POST',
+                  url: resource.resourceType,
+                },
+              })
+            ),
+          })
+        );
+      });
+
+      expect(fetch).toBeCalledTimes(3);
```
tonylee80, line 226: expecting 3 calls for 3 resources where `--numResourcesPerRequest` set to `1`

codyebberson (approved):

PR #2158, opened 2023-06-01
Fixes #2146 - invite user convert email to lower case

### LINKED ISSUE #2146: Capitalized email persisted in User resource

codyebberson:
This is an imperfect solution.  The better solution will be after the new validation code is released, and we will be able to add a constraint to make sure lower case `User.email` never gets into the system.

### DIFF
--- a/packages/server/src/admin/invite.ts
+++ b/packages/server/src/admin/invite.ts
@@ -76,8 +76,12 @@ export interface InviteRequest {
 export async function inviteUser(
   request: InviteRequest
 ): Promise<{ user: User; profile: ProfileResource; membership: ProjectMembership }> {
+  if (request.email) {
+    request.email = request.email.toLowerCase();
+  }
+
   const project = request.project;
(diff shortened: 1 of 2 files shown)


jamestouri (approved):

packages/server/src/admin/invite.ts
```diff
@@ -76,8 +76,12 @@ export interface InviteRequest {
 export async function inviteUser(
   request: InviteRequest
 ): Promise<{ user: User; profile: ProfileResource; membership: ProjectMembership }> {
+  if (request.email) {
+    request.email = request.email.toLowerCase();
+  }
+
```
tonylee80, line 82: Could this logic be added to [createUser](https://github.com/search?q=repo%3Amedplum%2Fmedplum%20createUser&type=code) helper functions?

codyebberson: Nice catch -- let me see if that covers everything.

tonylee80 (approved):

PR #2163, opened 2023-06-02
add GET support for system level export

tonylee80:
As per https://hl7.org/fhir/uv/bulkdata/export/index.html#endpoint---system-level-export, GET should be supported for system level exports.

image

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -2171,7 +2171,7 @@ export class MedplumClient extends EventTarget {
       url.searchParams.set('_since', since);
     }
 
-    options.method = exportLevel ? 'GET' : 'POST';
+    options.method = 'GET';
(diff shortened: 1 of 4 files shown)


mattwiller (approved):

mattwiller:
Thanks for catching this!

packages/core/src/client.ts
```diff
@@ -2171,7 +2171,7 @@ export class MedplumClient extends EventTarget {
       url.searchParams.set('_since', since);
     }
 
-    options.method = exportLevel ? 'GET' : 'POST';
+    options.method = 'GET';
```
codyebberson, line 2174: I think `GET` is the fetch default, so you might be able to delete this line entirely?

codyebberson (approved):
Nice 👍

codyebberson (approved):

Issue #2048, opened 2023-05-16
GraphQL mutations


codyebberson:
The FHIR GraphQL spec has optional support for mutations for "create", "update", and "delete": https://hl7.org/fhir/R4/graphql.html#mutations

At present, we do not support any GraphQL mutations.

mattwiller:
One irksome thing about the mutations is that you have to pass in specific [`input` types](https://graphql.org/learn/schema/#input-types), which are distinct from the regular type.  So, for each resource type, we'll probably have two versions of it in the GraphQL schema: one for queries, and one for mutations

vijaypraju04:
Are there plans to add support for this?

codyebberson:
Completed in https://github.com/medplum/medplum/pull/2169

PR #2169, opened 2023-06-05
GraphQL Mutations 

jamestouri:
Didn't want to add too much noise to this PR, so will do cleanup and code refactor in next one 

https://github.com/medplum/medplum/issues/2048

### DIFF
--- a/packages/fhir-router/src/graphql.ts
+++ b/packages/fhir-router/src/graphql.ts
@@ -64,7 +64,6 @@ import {
 } from 'graphql';
 import { FhirRequest, FhirResponse, FhirRouter } from './fhirrouter';
 import { FhirRepository } from './repo';
-import { Operation } from 'rfc6902';
 
 const typeCache: Record<string, GraphQLOutputType | undefined> = {
   base64Binary: GraphQLString,
@@ -241,6 +240,17 @@ function buildRootSchema(): GraphQLSchema {
       type: graphQLType,
       args: buildUpdateArgs(resourceType),
       resolve: resolveByUpdate,
+    };
+
+    mutationFields[resourceType + 'Delete'] = {
+      type: graphQLType,
+      args: {
+        id: {
+          type: new GraphQLNonNull(GraphQLID),
+          description: resourceType + ' ID',
+        },
+      },
+      resolve: resolveByDelete,
     };
   }
 
@@ -532,6 +542,7 @@ function buildUpdateArgs(resourceType: string): GraphQLFieldConfigArgumentMap {
       description: resourceType + ' ID',
     },
   };
+
   const searchParams = getSearchParameters(resourceType);
   if (searchParams) {
     for (const [code, searchParam] of Object.entries(searchParams)) {
@@ -581,9 +592,9 @@ function buildConnectionType(resourceType: ResourceType, resourceGraphQLType: Gr
   });
 }
 
-// create update and delete
+// create update
 async function resolveByUpdate(
-  source: any,
+  _: any,
   args: Record<string, string>,
   ctx: GraphQLContext,
   info: GraphQLResolveInfo
@@ -592,8 +603,19 @@ async function resolveByUpdate(
   const resourceType = fieldName.substring(0, fieldName.length - 'Update'.length) as ResourceType;
   const updatedResource = { ...args, resourceType };
   const resource = await ctx.repo.updateResource(updatedResource as Resource);
-  // console.log(resource);
   return resource;
+}
+
+async function resolveByDelete(
+  _: any,
+  args: Record<string, string>,
+  ctx: GraphQLContext,
+  info: GraphQLResolveInfo
+): Promise<void> {
+  const fieldName = info.fieldName;
+  const resourceType = fieldName.substring(0, fieldName.length - 'Delete'.length) as ResourceType;
+  await ctx.repo.deleteResource(resourceType, args.id as string);
+  return undefined
 }
 
 /**
--- a/packages/fhir-router/src/graphql.ts
+++ b/packages/fhir-router/src/graphql.ts
@@ -16,6 +16,7 @@ import {
   isLowerCase,
   isResourceTypeSchema,
   LRUCache,
+  MedplumClient,
   normalizeOperationOutcome,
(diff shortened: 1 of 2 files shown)


packages/fhir-router/src/graphql.ts
```diff
@@ -65,7 +69,7 @@ import {
 import { FhirRequest, FhirResponse, FhirRouter } from './fhirrouter';
 import { FhirRepository } from './repo';
 
-const typeCache: Record<string, GraphQLOutputType | undefined> = {
+const typeCache: Record<string, GraphQLObjectType | GraphQLInputType | undefined> = {
```
codyebberson, line 72: You might be able to use `GraphQLScalarType` here rather than `GraphQLObjectType | GraphQLInputType`

packages/fhir-router/src/graphql.ts
```diff
@@ -94,6 +98,14 @@ const typeCache: Record<string, GraphQLOutputType | undefined> = {
   'http://hl7.org/fhirpath/System.Time': GraphQLString,
 };
 
+const outputTypeCache: Record<string, GraphQLOutputType | undefined> = {
+  ...typeCache as Record<string, GraphQLOutputType | undefined>,
```
codyebberson, line 102: If you use `GraphQLScalarType` above, then this cast should not be necessary.

packages/fhir-router/src/graphql.ts
```diff
@@ -94,6 +98,14 @@ const typeCache: Record<string, GraphQLOutputType | undefined> = {
...
+};
+
+const inputTypeCache: Record<string, GraphQLInputType | undefined> = {
+  ...typeCache as Record<string, GraphQLInputType | undefined>,
```
codyebberson, line 106: If you use `GraphQLScalarType` above, then this cast should not be necessary.

packages/fhir-router/src/graphql.ts
```diff
@@ -703,6 +829,79 @@ async function resolveField(source: any, args: any, _ctx: GraphQLContext, info:
...
+ * @returns A Promise that resolves to the created resource, or undefined if the resource could not be found or updated.
+ */
+async function resolveByCreate(
+  _source: any,
+  args: Record<string, any>,
+  ctx: GraphQLContext,
+  info: GraphQLResolveInfo
+): Promise<Resource | undefined> {
+  const fieldName = info.fieldName;
+  const resourceType = fieldName.substring(0, fieldName.length - 'Update'.length) as ResourceType;
```
codyebberson, line 849: `'Update'.length === 6`

`'Create'.length === 6`

It works, but probably not what you meant ;)

packages/fhir-router/src/graphql.ts
```diff
@@ -703,6 +829,79 @@ async function resolveField(source: any, args: any, _ctx: GraphQLContext, info:
...
+  const resourceArgs = args.res;
+  const resourceObject = { ...resourceArgs, resourceType };
```
codyebberson, line 851: Adding `resourceType` here is interesting.  Is that in the spec?

In the REST API, we could arguably do the same thing, which would be convenient for users, but it is not spec compliant.  Per the spec, the user always has to pass `resourceType` in, even though it could be implicit.

I don't see anything in the [mutations](https://hl7.org/fhir/R4/graphql.html#mutations) spec to suggest otherwise, so I think we should probably remove `resourceType` :/

packages/fhir-router/src/graphql.ts
```diff
@@ -703,6 +829,79 @@ async function resolveField(source: any, args: any, _ctx: GraphQLContext, info:
...
+  _source: any,
+  args: Record<string, any>,
+  ctx: GraphQLContext,
+  info: GraphQLResolveInfo
+): Promise<Resource | undefined> {
+  const fieldName = info.fieldName;
+  const resourceType = fieldName.substring(0, fieldName.length - 'Update'.length) as ResourceType;
+  const resourceArgs = args.res;
+  const resourceId = args.id;
+  const updatedResource = { ...resourceArgs, resourceType, id: resourceId };
```
codyebberson, line 878: Similar to above comment in `resolveByCreate`, my interpretation of the spec is that the user is supposed to pass in `resourceType` and `id` on their own.

Unlike `create`, we may want to do a check here that the IDs match.

For example, here are the checks that we do in the REST "update" operation: https://github.com/medplum/medplum/blob/main/packages/fhir-router/src/fhirrouter.ts#L83-L95

```ts
// Update resource
async function updateResource(req: FhirRequest, repo: FhirRepository): Promise<FhirResponse> {
  const { resourceType, id } = req.params;
  const resource = req.body;
  if (resource.resourceType !== resourceType) {
    return [badRequest('Incorrect resource type')];
  }
  if (resource.id !== id) {
    return [badRequest('Incorrect ID')];
  }
  const result = await repo.updateResource(resource);
  return [allOk, result];
}
```

packages/fhir-router/src/graphql.ts
```diff
@@ -703,6 +829,79 @@ async function resolveField(source: any, args: any, _ctx: GraphQLContext, info:
...
+async function resolveByDelete(
+  _source: any,
+  args: Record<string, string>,
+  ctx: GraphQLContext,
+  info: GraphQLResolveInfo
+): Promise<void> {
+  const fieldName = info.fieldName;
+  const resourceType = fieldName.substring(0, fieldName.length - 'Delete'.length) as ResourceType;
+  await ctx.repo.deleteResource(resourceType, args.id as string);
+  return undefined;
```
codyebberson, line 902: You should be able to delete this line, it's effectively the default.

codyebberson (commented):
Nice!  This actually looks pretty clean!  I was worried that it was going to blow up.

Did you by chance check out the resulting size of the Introspection query?

codyebberson:
Woo

```
-rw-r--r--     1 cody  staff  49841834 Jun  7 13:57 schema-v6.json
-rw-r--r--     1 cody  staff   3055367 Jun  7 13:57 schema-v6.json.gz
```

49 MB uncompressed, 3 MB compressed.  That's only slightly larger than current schema.

codyebberson (approved):

Issue #2106, opened 2023-05-24
AuditEvent outcomeDesc should preserve newlines


codyebberson:
The motivating use case is `Bot` logs in the `outcomeDesc` property.  Especially when debugging a `Bot` with a stack trace in the output.

Open question about whether we want to special case `outcomeDesc` or preserve whitespace in all `string` properties.

PR #2177, opened 2023-06-06
Preserve newlines for all strings

jamestouri:
https://github.com/medplum/medplum/issues/2106

I set it so that all strings will preserve and display newlines. Can change to only OutcomeDesc if needed. 

Tested with "error: not found\ncannot find the value of data in result\nmodules by path line: 44"

Without the 'pre-wrap' 
Screenshot 2023-06-06 at 3 31 01 PM

With 'pre-wrap'
Screenshot 2023-06-06 at 3 31 48 PM

### DIFF
--- a/packages/react/src/ResourcePropertyDisplay/ResourcePropertyDisplay.tsx
+++ b/packages/react/src/ResourcePropertyDisplay/ResourcePropertyDisplay.tsx
@@ -59,9 +59,9 @@ export function ResourcePropertyDisplay(props: ResourcePropertyDisplayProps): JS
   switch (propertyType) {
     case PropertyType.boolean:
       return <>{value === undefined ? '' : Boolean(value).toString()}</>;
+    case PropertyType.SystemString:
     case PropertyType.string:
       return <div style={{ whiteSpace: 'pre-wrap' }}>{value}</div>;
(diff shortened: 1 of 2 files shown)


packages/react/src/ResourcePropertyDisplay/ResourcePropertyDisplay.tsx
```diff
@@ -59,12 +59,13 @@ export function ResourcePropertyDisplay(props: ResourcePropertyDisplayProps): JS
   switch (propertyType) {
     case PropertyType.boolean:
       return <>{value === undefined ? '' : Boolean(value).toString()}</>;
+    case PropertyType.string:
```
codyebberson, line 62: Can we include `PropertyType.SystemString` here as well?

codyebberson (commented):
Nice, surgical 👍

codyebberson:
FWIW, I recognize that using `style={{ ... }}` is suboptimal, and it would be better to use classes.  But we already do it in a number of other places throughout `packages/react`, so I'm not one to push the issue.

Maybe when Mantine 7 comes out, and it abandons Emotion CSS-in-JS, that would be a good time to go back and remove all of those?

codyebberson (approved):

PR #2178, opened 2023-06-06
Split custom Medplum search parameter definitions into separate file

mattwiller:


packages/core/src/types.ts
```diff
@@ -250,11 +250,7 @@ export function indexSearchParameterBundle(bundle: Bundle<SearchParameter>): voi
  * @see {@link IndexedStructureDefinition} for more details on indexed StructureDefinitions.
  */
 export function indexSearchParameter(searchParam: SearchParameter): void {
-  if (!searchParam.base) {
```
codyebberson, line 253: Why this change?  If `base` is `undefined`, then the function is a no-op, yeah?

packages/server/src/fhir/structure.ts
```diff
@@ -29,8 +29,6 @@ export function getSearchParameter(resourceType: string, code: string): SearchPa
 }
 
 function buildSearchParameters(): void {
-  const searchParams = readJson('fhir/r4/search-parameters.json') as Bundle<SearchParameter>;
-  for (const entry of searchParams.entry as BundleEntry<SearchParameter>[]) {
-    indexSearchParameter(entry.resource as SearchParameter);
-  }
+  indexSearchParameterBundle(readJson('fhir/r4/search-parameters.json') as Bundle<SearchParameter>);
```
codyebberson, line 32: Nice 👍

packages/server/src/index.ts
```diff
@@ -20,11 +17,6 @@ export async function main(configName: string): Promise<void> {
 
   const config = await loadConfig(configName);
 
-  // Preload the schema
```
codyebberson, line 23: Oy, we were double loading the schema? 🤦‍♂️   Nice catch.

IIRC, we did intentionally load the schema before initializing `express`, to the case where we lazy load the schema inside an HTTP request.  Maybe we were doing that anyway (sigh 🤦‍♂️), but I think the theory still holds?  So might be worth adding `getStructureDefinitions()` here?

codyebberson (commented):
Nice, this is a good change.

If you grep for `readJson('fhir/r4/search-parameters.json')`, there are quite a few places in the code base where we would need to add `search-parameters-medplum.json` too.  Probably not all of the tests, but we should add them to the `packages/generator/src/` files (docs.ts, migrate.ts, searchparams.ts, storybook.ts).

packages/core/src/types.ts
mattwiller: I kept it as a no-op, just streamlined the undefined check so it would instead iterate over an empty array

codyebberson (approved):

reshmakh:
Related:https://github.com/medplum/medplum/discussions/2006

Issue #1931, opened 2023-04-27
` fhirPathCriteria` extension for criteria


rahul1:
Subscriptions are a powerful feature, the the R4 implementation is quite coarse grained. It fires on every create and update operation for for a given resource type. 

The [FHIR R5 `SubscriptionTopic` resource](https://build.fhir.org/subscriptiontopic.html) adds many more fields for fine-grained triggers. The proposal here is to backport the `fhirPathCriteria` field as an extension onto our current `Subscription` resource. 

This this takes a fhirPath expression as a string. The expression should take two input vars, `%previous` and `%next`, and return a  boolean. 

This should unlock many interesting Subscription scenarios: 
- Trigger only on resource create (`%previous === null`)
- Trigger only when a specific field changes
- Trigger when a field enters / leaves a specific value.

PR #2190, opened 2023-06-08
Add global context for eval

jamestouri:
We need to evaluate an expression that takes in two inputs. The first step is to add a global variable `context` 

https://github.com/medplum/medplum/issues/1931

### DIFF
--- a/packages/core/src/fhirlexer/parse.ts
+++ b/packages/core/src/fhirlexer/parse.ts
@@ -1,7 +1,8 @@
 import { TypedValue } from '../types';
 import { Token } from './tokenize';
+
 export interface AtomContext {
-  parent: AtomContext | undefined;
+  parent?: AtomContext;
   variables: Record<string, TypedValue>;
 }
 export interface Atom {
--- a/packages/core/src/fhirlexer/parse.ts
+++ b/packages/core/src/fhirlexer/parse.ts
@@ -1,14 +1,17 @@
 import { TypedValue } from '../types';
 import { Token } from './tokenize';
-
+export interface AtomContext {
+  parent: AtomContext | undefined;
+  variables: Record<string, TypedValue>;
+}
 export interface Atom {
-  eval(context: TypedValue[]): TypedValue[];
+  eval(context: AtomContext, input: TypedValue[]): TypedValue[];
 }
 
 export abstract class PrefixOperatorAtom implements Atom {
   constructor(public readonly operator: string, public readonly child: Atom) {}
 
-  abstract eval(context: TypedValue[]): TypedValue[];
+  abstract eval(context: AtomContext, input: TypedValue[]): TypedValue[];
 
   toString(): string {
     return `${this.operator}(${this.child.toString()})`;
@@ -18,7 +21,7 @@ export abstract class PrefixOperatorAtom implements Atom {
 export abstract class InfixOperatorAtom implements Atom {
   constructor(public readonly operator: string, public readonly left: Atom, public readonly right: Atom) {}
 
-  abstract eval(context: TypedValue[]): TypedValue[];
+  abstract eval(context: AtomContext, input: TypedValue[]): TypedValue[];
 
   toString(): string {
     return `${this.left.toString()} ${this.operator} ${this.right.toString()}`;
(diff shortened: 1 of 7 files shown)


packages/core/src/fhirlexer/parse.ts
```diff
@@ -1,14 +1,17 @@
 import { TypedValue } from '../types';
 import { Token } from './tokenize';
-
+export interface AtomContext {
+  parent: AtomContext | undefined;
```
codyebberson, line 4: Let's change this to `parent?: AtomContext` so it's optional, and `undefined` is implied.

packages/core/src/fhirpath/atoms.test.ts
```diff
@@ -3,11 +3,18 @@ import { Bundle, Observation } from '@medplum/fhirtypes';
...
+import { AtomContext } from '../fhirlexer';
+
+let context: AtomContext;
 
 describe('Atoms', () => {
   beforeAll(() => {
     indexStructureDefinitionBundle(readJson('fhir/r4/profiles-types.json') as Bundle);
     indexStructureDefinitionBundle(readJson('fhir/r4/profiles-resources.json') as Bundle);
+    context = {
+      parent: undefined,
```
codyebberson, line 15: Then you can delete this line.

packages/core/src/fhirpath/parse.ts
```diff
@@ -248,8 +248,19 @@ export function evalFhirPath(expression: string, input: unknown): unknown[] {
...
-export function evalFhirPathTyped(expression: string, input: TypedValue[]): TypedValue[] {
-  return parseFhirPath(expression).eval(input);
+export function evalFhirPathTyped(
+  expression: string,
+  input: TypedValue[],
+  context: AtomContext | undefined = undefined
+): TypedValue[] {
+  if (!context) {
+    context = {
+      parent: undefined,
```
codyebberson, line 261: And this line.

packages/core/src/fhirpath/functions.ts
```diff
@@ -1052,106 +1101,71 @@ export const functions: Record<string, FhirPathFunction> = {
 
   /**
    * Returns true when the input string starts with the given prefix.
-   *
-   * If prefix is the empty string (''), the result is true.
```
codyebberson, line 1056: Curious why we deleted all of these comments?

packages/core/src/fhirpath/functions.test.ts
```diff
@@ -1,13 +1,18 @@
...
-import { Atom } from '../fhirlexer';
+import { Atom, AtomContext } from '../fhirlexer';
 import { PropertyType, TypedValue } from '../types';
 import { createReference, getReferenceString } from '../utils';
 import { LiteralAtom } from './atoms';
 import { FhirPathFunction, functions } from './functions';
 import { booleanToTypedValue, toTypedValue } from './utils';
 
+const context = {
+  parent: undefined,
```
codyebberson, line 10: After `parent` is made option, let's delete this line.

codyebberson (commented):
Woo!  That's a lot!  Thanks for cranking through these.

1. See suggestion about making `AtomContext.parent` optional, which removes a few `parent: undefined` lines
2. See note about deleted comments.

packages/core/src/fhirpath/parse.ts
```diff
@@ -248,8 +248,19 @@ export function evalFhirPath(expression: string, input: unknown): unknown[] {
...
  * @param input The resource or object to evaluate the expression against.
+ * @param context The context to use when evaluating the expression.
  * @returns The result of the FHIRPath expression against the resource or object.
  */
-export function evalFhirPathTyped(expression: string, input: TypedValue[]): TypedValue[] {
-  return parseFhirPath(expression).eval(input);
+export function evalFhirPathTyped(
+  expression: string,
+  input: TypedValue[],
+  context: AtomContext | undefined = undefined
```
mattwiller, line 257: Since this is the public interface, we might want to make it clearer what the user should pass in.  For example, maybe instead of asking them to construct an `AtomContext` (which is a purely internal concept), we could have them pass in a `Record<string, TypedValue>` for the variables

codyebberson: Yeah, that's a good call.  You could go further and say there should be `interface FhirPathContext extends AtomContext` or something.

Also @jamestouri you can use the shorter syntax of `context?: AtomContext` for an optional arg.

mattwiller (approved):

Issue #2148, opened 2023-05-31
Enable permalinks in graphiql


codyebberson:
See https://github.com/graphql/graphiql/issues/961

Aquilaafuadajo:
Hi @codyebberson can you please assign this to me? Already looked at the reference link and I think I understand what needs to be done.

codyebberson:
Hi @Aquilaafuadajo that would be great!

PR #2199, opened 2023-06-09
Enable permalinks in graphiql

Aquilaafuadajo:
#2148 
@codyebberson please help review 👀 

made a short video to help visualise how this works 👇🏽 

https://github.com/medplum/medplum/assets/46582086/3dd63cbb-8290-4dd6-af37-481fd18075e0


codyebberson:
Hi @Aquilaafuadajo - Thanks for submitting this!  I think the main structure all looks good.  I tested it out, and it works as expected! 🎉 

There are some style things that will fail the "lint" step in the build.  If you are using VS Code, I recommend adding the "ESLint" and "Prettier" extensions.  They will help identify some of these issues while developing.  For example:

image

Overall, this looks quite good.  Once we fix up the ESLint/Prettier issues, this is good to merge 👍

packages/graphiql/src/index.tsx
```diff
@@ -61,7 +61,7 @@ const theme: MantineThemeOverride = {
   },
 };
 
-function fetcher(params: FetcherParams): Promise<SyncExecutionResult> {
+async function fetcher(params: FetcherParams): Promise<SyncExecutionResult> {
```
codyebberson, line 64: The `async` keyword should not be necessary here.  You only need to add `async` if you use the `await` keyword in the function body.

packages/graphiql/src/index.tsx
```diff
@@ -88,10 +88,64 @@ const medplumPlugin: GraphiQLPlugin = {
   ),
 };
 
+// Parse the search string to get url parameters.
+const search = window.location.search;
```
codyebberson, line 92: Lines 94-102 can be simplified if you use the `URLSearchParams` class: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams

That way we don't have to do our own string parsing.

For example:

```ts
const searchParams = new URLSearchParams(location.href.search);
const parameters: FetcherParams = {
  query: searchParams.get('query'),
  variables: searchParams.get('variables'),
  operationName : searchParams.get('operationName'),
};
```

packages/graphiql/src/index.tsx
```diff
@@ -88,10 +88,64 @@ const medplumPlugin: GraphiQLPlugin = {
...
+
+function onEditOperationName(newOperationName: string) {
+  parameters.operationName = newOperationName;
+  updateURL();
+}
+
+function updateURL() {
+  const newSearch =
+    '?' +
+    Object.keys(parameters)
```
codyebberson, line 131: Likewise, we can use the URLSearchParams class to help construct the new URL without having to do our own string construction:

```ts
const newSearch = new URLSearchParams();
newSearch.set('query', parameters.query);
newSearch.set('variables', parameters.variables);
newSearch.set('operationName ', parameters.operationName);
history.replaceState(null, '', newSearch.toString());
```

packages/graphiql/src/index.tsx
```diff
@@ -88,10 +88,64 @@ const medplumPlugin: GraphiQLPlugin = {
...
+  history.replaceState(null, null, newSearch);
+}
+
 function App(): JSX.Element {
   const profile = useMedplumProfile();
   return profile ? (
-    <GraphiQL fetcher={fetcher} defaultQuery={HELP_TEXT} plugins={[medplumPlugin]} />
+    <GraphiQL
+      fetcher={fetcher}
+      defaultQuery={HELP_TEXT}
```
codyebberson, line 143: You will need to pass in the parsed values for GraphiQL to use them.  Something like this:

```tsx
      defaultQuery={parameters.query || HELP_TEXT}
      variables={parameters.variables}
      operationName={parameters.operationName || undefined}
```

codyebberson (changes requested):
Thanks @Aquilaafuadajo !  This looks like a good start.

See comments about using `URLSearchParams` to parse and format the URLs.

Also, I recommend adding the ESLint and Prettier plugins to VS Code to help catch some of the linter errors while you are developing:

image

codyebberson (approved):
Nice job.  Thanks @Aquilaafuadajo !

Issue #2194, opened 2023-06-08
Add support for `Subscription` and deleted resources


codyebberson:
This is explicitly *not* part of R4:

https://hl7.org/fhir/R4/subscription.html

> "The consequence of this is that there is no notification when a resource is deleted, or when a resource is updated so that it no longer meets the criteria."

But it is required for some customer use cases, such as global synching to a data warehouse.

R5 does support this: 

https://hl7.org/fhir/subscriptiontopic.html#resource-trigger-delete

TODO:  Figure out what to send in the request body?  Or how to include a reference to the deleted resource?

codyebberson:
R5 sends a `Bundle` with a new `SubscriptionStatus` resource: https://hl7.org/fhir/subscriptionstatus.html

![image](https://github.com/medplum/medplum/assets/749094/6aa9b25b-3336-4cf3-9181-07dce7ff5ae6)

See Smile CDR notes: https://smilecdr.com/docs/subscription/delivery.html#delivering-delete-events

Smile CDR can be configured to send `transaction` bundles with the full operation details: https://smilecdr.com/docs/subscription/delivery.html

codyebberson:
Proposal:

1. Immediate term:
    a. Add 'deleted' to supported interactions
    b. Send empty content body with `X-Medplum-Deleted-Resource` HTTP header
2. Medium term:
    a. Backport FHIR R4B `SubscriptionTopic` and `SubscriptionStatus`
    b. Add additional FHIR extensions to `Subscription` to support sending `SubscriptionStatus`
3. Long term:
    a. Use R5 subscription model

PR #2202, opened 2023-06-09
Support for Subscription on deleted resources 

jamestouri:
https://github.com/medplum/medplum/issues/2194

Cleanup issue here https://github.com/medplum/medplum/issues/2215

### DIFF
--- a/packages/server/src/workers/subscription.test.ts
+++ b/packages/server/src/workers/subscription.test.ts
@@ -244,6 +244,11 @@ describe('Subscription Worker', () => {
 
     // Update should not trigger the subscription
     expect(queue.add).not.toHaveBeenCalled();
+
+    // Delete the patient
+    await repo.deleteResource('Patient', patient.id as string);
+
+    expect(queue.add).not.toHaveBeenCalled();
   });
 
   test('Delete-only subscription', async () => {
@@ -296,7 +301,7 @@ describe('Subscription Worker', () => {
     expect(fetch).toHaveBeenCalledWith(
       url,
       expect.objectContaining({
-        method: 'POST',
+        method: 'DELETE',
         body: stringify({ id: patient.id }),
       })
     );
--- a/packages/server/src/workers/subscription.test.ts
+++ b/packages/server/src/workers/subscription.test.ts
@@ -246,6 +246,62 @@ describe('Subscription Worker', () => {
     expect(queue.add).not.toHaveBeenCalled();
   });
 
+  test('Delete-only subscription', async () => {
+    const url = 'https://example.com/subscription';
+
+    const subscription = await repo.createResource<Subscription>({
+      resourceType: 'Subscription',
+      reason: 'test',
+      status: 'active',
+      criteria: 'Patient',
+      channel: {
+        type: 'rest-hook',
+        endpoint: url,
+      },
+      extension: [
+        {
+          url: 'https://medplum.com/fhir/StructureDefinition/subscription-supported-interaction',
+          valueCode: 'delete',
+        },
+      ],
+    });
+    expect(subscription).toBeDefined();
+
+    // Clear the queue
+    const queue = getSubscriptionQueue() as any;
+    queue.add.mockClear();
+
+    // Create the patient
+    const patient = await repo.createResource<Patient>({
+      resourceType: 'Patient',
+      name: [{ given: ['Alice'], family: 'Smith' }],
+    });
+    expect(patient).toBeDefined();
+
+    // Create should trigger the subscription
+    expect(queue.add).not.toHaveBeenCalled();
+
+    // Update the patient
+    await repo.updateResource({ ...patient, active: true });
+
+    // Update should not trigger the subscription
+    expect(queue.add).not.toHaveBeenCalled();
+
+    // Delete the patient
+    await repo.deleteResource('Patient', patient.id as string);
+
+    expect(queue.add).toHaveBeenCalled();
+    const job = { id: 1, data: queue.add.mock.calls[0][1] } as unknown as Job;
+    await execSubscriptionJob(job);
+    expect(fetch).toHaveBeenCalledWith(
+      url,
+      expect.objectContaining({
+        method: 'POST',
+        body: stringify({ id: patient.id }),
+      })
+    );
+  });
+
   test('Send subscriptions with signature', async () => {
     const url = 'https://example.com/subscription';
     const secret = '0123456789';
--- a/packages/server/src/workers/subscription.test.ts
+++ b/packages/server/src/workers/subscription.test.ts
@@ -302,7 +302,11 @@ describe('Subscription Worker', () => {
       url,
       expect.objectContaining({
         method: 'POST',
-        body: stringify({ resourceType: 'Patient', id: patient.id }),
+        body: '{}',
+        headers: {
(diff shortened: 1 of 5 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -325,16 +329,16 @@ async function sendRestHook(
     logger.debug(`Ignore rest hook missing URL`);
     return;
   }
-
   const startTime = new Date().toISOString();
   const headers = buildRestHookHeaders(subscription, resource);
   const body = stringify(resource);
   let error: Error | undefined = undefined;
+  const fetchMethod = validDeleteInteraction(subscription) ? 'DELETE' : 'POST';
```
codyebberson, line 336: Ooo.... interesting....  Is this in the R5 spec?  Surprising

packages/server/src/workers/utils.ts
```diff
@@ -108,6 +108,17 @@ export async function createAuditEvent(
   });
 }
 
+export function validDeleteInteraction(subscription: Subscription): boolean {
```
codyebberson, line 111: For functions that return booleans, we try to use `is` prefixes.  So either `isValidDeleteInteraction` or just `isDeleteInteraction`

packages/server/src/workers/utils.ts
```diff
@@ -108,6 +108,17 @@ export async function createAuditEvent(
...
+  const supportedInteractionExtension = getExtension(
+    subscription,
+    'https://medplum.com/fhir/StructureDefinition/subscription-supported-interaction'
+  );
+  if (supportedInteractionExtension && supportedInteractionExtension.valueCode === 'delete') {
```
codyebberson, line 116: This could be:

```ts
return supportedInteractionExtension?.valueCode === 'delete';
```

packages/server/src/workers/subscription.ts
jamestouri: I don't see that in R5, I'll change this back

packages/server/src/workers/utils.ts
```diff
@@ -108,6 +108,17 @@ export async function createAuditEvent(
   });
 }
 
+export function isValidDeleteInteraction(subscription: Subscription): boolean {
```
codyebberson, line 111: Can we rename this to `isDeleteInteraction`

packages/server/src/workers/utils.ts
```diff
@@ -108,6 +108,17 @@ export async function createAuditEvent(
...
+  const supportedInteractionExtension = getExtension(
+    subscription,
+    'https://medplum.com/fhir/StructureDefinition/subscription-supported-interaction'
+  );
+  if (supportedInteractionExtension?.valueCode === 'delete') {
```
codyebberson, line 116: Replace with `return supportedInteractionExtension?.valueCode === 'delete';`

packages/server/src/workers/subscription.ts
```diff
@@ -263,34 +263,38 @@ export async function execSubscriptionJob(job: Job<SubscriptionJobData>): Promis
     // If the subscription has been disabled, then stop processing it.
     return;
   }
+  const deletedInteraction = isValidDeleteInteraction(subscription);
```
codyebberson, line 266: Can we call this `isDelete`?  `deleted` is past tense

packages/server/src/workers/subscription.ts
```diff
@@ -263,34 +263,38 @@ export async function execSubscriptionJob(job: Job<SubscriptionJobData>): Promis
...
-  if (job.attemptsMade > 0 && currentVersion.meta?.versionId !== versionId) {
-    // If this is a retry and the resource is not the current version, then stop processing it.
-    return;
   }
 
   try {
-    const resourceVersion = await systemRepo.readVersion(resourceType, id, versionId);
-
+    const resourceVersion = deletedInteraction
+      ? { resourceType, id }
```
codyebberson, line 290: This morning we discussed sending empty string - this will send content.  

Reading through... that might make this more complicated, because both `execBot` and `sendRestHook` require a resource...

This is spiralling.  Let's discuss offline.

packages/server/src/workers/subscription.ts
```diff
@@ -264,6 +264,17 @@ export async function execSubscriptionJob(job: Job<SubscriptionJobData>): Promis
     return;
   }
 
+  // If subscription is a delete interaction, then send the delete request and stop processing it.
+  const isDelete = isDeleteInteraction(subscription);
```
jamestouri, line 268: Instead of wrapping everything, added the check to see if it's an isDelete

packages/server/src/workers/subscription.ts
```diff
@@ -264,6 +264,17 @@ export async function execSubscriptionJob(job: Job<SubscriptionJobData>): Promis
...
+  if (isDelete) {
+    try {
+      await sendDeleteRestHook(job, subscription, { resourceType: resourceType, id: id } as Resource);
```
jamestouri, line 271: We are passing in the resourceType and id for the Header that will be passed

packages/server/src/workers/subscription.ts
```diff
@@ -440,3 +440,65 @@ async function execBot(subscription: Subscription, resource: Resource): Promise<
...
+): Promise<void> {
+  const url = subscription?.channel?.endpoint as string;
+  if (!url) {
+    // This can happen if a user updates the Subscription after the job is created.
+    logger.debug(`Ignore rest hook missing URL`);
+    return;
+  }
+
+  const startTime = new Date().toISOString();
+  const headers = buildRestHookHeaders(subscription, resource) as Record<string, string>;
```
jamestouri, line 463: Casting this removes the error in line 465
```
Element implicitly has an 'any' type because expression of type '"X-Medplum-Deleted-Resource"' can't be used to index type 'HeadersInit'.
```

packages/server/src/workers/subscription.ts
```diff
@@ -440,3 +440,65 @@ async function execBot(subscription: Subscription, resource: Resource): Promise<
...
+
+  headers['X-Medplum-Deleted-Resource'] = `${resource.resourceType}/${resource.id}`;
+
+  try {
+    logger.info('Sending rest hook to: ' + url);
+    logger.debug('Rest hook headers: ' + JSON.stringify(headers, undefined, 2));
+    const response = await fetch(url, { method: 'POST', headers, body: '{}' });
```
jamestouri, line 470: Body is empty object

packages/server/src/workers/subscription.test.ts
```diff
@@ -103,6 +103,14 @@ describe('Subscription Worker', () => {
 
     // Update should also trigger the subscription
     expect(queue.add).toHaveBeenCalled();
+
+    // Clear the queue
+    queue.add.mockClear();
+
+    // Delete the patient
+    await repo.deleteResource('Patient', patient.id as string);
```
jamestouri, line 111: @codyebberson added this line here, as of now the subscription will trigger from every delete resource. I can add to the `matchesCriteria` method an extra conditional that says that `https://medplum.com/fhir/StructureDefinition/subscription-supported-interaction` has to be there && the `valueCode` has to be `delete`

codyebberson (approved):
Nice, thanks @jamestouri !

Issue #2126, opened 2023-05-25
API to test user access


codyebberson:
Checking access policies on resources.  For example, "Can this user edit this patient?"

codyebberson:
P1: Client side test
P2: Server side endpoint

taylorqj:
Posted a question relating to this issue in Discord: https://discord.com/channels/905144809105260605/1116045790088003644

This feature will be very useful to us as we're hiding/showing functionality based on the current logged in user's access level. For now, we're fetching `ProjectMembership` and determining if the user's `AccessPolicy` will allow reading/writing on a specific resource. Interested to see where this goes.

In order to achieve the work-around I noticed that the SDK doesn't offer access to the logged in user's id, which is a requirement to query for `ProjectMembership`. I ended up having to decode the access token and using the subject as that's the only place I could find it. I _think_ it might be possible to grab it from initial login, but we're using Google and end up calling `processCode`, which doesn't return any context of the user object.

 I thought that might be useful information for anyone curious or for a future addition!

reshmakh:
AccessPolicy handling for general case will be supported, but will not be parity with server checks.

PR #2205, opened 2023-06-10
Adds 'userId' to 'security' in /me route

taylorqj:
Exposes the `userId` in the me route. This will allow user's to use their `userId` to query [ProjectMembership ](https://www.medplum.com/docs/api/fhir/medplum/projectmembership)and [User](https://www.medplum.com/docs/api/fhir/medplum/user). 

Not sure if this is the best place to put it, or if it's needed, but I'm unable to find my current user id. #2126

### DIFF
--- a/packages/server/src/auth/me.ts
+++ b/packages/server/src/auth/me.ts
@@ -17,24 +17,15 @@ interface UserSession {
 interface UserSecurity {
   mfaEnrolled: boolean;
   sessions: UserSession[];
+  userId: string | undefined;
 }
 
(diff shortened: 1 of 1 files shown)


codyebberson:
Hi @taylorqj - Thanks for your submission.

This functionality was just recently included in the update to `auth/me`.  You can get the user ID now in `response.membership.user.reference`:

image

If you're using `@medplum/core`, you can use `resolveId()` to extract the ID:

```ts
import { resolveId } from '@medplum/core';
const userId = resolveId(response.membership.user);
```

Apologies for the mixup!  Please let us know if you have any questions.

PR #2228, opened 2023-06-14
fhirPathCriteria match extension

jamestouri:


packages/core/src/fhirpath/atoms.ts
```diff
@@ -54,6 +54,9 @@ export class SymbolAtom implements Atom {
     if (this.name === '$this') {
       return input;
     }
+    if (this.name[0] === '%') {
+      return [context.variables[this.name.slice(1)]];
```
jamestouri, line 58: open to requiring the names to be exact e.g variables = {'%previous': {}, '%current': {}} vs. slicing

packages/core/src/fhirlexer/tokenize.ts
```diff
@@ -48,7 +48,7 @@ export class Tokenizer {
     this.keywords = keywords;
     this.operators = operators;
     this.dateTimeLiterals = !!options?.dateTimeLiterals;
-    this.symbolRegex = options?.symbolRegex ?? /[$\w]/;
+    this.symbolRegex = options?.symbolRegex ?? /[$\w%]/;
```
codyebberson, line 51: 1. Should variables use `SymbolAtom` or should we create a new `VariableAtom`?  No strong opinion on that
2. We should probably update the regex so that the `%` can only be at the beginning

We should make sure that we can handle expressions like `a%b`.  I believe (?) that should translate to `a % b` where `%` is a modulo operator.  But it might parse as a single symbol with value `a%b`?

packages/core/src/fhirlexer/tokenize.ts
```diff
@@ -113,7 +113,7 @@ export class Tokenizer {
       return this.consumeSymbol();
     }
 
-    if (c === '$' && next.match(/\w/)) {
+    if ((c === '$' || c === '%') && next.match(/\w/)) {
```
codyebberson, line 116: If we add a new `VariableAtom`, this would be a separate case.

packages/core/src/fhirpath/atoms.ts
codyebberson: 1. If we add a new `VariableAtom`, this would be handled there.
2. I like the slicing approach, but I'm also open to other opinions
3. What should we do on `undefined variable`?  Throw?

packages/server/src/workers/utils.ts
```diff
@@ -116,6 +116,28 @@ export function isDeleteInteraction(subscription: Subscription): boolean {
...
+    subscription,
+    'https://medplum.com/fhir/StructureDefinition/fhir-path-criteria-expression'
+  );
+  if (!criteria?.valueString) {
+    return true;
+  }
+  const history = await systemRepo.readHistory(currentResource.resourceType, currentResource?.id as string);
+  const previousResource = history.entry?.[1]?.resource as Resource;
+  // If there is no previous resource, then we can't compare and have the criteria automatically pass
+  if (!previousResource) {
```
codyebberson, line 130: The spec says to pass in "empty" when there is no history:

https://hl7.org/fhir/subscriptiontopic.html#resource-trigger-create

> When using fhirPathCriteria during a create, the %previous variable will be set to empty ({}). Authors should ensure that expressions can be evaluated without errors in all applicable situations.

packages/server/src/workers/utils.ts
```diff
@@ -116,6 +116,28 @@ export function isDeleteInteraction(subscription: Subscription): boolean {
...
+    return true;
+  }
+  const evalInput = { current: toTypedValue(currentResource), previous: toTypedValue(previousResource) };
+  const evalValue = evalFhirPathTyped(criteria.valueString, [toTypedValue(currentResource)], evalInput);
+  if (evalValue?.[0].value === true) {
```
codyebberson, line 135: FHIRPath uses a lot of "truthy" behavior.  We export a utility method called `toJsBoolean` which would be good here:

https://github.com/medplum/medplum/blob/main/packages/core/src/fhirpath/utils.ts#L39-L48

```ts
return toJsBoolean(evalValue);
```

Or even:

```ts
return toJsBoolean(evalFhirPathTyped(criteria.valueString, [toTypedValue(currentResource)], evalInput));
```

packages/core/src/fhirlexer/tokenize.ts
mattwiller: I'm in favor of a single `SymbolAtom` that covers this case.  The FHIRPath grammar actually calls these "external constants", so calling our atom "variable" might be slightly misaligned with respect to that

packages/server/src/workers/utils.ts
mattwiller: I think there's some benefit in being strict here: if a user gives us an expression that doesn't evaluate to a boolean, we can't really infer their intent and should consider treating that as an error that's reported to the subscriber

packages/core/src/fhirlexer/tokenize.ts
codyebberson: Actually.... looking at the way that we use `this.symbolRegex`, there's no good way to specify start of string.  That pushes me toward thinking we should add `VariableAtom` and a dedicated parsing handler.

codyebberson: Actually @jamestouri  - Now that I'm looking at the spec, our `$` is also incorrect.

Maybe we can do the simple thing for now as you've done it, and then we can do a fast follow to implement `$` and `%` prefixes properly in a separate commit.

Here is the official grammar: https://github.com/HL7/FHIRPath/blob/master/spec/fhirpath.g4

My reading is that the `$` not only must be at the start of the string, it also is only valid for a select number of keywords:

```
invocation                          // Terms that can be used after the function/member invocation '.'
        : identifier                                            #memberInvocation
        | function                                              #functionInvocation
        | '$this'                                               #thisInvocation
        | '$index'                                              #indexInvocation
        | '$total'                                              #totalInvocation
        ;
```

packages/server/src/workers/utils.ts
codyebberson: Sad that the `fhirPathCriteria` spec isn't more clear here.  I don't see any explicit definition.

The FHIRPath `iif` function includes the language "The criterion expression is expected to evaluate to a Boolean.", which is perhaps the best comparable.

https://hl7.org/fhirpath/#iifcriterion-expression-true-result-collection-otherwise-result-collection-collection

So yeah, @jamestouri , go ahead with @mattwiller 's suggestion and keep it as you had it.

codyebberson (approved):

PR #2233, opened 2023-06-15
Extract and complete package of lint rules

mattwiller:
Extracted the base set of ESLint rules for all `@medplum` packages into its own `@medplum/eslint-config` package, and added many of the rules from the suggested lists of both the base ESLint and TypeScript plugin rulesets


codyebberson (approved):
Nice, this is much appreciated, thanks for rolling up your sleeves 🙏

PR #2251, opened 2023-06-19
FhirPathCriteria Docs

jamestouri:
https://github.com/medplum/medplum/issues/1931

Screenshot 2023-06-19 at 10 34 00 AM

### DIFF
--- a/packages/docs/docs/subscriptions/subscription-extensions.md
+++ b/packages/docs/docs/subscriptions/subscription-extensions.md
@@ -227,7 +227,7 @@ If you use custom success codes, you will need to implement ALL of the HTTP stat
 
 ## fhirPathCritera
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/subscriptions/subscription-extensions.md
```diff
@@ -224,3 +224,32 @@ If you use custom success codes, you will need to implement ALL of the HTTP stat
   ]
 }
 ```
+
+## fhirPathCritera
+
+Medplum offers an extension for triggering subscriptions based on a string expression. This string expression takes in two variables: `%previous` and `%current`. These two variables are the same resource and are compared to one another. The expression should return either `True` or `False`.
```
codyebberson, line 230: `true` and `false` should be lowercase

codyebberson (approved):
Nice!  See minor comment about boolean capitalization, otherwise lgtm

PR #2256, opened 2023-06-19
Fixes #2253 - increase server keep alive timeout

### LINKED ISSUE #2253: Research response_processing_time -1
Roughly 0.005% (~1 / 20,000) of API queries result in HTTP 502.  I believe this is the root cause.

From: https://repost.aws/knowledge-center/elb-alb-troubleshoot-502-errors

> **The target closed the connection with a TCP RST or a TCP FIN while the load balancer had an outstanding request to the target**
> 
> The load balancer receives a request and forwards it to the target. The target receives the request and starts to process it, but closes the connection to the load balancer too early. This usually occurs when the duration of the keep-alive timeout for the target is shorter than the idle timeout value of the load balancer. Make sure that the duration of the keep-alive timeout is greater than the idle timeout value.
> 
> Check the values for the `request_processing_time`, `target_processing_time` and `response_processing_time` fields.
> 
> See the following example access log entry:
> 
> ```
> http 2022-04-15T16:52:50.757968Z app/my-loadbalancer/50dc6c495c0c9188 192.168.131.39:2817 10.0.0.1:80 0.001 4.205 -1 502 - 94 326 "GET http://example.com:80 HTTP/1.1" "curl/7.51.0" - - arn:aws:elasticloadbalancing:us-east-1:123456789012:targetgroup/my-targets/73e2d6bc24d8a067 "Root=1-58337262-36d228ad5d99923122bbe354"
> ```
> 
> Note: In this access log entry, the request_processing_time is 0.001, the target_processing_time is 4.205, and the response_processing_time is -1.

codyebberson:
See #2253 

AWS ALB idle timeout is 60 seconds.  Node.js server timeout must be greater than that to avoid 502 Bad Gateway errors.  This PR increases the default keep alive timeout to 90 seconds.

### DIFF
--- a/packages/server/src/index.test.ts
+++ b/packages/server/src/index.test.ts
@@ -6,7 +6,7 @@ jest.mock('ioredis');
 
 jest.mock('express', () => {
   const original = jest.requireActual('express');
-  const listen = jest.fn();
+  const listen = jest.fn(() => ({}));
   const fn = (): any => {
     const app = original();
     app.listen = listen;
@@ -28,7 +28,7 @@ jest.mock('pg', () => {
       if (sql === 'SELECT "version" FROM "DatabaseMigration"') {
         return { rows: [{ version: 1000000 }] };
       }
-      if (sql === 'SELECT "User"."id", "User"."content" FROM "User" WHERE "deleted"=$1 LIMIT 1') {
+      if (sql === 'SELECT "User"."id", "User"."content" FROM "User" WHERE "User"."deleted"=$1 LIMIT 2') {
         return { rows: [{ id: '1', content: '{}' }] };
       }
       return { rows: [] };
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -30,7 +30,8 @@ export async function main(configName: string): Promise<void> {
   loadDataTypes(resourceTypes);
 
   const app = await initApp(express(), config);
-  app.listen(config.port);
+  const server = app.listen(config.port);
+  server.keepAliveTimeout = config.keepAliveTimeout ?? 90000;
   logger.info('Server started on port', config.port);
 }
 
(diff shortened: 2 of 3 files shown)


packages/server/src/index.test.ts
```diff
@@ -28,7 +28,7 @@ jest.mock('pg', () => {
       if (sql === 'SELECT "version" FROM "DatabaseMigration"') {
         return { rows: [{ version: 1000000 }] };
       }
-      if (sql === 'SELECT "User"."id", "User"."content" FROM "User" WHERE "deleted"=$1 LIMIT 1') {
+      if (sql === 'SELECT "User"."id", "User"."content" FROM "User" WHERE "User"."deleted"=$1 LIMIT 2') {
```
codyebberson, line 31: This is the query to check "is the database seeded".  The existing query was out of date, so the test would unnecessarily re-seed the database.  By fixing this line, the test went from 52 seconds to 3 seconds.

reshmakh (approved):

packages/server/src/index.ts
```diff
@@ -30,7 +30,8 @@ export async function main(configName: string): Promise<void> {
   loadDataTypes(resourceTypes);
 
   const app = await initApp(express(), config);
-  app.listen(config.port);
+  const server = app.listen(config.port);
+  server.keepAliveTimeout = config.keepAliveTimeout ?? 90000;
```
tonylee80, line 34: Do we need to have set `servers.headersTimeout` to be greater than `keepAliveTimeout`?

tonylee80 (approved):

packages/server/src/index.ts
codyebberson: Great question.  My interpretation of the discussion in that ticket is that the headersTimeout thing was fixed.  I'll double check

codyebberson: > yokomotod commented on Jun 21, 2020
> Since https://github.com/nodejs/node/pull/32329 was merged, now I don't need to set headersTimeout bigger than keepAliveTimeout, right ?

> @yokomotod if you're asking for v12, v12.19.0 contains the fix

Issue #2058, opened 2023-05-17
Strict validation of string-backed fields


mattwiller:
Many resource fields are represented in JSON as strings, which must then be converted to the actual underlying system data type.  For example, `dateTime` values are represented like `"2023-05-17T13:58:00-07:00"` and binary data is encoded as [base64](https://datatracker.ietf.org/doc/html/rfc4648#section-4) (e.g. `"SGVsbG8sIHdvcmxkCg=="`).  The FHIR specification includes regular expressions that can be used to validate these strings to ensure they are formatted correctly, with two primary caveats:
1. These regexes are explicitly considered non-normative, and informational only; they do not guarantee that all values they match are valid, nor that they match all valid values
2. The raw regular expressions from the specification are unbounded: they can match at any point in the string.  This means that a value like `"1/(2022-12-10) = 0.0005"` would be accepted as a valid date, when it isn't

These are described in the specification in specific detail:
> Issues with the specified regexes:
> The regexes are provided to assist with tooling, but are informative, not normative. There are several issues with the regexes.
> The string regex has problems with Unicode - specifically, it might or might not allow Unicode whitespace to some degree depending on Unicode support in the regex engine being used. The regexes [\r\n\t\x{0020}-\x{FFFF}]* or [\r\n\t\u0020-\uFFFF]* are better expressions of the constraints on string, but poorly supported (see [Regex Tutorial ](https://www.regular-expressions.info/unicode.html)icon for details). The string regex also applies to markdown as well. The regex does not enforce the length limit
> The Unicode issues also apply to the regex for code.
> The regexes should be qualified with start of string and end of string anchors based on the regex implementation used (e.g. caret '^' and dollar-sign '$' for JavaScript, POSIX, XML and XPath; '\A' and '\Z' for .NET, Java, Python and others; please verify these definitions with the regex implementation used).
> The regexes may allow a broader set of values than are actually valid (e.g. leap years) so additional validation is always needed.

The [`validateString()`](https://github.com/medplum/medplum/blob/fbbbc390274a5e4e7ec01635d4f8a7679e837f46/packages/core/src/schema.ts#L294) function in `@medplum/core` should always bound any regex drawn from the spec: `new RegExp('^' + regex + '$'))` to ensure that the entire field matches the regex.

[...]

PR #2266, opened 2023-06-20
Stricter validation for Date and DateTime values

jamestouri:
Part of https://github.com/medplum/medplum/issues/2058

### DIFF
--- a/packages/core/src/typeschema/validation.ts
+++ b/packages/core/src/typeschema/validation.ts
@@ -225,7 +225,13 @@ class ResourceValidator {
       return;
     }
 
-    const regex = validationRegexes[type];
+    if (type === 'date' || type === 'dateTime') {
+      if (isNaN(Date.parse(str))) {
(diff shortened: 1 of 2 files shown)


packages/core/src/typeschema/validation.ts
```diff
@@ -225,9 +225,17 @@ class ResourceValidator {
       return;
     }
 
+    const structureIssue = createStructureIssue(path, 'Invalid ' + type + ' format');
```
codyebberson, line 228: There's a tradeoff here between "duplicate code" and unnecessary allocations.  This will always execute the code path of allocating memory for the issue even if it's not needed.  I'd suggest that we remove this line, and tolerate the duplicate `this.issues.push(createStructureIssue(path, 'Invalid ' + type + ' format'));` to avoid unnecessary execution.

codyebberson (commented):
Nice, this is great.  Thanks @jamestouri

tonylee80 (approved):

PR #2268, opened 2023-06-21
Dependency upgrades

codyebberson:
Notes:
- Holding back `node-fetch` for CJS support
- Holding back `prism-react-renderer` for Docusaurus compatibility
- Finally upgraded Mantine, because I found the source of the test failure (see comments)


packages/react/src/CodeableConceptInput/CodeableConceptInput.test.tsx
```diff
@@ -93,16 +93,8 @@ describe('CodeableConceptInput', () => {
...
-    for (let i = 0; i < 3; i++) {
-      await act(async () => {
-        fireEvent.keyDown(input, { key: 'ArrowDown', code: 'ArrowDown' });
-      });
-    }
-
-    // Press "Enter"
     await act(async () => {
-      fireEvent.keyDown(input, { key: 'Enter', code: 'Enter' });
+      fireEvent.mouseDown(screen.getByText('+ Create XYZ'));
```
codyebberson, line 97: 1. `fireEvent.click()` does not work on the Mantine `<SelectItem>` component.  I don't know if that's a Mantine issue or a JSDOM issue.  
2. I assume that I worked around that before by emulating down arrows and "Enter"
3. There was a change in Mantine where you cannot navigate to the "+ Create" option with keyboard controls (???)
4. I dug into the Mantine source code and found that `fireEvent.mouseDown()` works, so that's what we're doing.

tonylee80 (approved):
Tested resource search, looks good to me!

image


image

PR #2278, opened 2023-06-22
Suppress console errors on cancelled fetch

### LINKED ISSUE #1595: Suppress console errors on cancelled fetch
When using `<AsyncAutocomplete>` (i..e, `<CodeInput>`, `<CodingInput>`, `<CodeableConceptInput>`, etc)...

If a user enters input after a `fetch` has been issued, the active `fetch` call is aborted.  That is intended behavior.

Unfortunately, it currently pollutes the browser console:

image

This should be a matter of adding the appropriate `catch` clause and silently ignoring abort errors.

Paul-Taiwo:
Fixes #1595


codyebberson (approved):
Awesome!  Thanks @Paul-Taiwo!

PR #2279, opened 2023-06-22
Medplum sdk enhancements - POC

tonylee80:


packages/cli/src/auth.ts
```diff
@@ -13,6 +13,7 @@ export const whoami = createMedplumCommand('whoami');
 
 login.action(async (options) => {
   const medplum = await createMedplumClient(options);
+  medplum.clearActiveLogin();
```
tonylee80, line 16: Should we clear existing login if user wants to login with new credentials?

packages/core/src/client.ts
```diff
@@ -588,6 +597,9 @@ export class MedplumClient extends EventTarget {
       this.autoBatchQueue = undefined;
     }
 
+    // if (this.authType === 'basic') {
```
tonylee80, line 600: if CLI user switches to basic auth options, should we clear existing login so that there is no overlap between authentication?

packages/core/src/client.ts
```diff
@@ -2735,7 +2747,7 @@ export class MedplumClient extends EventTarget {
    * Starts a new OAuth2 client credentials flow.
    *
    * ```typescript
-   * await medplum.startClientLogin(process.env.MEDPLUM_CLIENT_ID, process.env.MEDPLUM_CLIENT_SECRET)
+   * await medplum.startClientLogin(clientId, clientSecret)
```
tonylee80, line 2750: updating documentation to direct users away from environment variables.

PR #2286, opened 2023-06-22
Move lint to separate GitHub Action

codyebberson:
This will run ESLint in parallel to the build for faster feedback, and faster overall build.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,30 @@
+name: Lint
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


tonylee80 (approved):

codyebberson:
In retrospect, this won't work.

The TypeScript rules depend on types in other packages.  If TypeScript cannot find a type definition, it swaps in `any`

Consider this lint error: https://github.com/medplum/medplum/actions/runs/5350988333/jobs/9704273907

```
/home/runner/work/medplum/medplum/packages/app/src/HomePage.tsx
Error:   215:47  error  'any' overrides all other types in this union type  @typescript-eslint/no-redundant-type-constituents
```

This is the actual line:

```ts
function getLastSearch(resourceType: string): SearchRequest | undefined {
```

The problem is that core has not been built, and therefore TypeScript treats SearchRequest as any.

I'm going to abort that PR.

PR #2287, opened 2023-06-22
Added Github Action for publish

codyebberson:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish
+# Based on: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+on:
+  push:
+    tags:
+      - 'v*'
```
codyebberson, line 6: This runs when we add a new git tag in the format `v*`.  This appears to be a somewhat common convention, and maps nicely to how we tag releases.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,37 @@
...
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        run: npm ci
+      - name: Build
+        run: npm run build
+        env:
+          KEEP_ENV_VARS: 'true'
+      - name: Publish
+        run: ./scripts/publish.sh
```
codyebberson, line 35: Uses existing publish.sh script which is pretty battle tested.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,37 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
codyebberson, line 37: `NPM_TOKEN` secret is already configured.

tonylee80 (approved):

Issue #2059, opened 2023-05-17
Validate fixed values in profiles


mattwiller:
Resource profiles often use fixed values to specialize resources: a given field is only permitted to have some certain value.  For example, the US Core Blood Pressure profile specializes Observation resources by requiring that the `code` is exactly LOINC code 85354-9 ("Blood pressure panel with all children optional").  These validation rules are sourced from the `fixed[x]` and `pattern[x]` fields on a field's `ElementDefinition`.  Both fields provide values that the field is expected to match, with slight differences in how the matching is performed: `fixed[x]` requires an exact match with no extraneous fields, whereas `pattern[x]` must match on the specified fields but may contain others.

When [validating a resource](https://github.com/medplum/medplum/blob/fbbbc390274a5e4e7ec01635d4f8a7679e837f46/packages/core/src/schema.ts#L229) using a profile, the `fixed[x]` and `pattern[x]` fields within the profile's `StructureDefinition` should be considered and matched against the resource.  We may want to consider using a library for the value matching logic, since it can be a bit subtle in JS: [lodash](https://lodash.com/) has [`_.isEqual()`](https://lodash.com/docs/4.17.15#isEqual) and [`_.isMatch()`](https://lodash.com/docs/4.17.15#isMatch) methods that appear to do what we would need

mattwiller:
I think that validating terminology bindings is probably orthogonal to this one, since fixed/pattern rules are just doing exact matching against the specified values.  For example, consider a rule that fixes the coding system for a `CodeableConcept` field:
```json
"patternCodeableConcept": {
  "coding": [
    { "system": "http://loinc.org" }
  ]
}
```
This doesn't set the binding for the field (which is handled by `ElementDefinition.binding`); instead, it simply requires there to be exactly one code in the field which has `system` set to that exact string value.  I think validating the terminology bindings additionally would be valuable, but this is probably only tractable for the base specification bindings at the moment.  Even then, not all of the value sets referenced by the base spec have a finite expansion; http://hl7.org/fhir/ValueSet/mimetypes is a notable exception.  For profiles in general, I think we might need a full terminology service in order to expand out bindings which may be specifying part or all of the massive coding systems like SNOMED

PR #2288, opened 2023-06-23
Validating Fixed and Pattern Values

jamestouri:
https://github.com/medplum/medplum/issues/2059

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -449,34 +449,33 @@ function deepEqualsObject(
 /**
  * Checks if object2 includes all fields and values of object1.
  * It doesn't matter if object2 has extra fields.
- * @param object1 The first object.
- * @param object2 The second object.
- * @returns True if object2 includes all fields and values of object1.
+ * @param value The object to test if contained in pattern.
+ * @param pattern The object to test against.
+ * @returns True if pattern includes all fields and values of value.
  */
-export function deepIncludes(object1: unknown, object2: unknown): boolean {
-  if (isEmpty(object1)) {
+export function deepIncludes(value: any, pattern: any): boolean {
+  if (isEmpty(value)) {
     return true;
   }
-  if (isEmpty(object2)) {
+  if (isEmpty(pattern)) {
     return false;
   }
-  if (Array.isArray(object1) && Array.isArray(object2)) {
-    return deepIncludesArray(object1, object2);
+  if (Array.isArray(value) && Array.isArray(pattern)) {
+    return deepIncludesArray(value, pattern);
   }
-  if (Array.isArray(object1) || Array.isArray(object2)) {
+  if (Array.isArray(value) || Array.isArray(pattern)) {
     return false;
   }
-  if (isObject(object1) && isObject(object2)) {
-    return deepIncludesObject(object1, object2);
-  }
-  if (isObject(object1) || isObject(object2)) {
+  if (isObject(value) && isObject(pattern)) {
+    return deepIncludesObject(value, pattern);
+  } else if (isObject(value) || isObject(pattern)) {
     return false;
   }
-  return object1 === object2;
+  return value === pattern;
 }
 
 function deepIncludesArray(array1: any[], array2: any[]): boolean {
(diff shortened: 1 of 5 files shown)


packages/core/src/utils.ts
```diff
@@ -446,6 +446,43 @@ function deepEqualsObject(
...
+/**
+ * Checks if object2 includes all fields and values of object1.
+ * It doesn't matter if object2 has extra fields.
+ * @param object1 The first object.
+ * @param object2 The second object.
+ * @returns True if object2 includes all fields and values of object1.
+ */
+export function deepIncludes(object1: unknown, object2: unknown): boolean {
+  if (isEmpty(object1)) {
+    return true;
```
jamestouri, line 458: If object1 is empty, will still return true. Can change this

mattwiller: That seems like the right behavior to me!

packages/core/src/utils.ts
```diff
@@ -446,6 +446,43 @@ function deepEqualsObject(
...
+  }
+  if (isEmpty(object2)) {
+    return false;
+  }
+  if (Array.isArray(object1) && Array.isArray(object2)) {
```
mattwiller, line 463: Minor style nit, but you could compact this whole block of code and reduce visual noise by making the set of conditions one big `if / else if`

packages/core/src/utils.ts
```diff
@@ -446,6 +446,43 @@ function deepEqualsObject(
...
+    return deepIncludesObject(object1, object2);
+  }
+  if (isObject(object1) || isObject(object2)) {
+    return false;
+  }
+  return object1 === object2;
+}
+
+function deepIncludesArray(array1: any[], array2: any[]): boolean {
+  return array1.every((value, index) => deepIncludes(value, array2[index]));
```
mattwiller, line 479: This is going to require that the array elements being in the same order, which [the FHIR spec does not require](https://hl7.org/fhir/elementdefinition-examples.html#pattern-array).  Instead, this should just verify that every value from `array1` is in `array2`, e.g. `array1.every((value) => array2.includes(value))`

packages/core/src/utils.ts
```diff
@@ -446,6 +446,43 @@ function deepEqualsObject(
...
 }
 
+/**
+ * Checks if object2 includes all fields and values of object1.
+ * It doesn't matter if object2 has extra fields.
+ * @param object1 The first object.
+ * @param object2 The second object.
+ * @returns True if object2 includes all fields and values of object1.
+ */
+export function deepIncludes(object1: unknown, object2: unknown): boolean {
```
mattwiller, line 456: Small nit on the naming: I usually try to match any verb in the name of a binary function to the relation between its arguments in order.  For example, this function signature says to me "object1 deepIncludes object2", when in reality it's testing the reverse.  One way around this would be to more clearly label the arguments: `object1` could be called `pattern` and `object2` just `value` or something like that.  What do you think of something like `deepMatches(value: any, pattern: any)`?

packages/core/src/utils.ts
jamestouri: The .includes will return false even if the objects are the same. I ended up calling deepEquals for this as a check

packages/core/src/utils.ts
```diff
@@ -449,34 +449,33 @@ function deepEqualsObject(
...
+  } else if (isObject(value) || isObject(pattern)) {
     return false;
   }
-  return object1 === object2;
+  return value === pattern;
 }
 
 function deepIncludesArray(array1: any[], array2: any[]): boolean {
-  return array1.every((value, index) => deepIncludes(value, array2[index]));
+  return array1.every((value1) => array2.some((value2) => deepEquals(value1, value2)));
```
mattwiller, line 478: Should this check use `deepIncludes()` instead of `deepEquals()`?

jamestouri: I was overthinking it, but when the suggestion for the .includes came, I was telling myself it needed to be an exact resemblance if it's inside an array. Don't know why, but I'll change that back

mattwiller (approved):
Awesome work, this will be super useful in profile validation!

PR #2290, opened 2023-06-23
Add non-null requirements to values in GraphQL array types

jamestouri:
In response to requiring Arrays to be non-null values, we will also require the elements in an array to be non-null as well


codyebberson (commented):
Love it.

Should we do the same thing for input types?

mattwiller (approved):

PR #2298, opened 2023-06-24
ESBuild

codyebberson:
Replaces Rollup and Webpack with ESBuild and Vite.

Deployed to staging: https://app.staging.medplum.com/

### What is ESBuild?

esbuild is a CLI, NPM package, and Go module that makes bundling JavaScript accessible and fast.

This PR uses ESbuild for building **libraries** (`@medplum/core`, `@medplum/react`, etc).

### What is Vite?

Vite is a build tool that aims to provide a faster and leaner development experience for modern web projects. 

This PR uses Vite for building **apps** (`@medplum/app`, `@medplum/graphiql`)

Note that we already use Vite in most of our example applications (`medplum-hello-world`, `foomedical`).

### Why make this change?

There are many reasons to adopt ESBuild and Vite.  The main two:

1. Performance - these tools are much faster than predecessors because (1) they're written with faster tools such as Go and (2) they don't have as much legacy baggage such as support for outdated versions of JS.
2. Modern alignment - these tools are more aligned with the future of JS/TS, such as better support for ESM.

We have actually tried to make this change many times in the past.  However, it has always got stuck for some reason or another (bugs, incompatibilities, dependency conflicts, etc).  This time it looks like it's going to work.

### Who uses these tools?

Notably, the TypeScript team themselves started using ESBuild for building TypeScript: https://github.com/microsoft/TypeScript/pull/51387

Vite is used by many organizations such as [...]

### Build time

Method:
- Run "time npm run build"
- Take the best of 3
- Calculate [speedup](https://en.wikipedia.org/wiki/Speedup) as T<sub>old</sub> / T<sub>new</sub>

image

2x app speedup to 8x react speedup!

### Package size

Method:
- Build
- Create tar of files shipped to browser

image

Package size is slightly larger (1-2%)

### Dev server start time

This is difficult to measure objectively.

Using `app`:

Webpack reports start time of 15.513 seconds:

image

Vite reports start time of 0.562 seconds:

image

### Foo Medical

Before:

image

Main index.js is 188.21 kB gzipped

After:

image

Main index.js is 208.70 kB gzipped

So 11% bigger.  Not great.  That might warrant some investigation.


reshmakh (approved):

reshmakh:
Glad to see the window opening for this change!  I recall it getting stuck 6 months ago and am surprised/pleased that the window opened up now.

mattwiller (approved):

PR #2303, opened 2023-06-26
Adding documentation and examples for CLI optional params to connect …

reshmakh:
…to other fhir servers

### DIFF
--- a/packages/docs/docs/cli/external-fhir-servers.md
+++ b/packages/docs/docs/cli/external-fhir-servers.md
@@ -1,0 +1,61 @@
+---
+sidebar_position: 2
+tags: [integration]
+---
+
+# Connecting to External FHIR Servers
+
+:::caution
+CLI connections to external FHIR servers is in Beta. Please join us on [Discord](https://discord.gg/medplum) with questions or email [hello@medplum.com](mailto:hello@medplum.com).
+:::
+
+When building an application, you many need to query or write data to an external FHIR server as part of your application's workflow. For example:
+
+- When a new patient is created, see if that patient already exists in another server
+- When an encounter is finalized, synchronize the encounter documentation to another FHIR server
+- On a bi-weekly basis, pull new FHIR data from a partner using a Bulk FHIR API
+
+To enable these scenarios, you will need a `clientId` or `clientSecret` to access the system you want to connect to. Please note that different systems have different levels of functionality, and so the commands in the CLI are not guaranteed to work.
+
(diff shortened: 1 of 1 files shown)


tonylee80 (approved):
Looks great!

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -0,0 +1,61 @@
...
+CLI connections to external FHIR servers is in Beta. Please join us on [Discord](https://discord.gg/medplum) with questions or email [hello@medplum.com](mailto:hello@medplum.com).
+:::
+
+When building an application, you many need to query or write data to an external FHIR server as part of your application's workflow. For example:
+
+- When a new patient is created, see if that patient already exists in another server
+- When an encounter is finalized, synchronize the encounter documentation to another FHIR server
+- On a bi-weekly basis, pull new FHIR data from a partner using a Bulk FHIR API
+
+To enable these scenarios, you will need a `clientId` or `clientSecret` to access the system you want to connect to. Please note that different systems have different levels of functionality, and so the commands in the CLI are not guaranteed to work.
```
codyebberson, line 18: This is not strictly true.  That is just one auth pattern.  You could say something like "The most straightforward authentication option is the OAuth client credentials flow using `clientId` and `clientSecret`."  or something like that

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -0,0 +1,61 @@
...
+
+The examples below use the [CLI optional flags](/docs/cli#optional-flags).
+
+## Example: Basic search
+
+In this example, we will show how to search for a patient by identifier using the command line.
+
+```bash
+medplum get 'Patient?identifier:contains=3SH0A00AA00' --base-url <base-url> --fhir-url-path <fhir-url-path> --token-url <token-url> --client-id <client-id> --client-secret <client-secret>
```
codyebberson, line 27: Hmmm... This is an unflattering example and quite verbose.

We should discuss offline.

1. As discussed previously, I think we should try to push this into a "profile" concept, so you only have to specify it once.
2. Many of these are optional and only required when the external FHIR server uses weird URL patterns.  It's great that we have them, but I don't think this is the right place to list everything.  That belongs more in a "Reference" section.

codyebberson (commented):
Let's discuss offline.  We may want to hold on this until some of the auth cleanup work is done.

PR #2305, opened 2023-06-26
Allow extra URL path content with Bot execute

codyebberson:
One of our integration partners appends paths to the webhook callback URL.  So, for example, if you create a subscription with the callback URL `https://example.com/Bot/123/$execute`, the actual requested URL could be `https://example.com/Bot/123/$execute/RequestGroup`.

This PR allows extra content after the `$execute` portion of the URL.


mattwiller (approved):
Could you add a comment explaining why we allow the wildcard paths in the router?

PR #2080, opened 2023-05-19
Fix export flaky test issue with _since params

tonylee80:
Address flaky export test issue:
https://github.com/medplum/medplum/actions/runs/5026350802/jobs/9014520800

Issue #1959, opened 2023-05-02
Export test intermittently fails


codyebberson:
I added a console.log to help debug:

```ts
    // Get the export content
    const outputLocation = new URL(output[0]?.url as string);
    const dataRes = await request(app)
      .get(outputLocation.pathname + outputLocation.search)
      .set('Authorization', 'Bearer ' + accessToken);
    expect(dataRes.status).toBe(200);
    console.log(dataRes.text);
```

```bash
$ npm t -- src/fhir/operations/export.test.ts

> @medplum/server@2.0.17 test
> jest --runInBand src/fhir/operations/export.test.ts

  console.log
    {"resourceType":"Observation","status":"final","code":{"text":"test"},"subject":{"reference":"Patient/59e24c76-2312-4a1a-bd41-bca686b86e12"},"id":"7082c6b0-aa19-4f51-b0e4-26481f5de22b","meta":{"versionId":"9acef927-7cd5-488c-816e-737242eae719","lastUpdated":"2023-05-02T00:11:50.837Z","author":{"reference":"ClientApplication/7d8f4d7f-7d8a-4eb2-9652-3e78b9943a07","display":"Test Client Application"},"project":"9edeb81d-1001-44e9-8035-21c2ca01fa16","compartment":[{"reference":"Project/9edeb81d-1001-44e9-8035-21c2ca01fa16"},{"reference":"Patient/59e24c76-2312-4a1a-bd41-bca686b86e12"}]}}

      at log (src/fhir/operations/export.test.ts:182:13)

 FAIL  src/fhir/operations/export.test.ts
  System export
    √ Success (2226 ms)
    × Parameters (624 ms)

  ● System export › Parameters

    expect(received).toEqual(expected) // deep equality

    Expected: "test2"
    Received: "test"

      186 |     const resourceJSON = dataRes.text.trim().split('\n');
      187 |     expect(resourceJSON).toHaveLength(1);
    > 188 |     expect(JSON.parse(resourceJSON[0])?.code?.text).toEqual('test2');
          |                                                     ^
      189 |   });
      190 | });
      191 |

      at toEqual (src/fhir/operations/export.test.ts:188:53)
      at call (src/fhir/operations/export.test.ts:2:1)
      at Generator.tryCatch (src/fhir/operations/export.test.ts:2:1)
      at Generator._invoke [as next] (src/fhir/operations/export.test.ts:2:1)
      at asyncGeneratorStep (src/fhir/operations/export.test.ts:2:1)
[...]

PR #2322, opened 2023-06-28
Fixes #2306 - remove console.log from export tests due to race conditions

### LINKED ISSUE #2306: Console logs in bulk export test
https://github.com/medplum/medplum/actions/runs/5383686843/jobs/9770741053?pr=2305

image

tonylee80:
Using `waitForAsyncJob` to wait for an async job like Bulk export to complete before the test cases ends.

Related PRs
#2022 
#2080 
#1959 

https://github.com/medplum/medplum/actions/runs/5404805700/jobs/9819491438?pr=2322#step:6:1405
image

### DIFF
--- a/packages/server/src/fhir/operations/export.test.ts
+++ b/packages/server/src/fhir/operations/export.test.ts
@@ -4,12 +4,10 @@ import express from 'express';
 import request from 'supertest';
 import { initApp, shutdownApp } from '../../app';
 import { loadTestConfig } from '../../config';
-import { createTestProject, initTestAuth, waitFor } from '../../test.setup';
+import { createTestProject, initTestAuth, waitFor, waitForJob } from '../../test.setup';
 import { systemRepo } from '../repo';
 import { exportResourceType } from './export';
 import { BulkExporter } from './utils/bulkexporter';
-
-jest.mock('../../logger');
 
 const app = express();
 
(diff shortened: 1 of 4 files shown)


codyebberson:
Hi @tonylee80  - I don't think this is the right fix for this.  The "database not setup" message implies that there is still something running in the background after the test completes.  That usually means we need to wait for a promise or a job to complete.

tonylee80:
> Hi @tonylee80 - I don't think this is the right fix for this. The "database not setup" message implies that there is still something running in the background after the test completes. That usually means we need to wait for a promise or a job to complete.

That's a really good point. I'll refactor to use `waitForAsyncJob` that's similarly used in [super.test.ts](https://github.com/medplum/medplum/blob/main/packages/server/src/admin/super.test.ts#L578) to ensure the bulk job is complete before continuing to other tests.

packages/server/src/fhir/operations/export.test.ts
```diff
@@ -68,15 +68,13 @@ describe('Export', () => {
 
     // Check the export status
     const contentLocation = new URL(initRes.headers['content-location']);
+    await waitForAsyncJob(initRes.headers['content-location'], app, accessToken);
 
-    let resBody: any;
-    await waitFor(async () => {
```
tonylee80, line 73: moving to waitForAsyncJob to ensure all jobs are completed in export tests. Introducing to this pattern for maintaining these export tests will no longer introduce race conditions where a test ends, but a job might still be running in the background.

codyebberson (approved):

PR #2324, opened 2023-06-28
Fixes #2311 - better expression parsing in getSearchParameterDetails

### LINKED ISSUE #2311: getSearchParameterDetails incorrect array case
`getSearchParameterDetails` is a Medplum function for returning more details about a `SearchParameter`

It works by parsing `SearchParameter.expression` and cross walking `StructureDefinition` to get the more detailed schema.

For example:
* What is the database column name
* What is the underlying `ElementDefinition` in the context of a resourceType
* Can it be an array

There is an awkward case that is not well handled: When `StructureDefinition.expression` has multiple entries for the same resource.  That case was not handled in the original implementation.  It turns out there are a handful of cases:

```
[
  'AllergyIntolerance code (2): ["AllergyIntolerance.code","AllergyIntolerance.reaction.substance"]',
  'DocumentManifest identifier (2): ["DocumentManifest.masterIdentifier","DocumentManifest.identifier"]',
  'DocumentReference identifier (2): ["DocumentReference.masterIdentifier","DocumentReference.identifier"]',
  'Location name (2): ["Location.name","Location.alias"]',
  'Observation combo-code (2): ["Observation.code","Observation.component.code"]',
  'Observation combo-data-absent-reason (2): ["Observation.dataAbsentReason","Observation.component.dataAbsentReason"]',
  'Observation combo-value-concept (2): ["Observation.value","Observation.component.value"]',
  'Observation combo-value-quantity (2): ["Observation.value","Observation.component.value"]',
  'Organization name (2): ["Organization.name","Organization.alias"]',
  'Substance code (2): ["Substance.code","Substance.ingredient.substance"]'
]
```

This will require a database migration, and extra care to handle the migration window.

codyebberson:
https://github.com/medplum/medplum/issues/2311

Before: We used simplistic string parsing to map `SearchParamter` to `ElementDefinition`

After: We use our FHIRPath parser to properly parse the expression, and map FHIRPath Atoms to details

The following search param columns need to be migrated to array columns:

| Resource Type | ID | Column Name |
| -- | -- | -- |
| AllergyIntolerance | clinical-code | code |
| Location | Location-name | name |
| Observation | Observation-combo-code | comboCode |
| Observation | Observation-combo-data-absent-reason | comboDataAbsentReason |
| Observation | Observation-combo-value-concept | comboValueConcept |
| Observation | Observation-combo-value-quantity | comboValueQuantity |
| Organization | Organization-name | name |
| Substance | Substance-code | code |


packages/core/src/search/details.ts
```diff
@@ -55,54 +72,110 @@ function setSearchParameterDetails(resourceType: string, code: string, details:
...
+  const propertyName = currAtom.toString();
+  let hasArrayIndex = false;
+  let nextIndex = index + 1;
+
+  if (nextIndex < atoms.length && atoms[nextIndex] instanceof IndexerAtom) {
+    hasArrayIndex = true;
+    nextIndex++;
+  }
+
+  const elementDefinition = getElementDefinition(baseType, propertyName);
```
mattwiller, line 129: Should this use the new `InternalTypeSchema` instead?  I'm also happy to leave it this way for now and convert it over later

mattwiller (approved):

packages/core/src/search/details.ts
codyebberson: Good question.  This file is already using `getElementDefinition`, so I'm inclined to wait until we're migrating everything after the validation cutover.  Good suggestion though

PR #2327, opened 2023-06-28
Add baseUrl to Medplum config file

mattwiller:


codyebberson (commented):
Do we want to do it this way?  Or should we update the `update-app` command to just use `config.apiDomainName` directly?

Change this line https://github.com/medplum/medplum/blob/main/packages/cli/src/aws/update-app.ts#L37

From current:

```ts
MEDPLUM_BASE_URL: config.baseUrl as string,
```

to:

```ts
MEDPLUM_BASE_URL: `https://${config.apiDomainName}/`,
```

Might be nice to avoid duplicates.

codyebberson (approved):

reshmakh:

PR #2328, opened 2023-06-28
Updating subscription documentation to reflect behavior when resources are deleted

reshmakh:


packages/docs/docs/subscriptions/subscription-extensions.md
```diff
@@ -64,7 +64,7 @@ Below are explanations of the different extensions Medplum Provides
 ## Interactions
 
 :::caution Note
-By default, FHIR Subscriptions will execute on "create", "update", "delete" operations.
+By default, FHIR Subscriptions will execute on "create" and "update" operations.
 :::
```
codyebberson, line 69: I think part of the confusion in this doc is that "create" and "delete" are not clearly separated.

I recommend we add a 3-# subheader here for `### Subscription for "create"-only events`

And then another one below for `### Delete subscriptions`

One limitation of the current model is that it appears we only support "create" or "delete", but not both.  We may want to add support for wildcard "*" or a comma separated list.

packages/docs/docs/subscriptions/subscription-extensions.md
```diff
@@ -116,7 +116,26 @@ The response for a deleted resource will contain:
 
 **_Executing only on delete_**
```
codyebberson, line 117: Let's change this from bold text to 3-# header `### Subscription for "delete" events`

codyebberson (approved):

PR #2337, opened 2023-06-29
Implement slice validation

mattwiller:
Checks the values in a sliced array property to ensure that each slice contains the required number of matching elements.  This currently supports the most common slice discriminator types: `value`/`pattern`, `exists`, and `type`.


packages/core/src/typeschema/validation.ts
```diff
@@ -149,10 +157,22 @@ class ResourceValidator {
         );
       }
 
-      this.checkSpecifiedValue(value, element, path);
+      if (!matchesSpecifiedValue(value, element)) {
+        this.issues.push(createStructureIssue(path, 'Value did not match expected pattern'));
+      }
+      const sliceCounts: Record<string, number> = Object.fromEntries(
```
codyebberson, line 163: I see at least 3 different checks for `element.slicing` - I wonder if you could do a minor refactor to do that check once to simplify

```ts
const slices = element.slicing?.slices;
const sliceCounts = slices.reduce((acc, s) => ({ ...acc, [s.name]: 0 }), {}) ?? [];

// ...

if (slices) {
  // ...
}
```

Using `?? []` makes for shorter code, but it leads to unnecessary allocations and unnecessary function calls which we may want to avoid in an inner inner loop like this.

packages/core/src/typeschema/validation.ts
```diff
@@ -360,3 +388,67 @@ function checkObjectForNull(obj: Record<string, unknown>, path: string, issues:
...
+  value: TypedValue | TypedValue[] | undefined,
+  discriminator: SliceDiscriminator,
+  slice: SliceDefinition
+): boolean {
+  const element = slice.fields[discriminator.path];
+  switch (discriminator.type) {
+    case 'value':
+    case 'pattern':
+      if (!element) {
+        throw new OperationOutcomeError(
```
codyebberson, line 411: You can just `throw new Error` rather than `throw new OperationOutcomeError(serverError(...))` - It will end up in the same place.

I'm trying to wrap my head around how we get here.... This would be the case where someone writes a FHIR Profile with an invalid slice expression?  Should that be a 500 or a 400?

packages/core/src/typeschema/validation.ts
```diff
@@ -360,3 +388,67 @@ function checkObjectForNull(obj: Record<string, unknown>, path: string, issues:
...
+
+function checkSliceElement(value: TypedValue, slicingRules: SlicingRules): string | undefined {
+  nextSlice: for (const slice of slicingRules.slices) {
+    for (const discriminator of slicingRules.discriminator) {
+      const discrimValues = arrayify(getNestedProperty(value, discriminator.path));
+      if (!discrimValues) {
+        return undefined;
+      }
+      if (!discrimValues.some((v) => matchDiscriminant(v, discriminator, slice))) {
+        continue nextSlice;
```
codyebberson, line 448: If you want to avoid the `nextSlice:` label, I think you could replace this with a simple `break;`, no?

mattwiller: Changing this to a `break` would alter the semantics of the loop due to the `return` below, but I can take another crack at refactoring to remove the label 👍

packages/core/src/typeschema/validation.ts
mattwiller: Thanks — that's a great callout on the allocation!  I can definitely refactor that out

packages/core/src/typeschema/validation.ts
mattwiller: This really should never happen; this is basically just a safety assertion so we get a better error message for this case if it ever pops up

codyebberson (approved):

reshmakh:

Issue #1423, opened 2023-01-20
Use aws-sdk-client-mock


codyebberson:
https://github.com/m-radzikowski/aws-sdk-client-mock

Looks much better than our manual mocks.

h/t @elliot-nelson

atifnimran:
Yes

taylorqj:
Have you taken a look at [LocalStack](https://github.com/localstack/localstack)? I've used this in the past as a replacement to mocking external cloud services.

Paul-Taiwo:
Hi @codyebberson. I would love to give this a try

mattwiller:
@Paul-Taiwo Please go right ahead!

brandonin:
Howdy @mattwiller :)

mattwiller:
Oh hey @brandonin! 👋

Paul-Taiwo:
Hey @mattwiller, I see @brandonin has a PR for the remaining manual mocks. Should I go ahead to work on it?

Paul-Taiwo:
Hi @codyebberson, I've completed all the tasks and updated the mocks. However, I noticed that the issue is still marked as open. Could you please review the changes and let me know if there's anything else I can do to help move this forward? Thank you!

mattwiller:
I went ahead and closed it — thanks so much @Paul-Taiwo and @brandonin 🎉🎉

PR #2341, opened 2023-06-30
move the SSM and Secrets Manager to use aws-sdk-client-mock

brandonin:
Looks like my branches are off-based. Squash and merge should solve this and I can fix this on my own repo later.

This can be used as a basis on how to write out mocks utilizing the new library.

@codyebberson let me know if there are any extra requirements that you may need.

I attempted to do the rest of the libraries but kept receiving this error even though I did `npm ci` at the root: 
```

 FAIL  src/logger.test.ts
  ● Test suite failed to run

    Cannot find module '@medplum/core' from 'src/test.setup.ts'

    Require stack:
      src/test.setup.ts
      src/logger.test.ts

    > 1 | import { createReference, getReferenceString, ProfileResource, sleep } from '@medplum/core';
        | ^
      2 | import {
      3 |   AccessPolicy,
      4 |   Bundle,

      at Resolver._throwModNotFoundError (../../node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.require (src/test.setup.ts:1:1)
      at Object.require (src/logger.test.ts:12:1)

```

I couldn't find anything in the docs or previous issues for this.

https://github.com/medplum/medplum/issues/1423


codyebberson:
Thanks @brandonin !

> I attempted to do the rest of the libraries but kept receiving this error even though I did `npm ci` at the root:

Did you happen to run `npm run build` from the root or in core?  My guess is that core isn't built, and therefore the server can't find it.

codyebberson:
I believe this was implemented in https://github.com/medplum/medplum/pull/2407

Thanks @brandonin !

PR #2349, opened 2023-07-02
Initial websocket connectivity support

codyebberson:
Integrates a websocket echo service.


packages/server/src/app.ts
```diff
@@ -174,4 +180,15 @@ export async function shutdownApp(): Promise<void> {
   await closeDatabase();
   closeRedis();
   closeRateLimiter();
+  closeWebSockets();
+
+  if (server) {
+    server.close();
+    setImmediate(() => {
+      if (server) {
+        server.emit('close');
```
mattwiller, line 189: Why does this need to be delayed from immediate execution?

packages/server/src/websockets.ts
```diff
@@ -0,0 +1,60 @@
...
+import { getRedis } from './redis';
+
+let wsServer: ws.Server | undefined = undefined;
+
+/**
+ * Initializes a websocket listener on the given HTTP server.
+ * @param server The HTTP server.
+ */
+export function initWebSockets(server: http.Server): void {
+  wsServer = new ws.Server({ noServer: true });
```
mattwiller, line 14: A thought on other options we should consider: `maxPayload` defaults to 100 MiB, which is a ton of data to be tossing around and could leave the server vulnerable to DoS attacks; we should tune this down to maybe 1 MiB or less?

mattwiller (approved):

packages/server/src/app.ts
codyebberson: Good question, ugly answer.

The problem was that the `ws` package prevented the server from cleanly shutting down.  The `setImmediate` hack is found in a few places.  The discussion here is not super satisfying, but it works: https://stackoverflow.com/questions/14626636/how-do-i-shutdown-a-node-js-https-server-immediately

packages/server/src/websockets.ts
codyebberson: Good call.

We have a server config setting for `maxJsonSize` which defaults to `1mb`.  Thoughts on using the same setting?  Convenience of one setting, vs explicitness of multiple settings.

packages/server/src/app.ts
mattwiller: Makes sense — would you mind adding that link as a comment alongside the code for future reference?

packages/server/src/websockets.ts
mattwiller: I like one setting for now; we can always split out a new setting if necessary to tune this maximum independently in the future

reshmakh (approved):

packages/server/src/app.ts
codyebberson: Just tested again without it, and it works now.... so I just removed it 🤷

Issue #2064, opened 2023-05-17
Evaluate FHIRPath constraints during resource validation


mattwiller:
The FHIR `StructureDefinition` resource that defines resource types can contain a  number of different validation rules, including very powerful expression-based rules called "constraints" or "invariants"; these can assert over multiple different fields simultaneously and perform more complex logic like math or string manipulation.  The `ElementDefinition.constraint` field contains information about the different constraints present on an element, for example:
```json
[
  {
    "key": "dom-6",
    "severity": "warning",
    "human": "A resource should have narrative for robust management",
    "expression": "text.`div`.exists()",
    "source": "http://hl7.org/fhir/StructureDefinition/DomainResource"
  },
  {
    "key": "obs-6",
    "severity": "error",
    "human": "dataAbsentReason SHALL only be present if Observation.value[x] is not present",
    "expression": "dataAbsentReason.empty() or value.empty()"
  },
  {
    "key": "obs-7",
    "severity": "error",
    "human": "If Observation.code is the same as an Observation.component.code then the value element associated with the code SHALL NOT be present",
    "expression": "value.empty() or component.code.where(coding.intersect(%resource.code.coding).exists()).empty()"
  }
]
```

This list includes constraints from the specific resource type, but may also includes ones sourced from base types.  Only constraints with a `severity` of "error" should be validated.  The `expression` field contains a FHIRPath expression that should be evaluated in the context of the resource being validated, using the [`evalFhirPathTyped()`](https://github.com/medplum/medplum/blob/fbbbc390274a5e4e7ec01635d4f8a7679e837f46/packages/core/src/fhirpath/parse.ts#L247-L255) method.  The expression is expected to return `true` if the resource satisfies the constraint.  Any other result value should cause the server to return a 400 Bad Request to the client, including the constraint `key` and description from the `human` field.

__NOTE:__ This is almost certainly a breaking change for existing customers, and testing should be performed to identify and mitigate the risk of breaking customer apps.  We should also consider benchmarking write operations before releasing this feature, as it is likely to significantly increase the time taken by validation as a whole.

PR #2352, opened 2023-07-03
ImpliesAtom

jamestouri:
When working on the constraint validation at https://github.com/medplum/medplum/issues/2064, there are several key features in FHIRPath logic that hasn't been implemented. One of them includes `Implies` which is needed for several rule constraints

http://hl7.org/fhirpath/N1/#implies

### DIFF
--- a/packages/core/src/fhirpath/atoms.test.ts
+++ b/packages/core/src/fhirpath/atoms.test.ts
@@ -53,17 +53,17 @@ describe('Atoms', () => {
   });
 
   test.each([
-    ['true implies true', true],
-    ['true implies false', false],
-    ['true implies null', undefined],
-    ['false implies true', true],
-    ['false implies false', true],
-    ['false implies null', true],
-    ['null implies true', true],
-    ['null implies false', undefined],
-    ['null implies null', undefined],
+    ['true implies true', [true]],
+    ['true implies false', [false]],
+    ['true implies {}', []],
+    ['false implies true', [true]],
+    ['false implies false', [true]],
+    ['false implies {}', [true]],
+    ['{} implies true', [true]],
+    ['{} implies false', []],
+    ['{} implies {}', []],
   ])('ImpliesAtom: %s to equal %s', (input: any, expected: any) => {
-    expect(evalFhirPath(input, [])).toEqual([expected]);
+    expect(evalFhirPath(input, [])).toEqual(expected);
   });
 
   test('AsAtom', () => {
--- a/packages/core/src/fhirpath/atoms.test.ts
+++ b/packages/core/src/fhirpath/atoms.test.ts
@@ -53,7 +53,16 @@ describe('Atoms', () => {
   });
 
   test.each([
-  ])('ImpliesAtom: %s %s', (input: any, expected: any) => {
+    ['true implies true', true],
+    ['true implies false', false],
+    ['true implies null', undefined],
+    ['false implies true', true],
+    ['false implies false', true],
+    ['false implies null', true],
+    ['null implies true', true],
+    ['null implies false', undefined],
+    ['null implies null', undefined],
+  ])('ImpliesAtom: %s to equal %s', (input: any, expected: any) => {
     expect(evalFhirPath(input, [])).toEqual([expected]);
   });
 
--- a/packages/core/src/fhirpath/atoms.test.ts
+++ b/packages/core/src/fhirpath/atoms.test.ts
@@ -52,6 +52,11 @@ describe('Atoms', () => {
     expect(evalFhirPath('x | y', [])).toEqual([]);
   });
 
+  test.each([
+  ])('ImpliesAtom: %s %s', (input: any, expected: any) => {
+    expect(evalFhirPath(input, [])).toEqual([expected]);
+  });
+
   test('AsAtom', () => {
     const obs1: Observation = {
       resourceType: 'Observation',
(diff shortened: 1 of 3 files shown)


packages/core/src/fhirpath/atoms.test.ts
```diff
@@ -52,6 +52,20 @@ describe('Atoms', () => {
     expect(evalFhirPath('x | y', [])).toEqual([]);
   });
 
+  test.each([
+    ['true implies true', true],
+    ['true implies false', false],
+    ['true implies null', undefined],
```
codyebberson, line 58: Can you add test cases for the "empty" column on this page: https://build.fhir.org/ig/HL7/FHIRPath/

You might need to refactor the `test.each()` array to be able to handle the empty array case.

```ts
['true implies {}', []],
['false implies {}', [true]],
['{} implies false', []],
['{} implies {}', []],
```

codyebberson (commented):
Nice, clean.  Thanks @jamestouri 

See note on some additional test cases from https://build.fhir.org/ig/HL7/FHIRPath/

codyebberson:
(duplication warning is ok with me -- it would be a gross refactor for minimal gain)

packages/core/src/fhirpath/atoms.ts
```diff
@@ -352,6 +352,35 @@ export class XorAtom extends InfixOperatorAtom {
...
+    super('implies', left, right);
+  }
+
+  eval(context: AtomContext, input: TypedValue[]): TypedValue[] {
+    const leftResult = this.left.eval(context, input);
+    const rightResult = this.right.eval(context, input);
+    if (leftResult.length === 0 && rightResult.length === 0) {
+      return [];
+    }
+    const leftValue = leftResult.length === 0 ? null : leftResult[0].value;
```
mattwiller, line 372: In the case where either argument collection has more than one element in it, this should probably produce an error to let the user know that the expression is incorrect.  In general, FHIRPath is pretty strict when coercing from a collection to a single value; that part of evaluation is covered explicitly here: http://hl7.org/fhirpath/N1/#singleton-evaluation-of-collections

packages/core/src/fhirpath/atoms.ts
```diff
@@ -352,6 +352,35 @@ export class XorAtom extends InfixOperatorAtom {
...
+    const rightValue = rightResult.length === 0 ? null : rightResult[0].value;
+    if (rightValue === true || leftValue === false || (leftValue === null && rightValue === true)) {
+      return booleanToTypedValue(true);
+    }
+    if (leftValue === true && rightValue === false) {
+      return booleanToTypedValue(false);
+    }
+    return [];
```
mattwiller, line 380: You already have a case above that returns `[]`, which I think could be combined if you reorganize the conditions like so:
```ts
if (leftValue === false || rightValue === true) {
  return true;
} else if (leftValue === null || rightValue === null) {
  return [];
} else {
  return false;
}
```

packages/core/src/fhirpath/parse.ts
```diff
@@ -210,7 +211,8 @@ export function initFhirPathParserBuilder(): ParserBuilder {
       (left, _, right) => new ArithemticOperatorAtom('mod', left, right, (x, y) => x % y)
     )
     .infixLeft('or', OperatorPrecedence.Is, (left, _, right) => new OrAtom(left, right))
-    .infixLeft('xor', OperatorPrecedence.Is, (left, _, right) => new XorAtom(left, right));
+    .infixLeft('xor', OperatorPrecedence.Is, (left, _, right) => new XorAtom(left, right))
+    .infixLeft('implies', OperatorPrecedence.Is, (left, _, right) => new ImpliesAtom(left, right));
```
mattwiller, line 215: Why wouldn't this be `OperatorPrecedence.Implies`?

packages/core/src/fhirpath/atoms.ts
```diff
@@ -352,6 +352,43 @@ export class XorAtom extends InfixOperatorAtom {
...
+    const leftResult = this.left.eval(context, input);
+    const rightResult = this.right.eval(context, input);
+    if (!this.isValidCollectionLength(leftResult, rightResult)) {
+      throw new Error(
+        `One or more collections are not of length 1 leftResult: ${leftResult.length} rightResult: ${rightResult.length}`
+      );
+    }
+    const leftValue = leftResult.length === 0 ? null : leftResult[0].value;
+    const rightValue = rightResult.length === 0 ? null : rightResult[0].value;
+    if (rightValue === true || leftValue === false || (leftValue === null && rightValue === true)) {
```
mattwiller, line 376: The rightmost part of this condition is superfluous; it can be rewritten equivalently as `rightValue === true || leftValue === false`

packages/core/src/fhirpath/atoms.ts
```diff
@@ -352,6 +352,43 @@ export class XorAtom extends InfixOperatorAtom {
...
+      return [];
+    }
+    return booleanToTypedValue(false);
+  }
+
+  private isValidCollectionLength(leftResult: TypedValue[], rightResult: TypedValue[]): boolean {
+    if (leftResult.length > 1 || rightResult.length > 1) {
+      return false;
+    }
+    return true;
```
mattwiller, line 388: This can be rewritten more concisely as `return (leftResult.length <= 1 && rightValue.length <= 1);`

packages/core/src/fhirpath/atoms.ts
jamestouri: Ah, that's right good catch

mattwiller (approved):

PR #2361, opened 2023-07-03
Split repo.ts and search.ts

codyebberson:
This should be a 100% re-organization PR.  `repo.ts` and `repo.test.ts` are two of our largest source files.  So much so that VS Code gets sluggish, and it's hard to surgically run a coherent block of tests together.  So this splits them up a bit.

Before:

- repo.ts - 2483 lines
- repo.test.ts - 3474 lines

After:

- repo.ts - 1782 lines
- repo.test.ts - 793 lines
- search.ts - 764 lines
- search.test.ts - 2667 lines

There is another chunk of code that could be split into another file, perhaps `indexer.ts`, which is the code that indexes a resource for searching.  All search tests are really a combination of "indexing" and "searching".

----

Addendum:  This will blow away git blame for search.ts and search.test.ts.  If anyone cares, it looks like it would be possible to use some git magic to preserve history across files.


mattwiller (approved):

tonylee80 (approved):

tonylee80:
Looks good, tested this locally, search is working as expected.

PR #2373, opened 2023-07-05
Fix sonar warnings in repo.ts and search.ts

codyebberson:
Previous PR moved a bunch of code around, which Sonar flagged as code smells.  Fixing the smells.

Previous PR: https://github.com/medplum/medplum/pull/2361

Sonar Code Smells: https://sonarcloud.io/project/issues?id=medplum_medplum&pullRequest=2361&resolved=false&types=CODE_SMELL

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -444,9 +444,7 @@ export function getElementDefinition(typeName: string, propertyName: string): El
  * @param value The object to check
  * @returns True if the input is of type 'object' and contains property 'resourceType'
  */
-export function isResource<T extends Resource = Resource>(
-  value: T | Reference<T> | string | undefined | null
-): value is T {
+export function isResource<T extends Resource = Resource>(value: unknown): value is T {
   return !!(value && typeof value === 'object' && 'resourceType' in value);
 }
 
@@ -455,9 +453,7 @@ export function isResource<T extends Resource = Resource>(
  * @param value The object to check
  * @returns True if the input is of type 'object' and contains property 'reference'
  */
-export function isReference<T extends Resource>(
-  value: T | Reference<T> | string | undefined | null
-): value is Reference<T> & { reference: string } {
+export function isReference<T extends Resource>(value: unknown): value is Reference<T> & { reference: string } {
   return !!(value && typeof value === 'object' && 'reference' in value);
 }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/types.ts
```diff
@@ -444,9 +444,7 @@ export function getElementDefinition(typeName: string, propertyName: string): El
  * @param value The object to check
  * @returns True if the input is of type 'object' and contains property 'resourceType'
  */
-export function isResource<T extends Resource = Resource>(
-  value: T | Reference<T> | string | undefined | null
-): value is T {
+export function isResource<T extends Resource = Resource>(value: unknown): value is T {
```
mattwiller, line 447: This type signature didn't immediately make sense to me; is the generic type parameter necessary slash could this just be `export function isResource(value: unknown): value is Resource`?

mattwiller (approved):

packages/core/src/types.ts
codyebberson: It can still be useful if you pass in a value with type `Patient | undefined`, then the type assertion will be `value is Patient` rather than `value is Resource`.

codyebberson: Yep, good call.  Just didn't clean it up all the way.

codyebberson: TypeScript is smart. 

```ts

interface Resource {
    resourceType: string;
}

interface Patient extends Resource {
    name: string;
}

interface Observation extends Resource {
    value: number;
}

function isResource1(input: unknown): input is Resource {
    return !!input && (typeof input === 'object') && ('resourceType' in input);
}

function isResource2<T extends Resource>(input: unknown): input is T {
    return !!input && (typeof input === 'object') && ('resourceType' in input);
}

const myInput = {} as Patient | undefined;

if (isResource1(myInput)) {
    console.log('1', myInput); // myInput is Patient, even though I assumed it would be Resource
}

if (isResource2(myInput)) {
    console.log('2', myInput); // myInput is Patient
}

```

codyebberson: Hm, `readReference` is not as smart.  Requires changes like this:

```
-    if (isReference<T>(value)) {
-      return medplum.getCachedReference(value);
+    if (isReference(value)) {
+      return medplum.getCachedReference(value as Reference<T>);
```

PR #2382, opened 2023-07-05
Cut over to new validation logic

mattwiller:
Switches the server to use the new strict validation function, and deprecates the legacy function with a note to use the new one

### DIFF
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -18,7 +18,7 @@ import {
 } from '@medplum/fhirtypes';
 import { readFileSync } from 'fs';
 import { resolve } from 'path';
-import { validate as validateResource } from './validation';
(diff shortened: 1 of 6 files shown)


packages/core/src/typeschema/validation.test.ts
```diff
@@ -18,7 +18,7 @@ import {
 } from '@medplum/fhirtypes';
 import { readFileSync } from 'fs';
 import { resolve } from 'path';
-import { validateResource } from './validation';
+import { validate as validateResource } from './validation';
```
codyebberson, line 21: Might be an ugly diff, but can we update the tests to avoid the alias?

codyebberson (approved):
Minor suggestion to replace `validateResource` with `validate` in `validation.test.ts`.  Your call.  lgtm 👍

reshmakh:

PR #2387, opened 2023-07-06
Minor perf improvement to validator

codyebberson:
@mattwiller - for your consideration.  see comments inline

### DIFF
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -18,7 +18,7 @@ import {
 } from '@medplum/fhirtypes';
 import { readFileSync } from 'fs';
 import { resolve } from 'path';
-import { validate as validateResource } from './validation';
+import { validate } from './validation';
 import { indexStructureDefinitionBundle } from '../types';
 import { readJson } from '@medplum/definitions';
 import { loadDataTypes } from './types';
@@ -41,10 +41,10 @@ describe('FHIR resource validation', () => {
 
   test('Invalid resource', () => {
     expect(() => {
-      validateResource(undefined as unknown as Patient);
+      validate(undefined as unknown as Patient);
     }).toThrow();
     expect(() => {
-      validateResource({} as unknown as Patient);
+      validate({} as unknown as Patient);
     }).toThrow();
   });
 
@@ -55,7 +55,7 @@ describe('FHIR resource validation', () => {
       birthDate: '1949-08-14',
     };
     expect(() => {
-      validateResource(patient);
+      validate(patient);
     }).not.toThrow();
   });
 
@@ -73,10 +73,10 @@ describe('FHIR resource validation', () => {
       },
     } as unknown as Patient;
     expect(() => {
-      validateResource(invalidMultiple);
+      validate(invalidMultiple);
     }).toThrow();
     expect(() => {
-      validateResource(invalidSingle);
+      validate(invalidSingle);
     }).toThrow();
   });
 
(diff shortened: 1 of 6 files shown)


packages/core/src/typeschema/validation.ts
```diff
@@ -316,27 +314,6 @@ function isChoiceOfType(
   return false;
 }
 
-function getNestedProperty(value: TypedValue, key: string): (TypedValue | TypedValue[] | undefined)[] {
-  const [firstProp, ...nestedProps] = key.split('.');
```
codyebberson, line 320: I ran a sanity test on this  `if (key.includes('.')) throw new Error('period');` and it didn't throw

Up above, it looks like you're already doing the key splitting and path management.  This function never enters into the `for` loop.

packages/core/src/typeschema/validation.ts
```diff
@@ -113,46 +113,44 @@ class ResourceValidator {
   }
 
   private checkProperty(parent: TypedValue, key: string, schema: InternalTypeSchema, path: string): void {
-    const propertyValues = getNestedProperty(parent, key);
+    const value = getTypedPropertyValue(parent, key);
```
codyebberson, line 116: By replacing `getNestedProperty` with a direct call to `getTypedPropertyValue`, you can avoid the outer array, and avoid the for loop on old line 121.

reshmakh (approved):

mattwiller:
The case that `getNestedPropertyValue()` was intended to handle was when profiles "walk into" types to place constraints on specific fields; for example, the [US Core Patient profile](http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient) makes `Patient.telecom.system` required.  Let me try adding a test case for that to ensure it's still working correctly

mattwiller (changes requested):
I added a test case on this branch with a Patient that almost conforms to the US Core Patient profile:
```js
{
  resourceType: 'Patient',
  identifier: [{
    system: 'http://example.com',
    value: 'foo',
  }],
  telecom: [{
    // Missing system property
    value: '555-555-5555',
  }],
  gender: 'unknown',
  name: [{
      given: ['Test'],
      family: 'Patient',
  }],
}
```
Unfortunately it fails:
```
Expected message: "Missing required property (Patient.telecom.system)"
Received message: "Missing required property (Patient.identifier.system); Missing required property (Patient.identifier.value); Missing required property (Patient.telecom.system); Missing required property (Patient.telecom.value)"
```
I verified that this same test passes on `main` and will create a PR to add it

packages/core/src/typeschema/validation.test.ts
```diff
@@ -790,16 +790,6 @@ describe('Legacy tests for parity checking', () => {
     }
   });
 
-  test('StructureDefinition', () => {
```
codyebberson, line 793: I believe this test is identical to the test on line 318

mattwiller: It is, great catch!

codyebberson:
> Unfortunately it fails:
> 
> ```
> Expected message: "Missing required property (Patient.telecom.system)"
> Received message: "Missing required property (Patient.identifier.system); Missing required property (Patient.identifier.value
> ```

Interesting.  Ok, I'll take a look at this.  Are "expected" and "received" reversed?  It seems like "received" is a superset of expected?  or are those other errors wrong?

mattwiller:
>  Are "expected" and "received" reversed? It seems like "received" is a superset of expected? or are those other errors wrong?

Nope, that's the correct labeling: the only error in the test resource should be the absence of `Patient.telecom.system`

codyebberson:
@mattwiller  Thanks for new test case! Makes more sense now.  Closing this out.

mattwiller:
I think there is still a possibility that we could get rid of the nested property accesses, but we might need to tweak the `InternalTypeSchema` representation to make that possible

PR #2408, opened 2023-07-08
Added jwt bearer token auth to MedplumClient

codyebberson:
1. Added support for `urn:ietf:params:oauth:grant-type:jwt-bearer`
2. Added a `isMedplumAccessToken` helper to determine if using `MedplumClient` with a Medplum server vs generic FHIR server
3. Only do special Medplum stuff (lookup profile, access policies, etc) if using a Medplum server

This is in service of accessing the Health Gorilla FHIR API using `MedplumClient`, although it would apply to any FHIR server that requires "client assertion" JWT bearer tokens.

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -572,6 +572,7 @@ describe('Client', () => {
     const client = new MedplumClient({ fetch });
     const result1 = await client.startJwtBearerLogin('test-client-id', 'test-client-secret', 'openid profile');
     expect(result1).toBeDefined();
+    expect(result1).toMatchObject({ resourceType: 'ClientApplication' });
     expect(fetch).toHaveBeenCalledTimes(2);
   });
 
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -554,6 +554,27 @@ describe('Client', () => {
     expect(fetch).toHaveBeenCalledTimes(4);
   });
 
+  test('JWT bearer token flow', async () => {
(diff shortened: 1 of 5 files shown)


packages/core/src/client.test.ts
```diff
@@ -140,7 +140,7 @@ describe('Client', () => {
     window.localStorage.setItem(
```
reshmakh, line 140: is JWT appropriate for client?  I thought not - but perhaps I'm missing something.

codyebberson: Hm, perhaps some naming confusion.

"JWT" - a format of access token, which includes a set of "claims" signed by the server

Medplum uses JWTs for access tokens, even when you're signing in with email/password or google auth or whatever

"JWT bearer grant type" - a special OAuth flow where the client signs their own JWT using the preconfigured client ID and client secret.

packages/core/src/client.test.ts
```diff
@@ -554,6 +554,27 @@ describe('Client', () => {
...
+      }
+      if (url.includes('/auth/me')) {
+        return { profile: { resourceType: 'ClientApplication' } };
+      }
+      return {};
+    });
+
+    const client = new MedplumClient({ fetch });
+    const result1 = await client.startJwtBearerLogin('test-client-id', 'test-client-secret', 'openid profile');
+    expect(result1).toBeDefined();
```
mattwiller, line 574: Is there a stronger assertion to be made here about `result1`?

mattwiller (approved):

PR #2420, opened 2023-07-10
Add validation benchmarks

mattwiller:


packages/core/src/typeschema/benchmarks.test.ts
```diff
@@ -0,0 +1,277 @@
+import { AuditEvent, Bundle, Patient, StructureDefinition } from '@medplum/fhirtypes';
+import { Bench } from 'tinybench';
+import { validate } from './validation';
+import { validateResource } from '../schema';
+import { readJson } from '@medplum/definitions';
+import { indexStructureDefinitionBundle } from '../types';
+import { loadDataTypes } from './types';
+
+describe.skip('Validation benchmarks', () => {
```
mattwiller, line 9: I wasn't sure how we might want to gather and run benchmark tests: it seemed ill-advised to run them with the other tests due to how long they take, and there's not an obvious way to track their performance over time

packages/core/package.json
```diff
@@ -17,7 +17,8 @@
   },
   "devDependencies": {
     "@medplum/definitions": "*",
-    "@medplum/fhirtypes": "*"
+    "@medplum/fhirtypes": "*",
+    "tinybench": "^2.5.0"
```
jamestouri, line 21: I know @codyebberson tends to be nitpicky about the ^ on dependency versions if it could be removed

jamestouri (approved):

PR #2437, opened 2023-07-11
Add server config for logLevel

rahul1:
### <samp>🤖 Generated by Copilot at e8ca9e3</samp>

This pull request adds a new `logLevel` configuration option for the server, which allows setting the logging level from a string value. It implements the `parseLogLevel` function in `logger.ts` and uses it in `index.ts` to apply the option. It also updates the `config.ts` interface, the `logger.test.ts` file, and the documentation in `config-settings.md`.


packages/docs/docs/self-hosting/config-settings.md
```diff
@@ -106,34 +106,35 @@ For example, if your environment name is "prod", then the "baseUrl" parameter na
...
+| `userInfoUrl`          | The OAuth userinfo URL. By default, Medplum server uses built in OAuth, so `userInfoUrl` should be `baseUrl` + `oauth2/userinfo`.                                                                     |          |            | `baseUrl` + `oauth2/userinfo`                        |
+| `appBaseUrl`           | The fully qualified URL of the user-facing app. This is used for CORS and system generated emails. For example, `https://app.example.com/`.                                                           | yes      | `init`     |                                                      |
+| `logLevel`             | Verbosity of logging: `'NONE'`, `'ERROR'`, `'WARN'`, `'INFO'`, `'DEBUG'`                                                                                                                              |          |            | `'ERROR'` if `NODE_ENV == 'test`. `'INFO'` otherwise |
```
codyebberson, line 119: For default value, I think we can just say `INFO` -- `NODE_ENV == 'test'` is only when running unit tests.

packages/server/src/logger.ts
```diff
@@ -72,3 +72,20 @@ function getCloudWatchLogger(config: MedplumServerConfig): CloudWatchLogger {
   }
   return cloudWatchLogger;
 }
+
+export function parseLogLevel(level: string): LogLevel {
```
codyebberson, line 76: You can use the enum to do this rather than duplicating values:

```
function parseLogLevel(level: string): LogLevel {
  const value = LogLevel[level as keyof typeof LogLevel];
  
  if (value === undefined) {
    throw new Error(`Invalid log level: ${level}`);
  }
  
  return value;
}
```

rahul1:
@codyebberson ready for re-review

codyebberson (approved):
Thanks! 👍

PR #2445, opened 2023-07-12
Add bundle to resource array in MedplumClient

codyebberson:
Context: https://medplum.slack.com/archives/C04A55B3VU5/p1689171579275589

This adds a `.bundle` property to search arrays.  Example usage:

```ts
const array = await medplum.searchResources('Patient', { name: 'alice' });

// You can now use .bundle.total to get total results
console.log('total', array.bundle.total);

// You can continue to use the array as normal
for (const patient of array) {
    console.log(patient);
}
```

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1496,6 +1496,8 @@ describe('Client', () => {
     expect(Array.isArray(result)).toBe(true);
     expect(result.length).toBe(1);
     expect(result[0].resourceType).toBe('Patient');
+    expect(result.bundle).toBeDefined();
+    expect(result.bundle.resourceType).toBe('Bundle');
   });
 
   test('Search resources with record of params', async () => {
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -246,6 +246,13 @@ export interface FetchLike {
  */
 export type QueryTypes = URLSearchParams | string[][] | Record<string, any> | string | undefined;
 
+/**
+ * ResourceArray is an array of resources with a bundle property.
+ * The bundle property is a FHIR Bundle containing the search results.
+ * This is useful for retrieving bundle metadata such as total, offset, and next link.
+ */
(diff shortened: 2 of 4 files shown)


mattwiller:
What's the advantage of this approach vs. using the `client.search()` method to retrieve the actual `Bundle`?

rahul1 (changes requested):
I was initially opposed, but I'm now convinced!

Reminder to document on the "Basic Search" and "Paginated Search" guides

codyebberson:
> What's the advantage of this approach vs. using the `client.search()` method to retrieve the actual `Bundle`?

See the Slack thread for customer description.

Using `client.search()` is a reasonable suggestion.  `Bundle` just has known friction (need to navigate `.entry`, need to cast `.resource`, etc).

codyebberson:
> Reminder to document on the "Basic Search" and "Paginated Search" guides

Done ✅

mattwiller:
I would prefer deprecating the `searchResources()` method and adding a new one that presents an async iterator over all search results with the required properties attached, rather than monkey-patching a builtin JS type.  If for some reason the language ever adds an `Array.prototype.bundle()` method, the approach in this PR would result in a conflict

codyebberson:
@mattwiller 

> I would prefer deprecating the `searchResources()` method

`searchResources()` is very popular and widely used (including by many of our own sample apps and demo bots).  This isn't really an option.

> adding a new one that presents an async iterator over all search results with the required properties attached

This kind of exists in `searchResourcePages()`: https://www.medplum.com/docs/sdk/classes/MedplumClient#searchresourcepages

> If for some reason the language ever adds an Array.prototype.bundle() method

This is valid, although unlikely IMHO.

Would `.fhirBundle` be better?  `.originalBundle`?  `._bundle`?

Note that "pagination" is only one part of the use case of "total".  There are many UI cases where you want to show a count.

rahul1 (approved):

PR #2456, opened 2023-07-13
Use eslint-config in examples

codyebberson:


examples/foomedical/package.json
```diff
@@ -30,17 +31,11 @@
     "@types/node": "20.3.3",
     "@types/react": "18.2.14",
     "@types/react-dom": "18.2.6",
-    "@typescript-eslint/eslint-plugin": "5.60.1",
```
codyebberson, line 33: Moving all eslint dependencies into `eslint-config`

package.json
```diff
@@ -24,17 +24,11 @@
     "@cyclonedx/cyclonedx-npm": "1.12.1",
     "@types/jest": "29.5.2",
     "@types/node": "20.3.3",
-    "@typescript-eslint/eslint-plugin": "5.60.1",
```
codyebberson, line 27: Moving all eslint dependencies into `eslint-config`

packages/eslint-config/index.cjs
```diff
@@ -138,6 +138,7 @@ module.exports = {
       },
     ],
     'jsdoc/require-jsdoc': 'off',
+    'jsdoc/tag-lines': 'off',
```
codyebberson, line 141: I'll submit a separate issue to fix these separately.  It's about 800 lines.

scripts/publish.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-PACKAGES=("app" "bot-layer" "cdk" "cli" "core" "definitions" "fhir-router" "fhirtypes" "mock" "react")
+PACKAGES=("app" "bot-layer" "cdk" "cli" "core" "definitions" "eslint-config" "fhir-router" "fhirtypes" "mock" "react")
```
codyebberson, line 3: This will publish `@medplum/eslint-config` in the next release.  That will be nice, because if anyone forks `foomedical`, they'll get the config.

package.json
mattwiller: I left this out because the [ESLint docs](https://eslint.org/docs/latest/extend/shareable-configs) seemed to suggest that it should be a `peerDependency`:
> If your shareable config depends on a plugin, you should also specify it as a peerDependency (plugins will be loaded relative to the end user’s project, so the end user is required to install the plugins they need). However, if your shareable config depends on a [custom parser](https://eslint.org/docs/latest/extend/custom-parsers) or another shareable config, you can specify these packages as dependencies in the package.json.

codyebberson: Ahh, good catch!

mattwiller (approved):
✨✨✨✨✨

PR #2461, opened 2023-07-13
Load and validate profiles for resources that claim conformance

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -523,11 +526,16 @@ export class Repository extends BaseRepository implements FhirRepository {
...
-  private validate(resource: Resource): void {
+  private async validate(resource: Resource): Promise<void> {
     if (this.context.strictMode) {
       const start = process.hrtime.bigint();
+      const profileURLs = resource.meta?.profile;
       try {
-        validate(resource);
+        if (profileURLs) {
+          await this.validateProfiles(resource, profileURLs);
+        } else {
```
codyebberson, line 536: Hm, interesting that this is `else` and not both.

I suppose the operating principle here is that the FHIR Profile `StructureDefinition` is supposed to include everything in the base definition, and therefore validating against the base definition would be redundant?

For FHIR Profiles published by HL7/FHIR, that seems reasonable.  For user submitted profiles, that means we will need a mechanism of validating that the submitted profile is compatible with the base `StructureDefinition`, otherwise this could be a backdoor to all kinds of shenanigans.

packages/server/src/fhir/repo.ts
```diff
@@ -523,11 +526,16 @@ export class Repository extends BaseRepository implements FhirRepository {
     return result;
   }
 
-  private validate(resource: Resource): void {
+  private async validate(resource: Resource): Promise<void> {
     if (this.context.strictMode) {
       const start = process.hrtime.bigint();
+      const profileURLs = resource.meta?.profile;
```
codyebberson, line 532: Nit: we use lower case abbreviations `profileUrls`, `userId`, etc.

packages/server/src/fhir/repo.ts
```diff
@@ -557,6 +565,63 @@ export class Repository extends BaseRepository implements FhirRepository {
...
+          descending: true,
+        },
+      ],
+    });
+
+    if (profile) {
+      // Store loaded profile in cache
+      await redis.set(
+        cacheKey,
+        JSON.stringify({ resource: profile, projectId: profile.meta?.project }),
```
codyebberson, line 617: Is there a reason we need `projectId` in this object, considering it's already in the Redis cache key?  Doesn't look like we're using it.

packages/server/src/fhir/repo.ts
mattwiller: Yeah, in the expected case the `snapshot` of the profile should contain all base element definitions included as well.  I agree though, it's likely better to be defensive and pay the cost of double validating here rather than trying to bulletproof the process of adding new `StructureDefinition` for use in validation

packages/server/src/fhir/repo.ts
mattwiller: I had added it just for consistency with the other objects stored in cache, but I'm cool with removing it if it's not necessary to maintain that parity

packages/core/src/fhirpath/fhirpath.test.ts
```diff
@@ -2976,103 +2976,81 @@ describe('FHIRPath Test Suite', () => {
 
   describe('testBooleanLogicOr', () => {
     test('testBooleanLogicOr1', () => {
-      expect(evalFhirPath('(true or true) = true', patient)).toEqual([true]);
+      expect(evalFhirPath('(true or true)', patient)).toEqual([true]);
```
codyebberson, line 2979: All of these test strings were originally ported from a FHIRPath test suite.  My preference would be to leave them as-is unless there is good reason.

For example, this line originates from here: https://github.com/HL7/FHIRPath/blob/master/tests/r4/tests-fhir-r4.xml#L972

codyebberson (commented):
See comment on FHIRPath tests.  Otherwise, everything else looks good.  Exciting!

packages/core/src/fhirpath/fhirpath.test.ts
mattwiller: Ah, sounds good, I'll change it back; I just found it easier to debug this way

codyebberson (approved):
Awesome.  Ty @mattwiller!

reshmakh:

Issue #2469, opened 2023-07-14
Profiles Subcommands 


jamestouri:
`medplum profile set`
`medplum profile remove`
`medplum profile list`
`medplum profile describe`

reshmakh:
I believe that should be documented here @jamestouri - let me know if that sounds correct to you: https://www.medplum.com/docs/cli/external-fhir-servers

PR #2477, opened 2023-07-15
Profile Subcommands

jamestouri:
https://github.com/medplum/medplum/issues/2469

### DIFF
--- a/packages/cli/src/profiles.ts
+++ b/packages/cli/src/profiles.ts
@@ -42,7 +42,9 @@ listProfiles.description('All profiles saved').action(async () => {
     const fileName = file.split('.')[0];
     const storage = new FileSystemStorage(fileName);
     const profile = storage.getObject('profile');
-    allProfiles.push({ profileName: fileName, profile });
+    if (profile) {
+      allProfiles.push({ profileName: fileName, profile });
+    }
   });
   console.log(allProfiles);
 });
--- a/packages/cli/src/profiles.ts
+++ b/packages/cli/src/profiles.ts
@@ -1,0 +1,29 @@
+import { Command } from 'commander';
+import { createMedplumCommand } from './util/command';
+
+const setProfile = createMedplumCommand('set');
+const removeProfile = createMedplumCommand('remove');
+const listProfiles = createMedplumCommand('list');
+const describeProfile = createMedplumCommand('describe');
+
+export const profile = new Command('profile')
+  .addCommand(setProfile)
+  .addCommand(removeProfile)
+  .addCommand(listProfiles)
+  .addCommand(describeProfile);
+
+setProfile.argument('<profileName>', 'Name of the profile').action(async (profileName, options) => {
+  //   const medplum = await createMedplumClient(options);
+});
+
+removeProfile.argument('<profileName>', 'Name of the profile').action(async (profileName, options) => {
+  //   await createMedplumClient(options);
+});
+
+listProfiles.action(async (options) => {
+  // reading all the data from the json files
+});
+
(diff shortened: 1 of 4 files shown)


reshmakh:
Perhaps add some quick docs here on how to use?  Need not be comprehensive, I can touch it up - https://www.medplum.com/docs/cli/external-fhir-servers

reshmakh:
@jamestouri - got it

packages/cli/src/profiles.ts
```diff
@@ -0,0 +1,59 @@
...
+removeProfile
+  .argument('<profileName>', 'Name of the profile')
+  .description('Remove a profile by name')
+  .action(async (profileName) => {
+    const storage = new FileSystemStorage(profileName);
+    storage.setObject('profile', undefined);
+    console.log(`${profileName} profile removed`);
+  });
+
+listProfiles.description('All profiles saved').action(async () => {
```
codyebberson, line 37: Description should probably start with a verb, "Lists all profiles saved"

packages/cli/src/profiles.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (profile) {
+      allProfiles.push({ profileName: fileName, profile });
+    }
+  });
+  console.log(allProfiles);
+});
+
+describeProfile
+  .argument('<profileName>', 'Name of the profile')
+  .description('Properties of a profile')
```
codyebberson, line 54: Description should probably start with a verb, "Describes a profile"

packages/cli/src/util/command.ts
```diff
@@ -7,5 +7,6 @@ export function createMedplumCommand(name: string): Command {
     .option('--base-url <baseUrl>', 'FHIR server base url')
     .option('--token-url <tokenUrl>', 'FHIR server token url')
     .option('--authorize-url <authorizeUrl>', 'FHIR server authorize url')
-    .option('--fhir-url-path <fhirUrlPath>', 'FHIR server url path');
+    .option('--fhir-url-path <fhirUrlPath>', 'FHIR server url path')
+    .option('--auth-type <authType>', 'Type of authentication', 'basic');
```
codyebberson, line 11: Is this where we can list all auth types?  We should determine our list, and the exact strings we want to accept?

We have this enum, which might be a good starting point:

```ts
export enum OAuthGrantType {
  ClientCredentials = 'client_credentials',
  AuthorizationCode = 'authorization_code',
  RefreshToken = 'refresh_token',
  JwtBearer = 'urn:ietf:params:oauth:grant-type:jwt-bearer',
  TokenExchange = 'urn:ietf:params:oauth:grant-type:token-exchange',
}
```

So maybe:

- `basic` - Basic auth
- `client_credentials` - Client credentials flow
- `authorization_code` - Authorization code flow (default)
- `jwt-bearer` - User provided presigned JWT
- `token-exchange` - Preconfigured external auth provider

packages/cli/src/profiles.ts
```diff
@@ -0,0 +1,59 @@
...
+  .addCommand(removeProfile)
+  .addCommand(listProfiles)
+  .addCommand(describeProfile);
+
+setProfile
+  .argument('<profileName>', 'Name of the profile')
+  .description('Create a new profile or replace it with the given name and its associated properties')
+  .action(async (profileName, options) => {
+    const storage = new FileSystemStorage(profileName);
+    storage.setObject('profile', options);
```
codyebberson, line 24: Thoughts on whether this property should be called "profile" vs "options" vs "config" vs ???

jamestouri: setting to "options"

codyebberson (approved):

PR #2480, opened 2023-07-17
Bot editor improvements

codyebberson:
* Removed absolute positioning
* Responsive layout
* Buttons now have loading state
* Fixed "Execute" success toast
* Fixed unnecessary React rerenders

Before:

image

After:

https://github.com/medplum/medplum/assets/749094/010780bb-897f-422d-bb99-90194c5617d0


reshmakh (approved):

rahul1:
You didn't mention it in your PR description, but I love the new button icons. 

I think global "loading" state makes sense to avoid confusion. If someone presses "execute" while deploying, it's ambiguous which Bot version will actually run. 

All the overflow/scrollbar/padding stuff is a huge improvement. 

I'm not sure I agree with having the Execute button in a different spot. I see what you're going for (execute _this data_), but I think it might read better to just have all 3 action buttons grouped, since they are all part of a single process (Save -> Deploy -> Execute)

codyebberson:
> I think global "loading" state makes sense to avoid confusion. If someone presses "execute" while deploying, it's ambiguous which Bot version will actually run.

K, I'll add.

There will still be some ambiguity.  Unfortunately lambdas take 5-10 seconds to fully propagate after deploying, so there is a period of uncertainty.  We could explore using the lambda "version" feature, but that is out of scope for this PR.

> I'm not sure I agree with having the Execute button in a different spot. I see what you're going for (execute _this data_), but I think it might read better to just have all 3 action buttons grouped, since they are all part of a single process (Save -> Deploy -> Execute)

Yeah... I'm mixed on it.  I can move it back.  It's based on the [Pipedream](https://pipedream.com/) UI, which I really like.

rahul1:
> My thinking is that these are not all part of a single process. "Save" and "Deploy" - yes, that's how you change the bot. The "Execute" button is more for testing, experimentation, running -- all of which is separate from change the bot.

Yeah good point, I agree with that logic. Given that, I'm really fine either way - I don't feel strongly about that button placement

rahul1:
> There will still be some ambiguity. Unfortunately lambdas take 5-10 seconds to fully propagate after deploying, so there is a period of uncertainty. We could explore using the lambda "version" feature, but that is out of scope for this PR.

ACK. Global seems directionally better, but yeah there's still a delay that's unaccounted for

Issue #2317, opened 2023-06-27
Add the ability to log and receive alerts for Auth0 requests failed 


reshmakh:
This logging/alerting setup should utilize any solution prepared for #2314.

PR #2487, opened 2023-07-17
Use 429 status for too many external auth requests

jamestouri:
https://github.com/medplum/medplum/issues/2317

### DIFF
--- a/packages/server/src/oauth/token.test.ts
+++ b/packages/server/src/oauth/token.test.ts
@@ -1496,7 +1496,7 @@ describe('OAuth2 Token', () => {
     });
     expect(res.status).toBe(429);
     expect(res.body.error).toBe('invalid_request');
-    expect(res.body.error_description).toBe('Failed to verify code - check your identity provider configuration');
+    expect(res.body.error_description).toBe('Too Many Requests');
   });
 
 
--- a/packages/server/src/oauth/token.test.ts
+++ b/packages/server/src/oauth/token.test.ts
@@ -1484,7 +1484,6 @@ describe('OAuth2 Token', () => {
   test('Too many requests', async () => {
     (fetch as unknown as jest.Mock).mockImplementation(() => ({
       status: 429,
-      son: () => ({ email }),
       headers: { get: () => 'application/json' },
     }));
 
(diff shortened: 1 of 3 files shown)


packages/server/src/oauth/token.ts
```diff
@@ -352,8 +352,10 @@ export async function exchangeExternalAuthToken(
   let userInfo;
   try {
     userInfo = await getExternalUserInfo(idp, subjectToken);
-  } catch (err) {
-    sendTokenError(res, 'invalid_request', normalizeErrorString(err));
+  } catch (err: any) {
+    // If the status is 429, then the error message will be Too Many Requests
+    const status = err.message === 'Too Many Requests' ? 429 : 400;
```
codyebberson, line 357: This is kinda brittle.  We have some helper utilities that might be good here:

* `function normalizeOperationOutcome(error: unknown): OperationOutcome`
* `function operationOutcomeToString(outcome: OperationOutcome): string`
* `function getStatus(outcome: OperationOutcome): number`

```ts
} catch (err) {
    const outcome = normalizeOperationOutcome(err);
    sendTokenError(res, 'invalid_request', operationOutcomeToString(outcome), getStatus(outcome));
}
```
}

packages/server/src/oauth/token.test.ts
```diff
@@ -1481,6 +1481,24 @@ describe('OAuth2 Token', () => {
     expect(res.body.error_description).toBe('Failed to verify code - check your identity provider configuration');
   });
 
+  test('Too many requests', async () => {
+    (fetch as unknown as jest.Mock).mockImplementation(() => ({
+      status: 429,
+      son: () => ({ email }),
```
mattwiller, line 1487:
```suggestion
      json: () => ({ email }),
```

packages/server/src/oauth/utils.ts
```diff
@@ -703,6 +704,11 @@ export async function getExternalUserInfo(
     throw new OperationOutcomeError(badRequest('Failed to verify code - check your identity provider configuration'));
   }
 
+  if (response.status === 429) {
+    logger.warn('Too many requests sent in a given amount of time', response.status);
```
mattwiller, line 708: This log message is more verbose than necessary; you could consider making it shorter and more information-dense like e.g. "Auth rate limit exceeded".  Also, adding `response.status` to the log probably isn't very useful, since it's guaranteed to always be `429` here

packages/server/src/oauth/token.test.ts
mattwiller: This line could probably also just be removed, since the test doesn't seem to be using it

codyebberson (approved):

PR #2500, opened 2023-07-19
First draft of agent

codyebberson:
This is ready for review, although it's very early and not particularly useful yet.

Next steps (out of scope for this PR, but sharing for visibility, and open to feedback):

- Configuration settings
    - Need to make decisions here
    - Simple option is to use local JSON file
    - More complex option is to use cloud persisted settings, i.e., new Medplum resource type
    - Simple option is to only support one port and one channel
    - More complex option is to allow multiple ports mapped to multiple channels
- Connect to Medplum server via websockets
- Enable sending control messages to agent
  - Start with simple ping to test connectivity
- Pass HL7 messages to configured Bot
- Return Bot response as HL7 message back to the edge device
- Package with [`vercel/pkg`](https://github.com/vercel/pkg) to create single file binary
- Sign with Microsoft [SignTool](https://learn.microsoft.com/en-us/windows/win32/seccrypto/signtool)


rahul1:
LGTM for the most part, but I'm a bit confused on the split between the `agent` and the `hl7` packages. Might make sense with your future planned work. It seems like the `agent` package is a thin wrapper that reads configs, sets up the server. 

Might it make sense to move the socket connection work into the agent, and then pass them into the HL7 server? This might be premature optimization, but it would potentially allow the HL7 server to listen on via other mechanisms (e.g. queues) in the future.

rahul1 (approved):

codyebberson:
> bit confused on the split between the `agent` and the `hl7` packages

Good question.

We have one-off agents custom developed for Medplum customers.  Some of those agents use the `simple-hl7` library for the HL7 TCP layer.  The `simple-hl7` library has worked well and is stable, but is not actively developed, does not have TypeScript definitions.  By publishing `@medplum/hl7`, we would be able to 100% replace `simple-hl7`.

Issue #2490, opened 2023-07-18
Medplum commands for basic auth


jamestouri:
`medplum get Patient` 
`medplum get Patient --profile <profileName>`

`medplum post Patient {patientBody}`
`medplum post Patient {patientBody} --profile <profileName>`

`medplum bulk export Patient --profile <profileName>`

PR #2502, opened 2023-07-19
Profile flags 

jamestouri:
For `get`, `post`, and `bulk export` 

Based on the commands here https://www.medplum.com/docs/cli/external-fhir-servers

https://github.com/medplum/medplum/issues/2490

For bulk export, I ran:

```
ts-node ~/Developer/medplum/packages/cli/src/index.ts profile set bcda --client-id 2462c96b-6427-4efb-aed7-118e20c2e997 --client-secret 8e87f0ebc50d10f1bc9734329a9900179b84ccd39e4d0920b905cc359cf6e94a6e760bbe3a0890c7 --base-url https://sandbox.bcda.cms.gov --fhir-url-path api/v2/ --token-url https://sandbox.bcda.cms.gov/auth/token
```

```
ts-node ~/Developer/medplum/packages/cli/src/index.ts bulk export -e Patient -p bcda
```

Docs:
Screenshot 2023-07-25 at 9 10 56 AM
Screenshot 2023-07-25 at 9 11 02 AM
Screenshot 2023-07-25 at 9 11 10 AM
Screenshot 2023-07-25 at 9 11 17 AM


jamestouri:
Will add documentation before changing this to "Ready for review"

reshmakh:
Not clear what to put for auth-type if anything, and that seems like an FAQ.  

Also, alignment on the docs seems off - maybe create an Example heading and then have the three examples as sub headings?

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,87 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
...
+- --auth-type
+- --client-id
+- --client-secret
+- --base-url
+- --token-url
+- --authorize-url
+- --fhir-url-path
+
+```bash
+medplum profile set <profileName> --base-url <base-url> --fhir-url-path <fhir-url-path> --token-url <token-url> --client-id <client-id> --client-secret <client-secret>
```
codyebberson, line 39: Let's call this "Syntax"

Can we use backslashes here to make this a multiline string?

```bash
medplum profile set <profileName> \
    --base-url <base-url> \
    --fhir-url-path <fhir-url-path> \
    --token-url <token-url> \
    --client-id <client-id> \
    --client-secret <client-secret>
```

And then we can have another block labeled "Example" with dummy values.

I think @reshmakh  is correct, that we will want clearly labeled examples for each of the main auth methods.

----

Example: Basic Auth

```bash
medplum profile set "example" \
    --base-url "https://api.example.com" \
    --fhir-url-path "fhir/R4" \
    --auth-type "basic" \
    --client-id "MY_CLIENT_ID" \
    --client-secret "MY_CLIENT_SECRET"
```

Example: Client Credentials

```bash
medplum profile set "example" \
    --base-url "https://api.example.com" \
    --fhir-url-path "fhir/R4" \
    --token-url "oauth2/token" \
    --auth-type "client_credentials" \
    --client-id "MY_CLIENT_ID" \
    --client-secret "MY_CLIENT_SECRET"
```

Example: User Login / Authorization Code

```bash
medplum profile set "example" \
    --base-url "https://api.example.com" \
    --fhir-url-path "fhir/R4" \
    --authorize-url "oauth2/authorize" \
    --token-url "oauth2/token"
```

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,87 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
 
 The examples below use the [CLI optional flags](/docs/cli#optional-flags).
 
+## Setting up your credentials
+
+Medplum CLI stores credentials to be used in a future period without needing it to be entered in every command. By using the `profile` command, this helps with the ability to work with multiple FHIR servers.
+
+### Example: Setting a Profile
```
codyebberson, line 26: Let's remove the word "Example:" here, and just have a section called "Setting a Profile"

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,87 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
...
+```bash
+medplum profile set <profileName> --base-url <base-url> --fhir-url-path <fhir-url-path> --token-url <token-url> --client-id <client-id> --client-secret <client-secret>
+```
+
+The profile will now be stored in a file directory in `~.medplum/<profileName>.json`
+
+Once you have a profile, you can connect with external FHIR servers with your profile using the `-p` flag. 
+
+
+Other profile commands include:
```
codyebberson, line 47: Add blank line before headings

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,87 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
...
+#### `describe`
+
+To see the state of your credentials in on profile 
+
+Example:
```
codyebberson, line 52: For each of these, let's do 2 blocks:

--------

Syntax:

```bash
medplum profile describe <profileName>
```

Example:

```bash
medplum profile describe "example"
```

--------

I know it's kinda verbose, but I think it's more explicit and technically correct.

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,87 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
...
 
+## Setting up your credentials
+
+Medplum CLI stores credentials to be used in a future period without needing it to be entered in every command. By using the `profile` command, this helps with the ability to work with multiple FHIR servers.
+
+### Example: Setting a Profile
+
+In this example, we will set up a profile using `medplum profile set <profileName>` with the flags below:
+
+- --auth-type
```
codyebberson, line 30: Agree with @reshmakh  - we should list out possible values for `--auth-type`.  I'm not sure if we should do it here, or put a table below.

packages/cli/src/utils.ts
```diff
@@ -180,3 +181,15 @@ export function getUnsupportedExtension(): Extension {
     ],
   };
 }
+
+export function checkForProfile(options: any): boolean {
```
codyebberson, line 185: Can we use a type here rather than `any`?

packages/cli/src/bulk.ts
```diff
@@ -28,9 +28,13 @@ bulkExportCommand
     'optional target directory to save files from the bulk export operations.'
   )
   .action(async (options) => {
-    const { exportLevel, types, since, targetDirectory } = options;
-    const medplum = await createMedplumClient(options);
+    if (!checkForProfile(options)) {
```
codyebberson, line 31: Can this `checkForProfile` call be done inside `createMedplumClient`?  That way we don't need to duplicate it on every command.  `createMedplumClient` can throw if the options are invalid.

packages/cli/src/util/client.ts
```diff
@@ -31,18 +29,22 @@ export async function createMedplumClient(options: MedplumClientOptions, profile
 }
 
 function getClientValues(options: MedplumClientOptions, storage: FileSystemStorage): Record<string, any> {
```
codyebberson, line 31: Is the return type of this method `MedplumClientOptions`?  Let's try to use types if they're available.

codyebberson (changes requested):

packages/cli/src/util/client.ts
```diff
@@ -20,31 +25,40 @@ export async function createMedplumClient(options: MedplumClientOptions, profile
...
-  return { baseUrl, fhirUrlPath, accessToken, tokenUrl, authorizeUrl, fetchApi };
+  return { baseUrl, fhirUrlPath, accessToken, tokenUrl, authorizeUrl, clientId, clientSecret };
 }
 
 export function onUnauthenticated(): void {
   console.log('Unauthenticated: run `npx medplum login` to sign in');
 }
+
+function checkForProfile(storage: FileSystemStorage, profile: string): void {
+  if (profile === 'default') {
```
jamestouri, line 57: Put the check for profile here, and skip if the profile is 'default'

packages/cli/src/auth.ts
```diff
@@ -13,6 +13,10 @@ export const whoami = createMedplumCommand('whoami');
 
 login.action(async (options) => {
   const medplum = await createMedplumClient(options);
+  if (options.authType && options.authType === 'basic') {
```
codyebberson, line 16: The first check is redundant with the 2nd.

packages/cli/src/util/client.ts
```diff
@@ -20,31 +25,40 @@ export async function createMedplumClient(options: MedplumClientOptions, profile
...
   if (clientId && clientSecret) {
     medplumClient.setBasicAuth(clientId, clientSecret);
     await medplumClient.startClientLogin(clientId, clientSecret);
   }
   return medplumClient;
 }
 
-function getClientValues(options: MedplumClientOptions, storage: FileSystemStorage): Record<string, any> {
+function getClientValues(options: MedplumClientOptions, storage: FileSystemStorage): MedplumClientOptions {
+  const storageOptions: any = storage.getObject('options');
```
codyebberson, line 36: Can we use `MedplumClientOptions` for the type here?

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,147 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
...
+    --base-url <base-url> \
+    --fhir-url-path <fhir-url-path> \
+    --token-url <token-url> \
+    --client-id <client-id> \
+    --client-secret <client-secret>
+```
+
+| Accepted Auth Type   |
+|--------------------|
+| basic            |
```
codyebberson, line 52: Can we format this table?  VS Code / prettier should fix it for you.

packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -19,34 +19,147 @@ To enable these scenarios, you will need a `clientId` or `clientSecret` to acces
...
+    --token-url "oauth2/token" \
+    --client-id "MY_CLIENT_ID" \
+    --client-secret "MY_CLIENT_SECRET"
+```
+
+### Example: User Login / Authorization Code
+
+```bash
+medplum profile set example \
+    --auth-type "authorization_code" \
```
codyebberson, line 87: This is the default value for `--auth-type`, correct?  So it can be omitted?

packages/cli/src/util/client.ts
codyebberson: Yep, this works 👍

codyebberson (commented):
One last thing, then I'll stop, I promise

packages/cli/src/util/client.ts
```diff
@@ -4,8 +4,13 @@ import { FileSystemStorage } from '../storage';
 export async function createMedplumClient(options: MedplumClientOptions, profileName?: string): Promise<MedplumClient> {
```
codyebberson, line 4: Sorry, I wish I would have caught this earlier -- In all of the calls to `createMedplumClient`, it looks like the `profileName` parameter always comes from `options`, so we should be able to just use `options.profile` here, right?  And remove the `profileName` parameter entirely?

packages/core/src/client.ts
```diff
@@ -172,6 +172,15 @@ export interface MedplumClientOptions {
...
   fetch?: FetchLike;
 
+  /**
+   * The profile name.
+   *
+   * This is for handling multiple profiles to access different FHIR Servers
+   *
+   * If undefined, the default profile is "default".
+   */
+  profile?: string;
```
jamestouri, line 182: Reason I had it as another parameter in createMedplumClient was because it was of MedplumClientOptions, and didn't think it was good to continue adding on to the type. After your review I added it here

packages/cli/src/util/command.ts
```diff
@@ -8,11 +8,12 @@ export function createMedplumCommand(name: string): Command {
     .option('--token-url <tokenUrl>', 'FHIR server token url')
     .option('--authorize-url <authorizeUrl>', 'FHIR server authorize url')
     .option('--fhir-url-path <fhirUrlPath>', 'FHIR server url path')
+    .option('-p, --profile <profile>', 'Profile name')
     .addOption(
       new Option('--auth-type <authType>', 'Type of authentication').choices([
         'basic',
-        'client_credentials',
-        'authorization_code',
+        'client-credentials',
```
jamestouri, line 15: Cody here are the changes

codyebberson (approved):

PR #2503, opened 2023-07-19
Move medplum-nextjs-demo to examples folder

jamestouri:
https://github.com/medplum/medplum/issues/2509


examples/medplum-nextjs-demo/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@emotion/react": "11.11.1",
+    "@emotion/server": "11.11.0",
+    "@mantine/core": "6.0.17",
```
codyebberson, line 15: `@mantine` dependences in `packages/app` and `packages/react` are all on `6.0.16` - can we align for now?  We'll upgrade them all at once in the next dependency upgrade.

examples/medplum-nextjs-demo/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "@mantine/hooks": "6.0.17",
+    "@mantine/next": "6.0.17",
+    "@mantine/notifications": "6.0.17",
+    "@medplum/core": "2.0.26",
+    "@medplum/react": "2.0.26",
+    "@tabler/icons": "2.26.0",
```
codyebberson, line 21: Not sure if this is correct?  When `@tabler/icons` went from v1 to v2, the package name changed to `@tabler/icons-react`.

Our other packages are on `"@tabler/icons-react": "2.25.0",`

codyebberson (commented):
See comments on dependency versions.  Otherwise lgtm 👍

examples/medplum-nextjs-demo/package.json
jamestouri: That's odd, definitely not correct. It didn't show an error when I ran `npm run build` and `npm run dev`. I ended up removing it and it still runs just fine, maybe we don't need that dependency? Let me know otherwise and I'll put it back.

codyebberson: Oh, heh, whoops.  I bet that's just a copy/paste error from the initial setup.  My bad!  Thanks for cleaning it up!

codyebberson (approved):

PR #2505, opened 2023-07-20
Use built-in EventTarget

codyebberson:
Extracted changes out of this PR: https://github.com/medplum/medplum/pull/2500

When the Medplum project started, the `EventTarget` constructor was not widely available.

Now it is: 
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget
- https://caniuse.com/mdn-api_eventtarget_eventtarget

Ripping out our version and using the built-in.


mattwiller (approved):
Nice!

PR #2512, opened 2023-07-21
Clarify error message when modifying read-only resources

rahul1:
User driven: https://discord.com/channels/905144809105260605/1131288083078328471

User was trying to submit a PUT and PATCH request to a resource with a "read-only" access policy and received a `403: Forbidden` error. Our docs didn't not explicitly define what the expected result would be.


codyebberson (approved):
lgtm 👍

PR #2522, opened 2023-07-21
Optimize bundles during batch upload

rahul1:
User driven: https://discord.com/channels/905144809105260605/905144809105260608/1131376049305948240

The user wanted to upload [this bundle](https://hl7.org/fhir/us/cancer-reporting/STU1/Bundle-us-pathology-content-bundle-example.json.html) using our batch upload page. However, this Bundle is of type `collection`, and isn't set up with proper references for ingestion.

This PR does two things to make the batch upload tool more useful: 
- On the client, it converts non-input bundles into `transaction` type bundles using our `convertToTransactionBundle` function
- It augments the `convertToTransactionBundle` with a topological sort, to make sure that resources are created before they are referenced. This will help more gracefully handle mal-ordered bundles

I've tested this PR with the bundle above. This should open us up to utilize much more example data from FHIR implementation guides


packages/core/src/bundle.ts
```diff
@@ -11,10 +13,21 @@ import { Bundle } from '@medplum/fhirtypes';
...
+  const idToUuid: Record<string, string> = {};
+  bundle = deepClone(bundle);
   for (const entry of bundle.entry || []) {
     delete entry.resource?.meta;
-    entry.fullUrl = 'urn:uuid:' + entry.resource?.id;
-    delete entry.resource?.id;
+    const id = entry.resource?.id;
+    // In some FHIR example bundles,they use human-readable ids, rather than true uuids.
+    if (id) {
+      if (isUUID(id)) {
```
codyebberson, line 23: We might as well skip the `isUUID` check and just always use `idToUuid`

packages/core/src/bundle.ts
```diff
@@ -11,10 +13,21 @@ import { Bundle } from '@medplum/fhirtypes';
...
+        idToUuid[id] = id;
+      } else {
+        idToUuid[id] = crypto.randomUUID();
```
codyebberson, line 26: This works in browser environments, but not in NodeJS environments.  

We have a cross-environment version in `fhir-router`: https://github.com/medplum/medplum/blob/main/packages/fhir-router/src/repo.ts#L371

We could move that into `core` if we need to.

packages/core/src/bundle.ts
```diff
@@ -27,15 +40,164 @@ export function convertToTransactionBundle(bundle: Bundle): Bundle {
...
+ * Topologically sorts a `batch` or `transaction` bundle to improve reference resolution.
+ * The bundle is sorted such that a resource is created _before_ references to that resource appear in the bundle.
+ *
+ * In the event of cycles, this function will first create a POST request for each resource in the cycle, and then will
+ * append a PUT request to the bundle. This ensures that each resources in the cycle is visited twice, and all
+ * references can be resolved
+ * @param bundle Input bundle with type `batch` or `transaction`
+ * @returns Bundle of the same type, with Bundle.entry reordered
+ */
+export function reorderBundle(bundle: Bundle): Bundle {
```
codyebberson, line 67: Ohhh boy.... Let's discuss this offline.  We've tried this in the past.  I'm slightly against adding this.

codyebberson: K, discussed offline, let's do it 👍

rahul1:
@codyebberson ready for re-review

Issue #2489, opened 2023-07-18
Login through profiles 


jamestouri:
Ability to manage multiple clients. 

When running the 'login' command with the --profile flag, it will also set a profile

PR #2535, opened 2023-07-25
Login with JWT and Token Exchange Profiles

jamestouri:
https://github.com/medplum/medplum/issues/2489


packages/docs/docs/cli/external-fhir-servers.md
```diff
@@ -44,14 +44,19 @@ medplum profile set <profileName> \
...
+    --scope <scope> \
+    --access-token <access-token>
 ```
 
 | Accepted Auth Type |
 | ------------------ |
 | basic              |
 | client_credentials |
 | authorization_code |
+| jwt-bearer         |
```
reshmakh, line 58: why are some dashes (-) and some underscore (_) - are those industry standard things?

jamestouri: Ahh good point with normalizing them. Dashes are more accepted with the CLI so I'll make changes for all of them to be that.

Issue #2545, opened 2023-07-26
Document /admin/{projectId}/client endpoint 


rahul1:
User Driven: https://medplum.slack.com/archives/C04A55B3VU5/p1690215065195849?thread_ts=1690215065.195849&cid=C04A55B3VU5

PR #2575, opened 2023-07-28
Reference documentation for all /admin/projects/... endpoints

rahul1:


rahul1:
Fixes #2545

packages/docs/docs/api/project-admin/members.md
```diff
@@ -0,0 +1,218 @@
...
+    reference: 'User/:userId',
+  },
+  profile: {
+    reference: 'Practitioner/:profileId',
+    display: 'George Washington'
+  },
+}
+```
+
+## POST `/admin/projects/:projectId/:membershipId`
```
codyebberson, line 57: Should be `/admin/projects/:projectId/members/:membershipId`

packages/docs/docs/api/project-admin/members.md
```diff
@@ -0,0 +1,218 @@
...
+    reference: 'User/:userId',
+  },
+  profile: {
+    reference: 'Practitioner/:profileId',
+    display: 'George Washington'
+  }
+}
+```
+
+## DELETE `/admin/projects/:projectId/:membershipId`
```
codyebberson, line 175: Should be `/admin/projects/:projectId/members/:membershipId`

codyebberson (commented):
Looks good

Although... I wonder if we should actually formally deprecate the following endpoints: 
- `/admin/projects/:projectId`
- `/admin/projects/:projectId/:membershipId`
- `/admin/projects/:projectId/secrets`
- `/admin/projects/:projectId/sites`

Before, project admins did not have access to the underlying `Project` and `ProjectMembership` resources, so you had to use those specialty endpoints.  Now all of it can be done with `/fhir/R4/Project` and `/fhir/R4/ProjectMembership`, which is much cleaner and easier to understand, works with search, works with batch, works with graphql, etc.

If we do that, you could delete the following files:

- `project-admin/members.md`
- `project-admin/project.md`
- `project-admin/secrets.md`
- `project-admin/sites.md`

The other endpoints are good and necessary, because they do heavy lifting of creating memberships, setting up permissions, etc.

Your call if we should document them now, and clean it all up later, or if we should start deprecating now.

Otherwise, 2 small incorrect URL paths.

rahul1:
If we don't need them, lets' formally deprecate. I'll remove the docs

codyebberson (approved):

PR #2578, opened 2023-07-31
Stream HL7 over websockets

codyebberson:
Implementation details discussed offline.

I added warnings to README and to docs that this is "alpha".

https://github.com/medplum/medplum/blob/ccd18ce77593b3fbda963e8c8f5e22d3a55481d7/packages/agent/README.md

https://medplum-www-git-cody-stream-hl7-over-websocket-medplum.vercel.app/docs/agent


reshmakh:
Related: https://github.com/medplum/medplum/discussions/1490

packages/agent/src/main.ts
```diff
@@ -1,25 +1,24 @@
...
 
   constructor(readonly medplum: MedplumClient, readonly bot: Bot) {
     this.log = new EventLogger({
       source: 'MedplumService',
       eventLog: 'SYSTEM',
     });
 
-    this.server = new Hl7Server();
-    this.server.addEventListener('message', (event) => this.handler(event));
+    this.server = new Hl7Server((connection) => {
```
codyebberson, line 18: I changed this from `addEventListener` to a callback, because the `Hl7Server` would receive messages before the event listener attached.  That meant the first message would often get dropped.  By using a callback, it becomes blocking, and the first message is not missed.

Alternatively, we could add message buffering to `Hl7Server`, but that has a bunch of tricky design trade offs (what does it mean to have multiple connection handlers?)

packages/hl7/src/client.ts
```diff
@@ -13,75 +12,44 @@ export class Hl7Client extends Hl7Base {
...
-          this.buffer += data.toString();
-          if (this.buffer.endsWith(FS + CR)) {
-            const msg = Hl7Message.parse(this.buffer.substring(1, this.buffer.length - 2));
-            this.dispatchEvent(new Hl7MessageEvent(socket, msg));
-            this.buffer = '';
-            this.awaitingResponse = false;
-          }
-        });
-        resolve(socket);
+        this.connection = new Hl7Connection(socket);
```
codyebberson, line 31: There was a lot of duplicate code between `Hl7Server` and `Hl7Client`.  I refactored it out to `Hl7Connection`, which seems like a better abstraction, and leads to better readability (IMHO).

reshmakh:
https://github.com/medplum/medplum/discussions/1490

reshmakh (approved):

PR #2586, opened 2023-08-01
Fix logging of some error messages

mattwiller:


packages/server/src/admin/invite.ts
```diff
@@ -54,7 +54,7 @@ export async function inviteHandler(req: Request, res: Response): Promise<void>
   } catch (err: any) {
     logger.info('Error inviting user to project', {
       project: projectId,
-      error: err,
+      error: err.toString(),
```
codyebberson, line 57: You may consider using `normalizeErrorString`: https://github.com/medplum/medplum/blob/main/packages/core/src/outcomes.ts#L276

codyebberson (approved):

PR #2587, opened 2023-08-01
JWT Bearer Login for CLI 

jamestouri:
Ran 

```
ts-node ~/Developer/medplum/packages/cli/src/index.ts login -p the-bearer \ 
--auth-type jwt-bearer \
--client-id {{HEALTH_GORILLA_CLIENT}} \ 
--client-secret {{HEALTH_GORILLA_SECRET}} \
--base-url "https://sandbox.healthgorilla.com/" \
--audience "oauth/token" \
--token-url "/oauth/token" \ 
--subject {{HEALTH_GORILLA_SUBJECT}} \
--issuer {{HEALTH_GORILLA_CLIENT_URI }}
```
Returned and stored access token

Documentation:
Screenshot 2023-08-01 at 10 07 02 AM
Screenshot 2023-08-01 at 10 06 48 AM

### DIFF
--- a/packages/cli/src/auth.ts
+++ b/packages/cli/src/auth.ts
@@ -1,9 +1,13 @@
-import { ContentType, getDisplayString, MedplumClient, normalizeErrorString } from '@medplum/core';
+import { ContentType, MedplumClient, encodeBase64, getDisplayString, normalizeErrorString } from '@medplum/core';
 import { exec } from 'child_process';
 import { createServer } from 'http';
 import { platform } from 'os';
 import { createMedplumClient } from './util/client';
 import { createMedplumCommand } from './util/command';
+import { Profile, createProfile, getProfileOptions } from './utils';
+import { FileSystemStorage } from './storage';
+import { ClientApplication } from '@medplum/fhirtypes';
+import { createHmac } from 'crypto';
 
 const clientId = 'medplum-cli';
 const redirectUri = 'http://localhost:9615';
@@ -12,12 +16,17 @@ export const login = createMedplumCommand('login');
 export const whoami = createMedplumCommand('whoami');
 
 login.action(async (options) => {
-  const medplum = await createMedplumClient(options);
+  if (!isThereExistingProfileName(options.profile)) {
+    console.log(`Creating new profile...`);
+    createProfile(options.profile, options);
+  }
   if (options.authType === 'basic') {
     console.log('Basic authentication does not require login');
     return;
   }
-  await startLogin(medplum);
+  const medplum = await createMedplumClient(options);
+  const profile = getProfileOptions(options.profile);
+  await startLogin(medplum, profile);
 });
 
 whoami.action(async (options) => {
@@ -25,15 +34,19 @@ whoami.action(async (options) => {
   printMe(medplum);
 });
 
-async function startLogin(medplum: MedplumClient): Promise<void> {
-  await startWebServer(medplum);
-  const loginUrl = new URL(medplum.getAuthorizeUrl());
-  loginUrl.searchParams.set('client_id', clientId);
-  loginUrl.searchParams.set('redirect_uri', redirectUri);
-  loginUrl.searchParams.set('scope', 'openid');
-  loginUrl.searchParams.set('response_type', 'code');
-  loginUrl.searchParams.set('prompt', 'login');
-  await openBrowser(loginUrl.toString());
+async function startLogin(medplum: MedplumClient, profile: Profile): Promise<void> {
+  if (!profile?.authType) {
+    await medplumAuthorizationCodeLogin(medplum);
+    return;
+  }
+  if (profile.authType === 'jwt-bearer') {
+    if (!profile.clientId || !profile.clientSecret) {
+      throw new Error('Missing values, make sure to add --client-id, --assertion, --scope for JWT login');
+    }
+    console.log('Starting JWT login...');
+    const clientApplication = await createClient(medplum, profile);
+    await jwtBearerLogin(medplum, profile, clientApplication);
+  }
 }
 
 async function startWebServer(medplum: MedplumClient): Promise<void> {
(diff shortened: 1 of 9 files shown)


packages/cli/src/util/command.ts
```diff
@@ -8,7 +8,13 @@ export function createMedplumCommand(name: string): Command {
...
-    .option('-p, --profile <profile>', 'Profile name')
+    .option('--scope <scope>', 'JWT scope')
+    .option('--access-token <accessToken>', 'Access token for token exchange authentication')
+    .option('--callback-url <callbackUrl>', 'Callback URL for authorization code flow')
+    .option('--subject <subject>', 'Subject for JWT authentication')
+    .option('--audience <audience>', 'Audience for JWT authentication')
+    .option('--issuer, <issuer>', 'Issuer for JWT authentication')
+    .option('-p, --profile <name>', 'Profile name')
     .addOption(
       new Option('--auth-type <authType>', 'Type of authentication').choices([
```
jamestouri, line 19: Perhaps change this from 'jwt-bearer' to 'health-gorilla'?

packages/cli/src/auth.ts
```diff
@@ -97,3 +111,64 @@ function printMe(medplum: MedplumClient): void {
...
+  const loginUrl = new URL(medplum.getAuthorizeUrl());
+  loginUrl.searchParams.set('client_id', clientId);
+  loginUrl.searchParams.set('redirect_uri', redirectUri);
+  loginUrl.searchParams.set('scope', 'openid');
+  loginUrl.searchParams.set('response_type', 'code');
+  loginUrl.searchParams.set('prompt', 'login');
+  await openBrowser(loginUrl.toString());
+}
+
+function isThereExistingProfileName(profileName?: string): boolean {
```
codyebberson, line 126: Let's use `profileExists` rather than `isThereExistingProfileName`

How is that different than `checkForProfile` in `packages\cli\src\util\client.ts`?

When creating util methods, we should make sure they are tight, single purpose, and clearly named.  We've been evolving these utils, and it may make sense to step back and come up with a clear set of helpers:

* `loadProfile` or `readProfile`
* `saveProfile` or `writeProfile`
* `profileExists`

It looks like you cleaned up most cases where we're calling `new FileSystemStorage(profileName)` (good!), but maybe not all?

packages/cli/src/util/client.ts
```diff
@@ -1,7 +1,11 @@
 import { MedplumClient, MedplumClientOptions } from '@medplum/core';
 import { FileSystemStorage } from '../storage';
+import { Profile } from '../utils';
 
-export async function createMedplumClient(options: MedplumClientOptions): Promise<MedplumClient> {
+export async function createMedplumClient(
+  options: MedplumClientOptions,
+  profileObject?: Profile
```
codyebberson, line 7: We should not use `Object` as a variable name suffix.  Let's just call this `profile?: Profile`

packages/cli/src/profile-auth.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    callback();
+  }),
+}));
+
+const testHomeDir = mkdtempSync(__dirname + sep + 'storage-');
+
+const originalWindow = globalThis.window;
+
+describe('Profiles Auth', () => {
+  beforeEach(async () => {
```
codyebberson, line 27: Delete empty block

packages/cli/src/utils.ts
```diff
@@ -192,3 +210,15 @@ export function getCodeContentType(filename: string): string {
   }
   return ContentType.TEXT;
 }
+
+export function createProfile(profileName: string, options: any): void {
```
codyebberson, line 214: See comment above - let's use consistent naming strategy here with `saveProfile` or `writeProfile`

Also, let's use the `Profile` type rather than `any`

packages/cli/src/utils.ts
```diff
@@ -192,3 +210,15 @@ export function getCodeContentType(filename: string): string {
...
+  const storage = new FileSystemStorage(profileName);
+  const optionsObject = { name: profileName, ...options };
+  storage.setObject('options', optionsObject);
+  console.log(`${profileName} profile created`);
+}
+
+export function getProfileOptions(profileName: string): any {
```
codyebberson, line 221: See comment above - let's use consistent naming strategy here with `loadProfile` or `readProfile`

Also, let's use the `Profile` for return type rather than `any`

packages/cli/src/util/command.ts
codyebberson: I'd like to avoid that unless we're absolutely convinced that this only works for health gorilla.

packages/cli/src/util/client.ts
```diff
@@ -1,7 +1,11 @@
...
+): Promise<MedplumClient> {
   const profile = options.profile ?? 'default';
```
codyebberson, line 9: This seems high risk of tying us in knots, because we have 2 separate arguments that specify profile, and it's not clear which one we're using.

I think we should probably be clear here whether we're using the `profile` argument or `options.profile`.

My recommendation would be something like this:

```ts

if (!profile) {
    profile = loadProfile(options.profile ?? 'default'); // or getProfile or readProfile or whatever
}
```

packages/cli/src/util/client.ts
```diff
@@ -25,7 +29,7 @@ export async function createMedplumClient(options: MedplumClientOptions): Promis
     medplumClient.setAccessToken(accessToken);
   }
 
-  if (clientId && clientSecret) {
+  if (clientId && clientSecret && !(profileObject?.authType === 'jwt-bearer')) {
```
codyebberson, line 32: This block was always a kludge.  It's incorrect.

I think the original intent was for it to be `profile.authType === 'client_credentials'`

jamestouri: I think it's still needed for basic so I changed the conditional to include an `else if profile.authType` === 'basic' 

I also changed client.test.ts `'setBasicAuth and startClientLogin'` to add a profile with an auth type of client_credentials

packages/cli/src/auth.ts
```diff
@@ -97,3 +113,51 @@ function printMe(medplum: MedplumClient): void {
...
+  formBody.set('grant_type', 'urn:ietf:params:oauth:grant-type:jwt-bearer');
+  formBody.set('client_id', profile.clientId as string);
+  formBody.set('assertion', signedToken);
+  formBody.set('scope', profile.scope ?? '');
+
+  const res = await medplum.post(profile.tokenUrl as string, formBody.toString(), 'application/x-www-form-urlencoded', {
+    credentials: 'include',
+  });
+
+  const obj = await JSON.parse(res);
```
codyebberson, line 161: Is this necessary?  `res` should be a JSON object already.

codyebberson (approved):
Nice, thanks @jamestouri .  One minor question on the result of `medplum.post`, otherwise lgtm 👍

packages/cli/src/auth.ts
jamestouri: I ran `console.log(typeof res)` and it returned string, and the active login will be empty without `JSON.parse(res)`

Here's the output without JSON.parse()

```
string
{"access_token":"{ACCESS_TOKEN}","scope":"create_users,place_orders,patient360,user/*.*","token_type":"Bearer","expires_in":604800,"hg_rest_api_version":"04-01-2023"}
```

Issue #2581, opened 2023-07-31
Allow uploading StructureDefinition resources for profile validation


mattwiller:
Enable users to upload their own `StructureDefinition` resources for use in validation.  This will require removing the [hardcoded block](https://github.com/medplum/medplum/blob/7fb02b94b3fb052da22d5b892ff0e951d11b1912/packages/core/src/access.ts#L59-L61) on writing `StructureDefinition` resources, but we will need to differentiate between user-uploaded profiles and resources comprising the base FHIR spec.  Currently, the base spec resources must be present in the database to support [dynamically loading resource type schema](https://github.com/medplum/medplum/blob/7fb02b94b3fb052da22d5b892ff0e951d11b1912/packages/core/src/client.ts#L1414) on the client side.  These resources must be protected from modification, while any other `StructureDefinition` should be writeable.

reshmakh:
Tagging here for recordkeeping https://github.com/medplum/medplum/discussions/2006

mattwiller:
Resolved by #2591

PR #2591, opened 2023-08-01
Move "public" resources into R4 Project

mattwiller:
Creates a special Project for holding the base R4 FHIR specification conformance resources.  The goal is to allow inclusion of those resources into other projects effectively via reference, so the hundreds of `StructureDefinition` and other resources do not live in each Project separately.  This change includes:
- Creation of an R4 project with a well-known ID (the V5 UUID generated for name "R4" under the [Nil UUID](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.7) namespace
- A database migration to move all "orphan" resources into the R4 project at server startup
- Updates to the seed script to create spec definition resources in the R4 project at first server startup
- Login in the `Repository` to add the R4 `Project` compartment to all DB read queries; blocking all writes across project boundaries

Related to #2581


reshmakh:
Discussed - test in staging.  @mattwiller @codyebberson

packages/server/src/fhir/operations/export.ts
```diff
@@ -113,12 +113,14 @@ function getResourceTypesByExportLevel(exportLevel: string): ResourceType[] {
   return getResourceTypes();
 }
 
+const unexportedResourceTypes = ['CodeSystem', 'SearchParameter', 'StructureDefinition', 'ValueSet'];
+
 function canBeExported(resourceType: string): boolean {
   if (resourceType === 'BulkDataExport') {
```
codyebberson, line 119: Now that we're introducing `unexportedResourceTypes`, should we move the `BulkDataExport` into that?

Then this method becomes `return !protectedResourcesTypes.includes(resourceType) && !unexportedResourceTypes.includes(resourceType)`

packages/server/src/migrations/v48.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+  await moveOrphanResourcesIntoProject('StructureDefinition', r4ProjectId, client);
+  await moveOrphanResourcesIntoProject('SearchParameter', r4ProjectId, client);
+  await moveOrphanResourcesIntoProject('ValueSet', r4ProjectId, client);
+}
+
+const updateQuery = `UPDATE "__TABLE__" SET
+    "projectId" = $1::UUID,
+    compartments = ARRAY[$1]::UUID[],
+    content = jsonb_set(content::jsonb, '{meta, project}'::text[], to_jsonb($1), true)
```
codyebberson, line 25: Clever

codyebberson (approved):
Nice, this looks good.

I'll deploy to staging when I get a chance.  Assuming that looks good, then this is good-to-go 👍

packages/server/src/migrations/v48.ts
mattwiller: 😅

codyebberson:
Staging deploy failed:

image

Does the migration handle the case where the `Project` is deleted, and the `content` column is empty?

packages/server/src/migrations/v48.ts
```diff
@@ -0,0 +1,30 @@
...
+  WHERE
+    "projectId" IS NULL OR "projectId" = $1`;
```
codyebberson, line 27: Might need to add `AND "content"<>''`

codyebberson (approved):

PR #2594, opened 2023-08-01
Combine Publish and Sync Repos

jamestouri:


.github/workflows/publish-and-sync-repos.yml
```diff
@@ -1,11 +1,12 @@
...
 # Based on: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
 on:
   push:
-    tags:
-      - 'v*'
+    branches: [main]
+
 jobs:
   build:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
```
jamestouri, line 9: Added conditional for the publish script to only run on tags with v

.github/workflows/publish-and-sync-repos.yml
```diff
@@ -1,11 +1,12 @@
-name: Publish
+name: Publish and Sync Repos
 # Based on: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
 on:
   push:
-    tags:
-      - 'v*'
+    branches: [main]
```
jamestouri, line 5: This will trigger all merges to main, below I added a conditional for publish based on the v tag

.github/workflows/publish-and-sync-repos.yml
```diff
@@ -1,11 +1,12 @@
-name: Publish
+name: Publish and Sync Repos
```
codyebberson, line 1: Let's keep the filename `publish.yml` and the original title.  This can all fall under the umbrella of "Publish"

.github/workflows/publish-and-sync-repos.yml
```diff
@@ -1,11 +1,12 @@
...
 # Based on: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
 on:
   push:
-    tags:
```
codyebberson, line 5: Let's keep this as is?

I think the goal is that we publish and sync repos at the same time on the same cadence.

.github/workflows/publish-and-sync-repos.yml
codyebberson, line 9: You can then remove this line

.github/workflows/publish-and-sync-repos.yml
```diff
@@ -39,3 +40,29 @@ jobs:
         run: ./scripts/publish.sh
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          
+  check_changes_and_update_repos:
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
codyebberson, line 45: You can then remove this line

.github/workflows/publish-and-sync-repos.yml
```diff
@@ -39,3 +40,29 @@ jobs:
...
+    name: Code Changes
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
codyebberson, line 54: If we combine this into the same `steps`, then you won't need to re-checkout.

.github/workflows/publish-and-sync-repos.yml
```diff
@@ -39,3 +40,29 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - name: Set script permissions
+        run: |
+          chmod +x ./scripts/check-changes.sh
```
codyebberson, line 59: We should fix this in the git repo.

https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git

.github/workflows/publish.yml
```diff
@@ -3,17 +3,25 @@ name: Publish
 on:
   push:
     tags:
-      - 'v*'
+      - "v*"
```
codyebberson, line 6: Let's sync to make sure your prettier is setup properly.  If you run code formatter in VS Code, it should change this back to single quotes and remove the blank lines.

.github/workflows/publish.yml
```diff
@@ -3,17 +3,25 @@ name: Publish
...
 jobs:
-  build:
-    name: Build
+  build_and_check:
```
codyebberson, line 8: We should probably name this `publish`.  It was incorrect when i called it `build`.

.github/workflows/publish.yml
```diff
@@ -35,7 +45,18 @@ jobs:
           MEDPLUM_REGISTER_ENABLED: '__MEDPLUM_REGISTER_ENABLED__'
           GOOGLE_CLIENT_ID: '__GOOGLE_CLIENT_ID__'
           RECAPTCHA_SITE_KEY: '__RECAPTCHA_SITE_KEY__'
+
       - name: Publish
         run: ./scripts/publish.sh
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Check for changes in specified path
```
codyebberson, line 54: Let's keep these names tighter.  `Check for changes`

.github/workflows/publish.yml
```diff
@@ -35,7 +45,18 @@ jobs:
...
+        id: check_changes
+        run: |
+          echo "changed_paths=$(./scripts/check-changes.sh ${{ github.event.before }} ${{ github.event.after }})" >> $GITHUB_ENV
+
+      - name: Update listening repos, create PR, and clean up
```
codyebberson, line 59: `Sync example repos`

codyebberson (approved):

PR #2597, opened 2023-08-02
Add support for _elements and _summary search parameters

### LINKED ISSUE #2332: Add support for _summary search parameter
The search API currently supports using `_summary=count` to retrieve the number of matching search results, but does not yet handle the other values for the [`_summary` parameter](http://hl7.org/fhir/R4/search.html#summary), which return a subset of the resulting resources' contents.  There are four additional cases we should implement:
- `_summary=true`: return only elements marked with `isSummary: true` in their `ElementDefinition`
- `_summary=text`: Return only `text`, `id`, `meta`, and any other required properties of the resource
- `_summary=data`: Return everything except `text`
- `_summary=false`: Default case, return everything (effectively a no-op)

In any case where properties are being stripped out of the resource for display, we should also add the following tag [coding](https://terminology.hl7.org/CodeSystem-v3-ObservationValue.html#v3-ObservationValue-SUBSETTED) to denote that what is shown is not the complete resource:
```json
"meta": {
  "tag": [
    {
      "code": "SUBSETTED",
      "system": "http://terminology.hl7.org/CodeSystem/v3-ObservationValue"
    }
  ]
}
```

mattwiller:
Added a general-purpose `subsetResource()` utility function for generating a version of a resource with selected properties removed, which correctly handles known edge cases and marks the result as a subset of the whole resource.  This utility forms the basis of the implementation of both `_elements` and `_summary` search parameters: the former allowing direct selection of subset properties, and the latter providing several predefined subsets.

Resolves #2332

### DIFF
--- a/packages/core/src/search/search.ts
+++ b/packages/core/src/search/search.ts
@@ -15,6 +15,7 @@ export interface SearchRequest<T extends Resource = Resource> {
   total?: 'none' | 'estimate' | 'accurate';
   include?: IncludeTarget[];
   revInclude?: IncludeTarget[];
+  summary?: 'true' | 'text' | 'data';
 }
 
 export interface Filter {
@@ -216,8 +217,12 @@ function parseKeyValue(searchRequest: SearchRequest, key: string, value: string)
       break;
 
     case '_summary':
-      searchRequest.total = 'accurate';
-      searchRequest.count = 0;
+      if (value === 'count') {
+        searchRequest.total = 'accurate';
+        searchRequest.count = 0;
+      } else if (value === 'true' || value === 'data' || value === 'text') {
+        searchRequest.summary = value;
+      }
       break;
 
     case '_include': {
@@ -247,6 +252,7 @@ function parseKeyValue(searchRequest: SearchRequest, key: string, value: string)
     }
 
     case '_fields':
+    case '_elements':
       searchRequest.fields = value.split(',');
       break;
 
--- a/packages/server/src/fhir/search.ts
+++ b/packages/server/src/fhir/search.ts
@@ -135,22 +135,29 @@ async function getSearchEntries<T extends Resource>(
   }
 
   for (const entry of entries) {
-    repo.removeHiddenFields(entry.resource as Resource);
+    if (!entry.resource) {
+      continue;
+    }
+    repo.removeHiddenFields(entry.resource);
     if (searchRequest.fields) {
-      entry.resource = subsetResource(entry.resource, searchRequest.fields);
+      const schema = getDataType(entry.resource.resourceType);
+      entry.resource = subsetResource(
+        entry.resource,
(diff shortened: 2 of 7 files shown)


packages/server/src/fhir/search.ts
```diff
@@ -143,6 +148,29 @@ async function getSearchEntries<T extends Resource>(
   };
 }
 
+function removeResourceFields(resource: Resource, repo: Repository, searchRequest: SearchRequest): void {
+  repo.removeHiddenFields(resource);
+  if (searchRequest.fields) {
```
codyebberson, line 153: I believe that `fields` and `elements` need to be tracked separately in `SearchRequest`.

Currently, `fields` is a mix of element names and search parameter codes.  For example, when looking at `Patient`, you can add `id`, `name`, and `telecom` (all element names) but you can also add `email` and `phone` which are not element names but instead search parameters.

On the client, we evaluate the search parameter to show the value.

https://app.staging.medplum.com/Patient?_count=20&_fields=id,_lastUpdated,name,birthDate,gender,telecom,email,phone&_offset=0&_sort=-_lastUpdated

image

packages/core/src/search/search.ts
```diff
@@ -247,6 +252,7 @@ function parseKeyValue(searchRequest: SearchRequest, key: string, value: string)
     }
 
     case '_fields':
+    case '_elements':
```
codyebberson, line 255: As mentioned earlier, let's break this out into 2 separate properties on `searchRequest`

codyebberson (changes requested):
See comments about separating `_fields` and `_elements`.

Also, we will need to update the auto-caching logic in `MedplumClient`.  In particular, look at the implementation of `MedplumClient.search`: https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts#L1153

```ts
  search<K extends ResourceType>(
    resourceType: K,
    query?: QueryTypes,
    options?: RequestInit
  ): ReadablePromise<Bundle<ExtractResource<K>>> {
    const url = this.fhirSearchUrl(resourceType, query);
    const cacheKey = url.toString() + '-search';
    const cached = this.getCacheEntry(cacheKey, options);
    if (cached) {
      return cached.value;
    }
    const promise = new ReadablePromise(
      (async () => {
        const bundle = await this.get<Bundle<ExtractResource<K>>>(url, options);
        if (bundle.entry) {
          for (const entry of bundle.entry) {
            this.cacheResource(entry.resource);
          }
        }
        return bundle;
      })()
    );
    this.setCacheEntry(cacheKey, promise);
    return promise;
  }
```

We'll need to add an extra check for the presence of `_elements`.  If `_elements` is specified, then we cannot cache the bundle entries.  Otherwise we would end up caching partial resources for future `readResource` calls.

For example, change 

```ts
if (bundle.entry) {
```

to

```ts
if (!searchRequest.elements && bundle.entry)( {
```

codyebberson (approved):

PR #2607, opened 2023-08-02
anuragPatch1 #2605 resolved

### LINKED ISSUE #2605: Default avatar icon wrong size
I suspect something changed in Mantine, so now the icon inside the default avatar looks oddly small:

image

See `ResourceAvatar.tsx`: https://github.com/medplum/medplum/blob/main/packages/react/src/ResourceAvatar/ResourceAvatar.tsx

See Mantine `Avatar` docs: https://mantine.dev/core/avatar

You can see this in Storybook: https://storybook.medplum.com/?path=/story/medplum-resourceavatar--letter-sizes

anuragdaksh7:
#2605 issue resolved

*need steps to run on my machine

### DIFF
--- a/packages/react/src/ResourceAvatar/ResourceAvatar.tsx
+++ b/packages/react/src/ResourceAvatar/ResourceAvatar.tsx
@@ -14,7 +14,7 @@ export function ResourceAvatar(props: ResourceAvatarProps): JSX.Element {
   const resource = useResource(props.value);
   const text = resource ? getDisplayString(resource) : props.alt ?? '';
   const imageUrl = (resource && getImageSrc(resource)) ?? props.src;
-  const radius = props.radius ?? 'xl';
+  const radius = props.radius ?? '4xl';
 
   const avatarProps = { ...props };
   delete avatarProps.value;


codyebberson:
Hi @anuragdaksh7  - thanks for submitting!

I don't think the issue is entirely fixed though.  You can see in the PR story book that the avatar sizes are still incorrect:

https://medplum-storybook-git-fork-anuragdaksh7-anuragpatch1-medplum.vercel.app/?path=%2Fstory%2Fmedplum-resourceavatar--letter-sizes

image

I looked into this a bit more, and I think it might actually be an upstream bug in Mantine: https://github.com/mantinedev/mantine/issues/4576

anuragdaksh7:
Hi @codyebberson 
can you guide me the steps to deploy this on my device

anuragdaksh7:
image
still localhost:6006 is unserved (I have done every step mentioned)
image

rahul1:
Hi @anuragdaksh7, based on your screenshots, it seems like you may have entered the `packages/server` directory, rather than the `packages/react` directory as cody mentioned above. 

If you try the following, starting from your medplum root directory, you should be able to storybook running at 6006

> > You can then start Storybook with the following commands:
> 
> cd packages/react
> npm run dev
>

anuragdaksh7:
when I am changing the size of this icon then it works but in case of placeholder icon these props just don't work

image

placeholder icon
image

Proposed solution:
using react icons

import { RxAvatar } from 'react-icons/rx';

<RxAvatar size="2rem" /> //adding this line inside Avatar tag would help
<Avatar ....><RxAvatar size="2rem" `/></Avatar>`

codyebberson:
Hi @anuragdaksh7 - Yeah, this is fixed in the next release of Mantine.  I think we should wait for that. https://github.com/mantinedev/mantine/pull/4600

Issue #2570, opened 2023-07-28
Batch upload tool doesn't handle multiple files


rahul1:
User Driven: https://discord.com/channels/905144809105260605/1134531308962467851/1134547021689274458

Right now our batch upload tool tells users to "Upload as many files as you like". But the app only processes the [first file that is uploaded.](https://github.com/medplum/medplum/blob/9d952bb739ce5b6314ffaf2d0fa68f0aca98cd07/packages/app/src/BatchPage.tsx#L56) 

Two options here: 
1. Change the copy on the Batch upload tool (easy)
2. Modify the Batch upload tool to process multiple files

Abi-Liu:
Hey can I take this issue if no one's working on?

codyebberson:
@Abi-Liu - Yep! 

I agree with @rahul1 that we should start with the simple fix of "Change the copy on the Batch upload tool (easy)"

If you're feeling ambitious, we could try to support multiple uploads.  That's really 2 subtasks:

1. In the `handleFiles` function in [BatchPage.tsx](https://github.com/medplum/medplum/blob/9d952bb739ce5b6314ffaf2d0fa68f0aca98cd07/packages/app/src/BatchPage.tsx#L56) process all files rather than `files[0]` (easy)
2. Figure out how to show the results of multiple uploaded files...  @rahul1 do you have any thoughts on what that should look like?
    a. Keep appending output to the same panel (easiest)
    b. Add output tabs for each file (medium)
    c. Full on UI for showing each uploaded file, upload state, and results (hardest)

Abi-Liu:
@codyebberson  Hey just wanted to make sure that I'm clear on everything.

For number 1 I could just do something like a for loop to iterate through and have ```reader.readAsText(files[i])``` or something like that right?

As for number 2, To just append all of the outputs to the same panel I could just map each item in the output array to the JSX element responsible for showing the output.

Am I missing anything?

rahul1:
@Abi-Liu I think that's the right direction. 

For #2, if you could somehow show which file produced which output, that would be nice for helping users debug

codyebberson:
Hi @Abi-Liu  - Yep, that all sounds great!

Abi-Liu:
Sounds good! I'll get started on the fix. Are there any files I can use for testing, or can I just use any pdf file?

rahul1:
Hi @Abi-Liu - the target files for this tool are JSON files that contain FHIR Bundle resources. 

[...]

PR #2614, opened 2023-08-03
Batch upload tool to handle multiple files and display the results

Abi-Liu:
[#2570](https://github.com/medplum/medplum/issues/2570)

I first created a for loop iterate over the submitted files in the handleFiles function and call submitBatch on each file. To display separate tabs for each file, I modified the function submitBatch to take an additional fileName variable as well as made the output state to be an object with the key being the file name and the value being the bundle. I then mapped over the object keys to create the tabs for each file and panels for the corresponding data.

This is what the final result looks like.
image


codyebberson:
Nice, thanks @Abi-Liu !  Tested & verified

image

codyebberson (approved):

PR #2631, opened 2023-08-05
Update Sweep yaml with a repo description that will write tests automatically.

wwzeng1:
# Summary
I was told that Medplum uses a testing pattern in the same directory so I experimented with our `description` feature to automatically prompt Sweep to write tests when it adds a new code path.

Notes:
I made `description` a really long string, wasn't sure of your preferred format. If you have a yaml formatting preference feel free to modify it! You can add newlines/change the content, as long as there is no unnecessary leading/trailing whitespace on the lines.

# Test
I tested it in this PR: https://github.com/wwzeng1/medplum/pull/4 (Made sure that it writes tests)
The PR may not make sense at all, but it did write tests.

### DIFF
--- a/sweep.yaml
+++ b/sweep.yaml
@@ -1,4 +1,4 @@
 # Reference: https://github.com/sweepai/sweep/blob/main/.github/sweep.yaml. If you use this be sure to frequently sync your default branch(main, master) to dev.
 branch: 'main'
 gha_enabled: True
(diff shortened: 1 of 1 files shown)


wwzeng1:
cc @reshmakh Related to our conversation earlier!

codyebberson (approved):

Issue #2606, opened 2023-08-02
Add "copy" button to resource id on resource page


codyebberson:
One thing I frequently do, and always find frustrating, is copying a resource ID from the resource page:

It would be handy to add a copy button right next to the ID.  

Mantine provides a `<CopyButton>` component: https://mantine.dev/core/copy-button/#timeout

The demo is basically what I would want on the resource page:

```tsx
import { CopyButton, ActionIcon, Tooltip } from '@mantine/core';
import { IconCopy, IconCheck } from '@tabler/icons-react';

function Demo() {
  return (
    <CopyButton value="https://mantine.dev" timeout={2000}>
      {({ copied, copy }) => (
        <Tooltip label={copied ? 'Copied' : 'Copy'} withArrow position="right">
          <ActionIcon color={copied ? 'teal' : 'gray'} onClick={copy}>
            {copied ? <IconCheck size="1rem" /> : <IconCopy size="1rem" />}
          </ActionIcon>
        </Tooltip>
      )}
    </CopyButton>
  );
}
```

Open question about whether we would add it to every element (might be too much?).  Only show on hover (might be too fancy?)

The simple solution would be to just add it to the `id` element when looping over properties in `BackboneElementDisplay.tsx`: packages/react/src/BackboneElementDisplay/BackboneElementDisplay.tsx

codyebberson:
Let's add a special case in `ResourcePropertyDisplay.tsx` for when `property.path.endsWith('.id')` to add the `<CopyButton>` as described above.

packages/react/src/ResourcePropertyDisplay/ResourcePropertyDisplay.tsx

You can test this in Storybook.

To run Storybook, first make sure you can run the repo: https://www.medplum.com/docs/contributing/run-the-stack

Then:

```bash
cd packages/react
npm run dev
```

Abi-Liu:
Hi @codyebberson I've looked through the code and have started implementing the changes. Can I be assigned this issue?

PR #2634, opened 2023-08-06
Added copy button to resource ID on resource page

Abi-Liu:
This PR adds a copy button to the resource ID on the resource page and fixed branch issues from last PR - https://github.com/medplum/medplum/issues/2606


codyebberson (approved):

codyebberson:
Hi @Abi-Liu  - Awesome, thanks for submitting!

Normally I'd push for tests.  However, this one is surprisingly tricky to write tests for, because Mantine does some weird stuff with hover events and clipboard events.  I sat down and tried to write the tests myself, but it looks like it would be quite annoying to do with `fireEvent`.  Internally, Mantine uses `user-event`, which we should probably migrate to at some point.  I filed a ticket for that seperately: https://github.com/medplum/medplum/issues/2637

Thanks again!

Abi-Liu:
Hey @codyebberson, I'll definitely keep tests in mind in the future. I need to get better at writing test cases so this will be good practice for me.

PR #2640, opened 2023-08-07
Medplum Agent data definitions

codyebberson:


packages/definitions/dist/fhir/r4/profiles-medplum.json
```diff
@@ -3849,5 +3849,356 @@
...
+              "targetProfile" : ["https://medplum.com/fhir/StructureDefinition/Bot"]
+            }],
+            "base" : {
+              "path" : "Agent.channel.target",
+              "min" : 1,
+              "max" : "1"
+            }
+          },
+          {
+            "id" : "Agent.channel.bindHost",
```
rahul1, line 4133: Might want to consider using the `Endpoint` resource for some of these properties related to the connection. Leverages the spec, and has terminology bindings for mimetype, etc.

Haven't done an in-depth analysis of the pros and cons, but having the Endpoint as a separate resource might make it easier to update across multiple agents


example

```json
{
      "resourceType" : "Endpoint",
      "id" : "77",
      "text" : {
        "status" : "generated",
        "div" : "<div xmlns=\"http://www.w3.org/1999/xhtml\">\n              v2 referral Requests, REF-I12<br/>\n              Address: mllp://127.0.0.1:4001</div>"
      },
      "status" : "active",
      "connectionType" : [{
        "coding" : [{
          "system" : "http://terminology.hl7.org/CodeSystem/endpoint-connection-type",
          "code" : "hl7v2-mllp"
        }]
      }],
      "name" : "v2 referral Requests",
      "managingOrganization" : {
        "display" : "Epic demo organization"
      },
      "payload" : [{
        "type" : [{
          "coding" : [{
            "code" : "REF-I12"
          }]
        }],
        "mimeType" : ["x-application/hl7-v2+er7"]
      }],
      "address" : "mllp://127.0.0.1:4001"
    }
```

packages/definitions/dist/fhir/r4/profiles-medplum.json
```diff
@@ -3849,5 +3849,356 @@
...
+              "targetProfile" : ["http://hl7.org/fhir/StructureDefinition/Device"]
+            }],
+            "base" : {
+              "path" : "Agent.channel.device",
+              "min" : 0,
+              "max" : "1"
+            }
+          },
+          {
+            "id" : "Agent.channel.target",
```
rahul1, line 4117: Instead of assuming this will always be a Bot, we might want to consider reusing the BackboneElement `SubscriptionChannel` (i.e. [Subscription.channel](https://hl7.org/fhir/R4/subscription.html)). That would at least allow us to send to arbitrary endpoints fi the use case arises.

codyebberson: I want to make the common case more obvious.  `Subscription.channel.endpoint` is a string, and then you need to know about magic `Bot/{id}` values, you don't get autocompletes, etc.

I'm open to considering `target[x]` choice-of-type, so you could do `targetUrl` to a URL or `targetReference` to a Bot or other resource.

codyebberson: I also want to be wary of blowing up the scope here - opening up `target` to arbitrary endpoints is not a requirement, and we do not have any plans of implementing it in the near future.

packages/definitions/dist/fhir/r4/profiles-medplum.json
codyebberson: If you don't think this will be too much of a training/support burden, then I support this as the "right" data model

reshmakh (approved):

PR #2652, opened 2023-08-10
Agent installer

codyebberson:
New installer features:

- Uses [NSIS](https://nsis.sourceforge.io/Main_Page) to create the installer
- NSIS captures: `baseUrl`, `clientId`, `clientSecret`, and `agentId`
- Uses [winsw](https://github.com/winsw/winsw) to manage Windows Service
- Launches the service with the 4 params
- NSIS sets up "Add/Remove Programs", creates uninstaller, etc

NSIS is showing its age, but still pretty nice.  I tried to keep all of the branding contained in 4 constants, which would be easy for us to change for whitelabeling purposes.

Lots of things to figure out:

- Most of these tools require MS Windows, so how do we build and test in CI/CD
- Installer not signed, so you get warnings about "unknown publisher"
- Logging is going to text files, not to EventLog

Build the installer:

image

Run the installer: 

image

Installer prompts for input:

image

Installer completes successfully:

image

Now you have a running windows service:

image

Example of sending an HL7 message from HAPI HL7:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ node_modules/
 
 # Output
 packages/agent/dist/
+packages/agent/medplum-agent-installer.exe
 packages/app/dist/
 packages/cdk/dist/
 packages/cli/dist/
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ node_modules/
 
 # Output
 packages/agent/dist/
-packages/agent/medplum-agent-installer.exe
 packages/app/dist/
 packages/cdk/dist/
 packages/cli/dist/
@@ -51,3 +50,6 @@ medplum-server.tar.gz
 
 # Rollup
 .rollup.cache/
+
+# All executables
+*.exe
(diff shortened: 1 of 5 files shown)


jamestouri (approved):

.gitignore
```diff
@@ -11,6 +11,7 @@ node_modules/
 
 # Output
 packages/agent/dist/
+packages/agent/medplum-agent-installer.exe
```
mattwiller, line 14: Should we just use e.g. `*.exe` here?

packages/agent/src/main.ts
```diff
@@ -165,10 +164,18 @@ export class AgentHl7ChannelConnection {
...
-  const medplum = new MedplumClient(config);
+  if (process.argv.length < 6) {
+    console.log('Usage: node medplum-agent.js <baseUrl> <clientId> <clientSecret> <agentId>');
+    process.exit(1);
+  }
+
+  const baseUrl = process.argv[2];
+  const clientId = process.argv[3];
+  const clientSecret = process.argv[4];
+  const agentId = process.argv[5];
```
mattwiller, line 175: Consider using destructuring to clean this up:
```ts
const [_node, _script, baseUrl, clientId, clientSecret, agentId] = process.argv;
```

PR #2658, opened 2023-08-10
Clean up temp binary storage for unit tests

codyebberson:
**A long time ago**: server unit tests were more verbose.  In `beforeAll`, each test file would setup subsystems one-by-one, including the "binary storage" system.

**Then a while later**: we refactored those tests to use `initApp`.  However, I didn't correctly update all of the cases of temp binary storage, which means that a lot of unnecessary temp files persisted after running unit tests.

**In this PR**: Cleaning that up more conclusively, by setting `config.binaryStorage` inside `loadTestConfig`.  

**Important**: in `shutdownApp`, if `config.binaryStorage` is a temp directory, as defined by (1) using local file system, and (2) path starts with `os.tmpdir()` + `medplum-temp-storage`, then we'll delete all of the temp files in the directory.

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -6,6 +6,7 @@ import { Express, json, NextFunction, Request, Response, Router, text, urlencode
 import { rmSync } from 'fs';
 import http from 'http';
 import { tmpdir } from 'os';
+import { join } from 'path';
 import { adminRouter } from './admin';
 import { asyncWrap } from './async';
 import { authRouter } from './auth';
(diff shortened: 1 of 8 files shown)


mattwiller:
Is it safe to delete the folder on server shutdown?  What happens if my local server gets restarted and a Binary resource in the DB no longer exists on disk?

codyebberson:
> Is it safe to delete the folder on server shutdown?

First, to be clear, this should only apply to unit tests.  Or, in the weird case where someone uses `os.tmpdir() + 'medplum-temp-storage'` as their binary storage directory.  The contract for `os.tmpdir()` (i.e., `/tmp`) in most Linux distributions is to empty the tmp directory after each reboot.  So.... temp directories are inherently ephemeral and at risk of deletion.

> What happens if my local server gets restarted and a Binary resource in the DB no longer exists on disk?

In the general case, if you configure Medplum `server` to use local file system for binary storage, and the local file is deleted, then you'll get a 404 when you try to receive the binary contents.  The `Binary` resource is still in-tact, but retrieving the contents will be broken.

When running on localhost, it's not totally uncommon, and the main way it manifests is broken images in the `app` UI.

mattwiller (approved):

PR #2667, opened 2023-08-11
'Add Page' to Questionnaire 

jamestouri:
https://github.com/medplum/medplum/assets/6913745/c8b1be09-8600-42d1-bfac-f5b2fcac2f8d

https://github.com/medplum/medplum/assets/6913745/3f818c81-b7fb-4e0b-873d-13dea13d07db

https://github.com/medplum/medplum/assets/6913745/531dd0a7-5c61-439e-a50a-922bee521108

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -530,7 +530,7 @@ function ButtonGroup(props: ButtonGroupProps): JSX.Element {
     return (
       <>
         <Button onClick={props.prevStep}>Back</Button>
-        <Button type="submit">{props.submitButtonText ?? 'OK'}</Button>
+        <Button onClick={props.nextStep} type="submit">{props.submitButtonText ?? 'OK'}</Button>
       </>
     );
   } else if (props.count === 0) {
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -109,17 +109,15 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element | 
       }}
     >
       {questionnaire.title && <Title>{questionnaire.title}</Title>}
-      {questionnaire.item && numberOfPages > 0 ? (
-        <QuestionnaireFormGroupPages
+
+      {questionnaire.item && (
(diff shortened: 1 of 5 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -120,6 +143,32 @@ function QuestionnaireFormItemArray(props: QuestionnaireFormItemArrayProps): JSX
     setResponseItems(newResponseItems);
     props.onChange(newResponseItems);
   }
+  const firstItem = props.items[0];
+  const firstItemValue = getExtension(firstItem, 'https://medplum.com/fhir/StructureDefinition/page-sequence');
+
+  if (firstItemValue && firstItemValue.valueString === 'page') {
```
rahul1, line 149: The logic here looks at the first item. If it's a page, this renders the whole questionnaire using a stepper. We drop items that are not in a page. 

This logic will produce surprising behavior for authors who might accidentally forget to set a group as a "page", or mix groups and questions at a top level. 

A more graceful approach would be to group all consecutive non-page items into their own pages. 

E.g
[q1, page1, page2, q3, q4, page 3]
woud yield the following pages: 
[q1]
[page1]
[page2]
[q3, q4]
[page3]

rahul1 (changes requested):

rahul1 (commented):
This is looking really good. This _could_ ship as is, but we can reduce a _bit_ more code repetition. Will discuss offline tomorrow

rahul1:
Sonar warnings can be dismissed, as they are complaining about extension URL strings

rahul1 (approved):

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -529,3 +633,11 @@ export function isQuestionEnabled(
     return true;
   }
 }
+
+function getNumberOfPages(items: QuestionnaireItem[]): number {
+  const pages = items.filter((item) => {
+    const extension = getExtension(item, 'http://hl7.org/fhir/R4B/extension-questionnaire-itemcontrol.html');
```
rahul1, line 639: URL for this extension should have been
http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl


[reference](http://hl7.org/fhir/R4/extension-questionnaire-itemcontrol.html)

PR #2671, opened 2023-08-14
Clarify user admin

rahul1:


reshmakh:
Another solid one - only minor feedback.  That diagram is going to help people.

reshmakh (approved):

PR #2682, opened 2023-08-17
fix(docs): correct client credentials flow python example

JaKXz:
As discussed here, this was a typo in the python example: https://discord.com/channels/905144809105260605/1141077048253747430


rahul1:
Thank you for the contribution @JaKXz !

PR #2684, opened 2023-08-17
Add FHIRPath criteria for write operations to AccessPolicy

### LINKED ISSUE #2647: Disallow certain write operations via access policy
Implement an element in `AccessPolicy` called `postUpdateCriteria`  (name subject to change) field. 

This element will specify a fhirpath expression on the resource being written, and will return a true/false value. If false, the write will be disallowed. 

This is will be used to disallow specific write operations.

mattwiller:
Adds a `resource.writeCriteria` field to `AccessPolicy` resources, which should contain a FHIRPath constraint expression to be evaluated either against the original state of the resource or the updated state, before it is written to the database.  This also introduces the full set of [resource interactions](http://hl7.org/fhir/codesystem-restful-interaction.html) as permissions, extending the simple read/write logic.

Resolves #2647


reshmakh:
Adding a note here mostly for myself: the use case for this feature is for workflow, I believe.  For example, we want to ensure that only users with the right permissions can edit final DiagnosticReports, but a different set can prepare and edit drafts.  @rahul1 - am I correct here?

packages/core/src/access.ts
```diff
@@ -90,35 +126,67 @@ export function canWriteResource(accessPolicy: AccessPolicy, resource: Resource)
...
+      ) {
         return true;
       }
     }
   }
   return false;
 }
 
+/**
+ * Check that there is an access policy permitting the given resource interaction, returning the matching policy object.
```
codyebberson, line 149: Super minor nit:  When writing jsdoc comments, I try to follow the recommended javadoc convention:

> Use 3rd person (descriptive) not 2nd person (prescriptive). The description is in 3rd person declarative rather than 2nd person imperative.
> 
> Gets the label. (preferred)
> 
> Get the label. (avoid)

https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#styleguide

packages/core/src/access.ts
```diff
@@ -90,35 +126,67 @@ export function canWriteResource(accessPolicy: AccessPolicy, resource: Resource)
...
+ * @param resource The resource being acted upon.
+ * @param interaction The interaction being performed on the resource.
+ * @param accessPolicy The relevant access policy for the current user.
+ * @returns The satisfied access policy, or undefined if the access policy does not permit the given interaction.
+ */
+export function satisfiedAccessPolicy(
```
codyebberson, line 155: Similar to above, I think this should be `satisfies` rather than `satisfied`?

packages/docs/static/data/medplumDefinitions/accesspolicy.json
```diff
@@ -289,7 +337,7 @@
       "depth": 2,
       "types": [
         {
-          "datatype": "string"
+          "datatype": "code"
```
codyebberson, line 340: Nice, good catch 👍

codyebberson (approved):
Nice, this is excellent @mattwiller, thank you!

See super minor comment on jsdoc verb conjugation.  Feel free to disregard 🙃

packages/docs/static/data/medplumDefinitions/accesspolicy.json
codyebberson: Actually, you found lots of good bugs in the structure definitions!  thank you!  🐑

packages/core/src/access.ts
mattwiller: Cool, that makes sense to me!

packages/core/src/access.ts
mattwiller: Since the function returns the resource policy that was satisfied for the interaction, I changed it from `satisfies` to `satisfied` — very open to suggestions on the name though

packages/docs/static/data/medplumDefinitions/accesspolicy.json
mattwiller: I just ran the generator and these changes were emitted; do we need to add something to the release process to ensure all relevant generator scripts get executed?

rahul1:
> Adding a note here mostly for myself: the use case for this feature is for workflow, I believe. For example, we want to ensure that only users with the right permissions can edit final DiagnosticReports, but a different set can prepare and edit drafts. @rahul1 - am I correct here?

Yes that's correct. Essentially, the use case to manage resources **state transitions** (e.g. preventing the setting of  `DiagnosticReport.status` to `final`

PR #2690, opened 2023-08-18
Add repeatable items and groups to Questionnaire

jamestouri:
https://github.com/medplum/medplum/issues/2689

Screenshot 2023-08-22 at 10 00 24 PM

The output:

Screenshot 2023-08-22 at 10 00 13 PM

### DIFF
--- a/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -190,7 +190,7 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
     } as T);
   }
 
-  function handleRepeatableValue(item: QuestionnaireItem): void {
+  function toggleRepeatable(item: QuestionnaireItem): void {
     props.onChange({
       ...props.item,
       item: props.item.item?.map((i) => (i === item ? { ...i, repeats: !i.repeats } : i)),
@@ -250,7 +250,7 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
             setHoverKey={props.setHoverKey}
             onChange={changeItem}
             onRemove={() => removeItem(i)}
-            onRepeatable={handleRepeatableValue}
+            onRepeatable={toggleRepeatable}
           />
         </div>
       ))}
--- a/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -135,6 +135,7 @@ interface ItemBuilderProps<T extends Questionnaire | QuestionnaireItem> {
   setHoverKey: (key: string | undefined) => void;
   onChange: (item: T) => void;
   onRemove?: () => void;
+  onRepeatable?: (item: QuestionnaireItem) => void;
 }
 
 function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBuilderProps<T>): JSX.Element {
@@ -187,6 +188,13 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
       ...itemRef.current,
       [property]: value,
     } as T);
+  }
+
+  function handleRepeatableValue(item: QuestionnaireItem): void {
+    props.onChange({
+      ...props.item,
+      item: props.item.item?.map((i) => (i === item ? { ...i, repeats: !i.repeats } : i)),
+    });
   }
 
   const className = cx(classes.section, {
@@ -242,6 +250,7 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
             setHoverKey={props.setHoverKey}
             onChange={changeItem}
             onRemove={() => removeItem(i)}
+            onRepeatable={handleRepeatableValue}
           />
         </div>
       ))}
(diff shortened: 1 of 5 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -641,3 +733,95 @@ function getNumberOfPages(items: QuestionnaireItem[]): number {
...
+        valueString: 'removable',
+      },
+    ],
+  };
+}
+
+function extensionForRemovingRepeatable(item: QuestionnaireItem): boolean {
+  const extension = getExtension(
+    item,
+    'http://hl7.org/fhir/R4/questionnaire-definitions.html#Questionnaire.item.repeats'
```
jamestouri, line 824: I needed a way to have the remove show for all of the repeatable components, except the original one, so I created an extension for it. I searched around the fhir specs for something close to it but couldn't find any. Something like "Can remove" or "Original Repeat"

packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -189,6 +190,13 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
     } as T);
   }
 
+  function handleRepeatableValue(item: QuestionnaireItem): void {
```
codyebberson, line 193: Would a better name for this function be `toggleRepeatable`?  Because it looks like it's used for both "make repeatable" and "remove repeatable"?

packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
```diff
@@ -367,6 +369,47 @@ export const EnableWhen = (): JSX.Element => (
   </Document>
 );
 
+export const Repeatables = (): JSX.Element => (
```
codyebberson, line 372: Let's rename this to "RepeatableItems".

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -88,6 +93,18 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
     setAnswers(getQuestionnaireAnswers(newResponse));
   }
 
+  function handleRepeatableItem(currentItem: QuestionnaireItem): void {
```
codyebberson, line 96: Would this be better named `addRepeatableItem`?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -109,14 +126,15 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+      {questionnaireItems && (
         <QuestionnaireFormItemArray
-          items={questionnaire.item}
+          items={questionnaireItems}
           answers={answers}
           onChange={setItems}
+          renderPages={numberOfPages > 1}
           activePage={activePage}
-          renderPages={numberOfPages > 0}
+          handleRepeatableItem={handleRepeatableItem}
```
codyebberson, line 136: Event handler props should use `on` prefix, not `handle`.  By internal convention, we name our functions with the `handle` prefix (Mantine does this as well), but the props should always start with `on`.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -63,6 +66,8 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
   const [response, setResponse] = useState<QuestionnaireResponse | undefined>();
   const [answers, setAnswers] = useState<Record<string, QuestionnaireResponseItemAnswer>>({});
   const [activePage, setActivePage] = useState(0);
+  const [questionnaireItems, setQuestionnaireItems] = useState(questionnaire?.item ?? []);
```
codyebberson, line 69: I don't think I follow this... Why do we need to modify the questionnaire items?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
codyebberson: Is there a reason this can't be done within `setItems()`?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -656,3 +748,95 @@ function addTargetTypes(item: QuestionnaireItem): string[] {
...
+
+function createRepeatableItem(item: QuestionnaireItem): QuestionnaireItem {
+  let newText = item.text ?? '';
+  if (!newText.endsWith(' continued')) {
+    newText += ' continued';
+  }
+  return {
+    ...item,
+    text: newText,
+    linkId: repeatableLinkId(item.linkId ?? ''),
```
codyebberson, line 825: I'm pretty sure we should not be modifying `linkId` inside of `QuestionnaireForm`.  `linkId` has a pretty well defined contract in FHIR Questionnaires.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
codyebberson: Let's discuss this offline.  I'm pretty sure this is not necessary.  I think wires are getting crossed between `QuestionnaireItem` and `QuestionnaireResponseItem`.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
codyebberson: See this discussion for `linkId` behavior with `repeats`: http://community.fhir.org/t/questionnaire-repeating-groups-what-is-the-correct-format/2276

codyebberson: "Repeating groups appear multiple times. Repeating questions have multiple answers"

codyebberson: https://hl7.org/fhir/R4/questionnaireresponse.html

codyebberson (approved):

codyebberson:
😠 
image

PR #2712, opened 2023-08-23
fix(client): read `clientSecret` from medplum client constructor options

JaKXz:
Hey team, thanks so much for all the help recently debugging.

I just noticed this line is missing from the medplum client constructor and I'm just opening this PR to ask what I might be missing with this.

I am doing a `medplumClient.createResource(bundle)` call, with my medplum client set up with my `clientId` and `clientSecret` via the constructor options, but I think this line being missing is causing `handleUnauthenticated` to not be able to get a `refreshPromise` from the `refresh` method.

Am I completely off base here? Sorry for the noise if so.


rahul1 (approved):
Thank you @JaKXz !

PR #2714, opened 2023-08-24
feat(QuestionnaireForm): add support for `exists`, `!=`, etc. operators

### LINKED ISSUE #2683: Add support for all Questionnaire `enableWhen` operators
Right now, when evaluating the `Questionnaire.item.enableWhen` logic, we assume that the user is checking for equality. In the example below, I'd also like to check for existence, which is not currently handled.

Would also be nice to for booleans to default eval to "false" if they don't exist, so two conditions are not needed

```json
{
  "resourceType": "Questionnaire",
  "name": "Merge Duplicate Patients",
  "status": "active",
  "item": [
    {
      "id": "id-15",
      "linkId": "g3",
      "type": "group",
      "text": "Disable Merge ",
      "item": [
        {
          "id": "id-23",
          "linkId": "q1",
          "type": "display",
          "text": "Check this box is the Patient records ARE NOT a match. This will add each Patient to the other's \"Do Not Match\" list"
        },
        {
          "id": "id-16",
          "linkId": "disableMerge",
          "type": "boolean",
          "text": "*Do not Merge these records*"
        }
      ]
    },
    {
      "id": "id-17",
      "linkId": "g5",
      "type": "group",
      "text": "Merging Patients",
      "item": [
        {
          "id": "id-18",
          "linkId": "replaceName",
          "type": "boolean",
          "text": "Merge names?\nIf selected, this will add the source patient's name to the target's list of names"
        },
        {
          "id": "id-19",
          "linkId": "replaceAddress",
          "type": "boolean",
          "text": "Merge Addresses?\nIf selected, this will add the source patient's address to the target's list of addresses"
        },
        {
          "id": "id-20",
          "linkId": "replaceData",
          "type": "boolean",
          "text": "Replace birthdate?\nIf selected, this will REPLACE the target patient's birthdate with the source patient's birthdate"
        }
      ],
      "enableWhen": [
        {
          "question": "disableMerge",
          "operator": "=",
          "answerBoolean": false
        },
        {
          "question": "disableMerge",
          "operator": "exists",
          "answerBoolean": false
        }
      ],
      "enableBehavior": "any"
    },
    {
      "id": "id-21",
      "linkId": "g9",
      "type": "group",
      "text": "Deleting Source Patient",
      "item": [
        {
          "id": "id-22",
          "linkId": "deleteSource",
          "type": "boolean",
          "text": "Would you like to delete the source patient after merging?"
        }
      ],
      "enableWhen": [
        {
          "question": "disableMerge",
          "operator": "=",
          "answerBoolean": false
        },
        {
          "question": "disableMerge",
          "operator": "exists",
          "answerBoolean": false
        }
      ],
      "enableBehavior": "any"
    }
  ],
  "subjectType": [
    "RiskAssessment"
  ],
  "title": "Merge Duplicate Patients",
  "id": "02055bc0-d0c9-4441-a6c4-94e490336f99",
  "meta": {
    "versionId": "68bbe973-2ca8-48c8-9ee7-7e57e3446485",
    "lastUpdated": "2023-08-17T17:37:12.410Z",
    "author": {
      "reference": "Practitioner/bda1c932-8e56-4c10-92fa-bbbb3fbe4b03",
      "display": "Rahul Agarwwal"
    },
    "project": "b65c0057-5926-4d59-9486-1039125784ac",
    "compartment": [
      {
        "reference": "Project/b65c0057-5926-4d59-9486-1039125784ac"
      }
    ]
  }
}
```

ThatOneBro:
This PR adds support for the `!=`, `>`, `>=`, `<`, `<=`, and `exists` operators for `QuestionnaireItem`. See: http://hl7.org/fhir/R4/valueset-questionnaire-enable-operator.html

The FHIR spec is a little unclear about what to do when comparison between types that are "potentially comparable" are made, such as `Time` and `DateTime`.

Currently the behavior is to ignore comparisons between types that are "potentially" comparable.

If the two types compared are not exactly equal and if the current answer is not `undefined`, then a match is not made. If the values are exactly the same type or the answer is `undefined` then the comparison is performed and the match is made based on the comparison.

This may be something that should be discussed in the future.

Please let me know if I missed anything obvious in this PR.

This closes #2683.

### DIFF
--- a/packages/core/src/fhirpath/utils.ts
+++ b/packages/core/src/fhirpath/utils.ts
@@ -1,4 +1,4 @@
-import { ElementDefinition, Period, Quantity } from '@medplum/fhirtypes';
+import { ElementDefinition, Period, Quantity, Coding } from '@medplum/fhirtypes';
 import { buildTypeName, getElementDefinition, isResource, PropertyType, TypedValue } from '../types';
 import { capitalize, isEmpty } from '../utils';
 
@@ -285,8 +285,11 @@ export function fhirPathArrayEquivalent(x: TypedValue[], y: TypedValue[]): Typed
  * @returns True if equivalent.
  */
 export function fhirPathEquivalent(x: TypedValue, y: TypedValue): TypedValue[] {
-  const xValue = x.value?.valueOf();
-  const yValue = y.value?.valueOf();
+  const { type: xType, value: xValueRaw } = x;
+  const { type: yType, value: yValueRaw } = y;
+  const xValue = xValueRaw?.valueOf();
+  const yValue = yValueRaw?.valueOf();
+
   if (typeof xValue === 'number' && typeof yValue === 'number') {
     // Use more generous threshold than equality
     // Decimal: values must be equal, comparison is done on values rounded to the precision of the least precise operand.
@@ -296,8 +299,25 @@ export function fhirPathEquivalent(x: TypedValue, y: TypedValue): TypedValue[] {
   if (isQuantity(xValue) && isQuantity(yValue)) {
     return booleanToTypedValue(isQuantityEquivalent(xValue, yValue));
   }
+
+  if (xType === 'Coding' && yType === 'Coding') {
+    if (typeof xValue !== 'object' || typeof yValue !== 'object') {
+      return booleanToTypedValue(false);
+    }
+    // "In addition, for Coding values, equivalence is defined based on the code and system elements only.
+    // The version, display, and userSelected elements are ignored for the purposes of determining Coding equivalence."
+    // Source: https://hl7.org/fhir/fhirpath.html#changes
+
+    // Basically checking if one of the `Coding` resources has a system defined. If so, then the two's system values must be compared
+    const systemsAreEquivalent =
+      (xValue as Coding).system || (yValue as Coding).system
+        ? (xValue as Coding).system === (yValue as Coding).system
+        : true;
+    return booleanToTypedValue((xValue as Coding).code === (yValue as Coding).code && systemsAreEquivalent);
+  }
+
   if (typeof xValue === 'object' && typeof yValue === 'object') {
-    return booleanToTypedValue(deepEquals(xValue, yValue));
+    return booleanToTypedValue(deepEquals({ ...xValue, id: undefined }, { ...yValue, id: undefined }));
   }
   if (typeof xValue === 'string' && typeof yValue === 'string') {
     // String: the strings must be the same, ignoring case and locale, and normalizing whitespace
--- a/packages/core/src/fhirpath/utils.ts
+++ b/packages/core/src/fhirpath/utils.ts
@@ -1,5 +1,5 @@
-import { ElementDefinition, Period, Quantity, Coding } from '@medplum/fhirtypes';
-import { buildTypeName, getElementDefinition, isResource, PropertyType, TypedValue } from '../types';
+import { Coding, ElementDefinition, Period, Quantity } from '@medplum/fhirtypes';
+import { PropertyType, TypedValue, buildTypeName, getElementDefinition, isResource } from '../types';
 import { capitalize, isEmpty } from '../utils';
 
 /**
@@ -308,12 +308,12 @@ export function fhirPathEquivalent(x: TypedValue, y: TypedValue): TypedValue[] {
     // The version, display, and userSelected elements are ignored for the purposes of determining Coding equivalence."
     // Source: https://hl7.org/fhir/fhirpath.html#changes
 
-    // Basically checking if one of the `Coding` resources has a system defined. If so, then the two's system values must be compared
-    const systemsAreEquivalent =
-      (xValue as Coding).system || (yValue as Coding).system
-        ? (xValue as Coding).system === (yValue as Coding).system
-        : true;
-    return booleanToTypedValue((xValue as Coding).code === (yValue as Coding).code && systemsAreEquivalent);
+    // We need to check if both `code` and `system` are equivalent.
+    // If both have undefined `system` fields, If so, then the two's `system` values must be compared.
+    // Essentially they must both be `undefined` or both the same.
+    return booleanToTypedValue(
+      (xValue as Coding).code === (yValue as Coding).code && (xValue as Coding).system === (yValue as Coding).system
+    );
   }
 
   if (typeof xValue === 'object' && typeof yValue === 'object') {
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.test.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.test.tsx
@@ -1359,5 +1359,94 @@ describe('QuestionnaireForm', () => {
         )
       ).toBe(false);
     });
+
+    test('enableBehavior=any, enableWhen `=` operator for `valueCoding`', () => {
+      const enableWhen = [
+        { question: 'q1', operator: '=', answerCoding: { code: 'MEDPLUM123' } },
+      ] satisfies QuestionnaireItemEnableWhen[];
+
+      expect(
+        isQuestionEnabled(
+          {
+            enableWhen,
+          },
+          {
+            q1: { valueCoding: { code: 'MEDPLUM123' } },
+          }
+        )
+      ).toBe(true);
+
+      expect(
+        isQuestionEnabled(
+          {
+            enableWhen,
+          },
+          {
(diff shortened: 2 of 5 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -597,30 +598,87 @@ function isDropDownChoice(item: QuestionnaireItem): boolean {
   );
 }
 
+/**
+ * This method is not called `areSameType` because it only checks if the two value's types are not explicitly different and allows one value to be `undefined`.
+ * @param value1 A value or `undefined`.
+ * @param value2 Another value to compare `value1`'s type to.
+ * @returns If the two values are NOT different types. It returns `true` if both are the same type or if `value1` is `undefined`.
+ */
+function areNotDifferentTypes(value1: TypedValue | undefined, value2: TypedValue): boolean {
```
jamestouri, line 607:
```suggestion
function areNotDifferentTypes(value1?: TypedValue, value2: TypedValue): boolean {
```

jamestouri (commented):
Nice! This all looks good. One thought have you tested with an answer with valueCoding? Previously the deepEquals didn't work with it so we may need a specific conditional for it, or perhaps even structuring the comparisons prior

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -597,30 +598,87 @@ function isDropDownChoice(item: QuestionnaireItem): boolean {
...
+  // We return true for `undefined` since we handle the undefined case inside the case blocks
+  if (value1 === undefined || value1.type === value2.type) {
```
jamestouri, line 609:
```suggestion
  if (!value1 || value1.type === value2.type) {
```

ThatOneBro: I think a strict comparison with `undefined` is the most correct here since we know that it will either be exactly `undefined` or a `TypedValue`

It should also be slightly faster since there is only one explicit comparison instead of the fuzzy compare via falsiness check with `!`

ThatOneBro:
> Nice! This all looks good. One thought have you tested with an answer with valueCoding? Previously the deepEquals didn't work with it so we may need a specific conditional for it, or perhaps even structuring the comparisons prior

Sure I think I can add some tests for `valueCoding`, I think we can make sure that the `=` and `!=` are working properly for it.

packages/react/src/QuestionnaireForm/QuestionnaireForm.test.tsx
```diff
@@ -968,97 +968,396 @@ describe('QuestionnaireForm', () => {
...
+        isQuestionEnabled(
+          {
+            enableWhen,
+          },
+          {
+            q1: { valueInteger: 4 },
+          }
+        )
+      ).toBe(false);
+    });
```
ThatOneBro, line 1361: Going to add tests for `valueCoding` with `=` and `!=` operators.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -597,30 +598,87 @@ function isDropDownChoice(item: QuestionnaireItem): boolean {
...
+    let match: boolean;
+    // We handle exists separately since its so different in terms of comparisons than the other mathematical operators
+    if (enableWhen.operator === 'exists') {
+      // if actualAnswer is not undefined, then exists: true passes
+      // if actualAnswer is undefined, then exists: false passes
+      match = !!actualAnswer === expectedAnswer.value;
+    } else if (!areNotDifferentTypes(actualAnswer, expectedAnswer)) {
+      match = false;
+    } else {
+      switch (enableWhen.operator) {
```
mattwiller, line 651: The comparison semantics for these values are defined in the [FHIRPath spec](https://hl7.org/fhirpath/#conversion); rather than reimplementing those in this component, we should use [`evalFhirPathTyped()`](https://github.com/medplum/medplum/blob/c7291e67ad636c03379d0c6a58fc12a7086b29f5/packages/core/src/fhirpath/parse.ts#L256) from `@medplum/core` to evaluate expressions like `%actual <= %expected` with the values passed in as `actual` and `expected` variables

ThatOneBro: Okay I made this change but it seems my tests for `valueCoding` when display names are missing or don't match. I'll check around in that `evalFhirPathTyped()` function and see what might be causing that.

ThatOneBro: Thanks for pointing this out by the way, this is way cleaner 👍

ThatOneBro:
Current status: fixing how comparisons are done for certain types of `TypedValue`s under the hood. Finishing #2717 first

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -604,23 +605,48 @@ export function isQuestionEnabled(
...
+    // We handle exists separately since its so different in terms of comparisons than the other mathematical operators
+    if (operator === 'exists') {
+      // if actualAnswer is not undefined, then exists: true passes
+      // if actualAnswer is undefined, then exists: false passes
+      match = !!actualAnswer === expectedAnswer.value;
+    } else if (actualAnswer === undefined) {
+      match = false;
+    } else {
+      const [{ value }] = evalFhirPathTyped(
+        `%actualAnswer ${operator?.replace('=', '~')} %expectedAnswer`,
```
rahul1, line 640: TIL the `~` operator in FHIRPath

ThatOneBro: Yeah Matt was super helpful on that. He noticed that the `=` operator for `enableWhen` should actually be `~` in FHIRPath. Funny.

rahul1 (commented):
This looks good. Will approve once the build is fixed!

packages/core/src/fhirpath/utils.ts
```diff
@@ -296,8 +299,25 @@ export function fhirPathEquivalent(x: TypedValue, y: TypedValue): TypedValue[] {
...
+  if (xType === 'Coding' && yType === 'Coding') {
+    if (typeof xValue !== 'object' || typeof yValue !== 'object') {
+      return booleanToTypedValue(false);
+    }
+    // "In addition, for Coding values, equivalence is defined based on the code and system elements only.
+    // The version, display, and userSelected elements are ignored for the purposes of determining Coding equivalence."
+    // Source: https://hl7.org/fhir/fhirpath.html#changes
+
+    // Basically checking if one of the `Coding` resources has a system defined. If so, then the two's system values must be compared
+    const systemsAreEquivalent =
```
mattwiller, line 312: If both are undefined, wouldn't the equality check return `true` even without this ternary?

ThatOneBro: You're right, one of the downsides to coding late at night is that you try to put unnecessary ternaries everywhere

mattwiller (approved):

packages/core/src/fhirpath/utils.ts
```diff
@@ -296,8 +299,25 @@ export function fhirPathEquivalent(x: TypedValue, y: TypedValue): TypedValue[] {
...
+    // If both have undefined `system` fields, If so, then the two's `system` values must be compared.
+    // Essentially they must both be `undefined` or both the same.
+    return booleanToTypedValue(
+      (xValue as Coding).code === (yValue as Coding).code && (xValue as Coding).system === (yValue as Coding).system
+    );
+  }
+
   if (typeof xValue === 'object' && typeof yValue === 'object') {
-    return booleanToTypedValue(deepEquals(xValue, yValue));
+    return booleanToTypedValue(deepEquals({ ...xValue, id: undefined }, { ...yValue, id: undefined }));
```
codyebberson, line 320: @ThatOneBro  Is this in the spec somewhere?

codyebberson: Woah!  TIL: https://build.fhir.org/fhirpath.html#changes

"Equivalence works in exactly the same manner, but with the addition that for complex types, equality requires all child properties to be equal, except for "id" elements.'

ThatOneBro: Yes it is, maybe I should add a comment about that too...

Here's what the spec says:
`Equivalence works in exactly the same manner, but with the addition that for complex types, equality requires all child properties to be equal, except for "id" elements.`

https://hl7.org/fhir/fhirpath.html#changes

PR #2716, opened 2023-08-24
Example on Lab Ordering demo

jamestouri:
This PR has two things: an example of how our lab ordering questionnaire will look as a JSON object, and the addition of getAllQuestionnaireAnswers - we will need to make this change as it will be needed for the enableWhen field. 

The flow of the questionnaire lab order is below. I removed all the changes in the QuestionnaireForm for future PR, so it will not work as the video shows

https://github.com/medplum/medplum/assets/6913745/6bfd35a4-de21-4217-9c1e-f20add4ab6f7

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -258,24 +258,51 @@ export function calculateAgeString(birthDateStr: string, endDateStr?: string): s
 }
 
 /**
- * @deprecated Returns all questionnaire answers as a map by link ID.
+ * Returns all questionnaire answers as a map by link ID.
  * @param response The questionnaire response resource.
  * @returns Questionnaire answers mapped by link ID.
  */
 export function getQuestionnaireAnswers(
   response: QuestionnaireResponse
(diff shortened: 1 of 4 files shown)


packages/core/src/utils.ts
```diff
@@ -270,11 +270,6 @@ export function getQuestionnaireAnswers(
   return result;
 }
 
-/**
- * Recursively builds the questionnaire answer items map.
- * @param items The current questionnaire response items.
- * @param result The cumulative result map.
- */
 function buildQuestionnaireAnswerItems(
```
jamestouri, line 273: QuestionnaireForms.tsx is still using this method for now until we have the isQuestionEnabled set for getAllQuestionnaireAnswers

codyebberson: Can you elaborate on this?  Is that something we need to fix?  It looks like the new methods are almost exact duplicates of the existing methods, so I'm inclined to just go ahead and fix whatever we need to fix in `isQuestionEnabled`

codyebberson (approved):

PR #2722, opened 2023-08-25
fix(admin/client): use `:projectId` param to create client as superadmin

### LINKED ISSUE #2717: ClientApplication endpoint ignores :projectId as SuperAdmin
The endpoint `/admin/projects/:projectId/client` currently ignores the `:projectId` param. This prevents users from inviting 

We should replicate the functionality In the `/invite` endpoint:
- Most of the time, we read the project id from res.locals
- But if the user is Superadmin, wer read the project id from the URL params

https://github.com/medplum/medplum/blob/c7291e67ad636c03379d0c6a58fc12a7086b29f5/packages/server/src/admin/invite.ts#L44-L50

ThatOneBro:
Fixes #2717

Originally wanted to use `meta` fields on `ClientApplication` to check for the project ID, but may have found a bug around the meta field being clobbered by addition of versioning info on initial creation of a resource, or at least the `ClientApplication`. (Will probably open an issue about this tomorrow once I have more info)

### DIFF
--- a/packages/server/src/admin/client.test.ts
+++ b/packages/server/src/admin/client.test.ts
@@ -1,7 +1,7 @@
 import { randomUUID } from 'crypto';
 import express from 'express';
 import request from 'supertest';
-import { getReferenceString, parseReference } from '@medplum/core';
+import { getReferenceString, resolveId } from '@medplum/core';
 
 import type { ClientApplication, BundleEntry, ProjectMembership } from '@medplum/fhirtypes';
 
@@ -97,7 +97,7 @@ describe('Client admin', () => {
     expect(clientMembership.resourceType).toBe('ProjectMembership');
 
     // Get the id of the project attached to this membership
-    const [, projectId] = parseReference(clientMembership.project) ?? [];
+    const projectId = resolveId(clientMembership.project);
 
     expect(projectId).toBe(project.id);
   });
--- a/packages/server/src/admin/client.test.ts
+++ b/packages/server/src/admin/client.test.ts
@@ -1,9 +1,14 @@
 import { randomUUID } from 'crypto';
 import express from 'express';
 import request from 'supertest';
+import { getReferenceString, parseReference } from '@medplum/core';
+
+import type { ClientApplication, BundleEntry, ProjectMembership } from '@medplum/fhirtypes';
+
 import { initApp, shutdownApp } from '../app';
 import { registerNew } from '../auth/register';
 import { loadTestConfig } from '../config';
+import { createTestProject, initTestAuth } from '../test.setup';
 
 const app = express();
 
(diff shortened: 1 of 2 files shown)


packages/server/src/admin/client.test.ts
```diff
@@ -58,4 +63,42 @@ describe('Client admin', () => {
...
+    // Find the membership
+    const clientMembership = res2.body.entry.find(
+      (e: BundleEntry<ProjectMembership>) =>
+        e.resource?.profile?.reference === getReferenceString(res.body as ClientApplication)
+    )?.resource;
+    expect(clientMembership).toBeDefined();
+    expect(clientMembership.resourceType).toBe('ProjectMembership');
+
+    // Get the id of the project attached to this membership
+    const [, projectId] = parseReference(clientMembership.project) ?? [];
```
rahul1, line 100: We actually have another helper function, `resolveId`, that just returns the id

packages/server/src/admin/client.test.ts
```diff
@@ -58,4 +63,42 @@ describe('Client admin', () => {
...
+  test('Create client as superadmin - verify projectID', async () => {
+    const { project } = await createTestProject();
+
+    // As a superadmin, create a new client
+    const superAdminAccessToken = await initTestAuth({ superAdmin: true });
+    const res = await request(app)
+      .post('/admin/projects/' + project.id + '/client')
+      .set('Authorization', 'Bearer ' + superAdminAccessToken)
+      .send({
+        name: 'Another one',
```
rahul1, line 76: We should keep these names descriptive. Feel free to reach out if you want to talk through some potential names

packages/server/src/admin/client.test.ts
```diff
@@ -58,4 +63,42 @@ describe('Client admin', () => {
...
+  test('Create client as superadmin - verify projectID', async () => {
+    const { project } = await createTestProject();
+
+    // As a superadmin, create a new client
+    const superAdminAccessToken = await initTestAuth({ superAdmin: true });
+    const res = await request(app)
+      .post('/admin/projects/' + project.id + '/client')
+      .set('Authorization', 'Bearer ' + superAdminAccessToken)
+      .send({
+        name: "Super Admin's test client",
```
ThatOneBro, line 76: Is this in the right direction?

packages/server/src/admin/client.ts
```diff
@@ -22,9 +22,17 @@ export async function createClientHandler(req: Request, res: Response): Promise<
     return;
   }
 
-  const client = await createClient(res.locals.repo as Repository, {
+  let project: Project;
+  const { project: localsProject, repo } = res.locals as { project: Project; repo: Repository };
+  if (localsProject.superAdmin) {
```
codyebberson, line 27: Yep 👍

packages/server/src/admin/client.test.ts
```diff
@@ -1,9 +1,14 @@
 import { randomUUID } from 'crypto';
 import express from 'express';
 import request from 'supertest';
+import { getReferenceString, resolveId } from '@medplum/core';
+
```
codyebberson, line 5: Remove blank lines & "Organize imports"

codyebberson (approved):
Nice, one formatting nit, otherwise lgtm 👍

packages/server/src/admin/client.test.ts
rahul1: Yes, looks great!

packages/server/src/admin/client.test.ts
codyebberson: Yeah... this is a gap in our ESLint config right now.  Currently we're just using the VS Code default behavior.  It would be nice to enforce with ESLint/Prettier.

PR #2723, opened 2023-08-25
feat(Storybook): enable viewing story source code

ThatOneBro:
I thought it may be beneficial to enable viewing source code for each story in Storybook so that devs checking out different components can easily inspect the source and even copy-paste it directly from the story. 

Thoughts?

Screen Shot 2023-08-25 at 4 43 33 AM

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -32,7 +32,7 @@
     "@storybook/addon-actions": "7.3.2",
     "@storybook/addon-essentials": "7.3.2",
     "@storybook/addon-links": "7.3.2",
-    "@storybook/addon-storysource": "^7.3.2",
+    "@storybook/addon-storysource": "7.3.2",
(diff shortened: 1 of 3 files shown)


rahul1:
Yes thank you! We had this feature enabled in an older version of storybook, but lost it somehow on an upgrade

rahul1 (approved):

packages/react/package.json
```diff
@@ -32,6 +32,7 @@
     "@storybook/addon-actions": "7.3.2",
     "@storybook/addon-essentials": "7.3.2",
     "@storybook/addon-links": "7.3.2",
+    "@storybook/addon-storysource": "^7.3.2",
```
jamestouri, line 35: Very small nit, I know Cody likes to remove the `^` in package.json versions

codyebberson (approved):

PR #2742, opened 2023-08-28
Fixes #2568 - move react native example into monorepo

### LINKED ISSUE #2568: Move medplum react native to examples

codyebberson:


codyebberson:
Before:

* `node_modules`: 1.3 GB
* `package-lock.json`: 1.7 MB, 44k lines

After:

* `node_modules`: 1.5 GB (+200 MB)
* `package-lock.json`: 2 MB, 52k lines

I'm not sure this is worth it :/

ThatOneBro:
LGTM, I have been working off this branch trying to get the app to run on an Android device. Mostly there, but I think that shouldn't block this PR.

It's unfortunate it adds more to `node_modules` but it does reduce some of the maintenance burden of upgrading deps and I think we should definitely keep this example maintained as it is definitely a good selling point for those looking to use `Medplum` for mobile.

Feel free to merge if you think it's worth the tradeoff.

ThatOneBro (approved):

PR #2748, opened 2023-08-29
Experiment: remove prettier from ESLint config, try out Biome (Rome fork)

ThatOneBro:
## Background
Our linting is pretty slow. I'm usually waiting around 8 seconds for red lines to disappear that shouldn't have popped up in the first place. There's a few reasons for that but that's not super important to this PR. The main thing to note is that `prettier` can run separately in VSCode pretty fast and adding `prettier` to an already slow `ESLint` doesn't make it any faster.

Even after removing it, however, the ESLint + `@typescript-eslint`  experience leaves a lot to be desired in terms of speed.  It's still super great at catching potential catastrophes, though, so it is still valuable despite being slow.

But Biome... yes, Biome.

## Biome? (Rome -> Biome? What was Rome anyways?)
Rome has been the talk of the town in terms of linting and formatting in a very niche, DX-obsessed sub-community on Twitter. It boasts super performance and a great set of defaults for linting and formatting.

The former Rome core team just officially forked and created Biome (literally today before I wrote this PR) as the Rome successor. We were already looking at using Rome as the house linter (and maybe formatter) but the fork and immediate 1.0 coincided with our attempt at adoption. See: https://biomejs.dev/blog/annoucing-biome

After some configuration to fit as close as possible to our current lint rules with `eslint` and `@typescript-eslint`, here is a comparison of the execution time of linting the entire codebase:

Screen Shot 2023-08-29 at 1 44 23 PM

*473 ms vs 2m17s... Nearly 200x faster!*

This would be a great improvement in DX and probably boost productivity and general wellbeing of everyone involved.

The only problem lies in that not all of our rules can be enforced by Biome currently... and some of the really good ones (and the most expensive ones) like `no-floating-promises` are missing.

## The proposal:
I think those who are adventurous enough should try using `Biome` for a while. Disable Eslint in VSCode, enable Biome, and continue as normal. Let me know what you think, if you run into any issues, etc. Then after a few weeks we can evaluate a project-wide recommendation and switch to `Biome` if we think it makes sense.

## What this PR does: 

1. Enable `prettier` as the default formatter for the project. We removed it from the eslint config to speed up eslint a little, but some people may use Eslint as a formatter. Now in `.vscode/settings.json`, we enable Prettier for the default formatter for relevant types of files for this project.
2. Add a `biome.json` to the project root. This contains the config for Biome and the Biome plugin will automatically detect the plugin. A lot of really good rules, especially ones like `noExcessiveComplexity` which we currently enforce with `Sonar` which previously would cause PRs to bounce after waiting for Sonar to run for many, many minutes. Now you find out in milliseconds. Amazing.
3. Fixes some errors based on the current Biome config.

### Next steps:
1. Add a step to run `prettier` on PR submission and report errors on the PR via a GitHub Action. Let the submitter know to enable `prettier` in their editor or run `prettier . --write` before submitting PRs. This will catch any strays missed by the lack of a `prettier` eslint enforcement.
2. Add a similar step for `eslint` once we decide to make the switch as a project to Biome

### Later:
* Re-evaluate if `Biome` should be the default linter in the project. We need to prove that it is consistent and doesn't cause too much friction if we are still running eslint on PRs. My guess is that it will be fine but we want to make sure before we try to get everyone on the bleeding edge with us.
* Consider switching to `Biome` for formatting as well? It could make a significant difference in the time to format larger files. But since most files are not that big anyways it may not matter as much for the project on that front... but would reduce the tooling surface area.
* Consider turning some of the current warnings into errors for things we really want to enforce. Some of them have been turned off due to existing errors in the codebase since these are essentially new rules we will need to apply retroactively. Especially things like `noUnusedVariable` (it seems ESLint is less sensitive than Biome about this rule), `noExplicitAny` (manual overrides only here, most of the time `any` is a code smell), etc.

## How to use Biome in VSCode *(WARNING: STILL EXPERIMENTAL)*:
Biome is pretty easy to get going in VSCode once you install it. For now, we will have a `biome.json` in the project root. You can disable the `ESLint` extension in VSCode, or uninstall it completely if you're that eager to be rid of it. Then you should:

1. Install `Biome` from `npm`. You can install it globally with `npm install -g --save-dev @biomejs/biome`
2. Install the [Biome VSCode plugin](https://marketplace.visualstudio.com/items?itemName=biomejs.biome)
3. Make sure the extension is enabled for the project at least.

And you should be good to go after that! Happy linting.

Please feel free to let me know your thoughts about `Biome` or if you need help getting setup and you want to try it, or if you run into any issues with it at all.

### NOTE:
~~If you're going to use Biome and disable `ESLint` in your editor, you should definitely run `npm run lint` to run `ESLint` from CLI before committing to the repo. This makes sure we still follow our `ESLint` rules but you only have to let `ESLint` run once instead of on every debounced keystroke.~~

### EDIT:
After some experimentation, it's definitely recommended to run both the `ESLint` VSCode plugin and the `Biome` plugin together in VSCode. 

So the recommended VSCode setup is:
* `Prettier` - formatter
* `ESLint` - main linter, most nuanced ruleset, slower feedback
* `Biome` - auxiliary linter, immediate feedback and some cool rulesets our `ESLint` config doesn't have, less picky on important rules that we have in `ESLint` though

In CI:
* `Sonar` - a lot of best practices and other static analysis for code smells and stuff that is a bit heavier and runs only in CI. Will catch stuff about cognitive complexity, code duplication, coverage, etc. Only runs in CI so you don't have to configure anything, just know that `Biome` may catch some of this stuff too now and can potentially save you from bounced PRs if you pay attention to `Biome`


packages/eslint-config/index.cjs
```diff
@@ -6,7 +6,6 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:@typescript-eslint/strict',
     'plugin:jsdoc/recommended-typescript-error',
-    'plugin:prettier/recommended',
```
codyebberson, line 9: Out of curiosity, why was this removed?

packages/eslint-config/package.json
```diff
@@ -22,21 +22,17 @@
...
+    "@typescript-eslint/eslint-plugin": "^6",
+    "@typescript-eslint/parser": "^6",
     "eslint": "^8",
     "eslint-plugin-jsdoc": "^46",
-    "eslint-plugin-prettier": "^4",
-    "eslint-plugin-react-hooks": "^4",
-    "prettier": "^2"
+    "eslint-plugin-react-hooks": "^4"
   },
-  "dependencies": {
```
codyebberson, line 32: 👍

package.json
```diff
@@ -33,6 +33,7 @@
     "jest-environment-jsdom": "29.6.3",
     "npm-check-updates": "16.12.2",
     "nyc": "15.1.0",
+    "prettier": "3.0.3",
```
codyebberson, line 36: Can you help me understand this change?

Before, Prettier and ESLint were driven by `@medplum/eslint-config`, which has `prettier` as a `devDependency`.  Will that no longer be the case?

packages/core/src/types.ts
```diff
@@ -159,11 +159,17 @@ export function indexStructureDefinition(structureDefinition: StructureDefinitio
 
   const elements = structureDefinition.snapshot?.element;
   if (elements) {
-    // First pass, build types
-    elements.forEach((element) => indexType(structureDefinition, element));
+    // @TODO(ThatOneBro 29 Aug 2023): any reason why this can't be done in one loop?
```
codyebberson, line 162: Good question, and something we should discuss as a team.  There are currently a diversity of opinions about "short functional programming" style vs "fast C-like" style.  I understand there are pros/cons to both.

In this particular case, `indexStructureDefinition` is a not on any critical performance paths, so it is unlikely to have any measurable difference.

But I like the idea that we're aligned as a team, and we could write up a doc that describes our philosophy, and even better, codify it in linter rules.

codyebberson (commented):
I'm going to try this out today.  I support it in theory.  Thanks @ThatOneBro!

packages/core/src/types.ts
ThatOneBro: I'm pretty flexible when it comes to how code is written in general, but I do thing `.forEach` is something we should be wary of, in general. I think it becomes easy to forget about potential perf hits especially when it doesn't seem like it may not impact the hot paths, but I think this kind of awareness of when something is okay and when it's not is something that must be fostered in a culture.

For objects you pay most of the cost in the accessing of fields itself, so the performance impact is not as great, but the main thing to be aware of is that there is a non-zero amount of overhead, especially when it comes to `.forEach`. Actually the overhead is quite high for non-objects / primitive values for `.forEach`:

Screen Shot 2023-09-01 at 2 26 30 PM

Over 30x slower for primitive values, insane!

I agree though, this is definitely a topic we should definitely discuss a group and come up with a proper style guide to follow. I think there is definitely a balance to strike and I think it's easy to get nickel'd-and-dime'd on both sides of the spectrum of performance and readability / complexity.

package.json
ThatOneBro: I think there were some problems when it came to running `prettier` is VSCode, so I added it to the root of the project. I think some files were not being formatted or checked by `prettier` previously, so that's why we had so many conflicts in #2758

ThatOneBro: We actually may be able to remove `prettier` from `@medplum/eslint-config` now since it's no longer running in `ESLint` after this PR...

packages/eslint-config/index.cjs
ThatOneBro: Basically this plugin runs `prettier` in `ESLint` but as of #2764, we won't need to do that anymore. It should also speed up `ESLint` a bit.

codyebberson (approved):

ThatOneBro:
During the build after rebasing, I ran into an issue with turbo: 
`ERROR Unable to find Go binary. Please report this issue at https://github.com/vercel/turbo/issues and include your package manager and version along with the following information:`
(See: https://github.com/medplum/medplum/actions/runs/6151160717/job/16690670709#step:6:43)

ThatOneBro:
After getting past the issue with `turbo`, ran into a problem with `docusaurus` getting the wrong version of `react-router` (v5 instead of v6), likely an incorrect hoisting of v6 from `storybook` (see: https://github.com/facebook/docusaurus/issues/7532).

Solved it by nuking again and running `npm install` in `packages/docs` before running it in the root.

PR #2756, opened 2023-08-30
feat(ResourceTimeline): add support for `note`s on `ResourceTimeline`

### LINKED ISSUE #2715: Add `Notes` to the timeline view
Currently, we have a few react components for displaying timeline views: 

- [`Timeline` (generic)](https://storybook.medplum.com/?path=/story/medplum-timeline--basic)
- [`PatientTimeline`](https://storybook.medplum.com/?path=/story/medplum-patienttimeline--patient)
- [`ResourceTimeline`](https://storybook.medplum.com/?path=/story/medplum-patienttimeline--patient)
- [`EncounterTimeline`](https://storybook.medplum.com/?path=/story/medplum-patienttimeline--patient)
- [`ServiceRequestTimeline`](https://storybook.medplum.com/?path=/story/medplum-servicerequesttimeline--basic)
- [`DefaultResourceTimeline`](https://storybook.medplum.com/?path=/story/medplum-servicerequesttimeline--basic)

Most of these components do some kind of search for resources related to the root resource, and then display them in chronological order. For instance, many search for `Communication` resources that are linked to the root resource, which are treated as "comments".

However, many resources also contain a `note` field on them, that could also be interpreted as "comments". Some examples: 
- `Observation`
- `ServiceRequest`
- `MediationRequest`
- ...

The request for this issue is to modify the appropriate timeline components to:
1. Check if the root resource has a `note` element
2. If so, add it to the timeline

ThatOneBro:
## Background
There are a few `Resource` types which have a `note` property. Currently the notes from this `note` property only show up in the parent `resource` in the timeline view.

As described in #2715, we want to be able to add the notes on these resources such as `Specimen` or `Observation` to the `ResourceTimeline` as a separate entry, sorted into the timeline based on when the `note` was recorded.

## What this PR does
* Adds notes to the `ResourceTimeline` and derivative timeline components when a `note` property is found on a `Resource`.
* Closes #2715
Screen Shot 2023-08-30 at 5 47 31 PM

## Notes about changes
There were are few things about this issue that made it a little more difficult than originally thought at first glance:
1. `ResourceTimeline` expected items to be `Resource`s
2. Sorting items in the timeline relied on a sorting function `sortByDateAndPriority()` which deals with sorting `Resource`s based on different fields and whether a resource is marked with a special `priority`. This makes it a bit tricky to sort `note`s which wouldn't have some of these fields necessarily and aren't shaped like `Resource`s.

The solution was to make `ResourceTimeline` generic on items which implement an interface `TimeSortable` which guarantees a datatype is a `Resource` or has a `time` property with a date string. We then created a `sortByDateAndPriorityGeneric()` which can sort these objects alongside resources. In the future, if needed, more logic can be added to extract different fields from `Resource`s and put them and sort them in the timeline as well.

### DIFF
--- a/packages/react/src/ResourceTimeline/ResourceTimeline.stories.tsx
+++ b/packages/react/src/ResourceTimeline/ResourceTimeline.stories.tsx
@@ -1,6 +1,6 @@
 import { createReference, MedplumClient, ProfileResource } from '@medplum/core';
 import { Attachment, Bundle, Encounter, ResourceType } from '@medplum/fhirtypes';
-import { HomerEncounter } from '@medplum/mock';
+import { HomerEncounter, HomerSimpsonSpecimen } from '@medplum/mock';
 import { Meta } from '@storybook/react';
 import React from 'react';
 import { Document } from '../Document/Document';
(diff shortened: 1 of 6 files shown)


ThatOneBro:
Closing this one for now after talking about product direction for the React component suite. This seems like it may be a bit too opinionated as is but we may come back to this later

Issue #2418, opened 2023-07-10
Documentation on project admin page 


rahul1:
Clarify how to set an existing user to admin state, if you are already an admin.

- Add content to this page in [this section ](https://www.medplum.com/docs/auth/user-management-guide#user-administration-via-medplum-app)
   - Add a section "Promote existing user to administrator", for both the App and API
   - Link back to [this page](https://www.medplum.com/docs/app/invite) to refer how to invite _new users_ as admins
- On [this page](https://www.medplum.com/docs/app/invite), add a bullet about how to invite a user as an admin
   - Link back to  [user admin guide ](https://www.medplum.com/docs/auth/user-management-guide#user-administration-via-medplum-app) with a pointer like "see here for more details"

ayobamiseun:
can i work on this i am a technical writer and also build react app

reshmakh:
On this one - we want to do the following:
1) Add content to this page in this section https://www.medplum.com/docs/auth/user-management-guide#user-administration-via-medplum-app
2) Add a section "Promote existing user to administrator"
3) Write the instruction on how to do that including links like so https://app.medplum.com/admin/users

rahul1:
Original thread here: https://medplum.slack.com/archives/C058YJKU955/p1689011081840229

I'll just add that we should add a stub article in the "Medplum Apps" section as well, that points to the new section. That stub can be called "Project Administration", and link to the user management guide as the source of truth

ksmith94:
I'll work on this issue

rahul1:
Fixed by #2757

PR #2757, opened 2023-08-31
Create docs for promoting a user to admin

ksmith94:
These changes are to provide documentation on how to promote an existing user to an admin in the app and via the API.


rahul1 (approved):
There is one incorrect URL, but this can be merged after that has been corrected

ksmith94:
Related to issue #2418

PR #2766, opened 2023-09-04
Use Turborepo cache in Github Actions

codyebberson:
Following recommendations here: https://turbo.build/repo/docs/ci/github-actions


ThatOneBro (approved):
LGTM when `prettier` check passes

PR #2767, opened 2023-09-04
config(prettier): only run `prettier` on JS/TS-like files from CLI

ThatOneBro:
## Background
I noticed `.mdx` and `.md` files do not format stably with `prettier` for some reason. Every time `prettier` runs on them it creates conflicts, usually due to whitespacing (adding or removing a space that it itself added before). 

So I figured since `prettier` is already going to run on `.mdx` files when edited in VSCode, we don't really need to have it on the CI or CLI `prettier` runs. 

This will create less noise and only error on JS-like files in CI (already configured this way) and only create changes to JS-like files when run from CLI.

## What this PR does
* Changes the `prettier` script to only write changes to JS/TS-like files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "build": "turbo run build",
     "test": "turbo run test",
     "lint": "eslint .",
-    "prettier": "prettier --write .",
+    "prettier": "prettier --write \"**/*.{ts,tsx,cts,mts,js,jsx,cjs,mjs}\"",
     "cyclonedx": "cyclonedx-npm --omit dev --omit optional --omit peer"
   },
   "workspaces": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "build": "turbo run build",
     "test": "turbo run test",
     "lint": "eslint .",
-    "prettier": "prettier --write .",
+    "prettier": "prettier --write \"**/*.{ts,tsx,cts,mts,js,jsx,cjs,mjs}\"",
```
codyebberson, line 12: We should definitely include `.json` and `.md`.

Would it be more maintainable to just exclude `.mdx`?

codyebberson (approved):

PR #2780, opened 2023-09-06
fix(react-native-example): get login working on mobile

ThatOneBro:
## Background
Previously the React Native example worked on desktop, however, as soon as you tried to run it on a real Android device, you would run into a few issues.

There are a lot of web APIs that `MedplumClient` uses that are not supported on React Native on Android devices. Namely these are the `crypto` global, `SessionStorage`, `TextEncoder`, `atob` and `btoa`, `window.location`. 

## What this PR does
This PR adds a new package in `packages/expo-polyfills` (prefixed with `expo` as per the `expo` convention.

This package should be published to NPM and users will be able to import `@medplum/expo-polyfills` in their Expo project and use it as such:
```ts
import { polyfillMedplumWebAPIs } from '@medplum/expo-polyfills';

polyfillMedplumWebAPIs();
```

### Web vs Mobile
On mobile this will polyfill the APIs required to use `MedplumClient` on iOS and Android.
On web these APIs are not required, so this call gets no-op'd.

The nice thing about this is that the user can use the same app code for both mobile and web (at least in the most basic case) and there is no need to create separate entrypoints for web and mobile.


packages/core/src/contenttype.ts
```diff
@@ -15,4 +15,6 @@ export const ContentType = {
   SVG: 'image/svg+xml',
   TEXT: 'text/plain',
   TYPESCRIPT: 'text/typescript',
-};
+} as const;
+
+export type ContentTypeString = (typeof ContentType)[keyof typeof ContentType];
```
codyebberson, line 20: Why was this necessary?

I don't think we want to treat `ContentType` as exhaustive or restrictive.  It's just meant to be a convenient set of constants.

packages/expo-medplum-polyfills/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+For bare React Native projects, you must ensure that you have [installed and configured the `expo` package](https://docs.expo.dev/bare/installing-expo-modules/) before continuing.
+
+### Add the package to your npm dependencies
+
+```
+npm install expo-medplum-polyfills
+```
+
+
```
codyebberson, line 24: Whitespace

packages/expo-medplum-polyfills/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@medplum/expo-medplum-polyfills",
```
codyebberson, line 2: By convention, we leave out `medplum` from project names because it's implicit with the `@medplum` namespace.  So, `@medplum/expo-polyfills`

packages/expo-medplum-polyfills/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "expo",
+    "expo-medplum-polyfills",
+    "ExpoMedplumPolyfills"
+  ],
+  "repository": "https://github.com/medplum/medplum",
+  "bugs": {
+    "url": "https://github.com/medplum/medplum/issues"
+  },
+  "author": "Medplum <dev@medplum.com> (https://github.com/medplum)",
+  "license": "MIT",
```
codyebberson, line 29: We use Apache 2

packages/expo-medplum-polyfills/src/ExpoMedplumPolyfillsModule.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    Object.defineProperty(window, 'crypto', {
+      configurable: true,
+      enumerable: true,
+      get: () => expoWebCrypto,
+    });
+  }
+
+  if (typeof window.location === 'undefined') {
+    const locationUrl = new URL('/', medplum.getBaseUrl()) as URL & { assign: () => void };
```
codyebberson, line 28: `getBaseUrl` returns the API base URL, not the current location.  

I'm not entirely sure what the default behavior of `window.location` should be.... probably something super generic such as `http://localhost:80`?

packages/expo-medplum-polyfills/src/ExpoMedplumPolyfillsModule.ts
```diff
@@ -0,0 +1,103 @@
...
+import 'react-native-url-polyfill/auto';
+import { TextEncoder } from 'text-encoding';
+
+type ExtendedExpoCrypto = typeof expoWebCrypto & {
+  subtle: {
+    digest: (algorithm: AlgorithmIdentifier, data: BufferSource) => Promise<ArrayBuffer>;
+  };
+};
+
+export function polyfillMedplumWebAPIs(medplum: MedplumClient): void {
```
codyebberson, line 14: I don't think this should require `medplum` as an argument?

packages/expo-medplum-polyfills/src/ExpoMedplumPolyfillsModule.ts
```diff
@@ -0,0 +1,103 @@
...
+      get: () => decode,
+    });
+  }
+}
+
+async function polyfilledDigest(algorithm: AlgorithmIdentifier, data: BufferSource): Promise<ArrayBuffer> {
+  return digest(algorithm as CryptoDigestAlgorithm, data);
+}
+
+class Storage {
```
codyebberson, line 73: We already have the equivalent of this in `@medplum/core` exported as `MemoryStorage`

codyebberson (commented):
See comments.

Let's also discuss offline w/ @rahul1  - I have mixed feelings about this overall.  I like the intent of supporting React Native, but I don't want to be in the business of supporting expo polyfills.  I want to make sure we've conducted an exhaustive search of 3rd party libs before signing up for maintaining this.

How can we test this?  Is it testable in CI/CD like Github Actions?  Or does it require android and/or xcode?

ThatOneBro:
Testing this new package ended up being a bit more involved than I realized...

Mocking almost any package that touches native bindings is a must in `jest`, and there were a few issues with setting the transforms up.

We must use `ts-jest` directly in order to process `React Native`-flavored TypeScript, and we use `@babel/jest` for JS like in all our others packages.

Also, while testing `expo-polyfills`, I noticed we aren't clearing `sessionStorage` on logout, which definitely seems unintentional. I went ahead and added a line to do that in `MedplumClient.clear()`, since this seems like the desired behavior. Correct me if I'm wrong there.

Otherwise I think this PR should be good to go

packages/core/test.setup.js
```diff
@@ -0,0 +1,7 @@
+/* globals module require globalThis */
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { MemoryStorage } = require('@medplum/core');
+
+module.exports = () => {
+  Object.defineProperty(globalThis.window, 'sessionStorage', { value: new MemoryStorage() });
```
codyebberson, line 6: Out of curiosity, why was this necessary?

packages/core/test.setup.js
```diff
@@ -0,0 +1,7 @@
+/* globals module require globalThis */
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
codyebberson, line 2: We should rename this file to use `.cjs`.  I believe there is a PR open to change `core` to `"type": "module"`

scripts/cp-file.mjs
codyebberson, line 1: It doesn't look like we're using this file?

packages/react/src/test.setup.ts
```diff
@@ -18,3 +18,5 @@ indexStructureDefinitionBundle(readJson('fhir/r4/profiles-resources.json') as Bu
 indexStructureDefinitionBundle(readJson('fhir/r4/profiles-medplum.json') as Bundle);
 indexSearchParameterBundle(readJson('fhir/r4/search-parameters.json') as Bundle<SearchParameter>);
 indexSearchParameterBundle(readJson('fhir/r4/search-parameters-medplum.json') as Bundle<SearchParameter>);
+
+Object.defineProperty(globalThis.window, 'sessionStorage', { value: new MemoryStorage() });
```
codyebberson, line 22: Why was this necessary?

packages/expo-polyfills/esbuild.mjs
```diff
@@ -0,0 +1,37 @@
...
+  bundle: true,
+  platform: '',
+  loader: { '.ts': 'ts' },
+  resolveExtensions: ['.ts'],
+  target: 'es2021',
+  tsconfig: 'tsconfig.json',
+  minify: true,
+  sourcemap: true,
+  plugins: [nodeExternalsPlugin()],
+  external: ['expo-crypto', 'expo-standard-web-crypto'],
```
codyebberson, line 18: Shouldn't all of the `dependencies` be in this list?

```json
  "dependencies": {
    "base-64": "1.0.0",
    "expo-crypto": "12.4.1",
    "expo-standard-web-crypto": "1.6.0",
    "react-native-url-polyfill": "2.0.0",
    "text-encoding": "0.7.0"
  },
```

packages/app/src/resource/BotEditor.tsx
```diff
@@ -64,11 +64,11 @@ export function BotEditor(): JSX.Element | null {
 
   const getCode = useCallback(() => {
     return sendCommand(codeFrameRef.current as HTMLIFrameElement, { command: 'getValue' });
-  }, [codeFrameRef]);
+  }, []);
```
codyebberson, line 67: This is a good change, but I'm curious how it came up in this PR 🙃

packages/expo-polyfills/babel.config.js
codyebberson, line 1: We should probably rename this file to `babel.config.cjs` too

packages/expo-polyfills/jest.config.js
codyebberson, line 1: Historically, we've tried to use JSON-formatted `jest.config.json` to avoid weird CJS/ESM compatibility issues.  That would be my preference unless necessary.  If necessary, then we should rename to `jest.config.cjs`

packages/react/src/test.setup.ts
ThatOneBro: `sessionStorage` does not exist in the `jsdom` environment

packages/app/src/resource/BotEditor.tsx
ThatOneBro: When I was removing the restrictions to `ContentTypeString` that I added in this PR before, I happened to catch it

packages/expo-polyfills/esbuild.mjs
ThatOneBro: Actually I don't think this is even needed, builds without this field on the config... using the `node-externals` plugin makes all imported packages external by default... cool.

packages/core/test.setup.js
ThatOneBro: sessionStorage does not exist in the jsdom environment

ThatOneBro: I think we weren't actually testing `sessionStorage` before

codyebberson (approved):

PR #2787, opened 2023-09-06
UI Clone for Lab Order Questionnaire

jamestouri:
https://github.com/medplum/medplum/assets/6913745/de8ede6e-16e5-454c-ba53-18c69fb85ec3

Add multi-select to repeatable story

https://github.com/medplum/medplum/assets/6913745/5c4058a1-b618-4447-875e-e51ddcd57e93

Sanity check for kitchen sink story

https://github.com/medplum/medplum/assets/6913745/5df57b39-d17b-470a-81f3-451853947b22

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -430,7 +430,7 @@ export function QuestionnaireFormItem(props: QuestionnaireFormItemProps): JSX.El
         <TextInput
           id={name}
           name={name}
-          defaultValue={initial?.valueString}
+          defaultValue={initial?.valueString ?? matchLinkIdValues(props.answers ?? [], item)}
           onChange={(e) => onChangeAnswer({ valueString: e.currentTarget.value }, index)}
         />
       );
@@ -487,7 +487,6 @@ export function QuestionnaireFormItem(props: QuestionnaireFormItemProps): JSX.El
                     { type: 'QuestionnaireItemAnswerOption', value: option },
                     'value'
                   ) as TypedValue;
-
                   return { valueCoding: optionValue.value };
                 });
                 const responses = props.responseItems?.filter((r) => r.linkId === item.linkId) ?? [];
@@ -719,7 +718,7 @@ function buildInitialResponseItems(items: QuestionnaireItem[] | undefined): Ques
 
 function buildInitialResponseItem(item: QuestionnaireItem): QuestionnaireResponseItem {
   return {
-    id: generateId(),
+    id: item.id ?? generateId(),
     linkId: item.linkId,
     text: item.text,
     item: buildInitialResponseItems(item.item),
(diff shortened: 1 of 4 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -790,3 +870,48 @@ function getResponseId(responses: QuestionnaireResponseItem[], index: number): s
...
+  const results = [];
+  for (const answer in answers) {
+    if (answer === item.id) {
+      for (const answerValue of answers[answer]) {
+        const itemValue = getTypedPropertyValue(
+          { type: 'QuestionnaireItemAnswerOption', value: answerValue },
+          'value'
+        ) as TypedValue;
+
+        if (itemValue?.type === 'Coding') {
```
codyebberson, line 908: Why is this special case necessary?

jamestouri: The defaultValue prop in the NativeSelect doesn't accept the object and wants a string. If we pass in the object it won't show in the next components as we step along the questionnaire. 

Since it might not just be valueCoding and passing in valueString or valueInteger I added this edge case.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -790,3 +870,48 @@ function getResponseId(responses: QuestionnaireResponseItem[], index: number): s
...
+  const option = (item.answerOption as QuestionnaireItemAnswerOption[])[0];
+  const optionValue = getTypedPropertyValue(
+    { type: 'QuestionnaireItemAnswerOption', value: option },
+    'value'
+  ) as TypedValue;
+  const propertyName = 'value' + capitalize(optionValue.type);
+
+  const data = (item.answerOption ?? []).map((a) => ({
+    value:
+      propertyName === 'valueCoding' ? a.valueCoding?.code : a[propertyName as keyof QuestionnaireItemAnswerOption],
```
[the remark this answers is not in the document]
jamestouri: Good question, the label needs to remain with a special casing to show the display, however `value` doesn't need this special casing. Will change this now

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
codyebberson: Hmm, ok, I think this is a bug in `<QuestionnaireFormItem>`.  We should not add a special case here, because we're going to tie ourselves in knots inside `<QuestionnaireFormItem>`.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -188,7 +190,7 @@ function QuestionnaireFormItemArray(props: QuestionnaireFormItemArrayProps): JSX
 
   if (props.renderPages) {
     return (
-      <Stepper active={props.activePage ?? 0} allowNextStepsSelect={false}>
+      <Stepper active={props.activePage ?? 0} allowNextStepsSelect={false} style={{ padding: 6 }}>
```
codyebberson, line 193: Stepper props: https://mantine.dev/core/stepper/?t=props

Can you try `p={6}`?

codyebberson (approved):

PR #2812, opened 2023-09-12
fix(auth/external): add some missing url params for external auth `PKCE` exchanges

### LINKED ISSUE #2810: `getExternalAuthRedirectUri` doesn't add PKCE parameters to URL
## Description
When signing in to an external auth provider with PKCE, some identity providers require that the PKCE parameters `code_challenge_method` and `code_challenge` are set as URL parameters

- [Okta](https://developer.okta.com/docs/guides/implement-grant-type/authcodepkce/main/#flow-specifics)
- [Auth0](https://auth0.com/docs/api/authentication#authorization-code-flow-with-pkce)

However, our `getExternalAuthRedirectUriI()` doesn't currently set those two params

https://github.com/medplum/medplum/blob/83432e6403b4d224a573cb459bd4bf91c924e02f/packages/core/src/client.ts#L1050-L1063

ThatOneBro:
Closes #2810

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -29,8 +29,6 @@ import {
   UserConfiguration,
   ValueSet,
 } from '@medplum/fhirtypes';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-/** @ts-ignore */
 import type { CustomTableLayout, TDocumentDefinitions, TFontDictionary } from 'pdfmake/interfaces';
 import { encodeBase64 } from './base64';
 import { LRUCache } from './cache';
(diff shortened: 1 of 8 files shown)


packages/core/src/client.ts
```diff
@@ -29,8 +29,6 @@ import {
   UserConfiguration,
   ValueSet,
 } from '@medplum/fhirtypes';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-/** @ts-ignore */
```
codyebberson, line 33: This is necessary for `pdfmake` to be an optional dependency in TypeScript projects.

packages/core/src/client.ts
```diff
@@ -1053,12 +1051,22 @@ export class MedplumClient extends EventTarget {
     redirectUri: string,
     loginRequest: BaseLoginRequest
   ): string {
+    const { codeChallenge, codeChallengeMethod } = loginRequest;
+    if (!codeChallengeMethod) {
```
codyebberson, line 1055: Is this always true?  I think PKCE is optional for external auth.

packages/core/src/client.ts
```diff
@@ -2716,7 +2724,7 @@ export class MedplumClient extends EventTarget {
    * @category Authentication
    * @returns The PKCE code challenge details.
    */
-  async startPkce(): Promise<{ codeChallengeMethod: string; codeChallenge: string }> {
+  async startPkce(): Promise<{ codeChallengeMethod: 'plain' | 'S256'; codeChallenge: string }> {
```
codyebberson: Might as well use the `CodeChallengeMethod` type here.

packages/core/src/client.ts
ThatOneBro: In the way that it's implemented there is always a call to `ensureCodeChallenge()` as far as I can tell. I'll double check

packages/core/src/client.ts
ThatOneBro: Ah yeah I think auto-organize messed things up and put it somewhere it didn't make sense. I'll fix that

packages/core/src/client.ts
ThatOneBro: This is currently the only place where `client.getExternalAuthRedirectUri()` is called:

https://github.com/medplum/medplum/blob/8be748f9ef5da63aa80eaa0d3154a06e82e634df/packages/core/src/client.ts#L1014-L1015

So in the current codebase it will always have a `codeChallenge` and `codeChallengeMethod`. I think in this case it's better to do this check because if they aren't there in the current codebase it is likely a bug and not intentional. 

I could see a case for it being more general maybe, but if it's not planned yet then this is probably better IMO.

packages/core/src/client.ts
ThatOneBro: Done

codyebberson (approved):

PR #2836, opened 2023-09-14
Persistent state for components in previous pages

jamestouri:
In order to persist state in the components when using the Stepper, we have to pass down the responses object, which stores all data and isn't flattened out. 

The `answers` state was primarily used for the isQuestionEnabled method, removing this state from the component meant having to use the response object. 

Added state - `groupSequence` 
   - Since the Groups are identical - we need to differentiate them when adding the value, otherwise all of the groups are going to show the same data

I'm open to discussing the use of `response` as a context instead of prop drilling it. However I'm curious to hear the thoughts of the trade offs when adding an extra layer abstraction and complexity. 

**Thoughts** - if the questionnaire is starting to get hefty, perhaps we look at a library that could help with the Questionnaire, even a lightweight one could help. 

https://github.com/medplum/medplum/assets/6913745/a8061bb8-c783-4184-8e02-3b50bb22c72c

And the enablement stories 

https://github.com/medplum/medplum/assets/6913745/edd11961-2ffd-44c8-8c6b-2c2680cbea6c

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -1,7 +1,6 @@
 import { Anchor, Button, Group, Stack, Stepper, Title } from '@mantine/core';
 import {
   createReference,
-  getAllQuestionnaireAnswers,
   getExtension,
   getReferenceString,
   IndexedStructureDefinition,
@@ -61,12 +60,8 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element | 
       resourceType: 'QuestionnaireResponse',
       item: newResponseItems,
     };
+
     setResponse(newResponse);
-    // const newAnswers = getAllQuestionnaireAnswers(newResponse);
-    // setAnswers((prevAnswers) => ({
-    //   ...prevAnswers,
-    //   ...newAnswers,
-    // }));
   }
 
   if (!schema || !questionnaire) {
@@ -118,6 +113,7 @@ interface QuestionnaireFormItemArrayProps {
   items: QuestionnaireItem[];
   // answers: Record<string, QuestionnaireResponseItemAnswer[]>;
   responses: QuestionnaireResponseItem[];
+  groupNumber?: number;
   renderPages?: boolean;
   activePage?: number;
   onChange: (newResponseItems: QuestionnaireResponseItem[]) => void;
@@ -148,6 +144,7 @@ function QuestionnaireFormItemArray(props: QuestionnaireFormItemArrayProps): JSX
             key={`${item.linkId}-${index}`}
             item={item}
             index={index}
+            groupNumber={props.groupNumber}
             responses={props.responses}
             responseItems={responseItems}
             setResponseItem={setResponseItem}
@@ -160,6 +157,7 @@ function QuestionnaireFormItemArray(props: QuestionnaireFormItemArrayProps): JSX
         key={`${item.linkId}-${index}`}
         item={item}
         index={index}
+        groupNumber={props.groupNumber}
         responses={props.responses}
         responseItems={responseItems}
         setResponseItem={setResponseItem}
@@ -182,6 +180,7 @@ interface QuestionnaireFormArrayContentProps {
   index: number;
   // answers: Record<string, QuestionnaireResponseItemAnswer[]>;
   responses: QuestionnaireResponseItem[];
+  groupNumber?: number;
   responseItems: QuestionnaireResponseItem[];
   setResponseItem: (responseId: string, newResponseItem: QuestionnaireResponseItem) => void;
 }
@@ -199,6 +198,7 @@ function QuestionnaireFormArrayContent(props: QuestionnaireFormArrayContentProps
         key={props.item.linkId}
         item={props.item}
         responses={props.responses}
+        groupNumber={props.groupNumber}
         responseItems={props.responseItems}
         onChange={(newResponseItem) => props.setResponseItem(newResponseItem.id as string, newResponseItem)}
       />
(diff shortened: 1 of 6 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -111,26 +109,27 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
     buildInitialResponseItems(props.items)
   );
 
   function setResponseItem(responseId: string, newResponseItem: QuestionnaireResponseItem): void {
-    const itemExists = responseItems.some((r) => r.id === responseId);
+    const itemExists = currentResponseItems.some((r) => r.id === responseId);
     let newResponseItems;
     if (itemExists) {
-      newResponseItems = responseItems.map((r) => (r.id === responseId ? newResponseItem : r));
+      newResponseItems = currentResponseItems.map((r) => (r.id === responseId ? newResponseItem : r));
```
jamestouri, line 128: I will be pushing a follow-up PR for this specifically which is to merge the previous response with the newResponseItem. Since the responses are nested - we have to recursively look into the inputs and merge them with the new response. 

As of now, when we step back into a previous page, if we make changes to a group, and we press next, it will clear all other inputs in the group that wasn't changed at that moment.

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -174,13 +175,14 @@ function QuestionnaireFormItemArray(props: QuestionnaireFormItemArrayProps): JSX
 interface QuestionnaireFormArrayContentProps {
   item: QuestionnaireItem;
   index: number;
-  answers: Record<string, QuestionnaireResponseItemAnswer[]>;
-  responseItems: QuestionnaireResponseItem[];
+  responses: QuestionnaireResponseItem[];
```
codyebberson, line 178: How should I interpret `responses` vs `currentResponseItems`?  

Is `responses` "all" responses for the entire Questionnaire, and `currentResponseItems` just the items for this particular item?

If we need both, maybe we can brainstorm on naming that clarifies.

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -63,15 +64,16 @@ export function QuestionnaireFormItem(props: QuestionnaireFormItemProps): JSX.El
...
   switch (type) {
     case QuestionnaireItemType.boolean:
       return (
         <CheckboxFormSection key={props.item.linkId} title={props.item.text} htmlFor={props.item.linkId}>
           <Checkbox
             id={props.item.linkId}
             name={props.item.linkId}
-            defaultChecked={initial?.valueBoolean}
+            defaultChecked={
+              initial?.valueBoolean ?? getDefaultAnswer(props.responses, item, index, props.groupSequence)
```
codyebberson, line 75: Should `getDefaultAnswer` be before `initial?.valueBoolean`?  

If a user enters an answer, that should take precedence, right?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
codyebberson: Or is this what you're referring to in your comment above?  If so, then disregard

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -429,3 +441,48 @@ function typedValueToString(typedValue: TypedValue | undefined): string | undefi
...
+  }
+  return result;
+}
+
+function getItemValue(answer: QuestionnaireResponseItemAnswer): any {
+  const itemValue = getTypedPropertyValue(
+    { type: 'QuestionnaireItemAnswerOption', value: answer },
+    'value'
+  ) as TypedValue;
+  if (itemValue?.type === 'Coding') {
```
codyebberson, line 467: I'm trying to remember our previous conversation about this... Is there a reason we can't handle this special case in the form item rather than here?

If nothing else, I'm nervous that this will not be sufficient.  Many `Coding` objects only have a `code`, and no `display`.

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -429,3 +457,45 @@ function typedValueToString(typedValue: TypedValue | undefined): string | undefi
...
+  return result;
+}
+
+function getItemValue(answer: QuestionnaireResponseItemAnswer): any {
+  const itemValue = getTypedPropertyValue(
+    { type: 'QuestionnaireItemAnswerOption', value: answer },
+    'value'
+  ) as TypedValue;
+  // formatCoding returns '' if nothing is there so we need to use ||
+  return formatCoding(itemValue?.value) || itemValue?.value;
```
jamestouri, line 484: Looked into CodingDisplay and used this, figure this is more elegant than an if statement but if you prefer that or a ternary operator I'll change it

codyebberson (approved):

codyebberson (changes requested):
This isn't working for me on Storybook.  expected?

https://medplum-storybook-git-james-persist-form-state-medplum.vercel.app/?path=%2Fstory%2Fmedplum-questionnaireform--page-sequence

jamestouri:
> This isn't working for me on Storybook. expected?
> 
> https://medplum-storybook-git-james-persist-form-state-medplum.vercel.app/?path=%2Fstory%2Fmedplum-questionnaireform--page-sequence

@codyebberson If you're talking about the Radio Group, just fixed that, Radio uses value instead of defaultValue so changed that. It is expected that if you go to the next page and back the values persist BUT if you change an input it will disappear. 

My follow up PR that makes changes to the setItems() method will fix this. 

Here is what it looks like at the moment

https://github.com/medplum/medplum/assets/6913745/86c71442-cf41-4c0a-84d3-db281a7be331

codyebberson (approved):
Nice 👍

PR #2837, opened 2023-09-14
Add request and trace ID to server logs

### LINKED ISSUE #2783: Log request and trace IDs
Attach request and trace IDs to all requests made to Medplum server, allowing clients to set their own (correctly-formatted) trace ID via the `X-Trace-Id` and/or `traceparent` headers.  These IDs should be added to all logs written as part of handling that request.

See #2759 for more information

mattwiller:
The server now generates request and trace IDs (UUIDs) for all server requests, and attaches them to all logs written as part of the request.  The request ID is always generated by the server, but the trace ID can be overridden by the client with a different valid UUID by sending an additional header with the request: either `X-Trace-Id: <trace UUID>` or the W3C standard [`traceparent` header](https://www.w3.org/TR/trace-context/#traceparent-header-field-values).

Instead of passing around these IDs explicitly, a new `RequestContext` object containing the full set of request-scoped values is attached to an [`AsyncLocalStorage`](https://nodejs.org/docs/latest-v18.x/api/async_context.html#class-asynclocalstorage) instance, which can then be used anywhere in the async call chain to retrieve the associated objects for the current request.  This required some extensive refactoring in server tests, to ensure that the async context is always present.

__Note:__ Ignoring whitespace changes should make the diff a bit more manageable to review

Fixes #2783


codyebberson:
@mattwiller  Nice, thanks for doing this.

Parsing this.  Initial impressions are all positive.  This touches a lot, so I want to be thorough here.

image

codyebberson:
@mattwiller How should we think about `getAuthenticatedContext` and `getRequestContext`?  Are they "free" from runtime performance perspective?  If you need to call them 100 or 1000 times, should they be pulled out of the loop?

packages/core/src/utils.ts
```diff
@@ -555,7 +555,7 @@ export function deepClone<T>(input: T): T {
  * @param input The input string.
  * @returns True if the input string matches the UUID format.
  */
-export function isUUID(input: string): boolean {
+export function isUUID(input: string): input is `${string}-${string}-${string}-${string}-${string}` {
```
codyebberson, line 558: Curious if this helps at a particular call site?

packages/server/src/admin/super.test.ts
```diff
@@ -7,12 +7,13 @@ import { initApp, shutdownApp } from '../app';
 import { registerNew } from '../auth/register';
 import { loadTestConfig } from '../config';
 import { systemRepo } from '../fhir/repo';
-import { logger } from '../logger';
+// import { logger } from '../logger';
```
codyebberson, line 10: Cleanup

packages/server/src/admin/super.test.ts
```diff
@@ -202,7 +204,7 @@ describe('Super Admin routes', () => {
     (rebuildR4StructureDefinitions as unknown as jest.Mock).mockImplementationOnce((): Promise<any> => {
       return Promise.reject(err);
     });
-    const loggerErrorSpy = jest.spyOn(logger, 'error').mockReturnValueOnce();
+    // const loggerErrorSpy = jest.spyOn(logger, 'error').mockReturnValueOnce();
```
codyebberson, line 207: Cleanup

packages/server/src/admin/super.test.ts
```diff
@@ -212,7 +214,9 @@ describe('Super Admin routes', () => {
       .send({});
 
     expect(res.status).toEqual(202);
-    expect(loggerErrorSpy).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({ error: err }));
+    const job = await waitForAsyncJob(res.headers['content-location'], app, adminAccessToken);
+    expect(job.status).toEqual('error');
+    // expect(loggerErrorSpy).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({ error: err }));
```
codyebberson, line 219: Cleanup

packages/server/src/app.test.ts
```diff
@@ -74,23 +74,18 @@ describe('App', () => {
...
-    expect(console.log).toHaveBeenCalledWith(
-      expect.stringMatching(
-        /^\{"level":"ERROR","timestamp":"\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d\.\d{3}Z","msg":"Unhandled error","error":"Error: Catastrophe!"\}$/
-      )
-    );
     await shutdownApp();
   });
 
-  test('Database disconnect', async () => {
+  test.skip('Database disconnect', async () => {
```
codyebberson, line 80: Is this due to `runInBand`?

packages/server/src/app.ts
```diff
@@ -109,10 +110,19 @@ function errorHandler(err: any, req: Request, res: Response, next: NextFunction)
...
-  logger.error('Unhandled error', err);
+  try {
+    getRequestContext().logger.error('Unhandled error', err);
+  } catch {
+    globalLogger.error('Unhandled error', err);
+  }
   res.status(500).json({ msg: 'Internal Server Error' });
 }
 
+async function attachRequestContext(req: Request, res: Response, next: NextFunction): Promise<void> {
```
codyebberson, line 121: Could this be moved into `middleware.ts` so all of the context code is in a single place?

packages/server/src/fhir/binary.ts
```diff
@@ -5,20 +5,21 @@ import internal from 'stream';
...
 import { asyncWrap } from '../async';
 import { sendOutcome } from './outcomes';
-import { Repository } from './repo';
 import { getPresignedUrl } from './signer';
 import { getBinaryStorage } from './storage';
+import { authenticateRequest } from '../oauth/middleware';
+import { getAuthenticatedContext } from '../context';
 
-export const binaryRouter = Router();
+export const binaryRouter = Router().use(authenticateRequest);
```
codyebberson, line 13: Was this not handled upstream in app.ts?

mattwiller:
> How should we think about `getAuthenticatedContext` and `getRequestContext`? Are they "free" from runtime performance perspective? If you need to call them 100 or 1000 times, should they be pulled out of the loop?

I benchmarked the call to `AsyncLocalStorage#getStore()` and it takes ~100 ns — so pretty speedy and not something I would be too worried about calling repeatedly, but definitely worth extracting from a hot loop since it's guaranteed not to change between loop iterations anyway.

packages/server/src/logger.ts
```diff
@@ -11,36 +12,97 @@ import { CloudWatchLogger } from './util/cloudwatch';
 
 export enum LogLevel {
   NONE = 0,
```
codyebberson, line 14: If you change `NONE` to 5, you can remove the special case.

codyebberson: Although, to me, the previous numeric values made more sense, in the colloquial sense of "use a higher log level" == "log more stuff"

codyebberson (commented):
Nice job, this is great work.

A bunch of comments, most are minor.  The 2 important ones are (1) consolidating context logic into middleware.ts, and (2) the disabled database test.

Don't worry about the "Missing rate limiting" errors.

Next steps, I'm going to deploy this to staging for manual testing.

Thanks @mattwiller !

codyebberson:
Deployed to https://app.staging.medplum.com/

Cursory review looks good

packages/core/src/utils.ts
mattwiller: Yeah, I remember something in the request ID middleware needed it

packages/server/src/app.test.ts
mattwiller: It was!

packages/server/src/logger.ts
mattwiller: Okay yeah that clicks for me — I'll switch this back

codyebberson (approved):
Awesome, this is good to go.

@mattwiller  - If it's ok with you, my preference would be to merge either after hours tonight or over the weekend just to let things shake out while traffic is low.

PR #2841, opened 2023-09-15
Fixes #2835 - resource crawler, download all attachments

### LINKED ISSUE #2835: Support URL Download for all `Attachement` types
## Use Case
We have a number of users who are migrating chat / SMS messages from Twilio to Medplum. 

Some of these messages have media attachments, and the best way to model these attachments is using `Communication.payload[n].contentAttachment`, which is of type `Attachment`.

It would greatly speed up customer migration pipelines if they could create simply create the appropriate `Communication` resources with external content urls, rather than having to download the media to the client, then re-upload to `Medplum`. 

### Current Workflow
```ts
const existingCommunication = await medplum.searchOne('Communication', `identifier=${twilioId}`);
if(!existingCommunication) {
  // Download the attachment from twiliio
 // Create the medplum binary
 // Create the communication
}
```

### Proposed Workflow
```ts
await medplum.executeBatch(
  {
    resourceType: 'Bundle', 
    entry: [
      //...
       {
        request: {
           ifNoneExist: `identifier=${twilioId}`
        }
         resource: {
            resourceType: 'Communciation', 
            payload: {
              contentAttachment: {url: twilioURL }
            }
          }
        }
     ]
  }
);
```

## Proposal

Our current implementation of the background downloader specifically looks for `Media` resources with external URLs to trigger download jobs

https://github.com/medplum/medplum/blob/83432e6403b4d224a573cb459bd4bf91c924e02f/packages/server/src/workers/download.ts#L106-L112

The proposal would be have this behavior apply to any `Attachment` on any resource type (e.g. `Communication.payload.contentAttachment`)

codyebberson:


packages/core/src/typeschema/crawler.ts
```diff
@@ -0,0 +1,113 @@
...
+import { getDataType, InternalTypeSchema } from './types';
+
+export interface ResourceVisitor {
+  onEnterObject?: (path: string, value: TypedValue, schema: InternalTypeSchema) => void;
+  onExitObject?: (path: string, value: TypedValue, schema: InternalTypeSchema) => void;
+  visitProperty?: (
+    parent: TypedValue,
+    key: string,
+    path: string,
+    propertyValues: (TypedValue | TypedValue[] | undefined)[],
```
mattwiller, line 14: The nested array is only necessary when retrieving data from a deep path: if this is only looking at a single direct property of one object in the JSON, this could probably be `TypedValue[] | undefined` or similar.  Maybe worth splitting out the common case from the weird profile case where nested paths are used — perhaps `visitProperty` and `visitDeepPath`?

packages/core/src/typeschema/crawler.ts
```diff
@@ -0,0 +1,113 @@
...
+        for (const value of arrayify(propertyValue) as TypedValue[]) {
+          this.crawlPropertyValue(value, path);
+        }
+      }
+    }
+  }
+
+  private crawlPropertyValue(value: TypedValue, path: string): void {
+    if (!isLowerCase(value.type.charAt(0))) {
+      // Recursively validate as the expected data type
```
mattwiller, line 86:
```suggestion
      // Recursively crawl as the expected data type
```

packages/core/src/typeschema/crawler.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  crawl(): void {
+    this.crawlObject(toTypedValue(this.rootResource), this.schema, this.initialPath);
+  }
+
+  private crawlObject(obj: TypedValue, schema: InternalTypeSchema, path: string): void {
+    if (this.visitor.onEnterObject) {
+      this.visitor.onEnterObject(path, obj, schema);
+    }
```
mattwiller, line 58: Should this handle the resource type check itself, and become e.g. `onEnterResource`?

mattwiller (commented):
Love the clean visitor pattern abstraction! 💖

packages/core/src/typeschema/crawler.ts
codyebberson: If ok with you, this is something I'd like to revisit in future.  I don't fully understand the cases where we need nested arrays.

I noticed that the case where evaluating an `ElementDefinition` without a value is still allocating empty arrays, which seems like low hanging fruit for optimizations.

I assume this was done to handle the slicing case?

I'm comfortable leaving as-is for now, and revisiting as we find more use cases for the crawler.

reshmakh (approved):

PR #2859, opened 2023-09-18
Merge schema/validation systems

codyebberson:
Back in May, we introduced a new [schema validation](https://github.com/medplum/medplum/pull/2119) system.  It is great, more accurate, more powerful, faster, etc, etc.  But we still have the old schema validation system.  We actually use the old schema system for a whole bunch of stuff, such as generating UI's, generating database tables, generating code, etc.  So this PR is intended to merge them once and for all.

- [x] `@medplum/core`
- [x] `@medplum/fhir-router`
- [x] `@medplum/mock`
- [x] `@medplum/generator`
- [x] `@medplum/react`
- [x] `@medplum/app`
- [x] `@medplum/server`
- [x] `base-schema.json`

--------

Deployed to staging: https://app.staging.medplum.com/


packages/core/src/client.test.ts
```diff
@@ -1563,36 +1564,16 @@ describe('Client', () => {
...
-    expect(schema1.types['Patient']).toBeDefined();
-
-    const schema2 = await request2;
-    expect(schema2).toBeDefined();
-    expect(schema2).toEqual(schema1);
-
-    const schema3 = await client.requestSchema('Patient');
-    expect(schema3).toEqual(schema1);
+    await request1;
+    expect(isDataTypeLoaded('Patient')).toBeDefined();
```
mattwiller, line 1576: Should this assertion be `.toBeTruthy()`?

rahul1:
Does this change how we unit test bots as well (no more `indexStructureDefinitionBundle`?)

https://www.medplum.com/docs/bots/unit-testing-bots

packages/generator/src/benchmarks.ts
```diff
@@ -14,9 +14,9 @@ interface Benchmark {
 }
 
 async function runBenchmarks(...benchmarks: Benchmark[]): Promise<void> {
-  indexStructureDefinitionBundle(typesData);
-  indexStructureDefinitionBundle(resourcesData);
-  indexStructureDefinitionBundle(medplumResourcesData);
+  loadDataTypes(typesData);
+  loadDataTypes(resourcesData);
+  loadDataTypes(medplumResourcesData);
```
mattwiller, line 19: These are redundant and can be removed

packages/react/src/FhirPathTable/FhirPathTable.tsx
```diff
@@ -36,7 +36,7 @@ export interface SmartSearchResponse {
  */
 export function FhirPathTable(props: FhirPathTableProps): JSX.Element {
   const medplum = useMedplum();
-  const [schema, setSchema] = useState<IndexedStructureDefinition | undefined>();
+  const [schemaLoaded, setSchemaLoaded] = useState<boolean>(false);
```
mattwiller, line 39: Is the generic type annotation needed here?  I would assume type inference should work correctly for this case

packages/server/src/fhir/structure.ts
```diff
@@ -15,9 +15,9 @@ export function getStructureDefinitions(): IndexedStructureDefinition {
     const dataTypes = readJson('fhir/r4/profiles-types.json') as Bundle<StructureDefinition>;
     const resourceTypes = readJson('fhir/r4/profiles-resources.json') as Bundle<StructureDefinition>;
     const medplumTypes = readJson('fhir/r4/profiles-medplum.json') as Bundle<StructureDefinition>;
-    indexStructureDefinitionBundle(dataTypes);
-    indexStructureDefinitionBundle(resourceTypes);
-    indexStructureDefinitionBundle(medplumTypes);
+    loadDataTypes(dataTypes);
+    loadDataTypes(resourceTypes);
+    loadDataTypes(medplumTypes);
```
mattwiller, line 20: Redundant calls

packages/core/src/typeschema/types.ts
```diff
@@ -17,6 +19,7 @@ export interface InternalTypeSchema {
 }
 
 export interface ElementValidator {
+  elementDefinition: ElementDefinition;
```
mattwiller, line 22: This addition is at odds with the purpose of creating `InternalTypeSchema`, which was to completely encapsulate the data from `ElementDefinition` so it didn't need to be included directly.  Which pieces of data would need to be added for parity?

packages/generator/src/benchmarks.ts
codyebberson: Oops, remnant of mass find/replace.

packages/core/src/typeschema/types.ts
codyebberson: Well, to be frank, this is why I was opposed to adding a new type here in the first place.

The point is that we already have 1000's of lines of code using `ElementDefinition`, which is a perfectly standardized data type.  I can't say off the top of my head which properties we use.  We use it all over the place (`<ResourceTable>` and `<BackboneElementDisplay>` to display properties, `<ResourceForm>` and `<BackboneElementInput>` to generate inputs, `@medplum/generator` to generate type definitions and migrations).

We ***could*** rewrite everything to a new data type.  Or we ***could*** leave all of that existing code alone, which is well tested and stable in production.

If anything, I'm inclined to go the opposite direction, and revert everything to the standardized `ElementDefinition`.

A key part of the Medplum strategy is "stick to standards".  This is a mini case study in why.  If we stick to standard data types, we reduce the risk of rewriting things.

codyebberson:
> Does this change how we unit test bots as well (no more `indexStructureDefinitionBundle`?)

We should discuss.

As currently implemented, yes, `indexStructureDefinitionBundle` should become `loadDataTypes`.

But we have existing customers and existing code using `indexStructureDefinitionBundle`, so we may want keep the existing function, either as a rename or an alias.

packages/core/src/typeschema/types.ts
mattwiller: The central issue I see is that `ElementDefinition` is not 1:1 with the totality of information for handling an element.  It works well to treat it that way in the simple limit for base resource types, but it has its warts (e.g. handling `ElementDefintion(.base).max` correctly) and in more complex scenarios with profiles the abstraction breaks down.  If we want to handle these kinds of complex cases in the React components, I think it makes sense to move more of the logic for handling the underlying FHIR schema information to one place and have as a side benefit a data structure that's simpler to work with in the ways we need to.  Rather than duplicating the data structure for `ElementDefinition`, I was hoping to explore raising the level of abstraction from `ElementDefintion` representing an individual field to accepting as a parameter only the whole type's `StructureDefinition`, which can be parsed ad hoc into the Medplum internal schema format

codyebberson: I understand, and I do see merit in having a custom type.

I also want to be pragmatic.  My view is that the first order of business is to solve the double schema issue.  Updating all references to `ElementDefinition` is a non-trivial task.  I don't consider that a blocker here.

codyebberson:
@mattwiller This is deployed to staging for testing.

https://app.staging.medplum.com/

ThatOneBro (approved):

PR #2861, opened 2023-09-18
Fix 'sign-in' vs 'sign in' in invite email

codyebberson:
This bothered me:

image

https://english.stackexchange.com/questions/49850/sign-in-signin-or-sign-in

> The verb is sign in.
> 
> The noun is sign-in.
> 
> The noun is better with the hyphen, because signin could be confused with an abbreviated signing that's lost its apostrophe.
> 
> Alternatively you could use the more common log in for the verb, and login for the noun.


mattwiller (approved):
✨

codyebberson:
Separate and longer conversation about whether we should standardize on "sign in" vs "log in".

PR #2862, opened 2023-09-18
Fixes #2792 - multiple patient compartments in resource

### LINKED ISSUE #2792: Add all comparment fields for `Communication`
## Use Case

The FHIR spec defines a Communication as with a Patient's compartment if they are referenced by `subject`, `sender`, or `recipient`. However our current implementation only adds a compartment for `subject`.

Implementing this would greatly simplify access policies for Patient Applications, esp. in pediatric care. For most resources, we can give access to the entire patient compartment for a child. However, currently `Communications` require a special search criteria to get right.

There are other resources, like `Coverage`, that also have multiple compartment participation fields, It would be good to include these as well, but `Communication` is the highest priority

codyebberson:


mattwiller (approved):

rahul1 (approved):

codyebberson:
One sec, fixing issue with duplicate values.

PR #2865, opened 2023-09-18
Fix bug on page sequence input changes 

jamestouri:
Currently with the Page Sequence tool, when we go back the input persists, however when we change one of the inputs, the others clear out automatically. Now we change the setItems method so that it merges the previous responses to the newer responses and keeps all of the inputs. 

https://github.com/medplum/medplum/assets/6913745/13f8aa01-04fc-450d-b681-e85fadf8147b


codyebberson (approved):
Nice 👍

PR #2868, opened 2023-09-19
Fixes ElementDefinition.contentReference in deep nested backbone element

codyebberson:
Discovered while working on https://github.com/medplum/medplum/pull/2859

`StructureMap.group.rule.rule` was incorrectly added as a field of `StructureMap` rather than `StructureMapGroupRule` because it was 3 contexts deep, and we only supported 2.

This fix supports any level of context nesting.

### DIFF
--- a/packages/core/src/typeschema/types.ts
+++ b/packages/core/src/typeschema/types.ts
@@ -168,6 +168,9 @@ class StructureDefinitionParser {
         }
 
         if (!parentContext) {
+          // Within R4 StructureDefinitions, there are 2 cases where StructureDefinition.name !== ElementDefinition.path.
+          // For SimpleQuantity and MoneyQuantity, the names are the names, but the root ElementDefinition.path is Quantity.
+          // We need to use StructureDefinition.name for the type name, and ElementDefinition.path for the path.
           const path = elementPath(element, this.root.path);
           if (element.isSummary) {
             this.resourceSchema.summaryProperties?.add(path.replace('[x]', ''));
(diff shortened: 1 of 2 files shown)


packages/core/src/typeschema/types.ts
```diff
@@ -157,12 +158,17 @@ class StructureDefinitionParser {
...
+        while (parentContext) {
+          if (element.path?.startsWith(parentContext.path + '.')) {
+            parentContext.type.fields[elementPath(element, parentContext.path)] = field;
+            break;
+          }
+          parentContext = parentContext.parent;
+        }
+
+        if (!parentContext) {
+          const path = elementPath(element, this.root.path);
```
codyebberson, line 171: The change from `elementPath(element, this.resourceSchema.name)`

to `elementPath(element, this.root.path)`

is not directly related to this PR (my bad), but still a change that we need to make in https://github.com/medplum/medplum/pull/2859

Within R4 `StructureDefinition`s, there are 2 cases where `StructureDefinition.name !== ElementDefinition.path`.  For `SimpleQuantity` and `MoneyQuantity`, the names are the names, but the root `ElementDefinition.path` is `Quantity`.  I guess they're kind of like mini profiles within the R4 spec?  Either way, we need to use `StructureDefinition.name` for the type name, and `ElementDefinition.path` for the path.

mattwiller: Would you mind adding this as a comment?  The context is super helpful around these kinds of edge cases in the code

mattwiller (approved):

PR #2870, opened 2023-09-19
Fixes #2847 - Create new resources from JSON input

### LINKED ISSUE #2847: Add JSON tab to "New resource" page
Context:
* Users want to create FHIR resources in the Medplum app (https://app.medplum.com/)
* Users often want to create by JSON, because they received a FHIR resource as JSON or found some example JSON
* The most common way to do that is using the "New" page (see `packages/app/src/CreateResourcePage.tsx`)
* At present, the "New" page only shows a form view with individual inputs for each resource
* There is no easy or straightforward way to paste in JSON and create the resource

Current "New Patient" page: https://app.medplum.com/Patient/new

image

Feature request:
* Add tabs to the "New" page using the [`<Tabs>` component](https://mantine.dev/core/tabs/) from Mantine
* There should be 2 tabs, "Form" and "JSON"
* The default tab should be "Form" with the current behavior using the `<ResourceForm>` component
* The new tab should be "JSON" which uses the [`<JsonInput>` component](https://mantine.dev/core/json-input/) from Mantine
* Both cases should use the same or similar submit handler based on the current handler which calls `medplum.createResource()`
* On success, both cases should redirect to the new resource page using existing logic

Additional references:

We already use the `<Tabs>` component in a few places which might be helpful references:
* `BatchPage.tsx`: https://github.com/medplum/medplum/blob/main/packages/app/src/BatchPage.tsx#L139
* `ProjectPage.tsx`: https://github.com/medplum/medplum/blob/main/packages/app/src/admin/ProjectPage.tsx#L36
* `ResourcePage.tsx`: https://github.com/medplum/medplum/blob/main/packages/app/src/resource/ResourcePage.tsx#L138

We already use the `<JsonInput>` component in a few places which might be helpful references:
* `BatchPage.tsx`: https://github.com/medplum/medplum/blob/main/packages/app/src/BatchPage.tsx#L172
* `JsonPage.tsx`: https://github.com/medplum/medplum/blob/main/packages/app/src/resource/JsonPage.tsx#L41

@mateolargo - interested in taking a look at this one?

mattlong:
I made a hook to encapsulate the common logic. I didn't see many other hooks; if it's preferable to remove the hook and duplicate the logic let me know!

### DIFF
--- a/packages/app/src/AppRoutes.tsx
+++ b/packages/app/src/AppRoutes.tsx
@@ -51,6 +51,8 @@ import { ResourcePage } from './resource/ResourcePage';
 import { ResourceVersionPage } from './resource/ResourceVersionPage';
 import { SubscriptionsPage } from './resource/SubscriptionsPage';
 import { TimelinePage } from './resource/TimelinePage';
+import { FormCreatePage } from './resource/FormCreatePage';
+import { JsonCreatePage } from './resource/JsonCreatePage';
 
 export function AppRoutes(): JSX.Element {
   return (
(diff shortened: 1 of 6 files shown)


codyebberson:
Thanks @mattlong 

This looks pretty good.  Sorry about the current state of SonarCloud integration -- it recently stopped working correctly on PR's from forks.

There is one issue identified by Sonar that I think we should fix: 
https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=medplum_medplum&open=AYqrZFVGNyH9ypT8q09T

image

My guess is that `handleSubmit` should not be `async`.

I.e., change line 24 to:

```ts
const handleSubmit = (newResource: T): void => {
```

Then should be good to go.

Thank you!

codyebberson (approved):
Thanks!

PR #2883, opened 2023-09-19
Dynamically move items in QuestionnaireBuilder

jamestouri:
Would love some UI feedback as I'm thinking of ways to move some of the actions around 

https://github.com/medplum/medplum/assets/6913745/033e2d40-7ff8-4c02-b28b-8a688e042cda

### DIFF
--- a/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -134,9 +134,6 @@ export function QuestionnaireBuilder(props: QuestionnaireBuilderProps): JSX.Elem
       return;
     }
     const reorderedItems = reorderItems(value.item ?? [], itemToMove, direction);
-    if (!reorderedItems) {
-      return;
-    }
 
     setValue({
       ...value,
--- a/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -737,11 +737,7 @@ function createPage(): QuestionnaireItem {
   } as QuestionnaireItem;
 }
 
-function reorderItems(
-  items: QuestionnaireItem[] | undefined,
-  itemIndex: number,
-  delta: number
-): QuestionnaireItem[] {
+function reorderItems(items: QuestionnaireItem[] | undefined, itemIndex: number, delta: number): QuestionnaireItem[] {
   const currentItems = items ?? [];
   const newIndex = itemIndex + delta;
   if (newIndex < 0 || newIndex >= currentItems.length) {
--- a/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -8,15 +8,16 @@ import {
   QuestionnaireItemAnswerOption,
   Reference,
 } from '@medplum/fhirtypes';
+import { IconArrowDown, IconArrowUp } from '@tabler/icons-react';
 import React, { useEffect, useRef, useState } from 'react';
 import { Form } from '../Form/Form';
 import { useMedplum } from '../MedplumProvider/MedplumProvider';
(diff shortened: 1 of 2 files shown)


packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -106,6 +129,18 @@ export function QuestionnaireBuilder(props: QuestionnaireBuilderProps): JSX.Elem
     };
   }, [defaultValue]);
 
+  function toggleItemOrder(itemToMove: QuestionnaireItem, direction: Direction): void {
```
codyebberson, line 132: If we move this from `QuestionnaireBuilder` into `ItemBuilder`, then I think you should be able to simplify the `reorderItems` method.

Currently, `reorderItems` recursively searches for the item to move.  If you do it in `ItemBuilder`, I believe you should be able to use the array directly, similar to how we do it with `addItem` and `removeItem`

packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -8,15 +8,21 @@ import {
...
+import { QuestionnaireFormItem } from '../QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem';
 import { getValueAndType } from '../ResourcePropertyDisplay/ResourcePropertyDisplay';
 import { ResourcePropertyInput } from '../ResourcePropertyInput/ResourcePropertyInput';
 import { useResource } from '../useResource/useResource';
 import { killEvent } from '../utils/dom';
 import { isChoiceQuestion, QuestionnaireItemType } from '../utils/questionnaire';
-import { QuestionnaireFormItem } from '../QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem';
+
+enum Direction {
+  UP,
```
codyebberson, line 23: I'm not sure if this is useful here, but I've found that using `delta` with values `1` or `-1` can be more elegant than direction enums.  Then it becomes less `if`/`else`, and just math.

codyebberson (commented):
Thanks for taking a stab at this.

See comments about moving logic into `ItemBuilder`

Aesthetically, I like the icon choice.  It's a little on the bold side.  Maybe shrink the icons down ~25%, and make them a more subtle gray?

packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -204,6 +230,15 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
     });
   }
 
+  function toggleItemOrder(itemToMove: QuestionnaireItem, delta: number): void {
```
codyebberson, line 233: It's not really a "toggle"... Maybe `moveItem`?

In this case, I think your code will end up cleaner if you accept the `itemIndex` as well rather than `itemToMove`.  When passing in the `QuestionnaireItem`, you have to do an `Array.find()` which is O(n) lookup.  You already have the item index on lines 295-296, so might as well use it directly.

packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -665,3 +736,21 @@ function createPage(): QuestionnaireItem {
     ],
   } as QuestionnaireItem;
 }
+
+function reorderItems(
+  items: QuestionnaireItem[] | undefined,
+  itemToMove: QuestionnaireItem,
+  delta: number
+): QuestionnaireItem[] | null {
```
codyebberson, line 744: We almost never return `null` from any of our methods, except in very rare circumstances.  Usually we'd return `undefined` for a "nothing" result.

However, in this case, I think we should always return the `QuestionnaireItem[]` array.  If the delta is invalid, then we can just return the original array.  Fewer edges cases to worry about (note that we're not checking for `null` return values at the call site).

As mentioned above, I think we should change `itemToMove` to `itemIndex` and pass in the numeric array index.  That should avoid the `findIndex` call.

Then you can remove the `index === -1` check.

```ts
const newIndex = itemIndex + delta;
if (newIndex < 0 || newIndex >= items.length) {
  return items;
}
...
```

codyebberson (changes requested):

codyebberson (approved):

PR #2885, opened 2023-09-19
Split tsquery handling by language config

### LINKED ISSUE #2744: String search ignores stop-words
User Driven: https://medplum.slack.com/archives/C05C5KXTPJA/p1693258742269849

## Repro
Create the following `Patient` resource 
```json
{
  "resourceType": "Patient",
  "name": [
    {
      "given": [
        "Justin"
      ],
      "family": "Bieber",
      "use": "usual"
    }
  ]
}
```

Search `Patient?name=Just`

## Root cause
This translates to into the query
```
SELECT "Patient"."id", "Patient"."content" FROM "Patient" LEFT JOIN "HumanName" AS "T1" ON ("Patient"."id"="T1"."resourceId" AND "T1"."name_tsv" @@ to_tsquery('english','Just:*')) WHERE ("Patient"."deleted"=false AND "T1"."resourceId" IS NOT NULL) GROUP BY "Patient"."id" ORDER BY "Patient"."lastUpdated" DESC, MIN("T1"."name") LIMIT 21
```

However:
`to_tsquery('english','Just:*')` filters out [english stopworlds](https://www.postgresql.org/docs/current/textsearch-dictionaries.html#TEXTSEARCH-STOPWORDS). 

This causes postgres to ignore the query and return 0 results: 
`NOTICE:  text-search query contains only stop words or doesn't contain lexemes, ignored`

mattwiller:
This PR is the initial phase of moving most `tsvector` columns from `english` language configuration to `simple`; only columns containing English text (not names or other proper nouns) should use the `'english'` stop word dictionary.  In addition to these code changes, the existing English `tsvector` generated columns will need to be separately migrated.

Broadly, I see two options for how to move forward with migrating the columns:
- Add and backfill new generated `tsvector` columns, with an index on the column
  - e.g. `ALTER TABLE "HumanName" ADD COLUMN name_tsv_simple tsvector GENERATED ALWAYS AS (to_tsvector('simple', name)) STORED` 
- Add new tsvector indices on the `text` columns themselves
  - e.g. `CREATE INDEX CONCURRENTLY IF NOT EXISTS name_tsv_idx_simple ON "HumanName" USING gin (to_tsvector('simple', name))`

Generated columns might be challenging to populate via schema migration on large tables, but will have slightly faster query performance over an index-only approach.  I checked the query plan on a table of 30k `HumanName` records, and saw same query plans and timing data for both index-only and generated column queries.  In light of this, I set up a schema migration to switch away from generated columns to `tsvector` indexes on the `text` columns.  See [Postgres docs](https://www.postgresql.org/docs/14/textsearch-tables.html) for more information

Resolves #2744


packages/server/src/migrations/schema/index.ts
```diff
@@ -49,3 +49,4 @@ export * as v47 from './v47';
 export * as v48 from './v48';
 export * as v49 from './v49';
 export * as v50 from './v50';
+export * as v51 from './v50';
```
codyebberson, line 52: `from './v51';`

codyebberson (changes requested):

codyebberson (approved):

codyebberson:
@mattwiller  - Let's deploy this to staging first to assess db impact of the migration.  Might want to deploy this one outside of peak.

codyebberson:
Migration ran in 10 seconds on staging:

image

Staging currently has 600k patients

PR #2895, opened 2023-09-20
Persisting currentValues for special cases 

jamestouri:
In the QuestionnaireItem values, we used QuestionnaireItemAnswerOption which didn't have boolean as a backbone element. QuestionnaireItemAnswer fixes the retained values of the checkbox

In the Multi-stepper, we are changing the data format to strings. In the onChange, we retrieve the values based on the option matched and getting the whole value in the response

Screenshot 2023-10-02 at 3 15 45 PM

https://github.com/medplum/medplum/assets/6913745/4bf6f0a5-36de-433a-9d2c-91d4e2a6a8c5

https://github.com/medplum/medplum/assets/6913745/b5b3226b-7139-41b2-823d-4f0fdf47ca3a


codyebberson (approved):
Nice, thanks @jamestouri 

Confirmed that this fixes https://github.com/medplum/medplum/issues/2959

PR #2902, opened 2023-09-21
Fixes #2891k - ID capitalization

### LINKED ISSUE #2891: Fix capitalization of "Id" in field selector
image

Field editor: `packages/react/src/SearchFieldEditor/SearchFieldEditor.tsx`

buildFieldNameString: `packages/react/src/SearchControl/SearchUtils.tsx`

It's calling `buildFieldNameString`, which includes a special case for `id`/`ID`, not sure why this isn't working.

capitalpb:
Looks like the buildFieldNameString is checking for id but not Id as well. Added this check into the method.


codyebberson (approved):
lgtm, thanks @capitalpb !

PR #2924, opened 2023-09-25
Fixes #2901 - cli command to update s3 bucket policies

### LINKED ISSUE #2901: OAI not added to S3 bucket policy in cross-region deployments
When deploying medplum outside of us-east-1...
1. We create one CDK stack in your primary region with the majority of AWS resource
2. We create another CDK stack in us-east-1 for "global" resources such as CloudFront distributions
3. We need to update the S3 bucket policy to grant access to CloudFront....

codyebberson:


packages/cli/src/index.ts
```diff
@@ -47,6 +47,7 @@ export async function main(argv: string[]): Promise<void> {
 
     await index.parseAsync(argv);
   } catch (err) {
+    console.error(err);
```
mattwiller, line 50: Should this be removed?

packages/cli/src/aws/update-bucket-policies.ts
```diff
@@ -0,0 +1,158 @@
...
+    await setPolicy(bucketName, bucketPolicy);
+    console.log('Bucket policy updated');
+
+    // Create a CloudFront invalidation to clear any cached responses
+    console.log('Creating CloudFront invalidation...');
+    await createInvalidation(distributionResource.PhysicalResourceId);
+    console.log('CloudFront invalidation created');
+  }
+
+  console.log(`${friendlyName} bucket policy updated`);
```
mattwiller, line 104: This might be confusing for users in dry-run mode, consider moving this log into the `else` case

mattwiller (approved):

PR #2949, opened 2023-09-28
CLI tools to send and receive HL7 messages

codyebberson:
Start an HL7 server on port 56001:

```bash
medplum hl7 listen 56001
```

Send an HL7 message to that server:

```bash
medplum hl7 send localhost 56001 'MSH|...'
```

Don't want to type in an HL7 message?  No problem.

Send an automatically generated message to the server:

```bash
medplum hl7 send localhost 56001 --generate
```

Send a message stored in a file:

```bash
medplum hl7 send localhost 56001 --file my-sample-hl7.txt
```

I briefly looked into supporting piped input, but it looks like we need to handle that at a lower level, so saving that for another day: https://github.com/tj/commander.js/issues/137

### DIFF
--- a/packages/cli/src/hl7.ts
+++ b/packages/cli/src/hl7.ts
@@ -1,0 +1,58 @@
+import { formatHl7DateTime, Hl7Message } from '@medplum/core';
+import { Hl7Client, Hl7Server } from '@medplum/hl7';
+import { Command } from 'commander';
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { createMedplumCommand } from './util/command';
+
+const HL7_NOW = formatHl7DateTime(new Date());
+const HL7_CONTROL_ID = Date.now().toString();
+const SAMPLE_HL7_MESSAGE = `MSH|^~\\&|ADTSYS|HOSPITAL|RECEIVER|DEST|${HL7_NOW}||ADT^A01|${HL7_CONTROL_ID}|P|2.5|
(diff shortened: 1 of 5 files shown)


packages/cli/src/hl7.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Command } from 'commander';
+import { readFileSync } from 'fs';
+import { createMedplumCommand } from './util/command';
+
+const send = createMedplumCommand('send')
+  .description('Send an HL7 v2 message via MLLP')
+  .argument('<host>', 'The destination host name or IP address')
+  .argument('<port>', 'The destination port number')
+  .argument('[body]', 'Optional HL7 message body')
+  .option('--generate', 'Generate a sample HL7 message')
```
mattwiller, line 12: We may want to consider renaming this to e.g. `example` to more clearly express what it does

packages/cli/src/hl7.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    const client = new Hl7Client({
+      host,
+      port: parseInt(port, 10),
+    });
+
+    const response = await client.sendAndWait(Hl7Message.parse(body));
+    console.log(response.toString().replaceAll('\r', '\n'));
+    client.close();
```
mattwiller, line 32: Does this need to be in a `try/finally` or similar to ensure the client is closed?

mattwiller (approved):

PR #2950, opened 2023-09-28
Evaluate X-Fhir-Query helper

mattwiller:


packages/core/src/search/search.ts
```diff
@@ -146,8 +147,13 @@ export function parseSearchRequest<T extends Resource = Resource>(
  * @returns A parsed SearchRequest.
  */
 export function parseSearchUrl<T extends Resource = Resource>(url: URL): SearchRequest<T> {
-  const resourceType = url.pathname.split('/').filter(Boolean).pop() as ResourceType;
```
codyebberson, line 149: Did this logic change?  I think it's the same?

packages/core/src/types.ts
```diff
@@ -467,3 +468,63 @@ export function isReference(value: unknown): value is Reference & { reference: s
...
+      return `${v.value.system}|${v.value.value}`;
+    case PropertyType.Coding:
+      return `${v.value.system}|${v.value.code}`;
+    case PropertyType.CodeableConcept:
+      return (v.value as CodeableConcept).coding?.map((c) => `${c.system}|${c.code}`).join(',') ?? v.value.text;
+    case PropertyType.HumanName:
+      if (v.value.text) {
+        return v.value.text;
+      }
+      return `${v.value.given?.join(' ')} ${v.value.family}`;
```
codyebberson, line 501: Consider using `formatHumanName` in format.ts: https://github.com/medplum/medplum/blob/main/packages/core/src/format.ts#L70

packages/core/src/types.ts
```diff
@@ -467,3 +468,63 @@ export function isReference(value: unknown): value is Reference & { reference: s
...
+      return v.value.value;
+    case PropertyType.Quantity:
+    case PropertyType.Age:
+    case PropertyType.Count:
+    case PropertyType.Duration:
+      return `${v.value.value}|${v.value.system}|${v.value.code || v.value.unit}`;
+    case PropertyType.date:
+    case PropertyType.dateTime:
+    case PropertyType.instant:
+      return (v.value as Date).toISOString();
```
codyebberson, line 521: Hm, does `evalFhirPathTyped` ever return `Date`?  I believe it will already return a well formatted string.

packages/core/src/search/search.ts
```diff
@@ -407,6 +411,28 @@ function addFilter(searchRequest: SearchRequest, filter: Filter): void {
...
+ * @param variables Values to pass into embedded FHIRPath expressions
+ * @returns The parsed search request
+ */
+export function parseXFhirQuery(query: string, variables: Record<string, TypedValue>): SearchRequest {
+  query = query.replaceAll(subexpressionPattern, (_, expr) => {
+    const replacement = evalFhirPathTyped(expr, [], variables);
+    if (replacement.length !== 1) {
+      return '';
+    }
+    return formatTypedValue(replacement[0]);
```
codyebberson, line 431: I'm not quite sure about the context here, but do we need to URL-escape the return value?  i.e. `encodeURIComponent`?

packages/core/src/search/search.ts
mattwiller: It should be the same: I was just trying to eliminate the throwaway array from `.filter().pop()`

packages/core/src/search/search.ts
mattwiller: My expectation is that the context is defined by the caller: they can encode if adding directly to a URL

packages/core/src/constants.ts
```diff
@@ -0,0 +1,7 @@
+// Common terminology systems, taken from https://terminology.hl7.org/external_terminologies.html
+export const UCUM = 'http://unitsofmeasure.org';
```
codyebberson, line 2: Nice

codyebberson (approved):

PR #2956, opened 2023-09-29
Added rules back to sweep.yaml

kevinlu1248:
Hey @codyebberson! Sorry about all the bad issues generated previously, we made several improvements since then:

* Sweep makes the PR directly and does not open up an issue to do this, decluttering the issues.
* Sweep is now more strict and careful about making these PRs.
* I clarified the formatting further in the rules to not change the extension for tests (.ts -> .test.ts).
* Several other reliability improvements with file creations and writing higher quality unit tests.


codyebberson:
Hi @kevinlu1248  - Thanks for submitting.  This feature was pretty noisy last time.  I think we're going to let the feature mature a little bit more before we try it again.

PR #2964, opened 2023-10-03
Initial FHIRCast setup

codyebberson:

PR #2965, opened 2023-10-03
Add `FHIRcast` client example

### LINKED ISSUE #2800: Make `FHIRcast` demo
`FHIRcast` is the hottest new FHIR-centric protocol on the block.

`FHIRcast` seems to be promising for use-cases such as having a [`PACS`](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1718393/) system driving a Medplum session. If the `PACS` tells the `Hub` (the messenger of the system) that it has changed to another patient's X-ray or similar diagnostic image, then `Medplum` will be notified that the context has changed and `Medplum` can handle the change appropriately (maybe opening up the notes or patient timeline for the patient attached to the X-ray.)

We want to prove out a `FHIRcast` hub implementation and see how it looks in practice, and estimate how long this sort of an integration might take. 

I'll be tracking the progress of that journey in this issue.

ThatOneBro:
## What this PR does
Adds a new example to demo Medplum's `FHIRcast` implementation

Depends on #2964 

Closes #2800


examples/medplum-fhircast-demo/src/components/Subscriber.tsx
```diff
@@ -0,0 +1,179 @@
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
+import { BASE_URL } from '../config';
+import { useClientId, useHubWSConnection } from '../hooks';
+import { FHIRcastMessagePayload, isWebSocketMessage, serializeHubSubscriptionRequest } from '../utils';
+import TopicLoader from './TopicLoader';
+
+type FhirCastMessageDisplayProps = {
+  eventNo: number;
+  message: object;
```
codyebberson, line 9: Is there a better type here than `object`?

examples/medplum-fhircast-demo/src/components/Subscriber.tsx
```diff
@@ -0,0 +1,179 @@
...
+          paddingLeft: 30,
+          paddingRight: 30,
+          paddingTop: 20,
+          paddingBottom: 20,
+          borderRadius: 10,
+          textAlign: 'left',
+        }}
+      >
+        <h3>Event No. {props.eventNo}</h3>
+        <pre>{JSON.stringify(props.message, null, 2) ?? ''}</pre>
```
codyebberson, line 27: Will `props.message` ever be `undefined`?  If not, I think we can remove the `?? ''`

examples/medplum-fhircast-demo/src/components/Subscriber.tsx
```diff
@@ -0,0 +1,179 @@
...
+  setFhirCastMessages: Dispatch<SetStateAction<FHIRcastMessagePayload[]>>;
+  setWebSocketStatus: (status: string) => void;
+  incrementEventCount: () => void;
+};
+
+function WebSocketHandler(props: WebSocketHandlerProps): null {
+  const { endpoint, setCurrentPatientId, setFhirCastMessages, setWebSocketStatus, incrementEventCount } = props;
+  const { websocket } = useHubWSConnection(endpoint, {
+    onConnect: () => {
+      console.log('here!');
```
codyebberson, line 46: Let's remove all debugging `console.log` statements

examples/medplum-fhircast-demo/src/components/Subscriber.tsx
```diff
@@ -0,0 +1,179 @@
...
+      setWebSocketStatus('CONNECTED');
+
+      // expect pong
+      websocket.on('pong', (message: Record<string, any>) => {
```
codyebberson, line 50: Are the `ping` and `pong` messages part of the spec?  If not, let's remove.

examples/medplum-fhircast-demo/src/hooks/index.ts
```diff
@@ -0,0 +1,100 @@
...
+/* eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents */
+type WrappedClientWebSocket = WrappedWebSocket<WebSocket> & {
+  on: (messageType: string, handler: (message: Record<string, any>) => void) => void;
+};
+
+type HubHookConnection = {
+  websocket: WrappedClientWebSocket;
+};
+
+const DEBUG = true;
```
codyebberson, line 18: Let's remove all debugging helpers.

examples/medplum-fhircast-demo/src/index.css
```diff
@@ -0,0 +1,77 @@
+:root {
```
codyebberson, line 1: Probably out of scope for this PR, but for the future:  When we add auth to everything, we'll probably want to use the `@medplum/react` React components, especially for `<SignInForm>` and some of the auth helper components.  Once we break the seal and start using Mantine, we might as well use it for all base styling.

We can discuss that offline.

examples/medplum-fhircast-demo/src/utils.ts
```diff
@@ -0,0 +1,166 @@
+type EventName = string;
+
+export type WorkflowEvent = {
```
codyebberson, line 3: I don't think this type is used?

examples/medplum-fhircast-demo/src/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+  name: EventName;
+  topic: string;
+  payload: Record<string, any>;
+};
+
+export type PatientOpenEvent = {
```
codyebberson, line 9: Not used?

examples/medplum-fhircast-demo/src/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+};
+
+export type SubscriptionConfirmationSerialized = {
+  'hub.mode': 'subscribe';
+  'hub.events': string;
+  'hub.topic': string;
+  'hub.lease_seconds': number;
+};
+
+export type FHIRResource = {
```
codyebberson, line 43: This can be replaced with `Resource` from `@medplum/fhirtypes`.

examples/medplum-fhircast-demo/src/hooks/index.ts
```diff
@@ -0,0 +1,100 @@
...
+import { FHIRcastMessagePayload, WebSocketMessage, WrappedWebSocket, isWebSocketMessage } from '../utils';
+import { getOrCreateWebSocketClient } from '../websocket-client';
+
+type HubWSConnectionOptions = {
+  onConnect: () => void;
+};
+
+/* eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents */
+type WrappedClientWebSocket = WrappedWebSocket<WebSocket> & {
+  on: (messageType: string, handler: (message: Record<string, any>) => void) => void;
```
codyebberson, line 11: Let's discuss offline whether we want to use websocket wrappers.  I'd like the understand the benefits of the abstraction vs the complexity.  In particular, I want to make sure we're thinking about this from the perspective of someone coming into this cold with zero context, and how they will process it.

examples/medplum-fhircast-demo/src/components/PACS.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const [driving, setDriving] = useState(false);
+  const [topic, setTopic] = useState<string | null>(null);
+  const [currentPatientId, setCurrentPatientId] = useState<string | null>(null);
+  const prevTopic = usePrevious(topic);
+
+  useEffect(() => {
+    // if had prev topic, and now we don't, setDriving false
+    // if didn't have and now we do, setDriving true
+    if (typeof prevTopic !== typeof topic) {
+      setDriving(!!topic);
```
codyebberson, line 55: Is there a reason we do this here in another `useEffect` call rather than directly when calling `setTopic`?

examples/medplum-fhircast-demo/src/components/PACS.tsx
```diff
@@ -0,0 +1,117 @@
...
+      ],
+    },
+  };
+}
+
+export default function PACS(): JSX.Element {
+  const [baseUrl, setBaseUrl] = useState(BASE_URL);
+  const [baseUrlInput, setBaseUrlInput] = useState(BASE_URL);
+  const [driving, setDriving] = useState(false);
+  const [topic, setTopic] = useState<string | null>(null);
```
codyebberson, line 47: By convention, we try to avoid `null` and use `undefined` unless there are very specific reasons (i.e., external library depends on it).

examples/medplum-fhircast-demo/src/hooks/index.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (fhirHandler) {
+          // @ts-expect-error This should be fine but union type of the handlers is probably not what we want for this (should be keyed on event type)
+          fhirHandler(message as FHIRcastMessagePayload);
+        }
+      } else if (wsHandlersRef.current[message.type]) {
+        // @ts-expect-error this is fine because FHIRcast is just an exception...
+        wsHandlersRef.current[message.type](message);
+      } else {
+        // eslint-disable-next-line no-unused-expressions
+        DEBUG && debug(`Failed to handle event of type: ${message.type}`);
```
codyebberson, line 62: Let's remove

examples/medplum-fhircast-demo/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "jsx": "react-jsx",
+
+    /* Linting */
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true
+  },
+  "include": ["src"],
+  "references": [{ "path": "./tsconfig.node.json" }]
```
codyebberson, line 24: I think we can merge `tsconfig.json` and `tsconfig.node.json` into a single file, yeah?  I don't know why Vite does this by default.

examples/medplum-fhircast-demo/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "devDependencies": {
+    "@types/react": "^18.2.15",
+    "@types/react-dom": "^18.2.7",
+    "@typescript-eslint/eslint-plugin": "^6.0.0",
+    "@typescript-eslint/parser": "^6.0.0",
+    "@vitejs/plugin-react": "4.0.4",
+    "eslint": "^8.45.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint-plugin-react-refresh": "^0.4.3",
```
codyebberson, line 24: TIL.  Interesting.  We should consider adding this to `packages/eslint-config`

codyebberson (changes requested):
Nice, thanks for putting this together @ThatOneBro!

There are a few bigger topics that we should discuss offline.  Some part of this PR, some not.

1. Using `MedplumClient` for config settings like "baseUrl" and auth
2. Using `@medplum/react` for auth components like `<SignInForm>`
3. Using `@medplum/fhirtypes` for `Resource` and `Patient` type definitions
4. Whether we should export FHIRcast type definitions from `@medplum/fhirtypes`
5. Some of the abstractions and helper utilities - There's a delicate balance when writing code specifically for examples and demos, where we are writing much more for the reader than ourselves.  A couple of the abstractions I want to really dig into, such as `WrappedWebSocket` and some of the custom hooks.  They can be convenient for us when writing code, but they can introduce a lot of cognitive overhead for the reader when they're in learning-mode.

We're going to need to figure out how to capitalize "FHIRcast" in class and function names 🙃

examples/medplum-fhircast-demo/tsconfig.json
ThatOneBro: I guess it's just for `vite.config.ts` since that runs in Node. Probably doesn't matter if your Vite config is small though since you're probably not doing a whole lot. 

I guess DOM vs Node env.

examples/medplum-fhircast-demo/src/components/PACS.tsx
ThatOneBro: I think it's generally the React way to let state driving changes via effects when there is a condition involved. In this case we only want to update `driving` if the current `topic` type has changed from `string` to `undefined` or vice versa when compared with the `prevTopic`. 

If it seems unclear we can change it but generally the "React way" is to let the state drive the changes

examples/medplum-fhircast-demo/src/hooks/index.ts
ThatOneBro: Removed the need for them in the latest changes. Should simplify things dramatically.

codyebberson (approved):
👍  Let's go!

Let's chat offline about next steps:
1. Adding auth
2. Adding `@medplum/react` for `<SignInForm>`
3. May want to use our `<AppShell>` component, but that's less obvious to me
4. Moving some of the `fetch` requests into `MedplumClient`

PR #2975, opened 2023-10-04
QuestionnaireForm style tweaks

codyebberson:
Fixed duplicate text labels on "boolean" questions:

image

Added vertical space between radio options:

image

Fixed spacing around Attachment input:

image

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -203,8 +203,21 @@ function QuestionnaireFormArrayContent(props: QuestionnaireFormArrayContentProps
       />
     );
   }
+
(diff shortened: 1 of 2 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -203,8 +203,21 @@ function QuestionnaireFormArrayContent(props: QuestionnaireFormArrayContentProps
       />
     );
   }
+
+  if (props.item.type === QuestionnaireItemType.boolean) {
+    return (
+      <QuestionnaireRepeatWrapper
```
codyebberson, line 209: Same as below, but without the `<FormSection>` wrapper.

reshmakh (approved):

PR #3015, opened 2023-10-10
New migration generator proof-of-concept

codyebberson:
Prototype for internal review

### What is this?

A new database migration generator that works in 3 steps:

1. Build an in-memory representation of the current database ("start")
2. Build an in-memory representation of the desired database from FHIR spec ("target")
3. Table by table, column by column, generate the SQL statements to move from "start" to "target"

The previous database migration generator was much less sophisticated.  For every migration, it required some manual tuning to output the desired migration.

This new migration generator is intended to be fully automated.

### Why do this?

Short term:

1. The new FHIRcast implementation would benefit from the FHIR R4B type `SubscriptionTopic`
2. I prototyped swapping in R4B (see https://github.com/medplum/medplum/tree/cody-r4b), which looks great
3. But to fully deploy R4B will require some heavy lifting from the migration generator

Longer term:

1. This could code move from `packages/generator` to `packges/server`
2. The server could run this automatically on first run to initialize a database
3. The server could run this automatically on new resource types / new search parameters
4. This would open up custom search parameters / custom indexes


packages/generator/src/migrate.ts
```diff
@@ -258,63 +325,169 @@ function getColumnType(details: SearchParameterDetails): string {
...
+}
+
+function migrateColumns(b: FileBuilder, startTable: TableDefinition, targetTable: TableDefinition): void {
+  for (const targetColumn of targetTable.columns) {
+    const startColumn = startTable.columns.find((c) => c.name === targetColumn.name);
+    if (!startColumn) {
+      writeAddColumn(b, targetTable, targetColumn);
+    } else if (
+      startColumn.type.replaceAll('TIMESTAMP WITH TIME ZONE', 'TIMESTAMPTZ').trim() !==
+      targetColumn.type.replaceAll(' PRIMARY KEY', '').replaceAll(' DEFAULT FALSE', '').trim()
```
mattwiller, line 434: What's going on here?

packages/generator/src/migrate.ts
```diff
@@ -258,63 +325,169 @@ function getColumnType(details: SearchParameterDetails): string {
...
+  result += '_idx ON "';
+  result += tableName;
+  result += '"';
+
+  if (index.gin) {
+    result += ' USING gin ';
+  }
+
+  result += '(';
+  result += index.columns.map((c) => '"' + c + '"').join(', ');
```
mattwiller, line 486:
```suggestion
  result += index.columns.map((c) => `"${c}"`).join(', ');
```

packages/generator/src/migrate.ts
```diff
@@ -258,63 +325,169 @@ function getColumnType(details: SearchParameterDetails): string {
...
+  result += index.columns.join('_');
+  result += '_idx ON "';
+  result += tableName;
+  result += '"';
+
+  if (index.gin) {
+    result += ' USING gin ';
+  }
+
+  result += '(';
```
mattwiller, line 485:
```suggestion
  result += '" ';

  if (index.gin) {
    result += 'USING gin ';
  }

  result += '(';
```

mattwiller (approved):
Love the approach: it's great to have this package getting into a robust state

packages/generator/src/migrate.ts
```diff
@@ -12,8 +12,30 @@ import { readJson } from '@medplum/definitions';
...
+}
+
+interface ColumnDefinition {
+  name: string;
+  type: string;
+}
+
+interface IndexDefinition {
+  columns: string[];
+  gin?: boolean;
```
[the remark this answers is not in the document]
codyebberson: Nice, like it.

packages/generator/src/migrate.ts
codyebberson: Agree

packages/generator/src/migrate.ts
codyebberson: Yeah...  This is a mismatch between "column definition" and "index definition"

I agree with your meta point, which is "this is confusing".... I'll move this logic to a helper function called "normalizeColumnDefinition" or something.

PR #3016, opened 2023-10-10
Remove barrel files

mattwiller:
Flattening the TypeScript module graph by removing intermediary files that serve only to re-export things is expected to improve performance


codyebberson (approved):
Love it 💪 ❤️

PR #3026, opened 2023-10-12
Test against multiple Postgres versions

mattwiller:


ThatOneBro (approved):

codyebberson (changes requested):
Later in `build.yml`, there are a bunch of checks like this:

```
if: github.repository == 'medplum/medplum' && github.ref == 'refs/heads/main' && matrix.node-version == '18.x'
```

The reason is that matrix strategies will end up duplicating the number of times we go through the deploy steps.

We need to do something similar for `matrix.pg-version == 'x'`

Rather than duplicating the checks over and over, it might make sense to establish a temp variable.  I don't know enough about Github Actions YAML to know if that's easy or hard.

codyebberson:
Don't worry about this:

image

Once the build name strings stabilize, we can update the repo settings for "expected checks".

Issue #2448, opened 2023-07-12
Reorg Auth Docs


rahul1:
- [x] Split out Authorization and Access Policies into 2 top level categories 
- [x] Consolidate all the authentication methods into a single folder
- [x] Implement vercel server-side redirects to avoid broken links
- [x] Update "Auth Integration" docs with a decision guide on how to pick the right auth method
- [x] Update title "Okta Integration" to indicate that it's really "domain level integration"

reshmakh:
@rahul1 assuming is also closed by https://github.com/medplum/medplum/pull/3074

PR #3029, opened 2023-10-13
Reorganize auth docs

rahul1:
Part of #2448 

This PR does 3 primary things:
1. Splits the "Auth and Identity" section into two sections
  a. Authentication and Identity
  b. Authorization and Access Control
2. Moves guides for all of our different authentication methods (Auth code, Client Credentials, Google Auth, etc.) into a subfolder of "Authentication and Identity"
3. Moves the "Auth Integration Guide" to be the index page for "Authentication Methods"

While the diff is large, this is primarily fixing broken links.


packages/docs/docs/auth/authentication-methods/_category_.json
```diff
@@ -0,0 +1,4 @@
+{
+  "label": "Authentication Methods",
```
mattwiller, line 2: The `/auth/authentication-methods/...` URLs look a bit chunky to me, would we want to slim down to e.g. `methods` or `integrations`?

rahul1: Good point re: URL. 

Will change to `/methods`

rahul1: Will resolve titles in a future PR, since there's a bit of article renaming that needs to be done

mattwiller (approved):

packages/docs/docs/access/access-policies.md
```diff
@@ -1,9 +1,9 @@
 ---
-sidebar_position: 4
+sidebar_position: 0
 tags: [auth]
 ---
 
-# Access Controls
```
reshmakh, line 6: One thing here that I'd like to preserve.  Access Controls is a widely known industry standard term that people understand... while "access policies" unfortunately is not.  I'd like to maintain Access Controls as the top level name for clarity while heavily referring to "Access Policies" as our implementation.

packages/docs/docs/auth/methods/index.md
```diff
@@ -1,48 +1,48 @@
 ---
-sidebar_position: 16
+sidebar_position: 1
 tags: [auth]
 ---
 
-# Auth Integration Patterns
+# Authentication Methods
```
reshmakh, line 6: Like this cleanup

reshmakh: Unfortunately, I do think something like "Patterns" is better here.  The reason is that methods can man other things as well such as "JWT", "Basic Auth" and the like.  This is more of what type of application is connecting.  Open to suggestions on the language but Methods I think is not quite right.

packages/docs/docs/access/access-policies.md
rahul1: I'm going to push back on this. I maintained the term "Access Control" at the top level to maintain the industry standard term. Below, I tried to list the different _tools_ we have for to implement access control, of which `AccessPolicies` are one. 


image

Additionally, a lot of our devs use the search term "Access Policy" on our docs

image

packages/docs/docs/auth/methods/index.md
rahul1: Makes sense. I can maintain patterns

PR #3033, opened 2023-10-13
Fix primitive extensions on choice types

### LINKED ISSUE #3018: Underscore primitives don't work with [x] choice of type
https://discord.com/channels/905144809105260605/1161125951703359539/1161125951703359539

Does not work, but should work:

```json
  "birthDate": "1995-09-27",
  "_birthDate": {
    "id": "birthdate1"
  },
  "deceasedBoolean": false,
  "_deceasedBoolean": {
    "id": "deceased1"
  },
```

Works, but should not work:

```json
  "birthDate": "1995-09-27",
  "_birthDate": {
    "id": "birthdate1"
  },
  "deceasedBoolean": false,
  "_deceased[x]": {
    "id": "deceased1"
  },
```

mattwiller:
Resolves #3018

### DIFF
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -129,8 +129,16 @@ describe('FHIR resource validation', () => {
       resourceType: 'Patient',
       deceasedBoolean: false,
     };
+    const choiceTypeExtension: Patient = {
+      resourceType: 'Patient',
+      deceasedBoolean: false,
+      _deceasedBoolean: {
+        id: '1',
+      },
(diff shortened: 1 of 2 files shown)


codyebberson (commented):
Nice.  Surgical.

Can we remove line 263 now?

```
!(key.startsWith('_') && key.slice(1) in properties) &&
```

I think your update to `isChoiceOfType` fixes that case.

And I think line 263 is what allowed the `"_deceased[x]"` property through.

We should perhaps add a test case to make sure `"_deceased[x]"` is invalid.

Otherwise lgtm 👍

PR #3049, opened 2023-10-16
fix(websocket): handle invalid WS conns gracefully

### LINKED ISSUE #3048: Handle connections to invalid `WebSocket` routes gracefully
Currently we are just destroying the socket (terminating the connection) whenever a route isn't matched against our `WebSocket` routes in the route map. This results in `5xx` errors at the ALB since the ALB gets stuck waiting for a response and eventually times out, leaving the client hanging until timeout and eventual `5xx`.

The fix is to send back `HTTP/1.1 404 Not Found\r\n\r\n` and give a proper HTTP response upon being unable to upgrade. See MDN for more about `Upgrade` protocol: 

> If the server decides to upgrade the connection, it sends back a [101 Switching Protocols](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/101) response status with an Upgrade header that specifies the protocol(s) being switched to. If it does not (or cannot) upgrade the connection, it ignores the Upgrade header and sends back a regular response (for example, a [200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)).

-- [MDN: Protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism)

ThatOneBro:
Fixes #3048

### DIFF
--- a/packages/server/src/websockets.test.ts
+++ b/packages/server/src/websockets.test.ts
@@ -2,7 +2,9 @@ import { allOk, ContentType, Hl7Message } from '@medplum/core';
 import { Agent, Bot } from '@medplum/fhirtypes';
 import express from 'express';
 import { Server } from 'http';
+import { AddressInfo } from 'net';
 import request from 'superwstest';
+import WebSocket from 'ws';
 import { initApp, shutdownApp } from './app';
 import { getConfig, loadTestConfig, MedplumServerConfig } from './config';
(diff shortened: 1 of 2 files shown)


codyebberson (approved):
Nice

I was surprised that you had to drop down to `socket.write()`, but that appears to be the recommended solution. 

👍 ship it

PR #3066, opened 2023-10-17
fix(client/auth): respect `accessToken` in constructor

### LINKED ISSUE #3065: `MedplumClient` constructor ignores `accessToken`
## Repro

- Log into Medplum app in browser
- Copy the medplum access token from local storage
- Create a Medplum client as follows
```
      fetch: async (url: string, options?: any) => await fetch(url, options),
      token: medplumToken 
    });
```
- Call `medplum.graphql(SOME_QUERY)`

## Expected Behavior
Query completes successfully

## Actual Behavior
Unauthenticated error

## Workaround
Call `medplum.setAccessToken`, rather than using constructor parameter

ThatOneBro:
Fixes #3065

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -666,6 +666,10 @@ export class MedplumClient extends EventTarget {
       this.refreshProfile().catch(console.log);
     }
 
+    if (options?.accessToken) {
+      this.accessToken = options.accessToken;
+    }
+
     this.setupStorageListener();
   }
 
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -667,7 +667,7 @@ export class MedplumClient extends EventTarget {
     }
 
     if (options?.accessToken) {
-      this.accessToken = options.accessToken;
+      this.setAccessToken(options.accessToken);
     }
 
     this.setupStorageListener();
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -660,14 +660,14 @@ export class MedplumClient extends EventTarget {
       this.autoBatchQueue = undefined;
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/client.ts
```diff
@@ -666,6 +666,10 @@ export class MedplumClient extends EventTarget {
       this.refreshProfile().catch(console.log);
     }
 
+    if (options?.accessToken) {
+      this.accessToken = options.accessToken;
```
mattwiller, line 670: We probably want to use `this.setAccessToken(options.accessToken)` here to ensure all the client state is set correctly

ThatOneBro: Good catch

packages/core/src/client.ts
```diff
@@ -666,6 +666,10 @@ export class MedplumClient extends EventTarget {
       this.refreshProfile().catch(console.log);
     }
 
+    if (options?.accessToken) {
+      this.setAccessToken(options.accessToken);
+    }
```
codyebberson, line 671: Yep, this is the meat of it.

I think we should move lines 663-667 into an `else` block though, to avoid confusion.  If user passes in an `accessToken`, we're going to ignore existing login.

In practice, this is unlikely to matter too much (the use cases are pretty different), but I'd prefer to be explicit.

codyebberson (commented):
Nice.  See comment on `getActiveLogin` code path.  Feel free to disagree.

ThatOneBro:
> Nice. See comment on `getActiveLogin` code path. Feel free to disagree.

Nope, that makes sense to me. Shouldn't bother calling it if we provide an `accessToken`, totally agree.

codyebberson (approved):

PR #3068, opened 2023-10-17
Add documentation about resource profiles

mattwiller:
Adding some documentation about how to use FHIR profiles to control resource validation.  This doesn't yet include anything about authoring profiles: it's complex and we should probably provide tooling for users so they can avoid hand-writing the JSON


packages/docs/docs/fhir-datastore/profiles.md
```diff
@@ -0,0 +1,67 @@
...
+Using a pre-existing profile is simple: placing the canonical URL of the profile in a resource's `meta.profile` field
+will cause the server to attempt to validate the resource against that profile when the resource is written.
+
+```json
+{
+  "resourceType": "Patient",
+  "meta": {
+    "profile": ["http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient"]
+  },
+  // Normally Patient has no required fields, but under the US Core Patient profile,
```
ThatOneBro, line 43: Technically comments in JSON make it invalid, which is why it's highlighted in red on Github. Maybe `jsonc` is a valid formatting here though?

ThatOneBro (approved):
Looks good to me, I learned a lot about profiles just from reading this!

packages/docs/docs/fhir-datastore/profiles.md
codyebberson: Elsewhere we've used just `js` which should do the trick.

packages/docs/docs/fhir-datastore/profiles.md
```diff
@@ -0,0 +1,67 @@
...
+
+[profiling]: http://hl7.org/fhir/profiling.html
+[us-core-bp]: http://hl7.org/fhir/us/core/StructureDefinition/us-core-blood-pressure
+[us-core]: https://www.hl7.org/fhir/us/core/#us-core-profiles
+[uscdi]: https://www.healthit.gov/isa/united-states-core-data-interoperability-uscdi
+
+## Profile adoption
+
+Using a pre-existing profile is simple: placing the canonical URL of the profile in a resource's `meta.profile` field
+will cause the server to attempt to validate the resource against that profile when the resource is written.
```
rahul1, line 35: Great example - I just want to break it down a little for a novice audience. 

- Can you describe which fields are required by the US Core patient resource?
- Can you add an example of the error message that users can expect when creating a `Patient` that _doesn't_ conform to the the profile (having the error message indexed by Algolia will help our users when troubleshooting their apps)
- Lastly, have this example below, to show what the _correct_  data would be

rahul1 (changes requested):

rahul1 (approved):

PR #3070, opened 2023-10-17
fix(build): put placeholder env vars in build step

ThatOneBro:
## Background
Currently we are seeing an issue with version 2.1.5 missing placeholder strings in the published NPM package (`@medplum/app`). The env vars like `MEDPLUM_BASE_URL` which are included in the build step as environment variables are missing from the published build. 

It seems like when we split our workflows in #3026 for `deploy` and `build`, we ended up creating an issue where a cached build was being built without the proper placeholder env vars set and it resulted in the published build not having the placeholder strings present in the build.

## What this PR does
This PR should fix the problem with the build missing placeholder strings like `MEDPLUM_BASE_URL`, etc.


codyebberson:
:'(

image

Let's discuss splitting `build.sh` into `build.sh` and `test.sh`.  The env vars should only be applied during build.

.github/workflows/build.yml
```diff
@@ -59,6 +59,16 @@ jobs:
             ${{ runner.os }}-
       - name: Build Project
         run: ./scripts/build.sh
+        env:
+          POSTGRES_HOST: localhost
```
codyebberson, line 63: `POSTGRES_HOST` and `POSTGRES_PORT` not required for build

codyebberson (approved):
One minor cleanup thing, otherwise lgtm 👍

PR #3080, opened 2023-10-19
feat(cdk): add initial support for external secrets

### LINKED ISSUE #2788: Allow external secrets in CDK config file
For https://www.medplum.com/docs/self-hosting/datadog

Users don't want to hardcode `DD_API_KEY` in the JSON file.

ThatOneBro:
## What this PR does
* Adds support for external secrets in `MedplumInfraConfig` during CDK deployments
* Closes #2788

## TODO:
- [x] Add tests
- [x] Refactor logic to make it less hard to read
- [x] Consider implications of serially fetching secrets vs. parallelization
- [ ] Update docs for CDK config?

### DIFF
--- a/packages/cdk/src/config.ts
+++ b/packages/cdk/src/config.ts
@@ -57,7 +57,7 @@ export function normalizeFetchedValue(
   if (!VALID_PRIMITIVE_TYPES.includes(typeOfVal)) {
     throw new OperationOutcomeError(
       validationError(
-        `Invalid value found for type, expected 'string' or 'boolean' or 'number', received: '${typeOfVal}'`
+        `Invalid value found for type; expected either ${VALID_PRIMITIVE_TYPES.join(', or')} but got ${typeOfVal}`
       )
     );
   }
@@ -67,7 +67,7 @@ export function normalizeFetchedValue(
     const normalized = (rawValue as string).toLowerCase() as 'true' | 'false';
     if (normalized !== 'true' && normalized !== 'false') {
       throw new OperationOutcomeError(
-        validationError(`Invalid value found for key '${key}', expected boolean value but got '${rawValue}'`)
+        validationError(`Invalid value found for key '${key}'; expected boolean value but got '${rawValue}'`)
       );
     }
     return normalized === 'true';
@@ -75,13 +75,13 @@ export function normalizeFetchedValue(
     const parsed = parseInt(rawValue as string, 10);
     if (Number.isNaN(parsed)) {
       throw new OperationOutcomeError(
-        validationError(`Invalid value found for key '${key}', expected integer value but got '${rawValue}'`)
+        validationError(`Invalid value found for key '${key}'; expected integer value but got '${rawValue}'`)
       );
     }
     return parsed;
   } else {
     throw new OperationOutcomeError(
-      validationError(`Invalid value found for type, expected '${expectedType}', received: '${typeOfVal}'`)
+      validationError(`Invalid value found for type; expected ${expectedType} value but got value of type ${typeOfVal}`)
     );
   }
 }
@@ -163,10 +163,8 @@ async function normalizeValueForKey(obj: Record<string, any>, key: string): Prom
 export async function normalizeObjectInInfraConfig(obj: Record<string, any>): Promise<Record<string, any>> {
   const normalizedObj = { ...obj };
   // walk config object
-  for (const key in normalizedObj) {
-    if (Object.hasOwn(normalizedObj, key)) {
-      await normalizeValueForKey(normalizedObj, key);
-    }
+  for (const key of Object.keys(normalizedObj)) {
+    await normalizeValueForKey(normalizedObj, key);
   }
   return normalizedObj;
 }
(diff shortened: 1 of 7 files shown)


packages/core/src/config.ts
```diff
@@ -1,3 +1,75 @@
...
+export type ExternalSecretPrimitiveType = 'string' | 'boolean' | 'number';
+
+export type ExternalSecret<T extends ExternalSecretPrimitiveType = ExternalSecretPrimitiveType> = {
+  system: ExternalSecretSystem;
+  key: string;
+  type: T;
+};
+
+export interface MedplumSourceInfraConfig {
+  name: string | ExternalSecret<'string'>;
```
codyebberson, line 16: Can we create a type for this common union?

```ts
type ValueOrExternalSecret<T> = T | ExternalSecret<T>;
```

packages/core/src/config.ts
```diff
@@ -1,3 +1,75 @@
...
+  stackName: string | ExternalSecret<'string'>;
+  accountNumber: string | ExternalSecret<'string'>;
+  region: string | ExternalSecret<'string'>;
```
codyebberson, line 19: I think `region` needs to be a string, so we know where to load the parameters from.

packages/cdk/src/config.ts
```diff
@@ -0,0 +1,214 @@
...
+  OperationOutcomeError,
+  badRequest,
+  validationError,
+} from '@medplum/core';
+
+const DEFAULT_AWS_REGION = 'us-east-1';
+const VALID_PRIMITIVE_TYPES = ['string', 'boolean', 'number'];
+
+export async function fetchParameterStoreSecret(path: string, key: string): Promise<string> {
+  const region = DEFAULT_AWS_REGION;
```
codyebberson, line 17: Region needs to be loaded from the config file for customers who don't use `us-east-1`.

packages/cdk/src/config.ts
```diff
@@ -0,0 +1,214 @@
...
+  const client = new SSMClient({ region });
+
+  let nextToken: string | undefined;
+  do {
+    const response = await client.send(
```
codyebberson, line 22: This will be O(n^2) lookup.

We can either get *all* the parameters for a given path prefix, and then match them up.

Or we can get parameters directly using `GetParametersCommand`:  https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-ssm/classes/getparameterscommand.html

packages/cdk/src/config.ts
```diff
@@ -0,0 +1,214 @@
...
+  return (
+    typeof obj === 'object' &&
+    typeof obj.system === 'string' &&
+    typeof obj.key === 'string' &&
+    typeof obj.type === 'string'
+  );
+}
+
+export function assertValidExternalSecret(obj: Record<string, any>): asserts obj is ExternalSecret {
+  if (
```
codyebberson, line 186: This could be `if (!isExternalSecret(obj))`, no?

packages/cdk/src/config.ts
```diff
@@ -0,0 +1,214 @@
...
+  const typeOfVal = typeof rawValue;
+  // Return raw type if type is string and value is of type string, or if type isn't string and typeof val isn't string
+  if (!VALID_PRIMITIVE_TYPES.includes(typeOfVal)) {
+    throw new OperationOutcomeError(
+      validationError(
+        `Invalid value found for type, expected 'string' or 'boolean' or 'number', received: '${typeOfVal}'`
+      )
+    );
+  }
+  if (expectedType === 'string' && typeOfVal === 'string') {
```
codyebberson, line 64: I think we can reduce the number of error checks if we boil this down to 3 possibilities:
1. `typeOfVal === expectedType`, so return as-is
2. `typeOfVal === 'string'`, so convert it
3. Otherwise it's an error

packages/cdk/src/config.ts
```diff
@@ -0,0 +1,214 @@
...
+  else if (typeof currentVal === 'object') {
+    obj[key] = await normalizeObjectInInfraConfig(currentVal);
+  }
+}
+
+export async function normalizeObjectInInfraConfig(obj: Record<string, any>): Promise<Record<string, any>> {
+  const normalizedObj = { ...obj };
+  // walk config object
+  for (const key in normalizedObj) {
+    if (Object.hasOwn(normalizedObj, key)) {
```
codyebberson, line 169: You can skip the `Object.hasOwn()` call if you iterate over `Object.keys` such as

```ts
for (const key of Object.keys(normalizedObj)) {
```

codyebberson:
@ThatOneBro  Nice, thanks for jumping on this!

packages/core/src/config.ts
ThatOneBro: Good idea

ThatOneBro:
@codyebberson Awesome, so I moved over to the `GetParameterCommand`, and made `region` only a string in `MedplumSourceInfraConfig`. Let me know if this looks good.

I considered batching all the requests after walking the config and collecting which keys need to be fetched, but I think that's probably an over-optimization and dumped that idea for the current setup.

codyebberson (approved):

PR #3087, opened 2023-10-19
Server update CLI

### LINKED ISSUE #2785: Add `medplum aws update` CLI command
See full discussion here: https://github.com/medplum/medplum/discussions/2778

- [ ] Add new CLI command for `medplum aws update`
    - [ ] Determine which versions will be required
    - [ ] Run CDK updates
    - [ ] Run server updates
    - [ ] Wait for server update to complete
    - [ ] Run data migrations
    - [ ] Wait for data migrations to complete
    - [ ] Loop if necessary

mattwiller:
Resolves #2785


packages/cli/src/aws/update-server.ts
```diff
@@ -1,32 +1,52 @@
...
+    .pop();
+}
+
+function deployServerUpdate(tag: string, config: MedplumInfraConfig): void {
+  const configFile = configFileName(tag);
+  writeConfig(config, tag);
+
+  const cmd = `npx cdk deploy -c config=${configFile}${config.region !== 'us-east-1' ? ' --all' : ''}`;
+  console.log('> ' + cmd);
+  const deploy = spawnSync(cmd);
```
codyebberson, line 46: Does `spawnSync()` allow interactivity with the spawned process?  I.e., if `cdk` prompts the user "Is this OK?", will you still be able to press "y" to proceed?

packages/cli/src/aws/index.ts
```diff
@@ -5,9 +5,16 @@ import { listStacksCommand } from './list';
 import { updateAppCommand } from './update-app';
 import { updateBucketPoliciesCommand } from './update-bucket-policies';
 import { updateServerCommand } from './update-server';
+import { createMedplumCommand } from '../util/command';
 
 export const aws = new Command('aws').description('Commands to manage AWS resources');
 
+const updateCmd = createMedplumCommand('update-server')
```
codyebberson, line 12: Did this change anything?  If not, preference to leave it as-is to be consistent with the other commands in the file.

packages/cli/src/aws/utils.ts
```diff
@@ -69,7 +70,11 @@ export async function getStackDetails(stackName: string): Promise<MedplumStackDe
   const result = {} as Partial<MedplumStackDetails>;
   await buildStackDetails(cloudFormationClient, stackName, result);
   if ((await cloudFormationClient.config.region()) !== 'us-east-1') {
-    await buildStackDetails(new CloudFormationClient({ region: 'us-east-1' }), stackName + '-us-east-1', result);
+    try {
+      await buildStackDetails(new CloudFormationClient({ region: 'us-east-1' }), stackName + '-us-east-1', result);
+    } catch {
+      // Fail gracefully
```
codyebberson, line 76: What is the case where this throws?  Can we print a warning?

The only reason I'm asking is that "printing more details" has be a theme during recent self-hosting customer support sessions.  See: https://github.com/medplum/medplum/issues/3117

packages/cli/src/aws/init.ts
```diff
@@ -270,7 +272,7 @@ export async function initStackCommand(): Promise<void> {
     await writeParameters(config.region, `/medplum/${config.name}/`, serverParams);
   } else {
     const serverConfigFileName = configFileName.replace('.json', '.server.json');
-    writeConfig(serverConfigFileName, serverParams);
+    writeConfig(config, config.name + '.server');
```
codyebberson, line 275: This is incorrect -- `serverParams` are the values for AWS Parameter Store, which is separate than the infra config.

It is totally valid critique that that distinction is bad, but that's the way it is for now :/

packages/cli/src/aws/update-server.ts
```diff
@@ -1,32 +1,52 @@
...
-    return;
+  const separatorIndex = config.serverImage.lastIndexOf(':');
+  const initialVersion = config.serverImage.slice(separatorIndex + 1);
+  let updateVersion = await nextUpdateVersion(initialVersion);
+  while (updateVersion) {
+    console.log(`Performing update to v${updateVersion}`);
+    config.serverImage = `${config.serverImage.slice(0, separatorIndex)}:${updateVersion}`;
+    deployServerUpdate(tag, config);
+    // Run data migrations
+    await client.post('/admin/super/migrate', '');
```
codyebberson, line 28: `/admin/super/migrate` is an async endpoint... I think we'll need to get the `AsyncJob` id from the response, and poll on it until it's done, yeah?

codyebberson (commented):
Nice!  Thanks for taking this @mattwiller, this is a pretty hairy one.

I added a bunch of comments, some total nits, but a few structural ones that we may need to discuss offline.

Thanks again, really really really appreciate you digging into this one 🙏

mattwiller:
@codyebberson Sorry about that, my branch got messed up trying to resolve merge conflicts; it should be better now

packages/cli/src/aws/update-server.ts
ThatOneBro: Actually you might wanna check out `inherit`, the default behavior will not keep the binding to the current `pty`

ThatOneBro:
```suggestion
  const deploy = spawnSync(cmd, { stdio: 'inherit' });
```

mattwiller: Thanks @ThatOneBro — I misread the Node docs 😅

PR #3092, opened 2023-10-20
Fixes #3083 - use ETag weak validator prefix

### LINKED ISSUE #3083: Returned ETag does not comply with FHIR spec (I think)
Hi!

I noticed Medplum is returning the `versionId` as is in the ETag header. The FHIR spec says the following:
> Note that the Version Id is considered a "weak" ETag and ETag headers should be prefixed with W/ and enclosed in quotes, for example:
> `ETag: W/"3141"`

See <https://hl7.org/fhir/R4/http.html#versioning>.

It only says "should", but I suppose there is no reason not to prefix it with `W/` ^^

Thank you!

codyebberson:


ThatOneBro (approved):
👍

PR #3093, opened 2023-10-20
Fixes #3081 - always show 'Add' button in filter editor

### LINKED ISSUE #3081: Filtering by boolean false flag in UI requires toggling checkbox
Hey team, noticed that currently if a user wants to filter by a boolean value being false in the Medplum app (for example patients that are not active), the filter UI currently requires toggling the checkbox to the on state before off state filter can then be added.

Please see here for a video demonstration:

https://github.com/medplum/medplum/assets/712727/887afaed-7765-4244-a597-5d0199451f46

I would have expected the "Add" button to appear immediately after selecting "equals" and not be required to toggle the checkbox on and then back off again. Thanks!

codyebberson:


ThatOneBro (approved):
Nice, checked it out. LGTM

Issue #2886, opened 2023-09-20
Fix Radio in Repeatable group


jamestouri:
Mantine's Radio Group gets Erratic if it's in a repeatable group. Reason for this is because Radio Group shows the selected value based on the name of the Radio Group and because they are matching with the exact same linkId or name it tends to render inconsistently.

jamestouri:
Closed from https://github.com/medplum/medplum/pull/3109

PR #3109, opened 2023-10-24
Refactor QuestionnaireForm

jamestouri:
Components interaction 

- `QuestionnaireForm`
  ↳ `QuestionnairePageSequence`

- `QuestionnairePageSequence`
  ↳ `QuestionnaireRepeatedGroup`
  ↳ `QuestionnaireRepeatableItem`
  
- `QuestionnaireRepeatedGroup`
  ↳ `QuestionnaireGroup`
  
- `QuestionnaireGroup`
  ↳ `QuestionnaireRepeatedGroup` (recursive)
  ↳ `QuestionnaireGroup` (recursive)
  ↳ `QuestionnaireRepeatableItem`
  
- `QuestionnaireRepeatableItem`
  ↳ `QuestionnaireFormItem`

- `QuestionnaireFormItem`

https://github.com/medplum/medplum/issues/2886


packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -383,21 +384,40 @@ function QuestionnaireChoiceRadioInput(props: QuestionnaireChoiceInputProps): JS
...
+function getCurrentRadioAnswer(options: [string, TypedValue][], defaultAnswer: TypedValue): string | undefined {
+  return options.find((option) => deepEquals(option[1].value, defaultAnswer?.value))?.[0];
+}
+
+function typedValueToString(typedValue: TypedValue | undefined): string | undefined {
+  if (!typedValue) {
+    return undefined;
+  }
+  if (typedValue.type === 'CodeableConcept') {
+    return typedValue.value.coding[0].display;
```
codyebberson, line 415: Can we use `formatCodeableConcept()` here?

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -383,21 +384,40 @@ function QuestionnaireChoiceRadioInput(props: QuestionnaireChoiceInputProps): JS
...
+  }
+  if (typedValue.type === 'Coding') {
+    return typedValue.value.display;
```
codyebberson, line 418: Can we use `formatCoding()` here?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -80,223 +87,269 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
-  index: number;
-  allResponses: QuestionnaireResponseItem[];
-  currentResponseItems: QuestionnaireResponseItem[];
-  groupSequence?: number;
-  setResponseItem: (responseId: string, newResponseItem: QuestionnaireResponseItem) => void;
+interface QuestionnaireRepeatableGroupProps {
+  readonly item: QuestionnaireItem;
+  readonly response: QuestionnaireResponseItem[];
+  readonly checkForQuestionEnabled: (item: QuestionnaireItem) => boolean;
+  readonly onChange: (responses: QuestionnaireResponseItem | QuestionnaireResponseItem[]) => void;
```
codyebberson, line 188: Is there a reason we need both single and array here?

codyebberson (commented):
Bunch of questions inline.

If we're going to go to this level of refactoring, I think we should also rip off the bandaid and move all of the various subcomponents into separate files.

I think my main point of confusion is around all of the cases of `value | array[]` -- It seems like we should be able to be tighter about that.  But I might just not be understanding something.

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -32,40 +32,44 @@ import {
...
+  const response = props.response;
+
+  if (!response) {
+    return null;
+  }
 
   function onChangeAnswer(
-    newResponseAnswer: QuestionnaireResponseItemAnswer | QuestionnaireResponseItemAnswer[],
-    repeatedIndex?: number
+    newResponseAnswer: QuestionnaireResponseItemAnswer | QuestionnaireResponseItemAnswer[]
```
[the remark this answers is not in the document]
jamestouri: MultiSelect in line 275 will have an array value for the `onChangeAnswer`

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
jamestouri: This is for a group nested inside another group 

e.g 
```
{
   type: group,
   text: groupOuter,
   linkId: groupOuter
   repeats: true, 
   item: [
         {
               type: group,
               text: groupInner,
               linkedId: groupInner,
               repeats: true, 
               item: [
                   { type: string,
                    text: string,
                    linkId: string
                  }
              ]
          }
     ]
}
              

```

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -32,40 +32,44 @@ import {
...
+  response?: QuestionnaireResponseItem;
   onChange: (newResponseItem: QuestionnaireResponseItem) => void;
 }
 
 export function QuestionnaireFormItem(props: QuestionnaireFormItemProps): JSX.Element | null {
   const item = props.item;
-  const index = props.index;
+  const response = props.response;
+
+  if (!response) {
```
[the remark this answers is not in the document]
jamestouri: Moved this to RepeatableItem to check. 

The QuestionnaireGroup is passing the prop down as 

```
   <QuestionnaireRepeatableItem
              key={item.linkId}
              item={item}
              response={response.item?.find((i) => i.linkId === item.linkId)}
              onChange={onSetGroup}
              checkForQuestionEnabled={checkForQuestionEnabled}
            />
```
So this is a check to make sure it's not undefined

codyebberson (approved):
Nice, this is looking promising @jamestouri !

I just did a bunch of testing on Storybook.

One thing I noticed is that we're now submitting empty answers for display items.  Is that intentional?

image

The nested repeating groups questionnaire looks solid.  Nice!

This is probably good to go.  Nice job 👍

jamestouri:
> One thing I noticed is that we're now submitting empty answers for display items. Is that intentional?
> 
> image

It's intentional (since we're already doing this in prod), we can change that though

PR #3111, opened 2023-10-24
Fixes #3063 Fixes #3106 - pushToAgent update and docs

### LINKED ISSUE #3063: `pushToAgent` initiates outbound connections
Currently, when `pushToAgent` assumes that an existing connection has been established by the listening device. However, many  listening HL7v2 systems expect the sender to establish the TCP connection. 

This task is to configure an Agent "channel" to initiate a TCP connection to a remote IP/port address, rathen than wait for an inbound connection

### LINKED ISSUE #3106: `pushToAgent` SDK method doesn't send correct payload
#3085 Recently updated the `Agent/$push` body to take 3 parameters: `body`, `destination`, and `contentType`. 

The `medplum.pushToAgent()` method needs to be updated to send this payload

codyebberson:
Link to new doc: https://medplum-www-git-cody-push-to-agent-fixes-and-docs-medplum.vercel.app/docs/agent/push


packages/core/src/utils.ts
```diff
@@ -54,17 +54,28 @@ export function createReference<T extends Resource>(resource: T): Reference<T> {
  * @param resource The FHIR resource.
  * @returns A reference string of the form resourceType/id.
  */
-export function getReferenceString(resource: Resource): string {
-  return resource.resourceType + '/' + resource.id;
+export function getReferenceString(resource: Reference | Resource): string {
+  const { reference, resourceType, id } = resource as any;
+  if (reference && typeof reference === 'string') {
```
rahul1, line 59: I think you can use the `isReference` [typeguard](https://github.com/medplum/medplum/blob/97a11229f78dba75116758af36813212f95b1100/packages/core/src/types.ts#L334-L337) here

packages/core/src/utils.ts
```diff
@@ -54,17 +54,28 @@ export function createReference<T extends Resource>(resource: T): Reference<T> {
...
+ * @param input A FHIR reference or resource.
  * @returns The ID portion of a reference.
  */
-export function resolveId(reference: Reference | undefined): string | undefined {
-  return reference?.reference?.split('/')[1];
+export function resolveId(input: Reference | Resource | undefined): string | undefined {
+  if (!input) {
+    return undefined;
+  }
+  const { reference, id } = input as any;
```
rahul1, line 74: Again, consider `isReference`

packages/docs/docs/agent/push.md
```diff
@@ -14,19 +14,82 @@ While the Medplum Agent traditionally connected healthcare facilities to cloud s
    - Relay a DICOM SR to a RIS using DICOM.
 4. **Simplified Infrastructure**: Eliminate the need for site-to-site VPNs, making setups more straightforward and secure.
 
+## How it works
+
+First, consider the most simple case: an HL7 device on the network:
```
rahul1, line 19: Suggestion: give an example of the real world use case
> an HL7 device on the network (e.g. a RIS -Radiology Information System)

packages/docs/docs/agent/push.md
```diff
@@ -14,19 +14,82 @@ While the Medplum Agent traditionally connected healthcare facilities to cloud s
...
+
+![Medplum Agent Push 1](./medplum-agent-push-1.webp)
+
+We can emulate this setup using the Medplum CLI:
+
+```bash
```
rahul1, line 25: I think it would read more cleanly if you put the emulation code for each step _before_ the diagram

packages/docs/docs/agent/push.md
```diff
@@ -14,19 +14,82 @@ While the Medplum Agent traditionally connected healthcare facilities to cloud s
...
+
+We can emulate this step using the Medplum CLI:
+
+```bash
+medplum hl7 send localhost 56000 --generate-example
+```
+
+This will send a message and print the reply to the terminal.
+
+The main challenge with HL7 interfaces is the need to be on the local network.
```
rahul1, line 41: Suggest adding some more context as to why the agent is a big deal

> The main challenge with HL7 interfaces is the need to be on the local network. Historically, this meant running the EHR software on-premise
>
> Instead, we will use the Medplum Agent to proxy traffic to the Medplum server hosted in AWS:

rahul1 (approved):
A few wordsmith suggestions on the doc, but non-blocking

PR #3125, opened 2023-10-25
Fixes #3118 - better upload error UX

### LINKED ISSUE #3118: Better error message on binary storage failure
Common case is during initial server setup when S3 bucket is not correctly configured

Currently the user gets `alert(undefined);` which is pretty bad

We should have some `OperationOutcome` from the server, and show in a nice Mantine notification.

codyebberson:
Before:

image

After:

image

### DIFF
--- a/packages/react/src/AttachmentButton/AttachmentButton.tsx
+++ b/packages/react/src/AttachmentButton/AttachmentButton.tsx
@@ -2,11 +2,13 @@ import { Attachment, Binary, OperationOutcome } from '@medplum/fhirtypes';
 import { useMedplum } from '@medplum/react-hooks';
 import React, { useRef } from 'react';
 import { killEvent } from '../utils/dom';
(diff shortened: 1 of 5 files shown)


packages/react/src/AttachmentButton/AttachmentButton.tsx
```diff
@@ -56,8 +58,10 @@ export function AttachmentButton(props: AttachmentButtonProps): JSX.Element {
           title: filename,
         });
       })
-      .catch((outcome: OperationOutcome) => {
-        alert(outcome.issue?.[0]?.details?.text);
+      .catch((err) => {
+        if (props.onUploadError) {
+          props.onUploadError(normalizeOperationOutcome(err));
+        }
```
mattwiller, line 64: Should there be some default behavior if the prop isn't set?

mattwiller (approved):

PR #3143, opened 2023-10-26
feat(client/fhircast): add support for `STU3` events

ThatOneBro:
## What this PR does 
* Adds support for **most** standard events for [`FHIRcast STU3`](https://build.fhir.org/ig/HL7/fhircast-docs/3_Events.html) (minus `DiagnosticReport-update` and `DiagnosticReport-select`, see #3138)
* Refactors the type system for `fhircastPublish` to support more flexible future extension and later custom types

## TODO
- [x] Add tests for new events
- [x] Update server-side `.well-known/fhircast-configuration` to include new events
- [x] Add tests for `syncerror`
- [x] Add tests for new validation paths

Special thanks to @colinhacks for figuring out how to extract `resourceType` for each resource individually from the `FHIRCAST_EVENT_RESOURCES` object via some TypeScript magic.


examples/medplum-fhircast-demo/src/hooks/index.ts
```diff
@@ -8,6 +8,7 @@ export function useClientId(): string {
 
 export function usePrevious<T>(value: T): T | undefined {
```
codyebberson, line 9: What is this doing?  Seems like it's just a passthrough?

packages/core/src/client.test.ts
```diff
@@ -1149,7 +1149,7 @@ describe('Client', () => {
   describe('FHIRcast', () => {
```
codyebberson, line 1149: Would it be ok to move these tests into a separate file?  It's a pretty clean cut on functionality, and would help get this file under control.

Not urgent, we can do that in separate PR.

codyebberson (approved):
Nice, looks pretty good.  2 minor comments.

examples/medplum-fhircast-demo/src/hooks/index.ts
ThatOneBro: Haha, yeah this a silly one but it basically keeps the value of the last render because of how `useEffect` works. If I remember correctly because of how `ref.current` is evaluated before the `useEffect` runs, it returns the value from the last render while storing the current value for the next render.

Basically it's useful for preventing running logic in hooks when nothing actually changed (something equivalent to `shouldComponentUpdate` functionality from class components).

Not sure why this hook isn't built in

PR #3161, opened 2023-10-27
Add sort-package-json tool

codyebberson:


ThatOneBro (approved):
Nice, it always annoys me when different sections are not in the same place throughout a monorepo. LGTM

PR #3171, opened 2023-10-29
Markdown fixes for Docusaurus 3

codyebberson:
Docusaurus 3 is coming soon.  One of the big changes is moving to MDX 3, which is considerably more strict than previous versions.  The Docusaurus team released a tool to help identify potential issues.  Read more here: https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3#common-mdx-problems

This PR fixes a bunch of the issues in our hand written MDX code:
* Replace `<https://example.com>` links with `[https://example.com](https://example.com)`
* Replace `<details><summary>` pattern with a `<DetailsBlock>` react component
* Escape `{` and `}` characters

Unfortunately, our autogenerated docs have a bunch of issues.  The autogenerated FHIR docs can be fixed with our custom tooling.  Autogenerated `typedoc` docs are a different story.  This may be what pushes us to kill TypeDoc.


reshmakh (approved):

packages/docs/src/components/DetailsBlock.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export interface DetailsBlockProps {
+  summary: string;
+  children?: React.ReactNode;
```
rahul1, line 5: request to add an optional "id" prop here, to allow us to send links to the details blocks. 

This is really helpful for sharing examples with customers.

```tsx
<summary id={pops.id}>
  <a href={props.id}>{props.summary}</a>
</summary>
```

codyebberson: I actually think I need to back out all of the `<details>` and `<summary>` changes, because they don't render correctly.

I don't quite understand why the `npx docusaurus-mdx-checker` tool complains about them.  1, they're valid JSX, and 2, they're a documented pattern, and 3, they don't work when used in a React component?

rahul1: That's fine. I was actually surprised you needed to this chagne in the first place

codyebberson: It turns out it doesn't like it when `<details>` and `<summary>` are on the same line 🤷  Moving them to new lines fixed the warnings.  It must be something about those particular tags.

PR #3172, opened 2023-10-29
Replace typedoc with api-documenter

codyebberson:
This is a proof-of-concept that replaces `typedoc` with `api-documenter`

We recently adopted `api-extractor` for generating TypeScript type definitions.  That ecosystem of tools also includes `api-documenter`, which supports extracting docs as Markdown.

In theory, that would be great.  `typedoc` is extremely slow.  This would speed up our build, and align with a more popular way of generating docs.

Unfortunately, `api-documenter` currently only outputs Markdown, not MDX.  Docusaurus always parses content as MDX, regardles of `.md` or `.mdx` file extension.  The generated Markdown does not escape curly braces or angle braces, which causes the MDX parser to choke.

There is an open issue to support MDX, but it hasn't seen much activity: https://github.com/microsoft/rushstack/issues/3625

For this to work, we would need to find an additional tool to convert Markdown to MDX.  That might be as simple as escaping curly braces and angle braces, but it might be much more than that 🤷 

--------

Update: Nov 2, 2023

I believe I was able to fix all issues with 2 big changes:
1. All JSDoc/TSDoc Markdown must be valid Markdown.  Before, we had a few cases of malformed numbered lists.
2. All code blocks must be prefixed with `@example`

After fixing those, everything builds and renders correctly.

Then there were a long list of broken links to change from the TypeDoc URL format to the API Documenter URL format.


ThatOneBro:
Maybe try this out? https://github.com/gabrielcsapo/docusaurus-plugin-api-extractor

reshmakh (approved):

PR #3195, opened 2023-11-01
feat: Pass resource to Repository logEvent on error

jdjkelly:
## Why
* `logAuditEvent` optionally supports passing a `Resource` but Repository doesn't take advantage of this in two places where it would be valuable: error cases for `createResource` and `updateResource`
* This is valuable because when an Update fails due to a validation in a Batch Transaction, being able to determine which resource failed to update is helpful
* More context for this PR available in this Discord discussion: https://discord.com/channels/905144809105260605/1149410163808284713

## What
* Pass resource to `this.logEvent` on the error cases for `createResource` and `updateResource`


codyebberson (approved):

codyebberson:
Thanks @jdjkelly!

Issue #2994, opened 2023-10-06
Log all server requests


mattwiller:
To improve observability of Medplum server, we should log all HTTP requests received and processed.  This would likely need to be added as a middleware during [server setup](https://github.com/medplum/medplum/blob/76adbaba6e1d5bdee532c75b8bbac2dbaf8379c6/packages/server/src/app.ts#L121).  The middleware should log at least the following data for each request handled:
- Timestamp the request was received (__not__ the timestamp when the log was written)
- HTTP request method and path
- User profile reference, or none if not authenticated
- Client IP address
- Time taken to process the request (from when it was first processed to when the response is rendered)
- Status code returned
- Number of bytes sent in the response to the client
- User-Agent string

codyebberson:
This should definitely be configurable.  In production systems, you get all of this for free through ALB logging.  See: https://www.medplum.com/docs/self-hosting/aws-athena-guide

Are there any standards/conventions that we can align with, especially thinking in context of this discussion https://github.com/medplum/medplum/discussions/2759 ?

Dharshan-K:
Hi, @mattwiller, I am interested in this issue to build a logging service.

mattwiller:
Hey @Dharshan-K, please feel free to get started with this one!  I see you've already taken a look at the broader [Observability plan](https://github.com/medplum/medplum/discussions/2759) — this is a great piece of that plan for you to contribute.  We're trying to take an iterative, piece-by-piece approach to adding observability features to Medplum server, so for this issue I have a couple recommendations for your implementation:
- As @codyebberson pointed out, this logging should be configurable: one should be able to turn off the access log if it's too verbose or redundant
- For now, prefer a simple implementation of the access log described above over broader changes to the server logging: it'll be easier to merge your PR if it's more focused and narrow in scope

vaibhavpnimkar:
Hello @mattwiller if no one is working on this i would like take this issue

mattwiller:
Hey @vaibhavpnimkar — I believe @Dharshan-K was planning to try working on this issue, but if they aren't still interested then it's all yours

Dharshan-K:
Hi @mattwiller I have written the logging middleware, which logs the necessary data mentioned in this issue. How to calculate the bytes sent and get the user reference. below is the code.


```javascript
const loggingMiddleware = (req: Request, res: Response, next: NextFunction) =>{
  const start = new Date();  
  next();
  const afterNext = new Date().getTime(); // Record the time after next() completes
[...]

PR #3196, opened 2023-11-01
Middleware to log necessary data for every request. 

Dharshan-K:
This PR is for the Logging Middleware addressed in the issue [2994](https://github.com/medplum/medplum/issues/2994).


packages/server/src/app.ts
```diff
@@ -205,3 +206,12 @@ export async function shutdownApp(): Promise<void> {
     rmSync(binaryStorage.replace('file:', ''), { recursive: true, force: true });
   }
 }
+
+const loggingMiddleware = (req: Request, res: Response, next: NextFunction):void =>{
+  const start = new Date();
+  next();
+  const afterNext = new Date().getTime(); // Record the time after next() completes
+  const totalTime = afterNext - start.getTime(); // Calculate the time taken including the time spent in next()  
+  globalLogger.log('INFO', 'logging data', {requestReceivedTimestamp:start,requestMethodAndPath: `${req.method} ${req.path}`,timeTaken: `${totalTime} ms`, IP_address: req.ip, status: res.statusCode, userAgentString: req.get('User-Agent')});
```
mattwiller, line 215: Instead of using the global logger here, it would be better to use the request-specific one in order to get the request ID and trace ID attached:
```ts
const ctx = getRequestContext();
ctx.logger.info(/* ... */);
```

Additionally, you can access the current user's profile information through the request context:
```ts
let userProfile: string | undefined;
if (ctx instanceof AuthenticatedRequestContext) {
  userProfile = ctx.profile.reference;
}
```

A few other minor notes:
- `INFO` is the correct level for this log, but the log message should be more meaningful: consider changing this from "logging data" to something like "Request served" to describe what the log entry is recording
- It's a bit hard to scan this long line with all the data properties, could you put each property on its own line?
- Since this log is going to be written many times and will likely make up a significant portion of overall log volume, I would like to try making the property names shorter to save space.  I know this might seem nitpicky, but tens of bytes can add up fast at scale!  I would suggest the following property names: `receivedAt`, `method`, `path`, `duration`, `ip`, `status`, and `ua` for brevity

mattwiller (commented):
Thanks for the PR @Dharshan-K — I left some comments inline, but this is a great start!  Please feel free to try making the requested changes, and let me know if there's anything I can clarify for you

packages/server/src/app.ts
Dharshan-K: Hi @mattwiller, Thank for your feedback. I have changed the code.

Dharshan-K:
Hi @mattwiller, Thank for your feedback. I have changed the code.

packages/server/src/app.ts
```diff
@@ -205,3 +206,25 @@ export async function shutdownApp(): Promise<void> {
...
+  const totalTime = afterNext - start.getTime(); // Calculate the time taken including the time spent in next()  
+  let userProfile: string | undefined;
+  if (ctx instanceof AuthenticatedRequestContext) {
+    userProfile = ctx.profile.reference;
+  }
+
+  ctx.logger.info('Request served', 
+  {
+    receivedAt:start,
+    requestMethodAndPath: `${req.method}: ${req.path}`,
```
mattwiller, line 224: Could you split this into two different properties: `method` and `path`?  That will make it easier to query the logs for users

packages/server/src/app.ts
```diff
@@ -205,3 +206,25 @@ export async function shutdownApp(): Promise<void> {
...
+    Duration: `${totalTime} ms`, 
+    IP: req.ip, status: res.statusCode, 
+    UA: req.get('User-Agent'),
+    Profile: userProfile
```
mattwiller, line 228: These should all have consistent casing:
```suggestion
    durationMs: `${totalTime}`, 
    ip: req.ip,
    status: res.statusCode, 
    ua: req.get('User-Agent'),
    profile: userProfile
```

mattwiller (commented):
This is looking really great, thank you following up @Dharshan-K!  It looks like you might have some formatting issues that prettier is complaining about in CI, and I left a couple very minor comments inline: if you could touch those items up we can get this PR merged!

Dharshan-K:
Hi @mattwiller, Changes were made to match Prettier workflow. Thank you

codyebberson (approved):

Dharshan-K:
Hi, @codyebberson, the build has failed. what should I do with prettier.

codyebberson:
Hi @Dharshan-K - it looks like an intermittent GitHub Actions error.  I'll push it through.  No action required.  Thank you!

Issue #3194, opened 2023-10-31
Add App Router to Nextjs Example Project


jamestouri:

codyebberson:
Fixed in https://github.com/medplum/medplum/pull/3197

PR #3197, opened 2023-11-01
Update Next Example project to use App Router

jamestouri:
https://github.com/medplum/medplum/assets/6913745/b5450035-74a2-45c3-97f2-2809a70691b9

Favicon and Title 

Screenshot 2023-11-01 at 1 57 27 PM

https://github.com/medplum/medplum/issues/3194


examples/medplum-nextjs-demo/app/root.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const medplum = new MedplumClient({
+  // Uncomment this to run against the server on your localhost
+  // baseUrl: 'http://localhost:8103/',
+
+  // Handle unauthenticated requests
+  onUnauthenticated: () => (window.location.href = '/'),
+
+  // Use Next.js fetch
+  fetch: (url: string, options?: any) => fetch(url, options),
```
ThatOneBro, line 15: I think we might need to test this a little, Next has done some monkey patching around `fetch` in Next 13+ https://nextjs.org/docs/app/api-reference/functions/fetch#:~:text=Next.js%20extends,persistent%20caching%20semantics.

So we should make sure everything works properly

examples/medplum-nextjs-demo/app/page.tsx
```diff
@@ -1,13 +1,16 @@
+'use client';
```
ThatOneBro, line 1: We should eventually try to see if we can support React Server Components, but this is sensible for now because of how browser-specific our `MedplumClient` is.

ThatOneBro (approved):

ThatOneBro:
Overall LGTM, some small notes but shouldn't block the PR I think.

examples/medplum-nextjs-demo/app/root.tsx
codyebberson: I don't know anything about Next.js 14 or current state, but at least at the time this was authored, this line was necessary to take advantage of the Next.js fetch monkey patch.  If this line does not exist, then `MedplumClient` tries to use the global fetch, which does not work (did not work?) with Next.

codyebberson (approved):

PR #3205, opened 2023-11-02
Fixed api-extractor warnings

codyebberson:
Last week, we started using `api-extractor` for better TypeScript definitions: https://github.com/medplum/medplum/pull/3152

`api-extractor` is more strict.  This PR fixes all warnings:
* Certain characters must be escaped (`@`, `<`, `>`, `{`, `}`) or inside of a backtick block.
* Any type that is referenced by an exported type must also be exported
* Added `@yields` as a custom jsdoc tag


mattwiller (approved):
✨

PR #3210, opened 2023-11-02
Fix "Id" capitolization issue in field selector

### LINKED ISSUE #2891: Fix capitalization of "Id" in field selector

mattlong:
Fixes #2891.

The `_id` version of the id field wasn't being handled in `buildFieldNameString`.

There was a related issue in `SearchFieldEditor` that this addresses as well where "ID" and "Id" would both be available:

Screenshot 2023-11-02 at 11 52 14 AM

Even with this change, it's still possible to (maunally) add both `_id` and `id` to the list of fields shown, e.g. [http://localhost:3000/Patient?_fields=_id,id](http://localhost:3000/Patient?_fields=_id,id). This doesn't seem like too big of a deal, but wanted to call it out.

### DIFF
--- a/packages/react/src/SearchControl/SearchUtils.test.tsx
+++ b/packages/react/src/SearchControl/SearchUtils.test.tsx
@@ -482,6 +482,9 @@ describe('SearchUtils', () => {
   test('buildFieldName', () => {
     expect(buildFieldNameString('id')).toBe('ID');
     expect(buildFieldNameString('Id')).toBe('ID');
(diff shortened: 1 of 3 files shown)


codyebberson (approved):
w00t

PR #3211, opened 2023-11-02
POC of HMR working in app for changes in packages/react

mattlong:
Based on Approach 2 from [this discussion](https://github.com/vitejs/vite/discussions/7155#discussioncomment-4121062).

This could of course be extended to other libraries besides `@medplum/react`.

One caveat is that it currently isn't possible to specify a different tsconfig file besides the default `tsconfig.json` in Vite which is kind of wild. If we actually wanted to pursue this, we'd need to use or build a plugin like the ones in [this discussion](https://github.com/vitejs/vite/discussions/8483) or get creative in our `package.json` scripts to swap in/rename the appropriate tsconfig file for dev vs building.

### DIFF
--- a/packages/react/src/utils/recaptcha.ts
+++ b/packages/react/src/utils/recaptcha.ts
@@ -1,5 +1,7 @@
 import { createScriptTag } from './script';
 
+declare let grecaptcha: undefined | ReCaptchaV2.ReCaptcha;
+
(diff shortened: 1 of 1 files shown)


codyebberson:
Finalized here: https://github.com/medplum/medplum/pull/3229

Issue #3009, opened 2023-10-08
Investigate noUncheckedIndexedAccess


codyebberson:
https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess

PR #3214, opened 2023-11-02
Support custom creation paths

### LINKED ISSUE #2888: Fix "New" button for Bot and ClientApplication
Creating Bots and ClientApplications is a slightly special case, that does not use the normal "New" workflow.

To avoid confusion, we omit the normal "New" button on the search page.  Instead, we should include the "New" button, but have it go to the correct destination.

image

See code here: https://github.com/medplum/medplum/blob/main/packages/app/src/HomePage.tsx#L73-L79

Rather than disabling the "New" button, the new button should go to the respective create pages.

Bots:  https://app.medplum.com/admin/bots/new

ClientApplications: https://app.medplum.com/admin/clients/new

mattlong:
Fixes #2888

### DIFF
--- a/packages/app/src/HomePage.utils.ts
+++ b/packages/app/src/HomePage.utils.ts
@@ -9,6 +9,11 @@ import {
 } from '@medplum/core';
 import { Bundle, ResourceType, UserConfiguration } from '@medplum/fhirtypes';
 
+/** Custom navigation paths when the user clicks New... */
+export const RESOURCE_TYPE_CREATION_PATHS: Partial<Record<ResourceType, string>> = {
+  Bot: '/admin/bots/new',
+  ClientApplication: '/admin/clients/new',
+};
 export function addSearchValues(search: SearchRequest, config: UserConfiguration | undefined): SearchRequest {
   const resourceType = search.resourceType || getDefaultResourceType(config);
   const fields = search.fields ?? getDefaultFields(resourceType);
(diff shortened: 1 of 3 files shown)


packages/app/src/HomePage.utils.ts
```diff
@@ -9,6 +9,11 @@ import {
 } from '@medplum/core';
 import { Bundle, ResourceType, UserConfiguration } from '@medplum/fhirtypes';
 
+/** Custom navigation paths when the user clicks New... */
+export const RESOURCE_TYPE_CREATION_PATHS: Partial<Record<ResourceType, string>> = {
```
mattwiller, line 13: I haven't seen this TS pattern before: `Partial<Record<K, V>>` — what does `Partial` do here?

mattwiller (approved):

packages/app/src/HomePage.utils.ts
mattlong: It makes all keys (i.e. ResourceTypes) optional in the object while still providing the type safety that every key specified is indeed a `ResourceType` to prevent you from fat-fingering `Patent` instead of `Patient`.

In the parts of the code I've come across so far, it looks like we tend to do `Record<string, ...>` when only a subset of resource types need to be included which doesn't offer the same type safety.

It comes at the expense of not being able to use `RESOURCE_TYPE_CREATION_PATHS.Bot` without typescript complaining, but that's arguably could be for the best depending on the situation.

mattwiller: Cool, thanks for the explanation! 🙌

codyebberson: Correct me if wrong, but I think this may have worked without `Partial` right now.  We have an open ticket to look into the `noUncheckedIndexedAccess` compiler flag, which would tighten that up. 
 https://github.com/medplum/medplum/issues/3009

codyebberson (approved):

PR #3219, opened 2023-11-03
Enables creating new bots/clients from the UI

### LINKED ISSUE #2888: Fix "New" button for Bot and ClientApplication

koox00:
This commit enables the button for creating new bots and client applications. The correct URLs for creating these 2 resources are behind admin routes.

Fixes #2888

### DIFF
--- a/packages/app/src/HomePage.utils.ts
+++ b/packages/app/src/HomePage.utils.ts
@@ -118,6 +118,24 @@ export function canCreate(resourceType: string): boolean {
   return resourceType !== 'Bot' && resourceType !== 'ClientApplication';
 }
 
+export function getNewResourceEndpoint(resourceType: string, isProjectAdmin: boolean): string {
+  if (canCreate(resourceType)) {
+    return `/${resourceType}/new`;
+  }
+
+  if (isProjectAdmin) {
+    if (resourceType === 'Bot') {
+      return `/admin/bots/new`;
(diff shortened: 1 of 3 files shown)


packages/app/src/HomePage.utils.ts
```diff
@@ -118,6 +118,24 @@ export function canCreate(resourceType: string): boolean {
   return resourceType !== 'Bot' && resourceType !== 'ClientApplication';
 }
 
+export function getNewResourceEndpoint(resourceType: string, isProjectAdmin: boolean): string {
```
koox00, line 121: I didn't dig much into the codebase, but could this benefit from better typing?

```suggestion
export function getNewResourceEndpoint(resourceType: ResourceType, isProjectAdmin: boolean): string {
```

The same goes with `canCreate`.

PR #3228, opened 2023-11-05
server centralize request validation for co-location and reduced redundancy

dillonstreator:


packages/server/src/auth/newuser.ts
```diff
@@ -31,12 +32,6 @@ export async function newUserHandler(req: Request, res: Response): Promise<void>
     return;
   }
 
-  const errors = validationResult(req);
-  if (!errors.isEmpty()) {
-    sendOutcome(res, invalidRequest(errors));
-    return;
-  }
-
```
dillonstreator, line 39: Not sure if this approach is viable given that the behavior is slightly different here in that request validation would now run prior to the config check

codyebberson:
Thanks @dillonstreator - this looks great.

dillonstreator:
Thank you @codyebberson!
Please let me know if there is anything I can do to support this through the pipeline. I see some workflow failures but I believe they're outside of my control. This is lower priority than other work so no rush!

packages/server/src/auth/newuser.ts
codyebberson: It's ok 👍

codyebberson (approved):
Nice, thanks @dillonstreator !

Love this:
image

PR #3229, opened 2023-11-05
HMR app on react changes

codyebberson:
Building on @mattlong's POC here: https://github.com/medplum/medplum/pull/3211

I couldn't get `@types/grecaptcha` to work with Vite aliasing.  Vite aliasing seems more valuable than those type definitions (we only use it in a couple places), so I'm inclined to push this through.

I was curious if we could go ahead and do this for all monorepo dependencies, most notably `@medplum/core`.  Unfortunately, the answer is "not yet", because `@medplum/core` is not yet using ESM in source.  For example, it still has references to `process.env` rather than `import.meta.env`.  ESM is a slippery slope, and has often caused issues with `@medplum/server`, so I'm leaving it alone for now.  Exciting future prospects though.

### DIFF
--- a/examples/foomedical/package.json
+++ b/examples/foomedical/package.json
@@ -36,7 +36,6 @@
     "@tabler/icons-react": "2.39.0",
     "@testing-library/jest-dom": "6.1.4",
     "@testing-library/react": "14.0.0",
-    "@types/grecaptcha": "3.0.5",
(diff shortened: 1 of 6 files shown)


mattwiller (approved):
Looks good to me!

codyebberson:
Deployed and tested on staging: https://app.staging.medplum.com/

PR #3232, opened 2023-11-05
feat(fhircast): add `context.priorVersionId` to `DiagnosticReport-update`

### LINKED ISSUE #3213: [FHIRcast] Add `priorVersionId` to `DiagnosticReport-update` on the Hub-distributed version
"The Hub SHALL distribute a corresponding event to all Subscribers. The Hub SHALL replace the context.versionId in the request with a new context.versionId generated and retained by the Hub. The prior version, context.priorVersionId of the context is also provided to ensure that a Subscriber is currently in sync with the latest context prior to applying the new updates. If the value of context.priorVersionId is not in agreement with the context.versionId last received by a Subscriber, it is recommended that the Subscriber issue a GET request to the Hub in order to retrieve the latest version of the context (note that the GET request returns the context, all existing content, and the current context.versionId)."

Source: https://build.fhir.org/ig/HL7/fhircast-docs/3-6-3-DiagnosticReport-update.html#diagnosticreport-update-event-example

ThatOneBro:
Closes #3213

### DIFF
--- a/packages/server/src/fhircast/routes.ts
+++ b/packages/server/src/fhircast/routes.ts
@@ -110,6 +110,13 @@ protectedRoutes.post(
     } else if (event['hub.event'].endsWith('-close')) {
       // We always close the current context, even if the event is not for the original resource... There isn't any mention of checking to see it's the right resource, so it seems it may be assumed to be always valid to do any arbitrary close as long as there is an existing context...
       await getRedis().del(`::fhircast::${req.params.topic}::latest::`);
+    } else if (event['hub.event'] === 'diagnosticreport-update') {
+      // See: https://build.fhir.org/ig/HL7/fhircast-docs/3-6-3-DiagnosticReport-update.html#:~:text=The%20Hub%20SHALL,the%20new%20updates.
+      event['context.priorVersionId'] = event['context.versionId'];
+      event['context.versionId'] = generateId();
+      stringifiedBody = JSON.stringify(req.body);
+      // TODO: Make sure this is actually supposed to be stored / overwrite open context? (ambiguous from docs, see: https://build.fhir.org/ig/HL7/fhircast-docs/2-9-GetCurrentContext.html)
+      await getRedis().set(`::fhircast::${req.params.topic}::latest::`, stringifiedBody);
     }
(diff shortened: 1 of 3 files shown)


packages/server/src/fhircast/routes.ts
```diff
@@ -110,6 +110,13 @@ protectedRoutes.post(
     } else if (event['hub.event'].endsWith('-close')) {
       // We always close the current context, even if the event is not for the original resource... There isn't any mention of checking to see it's the right resource, so it seems it may be assumed to be always valid to do any arbitrary close as long as there is an existing context...
       await getRedis().del(`::fhircast::${req.params.topic}::latest::`);
+    } else if (event['hub.event'] === 'diagnosticreport-update') {
+      // See: https://build.fhir.org/ig/HL7/fhircast-docs/3-6-3-DiagnosticReport-update.html#:~:text=The%20Hub%20SHALL,the%20new%20updates.
+      event['context.priorVersionId'] = event['context.versionId'];
+      event['context.versionId'] = generateId();
+      stringifiedBody = JSON.stringify(req.body);
+      // TODO: Make sure this is actually supposed to be stored / overwrite open context? (ambiguous from docs, see: https://build.fhir.org/ig/HL7/fhircast-docs/2-9-GetCurrentContext.html)
+      await getRedis().set(`::fhircast::${req.params.topic}::latest::`, stringifiedBody);
```
codyebberson, line 119: Out of curiosity, why double colons?   And why the double colon prefix?

In the rest of the server, we use single slash `/`, mostly because I didn't know any better when we started.

If I were to do it from scratch, I'd use single colon `:`, because a lot of Redis tools use that as a standard delimiter, which can help with visualization and debugging.

As far as I know, there is no actual behavior differences, Redis treats them all as opaque strings.  So just curious.

codyebberson (approved):
lgtm

One minor comment on Redis keys.  Feel free to disregard.

packages/server/src/fhircast/routes.ts
ThatOneBro: I guess the idea was to kind of make it clear this is an internal key when looking through Redis, but maybe that's unnecessary. If it doesn't sit well I can change it

PR #3234, opened 2023-11-06
Fixes #2892 - Use Mantine components for field editor

### LINKED ISSUE #2892: Field editor not using Mantine styles
image

`packages/react/src/SearchFieldEditor/SearchFieldEditor.tsx`

Note that it's using default styles and default font for a `<select>` element.

Instead, it should use the Mantine component for a multiline selector.

mattlong:
Main interesting bit of this was adding the special handling to avoid closing the modal along with the multiselect dropdown.

[Loom of new behavior](https://www.loom.com/share/364c2bcdd6bb4bb58f7530d2b3bcb1a8?sid=9f4b64be-45f9-4e3b-8819-9f95552e8a1c)

### DIFF
--- a/packages/docs/docs/search/basic-search.mdx
+++ b/packages/docs/docs/search/basic-search.mdx
@@ -81,7 +81,7 @@ The array returned by `searchResources` also includes a `bundle` property that c
 
 ## Searching Multiple Criteria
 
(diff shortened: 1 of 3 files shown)


codyebberson (approved):
Nice

This is hot:

image

image

PR #3239, opened 2023-11-06
fix-3238 gracefully shutdown server

### LINKED ISSUE #3238: Gracefully shut down main `packages/server` process
The server package defines a [`shutdownApp`](https://github.com/medplum/medplum/blob/fbf271ba1389e850d62d154c466cf4a3cda4b173/packages/server/src/app.ts#L190) function but it appears to only get called in tests to perform clean up. We should hook into common process kill signals and clean up before shutting down the server process.

https://www.npmjs.com/package/http-graceful-shutdown might be a good option to consider this functionality.

dillonstreator:
Fixes #3238

### DIFF
--- a/packages/server/src/config.ts
+++ b/packages/server/src/config.ts
@@ -47,6 +47,7 @@ export interface MedplumServerConfig {
   introspectionEnabled?: boolean;
   keepAliveTimeout?: number;
   vmContextBotsEnabled?: boolean;
+  shutdownTimeoutMilliseconds?: number;
 }
 
 /**
@@ -264,6 +265,7 @@ function addDefaults(config: MedplumServerConfig): MedplumServerConfig {
   config.botLambdaLayerName = config.botLambdaLayerName || 'medplum-bot-layer';
   config.bcryptHashSalt = config.bcryptHashSalt || 10;
   config.bullmq = { removeOnComplete: { count: 1 }, removeOnFail: { count: 1 }, ...config.bullmq };
+  config.shutdownTimeoutMilliseconds = config.shutdownTimeoutMilliseconds ?? 30000;
   return config;
 }
 
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1,7 +1,8 @@
 import express from 'express';
-import { initApp } from './app';
+import { initApp, shutdownApp } from './app';
 import { loadConfig } from './config';
 import { globalLogger, parseLogLevel } from './logger';
+import gracefulShutdown from 'http-graceful-shutdown';
 
 export async function main(configName: string): Promise<void> {
   process.on('unhandledRejection', (err: any) => {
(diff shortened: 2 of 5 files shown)


packages/server/src/config.ts
```diff
@@ -264,6 +265,7 @@ function addDefaults(config: MedplumServerConfig): MedplumServerConfig {
   config.botLambdaLayerName = config.botLambdaLayerName || 'medplum-bot-layer';
   config.bcryptHashSalt = config.bcryptHashSalt || 10;
   config.bullmq = { removeOnComplete: { count: 1 }, removeOnFail: { count: 1 }, ...config.bullmq };
+  config.shutdownTimeoutMilliseconds = config.shutdownTimeoutMilliseconds ?? 30000;
```
dillonstreator, line 268: This is redundant in its current state as the default timeout for `http-graceful-shutdown` is already 30 seconds https://github.com/sebhildebrandt/http-graceful-shutdown#options

I can remove, unless we'd like to adjust the default.

codyebberson: 30 seconds is appropriate.  I don't have strong feelings about whether to be explicit or implicit here.

codyebberson:
For my own curiosity, the AWS Fargate Task timeout is also 30 seconds: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_StopTask.html

In the common case, server shutdown should be fast (under 1 second).  The exceptions being:

1. Long running HTTP requests, such as a large FHIR `transaction` bundle
2. Long running BullMQ jobs, such as "rebuild structure definitions", "reindex resources", etc.

Those can take upwards of 30 seconds.

Overall, it feels good to be explicit.  Thanks @dillonstreator

packages/server/src/index.ts
```diff
@@ -22,6 +23,17 @@ export async function main(configName: string): Promise<void> {
   const server = app.listen(config.port);
   server.keepAliveTimeout = config.keepAliveTimeout ?? 90000;
   globalLogger.info('Server started', { port: config.port });
+  gracefulShutdown(server, {
+    timeout: config.shutdownTimeoutMilliseconds,
+    development: process.env.NODE_ENV !== "production",
```
dillonstreator, line 28: @codyebberson might be nice if the config defined environment with an enum of sorts🤔

codyebberson:
@dillonstreator  - Apologies for the inconvenience, but if you rebase & resubmit, the Github Actions problems should be fixed.  Or let me know, and we can do it on your behalf.

dillonstreator:
> @dillonstreator - Apologies for the inconvenience, but if you rebase & resubmit, the Github Actions problems should be fixed. Or let me know, and we can do it on your behalf.

All good - I've rebased on `upstream/main` and we should now be set 👍

codyebberson (approved):

Issue #2009, opened 2023-05-10
Create demo of Measure and MeasureReport


codyebberson:
- [x] Create a `Measure` with a numerator and denominator `Expression`
- [x] Create a `Bot` that executes the 2 counts and creates a `MeasureReport`
- [x] Optionally put the `Bot` on a cron
- [ ] Optionally create a React widget that shows the most recent `MeasureReport`
- [ ] Optionally create a React widget that shows the `MeasureReport` values over time

codyebberson:
Rather than using a `Bot`, we may also want to implement the `Measure` [`$evaluate-measure`](https://hl7.org/fhir/measure-operation-evaluate-measure.html) operation.

codyebberson:
@jamestouri - See "Stats with ring progress" for an example of what we want for a `<MeasureReportDisplay>` component

https://ui.mantine.dev/category/stats/#stats-ring

PR #3242, opened 2023-11-06
MeasureReportDisplay 

jamestouri:
https://github.com/medplum/medplum/issues/2009 part 4 of 5

Screenshot 2023-11-06 at 9 08 31 AM
Screenshot 2023-11-06 at 9 08 43 AM
Screenshot 2023-11-06 at 9 24 40 AM
Screenshot 2023-11-06 at 9 43 20 AM


packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
```diff
@@ -0,0 +1,106 @@
...
+    return 0;
+  }
+  if (score <= 1 && unit === '%') {
+    return score * 100;
+  }
+  return score;
+}
+
+function setColor(score: number): string {
+  if (score <= 33) {
```
mattwiller, line 99: Should users be able to customize these breakpoints?  I could see someone wanting to set up different thresholds for their own measures

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
```diff
@@ -0,0 +1,106 @@
...
+  );
+}
+
+function MeasureReportPopulation(props: any): JSX.Element {
+  const { group } = props;
+  const populations = group.population;
+  const numerator = populations?.find((p: any) => formatCodeableConcept(p.code) === 'numerator');
+  const denominator = populations?.find((p: any) => formatCodeableConcept(p.code) === 'denominator');
+
+  const value = (numerator?.count / denominator?.count) * 100;
```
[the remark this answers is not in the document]
jamestouri: I have them both equal to 1, although I can have a louder message that this is only a partial amount of data here

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
jamestouri: I've been trying to think through this... will it be for each group, or perhaps the entire MeasureReport will set up the same thresholds? It could be there for a larger discussion/next iteration

packages/react/src/MeasureReportDisplay/MeasureReportDisplay.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [measure, setMeasure] = React.useState<Measure | undefined>();
+
+  useEffect(() => {
+    medplum
+      .searchOne('Measure', { url: report?.measure })
+      .then((result: Measure | undefined) => {
+        setMeasure(result);
+      })
+      .catch(console.log);
+  }, [medplum, report?.measure]);
```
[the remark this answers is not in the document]
jamestouri: Testing it out report?.measure was valid as a dependency, however I can go ahead and change this to report

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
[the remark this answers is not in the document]
jamestouri: Screenshot 2023-11-07 at 10 27 45 AM

Next commit will have this

packages/react/src/MeasureReportDisplay/MeasureReportDisplay.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { MeasureReportDisplayGroup, MeasureTitle } from './MeasureReportGroup/MeasureReportGroup';
+
+export interface MeasureReportDisplayProps {
+  measureReport: MeasureReport | Reference<MeasureReport>;
+}
+
+export function MeasureReportDisplay(props: MeasureReportDisplayProps): JSX.Element | null {
+  const medplum = useMedplum();
+  const report = useResource(props.measureReport);
+  const [measure, setMeasure] = React.useState<Measure | undefined>();
```
codyebberson, line 14: You might need to test it, but I'm pretty sure you can use the new `useSearchOne` hook here.

Something like:

```tsx
const [measure] = useSearchOne('Measure',  { url: report?.measure });
```

Then you wouldn't need the extra `useState` and `useEffect`

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+interface MeasureGroupProps {
+  readonly group: MeasureReportGroup;
+}
+
+interface MeasureProps {
+  readonly measure: Measure;
+}
+
+export function MeasureReportDisplayGroup(props: MeasureGroupProps): JSX.Element | null {
```
codyebberson, line 15: For consistency purposes, let's use `{resourceType} + 'Display'`, so this would be called `MeasureReportGroupDisplay`.  The component, props, and filename should align.

packages/react/src/MeasureReportDisplay/MeasureReportDisplay.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }, [medplum, report]);
+
+  if (!report) {
+    return null;
+  }
+
+  return (
+    <Box>
+      {measure && <MeasureTitle measure={measure} />}
+      <SimpleGrid cols={3} spacing={'xs'}>
```
codyebberson, line 32: Is there a way to make this responsive for mobile?  3 column makes sense on desktop.  Mobile should probably just be vertically stacked.  I suspect Mantine has a pattern for this.

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
```diff
@@ -0,0 +1,127 @@
...
+function MeasureReportPopulation(props: MeasureGroupProps): JSX.Element {
+  const { group } = props;
+  const populations = group.population;
+  const numerator = populations?.find((p: any) => formatCodeableConcept(p.code) === 'numerator');
+  const denominator = populations?.find((p: any) => formatCodeableConcept(p.code) === 'denominator');
+
+  const numeratorCount = numerator?.count;
+  const denominatorCount = denominator?.count;
+
+  if (!numeratorCount || !denominatorCount) {
```
codyebberson, line 50: Beware the case where `numeratorCount === 0`.  We should explicitly check for `numeratorCount === undefined` here.

The case where `denominatorCount === 0` is slightly different, and may require a different special case?  It's not really "Insufficient Data", but we should guard against divide-by-zero errors.

jamestouri: I added 'Not Applicable'
Screenshot 2023-11-07 at 1 28 07 PM

packages/react/src/MeasureReportDisplay/MeasureReportDisplay.tsx
jamestouri: useMediaQuery will always return false for SSR, so made it default 1 col. Can change it back
Screenshot 2023-11-07 at 1 09 09 PM

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
codyebberson, line 1: Can we rename the folder and file to match the component name? `MeasureReportGroupDisplay/MeasureReportGroupDisplay.tsx`

packages/react/src/MeasureReportDisplay/MeasureReportGroup/MeasureReportGroup.tsx
```diff
@@ -0,0 +1,138 @@
+import { Box, Flex, Group, Paper, RingProgress, Text, Title } from '@mantine/core';
+import { formatCodeableConcept } from '@medplum/core';
+import { Measure, MeasureReportGroup } from '@medplum/fhirtypes';
+import React from 'react';
+import { QuantityDisplay } from '../../QuantityDisplay/QuantityDisplay';
+
+interface MeasureGroupProps {
```
codyebberson, line 7: Can we rename the props interface to match the component name? `MeasureReportGroupDisplayProps`.  I know it's a mouthful, but I think it's "technically correct".

packages/react/src/MeasureReportDisplay/MeasureReportDisplay.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const [measure] = useSearchOne('Measure', { url: report?.measure });
+
+  if (!report) {
+    return null;
+  }
+
+  return (
+    <Box>
+      {measure && <MeasureTitle measure={measure} />}
+      <SimpleGrid cols={!isMobile ? 3 : 1} spacing={'xs'}>
```
codyebberson, line 25: I think the recommended way to do it is to use the Mantine `breakpoints` prop: https://v6.mantine.dev/core/simple-grid/#breakpoints

```tsx
import { SimpleGrid } from '@mantine/core';

function Demo() {
  return (
    <SimpleGrid
      cols={4}
      spacing="lg"
      breakpoints={[
        { maxWidth: '62rem', cols: 3, spacing: 'md' },
        { maxWidth: '48rem', cols: 2, spacing: 'sm' },
        { maxWidth: '36rem', cols: 1, spacing: 'sm' },
      ]}
    >
      <div>1</div>
      <div>2</div>
      <div>3</div>
      <div>4</div>
      <div>5</div>
    </SimpleGrid>
  )
}
```

codyebberson (approved):
Thanks @jamestouri 👍

PR #3244, opened 2023-11-06
Log more details on external auth failure

codyebberson:


mattwiller (approved):

packages/server/src/auth/external.ts
```diff
@@ -187,10 +187,16 @@ async function verifyExternalCode(
       body: params.toString(),
     });
 
-    const tokens = await response.json();
-    return parseJWTPayload(tokens.id_token);
+    const responseBody = await response.json();
+
+    if (!response.ok) {
```
mattlong, line 192: @codyebberson Should this check go above `await response.json()`? As it is, if the response isn't valid JSON, the new error handling would be skipped.

packages/server/src/auth/external.ts
```diff
@@ -187,10 +187,16 @@ async function verifyExternalCode(
...
+      globalLogger.warn('Bad response from external auth check', { status: response.status, body: responseBody });
```
mattlong, line 193: @codyebberson  Related to my comment above, I think `body` should be set to `await response.text()` here.

PR #3249, opened 2023-11-06
Document "Security Best Practices" for Auth

### LINKED ISSUE #2693: Update the Auth Integration Patterns to include Security Best Practices
User driven
https://medplum.slack.com/archives/C05GD70FF55/p1692205551578159?thread_ts=1692205551.578159&cid=C05GD70FF55

Proposed changes should be in this article:
https://www.medplum.com/docs/auth/authentication-integration-patterns

Add a section - "Security Best Practices " with the following proposed:
* All three implementations will have tokens or client credentials with system access and the edge systems should follow best practices.
* Client Authentication
  * Local disk should be encrypted 
  * Workstation should be password locked
  * Consider disabling local storage on device for shared workstations or in accordance with institution policy
* Server Authentication
  * Run with as low level access policy as possible
  * Encrypt disk 
  * Restrict access to host via VPC or other mechanism, do not allow access from general internet
  * Use secrets management, do not store credentials on disk
  * Ensure host is patched and security updates applied
* Host authentication
  * Run with extremely limited access policy, such as read-only access to a specific Bot
  * Ensure host is patched and security updates applied
  * Restrict access to the host, do not allow access from the general internet

rahul1:
- Fix redirect for authentication-integration-patterns
- Add security best practices

Fixes #2693

### DIFF
--- a/packages/docs/docs/auth/methods/index.md
+++ b/packages/docs/docs/auth/methods/index.md
@@ -94,3 +94,27 @@ Device authentication is designed for true machine connectivity, for example con
 For machine connectivity, we encourage use of [ClientCredentials](/docs/auth/methods/client-credentials) with tightly scoped [access controls](/docs/access/access-policies), giving minimal access to the host at the edge.
 
 If OAuth2 client credentials based authentication is not an option due to device limitations, [Basic Authentication](https://www.medplum.com/docs/sdk/core.medplumclient.setbasicauth) can be used to connect to Medplum. In general, OAuth2 is preferred to basic authentication. [Consuming webhooks](/docs/bots/consuming-webhooks) is an example application where Basic Authentication makes sense.
+
+## Security Best Practices
+
+All three implementations types will have tokens or client credentials with system access, and edge systems should follow best practices.
+
+### Client Authentication
+
+- Local disk should be encrypted
+- Workstation should be password locked
+- Consider disabling local storage on device for shared workstations or in accordance with institution policy
+
+### Server Authentication
+
+- Run with as strict an[access policy](/docs/access/access-policies) as possible
+- Encrypt hard disk
+- Restrict access to host via VPC or other mechanism - do not allow access from general internet
+- Use a secrets management to store access keys - do not store credentials on disk
+- Ensure host is patched and has security updates applied regularly
+
+### Host authentication
+
+- Run with extremely an limited [access policy](/docs/access/access-policies). For example: read-only access to a specific Bot
+- Restrict network access to the host - do not allow access from the general internet
+- Ensure host is patched and has security updates applied regularly
--- a/packages/docs/docs/auth/methods/index.md
+++ b/packages/docs/docs/auth/methods/index.md
@@ -97,24 +97,34 @@ If OAuth2 client credentials based authentication is not an option due to device
 
 ## Security Best Practices
 
-All three implementations types will have tokens or client credentials with system access, and edge systems should follow best practices.
+All three implementations types will have tokens or client credentials with system access, and edge systems should follow best practices. Below are examples of security best practices for each implementation type.
+
+:::warning Compliance Note
+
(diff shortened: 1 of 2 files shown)


packages/docs/docs/auth/methods/index.md
```diff
@@ -94,3 +94,27 @@ Device authentication is designed for true machine connectivity, for example con
...
 
 If OAuth2 client credentials based authentication is not an option due to device limitations, [Basic Authentication](https://www.medplum.com/docs/sdk/core.medplumclient.setbasicauth) can be used to connect to Medplum. In general, OAuth2 is preferred to basic authentication. [Consuming webhooks](/docs/bots/consuming-webhooks) is an example application where Basic Authentication makes sense.
+
+## Security Best Practices
+
+All three implementations types will have tokens or client credentials with system access, and edge systems should follow best practices.
+
+### Client Authentication
+
+- Local disk should be encrypted
```
reshmakh, line 104: Mention password requirements, SSO and 2FA for credentials as best practice.

packages/docs/docs/auth/methods/index.md
```diff
@@ -94,3 +94,27 @@ Device authentication is designed for true machine connectivity, for example con
...
+- Workstation should be password locked
+- Consider disabling local storage on device for shared workstations or in accordance with institution policy
```
reshmakh, line 106: Mention organizations with mobile devices or laptops should enable a Mobile Device Management (MDM) solution for workstations

reshmakh: Mention that IP restrictions on an access policy can be considered for especially sensitive scenarios

packages/docs/docs/auth/methods/index.md
codyebberson: Yeah -- this section feels risky.  The most important point here is that Medplum assumes that the user is using a HIPAA compliant device.  We should lean on HIPAA here.  We can list these as "for example" bullet points, but it should be clear that these are just examples, and that your compliance team / legal team / system administrator is responsible for HIPAA compliance and security.

packages/docs/docs/auth/methods/index.md
```diff
@@ -94,3 +94,27 @@ Device authentication is designed for true machine connectivity, for example con
...
+
+### Server Authentication
+
+- Run with as strict an[access policy](/docs/access/access-policies) as possible
```
codyebberson, line 110: Missing a space after "an" ?

Maybe reference [principles of least privileges](https://en.wikipedia.org/wiki/Principle_of_least_privilege)

codyebberson (commented):
There are a bunch of things in here that don't feel like "auth", but generally "security" best practices.  

Anytime we're talking about security, we should be crystal clear that the customer is responsible for HIPAA compliance, and that we are not an authority on whether they are compliant or not.

packages/docs/docs/auth/methods/index.md
```diff
@@ -94,3 +94,37 @@ Device authentication is designed for true machine connectivity, for example con
...
 
 If OAuth2 client credentials based authentication is not an option due to device limitations, [Basic Authentication](https://www.medplum.com/docs/sdk/core.medplumclient.setbasicauth) can be used to connect to Medplum. In general, OAuth2 is preferred to basic authentication. [Consuming webhooks](/docs/bots/consuming-webhooks) is an example application where Basic Authentication makes sense.
+
+## Security Best Practices
+
+All three implementations types will have tokens or client credentials with system access, and edge systems should follow best practices. Below are examples of security best practices for each implementation type.
+
+:::warning Compliance Note
+
+**The practices below are only suggestions**. Please consult your compliance and legal team to ensure that your devices are HIPAA compliant.
```
codyebberson, line 104: 👍

codyebberson (approved):
Thanks!

PR #3251, opened 2023-11-06
Use lookup table for reference search

mattwiller:
In order to make chained search more efficient, this PR replaces the complicated string manipulation to join resource columns with dedicated lookup tables that can match to the correct resource IDs directly.  This change should preserve existing functionality for chained search, but should be tested on Staging before deploying in order to test the new SQL queries at scale


packages/server/src/fhir/lookups/lookuptable.ts
```diff
@@ -140,17 +140,23 @@ export abstract class LookupTable<T> {
...
+    values: Record<string, any>[],
+    ignoreConflict?: boolean
   ): Promise<void> {
     if (values.length === 0) {
       return;
     }
     const tableName = this.getTableName(resourceType);
-    await new InsertQuery(tableName, values).execute(client);
+    const insert = new InsertQuery(tableName, values);
+    if (ignoreConflict) {
```
codyebberson, line 156: What is the case for conflicts?

For better or worse, we currently delete all values before insert, so there should never be a conflict (at least that's my mental model).

packages/server/src/migrations/schema/v53.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Generated by @medplum/generator
+ * Do not edit manually.
+ */
+
+import { getResourceTypes } from '@medplum/core';
+import { PoolClient } from 'pg';
+
+export async function run(client: PoolClient): Promise<void> {
+  for (const resourceType of getResourceTypes()) {
```
codyebberson, line 10: Hm.... We may need to be careful with this.  The return value of `getResourceTypes` could change in the future, which would make this migration non-deterministic.

For all other migrations, we've intentionally duplicated resource types to ensure that migrations do the same thing.

Happy to discuss offline.  Depending on our assumptions about the new migration tooling, this may or not be a big deal.

packages/server/src/fhir/search.ts
```diff
@@ -905,33 +903,33 @@ function buildChainedSearch(selectQuery: SelectQuery, resourceType: string, para
...
-    } else {
-      return new Condition(nextColumn, 'REVERSE_LINK', currentColumn, resourceType);
-    }
-  } else if (link.details.array) {
-    return new ArraySubquery(
-      linkColumn,
-      new Condition(new Column(nextTable, 'id'), 'LINK', new Column(undefined, link.details.columnName))
-    );
+  if (link.reverse) {
+    const lookupJoin = new Condition(new Column(currentTable, 'id'), '=', new Column(lookupAlias, 'targetId'));
```
codyebberson, line 917: Oof.  This is intense.  It looks good, but intense.

packages/server/src/fhir/lookups/lookuptable.ts
mattwiller: This was necessary to make a test pass where an `AuditEvent` had the same `entity` represented multiple times.  This could happen if, for example, a given Practitioner was involved in an activity in multiple roles.  Since the primary key for the new lookup table covers all columns, ignoring on conflict has the same effect as overwriting the row

packages/server/src/fhir/search.ts
mattwiller: Intense probably means it should have more explanation, I'll add some comments!

PR #3252, opened 2023-11-07
process.env replacement broken with Vite aliases

mattlong:
Repro: Logout and go to http://localhost:3000/signin. You'll see `process is not defined`

`process.env.NODE_ENV` apparently gets special handling and continues to work in the `DateTimeInput`; e.g. when specifying specifying the start or end time on http://localhost:3000/Appointment/new.


mattlong:
This was fixed by #3259

Issue #1275, opened 2022-12-10
Implement autosave for the Questionnaire builder


reshmakh:
When building a questionnaire if you click in the wrong place before saving you can lose all your work.  

Autosave once a question is complete would help.

codyebberson:
Fixed in https://github.com/medplum/medplum/pull/6115

PR #3253, opened 2023-11-07
QuestionnaireBuilder Autosave

jamestouri:
https://github.com/medplum/medplum/issues/1275

https://github.com/medplum/medplum/assets/6913745/17b42737-d647-48c9-af78-37e99e0bdc30

https://github.com/medplum/medplum/assets/6913745/1b1607ab-7a3b-4247-ae73-aee2aff69a39

### DIFF
--- a/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.test.tsx
+++ b/packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.test.tsx
@@ -123,6 +123,33 @@ describe('QuestionnaireBuilder', () => {
     expect(onSubmit).toBeCalled();
   });
 
+  test('Handles AutoSave', async () => {
+    const onSubmit = jest.fn();
+
+    await setup({
+      questionnaire: {
+        resourceType: 'Questionnaire',
+        item: [
+          {
+            linkId: 'q1',
+            type: QuestionnaireItemType.string,
+            text: 'q1',
+          },
+        ],
+      },
+      onSubmit,
+      autoSave: true,
+    });
+
+    expect(screen.getByText('Save')).toBeDefined();
+
+    await act(async () => {
+      fireEvent.click(screen.getByText('Save'));
+    });
+
+    expect(onSubmit).toBeCalled();
+  });
+
   test('Sets ids', async () => {
     const onSubmit = jest.fn();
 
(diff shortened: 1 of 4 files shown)


packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -406,12 +423,15 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
...
-                  text: 'Group',
-                } as QuestionnaireItem);
+                addItem(
+                  {
+                    id: generateId(),
+                    linkId: generateLinkId('g'),
+                    type: 'group',
+                    text: 'Group',
+                  } as QuestionnaireItem,
+                  true
```
jamestouri, line 433: For adding Groups and Pages, there is an error for no nested items. I disabled the AutoSubmit for these two specifically
Screenshot 2023-11-06 at 4 12 53 PM

packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -272,15 +289,15 @@ function ItemBuilder<T extends Questionnaire | QuestionnaireItem>(props: ItemBui
               <TextInput
                 size="xl"
                 defaultValue={resource.title}
-                onChange={(e) => changeProperty('title', e.currentTarget.value)}
+                onChange={(e) => changeProperty('title', e.currentTarget.value, true)}
```
jamestouri, line 292: To avoid a 429, I disabled the autosubmit for text changes

mattwiller: Rather than disabling it for text, maybe `changeProperty()` could run on text input blur instead of each change?  Or, even better, the form submission should be [debounced](https://www.freecodecamp.org/news/javascript-debounce-example/) so changes get batched together

packages/react/src/QuestionnaireBuilder/QuestionnaireBuilder.test.tsx
```diff
@@ -189,6 +231,7 @@ describe('QuestionnaireBuilder', () => {
         target: { value: 'Renamed' },
       });
     });
+    fireEvent.blur(screen.getByDisplayValue('Renamed'));
```
jamestouri, line 234: Switched the textinputs from onChange to onBlur. Tested it in the medplum app and storybook. Didn't seem to change the user experience. Change change it back if needed

codyebberson (approved):

PR #3259, opened 2023-11-07
Convert @medplum/react into a module

mattlong:
Replaces #3252 

After [a recent change](https://github.com/medplum/medplum/pull/3229) to how `@medplum/app` works in dev mode, i.e. `npm run dev`, developers will encounter a `process is not defined` error on certain pages, e.g. http://localhost:3000/signin, since the special handling of environment variables at build time is no longer happening. To resolve this we convert `@medplum/react` into a module to more fully take advantage of [Vite's tooling](https://main.vitejs.dev/guide/env-and-mode.html#env-variables) for env vars via the `import.meta` namespace.


packages/react/babel.config.cjs
codyebberson, line 1: Do you mind adding this file to `sonar.exclusions` in `sonar-project.properties`?

codyebberson (approved):
Nice.  Tested and verified on localhost.  Thanks @mattlong 

If not too much trouble, could you add `babel.config.cjs` to sonar exclusions?  That will reduce noise in Sonar, and fix the coverage warnings.

Thanks!

PR #3263, opened 2023-11-07
Make @medplum/core a module

mattlong:


packages/core/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "module": "es2020",
+    "types": ["jest", "vite/client"],
```
mattlong, line 5: A couple things here:

It feels a bit weird to specify `"vite/client"` in core; but it avoids red squigglies for `import.meta.env`. It might be preferable to define our own interface instead as discussed in https://stackoverflow.com/questions/66039933/typescript-types-for-import-meta-env

I added `"jest"` here to resolve red squigglies in `.test.ts` files in core; I couldn't figure out why that wasn't necessary in `@medplum/react` or in any other packages in the repo. Anyone have any ideas on that?

codyebberson: > It feels a bit weird to specify "vite/client" in core; but it avoids red squigglies for

Fine by me

>  I couldn't figure out why that wasn't necessary in @medplum/react or in any other packages in the repo.

🤔  I'm not 100% sure, but I'd guess it's some of the setup code in `packages/react/src/test.setup.ts`

In particular:

```ts
import '@testing-library/jest-dom';
```

ThatOneBro: Adding `vite/client` to types is fine, however you can also do it via a `reference` in a `.d.ts` per the `vite-env.d.ts` convention:

https://github.com/medplum/medplum/blob/d5e2d61db792ad476c5de3b484360c10059a42c2/examples/medplum-fhircast-demo/src/vite-env.d.ts#L1

codyebberson: I would not be opposed to adding `"types": ["jest", "vite/client"],` to the base `tsconfig.json` as well.

mattlong: I'm leaning towards leaving it as is for now (specified in `packages/core/tsconfig.json`) since `import.meta.env` should only be used on modules.

codyebberson (approved):
LGTM (other than the merge conflict)

For what it's worth, I tried adding the `@medplum/core` alias to `packages/app/vite.config.ts`, and everything appears to work.  It was an awe-inspiring moment to make a change in `packages/core` and have the app auto-refresh.  I would be in favor of adding that.

codyebberson (approved):

PR #3279, opened 2023-11-09
feat(auth/external): make `PKCE` on external auth optional

### LINKED ISSUE #3266: Add option to disable PKCE on signInWithExternalAuth
Right now, our authentication methods, including signInWithExternalAuth, add a `code_challenge` and `code_challenge_method` to enable PKCE authentication. 

Some 3rd party identity providers, such as Azure SSO, decide whether or not to enable PKCE authentication based on whether the client initiates in their call to `/authorize`

To allow clients to _disable_ PKCE for these IDPs, we should allow them to disable sending PKCE params in the `signInWithExternalAuth` calls

ThatOneBro:
## What this PR does
* Adds optional last param  `pkceEnabled` to `signInWithExternalAuth`, defaulting to `true`
* Corrects some tests that were asserting `toBeDefined` against values which could only be `string | null`
* Closes #3266

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -362,6 +362,31 @@ describe('Client', () => {
     );
     expect(result).toBeUndefined();
     expect(assign).toBeCalledWith(expect.stringMatching(/authorize\?.+scope=/));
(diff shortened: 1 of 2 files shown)


codyebberson (approved):
Nice 👍

PR #3289, opened 2023-11-09
feat(server/fhircast): split `STU2` and `STU3` routes

### LINKED ISSUE #3276: Split routes for `FHIRcast` into `STU2+` and `STU3`
We should aim to support both `STU3` and a `PowerCast`-compatible `STU2`. There are some conflicts between the two but this can be managed by creating separate routers, exporting the same routes, only with the `STU2` router overriding `STU3` routes where the `PowerCast` implementation or `STU2` spec conflicts with `STU3`. These will be hosted on `/fhircast/STU2` and `/fhircast/STU3` respectively.

ThatOneBro:
## What this PR does
* Splits the routes for `FHIRcast` into `STU2` and `STU3`
* Adds a `protectedCommonRoutes` router, most routes can go here and will be available at both `STU2` and `STU3` base routes.
* Modifies tests to test for both `STU2` and `STU3` routes
* Closes #3276

### DIFF
--- a/packages/server/src/fhircast/routes.test.ts
+++ b/packages/server/src/fhircast/routes.test.ts
@@ -1,4 +1,10 @@
-import { ContentType, FhircastEventContext, FhircastEventPayload, createFhircastMessagePayload } from '@medplum/core';
+import {
+  ContentType,
+  FhircastEventContext,
+  FhircastEventPayload,
+  createFhircastMessagePayload,
+  generateId,
+} from '@medplum/core';
 import { randomUUID } from 'crypto';
 import express from 'express';
 import request from 'supertest';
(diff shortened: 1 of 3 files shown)


packages/server/src/fhircast/routes.test.ts
```diff
@@ -33,146 +53,197 @@ describe('FHIRCast routes', () => {
   });
 
   test('New subscription success', async () => {
-    const res = await request(app)
-      .post('/fhircast/STU3/')
-      .set('Content-Type', ContentType.JSON)
-      .set('Authorization', 'Bearer ' + accessToken)
-      .send({
+    for (const route of [STU2_BASE_ROUTE, STU3_BASE_ROUTE]) {
```
codyebberson, line 56: You can use `test.each` if you like.  It's mostly cosmetic, but jest will then report the test outcomes separately rather than a single pass/fail.

codyebberson (approved):

ThatOneBro:
Merged in with #3292 instead

PR #3294, opened 2023-11-09
Redirect to next from signin page if already authenticated

### LINKED ISSUE #3282: Automatically redirect from signin page if authenticated
If a user is on the Sign in page and is authenticated (e.g. they re-authenticated on a different tab), reloading the sign in page leaves them on the Sign in page. Instead, it'd be convenient if the user is automatically redirected to the path in the `next` query param if present or to the homepage if `next` is not specified.

mattlong:
Fixes #3282 

While I was there, I added a bit of validation of the `next` param to avoid maliciously-crafted redirect URLs.

### DIFF
--- a/packages/app/src/SignInPage.tsx
+++ b/packages/app/src/SignInPage.tsx
@@ -1,17 +1,31 @@
 import { Title } from '@mantine/core';
-import { Logo, SignInForm } from '@medplum/react';
-import React from 'react';
+import { Logo, SignInForm, useMedplum } from '@medplum/react';
+import React, { useCallback, useEffect } from 'react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 import { getConfig, isRegisterEnabled } from './config';
 
 export function SignInPage(): JSX.Element {
+  const medplum = useMedplum();
   const navigate = useNavigate();
   const [searchParams] = useSearchParams();
   const config = getConfig();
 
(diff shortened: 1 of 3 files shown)


mattwiller (commented):
Is it possible to add a test, for at least the security fix?  I'd love to ensure we don't regress on that front

packages/app/src/SignInPage.tsx
```diff
@@ -1,17 +1,31 @@
...
 
+  const navigateToNext = useCallback(() => {
+    // only redirect to next if it is a pathname to avoid redirecting
+    // to a maliciously crafted URL, e.g. /signin?next=https%3A%2F%2Fevil.com
+    const nextUrl = searchParams.get('next');
+    navigate(nextUrl?.startsWith('/') ? nextUrl : '/');
+  }, [searchParams, navigate]);
+
+  useEffect(() => {
+    if (medplum.getProfile()) {
```
codyebberson, line 21: We should only do this if `next` is present.

The use case to keep in mind is "Add another account".  If the user is already signed-in, they can use the top-right menu and click "Add another account".  That will navigate to "/signin" to add another session.

mattlong: @codyebberson mind taking another look?

codyebberson (approved):

PR #3297, opened 2023-11-09
Don't show copy button for ID if its empty/missing

mattlong:
Reduces visual clutter when you'd just be copying a value of `undefined`

**Before**
Screenshot 2023-11-09 at 1 38 49 PM

**After**
Screenshot 2023-11-09 at 1 39 16 PM


mattwiller (approved):
This could also use a test, if it's easy to do

codyebberson (approved):

PR #3305, opened 2023-11-10
NextAuth Example

awamser:


rahul1:
Thanks for the contribution @awamser! I think the reason this is failing is because the package.lock is not updated. I think if you run `npm i` from root, it should update for you. 

To make sure the diff is nice and tight, make sure:
- You're on a current version of NPM
- clear out your `node_modules` before running, to avoid any weirdness

PR #3307, opened 2023-11-11
feat(expo): add `ExpoClientStorage` for persisted store in RN

### LINKED ISSUE #3301: Fix auth not persisting in `React Native`

ThatOneBro:
Needs some cleanup but mostly done. Working well on Android device. Still need to add a few tests too.

Fixes #3301

## TODO:
- [x] Add tests
- [x] Change error to `OperationOutcomeError`
- [x] Clean up `.catch`s
- [x] Fix repeated call to `removeKey` invoking several calls to `setKeys` in `clear`
- [x] Test `MedplumProvider` setting and unsetting `loading`
- [x] Add docs for how to await `initialized` and what it means
- [x] Add docs for how to use the `loading` state from `useMedplumContext`


packages/expo-polyfills/package.json
```diff
@@ -30,6 +30,7 @@
   "dependencies": {
     "base-64": "1.0.0",
     "expo-crypto": "12.6.0",
+    "expo-secure-store": "12.3.1",
```
rhys-jmc, line 33: im wondering if this (and maybe some/all the other dependencies) should be peer deps? this version will quickly become desynced with expo's ~quarterly releases.

packages/core/src/client.ts
```diff
@@ -692,17 +693,36 @@ export class MedplumClient extends EventTarget {
...
-        this.refreshProfile().catch(console.log);
-      }
+      this.attemptResumeActiveLogin().catch(console.error);
+      this.initPromise = Promise.resolve();
     }
 
     this.setupStorageListener();
   }
 
+  get initialized(): Promise<void> {
```
codyebberson, line 713: Oof.  I understand the need for this, but we need a better name and heavy docs.

Can we make it a normal method called `getInitPromise()`?

packages/expo-polyfills/src/index.test.ts
```diff
@@ -1,36 +1,39 @@
...
-  setupURLPolyfill: () => {
-    // @ts-expect-error Just for testing
-    window.URL = NodeURL;
-    // @ts-expect-error Just for testing
-    window.URLSearchParams = NodeURLSearchParams;
-  },
-}));
+import { ExpoClientStorage, polyfillMedplumWebAPIs } from '.';
+
+// Commented out because test this mock was designed for is not working properly
```
codyebberson, line 3: Let's either fix the test or delete the mock.

packages/react-hooks/src/MedplumProvider/MedplumProvider.tsx
```diff
@@ -23,9 +23,21 @@ export function MedplumProvider(props: MedplumProviderProps): JSX.Element {
 
   const [state, setState] = useState({
     profile: medplum.getProfile(),
-    loading: false,
+    loading: true,
   });
 
+  useEffect(() => {
+    medplum.initialized
```
codyebberson, line 30: Remove curly braces

```ts
    medplum.getInitPromise()
      .then(() => setState((s) => ({ ...s, loading: false })))
      .catch(console.error);
```

packages/react-hooks/src/MedplumProvider/MedplumProvider.tsx
```diff
@@ -23,9 +23,21 @@ export function MedplumProvider(props: MedplumProviderProps): JSX.Element {
...
+      .then(() => {
+        setState((s) => ({ ...s, loading: false }));
+      })
+      .catch((err) => console.error(err));
+
+    return () => {
```
codyebberson, line 36: Remove curly braces

```ts
return () => setState((s) => ({ ...s, loading: true }));
```

examples/medplum-react-native-example/src/Home.tsx
```diff
@@ -1,23 +1,24 @@
...
-  const [profile, setProfile] = useState<ProfileResource | undefined>(undefined);
 
   function startLogin(): void {
     medplum.startLogin({ email, password }).then(handleAuthResponse).catch(console.error);
   }
 
   function handleAuthResponse(response: LoginAuthenticationResponse): void {
     if (response.code) {
-      handleCode(response.code);
+      medplum.processCode(response.code).catch((err) => console.error(err));
```
codyebberson, line 21: `.catch(console.error)`

packages/expo-polyfills/package.json
```diff
@@ -42,6 +43,7 @@
     "esbuild": "0.19.5",
     "esbuild-node-externals": "1.9.0",
     "jest": "29.7.0",
+    "jest-expo": "^49.0.0",
```
codyebberson, line 46: Why `jest-expo`?  I don't see any calls to it.  And it's big :(

packages/core/src/client.ts
```diff
@@ -692,17 +693,36 @@ export class MedplumClient extends EventTarget {
 
     if (options?.accessToken) {
       this.setAccessToken(options.accessToken);
+      this.initPromise = Promise.resolve();
+    } else if ((this.storage as AsyncBackedClientStorage).initialized !== undefined) {
```
codyebberson, line 697: Structurally, this is clever, and I think it nicely solves a really tricky problem.

This was already a pretty gnarly hot spot in MedplumClient, full of yucky complexity, so if it's ok with you I'd like us to be extra thoughtful about naming and API design.

Given that `storage` is already a custom Medplum type, how would you feel about making `getInitPromise()` an optional method on the existing interface?  Then we wouldn't need the extra type, the extra `as`, etc.

codyebberson (commented):
Clever solution to a hard problem.  Looks good from structural perspective.  A bunch of nitpicky style comments.

packages/expo-polyfills/package.json
ThatOneBro: It's the test environment for Expo, it actually fixes a lot of problems and better represents the various environments for `Android`, `iOS`, and `Web` as separate environments. See the `jest.config.ts`: 
https://github.com/medplum/medplum/blob/2c3fe80d47b96bf23806bcaa236c3d6f9d7d4d28/packages/expo-polyfills/jest.config.json#L2

packages/core/src/client.ts
ThatOneBro: I think optional methods are kind of weird for interfaces, having type hints for whether or not a method actually exists on a class is pretty good when possible IMO.

This condition is the one case where there is ambiguity since we aren't sure if the user passed in an `AsyncBackedClientStorage` or just a normal `ClientStorage`, but in all other circumstances it would be good to have proper typing on a class creating implementing `AsyncBackedClientStorage` vs just normal `ClientStorage` which should not have the method.

packages/expo-polyfills/src/index.ts
```diff
@@ -79,3 +80,162 @@ export function polyfillMedplumWebAPIs(): void {
     });
   }
 }
+
+class SyncSecureStorage implements Storage {
+  private readonly storage: Map<string, string>;
+  private readonly initPromise: Promise<void>;
+  private isInitialized = false;
```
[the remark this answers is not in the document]
ThatOneBro: True,

In this case was trying to avoid collision with `initialized` Promise on the object... but yeah you're 100% right

packages/core/src/client.ts
ThatOneBro: Yes 100%, was planning to document before merging.

I also think I understand your fear of having a `Promise` on the object rather than just calling a getter function. I've seen this pattern using getters more these days but I guess it still may be unintuitive to await a field on an object for some users.

packages/core/src/client.ts
```diff
@@ -692,17 +693,36 @@ export class MedplumClient extends EventTarget {
...
     }
 
     this.setupStorageListener();
   }
 
+  get initialized(): Promise<void> {
+    return this.initPromise;
+  }
+
+  async attemptResumeActiveLogin(): Promise<void> {
```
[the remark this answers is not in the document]
ThatOneBro: Good catch

ThatOneBro:
After the introduction of the `ExpoClientStorage`, this should be the minimal example of using `MedplumClient` in React Native.

If you want to wait to load components until after the `MedplumClient` has initialized, you can conditionally render based on the `loading` property from the `useMedplumContext` hook, like so:

```tsx
import { MedplumClient } from '@medplum/core';
import { MedplumProvider, useMedplumContext } from '@medplum/react-hooks';
import { polyfillMedplumWebAPIs, ExpoClientStorage } from '@medplum/expo-polyfills';

polyfillMedplumWebAPIs();

const medplum = new MedplumClient({ storage: new ExpoClientStorage() });

function Home(): JSX.Element {
  const { loading } = useMedplumContext();
  return loading ? <div>Loading...</div> : <div>Loaded!</div>;
}

function App(): JSX.Element {
  return (
    <MedplumProvider medplum={medplum}>
      <Home />
    </MedplumProvider>
  );
}
```

Any thoughts @codyebberson @rhys-jmc ?

codyebberson (approved):

packages/expo-polyfills/src/index.ts
```diff
@@ -79,3 +81,187 @@ export function polyfillMedplumWebAPIs(): void {
...
+ * This class is necessary for use with `MedplumClient` since `MedplumClient` expects `ClientStorage` to use a synchronous interface, and `SecureStore` uses an asynchronous one.
+ *
+ * On web, this class will instead wrap `localStorage` automatically, so there is no need to conditionally omit passing this to `MedplumClient` for React Native Web.
+ */
+export class ExpoClientStorage extends ClientStorage implements IExpoClientStorage {
+  // We keep a private reference to the storage we pass in so we can use it within this subclass too...
+  private secureStorage?: SyncSecureStorage;
+  constructor() {
+    // Metro should automatically prune these branches out at compile time
+    if (Platform.OS === 'web') {
```
ThatOneBro, line 245: Does this seem like a sane way to handle the `Web` case? These checks should compile out when `Metro` compiles the final app. @codyebberson

codyebberson: Seems reasonable - it's exported by Expo, so feels official.

codyebberson (approved):
Nice, thanks @ThatOneBro

PR #3308, opened 2023-11-11
fix-3240 rate-limit-redis persistent rate limit storage

### LINKED ISSUE #3240: redis-based store for rate limiter
https://github.com/medplum/medplum/blob/fbf271ba1389e850d62d154c466cf4a3cda4b173/packages/server/src/ratelimit.ts#L12

The rate limiter does not behave properly in a multi-node/replica environment due to being bound to the processes memory. Additionally, in the event of a process restart, all rate limits will be reset. I'm uncertain whether medplum has encountered the multi-replica use case, but moving the store out of the server process memory is worth considering nonetheless.

[`rate-limit-redis`](https://www.npmjs.com/package/rate-limit-redis) could be easy to adopt but I believe the routers will need to be lazily evaluated since simply importing the routers calls `getRateLimiter` and the redis connection will not yet be established.
https://github.com/medplum/medplum/blob/fbf271ba1389e850d62d154c466cf4a3cda4b173/packages/server/src/oauth/routes.ts#L11
Lazily evaluating should be easy enough by wrapping the router creation in a function call.

dillonstreator:
fix #3240


dillonstreator:
closing in favor of https://github.com/medplum/medplum/pull/4435

PR #3310, opened 2023-11-12
fix-3309 secrets editor textarea value with conceal/reveal

### LINKED ISSUE #2863: Secrets Editor strips out newlines
This breaks in .pem files, invalidating them and they fail createPrivateKey.

### LINKED ISSUE #3309: Secrets Editor treat secret as sensitive and conceal by default
The Project Secrets Editor should hide secrets. This is likely only relevant for string secrets. [Matine's PasswordInput](https://mantine.dev/core/password-input/) would be more appropriate but this doesn't work well with the issue related to capturing newlines https://github.com/medplum/medplum/issues/2863. A default blurred text-area with optional reveal/conceal could address both issues.

dillonstreator:
fix #3309
fix #2863

Option 1 with custom PasswordDots font:

https://github.com/medplum/medplum/assets/26043866/dea86039-7c77-49ef-84c2-1bc7aa0df54f

Option 2 with [`-webkit-text-security`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security)

medplum-sensitive-textarea-text-security


codyebberson:
Thanks @dillonstreator!  Did you by chance come across any implementation options that don't require additional fonts?  I'm reluctant to add `woff` files as a dependency.

dillonstreator:
@codyebberson the only other option that I came across involved blurring the font with a box-shadow and this was rendered useless in the conceal state when highlighting/selecting the text. Surprisingly, I did not see other similar use cases that are documented on the web. Ideally, `type="password"` would be natively supported by `textarea`.

Out of curiosity, is the issue you have with the introducing of the `woff` file that it is (arguably) an odd dependency of `@medplum/react`? FWIW, I added an [installation doc note in the README](https://github.com/medplum/medplum/pull/3310/files#diff-3dae6d110b675ef311b837cf10653c25945e787593bfc47d7bdbc3836df15abbR31) of the package.

dillonstreator:
@codyebberson - I actually just came across another approach that would not require a new font. There's a CSS property called [`-webkit-text-security`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security).

Browser support[[1](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security#browser_compatibility), [2](https://caniuse.com/mdn-css_properties_-webkit-text-security)] seems reasonably strong but this is a non-standard property so would understand hesitation to adopt.

Curious - are there any contribution guidelines regarding browser support requirements or is it handled on more of a case-by-case basis? If this were adopted, where (if at all) should this be documented? I imagine y'all have analytics on existing customer browsers and versions which could help make a data-driven decision on the approach to take.

codyebberson:
Thanks @dillonstreator !

Excellent find with "webkit-text-security"

https://github.com/medplum/medplum/pull/3642

PR #3316, opened 2023-11-14
fix request logging middleware status code reporting

dillonstreator:
With the `logRequests` config enabled, the request logging middleware reports 200 status code regardless of non-200 status codes being sent to the client.

[`res.on('finish', () => {})`](https://nodejs.org/api/http.html#event-finish_1) should be used to properly hook into response being sent and report the actual status code.

❗ Note this PR also changes the log property `requestMethod` -> `method`. @mattwiller called this out in his comment https://github.com/medplum/medplum/pull/3196#discussion_r1379415634 on the PR introducing this logging middleware. At this point, it would technically be considered a breaking change so I'd understand if this is not desired and is preferred to be kept as `requestMethod`.

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -214,22 +214,26 @@ export async function shutdownApp(): Promise<void> {
 const loggingMiddleware = (req: Request, res: Response, next: NextFunction): void => {
   const ctx = getRequestContext();
   const start = new Date();
-  next();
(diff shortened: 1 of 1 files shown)


mattwiller (approved):

mattwiller:
Thanks @dillonstreator! ✨

PR #3325, opened 2023-11-15
Create reference lookup tables

mattwiller:
Creating lookup tables for references between resources, to be populated via a data migration that performs a full reindex of all resource types


packages/server/src/fhir/lookups/reference.ts
```diff
@@ -0,0 +1,118 @@
...
+    const values = getSearchReferences(resource);
+    const existing = await getExistingValues(client, this.getTableName(resource.resourceType), resource.id as string);
+    if (compareArrays(values, existing)) {
+      // Nothing changed
+      return;
+    }
+    if (existing.length > 0) {
+      await this.deleteValuesForResource(client, resource);
+    }
+    await this.insertValuesForResource(client, resource.resourceType, values, true);
```
codyebberson, line 40: If we call `deleteValuesForResource`, do we still need to pass `true` for `ignoreConflict`?  Did you find a case where the `ignoreConflict` option was necessary?

packages/server/src/fhir/sql.ts
```diff
@@ -572,15 +572,21 @@ export class ArraySubquery implements Expression {
 
 export class InsertQuery extends BaseQuery {
   private readonly values: Record<string, any>[];
-  private merge?: boolean;
+  private onConflict: 'merge' | 'ignore' | 'error';
```
codyebberson, line 575: Can you help me understand the case where `ignore` was necessary?

packages/server/src/fhir/lookups/reference.ts
mattwiller: I think the context got lost when I split this out from the original PR: https://github.com/medplum/medplum/pull/3251#discussion_r1387011984

codyebberson: Hm - Ok, I think we should fix that either inside `insertValuesForResource` or elsewhere when building the list.  We should use a `Set<string>` to ensure there are no duplicate values.

PR #3327, opened 2023-11-15
Make Postgres version configurable in CDK

mattwiller:


packages/core/src/config.ts
```diff
@@ -100,6 +100,7 @@ export interface MedplumInfraConfig {
   rdsInstances: number;
   rdsInstanceType: string;
   rdsSecretsArn?: string;
+  postgresVersion?: string;
```
ThatOneBro, line 103: Can we add this to `MedplumSourceInfraConfig` as well?

```ts
postgresVersion?: ValueOrExternalSecret<string>;
```

ThatOneBro (changes requested):

packages/core/src/config.ts
codyebberson: How would you feel about `rdsInstanceVersion` for naming consistency?

I know this is used inside of `rds.DatabaseClusterEngine.auroraPostgres`, so it currently is only used for Postgres.  If we ever supported other RDS databases in the future, then this name becomes wrong or ambiguous.

ThatOneBro (approved):
Nice 👍

PR #3329, opened 2023-11-16
test(server): add `test.config.json` with port `8104`

### LINKED ISSUE #3326: Use a separate default port for running `@medplum/server` tests
Currently if you run tests and there is already a dev server running on the default port, the tests will fail, but only after partially running for usually several minutes.

ThatOneBro:
Fixes #3326


mattwiller (approved):
Awesome QOL fix, this bites me all the time

PR #3330, opened 2023-11-16
Add Project $init operation

### LINKED ISSUE #2699: Move Project initialization logic to `Project/$init` operation
User Driven:
-  https://medplum.slack.com/archives/C05C5KXTPJA/p1691688476394129
- https://discord.com/channels/905144809105260605/1141751558242050069/1141751558242050069

We've seen a number of use cases for self-hosted users, where they would like to programatically spin up new projects from each of their tenants. 

Right now the MedplumClient only has the `startNewProject` method, which is tightly coupled to the UI flow. 

**Proposal:**
Have an endpoint / SDK method to create a project _without_ requiring a partial loginState.

mattwiller:
Adding a new `/Project/$init` endpoint to centralize Project creation logic.  This should probably replace `/auth/newproject`

Resolves #2699


packages/server/src/fhir/operations/projectinit.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      .send({
+        resourceType: 'Parameters',
+        parameter: [
+          {
+            name: 'name',
+            valueString: projectName,
+          },
+          {
+            name: 'owner',
+            valueString: createReference(owner).reference,
```
codyebberson, line 138: Thoughts on making this `valueReference: createReference(owner)`?  Might be more FHIR-y?

packages/server/src/fhir/operations/projectinit.ts
```diff
@@ -0,0 +1,116 @@
...
+import { sendResponse } from '../routes';
+import { createClient } from '../../admin/client';
+import { createProfile, createProjectMembership } from '../../auth/utils';
+import { getUserByEmailWithoutProject } from '../../oauth/utils';
+import { createUser } from '../../auth/newuser';
+import { randomUUID } from 'crypto';
+
+interface ProjectInitParameters {
+  name: string;
+  owner?: string;
```
codyebberson, line 17: Maybe make this `owner?: Reference`?

codyebberson (approved):
Nice, this looks great, thanks @mattwiller  - this is long overdue.

One minor suggestion to change `owner` from `string` to `Reference`.  It's a little extra complexity, but it might make things more type-friendly.  I defer to you, your call.

mattwiller:
> One minor suggestion to change `owner` from `string` to `Reference`

I was strongly considering this myself, but wasn't sure about extending the simple parameters parser utility (which currently only works with `valueString`).  What I'd really like to have here is something that can look at an `OperationDefinition` and parse the parameters from that; this approach would make it easier to know which type the different params should be.  We'd also probably want to have the final output of the params parser be e.g. `Record<string, TypedValue>`.  I can take a stab at getting that set up!

mattwiller:
@codyebberson Some more food for thought: if we use `owner: Reference`, would we want to handle the email case with `Reference.identifier`?  For example:
```json
"owner": {
  "identifier": {
    "system": "urn:ietf:rfc:3986",
    "value": "mailto:owner.email@example.com"
  }
}
```

codyebberson:
> but wasn't sure about extending the simple parameters parser utility (which currently only works with `valueString`).

Ah, got it.  I think extending it to support any `value[x]` would be fine.

> We'd also probably want to have the final output of the params parser be e.g. `Record<string, TypedValue>`

The goal of the `parseParameters()` util was to convert the input into an ergonomic type for clean/easy usage.  It has a generic `<T>` type param, so the return value can be a type that is specialized for the operation.  (IMHO, I don't consider `TypedValue` ergonomic - it's kind of annoying to use 🥲)

I.e., I'd rather be able to use `params.owner` rather than `params.owner.value`, especially for the common case such as `owner` which is not a `[x]` choice-of-type.

>  if we use owner: `Reference`, would we want to handle the email case with `Reference.identifier`?

Good question...  I like that.  It probably makes sense to handle that inside `Repository.readReference`, yeah?

In that case, I'd recommend making that a separate ticket.

Ideally:

- Move `MemoryRepository.readReference` and server `Repository.readReference` into `BaseRepository` (in `packages/fhir-router/src/repo.ts`) - they have the same implementation right now anyway
- Detect the `Reference.identifier` case, and use `searchOne`

mattwiller:
@codyebberson I ended up not really liking the proposed usage of `Reference` in the end; it wasn't exactly a clean match to the way logical references are usually supposed to be handled.  Instead, I invested in creating utilities to streamline implementation of future FHIR operations by parsing input and sending output `Parameters` directly from the `OperationDefinition` specification.  Let me know if you have any thoughts on those interfaces!

packages/server/src/auth/routes.ts
```diff
@@ -19,23 +19,78 @@ import { revokeHandler, revokeValidator } from './revoke';
...
 authRouter.post('/changepassword', authenticateRequest, changePasswordValidator, asyncWrap(changePasswordHandler));
-authRouter.post('/resetpassword', resetPasswordValidator, asyncWrap(resetPasswordHandler));
+authRouter.post('/resetpassword', resetPasswordValidator, validateRecaptcha(), asyncWrap(resetPasswordHandler));
 authRouter.post('/setpassword', setPasswordValidator, asyncWrap(setPasswordHandler));
 authRouter.post('/google', googleValidator, asyncWrap(googleHandler));
 authRouter.post('/exchange', exchangeValidator, asyncWrap(exchangeHandler));
 authRouter.post('/revoke', authenticateRequest, revokeValidator, asyncWrap(revokeHandler));
 authRouter.get('/login/:login', statusValidator, asyncWrap(statusHandler));
+
+function validateRecaptcha(projectValidation?: (p: Project) => OperationOutcome | undefined): Handler {
```
codyebberson, line 48: Might make sense to move this into a separate util file.

codyebberson (approved):
Nice, structurally looks good to me.  This will be really helpful.

1. One minor nit comment about util file - feel free to ignore
2. Thoughts on documenting this?
    - We have a rough set of docs on Medplum FHIR Operations here: https://www.medplum.com/docs/api/fhir/operations
    - Maybe add a quick example doc here? `packages/server/src/fhir/operations`

PR #3331, opened 2023-11-16
Fix patient search parameter target type

mattwiller:
For some reason, the search parameter in the JSON file is out of sync with what's shown at http://hl7.org/fhir/R4/observation.html

Based on the `expression` for the search param, it's impossible for this parameter to point to a `Group`


ThatOneBro:
For context:
https://hl7.org/fhir/R4/observation.html#:~:text=patient%20TU,(Patient)

ThatOneBro (approved):

rahul1 (approved):

PR #3334, opened 2023-11-17
Tighter input validation on security endpoints

codyebberson:


packages/server/src/auth/newuser.ts
```diff
@@ -9,14 +9,17 @@ import { sendOutcome } from '../fhir/outcomes';
...
 import { makeValidationMiddleware } from '../util/validator';
+import { bcryptHashPassword, getProjectByRecaptchaSiteKey, verifyRecaptcha } from './utils';
 
 export const newUserValidator = makeValidationMiddleware([
-  body('firstName').notEmpty().withMessage('First name is required'),
-  body('lastName').notEmpty().withMessage('Last name is required'),
-  body('email').isEmail().withMessage('Valid email address is required'),
-  body('password').isLength({ min: 8 }).withMessage('Password must be at least 8 characters'),
+  body('firstName').isLength({ min: 1, max: 72 }).withMessage('First name must be between 8 and 72 characters'),
+  body('lastName').isLength({ min: 1, max: 72 }).withMessage('Last name must be between 8 and 72 characters'),
```
mattwiller, line 17: The error messages are out of sync with the actual limits here, but also these limits are probably not needed, since these inputs are not hashed

packages/server/src/auth/newuser.ts
```diff
@@ -9,14 +9,17 @@ import { sendOutcome } from '../fhir/outcomes';
...
+  body('email')
+    .isEmail()
+    .isLength({ min: 3, max: 72 })
+    .withMessage('Email must be valid email address between 3 and 72 characters'),
+  body('password').isLength({ min: 8, max: 72 }).withMessage('Password must be between 8 and 72 characters'),
```
ThatOneBro, line 22: We should be wary of the non-ASCII case (ie. non-English), this will break when passwords are 2 or 4 byte character glyphs

Not sure how we want to deal with that yet, but maybe this would be a good thing to consider

packages/server/src/auth/newuser.ts
codyebberson: Good catch on the numeric mismatch, will fix.

On whether the limits are actually required.... We received a bug report for DDoS risk due to open storage of these fields, on unauthenticated endpoints.  The reporter sent a video which uploaded a 1 MB text file in the name field.  It's a reasonable callout.

packages/server/src/auth/newuser.ts
codyebberson: Yeah, this is a fair callout.  I wouldn't say it "breaks", but the behavior does get weird.

image

In general, I'm comfortable with that.  If users start sending emojis or special characters, then the counts could get weird.

I would not want to block non-ASCII passwords, because that's kinda unfriendly to international users.

ThatOneBro: It breaks in the sense that it truncates the password, which I would think is more than just weird behavior. If the max was 25 let's say, and the user entered a password in Russian like this:
`"здравствуйте, Медплум"`

This is what happens:

Screenshot 2023-11-18 at 1 55 21 PM

Assuming that `bcrypt` truncated at 25 instead of 72, this is what the password would actually be in practice:
`"здравствуйте"`

Because of this:

Screenshot 2023-11-18 at 1 55 46 PM

Which leads to this becoming the de facto password as `bcrypt` sees it:

Screenshot 2023-11-18 at 1 58 05 PM


Which means the password would actually be much shorter than what the user is expecting, and actually the user doesn't even really know their own password, funny enough. I think we should make sure to handle this at some point otherwise this could come back to bite us 🤔

codyebberson: Hm, ok.

But the limit here is 72 bytes (not characters), so that should be compatible with bcrypt, no?

For 2-byte characters, that means a max password length of 36 characters, which seems... fine?

OWASP recommends 72 character/byte limit: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#:~:text=bcrypt%20has%20a%20maximum%20length,in%20use%20has%20smaller%20limits).

ThatOneBro: The `isLength` validator does not measure bytes, see my example

codyebberson: Ahh, I see.  Good call.  I'll change to 36.

codyebberson: `isByteLength()` ✅

codyebberson: See https://github.com/medplum/medplum/pull/3334/files#diff-902e8aa40885f63d7b3dcce2af92c11ebe5debbd9c40901a3b29ba635441cc65R137

ThatOneBro: Wow that was an easy fix. Didn't even see that. Good find

ThatOneBro (approved):

PR #3338, opened 2023-11-18
Use react-jsx transform

codyebberson:
Use [the new JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) in all React projects.

In practical terms:
* In `tsconfig.json`, use `"jsx": "react-jsx"` rather than `"jsx": "react"`
* Remove all `import React from 'react';` declarations

References
* https://medplum.slack.com/archives/C051SKJKZ50/p1700094668578869
* https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
* https://stackoverflow.com/questions/67776707/ts-config-jsx-reactjsx-vs-react

### DIFF
--- a/examples/foomedical/src/components/InfoButton.tsx
+++ b/examples/foomedical/src/components/InfoButton.tsx
@@ -1,4 +1,5 @@
 import { createStyles, Group, UnstyledButton } from '@mantine/core';
+import { ReactNode } from 'react';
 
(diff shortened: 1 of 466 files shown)


ThatOneBro (approved):
Nice, LGTM

reshmakh (approved):

PR #3346, opened 2023-11-20
Forms for resource profiles

mattlong:
### Overview

A new `Profiles` tab is added to the resource detail page. This likely isn't how we'll actually want to present profiles; it's more of a development sandbox.

Fields from a profile's extensions are rendered within the form and can be edited/saved. I've mostly been developing against [`USCorePatientProfile`](https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-patient.html). Currently, that is the only profile that is know to be well-supported. After this foundational PR lands, additional profiles will be checked for compatibility.

### Implementation notes

The React component hierarchy has been fleshed out to further mirror the structure of `InternalTypeSchema` and `InternalSchemaElement`. The high-level changes:

A new `SliceInput` component used in conjunction with `ResourceArrayInput` to facilitate rendering extension slices.

Extraction of a new `ElementsInput` from `BackboneElementInput` to allow reuse of the logic of iterating through a collection of `InternalSchemaElement` entries in slices.

A rework of `ExtensionInput` that drills into an extension's contents rather than presenting a JSON input. This abstraction doesn't feel quite right yet given that there are two structurally different types of extension element patterns encountered (so far):

1. Extension slices that reference a separate StructureDefinition via the `type.profile` field, e.g. `Patient.extension` of the [US Core Patient Profile](https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-patient.html)
2. Extensions slices that include the contents of nested slices directly, e.g. `Extension.extension` of the [US Core Race Extension](https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-race.html)

The first type described is still needs some finessing with regards to how Extensions interact with slices since slicing the values based on the discriminator requires first fetching the schema for each slice/sub-extension. I believe this will become more straightforward once #3398, assuming the work related to that issue would apply not only when validating extensions, but parsing them as well.

### Profile validation
Profile validation seems to not be fully enforced correctly yet. As an example, for the [`USCorePatientProfile`](https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-patient.html), a patient record including an entry of the [`USCoreRace` extension](https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-race.html) passes validation even though it is missing the required "text" element; `Extension.extension:text.value[x]` specifies `min: 1`. I haven't looked to closely nor have an idea on the scope of work that'll be necessary yet around profile validation, but wanted to call it out. This should also be related to #3398 

```
{
  "resourceType": "Patient",
  "id": "4c93dfd3-878f-4dbc-8c30-46d146105bc3",
  "meta": {
    "profile": [
      "http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient"
    ]
  },
  ... rest of the Patient resource omitted for brevity
  "extension": [
    {
      "extension": [
        {
          "valueCoding": {
            "system": "urn:oid:2.16.840.1.113883.6.238",
            "code": "2106-3",
            "display": "White"
          },
          "url": "ombCategory"
        }
        // a value for "url": "text"should be required here
      ],
      "url": "http://hl7.org/fhir/us/core/StructureDefinition/us-core-race"
    }
  ]
}
```

### US Core Implementation Guide

Most of the US Core IG was hastily included in this PR for the sake of having profiles to develop against.  The accompanying ValueSet JSON file is incomplete; it includes just the ValueSets needed for `USCorePatientProfile`. Some of those ValueSets have bogus/placeholder values at that. Deciding how to allow users to easily incorporate implementation guides into Medplum remains TBD.


packages/react/src/ResourcePropertyInput/ResourcePropertyInput.utils.ts
```diff
@@ -0,0 +1,9 @@
+import { OperationOutcome } from '@medplum/fhirtypes';
+
+export interface ComplexTypeInputProps<ValueType> {
```
codyebberson, line 3: Nice, like this

packages/core/src/client.ts
```diff
@@ -1616,6 +1617,44 @@ export class MedplumClient extends EventTarget {
...
+
+    const cacheKey = profileUrl + '-requestSchema';
+    const cached = this.getCacheEntry(cacheKey, undefined);
+    if (cached) {
+      return cached.value;
+    }
+
+    const promise = new ReadablePromise<void>(
+      (async () => {
+        // TODO{mattlong}: account for https://hl7.org/fhir/references.html#canonical-matching
```
codyebberson, line 1641: 👍

packages/core/src/fhirpath/utils.ts
```diff
@@ -101,20 +101,29 @@ function getTypedPropertyValueWithSchema(
   let resultValue: any = undefined;
   let resultType = 'undefined';
 
-  if (types.length === 1) {
-    resultValue = input.value[path];
+  if (path === 'value[x]') {
```
codyebberson, line 104: This probably should be `path.endsWith('[x]')` or just `types.length > 0`

(`types.length` should always be greater than zero if `path.endsWith('[x]')`)

packages/core/src/fhirpath/utils.ts
```diff
@@ -101,20 +101,29 @@ function getTypedPropertyValueWithSchema(
...
+    for (const type of types) {
+      const path2 = 'value' + capitalize(type.code);
+      if (path2 in value) {
+        resultValue = value[path2];
```
codyebberson, line 108: Actually, I think this block should be compatible with the `else` block below?

packages/react/src/AddressInput/AddressInput.tsx
```diff
@@ -15,18 +17,22 @@ function setLine(address: Address, index: number, str: string): Address {
...
 export function AddressInput(props: AddressInputProps): JSX.Element {
   const [value, setValue] = useState<Address>(props.defaultValue || {});
+  const { getNestedElement } = useContext(BackboneElementContext);
 
   const valueRef = useRef<Address>();
   valueRef.current = value;
 
+  const stateElement = useMemo(() => getNestedElement(props.path, 'state'), [getNestedElement, props.path]);
+  if (stateElement?.binding) {
+    // TODO{mattlong} is it worth the complexity of subbing in an autocomplete input when
```
codyebberson, line 31: This is a good question.

I think it's a good idea, and something we should consider.  I'm ok to leave that as a future todo though, as this PR is quite big.

I'm quite intrigued by the idea we discussed about `BackboneElementInput` in a horizontal row, which would potentially simplify all of that logic into a single place.  That would also solve this issue as well (i think?)

packages/react/src/BackboneElementInput/BackbonElementInput.utils.ts
```diff
@@ -0,0 +1,56 @@
+import { InternalSchemaElement, InternalTypeSchema } from '@medplum/core';
+import React from 'react';
+
+export function splitRight(str: string, delim: string): [string, string] {
+  const lastIndex = str.lastIndexOf(delim);
```
codyebberson, line 5: Does this need to check for `lastIndex === -1`?

packages/react/src/ContactPointInput/ContactPointInput.tsx
```diff
@@ -49,25 +57,31 @@ export function ContactPointInput(props: ContactPointInputProps): JSX.Element {
   }
 
   return (
-    <Group spacing="xs" grow noWrap>
+    <Group spacing="xs" grow noWrap align="flex-start">
       <NativeSelect
         data-testid="system"
         defaultValue={contactPoint?.system}
+        required={(systemElement?.min ?? 0) > 0}
```
codyebberson, line 64: What is the case when `systemElement` is `undefined`?

Not trying to needle.  We have a variety of different philosophies regarding "what should this component do if/when not all data is loaded?"  

1. `return null;`
2. `return <Loader />;`
3. Try your best with what data you do have

That 3rd option gets tricky, because it can lead to lots of content layout shifts and jumpy pages.

If this is by design, that's cool, just want to make sure we're tracking the best option.

packages/react/src/ElementsInput/ElementsInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        // mostly for Extension.url
+        if (key === 'url' && element.fixed) {
+          return null;
+        }
+
+        if (EXTENSION_KEYS.includes(key)) {
+          // TODO{mattlong} verify the following comment is accurate
+          // an extension property without slices has no sub-extensions nor content
```
codyebberson, line 65: 🤔 

Globally?  No, that is not true. Many resources use nested extensions where the outer extension has a fully qualified URL, and the inner extensions have a friendly name.

In the context of FHIR Profiles?  That might be true?  @mattwiller  do you know?

packages/react/src/HumanNameInput/HumanNameInput.tsx
```diff
@@ -22,33 +18,33 @@ export function HumanNameInput(props: HumanNameInputProps): JSX.Element {
...
 
   function setUse(use: 'temp' | 'old' | 'usual' | 'official' | 'nickname' | 'anonymous' | 'maiden' | undefined): void {
-    setValueWrapper({ ...valueRef.current, use: use || undefined });
+    setValueWrapper({ ...value, use: use || undefined });
   }
 
   function setPrefix(prefix: string): void {
     setValueWrapper({
-      ...valueRef.current,
+      ...value,
```
codyebberson, line 26: As discussed offline... I think this is fine, and probably an artifact of previous versions.  We just test the heck out of it to be sure.

packages/react/src/ResourceArrayInput/ResourceArrayInput.tsx
```diff
@@ -1,48 +1,276 @@
...
 export function ResourceArrayInput(props: ResourceArrayInputProps): JSX.Element {
-  const [values, setValues] = useState(
-    props.defaultValue && Array.isArray(props.defaultValue) ? props.defaultValue : []
-  );
+  const { property } = props;
+  const medplum = useMedplum();
+  const [loading, setLoading] = useState(true);
+  const [slices, setSlices] = useState<SupportedSliceDefinition[]>();
+  // props.defaultValue should NOT be used after this; prefer the defaultValue state
+  const [defaultValue] = useState<any[]>(() => (Array.isArray(props.defaultValue) ? props.defaultValue : []));
```
codyebberson, line 139: Should the `else` case be `[props.defaultValue]`?

I think we have an `arrayify` util somewhere in core.

codyebberson (commented):
Nice work @mattlong, this looks great.

Overall this feels right on track.  No serious comments.  Bunch of minor stuff.  I'm disregarding all `console.log` statements, which we probably need to figure out a better long term solution for anyway.

Probably the biggest question is about whether we actually want to commit `uscore` JSON files into the repo, or an alternate storage system...

My recommendation would be to do one cleanup pass, and then let's promote this to staging for a team bug bash.

Great work 🎉

packages/core/src/fhirpath/utils.ts
mattlong: It would, except the length of `types` for `Extension.value[x]` can have a length of 1, so the `else if (types.length === 1)` would catch it.

Here's an example from [this extension](https://hl7.org/fhir/us/core/STU6.1/StructureDefinition-us-core-birthsex.html)

```javascript
    {
      "id" : "Extension.value[x]",
      "path" : "Extension.value[x]",
      "type" : [{
        "code" : "code"
      }],
      "binding" : {
        "strength" : "required",
        "description" : "Code for sex assigned at birth",
        "valueSet" : "http://hl7.org/fhir/us/core/ValueSet/birthsex"
      },
    // ...Other properties excluded for brevity
    }
```

If we don't want to have the similar contents of the `if` and `else` blocks, it seems like the following would be equivalent to what we have now, but feels weird with that dangling else block...
```javascript
  if (path.endsWith('[x]')) {
    for (const type of types) {
      const path2 = path.replace('[x]', '') + capitalize(type.code);
      if (path2 in value) {
        resultValue = value[path2];
        resultType = type.code;
        break;
      }
    }
  } else if (types.length === 1) {
    resultValue = value[path];
    resultType = types[0].code;
  } else {
    // what to do?
  }
```

packages/react/src/AddressInput/AddressInput.tsx
mattlong: I don't think so?

The type of `address.state` is still string (even though it now has a binding), so `ElementDefinitionTypeInput` would still wind up returning a plain old `TextInput` instead of one of the autocomplete-based inputs that uses the binding.

Or am I not understanding your suggestion correctly?

packages/react/src/BackboneElementInput/BackbonElementInput.utils.ts
mattlong: I will check that when I get around to writing tests!

packages/react/src/ResourceArrayInput/ResourceArrayInput.tsx
mattlong: I wasn't sure about that; I wanted to keep the handling of `props.defaultValue` the same that it was before, which was:

```javascript
const [values, setValues] = useState(
    props.defaultValue && Array.isArray(props.defaultValue) ? props.defaultValue : []
  );
```

The change you're suggesting doesn't feel quite right since it'd imply you could pass a single, non-array value to a component that is array-based and it would "just work". I'd like to have that be more easily enforced by the type system, type-narrowing, etc., but that'll be it's own project since the usage of `any` is so prevalent right now.

mattlong: To be more explicit, it'd be great to stop using `any` throughout the repo (or just within `ResourceForm` to start) since it makes it difficult to impossible to use various features of typescript.

packages/react/src/ContactPointInput/ContactPointInput.tsx
mattlong: Good question. This gets at the heart of the new `BackboneElementContext.getNestedElement`. Right now, `getNestedElement`, only returns an element when one has been explicitly defined in the StructureDefinition being displayed. Otherwise, it returns `undefined`; e.g. if you are on the Edit tab instead of the profiles tab. So the idea is that when it is `undefined`, it is safe to assume that the element has the default definition of it's type. Which, in this particular example, [ContactPoint.system is not required](https://hl7.org/fhir/R4/datatypes-definitions.html#ContactPoint.system).

There should not be a "loading" phase here that would lead to  layout shifts since the form isn't rendered until `medplum.requestProfileSchema` or `medplum.requestSchema` returns in the `useEffect` in `ResourceForm.tsx`

packages/react/src/ElementsInput/ElementsInput.tsx
mattlong: Right, but isn't the only way to specify the nested extension by using a slice on the outer extension?

mattlong: e.g. `Extension.extension:ethnicity`?

packages/react/src/ContactPointInput/ContactPointInput.tsx
mattlong: I was exploring a version where `getNestedElement` would (always) return an element even when it wasn't being modified in some way by a profile, but decided against it since it made `BackbonElementContext` more complicated to implement for seemingly little gain.

packages/react/src/ElementsInput/ElementsInput.tsx
codyebberson: > the only way to specify the nested extension by using a slice on the outer extension

I don't know, but that seems likely.

I think I missed the key term here, "extension property", meaning a property defined in a FHIR Profile?

In that case, yes I believe the sentence "an extension property without slices has no sub-extensions nor content" is true.

packages/react/src/ExtensionInput/ExtensionInput.tsx
```diff
@@ -1,14 +1,113 @@
...
+  console.debug(typeSchema.name, { extensionHasValue });
+
+  /*
+    From the spec:
+    An extension SHALL have either a value (i.e. a value[x] element) or sub-extensions, but not both.
+    If present, the value[x] element SHALL have content (value attribute or other elements)
+  */
+
+  // It seems like the behavior of ExtensionInput should differ based on extensionHasValue. It likely
+  // isn't strictly necessary to do so given the recursive use of BackboneElementInput
```
codyebberson, line 83: This makes sense.  It's a good catch.  I don't know how much of an optimization it would be, but I support it 👍

packages/react/src/AddressInput/AddressInput.tsx
mattlong: Filed https://github.com/medplum/medplum/issues/3519

packages/app/src/resource/ProfilesPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+  ResourceForm,
+  isSupportedProfileStructureDefinition,
+  useMedplum,
+} from '@medplum/react';
+import { useNavigate, useParams } from 'react-router-dom';
+import { showNotification } from '@mantine/notifications';
+import { deepClone, normalizeErrorString, normalizeOperationOutcome } from '@medplum/core';
+import { Button, Code, Group, Stack, Tabs, ThemeIcon } from '@mantine/core';
+import { addProfileToResource, cleanResource } from './utils';
+import { IconCheck } from '@tabler/icons-react';
```
codyebberson, line 15: Minor nit: run VS Code "organize imports"

(longer conversation about why ESLint doesn't do this / check this)

packages/app/src/resource/ProfilesPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }, [resource, availableProfiles]);
+
+  if (!resource) {
+    return null;
+  }
+
+  return (
+    <Document>
+      <h2>Available {resourceType} profiles</h2>
```
codyebberson, line 64: Minor:  We've been trying to use Mantine's `<Title order={2}>` rather than `<h2>`

packages/app/src/resource/ProfilesPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+    },
+    [medplum, navigate, resourceType, id, profile.url]
+  );
+
+  const handleDelete = useCallback(() => {
+    navigate(`/${resourceType}/${id}/delete`);
+  }, [navigate, resourceType, id]);
+
+  return (
+    <div>
```
codyebberson, line 151: Curious if this `<div>` was necessary for anything?

packages/core/src/client.ts
```diff
@@ -1616,6 +1618,45 @@ export class MedplumClient extends EventTarget {
...
+    const cached = this.getCacheEntry(cacheKey, undefined);
+    if (cached) {
+      return cached.value;
+    }
+
+    const promise = new ReadablePromise<void>(
+      (async () => {
+        // Just sort by lastUpdated. Ideally, it would also be based on a logical sort of version
+        // See https://hl7.org/fhir/references.html#canonical-matching for more discussion
+        const response = await this.searchResources('StructureDefinition', {
```
codyebberson, line 1643: Minor: You could use `searchOne` and omit the `_count: 1` prop.

packages/core/src/fhirpath/utils.ts
codyebberson: 1. Offline discussions helped me understand the case where profile restricts the `types` array, so that makes sense
2. Note this other PR which may cause a merge conflict here :'( https://github.com/medplum/medplum/pull/3538

codyebberson (approved):

PR #3350, opened 2023-11-21
feat(fhircast): add `heartbeat` event

### LINKED ISSUE #3356: [FHIRcast] Add `heartbeat` server-side event
Currently our ELB will timeout inactive `FHIRcast` connections (and WebSocket connections broadly) after 60 seconds. This does not conform to our stated `FHIRcast` timeout of minimum 3600 seconds, so we should add the `heartbeat` event which is sent out every 10 seconds to all connected client to keep all `FHIRcast` connections active.

See: https://build.fhir.org/ig/HL7/fhircast-docs/branches/master/3-2-2-Heartbeat.html

ThatOneBro:
Closes #3356

### DIFF
--- a/packages/server/src/fhircast/websocket.ts
+++ b/packages/server/src/fhircast/websocket.ts
@@ -3,6 +3,27 @@ import { IncomingMessage } from 'http';
 import ws from 'ws';
 import { globalLogger } from '../logger';
 import { getRedis } from '../redis';
+import { setupHeartbeatTimer } from './utils';
+
+const heartbeatTimerCallbacks = new Map<string, () => void>();
+const cleanupPromises = [] as Promise<void>[];
+
+/**
+ * Cleans up heartbeat timers for FHIRcast WebSocket connections.
+ */
+export function cleanupHeartbeatTimers(): void {
+  for (const cleanup of heartbeatTimerCallbacks.values()) {
+    cleanup();
+  }
+}
+
+/**
+ *
+ * @returns Promise that resolves when all registered WebSocket cleanup Promises have resolved or rejected.
+ */
+export async function cleanupWebSockets(): Promise<unknown> {
+  return Promise.allSettled(cleanupPromises);
+}
 
 /**
  * Handles a new WebSocket connection to the FHIRCast hub.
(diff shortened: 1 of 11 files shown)


codyebberson (commented):
Let's discuss next week.  My assumption was that the client would send the heartbeat, not the server.  If we can avoid it, my preference would be to try to avoid timers in server whenever possible.  The whole `HeartbeatStore` class could be avoided if the client manages the timers.

ThatOneBro:
Just implementing it per the `FHIRcast` spec:
https://build.fhir.org/ig/HL7/fhircast-docs/3-2-2-Heartbeat.html#:~:text=The%20Heartbeat.html%20event%20is%20sent%20regularly%20by%20the%20Hub

ThatOneBro:
Potentially we could use one timer and just blast out the heartbeat to all the tracked topics.

codyebberson:
> Just implementing it per the `FHIRcast` spec:

Ahh, good find!  I stand corrected.  Ok, need to noodle on that.

ThatOneBro:
@codyebberson I refactored this to use only one timer that gets cleaned up if there are no active topics. Does this make more sense maybe?

packages/server/src/fhircast/websocket.ts
```diff
@@ -17,10 +73,34 @@ export async function handleFhircastConnection(socket: ws.WebSocket, request: In
   // According to Redis documentation: http://redis.io/commands/subscribe
   // Once the client enters the subscribed state it is not supposed to issue any other commands,
   // except for additional SUBSCRIBE, PSUBSCRIBE, UNSUBSCRIBE and PUNSUBSCRIBE commands.
-  const redisSubscriber = getRedis().duplicate();
+  const redis = getRedis();
+  const redisSubscriber = redis.duplicate();
+  const deferredCleanupPromise = createDeferredPromise();
+  cleanupPromises.push(deferredCleanupPromise.promise);
```
codyebberson, line 79: It looks like we're only adding to `cleanupPromises`, and never removing?  If lots of frequent connect/disconnects, would this array continue to grow?  Probably fine in the short term when not many users, but ideally we'd remove promises from the array on websocket close.

codyebberson (approved):
Nice, thanks @ThatOneBro 

See one comment on `cleanupPromises` array.  We can push that to separate commit if you prefer.

Also, as a thought exercise, do you think this could work with the Medplum Agent websocket connection?  It will need a similar heartbeat feature.  No need to force it if it's too unnatural, but if it fits, that'd be neat.

packages/server/src/fhircast/websocket.ts
ThatOneBro: Good point, I think in an earlier iteration I couldn't figure out how to safely remove promises reliably, but I guess when they are resolved they can probably just be removed. I'll fix that

ThatOneBro:
> Also, as a thought exercise, do you think this could work with the Medplum Agent websocket connection? It will need a similar heartbeat feature. No need to force it if it's too unnatural, but if it fits, that'd be neat.

I think that could work. Do you think it would be best just to turn this into an abstract module that gets pulled into both files? Would we want to drive both off of the same timer?

Issue #3340, opened 2023-11-20
Enhancement for Project-Specific Password Reset


ravindra-tc:
**Background**
We addressed a limitation in the password reset functionality within the Medplum system. The current process lacks project specificity, making it challenging to determine the intended project when resetting the password for a user associated with multiple projects.

**Issue Details:**

1. **Problem Statement:**
   - The password reset process in Medplum is not project-specific, causing difficulties when attempting to reset the password for a user associated with multiple projects.

2. **Global User Consideration:**
   - Exploring the option of creating a global user to cover all projects was discussed. However, the current system constraints, specifically the requirement for a project ID during user creation, make this approach unfeasible.

3. **Feasibility of Project-Specific:**
   - The objective is to enhance the password reset functionality to be project-specific, allowing for a more targeted and accurate process when dealing with users across multiple projects.

**Proposed Resolution:**
   - Implement a project-specific password reset mechanism, ensuring that the system can correctly identify and process password resets for users associated with multiple projects.

**Benefits:**
- Improved security and specificity in the password reset process.
- Enhanced user experience for administrators dealing with users across different projects.

rahul1:
From discord: https://github.com/medplum/medplum/blob/main/packages/server/src/auth/resetpassword.ts#L60-L69


The bug is centered in [these lines](https://github.com/medplum/medplum/blob/ab33abfba36b640aa372e59eb4a58d615fce91e0/packages/server/src/auth/resetpassword.ts#L60-L69). 

The system repo searches for a user by email address. If we have a two project-scoped users in two different projects that share an email address, this search doesn't guarantee returning the correct user

PR #3351, opened 2023-11-21
Fixed - Project specific reset password

ravindra-tc:
### Description

This pull request addresses the project-specific reset password issue in the Medplum project. Specifically, the fix involves adding the project to the user filters to resolve the reset password issue. 

### Test Cases

I have run test cases for the resetPassword functionality. However, I've identified an issue with the `Reset Password › Custom reCAPTCHA site key success` test case. Upon investigation, it appears that the test is failing because the `projectId` is provided in the request body, but the associated project is missing in the user resource. The root cause is that the project is not assigned to the user resource at the time of creating the user. 
[Test - Reset Password › Custom reCAPTCHA site key success](https://github.com/medplum/medplum/blob/01fdd05735ad1794884c349a876e68df01fe44b2/packages/server/src/auth/resetpassword.test.ts#L199-L245).

### Action Items

- [x] Fix the reset password issue by adding the project to the user filters.
- [ ] Review by @rahul1 and @codyebberson.

Please look at it and take it forward.

@rahul1 @codyebberson

#3340

### DIFF
--- a/packages/server/src/auth/resetpassword.ts
+++ b/packages/server/src/auth/resetpassword.ts
@@ -57,16 +57,29 @@ export async function resetPasswordHandler(req: Request, res: Response): Promise
     }
   }
 
-  const user = await systemRepo.searchOne<User>({
-    resourceType: 'User',
-    filters: [
-      {
-        code: 'email',
-        operator: Operator.EXACT,
-        value: email,
-      },
-    ],
+// Define filters for searching users
+const filters = [
+  {
+    code: 'email',
+    operator: Operator.EXACT,
+    value: email, // Set the email value for exact matching
+  },
+];
+
+// If a specific project is associated with the request, add a project filter
+if (req.body.projectId) {
+  filters.push({
+    code: 'project',
+    operator: Operator.EQUALS,
+    value: 'Project/' + req.body.projectId, // Set the project value for equality matching
   });
+}
+
+// Search for a user based on the defined filters
+const user = await systemRepo.searchOne<User>({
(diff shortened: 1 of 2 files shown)


packages/server/src/auth/resetpassword.ts
```diff
@@ -57,16 +57,29 @@ export async function resetPasswordHandler(req: Request, res: Response): Promise
...
+];
+
+// If a specific project is associated with the request, add a project filter
+if (req.body.projectId) {
+  filters.push({
+    code: 'project',
+    operator: Operator.EQUALS,
+    value: 'Project/' + req.body.projectId, // Set the project value for equality matching
   });
+}
```
rahul1, line 76: For safety, add an `else` clause here that filters for `project` is missing: 
```
    code: 'project',
    operator: Operator.MISSING,
    value: true
  }
```

This will enforce that a missing project parameter only resets server-level useres

rahul1 (changes requested):
Thanks so much for the contribution @ravindra-tc . 

This looks good, but I made one request for an `else` clause. 
It also looks like there is a broken test. 

Running `npm run prettier` as and fixing those issues will make sure the build passes

ravindra-tc:
@rahul1 I've incorporated an `else` clause for the project-based filter. Furthermore, I've resolved the issues in the broken test case and introduced two new test cases to cover the new scenario.

rahul1 (approved):

codyebberson (approved):

PR #3364, opened 2023-11-22
Default strict mode on super admin project

### LINKED ISSUE #3353: SuperAdmin Project should be strict-mode by default
Currently, when creating a new self-hosted instance of Medplum, the SuperAdmin project does not have strict mode enabled. This allows super admin users to make schema breaking changes to FHIR resources. 

We should default to strict mode, and allow users to opt-out if required

jogeshwar01:
Fixes #3353 

Updated Super Admin project to now have strictMode:true by default.
Updated current tests accordingly.


codyebberson (approved):

codyebberson:
Thanks @jogeshwar01 !

PR #3373, opened 2023-11-24
improve some doc cohesion and fix generator->definitions README link

dillonstreator:


packages/docs/docs/fhir-datastore/migration-logic.md
```diff
@@ -23,7 +23,7 @@ When considering which option is best for you, we must consider some system char
 - How easy or difficult is it to modify the existing system?
   - Are engineers actively updating the existing system?
   - How frequently does the existing system deploy new releases?
-  - Is the existing system stable, or are changes avoided?
+  - Is the existing system stable, or is change frequent?
```
dillonstreator, line 26: I'm not entirely confident that this change accurately captures the original intention, but it seems the intention was to convey opposites. In my view, a "stable system" typically implies avoiding changes.

I suppose this is context-dependent; for instance, when discussing the stability of an API, frequent change may suggest lower stability. On the other hand, in the context of internal component modifications, updates could potentially enhance stability.

codyebberson (approved):

PR #3378, opened 2023-11-25
Add autoClose false to persist error toasts

### LINKED ISSUE #3376: Error toast should persist until dismissed
At present, error toasts disappear after a few seconds, making them hard to read and follow up on.  See video:

https://github.com/medplum/medplum/assets/2314326/73a6c835-7cf5-4faf-9417-d5a36097344e

jogeshwar01:
Fixes #3376 

- Added autoClose:false in all error notifications.

### DIFF
--- a/packages/app/src/HomePage.tsx
+++ b/packages/app/src/HomePage.tsx
@@ -71,12 +71,12 @@ export function HomePage(): JSX.Element {
             .then((blob: Blob) => {
               window.open(window.URL.createObjectURL(blob), '_blank');
             })
-            .catch((err) => showNotification({ color: 'red', message: normalizeErrorString(err) }));
+            .catch((err) => showNotification({ color: 'red', message: normalizeErrorString(err), autoClose: false }));
         }}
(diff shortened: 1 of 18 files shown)


codyebberson (approved):

codyebberson:
Thanks @jogeshwar01 !

dillonstreator:
@jogeshwar01 @codyebberson unfortunate that this needs to be configured at each `showNotification` call. It would be nice if `@mantine/notifications` supported a global configuration for this type of rule. I've open an upstream PR to `@mantine/notifications` to allow this customization at the top-level on the `<Notifications />` component.
https://github.com/mantinedev/mantine/pull/5329
If this goes through, I think we should adopt to reduce the repetition of setting `autoClose` to `false` for all calls of `showNotification` where `color="red"`

jogeshwar01:
@dillonstreator That looks good. Lets see if the PR gets through. Then if the maintainers would approve this, we can go ahead and reduce the repetitions.

Issue #3389, opened 2023-11-26
NextJS demo build failing due to optional dependency


codyebberson:
https://vercel.com/medplum/medplum-nextjs-demo/HF9v6UFvhJQ2XFezmMxZyVYBQ96X

<img width="942" alt="image" src="https://github.com/medplum/medplum/assets/749094/0ae5b247-458d-407a-b3c3-08a3a6ada750">

In https://github.com/medplum/medplum/pull/3387 we added `rfc6902` as an optional peer dependency.  It should only matter if you try to use the `<ResourceDiffTable>` React component, which the NextJS demo does not.

The weird thing is that Vercel defaulted to yarn rather than npm 🤷‍♂️

This is probably easily fixed by copying `examples/foomedical/vercel.json` to `examples/medplum-nextjs-demo`, which instructs Vercel to use npm rather than yarn.

It may also be worth investigating why yarn didn't respect `"optional": true`

This page has different syntax than what we're using: https://classic.yarnpkg.com/lang/en/docs/dependency-types/

Not sure which one is correct 😔

codyebberson:
<img width="475" alt="image" src="https://github.com/medplum/medplum/assets/749094/56f491fc-b124-452a-8914-091a2340b73f">

ThatOneBro:
`peerDependenciesMeta` with `optional: true` seems to be correct for this situation. The error is likely not related to our dependencies setup, at least directly.



Yarn does seem to support it, I got install working properly. But I did get an error before that:
<img width="1128" alt="Screenshot 2023-11-26 at 4 17 28 PM" src="https://github.com/medplum/medplum/assets/1592008/faf71db9-520e-492e-b42d-d1171bef8f71">

So I had to add `private: true` to the top level repo. See: https://classic.yarnpkg.com/lang/en/docs/workspaces/#:~:text=Note%20that%20the%20private%3A%20true%20is%20required!%20Workspaces%20are%20not%20meant%20to%20be%20published%2C%20so%20we%E2%80%99ve%20added%20this%20safety%20measure%20to%20make%20sure%20that%20nothing%20can%20accidentally%20expose%20them.

codyebberson:
@ThatOneBro - Good catch, and that's a good fix.  I don't think that will fix the `medplum-nextjs-demo` project though.

You can repro by cloning the separate repo: https://github.com/medplum/medplum-nextjs-demo

[...]

PR #3390, opened 2023-11-27
chore(package.json): add `private: true` to root

ThatOneBro:
Required for `yarn` to work with the monorepo: https://classic.yarnpkg.com/lang/en/docs/workspaces/#:~:text=Note%20that%20the%20private%3A%20true%20is%20required!%20Workspaces%20are%20not%20meant%20to%20be%20published%2C%20so%20we%E2%80%99ve%20added%20this%20safety%20measure%20to%20make%20sure%20that%20nothing%20can%20accidentally%20expose%20them.

This prevents publishing the `root` package to NPM, which kind of makes sense.

May help in resolving #3389 ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "root",
   "version": "2.1.20",
+  "private": true,
   "workspaces": [
     "packages/*",
     "examples/*"


codyebberson:
> ⚠️	
> Changes were made to package.json, but not to package-lock.json - Perhaps you need to run `npm install`?

Otherwise lgtm 👍

ThatOneBro:
Looks like it might be a red herring, I don't think this change will update `package-lock.json` maybe?

Screenshot 2023-11-26 at 7 54 25 PM

Issue #3335, opened 2023-11-17
Allow Project admins to force set password of Project-scoped Users


mattwiller:
Currently, [Force Set Password](https://www.medplum.com/docs/self-hosting/super-admin-guide#force-set-password) is only available to the server superadmin.  For a User scoped to a specific Project (i.e. with `User.projectId` set), their Project admins should be able to force (re)set their password

rahul1:
This refers to the feature that allows admin users to Forcibly set passwords from the Medplum App

PR #3394, opened 2023-11-27
Project admins force set password for project-scoped users

jogeshwar01:
#3335 

- Added a new ProjectAdminConfigPage in the Admin menu (only for project admins - can be seen in the picture below).
- Created a new route /admin/projects/setpassword - force set password for users associated to the project by project admins. (for users with project id set)
- Added tests for both client and server side.

- There is one case here - Even though the menu option config will not show up for super admin, the super admin can go to the /admin/config page manually - as the check for project admin also includes a super admin for now. But they already have the super config so it shouldn't be an issue. We can change this if needed.
 
I have made a few decisions myself here. We can change the format as per need. If thats the case, we can consider this as a checkpoint PR and further build upon it.

image

### DIFF
--- a/packages/app/src/admin/ProjectAdminConfigPage.tsx
+++ b/packages/app/src/admin/ProjectAdminConfigPage.tsx
@@ -30,7 +30,6 @@ export function ProjectAdminConfigPage(): JSX.Element {
         <Stack>
           <TextInput name="email" label="Email" required />
           <PasswordInput name="password" label="Password" required />
-          <TextInput name="projectId" label="Project ID" />
           <Button type="submit">Force Set Password</Button>
         </Stack>
       </Form>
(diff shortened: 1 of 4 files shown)


packages/app/src/admin/ProjectAdminConfigPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Title order={2}>Force Set Password</Title>
+      <p>
+        Note that this applies to all projects for the user. Therefore, this should only be used in extreme
+        circumstances. Always prefer to use the "Forgot Password" flow first.
+      </p>
+      <Form onSubmit={forceSetPassword}>
+        <Stack>
+          <TextInput name="email" label="Email" required />
+          <PasswordInput name="password" label="Password" required />
+          <TextInput name="projectId" label="Project ID" />
```
codyebberson, line 33: I think we can remove the `projectId` input, because it is implicit in this case, correct?

jogeshwar01:
@codyebberson Got it. I have added both the changes now.

Issue #3402, opened 2023-11-27
Make clear that GraphQL respects Access Policies


ksmith94:
User Driven: https://discord.com/channels/905144809105260605/1154903645700243549

Add a note to the GraphQL Basic Queries doc to make clear that GraphQL respects the access policies of the user making the query

PR #3403, opened 2023-11-27
Add note about graphql respecting access policies

ksmith94:
Add a note that GraphQL queries respect Mepdlum's Access Policies. Related to issue #3402

### DIFF
--- a/packages/docs/docs/graphql/basic-queries.mdx
+++ b/packages/docs/docs/graphql/basic-queries.mdx
@@ -50,6 +50,10 @@ For example, to request a `Patient` by ID:
 
 This query retrieves the `resourceType`, `id`, `name`, and `address` of the specified `Patient`.
 
+:::note Access Policies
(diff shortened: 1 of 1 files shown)


packages/docs/docs/graphql/basic-queries.mdx
```diff
@@ -50,6 +50,10 @@ For example, to request a `Patient` by ID:
 
 This query retrieves the `resourceType`, `id`, `name`, and `address` of the specified `Patient`.
 
+:::note Access Policies
+When using GraphQL, [access policies](/docs/access/access-policies) are enforced, so users will not be able to read or edit any resources they do not have access to.
```
mattwiller, line 54: If we wanted to be extra precise here, I'd suggest the following — feel free to take or leave it at your discretion

```suggestion
When using GraphQL, [access policies](/docs/access/access-policies) are enforced, so users will not be able to read or edit any resources (or inner fields) they do not have access to.
```

mattwiller (approved):

codyebberson (approved):

PR #3411, opened 2023-11-28
`BinarySource` supporting `ReadableStream`

dillonstreator:
`BinarySource` supporting `ReadableStream` to better enable streaming

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -58,7 +58,7 @@
     "@medplum/definitions": "*",
     "@medplum/fhirtypes": "*",
     "jest-websocket-mock": "2.5.0",
-    "web-streams-polyfill": "^3.2.1"
+    "web-streams-polyfill": "3.2.1"
   },
   "peerDependencies": {
     "pdfmake": "^0.2.5"
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -57,7 +57,8 @@
   "devDependencies": {
     "@medplum/definitions": "*",
     "@medplum/fhirtypes": "*",
-    "jest-websocket-mock": "2.5.0"
+    "jest-websocket-mock": "2.5.0",
+    "web-streams-polyfill": "^3.2.1"
   },
   "peerDependencies": {
     "pdfmake": "^0.2.5"
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1402,7 +1402,12 @@ describe('Client', () => {
 
     const fetch = mockFetch(200, {});
     const client = new MedplumClient({ fetch });
-    const promise = client.createBinary(Readable.from('Hello World'), undefined, ContentType.TEXT, onProgress);
+    const promise = client.createBinary(
+      Readable.toWeb(Readable.from('Hello World')) as ReadableStream,
+      undefined,
+      ContentType.TEXT,
+      onProgress
+    );
     expect(xhrMock.open).toBeCalled();
     expect(xhrMock.setRequestHeader).toBeCalled();
 
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1,9 +1,10 @@
+import { randomUUID, webcrypto } from 'node:crypto';
+import { Readable } from 'node:stream';
+import { TextEncoder } from 'node:util';
 import { Bot, Bundle, Identifier, Patient, SearchParameter, StructureDefinition } from '@medplum/fhirtypes';
 import { MockAsyncClientStorage } from '@medplum/mock';
-import { randomUUID, webcrypto } from 'crypto';
 import PdfPrinter from 'pdfmake';
 import type { CustomTableLayout, TDocumentDefinitions, TFontDictionary } from 'pdfmake/interfaces';
-import { TextEncoder } from 'util';
 import { encodeBase64 } from './base64';
 import {
   DEFAULT_ACCEPT,
@@ -19,7 +20,6 @@ import { ContentType } from './contenttype';
 import { OperationOutcomeError, notFound, unauthorized } from './outcomes';
 import { isDataTypeLoaded } from './typeschema/types';
 import { ProfileResource, createReference } from './utils';
-import { Readable } from 'stream';
 
 const patientStructureDefinition: StructureDefinition = {
   resourceType: 'StructureDefinition',
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -19,6 +19,7 @@ import { ContentType } from './contenttype';
 import { OperationOutcomeError, notFound, unauthorized } from './outcomes';
 import { isDataTypeLoaded } from './typeschema/types';
 import { ProfileResource, createReference } from './utils';
+import { Readable } from 'stream';
 
 const patientStructureDefinition: StructureDefinition = {
   resourceType: 'StructureDefinition',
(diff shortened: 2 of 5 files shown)


packages/core/src/client.test.ts
```diff
@@ -1401,7 +1402,7 @@ describe('Client', () => {
 
     const fetch = mockFetch(200, {});
     const client = new MedplumClient({ fetch });
-    const promise = client.createBinary('Hello world', undefined, ContentType.TEXT, onProgress);
+    const promise = client.createBinary(Readable.from('Hello World'), undefined, ContentType.TEXT, onProgress);
```
dillonstreator, line 1405: Could move this to a separate test, if deemed necessary.
Could also do a more fine-grained test of progress tracking now that we can directly mock readable.

ThatOneBro: I think having both tests is good, just so we're hitting both paths through the codebase 👍

packages/core/src/client.ts
```diff
@@ -1885,7 +1886,17 @@ export class MedplumClient extends EventTarget {
       xhr.setRequestHeader('Cache-Control', 'no-cache, no-store, max-age=0');
       xhr.setRequestHeader('Content-Type', contentType);
       xhr.setRequestHeader('X-Medplum', 'extended');
-      xhr.send(data);
+      if (data instanceof Readable) {
```
ThatOneBro, line 1889: Instead of `Readable` and `Writable` which are Node APIs, can we use `ReadableStream` and `WritableStream` since these will work on both browser and in Node?

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream
https://developer.mozilla.org/en-US/docs/Web/API/WritableStream

They function a bit differently internally but you can easily convert to and from `node.Readable` and `ReadableStream` via some utils in `node:stream`, namely `Readable.fromWeb(ReadableStream)` and `Readable.toWeb(node.Readable)`.

https://nodejs.org/api/stream.html#streamreadabletowebstreamreadable-options

You can call `Readable.toWeb()` on your `Readable` before passing in the `ReadableStream` in as the `BinarySource`

I think it would be great to have file streaming on both the browser and Node platforms

ThatOneBro (changes requested):
Overall I think this will be a great change. Just one comment about using `ReadableStream` instead of `Readable` so that this will work in the browser as well 👍

packages/core/test.setup.cjs
```diff
@@ -1,6 +1,7 @@
 /* globals module require globalThis */
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { MemoryStorage } = require('@medplum/core');
+require('web-streams-polyfill');
```
ThatOneBro, line 4: Oh interesting, I guess this is necessary because `jsdom` doesn't have the WebStream API supported, right? I didn't even realize that

packages/core/src/client.ts
dillonstreator: Ah yes that makes more sense
Moved that direction in this commit https://github.com/medplum/medplum/pull/3411/commits/b27e68c30fd41c42ba0fb73ea2886ab484404732

packages/core/src/client.ts
```diff
@@ -1885,7 +1885,18 @@ export class MedplumClient extends EventTarget {
...
-      xhr.send(data);
+      if (data instanceof ReadableStream) {
+        const writableStream = new WritableStream({
+          write(chunk) {
+            xhr.send(chunk);
+          },
+        });
+
+        // eslint-disable-next-line no-void
+        void data.pipeTo(writableStream);
```
ThatOneBro, line 1896:
```suggestion
        data.pipeTo(writableStream).catch(console.error);
```
I think doing this should make the `void` here unnecessary and will make sure we don't have any floating promises

ThatOneBro (changes requested):

ThatOneBro:
Overall this looks really good! I'm excited to get this in

packages/core/package.json
```diff
@@ -57,7 +57,8 @@
   "devDependencies": {
     "@medplum/definitions": "*",
     "@medplum/fhirtypes": "*",
-    "jest-websocket-mock": "2.5.0"
+    "jest-websocket-mock": "2.5.0",
+    "web-streams-polyfill": "^3.2.1"
```
ThatOneBro, line 61:
```suggestion
    "web-streams-polyfill": "3.2.1"
```
Oh yeah we like to use exact versions here, you may have to rerun `npm install` again after this change.

ThatOneBro (changes requested):

packages/core/src/client.ts
dillonstreator: gotcha ya wasn't exactly sure what to do here given that resolve/reject is being handled in `xhr.onload`
can drop this in

packages/core/test.setup.cjs
dillonstreator: yup it was seemingly unavailable 🤷

packages/core/src/client.test.ts
```diff
@@ -1401,7 +1402,12 @@ describe('Client', () => {
 
     const fetch = mockFetch(200, {});
     const client = new MedplumClient({ fetch });
-    const promise = client.createBinary('Hello world', undefined, ContentType.TEXT, onProgress);
```
ThatOneBro, line 1404: Oh yeah, one last ask:

Can you add a separate test for the string version too?

My last nit, otherwise we'll get this in after tests pass on CI 👍

dillonstreator: ah yes, missed that!

ThatOneBro (approved):

ThatOneBro:
Thanks @dillonstreator! Going to merge once all the tests pass 👍 great addition

dillonstreator:
@ThatOneBro - just did some local testing with this and it is not working as intended. `xhr.send()` can't be repeatedly called with chunks of the payload.
I'm surprised to find very little online about streaming uploads with `XMLHttpRequest`. Understandably, most efforts appear to be placed into the `fetch` standard. As I understand it, we couldn't easily support progress tracking with `fetch` which is why we fallback to `XMLHttpRequest`.
We should revert this.

Any thoughts on how to approach supporting streaming?
I saw this doc comment that @codyebberson  wrote about a longer term approach to handling fetching https://github.com/medplum/medplum/blob/c80d69dd155f2253c40b8b5b9a2370791d805b60/packages/docs/docs/contributing/upgrading-dependencies.md?plain=1#L61-L70
Maybe a good time to continue the discussion of moving to Axios?

ThatOneBro:
The `node-fetch` dependency is something we should definitely look at again. As of Node 18, global `fetch` is now native to Node and so this dependency might be unnecessary since we recently set the minimum Node version to 18.

As for streaming, yeah you're right. I totally forgot that XHR has never supported file streaming. We could do multipart uploads (using `Content-Range`) but that wouldn't be quite the same as file streaming and would probably not have the desired performance.

In terms of `Axios`, I'm skeptical it will improve things much on the browser-side of things in regards to streaming. It looks like there are two backends for Axios, namely `node`, which uses `node:http` which does support streaming if I remember correctly, but I don't think it does it the same way that `fetch` does (uses HTTP/1.1 Chunked Transfer Encoding maybe?), and the second is `xhr`.  See: https://github.com/axios/axios/tree/v1.x/lib/adapters

So in the browser it is again probably not possible to do the file streaming while using `axios`. I think we will have to find a solution with `fetch` that still has upload progress (surely that exists by now... right?).

I did find this `TransformStream` solution, which may work, but I haven't tested it extensively. The example seems promising though:
https://stackoverflow.com/a/69400632
There is mention that `TransformStream` is not supported in browsers other than Chrome yet, but seems support has improved:
https://developer.mozilla.org/en-US/docs/Web/API/TransformStream

I'll probably play around with it later today.

In the meantime, yes we should revert this commit I guess and come back when we have a better solution. If you want to submit a PR reverting it I'll get that in. 👍

Thank you for your diligence in following up on this

dillonstreator:
Opened the revert https://github.com/medplum/medplum/pull/3421

Ah didn't know about the minimum Node version of 18 so ya agreed `node-fetch` is likely irrelevant.

Saw the same SO post and agree it seems quite promising.

ThatOneBro:
Yeah for sure! If you want to test it out, feel free. I think I'd also like to know @codyebberson's take on this too

PR #3418, opened 2023-11-28
Fixes #3401 - Handle no AWS credentials in AWS init tool

### LINKED ISSUE #3401: Error on aws init when AWS credentials not set
## Background 
Our [AWS Install docs](https://www.medplum.com/docs/self-hosting/install-on-aws#aws-cli-setup) indicate that setting up the AWS Credential Files is optional.  

However, if the user does not set up any AWS credentials (via either via env vars or credentials file), then the SIGNING KEY step of the AWS init 

## Actual Behavior

```bash
SIGNING KEY

Medplum uses AWS CloudFront Presigned URLs for binary content such as file uploads.
Error: Region is missing
```

## Expected Behavior
There are a few options: 
1. Update the docs to make the credentials step required (either via credentials file or environment vars)
2. Update the init script to skip this step if credentials are not found
3. Have the init tool prompt the user to log into the AWS console (similar to `medplum login`)?

codyebberson:


packages/cli/src/aws/init.ts
```diff
@@ -233,18 +239,21 @@ export async function initStackCommand(): Promise<void> {
   print('These values will be encrypted at rest.');
   print(`The values will be stored in the "/medplum/${config.name}" path.`);
 
-  const serverParams = {
+  const serverParams: Record<string, string | number> = {
```
ThatOneBro, line 242: Can this be a `Partial`?
```suggestion
  const serverParams: Partial<MedplumInfraConfig> = {
```

packages/cli/src/aws/init.ts
```diff
@@ -372,7 +388,7 @@ async function checkOk(text: string): Promise<void> {
  * @param configFileName - The config file name.
  * @param config - The config file contents.
  */
-function writeConfig(configFileName: string, config: MedplumInfraConfig): void {
+function writeConfig(configFileName: string, config: Record<string, any>): void {
```
ThatOneBro, line 391:
```suggestion
function writeConfig(configFileName: string, config: Partial<MedplumInfraConfig>): void {
```
Can this be a `Partial`?

ThatOneBro (commented):
Looks good for the most part, just curious if we can keep types a bit tighter with a `Partial` for the `MedplumInfraConfig`

rahul1 (approved):
LGTM, modulo @ThatOneBro comments re: types

codyebberson:
@ThatOneBro @rahul1 

> Looks good for the most part, just curious if we can keep types a bit tighter with a `Partial` for the `MedplumInfraConfig`

Unfortunately this is `MedplumServerConfig`, not `MedplumInfraConfig`.

At present, `MedplumServerConfig` lives in the `server` package.

I'm uncertain if we should move that to `core` as well 🤔

PR #3420, opened 2023-11-29
feat(@medplum/core) Add `AbortSignal` support to client's `createBinary` and `createAttachment` methods

dillonstreator:
Add `AbortSignal` support to `createBinary` and `createAttachment` by exposing an optional `signal: AbortSignal` that is propagated to the clients fetch interface as well as the `XMLHttpRequest`

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -1795,7 +1795,7 @@ export class MedplumClient extends EventTarget {
    * @param filename - Optional filename for the binary.
    * @param contentType - Content type for the binary.
    * @param onProgress - Optional callback for progress events.
-   * @param signal - Optional AbortSignal to short-circuit the request.
+   * @param signal - Optional AbortSignal to abort the request.
    * @returns The result of the create operation.
    */
   async createAttachment(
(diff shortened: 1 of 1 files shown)


packages/core/src/client.ts
```diff
@@ -1795,15 +1795,17 @@ export class MedplumClient extends EventTarget {
...
+   * @param signal - Optional AbortSignal to abort the request.
    * @returns The result of the create operation.
    */
   async createAttachment(
     data: BinarySource,
     filename: string | undefined,
     contentType: string,
-    onProgress?: (e: ProgressEvent) => void
+    onProgress?: (e: ProgressEvent) => void,
+    signal?: AbortSignal
```
codyebberson, line 1806: How would you feel about changing this param to `options: RequestInit = {}`?

My thinking:
1. It aligns with other methods in the client
2. It futureproofs the method for predictable future requests for setting HTTP headers, etc

There would be separate work to actually use the values of `RequestInit` with `XMLHttpRequest`, but I'm ok punting on that for now.

dillonstreator: > There would be separate work to actually use the values of RequestInit with XMLHttpRequest

Yeah this is tough because if someone is relying on an option of `RequestInit` and they then add an `onProgress` callback, they may unexpectedly change the behavior. If this were very explicitly documented somewhere, I could see getting behind it more.

dillonstreator: yup sounds good - added here https://github.com/medplum/medplum/pull/3420/commits/d8fb4c87a63455b1ce3d378b5981b57e61fee3dc

codyebberson (approved):

PR #3432, opened 2023-11-30
Stop reviving removed meta properties

mattlong:
Title pretty much says it all. Previously, attempting to completely remove `resource.meta.profile` or any other property within a resource's meta resulted in it silently being preserved/restored when updating the resource.

I believe all meta fields expected to exist after an update are recreated later on in the function.

### DIFF
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -145,6 +145,23 @@ describe('FHIR Repo', () => {
       expect(patient2.meta?.versionId).not.toEqual(patient1.meta?.versionId);
     }));
 
+  test('Update patient remove meta.profile', () =>
+    withTestContext(async () => {
+      const profileUrl = 'http://example.com/patient-profile';
+      const patient1 = await systemRepo.createResource<Patient>({
+        resourceType: 'Patient',
+        meta: { profile: [profileUrl] },
+        name: [{ given: ['Update1'], family: 'Update1' }],
+      });
+      expect(patient1.meta?.profile).toEqual(expect.arrayContaining([profileUrl]));
+      expect(patient1.meta?.profile?.length).toEqual(1);
+
+      const patientWithoutProfile = { ...patient1 };
+      delete (patientWithoutProfile.meta as any).profile;
+      const patient2 = await systemRepo.updateResource<Patient>(patientWithoutProfile);
+      expect('profile' in (patient2.meta as any)).toBe(false);
+    }));
+
   test('Update patient no changes', () =>
     withTestContext(async () => {
       const patient1 = await systemRepo.createResource<Patient>({
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -162,6 +162,32 @@ describe('FHIR Repo', () => {
       expect('profile' in (patient2.meta as any)).toBe(false);
     }));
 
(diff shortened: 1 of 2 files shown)


codyebberson (commented):
Oo, interesting.

I'm looking through all of our special meta properties...

The only one that concerns me is `project`.

For normal users, `project` will always be set by the repository context.

For super admins and `systemRepo` .... I fear that we might lose the `project` value?  The method `getProjectId` might need to take `existing` and `updated` similar to how `getAccount` works.

I'm not 100% sure -- but we should test

packages/server/src/fhir/repo.ts
```diff
@@ -1380,31 +1376,32 @@ export class Repository extends BaseRepository implements FhirRepository {
...
+    const submittedProjectId = updated.meta?.project;
     if (submittedProjectId && this.canWriteMeta()) {
       // If the resource has an project (whether provided or from existing),
       // and the current context is allowed to write meta,
       // then use the provided value.
       return submittedProjectId;
     }
 
-    return this.context.project;
+    return existing?.meta?.project ?? this.context.project;
```
mattlong, line 1404: @codyebberson I believe this is the desired order of preference. There's a slight logic change that results from this though: previously, an unchanged projectID would have been pulled into `submittedProjectId` above given the way meta from existing and updated were being merged. That arguably seems like it was semantically incorrect, but I think it points to a bigger issue that `canWriteMeta` is too broad since we want to allow non-superadmins to edit `meta.profile`.

Should `canWriteMeta` be renamed to `canWriteProtectedMeta` or something along those lines?

codyebberson: > I believe this is the desired order of preference.

Hmm 🤔 I'm trying to think what would be the setup where the order would matter...

When would `existing?.meta?.project` be `undefined`, but `this.context.project` has a value?

`existing?.meta?.project` is only `undefined` for resources outside of a `Project`, which generally means super admin resources, which means only super admins should ever be able to read/write them.

`this.context.project` is always set for all users other than super admins and `systemRepo`.

I think... I think order doesn't matter here, and they should never be in conflict?  So I think the way you have it is ok 👍 

> Should canWriteMeta be renamed to canWriteProtectedMeta or something along those lines?

Yes, I think that better represents the intent of the method.

packages/server/src/fhir/repo.test.ts
```diff
@@ -145,6 +145,49 @@ describe('FHIR Repo', () => {
...
+      expect(patient1.meta?.profile).toEqual(expect.arrayContaining([profileUrl]));
+      expect(patient1.meta?.profile?.length).toEqual(1);
+
+      const patientWithoutProfile = { ...patient1 };
+      delete (patientWithoutProfile.meta as any).profile;
+      const patient2 = await systemRepo.updateResource<Patient>(patientWithoutProfile);
+      expect('profile' in (patient2.meta as any)).toBe(false);
+    }));
+
+  test.only('meta.project preserved after attempting to remove it', () =>
```
codyebberson, line 165: I think we need to remove `.only` here, yeah?

codyebberson (changes requested):
LGTM other than `test.only`

packages/server/src/fhir/repo.test.ts
mattlong: oops, yes indeed!

codyebberson (approved):

PR #3437, opened 2023-12-01
Update AWS instructions

rahul1:


packages/docs/docs/self-hosting/install-on-aws.md
```diff
@@ -82,6 +87,12 @@ You will need permission to access the following AWS services:
 | Systems Manager (SSM)                | Store configuration details                                                           |
 | Web Application Firewall (WAF)       | Protect your web applications or APIs against common web exploits and bots            |
 
+### Configure your name servers
+
+Medplum strongly recommends configuring your domain to use [Amazon's Route 53](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/MigratingDNS.html) as your DNS service as your name server. This will make it much easier to set up SSL certificates for the [Medplum App](https://www.medplum.com/docs/app) and [Medplum Binary Storage](/docs/fhir-datastore/binary-data).
```
rahul1, line 92:
```suggestion
Medplum strongly recommends configuring your domain to use [Amazon's Route 53](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/MigratingDNS.html) as your custom DNS name server. This will make it much easier to set up SSL certificates for the [Medplum App](https://www.medplum.com/docs/app) and [Medplum Binary Storage](/docs/fhir-datastore/binary-data).
```
```suggestion
Medplum strongly recommends configuring your domain to use [Amazon's Route 53](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/MigratingDNS.html) as your custom DNS name server. This will make it much easier to set up SSL certificates for the [Medplum App](https://www.medplum.com/docs/app) and [Medplum Binary Storage](/docs/fhir-datastore/binary-data).
```

scripts/deploy-bot-layer.sh
```diff
@@ -41,7 +47,7 @@ zip -r -q medplum-bot-layer.zip .
 
 # Publish the bot layer
 aws lambda publish-layer-version \
-  --region us-east-1 \
+  --region $AWS_REGION \
```
codyebberson, line 50: We should just remove this line entirely.  The AWS CLI tools will pull the default value the exact same way `aws configure get region` does.  And that would also allow users to specify on the command line directly.

```bash
AWS_REGION=us-east-2 ./deploy-bot-layer.sh
```

codyebberson (commented):
Nice, lots of good finds in there.

See comment on the region param in `deploy-bot-layer.sh`, otherwise lgtm 👍

codyebberson (approved):

PR #3441, opened 2023-12-01
Create example bots that use a shared code library

ksmith94:
This updates the build system in the medplum-demo-bots directory to use esbuild so that bots can import functions from a shared library. It also provides two very simple example bots that share a function between them.

### DIFF
--- a/examples/medplum-demo-bots/medplum.config.json
+++ b/examples/medplum-demo-bots/medplum.config.json
@@ -74,13 +74,13 @@
     },
     {
       "name": "welcome-patient",
-      "id": "c7c4c2a2-d5fc-43e9-b892-37b94149f541",
+      "id": "",
       "source": "./src/bots-with-shared-library/welcome-patient.ts",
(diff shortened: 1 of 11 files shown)


examples/medplum-demo-bots/medplum.config.json
```diff
@@ -71,6 +71,18 @@
...
+    },
+    {
+      "name": "welcome-patient",
+      "id": "c7c4c2a2-d5fc-43e9-b892-37b94149f541",
+      "source": "./src/bots-with-shared-library/welcome-patient.ts",
+      "dist": "./dist/bots-with-shared-library/welcome-patient.js"
+    },
+    {
+      "name": "goodbye-patient",
+      "id": "3d123930-aa11-49eb-96fa-ac928d3eac3d",
```
rahul1, line 83: For the check-in version, just put an empty id `""`

rahul1 (approved):

PR #3443, opened 2023-12-01
Fixes #3391 - github action for prepare-release.sh

### LINKED ISSUE #3391: Create Github Action for `prepare-release.sh`
TIL about [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) and [Manually running a workflow](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow)

That would be a great fit for our `scripts/prepare-release.sh` script.  It is mostly automated, but you have to be on `main` with a clean branch, and you have to have the Github command line tools setup.  It's 95% fine, but 5% annoying.

We'll need to research a couple things:
* Need to make sure that a Github Action is allowed to create a new branch, create a PR, and create a draft release
* Need to make sure that the Github Action is only runnable by approved team members
* Might need to fix some bash issues -- I think that the `sed` call might not be fully cross platform agnostic

codyebberson:
Bonus: using `MEDPLUM_BOT_GITHUB_ACCESS_TOKEN` in `add-issue-to-project.yml` so it shows up as @medplumbot  rather than @codyebberson

### DIFF
--- a/.github/workflows/add-issue-to-project.yml
+++ b/.github/workflows/add-issue-to-project.yml
@@ -13,9 +13,9 @@ jobs:
       - uses: actions/add-to-project@v0.5.0
         with:
           project-url: https://github.com/orgs/medplum/projects/1
-          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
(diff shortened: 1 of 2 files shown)


ThatOneBro (approved):
Nice 👍

PR #3464, opened 2023-12-03
Update undefined string condition in ResourcePropertyInputs

### LINKED ISSUE #3463: Attachement Array Input not working for photos in resource create form
The ResourceProperty Input in the Resource Create Form contains the incorrect component for attachment arrays.
An example of it can be seen in the photo field in the new patient create form.

image

Here, despite the photos field being an Attachment array type, the rendered content seems to be a ResourceArrayInput containing multiple AttachmentInput.

The issue seems to be in the checks in ResourcePropertyInput.tsx

https://github.com/medplum/medplum/blob/63d62d922ceb3271582cbba3afe5a77d5008a52c/packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx#L39-L51

jogeshwar01:
Fixes #3463 

The problem is because a **string propertyType** is returned from getValueAndType which is used in BackboneElementInput.

https://github.com/medplum/medplum/blob/63d62d922ceb3271582cbba3afe5a77d5008a52c/packages/react/src/ResourcePropertyDisplay/ResourcePropertyDisplay.utils.ts#L13

https://github.com/medplum/medplum/blob/63d62d922ceb3271582cbba3afe5a77d5008a52c/packages/react/src/BackboneElementInput/BackboneElementInput.tsx#L50

And therefore this check in the ResourcePropertyInput is not enough and would fail for the **'undefined' string** _(not type)_
https://github.com/medplum/medplum/blob/63d62d922ceb3271582cbba3afe5a77d5008a52c/packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx#L41

I have updated it to now check for both undefined type as well as string.
Now this would behave as an AttachmentArrayInput instead of being a ResourceArrayInput with multiple AttachmentInputs.
This also prevents multiple empty uploads as seen in the first picture below.

Before -
image

After -
image

### DIFF
--- a/packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
+++ b/packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
@@ -38,7 +38,10 @@ export interface ResourcePropertyInputProps {
 
 export function ResourcePropertyInput(props: ResourcePropertyInputProps): JSX.Element {
   const property = props.property;
-  const propertyType = props.defaultPropertyType ?? property.type[0].code;
+  const propertyType =
+    props.defaultPropertyType && props.defaultPropertyType !== 'undefined'
+      ? props.defaultPropertyType
+      : property.type[0].code;
   const name = props.name;
   const value = props.defaultValue;
 


mattwiller:
Would it be better to update `getValueAndType()` instead, so it returns e.g. `[undefined, undefined]` or `[undefined, '']`?

jogeshwar01:
@mattwiller Yes, we can do that as well.
Just the thing is there are a few more places where we have **'undefined'** string as the property type being returned. 

https://github.com/medplum/medplum/blob/39cc849b421b9aa0b00b4324252bcc97bdb978ca/packages/core/src/fhirpath/utils.ts#L20-L22

https://github.com/medplum/medplum/blob/39cc849b421b9aa0b00b4324252bcc97bdb978ca/packages/core/src/fhirpath/utils.ts#L102

I thought there might have been a reason the string undefined was being used in the first place. So this seemed like a quick fix to me. If thats not the case, we can change it to `[undefined, undefined]` or `[undefined, '']` as you suggested.

PR #3465, opened 2023-12-03
Agent request response

codyebberson:
Support "request" and "response" pattern for Agent push messages.

Also includes proper types for a bunch of the message types -- the first implementation included too many `any` types, so this tightens all of that up.

### DIFF
--- a/packages/server/src/__mocks__/ioredis.ts
+++ b/packages/server/src/__mocks__/ioredis.ts
@@ -1,16 +1,16 @@
+const values = new Map<string, string>();
 const subscribers = new Map<string, Set<Redis>>();
 
 class Redis {
-  private values: Map<string, string>;
-  private listeners: Map<string, ((...args: any[]) => void)[]>;
+  private callback?: (...args: any[]) => void;
 
-  constructor(readonly config: any) {
-    this.values = new Map<string, string>();
-    this.listeners = new Map<string, ((...args: any[]) => void)[]>();
-  }
+  constructor(readonly config: any) {}
 
   on(event: string, callback: (...args: any[]) => void): void {
-    this.listeners.set(event, [callback]);
+    if (event !== 'message') {
+      throw new Error('ioredis mock only supports message events');
+    }
+    this.callback = callback;
   }
 
   async ping(): Promise<string> {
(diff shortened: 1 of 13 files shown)


packages/server/src/__mocks__/ioredis.ts
```diff
@@ -76,13 +76,13 @@ class Redis {
     }
   }
 
-  duplicate(): this {
-    // return new Redis(this.config);
-    return this;
+  duplicate(): Redis {
```
ThatOneBro, line 79: Nice, I was meaning to do this in my PR but forgot, glad to see you picked it up 👍

packages/server/src/__mocks__/ioredis.ts
```diff
@@ -1,57 +1,57 @@
...
-    this.values.delete(key);
+    values.delete(key);
   }
 
   async publish(channel: string, message: string): Promise<void> {
-    const listeners = this.listeners.get('message');
-    if (listeners) {
-      listeners.forEach((listener) => {
-        listener(channel, message);
+    const set = subscribers.get(channel);
```
ThatOneBro, line 51: Nice, good change

packages/server/src/agent/websockets.ts
```diff
@@ -29,24 +36,39 @@ export async function handleAgentConnection(socket: ws.WebSocket, request: Incom
...
             break;
 
+          // @ts-expect-error - Deprecated message type
           case 'transmit':
+          case 'agent:transmit:request':
             await handleTransmit(command);
             break;
+
+          case 'agent:transmit:response':
+            if (command.callback) {
```
ThatOneBro, line 54: Nice, this is a very clean way to handle this req/res model

packages/server/src/fhir/operations/agentpush.ts
```diff
@@ -12,6 +21,7 @@ export interface AgentPushParameters {
   body: string;
   contentType: string;
   destination: string;
+  waitForResponse?: boolean;
```
ThatOneBro, line 24: 👍

packages/server/src/fhir/operations/agentpush.ts
```diff
@@ -57,8 +67,49 @@ export const agentPushHandler = asyncWrap(async (req: Request, res: Response) =>
...
+    sendOutcome(res, allOk);
+    return;
+  }
+
+  // Otherwise, open a new redis connection in "subscribe" state
+  // message.callback = getReferenceString(agent) + '-' + randomUUID();
+  message.callback = randomUUID();
+
+  const redisSubscriber = getRedis().duplicate();
+  await redisSubscriber.subscribe(message.callback);
```
ThatOneBro, line 89: Cool, this is a pretty interesting way to use subscriptions. I'm curious what the overhead of duplicating / subscribing to a new topic in a high throughput HL7 session, but I think this probably works for how it's currently being used.

packages/server/src/fhir/operations/agentpush.ts
```diff
@@ -57,8 +67,49 @@ export const agentPushHandler = asyncWrap(async (req: Request, res: Response) =>
...
+    const response = JSON.parse(message);
+    res.status(200).type(response.contentType).send(response.body);
+    cleanup();
+  });
+
+  // Create a timer for 5 seconds for timeout
+  const timer = setTimeout(() => {
+    cleanup();
+    sendOutcome(res, badRequest('Timeout'));
+  }, 5000);
```
ThatOneBro, line 100: Makes sense, 5 seconds seems like a sensible req timeout.

ThatOneBro (commented):
So far it seems like a good solution, just some general comments and one open question about scalability, maybe something to think about longer term

packages/server/src/fhir/operations/agentpush.ts
codyebberson: Yeah, agree.  At minimum we should probably use a more sane channel name than `randomUUID`.

ChatGPT claims that Redis can handle this performantly.... I guess we'll see?

ThatOneBro: Yeah I can believe it.

Maybe one cool thing would be to set up a granular profiling system for testing particular parts / flows of the system that is easy to expand as the codebase expands. That would be super cool. Not sure what that would look like yet but I'll be thinking about it

PR #3474, opened 2023-12-04
Fix non-ASCII characters in definition file

mattwiller:


codyebberson:
TIL about `git diff --word-diff-regex=.`

image

codyebberson:
I think this one was supposed to be an apostrophe:

image

codyebberson (approved):
I noticed one case of non-whitespace, but it's all in `text` which we ignore anyway.

codyebberson:
image

PR #3477, opened 2023-12-05
Fixes #3476 - disjunctions in _filter param

### LINKED ISSUE #3476: _filter disjunction doesn't work with lookup tables
## Repro

Create 2 patients
```json
{
  "resourceType": "Patient",
  "name": [
    {
      "use": "official",
      "given": [
        "Alice"
      ],
      "family": "Doe"
    }
  ],
  "birthDate": "1974-12-25"
}
```

```json
{
  "resourceType": "Patient",
  "name": [
    {
      "given": [
        "George"
      ],
      "family": "Washington"
    }
  ]
}
```

Run search query 
```curl
Patient?_filter=birthdate eq 1974-12-25 or name co George
```

## Expected Behavior
The search should return both Patients

## Observed Behavior
Search only returns George Washington

## Root Cause
I'm not 100% sure, but I think this is the issue
- [This code](https://github.com/medplum/medplum/blob/63d62d922ceb3271582cbba3afe5a77d5008a52c/packages/server/src/fhir/search.ts#L654-L665) creates a `Disjunction` from the `or`
- In SQL, this translates to an `OR` query
- However, `name` is in the `Patient_token` lookup table, rather than the base `Patient` table, so the proper query would some kind of join

codyebberson:


reshmakh (approved):

rahul1:
@codyebberson will this also work in the case where one search parameter is on the lookup table, and one is on the base resource table?

PR #3486, opened 2023-12-05
Fixes #1918 - multiple projects sharing googleClientId

### LINKED ISSUE #1918: Better handling of shared `googleClientId`
Currently, if >=2 projects share a `googleClientId`, the server just returns the first one in a search. We have some users who share a `googleClientId` between dev, staging, and prod projects, this behavior is confusing and potentially unpredictable when using googleAuth.

Some options for more explicit ways of handling this: 
1. If 2 or more projects share a `googleClientId`, present users with a profile selector screen, as in Medplum app
2. Require all projects to have a unique `googleClientId`, and return an error when setting the `googleClientId` on the `Project` (as well as on login?)

[This is the code](https://github.com/medplum/medplum/blob/main/packages/server/src/auth/google.ts#L154-L159) where we make this selection.

codyebberson:


packages/server/src/auth/google.test.ts
```diff
@@ -204,31 +204,34 @@ describe('Google Auth', () => {
     expect(res2.body.code).toBeDefined();
   });
 
-  test('Custom Google client success', async () => {
```
codyebberson, line 207: This test was an exact line-for-line duplicate of the test immediately above it.  Repurposing the real estate for new test.

rahul1 (approved):

PR #3488, opened 2023-12-05
Implement ConceptMap translate operation

### LINKED ISSUE #3344: Implement ConceptMap $translate

mattwiller:
Adding the `$translate` operation on `ConceptMap` resources, available either as:
- `POST ConceptMap/:id/$translate` to translate using a given ConceptMap
- `POST ConceptMap/$translate` to use the `url` or `source` parameters to look up the ConceptMap

Currently, only simple enumerated mappings are supported via `ConceptMap.group`

Resolves #3344

### DIFF
--- a/packages/server/src/fhir/operations/README.md
+++ b/packages/server/src/fhir/operations/README.md
@@ -27,7 +27,8 @@ Utility functions that use the `OperationDefinition` to automate implementation 
 
 ```ts
 import { parseInputParameters, sendOutputParameters } from './utils/parameters';
-import { created, OperationDefinition, Reference } from '@medplum/core';
+import { created } from '@medplum/core';
+import { Reference, OperationDefinition } from '@medplum/fhirtypes';
 import { Request, Response } from 'express';
 
 const operation: OperationDefinition = {
@@ -57,7 +58,7 @@ interface ProjectInitParameters {
   ownerEmail?: string;
 }
 
-export function projectInitHandler(req: Request, res: Response): void {
+export async function projectInitHandler(req: Request, res: Response): Promise<void> {
   const params = parseInputParameters<ProjectInitParameters>(operation, req);
 
   // Handle operation business logic...
--- a/packages/server/src/fhir/operations/conceptmaptranslate.ts
+++ b/packages/server/src/fhir/operations/conceptmaptranslate.ts
@@ -129,22 +129,26 @@ function translateCodes(sourceCodes: Record<string, string[]>, groups: ConceptMa
   const matches: Match[] = [];
   for (const [system, codes] of Object.entries(sourceCodes)) {
     for (const group of groups.filter((g) => g.source === system)) {
-      const mappings = group.element?.filter((m) => codes.includes(m.code as string));
+      let mappings: Match[] | undefined = group.element
+        ?.filter((m) => codes.includes(m.code as string))
+        .flatMap(
+          (m) =>
+            m.target?.map((target) => ({
+              equivalence: target.equivalence,
+              concept: {
+                system: group.target,
+                code: target.code,
+                display: target.display,
+              },
+            })) ?? []
+        );
+
       if (!mappings?.length) {
-        continue;
+        mappings = handleUnmappedCodes(codes, group);
       }
-      const mappedCodes: Match[] = mappings.flatMap(
-        (m) =>
-          m.target?.map((target) => ({
-            equivalence: target.equivalence,
-            concept: {
-              system: group.target,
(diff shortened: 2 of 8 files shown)


packages/server/src/fhir/operations/README.md
```diff
@@ -57,7 +58,7 @@ interface ProjectInitParameters {
   ownerEmail?: string;
 }
 
-export function projectInitHandler(req: Request, res: Response): void {
+export async function projectInitHandler(req: Request, res: Response): Promise<void> {
```
codyebberson, line 61: Does this need to be `async`?

packages/server/src/fhir/operations/conceptmaptranslate.ts
```diff
@@ -0,0 +1,182 @@
...
+import { Operator, allOk, badRequest, notFound } from '@medplum/core';
+import { Request, Response } from 'express';
+import { getOperationDefinition } from './definitions';
+import { CodeableConcept, Coding, ConceptMap, ConceptMapGroup, OperationOutcome } from '@medplum/fhirtypes';
+import { getAuthenticatedContext } from '../../context';
+import { sendOutcome } from '../outcomes';
+
+const operation = getOperationDefinition('ConceptMap', 'translate');
+
+type ConceptMapTranslateParameters = {
```
codyebberson, line 11: Curious why `type` over `interface`?

Not a big deal, but hopefully as a team we can formulate a framework on how to choose.

In general, I've defaulted to `interface` in the past based on recommendations from the TypeScript's page on performance: https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

packages/server/src/fhir/operations/conceptmaptranslate.ts
```diff
@@ -0,0 +1,182 @@
...
+};
+
+export async function conceptMapTranslateHandler(req: Request, res: Response): Promise<void> {
+  const params = parseInputParameters<ConceptMapTranslateParameters>(operation, req);
+
+  const map = await lookupConceptMap(params, req.params.id);
+  if (isOutcome(map)) {
+    sendOutcome(res, map);
+    return;
+  } else if (!map.group) {
```
codyebberson, line 39: Another super minor thing, just for aligning coding styles, is using `else` when the parent block uses `return`  Functionally, totally equivalent, so this is purely a style question.  It'd just be nice to be consistent throughout the code base.

packages/server/src/fhir/operations/conceptmaptranslate.ts
```diff
@@ -0,0 +1,182 @@
...
+    sendOutcome(res, badRequest('ConceptMap does not specify a mapping group', 'ConceptMap.group'));
```
codyebberson, line 40: Excellent use of 2nd arg to `badRequest()`! 👍

codyebberson (approved):
Nice, excellent work @mattwiller - this looks great.

A handful of style questions, nothing serious or material.  The main motivation of the questions is to align on styles and philosophies so we can maintain a consistent feel.

packages/server/src/fhir/operations/README.md
mattwiller: Yep, it uses `await` on line 68

packages/server/src/fhir/operations/conceptmaptranslate.ts
mattwiller: I generally favor an early-return style for error cases, leaving the main code path unindented to highlight the "happy path" when reading the code — hence no `else` here.  Thoughts on that general idea?

packages/server/src/fhir/operations/conceptmaptranslate.ts
mattwiller: I was going off of [this line of thinking](https://blog.logrocket.com/types-vs-interfaces-typescript/#when-use-types-vs-interfaces), that `type` is more expressive and in theory works exactly the same for simple cases like this one.  I see the performance argument, but my read was that it only mattered when doing definition merging or extending, which isn't generally a major use case from what I've seen

packages/server/src/fhir/operations/conceptmaptranslate.ts
mattwiller: Ah, I see — I do like to group error cases into an `if`/`else if` as well, to visually cluster them together and further highlight the happy path.  If you'd prefer it the other way, we could enforce that style with https://eslint.org/docs/latest/rules/no-else-return by setting `allowElseIf: false`

PR #3501, opened 2023-12-06
Dependency upgrades

codyebberson:


.github/workflows/upgrade-dependencies.yml
```diff
@@ -0,0 +1,24 @@
+name: Upgrade dependencies
+
+on: workflow_dispatch
```
ThatOneBro, line 3: Nice 👍

ThatOneBro (approved):
Nice, looks like a sane way to do upgrades

PR #3516, opened 2023-12-08
fix-3515 `MemoryRepository.clear` to delete all resources and history

### LINKED ISSUE #3515: `@medplum/mock` easy way to clear memory store on `MockClient`?
When testing, it could be useful to have the ability to quickly clear/reset the memory store on the mock client. Today, this can be achieved by keeping track of the created resources and deleting them all at the end of the test.
```typescript
const medplum = new MockClient();

it('does the thing', async() => {
  const resource1 = await medplum.createResource(...);
  const resource2 = await medplum.createResource(...);
  const resource3 = await medplum.createResource(...);

  // ...

  await Promise.all([resource1, resource2, resource3].map(r => medplum.deleteResource(r)));
});
```

Having a mock client globally defined is not ideal but in some cases, without making large changes to an existing codebase, it may be the only option. In these cases, having the ability to quickly and easily clear all resources would be helpful in preserving memory and in some cases resolving bugs in tests that share identifiers.

I propose adding a `reset` or `clear` method on the MemoryRepository that looks something like the following:
```typescript
export class MemoryRepository extends BaseRepository implements FhirRepository {
  private readonly resources: Map<string, Map<string, Resource>>;
  private readonly history: Map<string, Map<string, Resource[]>>;

  constructor() {
    super();
    this.resources = new Map();
    this.history = new Map();
  }

  clear(): void {
    this.resources.clear();
    this.history.clear();
  }
// ...
```

This would enable developers using the `MockClient` to hook into a before each or after each (assuming their test runner exposes it) or simply call `clear` at the end of the test without needing to enumerate each of the resources created by the test.

```typescript
const medplum = new MockClient();

it('does the thing', async() => {
  const resource1 = await medplum.createResource(...);
  const resource2 = await medplum.createResource(...);
  const resource3 = await medplum.createResource(...);

  // ...

  medplum.repo.clear();
});
```
OR
```typescript
const medplum = new MockClient();

afterEach(() => {
  medplum.repo.clear();
});

it('does the thing', async() => {
  const resource1 = await medplum.createResource(...);
  const resource2 = await medplum.createResource(...);
  const resource3 = await medplum.createResource(...);

  // ...
});
```

dillonstreator:
Fixes #3515

### DIFF
--- a/packages/fhir-router/src/repo.test.ts
+++ b/packages/fhir-router/src/repo.test.ts
@@ -111,4 +111,29 @@ describe('MemoryRepository', () => {
     const bundle = await repo.search({ resourceType: 'Patient', sortRules: [{ code: 'xyz' }] });
     expect(bundle.entry).toBeDefined();
   });
+
+  test('clears all resources', async () => {
+    const patientCount = 10;
+    for (let i = 0; i < patientCount; i++) {
+      await repo.createResource<Patient>({ resourceType: 'Patient' });
+    }
+    const observationCount = 11;
+    for (let i = 0; i < observationCount; i++) {
+      await repo.createResource<Observation>({ resourceType: 'Observation' });
+    }
+
+    const [patientsBefore, observationsBefore] = await Promise.all([
+      repo.searchResources<Patient>({ resourceType: 'Patient' }),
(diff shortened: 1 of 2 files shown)


mattwiller (approved):
Thanks @dillonstreator!

PR #3524, opened 2023-12-08
Fix npm ci on Windows runner

codyebberson:
Last publish step failed: https://github.com/medplum/medplum/actions/runs/7146451123/job/19465886902

I tried re-running the job, but it failed in the same way.

I interpret that as an issue with npm node_modules caching.  We've never experienced that on Linux builders, so I'm assuming it's a Windows issue.


ThatOneBro (approved):
Approving, no caching hurts but if it's only on `publish` then I think it's ok

Issue #3529, opened 2023-12-11
Add a comment to FooMedical to show how to run it against localhost


ksmith94:
User Driven: https://discord.com/channels/905144809105260605/1113936455954346005/1146065474694692955

Adding a commented line setting `baseUrl: http://localhost:3000` in the `MedplumClient` constructor would make it clear how to run locally.

PR #3530, opened 2023-12-11
Add comment to show how to set baseUrl in FooMedical

ksmith94:
This will clarify how to run FooMedical locally. Related to issue #3529

### DIFF
--- a/examples/foomedical/src/main.tsx
+++ b/examples/foomedical/src/main.tsx
@@ -7,6 +7,8 @@ import { BrowserRouter } from 'react-router-dom';
 import { App } from './App';
 
 const medplum = new MedplumClient({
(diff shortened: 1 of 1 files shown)


examples/foomedical/src/main.tsx
```diff
@@ -7,6 +7,8 @@ import { BrowserRouter } from 'react-router-dom';
 import { App } from './App';
 
 const medplum = new MedplumClient({
+  // To run FooMedical locally, you can set the baseURL in this constructor
+  // baseUrl: http://localhost:3000
```
rahul1, line 11: I think the default port for the api is 8103
```suggestion
  // baseUrl: http://localhost:8103
```

rahul1 (changes requested):

rahul1 (approved):

PR #3532, opened 2023-12-11
`getExternalAuthRedirectUri` Should Respect Scope Parameter

### LINKED ISSUE #3531: `getExternalAuthRedirectUri` Should Respect Scope Parameter
The `getExternalAuthRedirectUri` method within `MedplumClient` does not respect the `BaseLoginRequest.scope` parameter, if it is set. It looks like the `url.searchParams.set('scope', 'openid profile email');` should be changed to `url.searchParams.set('scope', loginRequest.scope ?? 'openid profile email');`.

https://github.com/medplum/medplum/blob/70903e071e13efb44ce38517bd16f18d2e9a8c0f/packages/core/src/client.ts#L1183-L1194

atsjj:
Closes #3531

* Scope defaults to `openid profile email` if `loginRequest.scope` is undefined, otherwise it is set to the provided value.
* Added test to assert this is true

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -544,6 +544,22 @@ describe('Client', () => {
         )
       ).toThrow();
     });
+
+    test('should respect scope parameter', async () => {
+      const result = client.getExternalAuthRedirectUri(
+        'https://auth.example.com/authorize',
+        'external-client-123',
+        'https://me.example.com',
+        {
+          clientId: 'medplum-client-123',
+          scope: "profile email foo",
+        },
+        false
+      );
+
+      const { searchParams } = new URL(result);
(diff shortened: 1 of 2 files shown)


codyebberson (approved):

packages/core/src/client.test.ts
```diff
@@ -544,6 +544,22 @@ describe('Client', () => {
...
     });
+
+    test('should respect scope parameter', async () => {
+      const result = client.getExternalAuthRedirectUri(
+        'https://auth.example.com/authorize',
+        'external-client-123',
+        'https://me.example.com',
+        {
+          clientId: 'medplum-client-123',
+          scope: "profile email foo",
```
codyebberson, line 555:
```suggestion
          scope: 'profile email foo',
```

PR #3535, opened 2023-12-11
Profile snapshot operation

mattwiller:
Adding experimental support for the `StructureDefinition/$snapshot` operation, which populates the `snapshot` necessary for resource validation from a set of source profile differentials.  Locally, this take ~10s to run to generate the US Core Blood pressure profile, which recursively specializes `USCoreBloodPressure` -> `USCoreVitalSigns` -> `vitalsigns` -> `Observation`


codyebberson:
Closing for now, we can revisit when we get serious about authoring profiles.

Issue #3424, opened 2023-11-30
Use heartbeat system for all websocket connections


codyebberson:
1. FHIRCast (already exists)
2. FHIR Subscriptions
3. Agent

ThatOneBro:
Closed as completed in #3424

PR #3536, opened 2023-12-12
feat(subscriptions): add `WebSocket` subscriptions

ThatOneBro:
## TODO:
- [x] Cleanup some comments
- [x] Implement `bind-with-token` command over `WebSockets`
- [x] Add example for creating a subscription via `WebSockets`
- [x] Edit the readme for the example, create new example standalone repo clone

### In another PR:
- [ ] Write docs on how to use subscriptions via `WebSockets`
- [ ] Add some more tests off the happy path
- [ ] Add `handshake` message for every subscription added to a client's "channel"
- [x] Add `heartbeat` (see: #3424)

Included in this PR is a demo for `WebSocket` subscriptions. Here is what it looks like:
Screenshot 2023-12-20 at 11 03 10 AM

There are roughly 7 steps contained within the demo for the `WebSocket` subscription lifecycle, simulating a setup for a Patient-Practitioner chat, based on `Communication` resources:

trimmed_demo

1. Create 2 subscriptions. One `Subscription` will subscribe to incoming messages and one will subscribe to outgoing messages from the patient.
2. Call `Subscription/$get-ws-binding-token` operation for each `Subscription` created. This gets us a token to bind to after establishing a connection via a `WebSocket`, which is how the server knows what notifications to send over this `WebSocket` and what `Subscriptions` the user on the other side of this connection cares about.
3. Connect to the server via the `WebSockets` url: `{server_base_url}/ws/subscriptions-r4`.
4. Send a `bind-with-token` message over `WebSocket` for each token we received in step 2.
5. Create `Communication` resources matching our criteria for incoming or outgoing messages. We receive them over our `WebSocket` connection as `Bundle` resources containing a `SubscriptionStatus` resource explaining which `Subscription` originated this notification, and the resources that matched this `Subscription`.
6. Disconnect from `WebSocket` when finished.
7. Delete the `Subscription` resources we created to clean up.

Steps 2-4 are treated as one step from the perspective of the demo, as they are all executed in sequence after the user presses the `Connect via WebSocket` button.

The demo also lets the user see the formatted `Bundle` contained in the notification:
Screenshot 2023-12-20 at 11 57 26 AM

Overall this demo should serve as a good rough guide for implementers on how to actually use `WebSocket` subscriptions in their current state.

### Update to Demo
Screenshot 2023-12-22 at 5 20 33 PM

Demo is now tracking when a message is sent and received based on the `Communication.status` and `Communication.received`. The client will update the status once it receives the message as a subscription notification and the status is not `completed` or there is no `received` timestamp yet.


packages/server/src/fhir/operations/getwsbindingtoken.ts
```diff
@@ -0,0 +1,66 @@
...
+    ],
+  } satisfies Parameters;
+
+  // Create tentative binding for this user
+  // When user connects to WebSocket, get all of the tentative bindings
+  const subscriptionId = req.params.id;
+  const bindingsKey = `::subscriptions/r4::bindings::${textDecoder.decode(
+    await crypto.subtle.digest('sha-256', textEncoder.encode(accessToken))
+  )}`;
+  const existingBindings = await redis.get(bindingsKey);
```
codyebberson, line 58: Rather than doing a `get` and `set`, should we just use Redis Sets? https://redis.io/docs/data-types/sets/

packages/server/src/fhir/operations/getwsbindingtoken.ts
```diff
@@ -0,0 +1,66 @@
...
+      {
+        name: 'websocket-url',
+        valueUrl: `${baseUrl.replace('http://', 'ws://').replace('https://', 'wss://')}ws/subscriptions-r4`,
+      },
+    ],
+  } satisfies Parameters;
+
+  // Create tentative binding for this user
+  // When user connects to WebSocket, get all of the tentative bindings
+  const subscriptionId = req.params.id;
```
codyebberson, line 54: We should probably validate that this subscription ID is valid.  Even if just a quick `repo.readResource()` to make sure it exists.

packages/server/src/fhir/operations/getwsbindingtoken.ts
```diff
@@ -0,0 +1,66 @@
...
+    resourceType: 'Parameters',
+    parameter: [
+      {
+        name: 'token',
+        valueString: accessToken,
+      },
+      {
+        name: 'expiration',
+        valueDateTime: new Date(
+          (typeof tokenPayload.exp === 'string' ? parseInt(tokenPayload.exp, 10) : tokenPayload.exp) * 1000
```
codyebberson, line 42: Is it valid for `exp` to be a string?

packages/server/src/fhir/repo.ts
```diff
@@ -257,6 +258,8 @@ export class Repository extends BaseRepository implements FhirRepository {
 
     const resource = JSON.parse(rows[0].content as string) as T;
     await setCacheEntry(resource);
+    // TODO(ThatOneBro 01 Dec 23): Check if we can do this in the background
+    // setCacheEntry(resource).catch(console.error);
```
codyebberson, line 262: Interesting idea... 

It should be background-able / parallel-izable within the scope of the "read" operation.  I think we would still need to `await` before the operation could be considered "complete" though.  To do that would require some refactoring...  you'd basically need some representation of "promises that need to be settled before the operation is done".

packages/server/src/fhir/operations/getwsbindingtoken.ts
```diff
@@ -0,0 +1,66 @@
...
+        valueUrl: `${baseUrl.replace('http://', 'ws://').replace('https://', 'wss://')}ws/subscriptions-r4`,
+      },
+    ],
+  } satisfies Parameters;
+
+  // Create tentative binding for this user
+  // When user connects to WebSocket, get all of the tentative bindings
+  const subscriptionId = req.params.id;
+  const bindingsKey = `::subscriptions/r4::bindings::${textDecoder.decode(
+    await crypto.subtle.digest('sha-256', textEncoder.encode(accessToken))
```
codyebberson, line 56: What is the reason for encrypting the access token here?  Redis already contains sensitive data (including this exact access token), so I'm not sure we gain much security protection vs the complexity cost.

We could also use the `Login.id`, which should be 1:1 with the access token.

packages/server/src/workers/subscription.ts
```diff
@@ -268,6 +275,14 @@ async function getSubscriptions(resource: Resource): Promise<Subscription[]> {
       },
     ],
   });
+  // TODO: Can we include project ID in this key to make searches / deletes faster?
```
codyebberson, line 278: Yes, indexing by Project ID would make sense here.

codyebberson (commented):
Nice, overall this looks really good.  No structural concerns, just a bunch of minor questions.

I'll follow up offline about the "::" in the Redis keys... I feel like we need to develop a consistent philosophy on key naming conventions.  I couldn't find anything online about prefixing with double colons.  Using a single colon as a delimiter does seem to have some semantic value in Redis visualization tools, but using as a prefix and using 2 seems unnecessary.

Thanks @ThatOneBro

packages/server/src/fhir/operations/getwsbindingtoken.ts
codyebberson:
```
    console.log(ctx.login.id);
```

packages/server/src/fhir/operations/getwsbindingtoken.ts
ThatOneBro: As far as I can tell it is allowed to be a string based on the types

packages/server/src/fhir/repo.ts
ThatOneBro: Yeah that makes sense, maybe I can take a look at doing that...

ThatOneBro: Think it might not make much sense to do this now that I'm looking at it again. We probably won't get much out of it and will probably add a lot of complexity. Maybe worth looking at it again later but I'll probably just remove the TODO for now.

codyebberson (commented):
Nice, this is awesome!  This is going to unlock a bunch of really great use cases.

2 things I think we should change:

1. Unless necessary, I think we should revert the behavior changes in `parseReference()` in `packages/core/src/utils.ts`
2. As discussed offline, I'd like to use more standard naming convention for Redis keys, such as single colon `:` and/or using `medplum:` as the key prefix

Those aside, this looks great, and is basically good to go.  Thank you! 🙏 👍

packages/core/src/utils.ts
```diff
@@ -76,15 +77,13 @@ export function resolveId(input: Reference | Resource | undefined): string | und
...
-export function parseReference(reference: Reference | undefined): [ResourceType, string] | undefined {
+export function parseReference(reference: Reference | undefined): [ResourceType, string] {
   if (reference?.reference === undefined) {
-    return undefined;
+    throw new OperationOutcomeError(validationError('Reference missing reference property.'));
   }
   const [type, id] = reference.reference.split('/');
   if (type === '' || id === '' || id === undefined) {
-    return undefined;
+    throw new OperationOutcomeError(validationError('Unable to parse reference string.'));
```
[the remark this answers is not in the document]
ThatOneBro: I'm not sure anyone's actually using this function so I thought it would be okay to make this breaking change. I did a check through the codebase and no one is using it since I added it. 

It's also pretty much useless with the old signature since since the return value can be `undefined` if you destructure the return, it gives a TS error at compile time and a runtime error if we ever return `undefined` in the case of an invalid reference.

codyebberson (approved):

PR #3537, opened 2023-12-12
Create DB tables for terminology data

mattwiller:


mattwiller:
> How are you thinking about sequencing this with the code that uses the tables? I'm kinda inclined to ship them both together, unless we have very high confidence that the schema is stable and won't change

@codyebberson Having tested it end to end, I feel good about this schema being the final version.  Are there any use cases you'd like me to verify with the new schema before we commit to it?

mattwiller:
Combined with #3589

PR #3538, opened 2023-12-12
Fixes #3469 - handle resolved choice-of-type property names

### LINKED ISSUE #3469: New schema diff view does not work with Extension values
We added a new resource diff view in https://github.com/medplum/medplum/pull/3387

Overall, it has been a big improvement, especially on big resources.

Unfortunately, there are a few cases that trigger exceptions and the `<ErrorBoundary>` view.

One reproducible example is Extension values

image

codyebberson:
Follow-up to https://github.com/medplum/medplum/pull/3508

### DIFF
--- a/packages/core/src/fhirpath/utils.ts
+++ b/packages/core/src/fhirpath/utils.ts
@@ -1,5 +1,5 @@
 import { Coding, Period, Quantity } from '@medplum/fhirtypes';
-import { getElementDefinition, isResource, PropertyType, TypedValue } from '../types';
+import { PropertyType, TypedValue, getElementDefinition, isResource } from '../types';
 import { InternalSchemaElement } from '../typeschema/types';
 import { capitalize, isEmpty } from '../utils';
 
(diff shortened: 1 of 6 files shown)


packages/core/src/types.ts
```diff
@@ -324,7 +323,25 @@ export function getElementDefinition(typeName: string, propertyName: string): In
...
+
+  // The propertyName can be a "choice of type" property, such as "value[x]", but in resolved form "valueString".
+  // So we need to iterate through all the elements and find the one that matches.
+  for (const [elementName, element] of Object.entries(typeSchema.elements)) {
+    for (const type of element.type) {
+      if (elementName.replace('[x]', capitalize(type.code)) === propertyName) {
+        return element;
+      }
+    }
+  }
```
mattwiller, line 341: This is kinda inefficient for a function that might be called in some hot loops; would it be possible to try and find the correct definition directly by doing something like this?

```ts
const fieldParts = propertyName.split(/(?=[A-Z])/g);
for (let i = 0; i < fieldParts.length; i++) {
  const testProperty = fieldParts.slice(0, i+1).join('') + '[x]';
  const element = typeSchema.elements[testProperty];
  if (element) {
    return element;
  }
}
```

mattwiller (approved):

packages/core/src/types.ts
codyebberson: Neat, that is definitely faster.  That was counterintuitive to me due to the regex split. 

Old (not supporting `valueBoolean`): 1028ms
As implemented in PR: 1549ms
As implemented in suggestion: 1122ms

Strategy: For each, run 5 times, take the best time.

Test harness: 

```ts
import { readJson } from '@medplum/definitions';
import { Bundle } from '@medplum/fhirtypes';
import { indexStructureDefinitionBundle } from '../typeschema/types';
import { getTypedPropertyValue } from './utils';

indexStructureDefinitionBundle(readJson('fhir/r4/profiles-types.json') as Bundle);
indexStructureDefinitionBundle(readJson('fhir/r4/profiles-resources.json') as Bundle);

const startTime = Date.now();
let count = 0;

for (let i = 0; i < 100000; i++) {
  for (const propName of ['valueBoolean', 'value[x]', 'asdf']) {
    if (getTypedPropertyValue({ type: 'Extension', value: { [propName]: true } }, propName)) {
      count++;
    }
  }
}

const endTime = Date.now();
const duration = endTime - startTime;
console.log('Elapsed time: ' + duration + 'ms');
console.log('Count: ' + count);
```

codyebberson: Nerd sniped by optimization problem: 1074ms

```ts
  for (let i = 0; i < propertyName.length; i++) {
    const c = propertyName[i];
    if (c >= 'A' && c <= 'Z') {
      const testProperty = propertyName.slice(0, i) + '[x]';
      const element = typeSchema.elements[testProperty];
      if (element) {
        return element;
      }
    }
  }
```

packages/core/src/fhirpath/utils.ts
```diff
@@ -105,15 +105,19 @@ function getTypedPropertyValueWithSchema(
     resultValue = input.value[path];
     resultType = types[0].code;
   } else {
+    // The path parameter can be in both "value[x]" form and "valueBoolean" form.
+    // So we need to use the element path to find the type.
+    const elementPath = (element.path.split('.').pop() as string).replace('[x]', '');
```
mattlong, line 110: When is the path parameter in either form?

If that's the case, should the usage of `path` elsewhere in this function be replaced to stay consistent?

packages/core/src/types.ts
mattwiller: Love it — that's a full 30% faster than the O(N^2) solution even for a small input! 🔥

packages/core/src/fhirpath/utils.ts
codyebberson: Good call - there was indeed a bug with the "primitive extensions" logic.  Fixed now.  Thanks

PR #3550, opened 2023-12-14
Fixes #3539 - Agent installer fixes

### LINKED ISSUE #3539: Provide a SHASUM value for the Agent installer at Release
It would be helpful to have a SHASUM value for the Agent installer package to verify the contents have not been modified or the download is completed successfully. I have ran into an issue where the installer was corrupt downloading onto Windows and was seeing some issues with the service starting. A deletion of the installation and re-download of the installer resolved the issue.

codyebberson:
* Fixes #3539 - Add SHA256SUM checksum to release files
* Upgrade [Shawl](https://github.com/mtkennerly/shawl) to version 1.4.0 for better log rotation
* Added check for incompatible version upgrade (see screenshot)
* On upgrade, fully remove old service and install new service

image

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -158,3 +158,10 @@ jobs:
               name: "medplum-agent-installer-" + packageJson.version + ".exe",
               data: await fs.readFileSync("packages/agent/medplum-agent-installer-" + packageJson.version + ".exe")
             });
+            await github.rest.repos.uploadReleaseAsset({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              release_id: release.data.id,
+              name: "medplum-agent-installer-" + packageJson.version + ".exe.sha256",
+              data: await fs.readFileSync("packages/agent/medplum-agent-installer-" + packageJson.version + ".exe.sha256")
+            });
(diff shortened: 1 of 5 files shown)


ThatOneBro (approved):
LGTM!

One question, is the reason for opting for direct download of `shawl` over `node-shawl` is because `node-shawl` is not caught up with the latest release?

codyebberson:
> One question, is the reason for opting for direct download of `shawl` over `node-shawl` is because `node-shawl` is not caught up with the latest release?

Yeah... Actually, I was the one who set up `node-shawl` in the first place.  At the time, we didn't have a very complex script like `build-agent-installer.sh`.  The NSIS installer script would simply reference the file out of `node_modules`

Over the past couple months, we added `build-agent-installer.sh` for a whole bunch of complexity.... Downloading JSign, setting up Digicert, signing the executables, etc.  So adding in a simple `wget` call for Shawl felt more appropriate here.

Versus the alternative, of downloading and updating `node-shawl`.  

There's now a separate cleanup task of winding down `node-shawl`, or handing it off to someone else.  I suppose we could still maintain it even if we're not using it.  It doesn't release that frequently.

ThatOneBro:
> Yeah... Actually, I was the one who set up `node-shawl` in the first place. At the time, we didn't have a very complex script like `build-agent-installer.sh`. The NSIS installer script would simply reference the file out of `node_modules`

Oh yeah, that's right. I totally forgot. Good call, this makes sense.

PR #3552, opened 2023-12-15
Reference input profile support

mattlong:
Required to support [USCoreImplantableDevice](https://hl7.org/fhir/us/core/STU5.0.1/StructureDefinition-us-core-implantable-device.html) for `Device.patient`

```
{
        "id": "Device.patient",
        "path": "Device.patient",
        "short": "Patient to whom Device is affixed",
        "definition": "Patient information, If the device is affixed to a person.",
        "requirements": "If the device is implanted in a patient, then need to associate the device to the patient.",
        "min": 1,
        "max": "1",
        "base": {
          "path": "Device.patient",
          "min": 0,
          "max": "1"
        },
        "type": [
          {
            "code": "Reference",
            "targetProfile": [
              "http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient"
            ]
          }
        ]
}
```

Addresses https://github.com/medplum/medplum/discussions/2006#discussioncomment-5880165. Be sure to see my comments on that part of the discussion


packages/react/src/ReferenceInput/ReferenceInput.stories.tsx
```diff
@@ -18,3 +22,31 @@ export const FreeText = (): JSX.Element => (
...
   </Document>
 );
+
+const FishPatientProfileSD = FishPatientResources.getFishPatientProfileSD();
+
+export const PatientProfileAndPatient = (): JSX.Element => {
+  const medplum = useMedplum();
+  const [loaded, setLoaded] = useState(false);
+  useEffect(() => {
+    (async (): Promise<boolean> => {
```
ThatOneBro, line 32: Will this create these resources every time the story is loaded?

packages/react/src/ReferenceInput/ReferenceInput.test.tsx
```diff
@@ -178,4 +179,60 @@ describe('ReferenceInput', () => {
...
+    });
+
+    // Refocus the input; "B" still as the last value
+    await act(async () => {
+      fireEvent.focus(screen.getByPlaceholderText('My placeholder'));
+    });
+
+    // Wait for the autocomplete timeout
+    await act(async () => {
+      jest.advanceTimersByTime(1000);
```
ThatOneBro, line 231: Interesting, do you have to advance timers in an act? That actually might explain some problems I had before

packages/react/src/ReferenceInput/ReferenceInput.test.tsx
```diff
@@ -178,4 +179,60 @@ describe('ReferenceInput', () => {
     expect(screen.getByPlaceholderText('Resource Type')).toBeInTheDocument();
     expect(screen.queryByTestId('reference-input-resource-type-select')).not.toBeInTheDocument();
   });
+
+  test.only('Handle profile target type', async () => {
```
codyebberson, line 183: Remove `.only`?

packages/react/src/ReferenceInput/ReferenceInput.tsx
```diff
@@ -16,75 +17,253 @@ export interface ReferenceInputProps {
...
 export function ReferenceInput(props: ReferenceInputProps): JSX.Element {
-  const targetTypes = getTargetTypes(props.targetTypes);
-  const initialResourceType = getInitialResourceType(props.defaultValue, targetTypes);
+  const medplum = useMedplum();
   const [value, setValue] = useState<Reference | undefined>(props.defaultValue);
-  const [resourceType, setResourceType] = useState<ResourceType | undefined>(initialResourceType);
+  const [targetTypes, setTargetTypes] = useState<TargetType[] | undefined>(() => createTargetTypes(props.targetTypes));
+  const [state, setState] = useState<{
+    targetType?: TargetType;
+    resourceType?: ResourceType;
```
codyebberson, line 44: Is `resourceType` always `targetType.resourceType`?  If so, then might be more straightforward to make it a derived value.

packages/react/src/ReferenceInput/ReferenceInput.tsx
```diff
@@ -16,75 +17,253 @@ export interface ReferenceInputProps {
...
+
+  const setValueHelper = useCallback(
+    (item: Resource | undefined) => {
+      const newValue = item ? createReference(item) : undefined;
+      setValue(newValue);
+      if (props.onChange) {
+        props.onChange(newValue);
+      }
+    },
+    // exhaustive deps wants this to be [props] instead of [props.onChange], but it's
```
codyebberson, line 124: You can keep it happy by splatting it `const { onChange } = props;`

packages/docs/static/img/blinky-the-fish.png
codyebberson, line 1: 🐟

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -49,12 +48,13 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
     setTimer(undefined);
 
     const value = inputRef.current?.value.trim() || '';
-    if (value === lastValueRef.current) {
-      // Nothing has changed, move on
+    if (value === lastValueRef.current && loadOptions === lastLoadOptionsRef.current) {
```
codyebberson, line 51: Does `loadOptions` ever change?

packages/react/src/ReferenceInput/ReferenceInput.tsx
ThatOneBro, line 124: My guess is that `props` could possibly be a new object on every render (like in the case of prop spreading) so they probably try to make you include `props` whenever you include `props.{{SOME_PROP}}`. In this case I think it's fine.

packages/react/src/ReferenceInput/ReferenceInput.tsx
```diff
@@ -16,75 +17,253 @@ export interface ReferenceInputProps {
...
+          .catch((reason) => {
+            console.error(reason);
+            return { ...tt, error: reason };
+          });
+      } else {
+        return Promise.resolve(tt);
+      }
+    });
+
+    Promise.all(newTargetTypePromises)
```
codyebberson, line 95: Did you test it with multiple profiles?  Auto-batching should pick it up and combine it into a single HTTP call.  "should"

codyebberson (commented):
Neat

image

Looks pretty good overall.  A handful of minor comments.  Overall lgtm

packages/react/src/ReferenceInput/ReferenceInput.test.tsx
mattlong: Ugh, I'm bad about that. Thanks

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
mattlong: Yes, this was exposed by the test I added where the target types include a normal resourceType, e.g. `Patient` and an profile of `Patient`, e.g. `USCorePatientProfile`. Switching from one type to the other would cause this to return early even though the search criteria/results should be different, i.e. filtering for patients with the `_profile` query param or not.

mattlong: The same would be true if there were two different profiles of the same resourceType listed which seems like a more likely scenario.

packages/react/src/ReferenceInput/ReferenceInput.tsx
mattlong: I've been going back and forth on this...

There is one case where `resourceType !== targetType.resourceType`: When `props.targetTypes` isn't specified,  `state.targetType === undefined` and the `<ResourceTypeInput />` mode of this component is used which is totally based on `resourceType`  and avoids `targetTypes` and `targetType` altogether similar to how this component worked before this PR.

I just pushed a commit that just adds a wrapper `TargetType` for use in the `<ResourceTypeInput />` mode to make the simplification you suggest and I don't hate it!

packages/react/src/ReferenceInput/ReferenceInput.test.tsx
mattlong: It appears so! I had blindly copy/pasted this from elsewhere in the codebase. I just tried removing the `await act` wrapper and the test failed.

packages/react/src/ReferenceInput/ReferenceInput.tsx
mattlong: Yeah, must be something like that; I splatted it to avoid the comment & eslint-disable per Cody's suggestion.

packages/react/src/ReferenceInput/ReferenceInput.stories.tsx
mattlong: It really seems like it should be, but after adding some debug lines, no, it appears the resources available are cleared out when moving between stories. Now I'm curious too. @codyebberson could you point us at where the clearing of created resources happens when moving around in storybook?

packages/react/src/ReferenceInput/ReferenceInput.tsx
mattlong: They did not auto-batch since GQL requests use `POST` instead of `GET`.

packages/react/src/ReferenceInput/ReferenceInput.stories.tsx
codyebberson: This is a good question.  I *think* (we should confirm) that the Storybook `preview.tsx` is re-evaluated per preview, which would mean that each preview has an independent `MockClient`

See: `packages/react/.storybook/preview.tsx`

packages/react/src/ReferenceInput/ReferenceInput.test.tsx
codyebberson: @ThatOneBro - You should only need to call `jest.advanceTimers` when using [Jest Fake Timers](https://jestjs.io/docs/timer-mocks)

Aside: I have a long standing mental todo to go back and review our use of Jest fake timers.  I'm pretty sure there are some files that should use them, but currently do not, and therefore actually `sleep(1000)`, which adds up.  `packages/core/src/client.test.ts` in paticular

packages/react/src/ReferenceInput/ReferenceInput.tsx
codyebberson: Got it.  Sorry, I should have known that, sorry for the goose chase!

codyebberson (approved):

PR #3571, opened 2023-12-17
Fixed react tsc errors

codyebberson:
For context, here is `packages/react/package.json` `"scripts"`:

```json
  "scripts": {
    "api-extractor": "api-extractor run --local && cp dist/types.d.ts dist/cjs/index.d.ts && cp dist/types.d.ts dist/esm/index.d.ts",
    "build": "npm run clean && tsc --project tsconfig.build.json && node esbuild.mjs && npm run api-extractor",
    "clean": "rimraf dist storybook-static",
    "dev": "storybook dev -p 6006",
    "storybook": "storybook build",
    "test": "jest"
  }
```

Let's look at `"build"` one-by-one:

1. `npm run clean` - clean the `dist` directory
2. `tsc --project tsconfig.build.json` - type check and produce `.d.ts` files
3. `node esbuild.mjs` - build `.cjs` and `.mjs` output files
4. `npm run api-extractor` - build consolidated `.d.ts` output file

Note that we don't ever run bare `tsc`, we only run `tsc --project tsconfig.build.json`.  The reason is that `tsc` actually does 2 related but separate things:

1. Type checks
2. Generates `.d.ts` output

We want to type check the entire project

But we only want to generate `.d.ts` output for `.tsx` files, not `.test.tsx` files.

And hence it is possible to get type errors in `.test.tsx`

And hence we have type errors in `.test.tsx` files 😢

For future investigation:  Now that we're using `api-extractor`, this might not be necessary anymore?


packages/react/src/BackboneElementInput/BackboneElementInput.utils.ts
codyebberson, line 1: @mattlong - Sonar complaining about test coverage in this file.  It looks like the `splitOnceRight()` method never actually splits, at least in any of our tests.  Is the split logic actually required there?

mattlong: I just added a test for a more deeply nested profile that triggers the logic.

mattlong (approved):

packages/react/src/BackboneElementInput/BackboneElementInput.utils.ts
codyebberson: Nice, ty! 🙏

PR #3576, opened 2023-12-18
Update understanding-uscdi-dataclasses.md

reshmakh:
Updating documentation bug


packages/docs/docs/fhir-datastore/understanding-uscdi-dataclasses.md
```diff
@@ -43,7 +43,7 @@ We find the following table useful to summarize common data elements found in an
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | [Device](/docs/api/fhir/resources/device)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
mattlong, line 46: There are a lot of other links in markdown docs to `stu3.1.1` pages; should I update them all to STU6.1? The other options would be to future-proof it by using URLs for the _current version_ versus a hard-coded version URLs that we use now as shown below. I'm not sure which if preferable...

https://hl7.org/fhir/us/core/StructureDefinition-us-core-implantable-device.html
https://hl7.org/fhir/us/core/STU6.1/StructureDefinition-us-core-implantable-device.html

reshmakh: @mattlong seems desirable... but which version are we planning to maintain in production?  I'm assuming it won't constantly be updated

Also https://www.medplum.com/docs/fhir-datastore/understanding-uscdi-dataclasses - per our regulatory requirements, we are currently certified for US Core 5.0.1 / USCDI v2 - all of these are US Core 6+ - which means we should not support those.  

Closing this PR.

PR #3579, opened 2023-12-18
Remove tsconfig.build.json files

codyebberson:
Follow-up to https://github.com/medplum/medplum/pull/3571

For many of our TypeScript packages, the `"build"` step does the following:

1. `npm run clean` - clean the `dist` directory
2. `tsc --project tsconfig.build.json` - type check and produce `.d.ts` files
3. `node esbuild.mjs` - build `.cjs` and `.mjs` output files
4. `npm run api-extractor` - build consolidated `.d.ts` output file

Note that we don't run bare `tsc`, we run `tsc --project tsconfig.build.json`.  The reason is that `tsc` actually does 2 related but separate things:

1. Type checks
2. Generates `.d.ts` output

We want to type check the entire project

But we only want to generate `.d.ts` output for `.tsx` files, not `.test.tsx` files.

And hence it is possible to get type errors in `.test.tsx`

This PR removes `tsconfig.build.json` files, and always emits `.d.ts`, even for unwanted files like `.test.ts`.  The `api-extractor` tool is smart enough to ignore them when generating the consolidated `.d.ts` that we publish to NPM.

### DIFF
--- a/api-extractor.json
+++ b/api-extractor.json
@@ -3,7 +3,7 @@
   "mainEntryPointFilePath": "<projectFolder>/dist/types/index.d.ts",
   "bundledPackages": ["@medplum/react-hooks"],
   "compiler": {
-    "tsconfigFilePath": "<projectFolder>/tsconfig.build.json",
+    "tsconfigFilePath": "<projectFolder>/tsconfig.json",
     "skipLibCheck": true
   },
   "apiReport": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -35,7 +35,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "npm run clean && tsc --project tsconfig.build.json && node esbuild.mjs",
+    "build": "npm run clean && tsc && node esbuild.mjs",
     "clean": "rimraf dist",
     "medplum": "ts-node src/index.ts",
     "test": "jest"
--- a/packages/cli/tsconfig.build.json
+++ b/packages/cli/tsconfig.build.json
@@ -1,9 +1,0 @@
-{
-  "extends": "./tsconfig.json",
-  "compilerOptions": {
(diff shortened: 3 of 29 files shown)


mattlong:
Do we know if `api-extractor` is purposefully doing the right thing here or if that might change in the future and it'd start include `.d.ts` from `test.ts` files?

I'm curious how this will impact build times. I imagine the cost of generating `.d.ts` for `.test.ts` files just for them to be ignored is pretty minimal since they don't deal with many types, but there are quite a few `.test.ts` files that could add up.

mattlong (commented):
Do we know if api-extractor is purposefully doing the right thing here or if that might change in the future and it'd start include .d.ts from test.ts files?

I'm curious how this will impact build times. I imagine the cost of generating .d.ts for .test.ts files just for them to be ignored is pretty minimal since they don't deal with many types, but there are quite a few .test.ts files that could add up.

reshmakh (approved):

codyebberson:
@mattlong 

> Do we know if api-extractor is purposefully doing the right thing here or if that might change in the future and it'd start include .d.ts from test.ts files?

`api-extractor` starts at a single entry point, see `api-extractor.json`

```json
"mainEntryPointFilePath": "<projectFolder>/dist/types/index.d.ts",
```

And then follows `import` statements.  So as long as we don't import tests, it should do the right thing.

> I'm curious how this will impact build times. I imagine the cost of generating .d.ts for .test.ts files just for them to be ignored is pretty minimal since they don't deal with many types, but there are quite a few .test.ts files that could add up.

Yeah, that is a valid concern.  I'm not aware of a way to type check everything, but `noEmit` tests.  That would be the best case.

This PR did find a bunch of bugs in tests, so I feel like the build time hit is worth it for now.

PR #3593, opened 2023-12-20
Fixes #3137 - handle multiple memberships in invite

### LINKED ISSUE #3137: Throw error on  `/invite` when `ProjectMembership` exists
# Background
Currently, if you invite the same user to the same project with the same profile type, two `ProjectMembership` resources will be created. 

However, this causes problems when the user tries to log into the project. [This else clause](https://github.com/medplum/medplum/blob/c99af3fe729e55c8afb5386369c4917ae7dfa26f/packages/server/src/oauth/utils.ts#L184-L188) ignores the case of two memberships, and returns a `Login` with no `membership` property set, which eventually leads to a 500 error. 

## Proposal
The `/invite` endpoint should check to see if a `ProjectMembership` already exists for the current `User` in the current `Project`. 
- If so, it should just perform a PUT on the existing `ProjectMembership` resource
- If not, follow the existing logic and create a new `ProjectMembership` resource

codyebberson:


packages/server/src/admin/invite.ts
```diff
@@ -193,6 +190,63 @@ async function searchForExistingProfile(request: ServerInviteRequest): Promise<P
...
+      },
+      {
+        code: 'project',
+        operator: Operator.EQUALS,
+        value: getReferenceString(project),
+      },
+      {
+        code: 'profile',
+        operator: Operator.EQUALS,
+        value: getReferenceString(profile),
```
rahul1, line 244: I'm wondering if we should even allow the same `User` to have two different memberships into the same project. 

The main use case I can think of would be if someone is both a `Practitioner` and a `Patient`, but the user for the `Patient` would be project-scoped.

I'd be in favor of removing this check, and only allowing one (User, ProjectMemership) per Project

packages/server/src/auth/profile.test.ts
```diff
@@ -34,12 +34,19 @@ describe('Profile', () => {
...
+        name: [{ text: 'Other' }],
+      });
+
       const inviteResult = await inviteUser({
         project: registerResult.project,
         resourceType: 'Practitioner',
         firstName: 'Multi2',
         lastName: 'Multi2',
         email,
+        membership: { profile: createReference(otherProfile) },
```
rahul1, line 49: I'm not sure if this will work with the login flow. 

[this line ](https://github.com/medplum/medplum/blob/66f331240d7dc0aa776bd7eef810695a93f8e9cb/packages/server/src/oauth/utils.ts#L179-L188) only handles the case where there are 0 or 1 memberships per (user, project).  

With two memberships, it returns a `Login` with 0 memberships, which ends up generating a 500 down the line

rahul1:
I wanted to flag another issue that some users encountered that we might want to handle here. 

Some users were invited as a server-level user _and_ a project-level user to the same project, with the same email and same profileType. Later on, this seemed to break the sign-in flow (still investigating). 

@codyebberson I'm wondering if we want to disallow two users with the same email in the same project?

packages/server/src/auth/profile.test.ts
codyebberson: It works the same way as any `User` with multiple `ProjectMembership` resources - it returns the partial `Login` + list of profiles for the user to choose from.

codyebberson:
> Some users were invited as a server-level user _and_ a project-level user to the same project, with the same email and same profileType. Later on, this seemed to break the sign-in flow (still investigating).

Let's file that separately.

> I'm wondering if we want to disallow two users with the same email in the same project?

We can.  We can also file that separately, as it is independent of this PR.

packages/server/src/auth/profile.test.ts
rahul1: That will work when using our `SignInForm` component (or calling `/auth/login`) But it breaks the SSO flow through the token exchange, which is the flow the user was using for the original ticket. 

https://github.com/medplum/medplum/blob/66f331240d7dc0aa776bd7eef810695a93f8e9cb/packages/server/src/oauth/token.ts#L387-L390

rahul1 (approved):

PR #3594, opened 2023-12-20
Improve form based on US Core Implantable Device

mattlong:
https://hl7.org/fhir/us/core/STU5.0.1/StructureDefinition-us-core-implantable-device.html

**Simple array input formatting**
Screenshot 2023-12-20 at 12 46 06 PM

**Nested array input formatting**
Screenshot 2023-12-20 at 1 03 49 PM

**Support for `base64Binary` fields**
Screenshot 2023-12-20 at 4 02 15 PM

**Improved property name capitalization, e.g. `Carrier AIDC, Carrier HRF`**
Screenshot 2023-12-20 at 12 58 56 PM


packages/core/src/utils.ts
```diff
@@ -464,6 +464,24 @@ export function isEmpty(v: any): boolean {
   return (t === 'string' && v === '') || (t === 'object' && Object.keys(v).length === 0);
 }
 
+type CanBePopulated = { length: number } | object;
+/**
+ * Returns true if the value is a non-empty string, an object with a length property greater than zero, or a non-empty object
+ * @param arg - Any value
+ * @returns True if the value is a non-empty string, an object with a length property greater than zero, or a non-empty object
+ */
+export function isPopulated<T extends { length: number } | object>(arg: CanBePopulated | undefined | null): arg is T {
```
mattlong, line 473: I've been wanting this function to exist for a while; the type-narrowing is the really nice part here.

It would have been nice to redefine `isEmpty(arg)` to just be `!isPopulated(arg)`, but they have slightly different semantics.

I'm in favor of doing the work to make `isEmpty` the inverse of `isPopulated`. Or at least deprecating `isEmpty` for now in favor of using `isPopulated`

codyebberson: Neat.  I'd be curious to review the cases where `isEmpty(arg) != !isPopulated(arg)`.  `isEmpty()` has a lot of weird edge cases to satisfy FHIRPath test cases.  If there's a good clean conceptual model that fulfills it, that'd be perfect.

packages/core/src/types.ts
```diff
@@ -292,29 +292,53 @@ export function getPathDisplayName(path: string): string {
...
+
+    [A-Z]?[a-z]+
+    This part matches a single, optional, uppercase letter followed by one or more lowercase letters ([a-z]+).
+    This pattern is suitable for matching words in camelCase format, where a word begins with a lowercase letter
+    but can optionally start with an uppercase letter (like in the middle of camelCase).
+
+    \d+
+    Matches a sequence of one or more digits into their own word
+    */
+    words = propertyName.match(/[A-Z]+(?![a-z])|[A-Z]?[a-z]+|\d+/g) ?? [];
```
codyebberson, line 321: Sweet 👍

packages/react/src/ElementsInput/ElementsInput.tsx
```diff
@@ -43,9 +43,26 @@ export function ElementsInput(props: ElementsInputProps): JSX.Element {
...
   }
+
+  const orderedElements = useMemo<[string, InternalSchemaElement][]>(() => {
+    const result = Object.entries(elements);
+    result.sort((a, b) => {
+      const aElem = a[1];
+      const bElem = b[1];
+
+      // required elements go first
+      if (aElem.min > 0 || bElem.min > 0) {
```
codyebberson, line 54: We should discuss w/ @rahul1.  I definitely see the logic behind this, but it will a be a functional change for some users who use `app` daily (how many times they need to tab, etc).  It *should* be minor, but we should get a sanity check.

codyebberson (commented):
Nice, this looks great.

Let's sync w/ @rahul on changing the order of fields to put required fields first.

packages/react/src/ElementsInput/ElementsInput.tsx
mattlong: Ah, yeah, that's a good call. This is probably a more significant breaking change. I'll make an opt-in prop for now.

packages/core/src/utils.ts
mattlong: I made a few tweaks to where they're logically consistent now except for numerical, boolean, etc. are expected to return `false` for both `isEmpty` and `isPopulated`.

packages/react/src/ElementsInput/ElementsInput.tsx
mattlong: Err, scratch that; making it optional would require passing the prop down several layers (or introducing a new `ResourceFormContext` which doesn't seem worth it at this point. Gonna remove this and keep it on the back burner for now.

codyebberson (approved):
w00t 👍

Issue #3599, opened 2023-12-21
Missing redirectUri param in reset password


ravindra-tc:
When attempting to add a `redirectUri` from the `ResetPasswordPage`, it has been observed that there is currently no provision to pass the redirectUri parameter through the `/auth/resetpassword` API.

To enhance the `/auth/resetpassword` API functionality by introducing support for the `redirectUri` parameter, allowing for dynamic redirection after the password reset process.

<img width="896" alt="Screenshot 2023-12-21 at 2 46 49 PM" src="https://github.com/medplum/medplum/assets/111381399/8b3b23ad-1a3b-44a2-86ec-ea8e1977b184">

codyebberson:
Fixed with https://github.com/medplum/medplum/pull/3600

Thanks @ravindra-tc !

PR #3600, opened 2023-12-21
Added redirectUri to reset password request

ravindra-tc:
- Added `redirectUri` optional parameter in `resetPassword` method
- Added a test case for it.

#3599


codyebberson (approved):
Thanks @ravindra-tc !

One word of caution - do take extra care in your application to verify `redirectUri` and protect against open redirect vulnerabilities: https://cwe.mitre.org/data/definitions/601.html

The Medplum server does not use `PasswordChangeRequest.redirectUri`, other than basic storage/retrieval.

Thanks!

Issue #3598, opened 2023-12-21
Change password only shows errors on console


nikul-tc:
I faced an issue while attempting to change my password. Although no error appeared on the screen, the console indicated that the change password URL was already in use.

<img width="1461" alt="Screenshot 2023-12-21 at 2 08 55 PM" src="https://github.com/medplum/medplum/assets/111053419/b87b7607-94e6-49c6-9983-9a4ece19bfcf">

codyebberson:
Fixed with https://github.com/medplum/medplum/pull/3601

Thanks @nikul-tc !

PR #3601, opened 2023-12-21
Display error from the server

nikul-tc:
- Display error from the server on SetPasswordPage

Fixes the Issue
https://github.com/medplum/medplum/issues/3598


codyebberson (approved):

codyebberson:
Thanks @nikul-tc !

PR #3616, opened 2023-12-22
Tooling for consistent timestamps in storybook stories

mattlong:
A handful of stories change appearance on every chromatic build since they rely on the current time. To address that, wrap the initial, default resource creation in a `sinon.useFakeTimers` invocation as well as offer a storybook decorator `withMockedDate` that can be used in individual stories that also create resources with timestamps that wind up being shown in the UI.

Screenshot 2024-01-03 at 10 43 35 AM


packages/react/src/utils/MockDateWrapper.tsx
```diff
@@ -0,0 +1,13 @@
+import mockdate from 'mockdate';
+import { ReactNode, useEffect } from 'react';
+
+export function MockDateWrapper({ children }: { children: ReactNode }): JSX.Element {
+  useEffect(() => {
+    mockdate.set('2023-05-05T00:00:00.000Z');
```
codyebberson, line 6: Any objections to using `jest.setSystemTime(new Date('2017-01-01'))` instead of new dependency?

packages/react/src/DefaultResourceTimeline/DefaultResourceTimeline.stories.tsx
```diff
@@ -1,31 +1,35 @@
...
 
 export const Basic = (): JSX.Element | null => {
   const medplum = useMedplum();
   const [resource, setResource] = useState<DiagnosticReport>();
+
   useEffect(() => {
+    const uuid = createIdGenerator();
     medplum
       .createResource<DiagnosticReport>({
+        id: uuid(),
```
codyebberson, line 24: The `MemoryRepository` class (used inside `MockClient`) allows manually setting `id`, `meta.versionId`, and `meta.lastUpdated`.  I suspect that for most of the components, it's just about manually setting those values.

codyebberson:
For some components, especially `<Scheduler>`, we may want to just explicitly disable: https://www.chromatic.com/docs/ignoring-elements/

packages/react/src/utils/MockDateWrapper.tsx
mattlong: Unfortunately `jest` isn't available in storybook since storybook isn't in the test environment. HOw do you feel about making `sinon` an official top-level dependency of `@medplum/react` since it's already being install elsewhere an an indirect dependency?

codyebberson: ahh, right

codyebberson (approved):
Nice

Issue #3592, opened 2023-12-20
Storybook story for US Core Patient and Implantable Device


mattlong:

reshmakh:
Required for #3002 and #3001

reshmakh:
Completed by https://github.com/medplum/medplum/pull/3617

PR #3617, opened 2023-12-22
Add stories for US Core Patient Profile and Implantable Device

mattlong:


mattlong:
Related to #3592

codyebberson:
FYI - Sonar can't differentiate between URL as an identifier vs URL as a request target.  You can sign into Sonar and mark them as "Safe":

image

image


I've wondered if we should have a single constant somewhere in `@medplum/core` for "http://hl7.org", which we could then refer to everywhere to avoid all the false positives.  It would be kinda clunky, but perhaps less clunky than overriding.

codyebberson (approved):
Nice https://medplum-storybook-git-uscore-stories-medplum.vercel.app/?path=%2Fstory%2Fmedplum-resourceform--us-core-patient

reshmakh:
Very cool!

mattlong:
> I've wondered if we should have a single constant somewhere in `@medplum/core` for "http://hl7.org", which we could then refer to everywhere to avoid all the false positives. It would be kinda clunky, but perhaps less clunky than overriding.

I added it; it does indeed feel a little clunky, but better than the false-positives I'd say.

PR #3642, opened 2024-01-02
Fixes #3309 - secrets editor textarea value with conceal/reveal

### LINKED ISSUE #3309: Secrets Editor treat secret as sensitive and conceal by default

codyebberson:
Based on https://github.com/medplum/medplum/pull/3310

Big thanks to @dillonstreator  for the contribution!

https://github.com/medplum/medplum/assets/749094/e7e9f54a-517b-4f3a-8fba-8081c17a5a42


packages/react/src/SensitiveTextarea/SensitiveTextarea.tsx
```diff
@@ -0,0 +1,53 @@
...
+          root: {
+            ...(styles.root ?? {}),
+            flexGrow: 1,
+          },
+        }}
+        ref={ref}
+        autosize
+        minRows={1}
+        onFocus={() => setRevealed(true)}
+        onBlur={() => setRevealed(false)}
```
mattwiller, line 40: I think it would be better security practice to have an explicit show/hide control here, allowing people to securely type in a value without someone should surfing

codyebberson: That is how it worked in the original PR, but it creates a pretty confusing user experience.

This implementation is how AWS Secrets and AWS Parameter work, which feels more natural.

I'm open to discussion though.

mattwiller: Mmkay, I think this way is definitely much better than nothing — good to merge as is

rahul1: Weighing in here: 
Since this is an _editing_ compoent,  then I agree with the current implementation. Having a separate button when I'm _already_ in editing mode. Many UX implementations in the wild don't even hide secrets when editing.

I think the explicit show/hide makes more sense for a _display_ component, which is the more common practice

ThatOneBro (approved):
LGTM

PR #3671, opened 2024-01-05
Align the agent log file name with the service name

jmalobicky:
Remove the `shawl` application prefix on the log file. Log file will now be named `MedplumAgent_rCURRENT.log` within the Medplum Agent directory on Windows. The previous file name was constructed as `shawl-v1.4.0-win64_for_MedplumAgent_rCURRENT.log`


codyebberson (approved):
Thanks @jmalobicky !

PR #3673, opened 2024-01-05
Proof of concept storybook GQL mocking

mattlong:
The interface is still a bit unwieldy since you have to obtain the exact GQL query string that is generated. It'd probably be sufficient to instead just provide an array of responses that you want to returned each time the `MedplumClient.graphql` is invoked and ignore the query string altogether.


mattlong:
Closing since this was just meant to be a POC

PR #3690, opened 2024-01-08
Update hibp server dependency and include in future dependency upgrades

wKovacs64:
Hi, all 👋  - author of the `hibp` library here. I noticed your `server` workspace depends on `hibp` and that it was pinned to an older version, citing that the current major version (v14) requires ESM. This wasn't supposed to be the case, but it happened due to a publishing mistake on my part (so sorry about that!). I've since corrected the issue and published v14.0.3 which supports CommonJS again, so `hibp` is safe to include in your future dependency upgrades. Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -362,82 +362,82 @@
       }
     },
     "node_modules/@algolia/cache-browser-local-storage": {
-      "version": "4.20.0",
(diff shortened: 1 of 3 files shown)


codyebberson:
@wKovacs64 - awesome, thank you!  Honored by your contribution 🙏

codyebberson (approved):
lgtm 👍 

image

PR #3691, opened 2024-01-08
Update HG setup-subscriptions

rahul1:
This PR updates the `setup-subscriptions` bot to prevent duplicate Subscriptions in HG for the same target endpoint

### DIFF
--- a/examples/medplum-demo-bots/src/health-gorilla/send-to-health-gorilla.ts
+++ b/examples/medplum-demo-bots/src/health-gorilla/send-to-health-gorilla.ts
@@ -54,6 +54,15 @@ interface HealthGorillaConfig {
 
 // Available labs Health Gorilla IDs
 // These come from the Health Gorilla Organization resources
+
+// Sandbox
+// const availableLabs: Record<string, string> = {
+//   Testing: 'f-4f0235627ac2d59b49e5575c',
+//   Labcorp: 'f-388554647b89801ea5e8320b',
+//   Quest: 'f-7c075564349e1a592e53147a',
+// };
+
+// Prod
 const availableLabs: Record<string, string> = {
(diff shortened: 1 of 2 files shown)


examples/medplum-demo-bots/package.json
```diff
@@ -44,6 +45,7 @@
     "pdfmake": "0.2.8",
     "ssh2-sftp-client": "9.1.0",
     "stripe": "14.7.0",
+    "tsx": "^4.7.0",
```
codyebberson, line 48: Was left in by mistake?  `tsx` has a lot of ripple considerations.  it doesn't look like you're using it anywhere.  Let's remove it, and revert changes to `package-lock.json`

`tsx` has some nice properties to it, and I'm open to considering it, but if we're going to use it, we should discuss at large

codyebberson (changes requested):

examples/medplum-demo-bots/package.json
rahul1: I don't use it anywhere in the code, but just to be able to run the script. I'm fine to remove it and do a global install of `tsx` on my machine instead. 

I didn't really want to use it. my goal was to use
`npx ts-node ....`
But I was running into [this bug](https://github.com/TypeStrong/ts-node/issues/1997)  when using `ts-node`

codyebberson: Sigh, yeah

Here's what I do for my one-off scripts:

```bash
node --no-warnings=ExperimentalWarning --loader ts-node/esm src/index.ts
```

Tragically complicated.

(*cough* wouldn't be a problem with Bun *cough*)

rahul1:
@codyebberson this is ready for re-review

PR #3702, opened 2024-01-10
Implement CodeSystem lookup operation

### LINKED ISSUE #3365: Implement CodeSystem/$lookup

mattwiller:
Resolves #3365


packages/server/src/fhir/operations/codesystemlookup.ts
```diff
@@ -0,0 +1,88 @@
...
+  );
+  lookup
+    .column(new Column(codeSystemTable, 'title'))
+    .column(new Column('Coding', 'display'))
+    .column(new Column(csPropTable, 'code'))
+    .column(new Column(csPropTable, 'type'))
+    .column(new Column(csPropTable, 'description'))
+    .column(new Column(propertyTable, 'value'))
+    .where(new Column(codeSystemTable, 'url'), '=', coding.system)
+    .where(new Column('Coding', 'code'), '=', coding.code);
```
codyebberson, line 59: How do project permissions work with this query?

mattwiller: That's actually originally how I had structured this, but I changed it and the permissions got lost in the shuffle; I'm open to reverting back to this approach.  I also tried to add a simpler check to verify that the CodeSystem is in the user's project: I could see an argument that execution permission on the operation is different from reading the resource, and not being able to read the resource itself shouldn't necessarily block someone from leveraging the terminology service features.  Thoughts either way?

codyebberson (approved):

PR #3707, opened 2024-01-12
Dependency upgrades

codyebberson:
Dependency upgrades

The big one is Docusaurus 3.1, which includes more sophisticated link validation.  It found a bunch of broken `#` links, which I made best effort to fix.

@reshmakh - FYI - It appears that many of the ONC `#` links pointed to content that was removed or refactored, so I truncated many of those links to simply point to the ONC page.


reshmakh (approved):

reshmakh:
@codyebberson thanks for fixing those ONC links.  After Medplum certification we removed the items for which we were not certified, causing the links to break.

ThatOneBro (approved):
LGTM

PR #3723, opened 2024-01-15
bug: Prevent isRegisterEnabled optimization

### LINKED ISSUE #3722: MEDPLUM_REGISTER_ENABLED not configurable in @medplum/app except by building from source
## Problem

In the build on npm, `isRegisterEnabled` is optimized away by Rollup (because the function always evaluates to true, and so registration cannot be disabled in the frontend except by building the app from source (at least via the __placeholder__ replacement strategy used by @medplum/cli).

Here's how this bug works:

`__MEDPLUM_REGISTER_ENABLED__` is used a placeholder for the `MEDPLUM_REGISTER_ENABLED` environment variable in the build on npm: https://github.com/medplum/medplum/blob/3d6f1b55fa90f49bdb3db4ca13a1038858999991/.github/workflows/publish.yml#L58

Which is assigned to the `config` const here: https://github.com/medplum/medplum/blob/3d6f1b55fa90f49bdb3db4ca13a1038858999991/packages/app/src/config.ts#L9-L15

And then evaluated in `isRegisterEnabled`, which is always `true` given `__MEDPLUM_REGISTER_ENABLED__` is truthy: https://github.com/medplum/medplum/blob/3d6f1b55fa90f49bdb3db4ca13a1038858999991/packages/app/src/config.ts#L21-L24

Then, `isRegisterEnabled` is optimized away in the build and is missing from where it is called in the source. Here's what appears in the source:

https://github.com/medplum/medplum/blob/3d6f1b55fa90f49bdb3db4ca13a1038858999991/packages/app/src/SignInPage.tsx#L30

And here's what appears in the build on npm:

```typescript
onRegister:()=>e("/register")
```

## Solution

Since Rollup deoptimizes inside of try / catch statements, we could do that here -- it's a little wonky looking: [tryCatchDeoptimization](https://rollupjs.org/configuration-options/#treeshake-trycatchdeoptimization)

```typescript
export function isRegisterEnabled(): boolean {
  try {
    return config.registerEnabled !== false && config.registerEnabled !== 'false';
  } catch {
    return true;
  }
}
```

If this sounds good, I'll submit a PR.

jdjkelly:
## Why
* Closes #3722

## What
* Prevents Rollup from optimizing the registerEnabled check away by use of a try/catch:
* Correctly compiles the function now when running the build, example from my local build:

```
function LA(){try{return Wy.registerEnabled!==!1&&Wy.registerEnabled!=="false"}catch{return!0}}
```

### DIFF
--- a/packages/app/src/config.ts
+++ b/packages/app/src/config.ts
@@ -19,6 +19,10 @@ export function getConfig(): MedplumAppConfig {
 }
 
 export function isRegisterEnabled(): boolean {
-  // Default to true
-  return config.registerEnabled !== false && config.registerEnabled !== 'false';
+  // This try/catch exists to prevent Rollup optimization from removing this function
+  try {
+    return config.registerEnabled !== false && config.registerEnabled !== 'false';
+  } catch {
+    return true;
+  }
 }


codyebberson (approved):
Thanks @jdjkelly !

PR #3731, opened 2024-01-16
Add lookup table to index CodeSystem codings

mattwiller:
Adds a new `LookupTable` implementation to the repo, which automatically indexes contained in `CodeSystem` resources with `content = 'complete'`.  This should allow bringing all the base FHIR code systems into the new DB tables, including implicit hierarchy properties


codyebberson (approved):
Nice 👍

PR #3733, opened 2024-01-16
Add array append helper

mattwiller:
This pattern (i.e. appending to a possibly-undefined array) was used in a couple places, but the existing helper function was [not very efficient](https://jsperf.app/keboda); this adds a new `append(array, element)` helper function to `@medplum/core` and attempts to standardize usage throughout the codebase


codyebberson (approved):
Nice.  Good eye

PR #3740, opened 2024-01-17
feat(subscriptions): add `heartbeat` for WS subs

ThatOneBro:
## TODO:
- [x] Add tests for heartbeat functionality

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -42,7 +42,7 @@ export interface MedplumSourceInfraConfig {
   maxAzs: ValueOrExternalSecret<number>;
   rdsInstances: ValueOrExternalSecret<number>;
   rdsInstanceType: ValueOrExternalSecret<string>;
-  rdsInstanceVersion: ValueOrExternalSecret<string>;
+  rdsInstanceVersion?: ValueOrExternalSecret<string>;
   rdsSecretsArn?: ValueOrExternalSecret<string>;
   cacheNodeType?: ValueOrExternalSecret<string>;
   desiredServerCount: ValueOrExternalSecret<number>;
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -196,9 +196,7 @@ export function initAppServices(config: MedplumServerConfig): Promise<void> {
     await initKeys(config);
     initBinaryStorage(config.binaryStorage);
     initWorkers(config);
-    if (!config.heartbeatDisabled) {
-      initHeartbeat(config);
-    }
+    initHeartbeat(config);
   });
 }
 
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -196,7 +196,9 @@ export function initAppServices(config: MedplumServerConfig): Promise<void> {
     await initKeys(config);
     initBinaryStorage(config.binaryStorage);
     initWorkers(config);
-    initHeartbeat();
+    if (!config.heartbeatDisabled) {
+      initHeartbeat(config);
+    }
   });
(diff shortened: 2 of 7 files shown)


packages/core/src/config.ts
```diff
@@ -42,7 +42,7 @@ export interface MedplumSourceInfraConfig {
   maxAzs: ValueOrExternalSecret<number>;
   rdsInstances: ValueOrExternalSecret<number>;
   rdsInstanceType: ValueOrExternalSecret<string>;
-  rdsInstanceVersion: ValueOrExternalSecret<string>;
+  rdsInstanceVersion?: ValueOrExternalSecret<string>;
```
ThatOneBro, line 45: It seems like this is intended to be optional, this was causing a type error in the tests since it was marked as required, though in the `MedplumInfraConfig` it is treated as optional

packages/server/src/config.ts
```diff
@@ -47,6 +47,8 @@ export interface MedplumServerConfig {
   keepAliveTimeout?: number;
   vmContextBotsEnabled?: boolean;
   shutdownTimeoutMilliseconds?: number;
+  heartbeatMilliseconds?: number;
+  heartbeatDisabled?: boolean;
```
ThatOneBro, line 51: I made this `heartbeatDisabled` instead of `heartbeatEnabled` to make it somewhat implicit that `heartbeat` is enabled by default. I think this could be a useful convention for optional boolean config values, but open to disagreement

packages/core/src/config.ts
codyebberson: Nice, good catch

packages/server/src/app.ts
```diff
@@ -196,7 +196,9 @@ export function initAppServices(config: MedplumServerConfig): Promise<void> {
     await initKeys(config);
     initBinaryStorage(config.binaryStorage);
     initWorkers(config);
-    initHeartbeat();
+    if (!config.heartbeatDisabled) {
+      initHeartbeat(config);
```
codyebberson, line 200: Super minor preference to move this conditional into `initHeartbeat` so we can keep this function clean.

packages/server/src/config.ts
codyebberson: Yeah, tough one.  I understand where you're coming from.  I don't have strong preferences, but I know there is a school of thought to never use "disabled" in boolean names to avoid double negatives.

We do already have precedence for "enabled" with default true, `registerEnabled`, `logAuditEvents`

packages/server/src/heartbeat.ts
```diff
@@ -1,17 +1,26 @@
-import { EventTarget } from '@medplum/core';
+import { TypedEventTarget } from '@medplum/core';
+import { MedplumServerConfig } from './config';
 
-export const heartbeat = new EventTarget();
+export type HeartbeatEventMap = {
+  heartbeat: { type: 'heartbeat' };
+};
+
+export const heartbeat = new TypedEventTarget<HeartbeatEventMap>();
```
codyebberson, line 8: Separate conversation, out of scope for this PR, but I wonder if we could combine `TypedEventTarget` and `EventTarget` into a single class to avoid the wrapping.  Not a big deal, we can discuss offline.

codyebberson (approved):
Nice, lgtm 👍 

Nothing blocking, but a handful of comments for your consideration

packages/server/src/heartbeat.ts
ThatOneBro: Probably could, shouldn't be an issue

PR #3745, opened 2024-01-18
ci(madge): add `madge` job for PRs against `medplum/test-actions`

ThatOneBro:
Adds `madge` CI job to detect circular dependencies in modules like the ones found and fixed in #3742.

There are still some circular dependencies in the codebase but this job will only run in CI for PRs against the `medplum/test-actions` branch for now. This is so we can test it before it gets accepted into the standard suite of CI jobs that run on PRs against `main`.

### DIFF
--- a/.github/workflows/madge.yml
+++ b/.github/workflows/madge.yml
@@ -1,0 +1,96 @@
+name: madge
+
+concurrency:
+  group: ci-madge-${{ github.ref_name == 'main' && github.sha || github.ref }}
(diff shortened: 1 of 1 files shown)


ThatOneBro:
@codyebberson Nope, this is only enabled on the `medplum/test-actions` branch for now. I wanted to test it first and then once we resolve the circular dependencies we can easily enable it on all PRs

codyebberson (approved):

Issue #3663, opened 2024-01-04
React NotesDisplay is not exported


ksmith94:
The [`NotesDisplay`](https://github.com/medplum/medplum/blob/main/packages/react/src/NoteDisplay/NoteDisplay.tsx) component in the React package is not exported as part of `@medplum/react`. It's in the [storybook](https://storybook.medplum.com/?path=/story/medplum-notesdisplay--simple), so I believe it should be an exported member.

PR #3759, opened 2024-01-19
Export NotesDisplay from @medplum/react

ksmith94:
This will add `NotesDisplay` as an exported member of the medplum react package.


ksmith94:
Fixes #3663

rahul1 (approved):

PR #3760, opened 2024-01-19
Restructure elements context and slices

mattlong:
This PR tightens up the abstractions around the `ElementsContext`, formerly called `BackboneElementContext`. One big improvement is the proper integration of parent contexts into sub-contexts, which will be foundational to properly incorporating fixed/pattern values from structure definitions into resources. Related and a more noticeable developer-facing change is the addition of the required `path` property on most input components to facilitate usage of `ElementsContext`. 

Most of the work for applying pattern/fixed values is mostly ready to go, but I wanted to break it up into a separate PRs to keep them more digestible.


packages/mock/src/mocks/uscore/index.ts
```diff
@@ -1,9 +1,9 @@
-import { Device, Patient } from '@medplum/fhirtypes';
+import { Device, Patient, StructureDefinition } from '@medplum/fhirtypes';
 import StructureDefinitionList from './uscore-v5.0.1-structuredefinitions.json';
 import { HTTP_HL7_ORG, deepClone } from '@medplum/core';
 import { HomerSimpson } from '../simpsons';
 
-export const USCoreStructureDefinitionList = StructureDefinitionList;
+export const USCoreStructureDefinitionList: StructureDefinition[] = StructureDefinitionList as StructureDefinition[];
```
mattwiller, line 6: [nit] You probably don't need both type assertions here

packages/react/src/PatientSummary/Medications.tsx
```diff
@@ -80,9 +80,11 @@ export function Medications(props: MedicationsProps): JSX.Element {
           <Stack h={275}>
             <CodeableConceptInput
               name="request"
+              path="MedicationRequest.medication[x]"
```
mattwiller, line 83: I think in an ideal world the user wouldn't need to know that the element is choice type or not: would it be possible to use the element name (e.g. `MedicationRequest.medication`) here instead?

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.test.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      expect(slices.length).toBe(2);
+      expect(slicedValues.map((sliceValues) => sliceValues.length)).toEqual([1, 1, 0]);
+    });
+  });
+
+  // test('extensions', () => {});
+});
+
+// const values = [
```
mattwiller, line 225: Should remove or uncomment

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.ts
```diff
@@ -1,41 +1,62 @@
...
+  //     lastNestedProp = getTypedPropertyValueWithSchema(typedValue, discriminator.path, lastEd);
+  //   }
+  // }
+  // nestedProp = lastNestedProp;
+
+  // let nestedProp: TypedValue | TypedValue[] | undefined;
+  // const ed = getElementDefinitionFromElements(elementList, discriminator.path);
+  // if (ed) {
+  // nestedProp = getTypedPropertyValueWithSchema(typedValue, discriminator.path, ed);
+  // }
```
mattwiller, line 41: Remove?

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.ts
```diff
@@ -1,41 +1,62 @@
...
+
+  const nestedProp = getNestedProperty(typedValue, discriminator.path, profileUrl);
+
+  if (nestedProp) {
+    const result =
+      arrayify(nestedProp)?.some((v: any) => matchDiscriminant(v, discriminator, slice, elementList)) ?? false;
+    return result;
```
mattwiller, line 48:
```suggestion
    return
      arrayify(nestedProp)?.some((v: any) => matchDiscriminant(v, discriminator, slice, elementList)) ?? false;
```

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.ts
```diff
@@ -1,41 +1,62 @@
...
+  } else {
+    console.assert(false, 'getNestedProperty[%s] in isDiscriminatorComponentMatch missed', discriminator.path);
```
mattwiller, line 50: Is this a pattern we want to use in the React components for error reporting?  Seems pretty handy

packages/react/src/hooks/useCallbackState.ts
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * [1]. https://github.com/theKashey/use-callback-state
+ * [2]. https://medium.com/geekculture/usecallbackstate-the-hook-that-let-you-run-code-after-a-setstate-operation-finished-25f40db56661
+ * @param options - An object
+ * @param options.initialState - The value you want the state to be initially. It can be a value of any type, but there is a special behavior for functions. This argument is ignored after the initial render.
+ * @param options.debugMode - enable debug console messages
+ * @param options.debugName - name to be used in debug messages
+ * @returns [state, setState]
+ */
+function useCallbackState<T>({
```
codyebberson, line 21: Is this primarily for dev and testing?

I didn't see any calls to this hook using `debugMode` or `debugName`.

Perf impact at runtime?

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.ts
```diff
@@ -69,7 +95,10 @@ export function assignValuesIntoSlices(
   }
 
   for (const value of values) {
-    const sliceName = getValueSliceName(value, slices, slicing.discriminator);
+    const sliceName = getValueSliceName(value, slices, slicing.discriminator, profileUrl);
+    if (!isPopulated(sliceName)) {
+      console.debug('slice value assigned to default slice', value, slicing);
```
mattwiller, line 100: Should this be left in for users, or is it for troubleshooting only?

packages/core/src/fhirpath/utils.ts
```diff
@@ -66,30 +66,35 @@ export function singleton(collection: TypedValue[], type?: string): TypedValue |
...
  * @param input - The base context (FHIR resource or backbone element).
  * @param path - The property path.
+ * @param profileUrl - (optional) Profile URl
  * @returns The value of the property and the property type.
  */
-export function getTypedPropertyValue(input: TypedValue, path: string): TypedValue[] | TypedValue | undefined {
+export function getTypedPropertyValue(
+  input: TypedValue,
+  path: string,
+  profileUrl?: string
```
codyebberson, line 75: Sensible, but I wonder if we should generalize this to `context` or `options` so that we don't end up with many extra args.

Separately, I'll follow up with you on an idea from @mattwiller  about a `Proxy` based solution.  I've been resistant to it, because it will add a layer of complexity, but changes like this make me wonder if it's necessary complexity.

mattwiller (commented):
This all looked great to me!  I'll just defer to someone more familiar with React for approval on the finer details of the PR

codyebberson (approved):
Overall lgtm - see inline comments, not blocking.

packages/react/src/hooks/useCallbackState.ts
mattlong: So, the reason I had introduced `useCallbackState` is complicated to say the least.

We often follow the pattern of propagating the chain of `onChange` callbacks with something like this

```typescript
function setValueWrapper(newValue: any): void {
  setValue(newValue);
  if (props.onChange) {
    props.onChange(newValue);
  }
}

return (
...
<ChildInput onChange={setValueWrapper} />
...
);
```

In the before times, where every change in the form was triggered by the user value, this pattern works out just fine.

Now, consider the after times where fixed (and pattern) values need to be added in to the resource's value at various levels of the hierarchy. These fixed values are determined shortly after (importantly, a later render cycle) input components are rendered. A common scenario for fixed values are extension URLs. This is seemingly straightforward enough still when looking at an individual Extension, but what about an array of Extensions?

A `ResourceArrayInput` will have several `SliceInput`/`ExtensionInput` below it, all of which attempt to apply fixed values in the same render cycles. Finally we arrive at the issue: using the pattern above, a race condition exists since each call of `setValue` on the `ResourceArrayInput` would mask the previous invocations since they're working on a stale version of `value`, resulting in only the changes from whatever the last extension in the array happens to be persisting to the `ResourceArrayInput` value:

```
function setValuesWrapper(newValue: any, valueIndex: number) {
  const newValues = [...values];
  newValues[valueIndex] = newValue;
  setValues(newValues);
  if (props.onChange) {
    props.onChange(newValue);
  }
}
```

React has the masking aspect of this covered with the callback version of `setState`. However, a new problem arises: calling `onChange` in a setState callback as show below can lead to errors like the one shown below depending on the exact interplay of events, effects, and render cycles:

```typescript
function setValuesWrapper(newValue: any, valueIndex: number) {
  setValues((oldValues) {
    const newValues = [...oldValues];
    newValues[valueIndex] = newValue;

    if (props.onChange) {
      // Danger! Calling onChange in a setState callback can lead to React errors
      props.onChange(newValue);
    }
    return newValues;
  });
}
```

Screenshot 2024-01-22 at 7 22 51 PM

Error message from the screenshot above:

Cannot update a component (`BackboneElementInput`) while rendering a different component (`ElementsInput`). To locate the bad setState() call inside `ElementsInput`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

Instead, a `useEffect` needs to be added to `values` for changes and only then calling `onChanges`. Putting all that together gets this most of the way there, but the onChange callback chain approach is definitely starting to feel brittle. As we've discussed, at some point I think it'd be worth investigating switching to an uncontrolled form approach instead.

Oh, also, `useCallbackState` turned out to not even be effective at what I thought it was doing. So that'll be going away.

packages/react/src/PatientSummary/Medications.tsx
mattlong: It's important that `path` property matches the paths specified in a StructureDefinition (and the keys of`InternalTypeSchema.elements`). The utility from this is so that `ElementsContext.getModifiedNestedElement` and can be used as is being done in [HumanNameInput](https://github.com/medplum/medplum/blob/6e353e56ae8dca2581b8c6ec09f2135d6065c0a6/packages/react/src/IdentifierInput/IdentifierInput.tsx#L16) without necessarily knowing the chosen type.

mattlong:
Replaced by #3869

PR #3774, opened 2024-01-20
Dependency upgrades

medplumbot:
Dependency upgrades


codyebberson:
@ThatOneBro - I had to manually add `metro-react-native-babel-preset` to `packages/expo-polyfills/package.json`

It looks like before, it was included as part of `babel-preset-expo`, but now it's not?

Before / on `main`:

image

Should we instead change the `babel.config.js` to use `babel-preset-expo`?

See: https://github.com/medplum/medplum/pull/3774/commits/620a039cd1c66653c91484b9e49dc89c246a95cc

ThatOneBro:
I think it may be ok to just use `babel-preset-expo`, I tested it and seems like it still builds correctly. We can just eliminate usage / dependency of `metro-react-native-babel-preset`

ThatOneBro (approved):

PR #3778, opened 2024-01-20
Testing `madge` CI job

ThatOneBro:


ThatOneBro:
Looks like it works in the expected failure case 👍

PR #3784, opened 2024-01-22
Document using vm context bots

ksmith94:
This will add documentation on using VM Context bots to use bots in self-hosted/local environments

### DIFF
--- a/packages/docs/docs/bots/vm-context-bots.md
+++ b/packages/docs/docs/bots/vm-context-bots.md
@@ -1,0 +1,18 @@
+# VM Context Bots
+
+To set up the Medplum Bot framework locally, Medplum offers VM Context Bots. VM Context allows bots to spin up a local thread inside your server, rather than using an isolated lambda.
+
+There are two steps to set up VM context bots:
+
+1. Enable VM Context Bots on your server config.
+2. Set your Bot's runtime version to VM Context.
+
+To enable VM Context bots on your server, set `vmContextBotsEnabled: true` in both the AWS parameter store _and_ your local `config.json` file.
+
+All Bots have a field for `runtimeVersion`, which can be set to either `awslambda` or `vmcontext`. To use your bot locally, set this field to `vmcontext`.
+
+:::danger
+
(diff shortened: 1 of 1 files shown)


packages/docs/docs/bots/vm-context-bots.md
```diff
@@ -0,0 +1,22 @@
...
+
+To enable VM Context [`Bots`](/docs/api/fhir/medplum/bot) on your server, set `vmContextBotsEnabled: true` in both the AWS parameter store and your local `config.json` file.
+
+All [`Bots`](/docs/api/fhir/medplum/bot) have a field for `runtimeVersion`, which can be set to either `awslambda` or `vmcontext`. To use your [`Bot`](/docs/api/fhir/medplum/bot) locally, set this field to `vmcontext`.
+
+Using VM Context allows you to use resources more efficiently, however it can also have security and isolation concerns if used in an untrusted environment. For this reason, it is important to only use VM Context [`Bots`](/docs/api/fhir/medplum/bot) in trusted environments (e.g. localhost).
+
+:::danger
+
+WE STRONGLY CAUTION THAT VM CONTEXT BOTS SHOULD ONLY BE USED IN TRUSTED ENVIRONMENTS (e.g. localhost), AND NOT IN PRODUCTION.
```
codyebberson, line 18: We should be more nuanced with this line.

"Trusted Environments" means "environments where only trusted and authorized users are allowed to deploy bots".

That can include "production" under the right circumstances.

We have enterprise customers who use VM Context bots in production, because they have put to appropriate safety mechanisms in place.

Here is the language that Node.js uses:

https://nodejs.org/docs/latest-v20.x/api/vm.html#vm-executing-javascript

> The node:vm module is not a security mechanism. Do not use it to run untrusted code.

ksmith94: Makes sense, I changed the wording to clarify that it can be used as long as you the necessary safety measures are takenx

rahul1 (approved):

Issue #1369, opened 2022-12-20
Better FHIR Bundle transaction support


rahul1:
We might want to consider rolling back changes to FHIR bundles that have partial failures to avoid getting into an inconsistent state. 

An example issue: when a batch request fails to create a resource, but successfully creates a resource with a reference to the failed resource.

mattwiller:
I did some investigation around what it would take to more fully support `transaction` Bundles: we'll likely need to move the existing batch implementation from `@medplum/fhir-router` into `@medplum/server` and refactor it significantly.

The new logic for handling `batch` and `transaction` Bundles should follow the algorithm outlined in the [FHIR HTTP spec][fhir-transaction].  As I understand it, the procedure is as follows:
- For both kinds of Bundle, group and order the entries by `Bundle.entry.request.method`
  - Ordered as: deletions, creations, updates, reads
- While grouping the entries, also resolve each delete, create, or update entry to a relative reference (e.g. `Patient/12345`), and record a map from `Bundle.entry.fullUrl` to that reference
  - For `DELETE` entries, the ID is undefined
  - For `POST` entries, generate a new ID for the resource
  - For `PUT` and `PATCH` entries, use the ID of the resource presented, or resolve the search query in `Bundle.entry.request.url` for [conditional update][cond-update]
  - If any two entries resolve to the same reference, return an error for the entire operation
- [Rewrite all references][bundle-ref-rewrite] in the Bundle to point to the correctly-resolved reference strings:
  - Replace all instances within each `Bundle.entry.resource` of each mapped `fullUrl` to its corresponding relative reference
    - Not just in `Reference` elements, but also `Resource.id` and elements of type `uri`, `url`, `oid`, and `uuid`
    - Elements of type `canonical` must __NOT__ be rewritten
  - Also resolve [conditional references][conditional-refs] by performing the given search query and ensuring it returns a single item — otherwise an error is returned for that entry
- If Bundle is a `transaction`, start the database transaction
- Process the entries in order of the groups defined above as they would normally be, issuing DB queries for each one
  - Record the results for each entry to construct the Bundle response
- If `transaction`, commit the database transaction

[fhir-transaction]: http://hl7.org/fhir/http.html#brules
[cond-update]: http://hl7.org/fhir/http.html#cond-update
[bundle-ref-rewrite]: http://hl7.org/fhir/http.html#url-fixing
[conditional-refs]: http://hl7.org/fhir/http.html#vs-refs

taylorqj:
[...]

PR #3788, opened 2024-01-23
Nested transactions

codyebberson:
Formalizes our use of database transactions across all database usage.

This is in service of proper FHIR Bundle transaction support, described here: https://github.com/medplum/medplum/issues/1369

Once this is shipped, implementing `Bundle.type = 'transaction'` is this:

image

Old conventions:
* The only concrete transactions were in the multi-table write operations
* Otherwise, grab a database connection from `getClient()` as needed

New conventions:
* By default, you should only ever use the server `Repository` class for database connections
* If you don't need a transaction, use `repo.getClient()` for database access
  * If in an existing transaction, this will reuse the same connection and be part of the transaction
  * If not, it falls back to the existing behavior and executes a one-off query
* If you need a transaction, wrap with `repo.withTransaction((client) => { ... });`
  * If already in a transaction, this updates a depth counter and reuses the connection
  * If not, it creates a new transaction
  * This adds the necessary wrappers for `commit`, `rollback` and release

Exceptions:
* We have a few cases of database cursors for processing all rows in a table
  * For example, "reindex all resources"
  * Full table cursors are incompatible with this model, and will lock the connection
  * So we still call `getClient()` (not `repo.getClient()`) in those cases
* Specialty operations
  * For example, `Project/$expunge` and `CodeSystem/$import`
  * These have a high degree of special logic
  * Perhaps someday they should be converted to this same model, but that can be separate work

Known issues:
* There is one serious known issue with Redis cache cleanup in transaction rollback
* As a team, we have discussed a number of possible solutions
    * For example, track resources to add to Redis, and only add on final `COMMIT`
    * Or, track all affected resource IDs, and flush those from the cache on `ROLLBACK`
    * Or, revisit our use of Redis cache entirely
* This issue does not affect any of our current uses of `withTransaction()`, so it does not effect any current runtime implementation
* However, it is a blocker before we can fully implement FHIR transaction bundles
* The next step is to run some experiments and performance tests on the various strategy

Also:
 * There were a few unexpected warnings in the postgres logs for a clean run of server tests
 * This PR cleans up all of them except for a "unique constraint" error, which is expected and correct

Before:

image

After: 

image


packages/fhir-router/src/repo.ts
```diff
@@ -145,6 +145,12 @@ export interface FhirRepository {
    * @returns Promise to the array of search results.
    */
   searchResources<T extends Resource>(searchRequest: SearchRequest<T>): Promise<T[]>;
+
+  beginTransaction(): Promise<void>;
+
+  commitTransaction(): Promise<void>;
+
+  rollbackTransaction(): Promise<void>;
```
mattwiller, line 153: I'd also propose for consideration something like this:
```ts
withTransaction<T>((db: PoolClient) => T): Promise<T>;
```

codyebberson: Yep, that is much cleaner.  Thanks for suggestion 🙏

packages/fhir-router/src/repo.ts
```diff
@@ -363,6 +363,7 @@ export class MemoryRepository extends BaseRepository implements FhirRepository {
   }
 
   withTransaction<TResult>(callback: (client: unknown) => Promise<TResult>): Promise<TResult> {
+    // MockRepository currently does not support transactions
```
ThatOneBro, line 366: Should we log something to console too maybe, since it could be a red herring pointing to a missing client / potentially silently "succeed"?

packages/server/src/fhir/operations/codesystemimport.ts
```diff
@@ -76,7 +75,9 @@ export async function codeSystemImportHandler(req: Request, res: Response): Prom
   const codeSystem = codeSystems[0];
 
   try {
-    await importCodeSystem(codeSystem, params.concept, params.property);
+    await ctx.repo.withTransaction(async (db) => {
+      await importCodeSystem(db, codeSystem, params.concept, params.property);
+    });
```
mattwiller, line 80: Love that the diff is this simple 👍

packages/server/src/fhir/repo.ts
```diff
@@ -802,8 +793,10 @@ export class Repository extends BaseRepository implements FhirRepository {
       throw new OperationOutcomeError(forbidden);
     }
 
-    const resource = await this.readResourceImpl<T>(resourceType, id);
-    return this.reindexResourceImpl(resource);
+    await this.withTransaction(async () => {
+      const resource = await this.readResourceImpl<T>(resourceType, id);
+      return this.reindexResourceImpl(resource);
+    });
```
mattwiller, line 799: Nice find!  How did you identify that this should get a transaction?

packages/server/src/fhir/transaction.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      expect(readCheck1).toBeDefined();
+
+      // Search for patient by ID
+      // This should succeed
+      const searchCheck1 = await repo.search<Patient>({
+        resourceType: 'Patient',
+        filters: [{ code: '_id', operator: Operator.EQUALS, value: patient?.id as string }],
+      });
+      expect(searchCheck1).toBeDefined();
+      expect(searchCheck1.entry).toHaveLength(1);
```
mattwiller, line 52: These could optionally be combined, up to your preference
```suggestion
      expect(searchCheck1?.entry).toHaveLength(1);
```

packages/server/src/fhir/transaction.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      // TODO: Currently not failing due to cache bug
+      // try {
+      //   await repo.readResource('Patient', (patient as Patient).id as string);
+      //   throw new Error('Expected error');
+      // } catch (err) {
+      //   expect((err as OperationOutcomeError).outcome).toMatchObject(notFound);
+      // }
+
+      // Search for patient by ID
+      // This should succeed within the transaction
```
mattwiller, line 113:
```suggestion
      // This should return zero results because the transaction was rolled back
```

mattwiller (approved):

ThatOneBro (approved):

packages/server/src/fhir/repo.ts
codyebberson: It's not strictly required, but is the more "correct" version

PR #3790, opened 2024-01-23
Additional logging

mattwiller:


packages/docs/docs/self-hosting/config-settings.md
```diff
@@ -135,7 +135,6 @@ You will also be prompted for a parameter "Type". The default option is "String"
...
 | `redis`                | The redis connection details as a JSON object. Only available when using JSON config file. See [AWS Secrets](#aws-secrets).                                                                                                                                                       |          |            |                                     |
 | `RedisSecrets`         | The AWS Secret ID containing Redis connection details (created automatically by CDK). Only available when using AWS Parameter Store config. See [AWS Secrets](#aws-secrets).                                                                                                      |          | `cdk`      |                                     |
-| `logRequests`          | Optional flag to log individual HTTP requests.                                                                                                                                                                                                                                    |          |            | `false`                             |
```
codyebberson, line 138: I think we should keep the config setting around, and just flip the switch in our prod environments.

This feature is very noisy on localhost / dev mode.

codyebberson (changes requested):
I think we should keep the config setting around, and just flip the switch in our prod environments.

This feature is very noisy on localhost / dev mode.

codyebberson (approved):

PR #3791, opened 2024-01-23
Use tsvector index for token text search

mattwiller:
To align `:text` modifier search with regular string search semantics and improve query performance, this query pattern will use a `gin(tsvector)` index instead of string wildcard matching


packages/server/src/migrations/schema/index.ts
```diff
@@ -54,3 +54,5 @@ export * as v52 from './v52';
 export * as v53 from './v53';
 export * as v54 from './v54';
 export * as v55 from './v55';
+export * as v56 from './v56';
```
codyebberson, line 57: 😬

codyebberson (approved):
lgtm

Any estimates on migration impact?  Do we need to deploy this outside peak hours?

packages/server/src/migrations/schema/v57.ts
```diff
@@ -0,0 +1,182 @@
...
+  'ValueSet',
+  'VerificationResult',
+  'VisionPrescription',
+];
+
+export async function run(db: PoolClient): Promise<void> {
+  for (const resourceType of resourceTypes) {
+    const start = Date.now();
+    await db.query(
+      `CREATE INDEX CONCURRENTLY IF NOT EXISTS "${resourceType}_text_tsv_idx" on "${resourceType}_Token" USING gin (to_tsvector('simple', value)) WHERE system = 'text'`
```
codyebberson, line 178: I think this index name should be `${resourceType}_Token_text_idx_tsv` -

1. Add `_Token`
2. Align with our other indexes, where if the `_tsv` is implicit, we add it after `_idx`

codyebberson (changes requested):
Sorry for last minute change, but I think we should fix the index name.

codyebberson (approved):

PR #3806, opened 2024-01-24
fix-3803 add prefix to client storage

### LINKED ISSUE #3803: Namespace Medplum-managed local storage keys
Currently, calling `MedplumClient#clear` is quite destructive and blows away **all** local storage entries regardless of if they're Medplum-managed.

https://github.com/medplum/medplum/blob/1ac475829270460fdcc008e60a30e93c2d62d2ad/packages/core/src/client.ts#L805-L814

This is a side-effect of logging out:
https://github.com/medplum/medplum/blob/1ac475829270460fdcc008e60a30e93c2d62d2ad/packages/core/src/client.ts#L1116-L1120

I'd like to propose introducing a prefix to all local storage keys that are created by Medplum to enable `#clear` to be more surgical and only remove Medplum created/managed local storage entries. Additionally, this would reduce the likelihood of key clashing with other systems/extensions that leverage local storage.
i.e. `@medplum:${key}`
`@medplum:logins`
`@medplum:activeLogin`

This would be a breaking change if it became the default behavior and FWIW, the `MedplumClient` does already expose storage as an optional input so clients could specify their own local storage wrapper that applies key prefixes and a more surgical removal.

dillonstreator:
Fixes #3803

> [!WARNING]
> This is a breaking change for any client that depends on the local storage key names `activeLogin` & `logins` and as such should be highly visible in release notes, assuming there is alignment on proceeding with this change.

### DIFF
--- a/packages/core/src/storage.test.ts
+++ b/packages/core/src/storage.test.ts
@@ -1,6 +1,20 @@
 import { ClientStorage, MemoryStorage } from './storage';
 
 describe('Storage', () => {
+  test('clearing should keep non-medplum intact', () => {
+    const store = new MemoryStorage();
+    const storage = new ClientStorage(store);
+
+    store.setItem('foo', 'bar');
+    storage.setString('baz', 'qux');
+
+    expect(store.length).toEqual(2);
+
+    storage.clear();
+    expect(store.length).toEqual(1);
+    expect(store.getItem('foo')).toEqual('bar');
+  });
+
   test('Using localStorage', () => {
     const storage = new ClientStorage();
 
--- a/packages/core/src/storage.ts
+++ b/packages/core/src/storage.ts
@@ -18,34 +18,46 @@ export interface IClientStorage {
  */
 export class ClientStorage implements IClientStorage {
   private readonly storage: Storage;
+  private readonly prefix: string;
 
-  constructor(storage?: Storage) {
+  constructor(storage?: Storage, prefix?: string) {
     this.storage = storage ?? (typeof localStorage !== 'undefined' ? localStorage : new MemoryStorage());
+    this.prefix = prefix ?? '@medplum:';
+  }
+
+  private makeKey(key: string): string {
+    return this.prefix + key;
   }
 
   clear(): void {
-    this.storage.clear();
+    const keys = this.storage instanceof MemoryStorage ? this.storage.keys() : Object.keys(this.storage);
+
+    keys
+      .filter((key) => key.startsWith(this.prefix))
+      .forEach((key) => {
+        this.storage.removeItem(key);
+      });
   }
 
   getString(key: string): string | undefined {
-    return this.storage.getItem(key) ?? undefined;
+    return this.storage.getItem(this.makeKey(key)) ?? undefined;
   }
 
   setString(key: string, value: string | undefined): void {
     if (value) {
-      this.storage.setItem(key, value);
+      this.storage.setItem(this.makeKey(key), value);
     } else {
-      this.storage.removeItem(key);
+      this.storage.removeItem(this.makeKey(key));
     }
   }
 
   getObject<T>(key: string): T | undefined {
-    const str = this.getString(key);
+    const str = this.getString(this.makeKey(key));
     return str ? (JSON.parse(str) as T) : undefined;
   }
 
   setObject<T>(key: string, value: T): void {
-    this.setString(key, value ? stringify(value) : undefined);
+    this.setString(this.makeKey(key), value ? stringify(value) : undefined);
   }
 }
 
@@ -111,6 +123,10 @@ export class MemoryStorage implements Storage {
    */
   key(index: number): string | null {
     return Array.from(this.data.keys())[index];
+  }
+
+  keys(): string[] {
+    return Array.from(this.data.keys());
   }
 }
 


dillonstreator:
Test flake or did I break something 🤔 
image

packages/core/src/storage.ts
```diff
@@ -18,34 +18,46 @@ export interface IClientStorage {
...
+    this.prefix = prefix ?? '@medplum:';
+  }
+
+  private makeKey(key: string): string {
+    return this.prefix + key;
   }
 
   clear(): void {
-    this.storage.clear();
+    const keys = this.storage instanceof MemoryStorage ? this.storage.keys() : Object.keys(this.storage);
```
dillonstreator, line 33: While `Object.keys(this.storage)` works for localStorage, this is not safe for any `Storage` implementation. Might need to create a new interface that extends `Storage` and adds a key method.

packages/core/src/storage.ts
```diff
@@ -112,6 +124,10 @@ export class MemoryStorage implements Storage {
   key(index: number): string | null {
     return Array.from(this.data.keys())[index];
   }
+
+  keys(): string[] {
+    return Array.from(this.data.keys());
+  }
```
dillonstreator, line 130: need to add this to [`SyncSecureStorage`](https://github.com/medplum/medplum/blob/e283645d5444c344844d897b29120455da1ba2f1/packages/expo-polyfills/src/index.ts#L172) as well

packages/core/src/storage.ts
dillonstreator: https://github.com/medplum/medplum/pull/3806/commits/94daee94c5d7c35cd17842d35ac52ede45312e81 is an attempt at solving this issue

packages/core/src/client.ts
```diff
@@ -4196,7 +4196,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
         // Refresh the page to ensure the active login is up to date.
         if (e.key === null) {
           locationUtils.reload();
-        } else if (e.key === 'activeLogin') {
+        } else if (e.key === this.storage.makeKey('activeLogin')) {
```
ThatOneBro, line 4206: Had to make the storage listener aware of the prefix, otherwise this condition would never be hit

packages/core/src/storage.test.ts
```diff
@@ -2,6 +2,31 @@
 // SPDX-License-Identifier: Apache-2.0
 import { ClientStorage, MemoryStorage } from './storage';
 
+class TestStorage implements Storage {
```
codyebberson: Do we need this now?  It looks pretty close to `MemoryStorage`?

ThatOneBro: Removed all of these tests with the change to remove configurable prefix 👍

packages/core/src/storage.ts
```diff
@@ -20,24 +21,42 @@ export interface IClientStorage {
  */
 export class ClientStorage implements IClientStorage {
   private readonly storage: Storage;
+  private readonly prefix: string = '';
 
   constructor(storage?: Storage) {
-    this.storage = storage ?? (typeof localStorage !== 'undefined' ? localStorage : new MemoryStorage());
+    this.storage =
+      storage ?? (typeof globalThis.localStorage !== 'undefined' ? globalThis.localStorage : new MemoryStorage());
+    this.prefix = this.storage === globalThis.localStorage ? '@medplum:' : '';
```
codyebberson: We should probably preserve default behavior in `@medplum/core`.

We can change default behavior in `app`, as that doesn't have any backward compatibility requirements, but `core` by itself might have people depending on the current behavior.

packages/core/src/storage.ts
```diff
@@ -20,24 +21,42 @@ export interface IClientStorage {
...
+
+  makeKey(key: string): string {
+    return this.prefix + key;
   }
 
   clear(): void {
-    this.storage.clear();
+    // We only care about checking the keys for localStorage
+    // We only want to clear out keys that are Medplum keys
+    if (this.storage === globalThis.localStorage) {
```
codyebberson: This could also check `this.prefix`

```suggestion
    if (this.storage === globalThis.localStorage && this.prefix) {
```

packages/core/src/storage.ts
```diff
@@ -20,24 +21,42 @@ export interface IClientStorage {
  */
 export class ClientStorage implements IClientStorage {
   private readonly storage: Storage;
+  private readonly prefix: string = '';
 
   constructor(storage?: Storage) {
-    this.storage = storage ?? (typeof localStorage !== 'undefined' ? localStorage : new MemoryStorage());
+    this.storage =
+      storage ?? (typeof globalThis.localStorage !== 'undefined' ? globalThis.localStorage : new MemoryStorage());
```
codyebberson: If we use the `globalThis.` prefix, do we need to use `typeof`?

```suggestion
      storage ?? globalThis.localStorage ?? new MemoryStorage();
```

packages/core/src/client.ts
```diff
@@ -348,6 +348,13 @@ export interface MedplumClientOptions {
    * This can be used to set custom headers such as Cookies or Authorization headers.
    */
   defaultHeaders?: Record<string, string>;
+
+  /**
+   * Prefix to add to all keys when using `localStorage` as the backing store for `ClientStorage` (the default option in the browser).
+   *
+   * Default is `''` (no prefix).
+   */
+  storagePrefix?: string;
```
ThatOneBro, line 357: Added `storagePrefix` to `MedplumClient` constructor as discussed

packages/app/src/index.tsx
```diff
@@ -19,6 +19,7 @@ export async function initApp(): Promise<void> {
   const medplum = new MedplumClient({
     baseUrl: config.baseUrl,
     clientId: config.clientId,
+    storagePrefix: '@medplum:',
```
ThatOneBro, line 22: Added `storagePrefix` of `@medplum:` to `app`

PR #3808, opened 2024-01-25
fix-3807 rectify fhirpath `exists` and `empty`

### LINKED ISSUE #3807: Unexpected fhirpath result for empty `%previous` value with `exists()` and `empty()`
I believe there might be a bug in the fhirpath/functions `empty` & `exists` or in the way that an expected empty value is passed to `evalFhirPathTyped`.

I’m noticing the potential bug with the Subscription extension fhir-path-criteria-expression

We can see the `%previous` is specified as the following when there is no previous resource in the subscription extension:
https://github.com/medplum/medplum/blob/60425c132cf7e9492d6cff5c6ce5b453c1816ba1/packages/server/src/workers/utils.ts#L120

However, while `%previous.empty()` should be `true`, it is actually `false`.

We can see this with the following test:
```ts
  test('%previous.empty() returns true for an empty %previous value', () => {
    const patient: Patient = {
      resourceType: 'Patient',
    };
    const result = evalFhirPathTyped('%previous.empty()', [toTypedValue(patient)], { '%previous': toTypedValue({}) });

    expect(result).toEqual([
      {
        type: PropertyType.boolean,
        value: true,
      },
    ]);
  });
```

Additionally, `%previous.exists().not()` should also return `true` but it does not.

dillonstreator:
Fixes #3807


packages/core/src/fhirpath/functions.ts
```diff
@@ -34,7 +34,7 @@ export const functions: Record<string, FhirPathFunction> = {
    * @returns True if the input collection is empty ({ }) and false otherwise.
    */
   empty: (_context: AtomContext, input: TypedValue[]): TypedValue[] => {
-    return booleanToTypedValue(input.length === 0);
+    return booleanToTypedValue(input.length === 0 || input.every((e) => isEmpty(e.value)));
```
dillonstreator, line 37: This is effectively the same as the following:
```ts
    return booleanToTypedValue(input.every((e) => isEmpty(e.value)));
```

However, the current change is a bit more obvious IMO. Defer to maintainers on preference.

rahul1:
HI @dillonstreator, if you'd like to fix the prettier errors and fix conflicts, we can take another look

codyebberson (approved):

PR #3810, opened 2024-01-25
fix-3809 race condition in Subscription extension fhir-path-criteria-expression %previous value lookup

### LINKED ISSUE #3809: Subscription extension `fhir-path-criteria-expression` race condition could cause unexpected `%previous` value
https://github.com/medplum/medplum/blob/e283645d5444c344844d897b29120455da1ba2f1/packages/server/src/workers/utils.ts#L109-L125

In this code, there is a race condition that exists which could theoretically cause `%previous` to point to the same as `%current` or even newer records.

In a system with many subscriptions for the same resource type all using the `fhir-path-criteria-expression` could hit this race condition if updating the same resource concurrently.

https://github.com/medplum/medplum/blob/e283645d5444c344844d897b29120455da1ba2f1/packages/server/src/workers/subscription.ts#L143-L157

The problem here being that we could end up with unexpected Subscription calls (or lack there of).

The search for the previous entry should be a bit more sophisticated by leveraging last updated timestamp or version id to access the correct ‘previous’ value.
https://github.com/medplum/medplum/blob/e283645d5444c344844d897b29120455da1ba2f1/packages/server/src/fhir/repo.ts#L1101-L1102

dillonstreator:
Fixes #3809

### DIFF
--- a/packages/server/src/workers/utils.ts
+++ b/packages/server/src/workers/utils.ts
@@ -114,20 +114,27 @@ export async function isFhirCriteriaMet(subscription: Subscription, currentResou
   if (!criteria?.valueString) {
     return true;
   }
-  const history = await systemRepo.readHistory(currentResource.resourceType, currentResource?.id as string);
+  const previous = await getPreviousResource(currentResource);
   const evalInput = {
     '%current': toTypedValue(currentResource),
-    '%previous': toTypedValue({}),
+    '%previous': toTypedValue(previous ?? {}),
   };
-  const previousResource = history.entry?.[1]?.resource as Resource;
-  if (previousResource) {
-    evalInput['%previous'] = toTypedValue(previousResource);
-  }
   const evalValue = evalFhirPathTyped(criteria.valueString, [toTypedValue(currentResource)], evalInput);
-  if (evalValue?.[0]?.value === true) {
-    return true;
+  return evalValue?.[0]?.value === true;
+}
+
+async function getPreviousResource(currentResource: Resource): Promise<Resource | undefined> {
+  const history = await systemRepo.readHistory(currentResource.resourceType, currentResource?.id as string);
+  if ((history.entry?.length ?? 1) === 1) {
+    return undefined;
   }
-  return false;
+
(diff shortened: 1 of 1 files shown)


packages/server/src/workers/utils.ts
```diff
@@ -114,20 +114,27 @@ export async function isFhirCriteriaMet(subscription: Subscription, currentResou
...
   const evalValue = evalFhirPathTyped(criteria.valueString, [toTypedValue(currentResource)], evalInput);
-  if (evalValue?.[0]?.value === true) {
-    return true;
+  return evalValue?.[0]?.value === true;
+}
+
+async function getPreviousResource(currentResource: Resource): Promise<Resource | undefined> {
+  const history = await systemRepo.readHistory(currentResource.resourceType, currentResource?.id as string);
+  if ((history.entry?.length ?? 1) === 1) {
+    return undefined;
```
dillonstreator, line 129: this is redundant and can be removed as the following find implementation would already return `undefined` in this case

packages/server/src/workers/utils.ts
```diff
@@ -114,20 +114,25 @@ export async function isFhirCriteriaMet(subscription: Subscription, currentResou
   if (!criteria?.valueString) {
     return true;
   }
-  const history = await systemRepo.readHistory(currentResource.resourceType, currentResource?.id as string);
+  const previous = await getPreviousResource(currentResource);
   const evalInput = {
     '%current': toTypedValue(currentResource),
-    '%previous': toTypedValue({}),
+    '%previous': toTypedValue(previous ?? {}),
```
mattwiller, line 120: I'm curious if something this might work better:
```suggestion
    '%previous': previous ? toTypedValue(previous) : undefined,
```

dillonstreator: We may want to address that change in https://github.com/medplum/medplum/pull/3808 but FWIW, the `variables` typing as input to `evalFhirPathTyped` would not allow a value of `undefined`. It must be a `TypedValue`.

codyebberson (approved):

PR #3824, opened 2024-01-27
Separate search popup menus for 'text' and 'token'

codyebberson:
Removes "sort" and "contains" from `token` search parameters.

"Priority":

image

Confirmed that our "Priority Order" search params still work as expected:

image


reshmakh (approved):

rahul1:
Why are we removing sorting for token parameters? This is a valid FHIR search

codyebberson:
> Why are we removing sorting for token parameters? This is a valid FHIR sort

Yes, it is valid FHIR sort, and the API will continue to support it.

It is a relatively expensive query.  Yesterday, users clicked on it incidentally, just exploring.  I.e., sorting `Observation` by `code` when looking at 3 million rows.  That can cause significant server load without anyone realizing it.

The better long term solution is to improve the performance of those queries, but for now the goal is to subtly steer people away from it.

PR #3834, opened 2024-01-29
fix-3815 cache presigned s3 binary urls

### LINKED ISSUE #3815: `Communication.payload[x].contentAttachment.url` new signed S3 url on every request regardless of expiration
Fetching the `Communication` resource returns new signed S3 URLs for every attachment regardless of URL expiration. In applications that implement communication polling and just generally less sophisticated means of updating chat, this can result in large amounts of unnecessary network bandwidth consumption as the native browser caching mechanism is bypassed.

A quick and dirty workaround that I implemented was to wrap `@medplum/react#AttachmentDisplay` to add a simple caching mechanism:

```tsx
import React, { useMemo } from 'react';
import { AttachmentDisplay, AttachmentDisplayProps } from '@medplum/react';

// Maintain a cache of urls to avoid unnecessary re-download of attachments
// The following is a workaround for the fact that each `Communication` request
// returns a NEW signed S3 URL for each attachment bypassing the native browser caching mechanism
// resulting in unnecessary bandwidth consumption.
// https://www.medplum.com/docs/fhir-datastore/binary-data#consuming-a-fhir-binary-in-an-application
// https://github.com/medplum/medplum/issues/3815
const urls = new Map<string, string>();

export const AttachmentDisplayCached = ({ value, ...rest }: AttachmentDisplayProps): JSX.Element => {
  const url = value?.url;
  const finalUrl = useMemo(() => {
    if (!url) {
      return url;
    }

    const resourceUrl = url.split('?')[0];

    const cachedUrl = urls.get(resourceUrl);
    if (cachedUrl) {
      const searchParams = new URLSearchParams(new URL(cachedUrl).search);
      // This is fairly brittle as it relies on the current structure of the Medplum returned URL
      const expires = searchParams.get('Expires');
      // `expires` is in seconds, Date.now() is in ms
      if (expires && parseInt(expires, 10) * 1000 > Date.now()) {
        return cachedUrl;
      }
    }

    urls.set(resourceUrl, url);
    return url;
  }, [url]);

  return <AttachmentDisplay {...rest} value={{ ...value, url: finalUrl }} />;
};
```

Curious - is there any appetite to include this caching mechanism into the `AttachmentDisplay` component by default?

dillonstreator:
Fixes #3815


packages/react-hooks/src/useCachedS3BinaryUrl/useCachedS3BinaryUrl.ts
```diff
@@ -0,0 +1,42 @@
...
+// https://www.medplum.com/docs/fhir-datastore/binary-data#consuming-a-fhir-binary-in-an-application
+// https://github.com/medplum/medplum/issues/3815
+
+// TODO: possibly retain this cache in and restore from local storage to persist across page refreshes?
+// The S3 presigned URLs expire after 1 hour with the default configuration and hard refreshes are not uncommon even in SPAs so this
+// could be a good way to get additional cache hits
+// This would require additional logic for initialization, saving, and purging of expired keys
+const urls = new Map<string, string>();
+
+export const useCachedS3BinaryUrl = (s3Url: string | undefined): string | undefined => {
```
codyebberson, line 16: Any objection to using `useCachedBinaryUrl`? (i.e., remove the "S3")

Given that this will be part of our public API, I'm compelled to make it backend agnostic.

codyebberson, line 16: Maybe rename `s3Url` to `binaryUrl`?

packages/react-hooks/src/useCachedS3BinaryUrl/useCachedS3BinaryUrl.ts
```diff
@@ -0,0 +1,42 @@
...
+    return useMemo(() => {
+        if (!s3Url) {
+            return undefined;
+        }
+
+        const binaryResourceUrl = s3Url.split('?')[0];
+        if (!binaryResourceUrl) {
+            return undefined;
```
codyebberson, line 24: I believe this should return `s3Url` (or `originalUrl` or whatever) - the fallback should be to the original value.

packages/react-hooks/src/useCachedS3BinaryUrl/useCachedS3BinaryUrl.ts
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        const cachedUrl = urls.get(binaryResourceUrl);
+        if (cachedUrl) {
+            const searchParams = new URLSearchParams(new URL(cachedUrl).search);
+            // This is fairly brittle as it relies on the current structure of the Medplum returned URL
```
codyebberson, line 30: Good callout...

If you want to tighten it up, we could add some additional checks:
* Check for presence of `Key-Pair-Id` query param
* Check for presence of `Signature` query param
* Validate that `Expires` is numeric
* Validate that `Expires` is within some sensible range

I don't think any of those are strictly necessary.  Your call.

packages/react/src/AttachmentDisplay/AttachmentDisplay.tsx
```diff
@@ -1,13 +1,15 @@
...
 
 export interface AttachmentDisplayProps {
   value?: Attachment;
   maxWidth?: number;
 }
 
 export function AttachmentDisplay(props: AttachmentDisplayProps): JSX.Element | null {
-  const { contentType, url, title } = props.value ?? {};
+  const { contentType, url: uncachedUrl, title } = props.value ?? {};
+  const url = useCachedS3BinaryUrl(uncachedUrl);
```
codyebberson, line 12: Nice, this looks clean.

package-lock.json
codyebberson, line 1: I think you can revert changes to `package-lock.json`, given that there were no changes to any `package.json` files.

codyebberson (commented):
Thanks @dillonstreator  - this looks great

2 requests:
* Let's rename `useCachedS3BinaryUrl` to `useCachedBinaryUrl` to try to be backend agnostic.
* Let's change the fallback behavior to return the original value

Otherwise lgtm 👍 

Thank you!

codyebberson (approved):

PR #3835, opened 2024-01-29
Search control counts fallback

rahul1:


codyebberson (commented):
We should do this in the server, not in the search control.  We should try to minimize the amount of logic in the search control component.

rahul1:
Closing this PR as obsolete. Smaller pieces of this might make it into main, but not the whole thing as-is

PR #3850, opened 2024-01-30
Fixes #3794 - MeasureReport.period search

### LINKED ISSUE #3794: Properly search over Period elements
## Background
User Driven: https://medplum.slack.com/archives/C057GEVQCRE/p1706043851251609

The user requested query is "find all measure reports where `MeasureReport.period` includes today"

Currently, when storing `Period` elements (e.g. MeasureReport.period), we only index the `start` of the time range. This means that we can't implement the proper time overlap search semantics defined [here](https://hl7.org/fhir/R5/search.html#date)

## Proposal
- Index `Period.end` for search
- Implement FHIR search semantics for period: return all resources where the `Period` element overlaps the queried range

codyebberson:
Implementing proper `SearchParameter.type="date"` search behavior, but only for `MeasureReport.period`

Based on @mattwiller's doc here: https://gist.github.com/mattwiller/6167035b0e5e3e53fad1eb2141b876df

The intent is for this behavior to be rolled out to all `SearchParameter.type="date"` search parameters, but that will require a series of database migrations and reindexing steps.


packages/core/src/fhirpath/utils.ts
```diff
@@ -411,14 +411,61 @@ export function fhirPathIs(typedValue: TypedValue, desiredType: string): boolean
...
+ * @param input - Unknown input value.
+ * @returns A Period object or undefined.
+ */
+export function toPeriod(input: unknown): Period | undefined {
+  if (!input) {
+    return undefined;
+  }
+
+  if (isDateString(input)) {
+    return { start: input + 'T00:00:00.000Z', end: input + 'T23:59:59.999Z' };
```
mattwiller, line 457: This won't correctly handle the case where the `date` is specified only to month precision:
```
postgres=# select '2024-01T12:34:56Z'::timestamptz;
ERROR:  invalid input syntax for type timestamp with time zone: "2024-01T12:34:56Z"
```

rahul1 (approved):

PR #3868, opened 2024-02-01
feat: Support input Parameters for Patient-everything

jdjkelly:
This is a draft PR, just looking for some feedback on implementing input Parameters for an Operation. I'll write test coverage before moving out of draft.

## Why
* Being able to use `_since` on Patient-everything is very useful and it's supported in the OperationDefinition
* It would be nice in general if the implementations of Operations covered as much of the OperationDefinition input parameter spec as possible

## What
* Define a `PatientEverythingParameters` type and read the OperationDefinition, modelled after the implementations of `CodeSystemLookup` and `CodeSystemValidate`
* Implements just `_since` to start because it's the easiest
* Read parameters with `parseInputParameters` and extend `getPatientEverything` to support params input

### DIFF
--- a/packages/server/src/fhir/operations/patienteverything.ts
+++ b/packages/server/src/fhir/operations/patienteverything.ts
@@ -13,6 +13,14 @@ import { getAuthenticatedContext } from '../../context';
 import { getPatientCompartments } from '../patient';
 import { Repository } from '../repo';
 import { sendResponse } from '../response';
+import { parseInputParameters } from './utils/parameters';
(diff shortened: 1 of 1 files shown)


codyebberson:
Strategy looks good to me

@mattwiller  - As resident expert on FHIR operation input params, care to comment?

mattwiller:
All looks good to me as well!

codyebberson:
Merged with https://github.com/medplum/medplum/pull/3908

PR #3869, opened 2024-02-01
Refactor elements context and slicing logic

mattlong:
This PR is a replacement for and just about a strict subset of #3760. It excludes the changes to various `onChange` callbacks which had been made in preparation for handling fixed/pattern values.

This PR tightens up the abstractions around the `ElementsContext`, formerly called `BackboneElementContext`. One big improvement is the proper integration of parent contexts into sub-contexts, which will be foundational to properly incorporating fixed/pattern values from structure definitions into resources. Related and a more noticeable developer-facing change is the addition of the required `path` property on most input components to facilitate usage of `ElementsContext`.


packages/react/src/ElementsInput/ElementsInput.tsx
```diff
@@ -1,91 +1,94 @@
...
   readonly onChange: ((value: any) => void) | undefined;
   readonly testId?: string;
 }
 
 export function ElementsInput(props: ElementsInputProps): JSX.Element {
-  const { elements } = props;
   const [value, setValue] = useState<any>(props.defaultValue ?? {});
+  const elementsContext = useContext(ElementsContext);
+  const elementsToRender = useMemo(() => {
+    const result = Object.entries(elementsContext.elements).filter(([key, element]) => {
```
mattwiller, line 30: Minor recommendation to factor this big filter lambda into a named function

packages/react/src/ElementsInput/ElementsInput.tsx
```diff
@@ -1,91 +1,94 @@
...
-    const result: { [key: string]: InternalSchemaElement & { fixed: TypedValue } } = Object.create(null);
-    for (const [key, property] of Object.entries(elements)) {
-      if (property.fixed) {
-        result[key] = property as any;
+      if (element.max === 0) {
+        return false;
       }
-    }
+
+      // mostly for Extension.url
```
mattwiller, line 39: Would it be possible to explicitly do this only for `Extension.url`?

packages/react/src/ElementsInput/ElementsInput.utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  elements: InternalTypeSchema['elements'] | undefined;
+  parentPath: string;
+  parentContext: ElementsContextType | undefined;
+  parentType: string | undefined;
+  profileUrl?: string;
+  debugMode?: boolean;
+};
+
+function hasFixed(element: InternalSchemaElement): element is InternalSchemaElement & { fixed: TypedValue } {
+  return Boolean(element.fixed);
```
mattwiller, line 60: This check could be stricter to align with the predicate in the signature:
```suggestion
  return element.fixed?.type && Object.hasOwnProperty(element.fixed, 'value');
```

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.test.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      expect(slices.length).toBe(2);
+      expect(slicedValues.map((sliceValues) => sliceValues.length)).toEqual([1, 1, 0]);
+    });
+  });
+
+  // test('extensions', () => {});
+});
+
+// const values = [
```
mattwiller, line 225: Could be removed?

mattwiller (approved):

PR #3873, opened 2024-02-01
Document how to switch to super admin project in projects

ksmith94:
This will add a note to the projects docs on how to switch to the super admin project or check if you are already in it.


rahul1 (approved):
One comment, then this is good to go!

PR #3875, opened 2024-02-01
Expand profile operation

mattlong:
When working with a profile, it is often necessary to also have the schemas of any included extension profiles. To reduce server roundtrips, we add a new operation `StructureDefinition$expandProfile` that iterates through a given profile looking for other profiles referenced in elements' `type`.

Don't be too scared of the huge diff in `search.test.ts`; it is mostly a change from `systemRepo` to `repo`

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -1682,16 +1682,11 @@ export class MedplumClient extends EventTarget {
         if (options?.expandProfile) {
           const url = this.fhirUrl('StructureDefinition', '$expand-profile');
           url.search = new URLSearchParams({ url: profileUrl }).toString();
-          const sds = await new ReadablePromise(
-            this.get<Bundle<StructureDefinition>>(url.toString()).then(bundleToResourceArray)
-          );
-
-          const result: string[] = [];
-          for (const sd of sds) {
+          const sdBundle = await this.get<Bundle<StructureDefinition>>(url.toString());
+          return bundleToResourceArray(sdBundle).map((sd) => {
             loadDataType(sd, sd.url);
-            result.push(sd.url);
-          }
-          return result;
+            return sd.url;
+          });
         } else {
           // Just sort by lastUpdated. Ideally, it would also be based on a logical sort of version
           // See https://hl7.org/fhir/references.html#canonical-matching for more discussion
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -67,7 +67,7 @@ import {
 import { ReadablePromise } from './readablepromise';
 import { ClientStorage, IClientStorage } from './storage';
 import { indexSearchParameter } from './types';
-import { indexStructureDefinitionBundle, isDataTypeLoaded, isProfileLoaded } from './typeschema/types';
+import { indexStructureDefinitionBundle, isDataTypeLoaded, isProfileLoaded, loadDataType } from './typeschema/types';
 import {
   CodeChallengeMethod,
   ProfileResource,
@@ -588,6 +588,11 @@ export interface ValueSetExpandParams {
   date?: string;
   offset?: number;
   count?: number;
+}
+
+export interface RequestProfileSchemaOptions {
+  /** (optional) Whether to include nested profiles, e.g. from extensions. Defaults to false. */
+  expandProfile?: boolean;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


packages/core/src/client.ts
```diff
@@ -1658,32 +1663,50 @@ export class MedplumClient extends EventTarget {
...
-          url: profileUrl,
-          _sort: '-_lastUpdated',
-        });
-
-        if (!sd) {
-          console.warn(`No StructureDefinition found for ${profileUrl}!`);
+        if (options?.expandProfile) {
+          const url = this.fhirUrl('StructureDefinition', '$expand-profile');
+          url.search = new URLSearchParams({ url: profileUrl }).toString();
+          const sds = await new ReadablePromise(
```
codyebberson, line 1685: I think this could be simplified (maybe)

If you omit `then(bundleToResourceArray)`, then `this.get()` will return a `Bundle<StructureDefinition>`

Then you could also omit `new ReadablePromise()`

Then you could replace the `for (const sd of sds)` with a call to `indexStructureDefinitionBundle()` (which basically just calls `loadDataType()` in a for loop.

The one thing you would still need to do is get a list of the URLs, but that might be simple enough.

```ts
return sdBundle.entry?.map((e) => e.resource.url);
```

packages/server/src/fhir/operations/structuredefinitionexpandprofile.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  return response;
+}
+
+function bundleResults(profiles: StructureDefinition[]): Bundle<StructureDefinition> {
+  const entry: BundleEntry<StructureDefinition>[] = [];
+
+  for (const profile of profiles) {
+    entry.push({
+      fullUrl: `${getConfig().baseUrl}fhir/R4/StructureDefinition/${profile.id}`,
```
codyebberson, line 101: There is a `getFullUrl()` helper in `packages/server/src/fhir/response.ts`

packages/server/src/fhir/operations/structuredefinitionexpandprofile.ts
```diff
@@ -0,0 +1,111 @@
...
+      ?.map((t) => t.profile)
+      .flat()
+      .filter((p): p is NonNullable<string> => p !== undefined);
+
+    profileUrls?.forEach((p) => {
+      profilesUrlsToLoad.push(p);
+    });
+  });
+
+  const promises: Promise<StructureDefinition | undefined>[] = profilesUrlsToLoad.map((url) =>
```
codyebberson, line 75: This could be combined into a single query using comma separated values to the `url` parameter.

packages/server/src/fhir/operations/structuredefinitionexpandprofile.ts
```diff
@@ -0,0 +1,111 @@
...
+  const sds = await Promise.all(promises);
+  for (const result of sds) {
+    if (result === undefined) {
+      continue;
+    }
+
+    response.push(result);
+
+    // recursive loop
+    const nested = await loadNestedStructureDefinitions(repo, result);
```
codyebberson, line 89: Not urgent, but we may want to add a circuit breaker here.

At minimum, we should make sure that we don't search for URLs that we've already searched for.

We may want to add a "max depth", with some sensible limit like 10 or 20 deep.

codyebberson (commented):
Structurally looks good.  Mostly minor suggestions

The circuit breaker on recursive searches is probably the most important one

I'm also assuming that the big diff to `search.test.ts` is mostly whitespace?  I checked a handful of tests, and they all appeared to be.

packages/core/src/client.ts
mattlong: That would result in all the StructureDefinitions being loaded within the top-level profile's URL namespace which isn't how it works right now; each extension has `loadDataType` called with its own url. That decision had been made a while back without enough consideration...

I'd need to think through it a bit more, but I believe we could  actually make the change to have all extensions loaded/index behind the top-level profile's URL. Either way, I'd like to hold off on that change from this PR since it'd require touching some other code.

packages/server/src/fhir/operations/structuredefinitionexpandprofile.ts
codyebberson: Yep yep, makes sense.

packages/core/src/client.ts
codyebberson: Copy, sounds good.  Thanks

codyebberson (approved):

PR #3877, opened 2024-02-02
Maintain full `traceparent` header value for `traceId`

dillonstreator:
Addresses #3878

This PR proposes moving to maintain the full [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) header value in logs as opposed to stripping just the traceId and converting it to hyphen separated UUID. By having the Medplum server extract just the `traceId` portion of the `traceparent` header *and* changing its format makes it much more difficult to tie logs together between distributed systems.

Additionally, this PR proposes allowing a slight divergence from the `traceparent` spec and allowing the `version` prefix and `flags` suffix to be optional. Allowing this will enable clients using Sentry to more easily adopt distributed tracing as Sentry's tracing integrations do not currently strictly adhere to the `traceparent` format/spec and does not respect the `version` prefix.
https://github.com/getsentry/sentry-javascript/blob/84baeb1847b0cd90af43ba8b29517b1b6cb24854/packages/utils/src/tracing.ts#L117-L131
https://github.com/getsentry/sentry-javascript/blob/84baeb1847b0cd90af43ba8b29517b1b6cb24854/packages/utils/src/tracing.ts#L6-L14
Another note about Sentry is that it does not currently follow the spec w/r/t the `flags` format as it only suffixes with either `-1` or `-0`.
Ideally, Sentry SDKs would properly adhere to the `traceparent` spec (https://github.com/getsentry/team-sdks/issues/41) and alternatively, Medplum calling clients could intercept and properly format but this seems like a reasonable middle-ground. Ultimately, I'll understand if Medplum does not want to open itself to this kind of scope creep and maintenance of vendor specific inconsistencies.


packages/server/src/context.ts
```diff
@@ -101,29 +101,48 @@ export function closeRequestContext(): void {
...
       id.substring(20, 32),
     ].join('-');
-    if (isUUID(uuid)) {
-      traceId = uuid;
+
+    return isUUID(uuid) && uuid;
+  },
+  // https://github.com/getsentry/sentry-javascript/blob/6dfddc2dfc4b74cbb70dc325b4af165e25904ed4/packages/utils/src/tracing.ts#L117-L130
+  // {uuidv4}-{uuidv4.substring(16)}[-(0|1)]
+  'sentry-trace': (traceId: string) => {
```
dillonstreator, line 126: This is basically a `traceparent` value minus the `version`. The client should just send a valid traceparent value.

packages/server/src/context.ts
```diff
@@ -101,29 +101,48 @@ export function closeRequestContext(): void {
...
+    return (traceId.length === 51 || traceId.length === 49) && traceId;
```
dillonstreator, line 127: Should implement a true traceparent parsing maybe with optional version and flag?

packages/server/src/traceparent.ts
```diff
@@ -0,0 +1,30 @@
...
+type twohex = `${hex}${hex}`;
+
+export type Traceparent =
+  | { valid: false }
+  | {
+      valid: true;
+      version: twohex | null;
+      traceId: string;
+      parentId: string;
+      flags: twohex | null;
```
dillonstreator, line 13: `twohex` is an invalid typing for `flags` as the regular expression allows 1 character in it's current form

mattwiller (approved):
Nice, this looks great — thanks @dillonstreator!

codyebberson (approved):

PR #3884, opened 2024-02-02
Remove global systemRepo

codyebberson:


packages/server/src/fhir/repo.ts
```diff
@@ -1811,6 +1812,10 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
     this.closed = true;
   }
 
+  [Symbol.dispose](): void {
+    this.close();
+  }
```
mattwiller, line 1817: Nice, I didn't know this pattern existed

mattwiller (approved):

PR #3886, opened 2024-02-04
fix-3885 propagate `traceId` through asynchronous jobs

### LINKED ISSUE #3885: Propagate trace ID through to asynchronous jobs
I think it would be valuable to propagate the `traceId` through to the download and subscription asynchronous jobs for the logs that they generate as well as to drill it through to the external requests that they make. This would further tie together logs and enable a clearer view into a requests full life-cycle and impact in the system.

PS - I love all of  the work that has been done so far to enable observability! https://github.com/medplum/medplum/discussions/2759

dillonstreator:
Fixes #3885

This is based off of https://github.com/medplum/medplum/pull/3877 to include the work done to better support `traceparent` and slight variations seen in the wild but it does not need to depend on it.

- [x] subscriptions: rest-hooks
- [x] subscriptions: bots lambda
- [x] subscriptions: bots vmcontext
- [x] downloads

Questions:
- Thoughts on including changes to the documentation to specify that bots and rest subscription calls would be receiving an `x-trace-id` & `traceparent` header? Could be a future documentation change consideration.
- How to handle crons? I feel like it makes less sense to propagate the initial traceId that creates a cron to all future cron executions. Should this just be ignored?


codyebberson:
> Questions:
> 
> * Thoughts on including changes to the documentation to specify that bots and rest subscription calls would be receiving an `x-trace-id` & `traceparent` header? Could be a future documentation change consideration.

Good question.  I think we need to add more documentation on tracing in general, so I'm ok to punt to future work.

> * How to handle crons? I feel like it makes less sense to propagate the initial traceId that creates a cron to all future cron executions. Should this just be ignored?

Agree, I think cron is out of scope for trace ID, let's just ignore.

Thanks @dillonstreator !

codyebberson (commented):
Thanks @dillonstreator  - this all looks good to me.  My focus was on the download and subscription workers, as those are quite critical to production workflows.

@mattwiller  - can you double check the request ID and trace ID parts?  They look good to me, but you're the expert on those.

mattwiller (approved):
All looks good to me!  One thing to consider for the future is how we might centralize some of the async job handling, since this PR required changing multiple places to ensure everything got wired up correctly

PR #3887, opened 2024-02-05
Bugfix: Search Control Pagination End

rahul1:


packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -597,5 +601,5 @@ function getTotal(search: SearchRequest, lastResult: Bundle): number {
       (lastResult.entry?.length ?? 0) +
       (lastResult.link?.some((l) => l.relation === 'next') ? 1 : 0);
   }
-  return total;
+  return total ?? 0;
```
codyebberson, line 604: Minor, but I think this is unnecessary?  `total` should always be a number at this point.

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -577,15 +577,19 @@ function getPage(search: SearchRequest): number {
...
+  return total > 0 ? Math.ceil(total / pageSize) : 1;
 }
 
 function getStart(search: SearchRequest, lastResult: Bundle): number {
   return Math.min(getTotal(search, lastResult), (search.offset ?? 0) + 1);
 }
 
 function getEnd(search: SearchRequest, lastResult: Bundle): number {
-  return Math.min(getTotal(search, lastResult), ((search.offset ?? 0) + 1) * (search.count ?? DEFAULT_SEARCH_COUNT));
+  return Math.min(
```
codyebberson, line 589: What was the change here?

I don't have any memory of why this function needs to call `getTotal()`... Is it necessary here?

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -577,15 +577,19 @@ function getPage(search: SearchRequest): number {
 
 function getTotalPages(search: SearchRequest, lastResult: Bundle): number {
   const pageSize = search.count ?? DEFAULT_SEARCH_COUNT;
-  return Math.ceil(getTotal(search, lastResult) / pageSize);
+  const total = getTotal(search, lastResult);
+  return total > 0 ? Math.ceil(total / pageSize) : 1;
```
codyebberson, line 581: What is the change here?  It seems like the only time the return value would be different is if `total === 0`, before it would return `0`, now it will return `1`?  Is that desired?

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -333,8 +333,8 @@ export function SearchControl(props: SearchControlProps): JSX.Element {
           </Group>
           <Group gap={2}>
             {lastResult && (
-              <Text size="xs" c="dimmed">
-                {getStart(search, lastResult)}-{getEnd(search, lastResult)}
+              <Text size="xs" c="dimmed" data-testid="count-display">
+                {getStart(search, lastResult).toLocaleString()}-{getEnd(search, lastResult).toLocaleString()}
```
codyebberson, line 337: This is a good catch 👍

packages/mock/src/client.ts
```diff
@@ -229,15 +230,21 @@ round-trip min/avg/max/stddev = 10.977/14.975/23.159/4.790 ms
...
   initialized = false;
   initPromise?: Promise<void>;
 
   constructor(
     readonly router: FhirRouter,
     readonly repo: MemoryRepository,
-    readonly debug = false
-  ) {}
+    readonly debug = false,
+    readonly mockHandlerOverride?: MockRequestHandler
```
codyebberson, line 242: 🫠 very mixed feelings about adding more complexity to `MockRepository`.... When it was just something that we used internally for testing, that was one thing.  As more people are using it in more places, I think we need to be a lot more thoughtful about API design here.

packages/react/src/SearchControl/SearchControl.tsx
rahul1: Confirmed that this is not desired - i had a misunderstanding about how Mantine `Pagination` handles 0 pages

rahul1: Will revert

packages/react/src/SearchControl/SearchControl.tsx
rahul1: The main change: 
```ts
- ((search.offset ?? 0) + 1) * (search.count ?? DEFAULT_SEARCH_COUNT)
+ getStart(search, lastResult) + (search.count ?? DEFAULT_SEARCH_COUNT) - 1
``` 

The multiplication logic was incorrect for larger numbers: 

e.g. 
- pageSize: 20
- offset 100K
- total: 200K

Before: 
end = (100K) * (20) = 2M

After:
end = 100K + 20 = 100,020

packages/mock/src/client.ts
rahul1: Yeah I don't have strong opinions on: 
1. This as the API design
2. Allowing this kind of extension at all

The main thing I was solving for was how to mock out really large counts for the `SearchControl`, without actually creating a ton of resources. 

Open to suggestions

packages/react/src/SearchControl/SearchControl.tsx
rahul1: VSCode red squigglies did me dirty - not sure why the flagged this

packages/react/src/SearchControl/SearchControl.tsx
rahul1: getTotal() is needed in this context: 

count: 20
offset 10
total: 25


Desired output: 
start=10
end=25 (not 30)

codyebberson (approved):

PR #3890, opened 2024-02-06
feat(subscriptions): add client utils for WebSocket subscriptions

### LINKED ISSUE #3804: Add `subscribeToCriteria()` to `MedplumClient`; add `useSubscription()` hook to `@medplum/react-hooks`

ThatOneBro:
## Done:
- [x] Finish tests

## In another PR:
- [ ] Add docs for `useSubscription`, `SubscriptionManager`, `SubscriptionEmitter` (in followup PR)

This PR closes #3804

### DIFF
--- a/examples/medplum-live-chat-demo/src/components/Chat.tsx
+++ b/examples/medplum-live-chat-demo/src/components/Chat.tsx
@@ -1,15 +1,9 @@
 import { ActionIcon, Avatar, Group, Paper, ScrollArea, Stack, TextInput } from '@mantine/core';
 import { showNotification } from '@mantine/notifications';
-import {
-  MedplumClient,
-  ProfileResource,
-  createReference,
-  getReferenceString,
-  normalizeErrorString,
-} from '@medplum/core';
-import { Bundle, Communication, Parameters, Practitioner, Reference, Subscription } from '@medplum/fhirtypes';
+import { ProfileResource, createReference, getReferenceString, normalizeErrorString } from '@medplum/core';
+import { Bundle, Communication, Practitioner, Reference } from '@medplum/fhirtypes';
 import { DrAliceSmith } from '@medplum/mock';
-import { Form, useMedplum } from '@medplum/react';
+import { Form, useMedplum, useSubscription } from '@medplum/react';
 import { IconArrowRight, IconChevronDown, IconMessage } from '@tabler/icons-react';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import classes from './Chat.module.css';
@@ -49,44 +43,6 @@ function upsertCommunications(
   setCommunications(newCommunications);
 }
 
-async function listenForSub(
-  medplum: MedplumClient,
-  subscription: Subscription,
-  setWebSocket: (ws: WebSocket) => void,
-  onNewMessage: (c: Communication) => void
-): Promise<void> {
-  const { parameter } = (await medplum.get(
-    `/fhir/R4/Subscription/${subscription.id}/$get-ws-binding-token`
-  )) as Parameters;
-  const token = parameter?.find((param) => param.name === 'token')?.valueString;
-  const url = parameter?.find((param) => param.name === 'websocket-url')?.valueUrl;
-  if (!token) {
-    throw new Error('Failed to get token!');
-  }
-
-  if (!url) {
-    throw new Error('Failed to get URL from $get-ws-binding-token!');
-  }
-
-  const ws = new WebSocket(url);
-
-  ws.addEventListener('open', () => {
-    ws.send(JSON.stringify({ type: 'bind-with-token', payload: { token } }));
-  });
-
-  ws.addEventListener('message', (event: MessageEvent<string>) => {
-    const bundle = JSON.parse(event.data) as Bundle;
-    const communication = bundle.entry?.[1]?.resource;
-    if (!communication || communication.resourceType !== 'Communication') {
-      console.error('Invalid chat bundle!');
-      return;
-    }
-    onNewMessage(communication);
-  });
-
-  setWebSocket(ws);
-}
-
 export function Chat(): JSX.Element | null {
   const medplum = useMedplum();
   const [open, setOpen] = useState(false);
@@ -94,22 +50,28 @@ export function Chat(): JSX.Element | null {
   const inputRef = useRef<HTMLInputElement>(null);
   const scrollAreaRef = useRef<HTMLDivElement>(null);
   const [profile, setProfile] = useState(medplum.getProfile());
-  const [subscription, setSubscription] = useState<Subscription | undefined>(undefined);
-  const [webSocket, setWebSocket] = useState<WebSocket | undefined>();
-
-  const creatingSubRef = useRef(false);
-  const deleteSubTimerRef = useRef<NodeJS.Timeout | undefined>();
 
   const profileRefStr = useMemo<string>(
     () => (profile ? getReferenceString(medplum.getProfile() as ProfileResource) : ''),
     [profile, medplum]
   );
 
+  useSubscription(
+    `Communication?sender=${profileRefStr},${DR_ALICE_SMITH.reference}&recipient=${DR_ALICE_SMITH.reference},${profileRefStr}`,
+    (bundle: Bundle) => {
+      if (!bundle.entry?.[1]) {
+        return;
+      }
+      const communication = bundle.entry?.[1]?.resource as Communication;
+      upsertCommunications(communicationsRef.current, [communication], setCommunications);
+    }
+  );
+
   // Disabled because we can make sure this will trigger an update when local profile !== medplum.getProfile()
   // eslint-disable-next-line react-hooks/exhaustive-deps
   useEffect(() => {
     const latestProfile = medplum.getProfile();
-    if (profile !== latestProfile) {
+    if (profile?.id !== latestProfile?.id) {
       setProfile(latestProfile);
     }
   });
@@ -117,9 +79,6 @@ export function Chat(): JSX.Element | null {
   const communicationsRef = useRef<Communication[]>(communications);
   communicationsRef.current = communications;
   const prevCommunicationsRef = useRef<Communication[]>(communications);
-
-  const openRef = useRef<boolean>();
-  openRef.current = open;
 
   const scrollToBottomRef = useRef<boolean>(false);
 
@@ -135,66 +94,6 @@ export function Chat(): JSX.Element | null {
     );
     upsertCommunications(communicationsRef.current, searchResult, setCommunications);
   }, [medplum, profileRefStr]);
-
-  useEffect(() => {
-    // Create subscription...
-    // Check for creatingSubRef
-    if (!(profile && profileRefStr) || creatingSubRef.current) {
-      return () => undefined;
-    }
-    if (!subscription) {
-      creatingSubRef.current = true;
-      medplum
-        .createResource<Subscription>({
-          resourceType: 'Subscription',
-          criteria: `Communication?sender=${profileRefStr},${DR_ALICE_SMITH.reference}&recipient=${DR_ALICE_SMITH.reference},${profileRefStr}`,
-          status: 'active',
-          reason: `Watch for Communications between ${profileRefStr} and ${DR_ALICE_SMITH.reference}.`,
-          channel: {
-            type: 'websocket',
-          },
-        })
-        .then((subscription) => {
-          setSubscription(subscription);
-          listenForSub(medplum, subscription, setWebSocket, (communication) => {
-            upsertCommunications(communicationsRef.current, [communication], setCommunications);
-            // NOTE: We may normally want to do a guard like this to prevent our client from updating messages that we have sent ourselves, but in this case
-            // We allow them to be updated anyways so that we have received timestamps on our sent messages
-            // const senderId = resolveId(communication.sender);
-            // if (senderId === profile.id) {
-            //   return;
-            // }
-            // You may want to update received time and "completed" status independently, but for the purposes of this demo we are updating them together
-            if (!(communication.received && communication.status === 'completed')) {
-              medplum
-                .updateResource<Communication>({
-                  ...communication,
-                  received: new Date().toISOString(), // Mark as received
-                  status: 'completed', // Mark as read
-                  // See: https://www.medplum.com/docs/communications/organizing-communications#:~:text=THE%20Communication%20LIFECYCLE
-                  // for more info about recommended `Communication` lifecycle
-                })
-                .catch(console.error);
-            }
-          })
-            .then(() => {
-              creatingSubRef.current = false;
-            })
-            .catch(console.error);
-        })
-        .catch(console.error);
-    }
-    if (deleteSubTimerRef.current) {
-      clearTimeout(deleteSubTimerRef.current);
-      deleteSubTimerRef.current = undefined;
-    }
-    return () => {
-      if (deleteSubTimerRef.current) {
-        return;
-      }
-      deleteSubTimerRef.current = setTimeout(() => {}, 1000);
-    };
-  }, [medplum, profile, profileRefStr, subscription]);
 
   const sendMessage = useCallback(
     async (formData: Record<string, string>) => {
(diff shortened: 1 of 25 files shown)


examples/medplum-live-chat-demo/vite.config.ts
```diff
@@ -11,4 +12,11 @@ export default defineConfig({
     host: 'localhost',
     port: 3000,
   },
+  resolve: {
+    alias: {
+      '@medplum/core': path.resolve(__dirname, '../../packages/core/src'),
```
codyebberson, line 17: This is useful during development, but we can't commit it / can't publish it to the example repo.  Not sure if there's a way to make this conditional?  Simple option is to just remove before finalizing.

packages/react-hooks/src/useSubscription/useSubscription.ts
```diff
@@ -0,0 +1,55 @@
+import { MedplumClient, SubscriptionEmitter, SubscriptionEventMap } from '@medplum/core';
+import { Bundle } from '@medplum/fhirtypes';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { useMedplum } from '../MedplumProvider/MedplumProvider.context';
+
+const SUBSCRIPTION_DEBOUNCE_MS = 3000;
+
+export function useSubscription(criteria: string, callback: (bundle: Bundle) => void): void {
```
codyebberson, line 8: This is getting kinda heavy with 3 `useEffect`... Feels like we should be able to slim this down.  In my mind, it should be possible for this to be a very lightweight operation, so that we can freely call `useSubscription` without worrying about performance overhead.

packages/react-hooks/src/useSubscription/useSubscription.ts
```diff
@@ -0,0 +1,55 @@
...
+        setEmitter(undefined);
+      }, SUBSCRIPTION_DEBOUNCE_MS);
+    };
+  }, [medplum, criteria]);
+
+  useEffect(() => {
+    prevCriteriaRef.current = criteria;
+  }, [criteria]);
+
+  const emitterCallback = useCallback((event: SubscriptionEventMap['message']) => {
```
codyebberson, line 38: What is the advantage of this over calling `callbackRef.current()` directly?

packages/react-hooks/src/useSubscription/useSubscription.ts
```diff
@@ -0,0 +1,55 @@
+import { MedplumClient, SubscriptionEmitter, SubscriptionEventMap } from '@medplum/core';
+import { Bundle } from '@medplum/fhirtypes';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { useMedplum } from '../MedplumProvider/MedplumProvider.context';
+
+const SUBSCRIPTION_DEBOUNCE_MS = 3000;
+
+export function useSubscription(criteria: string, callback: (bundle: Bundle) => void): void {
+  const medplum = useMedplum() as MedplumClient;
+  const [emitter, setEmitter] = useState<SubscriptionEmitter>();
```
codyebberson, line 10: Why do we need the emitter?  Why not just call callback directly?

examples/medplum-live-chat-demo/src/components/Chat.tsx
```diff
@@ -49,67 +43,32 @@ function upsertCommunications(
...
-
-  const creatingSubRef = useRef(false);
-  const deleteSubTimerRef = useRef<NodeJS.Timeout | undefined>();
 
   const profileRefStr = useMemo<string>(
     () => (profile ? getReferenceString(medplum.getProfile() as ProfileResource) : ''),
     [profile, medplum]
   );
 
+  useSubscription(
```
codyebberson, line 59: This is pretty slick, love this API.

packages/core/src/subscriptions/index.ts
```diff
@@ -0,0 +1,249 @@
...
+export class SubscriptionEmitter extends TypedEventTarget<SubscriptionEventMap> {
+  private criteria: Set<string>;
+  constructor(...criteria: string[]) {
+    super();
+    this.criteria = new Set(criteria);
+  }
+  getCriteria(): Set<string> {
+    return this.criteria;
+  }
+  [kAddCriteria](criteria: string): void {
```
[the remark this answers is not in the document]
ThatOneBro: The idea was to to not expose this method to users, adding criteria here can lead to getting into confusing states

packages/react-hooks/src/useSubscription/useSubscription.ts
ThatOneBro: Having to rebind any time the callback changes, otherwise every rerender the event listener has a callback bound and unbound. In this way, we can bind one time and then it doesn't matter even if the callback is a new function every render

examples/medplum-live-chat-demo/vite.config.ts
ThatOneBro: I'll remove them for now and we can add them back in a separate PR and maybe add a step to remove them when syncing the example repos? Just saves so much time

packages/react-hooks/src/useSubscription/useSubscription.ts
ThatOneBro: Talked about this offline:

Good call raising concern about multiple `useEffect`, they add up quickly. We determined that these should be pretty light and shouldn't rerun except when logging out or if the criteria changes... 

I did reduce the amount of effects by 33% (3 -> 2) so that should help a little 👍

packages/react-hooks/src/useSubscription/useSubscription.ts
ThatOneBro: Talked about this offline 👍 think we're all good here

examples/medplum-live-chat-demo/src/components/Chat.tsx
ThatOneBro: Yeah I think this should be a pretty powerful abstraction while remaining relatively performant. Can't wait to see what comes out of this

packages/core/src/subscriptions/index.ts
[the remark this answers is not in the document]
codyebberson: Can we just market `private`?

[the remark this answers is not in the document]
ThatOneBro: If it's private it wouldn't be accessible to the `SubscriptionManager` which is the class that calls it 🤔 . Using a Symbol hides it from the public interface while still making it accessible to those who know to use its symbol, this is how Node deals with private methods that need to be accessible to other internal classes

packages/core/src/client.ts
```diff
@@ -2222,13 +2225,13 @@ export class MedplumClient extends EventTarget {
...
   ): Promise<any> {
-    let url;
+    let url: string | URL;
     if (typeof idOrIdentifier === 'string') {
       const id = idOrIdentifier;
       url = this.fhirUrl('Bot', id, '$execute');
     } else {
       const identifier = idOrIdentifier;
-      url = this.fhirUrl('Bot', '$execute') + `?identifier=${identifier.system}|${identifier.value}`;
+      url = this.fhirUrl('Bot', '$execute').toString() + `?identifier=${identifier.system}|${identifier.value}`;
```
codyebberson, line 2234: Good catch.

We should probably use the `URL` for this rather than concatenation query params due to escaping concerns.

I.e..,

```ts
      url = this.fhirUrl('Bot', '$execute');
      url.searchParams.set('identifier', identifier.system + '|' + identifier.value);
```

packages/core/src/client.ts
```diff
@@ -2903,7 +2906,8 @@ export class MedplumClient extends EventTarget {
       }
       await sleep(retryDelay);
     }
-    return this.parseResponse(resultResponse as Response, 'POST', statusUrl);
+    // @ts-expect-error resultResponse will be defined by this point, since loop will spin until a result response is received
+    return this.parseResponse(resultResponse, 'POST', statusUrl);
```
codyebberson, line 2910: I think this is just a cosmetic change?  This will be a merge conflict with https://github.com/medplum/medplum/pull/3937 but I think that's ok.

packages/core/src/subscriptions/index.ts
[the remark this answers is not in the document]
codyebberson: I understand the motivation here, but I just want to flag it as unnecessary complexity.  Sometimes the additional protections are not worth the cognitive overhead.  Your call.

packages/core/src/subscriptions/index.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+
+export class SubscriptionManager {
+  private readonly medplum: MedplumClient;
+  private ws: WebSocket;
+  private masterSubEmitter?: SubscriptionEmitter;
+  private subEmitters: Map<string, SubscriptionEmitter>; // Map<criteria, SubscriptionEmitter>
+  private refCounts: Map<string, number>; // Map<criteria, refCount>
+  private subscriptionCriteriaLookup: Map<string, string>; // Map<subscriptionId, criteria>
+  private criteriaSubscriptionLookup: Map<string, string>; // Map<criteria, subscriptionId>
```
codyebberson, line 57: Rather than using `Map` for everything, I wonder if we could add a helper class or type which would contain the logic.  It could be quite small, be private/internal, and manage this state to avoid the lookups.

```ts
class SubscriptionCriteria {
  readonly criteria: string;
  readonly subscriptionId: string;
  readonly emitter: SubscriptionEmitter;
  refCount: number;
}
```

packages/core/src/subscriptions/index.ts
```diff
@@ -0,0 +1,259 @@
...
+  private wsClosed: boolean;
+
+  constructor(medplum: MedplumClient, wsOrUrl: WebSocket | string) {
```
codyebberson, line 60: Looking at all of the call sites for `new SubscriptionManager`, and specifically cases where the 2nd arg is a `WebSocket`...

Is the `WebSocket` option ever required?  It looks like this could always take a `string`, and then you could simplify the constructor and eliminate a bunch of tests.

packages/core/src/subscriptions/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    this.setupWebSocketListeners();
+  }
+
+  private setupWebSocketListeners(): void {
+    const ws = this.ws;
+
+    ws.addEventListener('message', (event: MessageEvent) => {
+      const bundle = JSON.parse(event.data) as Bundle;
+      // Get criteria for event
+      const status = bundle?.entry?.[0]?.resource as SubscriptionStatus;
```
codyebberson, line 97: 1. We should probably treat this as `SubscriptionStatus | undefined` to be safe.  Maybe we can count on Medplum server to always include the status, but probably not other servers.
2. Out of curiosity, what happens if this callback function throws?  Does it gracefully recover, or does it kill the connection?

codyebberson (commented):
Nice, thanks @ThatOneBro  - there's a lot going on here, and it's a pretty solid implementation.

Most of my comments are nit picky / style things.  Let me know if you'd like to discuss any.

Tangentially related, maybe future work, but I wonder if we could use this in `packages/agent` as well, given that there's quite a bit of overlap in some of the connection management / reconnect / retry stuff.

Overall, this looks quite good.  Let me know if you have any questions on the superficial stuff.

packages/core/src/subscriptions/index.ts
ThatOneBro: Yes sometimes you need the caller to have a handle on the `WebSocket`, particularly for testing purposes, but I could see that you might want that in general as well.

packages/core/src/subscriptions/index.ts
ThatOneBro: We still will need the lookups, there is no other way to associate `criteria` and `subscriptionId` since these exist separately and you cannot get the `criteria` for `subscription` without the lookup and vice versa.

ThatOneBro: The `subscription -> criteria` is needed since the `SubscriptionStatus` only contains a reference string and not the full resource as a convention. So we need to cache the associated `criteria` so we know which callback to call when a notification comes in

packages/core/src/client.ts
ThatOneBro: Sonar labels this as a useless type assertion, using `@ts-expect-error` both makes it more clear why we can ignore this type error and makes sure that if it's no longer necessary it will throw a type error on the comment

packages/core/src/subscriptions/index.ts
ThatOneBro: 1. It's part of the [spec](https://build.fhir.org/bundle.html#subscription-notification:~:text=A%20subscription%20notification%20(type%20%3D%20%22subscription%2Dnotification%22)%20consists%20of%20a%20series%20of%201%20or%20more%20entries%2C%20the%20first%20of%20which%20is%20a%20SubscriptionStatus.) that `SubscriptionStatus` is the first element in the bundle, and without the `SubscriptionStatus` it's not possible to resolve which callback should be called. We can handle it as an error but it complicates the code and would only be possible if communicating with a non-compliant server.

ThatOneBro: 2. This is typically handled by the `EventEmitter` (or `EventTarget`) implementation, but that's a valid question. I'll add a test for that

packages/core/src/subscriptions/index.ts
ThatOneBro: Actually looks like I might have removed the tests where I was doing that, good call. We can probably simplify that.

packages/core/src/subscriptions/index.ts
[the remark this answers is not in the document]
ThatOneBro: Would you prefer if we just prefix with an underscore, ie. `_addCriteria`?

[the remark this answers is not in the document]
ThatOneBro: The point is to flag that this is meant to be an internal method and using it might result in undesired behavior.

packages/core/src/subscriptions/index.ts
ThatOneBro: The class might still be useful though

packages/core/src/subscriptions/index.ts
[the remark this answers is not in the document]
ThatOneBro: Talked offline: going to use JSDoc `@internal` and `_` prefix instead 👍

packages/core/src/subscriptions/index.ts
ThatOneBro: Now we accept `URL` or `string` here, and simplified the code 👍

packages/core/src/subscriptions/index.ts
ThatOneBro: Apparently `EventTarget` lets uncaught exceptions get thrown to `UncaughtException` handler, so I don't think we should add logic like that (Node's `EventEmitter` does emit errors to `error` event listeners if they exist before throwing them to the `UncaughtException` handler).

Instead I wrapped everything in a try catch within the handler and emitted `error` from our callback in the case of something like an invalid message being received from the WebSocket.

packages/core/src/client.ts
codyebberson: Oo, nice, that's a good trick.  TIL

packages/core/src/subscriptions/index.ts
```diff
@@ -0,0 +1,342 @@
...
+export type SubscriptionEventMap = {
+  connect: { type: 'connect'; payload: { subscriptionId: string } };
+  disconnect: { type: 'disconnect'; payload: { subscriptionId: string } };
+  error: { type: 'error'; payload: Error };
+  message: { type: 'message'; payload: Bundle };
+  close: { type: 'close' };
+  heartbeat: { type: 'heartbeat'; payload: Bundle };
+};
+
+export type RobustWebSocketEventMap = {
```
codyebberson, line 16: `Robust` 🤔 

Is this a nod to partysocket? 😛

codyebberson (approved):
Nice work @ThatOneBro  - this really came together, looks super clean.  Thank you 🙏

packages/core/src/subscriptions/index.ts
ThatOneBro: Possibly, its definitely inspired by `partysocket`

PR #3891, opened 2024-02-07
Fix return documentation for client read methods

josh-:
This PR fixes the `@returns` comment for several methods in the medplum client.

For example, `readResource` is typed as returning `ReadablePromise<ExtractResource<K>>` and it's promise will reject when a resource isn't available.

for example, the following will print `promise rejected` instead of fulling the promise and printing `undefined` (as the `@returns` comment currently implies)

```ts
medplum
  .readResource('Patient', 'invalid-id-here')
  .then((patient) => {
    console.log(patient);
  })
  .catch(() => {
    console.log('promise rejected');
  });
```


mattwiller (approved):
✨

PR #3909, opened 2024-02-08
Enable linking between Projects

mattwiller:
This PR adds the ability to link one Project to another.  Linking a Project makes the resources contained in the linked Project available (read-only) to the host Project.  Only Super Admin is allowed to modify this field


packages/server/src/fhir/accesspolicy.ts
```diff
@@ -18,30 +19,36 @@ import { applySmartScopes } from './smart';
  * This method ensures that the repository is setup correctly.
  * @param login - The user login.
  * @param membership - The active project membership.
- * @param strictMode - Optional flag to enable strict mode for in-depth FHIR schema validation.
+ * @param project - The Project the current user is a member of.
```
codyebberson, line 22: 👍 

This function is ripe for a cleanup refactor.  Thanks

packages/server/src/fhir/repo.ts
```diff
@@ -107,11 +107,11 @@ export interface RepositoryContext {
   remoteAddress?: string;
 
   /**
-   * The current project reference.
-   * This should be the ID/UUID of the current project.
+   * Projects that the Repository is allowed to access.
+   * This should include the ID/UUID of the current project, but may also include other accessory Projects.
```
codyebberson, line 111: We may want to extend this comment, because there's a fair amount of hidden complexity here.

1. `projects` can be undefined or empty, which is only allowed for system user (and super admins?)
2. `projects` can be length 1, indicating the simple classic case of a user without any linked projects
3. `projects` can have additional length, indicating a primary project and additional linked projects

In any case where `projects.length > 0`, the first project is the "primary" project?

packages/server/src/fhir/repo.ts
```diff
@@ -602,13 +602,15 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
-
-    if (projectId) {
-      // Try retrieving from cache
-      const cachedProfile = await getProfileCacheEntry(projectId, url);
-      if (cachedProfile) {
-        return cachedProfile.resource;
+    const projectIds = this.context.projects;
+
+    if (projectIds?.length) {
+      for (const projectId of projectIds) {
```
codyebberson, line 608: Low priority, but this could be converted to a Redis `mget` for a single parallelized operation.

packages/server/src/fhir/repo.ts
```diff
@@ -1000,8 +1002,8 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
    * @param builder - The select query builder.
    */
   private addProjectFilters(builder: SelectQuery): void {
-    if (this.context.project) {
-      builder.where('compartments', 'ARRAY_CONTAINS', [this.context.project, r4ProjectId], 'UUID[]');
+    if (this.context.projects?.length) {
+      builder.where('compartments', 'ARRAY_CONTAINS', [...this.context.projects, r4ProjectId], 'UUID[]');
```
codyebberson, line 1006: Ok, i wondered about this -- Here we're going to just always add `r4ProjectId` at filter time.

Would it make more sense to add `r4ProjectId` to `this.context.projects` in the constructor?

packages/server/src/fhir/repo.ts
```diff
@@ -1597,7 +1599,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
     if (protectedResourceTypes.includes(resourceType)) {
       return false;
     }
-    if (resource.meta?.project !== this.context.project) {
+    if (resource.meta?.project !== this.context.projects?.[0]) {
```
codyebberson, line 1602: 👍  - can only write to "primary" project ✅

codyebberson (approved):
Nice, this looks good to me.  Clean and surgical

A bunch of minor non-blocking comments.  Probably the biggest one would be to add some additional commentary on the nuances of the `projects` array, as there is some implicit hidden complexity in there.

PR #3918, opened 2024-02-09
Apply pattern and fixed values to profile resources

### LINKED ISSUE #3732: Apply fixed/pattern values in ResourceForm
How should fixed/pattern values be applied to a resource in `ResourceForm`? At a high level, the two most likely approaches would either be:

1. at load time, i.e. as the `ResourceForm` and the tree of input components included in it are setup. 
2. at save time, i.e. when the user clicks OK to save the resource, merge in all fixed/pattern element values into the resource before submitting it to the server to be validated and saved.

The load-time option is preferential since it would allow feedback to the user of which values are fixed and should not (or cannot) be edited so that they don't attempt to fill in values that would only inevitably lead to validation errors. At a high level, the algorithm would be as follows:

* For required array elements (`element.min > 0`), create up to `min` placeholder values
	* For elements with fixed/pattern values within a required element, apply the fixed/pattern value to the placeholder.
* For optional array elements (`element.min === 0`), do NOT create any placeholder values
	* For elements with fixed/pattern values within an optional element, only apply them to existing entries
* When a new array/slice entry is added (i.e. the user clicks the Add button), for both required and optional elements, apply same logic as above.
* For non-array elements, (`element.max === 1`), apply fixed/pattern values, but track that they were applied be default and do not include them in a save request if the user doesn't change anything else in the element.

The application of fixed/pattern values should be centralized rather controlled by individual inputs. The most obvious location is within `ElementContext` since all elements shown in a form are contained within an `ElementContext`. Since applying cardinality & fixed/pattern values requires profile & extensions schemas to be loaded, it must be possible for fixed/pattern values to be applied at various depths within a resource; not just at `ResourceForm` or the top-most `BackboneElementInput`. `ElementContext` also fits nicely with this requirement since a new context is already necessary when a type schema is fetched from the server.

When displaying fixed/pattern elements:
* Display the input in a disabled/read-only state with a tooltip explaining "This value is required by the US Core Patient profile"
* If the resource has an existing value that does NOT conform to the fixed/pattern value, highlight it as an error with a message/question mark icon explaining it does not conform along with a button to set the value to the fixed/pattern value. The guiding principle here is that an existing value should never be changed automatically; the user must explicitly opt-in to the change.

What to do with custom input components that ignore some sub-elements that could have a non-zero `min` or fixed/pattern values? For example, `QuantityInput` shows `value, comparator, unit` but ignores `system, code`. They should mostly just work as if they were hidden form fields. The one exception being when the existing element values that do NOT conform to fixed/pattern values. In that case, a message should be shown for the input similarly as described above: a message is shown for the `QuantityInput` explaining that there the non-conforming values along with a button to set the hidden values to the required fixed/pattern values.

mattlong:
There are a few different scenarios for working with pattern & fixed values that this PR addresses. The [US Core Blood Pressure Profile](https://hl7.org/fhir/us/core/STU5.0.1/StructureDefinition-us-core-blood-pressure.html) conveniently demonstrates most of them.

### Scenario 1: Fixed/pattern values on required elements
To conform to a given profile, these values must be included in the resource; i.e. the chain of elements to them all have `element.min > 0`. From the blood pressure profile, this includes `Observation.code.coding.{system,code}`:

```
"code": {
    "coding": [
      {
        "system": "http://loinc.org",
        "code": "85354-9"
      }
    ]
  }
```

As well as `Observation.category.coding.{system,code}` within the `VSCat` slice of `Observation.category`.

```
"category": [
    {
      "coding": [
        {
          "system": "http://terminology.hl7.org/CodeSystem/observation-category",
          "code": "vital-signs"
        }
      ]
    }
  ]
```

And finally, `Observation.component.code.coding.{system,code}` in the `systolic` and `diastolic` slices of `Observation.component`:

```
"component": [
    {
      "code": {
        "coding": [
          {
            "system": "http://loinc.org",
            "code": "8480-6",
          }
        ]
      }
    },
    {
      "code": {
        "coding": [
          {
            "system": "http://loinc.org",
            "code": "8462-4",
          }
        ]
      }
    }
  ]
```

The values for all these elements in included when creating a new, conforming blood pressure observation. This is achieved with combination of a new `SchemaCrawler` and `DefaultValueVistor` that traverses the elements and slices of a StructureDefinition in a hierarchical way, applying all required pattern/fixed values along the way.

Note, there are other required elements (with `element.min > 1`), `Observation.{subject,effective[x]}`, but since they do not have fixed nor pattern values, no values are pre-populated for them. If the user does not enter a value, they will be caught at validation time.

### Scenario 2: Fixed/pattern values on optional elements

To conform to a given profile, these values must be included in the resource **only when related/peer elements are specified**. Put another way, while the leaf elements do indeed have `element.min > 0`, somewhere along chain of parent elements above them are not required, i.e. they have `element.min === 0`.  From the blood pressure profile, this includes `Observation.component.value[x].{system,code}` in the `systolic` and `diastolic` components. In english, a blood pressure observation isn't required to include actual readings (a `dataAbsentReason` can be specified instead), but if the readings **are** included, they must be done in `mm[Hg]` units:

```
"component": [
    {
      "valueQuantity": {
        "code": "mm[Hg]",
        "system": "http://unitsofmeasure.org"
      }
    },
    {
      "valueQuantity": {
        "code": "mm[Hg]",
        "system": "http://unitsofmeasure.org"
      }
    }
  ]
```

These elements receive a different treatment from Scenario 1 since they should NOT be included in the resource when being saved UNLESS the user fills out the peer required elements `Observation.component.value[x].{value,unit}`. However, the values should still be visually shown in the form since they would indeed be required values should the other inputs in the form be edited.

To achieve this, these values are added to the `defaultValue` property that `ElementDefinitionTypeInput` specifies for all complex child components. As such, the fixed/pattern values will only be rolled up into the resource if/when the user edits peer elements of the complex type (e.g. enters 124 and 76 for the systolic and diastolic readings). When the edits happen, the `onChange` callbacks includes the fixed/pattern values.

#### Action item
Something that has not yet been implemented is handling the case when a user edits the peer elements (thus rolling up the fixed/pattern values into the resource in `ResourceForm`) but then changes their mind and deletes the peer values. In this case, the fixed/pattern values will remain on the resource even though that is very likely not the intention of the user. e.g. they could have a technically valid blood pressure observation, but `Observation.component.value[x].value` is missing making it effectively an incomplete observation. To address this, we'd need to do something like adding logic in `onChange` callbacks of complex types that evaluates if only fixed/pattern fields remain and remove them as well if no other optional elements have values.

### Scenario 3: Fixed/pattern values on optional slice/array elements

This scenario is quite similar to Scenario 2, but with the twist that there is not a static slice/array entry count and the user needs to be able to hit the "Add" button and have each newly added array element have fixed/pattern values properly applied. Consider the various extension on the [US Core Patient Profile](https://hl7.org/fhir/us/core/STU5.0.1/StructureDefinition-us-core-patient.html). Within the [US Core Race Extension](https://hl7.org/fhir/us/core/STU5.0.1/StructureDefinition-us-core-race.html), up to five entries can exist for the `ombCategory` extension and an unlimited number of entries can exist for the `detailed` extension. For each entry that does exist, their respective `Patient.extension.extension.url` elements are fixed and must be included.

Fortunately, given the React component structure and element context, this can now be handled by the exact same mechanisms as Scenario 2. Before, it required special handling of applying the fixed `url` in the chain of `onChange` callbacks.

### Other notes
The new `SchemaCrawler` is somewhat similar in spirit to `ResourceCrawler`. It could perhaps eventually make sense to merge the two, but for now they are different enough that perhaps the only action item is clarifying the naming of the two.

The code to support applying fixed/pattern values was put in `@medplum/core` since it is something we eventually would want to expose as a server operation in addition to usage with `@medplum/react`. That being said, there is no reason the code couldn't be moved to `@medplum/react` for now if there is concern over bloating core. If nothing else, the current separation from React code helps drive home the point that it is not and should not be tangled in with React components, design patterns, etc.

Fixes #3732


packages/generator/src/storybook.ts
```diff
@@ -97,10 +97,10 @@ export function main(): void {
     // To build USCore, download and expand a USCore Implementation Guide package file,
     // such as https://hl7.org/fhir/us/core/STU5.0.1/package.tgz which is linked to
     // from https://hl7.org/fhir/us/core/STU5.0.1/downloads.html
-    buildUSCoreStructureDefinitions(
-      '/absolute/path/to/expanded/package-file',
-      resolve(__dirname, '../../mock/src/mocks/uscore/uscore-v5.0.1-structuredefinitions.json')
-    );
+    buildUSCoreStructureDefinitions('/absolute/path/to/expanded/package-file', [
+      resolve(__dirname, '../../mock/src/mocks/uscore/uscore-v5.0.1-structuredefinitions.json'),
```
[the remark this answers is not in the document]
codyebberson: Yeah... 🤔 

I support this as-is for now.  Let's make a note of this as part of a planning exercise for figuring out to do with core+mock.

packages/react/src/BackboneElementInput/BackboneElementInput.tsx
```diff
@@ -22,33 +23,35 @@ export interface BackboneElementInputProps {
...
-        path={props.path}
-        type={type}
-        defaultValue={defaultValue}
-        onChange={props.onChange}
-        outcome={props.outcome}
-      />
-    </ElementsContext.Provider>
+  return maybeWrapWithContext(
+    ElementsContext.Provider,
+    contextValue,
```
codyebberson, line 48: Trying to parse when `contextValue === undefined` here

The result of the `useMemo` will only be undefined when `typeSchema` is undefined, yeah?

And we're already checking for `!typeSchema` above?

So we could probably add `if (!typeSchema || !contextValue)` to that check?

packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
```diff
@@ -41,15 +51,19 @@ export interface ResourcePropertyInputProps {
...
+  const { property, name, onChange } = props;
   const defaultPropertyType =
     props.defaultPropertyType && props.defaultPropertyType !== 'undefined'
       ? props.defaultPropertyType
       : property.type[0].code;
-
   const propertyTypes = property.type as ElementDefinitionType[];
 
-  if ((property.isArray || property.max > 1) && !props.arrayElement) {
+  const isArrayInput = (property.isArray || property.max > 1) && !props.arrayElement;
```
codyebberson, line 61: At this point, is `property` the `InternalElementDefinition` from the original `StructureDefinition`?  Or is it from the profile?  I'm just trying to wrap my head around which checks we may / may not need.

I'm pretty sure that `property.isArray` implies `property.max > 1`, but I'm not 100% sure how profiles fit into that.

packages/core/src/utils.ts
```diff
@@ -683,6 +683,24 @@ export function isLowerCase(c: string): boolean {
   return c === c.toLowerCase() && c !== c.toUpperCase();
 }
 
+export function isComplexTypeCode(code: string): boolean {
```
codyebberson, line 686: Here "complex type" just means "not a primitive"?

I know this came up in Slack the other day, but we may want to do a sanity check on `System.String`, which may not behave as expected here.  I believe (??) that we normalize `System.String` to `string` in the schema loader, but that might be worth double checking.

packages/core/src/utils.ts
```diff
@@ -606,7 +606,7 @@ function deepIncludesObject(value: { [key: string]: unknown }, pattern: { [key:
  * @returns A deep clone of the input.
  */
 export function deepClone<T>(input: T): T {
-  return JSON.parse(JSON.stringify(input)) as T;
+  return input === undefined ? input : (JSON.parse(JSON.stringify(input)) as T);
```
codyebberson, line 609: Nice catch 👍

packages/core/src/elements-context.ts
```diff
@@ -0,0 +1,85 @@
+import { InternalSchemaElement } from './typeschema/types';
+import { getPathDifference } from './utils';
+
+export type ElementsContextType = {
```
codyebberson, line 4: Can we add some comments to this?  This type seems important.

packages/core/src/elements-context.ts
```diff
@@ -0,0 +1,85 @@
...
+  path: string;
+  profileUrl: string | undefined;
+  elements: Record<string, InternalSchemaElement>;
```
codyebberson, line 7: I need to dig into this one more... It seems like this would be duplicative to all of the existing code we have with `InternalTypeSchema.elements`

Or, in other words... is this something that could have been added to `InternalTypeSchema` directly to avoid creating more internal schema types?

packages/core/src/utils.ts
mattlong: Yeah, I added a test `expect(isComplexTypeCode(PropertyType.SystemString)).toEqual(false);` to make sure it's handled correctly. It didn't require any special handling in the function itself since `PropertyType.SystemString === 'http://hl7.org/fhirpath/System.String'` which does indeed start with a lower-case letter 🙃

packages/core/src/elements-context.ts
mattlong: Yes, that's a great idea. The usage of ElementsContext has slowly expanded and definitely deserves some documentation TLC.

packages/core/src/elements-context.ts
mattlong: I added more documentation to ElementsContext like you suggested elsewhere would help clear this up. ElementsContext exists to normalize working with elements across resources, extensions, and slices...really just normalizing traversing values either outside of or within slices. I don't believe there's a straightforward change that could be made to `InternalTypeSchema` to achieve anything similar. All the functions like `getElementDefinition`, `getDataType` revolve around `InternalTypeSchema`, but that just fundamentally doesn't work in slices where there isn't a corresponding `InternalTypeSchema`.

We could do something like make a pseudo `InternalTypeSchema` for slices that also get injected into DATA_TYPES, `getDataType`, etc., but that would definitely be violating the abstraction and pretty blatantly lying to the computer.

packages/react/src/BackboneElementInput/BackboneElementInput.tsx
mattlong: Maybe the naming needs to be improved or a more explicit helper function needs to be added, but `buildElementsContext` can return `undefined` as well if it detects that the context it is being asked to build would be identical to it's parent context.

packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
mattlong: Checking both `property.isArray` and `property.max > 1` pre-dated all the profile work.

`isArray` [is defined](https://github.com/medplum/medplum/blob/b770cf9c8e50ed829ccb98b91805dc9ca975abb1/packages/core/src/typeschema/types.ts#L483) as `ed.base?.max ?? ed.max > 1` which lines up with [the spec](https://hl7.org/fhir/R4/profiling.html#cardinality):

> When a profile is constraining another profile where there are more cardinality options (e.g. low is not just 0 or 1, and high is not just 1 or *), the same principles still apply: the constraining profile can only allow what the base profile allows.

> Note that though a profile can constrain an element from x..* to x..1, this doesn't make any difference to the representation in the JSON format - the element will still be represented in an array. As an example, take Patient.name which has a cardinality of 0..*. In an unprofiled Patient, this will be represented as:

```javascript
{
  "resourceType" : "Patient",
	"name" : [{
	   "text" : "Peter James"
	}]
}
```

> Even if a profile is created on the resource that narrows the cardinality to 1..1, applications will still process the resource without knowledge of the profile. For this reason the representation will still be the same.

So we could simplify the check to the following if we wanted to.

```
const isArrayInput = property.isArray && !props.arrayElement;
```

mattlong: And to answer your original question, `property` could be coming from either a profile or the base `StructureDefinition`.

packages/core/src/elements-context.ts
mattlong: I think the eventual final form is that something like `ElementsContextType` is eventually merged into `InternalTypeSchema` and `SliceDefinition` to unify the two.

packages/core/src/utils.ts
codyebberson: ah, right, lol

packages/core/src/default-values.ts
```diff
@@ -0,0 +1,439 @@
...
+  }
+
+  if (value === undefined) {
+    delete parent[resolvedKey];
+  } else {
+    parent[resolvedKey] = value;
+  }
+}
+
+function getValueAtKey(
```
codyebberson, line 308: We can revisit this in future PR, but this method has a lot of overlap with `getTypedPropertyValue`

packages/core/src/default-values.ts
```diff
@@ -0,0 +1,439 @@
...
+  elements: Record<string, InternalSchemaElement>
+): any {
+  const keyParts = key.split('.');
+  let last: any = value;
+  let answer: any;
+  for (let i = 0; i < keyParts.length; i++) {
+    let keyPart = keyParts[i];
+    if (keyPart.includes('[x]')) {
+      const keyPartElem = elements[keyParts.slice(0, i + 1).join('.')];
+      // should this loop through all possible types instead of using type[0]?
```
codyebberson, line 321: Yes, which would be handled with `getTypedPropertyValue`

codyebberson (approved):
Nice, this looks good to ship.  Thanks for powering through

Non-blocking follow-ups:

1. Explore using `getTypedPropertyValue` in `setValueAtKey`
2. Explore moving the US Core JSON into a shared mock directory.  This might be a bigger conversation about merging `mock` and `fhir-router` into `core` as sub-exports.
3. Explore merging `SchemaCrawler` and `ResourceCrawler` - I agree with your call in this PR, that they are sufficiently different.  I do wonder if they could be merged in a way that doesn't feel too clunky

You may want to poke @mattwiller for an additional review on some of the changes in `typeschema`.  I'm less familiar with that code.

packages/core/src/default-values.ts
mattlong: Yeah, part of the problem (and why I added `getValueAtKey` is that `getTypedPropertyValue` doesn't return the `Object.is` equivalent value in some cases (e.g. `isEmpty(resultValue)` and `Array.isArray(resultValue)` which was important for this code.

PR #3919, opened 2024-02-10
Fixes #3906 - Respect Bot.runAsUser in $execute operation

### LINKED ISSUE #3906: Bot.runAsUser has no effect when executed via $execute endpoint
As far as I can tell, the `runAsUser` flag only has an effect when the Bot is executed as a result of a subscription. Bots that are executed using the `$execute` endpoint seem to always use the caller's credentials, and any access policy set on the Bot has no effect. Is this intentional, or am I just configuring something incorrectly? It seems out of line with the AccessPolicy strategy described at https://www.medplum.com/docs/bots/consuming-webhooks. Thanks for your help!

natemerrill:
Currently, bot access policies have no effect when bot are run with the `$execute` endpoint. This PR brings the behavior of the Bot `$execute` endpoint in line with what is implied at https://www.medplum.com/docs/bots/consuming-webhooks#creating-access-policies. 
- The user's access policy is used to determine if the user has permission to execute the bot
- If `runAsUser` is enabled on the bot, the user's token and access policy is also used by the code inside the bot
- If `runAsUser` is not enabled on the bot, the bot's access policy (if any) is used by the bot.

I think this would be the desired behavior, but it could also result in bots that have no direct access policy configured having more access than they used to, since the user's access policy was previously used by default. Interested to hear your take. Thanks for all your work on Medplum, this project is fantastic!

### DIFF
--- a/packages/server/src/fhir/operations/execute.ts
+++ b/packages/server/src/fhir/operations/execute.ts
@@ -38,7 +38,7 @@ import { generateAccessToken } from '../../oauth/keys';
 import { recordHistogramValue } from '../../otel/otel';
 import { AuditEventOutcome, logAuditEvent } from '../../util/auditevent';
 import { MockConsole } from '../../util/console';
-import { createAuditEventEntities } from '../../workers/utils';
+import { createAuditEventEntities, findProjectMembership } from '../../workers/utils';
 import { sendOutcome } from '../outcomes';
 import { getSystemRepo } from '../repo';
 import { getBinaryStorage } from '../storage';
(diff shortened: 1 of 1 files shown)


codyebberson:
Thanks @natemerrill - This is a good callout and a good PR.

For some context:
* In the early days of bots, the only way to execute a bot was via subscription (i.e., in response to a resource change)
* The `runAsUser` flag was introduced so that the bot would run as the user who changed the original resource
* For example, if user Alice modified a `Patient`, and a FHIR subscription triggered a bot, and the bot subsequently made more changes, those changes would appear in history as `Alice`
* The bot `$execute` operation was added later, and has always implicitly been `runAsUser`

I agree with your interpretation of the values, and the behavior in your PR.

My only hesitation is that this could be a breaking change for existing bot users.  We need to do a quick review of how this would impact existing bot `$execute` usage.

Stay tuned, we'll perform that analysis and report back here.

Thanks again 🙏

codyebberson:
Update: We have analyzed all current usage of `$execute` and identified a handful of customer cases that require follow-up.  The fix is trivial, just set `runAsUser = true`, but we nonetheless want to follow proper protocols.

natemerrill:
Thanks for the context, the current behavior makes sense from an auditing perspective. Ideally, any changes the bot made when triggered by `$execute` would show as having been made by the user, but would run with the Bot's AccessPolicy.

I run into a lot of use cases like this where I'd like to allow patients to create certain resources through a bot, where I can control how and if the resource is created, without giving them permission to create/update/delete the resource directly through the Medplum API.

From what I can tell enabling this would be a much bigger change, though, and fixing `runAsUser` will do what I need to for now. 

Thanks so much for your help!

codyebberson:
@rahul1  - assigning to you for tracking while we complete customer follow-ups.

rahul1 (approved):

PR #3920, opened 2024-02-10
fix-3914 add `traceId` to `AuditEvent`s as extension

dillonstreator:
Addresses #3914

Add an extension (`https://medplum.com/fhir/StructureDefinition/tracing`) on `AuditEvent` that persists the associated `traceId` value.

This PR also Initialize a _new_ `traceId` and `requestId` for cron bot executions with bullmq opentelemetry instrumentation.

Questions:
- [x] [`getNodeAutoInstrumentations`](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node) covers `redis` and `ioredis` but not sure if this also covers `bullmq`.
  - `bullmq` is not covered by the auto instrumentation. There are community-driven libraries that provide compatible instrumentation such as https://github.com/jenniferplusplus/opentelemetry-instrumentation-bullmq

### DIFF
--- a/packages/server/src/fhir/operations/execute.ts
+++ b/packages/server/src/fhir/operations/execute.ts
@@ -574,12 +574,12 @@ async function createAuditEvent(
     entity: createAuditEventEntities(bot, request.input, request.subscription, request.agent, request.device),
     outcome,
     outcomeDesc,
-    extension: ctx.traceId ? [
+    extension: [
       {
-        url: "https://medplum.com/fhir/StructureDefinition/trace-id",
+        url: 'https://medplum.com/fhir/StructureDefinition/trace-id',
         valueString: ctx.traceId,
-      }
-    ] : undefined,
+      },
+    ],
   };
 
   const destination = bot.auditEventDestination ?? ['resource'];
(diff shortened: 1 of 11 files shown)


packages/server/src/fhir/operations/execute.ts
```diff
@@ -573,6 +574,15 @@ async function createAuditEvent(
     entity: createAuditEventEntities(bot, request.input, request.subscription, request.agent, request.device),
     outcome,
     outcomeDesc,
+    extension: [
+      {
+        url: 'https://medplum.com/fhir/StructureDefinition/tracing',
+        extension: [
+          { url: 'requestId', valueUuid: ctx.requestId },
+          { url: 'traceId', valueUuid: ctx.traceId },
```
dillonstreator, line 582: `valueUuid` is not always accurate for `traceId`. Thoughts on a key rename?

dillonstreator:
Given the work done on https://github.com/medplum/medplum/pull/4031, I will close this out. I do think that the `bullmq` tracing is a valuable addition so will split that out to separate PR.

Issue #3935, opened 2024-02-12
Add search result parameters to CapabilityStatement `searchParam`


larry-flexpa:
Medplum implements the following FHIR [search result parameters](https://github.com/medplum/medplum/blob/6ae909459b2439b353b2c4e4f7817ce17e84d274/packages/server/src/fhir/metadata.ts#L148):
- `_summary`
- `_elements`
- `_total`
- `_sort`
- `_include`
- `_revinclude`
- `_count`

However, these are not listed in the Capability Statement under `CapabilityStatement.rest.searchParam`. Would like to add these parameters to the Capability Statement to better expose these to end users. Thanks!

PR #3936, opened 2024-02-12
Add search result parameters to CapabilityStatement

larry-flexpa:
## Why
* Addresses #3935 
* In order to surface [search result parameters](https://www.hl7.org/fhir/search.html#modifyingresults) to end users, update `CapabilityStatement.rest.searchParam` with these parameters (what Medplum calls advanced search parameters)

## What
* Add the supported advanced search parameters to `CapabilityStatement.rest.searchParam` in `packages/server/src/fhir/metadata.ts` such that these parameters are returned.


codyebberson (approved):
lgtm, thanks @larry-flexpa !

PR #3978, opened 2024-02-16
feat(subscriptions): cleanup WS Subscription resources, add feature flag

ThatOneBro:
## TODO:
- [x] Check if backend WebSocket emits `close` when client WebSocket drops connection temporarily
- [x] Delete Subscription after `close`
- [x] Add test for checking that `Subscription` is cleaned up
- [x] Add feature flag for `WebSocket Subscriptions` on `Project`
- [x] Add logging for when `AccessPolicy` for a given `Subscription` is not satisfied but notification is made

### DIFF
--- a/packages/server/src/fhir/operations/getwsbindingtoken.ts
+++ b/packages/server/src/fhir/operations/getwsbindingtoken.ts
@@ -1,5 +1,5 @@
 import { allOk, badRequest, resolveId } from '@medplum/core';
-import { Parameters } from '@medplum/fhirtypes';
+import { Parameters, Subscription } from '@medplum/fhirtypes';
 import { Request, Response } from 'express';
 import { getConfig } from '../../config';
 import { getAuthenticatedContext } from '../../context';
@@ -12,6 +12,7 @@ const ONE_HOUR = 60 * 60 * 1000;
 
 export type AdditionalWsBindingClaims = {
   subscription_id: string;
+  project_id: string;
 };
 
 /**
@@ -40,8 +41,16 @@ export async function getWsBindingTokenHandler(req: Request, res: Response): Pro
   }
 
   const subscriptionId = req.params.id;
-  const subExists = await redis.exists(`Subscription/${subscriptionId}`);
-  if (!subExists) {
+  const cacheEntryStr = await redis.get(`Subscription/${subscriptionId}`);
+  if (!cacheEntryStr) {
+    await sendOutcome(res, badRequest('Content could not be parsed'));
+    return;
+  }
+
+  let cacheEntry: { resource: Subscription; projectId: string };
+  try {
+    cacheEntry = JSON.parse(cacheEntryStr);
+  } catch (err: unknown) {
     await sendOutcome(res, badRequest('Content could not be parsed'));
     return;
   }
@@ -57,6 +66,7 @@ export async function getWsBindingTokenHandler(req: Request, res: Response): Pro
     },
     {
       subscription_id: subscriptionId,
+      project_id: cacheEntry.projectId,
     } satisfies AdditionalWsBindingClaims
   );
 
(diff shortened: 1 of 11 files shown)


packages/server/src/fhir/operations/getwsbindingtoken.ts
```diff
@@ -40,8 +41,16 @@ export async function getWsBindingTokenHandler(req: Request, res: Response): Pro
   }
 
   const subscriptionId = req.params.id;
-  const subExists = await redis.exists(`Subscription/${subscriptionId}`);
-  if (!subExists) {
+  const cacheEntryStr = await redis.get(`Subscription/${subscriptionId}`);
```
codyebberson, line 44: Hm, we probably should not be reaching into Redis directly here.  We should go through `repo` to make sure that it's respecting user and project permissions.

packages/server/src/fhir/repo.ts
```diff
@@ -541,19 +540,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
-      const existingIdx = currentWsSubscriptions.findIndex(
-        (sub: Subscription) => (sub.id as string) === (result.id as string)
-      );
-      if (existingIdx !== -1) {
-        currentWsSubscriptions[existingIdx] = result;
-      } else {
-        currentWsSubscriptions.push(result);
-      }
-      await redis.set(`medplum:subscriptions:r4:project:${project}`, JSON.stringify(currentWsSubscriptions));
+      await redis.sadd(`medplum:subscriptions:r4:project:${project}:active`, `Subscription/${result.id}`);
```
codyebberson, line 543: 👍

packages/server/src/workers/subscription.ts
```diff
@@ -290,10 +290,25 @@ async function getSubscriptions(resource: Resource): Promise<Subscription[]> {
...
   });
-  const inMemorySubscriptionsStr = await getRedis().get(`medplum:subscriptions:r4:project:${project}`);
-  if (inMemorySubscriptionsStr) {
-    const inMemorySubscriptions = JSON.parse(inMemorySubscriptionsStr) as Subscription[];
-    subscriptions.push(...inMemorySubscriptions);
+  const redisOnlySubRefStrs = await getRedis().smembers(`medplum:subscriptions:r4:project:${project}:active`);
+  const redisOnlySubStrs = await getRedis().mget(redisOnlySubRefStrs);
+  if (redisOnlySubStrs.length) {
+    const arrLen = redisOnlySubStrs.length;
+    let subArrStr = '[';
```
codyebberson, line 297: Not sure, but I'd wager that using `.join(',')` would be faster than for loop here

```ts
const subArrStr = '[' + redisOnlySubStrs.filter(Boolean).join(',') + ']';
```

packages/server/src/fhir/operations/getwsbindingtoken.ts
codyebberson: i.e.,

image

This also protects against implementation detail changes (how resources are cached, whether caching is enabled, etc).

ThatOneBro: Yeah good call. 

I don't think a user should be able to get here without being able to access this subscription resource anyways but it's probably better for a number of reasons

packages/server/src/fhir/operations/getwsbindingtoken.ts
```diff
@@ -28,21 +27,21 @@ export type AdditionalWsBindingClaims = {
...
   }
 
   const subscriptionId = req.params.id;
-  const subExists = await redis.exists(`Subscription/${subscriptionId}`);
-  if (!subExists) {
-    await sendOutcome(res, badRequest('Content could not be parsed'));
+  try {
+    await repo.readResource<Subscription>('Subscription', subscriptionId);
+  } catch (_err: unknown) {
+    sendOutcome(res, badRequest('Subscription does not exist or does not have a project ID associated with it'));
```
codyebberson, line 44: Error message recommendation:

```ts
sendOutcome(res, badRequest(`Error reading subscription: ${normalizeErrorString(err)}`));
```

codyebberson (approved):
Minor feedback on error message, otherwise lgtm 👍

PR #4003, opened 2024-02-21
feat(subscriptions): cleanup WS `Subscription`s, log access policy satisfied

### LINKED ISSUE #3969: WebSocket subscription redis maintenance
1. Delete the in-Redis `Subscription` when the websocket connection disconnects
2. Let's use a Redis `SET` rather than stringified list: https://redis.io/docs/data-types/sets/

ThatOneBro:
Update of #3978 with a fix.

* Wraps everything inside of `checkAccessPolicy()`  in a `try... catch`
* Adds regression test for the bug it originally introduced

Basically the `buildAccessPolicy` step within `checkAccessPolicy()` could fail on malformed subscriptions with broken references, and it would cause the for loop on `subscriptions` to terminate early and not process later `subscriptions` in the list. Now that we catch anything within `checkAccessPolicy()` this is no longer an issue.

Closes #3969

### DIFF
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -293,18 +293,7 @@ async function getSubscriptions(resource: Resource): Promise<Subscription[]> {
   const redisOnlySubRefStrs = await getRedis().smembers(`medplum:subscriptions:r4:project:${project}:active`);
   const redisOnlySubStrs = await getRedis().mget(redisOnlySubRefStrs);
   if (redisOnlySubStrs.length) {
-    const arrLen = redisOnlySubStrs.length;
-    let subArrStr = '[';
-    for (let i = 0; i < arrLen - 1; i++) {
-      const result = redisOnlySubStrs[i];
-      if (result === null) {
-        continue;
-      }
-      subArrStr += result;
-      subArrStr += ',';
-    }
-    subArrStr += redisOnlySubStrs[arrLen - 1];
-    subArrStr += ']';
+    const subArrStr = '[' + redisOnlySubStrs.filter(Boolean).join(',') + ']';
     const inMemorySubs = JSON.parse(subArrStr) as { resource: Subscription; projectId: string }[];
     for (const { resource } of inMemorySubs) {
       subscriptions.push(resource);
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -3,6 +3,7 @@ import {
   createReference,
   getExtension,
   getExtensionValue,
+  getReferenceString,
   isGone,
   matchesSearchRequest,
   normalizeOperationOutcome,
@@ -12,7 +13,7 @@ import {
   serverError,
   stringify,
 } from '@medplum/core';
-import { Bot, ProjectMembership, Reference, Resource, Subscription } from '@medplum/fhirtypes';
+import { Bot, Project, ProjectMembership, Reference, Resource, Subscription } from '@medplum/fhirtypes';
 import { Job, Queue, QueueBaseOptions, Worker } from 'bullmq';
 import { createHmac } from 'crypto';
 import fetch, { HeadersInit } from 'node-fetch';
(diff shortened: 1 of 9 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -126,6 +130,53 @@ export function getSubscriptionQueue(): Queue<SubscriptionJobData> | undefined {
...
+ */
+async function checkAccessPolicy(resource: Resource, project: Project, subscription: Subscription): Promise<void> {
+  try {
+    // Check access policy
+    const subAuthor = subscription.meta?.author;
+    if (subAuthor) {
+      const membership = await findProjectMembership(project.id as string, subAuthor);
+      if (membership) {
+        const accessPolicy = await buildAccessPolicy(membership);
+        const satisfied = !!satisfiedAccessPolicy(resource, AccessPolicyInteraction.READ, accessPolicy);
```
codyebberson, line 153: Double exclamations not necessary

packages/server/src/workers/subscription.ts
```diff
@@ -126,6 +130,53 @@ export function getSubscriptionQueue(): Queue<SubscriptionJobData> | undefined {
...
+        if (!satisfied) {
+          globalLogger.warn(
+            `[Subscription Access Policy]: Access Policy not satisfied on '${getReferenceString(resource)}' for '${getReferenceString(project)}'`,
```
codyebberson, line 156: Preference to include `getReferenceString(subscription)` rather than `getReferenceString(project)` - subscription is more explicit.

packages/server/src/workers/subscription.ts
```diff
@@ -147,12 +198,30 @@ export async function addSubscriptionJobs(resource: Resource, context: Backgroun
...
+
+  const systemRepo = getSystemRepo();
+  let project: Project | undefined;
+  try {
+    const projectId = resource.meta?.project;
+    if (projectId) {
+      project = await systemRepo.readResource<Project>('Project', projectId);
+    }
+  } catch (_err: unknown) {
+    project = undefined;
```
codyebberson, line 210: I'm wondering if we should actually swallow this error, or if we should actually abort if project is gone?  Is there a valid case where `projectId` is an ID, but `readResource` fails? 🤔

packages/server/src/workers/subscription.ts
```diff
@@ -126,6 +130,53 @@ export function getSubscriptionQueue(): Queue<SubscriptionJobData> | undefined {
...
+    } else {
+      // Log it if there is no author for this Subscription (this is not good)
+      globalLogger.warn(
+        `[Subscription Access Policy]: No author for subscription '${getReferenceString(subscription)}'`
+      );
+    }
+  } catch (err: unknown) {
+    globalLogger.warn(
+      `[Subscription Access Policy]: Error occurred while checking access policy for resource '${getReferenceString(resource)}' against '${getReferenceString(subscription)}'`,
+      { error: err, subscription }
```
codyebberson, line 175: Can we log the subscription reference rather than the entire subscription?

packages/server/src/workers/subscription.ts
```diff
@@ -126,6 +130,53 @@ export function getSubscriptionQueue(): Queue<SubscriptionJobData> | undefined {
...
+        if (!satisfied) {
+          globalLogger.warn(
+            `[Subscription Access Policy]: Access Policy not satisfied on '${getReferenceString(resource)}' for '${getReferenceString(project)}'`,
+            { subscription, project: createReference(project), accessPolicy }
+          );
+        }
+      } else {
+        globalLogger.warn(
+          `[Subscription Access Policy]: No membership for subscription author '${getReferenceString(subAuthor)}' in project '${getReferenceString(project)}'`,
+          { subscription }
```
codyebberson, line 163: Can we log the subscription reference rather than the entire subscription?

packages/server/src/workers/subscription.ts
ThatOneBro: Alright, we're logging that now 👍 

Should probably never hit that log... probably...

codyebberson (approved):

PR #4010, opened 2024-02-22
Cleanup core search parse utils

codyebberson:
`@medplum/core` includes 4 utility methods which all do basically the same thing.

`packages/core/src/search/search.ts`

1. `parseSearchRequest(resourceType, query)`
2. `parseSearchUrl(url: URL)`
3. `parseSearchDefinition(url: string)`
4. `parseCriteriaAsSearchRequest(criteria: string)`

This PR attempts to clean that up.

1. Make `parseSearchRequest(url, query?)` the singular preferred utility method
    a. The `url` param can be `T['resourceType']`, `URL`, or `string`
    b. Fully support backwards compatibility
    c. Handle all of the permutations with minimal excess parsing
2. Deprecate all of the other utility methods
3. Replace all usage of the other utility methods throughout the monorepo

### DIFF
--- a/examples/medplum-chart-demo/src/pages/SearchPage.tsx
+++ b/examples/medplum-chart-demo/src/pages/SearchPage.tsx
@@ -3,7 +3,7 @@ import {
   DEFAULT_SEARCH_COUNT,
   Filter,
   formatSearchQuery,
-  parseSearchDefinition,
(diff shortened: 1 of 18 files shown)


mattwiller (approved):
✨✨✨

PR #4011, opened 2024-02-22
Cross-Project reference checks

### LINKED ISSUE #3927: Project linking with `checkReferencesOnWrite`
`checkReferencesOnWrite` is an optional project setting.  When enabled, on every resource write (create or update), the server validates all `Reference` values.

The current implementation does not allow references to resources in linked projects.

See `packages/server/src/fhir/references.ts`

The fix should be straightforward.  We will need to pass in the array of valid project IDs rather than requiring exact match.

mattwiller:
Resolves #3927


packages/server/src/fhir/references.ts
```diff
@@ -79,11 +81,21 @@ export class FhirReferenceValidator<T extends Resource> {
...
     }
+    if (!this.root.meta?.project) {
+      return;
+    }
 
     try {
       const systemRepo = getSystemRepo();
-      const existing = await systemRepo.readReference(reference);
-      if (existing.meta?.project && this.root.meta?.project && existing.meta.project !== this.root.meta.project) {
+      const target = await systemRepo.readReference(reference);
```
codyebberson, line 90: `validateReferences` is only ever called from `repo.ts`, where I believe we always have the `allowedProjects` array loaded?

Could we make `allowedProjects` an extra param to the `validateReferences` method?

packages/server/src/fhir/references.ts
```diff
@@ -79,11 +84,18 @@ export class FhirReferenceValidator<T extends Resource> {
...
```
codyebberson, line 93: (non-blocking)

Out of scope for this PR, but I'm realizing the inefficiencies of the original implementation of this file.

Currently it's doing the naive thing, crawling the resource, and synchronously checking the references.

The server `Repository` class already has a `readReferences(references: Reference[])` method.  We could easily modify this file to first gather all the references, and then do it as a batch operation.

The implementation of `readReferences` uses a batch Redis operation.

The implementation of `readReferences`  could itself be optimized to read multiple references in a single SQL query.  That is more complicated though.

codyebberson (approved):
Nice 👍

packages/server/src/fhir/references.ts
mattwiller: That sounds good to me: I can knock the efficiency boost out in a fast follow PR

PR #4016, opened 2024-02-22
Update peer dep major versions

rahul1:


codyebberson (changes requested):
Actually, you need to run `npm i` after making this change to update `package-lock.json`

codyebberson (approved):

PR #4018, opened 2024-02-22
Log batch processing statistics

### LINKED ISSUE #3861: Log FHIR bundles stats
Before we can work on https://github.com/medplum/medplum/issues/1369 (better transaction support)

There are 2 transaction bugs/features that need to be fixed/deprecated

1. Using the `urn:uuid:` syntax with `Bundle.type="batch"` - that should only be supported in `type=transaction`
2. Ignoring certain errors with `type=transaction` - all errors should invalidate the bundle and rollback

mattwiller:
Resolves #3861

### DIFF
--- a/packages/fhir-router/src/batch.ts
+++ b/packages/fhir-router/src/batch.ts
@@ -8,6 +8,7 @@ import {
   OperationOutcomeError,
   parseSearchUrl,
   resolveId,
+  Event,
 } from '@medplum/core';
 import { Bundle, BundleEntry, BundleEntryRequest, OperationOutcome, Resource } from '@medplum/fhirtypes';
 import { FhirRouter } from './fhirrouter';
@@ -87,12 +88,14 @@ class BatchProcessor {
       }
     }
 
-    this.router.emit('batch', {
+    const event: BatchEvent = {
+      type: 'batch',
       bundleType,
       count,
       errors,
       size: JSON.stringify(this.bundle).length,
-    });
+    };
+    this.router.dispatchEvent(event);
 
     return {
       resourceType: 'Bundle',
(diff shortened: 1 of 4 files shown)


packages/fhir-router/src/batch.ts
```diff
@@ -68,24 +68,35 @@ class BatchProcessor {
     }
 
     const resultEntries: BundleEntry[] = [];
+    let count = 0;
+    let errors = 0;
     for (const entry of entries) {
       const rewritten = this.rewriteIdsInObject(entry);
-      // If the resource 'id' element is specified, we want to replace teh `urn:uuid:*` string and
+      // If the resource 'id' element is specified, we want to replace the `urn:uuid:*` string and
```
codyebberson, line 75: 👍

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -5,6 +5,7 @@ import { processBatch } from './batch';
 import { graphqlHandler } from './graphql';
 import { FhirRepository } from './repo';
 import { HttpMethod, Router } from './urlrouter';
+import EventEmitter from 'events';
```
codyebberson, line 8: I think this is a Node.js-only import?  `fhir-router` is used in Storybook, which will probably choke on this.

The good news is that you should be able to use `EventTarget` from `@medplum/core`

Then replace `emit` with `dispatchEvent` and `on` with `addEventListener`

packages/server/src/fhir/routes.ts
```diff
@@ -209,6 +210,16 @@ protectedRoutes.use(
     const ctx = getAuthenticatedContext();
     if (!internalFhirRouter) {
       internalFhirRouter = new FhirRouter({ introspectionEnabled: getConfig().introspectionEnabled });
+      internalFhirRouter.on('warn', (msg, data) => ctx.logger.warn(msg, data));
+      internalFhirRouter.on('batch', ({ count, errors, size, bundleType }) => {
+        recordHistogramValue('medplum.batch.entries', count, { bundleType });
```
codyebberson, line 215: 👍

codyebberson:
Vercel config:

https://vercel.com/medplum/medplum-storybook/settings/git

image

Before:

```
git diff --quiet HEAD^ HEAD ./packages/core/ ./packages/react/
```

After:

```
git diff --quiet HEAD^ HEAD ./packages/core/ ./packages/fhir-router/ ./packages/react/ ./packages/react-hooks/
```

codyebberson (approved):

PR #4028, opened 2024-02-26
Fix error message for Google auth, user not found

codyebberson:
Somewhere in recent refactoring we lost error messages for Google auth + "User not found".

Simple fix: catch error, show `<OperationOutcomeAlert issues={issues} />` if there are any issues.

### DIFF
--- a/packages/react/src/auth/SignInForm.test.tsx
+++ b/packages/react/src/auth/SignInForm.test.tsx
@@ -1,10 +1,10 @@
 import { Title } from '@mantine/core';
 import { allOk, badRequest, GoogleCredentialResponse, MedplumClient } from '@medplum/core';
-import { act, fireEvent, render, screen, waitFor } from '../test-utils/render';
+import { MedplumProvider } from '@medplum/react-hooks';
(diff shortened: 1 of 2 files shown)


packages/react/src/auth/SignInForm.test.tsx
```diff
@@ -649,6 +655,59 @@ describe('SignInForm', () => {
...
+    const onSuccess = jest.fn();
+
+    await act(async () => {
+      await setup({
+        onSuccess,
+        googleClientId: clientId,
+      });
+    });
+
+    await waitFor(() => expect(screen.getByText('Sign in with Google')).toBeInTheDocument());
```
ThatOneBro, line 697:
```suggestion
    await expect(screen.findByText('Sign in with Google')).resolves.toBeInTheDocument();
```
Maybe worth mentioning that this is the same thing, just in case you're not aware of the `screen.findBy{...}` APIs. They are the async versions of the `getBy{...}` APIs

ThatOneBro (approved):
LGTM

PR #4036, opened 2024-02-26
Add note for common postgres error

ksmith94:
This adds a note to tell how to fix a common issue when following the run the stack guide in the contributing docs

### DIFF
--- a/packages/docs/docs/contributing/run-the-stack.md
+++ b/packages/docs/docs/contributing/run-the-stack.md
@@ -146,7 +146,7 @@ If everything is working, you should see the following in your browser:
 
 :::warning Troubleshooting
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/contributing/run-the-stack.md
```diff
@@ -144,6 +144,12 @@ If everything is working, you should see the following in your browser:
 { "ok": true, "version": "<MEDPLUM_VERSION>", "postgres": true, "redis": true }
 ```
 
+:::warning Troubleshooting
+
+If you get an error such as `error: role "admin@example.com" does not exist` or `password authentication failed for user "medplum"` this usually means you have another instance of Postgres running. Closing the other instance should resolve this issue.
```
codyebberson, line 149: I think this should be `role "medplum" does not exist`

"admin@example.com" is in our application layer.  The database username is typically "medplum".

rahul1 (approved):

PR #4038, opened 2024-02-26
Add default DB query timeout

### LINKED ISSUE #3859: Add default query timeouts

mattwiller:
Resolves #3859

### DIFF
--- a/packages/server/src/config.ts
+++ b/packages/server/src/config.ts
@@ -82,6 +82,7 @@ export interface MedplumDatabaseConfig {
   username?: string;
   password?: string;
   ssl?: MedplumDatabaseSslConfig;
+  idleTimeout?: number;
 }
 
 export interface MedplumRedisConfig {
--- a/packages/server/src/config.ts
+++ b/packages/server/src/config.ts
@@ -82,7 +82,7 @@ export interface MedplumDatabaseConfig {
   username?: string;
   password?: string;
   ssl?: MedplumDatabaseSslConfig;
-  idleTimeout?: number;
+  queryTimeout?: number;
 }
 
 export interface MedplumRedisConfig {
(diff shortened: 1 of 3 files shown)


packages/server/src/database.ts
```diff
@@ -26,6 +26,7 @@ export async function initDatabase(serverConfig: MedplumServerConfig, runMigrati
     user: config.username,
     password: config.password,
     ssl: config.ssl,
+    idleTimeoutMillis: config.idleTimeout ?? 60_000,
```
codyebberson, line 29: I think this does something different: https://node-postgres.com/apis/pool

```
  // number of milliseconds a client must sit idle in the pool and not be checked out
  // before it is disconnected from the backend and discarded
  // default is 10000 (10 seconds) - set to 0 to disable auto-disconnection of idle clients
  idleTimeoutMillis?: number
```

My understanding of the original issue is that we want something more like `statement_timeout`: https://www.postgresql.org/docs/9.6/runtime-config-client.html

> `statement_timeout (integer)`
> 
> Abort any statement that takes more than the specified number of milliseconds, starting from the time the command arrives at the server from the client. If log_min_error_statement is set to ERROR or lower, the statement that timed out will also be logged. A value of zero (the default) turns this off.
> 
> Setting statement_timeout in postgresql.conf is not recommended because it would affect all sessions.

packages/server/src/config.ts
```diff
@@ -82,6 +82,7 @@ export interface MedplumDatabaseConfig {
   username?: string;
   password?: string;
   ssl?: MedplumDatabaseSslConfig;
+  queryTimeout?: number;
```
codyebberson, line 85: Can we document this in `packages/docs/docs/self-hosting/config-settings.md` ?

codyebberson (approved):
Code looks good

See comment on docs

image

PR #4039, opened 2024-02-26
Resource table profile support

mattlong:
`ResourceTable` now displays array slices and extensions (when the supporting profile is available).

Each slice of an array appears as its own separate row in the table. Achieving this somewhat complicated the props of `ResourcePropertyDisplay` and some other components, but it was all done in what I believe was a fully backwards compatible way.

Slice and extension support reused a lot of code from their `ResourceForm` equivalents which was nice to see. I didn't bother to DRY up all the duplicated code since it was all fairly straightforward blocks of code that I didn't think merit abstracting away just yet.

I also tweaked the general table display to not put a border above the first row and have more symmetric padding at the top/bottom of the table. However, in reviewing the storybook diffs, I'm not so sure removing the border above the first row is desirable across the board. e.g. in the timeline components, the top border might actually be beneficial. On `DetailPage`, where `ResourceTable` is displayed standalone, it looks cleaner without that top border. Open to changing it back or adding an optional prop on `ResourceTable` to control inclusion of the top border.

Also, notice the new UI in the top right of the screenshot below for selecting the profile to display on the details page. The select input is only displayed when profiles exist in `resource.meta.profile`.

Screenshot 2024-02-27 at 4 14 27 PM

### DIFF
--- a/packages/react/src/BackboneElementDisplay/BackboneElementDisplay.tsx
+++ b/packages/react/src/BackboneElementDisplay/BackboneElementDisplay.tsx
@@ -14,8 +14,8 @@ import { useContext, useMemo } from 'react';
 import { ElementsContext } from '../ElementsInput/ElementsInput.utils';
 import { maybeWrapWithContext } from '../utils/maybeWrapWithContext';
 
-const EXTENSION_KEYS = new Set(['extension', 'modifierExtension']);
-const IGNORED_PROPERTIES = new Set(DEFAULT_IGNORED_PROPERTIES.filter((prop) => !EXTENSION_KEYS.has(prop)));
+const EXTENSION_KEYS = ['extension', 'modifierExtension'];
+const IGNORED_PROPERTIES = DEFAULT_IGNORED_PROPERTIES.filter((prop) => !EXTENSION_KEYS.includes(prop));
 
 export interface BackboneElementDisplayProps {
   readonly value: TypedValue;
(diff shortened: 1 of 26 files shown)


packages/react/src/BackboneElementDisplay/BackboneElementDisplay.tsx
```diff
@@ -1,25 +1,56 @@
...
 import { DEFAULT_IGNORED_NON_NESTED_PROPERTIES, DEFAULT_IGNORED_PROPERTIES } from '../constants';
 import { DescriptionList, DescriptionListEntry } from '../DescriptionList/DescriptionList';
 import { ResourcePropertyDisplay } from '../ResourcePropertyDisplay/ResourcePropertyDisplay';
 import { getValueAndType } from '../ResourcePropertyDisplay/ResourcePropertyDisplay.utils';
+import { useContext, useMemo } from 'react';
+import { ElementsContext } from '../ElementsInput/ElementsInput.utils';
+import { maybeWrapWithContext } from '../utils/maybeWrapWithContext';
+
+const EXTENSION_KEYS = new Set(['extension', 'modifierExtension']);
+const IGNORED_PROPERTIES = new Set(DEFAULT_IGNORED_PROPERTIES.filter((prop) => !EXTENSION_KEYS.has(prop)));
```
codyebberson, line 18: Suuuper minor, but we microbenchmarked this elsewhere and found that `array.includes()` was faster than `set.has()` for sets under 20 elements.

codyebberson (commented):
Code all looks great, no concerns there.

UI for displaying the profile is great 👌 

UI for profile selector is definitely good enough for a v1.

(idle thoughts for future PR)

We could probably make it a little more subtle.  Something like the version selector in Mantine docs:

image

Or maybe the "Basic Button" example from the Mantine Combobox demos page: https://mantine.dev/combobox/?e=BasicButton

image

Definitely non-blocking.

Looks like we could use some basic tests for `ExtensionDisplay.tsx` and `SliceDisplay.tsx`, then good to go 👍

mattlong:
Fixes #3504

codyebberson (approved):

PR #4041, opened 2024-02-27
Added initial key value API

codyebberson:
See https://github.com/medplum/medplum/discussions/3995

TODO:
- [x] Cap users at 10 key/value pairs
- [x] Add utility methods to `MedplumClient`

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -79,6 +79,7 @@ import {
   resolveId,
   sleep,
 } from './utils';
+import { MedplumKeyValueClient } from './keyvalue';
 
 export const MEDPLUM_VERSION = import.meta.env.MEDPLUM_VERSION ?? '';
 export const DEFAULT_ACCEPT = ContentType.FHIR_JSON + ', */*; q=0.1';
@@ -689,6 +690,7 @@ export class MedplumClient extends EventTarget {
   private basicAuth?: string;
   private initPromise: Promise<void>;
   private initComplete = true;
+  private keyValueClient?: MedplumKeyValueClient;
 
   constructor(options?: MedplumClientOptions) {
     super();
@@ -2710,6 +2712,17 @@ export class MedplumClient extends EventTarget {
     headers['Prefer'] = 'respond-async';
 
     return this.request('POST', url, options);
+  }
+
+  /**
+   * Returns the key value client.
+   * @returns The key value client.
+   */
+  get keyValue(): MedplumKeyValueClient {
+    if (!this.keyValueClient) {
+      this.keyValueClient = new MedplumKeyValueClient(this);
+    }
+    return this.keyValueClient;
   }
 
   //
--- a/packages/server/src/keyvalue/store.ts
+++ b/packages/server/src/keyvalue/store.ts
@@ -73,8 +73,9 @@ function validateKey(key: string): string {
 }
 
 /**
- * Returns the Redis cache key for the given project and key.
- * Validates the key string is an alphanumeric string from 1-255 characters.
+ * Returns the Redis cache key for the user's project membership.
+ * The Redis cache key points to a Redis "hash" type: https://redis.io/docs/data-types/hashes/
+ * All of the user's key/value pairs are entries in that hash object.
  * @param ctx - The request context.
  * @returns The Redis cache key.
  */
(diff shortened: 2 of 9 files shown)


packages/core/src/client.ts
```diff
@@ -2712,6 +2714,17 @@ export class MedplumClient extends EventTarget {
     return this.request('POST', url, options);
   }
 
+  /**
+   * Returns the key value client.
+   * @returns The key value client.
+   */
+  get keyValue(): MedplumKeyValueClient {
```
ThatOneBro, line 2721: Love to see it. Just the right amount of magic here.

I think we can also lazy load the `./keyvalue` module here but not blocking this PR.

We should definitely do this to the `FHIRcast` client and `SubscriptionManager` too I think. If we actually bundle them as (a) separate file(s) potentially could reduce bundle size and initial load times a lot for those who are not actually using any of those fancy features.

packages/server/src/keyvalue/store.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+  if (value.length > MAX_VALUE_LENGTH) {
+    throw new OperationOutcomeError(badRequest('Value too long'));
+  }
+  const ctx = getAuthenticatedContext();
+  const cacheKey = getCacheKey(ctx);
+  const redis = getRedis();
+  const exists = await redis.hexists(cacheKey, key);
+  if (!exists) {
+    const length = await redis.hlen(cacheKey);
```
ThatOneBro, line 42: I see the switch to using raw Redis in tests may have paid off here

packages/server/src/keyvalue/store.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Returns the Redis cache key for the given project and key.
+ * Validates the key string is an alphanumeric string from 1-255 characters.
+ * @param ctx - The request context.
+ * @returns The Redis cache key.
+ */
+function getCacheKey(ctx: AuthenticatedRequestContext): string {
+  return `medplum:keyvalue:${ctx.membership.id}`;
```
ThatOneBro, line 82: Is this per project or per project per user here? Should it not be `ctx.project.id`? Or are these KVs per user? If they are per user, is there any way to share keys for a whole project with this API?

ThatOneBro (commented):
Looks good for the most part, one major clarification to make sure I understand what the scoping is for this system

packages/server/src/keyvalue/store.ts
```diff
@@ -0,0 +1,83 @@
...
+    throw new OperationOutcomeError(badRequest('Invalid key'));
+  }
+  if (!/^[a-zA-Z0-9_-]+$/.test(key)) {
+    throw new OperationOutcomeError(badRequest('Invalid key'));
+  }
+  return key;
+}
+
+/**
+ * Returns the Redis cache key for the given project and key.
```
ThatOneBro, line 76: Mostly I think I was confused by the description here: `for the given project and key`

packages/server/src/keyvalue/store.ts
```diff
@@ -0,0 +1,83 @@
...
+  const cacheKey = getCacheKey(ctx);
+  const redis = getRedis();
+  const exists = await redis.hexists(cacheKey, key);
+  if (!exists) {
+    const length = await redis.hlen(cacheKey);
+    if (length >= MAX_ITEMS) {
+      throw new OperationOutcomeError(badRequest('Max items exceeded'));
+    }
+  }
+  await redis.hset(cacheKey, key, value);
```
mattlong, line 47: There's a small race condition here where someone could spam a lot of set requests at the same time and have more than `MAX_ITEMS` wind up being set. We could avoid it by obtaining/releasing a lock around the operation, but I'm not sure if it's worth the complexity.

packages/server/src/keyvalue/store.ts
codyebberson: It is currently per user, by design.  My thinking was that the primary use case for this is temp storage for bots, not for any data sharing.  I feel like the moment we open it up to sharing, it really spirals in complexity (access policies, hidden values, readonly values, auditing, logging, etc).  Purely per user avoids all of that complexity.

packages/server/src/keyvalue/store.ts
codyebberson: Ah, good catch, that comment is outdated.  I will update.

packages/server/src/keyvalue/store.ts
codyebberson: Acknowledged.  If someone gets 11 (or even 20) keys, that's not going to kill us.

ThatOneBro (approved):

mattlong (approved):

PR #4046, opened 2024-02-27
fix: Handle Identifier case for isReference

jdjkelly:
## Why

Sometimes a `reference` is not actually a `Reference`, sometimes it is an `Identifier`.

Screenshot 2024-02-27 at 10 49 08 AM

## What
* Check the type of `reference` and validate that it's a string - this disambiguates it from an `Identifier`. 
* Adds a test


codyebberson:
Merged here: https://github.com/medplum/medplum/pull/4046

PR #4049, opened 2024-02-27
Add strict handling for query string Operation parameters

mattwiller:
Ensures that query string parameters are parsed as the correct type, and adds an explicit check for complex parameter types

### DIFF
--- a/packages/server/src/fhir/operations/utils/parameters.ts
+++ b/packages/server/src/fhir/operations/utils/parameters.ts
@@ -70,29 +70,39 @@ function parseQueryString(
     if (!value) {
       continue;
     }
-    if (param.type?.match(/^[A-Z]/)) {
+    if (param.part || param.type?.match(/^[A-Z]/)) {
       // Query parameters cannot contain complex types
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/operations/utils/parameters.ts
```diff
@@ -60,6 +60,48 @@ export function parseInputParameters<T>(operation: OperationDefinition, req: Req
...
+      case 'integer':
+      case 'positiveInt':
+      case 'unsignedInt':
+        parsed[param.name] = parseInt(value, 10);
+        break;
+      case 'decimal':
+        parsed[param.name] = parseFloat(value);
+        break;
+      case 'boolean':
+        if (value === 'true') {
```
codyebberson, line 90: Super minor, but I suspect this pattern is repeated enough that we could benefit from a `parseBoolean` method in `@medplum/core`.

codyebberson (approved):
overall lgtm 👍 

May want to add some tests, especially for the various error cases.

PR #4051, opened 2024-02-27
Fixes #4043 - responsive auth form

### LINKED ISSUE #4043: Login scopes and responsive issues on mobile.
Check the following Screenrecord, tested with iPhone 12 mini, iOS 17.3.1:

https://github.com/medplum/medplum/assets/28322469/30cffc18-8af9-48b8-a43b-b2b1933f401d

## What's happening
When login on Medplum App from a mobile:
1. The login form doesn't fit well in the screen, the user has to scroll horizontally to get to the **Next** and **Sign in** buttons.
2. The login shows a last step screen after user password, asking to **Choose scope** and a list of pre-checked checkboxes.

## What I expect to happen
1. The login form should fit the mobile screen so the user doesn't have to scroll horizontally to get to the buttons.
2. Scopes should not be asked to confirm if they are already set on Authentication request, like this request:
```dart
GET https://api.medplum.com/oauth2/authorize?
    response_type=code&
    client_id=XXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXX&
    redirect_uri=https://my.app.com/redirect&
    state=XXXXXXXXXXXX&
    scope=offline_access openid email phone profile
```

codyebberson:


packages/react/src/Panel/Panel.tsx
```diff
@@ -2,11 +2,6 @@ import { Paper, PaperProps } from '@mantine/core';
 import cx from 'clsx';
 import classes from './Panel.module.css';
 
-export interface PanelStylesParams {
```
codyebberson, line 5: This appears to be carryover from Mantine 6.  It is not used anywhere else in the codebase.

ThatOneBro (approved):

PR #4063, opened 2024-02-28
refactor(ci): parallelize `eslint` and running tests in CI

ThatOneBro:


scripts/build-docs.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Diagnostics
+node --version
+npm --version
+
+# Install
+[ ! -d "node_modules" ] && npm ci --maxsockets 1
+
+# Build
+npm run build:all
```
codyebberson, line 22: We could be more surgical here if you wanted to

`npm run build:all -- --filter=@medplum/docs`

or we could add that as a `script` in the root `package.json`

.github/workflows/build.yml
```diff
@@ -98,7 +216,57 @@ jobs:
...
           path: coverage/lcov.info
+
+  build-docs:
+    name: Build the docs
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    strategy:
+      matrix:
+        node-version: [18, 20]
```
codyebberson, line 227: I'm ok to drop Node 18 support for docs.  Matrix strategy not required here.

codyebberson (approved):
Nice, this should be a big win.  See optional comments

PR #4064, opened 2024-02-28
Fixes #2021 - support _type search parameter

### LINKED ISSUE #2021: Support `_type` search parameter
The FHIR search spec allows for searching multiple resource types in a single query using the `_type` parameter. We have users that would like to support this in order to do unified sort / pagination across resource types. An example use case is building a "feed" of all resources related to a patient. 

https://build.fhir.org/search.html#type

> If no type is specified, the only search parameters that can be used in global search like this are the base parameters that apply to [all resources](https://build.fhir.org/search.html#all).
>
> If a single type is specified, all search parameters for that type are allowed.
>
> If multiple types are specified, all search parameters that can be referenced commonly between resources can be used (e.g., listed with the same name in the CapabilityStatement, typically the value of [SearchParameter.code](https://build.fhir.org/searchparameter-definitions.html#SearchParameter.code)) TU. For example, both [Patient](https://build.fhir.org/patient.html#search) and [Person](https://build.fhir.org/person.html#search) define a search parameter referenced by name. Assuming a server implements both parameters and uses name to do so, the name parameter can be used to search across both Patient and Person resources. Note that this is allowed even though the search parameters do not share a common definition. If a request references multiple parameters that are not the same type (e.g., if one parameter is a number type and another is a token type, despite sharing the same name), this is an error and servers SHOULD return a 400 status TU.

codyebberson:
Quick and dirty implementation of `_type`

I anticipate the most controversial choice here was to add `types` to `SearchRequest` rather than changing `resourceType` to an array.  We could totally do that, but it would probably 10x the size of this diff.

As implemented, this treats `SearchRequest.types` as an optional override.  If present, then `SearchRequest.resourceType` is basically ignored.

Everything else proved to be pretty natural and clean.

### DIFF
--- a/packages/core/src/search/search.ts
+++ b/packages/core/src/search/search.ts
@@ -20,6 +20,7 @@ export interface SearchRequest<T extends Resource = Resource> {
   summary?: 'true' | 'text' | 'data';
   format?: string;
   pretty?: boolean;
+  types?: T['resourceType'][];
 }
 
 export interface Filter {
@@ -303,6 +304,10 @@ function parseKeyValue(searchRequest: SearchRequest, key: string, value: string)
     case '_fields':
     case '_elements':
       searchRequest.fields = value.split(',');
+      break;
+
+    case '_type':
+      searchRequest.types = value.split(',') as Resource['resourceType'][];
       break;
 
     case '_format':
--- a/packages/core/src/search/parse.test.ts
+++ b/packages/core/src/search/parse.test.ts
@@ -674,4 +674,11 @@ describe('Search parser', () => {
       pretty: false,
     });
   });
+
+  test('_type', () => {
+    expect(parseSearchRequest('Patient', { _type: 'Patient,Observation' })).toMatchObject<SearchRequest>({
+      resourceType: 'Patient',
+      types: ['Patient', 'Observation'],
+    });
+  });
 });
--- a/packages/fhir-router/src/fhirrouter.test.ts
+++ b/packages/fhir-router/src/fhirrouter.test.ts
@@ -188,4 +188,21 @@ describe('FHIR Router', () => {
     expect(res).toMatchObject(allOk);
     expect(bundle).toBeDefined();
   });
+
+  test('Search multiple types', async () => {
+    const [res, bundle] = await router.handleRequest(
+      {
+        method: 'GET',
+        pathname: '/',
+        body: {},
+        params: {},
+        query: {
+          _type: 'Patient,Observation',
+        },
+      },
+      repo
+    );
+    expect(res).toMatchObject(allOk);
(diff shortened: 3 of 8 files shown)


ThatOneBro (commented):
This looks good so far but I want @mattwiller to take a look before approving

packages/core/src/search/search.ts
```diff
@@ -20,6 +20,7 @@ export interface SearchRequest<T extends Resource = Resource> {
   summary?: 'true' | 'text' | 'data';
   format?: string;
   pretty?: boolean;
+  types?: T['resourceType'][];
```
mattwiller, line 23: This all looks good to me, but should we try to move towards this as the source of truth property and deprecate `resourceType`?

mattwiller (approved):

packages/core/src/search/search.ts
codyebberson: > should we try to move towards this as the source of truth property and deprecate resourceType?

That's exactly the right question.

I'm open to feedback, but my thinking is "yes, but not right now"

Considerations:
* Overall API design
    * Currently a bunch of the search methods use TypeScript magic to map `resourceType` to the return type.
    * Both in `MedplumClient` and `Repository`, which are both used all over the place.
    * There is probably a way to do that with an array field, but it might be pretty gross.  
    * I'd like to think about it in the context of `_include` and `_revinclude` which also breaks the current abstraction
    * I'd like to think about it in the context of R5, which would be a different return type for the same `resourceType`
* Migration cost
    * That would be a breaking API change at a pretty critical depth.  It would probably necessitate a major version bump.  
    * Furthermore, internally we have a ton of code using the simple case of singular `resourceType`.  It would be a straightforward migration, but it would be a lot of wood to chop.

For better or worse, for right now, the common case is 1:1 single resource type input, single resource type output.  I think it makes sense to optimize for the common case.

I think the future state is probably something like this:

```ts
interface BaseSearchRequest<T extends BaseResource> { ... }

interface SingleTypeSearchRequest<T extends BaseResource> extends BaseSearchRequest<T> { ... };

interface MultiTypeSearchRequest<T extends BaseResource> extends BaseSearchRequest<T> { ... };

type SearchRequest<T extends BaseResource> = SingleTypeSearchRequest<T> | MultiTypeSearchRequest<T>;

type R4SearchRequest = SearchRequest<R4BaseResource>;

type R5SearchRequest = SearchRequest<R5BaseResource>;

// etc
```

So, back to your original question:  Yes, I think we should do that eventually, but I fear that it is a pretty big can of worms, and we should be ready for an onslaught of work to implement it soup to nuts.  By adding a single optional override member, we unblock the `_type` feature in a pretty surgical way.

tl;dr - I think this is acceptable tech debt.

PR #4065, opened 2024-02-28
feat(core/client): expose a wrapped `fetch`-adhering method from `MedplumClient`

dillonstreator:
Expose `fetch` primitive that wraps token refresh from the `MedplumClient`. This will enable an adapter pattern to provide different graphql clients their fetch method. Namely, Apollo via `createHttpLink#fetch`.


packages/core/src/client.ts
```diff
@@ -2866,7 +2879,7 @@ export class MedplumClient extends EventTarget {
     ) {
       const contentLocation = await tryGetContentLocation(response, obj);
       if (contentLocation) {
-        return this.request('GET', contentLocation, { ...options, body: undefined });
+        return this.request(contentLocation, { ...options, body: undefined });
```
dillonstreator, line 2882: Need to set `method` to 'GET' in options?

dillonstreator: https://github.com/medplum/medplum/pull/4065/commits/ae3e0d9e775a0584b29b1abdfcbd6ec046845c96

codyebberson:
Continuing here: https://github.com/medplum/medplum/pull/9357

Thanks @dillonstreator

PR #4066, opened 2024-02-28
chore(otel): add `bullmq` open telemetry instrumentation

dillonstreator:
Followup of https://github.com/medplum/medplum/pull/3920

`@opentelemetry/auto-instrumentations-node` does not include `bullmq`. As it is a core component of the system, it should be instrumented.
https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node#supported-instrumentations

https://github.com/jenniferplusplus/opentelemetry-instrumentation-bullmq looks like a good option and supports the bullmq v5.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10290,6 +10290,79 @@
         "node": ">=8"
       }
     },
+    "node_modules/@jenniferplusplus/opentelemetry-instrumentation-bullmq": {
+      "version": "0.5.0",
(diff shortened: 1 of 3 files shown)


ThatOneBro:
Update: it seems like if we get back to enabling traces again we can use this package which is derived from jenniferplusplus's package, with 32,000 weekly downloads now:
https://www.npmjs.com/package/@appsignal/opentelemetry-instrumentation-bullmq

The-Alchemist:
BullMQ instruction was added as part of 877ab4a272db0b5b8d88583d2edb7863736f5fa6

PR #4079, opened 2024-03-02
Escape equals sign in CSV export

codyebberson:


ThatOneBro (approved):
TIL: https://owasp.org/www-community/attacks/CSV_Injection

codyebberson:
> TIL: https://owasp.org/www-community/attacks/CSV_Injection

Hm, we should probably include all of those special characters, not just `=`

> Equals to (=)
> Plus (+)
> Minus (-)
> At (@)
> Tab (0x09)
> Carriage return (0x0D)

PR #4082, opened 2024-03-04
Dependency upgrades

medplumbot:
Dependency upgrades


codyebberson:
Build failing, I believe due to this: https://github.com/facebook/docusaurus/discussions/9456#discussioncomment-7439620

In particular, https://github.com/facebook/docusaurus/discussions/9456#discussioncomment-7497789

> [yhuard](https://github.com/yhuard)
> [on Nov 7, 2023](https://github.com/facebook/docusaurus/discussions/9456#discussioncomment-7497789)
> Thank you @ashleykolodziej-moodys for sharing your issue, it helped me solve mine. 👍
> I was getting the following error:
> 
> Expected component `Details` to be defined: you likely forgot to import, pass, or provide it.
> I checked my lockfile and found that @mdx-js/react version 2.3.0 was being imported by @storybook/addon-docs, and it created conflicts with Docusaurus v3.

`@mdx-js/react` v3 is installed in `packages/docs/node_modules`, but Webpack is in the root `node_modules`, and it appears that is causing problems.  I'm going to research if we can get npm to put Webpack in `packages/docs/node_modules` (because I believe Docusaurus is the only project that uses it).

mattlong (approved):

codyebberson:
@mattlong - Nice!  

I'm looking at https://github.com/medplum/medplum/pull/4082/commits/fa6ffc95a3bbac879a28d7b13143e9c26d1b6607

If we add `@mdx-js/react` to the root `package.json` and run `npm i`, would that also solve it?

Just slightly nervous that if we only fix it in `package-lock.json`, we might have the same issue next week.

mattlong:
@codyebberson To clarify, I got the lock file to its current state by adding and removing `@mdx-js/react` to the root `package.json`. The only harm of leaving it around in root would be not remembering why it's there.

codyebberson:
> I'd suggest we ship it as is for now and wait and see what happens over the next couple weeks of dep upgrades.

sounds good to me 👍

PR #4088, opened 2024-03-05
IF-MATCH header for resource contention

### LINKED ISSUE #3313: Implement IF-MATCH header for resource contention
The FHIR spec defines an `IF-MATCH` header for PUT operations, which causes the operation to fail if the latest version of the resource on the server : http://hl7.org/fhir/R4/http.html#concurrency

Implementing this header, and propagating this option to the SDK, will make it easier to handle high-concurrency workflows without having to resort to PATCH for updating

Revanee:
This PR implements [optimistic locking](https://hl7.org/fhir/R4/http.html#concurrency) using the `if-match` header for `PUT` operations.

Possible issues I've noticed:
- `lock-error` may not be the correct code to use for a `412 Precondition Failed`
- Using `IncomingHttpHeaders` in `FhirRequest` may not be the best approach
- Extracting the `versionId` with `versionId = if_match.slice(3, if_match.length - 1);` may not be robust enough

Closes #3313

### DIFF
--- a/packages/core/src/outcomes.test.ts
+++ b/packages/core/src/outcomes.test.ts
@@ -19,6 +19,7 @@ import {
   notFound,
   notModified,
   operationOutcomeToString,
+  preconditionFailed,
   tooManyRequests,
   unauthorized,
 } from './outcomes';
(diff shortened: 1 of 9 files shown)


codyebberson (approved):

codyebberson:
Thank you @Revanee !

PR #4110, opened 2024-03-06
Fixes #3899 - agent status api

### LINKED ISSUE #3899: Medplum Agent Command: Get Status

codyebberson:
Adds `Agent/{id}/$status` operation

Returns `status`:

- `connected` - The websocket connection is active
- `disconnected` - The websocket connection is closed
- `unknown` - No websocket connection, and no known status.  Either the agent has never connected, or it disconnected more than 24 hours ago.  (Open to better names for this)

This can be used with the Medplum CLI:

```bash
$ medplum get 'Agent/b6f04fb8-acac-44dd-a28f-423f8297ce70/$status'
{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "status",
      "valueCode": "disconnected"
    },
    {
      "name": "lastUpdated",
      "valueInstant": "2024-03-06T22:47:36.451Z"
    }
  ]
}
```

(Note that the operation uses a `$` character, so take special care to wrap with single quotes and not double quotes)

### DIFF
--- a/packages/app/src/resource/ToolsPage.tsx
+++ b/packages/app/src/resource/ToolsPage.tsx
@@ -96,7 +96,7 @@ export function ToolsPage(): JSX.Element | null {
           name="ip"
           placeholder="IP Address"
           rightSection={
-            <ActionIcon size={24} radius="xl" variant="filled" type="submit" aria-label="Ping">
+            <ActionIcon size={24} radius="xl" variant="filled" type="submit" aria-label="Ping" loading={pinging}>
               <IconRouter style={{ width: '1rem', height: '1rem' }} stroke={1.5} />
             </ActionIcon>
           }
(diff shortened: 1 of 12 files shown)


packages/app/src/resource/ToolsPage.tsx
```diff
@@ -1,76 +1,107 @@
...
+      <p>
+        Send a ping command from the agent to an IP address. Use this tool to troubleshoot local network connectivity.
+      </p>
+      <Form onSubmit={handlePing}>
+        <TextInput
+          id="ip"
+          name="ip"
+          placeholder="IP Address"
+          rightSection={
+            <ActionIcon size={24} radius="xl" variant="filled" type="submit" aria-label="Ping">
```
ThatOneBro, line 99: I think we have an unintentional UX regression after refactoring this button. Before it would show a spinner and disable the button while the ping was in progress. Now there is no indicator that a ping is in progress and you can queue up several concurrent ping requests without knowing it if you keep clicking the button.

ThatOneBro (commented):
Looks good for the most part, one comment about UX regression

packages/app/src/resource/ToolsPage.tsx
codyebberson: Ah, good catch.  Will fix

ThatOneBro (approved):
LGTM 👍

PR #4114, opened 2024-03-07
feat(server): add `Communication.topic` as search param

### LINKED ISSUE #4105: Add `Communication.topic` as a `SearchParameter`
`Communication.topic` acts a sort of subject line for a message or thread of messages. In [R4](https://hl7.org/fhir/R4/communication.html#search) it was not a search parameter but it is in [R5](https://hl7.org/fhir/R5/communication.html#search) and it would be useful if we had it for identifying threads in search.

TODO:
- [ ] Also update `medplum-live-chat-demo` to query by `topic` when searching for thread to use

ThatOneBro:
Closes #4105

Also adds autoincrementing for migration versions vs. previously hardcoded file names, as well as automatically exporting new migrations in the `schema/index.ts` barrel file.


packages/generator/src/migrate.ts
```diff
@@ -524,3 +528,12 @@ function buildIndexSql(tableName: string, index: IndexDefinition): string {
 if (process.argv[1].endsWith('migrate.ts')) {
   main().catch(console.error);
 }
+
+function getNextSchemaVersion(): number {
```
codyebberson, line 532: Nice, this was probably overdue 🙃

packages/server/src/migrations/schema/v62.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Generated by @medplum/generator
+ * Do not edit manually.
+ */
+
+import { PoolClient } from 'pg';
+
+export async function run(client: PoolClient): Promise<void> {
```
codyebberson, line 8: If we change `"string"` to `"token"`, this migration shouldn't be necessary.

codyebberson (commented):
The main thing is we should switch `"string"` to `"token"`

This model would have worked, but unfortunately you would need to manually add `v62.ts` to `packages/server/src/migrations/schema/index.ts`

codyebberson (approved):

Issue #4118, opened 2024-03-07
Update Documentation on Messaging


rahul1:
Our guidance on async encounters and communications are a bit out of date. Requesting the following changes: 

[Organizing Communication Threads](https://www.medplum.com/docs/communications/organizing-communications)
- [x] Make this page appear first in the sidebar (within comms and messaging)
- [x] Remove all references to "selecting a grouping resource", and any references to using `Encounter` as the grouping resource. 
- [x] Simplify this guidance to _only_ recommend `Communication` as the top-level resource for a thread. 
- [x] Standardize terminology. Call this top-level `Communication` the "thread header".

[Representing Async Encounters](https://www.medplum.com/docs/communications/async-encounters)
- [x] Update the guidance here to recommend _only_ setting `Communication.encounter` for the thread header 
- [x] Update the diagrams to reflect this new setup. Top-level `Communication` pointed to `Encounter`, with messages underneath

PR #4135, opened 2024-03-08
Update documentation on messaging

ksmith94:
Related to #4118 this will update the communications docs to recommend creating threads only with `Communication` resources.


packages/docs/docs/communications/async-encounters/async-encounters.md
```diff
@@ -74,21 +74,21 @@ The thread of [`Communication`](/docs/api/fhir/resources/communication) resource
 
 flowchart TD
```
rahul1, line 75: nit: make this `RL` as well

rahul1 (approved):

PR #4139, opened 2024-03-09
Add tests for ElementsInput component and utils.

willkrakow:
This also moves some helper functions / constants from `ElementsInput.tsx` to `ElementsInput.utils.ts` to better define the test suites.


mattlong (approved):
@willkrakow Thanks so much for this! I appreciate you taking the time to add some testing around this!

PR #4154, opened 2024-03-12
Fix uniqueness constraint on terminology properties

mattwiller:
Postgres treats `NULL`s as distinct by default, so multiple identical simple attribute values would produce duplicate rows in the DB


codyebberson (commented):
Why `simple`?

Should the existing `Coding_Property_idx` index be removed?


```
medplum=# \d "Coding_Property";
           Table "public.Coding_Property"
  Column  |  Type  | Collation | Nullable | Default
----------+--------+-----------+----------+---------
 coding   | bigint |           | not null |
 property | bigint |           | not null |
 target   | bigint |           |          |
 value    | text   |           |          |
Indexes:
    "Coding_Property_coding_idx" btree (coding)
    "Coding_Property_idx" UNIQUE, btree (coding, property, target, value)
```

mattwiller:
> Why simple?

Since it corresponds to UMLS "simple attributes", as opposed to relationships with a `target`

> Should the existing Coding_Property_idx index be removed?

Nope, it's still necessary to index relationship properties

codyebberson (approved):

PR #4160, opened 2024-03-13
Allow initializing Project without owner

mattwiller:


packages/server/src/fhir/operations/projectinit.ts
```diff
@@ -118,16 +117,16 @@ export async function projectInitHandler(req: Request, res: Response): Promise<v
...
 export async function createProject(
   projectName: string,
-  admin: User
+  admin?: User
 ): Promise<{
   project: Project;
-  profile: ProfileResource;
-  membership: ProjectMembership;
   client: ClientApplication;
+  profile?: ProfileResource;
```
codyebberson, line 128: If you wanted to get fancy / abuse TypeScript, you could make the return type conditional on the `admin` param.

See packages/server/src/test.setup.ts for an example

Something like:

```ts
export type CreateProjectResult<T extends User | undefined> = {
  project: Project;
  client: ClientApplication;
  profile: T extends User ? ProfileResource : undefined;
  membership: T extends User ? ProjectMembership : undefined;
};

export async function createProject<T extends User | undefined>(
  projectName: string,
  admin: T
): Promise<CreateProjectResult<T>> {
```

Totally optional.  It might clean up some of the casts that you had to add.

codyebberson (approved):
Nice.  See suggestion on TS abuse.  lgtm 👍

packages/server/src/fhir/operations/projectinit.ts
mattwiller: Ooooh nice!  I had attempted something similar initially, but didn't find this correct incantation

PR #4170, opened 2024-03-14
Bugfix: domain-level IDP should be case-insensitive

rahul1:
As currently implented, domain level IdPs treat the email domain as case-sensitive
(e.g. mydomain.com is different than MyDomain.com). 

This creates problems when the Idp (e.g. Azure SSO) stores the user info with different casing than they way the user enters their email


charlesbthomas:
Thanks @rahul1! You all are the best. Top tier support! 🙂

codyebberson (approved):

PR #4176, opened 2024-03-16
Add heap size to healthcheck

codyebberson:
Hold on this - testing in staging


ThatOneBro (approved):

codyebberson:
Upon further reflection, I don't think we want this in the public healthcheck endpoint.  It would be valuable in a super admin "server stats" endpoint though.

PR #4182, opened 2024-03-18
Implement Binary.securityContext check

codyebberson:


packages/server/src/fhir/rewrite.ts
```diff
@@ -157,6 +157,9 @@ class Rewriter {
       } else {
         binary = await this.repo.readResource<Binary>('Binary', id);
       }
+      if (binary.securityContext) {
+        await this.repo.readReference(binary.securityContext);
```
rahul1, line 161: Assume the idea here is that this will throw an error if you don't have access to the security context? 

Clever! I just suggest a comment to explain

rahul1 (approved):

ThatOneBro (approved):

PR #4187, opened 2024-03-18
Disable search for FHIR Binary

codyebberson:
Future work:

- Track and add `Binary` resources in `$export` operation
    - I'm not sure how valuable this is in current form, because we currently do not export `Binary` contents anyway
    - Internally, Medplum stores `Binary` contents in a `BinaryStorage` (i.e., AWS S3 or the file system)
    - The FHIR Bulk Export API does not mention anything about how to handle this case
    - We could try to jam it into `Binary.data` as a base64 encoded string
    - But we routinely process massive 1GB+ files, which would not work at all
    - We could try to set some kind of threshold?
- Review behavior of `includeIds` in `Project/{id}/$clone` operation
    - The existing API assumes that IDs are unique across resource types
    - That is true today, but we are actively preparing for that not to be true

### DIFF
--- a/packages/server/src/fhir/operations/export.ts
+++ b/packages/server/src/fhir/operations/export.ts
@@ -115,7 +115,14 @@ function getResourceTypesByExportLevel(exportLevel: string): ResourceType[] {
   return getResourceTypes();
 }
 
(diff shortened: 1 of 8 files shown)


ThatOneBro (approved):

ThatOneBro:
One thing, can we get a test for `getBinaryIds()` for coverage? 😄

codyebberson:
@ThatOneBro  ✅

PR #4192, opened 2024-03-19
Fixes #4186 - resource type validation for search with _type

### LINKED ISSUE #4186: Search for multiple types without base resource is not possible
I think this should be possible `GET [base]/?_type=Task,CarePlan` but currently leads to the OperationOutcome `Unknown resource type`. This is possible `GET [base]/Task?_type=Task,CarePlan` but rather weird.

https://www.hl7.org/fhir/R4/search.html#_type

codyebberson:


ThatOneBro (approved):

packages/server/src/fhir/routes.test.ts
```diff
@@ -587,4 +587,39 @@ describe('FHIR Routes', () => {
...
+        .send({
+          resourceType: 'Observation',
+          status: 'final',
+          code: { text: 'test' },
+          subject: { reference: `Patient/${res1.body.id}` },
+        });
+      expect(res2.status).toBe(201);
+
+      const res3 = await request(app)
+        .get('/fhir/R4?_type=Patient,Observation')
```
ThatOneBro, line 613: Actually is `/fhir/R4/?_type=Patient,Observation` also supported here?

codyebberson: Ah, the trailing slash?  Good question.  I'll investigate

codyebberson: Confirmed, works.  Spec is ambiguous.  I added a test to lock it in with this behavior.

PR #4193, opened 2024-03-19
feat(getwsbindingtoken): add feature gate on `websocket-subscriptions`

### LINKED ISSUE #4190: Return 400 on `Subscription/$get-ws-binding-token` if project doesn't have WebSocket Subscriptions enabled

ThatOneBro:
Closes #4190 

This should save a lot of debugging time by preventing connecting to the `WebSocket Subscription` endpoint early by returning an `OperationOutcome` error when a project doesn't have the feature enabled.


codyebberson (approved):
Nice 👍

PR #4201, opened 2024-03-19
Add support for US Core Search Params

mattlong:
Adds the seven Search Parameters listed in [section 6.2.1](https://hl7.org/fhir/us/core/STU5.0.1/search-parameters-and-operations.html#search-parameters-defined-by-this-implementation-guide) of the US Core Implementation Guide.

The search parameter are defined at the system level regardless of the presence of the related US Core resources. This is acceptable since the expressions of each search parameter will only evaluate to a non-empty value when encountering resources conforming the the related US Core profiles.

Since the US Core profiles are not available at the time the search parameter definitions are parsed, a couple special cases are introduced in `packages/core/src/search/details.ts`. As suggested by @mattwiller, we could instead add custom extensions to the SearchParameter definitions that `packages/core/src/search/details.ts` looks for to override default behavior rather than having one-off code. Happy to go that route if we think that's a cleaner approach.


codyebberson (approved):
Nice, this works for me. 👍 

I understand the discomfort with `details.ts`, but I think the current implementation makes sense.

I'm on the fence about adding these to `search-parameters-medplum.json` vs a new `search-parameters-uscore.json` file.  I would definitely prefer the hygeine of a new file, but there is a kinda obnoxious setup right now where you would need to make more calls to `indexSearchParameterBundle()` in a few places.  I'm ok with as-is.

packages/core/src/types.ts
```diff
@@ -138,6 +138,16 @@ export interface TypeInfo {
   searchParamsDetails?: Record<string, SearchParameterDetails>;
 }
 
+/**
+ * The list of all known search parameter definition bundles
+ * filenames from `@medplum/definitions`.
+ */
+export const SEARCH_PARAMETER_BUNDLE_FILES = [
```
codyebberson, line 145: Would this make more sense in `@medplum/definitions` - `packages/definitions/src/index.ts`

PR #4202, opened 2024-03-19
FHIR search lookup table optimizations

codyebberson:
Consider the process of writing a resource:

1. Start a postgres transaction
2. Write the resource to the main resource table
3. Write a version to the history table
4. Update lookup tables
5. Commit the transaction

Now focus on step 4: "Update lookup tables"

Most FHIR search parameters are implemented as precomputed values directly on the main resource table.  For example, `Patient-birthdate` is a simple `date` search parameter, and therefore is a simple column on the `Patient` table.

Some FHIR search parameters have more complex behaviors, and require additional tables.  For example, complex `token` parameters need lookup tables to fully support all of the different syntaxes of `token` search (i.e., `value`, `system|value`, `system|`, `|value`, etc).

So, when writing a resource, we iterate over all of the lookup table implementations:

1. Does this lookup table apply to this resource + search parameter?
2. If yes, continue
3. Get existing values from the lookup table
4. Compare the existing values with the values in the updated resource
5. If changed, delete the existing values and write the new values

In practice, that is suboptimal.

1. We do more "creates" than "updates", so the create flow should be optimized
2. In practice, conditionally updating the lookup values rarely saved us anything
3. `SELECT`+`DELETE`+`INSERT` is slower than `DELETE`+`INSERT`

So the new process implemented in this PR is:

1. Does this lookup table apply to this resource + search parameter?
2. If yes, continue
3. If creating, delete existing values
4. Write new values

Also, bonus, expliclty excluding `_compartment` search parameter from `references.ts`.  The `_compartment` search parameter is an internal Medplum thing and handled separately.

### DIFF
--- a/packages/server/src/fhir/lookups/address.ts
+++ b/packages/server/src/fhir/lookups/address.ts
@@ -99,20 +99,18 @@ export class AddressTable extends LookupTable {
     const resourceId = resource.id as string;
     const values = [];
 
-    for (let i = 0; i < addresses.length; i++) {
-      const address = addresses[i];
-      values.push({
-        id: randomUUID(),
-        resourceId,
-        index: i,
-        content: stringify(address),
-        address: formatAddress(address),
-        city: address.city?.trim(),
-        country: address.country?.trim(),
-        postalCode: address.postalCode?.trim(),
-        state: address.state?.trim(),
-        use: address.use?.trim(),
-      });
+const values = addresses.map((address, index) => ({
+  id: randomUUID(),
+  resourceId,
+  index,
+  content: stringify(address),
+  address: formatAddress(address),
+  city: address.city?.trim(),
+  country: address.country?.trim(),
+  postalCode: address.postalCode?.trim(),
+  state: address.state?.trim(),
+  use: address.use?.trim(),
+}));
     }
 
     await this.insertValuesForResource(client, resourceType, values);
--- a/packages/server/src/fhir/lookups/address.ts
+++ b/packages/server/src/fhir/lookups/address.ts
@@ -97,21 +97,18 @@ export class AddressTable extends LookupTable {
 
     const resourceType = resource.resourceType;
     const resourceId = resource.id as string;
-    const values = [];
-
-const values = addresses.map((address, index) => ({
-  id: randomUUID(),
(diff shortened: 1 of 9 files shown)


codyebberson:
Using `pg_stat_statements` and a clean run of server tests `npm t`:

```sql
SELECT
  CASE
    WHEN query ILIKE 'CREATE%' THEN 'CREATE'
    WHEN query ILIKE 'SELECT%' THEN 'SELECT'
    WHEN query ILIKE 'INSERT%' THEN 'INSERT'
    WHEN query ILIKE 'UPDATE%' THEN 'UPDATE'
    WHEN query ILIKE 'DELETE%' THEN 'DELETE'
    ELSE 'OTHER'
  END AS query_type,
  COUNT(*) AS total_queries
FROM pg_stat_statements
GROUP BY query_type
ORDER BY total_queries DESC;
```

| Query Type | Before | After | Change |
| --- | --- | --- | --- |
| `SELECT` | 1056 | 950 | -10% |
| `INSERT` | 325 | 283 | -13% |
| `UPDATE` | 172 | 172 | 0% |
| `DELETE` | 134 | 137 | +3% |

packages/server/src/fhir/lookups/address.ts
```diff
@@ -83,43 +82,40 @@ export class AddressTable extends LookupTable<Address> {
...
+        resourceId,
+        index: i,
+        content: stringify(address),
+        address: formatAddress(address),
+        city: address.city?.trim(),
+        country: address.country?.trim(),
+        postalCode: address.postalCode?.trim(),
+        state: address.state?.trim(),
+        use: address.use?.trim(),
+      });
```
mattwiller, line 115: This looks like a case where a straightforward `map()` more clearly expresses the intent of the code:

```suggestion
const values = addresses.map((address, index) => ({
  id: randomUUID(),
  resourceId,
  index,
  content: stringify(address),
  address: formatAddress(address),
  city: address.city?.trim(),
  country: address.country?.trim(),
  postalCode: address.postalCode?.trim(),
  state: address.state?.trim(),
  use: address.use?.trim(),
}));
```

mattwiller (approved):

PR #4207, opened 2024-03-20
fix(subscription): enable checking `AccessPolicy` for WS subs

### LINKED ISSUE #4205: Access Policies for Websockets Subscriptions only
Enforce AccessPolicies on websocket subscriptions only.
These subscriptions should use the AccessPolicy of ProjectMemberhsip that created them.

ThatOneBro:
Closes #4205


packages/server/src/workers/subscription.ts
```diff
@@ -134,13 +134,19 @@ export function getSubscriptionQueue(): Queue<SubscriptionJobData> | undefined {
  *
  * Currently we log if the `AccessPolicy` is not satisfied only.
  *
- * TODO: Actually prevent notifications for `Subscriptions` where the `AccessPolicy` is not satisfied.
+ * TODO: Actually prevent notifications for `Subscriptions` where the `AccessPolicy` is not satisfied (for rest-hook subscriptins)
```
rahul1, line 137: type
```suggestion
 * TODO: Actually prevent notifications for `Subscriptions` where the `AccessPolicy` is not satisfied (for rest-hook subscriptions)
```

packages/server/src/workers/subscription.ts
```diff
@@ -157,6 +163,9 @@ async function checkAccessPolicy(resource: Resource, project: Project, subscript
             `[Subscription Access Policy]: Access Policy not satisfied on '${resourceReference}' for '${subReference}'`,
             { subscription: subReference, project: projectReference, accessPolicy }
           );
+          if (subscription.channel.type === 'websocket') {
```
rahul1, line 166: Consolidate with a single ` if (subscription.channel.type === 'websocket') {}` at the end. 

This will make it easier to remove this case later

packages/server/src/workers/subscription.ts
ThatOneBro: Good catch

packages/server/src/workers/subscription.ts
```diff
@@ -180,7 +181,10 @@ async function checkAccessPolicy(resource: Resource, project: Project, subscript
       `[Subscription Access Policy]: Error occurred while checking access policy for resource '${resourceReference}' against '${subReference}'`,
       { error: err, subscription: subReference }
     );
+    satisfied = false;
   }
+  // Always return true if channelType !== websocket for now
+  return subscription.channel.type === 'websocket' ? satisfied : true;
```
codyebberson, line 187: Nice, I think this is the safe strategy for now 👍

codyebberson (approved):

PR #4223, opened 2024-03-22
fix(server): Fix server tests failing intermittently due to `ioredis`

ThatOneBro:
A few contributors have reported flakiness of tests in `packages/server/src/app.test.ts`, specifically since using real Redis in tests. There were a few things that indicated it may be to weird interplay between `jest` and `redis`, although the details are unclear. I tweaked a few things that makes it feel more stable (haven't had another test failure yet) but not 100% confident this is the fix.

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -206,10 +206,10 @@ export function initAppServices(config: MedplumServerConfig): Promise<void> {
 }
 
 export async function shutdownApp(): Promise<void> {
-  await closeWorkers();
-  await closeDatabase();
   cleanupHeartbeat();
   await closeWebSockets();
+  await closeWorkers();
+  await closeDatabase();
   await closeRedis();
   closeRateLimiter();
 
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -214,7 +214,9 @@ export async function shutdownApp(): Promise<void> {
   closeRateLimiter();
 
   if (server) {
-    server.close();
+    await new Promise((resolve) => {
+      (server as ReturnType<typeof http.createServer>).close(resolve);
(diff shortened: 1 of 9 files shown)


packages/server/src/app.ts
```diff
@@ -214,7 +214,9 @@ export async function shutdownApp(): Promise<void> {
   closeRateLimiter();
 
   if (server) {
-    server.close();
+    await new Promise((resolve) => {
+      (server as ReturnType<typeof http.createServer>).close(resolve);
```
codyebberson, line 218:
```suggestion
      (server as http.Server).close(resolve);
```

codyebberson (approved):

packages/server/src/websockets.ts
```diff
@@ -34,12 +41,34 @@ export function initWebSockets(server: http.Server): void {
     // See: https://github.com/websockets/ws/blob/master/doc/ws.md#websocketbinarytype
     socket.binaryType = 'nodebuffer';
 
+    if (!wsState?.sockets.size) {
+      let socketsClosedResolve!: () => void;
+      const socketsClosedPromise = new Promise<void>((resolve) => {
+        socketsClosedResolve = resolve;
+      });
+      wsState = { sockets: new Set(), socketsClosedPromise, socketsClosedResolve };
```
codyebberson, line 49: Tricky

codyebberson (approved):
lgtm 👍 

This only affects tests & server shutdown, right?  Looks safe.

ThatOneBro:
> lgtm 👍
> 
> This only affects tests & server shutdown, right? Looks safe.

Yeah this is only regarding tests and server shutdown. Thanks 🙏

PR #4226, opened 2024-03-22
Use FhirRouter for FHIR operations

codyebberson:
Before: Most FHIR operation routes were handled by the Express router, not by the `FhirRouter`

That wasn't inherently bad, but it meant that those FHIR operations were not accessible via FHIR batch operations.

After:  Most FHIR operation routes are now handled by `FhirRouter`, and therefore are available via FHIR batch.

This paves the way in the future for projects to be able to configure Medplum Bots as FHIR operation handlers...

Note I said "most", as some operations are currently not compatible with the `FhirRouter` API:

* `$csv` - exports a CSV representation of search results, and therefore sets content-type, etc
* `Bot/$execute` - Bot execution can return HL7, DICOM, etc
* `Agent/$push` - Agents can return HL7, DICOM, etc

The diff is large, but it's actually pretty straightforward and procedural changes.

Two of our older FHIR operations predate @mattwiller's FHIR `Parameter` utils, so I went ahead and ported those over:

* `Measure/$evaluate-measure`
* `PlanDefinition/$apply`

In both cases, the transition was pretty straightforward, just different error messages.

### DIFF
--- a/packages/server/src/fhir/operations/README.md
+++ b/packages/server/src/fhir/operations/README.md
@@ -18,7 +18,7 @@ Utility functions that use the `OperationDefinition` to automate implementation 
 - [`parseInputParameters`](./utils/parameters.ts) takes either standard `Parameters` input or plain JSON key-value pairs
   for compatibility, and produces an object containing all defined input parameter values
   - Validates number of parameter values, including required parameters
-- [`sendOutputParameters`](./utils/parameters.ts) handles sending a set of output values as the response
+- [`buildOutputParameters`](./utils/parameters.ts) handles sending a set of output values as the response
   - Only sends parameters defined in the `OperationDefinition`
   - Validates parameter cardinality and throws a server error when response wouldn't match `OperationDefinition`
   - Can be passed a `Resource` when there is one matching output parameter named `return`
@@ -26,7 +26,7 @@ Utility functions that use the `OperationDefinition` to automate implementation 
 ### Example: Project $init
 
 ```ts
-import { parseInputParameters, sendOutputParameters } from './utils/parameters';
+import { buildOutputParameters, parseInputParameters } from './utils/parameters';
 import { created } from '@medplum/core';
 import { Reference, OperationDefinition } from '@medplum/fhirtypes';
 import { Request, Response } from 'express';
@@ -58,13 +58,13 @@ interface ProjectInitParameters {
   ownerEmail?: string;
 }
 
-export async function projectInitHandler(req: Request, res: Response): Promise<void> {
+export async function projectInitHandler(req: FhirRequest): Promise<FhirResponse> {
   const params = parseInputParameters<ProjectInitParameters>(operation, req);
 
   // Handle operation business logic...
   const project = doProjectInit(params);
 
   // Special case: single `return` output parameter means respond with the Project resource directly
-  await sendOutputParameters(req, res, operation, created, project);
+  return [created, project];
 }
 ```
--- a/packages/server/src/fhir/operations/evaluatemeasure.ts
+++ b/packages/server/src/fhir/operations/evaluatemeasure.ts
@@ -1,4 +1,5 @@
-import { Operator, badRequest, created, parseSearchRequest } from '@medplum/core';
+import { Operator, created, parseSearchRequest } from '@medplum/core';
+import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
 import {
   Measure,
   MeasureGroup,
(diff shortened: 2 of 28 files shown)


dangerfile.ts
```diff
@@ -16,7 +16,7 @@ if (packageChanged && !lockfileChanged) {
 const modifiedFiles = danger.git.modified_files.filter((path) => /\/src\/.+\.tsx?/.exec(path));
 
 // Check for console.log statements
-const statements = ['console.debug', 'console.log', 'console.warn', 'describe.only', 'test.only'];
```
mattwiller, line 19: Was this removal intentional?

packages/server/src/fhir/operations/README.md
```diff
@@ -58,13 +58,13 @@ interface ProjectInitParameters {
   ownerEmail?: string;
 }
 
-export async function projectInitHandler(req: Request, res: Response): Promise<void> {
+export async function projectInitHandler(req: FhirRequest): Promise<FhirResponse> {
```
mattwiller, line 61: Thanks for updating the docs as well!

packages/server/src/fhir/operations/evaluatemeasure.ts
```diff
@@ -1,18 +1,41 @@
...
+    { name: 'periodStart', use: 'in', min: 1, max: '1', type: 'date' },
+    { name: 'periodEnd', use: 'in', min: 1, max: '1', type: 'date' },
+    { name: 'measure', use: 'in', min: 0, max: '1', type: 'Reference' },
+    { name: 'reportType', use: 'in', min: 0, max: '1', type: 'code' },
+    { name: 'subject', use: 'in', min: 0, max: '1', type: 'Reference' },
+    { name: 'practitioner', use: 'in', min: 0, max: '1', type: 'Reference' },
+    { name: 'lastReceivedOn', use: 'in', min: 0, max: '1', type: 'dateTime' },
+    { name: 'return', use: 'out', min: 1, max: '1', type: 'MeasureReport' },
+  ],
+};
```
mattwiller, line 38: Why not use the [actual FHIR spec definition](http://hl7.org/fhir/R4/measure-operation-evaluate-measure.html) of this one?

```ts
const operation = getOperationDefinition('Measure', 'evaluate-measure');
```

packages/server/src/fhir/operations/plandefinitionapply.ts
```diff
@@ -1,41 +1,49 @@
...
+    { name: 'practitioner', use: 'in', min: 0, max: '1', type: 'string', searchType: 'reference' },
+    { name: 'organization', use: 'in', min: 0, max: '1', type: 'string', searchType: 'reference' },
+    { name: 'userType', use: 'in', min: 0, max: '1', type: 'CodeableConcept' },
+    { name: 'userLanguage', use: 'in', min: 0, max: '1', type: 'CodeableConcept' },
+    { name: 'userTaskContext', use: 'in', min: 0, max: '1', type: 'CodeableConcept' },
+    { name: 'setting', use: 'in', min: 0, max: '1', type: 'CodeableConcept' },
+    { name: 'settingContext', use: 'in', min: 0, max: '1', type: 'CodeableConcept' },
+    { name: 'return', use: 'out', min: 1, max: '1', type: 'CarePlan' },
+  ],
+};
```
mattwiller, line 43:
```suggestion
const operation = getOperationDefinition('PlanDefinition', 'apply');
```

packages/server/src/fhir/operations/utils/parameters.ts
```diff
@@ -223,13 +187,8 @@ export async function sendOutputParameters(
     }
   }
 
-  try {
-    validateResource(response);
-    res.status(getStatus(outcome)).json(response);
-  } catch (err: any) {
-    getLogger().error('Malformed operation output Parameters', { error: err.toString() });
```
mattwiller, line 230: Up to you whether we want to keep this log; it was intended to help catch server issues

mattwiller (approved):

dangerfile.ts
codyebberson: Yes...  We just added this yesterday.  I was on the fence about `console.log` and all, and this PR triggered a bunch of false positives.

The "recipe" that we're currently using scans the entire file if you touch a single line.  So, in this case, I touched `deploy.ts` which has valid use of `console.log()` (inside the bot), and no way to escape it / ignore it.

Ideally, (1) Danger.js would only look at lines changed and (2) Danger.js would provide an escape hatch for "ignore".

That's low pri future work though.

packages/server/src/fhir/operations/evaluatemeasure.ts
codyebberson: Oh, nice.

packages/server/src/fhir/operations/README.md
codyebberson: const operation = getOperationDefinition('Measure', 'evaluate-measure');

codyebberson: "Replace in files" found it 🤣

PR #4237, opened 2024-03-25
Index-aware validation expressions

### LINKED ISSUE #3447: Array-aware validation outcome expressions
The expressions produced during validation should be specific to the array indices that have issues.

Consider the scenario where `Patient.identifier.system` is required and more than one identifiers exist; some of which have `identifier.system` specified . Since only an expression for `Patient.identifier.system` is included in the outcome, the options are to either highlight all system inputs or disable error highlighting on array inputs:

Screenshot 2023-11-28 at 11 37 38 AM

Screenshot 2023-11-28 at 11 37 15 AM

Ideally, the expression returned from validation in the above scenario would instead be `Patient.identifier[1].system`. Assuming that were available, on the frontend, `ResourceArrayInput` would need to pass on the array index a child element input is at and `getErrorsForInput` would need to take that index (if applicable) as a parameter.

mattlong:
Fixes #3447.

This turned out to be a more invasive on the validation code than expected, mostly due to how `ResourceCrawler` and `getNestedProperty` flattens the set of values found for a given key.

The changes in `@medplum/react` were tedious but all pretty relatively straightforward. I used this as an opportunity to further centralize a lot of the repeated prop definitions in `ResourcePropertyInput` for primitive types. Note that slices are not yet supported for index-aware validation expressions. The previous behavior is preserved within slices by the new `indexedPath` property not being propagated to children in `SliceInput`.


packages/react/src/ResourceArrayInput/ResourceArrayInput.tsx
```diff
@@ -101,6 +100,7 @@ export function ResourceArrayInput(props: ResourceArrayInputProps): JSX.Element
                 property={props.property}
                 name={props.name + '.' + valueIndex}
                 path={props.path}
+                indexedPath={getIndexedPath(props.path, props.indexedPath, valueIndex)}
```
codyebberson, line 103: Can you help me build intuition here regarding `path` vs `indexedPath`?  I assume `indexedPath` has something to do with slices?

packages/core/src/typeschema/crawler.ts
```diff
@@ -103,27 +112,78 @@ class ResourceCrawler {
...
 ): (TypedValue | TypedValue[] | undefined)[] {
+  if (value === undefined) {
+    return [undefined];
+  }
+
   if (key === '$this') {
     return [value];
   }
+
+  const propertyGetter = options?.withIndexedPath ? getTypedPropertyValueWithIndexedPath : getTypedPropertyValue;
```
codyebberson, line 137: Similar question here.... what are the conditions we would use `getTypedPropertyValueWithIndexedPath` vs `getTypedPropertyValue`?

Would there be any harm if we simply always returned the path?  (aside from perf hit)

codyebberson (commented):
Overall, looks good to me

I'll follow-up with you offline on `path` vs `indexedPath`, I don't think I fully understand the distinction.

I think this has the potential to be used in many different places.  I'd be open to making the `withIndexedPath` on by default assuming the perf hit isn't too bad.

My naive assumption is that the type would have been called `TypedValueWithPath`, but I think that gets back to not understanding `path` vs `indexedPath`.

We should get @mattwiller  to look at this as well.

packages/core/src/typeschema/crawler.ts
mattlong: On the surface, no, I don't think there'd be an issue with always returning a path besides `TypedValue` being a bit of a misnomer since the object includes more than just `type` and `value`.

Another concern with making that blanket change is the behavior around `isEmpty`/`undefined` values being represented as `{ type: 'undefined', value: undefined}` versus simply `undefined`. I think the latter would be preferable since return types of signatures could be a bit simpler: `TypedValue | TypedValue[]` instead of the current pattern of `TypedValue | TypedValue[] | undefined`. But that would be another smaller performance hit.

packages/react/src/ResourceArrayInput/ResourceArrayInput.tsx
mattlong: UPDATE: I went through and replace all the naming from `indexedPath` to `expression`.

packages/core/src/typeschema/crawler.ts
```diff
@@ -103,27 +107,78 @@ class ResourceCrawler {
...
+function getTypedPropertyValueWithExpression(
+  input: TypedValue | TypedValueWithExpression,
+  path: string,
+  options?: GetTypedPropertyValueOptions
+): TypedValueWithExpression[] | TypedValueWithExpression {
+  const parentExpression = (input as TypedValueWithExpression).expression;
+  return withExpression(getTypedPropertyValue(input, path, options), parentExpression, path);
+}
+
+export type TypedValueWithExpression = TypedValue & { expression: string };
```
mattwiller, line 163: I would suggest "path" over "expression" for this concept to more accurately describe what's being tracked, given that these are explicitly paths into the resource

mattwiller (approved):

packages/core/src/typeschema/crawler.ts
mattlong: Summarizing an offline discussion: since `path` would collide with the same-named property on React component inputs, `expression` is the next most specific option, so going with that for now.

PR #4254, opened 2024-03-26
Add @medplum/definitions to bot layer

rahul1:


codyebberson (commented):
I think you need to run `npm i` to update `package-lock.json`

codyebberson (approved):

PR #4257, opened 2024-03-26
Email verification

### LINKED ISSUE #2515: Support Email Verification Link
User Driven - https://discord.com/channels/905144809105260605/1131849169095639084/1131849169095639084

Our user has cloned foo medical and successfully set up the invite process. They would now like to add an email verification link to the welcome email the patient receives. While we do currently support [custom welcome emails](https://www.medplum.com/docs/auth/custom-welcome-emails), I don't think we have all the functionality to implement this via Bots. 

Some thoughts on implementation: 
- Allow bots to be executed by GET requests. Then users can create custom emails with a link to an email verification Bot
- Support this a native part of the `/newuser` flow somehow?

mattlong:
Fixes #2515 

To support email verification flow, e.g. including a link in a custom welcome email to new users, the `/auth/verifyemail` endpoint is added in conjunction with the new `UserSecurityRequest` resource type.

`UserSecurityRequest` is just about identical to `PasswordChangeRequest`, but with the addition of `verify-email`to `UserSecurityRequest.type`. 

The `/auth/verifyemail` endpoint is quite similar to `/auth/setpassword` with the critical difference of course being that it does NOT attempt to change a user's password; it simply sets `User.emailVerified` to `true` if the `UserSecurityRequest` is valid.

The `PasswordChangeRequest` type is now DEPRECATED; `UserSecurityRequest` is a drop-in replacement for it.

For now, email verification `UserSecurityRequest` resources must be created manually; there are no endpoints that will create one for a given user. A random and sufficiently long `UserSecurityRequest.secret` should be specified, such as [16 random bytes formatted as a hex string](https://github.com/medplum/medplum/blob/41ea0aee9ee97103926994d089cc41caa16f061e/packages/server/src/oauth/keys.ts#L179-L181).


packages/server/src/auth/setpassword.ts
```diff
@@ -17,19 +17,31 @@ export const setPasswordValidator = makeValidationMiddleware([
 
 export async function setPasswordHandler(req: Request, res: Response): Promise<void> {
   const systemRepo = getSystemRepo();
-  const pcr = await systemRepo.readResource<PasswordChangeRequest>('PasswordChangeRequest', req.body.id);
+  let securityRequest: UserSecurityRequest | PasswordChangeRequest;
 
-  if (pcr.used) {
+  // PasswordChangeRequest is deprecated but still supported. When it is removed, this try/catch can be removed
```
codyebberson, line 22: 👍

codyebberson (approved):
Nice, clean

PR #4265, opened 2024-03-27
Replace pg-cursor with iterative search

codyebberson:
Before, we used `pg-cursor` for 2 similar use cases:
1. Reindex all resources - a super admin function to rebuild the search index for a resource type
2. Rebuild compartments - a super admin function to rebuild compartment indexes for a resource type

In both cases, the server would process all rows of a resource type table to perform the internal work.

We used `pg-cursor` to use Postgres cursor functionality.  This guaranteed processing of all rows in a table, no duplicates, no gaps.  However, for large tables, it required too much temp memory, which would often lead to paging and system instability.

In this PR, we swap the `pg-cursor` implementation with a simple iterative search.  The logic is:
1. Search for resources ordered by last updated time since `currentTimestamp` (initially 0)
2. For each resource in the search result, execute the appropriate callback
3. Update `currentTimestamp` with the maximum `meta.lastUpdated` time
4. If there are more resources, go back to step 1

The operating principle here is that `lastUpdated` time can only go up, and that we're marching forward in time, and maintaining `currentTimestamp` along the way.

For small tables, this will be slower than `pg-cursor`.  But speed is not really the issue here.  These features are used infrequently.  The main objective is predictability and stability.


packages/server/src/admin/super.ts
```diff
@@ -37,10 +37,7 @@ superAdminRouter.post(
     requireSuperAdmin();
     requireAsync(req);
 
-    await sendAsyncResponse(req, res, async () => {
-      await rebuildR4ValueSets();
-      await getSystemRepo().reindexResourceType('CodeSystem');
-    });
+    await sendAsyncResponse(req, res, async () => rebuildR4ValueSets());
```
mattwiller, line 40: Why was the CodeSystem reindex removed?  I think this will be necessary to transition self-hosted customers to the new terminoloogy service implementation.

mattwiller (approved):

packages/server/src/admin/super.ts
codyebberson: I thought the terminology service results are built dynamically at query-time, so that was not necessary?

`reindexResourceType('CodeSystem')` is big and slow, so tests were timing out.  I was under the impression that it is not a necessary step after `rebuildR4ValueSets()`.

mattwiller: No, I believe it is necessary to transition self-hosted customers.  How else will all the R4 CodeSystem resources get loaded into the `Coding` table?

PR #4271, opened 2024-03-28
fix(expo-polyfills): init and recover when can't get `___keys___`

ThatOneBro:


packages/expo-polyfills/src/index.ts
```diff
@@ -210,7 +210,15 @@ class SyncSecureStorage implements Storage {
             })
             .catch(reject);
         })
-        .catch(reject);
+        .catch((err: Error) => {
+          console.error(err);
+          SecureStore.deleteItemAsync('___keys___')
```
codyebberson, line 215: 👍 

I don't know if it's worth trying to catch this one specific error?  Probably not worth the risk of false positives / false negatives.

ThatOneBro: What do you mean exactly? Which specific error? Or what is the alternative

ThatOneBro: Oh I think I understand, like looking at the `error.message` for only this error?

Yeah I think if `getItemAsync` fails it's more robust just to start over 👍

packages/expo-polyfills/src/index.ts
```diff
@@ -176,42 +176,46 @@ class SyncSecureStorage implements Storage {
...
-              resolve();
-            })
-            .catch(reject);
-        })
-        .catch(reject);
-    });
+    this.initPromise = this.initStorage();
+  }
+
+  private async initStorage(): Promise<void> {
```
ThatOneBro, line 182: The init logic before was admittedly very confusing with the nested promises

Thank god for async fns

packages/expo-polyfills/src/index.ts
```diff
@@ -176,42 +176,46 @@ class SyncSecureStorage implements Storage {
...
+      // We've chose `___keys___` to prevent collisions
+      // Here we get the keys and attempt to restore them
+      keysStr = await SecureStore.getItemAsync('___keys___');
+    } catch (err: unknown) {
+      // If an error is thrown here, it's likely an issue with encryption...
+      // We may be unable to decrypt existing keys if the keystore itself has changed
+      // In that case we should dump the keys and start fresh
+      // We can just initialize and resolve the `initPromise`
+      console.error(err);
+      await SecureStore.deleteItemAsync('___keys___');
```
codyebberson, line 196: Do we need to wrap this in `try`/`catch` too?

At least now that the method is `async`, if this fails the promise will reject, so it's not as urgent.

codyebberson (approved):

packages/expo-polyfills/src/index.ts
ThatOneBro: Yeah I think if delete fails we probably want it to reject the promise. Something is probably wrong in that case

PR #4274, opened 2024-03-29
Update danger.js to handle null modified_files

codyebberson:
😠 

https://github.com/medplum/medplum/actions/runs/8478316335/job/23230462576

image

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -4,6 +4,9 @@ import { readFileSync, statSync } from 'fs';
 
 // Gather changes
 const modifiedFiles = danger.git.modified_files ?? [];
+if (!modifiedFiles.length) {
+  warn('No files were modified by this PR. Make sure this is expected!');
+}
 const modifiedSourceFiles = modifiedFiles.filter((path) => /\/src\/.+\.tsx?/.exec(path));
 
 // Keep package-lock.json up to date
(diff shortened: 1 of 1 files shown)


ThatOneBro (approved):

ThatOneBro:
There is a separate issue hidden here, the fact that it still passes even though danger reported failures

dangerfile.ts
```diff
@@ -2,22 +2,23 @@ import { execSync } from 'child_process';
 import { danger, message, warn } from 'danger';
 import { readFileSync, statSync } from 'fs';
 
+// Gather changes
+const modifiedFiles = danger.git.modified_files ?? [];
+const modifiedSourceFiles = modifiedFiles.filter((path) => /\/src\/.+\.tsx?/.exec(path));
+
```
ThatOneBro, line 8:
```suggestion

if (!modifiedFiles.length) {
  warn('No files were modified by this PR. Make sure this is expected!'); 
}
```
I think in most cases this is indication that something is not right. This should almost never happen and I think the failure might actually be warranted. Could be that it's not actually running on the right branch.

ThatOneBro (changes requested):

dangerfile.ts
ThatOneBro: Maybe this should be a fail and not a warn too

codyebberson: Yes, agree, I think this implies something isn't configured correctly.

My short term goal is to just stabilize this and get it working consistently again.

Longer term, my honeymoon period with Danger.js is over, and I'm not sure this is the right tool for us.  I like the idea of having something like this for medplum specific checks where we can catch common issues.  There might be a simpler option for that though.

ThatOneBro (approved):

PR #4283, opened 2024-03-29
Fixes #4276 - allow super admin bots

### LINKED ISSUE #4276: Super Admin Bots can't search multiple Projects
## Repro

On a server to > 1 project

- Create a new project
- Mark the new project as Super Admin
- Create a Bot with the code below
- Mark the Bot as `admin`

```ts
import { BotEvent, MedplumClient, getReferenceString } from '@medplum/core';

export async function handler(medplum: MedplumClient, event: BotEvent): Promise<any> {
  const profile = await medplum.getProfileAsync();
  console.log("Profile: ", profile ? getReferenceString(profile) : 'no profile')  
  const projects = await medplum.searchResources('Project');
  return {
    count: projects.length, 
    name: projects?.[0]?.name, 
    superAdmin: medplum.isSuperAdmin(), 
    projectAdmin: medplum.isProjectAdmin()
  }
}
```

## Expected Behavior
Returned `count` should be > 1

## Observed Behavior
Returned `count` is 1, referring only to the current project. 

This is true regardless of whether or not `runAsUser` is set

codyebberson:
The issue is that the "is super admin?" check looks as `Login.superAdmin`, which is assigned when creating the `Login`.  The bot `$execute` operation has a separate code path for creating the `Login` which never set `superAdmin`.

This PR deprecates `Login.superAdmin` and replaces all uses with `Project.superAdmin`.

--------

Original:

This PR follows the existing logic from `src/oauth/token.ts` which copies `superAdmin` from `Project` to `Login`. This is the most surgical fix to the problem.

Arguably it might be better to remove `Login.superAdmin` and instead only rely on `Project.superAdmin`?  We originally added `Login.superAdmin` a long long time ago, before we had Redis caching, and before we loaded `Project` on every request.  It was definitely a case of premature optimization, and wanting to avoid unnecessarily loading the `Project`.


codyebberson:
I got curious, and quick scoped how hard it would be to remove `Login.superAdmin`, and it will be fast.

Updating PR

ThatOneBro (approved):

PR #4286, opened 2024-03-30
fix(generator): rm `base` from SDs to compress min SDs for `MockClient`

### LINKED ISSUE #4287: Make sure we infer `contentReference.base` properly when missing
Example case: 

`Observation.component.referenceRange` should be of type `ObservationReferenceRange` but if you read the `StructureDefinition` and `base` was missing, it looks for DataType `ObservationComponentReferenceRange` which does not exist.

ThatOneBro:
We can remove `base` property when it matches the `min`, `max`, and `path` of the `ElementDefinition` to compress space for the minimal `MockClient` version.

It seems like we may need to fix how `contentReference` is handled in some cases though, as some tests are failing after removing `base` when resolving `contentReference` properly seems like it would have prevented the failure. See:  #4287

Closes #4287


packages/core/src/typeschema/types.ts
```diff
@@ -373,7 +373,7 @@ class StructureDefinitionParser {
         } while (this.backboneContext && !pathsCompatible(this.backboneContext.path, element?.path));
       } else {
         this.innerTypes.push(this.backboneContext.type);
-        delete this.backboneContext;
+        this.backboneContext = undefined;
```
codyebberson, line 376: Out of curiosity, was this required for something, or more of a stylistic choice?

codyebberson (approved):

packages/core/src/typeschema/types.ts
ThatOneBro: Screenshot 2024-04-01 at 9 37 45 AM
Not really stylistic, `delete` is super slow because it causes changes in `HiddenClass` structure

PR #4290, opened 2024-04-01
Remove build.sh

codyebberson:
The build step used to be much bigger and more complicated.  Over time, we have simplified it down to just `npm run build`.  This PR is the final step of killing `build.sh` entirely.


mattlong (approved):

.github/workflows/build.yml
```diff
@@ -53,8 +54,10 @@ jobs:
           echo "TURBO_TOKEN=${{ secrets.TURBO_TOKEN }}" >> $GITHUB_ENV
           echo "TURBO_TEAM=${{ secrets.TURBO_TEAM }}" >> $GITHUB_ENV
           echo "TURBO_REMOTE_ONLY=${{ secrets.TURBO_REMOTE_ONLY }}" >> $GITHUB_ENV
+      - name: Install
+        run: npm ci --maxsockets 1
```
ThatOneBro, line 58: Can I suggest also keep stats, just in case:
```suggestion
        run: npm ci --maxsockets 1
      - name: Log npm stats
        run: |
          echo "node version: $(node --version)"
          echo "npm version: $(npm --version)"
```

ThatOneBro (approved):

.github/workflows/build.yml
```diff
@@ -53,8 +54,10 @@ jobs:
           echo "TURBO_TOKEN=${{ secrets.TURBO_TOKEN }}" >> $GITHUB_ENV
           echo "TURBO_TEAM=${{ secrets.TURBO_TEAM }}" >> $GITHUB_ENV
           echo "TURBO_REMOTE_ONLY=${{ secrets.TURBO_REMOTE_ONLY }}" >> $GITHUB_ENV
+      - name: Install
```
ThatOneBro, line 57: Can I suggest we keep stats, just in case?

```suggestion
      - name: Log npm stats
        run: |
          echo "node version: $(node --version)"
          echo "npm version: $(npm --version)"
      - name: Install
```

PR #4294, opened 2024-04-01
Drop unused columns from lookup tables

codyebberson:
This is a continuation of https://github.com/medplum/medplum/pull/4222 and https://github.com/medplum/medplum/pull/4202

Some of our FHIR search lookup tables are *very* old - notably `Address` and `HumanName` were authored during the earliest prototypes.  And some of those lookup tables include columns and ideas that were never realized, which are now dead weight.

In https://github.com/medplum/medplum/pull/4222 we discontinued all use of the columns.  They can now be deleted.

Here are the columns that will be deleted:

```ts
export async function run(client: PoolClient): Promise<void> {
  await client.query('ALTER TABLE "Address" DROP COLUMN "id"');
  await client.query('ALTER TABLE "Address" DROP COLUMN "index"');
  await client.query('ALTER TABLE "Address" DROP COLUMN "content"');

  await client.query('ALTER TABLE "HumanName" DROP COLUMN "id"');
  await client.query('ALTER TABLE "HumanName" DROP COLUMN "index"');
  await client.query('ALTER TABLE "HumanName" DROP COLUMN "content"');

  await client.query('ALTER TABLE "ValueSetElement" DROP COLUMN "id"');

  for (const resourceType of resourceTypes) {
    await client.query(`ALTER TABLE "${resourceType}_Token" DROP COLUMN "index"`);
  }
}
```


packages/server/src/fhir/operations/expunge.test.ts
```diff
@@ -193,6 +193,9 @@ async function existsInDatabase(tableName: string, id: string | undefined): Prom
 }
 
 async function existsInLookupTable(tableName: string, id: string | undefined): Promise<boolean> {
-  const rows = await new SelectQuery(tableName).column('id').where('resourceId', '=', id).execute(getDatabasePool());
+  const rows = await new SelectQuery(tableName)
+    .column('resourceId')
+    .where('resourceId', '=', id)
+    .execute(getDatabasePool());
```
mattwiller, line 199: This is unsafe to deploy in the same PR as the DB migration: the migration will occur with old instances still running and potentially referencing the deleted column

codyebberson: This is just a test.  The test was verifying that the rows were actually gone from the db.

mattwiller (approved):

codyebberson:
Deploying to staging...

codyebberson:
Took ~600ms on staging

image

PR #4297, opened 2024-04-02
Added sortStringArray utility function

codyebberson:
Prep work for https://github.com/medplum/medplum/pull/4296

Failed on sonar warning:

image

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1058,7 +1058,10 @@ export function getWebSocketUrl(path: string, baseUrl: URL | string): string {
 }
 
 /**
- * Sorts an array of strings using the localeCompare method.
+ * Sorts an array of strings in place using the localeCompare method.
+ *
+ * This method will mutate the input array.
+ *
  * @param array - The array of strings to sort.
  * @returns The sorted array of strings.
  */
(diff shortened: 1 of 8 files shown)


packages/core/src/utils.ts
```diff
@@ -1056,3 +1056,12 @@ export function append<T>(array: T[] | undefined, value: T): T[] {
 export function getWebSocketUrl(path: string, baseUrl: URL | string): string {
   return new URL(path, baseUrl).toString().replace('http://', 'ws://').replace('https://', 'wss://');
 }
+
+/**
+ * Sorts an array of strings using the localeCompare method.
```
ThatOneBro, line 1061: Can we also note that it mutates and returns the input array?

packages/server/src/fhir/patient.ts
```diff
@@ -72,9 +72,7 @@ export function getPatients(resource: Resource): Reference<Patient>[] {
       }
     }
   }
-  return Array.from(result)
-    .sort((a, b) => a.localeCompare(b))
-    .map((reference) => ({ reference }));
+  return Array.from(result).map((reference) => ({ reference }));
```
ThatOneBro, line 75: Is it intentional that the sort was removed? 👀

codyebberson: There's no logical reason for this method to return sorted results.  It's sorting on `reference`, which is effectively sorting on v4 UUIDs, which is kinda arbitrary.

It does look like the tests assumed sorted results.... I will look into whether we need to update those.

packages/core/src/utils.ts
codyebberson: ✅  comments updated

packages/server/src/fhir/patient.ts
codyebberson: ✅  tests updated

ThatOneBro (approved):

PR #4301, opened 2024-04-02
Storybook Themes

rahul1:
Adding a new theme


packages/react/.storybook/main.ts
```diff
@@ -16,6 +16,7 @@ const config: StorybookConfig = {
         },
       },
     },
+    'storybook-addon-mantine',
```
ThatOneBro, line 19: Looks like this might be missing from `package.json`

packages/react/.storybook/preview.tsx
```diff
@@ -43,15 +44,157 @@ const theme: MantineThemeOverride = {
...
+      h2: { fontSize: '24px', lineHeight: '1.35' },
+      h3: { fontSize: '20px', lineHeight: '1.3' },
+      h4: { fontSize: '18px', lineHeight: '1.25' },
+      h5: { fontSize: '16px', lineHeight: '1.2' },
+      h6: { fontSize: '14px', lineHeight: '1.15' },
+    },
+  },
+});
+
+const bonFoo = createTheme({
```
ThatOneBro, line 113: Is this used somewhere?

ThatOneBro (changes requested):
Cool, mostly looks good. Just a few cleanup comments

Does this mean there's a way to import a different theme on a per story basis?

PR #4307, opened 2024-04-03
Implement conditional update

mattwiller:
Implementing FHIR's [conditional update](http://hl7.org/fhir/R4/http.html#cond-update) functionality, which allows clients to upsert a resource based on search criteria


packages/core/src/client.ts
```diff
@@ -1969,6 +1970,31 @@ export class MedplumClient extends EventTarget {
...
 
+  /**
+   * Upsert a resource: update it in place if it exists, otherwise create it.  This is done in a single, transactional
+   * request to guarantee data consistency.
+   * @param resource - The resource to update or create.
+   * @param query - A FHIR search query to uniquely identify the resource if it already exists.
+   * @param options  - Optional fetch options.
+   * @returns The updated/created resource.
+   */
+  async upsert<T extends Resource>(resource: T, query: string, options?: MedplumRequestOptions): Promise<T> {
```
codyebberson, line 1981: @rahul1  - thoughts on function name?

At minimum, I think we should use `upsertResource()` to follow the pattern of `createResource()`, `updateResource()`, etc, and to give us room in the future for other potential upsert methods (??).

I'd also be ok to reuse the existing `createResourceIfNoneExist()` for smoother migration.  Not sure if the API contract is similar enough to reuse though.

packages/fhir-router/src/repo.ts
```diff
@@ -92,6 +94,19 @@ export interface FhirRepository<TClient = unknown> {
...
 
+  /**
+   * Creates or updates a resource based on a search criteria.
+   *
+   * See: https://www.hl7.org/fhir/http.html#cond-update
+   * @param resource - The FHIR resource to create or update.
+   * @param search - The search criteria to find an existing resource to update.
+   * @returns The updated resource.
+   */
+  conditionalUpdate<T extends Resource>(
```
codyebberson, line 105: Hm, all resource-oriented operations currently include `Resource` in the function name.

`conditionallyUpdateResource` is probably the most technically accurate but long and an eyesore.

Objections to `upsertResource()`, which would then align with the name in `MedplumClient`?

reshmakh:
Add log line

packages/core/src/client.ts
```diff
@@ -1969,6 +1969,31 @@ export class MedplumClient extends EventTarget {
...
 
+  /**
+   * Upsert a resource: update it in place if it exists, otherwise create it.  This is done in a single, transactional
+   * request to guarantee data consistency.
+   * @param resource - The resource to update or create.
+   * @param query - A FHIR search query to uniquely identify the resource if it already exists.
+   * @param options  - Optional fetch options.
+   * @returns The updated/created resource.
+   */
+  async upsertResource<T extends Resource>(resource: T, query: string, options?: MedplumRequestOptions): Promise<T> {
```
rahul1, line 1980: @mattwiller , for the `query` param, can we support the `query: QueryTypes`, similar to what we use in `searchResources`. 

Being able to pass a dictionary in can be very handy

codyebberson: This is a good call.  See `searchResources`.  You can then use `fhirSearchUrl` to generate the URL as well.

```ts
    const url = this.fhirSearchUrl(resourceType, query);
```

codyebberson (approved):
👍 🎉

PR #4317, opened 2024-04-04
fix(core): account for `baseUrl`s containing subpaths

### LINKED ISSUE #4298: Failing WebSocket connection in medplum-live-chat-demo
Issue: when running medplum-live-chat-demo the WebSocket connection fails due to calling the wrong URL.

Reproduction: given the following baseUrl for MedplumClient: `https://foo.bar/one/two/` the resulting WebSocket URL is `wss://foo.bar/ws/subscriptions-r4`, when it should be `wss://foo.bar/one/two/ws/subscriptions-r4`.

Suggested fix: in `medplum/packages/core/src/client.ts`, in line `this.subscriptionManager = new SubscriptionManager(this, getWebSocketUrl('/ws/subscriptions-r4', this.baseUrl))` the path string `'/ws/subscriptions-r4'` should not begin with `'/'`, instead it should be just `'ws/subscriptions-r4'`.

ThatOneBro:
Fixes #4298

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -74,7 +74,9 @@ import {
   CodeChallengeMethod,
   ProfileResource,
   arrayBufferToBase64,
+  concatUrls,
   createReference,
+  ensureTrailingSlash,
   getReferenceString,
   getWebSocketUrl,
   isObject,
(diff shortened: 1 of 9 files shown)


packages/core/src/client.ts
```diff
@@ -2465,7 +2467,7 @@ export class MedplumClient extends EventTarget {
    * @returns The FHIR batch/transaction response bundle.
    */
   executeBatch(bundle: Bundle, options?: MedplumRequestOptions): Promise<Bundle> {
-    return this.post(this.fhirBaseUrl.slice(0, -1), bundle, undefined, options);
+    return this.post(this.fhirBaseUrl, bundle, undefined, options);
```
codyebberson, line 2470: Cool, I verified that both with and without trailing slashes work as expected

```bash
curl 'https://api.medplum.com/fhir/R4' ...
curl 'https://api.medplum.com/fhir/R4/' ...
```

codyebberson (approved):
Nice, this turned out clean

PR #4318, opened 2024-04-04
Fixes #4314 - jsdom version 24

### LINKED ISSUE #4314: Investigate upgrading to latest `jsdom` version
We're on `20` and `24` was released in January. Maybe worth seeing what new features / perf improvements / bug fixes there are.

We just need to make `jsdom` version explicit in `package.json` if we want to upgrade.

https://www.npmjs.com/package/jsdom
https://github.com/jsdom/jsdom/releases

codyebberson:
Not totally sure if this is a good idea, because Jest is using a bunch of the same dependencies but on different versions, so we're dual installing a bunch of stuff:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "identity-obj-proxy": "3.0.0",
         "jest": "29.7.0",
         "jest-environment-jsdom": "29.7.0",
+        "jsdom": "24.0.0",
         "npm-check-updates": "16.14.18",
         "nyc": "15.1.0",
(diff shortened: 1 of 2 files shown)


ThatOneBro:
> because Jest is using a bunch of the same dependencies but on different versions, so we're dual installing a bunch of stuff

Ouch, well still maybe worth it?

I'm okay with it if v24 of `jsdom` is supposed to be more correct

ThatOneBro (approved):

PR #4322, opened 2024-04-05
remove the max-width of form

aniketkumar7:
In reference to this: 

I believe this should be a matter of removing the unnecessary style={{ maxWidth: 400 }} in a few places:

packages/react/src/auth/AuthenticationForm.tsx line 163
packages/app/src/ChangePasswordPage.tsx line 15
packages/app/src/ResetPasswordPage.tsx line 35
packages/app/src/SetPasswordPage.tsx line 28

I have removed the max-width of the form which will disbalance the UI of the form.


codyebberson (approved):
Thanks @aniketkumar7 !

PR #4324, opened 2024-04-05
Fixes #4323 - add bot to BotEvent

### LINKED ISSUE #4323: Access "Bot" resource inside Bots
User Driven: https://medplum.slack.com/archives/C05GD70FF55/p1712328064737169

Users want to be able to access the `Bot` resource of the current bot that is executing. Specifically to read the identifiers on the Bot. 

Possible implementations: 

- Patch the `this` parameter of `handler`
- Pass in a reference to the Bot in the `BotEvent` struct
- modify the behavior of `getProfileAsync`? (probably not recommended)

codyebberson:


rahul1:
I would suggest making this a reference to the `Bot`, rather than the Bot resource. Will make it easier to enforce access policies (e.g. the `Bot` has an access policy that restricts access to `Bots`)

ThatOneBro (approved):

PR #4328, opened 2024-04-06
fix(agent): check for `ping` command, report transmit errors back on `Agent/$push`

### LINKED ISSUE #4007: Agent Ping $push fails when `ping` is not installed
Running the Agent in the `debain:bullseye-slim` image, `ping` is not installed by default. It would be good to check if `ping` is installed before issuing the `$push` - https://github.com/medplum/medplum/blob/v3.0.3/packages/agent/src/app.ts#L284

This leads to an inconsistent error message on the UI
```
Error
"$push" operation timed out. Agent may be unreachable
```

Though the Agent is connected and received the event. Below is the Agent log when trying to ping the docker gateway.
```
{"level":"INFO","timestamp":"2024-02-22T14:05:53.826Z","msg":"Medplum service started successfully"}
{"level":"ERROR","timestamp":"2024-02-22T14:09:48.075Z","msg":"Error during ping attempt to 172.17.0.1: Command failed: ping -c 4 172.17.0.1\n/bin/sh: 1: ping: not found\n"}
{"level":"ERROR","timestamp":"2024-02-22T14:09:48.076Z","msg":"Unknown message type: agent:error"}
```

### LINKED ISSUE #4130: Errors occurring during pinging cause `Agent/$push` operation to timeout

### LINKED ISSUE #4331: Add `count` as parameter to Agent `ping` feature in UI

ThatOneBro:
* Fixes #4007
* Fixes #4130
* Fixes #4331

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -348,7 +348,7 @@ export class App {
         contentType: ContentType.PING,
         remote: message.remote,
         callback: message.callback,
-        status: 'ok',
+        statusCode: 200,
         body: result,
       } satisfies AgentTransmitResponse);
     } catch (err) {
@@ -359,7 +359,7 @@ export class App {
         contentType: ContentType.TEXT,
         remote: message.remote,
         callback: message.callback,
-        status: 'error',
+        statusCode: 500,
         body: normalizeErrorString(err),
       } satisfies AgentTransmitResponse);
     }
@@ -400,7 +400,7 @@ export class App {
           remote: message.remote,
           callback: message.callback,
           contentType: ContentType.HL7_V2,
-          status: 'ok',
+          statusCode: 200,
           body: response.toString(),
         } satisfies AgentTransmitResponse);
       })
@@ -412,7 +412,7 @@ export class App {
           remote: message.remote,
           callback: message.callback,
           contentType: ContentType.TEXT,
-          status: 'ok',
+          statusCode: 500,
           body: normalizeErrorString(err),
         } satisfies AgentTransmitResponse);
       })
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -282,9 +282,9 @@ export class App {
       case 'win32':
         return 'ping';
       case 'linux':
-        return 'ping --help';
+        return 'ping -h';
       default:
-        return 'ping --help';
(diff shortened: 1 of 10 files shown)


packages/agent/src/app.ts
```diff
@@ -12,16 +11,27 @@ import {
...
-import { promisify } from 'node:util';
-import { platform } from 'os';
+import { platform } from 'node:os';
 import WebSocket from 'ws';
 import { Channel } from './channel';
 import { AgentDicomChannel } from './dicom';
 import { AgentHl7Channel } from './hl7';
 
-const exec = promisify(_exec);
+async function execAsync(command: string, options: ExecOptions): Promise<{ stdout: string; stderr: string }> {
```
codyebberson, line 22: Out of curiosity, what does this do differently than `promisify()`?

packages/agent/src/app.ts
```diff
@@ -265,10 +276,48 @@ export class App {
...
+      this.pingUtilAvailable = false;
+    }
+    return this.pingUtilAvailable;
+  }
+
   private async tryPingIp(message: AgentTransmitRequest): Promise<void> {
     try {
-      if (message.body && message.body !== 'PING') {
-        const warnMsg = 'Message body present but unused. Body should be empty for a ping request.';
+      if (!(await this.isPingUtilAvailable())) {
```
codyebberson, line 313: What are the advantages of using this, vs just trying the ping and reporting the error?

I think https://github.com/medplum/medplum/issues/4007 would still be fixed by this PR, even without this check, just by virtue of the better error handling, yeah?

packages/agent/src/app.ts
ThatOneBro: `node` uses a special promisify for a lot of the node core modules, making it hard to mock. The easiest way to mock it would be to mock `exec` directly, but the special internal version doesn't directly wrap `exec` 😬 so it becomes difficult to test without mocking the promisify function too and just is really unclear. This is much simpler than what I tried to do originally with mocking 😅

packages/agent/src/app.ts
ThatOneBro: Good point, removed those functions as they're kind of redundant

codyebberson (approved):
🔥

PR #4332, opened 2024-04-07
fix(client): make `QueryTypes` more accurate

### LINKED ISSUE #4325: Make `QueryTypes` more accurate (no `any`)

ThatOneBro:
Restricts values in objects passed in as `QueryTypes` to `string`, `number`, and `boolean` types, from previous allowance of `any`. Fixes #4325

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -22,6 +22,18 @@ import { formatCodeableConcept, formatHumanName } from './format';
 import { OperationOutcomeError, validationError } from './outcomes';
 import { isReference } from './types';
 
+/**
+ * QueryTypes defines the different ways to specify FHIR search parameters.
+ *
+ * Can be any valid input to the URLSearchParams() constructor.
+ *
+ * TypeScript definitions for URLSearchParams do not match runtime behavior.
+ * The official spec only accepts string values.
+ * Web browsers and Node.js automatically coerce values to strings.
+ * See: https://github.com/microsoft/TypeScript/issues/32951
+ */
+export type QueryTypes = URLSearchParams | string[][] | Record<string, string | number | boolean> | string | undefined;
+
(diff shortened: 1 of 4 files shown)


packages/core/src/utils.ts
```diff
@@ -1096,6 +1113,29 @@ export function concatUrls(baseUrl: string | URL, path: string): string {
...
+
+/**
+ * Converts the given `query` to a string.
+ *
+ * @param query - The query to convert. The type can be any member of `QueryTypes`.
+ * @returns The query as a string.
+ */
+export function getQueryString(query: QueryTypes): string {
+  if (typeof query === 'object' && !Array.isArray(query) && !(query instanceof URLSearchParams)) {
+    query = Object.fromEntries(Object.entries(query).filter((entry) => entry[1] !== undefined));
```
codyebberson, line 1135: I'm trying to think if this could cause any behavioral differences

In the realm of URLs, a key without equals sign is different than key with equals sign and empty value.

i.e., `example.com?foo` is not the same as `example.com?foo=`

I don't know if there are any cases in FHIR where that distinction matters.... 🤔 

The FHIR spec specifically says " the parameters are a series of name=[value] pairs" (https://hl7.org/fhir/R4/search.html#Introduction) so this is probably ok.

codyebberson (approved):

packages/core/src/utils.ts
ThatOneBro: Not sure what you mean here, this is filtering out `undefined`s in the map that would cause otherwise unintended behavior, specifically when we had some cases of `{ hello: foo?.bar }` which would result in: `hello=undefined` in the query string

ThatOneBro: But I see your point, its not really possible for a user to input `?foo=` with this function... maybe a feature and not a bug?

PR #4334, opened 2024-04-08
perf(seed): run rebuilds in parallel, add perf logs

ThatOneBro:
# TODO 
- [ ] Ensure low load on when rebuilding (test on staging)


packages/server/src/seeds/searchparameters.ts
```diff
@@ -14,11 +14,13 @@ export async function rebuildR4SearchParameters(): Promise<void> {
 
   const systemRepo = getSystemRepo();
 
+  const promises = [];
   for (const filename of SEARCH_PARAMETER_BUNDLE_FILES) {
     for (const entry of readJson(filename).entry as BundleEntry[]) {
-      await createParameter(systemRepo, entry.resource as SearchParameter);
+      promises.push(createParameter(systemRepo, entry.resource as SearchParameter));
```
codyebberson, line 20: lol, curious how Postgres handles this

codyebberson (approved):
Nice.  It'll make CPU's heat up, but definitely worth the experiment.

packages/server/src/seeds/searchparameters.ts
ThatOneBro: We're capped at 10 concurrent connections with the pool, I considered trying to up it and see what happens, but seemed to run fine on my local machine as is.

packages/server/package.json
```diff
@@ -18,7 +18,9 @@
     "clean": "rimraf dist",
     "dev": "ts-node-dev --poll --respawn --transpile-only --require ./src/otel/instrumentation.ts src/index.ts",
     "start": "node --require ./dist/otel/instrumentation.js dist/index.js",
-    "test": "jest"
+    "test:seed:serial": "jest seed-serial.test.ts --config jest.seed.config.ts --coverageDirectory \"<rootDir>/coverage/seed/serial\"",
+    "test:seed:parallel": "jest seed.test.ts --config jest.seed.config.ts --coverageDirectory \"<rootDir>/coverage/seed/parallel\"",
+    "test": "npm run test:seed:parallel && jest"
```
codyebberson, line 23: I understand this, but I don't love it.

One of my main testing patterns is to run individual tests using `npm t -- src/file.ts -t 'test name'` - that will be significantly slower, because it will always run `seed.test.ts` first, right?

Given that the test seed is a rare event, I'd rather just add it to the developer setup docs.

packages/server/seed-tests/seed.test.ts
codyebberson, line 1: Why move this file?  Strong preference for keeping all `.ts` files in `src/`

packages/server/src/seeds/searchparameters.ts
```diff
@@ -4,19 +4,32 @@ import { getDatabasePool } from '../database';
...
-    for (const entry of readJson(filename).entry as BundleEntry[]) {
-      await createParameter(systemRepo, entry.resource as SearchParameter);
+  if (finalOptions.parallel) {
+    const promises = [];
+    for (const filename of SEARCH_PARAMETER_BUNDLE_FILES) {
+      for (const entry of readJson(filename).entry as BundleEntry[]) {
+        promises.push(createParameter(systemRepo, entry.resource as SearchParameter));
+      }
+    }
+    await Promise.all(promises);
```
codyebberson, line 27: Consider always building the array, and then using the `parallel` option in the last step

```ts
if (finalOptions.parallel) {
  await Promise.all(promises);
} else {
  for (const promise of promises) {
    await promise;
  }
}
```

That way we don't need to worry about the logic getting out of sync.

packages/server/src/seeds/searchparameters.ts
```diff
@@ -4,19 +4,32 @@ import { getDatabasePool } from '../database';
...
 import { r4ProjectId } from '../seed';
+import { RebuildOptions, buildRebuildOptions } from './common';
 
 /**
  * Creates all SearchParameter resources.
+ * @param options - Optional options for how rebuild should be done.
  */
-export async function rebuildR4SearchParameters(): Promise<void> {
+export async function rebuildR4SearchParameters(options?: Partial<RebuildOptions>): Promise<void> {
+  const finalOptions = buildRebuildOptions(options);
```
codyebberson, line 14: A long time ago, I was scolded in a code review for using the variable name "finalOption", because it sounded too close to "final solution", which has some horrific connotations. 🤷

packages/server/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src/**/*.ts"]
+  "include": ["src/**/*.ts", "seed-tests/**/*.ts"]
```
codyebberson, line 3: Yeah, prefer to just move those tests to `src/`

codyebberson (commented):
Thanks for doing this investigation.

How much time savings do we get from this?

I'd like to understand full picture. This increases the complexity of test setup (extra instance of postgres, double seed logic, etc).  I just want to be sure that complexity is justified.

packages/server/src/seeds/searchparameters.ts
ThatOneBro: Open to suggestions to indicate the finality of the options without using this verbiage

packages/server/src/seeds/searchparameters.ts
ThatOneBro: Good point 👍

packages/server/seed-tests/seed.test.ts
ThatOneBro: It was initially due to weird glob matching not excluding the seed tests properly from main line of tests... it was cleaner to just separate them since they are somewhat logically separate...

I put them back though, found a sort of hack around it with some `**` paths that works for some reason. Made a note about it in the file

packages/server/src/seeds/searchparameters.ts
ThatOneBro: Actually this would work. Filling the array in the serial case doesn't actually execute the promises serially, only waits for them serially. We have to create the promise in the loop where it is awaited in order to not unintentionally parallelize them

packages/server/src/seeds/searchparameters.ts
ThatOneBro: Went with `rebuildOptions`

packages/server/src/config.ts
```diff
@@ -261,7 +261,7 @@ function addDefaults(config: MedplumServerConfig): MedplumServerConfig {
   config.accurateCountThreshold = config.accurateCountThreshold ?? 1000000;
   config.defaultBotRuntimeVersion = config.defaultBotRuntimeVersion ?? 'awslambda';
   config.defaultProjectFeatures = config.defaultProjectFeatures ?? [];
-  config.emailProvider = config.emailProvider || (config.smtp ? 'smtp' : 'awsses');
+  config.emailProvider = config.emailProvider ?? (config.smtp ? 'smtp' : 'awsses');
```
mattlong, line 264: Would `||` actually be preferable here to guard against `config.emailProvider === ''`?

packages/server/src/seeds/searchparameters.ts
mattlong: I don't know if it's worth introducing a new dependency, but [p-limit](https://github.com/sindresorhus/p-limit) seems like a pretty elegant and simple solution to abstract away the max concurrency that we want to allow and avoid the `if/else` code paths.

mattlong (approved):
LGTM. Just a couple comments to consider.

PR #4340, opened 2024-04-08
fix: registration by not passing the `new` `projectId`

joennlae:
First of all really fantastic project, an absolute banger 🚀 

When trying to register on [app.medplum.com](https://app.medplum.com/register)

I get the following error.

Screenshot 2024-04-08 at 18 40 22

I then decided to install it locally and got the same one. I do believe that this error was introduced with this PR: [#4321](https://github.com/medplum/medplum/pull/4321)

This PR should prevent the propagation of `new` to `createUser`.

### DIFF
--- a/packages/server/src/auth/newuser.test.ts
+++ b/packages/server/src/auth/newuser.test.ts
@@ -50,6 +50,7 @@ describe('New user', () => {
         email: `alex${randomUUID()}@example.com`,
         password: 'password!@#',
         recaptchaToken: 'xyz',
+        projectId: 'new',
       });
 
     expect(res.status).toBe(200);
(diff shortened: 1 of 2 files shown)


codyebberson:
Eek, thanks @joennlae  - probably a regression due to https://github.com/medplum/medplum/pull/4321

Looking now 👀

codyebberson:
Thanks @joennlae  - I made a slight tweak and added a test.

We try to avoid mutating `req.body` for logging and analytics purposes.  This pushes the check for `projectId='new'` a little later.

Thank you so much for your contribution and your kind words! 🙏

joennlae:
> Thanks @joennlae - I made a slight tweak and added a test.
> 
> We try to avoid mutating `req.body` for logging and analytics purposes. This pushes the check for `projectId='new'` a little later.
> 
> Thank you so much for your contribution and your kind words! 🙏
 
Yeah, all good :-)

I was pretty confident that this proposal was not the final solution, as I was not able to grasp all of the code in this short time :-) But I will definitely invest more time into it :-)

PR #4351, opened 2024-04-10
Added server config setting for approved email senders

codyebberson:


ThatOneBro:
Can we get a test for unapproved sender too, please? 😄

ThatOneBro (changes requested):
Can we get a test for unapproved sender please? 😄

codyebberson:
> Can we get a test for unapproved sender please? 😄

I snuck it into an existing test, because that is effectively the default behavior.  Let me know if you'd like that to be more explicit:

image

ThatOneBro (approved):

PR #4352, opened 2024-04-10
Fixes #4343 - ONC service base url

### LINKED ISSUE #4343: Medplum's capability statement not matching what's expected by ONC
ONC has a system called "lantern" that tracks the status of (g)(10) implementations for CHPL listings.  Ours needs to pass and it relates to our capability statement. 

This is the Inferno tool: https://inferno.healthit.gov/test-kits/service-base-url/

codyebberson:


codyebberson:
Fixed in https://github.com/medplum/medplum/pull/4869

PR #4358, opened 2024-04-11
Implement full batch and transaction support

### LINKED ISSUE #4234: `createResourceIfNoneExist` uniqueness guarantees
The current implementation of [`createResourceIfNoneExist`](https://github.com/medplum/medplum/blob/ee642056ebcc58a850022d47d6c8b08c9e518cf1/packages/core/src/client.ts#L1964-L1971) is subject to a race condition that could unexpectedly result in duplicate records. If it is called concurrently with the same `query` , `searchOne` may return `undefined` for multiple concurrent calls and subsequently call `createResource` resulting in duplicates.

A `Bundle#transaction` may not inherently enforce uniqueness, depending on the locking mechanism of the implementation https://github.com/medplum/medplum/pull/4358#issuecomment-2052519584.

mattwiller:
Resolves #4234


codyebberson (commented):
Structurally, this looks great to me.

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -143,31 +183,43 @@ async function patchResource(req: FhirRequest, repo: FhirRepository): Promise<Fh
   return [allOk, resource];
 }
 
+export type RestInteraction =
```
codyebberson, line 186: Low pri mental note: we may want to try to merge this with `packages/server/src/util/auditevent.ts`, either here in `fhir-router` or in `core`.

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -143,31 +183,43 @@ async function patchResource(req: FhirRequest, repo: FhirRepository): Promise<Fh
...
+  | CapabilityStatementRestInteraction['code']
```
codyebberson, line 187: Low pri mental note: This is another case where having a proper `type` for a ValueSet-derived string union would have been helpful, so you could use `CapabilityStatementRestInteractionCode` rather than `CapabilityStatementRestInteraction['code']` -- I'm pretty sure the former would be faster for tsc, and probably more discoverable.

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -143,31 +183,43 @@ async function patchResource(req: FhirRequest, repo: FhirRepository): Promise<Fh
...
-    this.router.add('POST', ':resourceType', createResource);
-    this.router.add('GET', ':resourceType/:id', readResourceById);
-    this.router.add('GET', ':resourceType/:id/_history', readHistory);
-    this.router.add('GET', ':resourceType/:id/_history/:vid', readVersion);
-    this.router.add('PUT', ':resourceType', conditionalUpdate);
-    this.router.add('PUT', ':resourceType/:id', updateResource);
-    this.router.add('DELETE', ':resourceType/:id', deleteResource);
-    this.router.add('PATCH', ':resourceType/:id', patchResource);
-    this.router.add('POST', '$graphql', graphqlHandler);
+    this.router.add('GET', '', searchMultipleTypes, { interaction: 'search-system' });
```
codyebberson, line 202: This is definitely starting to paint a clearer picture for the future.

Something like:
* New outer type such as `FhirProcessor` which wraps `FhirRouter` and `FhirRepository`
* `registerHandler(handler: OperationHandler)` which adds routes and whatnot

I'm not sure if FHIR has a good term that encompasses both `interaction` and `operation` ?

dillonstreator:
This is exciting to see taking shape!

I have a few questions/concerns to better understand and expectation set.

- How does [`ifNotExists`](https://hl7.org/fhir/R4/bundle-definitions.html#Bundle.entry.request.ifNoneExist) tie into the uniqueness guarantees of resource creation within a transaction? https://github.com/medplum/medplum/issues/1369#issuecomment-1904610226
	- Is implicit table locking the mechanism by which uniqueness is guaranteed?
- How will implicit table locking impact the multi-tenant nature of the database https://arc.net/l/quote/bochihxa?
	- Is this a 'noisy neighbor' or worse yet a 'DOS' in the making with large/long-running `Bundle#transaction`'s?
	- Potential for deadlocks? https://arc.net/l/quote/ryjtghvc

mattwiller:
@dillonstreator Great questions!  At this point, strict uniqueness guarantees for conditional create aren't part of the implementation: it's a hard problem to solve perfectly, as you've pointed out with the concerns around table locking.  The [FHIR spec](http://hl7.org/fhir/http.html#ccreate) is somewhat vague about exactly what the guarantees for this functionality should be, but it's something I'm planning to explore as a fast follow after the initial implementation is complete.  In any case, I don't think that we're likely to go with a heavyweight solution like whole table locking — for exactly the scale/performance concerns you mentioned.  I'll need to spend some time thinking through other options and the correctness/performance tradeoffs around them

dillonstreator:
Thank you for the transparency, @mattwiller. I appreciate the thoughtfulness being put into this and the commitment to addressing the uniqueness needs.
As a stop-gap, to enable the uniqueness guarantees that we need, we're looking at standing up a Redis instance to employ arbitrary locking.
On the Medplum side, this might be a good pattern to emulate table locking and could be namespaced on project id to prevent the noisy neighbor situation.

packages/fhir-router/src/batch.ts
```diff
@@ -99,86 +98,239 @@ class BatchProcessor {
...
-    if (!entry.request.url) {
-      throw new OperationOutcomeError(badRequest('Missing entry.request.url'));
+  private async performBatchOperation(entry: BundleEntry): Promise<FhirResponse> {
+    const urlParts = splitN(entry.request?.url as string, '?', 2);
+    const requestPath = urlParts[0];
+    const queryParams = urlParts[1];
+    const route = this.router.find(entry.request?.method as HttpMethod, requestPath);
+    const params = route?.params;
+
+    switch (route?.data?.interaction) {
```
codyebberson, line 257: Hm, I need to review this more, but I would have assumed that most of this `switch` statement would have been handled by `handleRequest` - if nothing else, a lot of these checks are duplicated, which would be nice to clean up if possible.

```ts
    const result = await this.router.handleRequest(
      {
        method: request.method as HttpMethod,
        pathname: url.pathname,
        params: Object.create(null),
        query: Object.fromEntries(url.searchParams),
        body,
      },
      this.repo
    );
```

packages/server/src/config.ts
```diff
@@ -98,6 +98,7 @@ export interface MedplumDatabaseConfig {
   password?: string;
   ssl?: MedplumDatabaseSslConfig;
   queryTimeout?: number;
+  maxInstanceConnections?: number;
```
codyebberson, line 101: We should probably give this a name that clearly indicates "database connections".  We've been trending toward longer and more explicit config setting names.

```suggestion
  maxDatabaseConnectionsPerServer?: number;
```

codyebberson (commented):
Overall this looks great.  Thanks for slogging through 🙏 

Added initial feedback.  I want to review more in-depth when I get some time.

One immediate thing we need to add is disabling "read from Redis" when in the middle of a transaction.  Otherwise things could get weird (I think?)

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -26,7 +32,16 @@ export type FhirRequest = {
 
 export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
 
-export type FhirRouteHandler = (req: FhirRequest, repo: FhirRepository, router: FhirRouter) => Promise<FhirResponse>;
+export type FhirRouteOptions = {
+  batch?: boolean;
+};
+
+export type FhirRouteHandler = (
```
codyebberson, line 39: I was thinking about this more, and I think we should have a team session on this, because it's going to be a pretty central concept throughout the entire code base.

This doesn't necessarily need to be part of this PR, just thinking out loud.

A couple thoughts:
1. Given that we're already at 4 args, I suspect we should just make this into a single argument "context" type
2. We may want to consider adding (moving?) a bunch of the `AuthenticatedRequestContext` properties to that "context" type
3. As you've pointed out in the past, `FhirResponse` should probably be an object rather than an array.

examples/medplum-chart-demo/src/bots/core/charting-utils.ts
```diff
@@ -277,7 +276,7 @@ export function createBundle(resources: (Condition | Observation | ClinicalImpre
   const entries: BundleEntry[] = resources.map((resource) => {
     const entry: BundleEntry = {
       fullUrl: `urn:uuid:${randomUUID()}`,
-      request: { method: 'PUT', url: getUpsertUrl(resource) },
+      request: { method: 'POST', url: resource.resourceType },
```
mattwiller, line 279: I removed the upsert from the demo in order to make it work with our conservative/defensive limits on `serializable` transactions.  If the intent is just to represent an all-or-nothing create of a set of resources that couldn't possibly exist any other way, the upsert isn't really necessary

packages/server/src/config.ts
```diff
@@ -112,6 +112,7 @@ export interface MedplumDatabaseConfig {
   password?: string;
   ssl?: MedplumDatabaseSslConfig;
   queryTimeout?: number;
+  maxConnectionsPerServer?: number;
```
codyebberson, line 115: This might be artifact of old PR?  This interface already contains `maxConnections`

packages/server/src/database.ts
```diff
@@ -47,7 +47,7 @@ async function initPool(config: MedplumDatabaseConfig, proxyEndpoint: string | u
     user: config.username,
     password: config.password,
     ssl: config.ssl,
-    max: config.maxConnections ?? 100,
+    max: config.maxConnectionsPerServer ?? 100,
```
codyebberson, line 50: I think revert this?

packages/server/src/fhir/references.ts
```diff
@@ -11,7 +11,7 @@ import {
 } from '@medplum/core';
 import { OperationOutcomeIssue, Reference, Resource } from '@medplum/fhirtypes';
 import { randomUUID } from 'crypto';
-import { getSystemRepo, Repository } from './repo';
+import { Repository, getSystemRepo } from './repo';
```
codyebberson, line 14: We need to align our import sorting config 😔 

Normally I ignore these, but given that this is the only change in the file, I'd recommend we revert.

packages/server/src/fhir/transaction.test.ts
```diff
@@ -255,6 +253,147 @@ describe('FHIR Repo Transactions', () => {
       });
     }));
 
+  test.skip('Nested transaction rollback from DB error', () =>
```
codyebberson, line 256: Remove skip?

packages/server/src/fhir/repo.ts
```diff
@@ -2151,6 +2106,166 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
+  }
+
+  /**
+   * Writes a cache entry to Redis.
+   * @param resource - The resource to cache.
+   */
+  private async setCacheEntry(resource: Resource): Promise<void> {
+    // No cache access allowed mid-transaction
+    if (this.transactionDepth) {
+      return;
```
codyebberson, line 2187: Should this add `() => setCacheEntry(resource)` to the post commit array?

packages/server/src/fhir/repo.ts
```diff
@@ -2151,6 +2106,166 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
+
+  /**
+   * Deletes a cache entry from Redis.
+   * @param resourceType - The resource type.
+   * @param id - The resource ID.
+   */
+  private async deleteCacheEntry(resourceType: string, id: string): Promise<void> {
+    // No cache access allowed mid-transaction
+    if (this.transactionDepth) {
+      return;
```
codyebberson, line 2210: Should this add `deleteCacheEntry` to the post commit array?

packages/server/src/fhir/repo.ts
```diff
@@ -2151,6 +2106,166 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
+
+  /**
+   * Deletes cache entries from Redis.
+   * @param resourceType - The resource type.
+   * @param ids - The resource IDs.
+   */
+  private async deleteCacheEntries(resourceType: string, ids: string[]): Promise<void> {
+    // No cache access allowed mid-transaction
+    if (this.transactionDepth) {
+      return;
```
codyebberson, line 2224: Same as above

packages/server/src/oauth/userinfo.test.ts
```diff
@@ -48,7 +48,7 @@ describe('OAuth2 UserInfo', () => {
     expect(res3.body.email).toBe('admin@example.com');
   });
 
-  test('Get userinfo with phone', async () => {
+  test.only('Get userinfo with phone', async () => {
```
codyebberson, line 51: Remove only?

packages/fhir-router/src/batch.ts
```diff
@@ -22,47 +37,299 @@ import { HttpMethod } from './urlrouter';
  * @param router - The FHIR router.
  * @param repo - The FHIR repository.
  * @param bundle - The input bundle.
+ * @param req - The request for the batch.
  * @returns The bundle response.
  */
-export async function processBatch(router: FhirRouter, repo: FhirRepository, bundle: Bundle): Promise<Bundle> {
-  return new BatchProcessor(router, repo, bundle).processBatch();
+export async function processBatch(
```
codyebberson, line 43: Minor nit: we use the following param ordering throughout `fhir-router`:

```ts
(req: FhirRequest, repo: FhirRepository, router: FhirRouter)
```

Not saying it's the "best" ordering, but just rather that we've tried to be consistent up to this point

packages/fhir-router/src/batch.ts
```diff
@@ -22,47 +37,299 @@ import { HttpMethod } from './urlrouter';
...
+          // Conditional update requires strong (serializable) transaction to
+          // guarantee uniqueness of possibly-created resource
+          requiresStrongTransaction = true;
+        }
+        updates++;
+      } else if (interaction === 'delete' && entry.request?.url.includes('?')) {
+        // Conditional delete requires strong (serializable) transaction
+        requiresStrongTransaction = true;
+      }
+      bucketedEntries[interaction]?.push(i);
```
codyebberson, line 174: Can we be sure that `interaction` is valid here?

packages/fhir-router/src/batch.ts
```diff
@@ -114,105 +381,78 @@ class BatchProcessor {
...
-      this.router.log('warn', 'Conditional mutation in batch', {
-        method: request.method,
-        url: request.url,
-      });
-    }
-
-    let body = entry.resource;
-    if (request.method === 'PATCH') {
-      body = this.parsePatchBody(entry);
+    const [requestPath] = splitN(entry.request?.url as string, '?', 2);
```
codyebberson, line 384: We use this pattern multiple times.  1, URL manipulation always makes me nervous.  2, I wonder if we should handle this in `router.find()` or add a new `router.findByUrl()`

packages/docs/docs/self-hosting/config-settings.md
```diff
@@ -228,7 +228,8 @@ Example `DatabaseSecrets` value:
   "port": 5432,
   "host": "my-cluster.us-east-1.rds.amazonaws.com",
   "username": "clusteradmin",
-  "queryTimeout": "60000"
+  "queryTimeout": 60000,
+  "maxInstanceConnections": 50
```
codyebberson, line 232: I think this might be old PR artifact?

packages/fhir-router/src/batch.ts
```diff
@@ -114,105 +381,78 @@ class BatchProcessor {
...
-        method: request.method as HttpMethod,
-        pathname: url.pathname,
-        params: Object.create(null),
-        query: Object.fromEntries(url.searchParams),
-        body,
-      },
-      this.repo
-    );
+    const request = this.parseBatchRequest(entry, route?.params);
+    const [outcome, resource] = await route.handler(request, this.repo, this.router, { batch: true });
```
codyebberson, line 391: Why is `{ batch: true }` needed here?  Isn't that implicit in the handler?

packages/fhir-router/src/batch.ts
```diff
@@ -114,105 +381,78 @@ class BatchProcessor {
...
     }
 
     const body = JSON.parse(Buffer.from(patchResource.data, 'base64').toString('utf8'));
-    if (!body) {
-      throw new OperationOutcomeError(badRequest('Empty patch body'));
-    }
-
     if (!Array.isArray(body)) {
-      throw new OperationOutcomeError(badRequest('Patch body must be an array'));
+      throw new OperationOutcomeError(badRequest('Patch operation body must be an array'));
```
codyebberson, line 450: I don't think "patch" is technically an "operation"

packages/fhir-router/src/batch.ts
```diff
@@ -114,105 +381,78 @@ class BatchProcessor {
...
   private parsePatchBody(entry: BundleEntry): any {
     const patchResource = entry.resource;
-    if (!patchResource) {
-      throw new OperationOutcomeError(badRequest('Missing entry.resource'));
-    }
-
-    if (patchResource.resourceType !== 'Binary') {
-      throw new OperationOutcomeError(badRequest('Patch resource must be a Binary'));
+    if (patchResource?.resourceType !== 'Binary') {
+      throw new OperationOutcomeError(badRequest('Patch operation must include a Binary resource'));
```
codyebberson, line 442: I don't think "patch" is technically an "operation"

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -78,15 +95,44 @@ async function searchByPost(req: FhirRequest, repo: FhirRepository): Promise<Fhi
...
-async function createResource(req: FhirRequest, repo: FhirRepository): Promise<FhirResponse> {
+async function createResource(
+  req: FhirRequest,
+  repo: FhirRepository,
+  _router: FhirRouter,
+  options?: FhirRouteOptions
+): Promise<FhirResponse> {
   const { resourceType } = req.params;
   const resource = req.body as Resource;
+  const assignedId = Boolean(options?.batch);
```
codyebberson, line 106: Can we just call the option `assignedId` to make it more consistent?  I didn't really understand what `batch` would mean as an option.

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -78,15 +95,44 @@ async function searchByPost(req: FhirRequest, repo: FhirRepository): Promise<Fhi
...
+
+  if (req.headers?.['if-none-exist']) {
+    let ifNoneExist = req.headers['if-none-exist'];
+    if (Array.isArray(ifNoneExist)) {
```
codyebberson, line 110: I feel like we probably have at least one `unarrayify()` helper somewhere.

packages/server/src/fhir/sql.ts
```diff
@@ -396,17 +396,25 @@ export class SqlBuilder {
...
+export function normalizeDatabaseError(err: any): OperationOutcomeError {
+  if (err?.code === '23505') {
+    // Catch duplicate key errors and throw a 409 Conflict
+    // See https://github.com/brianc/node-postgres/issues/1602
+    // See https://www.postgresql.org/docs/10/errcodes-appendix.html
+    return new OperationOutcomeError(conflict(err.detail));
+  }
+  if (err?.code === '40001') {
+    // Catch transaction serialization errors and throw a 409 Conflict
+    return new OperationOutcomeError(conflict(err.message));
```
codyebberson, line 413: Should these 2 special cases be combined?  Not sure what the difference is between `err.message` and `err.detail`?

packages/definitions/dist/fhir/r4/fhir.schema.json
```diff
@@ -59503,7 +59503,8 @@
               "graphql-introspection",
               "terminology",
               "websocket-subscriptions",
-              "reference-lookups"
+              "reference-lookups",
+              "transactions"
```
codyebberson, line 59507: How would you feel about renaming this to `transaction-bundles`?

We will use `transactions` for conditional create, conditional update.  This feature flag is just for the `transaction` bundles.

packages/fhir-router/src/batch.test.ts
```diff
@@ -370,22 +418,27 @@ describe('Batch', () => {
   test.skip('Process batch create ifNoneExist invalid resource type', async () => {
```
codyebberson, line 418: Should this `skip` be removed now?

packages/fhir-router/src/batch.test.ts
```diff
@@ -1063,37 +1214,42 @@ describe('Batch', () => {
...
-      const bundle = await processBatch(router, repo, {
-        resourceType: 'Bundle',
-        type: 'transaction',
-        entry: [
-          {
-            fullUrl: 'urn:uuid:e95d01cf-60ae-43f7-a8fc-0500a8b045bb',
-            request: {
-              method: 'POST',
-              url: 'Patient',
+    test.skip('Transaction update after create', async () => {
```
codyebberson, line 1217: Remove `skip`?

packages/fhir-router/src/batch.test.ts
```diff
@@ -1109,65 +1265,70 @@ describe('Batch', () => {
...
-            resource: {
-              resourceType: 'Organization',
-              name: 'Texas',
-              identifier: [
-                {
-                  system: 'https://foomedical.org/organization',
-                  value: 'org:bus:1',
-                },
-              ],
+    test.skip('urn:uuid in PATCH', async () => {
```
codyebberson, line 1268: Remove `skip`?

packages/fhir-router/src/batch.ts
mattwiller: I'll add a guard above to ensure that it is

packages/server/src/fhir/sql.ts
mattwiller: It's mostly a choice of which has the more useful information: for error `23505`, `err.detail` contains the specific conflict key/value, but for error `40001` it would be a relatively arbitrary SQL statement from the transaction in conflict

packages/fhir-router/src/batch.test.ts
mattwiller: This test is no longer valid

packages/fhir-router/src/batch.test.ts
mattwiller: This test is no longer valid

packages/fhir-router/src/batch.test.ts
mattwiller: Test moved to server package

codyebberson (approved):

Issue #3888, opened 2024-02-05
Throw error if `PATCH` tries to change `resourceType` or `id`


codyebberson:
Consider a user submitting a `PATCH` request which attempts to change a resource `id`

Current behavior: it would be the equivalent of:
1. Read resource A
2. Write as resource B

It would still perform all security and permission checks, so no security concerns... just might lead to unexpected behavior.

The FHIR patch spec does not comment on this: https://hl7.org/fhir/r4/http.html#patch

The FHIR update spec says:

> The request body SHALL be a Resource of the named type with an id element that has an identical value to the [id] in the URL. If no id element is provided, or the id disagrees with the id in the URL, the server SHALL respond with an HTTP 400 Bad Request error code, and SHOULD provide an OperationOutcome identifying the issue.

Desired behavior: we should throw an error in this case.

Implementor tips:
* This should be fixed in `packages/server/src/fhir/repo.ts`
* In the `patchResource()` method
* We can copy the same checks from `updateResource()`:

```ts
  if (resource.resourceType !== resourceType) {
    throw new OperationOutcomeError(badRequest('Incorrect resource type'));
  }
  if (resource.id !== id) {
    throw new OperationOutcomeError(badRequest('Incorrect ID'));
  }
```

rachana-uniyal:
@zulipbot claim

rachana-uniyal:
Hi, @codyebberson I would like to work on this issue.

codyebberson:
Hi @rachana-uniyal  - all yours!  let me know if you have any questions

PR #4360, opened 2024-04-11
Throw an error when a patch attempts to modify resourceType or Id

gusIreland:
This PR addresses https://github.com/medplum/medplum/issues/3888

When a resource is PATCHed and the field attempted to be updated is either `id` or `resourceType`, an error is thrown

### DIFF
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -241,6 +241,63 @@ describe('FHIR Repo', () => {
       expect(patient2.name?.[1]?.family).toEqual('Jones');
     }));
 
+  test('Patch patient', () =>
+    withTestContext(async () => {
+      const patient = await systemRepo.createResource<Patient>({
+        resourceType: 'Patient',
+        name: [{ given: ['Suzy'], family: 'Smith' }],
+      });
+      const result = await systemRepo.patchResource('Patient', patient.id as string, [
+        {
+          op: 'add',
+          path: '/active',
+          value: true,
+        },
+      ]);
+
+      expect(result.id).toEqual(patient.id);
+      expect(result.meta?.versionId).not.toEqual(patient.meta?.versionId);
+      expect((result as Patient).active).toEqual(true);
+    }));
+
+  test('Patch patient updating ID field', () =>
+    withTestContext(async () => {
+      const patient1 = await systemRepo.createResource<Patient>({
+        resourceType: 'Patient',
+        name: [{ given: ['Suzy'], family: 'Smith' }],
+      });
+      try {
+        await systemRepo.patchResource('Patient', patient1.id as string, [
+          {
+            op: 'replace',
+            path: '/id',
+            value: randomUUID(),
+          },
+        ]);
+      } catch (err) {
+        expect((err as OperationOutcomeError).outcome).toMatchObject(badRequest('Incorrect Id'));
+      }
+    }));
+
+  test('Patch patient updating resourceType field', () =>
+    withTestContext(async () => {
+      const patient1 = await systemRepo.createResource<Patient>({
+        resourceType: 'Patient',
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/repo.test.ts
```diff
@@ -241,6 +241,63 @@ describe('FHIR Repo', () => {
...
+      });
+      const result = await systemRepo.patchResource('Patient', patient.id as string, [
+        {
+          op: 'add',
+          path: '/active',
+          value: true,
+        },
+      ]);
+
+      expect(result.id).toEqual(patient.id);
```
gusIreland, line 258: I added this spec to wrap my head around the `patchResorce` function - I copied the spec from throw-error-on-patch-issue-3888. I'm a little unclear as I thought the patch would modify the resource in place, instead of having to use and check the `result` object

packages/server/src/fhir/repo.ts
```diff
@@ -1019,6 +1019,13 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
     try {
       const resource = await this.readResourceImpl(resourceType, id);
 
+      if (patch.find((operation) => operation.path === '/resourceType')) {
+        throw new OperationOutcomeError(badRequest('Incorrect Resource Type'));
+      }
+      if (patch.find((operation) => operation.path === '/id')) {
+        throw new OperationOutcomeError(badRequest('Incorrect Id'));
+      }
```
mattwiller, line 1027: I'd prefer for these checks to live in `updateResourceImpl()`, which several different `Repository` methods call into, so that this issue can be prevented in depth for the future.  @gusIreland If you don't mind making that change I'd appreciate it, but I'm also happy to handle it if you're short on time — let me know either way!

gusIreland: That makes sense! I'll update it and ping you when it's ready

packages/server/src/fhir/repo.ts
```diff
@@ -1019,6 +1019,13 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
+
       try {
         const patchResult = applyPatch(resource, patch).filter(Boolean);
```
gusIreland, line 1030: @mattwiller I moved the checks to `updateResourceImpl()`, but my new spec fails with a validation error(`Invalid additional property`) from the `applyPatch` function, which gets called before `updateResourceImpl()`.

Would it make sense to have checks at both levels? One to check that the patch call is valid, and another at the `updateResourceImpl` to prevent the issue in depth

packages/server/src/fhir/repo.ts
```diff
@@ -509,6 +509,12 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
     if (!validator.isUUID(id)) {
       throw new OperationOutcomeError(badRequest('Invalid id'));
     }
+    if (resource.resourceType !== resourceType) {
+      throw new OperationOutcomeError(badRequest('Incorrect resource type'));
+    }
+    if (resource.id !== id) {
+      throw new OperationOutcomeError(badRequest('Incorrect ID'));
+    }
```
mattwiller, line 517: These checks can never fail:
```ts
const { resourceType, id } = resource; // <-- resourceType and id come from resource
// ...
if (resource.resourceType !== resourceType) { // <-- comparing back to resource must succeed
  throw new OperationOutcomeError(badRequest('Incorrect resource type'));
}
if (resource.id !== id) { // <-- same here
  throw new OperationOutcomeError(badRequest('Incorrect ID'));
}
```

In order to make this work correctly, the updated `resourceType` and `id` should be compared against the existing ones in a stored resource to ensure they will not be changed in the update.  This might require some deeper refactoring; I can take a look in the next day or two to evaluate

gusIreland: Ah ok, I was trying to add some specs for that change and running into that problem - I'll start poking at it on my end

PR #4369, opened 2024-04-15
Dependency upgrades 2024-04-15

medplumbot:
Dependency upgrades


codyebberson (approved):
Tested on staging 👍 

Notable changes:
* `tar` upgraded to v7, which had some breaking API changes
* Something changed in Next.js/PostCSS which was only fixed by changing `postcss.config.mjs` to `postcss.config.cjs` 🤷‍♂️

Issue #4241, opened 2024-03-25
Default Resource Page should include edit and history on provider.medplum.com


reshmakh:
This is the [page](https://github.com/medplum/medplum/blob/main/examples/medplum-provider/src/pages/ResourcePage.tsx) in question

PR #4376, opened 2024-04-15
[Medplum Provider] Add Edit and History functionality to resource pages

mattlong:
Addresses #4241. Add an Edit and History view in addition to the existing details view for resource both in a patient context and standalone.

I used the pill variant of mantine tabs for navigating between the three views. I don't think there is anywhere else in Medplum where we have nested navigation like this to share a UI pattern with. For the standalone version, we could use the default tab variant instead of pills since there aren't two levels of navigation. Very open to feedback on the UI.

### Within a patient context
**Detail**
Screenshot 2024-04-15 at 3 53 10 PM

**Edit**
Screenshot 2024-04-15 at 3 53 45 PM

**History**
Screenshot 2024-04-15 at 3 53 57 PM

### Standalone resources
**Detail**
Screenshot 2024-04-15 at 4 29 36 PM

**Edit**
Screenshot 2024-04-15 at 4 29 42 PM

**History**
Screenshot 2024-04-15 at 4 29 53 PM


reshmakh:
Does this count as nested navigation?  
Is the thinking that another layer of tabs would be confusing given that the tabs are just above? 

I like the proposed, and think it makes sense.

image

mattlong:
> Does this count as nested navigation? Is the thinking that another layer of tabs would be confusing given that the tabs are just above?

I'm not entirely sure what the correct terminology is...I was referring to needing a second list of tabs (or some other way of navigating) below/within a list of tabs like the ones in your screenshot.

reshmakh:
To be clear - I'm good with this :)

codyebberson (approved):

Issue #3959, opened 2024-02-14
Explain Binary URLs + CORS


rahul1:
If using `<a>` or `<object>` tags - use binary url as is

If using JS to download the file: Parse the URL to get the `Binary/id` and call `medplum.download()`

PR #4377, opened 2024-04-16
Document Binary URLs and CORS

ksmith94:
Related to #3959. This will document potential CORS issues when downloading Binary attachments and how to avoid them.

### DIFF
--- a/packages/examples/src/fhir-datastore/binary-data.ts
+++ b/packages/examples/src/fhir-datastore/binary-data.ts
@@ -9,7 +9,7 @@ const patient: Patient = {
   photo: [
     {
       contentType: 'image/jpeg',
-      url: 'https://storage.medplum.com/12345?token=...',
+      url: 'https://storage.medplum.com/binary/12345',
     },
   ],
 };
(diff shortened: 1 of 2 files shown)


packages/examples/src/fhir-datastore/binary-data.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Patient } from '@medplum/fhirtypes';
+// end-block imports
+
+// start-block downloadBinary
+const patient: Patient = {
+  resourceType: 'Patient',
+  photo: [
+    {
+      contentType: 'image/jpeg',
+      url: 'https://storage.medplum.com/12345?token=...',
```
rahul1, line 12: example should be `https://storage.medplum.com/binary/12345`

packages/examples/src/fhir-datastore/binary-data.ts
```diff
@@ -0,0 +1,27 @@
...
+const medplum = new MedplumClient();
+
+// A function to return the url in 'Binary/{id}' format
+const binaryUrl = parsePresignedUrl(patient.photo?.[0].url);
+// Download the binary
+await medplum.download(binaryUrl);
+// end-block downloadBinary
+
+function parsePresignedUrl(url?: string): string {
+  return `Binary/${url ?? 'example-id'}`;
```
rahul1, line 26: update this function to just extract the id from the `storage.medplum.com/binary` path

rahul1 (changes requested):

PR #4379, opened 2024-04-16
feat(agent): add `Agent/$bulk-status`

### LINKED ISSUE #4194: `Agent/$status` does not support a bulk GET
As the number of Agents increase in the fleet, and needing to check the Agent status via some external mechanism, it would be preferred to issue a GET  on the Agent endpoint that would return a list of all known Agents  and Parameters and as well support the normal query parameters for resources.

ThatOneBro:
Closes #4194

### DIFF
--- a/packages/server/src/fhir/operations/agentbulkstatus.ts
+++ b/packages/server/src/fhir/operations/agentbulkstatus.ts
@@ -1,6 +1,6 @@
 import { allOk, badRequest, isOk, serverError } from '@medplum/core';
 import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
-import { Bundle, BundleEntry, OperationDefinition, OperationOutcome, Parameters } from '@medplum/fhirtypes';
+import { Agent, Bundle, BundleEntry, OperationDefinition, OperationOutcome, Parameters } from '@medplum/fhirtypes';
 import { getAuthenticatedContext } from '../../context';
 import { agentStatusHandler } from './agentstatus';
 import { getAgentsForRequest } from './agentutils';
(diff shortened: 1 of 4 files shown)


packages/server/src/fhir/operations/agentbulkstatus.ts
```diff
@@ -0,0 +1,73 @@
...
+ * First reads the agent and makes sure it is valid and the user has access to it.
+ * Then tries to get the agent status from Redis.
+ * Returns the agent status details as a Parameters resource.
+ *
+ * @param req - The FHIR request.
+ * @returns The FHIR response.
+ */
+export async function agentBulkStatusHandler(req: FhirRequest): Promise<FhirResponse> {
+  const { repo } = getAuthenticatedContext();
+  const agents = await getAgentsForRequest(req, repo);
```
codyebberson, line 33: Low pri: should we consider adding an upper bound on number of agents?

Pagination should "just work" with your current implementation, because `_count` and `_offset` should flow through automatically.

I'd be ok with a reasonably high limit here, like 100

packages/server/src/fhir/operations/agentbulkstatus.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (!agents?.length) {
+    return [badRequest('No agent(s) for given query')];
+  }
+
+  const promises = agents.map((agent) => agentStatusHandler({ ...req, params: { id: agent.id as string } }));
```
codyebberson, line 38: Low pri: there is opportunity for optimization here, as the `agentStatusHandler` basically just reads a value from Redis.  We could use Redis `mget` to make this a single Redis call.

That's probably premature optimization, so we could do that in the future.

codyebberson (approved):
lgtm after tests pass 👍

packages/server/src/fhir/operations/agentbulkstatus.ts
ThatOneBro: Yeah was thinking about this too, but my conclusion was yeah, maybe too early and don't want to potentially expose ourselves to diverging implementations yet

PR #4381, opened 2024-04-16
ci(actions): test large runners

ThatOneBro:


package.json
```diff
@@ -16,7 +16,8 @@
     "lint": "eslint .",
     "prettier": "prettier --write \"**/*.{ts,tsx,cts,mts,js,jsx,cjs,mjs,json}\"",
     "sort-package-json": "sort-package-json package.json \"packages/*/package.json\" \"examples/*/package.json\"",
-    "test": "turbo run test"
+    "test": "turbo run test --filter=!@medplum/docs",
+    "test:docs": "turbo run test --filter=@medplum/docs"
```
codyebberson, line 20: The docs package.json doesn't have a "test" script, so this should be a no-op, yeah?

scripts/test.sh
```diff
@@ -7,23 +7,10 @@ set -e
...
-  fi
-done
-
-for dir in `ls examples`; do
-  if test -f "examples/$dir/package.json" && grep -q "\"test\":" "examples/$dir/package.json"; then
-    npx turbo run test --filter=./packages/$dir
-  fi
-done
-
+npx turbo run test -- --coverage
```
codyebberson, line 13: 🎉

codyebberson (approved):
Good to go after removing `--force`

package.json
ThatOneBro: Good point, I guess I somehow hallucinated it was there at one point, probably because running `npm run test` still triggers a build of `@medplum/docs` currently if the `build` cache gets busted... not sure why. Going to see if `--filter` applies to jobs that get triggered by the current job too

ThatOneBro: Yeah looks like when `build` gets triggered, it still considers `@medplum/docs` within scope and runs `build` for `@medplum/docs` too.

Keeping `--filter=!@medplum/docs` so we don't end up building it when running `npm run test` but can remove `test:docs` of course

ThatOneBro:
Closed due to inactivity. We have decided against using the provided larger runners in favor of (if needed) spinning up our own self-hosted runners on a managed Kubernetes cluster (EKS, AKS, etc.)

PR #4393, opened 2024-04-17
fix: `AsyncAutocomplete` multi-select support

dillonstreator:
- options list is closed after each select
- search is cleared after each select
- `onChange` is fired with only the latest selection as a single element array
- re-selecting a selected option when the dropdown is open does not de-select
- default item component doesn't have indicator for which options are selected in dropdown


mattlong:
Changes from this were added from this to #4858 . Thanks!

PR #4400, opened 2024-04-18
Add note that practitioners cannot have subscriptions

ksmith94:
This will update the note in the subscription docs to include Practitioners as a resource that cannot be used as a subscription criteria.


rahul1:
I actually don't think this is accurate. The originating discord issue was actually based on a different misconception

rahul1:
I actually don't think this is accurate. The originating discord issue was actually based on a different misconception

PR #4412, opened 2024-04-20
Per-project rate limiter config

codyebberson:
Draft PR for discussion

Project now has 4 settings/secrets properties:

1. `Project.setting` - project admins, visible
2. `Project.secret` - project admins, hidden
3. `Project.systemSetting` - super admins, visible
4. `Project.systemSecret` - super admins, hidden

In the PR, `rateLimit` and `authRateLimit` are implemented as `Project.systemSetting` - so project admins can read, but only super admins can write

Some thoughts for discussion:

1. This works, and I'd be ok to ship as-is
2. The main conceptual consideration here is whether the rate limit settings should be on `Project` or on `AccessPolicy`
   * I think even if we add them to `AccessPolicy`, we will still need some super-admin only setting on `Project`
   * Therefore, I think this PR could be a stepping stone to that model
3. As implemented, the rate limiter is always *after* the context middleware
   * As a consequence, we always do some "auth" work before rate limiter
   * ... and CodeQL does not like that
   * We could maybe play tricks with a 2-tier rate limiter model?
   * I'm inclined to just mark it as ignored
   * The "auth" work is pretty minimal, and doesn't touch the db at all
      * First, validate the JWT
      * If JWT valid, perform up to 3 lookups from Redis (could be reduced to 2 with `mget`)
4. As a tactical implementation detail, I used `contentReference` in the `StructureDefinition`, so all 4 of these share a TypeScript type `ProjectSetting`
   * The type is the exact same as the previous `ProjectSecret`
   * However, to be clear, this is a "breaking change" in that the old `ProjectSecret` type was removed
   * That might show up for people who use the `ProjectSecret` type in tests with bots, because "secrets" is a prop on `BotEvent`
   * We may want to add `type ProjectSecret = ProjectSetting` just as a temp type alias, marked as deprecated, to be removed in v4


ThatOneBro (approved):

rahul1:
The type alias would be appreciated to maintain some of the tests I've written for customers!

packages/server/src/ratelimit.ts
```diff
@@ -1,10 +1,15 @@
...
+import { Request } from 'express';
 import rateLimit, { MemoryStore, RateLimitRequestHandler } from 'express-rate-limit';
-import { OperationOutcomeError, tooManyRequests } from '@medplum/core';
+import { AuthenticatedRequestContext, getRequestContext } from './context';
 
 /*
  * MemoryStore must be shutdown to cleanly stop the server.
  */
 
+const DEFAULT_RATE_LIMIT_PER_15_MINUTES = 10000;
```
rahul1, line 10: Could you add a brief note [here](https://www.medplum.com/docs/fhir-datastore) with the new default rate limit>

ThatOneBro:
New project model seems to make a lot of sense. Nice that you can use `contentReference` to dedupe `StructureDefinition`s 👍

Regarding 2: I think it's fine to live on the Project for now, we need this secrets model anyways. Could also be the place for project default with overrides in `AccessPolicy` added later on?

Regarding 3: Not too concerned about auth stuff being before rate limiter, JWT verification is not terribly expensive (though it may become a bottleneck somewhere down the line for mitigation)

rahul1:
One request for docs on the default limit, otherwise LGTM. 

In a follow-up PR, we should add this as a config settings for self-hosting users

codyebberson:
Verified that the `ProjectSecret` type still works:

image

PR #4419, opened 2024-04-21
Remove runInBand from server tests

codyebberson:
Running `npm t` in `packages/server` on my local dev machine:
* Before: ~470 sec
* After: ~140 sec (3.3x speedup!)

This is all possible due to our existing project level isolation features.

The main changes required:
* Refactored old tests that called the db directly
* Moved some test setup from `beforeAll` to `beforeEach` to avoid test pollution
* Remove calls to `redis.flushdb()` - many tests rely on data in redis, so tests need to surgically clean up after themselves
* Updated `test.sh` to run `seed.test.ts` first in isolation, then everything else concurrently

--------

Future work:

The long pole now is `src/fhir/repo.test.ts`, which just needs to be split up into multiple files.

I'm considering adding a new directory for `src/fhir/interactions/` (similar to `src/fhir/operations/`), and moving all of the standard [FHIR Interactions](https://hl7.org/fhir/r4/http.html) there, (i.e., `create.ts`, `read.ts`, `update.ts`, etc).  That would (1) improve the readability and (2) improve parallelization of test runner.

We already did something similar for `search.ts`, which was about half of `repo.ts` at the time.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -18,7 +18,7 @@
     "clean": "rimraf dist",
     "dev": "ts-node-dev --poll --respawn --transpile-only --require ./src/otel/instrumentation.ts src/index.ts",
     "start": "node --require ./dist/otel/instrumentation.js dist/index.js",
-    "test": "jest --runInBand"
+    "test": "jest"
   },
   "dependencies": {
     "@aws-sdk/client-cloudwatch-logs": "3.554.0",
--- a/packages/server/src/agent/websockets.test.ts
+++ b/packages/server/src/agent/websockets.test.ts
@@ -21,7 +21,7 @@ describe('Agent WebSockets', () => {
   beforeAll(async () => {
     config = await loadTestConfig();
     config.vmContextBotsEnabled = true;
-    config.heartbeatMilliseconds = 100000;
+    config.heartbeatMilliseconds = 5000;
 
     server = await initApp(app, config);
     accessToken = await initTestAuth({ membership: { admin: true } });
--- a/packages/server/src/agent/websockets.test.ts
+++ b/packages/server/src/agent/websockets.test.ts
@@ -21,7 +21,7 @@ describe('Agent WebSockets', () => {
   beforeAll(async () => {
     config = await loadTestConfig();
     config.vmContextBotsEnabled = true;
-    config.heartbeatMilliseconds = 500;
+    config.heartbeatMilliseconds = 100000;
 
     server = await initApp(app, config);
     accessToken = await initTestAuth({ membership: { admin: true } });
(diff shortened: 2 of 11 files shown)


packages/server/src/oauth/keys.test.ts
```diff
@@ -25,25 +22,6 @@ describe('Keys', () => {
     await shutdownApp();
   });
 
-  test('Init keys', () =>
```
codyebberson, line 28: This is a very old test, which predates almost everything.  All code paths here are already covered by `initAppServices()` which runs in `seed.test.ts`, and most other server tests.

packages/server/src/agent/websockets.test.ts
```diff
@@ -21,7 +21,7 @@ describe('Agent WebSockets', () => {
   beforeAll(async () => {
     config = await loadTestConfig();
     config.vmContextBotsEnabled = true;
-    config.heartbeatMilliseconds = 500;
+    config.heartbeatMilliseconds = 5000;
```
ThatOneBro, line 24: Why does this need to be increased? Test was flaky?

packages/server/package.json
```diff
@@ -18,7 +18,7 @@
     "clean": "rimraf dist",
     "dev": "ts-node-dev --poll --respawn --transpile-only --require ./src/otel/instrumentation.ts src/index.ts",
     "start": "node --require ./dist/otel/instrumentation.js dist/index.js",
-    "test": "jest --runInBand"
+    "test": "jest"
```
ThatOneBro, line 21: Nice

scripts/test.sh
```diff
@@ -9,6 +9,18 @@ set -x
...
 
+# Clear old code coverage data
+rm -rf coverage
+mkdir -p coverage/packages
+mkdir -p coverage/combined
+
+# Seed the database
+# This is a special "test" which runs all of the seed logic, such as setting up structure definitions
+# On a normal developer machine, this is run only rarely when setting up a new database
+# This test must be run first, and cannot be run concurrently with other tests
```
ThatOneBro, line 20: > This test must be run first, and cannot be run concurrently with other tests

*yet*

Got a PR to follow this up that will allow them to be run in parallel thanks to compressed schema 😉

ThatOneBro (approved):

packages/server/src/agent/websockets.test.ts
codyebberson: Yeah ..  the heartbeat would come in before the expected "transmit" message.  Ideally there would be a way to say "expect these messages in any order", but didn't see that 

Alternatively, the test file could be refactored to support enabling/disabling the heartbeat as necessary.

The-Alchemist: Wondering if this change might help too: https://github.com/medplum/medplum/pull/9533

PR #4422, opened 2024-04-21
Added --update option to reinstall.sh

codyebberson:
Before:

1. `reinstall.sh` did not take any options, and always deleted `package-lock.json` and always used `npm i`

After:

1. By default, `reinstall.sh` now uses `npm ci` which uses `package-lock.json`, which should put the project in a clean state
2. You can now pass in `reinstall.sh --update` to delete `package-lock.json`, use `npm i`, which creates a fresh new `package-lock.json`

### DIFF
--- a/scripts/reinstall.sh
+++ b/scripts/reinstall.sh
@@ -7,7 +7,6 @@ set -e
 set -x
 
 rm -rf node_modules
-rm -rf package-lock.json
 
 for dir in `ls packages`; do
   if test -d "packages/$dir/node_modules"; then
@@ -21,4 +20,10 @@ for dir in `ls examples`; do
   fi
 done
 
-npm i --strict-peer-deps
+# If called with "--update", then use npm i
+if [ "$1" == "--update" ]; then
+  rm -rf package-lock.json
+  npm i --strict-peer-deps
+else
+  npm ci --strict-peer-deps
+fi
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -40,7 +40,7 @@ git push origin "$BRANCH_NAME"
 gh pr create --title "Dependency upgrades $DATE" --body "Dependency upgrades" --draft
 
 # Reinstall all dependencies
-./scripts/reinstall.sh
+./scripts/reinstall.sh --update
 
 # Commit and push after running NPM install
 git add -u .


scripts/reinstall.sh
```diff
@@ -7,7 +7,6 @@ set -e
 set -x
 
 rm -rf node_modules
```
ThatOneBro, line 9: I believe removing `node_modules` in the new default case (`npm ci`) doesn't do anything but prevent the fast path from being hit where Node can diff `package-lock.json` and the local `.package-lock.json` in `node_modules`. https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json#hidden-lockfiles

`npm ci` should end up with the exact same `node_modules` no matter if you delete it or not, because at the hash of the `node_modules` has to match the hash in `package-lock.json` (at least before `postinstall` step)

What is the desired behavior from `npm ci` after deleting the `node_modules`?

ThatOneBro (changes requested):

scripts/reinstall.sh
codyebberson: Hm, interesting.  I'll read more.  Maybe I've confounded things, but I feel like I've landed in uncertain state from this before.

If the contract of `npm ci` is "your directory will be exactly as it is supposed to be, including deleting unwanted files/folders/versions", then yeah, this pr is unnecessary

ThatOneBro: Yeah I thought the same thing but seems like `npm install` and `npm ci` have been pruning stuff for a while now. We should test it to be sure though, not 100% convinced that the theoretical matches the practical either...

ThatOneBro: Ah actually I know why this may work... 

`.turbo` and other `.x` dirs in `node_modules` are not wiped by `npm ci`...
We could then perhaps be more surgical, but maybe best just to leave it as is.

ThatOneBro (approved):

scripts/reinstall.sh
codyebberson: I debated adding `rm -rf .next` and `rm -rf .turbo` to each of the subdirectories.

That's more of a "clean" than "reinstall" though.  Maybe another script for another day.

PR #4435, opened 2024-04-23
Fixes #3240 - rate-limit-redis persistent rate limit storage

### LINKED ISSUE #3240: redis-based store for rate limiter

codyebberson:
Rebased https://github.com/medplum/medplum/pull/3308 and fixed all merge conflicts.

Thank you @dillonstreator !


packages/server/src/ratelimit.ts
```diff
@@ -1,21 +1,23 @@
...
 let handler: RateLimitRequestHandler | undefined = undefined;
-let store: MemoryStore | undefined = undefined;
+let store: RedisStore | undefined = undefined;
 
 export function getRateLimiter(): RateLimitRequestHandler {
   if (!handler) {
-    store = new MemoryStore();
+    const client = getRedis();
+    store = new RedisStore({
+      // @ts-expect-error - Pass string args direct to ioredis
```
ThatOneBro, line 18: A more descriptive error would be useful for future readers
```suggestion
      // See: https://www.npmjs.com/package/rate-limit-redis#:~:text=//%20%40ts%2Dexpect%2Derror%20%2D%20Known%20issue%3A
      // @ts-expect-error - Known issue: the `call` function is not present in @types/ioredis
```

ThatOneBro (approved):

packages/server/src/ratelimit.ts
codyebberson: K, will do.

"Known issue: the `call` function is not present in @types/ioredis" is no longer correct.  The `call` function is present, it just has a bunch of TypeScript shenanigans for the proper ordering of Redis commands.

ThatOneBro: I see, fair. I guess the readme hasn't been updated yet

PR #4443, opened 2024-04-24
Super admin endpoint for database stats

codyebberson:


packages/server/src/admin/super.ts
```diff
@@ -221,6 +222,41 @@ superAdminRouter.post(
   })
 );
 
+// POST to /admin/super/dbstats
+// to query database statistics.
+superAdminRouter.post(
+  '/dbstats',
```
mattwiller, line 228: Should this be an operation endpoint?

packages/server/src/admin/super.ts
```diff
@@ -221,6 +222,41 @@ superAdminRouter.post(
...
+      `;
+
+      const result = await client.query(sql);
+
+      const contentString = result.rows
+        .map((row) => `${row.table_schema}.${row.table_name}: ${row.table_size}`)
+        .join('\n');
+
+      await systemRepo.createResource<Communication>({
+        resourceType: 'Communication',
```
mattwiller, line 252: Why `Communication` here, and why is it persisted to the DB?

mattwiller (approved):

packages/server/src/admin/super.ts
codyebberson: With the existing `/admin/super/x` APIs, they all require async, and don't have any representation of return value.

It's a good question: how should we capture the return value of a async operation?  We could add the output `Parameters` to the `AsyncJob`?  It would still be persisted in the db, but at least it wouldn't be as contrived as `Communication`.

mattwiller: Yeah, that would be my suggestion — persisted `Parameters` sounds good to me

ThatOneBro: You broke the one rule!
https://hl7.org/fhir/R4/parameters.html#:~:text=Scope%20and%20Usage-,This%20resource%20is%20a%20non%2Dpersisted%20resource,-used%20to%20pass

But I mean yeah I think it makes the most sense

PR #4455, opened 2024-04-26
feat(agent): support `Prefer: respond-async` for `Agent/$push`

### LINKED ISSUE #4106: Support "prefer async" in agent push

ThatOneBro:
Related to #4106

### DIFF
--- a/packages/server/src/fhir/operations/agentpush.ts
+++ b/packages/server/src/fhir/operations/agentpush.ts
@@ -7,7 +7,7 @@ import {
   getReferenceString,
   serverError,
 } from '@medplum/core';
-import { Agent } from '@medplum/fhirtypes';
+import { Agent, OperationOutcome, Parameters } from '@medplum/fhirtypes';
 import { Request, Response } from 'express';
 import { randomUUID } from 'node:crypto';
 import { asyncWrap } from '../../async';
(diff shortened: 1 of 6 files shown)


packages/server/src/fhir/operations/agentpush.ts
```diff
@@ -35,45 +36,78 @@ const MAX_WAIT_TIMEOUT = 60000;
  * Returns the outcome of the agent execution.
  */
 export const agentPushHandler = asyncWrap(async (req: Request, res: Response) => {
+  if (req.header('Prefer') === 'respond-async') {
```
codyebberson, line 39: It's not immediately obvious to me, but I suspect there's a pattern that we could develop

```ts
function maybeAsync(req: Request, res: Response, handler: SomeFunctionType): void
if (isPreferAsync(req)) {
  sendAsyncResponse(req, res, handler);
} else {
  sendResponse(handler(req, res));
}
```

+/- `async`, `await`, and a bunch of error handling and whatnot
 
With the long term goal that 100% of FHIR operations could automatically get this exact same behavior.

I'm not totally sure if we have all the right primitives and utilities to make that happen, so I'm totally ok to ship this, and let the patterns bake a little bit more.

codyebberson (approved):
Nice, lgtm, thanks 👍 🙏 

See comment about how we might be able to abstract and generalize this.  I think it's responsible to ship this first, and let our minds marinate on it.

PR #4457, opened 2024-04-26
feat(agent): `Agent/$reload-config` operation

### LINKED ISSUE #3898: Medplum Agent Command: Reload Config
Right now the the physical agent only reads the config in the `Agent` resource upon startup. We would like to be able to update this resource, and then push the configuration to the running agent

ThatOneBro:
Fixes #3898

### DIFF
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -56,13 +56,17 @@ export interface AgentReloadConfigResponse extends BaseAgentMessage {
   statusCode: number;
 }
 
-export type AgentMessage =
-  | AgentError
+export type AgentRequestMessage =
   | AgentConnectRequest
-  | AgentConnectResponse
   | AgentHeartbeatRequest
-  | AgentHeartbeatResponse
   | AgentTransmitRequest
+  | AgentReloadConfigRequest;
+
+export type AgentResponseMessage =
+  | AgentConnectResponse
+  | AgentHeartbeatResponse
   | AgentTransmitResponse
-  | AgentReloadConfigRequest
-  | AgentReloadConfigResponse;
+  | AgentReloadConfigResponse
+  | AgentError;
+
+export type AgentMessage = AgentRequestMessage | AgentResponseMessage;
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -12,6 +12,10 @@ export interface AgentError extends BaseAgentMessage {
   body: string;
 }
 
+export interface AgentSuccess extends BaseAgentMessage {
+  type: 'agent:success';
+}
+
 export interface AgentConnectRequest extends BaseAgentRequestMessage {
   type: 'agent:connect:request';
   agentId: string;
@@ -53,6 +57,7 @@ export interface AgentReloadConfigRequest extends BaseAgentMessage {
 
 export type AgentMessage =
   | AgentError
+  | AgentSuccess
   | AgentConnectRequest
   | AgentConnectResponse
   | AgentHeartbeatRequest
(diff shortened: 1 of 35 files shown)


packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -0,0 +1,202 @@
...
+  agent: Agent,
+  message: BaseAgentRequestMessage,
+  options?: AgentMessageOptions
+): Promise<[OperationOutcome] | [OperationOutcome, T | AgentError]> {
+  if (options?.waitForResponse) {
+    let resolve: (response: [OperationOutcome, T | AgentError]) => void;
+    let reject: (err: OperationOutcomeError) => void;
+
+    // If a callback doesn't already exist on the message, tie callback to the associated agent and assign a random ID
+    message.callback ??= getReferenceString(agent) + '-' + randomUUID();
```
codyebberson, line 171: Do we set `message.callback` anywhere else?  Sorry if I missed it.  Might be nice to only set it exactly once here, so we don't need the conditional logic.

packages/server/src/fhir/operations/agentpush.ts
```diff
@@ -118,47 +107,26 @@ async function pushToAgent(
...
-    sendOperationResponse(allOk);
+  // Publish the message to the agent channel
+  const [outcome, response] = await publishAgentMessage<AgentTransmitResponse>(
+    agent,
+    message,
+    params.waitForResponse ? { waitForResponse: true, timeout: waitTimeout } : undefined
+  );
+
+  if (!response) {
+    sendOperationResponse(outcome);
```
codyebberson, line 118: Non-blocking, future work: It might be nice to revisit whether this could have been done with `async` rather than promises and callbacks.

packages/core/src/agent.ts
```diff
@@ -12,6 +12,10 @@ export interface AgentError extends BaseAgentMessage {
   body: string;
 }
 
+export interface AgentSuccess extends BaseAgentMessage {
```
codyebberson, line 15: 🤔 

I see the appeal of having a singular "success" response, but I wonder if we should future-proof this with a dedicatd `AgentReloadResponse` type, in anticipation of additional data coming eventually (i.e., what changed, how many channels restarted, etc)

packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -0,0 +1,202 @@
...
+      reject(new OperationOutcomeError(badRequest('Timeout')));
+    }, options?.timeout ?? 5000);
+
+    const cleanup = (): void => {
+      redisSubscriber.disconnect();
+      clearTimeout(timer);
+    };
+
+    await getRedis().publish(getReferenceString(agent), JSON.stringify(message));
+    return new Promise<[OperationOutcome, T | AgentError]>((_resolve, _reject) => {
```
codyebberson, line 194: Any reason we can't follow the normal convention of putting everything in the body of the promise?

packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -0,0 +1,202 @@
...
+      resolve = _resolve;
+      reject = _reject;
+    });
+  }
+
+  await getRedis().publish(getReferenceString(agent), JSON.stringify(message));
```
codyebberson, line 200: Recommend we bring back the `publishMessage()` helper which wraps this logic.  The convention of using the agent reference string as the redis channel name might not be totally obvious.

codyebberson (approved):
Nice, this turned out pretty clean.

Handful of non-blocking questions.

If ok with you, I think we should do a round of testing on a Windows machine before releasing.

PR #4466, opened 2024-04-27
Update default project access policy for new systemSetting

codyebberson:
1. Updated default project admin access policy
    * Added `systemSetting` to `readonlyFields`
    * Added `systemSecret` to `hiddenFields`
2. Fixed a couple bugs in JSONSchema generator
    * Two lingering bugs that have been hiding since the switch from `ElementDefinition` to `InternalElementDefinition`
    * `medplumTypes` was wrong, so no JSONSchema updates were ever coming through
    * `excludedValueSets` check was inverted, so we never used JSONSchema `enum`
    * In practice, this only affects projects not using strict mode, which is increasingly rare, which is why this hid for as long as it did
3. Replaced all `DEPRECATED` with `@deprecated` for TypeScript TSDoc annotations
    * This is mostly a cosmetic change, but nice to be consistent


ThatOneBro:
One small typo, everything else looks good

PR #4474, opened 2024-04-29
Add Extension to Bot resource

### LINKED ISSUE #4473: Bot resource missing Extension
Was there a design reason that the Bot resource does not support an Extension? With the addition of [https://github.com/medplum/medplum/pull/4324](https://github.com/medplum/medplum/pull/4324) with the Bot reference on the Event, it would allow adding Bot specific config via Extensions instead of needing Project secrets for config values.

jmalobicky:
Fixes #4473 

Having an issue fixing the below on the Timeline view.
image

### DIFF
--- a/packages/definitions/dist/fhir/r4/fhir.schema.json
+++ b/packages/definitions/dist/fhir/r4/fhir.schema.json
@@ -59273,6 +59273,31 @@
           "description": "The base language in which the resource is written.",
           "$ref": "#/definitions/code"
         },
+        "text": {
+          "description": "A human-readable narrative that contains a summary of the resource and can be used to represent the content of the resource to a human. The narrative need not encode all the structured data, but is required to contain sufficient detail to make it \"clinically safe\" for a human to just read the narrative. Resource definitions may define what content should be represented in the narrative to ensure clinical safety.",
+          "$ref": "#/definitions/Narrative"
+        },
+        "contained": {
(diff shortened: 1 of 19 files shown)


jmalobicky:
@rahul1 @ThatOneBro Any help on how to resolve the Timeline display issue?

codyebberson:
Thanks @jmalobicky 

For context:

FHIR resources can "extend" either [`Resource`](https://hl7.org/fhir/R4/resource.html) or [`DomainResource`](https://hl7.org/fhir/R4/domainresource.html) (`DomainResource` extends `Resource`).

`extension` is defined on `DomainResource` but not `Resource`

It looks like when we wrote the original `Bot` `StructureDefinition`, we extended `Resource`, not `DomainResource`.

As @mattwiller pointed out:

> This resource extends the base Resource. All of the listed Resources except Bundle, Parameters and Binary extend this resource.

So, it seems like it’s the rule with few exceptions.

I support this change, but I think we should go all the way and add all of the `DomainResource` elements (`text`, `contained`, `extension`, and `modifierExtension`).

jmalobicky:
Thanks for the context. Let me see if I can make those changes.

codyebberson:
Whoops!  Had already started on this!  Feel free to reject

jmalobicky:
You are good. You know way more about this space than I do.

ThatOneBro (approved):

PR #4478, opened 2024-04-29
[Medplum Provider app] Integrate specialized Tasks tab from examples/medplum-task-demo

### LINKED ISSUE #4249: Merge Task Demo and provider.medplum.com page for Task resource
`/Patient/:id/task` should include middle and side buttons from Task Demo

This will require making a custom Task route in provider.medplum.com

mattlong:
* I was almost entirely able to copy/paste in components from `examples/medplum-task-demo` and use them as is.
* I cleaned up and normalized the styling of some of the modals shown for various task actions to be consistent with the styling in the other modals which are derived from `QuestionnaireForm`. I have NOT back-ported those to `examples/medplum-task-demo`. If that's desirable, we should make an issue for it. It's a bit unfortunate that we'll have duplicate code in separate examples that'll start drifting apart over time...
* Some complexity had to be added when deleting a task to correctly wind up back at list view of a patient's task with the correct search parameters set, e.g. `/Patient/6391816a-5fc9-405f-9097-b51c2ca9315c/Task?_count=20&_fields=_lastUpdated,code,status,focus&_offset=0&_sort=-_lastUpdated&patient=6391816a-5fc9-405f-9097-b51c2ca9315c`.

### Task detail view:
Screenshot 2024-04-29 at 2 19 50 PM

#### Claim Task

| Before | After |
|--------|--------|
| Screenshot 2024-04-29 at 2 36 00 PM | Screenshot 2024-04-29 at 3 01 54 PM |

#### Delete Task
| Before | After |
|--------|--------|
| Screenshot 2024-04-29 at 2 36 14 PM | Screenshot 2024-04-29 at 2 30 12 PM |

### Patient's task list view remains unchanged:
Screenshot 2024-04-29 at 2 19 31 PM

Fixes #4249

### DIFF
--- a/examples/medplum-provider/src/components/tasks/TaskList.tsx
+++ b/examples/medplum-provider/src/components/tasks/TaskList.tsx
@@ -173,10 +173,14 @@ function TaskTitle(props: TaskCellProps): JSX.Element {
       setTitle(<>{props.resource.title}</>);
     } else if (props.resource.resourceType === 'QuestionnaireResponse') {
       fetchQuestionnaireTitle().catch(console.error);
+    } else if (props.task.code) {
+      setTitle(<CodeableConceptDisplay value={props.task.code} />);
     } else {
+      // TODO what should this else actually be?
       setTitle(<>{props.task.code}</>);
     }
   }, [props.resource, props.task, medplum]);
 
+  console.log('TaskTitle', title);
   return title ?? <></>;
 }
--- a/examples/medplum-provider/src/components/tasks/TaskList.tsx
+++ b/examples/medplum-provider/src/components/tasks/TaskList.tsx
@@ -1,5 +1,5 @@
 import { Box, Card, Divider, Flex, Group, Text, Title } from '@mantine/core';
-import { formatDate } from '@medplum/core';
+import { formatDate, getDisplayString } from '@medplum/core';
 import { CodeableConcept, Questionnaire, Reference, Resource, Task } from '@medplum/fhirtypes';
 import {
   CodeableConceptDisplay,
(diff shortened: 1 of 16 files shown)


examples/medplum-provider/src/components/tasks/TaskList.tsx
```diff
@@ -173,10 +173,14 @@ function TaskTitle(props: TaskCellProps): JSX.Element {
       setTitle(<>{props.resource.title}</>);
     } else if (props.resource.resourceType === 'QuestionnaireResponse') {
       fetchQuestionnaireTitle().catch(console.error);
+    } else if (props.task.code) {
+      setTitle(<CodeableConceptDisplay value={props.task.code} />);
     } else {
+      // TODO what should this else actually be?
```
mattlong, line 179: Previously, the else block was causing throwing an error since `props.task.code` (an object) is not a valid ReactNode. What would be an appropriate wayto cast a task to a string here if none of the other blocks trigger? Perhaps `Task/${id}`?

examples/medplum-provider/src/components/tasks/actions/AddDueDate.tsx
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      const result = await medplum.patchResource('Task', taskId, ops);
+      notifications.show({
+        icon: <IconCircleCheck />,
+        title: 'Success',
+        message: 'Due-date updated.',
+      });
+      onChange(result);
+    } catch (error) {
+      notifications.show({
```
codyebberson, line 52: Maybe a topic for discussion, but we use this pattern all over the place.  Probably overdue for a helper utility for "show success message" / "show error message".

codyebberson (approved):

PR #4499, opened 2024-05-02
Fixed horizontal scrollbar on provider app patient pages

codyebberson:


codyebberson:
Included in https://github.com/medplum/medplum/pull/4521

PR #4501, opened 2024-05-02
fix: `AttachmentDisplay` use uncached url for download link

dillonstreator:


packages/react/src/AttachmentDisplay/AttachmentDisplay.tsx
```diff
@@ -39,7 +39,8 @@ export function AttachmentDisplay(props: AttachmentDisplayProps): JSX.Element |
       )}
       <div data-testid="download-link" style={{ padding: '2px 16px 16px 16px' }}>
         <Anchor
-          href={url}
+          // use the `uncachedUrl` to download the file as the cached URL may expire by the time the user clicks the download link
+          href={uncachedUrl}
```
dillonstreator, line 43: `useCachedBinaryUrl` _could_ additionally return a function that signals if the url is expired which would allow branching to use the preferred url i.e.
```ts
const {url, checkIsExpired} = useCachedBinaryUrl(...);
//...
href={checkIsExpired() ? uncachedUrl : url}
```

codyebberson (approved):

PR #4502, opened 2024-05-02
add example for self hosting in docker compose with SSL enabled and n…

geoHeil:
…ice DNS


examples/hosting-docker-compose/README.md
```diff
@@ -0,0 +1,48 @@
...
+## setup prerequisites:
+
+- git
+- docker
+- make
+- sops, age
+- keyfile to decrypt the secrets
+- mkcert & permission to load the SSL certificate in your broser
+- a DNS name i.e. myname.com
+    - replace any occurence of myname.com with your chosen name
```
geoHeil, line 14: maybe 127.0.0.1 redirect shuld be mentioned as a quick dev setup

geoHeil:
that key files was just dummy certs - but maybe it is better to delete them

rahul1:
@geoHeil thanks so much for the PR! The community will really appreciate this as a default setup. 

After discussing with the team, I think the best place for this would be a separate repo inside your account. Once that's set up, we can link to it from our documentation to point community members to it

geoHeil:
hey @rahul1 is this https://github.com/geoHeil/medplum-self-hosted what you mean?

PR #4515, opened 2024-05-06
Dependency upgrades 2024-05-06

medplumbot:
Dependency upgrades


codyebberson (approved):
Notable changes:
* Holding back `react` and `react-dom` due to incompatibility with `react-native`
* Holding back `supertest` due to incompatibility with `superwstest`
* Removed `chromatic` from exclude list, upgraded to latest

PR #4516, opened 2024-05-06
Fixes #4509 - Improve exact match search support for token parameters in `matchesTokenFilter`

### LINKED ISSUE #4509: `MockClient` returns incorrect data when searching with `token` parameters due to the way `matchesTokenFilter` operates
## Issue

`MockClient` returns incorrect data when searching with `token` parameters because the current implementation of the `matchesTokenFilter` function does not support **exact** match searches, as specified in the [FHIR documentation](https://build.fhir.org/search.html#token). 

## Problem

In the original implementation, the `matchesTokenFilter` function uses the `includes` method to check if the token value is included in the search parameter. This implementation does not support exact match searches, as specified in the [FHIR documentation](https://build.fhir.org/search.html#token).

Taking into account that the `MockClient` uses the `MemoryRepository` which uses the `matchesSearchRequest` function to filter the resources, we can get incorrect results when searching with `token` parameters. For example, if `mockClient.upsertResource` is called with an observation with code `12345-6` and then another `mockClient.upsertResource` is called with an observation with code `1234`, the first will be replaced by the second one because `MemoryRepository` will be used and it will consider they are the same resource because the `matchesSearchRequest` function will return `true` when it should return `false`.

### Example 1

When searching for a patient with the `identifier` value `1234567890`, the `matchesTokenFilter` function will return `true` for any `identifier` value that includes `1234567890`, such as `1234`, `12345`, etc.

When we add the following test to packages/core/src/search/match.test.ts, it fails when it should pass:

<details>

<summary><strong>:arrow_right: Click to expand and see the code</strong></summary>

```typescript
test('[Token] Datatype: Identifier', () => {
  const identifier = '1234567890';
  const identifierSubstring = identifier.substring(0, 4);
  const resource: Patient = {
    resourceType: 'Patient',
    identifier: [
      {
        system: 'http://example.com',
        value: identifier,
      },
      {
        system: 'http://test.com',
        value: identifier,
      },
      {
        system: 'http://test.com',
        value: 'foo',
      },
    ],
  };

  expect(
    matchesSearchRequest(resource, {
      resourceType: 'Patient',
      filters: [
        { code: 'identifier', operator: Operator.EQUALS, value: 'http://example.com|' + identifierSubstring },
      ],
    })
  ).toBe(false); // FAIL! It currently returns true, but it should return false

  expect(
    matchesSearchRequest(resource, {
      resourceType: 'Patient',
      filters: [{ code: 'identifier', operator: Operator.EQUALS, value: identifierSubstring }],
    })
  ).toBe(false); // FAIL! It currently returns true, but it should return false
});
```
</details>

### Example 2

When searching for an observation with the `code` value `12345-6`, the `matchesTokenFilter` function will return `true` for any `code` value that includes `12345-6`, such as `12345`, `12345-`, etc.

When we add the following test to packages/core/src/search/match.test.ts, it fails when it should pass:

<details>

<summary><strong>:arrow_right: Click to expand and see the code</strong></summary>

```typescript
test('[Token] Datatype: CodeableConcept', () => {
  const identifier = '12345-6';
  const identifierSubstring = identifier.substring(0, 4);
  const resource: Observation = {
    resourceType: 'Observation',
    status: 'final',
    code: {
      coding: [
        {
          system: 'http://example.com',
          code: identifier,
        },
        {
          system: 'http://test.com',
          code: identifier,
        },
        {
          system: 'http://test.com',
          code: 'foo',
        },
      ],
      text: 'test',
    },
  };

  expect(
    matchesSearchRequest(resource, {
      resourceType: 'Observation',
      filters: [{ code: 'code', operator: Operator.EQUALS, value: 'http://example.com|' + identifierSubstring }],
    })
  ).toBe(false); // FAIL! It currently returns true, but it should return false

  expect(
    matchesSearchRequest(resource, {
      resourceType: 'Observation',
      filters: [{ code: 'code', operator: Operator.EQUALS, value: identifierSubstring }],
    })
  ).toBe(false); // FAIL! It currently returns true, but it should return false
});

```
</details>

## Proposed solution

That being said, I thought about a partial solution that would allow us to support exact match searches for `token` parameters. I say partial because it would address the identifier and code types. Here is a code snippet:

<details>

<summary><strong>:arrow_right: Click to expand and see the code</strong></summary>

```typescript
function matchesStringFilter(
  resource: Resource,
  filter: Filter,
  searchParam: SearchParameter,
  asToken?: boolean
): boolean {
  const details = getSearchParameterDetails(resource.resourceType, searchParam);
  const searchParamElementType = details.elementDefinitions?.[0].type?.[0].code;
  const resourceValues = evalFhirPath(searchParam.expression as string, resource);
  const filterValues = filter.value.split(',');
  const negated = isNegated(filter.operator);
  for (const resourceValue of resourceValues) {
    for (const filterValue of filterValues) {
      let match;
      if (searchParamElementType === PropertyType.Identifier) {
        match = matchesTokenIdentifierValue(resourceValue as Identifier, filter.operator, filterValue);
      } else if (searchParamElementType === PropertyType.CodeableConcept) {
        match = matchesTokenCodeableConceptValue(resourceValue as Coding, filter.operator, filterValue);
      } else {
        match = matchesStringValue(resourceValue, filter.operator, filterValue, asToken);
      }
      if (match) {
        return !negated;
      }
    }
  }
  // If "not equals" and no matches, then return true
  // If "equals" and no matches, then return false
  return negated;
}

function matchesTokenCodeableConceptValue(
  resourceValue: CodeableConcept,
  operator: Operator,
  filterValue: string
): boolean {
  if (filterValue.includes('|')) {
    const [system, code] = filterValue.split('|');
    return (
      resourceValue.coding?.some(
        (coding) =>
          coding.system?.toLowerCase() === system.toLowerCase() && coding.code?.toLowerCase() === code.toLowerCase()
      ) ?? false
    );
  }

  return (
    resourceValue.text?.toLowerCase() === filterValue.toLowerCase() ||
    (resourceValue.coding?.some((coding) => coding.code?.toLowerCase() === filterValue.toLowerCase()) ?? false)
  );
}

```
</details>

I was working on this solution, however I wonder if:

- It would be enough to have a partial solution because I understand there are a lot of nuances in FHIR standards about search such as operators and modifiers.
- It would be a overkill to implement a full solution considering (I suppose) that the main usage is for testing purposes. 
- It might be a breaking change to other `matches*` functions that are using the `matchesStringFilter` function. For example, `matchesAccessPolicyResourcePolicy` and `matchesCriteria`.
- It might be a breaking change to customers that are already using the `MockClient` class.

So I would like to know your opinion about all of this.

Please let me know if you have any questions or need further information. I am looking forward to your feedback. Thank you!

pamella:
## Description

I opened this PR so you folks can have a closer look at the proposed partial solution mentioned in #4509.

> ## Proposed solution
> 
> That being said, I thought about a partial solution that would allow us to support exact match searches for token parameters. I say partial because it would address the identifier and code types.
> 
> I was working on this solution, however I wonder if:
> 
> - It would be enough to have a partial solution because I understand there are a lot of nuances in FHIR standards about search such as operators and modifiers.
> - It would be a overkill to implement a full solution considering (I suppose) that the main usage is for testing purposes. 
> - It might be a breaking change to other `matches*` functions that are using the `matchesStringFilter` function. For example, `matchesAccessPolicyResourcePolicy` and `matchesCriteria`.
> - It might be a breaking change to customers that are already using the `MockClient` class.
> 
> So I would like to know your opinion about all of this.
> 
> Please let me know if you have any questions or need further information. I am looking forward to your feedback. Thank you!


codyebberson:
Hi @pamella  - thank you for submitting!

> It would be enough to have a partial solution because I understand there are a lot of nuances in FHIR standards about search such as operators and modifiers.
> It would be a overkill to implement a full solution considering (I suppose) that the main usage is for testing purposes.

Yes, this is an inherent challenge with `MockClient`.  Our general philosophy has been "do as good as we can with code, but accept data limits".

This draft PR definitely qualifies for "do as good as we can with code".

The "accept data limits" disclaimer refers to the impossibility of implementing everything without 100's of MB of `StructureDefinition`, `SearchParameter`, `CodeSystem`, and `ValueSet` resources.

> It might be a breaking change to other matches* functions that are using the matchesStringFilter function. For example, matchesAccessPolicyResourcePolicy and matchesCriteria.
> It might be a breaking change to customers that are already using the MockClient class.

This is a fair callout.  We generally accept anything that moves us closer to spec compliance, and I believe this PR qualifies for that.

Thank you for your submission.  Cursory review of the code looks good to me.  I'll take another deeper look once all automated tests have passed.  Please let us know if you have any questions.

pamella:
Hi @codyebberson, thank you for looking at this and sharing the context around the philosophy!

---

I noticed that a test failed but passed when I ran it locally. Would it be the case to rerun the [related action job](https://github.com/medplum/medplum/actions/runs/8971611656/job/24642143839?pr=4516)?

image

pamella:
I was able to reproduce the failing test locally after rerunning `npm ci` and `npm run build:fast`. But, curiously, it also fails in the `main` branch. Do you folks have an initial guess as to what could be causing this?

codyebberson:
Thanks @pamella  - it looks like the test failure is consistent.  We are investigating.

pamella:
Thank you, Cody! In case there is anything I should adjust in the PR, please let me know. Meanwhile, I'm looking here to see if there's anything I can solve on my end.

pamella:
Recent finding:
- On `main` branch, the test passes after running `npm ci` and `npm run build:fast`.
- On this PR branch, the test fails after running `npm ci` and `npm run build:fast`. 

So indeed the changes seem to have had an impact on the test case.

If I make the following changes to the test (see lines with `// Before: 'Test'`), it passes, but I am not sure if that's the right move here.

```typescript
test('Reference filter', async () => {
  await setup({
    questionnaire: {
      resourceType: 'Questionnaire',
      status: 'active',
      id: 'reference-filter',
      item: [
        {
          linkId: 'q1',
          type: 'reference',
          text: 'Question',
          extension: [
            {
              url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-referenceResource',
              valueCode: 'Observation',
            },
            {
              url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-referenceFilter',
              valueString: 'subject=$subj',
            },
          ],
        },
      ],
    },
    subject: { reference: 'Patient/123' },
    onSubmit: jest.fn(),
  });

  // Add a spy on medplum.searchResources
  const searchResources = jest.spyOn(medplum, 'searchResources');

  // Get the search input
  const input = screen.getByRole('searchbox') as HTMLInputElement;

  // Enter "Simpson"
  await act(async () => {
    fireEvent.change(input, { target: { value: 'Test 1' } }); // Before: 'Test'
  });

  // Wait for the drop down
  await act(async () => {
    jest.advanceTimersByTime(1000);
  });

  expect(screen.getByText('Test 1')).toBeDefined();
  expect(searchResources).toHaveBeenCalledTimes(1);
  expect(searchResources.mock.calls[0][0]).toBe('Observation');
  expect(searchResources.mock.calls[0][1]).toBeInstanceOf(URLSearchParams);

  const params = searchResources.mock.calls[0][1] as URLSearchParams;
  expect(params.get('subject')).toBe('Patient/123');
  expect(params.get('code')).toBe('Test 1'); // Before: 'Test'
});
```

codyebberson:
It looks like that test was relying on the incorrect behavior.  Fix here: https://github.com/medplum/medplum/pull/4516/commits/e9c122ba4778553b5d35aeac603193390e56158d

pamella:
Thank you for the fix!

codyebberson (approved):

codyebberson:
Thanks @pamella !

PR #4520, opened 2024-05-06
fix(agent): return non-200 status when > 0 failures in bulk op

ThatOneBro:


packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -121,9 +123,13 @@ export async function handleBulkAgentOperation(
           agents[i]
         )
       );
+      failureCount += 1;
       continue;
     }
     const [outcome, params] = result.value;
+    if (!isOk(outcome)) {
+      failureCount += 1;
```
codyebberson, line 131: I recommend we save the `outcome` here and use it to determine status.

`++`?

packages/server/src/fhir/operations/agentbulkstatus.test.ts
```diff
@@ -233,7 +233,7 @@ describe('Agent/$bulk-status', () => {
       .get('/fhir/R4/Agent/$bulk-status')
       .query({ name: 'Test Agent 2' })
       .set('Authorization', 'Bearer ' + accessToken);
-    expect(res.status).toBe(200);
+    expect(res.status).toBe(500);
```
codyebberson, line 236: We should never intentionally return a 500 - that should be reserved for unexpected errors.

packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -132,7 +138,7 @@ export async function handleBulkAgentOperation(
   }
 
   return [
-    allOk,
+    failureCount === 0 ? allOk : serverError(new Error(`${failureCount} sub-operations failed`)),
```
codyebberson, line 141: Use the saved `outcome` here rather than `serverError`

codyebberson (changes requested):

packages/server/src/fhir/operations/utils/agentutils.ts
ThatOneBro: It's possible that you get multiple different types of outcomes, should we just save the first one?

PR #4523, opened 2024-05-06
feat(agent): respect `Agent.status` and `Agent.channel.endpoint.status` being `off`

### LINKED ISSUE #4459: Setting an Endpoint status to `off` does not disable the incoming data
Changing the Endpoint.status from `active` to `off` does not disable the Endpoint from accepting data. This is true even when the Agent service is restarted.

Expected that the Endpoint would not accept data on the defined port and/or the Medplum server (if the Agent has gotten the updated Endpoint configuration) with some TimeoutError. Data should not be sent to the channel target for processing.

### LINKED ISSUE #4460: Setting the Agent status to `off` does not disable the Agent
Similar to #4459, setting the Agent.status to `off` does not stop the Agent from receiving and processing inbound data or sending outbound data. Even after restarting the Agent service data is still processed.

ThatOneBro:
Fixes #4460
Fixes #4459

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -1,4 +1,5 @@
 import {
+  AgentError,
   AgentMessage,
   AgentReloadConfigResponse,
   AgentTransmitRequest,
@@ -158,17 +159,7 @@ export class App {
           case 'transmit':
           case 'agent:transmit:response':
             if (this.config?.status !== 'active') {
-              const err = new Error('Agent.status is currently set to off');
-              this.log.error(normalizeErrorString(err));
-              this.addToWebSocketQueue({
-                type: 'agent:transmit:response',
-                channel: command.channel,
-                remote: command.remote,
-                callback: command.callback,
-                contentType: ContentType.TEXT,
-                statusCode: 400,
-                body: normalizeErrorString(err),
-              } satisfies AgentTransmitResponse);
+              this.sendAgentDisabledError(command);
             } else {
               this.addToHl7Queue(command);
             }
@@ -177,17 +168,7 @@ export class App {
           case 'push':
           case 'agent:transmit:request':
             if (this.config?.status !== 'active') {
-              const err = new Error('Agent.status is currently set to off');
-              this.log.error(normalizeErrorString(err));
-              this.addToWebSocketQueue({
-                type: 'agent:transmit:response',
-                channel: command.channel,
-                remote: command.remote,
-                callback: command.callback,
-                contentType: ContentType.TEXT,
-                statusCode: 400,
-                body: normalizeErrorString(err),
-              } satisfies AgentTransmitResponse);
+              this.sendAgentDisabledError(command);
             } else if (command.contentType === ContentType.PING) {
               await this.tryPingHost(command);
             } else {
@@ -518,6 +499,16 @@ export class App {
       message.accessToken = this.medplum.getAccessToken() as string;
     }
     this.webSocket.send(JSON.stringify(message));
+  }
+
+  private sendAgentDisabledError(command: AgentTransmitRequest | AgentTransmitResponse): void {
+    const errMsg = 'Agent.status is currently set to off';
+    this.log.error(errMsg);
+    this.addToWebSocketQueue({
+      type: 'agent:error',
+      callback: command.callback,
+      body: errMsg,
+    } satisfies AgentError);
   }
 
   private pushMessage(message: AgentTransmitRequest): void {
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -18,7 +18,7 @@ import { ExecException, ExecOptions, exec } from 'node:child_process';
 import { isIPv4, isIPv6 } from 'node:net';
 import { platform } from 'node:os';
 import WebSocket from 'ws';
-import { Channel } from './channel';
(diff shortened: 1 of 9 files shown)


packages/agent/src/app.ts
```diff
@@ -156,12 +157,38 @@ export class App {
           // @ts-expect-error - Deprecated message type
           case 'transmit':
           case 'agent:transmit:response':
-            this.addToHl7Queue(command);
+            if (this.config?.status !== 'active') {
+              const err = new Error('Agent.status is currently set to off');
```
codyebberson, line 161: Rather than creating a `new Error()` you can just use the string in both `this.log.error()` and `body:`

packages/agent/src/app.ts
```diff
@@ -156,12 +157,38 @@ export class App {
...
+              this.log.error(normalizeErrorString(err));
+              this.addToWebSocketQueue({
+                type: 'agent:transmit:response',
```
codyebberson, line 164: This should probably be `type: 'agent:error'`, no?

packages/agent/src/app.ts
```diff
@@ -156,12 +157,38 @@ export class App {
...
+            } else {
+              this.addToHl7Queue(command);
+            }
             break;
           // @ts-expect-error - Deprecated message type
           case 'push':
           case 'agent:transmit:request':
-            if (command.contentType === ContentType.PING) {
+            if (this.config?.status !== 'active') {
+              const err = new Error('Agent.status is currently set to off');
```
codyebberson, line 180: Given similarity to `case 'agent:transmit:response`, would it be possible to move this to helper function?

packages/agent/src/app.ts
```diff
@@ -209,10 +244,28 @@ export class App {
     const endpoints = await Promise.all(endpointPromises);
     this.validateAgentEndpoints(channels, endpoints);
 
-    // Remove all channels from list that are present in the new definition
-    // We will remove the channels that are left over -- channels that are not part of the new config
-    for (const definition of channels) {
-      pendingRemoval.delete(definition.name);
+    const filteredChannels = [] as AgentChannel[];
```
codyebberson, line 247: Doesn't look like `filteredChannels` or `filteredEndpoints` is used anywhere?

packages/agent/src/net-utils.test.ts
```diff
@@ -197,7 +199,7 @@ describe('Agent Net Utils', () => {
       await expect(messageReceived).resolves.toMatchObject<Partial<AgentTransmitResponse>>({
         type: 'agent:transmit:response',
         contentType: ContentType.TEXT,
-        statusCode: 500,
+        statusCode: 400,
```
codyebberson, line 202: Nice, ty 🙏

packages/hl7/src/client.ts
```diff
@@ -1,5 +1,5 @@
 import { Hl7Message } from '@medplum/core';
-import { connect } from 'net';
+import { connect } from 'node:net';
```
codyebberson, line 2: 👍

packages/hl7/src/client.ts
```diff
@@ -26,11 +26,18 @@ export class Hl7Client extends Hl7Base {
...
 
-    return new Promise((resolve) => {
+    return new Promise((resolve, reject) => {
+      const connectErrorListener = (err: Error): void => {
+        reject(err);
+      };
+
       const socket = connect({ host: this.host, port: this.port }, () => {
         this.connection = new Hl7Connection(socket);
+        socket.off('error', connectErrorListener);
```
codyebberson, line 36: Could we just pass in `reject` directly here?

```suggestion
        socket.off('error', reject);
```

packages/hl7/src/client.ts
```diff
@@ -26,11 +26,18 @@ export class Hl7Client extends Hl7Base {
...
         resolve(this.connection);
       });
+
+      socket.on('error', connectErrorListener);
```
codyebberson, line 40: And here?

```suggestion
      socket.on('error', reject);
```

packages/server/src/fhir/operations/agentpush.ts
```diff
@@ -114,3 +121,17 @@ async function pushToAgent(req: Request): Promise<[OperationOutcome] | [Operatio
   // 1. The agent will respond with a message on the channel
   // 2. The timer will expire and the request will timeout
 }
+
+function validateParams(params: AgentPushParameters): void {
```
codyebberson, line 125: Mental note, as discussed offline, in the future these checks should be handled by `parseParameters()` for required fields.

codyebberson (commented):
Nice, thanks @ThatOneBro ! 🙏 

As discussed offline, in Medplum v4, we should review the value set for `Agent.status`.  I think this looks good for now.

Bunch of minor nitpicky comments.  Overall structure looks good to me 👍

packages/agent/src/app.ts
ThatOneBro: Ah you're right. I think there's technically a bug here, and we should be using the filtered arrays instead of only filtering channels. Going to add a test for multiple channels to catch this

packages/agent/src/app.ts
ThatOneBro: Probably yes you're right

packages/agent/src/app.ts
ThatOneBro: Done 👍

Issue #4409, opened 2024-04-19
Allow Project Admins to configure refresh token expiry


rahul1:
Different apps will have different requirements around how long a refresh token should last. We should allow project admins to configure the expiry time for their refresh tokens. 

**Open question: ** Should this be a project-level setting, or a setting on the `ClientApplication`?

reshmakh:
Patient auth and provider auth often have different requirements both from the regs (d1 vs g10) and on the practice policy side - so `ClientApplication` seems pragmatic for that reason.

rahul1:
@codyebberson mentioned that it could also make sense for this to live on the `AccessPolicy`

ThatOneBro:
Closed as completed via #4525

PR #4525, opened 2024-05-07
Feat: Configure refresh token expiry

thejwuscript:
Related to issue #4409

This PR allows project admins to optionally configure the expiry of a refresh token for a `ClientApplication`.

A `refreshTokenExpiry` field is added to the `ClientApplication` resource. The field is optional to allow backward compatibility. The value is used to generate the `exp` claim on a refresh token.

UI changes to the Medplum App is not addressed in this PR; it can be included here if necessary, or be made on a separate PR.


packages/definitions/dist/fhir/r4/profiles-medplum.json
```diff
@@ -823,6 +823,21 @@
...
             }
+          },
+          {
+            "id" : "ClientApplication.refreshTokenExpiry",
+            "path" : "ClientApplication.refreshTokenExpiry",
+            "definition" : "Optional configuration to set the refresh token expiry",
+            "min" : 0,
+            "max" : "1",
+            "type" : [{
+              "code" : "date"
```
josh-, line 834: for our use case we'd like to specify a value here which is relative to the current time, not a fixed date parameter.

the expiration value is passed to the [`jose` npm package](https://github.com/panva/jose), which treats `number` and `Date` values as the fixed expiration time in the claim ([reference here](https://github.com/panva/jose/blob/4261556a123ae2dc5c5f238465eff7eb9404b293/src/jwt/produce.ts#L103-L105)), however `string` values are converted to a time span relative to the current time ([reference here](https://github.com/panva/jose/blob/4261556a123ae2dc5c5f238465eff7eb9404b293/src/jwt/produce.ts#L106-L115)) - which matches the existing behaviour with the `2w` default.

---

i think ideally this property would be a `string` which could either support the [full range of jose formats](https://github.com/panva/jose/blob/4261556a123ae2dc5c5f238465eff7eb9404b293/src/jwt/produce.ts#L109-L115), or perhaps a subset of this (to allow for easier validation in the medplum middleware)

ThatOneBro: Yes I think if we can change the name of this field to `refreshTokenLifetime` and make it an expiry duration as a string (eg. '2w', '1h', etc.) that maps to `jose` formats as @josh- pointed out, that would be ideal 👍

rahul1: @thejwuscript  - just wanted to make sure you saw the above comment from @ThatOneBro . This is our main feedback on the PR

thejwuscript: Yes, thanks very much for the feedback. I'll be making the changes.

josh-: @thejwuscript do you know when you'll be able to make those changes? let me know if I can help. eager to be able to use this.

thejwuscript: @josh- I'll try get it done by Monday.

packages/server/src/fhir/routes.ts
```diff
@@ -117,6 +118,9 @@ protectedRoutes.use('/bulkdata', bulkDataRouter);
 // Async Job
 protectedRoutes.use('/job', jobRouter);
 
+// Validate data before updating ClientApplication
+protectedRoutes.put('/ClientApplication/:id', saveClientValidator);
```
ThatOneBro, line 122: I'm thinking this probably should not be directly in an `express` route, should probably be done at the `FhirRouter` level potentially? What do you think, @mattwiller?

I know we have talked about doing something like this across the board long term (having specific validation on a per resource basis, maybe via some "prewrite" pass

That's definitely out of scope for this PR but would be nice to move in a more uniform direction

ThatOneBro (commented):
Looks good for the most part! Consulting with @mattwiller about where the validation should occur. Once we figure that out we should be good to go! Thanks again for your help @thejwuscript! 😄

packages/server/src/fhir/routes.ts
mattwiller: I would actually propose moving this validation into the field definition itself, using a FHIRPath condition like this:
```json
 {
  "id" : "ClientApplication.refreshTokenLifetime",
  "path" : "ClientApplication.refreshTokenLifetime",
  "definition" : "Optional configuration to set the refresh token duration",
  "min" : 0,
  "max" : "1",
  "type" : [{
    "code" : "string"
  }],
  "constraint" : [{
    "key" : "clapp-1",
    "severity" : "error",
    "human" : "Token lifetime must be a valid string representing time duration (eg. 2w, 1h)",
    "expression" : "$this.matches('^[0-9]+[smhdwy]$')"
  }],
  "base": {
    "path" : "ClientApplication.refreshTokenLifetime",
    "min" : 0,
    "max" : "1"
  }
}
```
That way, this validation gets performed at the `Repository` level, any time the resource is updated anywhere in the system.

ThatOneBro: Yeah I actually agree, this is probably the right way to do it

ThatOneBro:
Okay I pushed a tweak applying @mattwiller's suggestion and some tests for some specific cases, after someone else approves we can push this through. Thank you @thejwuscript for taking this issue 🙏

packages/server/src/admin/client.ts
```diff
@@ -16,6 +16,10 @@ import { makeValidationMiddleware } from '../util/validator';
 
 export const createClientValidator = makeValidationMiddleware([
   body('name').notEmpty().withMessage('Client name is required'),
+  body('refreshTokenLifetime')
```
rahul1, line 19: @ThatOneBro - do we need this, or can we leave it to resource schema validation

packages/server/src/oauth/keys.ts
```diff
@@ -205,10 +205,13 @@ export function generateAccessToken(
 /**
  * Generates a refresh token JWT.
  * @param claims - The refresh token claims.
+ * @param refreshLifetime - The refresh token duration.
  * @returns A well-formed JWT that can be used as a refresh token.
  */
-export function generateRefreshToken(claims: MedplumRefreshTokenClaims): Promise<string> {
-  return generateJwt('2w', claims);
+export function generateRefreshToken(claims: MedplumRefreshTokenClaims, refreshLifetime?: string): Promise<string> {
+  const duration = refreshLifetime ?? '2w';
```
rahul1, line 212: Think it's worth pulling out the `2w` into a top-level constant for readability

rahul1 (approved):

PR #4530, opened 2024-05-07
Fixes #4508 - MeasureReport-subject search param backport

### LINKED ISSUE #4508: Backport MeasureReport.subject from R5
## Use Case
In many VBC use cases, quality measures (e.g. cost score, risk score, quality score) are computed not only on `Practitioners`, but on the `Organization` level (e.g. Hospital cost score). 

In R4 and R4b,  `MeasureReport.subject` can point to a `Patient` and `Practitioner` (amongst other resourceTypes), but not to a `Organization`. This was rectified in R5

## Proposal
- Backport the choice of types for `MeasureReport.subject` from R5 -> R4. I acknowledge that this might be controversial, because we (rightly) want to avoid creating a chimera of mixed R4/R5 models - so I'm open to alternatives. 

## Alternatives Considered
- **Using `Observation` instead of `MeasureReport`:** This has the same problem `MeasureReport` - `subject` can be an `Organization` in R5, but not R4
- **Setting `MeasureReport.subject` to a `Location`** : This breaks the semantics of `Location` - In general an organization (e.g. hospital system) might have 2+ locations (facilities, buildings). So associating the the `MeasureReport` with a `Location` doesn't have the right cardinality, unless we enforce making a "dummy" `Location`

codyebberson:
R4: https://hl7.org/fhir/r4/measurereport.html

R5: https://hl7.org/fhir/R5/measurereport.html

### DIFF
--- a/packages/definitions/dist/fhir/r4/search-parameters.json
+++ b/packages/definitions/dist/fhir/r4/search-parameters.json
@@ -39203,8 +39203,11 @@
       "xpathUsage" : "normal",
       "target" : ["Practitioner",
       "Group",
+      "Organization",
+      "CareTeam",
       "Device",
       "Patient",
+      "HealthcareService",
       "PractitionerRole",
       "RelatedPerson",
       "Location"]
--- a/packages/fhirtypes/dist/MeasureReport.d.ts
+++ b/packages/fhirtypes/dist/MeasureReport.d.ts
@@ -3,10 +3,12 @@
  * Do not edit manually.
  */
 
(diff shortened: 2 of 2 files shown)


reshmakh (approved):

rahul1 (changes requested):
@codyebberson The original ticket was about changing the reference type, so I think you'll need to adjust the StructureDefinition as well

rahul1 (commented):
@codyebberson The original ticket was about changing the reference type, so I think you'll need to adjust the StructureDefinition as well

PR #4535, opened 2024-05-09
`PatientSummary` Problem List uses US Core profile

mattlong:
A notable underlying change to facilitate this:

In order to get discriminators with a path of `$this` working, `SliceDefinition` now extends `InternalSchemaElement`. Previously they already had an awful lot of overlap, so this feels like a low risk change to make. The redundancy of `SliceDefinition.definition` and `InternalSchemaElement.description` is unfortunate. We could get rid of `SliceDefinition.definition` and normalize on `description`, but that would of course be a breaking change which would probably be best to leave for the next major release. If/when this PR gets merged, I'll make an issue to track that.


packages/core/src/typeschema/validation.ts
```diff
@@ -589,7 +594,19 @@ export function matchDiscriminant(
       if (!value || !sliceElement) {
         return false;
       }
-      if (matchesSpecifiedValue(value, sliceElement)) {
+      if (sliceElement.pattern /*&& deepIncludes(value, sliceElement.pattern)*/) {
```
mattlong, line 597: This is a subtle but potentially breaking validation change for malformed slice definitions. Since `matchesSpecifiedValue` returns `true` by default, checks on elements not containing either `.fixed` nor `.pattern` now return `false` whereas they used to return `true`. The only time fixed and pattern shouldn't be defined is when we're dealing with slicing by ValueSet which is handled below and retains the permissiveness for now.

packages/core/src/typeschema/validation.ts
```diff
@@ -589,7 +594,19 @@ export function matchDiscriminant(
...
+        // return true;
+        return deepIncludes(value, sliceElement.pattern);
+      }
+      if (sliceElement.fixed /*&& deepEquals(value, sliceElement.fixed)*/) {
+        // return true;
+        return deepEquals(value, sliceElement.fixed);
+      }
+
+      if (sliceElement.binding?.strength === 'required' && sliceElement.binding?.valueSet) {
```
mattlong, line 606: @mattwiller Based on our offline conversation, I believe this if statement should be changed to:

```typescript
if (sliceElement.binding?.valueSet)
```

mattwiller: I think the `required` strength is also necessary here: we don't want to validate optional bindings since technically any other value would also be acceptable

packages/core/src/typeschema/validation.ts
```diff
@@ -589,7 +594,17 @@ export function matchDiscriminant(
       if (!value || !sliceElement) {
         return false;
       }
-      if (matchesSpecifiedValue(value, sliceElement)) {
+      if (sliceElement.pattern) {
```
mattlong, line 597: This is a subtle but potentially breaking validation change for malformed slice definitions. Since `matchesSpecifiedValue` returns `true` by default, checks on elements not containing either `.fixed` nor `.pattern` now return `false` whereas they used to return `true`. The only time fixed and pattern shouldn't be defined is when we're dealing with slicing by ValueSet which is handled below and retains the permissiveness for now.

packages/core/src/typeschema/types.ts
```diff
@@ -66,15 +66,14 @@ export interface SlicingRules {
   slices: SliceDefinition[];
 }
 
-export interface SliceDefinition {
+export interface SliceDefinition extends InternalSchemaElement {
+  // narrowed InternalSchemaElement fields
+  slicing?: never;
```
codyebberson, line 71: Would this be more accurate? 

```ts
export interface SliceDefinition extends Omit<InternalSchemaElement, 'slicing'> {
```

packages/core/src/typeschema/validation.ts
```diff
@@ -580,7 +580,12 @@ export function matchDiscriminant(
     return false;
   }
 
-  const sliceElement: InternalSchemaElement = (elements ?? slice.elements)[discriminator.path];
+  let sliceElement: InternalSchemaElement | undefined;
+  if (discriminator.path === '$this') {
```
codyebberson, line 584: Clarifying question:  My understanding is that `discriminator.path` can be a full FHIRPath expression, so we're currently adding support for the special case of `=== '$this'`, but still have open future work here?

packages/core/src/typeschema/validation.ts
codyebberson: @mattwiller  - can you comment on this?  Seems ok to me, but this is kinda out of my comfort zone.

codyebberson (commented):
lgtm.  I think we should wait for @mattwiller  to clarify the `matchesSpecifiedValue()` change

packages/core/src/typeschema/validation.ts
mattwiller: Yep, Matt and I discussed this and I think this change should be fine.  In general, the possibility of malformed profile `StructureDefinition` resources makes validation tricky unless we assume the underlying profile is valid.  Eventually, we may want to add explicit validation checks on the profile before using it for validation of another resource: at least then we could abort and display the issue with the profile

packages/core/src/typeschema/validation.ts
mattlong: Yes, that's correct; the [Restricted Subset ("Simple")](https://hl7.org/fhir/R4/fhirpath.html#simple):

* All statements SHALL start with the name of the context element (e.g. on a Patient resource, Patient.contact.name.), or SHALL be simply "$this" to refer to the element that has focus
* Operators SHALL NOT be used
* Only the following functions may be used:
  * .resolve()
  * .extension("url")
  * .ofType(type)
* All other functions SHALL NOT be used

codyebberson (approved):

PR #4540, opened 2024-05-11
feat(Subscription): add `author` as a `SearchParameter`

### LINKED ISSUE #4385: Access Policies for Websocket Subscriptions
## Problem
- Patient user creates a websocket `Subscription` resource to recieve incoming chats
- Admin would like to create an access policy to prevent Patient user from manipulating _other_ `Subscription` resources
- The following policy won't work, because `Subscriptions` are not added to a `Patient` compartment
```
    {
      resourceType: 'Subscription',
      criteria: 'Subscription?type=websocket&_compartment=%profile',
    }
```

## Possible Solutions
- Add an `author` search param (only to `Subscriptions`) to every resource that points to `meta.author`. Then we can restrict access as such: 
```
    {
      resourceType: 'Subscription',
      criteria: 'Subscription?type=websocket&author=%profile',
    }
```
- Add Subscriptions to the `Patient` compartment (we would have to handle the case of `Practitioner` users seperately)
- ???

ThatOneBro:
Fixes #4385


packages/definitions/dist/fhir/r4/search-parameters-medplum.json
```diff
@@ -693,6 +693,24 @@
...
+        "version": "4.0.1",
+        "name": "author",
+        "status": "draft",
+        "publisher": "Medplum",
+        "description": "The author of the Subscription resource",
+        "code": "author",
+        "base": ["Subscription"],
+        "type": "reference",
+        "expression": "Subscription.meta.author",
+        "target": ["Patient", "Practitioner", "RelatedPerson", "ClientApplication", "Bot"]
```
ThatOneBro, line 712: Not sure if this is the right enum for all the possible `author` resource types, feedback would be appreciated

packages/server/src/migrations/schema/v69.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Generated by @medplum/generator
+ * Do not edit manually.
+ */
+
+import { PoolClient } from 'pg';
+
+export async function run(client: PoolClient): Promise<void> {
+  await client.query('ALTER TABLE IF EXISTS "Subscription" ADD COLUMN IF NOT EXISTS "author" TEXT');
```
codyebberson, line 9: Hm, this doesn't seem right... I think this should have used the `Subscription_References` table, not added a new column 🤔 

This may be a bug in the migration generator, due to our custom `Meta` types 🤔

mattwiller: I don't think the migration generator has been updated to remove the `Reference` columns yet; it's probably good to keep both until we fully deprecate the old columns altogether

packages/definitions/dist/fhir/r4/search-parameters-medplum.json
codyebberson: That looks correct to me.  

The "FHIR"-y way to manage that would probably be to have a proper `CodeSystem` and `ValueSet`.  Low pri, future work.

PR #4542, opened 2024-05-11
fix(agent): unwrap response for `$reload-config` by id

### LINKED ISSUE #4541: Calling `Agent/:id/$reload-config` should not return a bundle
We should establish a pattern that the bulk version of operations have a distinct payload from the single instance-level versions.

ThatOneBro:
Fixes #4541 

Later we may want to consider moving the bulk version of all endpoints to something like a `$bulk` operation which takes an `op` query parameter.

For example, `Agent/$bulk?op=reload-config&name:contains=Medplum` would perform a bulk reload-config on Agents with names containing `Medplum`...

I'll open a discussion about this later, but this PR should at least move in the right direction for now.

### DIFF
--- a/packages/core/src/outcomes.ts
+++ b/packages/core/src/outcomes.ts
@@ -181,7 +181,7 @@ export function badRequest(details: string, expression?: string): OperationOutco
         details: {
           text: details,
         },
-        expression: expression ? [expression] : undefined,
+        ...(expression ? { expression: [expression] } : undefined),
       },
     ],
   };
--- a/packages/server/src/fhir/operations/agentreloadconfig.ts
+++ b/packages/server/src/fhir/operations/agentreloadconfig.ts
@@ -1,4 +1,4 @@
-import { AgentReloadConfigResponse, OperationOutcomeError, serverError } from '@medplum/core';
+import { AgentReloadConfigResponse, OperationOutcomeError, badRequest, serverError } from '@medplum/core';
 import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
 import { Agent, OperationDefinition } from '@medplum/fhirtypes';
 import { handleBulkAgentOperation, publishAgentRequest } from './utils/agentutils';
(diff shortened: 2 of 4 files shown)


packages/core/src/outcomes.ts
```diff
@@ -181,7 +181,7 @@ export function badRequest(details: string, expression?: string): OperationOutco
         details: {
           text: details,
         },
-        expression: expression ? [expression] : undefined,
+        ...(expression ? { expression: [expression] } : undefined),
```
codyebberson, line 184: 🤔 

I suppose this avoids setting `expression: undefined`, but does that change anything in practice?

packages/server/src/fhir/operations/agentreloadconfig.ts
```diff
@@ -47,7 +47,7 @@ async function reloadConfig(agent: Agent): Promise<FhirResponse> {
   }
 
   if (result.type === 'agent:error') {
-    throw new OperationOutcomeError(serverError(new Error(result.body)));
+    throw new OperationOutcomeError(badRequest(result.body));
```
codyebberson, line 50: 👍

packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -108,6 +108,10 @@ export async function handleBulkAgentOperation(
     return [badRequest('No agent(s) for given query')];
   }
 
+  if (req.params.id) {
```
codyebberson, line 111: 👍

codyebberson (approved):

packages/core/src/outcomes.ts
ThatOneBro: Yes, when I tried to compare a `badRequest` from over the wire with a `badRequest` instantiated, the strict equals failed since the newly created one had the extra `expression` key

ThatOneBro: Since we remove keys with `undefined` values when sending over the wire

PR #4544, opened 2024-05-13
Dependency upgrades 2024-05-13

medplumbot:
Dependency upgrades


packages/expo-polyfills/package.json
```diff
@@ -49,21 +49,21 @@
...
+    "rimraf": "5.0.7",
     "ts-jest": "29.1.2"
   },
   "peerDependencies": {
     "@medplum/core": "3.1.5",
     "expo": "*",
-    "expo-crypto": "^12.6.0",
-    "expo-secure-store": "^12.3.1",
-    "expo-standard-web-crypto": "^1.7.0",
+    "expo-crypto": "^13.0.0",
```
ThatOneBro, line 64: Not convinced our test coverage is enough to prove these are working as intended, we'll probably want to manually test on real mobile devices before merging this

PR #4558, opened 2024-05-15
Optimize database reindex

mattwiller:
In testing on localhost over resource types with 15k – 100k resources in the table, this combination of transaction batching and concurrent handling of the lookup tables reduced the overall full reindex time by 30–50% (saving on the order of several minutes for my small data set).

This PR also moves the reindex logic into an async worker that processes a batch of rows and then enqueues another job to handle the next chunk.  This makes the job much more robust, and allows it to more safely run for extended periods of time.


packages/server/src/fhir/repo.ts
```diff
@@ -853,37 +865,42 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
-
-      const bundle = await this.search(searchRequest);
-      if (bundle.entry) {
-        for (const entry of bundle.entry) {
-          const resource = entry.resource as T;
-          await callback(resource);
-
-          const lastUpdated = resource.meta?.lastUpdated as string;
-          currentTimestamp = lastUpdated;
+      await this.withTransaction(async (conn) => {
```
codyebberson, line 890: Is this transaction necessary?  Trying to imagine what it's needed for, versus letting the callback handle it.

codyebberson: Hm, I see, you're using this to avoid per-resource transactions in the reindex step.

imho, that kinda breaks the abstraction of `forAllResources`

1. I'm not sure the transaction is fully necessary in `reindexResource`
2. If it is both necessary and if this is a major performance win, then i think we should probably modify the API contract of `forAllResources` (maybe a "page" callback and a "resource" callback?)

codyebberson (commented):
Overall looks good, nice perf wins 🎉 

See questions on `forAllResources` and transactions

packages/server/src/fhir/repo.ts
```diff
@@ -852,16 +858,22 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
-    await this.forAllResources(resourceType, async (resource) => {
-      try {
-        await this.reindexResourceImpl(resource);
-      } catch (err) {
-        getLogger().error('Failed to reindex resource', { error: normalizeErrorString(err) });
-      }
+    const startTime = Date.now();
+    let count = 0;
+    await this.forAllResources(resourceType, async (conn, resources) => {
+      await this.reindexResources(conn, ...resources);
```
mattlong, line 870: Regarding @codyebberson comments about the transaction in `forAllResources`, could the transaction be moved to wrap this line?

mattwiller: That line is logically inside the transaction, with starts inside `forAllResources()` before the callback is invoked and also includes the search that returns `resources`

codyebberson (approved):
nice 👍

PR #4559, opened 2024-05-15
Added resource validation to MockClient

rahul1:
This PR adds a call to `validateResource` on the `MemoryRepo` class, primarily used by MockClient. 

This actually discovered a bunch of invalid resources in our preloaded `MockClient` resources, as well as a bug in AllergyDialog.tsx

Unfortunately, it also meant that we had to revert some of the work from #4286, as this compression created invalid StructureDefinitions that wouldn't validate


packages/agent/src/app.test.ts
```diff
@@ -155,60 +161,6 @@ describe('App', () => {
     mockServer2.stop();
   });
 
-  test('Empty endpoint URL', async () => {
```
rahul1, line 158: Removed this test b/c the resource validator prevents Endpoints w/ empty addresses from being created

rahul1:
Closed due to inactivity

PR #4570, opened 2024-05-17
Apply filter to ValueSet.expansion.contains

mattlong:
The expand operator now applies the `filter` parameter on a `ValueSet` with `expansion.contains` populated instead of `compose.include`.

### DIFF
--- a/packages/server/src/fhir/operations/expand.ts
+++ b/packages/server/src/fhir/operations/expand.ts
@@ -1,6 +1,6 @@
 import { allOk, badRequest, OperationOutcomeError } from '@medplum/core';
 import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
-import { CodeSystem, ValueSet, ValueSetComposeInclude, ValueSetExpansionContains } from '@medplum/fhirtypes';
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/operations/expand.ts
```diff
@@ -244,9 +252,10 @@ async function computeExpansion(
     const codeSystem = codeSystemCache[include.system] ?? (await findTerminologyResource('CodeSystem', include.system));
     codeSystemCache[include.system] = codeSystem;
     if (include.concept) {
-      const concepts = await Promise.all(include.concept.flatMap((c) => validateCoding(codeSystem, c)));
-      for (const c of concepts) {
-        if (c && (!filter || c.display?.includes(filter))) {
+      const filteredCodings = filterCodings(include.concept, filter);
+      const validCodings = await Promise.all(filteredCodings.map((c) => validateCoding(codeSystem, c)));
```
codyebberson, line 256: 👍 

Aside: we should really prioritize optimizing `validateCoding()` - low hanging fruit here.

codyebberson (approved):

PR #4573, opened 2024-05-17
More efficiently validate included concepts

mattlong:


packages/server/src/fhir/operations/codesystemvalidatecode.ts
```diff
@@ -62,19 +62,38 @@ export async function codeSystemValidateCodeHandler(req: FhirRequest): Promise<F
...
+      eligible[i] = true;
+    }
   }
 
-  const query = new SelectQuery('Coding')
-    .column('id')
-    .column('display')
-    .where('code', '=', coding.code)
-    .where('system', '=', codeSystem.id);
+  let result: any[] | undefined;
```
codyebberson, line 82: You could use `Coding` rather than `any` here, because it has `id`, `code`, and `display` as optional string fields.

packages/server/src/fhir/operations/codesystemvalidatecode.ts
```diff
@@ -62,19 +62,38 @@ export async function codeSystemValidateCodeHandler(req: FhirRequest): Promise<F
...
+    result = await query.execute(db);
+  }
 
-  const db = getDatabasePool();
-  const result = await query.execute(db);
-  return result.length
-    ? { id: result[0].id, system: codeSystem.url, code: coding.code, display: result[0].display }
-    : undefined;
+  return codings.map((c, idx) => {
+    const row = eligible[idx] && result?.find((r: any) => r.code === c.code);
```
codyebberson, line 96: I believe if you use `Coding` on line 82, you could remove the type from `(r)` entirely.

codyebberson (approved):
Nice, looks great.

Minor suggestions on `any` vs `Coding`, totally optional.

PR #4575, opened 2024-05-20
fix(client/keyvalue): set keyvalue content-type text

dillonstreator:


codyebberson (approved):

codyebberson:
Thanks @dillonstreator !

PR #4577, opened 2024-05-22
fix(subscriptions): don't retry ws subs if sub is deleted

ThatOneBro:


packages/server/src/workers/subscription.ts
```diff
@@ -446,12 +449,20 @@ export async function execSubscriptionJob(job: Job<SubscriptionJobData>): Promis
...
-    return await systemRepo.readResource<Subscription>('Subscription', subscriptionId);
+    return await systemRepo.readResource<Subscription>('Subscription', subscriptionId, {
+      checkCacheOnly: channelType === 'websocket',
+    });
   } catch (err) {
     const outcome = normalizeOperationOutcome(err);
-    if (isGone(outcome)) {
+    // If the Subscription was marked as deleted in the database, this will return "gone"
+    // However, deleted WebSocket subscriptions will return "not found"
+    if (isGone(outcome) || isNotFound(outcome)) {
```
codyebberson, line 465: good catch 👍 

This would have also applied to "purged" or "expunged" subscriptions, which are hard deleted from the database.

codyebberson (approved):
Nice, clean.  

All of the `T['resourceType']` cleanup is nice too 👌

PR #4583, opened 2024-05-22
Use nested transactions with stronger isolation

mattwiller:
This PR un-reverts #3866 and adds some fixes, plus hardcodes the transaction isolation level to `repeatable read`, which prevents against concurrent modifications.

We should merge this ahead of both #4358 and #4558


packages/server/src/fhir/repo.ts
```diff
@@ -1906,32 +1907,58 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
+      this.endTransaction();
     }
   }
 
   close(): void {
     this.assertNotClosed();
+    this.releaseConnection();
+    if (this.transactionDepth > 0) {
+      getRequestContext().logger.error('Closing Repository with active transaction');
+    }
```
codyebberson, line 1961: Minor suggestion, probably paranoia, for your consideration

```suggestion
    if (this.transactionDepth > 0) {
      // Bad state, remove connection from pool
      getRequestContext().logger.error('Closing Repository with active transaction');
      this.releaseConnection(new Error('Closing Repository with active transaction'));
    } else {
      // Good state, return healthy connection to pool
      this.releaseConnection();
    }
```

codyebberson (approved):
lgtm 👍 

I added a small suggestion to `Repository.close()`, w.r.t. purging bad connections.  Your call.

PR #4585, opened 2024-05-23
refactor(subscriptions): skip `bullmq` for WS subs, push all WS subs thru one Redis channel

ThatOneBro:


packages/server/src/subscriptions/websockets.ts
```diff
@@ -22,12 +22,86 @@ export interface BindWithTokenMsg extends BaseSubscriptionClientMsg {
...
+    heartbeatHandler = (): void => {
+      for (const [ws, subscriptionIds] of wsToSubLookup.entries()) {
+        ws.send(JSON.stringify(createSubHeartbeatEvent(subscriptionIds)));
+      }
+    };
+    heartbeat.addEventListener('heartbeat', heartbeatHandler);
+  }
+}
+
+// function unsubscribeWsFromSubscription(ws: ws.WebSocket, subscriptionId: string): void {
```
codyebberson, line 71: Cleanup

packages/server/src/subscriptions/websockets.ts
```diff
@@ -22,12 +22,86 @@ export interface BindWithTokenMsg extends BaseSubscriptionClientMsg {
...
+      for (const socket of subToWsLookup.get(subscriptionId) ?? []) {
+        socket.send(JSON.stringify(bundle), { binary: false });
+      }
+    }
+  });
+  await redisSubscriber.subscribe('medplum:subscriptions:r4:websockets');
+}
+
+function subscribeWsToSubscription(ws: ws.WebSocket, subscriptionId: string): void {
+  let wsSet = subToWsLookup.get(subscriptionId);
```
codyebberson, line 46: 🤔  Is it allowed to use the same `Subscription` on multiple websocket connections? 🤔

packages/server/src/subscriptions/websockets.ts
```diff
@@ -22,12 +22,86 @@ export interface BindWithTokenMsg extends BaseSubscriptionClientMsg {
...
+    return;
+  }
+  for (const subscriptionId of subscriptionIds) {
+    if (!subToWsLookup.has(subscriptionId)) {
+      globalLogger.error(`[WS] Subscription binding to subscription ${subscriptionId} for this WebSocket is missing`);
+      continue;
+    }
+    const wsSet = subToWsLookup.get(subscriptionId) as Set<ws.WebSocket>;
+    wsSet.delete(ws);
+    if (wsSet.size === 0) {
```
codyebberson, line 95: Clever

codyebberson (approved):
Nice, this is clean.

1. Minor, some commented out code
2. What is the case for multiple websocket connections sharing a `Subscription`?  I guess it's neat that we support it, I just didn't think it was something that would be possible.  Maybe if the Subscription id is in localStorage, and you have multiple tabs open?

packages/server/src/subscriptions/websockets.ts
ThatOneBro: It's technically allowed by the spec, and  Currently our client doesn't support it but it shouldn't be hard to add support for it in the future. I kinda left it open more for other client implementers / to be more FHIR spec-compliant

Issue #4470, opened 2024-04-28
Use AccessPolicy in ResourceForm: hidden fields should be hidden


codyebberson:

mattlong:
Fixed in #4588

PR #4588, opened 2024-05-23
Respect AccessPolicy.resource.{readonlyFields,hiddenFields} in ResourceForm

### LINKED ISSUE #4469: Use AccessPolicy in ResourceForm: readonly fields should not use input elements

mattlong:
`ResourceForm` can now consider the user's `AccessPolicy` and hide or show inputs in a disabled state for `AccessPolicy.resource.hiddenFields`  and `AccessPolicy.resource.readonlyFields` respectively.

For nested elements that have a custom input component, e.g. `<HumanNameInput/>`, we do not attempt to hide `hiddenFields` such as `name.given` since it would complicate managing the layout of the inputs. Instead, the fields are shown in a disabled state as an affordance to the user that they cannot edit/supply a value for those fields.

Note: choice-of-type elements, e.g. `value[x]`, are not technically supported since they are not valid values in `AccessPolicy.resource.readonlyFields`. A followup project to this PR is adding support for specifying choice-of-type elements in access policies at which point they should generally _just work_ in `<ResourceForm />`. Specifying explicitly typed choice-of-type like `valueBoolean` as is currently supported in `AccessPolicy.resource.{readonlyFields,hiddenFields}` will not be reflected correctly in `<ResourceForm />`.

#### Patient form with various fields in readonly
Screenshot 2024-05-23 at 3 41 19 PM

#### Patient form with various fields hidden, e.g. `'identifier', 'name.use', 'name.prefix', 'name.suffix', 'birthDate'`
Screenshot 2024-05-23 at 3 52 12 PM

Fixes #4469 #4470

### DIFF
--- a/packages/core/src/elements-context.test.ts
+++ b/packages/core/src/elements-context.test.ts
@@ -1,4 +1,4 @@
-import { buildElementsContext } from './elements-context';
+import { ExtendedInternalSchemaElement, buildElementsContext } from './elements-context';
 import { HTTP_HL7_ORG } from './constants';
 import { isPopulated } from './utils';
 import {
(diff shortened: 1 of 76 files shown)


packages/core/src/elements-context.test.ts
```diff
@@ -17,6 +26,8 @@ describe('buildElementsContext', () => {
   }
   beforeAll(() => {
     USCoreStructureDefinitions = readJson('fhir/r4/testing/uscore-v5.0.1-structuredefinitions.json');
+    indexStructureDefinitionBundle(readJson('fhir/r4/profiles-resources.json') as Bundle);
+    // indexStructureDefinitionBundle(readJson('fhir/r4/profiles-types.json') as Bundle);
```
codyebberson, line 30: Cleanup 

```suggestion
```

packages/core/src/elements-context.ts
```diff
@@ -107,3 +177,61 @@ function mergeElementsForContext(
...
+    return elements;
+  }
+
+  const prefix = keyPrefix ? keyPrefix + '.' : '';
+  return Object.fromEntries(
+    Object.entries(elements).filter(([key]) => !matchesKeyPrefixes(prefix + key, accessPolicyResource.hiddenFields))
+  );
+}
+
+function markReadonlyFields(
```
codyebberson, line 196: Just tagging this as a pretty important detail of the implementation.  The in-memory schema is now mutable, and a function of the user's access policy.

I'm trying to think through the implications of that...  For example, in a multi profile setting, if I change my profile, and therefore change my access policy, does that invalidate the cache of structure definitions and `ExtendedInternalSchemaElement` values?

codyebberson (approved):
Oof, that looks like it was a slog. Thank you for doing that 🙏 

Structurally looks good to me.

Given the breadth of the PR, we should definitely do a mini bug bash on staging.

The only question I had was regarding `ExtendedInternalSchemaElement` and how that works when switching profiles / switching access policies.

Otherwise, everything lgtm 👍

packages/core/src/elements-context.ts
mattlong: I considered that and should have added a comment/test about it since it is quite important, but the in-memory schema isn't being mutated since the element object is spread when setting `readonly` and the `result` object is a newly create object as well. So unless I'm missing something, this should be fine? I'm still going to add a test asserting as much to make sure that doesn't drift over time.

```js
    if (isReadonly) {
      result[key] = { ...element, readonly: true };
    } else {
      result[key] = element;
    }
```

PR #4594, opened 2024-05-27
Fixes #4576 - Windows 1252 character encoding

### LINKED ISSUE #4576: Add support for `windows-1252` character encoding
`iconv-lite` supports `windows-1252` and is already in the dependency tree for `nodemailer`: https://www.npmjs.com/package/iconv-lite

### Create Middleware for Encoding Conversion

You need to create a middleware that checks the `Content-Type` of the response and converts the content to CP1252 if required.

```javascript
const iconv = require('iconv-lite');

function cp1252Middleware(req, res, next) {
    const originalSend = res.send;

    res.send = function (data) {
        if (res.getHeader('Content-Type') === 'text/plain; charset=windows-1252') {
            if (typeof data === 'string') {
                data = iconv.encode(data, 'cp1252');
                // Set the correct Content-Length for the encoded data
                res.setHeader('Content-Length', Buffer.byteLength(data));
            }
        }
        
        originalSend.apply(res, [data]);
    };

    next();
}
```

### Use the Middleware in Your Express Application

You must apply this middleware to your Express application. It should be added before any routes that may send `text/plain; charset=windows-1252` responses.

```javascript
const express = require('express');
const app = express();

// Use the cp1252 encoding middleware
app.use(cp1252Middleware);

// Example route
app.get('/', (req, res) => {
    res.setHeader('Content-Type', 'text/plain; charset=windows-1252');
    res.send('Hello, this is some text that will be encoded in CP1252!');
});

// Start the server
const PORT = 3000;
app.listen(PORT, () => {
    console.log(`Server is running on http://localhost:${PORT}`);
});
```

codyebberson:
- `packages/hl7`
    - Replaced the `encoding: BufferEncoding` param with `encoding: string`
    - Before, the `BufferEncoding` was passed to the Node.js socket `setEncoding` and defaulted to UTF-8
    - After, the Node.js socket never calls `setEncoding` and relies on `Buffer` for everything
    - We encode/decode all data with `iconv-lite`
    - We still default to UTF-8 if encoding is not specified
- `packages/agent`
    - For HL7 listener, we use the encoding from the `encoding` query param in `Endpoint.address`
    - For HL7 push, we use the encoding from the `encoding` query param in `Device.url`
    - For example, `mllp://localhost:2575?encoding=windows-1252`
- `packages/cli`
    - Added a new optional `--encoding` argument to `medplum hl7 listen`
    - Added a new optional `--encoding` argument to `medplum hl7 send`

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20302,9 +20302,12 @@
       }
     },
     "node_modules/async": {
-      "version": "3.2.5",
-      "resolved": "https://registry.npmjs.org/async/-/async-3.2.5.tgz",
-      "integrity": "sha512-baNZyqaaLhyLVKm/DlvdW051MSgO6b8eVfIezl9E5PqWxFgzLm/wQntEW4zOytVburDEr0JlALEpdOFwvErLsg=="
+      "version": "2.6.4",
+      "resolved": "https://registry.npmjs.org/async/-/async-2.6.4.tgz",
+      "integrity": "sha512-mzo5dfJYwAn29PeiJ0zvwTo04zj8HDJj0Mn8TD7sno7q12prdbnasKJHhkm2c1LgrhlJ0teaea8860oxi51mGA==",
+      "dependencies": {
+        "lodash": "^4.17.14"
+      }
     },
     "node_modules/async-eventemitter": {
       "version": "0.2.4",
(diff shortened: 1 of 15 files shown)


ThatOneBro (approved):
Nice, clean. LGTM

PR #4597, opened 2024-05-27
Support Database SSL config from Env

jdjkelly:
## Why
- Similar to #3787, it would be nice to be able to configure Database SSL from ENV values

## What
- Implemented the same way, the key we do an object check on is `ssl` - when the key is `ssl` we JSON.parse() before setting the value


jdjkelly:
Annoyingly, I can't address the edits by maintainers issue - due to: https://github.com/orgs/community/discussions/5634

This setting doesn't exist for org forks... I guess we'll open PRs from personal forks in the future

codyebberson (approved):

PR #4391, opened 2024-04-17
Log message when MockClient cannot find resource in storybook

thejwuscript:
Related to issue #4269 

This PR will log helpful error messages in the console when `MockClient` cannot find  a used `SearchParameter` or `StructureDefinition` in `storybook`. They indicate the JSON files that would need to be updated.

Showing how to add `SearchParameter` or `StructureDefinition` could perhaps be addressed in another PR.

Issue #3627, opened 2023-12-29
Add support for the `_filter` search parameter in `Subscription.criteria`


ThatOneBro:
Currently it's not possible to use `_filter` search parameter for `Subscription.criteria`. This is because the internal function `matchesCriteria()` makes use of the in-memory search, which currently has some limitations, one of which is the inability to use `_filter`.

`_filter` can be really useful, especially for complex queries where a logical OR is needed, so it would be great if we could get this added to `Subscription.criteria`, especially once `WebSocket` subscriptions are available for usage.

For example, a criteria for `Communication` resources representing a live chat session could be written as such:
`Communication?patient={patientId}&_filter=(sender re Practitioner/123 or recipient re Practitioner/123)`

However in the current system without `_filter`, we would have to create two `Subscription`s, each with a criteria representing either side of the `or` operator in the `_filter` parameter above. That means adding the `_filter` search parameter for `Subscription.criteria` could reduce total subscriptions across all clients by half for a chat service modeled by Medplum's WebSocket subscriptions.

I can see two solutions; either seems like a decent path forward for.

1. Add the appropriate `_filter` query parameter to the in-memory search implementation. Probably better for performance.
2. Migrate subscription `matchesCriteria()` to database search. Maybe less work (?)

Dharshan-K:
Hi, @ThatOneBro , I am interested in this issue. Can you guide to implement this feature.

PR #4604, opened 2024-05-29
fix(search): Log message when missing search parameter in `MockClient`, add tests

ThatOneBro:
Working off of #4391

We need to ensure that no weird behavior comes from throwing from `matchesSearchRequest`
Adding some tests for that

In another PR:
- [ ] Link docs from error about how to index SearchParameters in `MockClient`

### DIFF
--- a/examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
+++ b/examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
@@ -463,7 +463,7 @@ export async function addFamilyMemberHistory(
     },
     {
       patient: getReferenceString(patient),
-      condition: `${condition.system}|${condition.code}`,
+      code: `${condition.system}|${condition.code}`,
       relationship: `${relationship.system}|${relationship.code}`,
     }
   );
--- a/examples/medplum-task-demo/src/bots/example/create-respond-to-message-task.test.ts
+++ b/examples/medplum-task-demo/src/bots/example/create-respond-to-message-task.test.ts
@@ -1,6 +1,6 @@
 import { indexSearchParameterBundle, indexStructureDefinitionBundle } from '@medplum/core';
-import { SEARCH_PARAMETER_BUNDLE_FILES, readJson } from '@medplum/definitions';
-import { Bundle, SearchParameter } from '@medplum/fhirtypes';
+import { readJson, SEARCH_PARAMETER_BUNDLE_FILES } from '@medplum/definitions';
+import { Bundle, Communication, SearchParameter } from '@medplum/fhirtypes';
 import { MockClient } from '@medplum/mock';
 import { beforeAll, describe, expect, test, vi } from 'vitest';
 import {
@@ -27,9 +27,18 @@ describe('Create Respond to Message Task', async () => {
 
     await medplum.executeBatch(noMessagesInLast30Minutes);
 
+    const searchResourcesSpy = vi.spyOn(medplum, 'searchResources').mockImplementation((async () => {
+      return [] as Communication[];
+    }) as any);
+
     const result = await handler(medplum);
     expect(result).toBe(false);
     expect(console.log).toHaveBeenCalledWith('No messages in the last 30 minutes that require a response.');
+    expect(searchResourcesSpy).toHaveBeenCalledWith('Communication', {
+      sent: expect.stringMatching(/^lt/),
+      'part-of:Communication.status': 'in-progress',
+    });
+    searchResourcesSpy.mockRestore();
   });
 
   test('Messages in the last 30 minutes not sent by patients', async () => {
@@ -38,9 +47,18 @@ describe('Create Respond to Message Task', async () => {
 
     await medplum.executeBatch(messagesNotSentByPatients);
 
+    const searchResourcesSpy = vi.spyOn(medplum, 'searchResources').mockImplementation((async () => {
+      return [] as Communication[];
+    }) as any);
+
     const result = await handler(medplum);
     expect(result).toBe(false);
     expect(console.log).toHaveBeenCalledWith('No messages in the last 30 minutes that require a response.');
+    expect(searchResourcesSpy).toHaveBeenCalledWith('Communication', {
+      sent: expect.stringMatching(/^lt/),
+      'part-of:Communication.status': 'in-progress',
+    });
+    searchResourcesSpy.mockRestore();
   });
 
   // Skipping until chained search is implemented in MockClient
--- a/packages/core/src/access.ts
+++ b/packages/core/src/access.ts
@@ -12,7 +12,7 @@ const universalAccessPolicy: AccessPolicyResource = {
  * Protected resource types are in the "medplum" project.
  * Reading and writing is limited to the system account.
  */
-export const protectedResourceTypes = ['DomainConfiguration', 'JsonWebKey', 'Login'];
+export const protectedResourceTypes = ['DomainConfiguration', 'JsonWebKey', 'Login'] as const;
 
 /**
  * Project admin resource types are special resources that are only
--- a/packages/core/src/access.ts
+++ b/packages/core/src/access.ts
@@ -1,6 +1,8 @@
 import { AccessPolicy, AccessPolicyResource, Resource, ResourceType } from '@medplum/fhirtypes';
+import { normalizeErrorString } from './outcomes';
 import { matchesSearchRequest } from './search/match';
 import { parseSearchRequest } from './search/search';
+import { getReferenceString } from './utils';
 
 const universalAccessPolicy: AccessPolicyResource = {
   resourceType: '*',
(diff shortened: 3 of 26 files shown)


examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -463,7 +463,7 @@ export async function addFamilyMemberHistory(
     },
     {
       patient: getReferenceString(patient),
-      condition: `${condition.system}|${condition.code}`,
+      code: `${condition.system}|${condition.code}`,
```
ThatOneBro, line 466: `FamilyMemberHistory.condition` is actually `FamilyMemberHistory.code`

examples/medplum-task-demo/src/bots/example/create-respond-to-message-task.test.ts
```diff
@@ -27,9 +27,18 @@ describe('Create Respond to Message Task', async () => {
 
     await medplum.executeBatch(noMessagesInLast30Minutes);
 
+    const searchResourcesSpy = vi.spyOn(medplum, 'searchResources').mockImplementation((async () => {
```
ThatOneBro, line 30: Need to mock this since it relies on `part-of:Communication.status` chaining

packages/react-hooks/src/MedplumProvider/MedplumProvider.test.tsx
```diff
@@ -402,8 +402,8 @@ describe('MedplumProvider', () => {
       expect(screen.getByText('Loading...')).toBeInTheDocument();
       expect(medplum.isLoading()).toEqual(true);
 
-      await expect(screen.findByText('Parent count: 2')).resolves.toBeInTheDocument();
-      expect(screen.getByText('Child count: 1')).toBeInTheDocument();
+      await expect(screen.findByText('Child count: 1')).resolves.toBeInTheDocument();
```
ThatOneBro, line 405: Need to check child first since it's dependent on an async change that is dependent on parent change (this order ensures both conditions are met)

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
ThatOneBro: As revealed by validating search parameters

packages/core/src/access.ts
```diff
@@ -121,14 +123,21 @@ export function canWriteResource(accessPolicy: AccessPolicy, resource: Resource)
...
+            resourcePolicy
+          )
+        ) {
+          return true;
+        }
+      } catch (err: unknown) {
+        // TODO(ThatOneBro 24 Oct 2024): Consider removing try-catch once invalid access policies are not allowed to be created (validating search parameters are valid)
+        console.error(
+          `[Access Policy]: Got error "${normalizeErrorString(err)}" while evaluating resource against ${getReferenceString(accessPolicy)}`
+        );
```
mattwiller, line 140: I would suggest letting this `@medplum/core` function throw, and handle catching the error and logging specifically in the access policy logic in `@medplum/server`, but I'm content either way

packages/mock/src/client.ts
```diff
@@ -745,10 +756,31 @@ export class MockFetchClient {
 
     const result = await this.router.handleRequest(request, this.repo);
     if (result.length === 1) {
+      const issueDetails = result[0].issue[0]?.details;
```
mattwiller, line 759: I've seen destructuring used in several other places in the code to clarify like this:
```suggestion
    const [outcome, resource] = await this.router.handleRequest(request, this.repo);
    if (!resource) {
      const issueDetails = outcome.issue[0]?.details;
```

packages/mock/src/client.ts
```diff
@@ -720,6 +720,17 @@ export class MockFetchClient {
...
     }
 
+    // Special case for ServiceRequestTimeline and DefaultResourceTimeline Task query
+    // Since we don't support _filter yet
+    // TODO(ThatOneBro 24 Oct 2024): Remove this once we support _filter in the in-memory search implementation
+    if (
+      /^fhir\/R4\/Task\?_filter=based-on\+eq\+[a-zA-Z]+%2F[0-9a-zA-Z-]+\+or\+focus\+eq\+[a-zA-Z]+%2F[0-9a-zA-Z-]+\+or\+subject\+eq\+[a-zA-Z]+%2F[0-9a-zA-Z-]+/.test(
+        url
+      )
+    ) {
```
mattwiller, line 730: Is there a less brittle way we might be able to handle this?

ThatOneBro: Open to suggestions, these are hardcoded queries that show up everywhere in `examples` and `app` tests since they appear in both `ServiceRequestTimeline` and `DefaultResourceTimeline`

https://github.com/medplum/medplum/blob/main/packages/react/src/ServiceRequestTimeline/ServiceRequestTimeline.tsx#L23
https://github.com/medplum/medplum/blob/main/packages/react/src/DefaultResourceTimeline/DefaultResourceTimeline.tsx#L19

The idea is we don't want to return anything in too generic of a case so that it fails and triggers the error

packages/core/src/access.ts
ThatOneBro: I think the problem is that users might be relying on calling this function directly in other places too and could suddenly start throwing on them 😓 

In principle I agree though. Thoughts @codyebberson, @mattlong , @rahul1  ?

packages/server/src/fhir/repo.ts
```diff
@@ -1174,6 +1180,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
 
     for (const policy of accessPolicy.resource) {
       if (policy.resourceType === resourceType) {
+        console.log({ policy });
```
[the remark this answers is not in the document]
ThatOneBro: Oops yeah I removed these during the process of your review 😛

packages/server/src/fhir/repo.ts
```diff
@@ -200,7 +200,7 @@ export interface InteractionOptions {
 }
 
 export interface ReadResourceOptions extends InteractionOptions {
-  checkCacheOnly?: boolean;
+  allowReadFrom?: ('cache' | 'database')[];
```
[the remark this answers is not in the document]
ThatOneBro: Good point, although now I'm looking at this change, which was originally made to support a test case where I wanted to skip the cache, I ended up removing that test case and going with another one which didn't require the database-only `readResource`. Think I'm going to opt for removing this for now and if we do need the `database-only` option in the future we should use the object version you came up with

packages/fhir-router/src/repo.ts
```diff
@@ -456,6 +456,12 @@ export class MemoryRepository extends FhirRepository<undefined> {
   async search<T extends Resource>(searchRequest: SearchRequest<T>): Promise<Bundle<T>> {
     const { resourceType } = searchRequest;
     const resources = this.resources.get(resourceType) ?? new Map();
+    if (resources.size === 0) {
```
ThatOneBro, line 459: Needed to add this since in the case when no resources were in `MemoryRepository` for a given resource type, we would always return an empty searchset `Bundle` even when given an invalid query with unknown search parameters

mattwiller (approved):

packages/fhir-router/src/repo.ts
```diff
@@ -457,10 +474,24 @@ export class MemoryRepository extends FhirRepository<undefined> {
     const { resourceType } = searchRequest;
     const resources = this.resources.get(resourceType) ?? new Map();
     const result = [];
-    for (const resource of resources.values()) {
-      if (matchesSearchRequest(resource, searchRequest)) {
-        result.push(resource);
+    try {
+      validateSearchResourceTypes(this, searchRequest);
```
ThatOneBro, line 478: Moved resource type validation here to mirror the server respository implementation

packages/fhir-router/src/repo.ts
```diff
@@ -457,10 +474,24 @@ export class MemoryRepository extends FhirRepository<undefined> {
...
+        // Instead of silently failing and returning an empty search set as if it was a valid search
+        matchesSearchRequest({ resourceType } as T, searchRequest);
       }
+      for (const resource of resources.values()) {
+        if (matchesSearchRequest(resource, searchRequest)) {
+          result.push(resource);
+        }
+      }
+    } catch (err) {
+      if (this.strictMode) {
```
ThatOneBro, line 491: Added "strict mode" check to add an escape hatch when we don't want `MockClient` / `MemoryRepository` to throw, especially in the case of `storybook`

packages/fhir-router/src/repo.ts
```diff
@@ -515,6 +546,50 @@ export class MemoryRepository extends FhirRepository<undefined> {
...
+    return true;
+  }
+}
+
+/**
+ * Validates that the resource type(s) are valid and that the user has permission to read them.
+ * @param repo - The user's repository.
+ * @param searchRequest - The incoming search request.
+ */
+export function validateSearchResourceTypes(repo: FhirRepository, searchRequest: SearchRequest): void {
```
ThatOneBro, line 568: Moved these to `fhir-router` since we can reuse them in `MemoryRepository`

packages/react/.storybook/preview.tsx
```diff
@@ -21,7 +21,7 @@ export const parameters = {
 // so that resources created in MockFetchClient#initMockRepo have
 // consistent timestamps between storybook runs
 const clock = createGlobalTimer();
-const medplum = new MockClient();
+const medplum = new MockClient({ strictMode: false });
```
ThatOneBro, line 24: We need to disable `strictMode` in `storybook` since we have some searches that reference types that we don't have the `StructureDefinition` or `SearchParameter`s for

packages/mock/src/client.ts
codyebberson: Yeah... I think this is might be too much of a special case.

We could silently ignore all `_filter`, but that seems like a weird middle ground.

I think the right answer is that we should implement `_filter` in `MemoryRepository` before we can ship this.

packages/mock/src/client.ts
```diff
@@ -743,12 +761,33 @@ export class MockFetchClient {
...
-    } else {
-      return result[1];
+    const [outcome, resource] = await this.router.handleRequest(request, this.repo);
+    if (!resource) {
+      const issueDetails = outcome.issue[0]?.details;
+      const issueText = issueDetails?.text;
+      if (!issueText) {
+        return outcome;
+      }
+      // Special case for unknown search parameters and resource types
```
codyebberson, line 771: We can discuss as a team.  I'm kinda philosophically opposed to having these kinds of extra layers of error logic.

I understand the utility of helping users who are fighting mysterious error messages.  That is a good and noble goal.  I'm just not sure this is the right place or the right way to do it.

This is brittle, because it depends on exact error message strings.  It also duplicates errors in multiple places.

I.e., I'd rather make it easier to automatically call `indexStructureDefinitionBundle` and `indexSearchParameterBundle` with a single medplum test setup utility or something.

Or, we could update all of the `throw new OperationOutcomeError` call sites to have short messages and longer detailed messages.

I'm open to improving the dev ex here, I'm just concerned about the sustainability and maintainability of this approach.

codyebberson (commented):
There's a lot of good stuff in here, but there are a few changes that we should discuss as a team.

ThatOneBro:
Closed due to bitrot and a dependency on #3627

We should re-evaluate this after more work is done there

PR #4613, opened 2024-05-31
feat(agent): `Agent/$upgrade` operation

### LINKED ISSUE #3900: Medplum Agent Command: Force Upgrade

### LINKED ISSUE #4225: Find alternative to `pkg` for agent
The `pkg` project was deprecated: https://github.com/vercel/pkg 😭 

We have a little grace period here, but ultimately we need to find a replacement.

Consider Node.js single file binary: https://nodejs.org/api/single-executable-applications.html

ThatOneBro:
Closes #3900
Closes #4225


packages/agent/src/upgrader-utils.ts
```diff
@@ -0,0 +1,161 @@
...
+export const GITHUB_RELEASES_URL = 'https://api.github.com/repos/medplum/medplum/releases';
+export const RELEASES_PATH = resolve(__dirname);
+
+const releaseManifests = new Map<string, ReleaseManifest>();
+
+export function clearReleaseCache(): void {
+  releaseManifests.clear();
+}
+
+export function getOsString(): SupportedOs {
```
codyebberson, line 24: Could we just use `platform()` directly here, and cut `getOsString()` and `SupportedOs`?

packages/agent/src/upgrader-utils.ts
```diff
@@ -0,0 +1,161 @@
...
+  // Get download url
+  const downloadUrl = parseDownloadUrl(release, getOsString());
+
+  // Write file to RELEASE_INSTALLER_FOLDER
+  const { body } = await fetch(downloadUrl);
+  if (!body) {
+    throw new Error('Body not present on Response');
+  }
+
+  // @ts-expect-error Slight type mismatch
```
codyebberson, line 90: Can we use cast rather than `@ts-expect-error`?

packages/app/src/resource/ToolsPage.test.tsx
```diff
@@ -229,4 +227,46 @@ describe('ToolsPage', () => {
...
+      fireEvent.click(screen.getByRole('button', { name: /upgrade/i }));
+    });
+
+    await expect(screen.findByText('Success')).resolves.toBeInTheDocument();
+  });
+
+  test('Upgrade -- Error', async () => {
+    medplum = new MockClient();
+    medplum.router.router.add('GET', 'Agent/:id/$upgrade', async () => [serverError(new Error('Something is broken'))]);
+    agent = await medplum.createResource<Agent>({ resourceType: 'Agent', name: 'Agente', status: 'active' });
```
codyebberson, line 258: lol, `Agente` always catches me off guard 😛

codyebberson (commented):
Big picture looks good.  See comments and questions.  I'm going to test and run it on some windows machines before approval, but nothing else blocking.

Thanks for hard work on this, I imagine it was a slog 🙏

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -198,3 +198,9 @@ function parseIfMatchHeader(ifMatch: string | undefined): string | undefined {
   const match = /"([^"]+)"/.exec(ifMatch);
   return match ? match[1] : undefined;
 }
+
+export function getFhirRequestParams<T extends Record<string, string | undefined> = Record<string, string | undefined>>(
```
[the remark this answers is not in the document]
ThatOneBro: It's about type info, technically a key can be missing from the object and if you forget to check the current types will tell you that something is defined. Using this function will make sure all values are checked for `undefined`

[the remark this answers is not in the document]
ThatOneBro: I'll move this to another PR potentially since it's not necessarily important for this PR.
Just came up while I was working on checking the parameters in the operation

packages/agent/src/upgrader-utils.ts
ThatOneBro: Originally tried assert, didn't work naively. But I found that asserting `ReadableStream` from `node:stream/web` worked 👍 done

packages/agent/src/agent-main.ts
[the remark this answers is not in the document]
ThatOneBro: `main.ts` now is the entry point for two different "binaries"... I think it made since to split the main entry point from the entry point of each of the two...

Also when testing the `main`, I need to mock `agentMain` in order to test that it actually calls the right main between `agentMain` and `upgraderMain`, which is not possible if `agentMain` is in the same module as `main` itself.

scripts/build-agent-sea-win64.sh
```diff
@@ -0,0 +1,34 @@
...
+# fi
+
+# Generate blob to inject into node executable
+node --experimental-sea-config sea-config.json
+
+# Copy the local node binary
+node -e "require('fs').copyFileSync(process.execPath, 'dist/medplum-agent-$MEDPLUM_VERSION-win64.exe')" 
+
+# Remove signature from binary
+signtool remove -s dist/medplum-agent-$MEDPLUM_VERSION-win64.exe
```
ThatOneBro, line 29: Add comment about how to install `signtool` from Windows SDK, if missing:

References: 
https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/
https://stackoverflow.com/a/52963704

packages/agent/src/app.ts
```diff
@@ -488,6 +550,101 @@ export class App {
...
+        body: errMsg,
+      } satisfies AgentError);
+      return;
+    }
+
+    try {
+      const command = __filename;
+      const logFile = openSync(UPGRADER_LOG_PATH, 'w+');
+      child = spawn(command, ['--upgrade'], { detached: true, stdio: ['ignore', logFile, logFile, 'ipc'] });
+      child.unref();
```
ThatOneBro, line 584: Add comment about why we unref (to allow process to close before child closes)

packages/agent/src/agent-main.ts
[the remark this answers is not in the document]
ThatOneBro: As a suggestion I'm not completely sold on but might bring more uniformity, should we rename `agent-main.ts` to `agent.ts`? So it's `upgrader.ts` and `agent.ts` ?

codyebberson (approved):

PR #4624, opened 2024-06-03
AWS Textract demo

codyebberson:


reshmakh (approved):

packages/app/src/config.ts
```diff
@@ -4,6 +4,7 @@ export interface MedplumAppConfig {
   googleClientId?: string;
   recaptchaSiteKey?: string;
   registerEnabled?: boolean | string;
+  aiEnabled?: boolean | string;
```
reshmakh, line 7: maybe this name is too general?  "comprehend enabled" or "textract enabled"?

codyebberson: Yeah, I wondered about that.  It's 100% a granularity question.  We have a Slack thread internally about using data from `/auth/me` such as AccessPolicy and Project.features to determine this dynamically.  My assumption is that this config setting will go away when the UI becomes more aware of server side config.

With that said, I'm happy to break it apart into smaller settings 👍

codyebberson: Renamed it to `awsTextractEnabled` 👍

reshmakh (approved):

PR #4647, opened 2024-06-10
Allow chained search in _filter

### LINKED ISSUE #3224: Allow using chained search parameters in _filter
Currently, chained search parameters do not work within a `_filter` search expression.  For example, this `Task?_filter` query should work:
```
patient._has:CareTeam:subject:participant re Practitioner/123 or patient.general-practitioner re Practitioner/123
```

Currently, chained search parameters are [handled as a special case, before other parameters][code], but `_filter` is handled within [`buildSearchFilterExpression()`][code2] — which is recursively called to parse the search parameters inside the `_filter` query.  This will need to be refactored to include handling of chained search parameters.

[code]: https://github.com/medplum/medplum/blob/fbf271ba1389e850d62d154c466cf4a3cda4b173/packages/server/src/fhir/search.ts#L486-L497
[code2]: https://github.com/medplum/medplum/blob/fbf271ba1389e850d62d154c466cf4a3cda4b173/packages/server/src/fhir/search.ts#L520

mattwiller:
Fixes #3224

### DIFF
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -2906,6 +2906,39 @@ describe('FHIR Search', () => {
         expect(result.entry?.[0]?.resource?.id).toEqual(patient.id);
       }));
 
+    test('_filter with chained search', () =>
+      withTestContext(async () => {
+        const mrn = randomUUID();
+        const patient = await repo.createResource<Patient>({
+          resourceType: 'Patient',
+          name: [{ given: ['Eve'] }],
+          identifier: [{ system: 'http://example.com/mrn', value: mrn }],
+        });
+
+        const observation = await repo.createResource<Observation>({
+          resourceType: 'Observation',
+          status: 'final',
+          code: {
+            text: 'Strep test',
+          },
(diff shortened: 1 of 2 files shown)


ThatOneBro (approved):

packages/server/src/fhir/search.test.ts
```diff
@@ -2906,6 +2906,39 @@ describe('FHIR Search', () => {
...
+              operator: Operator.EQUALS,
+              value: `subject:Patient.identifier eq http://example.com/mrn|${mrn}`,
+            },
+          ],
+        });
+
+        expect(result.entry).toHaveLength(1);
+        expect(result.entry?.[0]?.resource?.id).toEqual(observation.id);
+      }));
+
```
rahul1, line 2941: The canonical use case of this is to use `or`, so it's with having a test case for this. 

From [this thread](https://medplum.slack.com/archives/C04A55B3VU5/p1715703564822689):
"I want to search Tasks where the `Task.patient` has an identifier with system `http://example.com|123` OR the patient has no identifier with system `http://example.com`

PR #4651, opened 2024-06-10
Deprecate non-strict mode

mattwiller:


ThatOneBro (commented):
In general I think this is a good change, but want to get more consensus on this 👍

codyebberson:
Curious what motivated this change?

In a perfect world, I agree, and we would appeal to maximum strictness all the time.

In practice, we have significant customers who depend on non-strict capabilities, so we cannot actually remove it in the immediate future.

@rahul1 - do you have any thoughts here?

mattwiller:
@codyebberson The intent was to only deprecate the setting, so that any customers already using the escape hatch still have it.  This change would just make it so non-Super Admins cannot turn the feature on anymore, guiding new customers towards the more correct option

codyebberson (approved):

PR #4657, opened 2024-06-12
feat(agent): add `keepAlive` setting to `Agent`

### LINKED ISSUE #4606: Agent Persistent connections?
When sending messages to the Agent service, is a persistent connection possible or is the connection opened/closed each time? This would be useful to know for both inbound messages sent to the Agent and messages flowing outbound from the Agent to so other system. I know the Websocket to connect the Agent to the Medplum Server is persistent but this is related to the Agent and outside connections - specifically using MLLP

ThatOneBro:
Closes #4606


packages/agent/package.json
```diff
@@ -27,14 +27,12 @@
     "@medplum/hl7": "3.1.8",
     "dcmjs-dimse": "0.1.27",
     "iconv-lite": "0.6.3",
-    "node-windows": "1.0.0-beta.8",
```
codyebberson, line 30: Yeah, this has been hanging around for a while.  It's ok to remove it, but we may need to bring it back.

Context:  The Medplum "Agent" product is kinda like the 3rd or 4th version, descending from custom proprietary projects that we have custom built for customers.  One of the long term goals is to deprecate all of those old proprietary projects, and standardize everything on the agent.

One of the required features of one of those custom proprietary projects is logging to the Windows Event Log, which is why the `node-windows` dep was here.  

We can revisit that in the future when we want to make the final push to converge everything.

packages/agent/src/app.ts
```diff
@@ -518,11 +545,37 @@ export class App {
...
+      client = new Hl7Client({
+        host: address.hostname,
+        port: Number.parseInt(address.port, 10),
+        encoding,
+        keepAlive,
+      });
+      this.log.info(`Client created for remote '${message.remote}'`, { keepAlive, encoding });
+
+      if (client.keepAlive) {
+        this.hl7Clients.set(message.remote, client);
```
codyebberson, line 565: Mental note - `message.remote` should include host, port, and other connection details, so this is a good unique identifier 👍

codyebberson (approved):
Nice, clean, lgtm 👍

packages/agent/package.json
ThatOneBro: Yeah that makes sense, this is good context

packages/agent/esbuild.mjs
```diff
@@ -12,7 +12,7 @@ const options = {
   resolveExtensions: ['.js', '.ts'],
   target: 'es2021',
   tsconfig: 'tsconfig.json',
-  external: ['iconv-lite', 'pdfmake'],
```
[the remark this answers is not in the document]
ThatOneBro: In the Node single executable we can't keep `iconv-lite` because it doesn't do any of the auto-bundling magic that `pkg` did and only works with a single bundled file. `iconv-lite` is a completely JS module so this works though we may have to figure out how native modules work in the Node SEA env since I don't think it's quite as "magical" as the `pkg` version, though we saw that `pkg` itself had its issue with native stuff

PR #4659, opened 2024-06-12
Re-add example-bots.json

Bjornkjohnson:
Fix local dev currently broken due to example-bot.json reference in examples/medplum-chart-demo/src/pages/UploadDataPage.tsx

image


rahul1:
HI @Bjornkjohnson , thanks for flagging this!

I've just opened up [this pr](https://github.com/medplum/medplum/pull/4660) with some guidance on how to fix. This `example-bots.json` file is actually dynamically generated by the `build:bots` script, so running this first should solve the problem

PR #4663, opened 2024-06-13
Update useSearch.ts

claycoleman:
The `useSearchImpl` function never sets its `loading` state to true; it is initialized to `false` (likely the bug) and then set to `false` after a successful load or an error response. 

This commit initializes `loading` to `true` so the flag is true before any response is returned.

### DIFF
--- a/packages/react-hooks/src/useSearch/useSearch.ts
+++ b/packages/react-hooks/src/useSearch/useSearch.ts
@@ -60,7 +60,7 @@ function useSearchImpl<K extends ResourceType, ReturnType>(
 ): [ReturnType | undefined, boolean, OperationOutcome | undefined] {
   const medplum = useMedplum();
   const [searchKey, setSearchKey] = useState<string>();
-  const [loading, setLoading] = useState<boolean>(false);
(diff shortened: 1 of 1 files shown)


claycoleman:
cc @rahul1 and @ThatOneBro for a quick review 🙏

ThatOneBro (approved):
Good catch! Looks like `loading` is never true currently, huh? 😅 

LGTM thanks for finding that @claycoleman

claycoleman:
amazing, thanks @ThatOneBro for taking a look so fast! I'm new to using medplum, what is your typical release cadence for the medplum packages?

ThatOneBro:
@claycoleman We try to release at least every 2 weeks or so, though often times we get a release per week or even more frequently. We're mostly in the 1-2 week range but it kind of depends on how many PRs we get merged, if there are any major bug fixes we need to push out, etc.

Ideally we would release every week or so in my opinion, but we haven't formalized any kind of release cadence at this point

Again, thanks again for your contribution! And feel free to let us know any pain points or questions you may have

PR #4664, opened 2024-06-13
Export QuestionnaireFormContext and QuestionnairePageSequence from QuestionnaireForm

claycoleman:
Hi there,

I'd like to create an extension of `QuestionnaireForm` in my own repository by forking the code of `QuestionnaireForm`, but two files are not exported from the `@medplum/react` package: `QuestionnaireFormContext` and `QuestionnairePageSequence`. I'd like to use the exact context from `QuestionnaireFormContext` rather than recreating it so all nested components can continue to work with it. This PR just exports these two files from `QuestionnaireForm` so they're picked up in the barrel `index.ts` file.

I'm currently using a pnpm patch package to do this, but it's fairly janky to do this as `@medplum/react` exports a minified .mjs file and creates a massive patch file!

Another approach here could be adding the Context and PageSequence exports into the barrel file, not sure which you'd prefer.

Thanks for taking a look!

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -15,6 +15,9 @@ import { buildInitialResponse, getNumberOfPages, isQuestionEnabled } from '../ut
 import { QuestionnaireFormContext } from './QuestionnaireForm.context';
 import { QuestionnairePageSequence } from './QuestionnaireFormComponents/QuestionnaireFormPageSequence';
 
+export { QuestionnaireFormContext } from './QuestionnaireForm.context';
+export { QuestionnairePageSequence } from './QuestionnaireFormComponents/QuestionnaireFormPageSequence';
+
 export interface QuestionnaireFormProps {
   readonly questionnaire: Questionnaire | Reference<Questionnaire>;
   readonly subject?: Reference;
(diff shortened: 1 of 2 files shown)


claycoleman:
cc @codyebberson for a quick review 🙏

claycoleman:
also @ThatOneBro , if you have another quick second, this one is a little more opinionated but would love to get this one into the next release as well

ThatOneBro:
Hi @claycoleman, thanks for your contributions! 

I can definitely see the logic behind this, and this is definitely a use-case we want to support. 

I think we typically try to export one component / thing per file, mostly due to how to [React fast refresh](https://www.npmjs.com/package/eslint-plugin-react-refresh) works in dev mode, and that it requires you only export components from files containing components to be eligible for fast refresh. (We use that rule in `@medplum/eslint-config`)

Can you change this PR to directly export the context and `QuestionnairePageSequence` directly from the barrel file?

Again, thanks for contributing

claycoleman:
@ThatOneBro done! tested by going to a file that imported from `@medplum/react` and saw using TS autocomplete that the `Context` and `PageSequence` are being exported.

ThatOneBro (approved):
@claycoleman Awesome, thanks a ton! Approved

We should cut a release soon so keep an eye out. Patching packages is never fun and we strive to never have Medplum users doing that for stuff like this

Let us know if you have any other concerns!

PR #4668, opened 2024-06-14
Revert "Export  QuestionnairePageSequence from QuestionnaireForm (#4664)"

rahul1:


rahul1:
@claycoleman - so sorry for the swirl here. The `QuestionnairePageSequence` was meant to be an internal implementation detail, and the API isn't quite ready for the real world. 

For now, if you wouldn't mind forking the `QuestionnairePageSequence`, we're working on some solutions that will hopefully make it easier to customize questionnaire forms in the future

PR #4669, opened 2024-06-14
revert(react): remove export of `QuestionnairePageSequence`

ThatOneBro:
After more thought from the team, we might not want the implementation details of `QuestionnairePageSequence` to be leaked in its current form.

We are experimenting with a different route and encapsulating most of the component state / logic in a hook since `QuestionnairePageSequence` may soon be factored out of the `QuestionnaireForm` altogether. See: #4667

This route ultimately seems more stable and maintainable long term. Sorry about the confusion @claycoleman

For now we will only export the `QuestionnaireForm.context` and not `QuestionnairePageSequence`


rahul1 (approved):

claycoleman:
darn, so seems like there is no officially supported way to fork this component for now seeing as `QuestionnairePageSequence` renders the bulk of the form

PR #4672, opened 2024-06-14
feat(subscriptions): add `unbind-from-token` message for WebSocket subscriptions

### LINKED ISSUE #4133: Add `unbind-from-token` message to FHIR `WebSocket Subscription` protocol
The current workflow for `WebSocket Subscriptions` is as follows:
1. Create `Subscription` with `channelType` of `websocket`
2. Call `Subscription/$get-ws-binding-token` operation on created Subscription to get JWT for this `Subscription`
3. Connect to WebSocket
4. Send `bind-to-token` message with the given token, indicating that you want to receive notifications for the connected `Subscription`
5. Receive notifications for this `Subscription` for the rest of the life of this WebSocket connection, or until the `Subscription` is deleted

It would be nice to have an `unbind-from-token` message, especially when quickly creating and deleting Subscriptions with different criteria from page to page, so that you can opt out of notifications for a given `Subscription` explicitly.

As of now, we could still achieve the same behavior by deleting a `Subscription`, but this is not exactly the cleanest way and not as flexible as being able to explicitly unbind a particular WebSocket from the `Subscription` without deleting it (in the case of having multiple WebSockets, maybe across multiple tabs, reusing the same `Subscription`, for example.

ThatOneBro:
Fixes #4133

### DIFF
--- a/packages/core/src/subscriptions/index.test.ts
+++ b/packages/core/src/subscriptions/index.test.ts
@@ -503,50 +503,54 @@ describe('SubscriptionManager', () => {
       console.warn = originalWarn;
     });
 
-    test('should not clean up a criteria if there are outstanding listeners', (done) => {
+    test('should not clean up a criteria if there are outstanding listeners', async () => {
       let success = false;
       const emitter = manager.addCriteria('Communication');
       expect(emitter).toBeInstanceOf(SubscriptionEmitter);
 
+      let receivedDisconnect = false;
       const handler = (): void => {
         emitter.removeEventListener('disconnect', handler);
         if (!success) {
-          done(new Error('Received `disconnect` when not expected'));
+          receivedDisconnect = true;
         }
       };
       emitter.addEventListener('disconnect', handler);
 
       manager.removeCriteria('Communication');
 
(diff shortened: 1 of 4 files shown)


packages/mock/src/subscription-manager.ts
```diff
@@ -18,13 +20,19 @@ class MockRobustWebSocket extends TypedEventTarget<RobustWebSocketEventMap> impl
   }
 }
 
+type MockCriteriaEntry = {
```
codyebberson, line 23: Does this need to be exported?

codyebberson (approved):
Nice, lgtm

Non-blocking, idle thoughts... Would we have any way to track how many open connections there are?  Just trying to think through the dev ops scenario, and how we could get observability here.  Might already be possible via otel?

ThatOneBro:
> Non-blocking, idle thoughts... Would we have any way to track how many open connections there are? Just trying to think through the dev ops scenario, and how we could get observability here. Might already be possible via otel?

We should definitely add an issue, I think reporting via otel should be relatively simple and would definitely be a win for observability

PR #4677, opened 2024-06-17
fix(react-hooks): make `loading` track `MedplumClient#isLoading()`

ThatOneBro:
This aligns `useMedplumContext().loading` and `MedplumClient#isLoading()` so that it should be less confusing and more useful for users who want to be able use `loading` as a dependency to other hooks

### DIFF
--- a/packages/react-hooks/src/MedplumProvider/MedplumProvider.tsx
+++ b/packages/react-hooks/src/MedplumProvider/MedplumProvider.tsx
@@ -23,7 +23,7 @@ export function MedplumProvider(props: MedplumProviderProps): JSX.Element {
 
   const [state, setState] = useState({
     profile: medplum.getProfile(),
-    loading: !medplum.isInitialized,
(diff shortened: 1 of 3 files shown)


packages/react-hooks/src/MedplumProvider/MedplumProvider.tsx
```diff
@@ -23,7 +23,7 @@ export function MedplumProvider(props: MedplumProviderProps): JSX.Element {
 
   const [state, setState] = useState({
     profile: medplum.getProfile(),
-    loading: !medplum.isInitialized,
+    loading: medplum.isLoading(),
```
codyebberson, line 26: Nice catch👍   

For my pea brain when we come back to this:

```ts
  isLoading(): boolean {
    return !this.isInitialized || !!this.profilePromise;
  }
```

codyebberson (approved):
🔥

PR #4682, opened 2024-06-17
fix(build) Update deprecated import assertion into import attribute

sahinf:
It looks like "assert" was deprecated. 
image
https://nodejs.org/docs/latest-v20.x/api/esm.html#json-modules

I'm not sure if this is a problem for others but building fails for me because of assert:
image

Found 4 occurrences. 
image


ThatOneBro (approved):
Nice,  noticed this change a while back but didn't realize we had the old assertions here

Thank you

PR #4683, opened 2024-06-18
Implements FHIRPath string join (stu)

codyebberson:
Note that `join` is not in the published version: https://hl7.org/fhirpath/

It is in the current STU version: https://build.fhir.org/ig/HL7/FHIRPath/#joinseparator-string--string

For better or worse, it is referenced by the SQL-on-FHIR test suite

### DIFF
--- a/packages/core/src/fhirpath/functions.ts
+++ b/packages/core/src/fhirpath/functions.ts
@@ -1355,7 +1355,7 @@ export const functions: Record<string, FhirPathFunction> = {
    * @returns The joined string.
    */
   join: (context: AtomContext, input: TypedValue[], separatorAtom: Atom): TypedValue[] => {
-    const separator = separatorAtom?.eval(context, input)[0]?.value ?? '';
+    const separator = separatorAtom?.eval(context, getRootInput(context))[0]?.value ?? '';
     if (typeof separator !== 'string') {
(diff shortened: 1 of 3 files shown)


packages/core/src/fhirpath/functions.ts
```diff
@@ -1327,6 +1327,41 @@ export const functions: Record<string, FhirPathFunction> = {
...
+   *
+   * See: https://build.fhir.org/ig/HL7/FHIRPath/#joinseparator-string--string
+   *
+   * @param context - The evaluation context.
+   * @param input - The input collection.
+   * @param separatorAtom - Optional separator atom.
+   * @returns The joined string.
+   */
+  join: (context: AtomContext, input: TypedValue[], separatorAtom: Atom): TypedValue[] => {
+    const separator = separatorAtom?.eval(context, input)[0]?.value ?? '';
```
mattlong, line 1358: Should separator be computed from the root input instead?

mattlong:
```suggestion
    const separator = separatorAtom?.eval(context, getRootInput(context));[0]?.value ?? '';
```

mattwiller (approved):

mattlong (approved):

PR #4702, opened 2024-06-21
More SQL-on-FHIR tests

codyebberson:


codyebberson:
After merging, the `test_report.json` will be available for submission to the implementations page.

`test_report.json`: https://raw.githubusercontent.com/medplum/medplum/cody-more-sql-on-fhir/packages/core/src/sql-on-fhir/test_report.json

https://github.com/FHIR/sql-on-fhir-v2/blob/master/test_report/public/implementations.json

packages/core/src/sql-on-fhir/eval.ts
```diff
@@ -59,12 +59,23 @@ function processResource(v: ViewDefinition, r: Resource): OutputRow[] {
...
+
   const typedResource = toTypedValue(r);
 
   if (v.where) {
     for (const where of v.where) {
-      const whereResult = evalFhirPathTyped(where.path, [typedResource]);
-      if (whereResult.length !== 1 || whereResult[0].type !== 'boolean') {
+      const whereResult = evalFhirPathTyped(where.path, [typedResource], variables);
+      if (whereResult.length !== 1) {
+        return [];
```
mattwiller, line 76: Should this throw an error instead of failing silently?

packages/core/src/fhirpath/functions.ts
```diff
@@ -1791,6 +1791,81 @@ export const functions: Record<string, FhirPathFunction> = {
...
+   * @param _context - The evaluation context.
+   * @param input - The input collection.
+   * @returns The resource key.
+   */
+  getResourceKey: (_context: AtomContext, input: TypedValue[]): TypedValue[] => {
+    const resource = input[0].value as Resource;
+    if (!resource?.id) {
+      return [];
+    }
+    return [{ type: PropertyType.id, value: resource.id }];
```
mattwiller, line 1819: Do these need to be globally unique, or just per resource type?

mattwiller (approved):

packages/core/src/sql-on-fhir/eval.ts
codyebberson: Good question.  In this case, no, there are tests explicitly for each of these cases.

For example, there is a test that uses simply `Patient.active` and expects to gracefully handle the case where `Patient.active` is missing.

packages/core/src/fhirpath/functions.ts
codyebberson: Good question.  Using `id` is the recommended implementation pattern from the spec: https://build.fhir.org/ig/FHIR/sql-on-fhir-v2/StructureDefinition-ViewDefinition.html#getresourcekey--keytype

The spec does include language about how that could lead to ambiguity if ID's are not globally unique.

I'm kinda mixed on the pros/cons.  Given that the spec recommended using `id`, I'm inclined to go with it 🤷

mattwiller: Cool, I'm alright with using just `id` — especially since this might be less of a risk if we switch to V5 UUIDs in the future for more "guaranteed" global uniqueness

PR #4712, opened 2024-06-23
refactor(core): merge `TypedEventTarget` and `EventTarget` classes

### LINKED ISSUE #4710: Merge `TypedEventTarget` and `EventTarget`
It would be nice to flatten the definitions of these two. Seems possible but a little more work than it would seem at first glance

ThatOneBro:
Closes #4710

### DIFF
--- a/packages/core/src/eventtarget.ts
+++ b/packages/core/src/eventtarget.ts
@@ -2,29 +2,51 @@
  * Based on: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget
  */
 
-export interface Event {
-  readonly type: string;
+export interface Event<T extends string = string> {
+  readonly type: T;
   readonly defaultPrevented?: boolean;
 }
 
-export type EventListener = (e: Event) => void;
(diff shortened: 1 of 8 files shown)


packages/core/src/eventtarget.ts
```diff
@@ -2,29 +2,51 @@
...
-export type EventListener = (e: Event) => void;
+export type TypedEvent<T extends string> = globalThis.Event & {
+  type: T;
+};
 
-export class EventTarget {
-  private readonly listeners: Record<string, EventListener[]>;
+export type EventMap<T> = { [K in keyof T]: K extends string ? Event<K> | T[K] : never } | NoEventMap;
+export type NoEventMap = [never];
+export type Key<K, T> = T extends NoEventMap ? string : K | keyof T;
```
codyebberson, line 16: Clever 👍

ThatOneBro:
Attempted a couple of different approaches here, couldn't find a way to support both default unmapped and mapped events in one class. Will come back to this later but going to close this for now since it's kind of just noise.

PR #4722, opened 2024-06-25
Fix recursive filter application in ValueSet expand

mattwiller:
The logic for the `ValueSet/$expand` operation was applying the text filter at each level of recursion when expanding out a hierarchy; instead, the filter should only be applied at the end


codyebberson (approved):

reshmakh:
@mattwiller - link the new performance issue here.

mattwiller:
Related: https://github.com/medplum/medplum/issues/4736

PR #4723, opened 2024-06-25
feat(core): add `maxRetries` opt to `MedplumRequestOptions`

ThatOneBro:
- Adds `maxRetries` option to `MedplumRequestOptions`, so that you can tune the amount of retries on a per-request basis.
- Refactors `MedplumClient#fetchWithRetry` so that it's a bit easier to follow
- Adds tests for various cases within `MedplumClient#fetchWithRetry`

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1,4 +1,12 @@
-import { Bot, Bundle, Identifier, Patient, SearchParameter, StructureDefinition } from '@medplum/fhirtypes';
+import {
+  Bot,
+  Bundle,
+  Identifier,
+  OperationOutcome,
+  Patient,
+  SearchParameter,
+  StructureDefinition,
+} from '@medplum/fhirtypes';
 import { randomUUID, webcrypto } from 'crypto';
 import PdfPrinter from 'pdfmake';
 import type { CustomTableLayout, TDocumentDefinitions, TFontDictionary } from 'pdfmake/interfaces';
@@ -15,10 +23,19 @@ import {
 } from './client';
 import { createFakeJwt, mockFetch, mockFetchResponse } from './client-test-utils';
 import { ContentType } from './contenttype';
-import { OperationOutcomeError, accepted, allOk, forbidden, notFound, unauthorized } from './outcomes';
+import {
+  OperationOutcomeError,
+  accepted,
+  allOk,
+  badRequest,
+  forbidden,
+  notFound,
+  serverError,
+  unauthorized,
+} from './outcomes';
 import { MockAsyncClientStorage } from './storage';
 import { getDataType, isDataTypeLoaded, isProfileLoaded } from './typeschema/types';
-import { ProfileResource, createReference } from './utils';
+import { ProfileResource, createReference, sleep } from './utils';
 
 const patientStructureDefinition: StructureDefinition = {
   resourceType: 'StructureDefinition',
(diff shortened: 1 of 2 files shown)


packages/core/src/client.ts
```diff
@@ -320,6 +320,10 @@ export interface MedplumRequestOptions extends RequestInit {
    * Default value is 1000 (1 second).
    */
   pollStatusPeriod?: number;
+  /**
+   * Optional max number of retries that should be made in the case of a failed request. Default is `2`.
+   */
+  maxRetries?: number;
```
mattwiller, line 326: Users might also want to customize the retry delay or other characteristics of the retry loop; one possible way we could accommodate this would be to allow them to pass a function option like this:
```ts
retryStrategy?: (url: string, retryCount: number, response: Response) => number | undefined;
```

packages/core/src/client.test.ts
```diff
@@ -1992,6 +2009,101 @@ describe('Client', () => {
...
+      const fetch = mockFetch(500, serverError(new Error('Something is broken')));
+      const client = new MedplumClient({ fetch });
+
+      await expect(client.get(client.fhirUrl('Patient', '123'), { maxRetries: 0 })).rejects.toThrow(
+        'Internal server error (Error: Something is broken)'
+      );
+      expect(fetch).toHaveBeenCalledTimes(1);
+    });
+
+    test.each([400, 401, 404])('%d status code is not retried', async (statusCode) => {
```
mattwiller, line 2033: What about 429 — should that get retried, or nah?

mattwiller (approved):

packages/core/src/client.test.ts
ThatOneBro: That's a good callout... probably not...
We might want to include that one

packages/core/src/client.ts
ThatOneBro: I think it's probably a good idea to have this at some point, want to open an MQ issue?

packages/core/src/client.ts
```diff
@@ -3129,27 +3133,42 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
+
+        // If we got an abort error or exceeded retries, then throw immediately
+        if ((err as Error).name === 'AbortError' || attemptNum === maxRetries) {
+          throw err;
+        }
       }
+
       await sleep(retryDelay);
     }
-    return response as Response;
```
codyebberson, line 3152: Before, we returned the 500 response after max retries, as that could potentially have some debugging details.  That might be worth retaining?

packages/core/src/client.test.ts
codyebberson: The 429 response should include how much time until they're allowed to make the request.  Ideal behavior would be to read that header, and sleep accordingly.  That could be future PR though.

codyebberson (approved):

packages/core/src/client.ts
ThatOneBro: 👍 It is still returned after max retries, see:
https://github.com/medplum/medplum/pull/4723/files#diff-858339789c798051d9c13620bffce8ce860e7ad8ba18dba1c8711eb1f7d51a55R3153

We return before the last iteration awaits so that don't sleep an extra 200ms

packages/core/src/client.test.ts
ThatOneBro: We would probably want some more intelligent logic than just sleeping based on time given, ie. if the time is really long we probably don't want to wait for it in most cases

PR #4737, opened 2024-06-27
feat(otel): add basic metrics for pg

### LINKED ISSUE #4733: Add metrics on db connection pool time
add metrics on db connection pool time for hypothesis on connection starvation

ThatOneBro:
Closes #4733


packages/server/src/healthcheck.ts
```diff
@@ -1,16 +1,30 @@
...
+
+  let startTime = Date.now();
+  const postgresOk = await testPostgres();
+  const dbRoundtripMs = Date.now() - startTime;
+  setGauge('medplum.db.roundtripMs', dbRoundtripMs);
+
+  startTime = Date.now();
+  const redisOk = await testRedis();
+  const redisRoundtripMs = Date.now() - startTime;
+  setGauge('medplum.redis.roundtripMs', redisRoundtripMs);
```
mattwiller, line 19: For consistency, we might want to report all time metrics like this in whole seconds

packages/server/src/healthcheck.ts
```diff
@@ -1,16 +1,30 @@
...
 import { getRedis } from './redis';
 
 export async function healthcheckHandler(_req: Request, res: Response): Promise<void> {
+  setGauge('medplum.db.idleConnections', getDatabasePool().idleCount);
+  setGauge('medplum.db.queriesAwaitingClient', getDatabasePool().waitingCount);
+
+  let startTime = Date.now();
+  const postgresOk = await testPostgres();
+  const dbRoundtripMs = Date.now() - startTime;
+  setGauge('medplum.db.roundtripMs', dbRoundtripMs);
```
mattwiller, line 14: For clarity, we might want to name both of these roundtrip metrics to indicate that they're only for the healthcheck, and aren't measuring all queries — e.g. `medplum.db.healthcheckLatency`.

Additionally, we should figure out how to segment these metrics by host

packages/server/src/healthcheck.ts
ThatOneBro: Since this metric is almost always going to be sub ~100ms, isn't it easier to read in milliseconds? or is there some kind of convention on reporting time in OTel?

mattwiller: It's a convention for metrics reporting, since the observability frontend is usually responsible for displaying the data points with whatever sub-unit (e.g. `s`, `ms` or `µs`) is appropriate

packages/server/src/cloud/aws/metadata.ts
```diff
@@ -0,0 +1,32 @@
+import { randomUUID } from 'crypto';
+import { globalLogger } from '../../logger';
+
+let instanceId: string | undefined = undefined;
+
+export async function fetchInstanceId(): Promise<string> {
```
codyebberson, line 6: Can we use hostname rather than this env var?  I believe the hostname will have largely have the same value (maybe not the exact same string, but something that could be correlated to ECS instance), and that would then be more general purpose for other cloud providers.

Starting to think more about this as we have some contractual obligations for GCP and Azure on the horizon.

packages/server/src/healthcheck.ts
```diff
@@ -1,16 +1,33 @@
...
+  let startTime = Date.now();
+  const postgresOk = await testPostgres();
+  const dbRoundtripMs = Date.now() - startTime;
+  setGauge('medplum.db.healthcheckRTT', dbRoundtripMs / 1000, { instanceId });
+
+  startTime = Date.now();
+  const redisOk = await testRedis();
+  const redisRoundtripMs = Date.now() - startTime;
+  setGauge('medplum.redis.healthcheckRTT', redisRoundtripMs / 1000, { instanceId });
+
```
codyebberson, line 23: Nice, clean 👍

packages/server/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "module": "NodeNext",
-    "moduleResolution": "NodeNext"
+    "moduleResolution": "NodeNext",
+    "lib": ["ESNext"]
```
codyebberson, line 6: 🤔  Do you remember what required this?  Should we add it to the root `tsconfig.json` instead?

I have a soft goal of keeping the per-package `tsconfig.json` files as minimal as possible, to promote unified config as much as possible.

ThatOneBro: There are some APIs that differ between different lib versions, ie. in this case I couldn't use the `cause` field on Error because it requires lib specified to `ES2021` or later

ThatOneBro: We could add it to root as it's probably a sane default since we already have it in `core`

packages/server/src/healthcheck.ts
codyebberson: What does "RTT" stand for?

ThatOneBro: Roundtrip time, if it's not clear we can unabbreviate 🙃

codyebberson (approved):

PR #4740, opened 2024-06-27
Fixes #3892 - "On Behalf Of" header

### LINKED ISSUE #3892: Medplum Auth "On Behalf Of" header
## Use Case
For use when a client is wrapping the Medplum API with their own API service.

On reach request, the client API service will: 
- Create a new `MedplumClient` object
- Use Basic Authentication
- Pass on a new header, `X-Medplum-On-Behalf-Of`, to specify the `ProjectMembership` that of the user they are logging in as. 

## Design Considerations
- Should only be available to `ClientApplications` with `admin` priviledges

## Tasks
- [ ] Server parses new header
- [ ] Add `meta.onBehalfOf` field for history tracking
- [ ] Docs on API passthrough

codyebberson:
Implements "On-Behalf-Of" for Basic Auth

Context: https://github.com/medplum/medplum/issues/3892

Design doc: https://docs.google.com/document/d/118LLwHJ75aPCjXA6QI1j6j3mqlmCJ3-5FXO7v6a5_QQ/edit#heading=h.a3j2r6dlljij

## What is `On-Behalf-Of`?

An emerging pattern we have observed with customers is using Medplum as a backing store, and putting an additional API server or web server in front.

Consider this common Medplum architectural pattern:

image

Here, the Medplum "Customer" has a custom End User App and a Server Side App to enable their business requirements.

In this model, the Customer Server Side App is the only system component that interacts with Medplum Server. In this model, the Medplum Server is more like an internal data store.

The challenge with this model is that it is difficult to take advantage of many Medplum features related to Medplum authentication (author tracking, access policies, etc).

### Option 1: Try to use Medplum Authentication

The Customer Server Side App can attempt to link its "native" authentication method (which can be virtually anything, depending on the technology stack and auth services) and Medplum authentication.  The result would typically be a Medplum access token stored in the Server Side App's "session storage".  Then, when making API requests, the Server Side App would use the Medplum access token.

This has a number of challenges:

* Medplum user authentication was not really designed for this use case (PKCE, etc)
* The Server Side App must manage short lived JWT access tokens and/or refresh tokens
* General anti pattern of trying to managing auth state in two places

### Option 2: Do not use Medplum Authentication

Instead, the Customer Server Side App can simply skip Medplum Authentication and do everything as a single ClientApplication.  This dramatically simplifies the the setup and management of the integration, but then loses out on many key Medplum features:

* No author tracking
* No access control checks

### Option 3: Medplum "On-Behalf-Of"

1. The Customer Server Side App uses a single ClientApplication
   a. Similar to how a web server would use a single database credential
   b. The ClientApplication must be provisioned with "project admin" rights
2. When the Customer Server Side App sends a request, it includes a ProjectMembership id
   a. This ID represents a Medplum ProjectMembership id
   b. It would presumably be stored in the Customer Server Side App database
3. Medplum server would first authenticate the ClientApplication as normal
4. Medplum server would then check for the X-Medplum-On-Behalf-Of header
5. If present, it sets up the Repository instance accordingly
   a. Sets onBehalfOf
   b. Uses the ProjectMembership access policy

## How do you use `On-Behalf-Of`?

When making requests, as a project admin, you can now pass in an extra `X-Medplum-On-Behalf-Of` header.  The value of this header should be a FHIR reference string.  It can be a reference to either:

1. A `ProjectMembership` resource, specifying an exact `ProjectMembership`, including `Project`, profile, and access policy
2. A profile resource such as `Patient` or `Practitioner`.  The server will then search for the `ProjectMembership` accordingly.

Check out the new tests in [onbehalfof.test.ts
](https://github.com/medplum/medplum/pull/4740/files#diff-f200b32443cd02e0edb209e65d41ec838a8b57c9a3eb83187c01f5d02058e1bc)

For example:

```ts
      // Create a patient on behalf of test account 1
      const res1 = await request(app)
        .post(`/fhir/R4/Patient`)
        .set('Authorization', basicAuth)
        .set('X-Medplum', 'extended')
        .set('X-Medplum-On-Behalf-Of', getReferenceString(testAccount1.profile))
        .set('Content-Type', ContentType.FHIR_JSON)
        .send({ resourceType: 'Patient' });
      expect(res1.status).toBe(201);
      expect(res1.body.resourceType).toEqual('Patient');
      expect(res1.headers.location).toContain('Patient');
      expect(res1.headers.location).toContain(res1.body.id);

      const patient1 = res1.body;
      expect(patient1.resourceType).toBe('Patient');
      expect(patient1.meta.author.reference).toEqual(getReferenceString(adminAccount.client));
      expect(patient1.meta.onBehalfOf.reference).toEqual(getReferenceString(testAccount1.profile));
```

## Why implement it with Basic Auth first?

We will add support for `On-Behalf-Of` with access tokens in a future PR.

Based on initial customer interviews, we anticipate that the most common usage will be with Basic Auth, to avoid the complexity of managing expiring access tokens.

Basic Auth sends the Client ID and Client Secret in the `Authorization` header on every request.

## Limitations

* Does not implement "On-Behalf-of" for Bearer tokens / Access tokens.
* Not supported for `/admin/*` APIs
* Not supported for `/auth/*` APIs
* Not supported for `/oauth2/*` APIs

## Implementation Notes

In the PR, you'll notice that we're formalizing the notion of `AuthState`.  There were a bunch of internal helper utilities that took `login`, `project`, and `membership` as parameters.  This PR would have required adding `onBehalfOf` and `onBehalfOfMembership` to those utility methods.  Instead, it seemed that the `AuthState` type already represented all of that.

In the future, I'd like to submit a cleanup PR that fully embraces `AuthState`, and centralizes all of the related utility methods into a single file in `server`.  Open for feedback on that.

### DIFF
--- a/packages/server/src/context.ts
+++ b/packages/server/src/context.ts
@@ -6,7 +6,7 @@ import { NextFunction, Request, Response } from 'express';
 import { getConfig } from './config';
 import { getRepoForLogin } from './fhir/accesspolicy';
 import { Repository, getSystemRepo } from './fhir/repo';
-import { authenticateTokenImpl, isExtendedMode } from './oauth/middleware';
+import { AuthState, authenticateTokenImpl, isExtendedMode } from './oauth/middleware';
 import { parseTraceparent } from './traceparent';
 
 export class RequestContext {
@@ -32,29 +32,28 @@ export class RequestContext {
 const systemLogger = new Logger(write, undefined, LogLevel.ERROR);
 
 export class AuthenticatedRequestContext extends RequestContext {
-  readonly repo: Repository;
-  readonly project: Project;
-  readonly membership: ProjectMembership;
-  readonly login: Login;
-  readonly profile: Reference<ProfileResource>;
-  readonly accessToken?: string;
-
   constructor(
     ctx: RequestContext,
-    login: Login,
-    project: Project,
-    membership: ProjectMembership,
-    repo: Repository,
-    accessToken?: string
+    readonly authState: Readonly<AuthState>,
+    readonly repo: Repository
   ) {
     super(ctx.requestId, ctx.traceId, ctx.logger);
+  }
 
-    this.repo = repo;
-    this.project = project;
-    this.membership = membership;
-    this.login = login;
-    this.profile = membership.profile as Reference<ProfileResource>;
-    this.accessToken = accessToken;
+  get project(): Project {
+    return this.authState.project;
+  }
+
+  get membership(): ProjectMembership {
+    return this.authState.membership;
+  }
+
+  get login(): Login {
+    return this.authState.login;
+  }
+
+  get profile(): Reference<ProfileResource> {
+    return this.authState.membership.profile as Reference<ProfileResource>;
   }
 
   close(): void {
@@ -64,9 +63,7 @@ export class AuthenticatedRequestContext extends RequestContext {
   static system(ctx?: { requestId?: string; traceId?: string }): AuthenticatedRequestContext {
     return new AuthenticatedRequestContext(
       new RequestContext(ctx?.requestId ?? '', ctx?.traceId ?? '', systemLogger),
-      {} as unknown as Login,
-      {} as unknown as Project,
-      {} as unknown as ProjectMembership,
+      {} as unknown as AuthState,
       getSystemRepo()
     );
   }
@@ -95,18 +92,21 @@ export function getAuthenticatedContext(): AuthenticatedRequestContext {
 }
 
 export async function attachRequestContext(req: Request, res: Response, next: NextFunction): Promise<void> {
-  const { requestId, traceId } = requestIds(req);
+  try {
+    const { requestId, traceId } = requestIds(req);
 
-  let ctx = new RequestContext(requestId, traceId);
+    let ctx = new RequestContext(requestId, traceId);
 
-  const authState = await authenticateTokenImpl(req);
-  if (authState) {
-    const { login, membership, project, accessToken } = authState;
-    const repo = await getRepoForLogin(login, membership, project, isExtendedMode(req));
-    ctx = new AuthenticatedRequestContext(ctx, login, project, membership, repo, accessToken);
+    const authState = await authenticateTokenImpl(req);
+    if (authState) {
+      const repo = await getRepoForLogin(authState, isExtendedMode(req));
+      ctx = new AuthenticatedRequestContext(ctx, authState, repo);
+    }
+
+    requestContextStore.run(ctx, () => next());
+  } catch (err) {
+    next(err);
   }
-
-  requestContextStore.run(ctx, () => next());
 }
 
 export function closeRequestContext(): void {
--- a/packages/server/src/fhir/onbehalfof.test.ts
+++ b/packages/server/src/fhir/onbehalfof.test.ts
@@ -1,0 +1,191 @@
+import { ContentType, getReferenceString, unauthorized } from '@medplum/core';
+import { randomUUID } from 'crypto';
+import express from 'express';
+import request from 'supertest';
+import { initApp, shutdownApp } from '../app';
+import { loadTestConfig } from '../config';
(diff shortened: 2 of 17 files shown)


packages/server/src/context.ts
```diff
@@ -32,29 +32,28 @@ export class RequestContext {
...
-  readonly accessToken?: string;
-
   constructor(
     ctx: RequestContext,
-    login: Login,
-    project: Project,
-    membership: ProjectMembership,
-    repo: Repository,
-    accessToken?: string
+    readonly authState: AuthState,
```
rahul1, line 37: Do you want to consider making this ?
```ts
readonly authState: ReadOnly<AuthState>
```

packages/server/src/fhir/onbehalfof.test.ts
```diff
@@ -0,0 +1,161 @@
...
+        .set('X-Medplum-On-Behalf-Of', getReferenceString(testAccount1.profile));
+      expect(res2.status).toBe(200);
+
+      const patient2 = res2.body;
+      expect(patient2.resourceType).toBe('Patient');
+      expect(patient2.id).toBe(patient1.id);
+      expect(patient2.meta.author.reference).toEqual(getReferenceString(adminAccount.client));
+      expect(patient2.meta.onBehalfOf.reference).toEqual(getReferenceString(testAccount1.profile));
+
+      // Create a patient on behalf of test account 1
```
rahul1, line 86:
```suggestion
      // Create a patient on behalf of test account 2
```

packages/server/src/fhir/onbehalfof.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      expect(res3.body.resourceType).toEqual('Patient');
+      expect(res3.headers.location).toContain('Patient');
+      expect(res3.headers.location).toContain(res3.body.id);
+
+      const patient3 = res3.body;
+      expect(patient3.resourceType).toBe('Patient');
+      expect(patient3.meta.author.reference).toEqual(getReferenceString(adminAccount.client));
+      expect(patient3.meta.onBehalfOf.reference).toEqual(getReferenceString(testAccount2.profile));
+
+      // Read the patient on behalf of test account 1
```
rahul1, line 104:
```suggestion
      // Read the patient on behalf of test account 2
```

packages/server/src/oauth/middleware.ts
```diff
@@ -76,7 +87,34 @@ async function authenticateBasicAuth(req: IncomingMessage, token: string): Promi
...
+    if (onBehalfOfHeader.startsWith('ProjectMembership/')) {
+      onBehalfOfMembership = await systemRepo.readReference<ProjectMembership>({ reference: onBehalfOfHeader });
+    } else {
+      onBehalfOfMembership = await systemRepo.searchOne({
+        resourceType: 'ProjectMembership',
+        filters: [
+          { code: 'profile', operator: Operator.EQUALS, value: onBehalfOfHeader },
+          { code: 'project', operator: Operator.EQUALS, value: getReferenceString(project) },
+        ],
+      });
```
rahul1, line 108: out of curiosity, why do we need system repo here? We're only looking for resources scoped to the current project, and the caller should be an admin, so they should have access to `ProjectMemberships`. 

Would save us from having to remember to filter on `project`

rahul1 (approved):

packages/server/src/oauth/middleware.ts
codyebberson: Good suggestion, and that would work in theory.  However, this is in the middle of the auth flow.  We don't have a repo for the current user yet, because this is the code where we're setting up the current user.

We could create an intermediate `Repository` using `getRepoForLogin` yada yada, but it would be a bunch of ceremony for what logically boils down to the current code.

codyebberson: Thought about this more, and I think you're right that the extra ceremony might be worth it, especially when thinking about future proofing for cases where project admins might have access policies that restrict which users they can see.

PR #4741, opened 2024-06-27
`useResource` compatible with a disabled MedplumClient cache

mattlong:
Previously `useResource` with a reference, e.g. `useResrouce({reference: 'Patient/some-id'})`, in conjunction with a `MedplumClient` configured with a very low TTL or disabled cache would result in an infinite render loop due to the way `useReference` interacts with the cache and competing calls to `setResource`.

### DIFF
--- a/packages/react-hooks/src/useResource/useResource.ts
+++ b/packages/react-hooks/src/useResource/useResource.ts
@@ -15,7 +15,9 @@ export function useResource<T extends Resource>(
   setOutcome?: (outcome: OperationOutcome) => void
 ): T | undefined {
   const medplum = useMedplum();
-  const [resource, setResource] = useState<T | undefined>(getInitialResource(medplum, value));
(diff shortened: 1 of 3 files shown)


codyebberson:
Deploying to staging...

codyebberson (approved):

packages/react-hooks/src/useResource/useResource.ts
```diff
@@ -49,6 +48,8 @@ export function useResource<T extends Resource>(
             }
           }
         });
+    } else {
+      setResourceIfChanged(newValue);
     }
 
     return (() => (subscribed = false)) as () => void;
```
yang-seen, line 55: why is resource in the dependency array here in useEffect? It makes sense this should depend on value. Reference is not referenced in the useEffect body

mattlong: Good catch, @yang-seen! Pushing out a followup PR to remove it.

PR #4747, opened 2024-06-28
fix(subscriptions): don't log if AP not satisfied for WS subs

ThatOneBro:
Since we already enforce `AccessPolicy`s for WebSocket Subscriptions, we don't really need to log when an `AccessPolicy` is not satisfied since the Subscription does not create a notification when `subscription.channel.type === 'websocket'`


codyebberson (approved):
👍

PR #4749, opened 2024-06-28
fix(core): add `storageInitFailed` event to `MedplumClient`

ThatOneBro:


packages/core/src/client.ts
```diff
@@ -844,7 +845,11 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
           this.initComplete = true;
           this.dispatchEvent({ type: 'storageInitialized' });
         })
-        .catch(console.error);
+        .catch((err: Error) => {
+          console.error(err);
```
codyebberson, line 849: Mental note - we should have a team discussion on our philosophy on `console.log` and `console.error` throughout `@medplum/core` and `@medplum/react`.  I'm guilty of most of them.  It'd be nice to have a clear recommended pattern here (logger, etc).

codyebberson (approved):
Nice 👍

packages/core/src/client.ts
ThatOneBro: I think although `Logger` is part of `@medplum/core`, we don't have any usage of it anywhere within the package except the recently ported function from `@medplum/server`, `resourceMatchesSubscriptionCriteria`. Interesting. I have always used `console.{ log, error }` in `@medplum/core` because of that

Issue #4681, opened 2024-06-17
Clean up CLI global options


ThatOneBro:
Currently there are a lot of global options that show up on every command which make the help menu for commands very noisy and not as useful as it could be. There is probably a better place for these options and/or a more tightly scoped subset of commands where they should live

<img width="912" alt="Screenshot 2024-06-17 at 1 38 16 PM" src="https://github.com/medplum/medplum/assets/1592008/6d5ad7d5-63c8-4651-8225-59c0e0795de2">

ThatOneBro:
Ast of #4750, this is now the structure of the `--help` command, with global options separate from the command-specific options

![Image](https://github.com/user-attachments/assets/ea7fbe19-473b-4cc8-8907-33af8124ae00)

PR #4750, opened 2024-06-28
feat(cli): add agent commands

### LINKED ISSUE #3901: Medplum Agent Commands exposed in CLI
## Use Case
Some DevOps engineers primarily work with CLI tooling and shell scripts, rather than rest API calls. 

To help them, we want to expose all the Agent maintenance commands as CLI commands. This will also help to integrate into monitoring tools such as [BetterStack](https://betterstack.com/).

## Proposal

```
medplum agent push <args..> [--wait-for-response]
medplum agent ping <host> [pingCount]
medplum agent status <agentId>
medplum agent bulk-status [criteria]

# When added:
medplum agent reload-config <agentId1>[agentId2..agentIdN]
medplum agent upgrade <version> <agentId1>[agentId2..agentIdN]
```

ThatOneBro:
Closes #3901 and #4681


reshmakh:
Planned for today

packages/cli/src/agent.ts
```diff
@@ -0,0 +1,487 @@
...
+    let pingResult: string;
+    try {
+      pingResult = (await medplum.pushToAgent(agentRef, ipOrDomain, `PING ${count}`, ContentType.PING, true, {
+        maxRetries: 0,
+      })) as string;
+    } catch (err) {
+      throw new Error('Unexpected response from agent while pinging', { cause: err });
+    }
+
+    console.info(pingResult);
```
mattwiller, line 124: Suggest tucking statements inside `try` for cohesion:
```suggestion
    try {
      const pingResult = (await medplum.pushToAgent(agentRef, ipOrDomain, `PING ${count}`, ContentType.PING, true, {
        maxRetries: 0,
      })) as string;
      console.info(pingResult);
    } catch (err) {
      throw new Error('Unexpected response from agent while pinging', { cause: err });
    }
```

packages/cli/src/agent.ts
```diff
@@ -0,0 +1,487 @@
...
+      severity: issue.severity,
+      code: issue.code,
+      details: issue.details?.text ?? 'No details to show',
+    } satisfies FailedRow;
+    failedRows.push(row);
+  }
+
+  console.info();
+  console.info(`${successfulRows.length} successful response(s):`);
+  console.info();
```
mattwiller, line 304:
```suggestion
  console.info(`\n${successfulRows.length} successful response(s):\n`);
```

packages/cli/src/agent.test.ts
```diff
@@ -0,0 +1,1326 @@
...
+    jest.clearAllMocks();
+    medplum = new MockClient();
+    (createMedplumClient as unknown as jest.Mock).mockImplementation(async () => medplum);
+
+    process.env = { ...env };
+    medplumGetSpy = jest.spyOn(medplum, 'get');
+  });
+
+  afterEach(() => {
+    process.env = env;
```
mattwiller, line 69: Are both these assignments to `process.env` needed?

packages/cli/src/agent.test.ts
```diff
@@ -0,0 +1,1326 @@
...
+                { name: 'status', valueString: 'unknown' },
+                { name: 'version', valueString: 'unknown' },
+              ],
+            } satisfies Parameters,
+          ];
+        });
+
+        await expect(main(['node', 'index.js', 'agent', 'status', agentId])).resolves.toBeUndefined();
+        expect(medplumGetSpy).toHaveBeenCalledWith(
+          // NOTE(ThatOneBro 28 Sept 2024):
```
mattwiller, line 99: 👍

mattwiller (approved):

packages/cli/src/agent.test.ts
ThatOneBro: I think this `afterEach` should actually be an `afterAll` probably, to reset the `process.env` at the end

Issue #4726, opened 2024-06-25
Research CCDA libraries


codyebberson:

codyebberson:
After researching many options, we elected to DIY.  `@medplum/ccda` is now published.

Issue #3005, opened 2023-10-08
Certify and test Transitions of care § 170.315(b)(1) 


reshmakh:
* [Official documentation](https://www.healthit.gov/topic/certification-ehrs/2015-edition-test-method/2015-edition-cures-update-base-electronic-health-record-definition)
* [Proctor sheet](https://docs.google.com/document/d/1i9W0U3BlGUuvY_Dm5Phni8GIIeTL2ls7jFxfR57-0eE/edit)

reshmakh:
(b)(1) Specific sheet: https://www.healthit.gov/test-method/transitions-care
Test data: https://codeload.github.com/onc-healthit/2015-edition-cures-update-data/zip/master
Edge Testing Tool: https://ett.healthit.gov/ett/#/validators/ccdar3

reshmakh:
# Requirement Summary
* **Ability to Send and Receive Transitions of Care/Referral Summaries**: The Health IT must be capable of sending and receiving transition of care (ToC)/referral summaries using specified protocols such as IHE XDR or SMTP for sending and IHE XDR, SMTP, POP3, or IMAP4 for receiving.

* **Compliance with Security Protocols**: The exchange of ToC/referral summaries should be conducted using secure protocols like TLS/SASL or a secure network, such as a VPN.

* **Support for XDM Packages**: If the Health IT Module is certified to an SMTP-based edge protocol, it must be able to receive and make available the contents of an XDM package created in accordance with the IHE IT Infrastructure Technical Framework.

* **Detection of Valid and Invalid ToC/Referral Summaries**: The system should be capable of detecting valid and invalid ToC/referral summaries upon receipt of C-CDA documents formatted to C-CDA Release 1.1 and 2.1.

* **Display of Human-readable C-CDA**: The Health IT should be able to display a human-readable C-CDA for both C-CDA Releases 1.1 and 2.1.

* **User Interface Customization**: The system should allow users to choose to display only the data within a particular C-CDA section, set a preference for section display order, and set the initial number of sections to be displayed.

* **Creation of C-CDA Documents**: The Health IT must be able to create a C-CDA document (formatted to Release 2.1) that includes USCDI data, encounter diagnoses, cognitive status, functional status, and, in certain settings, additional information like the reason for referral or discharge instructions.

* **Data Elements for Patient Matching**: The system should include certain data elements to assist with patient matching in ToC/referral summary documents.

## Resources 
[...]

PR #4768, opened 2024-07-01
FHIR Mapper shorthand rules using 'types' groups

codyebberson:
In service of https://github.com/medplum/medplum/issues/3005 and https://github.com/medplum/medplum/issues/4726

The new test "Using types" is based on the FHIR Mapping Language tutorial step 10: https://build.fhir.org/mapping-tutorial.html#step10

Consider this map:

```
      group tutorial(source src : TLeft, target tgt : TRight) {
        src.aa -> tgt.aa;
      }
      group ab_content(source src : TLeftInner, target tgt : TRightInner) <<types>> {
        src.ab -> tgt.ac;
      }
```

The mapping engine needs to be smart enough to do the following:
1. Recognize that `src.aa` is type `TLeftInner`
2. Recognize that `tgt.aa` is type `TRightInner`
3. Recognize that the rule `src.aa -> tgt.aa;` is a shorthand rule which should automatically search for a `<<types>>` group
4. Search for the `<<types>>` group with corresponding types
5. Apply that group

### DIFF
--- a/packages/core/src/fhirmapper/transform.errors.test.ts
+++ b/packages/core/src/fhirmapper/transform.errors.test.ts
@@ -77,8 +77,8 @@ describe('FHIR Mapper transform - errors', () => {
       }
     `;
 
-    const input = [toTypedValue({ a: 'abc' })];
-    const expected = [toTypedValue({})];
+    const input = [{ type: 'TLeft', value: { a: 'abc' } }];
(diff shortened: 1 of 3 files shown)


rahul1 (approved):
I love this feature. This has the potential to really level up our HL7 integrations as well

PR #4772, opened 2024-07-02
Use reader database instance for readonly graphql queries

codyebberson:
### Update 2024-07-03: Start with `$graphql` only

* Based on team discussions, we want to start using the reader instance only for readonly graphql queries
* This is a little tricky, due to the code split between `server` and `fhir-router`
* The `FhirRepository` interface now has a `setMode()` method which accepts `RepositoryMode.READER` or `RepositoryMode.WRITER`
* The server `Repository` starts in `RepositoryMode.WRITER`, and will therefore use the RDS writer instance by default for everything
* If the request is a `$graphql` request, we put the `Repository` in `RepositoryMode.READER`, meaning that it is eligible for the RDS reader instance
* Inside the graphql handler, we check for "mutations", and switch back to `RepositoryMode.WRITER` if necessary
* The net results is that we should only ever consider the reader instance under the following conditions:
   * Reader instance is configured in the server config settings
   * The request is a `$graphql` query
   * The GraphQL body does not include "mutation"

One slightly ugly part of this is there are now two enums, `RepositoryMode` and `DatabaseMode`, both with `READER` and `WRITER`.  These could be combined?

I don't love the flip-flopping between reader/writer modes, but I think it came out the most straightforward from API level.  Open to feedback.

The long term intent, as we gain confidence here, is that eventually server `Repository` could start in `READER` mode, and only switch to `WRITER` if the request makes any writes (as described below).  I think the current code is scaffolded nicely for that goal state.

--------

### Original remarks

How it works:
* During server startup, check for presence of a "reader" database config
* If there is a "reader" database config, setup a 2nd database pool for that instance
* Anytime the server requests a database client, it must specify "reader" or "writer"
* If the server requests a "reader" and the "reader" pool is available, then use a "reader" connection
* Database transactions always use "writer", including "reader" queries once inside a transaction

Needs more testing:
* We should be able to test this easily by creating a 2nd postgres user `medplum_readonly` or whatever
* I want to do some deeper testing on requests that potentially use a mix of "reader" and "writer"
   * The fear is the following situation:
      1. Create a resource using a "writer" connection
      2. Immediately search for a resource using a "reader" connection
      3. But the "reader" search executes during the brief window where "reader" and "writer" are out of sync
      4. Weird results follow
   * To resolve that, we may want to add some kind of logic such as: "once you use a writer connection, you have to use writer connections for the rest of the request"

### DIFF
--- a/packages/server/src/config.ts
+++ b/packages/server/src/config.ts
@@ -26,6 +26,8 @@ export interface MedplumServerConfig {
   approvedSenderEmails?: string;
   database: MedplumDatabaseConfig;
   databaseProxyEndpoint?: string;
+  readonlyDatabase: MedplumDatabaseConfig;
+  readonlyDatabaseProxyEndpoint?: string;
   redis: MedplumRedisConfig;
   emailProvider?: 'none' | 'awsses' | 'smtp';
   smtp?: MedplumSmtpConfig;
@@ -187,6 +189,11 @@ export async function loadTestConfig(): Promise<MedplumServerConfig> {
   config.database.port = process.env['POSTGRES_PORT'] ? Number.parseInt(process.env['POSTGRES_PORT'], 10) : 5432;
   config.database.dbname = 'medplum_test';
   config.database.runMigrations = false;
+  config.readonlyDatabase = {
+    ...config.database,
+    username: 'medplum_test_readonly',
+    password: 'medplum_test_readonly',
+  };
   config.redis.db = 7; // Select logical DB `7` so we don't collide with existing dev Redis cache.
   config.redis.password = process.env['REDIS_PASSWORD_DISABLED_IN_TESTS'] ? undefined : config.redis.password;
   config.approvedSenderEmails = 'no-reply@example.com';
--- a/packages/server/src/fhir/operations/dbstats.ts
+++ b/packages/server/src/fhir/operations/dbstats.ts
@@ -2,7 +2,7 @@ import { allOk } from '@medplum/core';
 import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
 import { OperationDefinition } from '@medplum/fhirtypes';
 import { requireSuperAdmin } from '../../admin/super';
-import { getDatabasePool } from '../../database';
+import { DatabaseMode, getDatabasePool } from '../../database';
 import { buildOutputParameters } from './utils/parameters';
 
 const operation: OperationDefinition = {
@@ -21,7 +21,7 @@ const operation: OperationDefinition = {
 export async function dbStatsHandler(_req: FhirRequest): Promise<FhirResponse> {
   requireSuperAdmin();
 
-  const client = getDatabasePool();
+  const client = getDatabasePool(DatabaseMode.READER);
   const sql = `
       SELECT * FROM (
         SELECT table_schema, table_name, pg_relation_size('"'||table_schema||'"."'||table_name||'"') AS table_size
--- a/packages/server/src/fhir/operations/dbstats.ts
+++ b/packages/server/src/fhir/operations/dbstats.ts
@@ -21,7 +21,7 @@ const operation: OperationDefinition = {
 export async function dbStatsHandler(_req: FhirRequest): Promise<FhirResponse> {
   requireSuperAdmin();
 
-  const client = getDatabasePool(DatabaseMode.READER);
+  const client = getDatabasePool(DatabaseMode.WRITER);
   const sql = `
       SELECT * FROM (
         SELECT table_schema, table_name, pg_relation_size('"'||table_schema||'"."'||table_name||'"') AS table_size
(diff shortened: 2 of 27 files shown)


packages/server/src/config.ts
```diff
@@ -26,6 +26,7 @@ export interface MedplumServerConfig {
   approvedSenderEmails?: string;
   database: MedplumDatabaseConfig;
   databaseProxyEndpoint?: string;
+  readonlyDatabase: MedplumDatabaseConfig;
```
mattlong, line 29: Seems like we'd want a readonly equivalent of `databaseProxyEndpoint`?

mattlong (commented):
After a quick first readthrough, this all seems reasonable.

> I want to do some deeper testing on requests that potentially use a mix of "reader" and "writer"
> To resolve that, we may want to add some kind of logic such as: "once you use a writer connection, you have to use writer connections for the rest of the request"

Replication lag could of course be longer than a single request. To be more conservative in the short term, what do we think about utilizing Redis to peg clients that have used the writer in the last N milliseconds and such that they'll still get the writer even if they request the reader?

packages/server/src/config.ts
codyebberson: Yeah, agreed.  I was curious if RDS Proxy has any of this built in.  That's on my list of things to research.

ThatOneBro:
Idea I'm not married too but might make sense:

What if we tied the readonly / read-write behavior to the `Repository` itself? That way it is explicit if you are in `read-only` mode or `read-write` mode

codyebberson:
Tests failing because Github Actions doesn't use `init_test.sql` 🤔 

https://github.com/medplum/medplum/actions/runs/9765092525/job/26955125449?pr=4772

image

There's probably a way to do this using "Initialization Scripts", but need to research how to set that up: https://hub.docker.com/_/postgres

codyebberson:
> What if we tied the readonly / read-write behavior to the `Repository` itself? That way it is explicit if you are in `read-only` mode or `read-write` mode
> 
> Repository defaults to `read-only` mode but falls back to `read-write` if any create, update, or delete is detected. Only searches or readResource are allowed in `read-only`

Yeah, I think that is where this is headed.

The `getDatabasePool(mode)` method will still need to exist, for the handful of cases that execute SQL queries outside of `Repository`.

Within `Repository`, I think you're right that you can request `READER` or `WRITER`, but as soon as you use `WRITER` once, you get a `WRITER` for the duration of the `Repository` instance.

We still may need to set a similar flag at the request context level, in particular for the case where we create a temp `systemRepo`, and it has to play nice with the user's `repo`.  I don't have a great example off the top of my head, but something like:

1. `systemRepo` creates a resource
2. `repo` needs to read that resource, and executes a search
   * With current logic, that search may be in `READER` mode, which could lead to incoherent results
   * Arguably the "right" thing to do here is never rely on create-then-search behavior, and just pass around the resources
   * But if that proves too thorny, then the model would be to inspect request context to see if there have been any writes

packages/server/src/fhir/repo.ts
```diff
@@ -216,12 +217,21 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
+    // However, for now, we default to writer and only use reader mode for requests guaranteed not to have consistency risks
+    this.mode = RepositoryMode.WRITER;
   }
 
   clone(): Repository {
     return new Repository(this.context);
   }
 
+  setMode(mode: RepositoryMode): void {
+    this.mode = mode;
```
mattlong, line 232: Given the comment below, should `setMode(RepositoryMode.READER)` throw if `getDatabaseClient` has already been called when `this.mode === RepositoryMode.WRITER`?

mattlong (changes requested):

packages/server/src/fhir/operations/dbstats.ts
```diff
@@ -21,7 +21,7 @@ const operation: OperationDefinition = {
 export async function dbStatsHandler(_req: FhirRequest): Promise<FhirResponse> {
   requireSuperAdmin();
 
-  const client = getDatabasePool();
+  const client = getDatabasePool(DatabaseMode.READER);
```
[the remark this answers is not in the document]
ThatOneBro: Maybe we should have both?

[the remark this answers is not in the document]
codyebberson: Good catch.

packages/server/src/fhir/repo.ts
codyebberson: Hm, interesting.

At present, the `mode` is initialized to `WRITER` (pessimistically).  We then set it to `READER` when we detect that it's a graphql request.

We could instead initialize to `undefined`?  Not sure that's much better? 🤷

ThatOneBro: Yeah I think once the mode is set to `WRITER` it should probably be sticky. Initializing to `READER` but as soon as we see we are not in a "readonly gql query" we set it to `WRITER` where it is stuck for the rest of this Repository's lifetime might be want we want

ThatOneBro (approved):

PR #4776, opened 2024-07-02
tweak(repo): each repo can only do one query at a time

ThatOneBro:
This is just testing what it would be like if each `Repository` could only issue one query at a time


ThatOneBro:
Closed due to inactivity / bitrot

PR #4780, opened 2024-07-02
Temporarily remove extraneous tracing spans

mattwiller:


codyebberson (approved):

codyebberson:
Curious if there's a good way to make this config-file driven.  For example, how to make it easy to enable/disable the AWS monitoring when not running on AWS, etc.

PR #4781, opened 2024-07-02
Add containerInsights to CDK config options

mattlong:


packages/core/src/config.ts
```diff
@@ -143,6 +143,7 @@ export interface MedplumInfraConfig {
       [key: string]: string;
     };
   }[];
+  containerInsights?: boolean;
```
codyebberson, line 146: This needs a corresponding entry in `MedplumSourceInfraConfig` using the `ValueOrExternalSecret` thing.

Some customers have built adapters where all infra config settings are loaded from external stores such as Hashicorp Vault, so we need to expose all config settings through that interface as well.

@ThatOneBro  - Please correct me if that's wrong.

codyebberson (changes requested):

packages/core/src/config.ts
mattlong: Added

ThatOneBro (approved):

codyebberson (approved):

PR #4791, opened 2024-07-03
Document future GraphQL query limits

mattwiller:


codyebberson:
Direct link to preview: https://medplum-www-git-graphql-limit-docs-medplum.vercel.app/docs/graphql/basic-queries#query-limits

packages/docs/docs/graphql/basic-queries.mdx
```diff
@@ -296,6 +296,22 @@ This query searches for a list of `Patients` named `"Eve"`, living in `"Philadel
   </MedplumCodeBlock>
 </details>
 
+## Query Limits
+
+Because complex GraphQL queries may require many database operations, the cost of the query is estimated before execution,
```
codyebberson, line 301:
```suggestion
Because complex GraphQL queries may be computationally expensive, the cost of the query is estimated before execution,
```

codyebberson (commented):
Big picture:
1. I think this should be a separate page inside the `docs/graphql` folder
2. I think we should probably write a longer form article, such as https://shopify.engineering/rate-limiting-graphql-apis-calculating-query-complexity
3. We may even want to reference the Shopify guidelines explicitly?  It may be somewhat weird, but I anticipate some pushback from some heavy graphql users, and I would like to make an appeal to authority / point to this as "industry best practices".

With regards to the proposed formula itself...
* I appreciate the mathematical rigor
* The formula seems quite complicated, and I fear that it will be difficult to understand intuitively for users
* What is the proposed relationship between `_count` and `C`?
* If `C` is already a separate formula, then moving the `1.5` term into that formula (I think that works, right?  `C^(1.5*D) === C^1.5*C^D`?) might make this look cleaner to really emphasize it's `C^D`, and depth is expensive

codyebberson:
Hm, re-reading the Shopify doc, it's probably "yes and"

The Shopify doc doesn't have any mention of "nested" or "depth".  So perhaps we could start with their doc, and include your proposed language on how depth causes exponential growth in cost.

https://shopify.engineering/rate-limiting-graphql-apis-calculating-query-complexity

packages/docs/docs/graphql/basic-queries.mdx
```diff
@@ -296,6 +296,22 @@ This query searches for a list of `Patients` named `"Eve"`, living in `"Philadel
...
+following general formula:
+
+- Each reference link (i.e. `Reference.resource` field) in the query costs _D_ units, where D is the nesting depth of the
+  reference field in the overall query
+- Each search operation (e.g. `*List` fields that take a `_count` argument) costs `C^(1.5 * D)` units, where C is a small
+  number (between 1 and 10) that grows with the requested `_count`, and D is the nesting depth of the search field in the
+  overall query
+
+These costs are added together, and compared to a configurable server limit in order to determine if the query should be
+executed against the database.
```
rahul1, line 313: One exampel with a 2-3 level deep query will help solidify this. Can you add a code block inside a <details></details> tag, with the total score up top, and comments for how the depth adds to the complexity?

rahul1 (approved):

PR #4811, opened 2024-07-05
Use DataLoader for nested GraphQL searches

mattlong:
### Update 2
Disregard Update 1; it used a different approach than is described here.

As mentioned in the last couple comments on this PR's thread, we use a lateral join in conjunction with a `VALUES` expression to supply the reference value inputs and build up the inner `SelectQuery` on the resource being fetched using just about all of the same code paths and logic as a regular search.

### Update 1
1. Introduced an `AbstractSelectQuery` that both `SelectQuery` and the new `SelectPerReferenceQuery` implement; albeit with a lot of the methods simply throwing a `NotSupportedError`.
2. The signatures of most functions of the search-related code that used to accept a `SelectQuery` parameter now accept an `AbstractSelectQuery` instead. This allows the new query type to be passed around to all the search helper functions _with the expectation that it may result in an error being thrown_.
3. `searchByReferenceImpl` was reworked to catch `NotSupportError` and fallback to providing the results via the original `searchImpl`, i.e. in N queries instead of one.

I'm feeling pretty good about this approach as it allows us to incrementally expand the types of searches that `searchByReference` can handle while allowing us to push it out now to handle the simple nested resource queries that have been hammering us lately.

Remaining action items:

- [ ] There are a few places where `SelectQuery.predicate.expressions` is accessed directly to add new expressions. I'd like to make `BaseQuery.predicate` private and introduce a method to push new expressions instead so that behavior can also be marked as unsupported in `SelectPerReferenceQuery` for now.
- [ ] A handful of TODO{mattlong} in the code that should be relatively easy lifts.
- [ ] Add docstrings to the code.
- [ ] Write some more tests
- [ ] Implement `MemoryRepository.searchByReference`
- [ ] Add some logging/metrics; especially around the fallback logic to help guide future improvements.

### Original description
This is a very experimental proof of concept.

Nested searches are done in 1 sql query per nested search instead of N queries by using the JOIN LATERAL and ROW_NUMBER() features of postgres in conjunction with more `DataLoader`-based code in our GraphQL resolvers. While the  Lots of load testing and validating that this approach does indeed yield performance improvements at scale.

One of the cooler parts was being able to build the new query by building on top of `SelectQuery`.

Sample GQL query:
```GQL
{
  PatientList(_count: 1000) {
    id
    ObservationList(_reference: subject, _count: 3, _sort: "_lastUpdated") {
      id
      status
      meta {
        lastUpdated
      }
      code {
        text
      }
    }
    TaskList(_reference: patient) {
      id
      status
    }
  }
}
```

Share of resulting SQL:
```SQL
SELECT
  "results"."id",
  "results"."content",
  "references"."reference"
FROM
  (
    SELECT DISTINCT "subject" AS "reference"
    FROM "Observation"
    WHERE ("Observation"."subject" IN ('Patient/4c93dfd3-878f-4dbc-8c30-46d146105bc3', 'Patient/afc5f9a2-529a-4ae4-a3f4-263de813a3a0', 'Patient/fc6ac399-4e16-4454-897c-7dba5a04120b') AND
           "Observation"."deleted" = false
          )
  ) AS "references"
  JOIN LATERAL
  (
    SELECT
      "ranked"."id",
      "ranked"."content"
    FROM
      (
        SELECT
          "Observation"."id",
          "Observation"."content",
          ROW_NUMBER() OVER (ORDER BY "lastUpdated") AS rn
        FROM "Observation"
        WHERE (
          "Observation"."subject" = "references"."reference" AND
          "Observation"."deleted" = false
        )
      ) AS "ranked"
    WHERE
      "ranked"."rn" <= 3
  ) AS "results" ON true
```

### DIFF
--- a/packages/fhir-router/src/graphql/graphql.test.ts
+++ b/packages/fhir-router/src/graphql/graphql.test.ts
@@ -15,6 +15,7 @@ import {
   ExplanationOfBenefit,
   Extension,
   HumanName,
+  Observation,
   OperationOutcome,
   Patient,
   SearchParameter,
@@ -28,6 +29,7 @@ import { getRootSchema, graphqlHandler } from './graphql';
 const repo = new MemoryRepository();
 let binary: Binary;
 let patient: Patient;
+let patient2: Patient;
 let serviceRequest: ServiceRequest;
 let encounter1: Encounter;
 let encounter2: Encounter;
@@ -55,6 +57,16 @@ describe('GraphQL', () => {
         {
           contentType: 'image/jpeg',
           url: getReferenceString(binary),
+        },
+      ],
+    });
+
+    patient2 = await repo.createResource<Patient>({
+      resourceType: 'Patient',
+      name: [
+        {
+          given: ['Bill'],
+          family: 'Miller',
         },
       ],
     });
@@ -488,6 +500,30 @@ describe('GraphQL', () => {
   });
 
   test('Reverse lookup with _reference', async () => {
+    await repo.createResource<Observation>({
+      resourceType: 'Observation',
+      status: 'final',
+      subject: createReference(patient),
+      code: {
+        text: 'Blood pressure',
+      },
+    });
+    await repo.createResource<Observation>({
+      resourceType: 'Observation',
+      status: 'final',
+      subject: createReference(patient2),
+      code: {
+        text: 'Blood pressure',
+      },
+    });
+    await repo.createResource<Observation>({
+      resourceType: 'Observation',
+      status: 'final',
+      subject: createReference(patient2),
+      code: {
+        text: 'Heart rate',
+      },
+    });
     const request: FhirRequest = {
       method: 'POST',
       pathname: '/fhir/R4/$graphql',
@@ -496,9 +532,9 @@ describe('GraphQL', () => {
       body: {
         query: `
       {
-        PatientList(_count: 1) {
+        PatientList(_count: 2) {
           id
-          ObservationList(_reference: subject) {
+          ObservationList(_reference: subject, _sort: "_lastUpdated") {
             id
             status
             code {
--- a/packages/fhir-router/src/repo.test.ts
+++ b/packages/fhir-router/src/repo.test.ts
@@ -1,5 +1,7 @@
 import {
   badRequest,
+  createReference,
+  getReferenceString,
   indexSearchParameterBundle,
   indexStructureDefinitionBundle,
   notFound,
@@ -7,7 +9,7 @@ import {
   parseSearchRequest,
 } from '@medplum/core';
 import { readJson } from '@medplum/definitions';
-import { Bundle, Observation, Patient, ResourceType, SearchParameter } from '@medplum/fhirtypes';
+import { Bundle, Observation, Patient, Resource, ResourceType, SearchParameter } from '@medplum/fhirtypes';
 import { randomInt, randomUUID } from 'crypto';
 import { MemoryRepository } from './repo';
 
@@ -138,5 +140,103 @@ describe('MemoryRepository', () => {
     const resourcesListAfter = await Promise.all(resourceTypes.map((rt) => repo.searchResources({ resourceType: rt })));
     const actualResourceCountAfter = resourcesListAfter.reduce((count, resources) => (count += resources.length), 0);
     expect(actualResourceCountAfter).toBe(0);
+  });
+
+  describe('searchByReference', () => {
+    async function createPatients(repo: MemoryRepository, count: number): Promise<Patient[]> {
+      const patients = [];
+      for (let i = 0; i < count; i++) {
+        patients.push(await repo.createResource<Patient>({ resourceType: 'Patient' }));
+      }
+      return patients;
+    }
+
+    async function createObservations(repo: MemoryRepository, count: number, patient: Patient): Promise<Observation[]> {
+      const resources = [];
+      for (let i = 0; i < count; i++) {
+        resources.push(
+          await repo.createResource<Observation>({
+            resourceType: 'Observation',
+            subject: createReference(patient),
+            valueString: i.toString(),
+          } as Observation)
+        );
+      }
+      return resources;
+    }
+
+    function zip<A, B>(a: A[], b: B[]): [A, B][] {
+      return a.map((k, i) => [k, b[i]]);
+    }
+
+    function expectResultsContents<Parent extends Resource, Child extends Resource>(
+      parents: Parent[],
+      childrenByParent: Child[][],
+      { count, offset }: { count: number; offset: number },
+      results: Record<string, Child[]>
+    ): void {
+      expect(Object.keys(results)).toHaveLength(parents.length);
+      for (const [parent, children] of zip(parents, childrenByParent)) {
+        const result = results[getReferenceString(parent)];
+        expect(result).toHaveLength(Math.min(children.length - offset, count));
+        for (const child of result) {
+          expect(children.map((c) => c.id)).toContain(child.id);
+        }
+      }
+    }
+
+    test('basic search by reference', async () => {
+      const patients = await createPatients(repo, 3);
+      const patientObservations = [
+        await createObservations(repo, 2, patients[0]),
+        await createObservations(repo, 3, patients[1]),
+        await createObservations(repo, 4, patients[2]),
+      ];
+
+      const count = 3;
+      const offset = 1;
+      const observation = patientObservations[0][offset];
+      const result = await repo.searchByReference<Observation>(
+        { resourceType: 'Observation', count, offset },
+        'subject',
+        patients.map((p) => getReferenceString(p))
+      );
+
+      expectResultsContents(patients, patientObservations, { count, offset }, result);
+      const resultRepoObservation = result[getReferenceString(patients[0])][0];
+      expect(resultRepoObservation).toEqual(observation);
+      expect(resultRepoObservation.meta?.tag).toBeUndefined();
+    });
+
+    test('with sort', async () => {
+      const patients = await createPatients(repo, 2);
+      const patientObservations = [
+        await createObservations(repo, 3, patients[0]),
+        await createObservations(repo, 2, patients[1]),
+      ];
+      const count = 3;
+      const offset = 0;
+
+      // descending
+      const resultDesc = await repo.searchByReference<Observation>(
+        { resourceType: 'Observation', count, offset, sortRules: [{ code: 'value-string', descending: true }] },
+        'subject',
+        patients.map((p) => getReferenceString(p))
+      );
+
+      expectResultsContents(patients, patientObservations, { count, offset }, resultDesc);
+      expect(resultDesc[getReferenceString(patients[0])].map((o) => o.valueString)).toEqual(['2', '1', '0']);
+      expect(resultDesc[getReferenceString(patients[1])].map((o) => o.valueString)).toEqual(['1', '0']);
+
+      // ascending
+      const resultAsc = await repo.searchByReference<Observation>(
+        { resourceType: 'Observation', count, sortRules: [{ code: 'value-string', descending: false }] },
+        'subject',
+        patients.map((p) => getReferenceString(p))
+      );
+      expectResultsContents(patients, patientObservations, { count, offset }, resultAsc);
+      expect(resultAsc[getReferenceString(patients[0])].map((o) => o.valueString)).toEqual(['0', '1', '2']);
+      expect(resultAsc[getReferenceString(patients[1])].map((o) => o.valueString)).toEqual(['0', '1']);
+    });
   });
 });
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -1,5 +1,7 @@
 import {
+  Filter,
   OperationOutcomeError,
+  Operator,
   SearchRequest,
   SortRule,
   allOk,
@@ -150,6 +152,12 @@ export interface FhirRepository<TClient = unknown> {
    * @returns The search results.
    */
   search<T extends Resource>(searchRequest: SearchRequest<T>): Promise<Bundle<T>>;
+
+  searchByReference<T extends Resource>(
+    searchRequest: SearchRequest<T>,
+    referenceField: string,
+    references: string[]
+  ): Promise<Record<string, T[]>>;
 
   /**
    * Searches for a single FHIR resource.
@@ -407,6 +415,17 @@ export class MemoryRepository extends BaseRepository implements FhirRepository {
       entry,
       total: result.length,
     };
+  }
+
+  async searchByReference<T extends Resource>(
+    searchRequest: SearchRequest<T>,
+    referenceField: string,
+    references: string[]
+  ): Promise<Record<string, T[]>> {
+    const combinedFilter: Filter = { code: referenceField, operator: Operator.IN, value: references.join(',') };
+    searchRequest.filters ??= [];
+    searchRequest.filters.push(combinedFilter);
+    return {};
   }
 
   async deleteResource(resourceType: string, id: string): Promise<void> {
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -1,5 +1,4 @@
 import {
-  Filter,
   OperationOutcomeError,
   Operator,
   SearchRequest,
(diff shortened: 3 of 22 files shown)


packages/fhir-router/src/repo.ts
```diff
@@ -409,6 +417,17 @@ export class MemoryRepository extends BaseRepository implements FhirRepository {
...
 
+  async searchByReference<T extends Resource>(
+    searchRequest: SearchRequest<T>,
+    referenceField: string,
+    references: string[]
+  ): Promise<Record<string, T[]>> {
+    const combinedFilter: Filter = { code: referenceField, operator: Operator.IN, value: references.join(',') };
+    searchRequest.filters ??= [];
+    searchRequest.filters.push(combinedFilter);
+    return {};
```
codyebberson, line 428: Hm.. how hard would it be to implement this in `MemoryRepository`?

I suppose we could just write the brute force version, as this would really only be used in unit tests and Storybook graphql examples.

packages/server/src/fhir/sql.ts
```diff
@@ -613,6 +621,116 @@ export class SelectQuery extends BaseQuery implements Expression {
   }
 }
 
+export class SelectPerReferenceQuery extends BaseQuery implements Expression {
```
codyebberson, line 624: Hm.  This mixes a lot of FHIR business logic with SQL builder, and one of the goals of SQL builder was to keep it general purpose.  How hard would it have been to use the normal `SelectQuery` for this?

packages/server/src/fhir/sql.test.ts
```diff
@@ -145,6 +153,25 @@ describe('SqlBuilder', () => {
     expect(sql.toString()).toBe('SELECT "MyTable"."id" FROM "MyTable" WHERE LOWER("MyTable"."name") NOT LIKE $1');
   });
 
+  test.only('Select per reference', () => {
```
codyebberson, line 156: `.only`

packages/fhir-router/src/graphql/graphql.test.ts
```diff
@@ -487,7 +499,31 @@ describe('GraphQL', () => {
     expect(data.Encounter.subject.resource).toBeNull();
   });
 
-  test('Reverse lookup with _reference', async () => {
+  test.only('Reverse lookup with _reference', async () => {
```
codyebberson, line 502: `.only`

packages/server/src/fhir/search.ts
```diff
@@ -113,6 +114,57 @@ export async function searchImpl<T extends Resource>(
...
+  type Row = { id: string; content: string; reference: string };
+  const rows: Row[] = await builder.execute(repo.getDatabaseClient(DatabaseMode.READER));
+
+  for (const row of rows) {
+    results[row.reference] ??= [];
+    results[row.reference].push(JSON.parse(row.content));
+  }
+
+  // fill in with empty arrays for references that have no results
+  for (const ref of references) {
```
codyebberson, line 161: Super minor code golf, if you do this loop first, you could remove line 156.

packages/server/src/fhir/search.ts
```diff
@@ -113,6 +114,57 @@ export async function searchImpl<T extends Resource>(
...
+  if ((searchRequest.offset ?? 0) > 0) {
+    throw new Error('Cannot use offset with searchByReference');
+  }
+
+  type Row = { id: string; content: string; reference: string };
+  const rows: Row[] = await builder.execute(repo.getDatabaseClient(DatabaseMode.READER));
+
+  for (const row of rows) {
+    results[row.reference] ??= [];
+    results[row.reference].push(JSON.parse(row.content));
```
codyebberson, line 157: Need to go through all of the standard post-processing (removing hidden fields, etc).

packages/server/src/fhir/search.ts
```diff
@@ -113,6 +114,57 @@ export async function searchImpl<T extends Resource>(
...
+  const builder = new SelectPerReferenceQuery(
+    searchRequest.resourceType,
+    new Column(resourceType, referenceField),
+    references
+  );
+  builder
+    .column({ tableName: resourceType, columnName: 'id' })
+    .column({ tableName: resourceType, columnName: 'content' });
+  repo.addDeletedFilter(builder as unknown as SelectQuery);
+  repo.addSecurityFilters(builder as unknown as SelectQuery, resourceType);
```
codyebberson, line 142: Given that the `SelectPerReferenceQuery` actually has several `SelectQuery` internally, we'd need to make sure that we're propagating all of these security filters to all of the inner queries.

I haven't parsed that fully, maybe you're already handling those?

```ts
    this.referenceQuery = new SelectQuery(this.tableName);
    this.resultsInnerQuery = new SelectQuery(this.tableName);
    this.resultsQuery = new SelectQuery('ranked', this.resultsInnerQuery);
    this.outerQuery = new SelectQuery('references', this.referenceQuery);
```

codyebberson (commented):
Big picture:
1. Given how much load we're getting from GraphQL fan-out queries, this has a ton of potential for protecting database CPU, so definitely excited to explore this.
2. We had to break a bunch of abstractions to get there, which increases the surface area.  We would need to add a ton of tests to make sure that we're covering all of the necessary access policy permission cases (security filters on all of the nested resource types, apply hidden fields to all resources).

packages/fhir-router/src/repo.ts
mattlong: I'm fairly certain this was actually working as-is in `MemoryRepository` since it already support matching on `Operator.IN`. I don't quite remember why I left this function not functional...probably because it was not on the critical path of testing out the SQL-based functionality which is all I was focused on at the time.

packages/server/src/fhir/sql.ts
mattlong: `SelectPerReferenceQuery` should be moved from `sql.ts` into `packages/server/src/fhir/search.ts` as an implementation detail since it is indeed not at all generic.

mattlong: Now that I have a better grasp on `SelectQuery` and friends, it probably wouldn't be too bad to rewrite `SelectPerReferenceQuery` as just a function if your concern is around structuring it as a class.

packages/server/src/fhir/search.ts
mattlong: The security filters are propagated correctly. To do this the "right" way, it would be ideal to have a targeted interface that both `SelectQuery` and `SelectPerReferenceQuery` implements. Currently, I believe the interface would just need to specify the `where` method.

packages/server/src/fhir/sql.ts
```diff
@@ -613,6 +621,116 @@ export class SelectQuery extends BaseQuery implements Expression {
...
+    this.outerQuery.column(new Column('results', getColumnName(column)));
+    this.resultsQuery.column(getColumnName(column));
+    this.resultsInnerQuery.column(column);
+    return this;
+  }
+
+  where(column: Column | string, operator?: keyof typeof Operator, value?: any, type?: string): this {
+    this.referenceQuery.where(column, operator, value, type);
+    this.resultsInnerQuery.where(column, operator, value, type);
+    return this;
```
mattlong, line 670: @codyebberson Per your comments on security filters, they are propagated into the inner queries here.

codyebberson:
Taking notes for myself to wrap my head around this.  Looking at this existing test:

```ts
  test('Reverse lookup with _reference', async () => {
      .post('/fhir/R4/$graphql')
      .set('Authorization', 'Bearer ' + accessToken)
      .set('Content-Type', ContentType.JSON)
      .send({
        query: `
      {
        PatientList(_count: 1) {
          id
          ObservationList(_reference: subject) {
            id
            status
            code {
              text
            }
          }
        }
      }
    `,
      });
    expect(res.status).toBe(200);
    expect(res.body.data.PatientList).toBeDefined();
    expect(res.body.data.PatientList[0].ObservationList).toBeDefined();
  });
```

This searches for `Patient`, and then for each match searches for `Observation`.

Focusing on the `Observation` searches:

Before SQL, run once per `Patient`:

```sql
SELECT
  "Observation"."id",
  "Observation"."content"
FROM
  "Observation"
WHERE
  (
    "Observation"."deleted" = $1
    AND (
      "Observation"."compartments" IS NOT NULL
      AND "Observation"."compartments" & & ARRAY [$2,$3] :: UUID []
    )
    AND "Observation"."subject" = $4
  )
LIMIT
  21;
```

After SQL, run once:

```sql
SELECT
  "results"."id",
  "results"."content",
  "references"."reference"
FROM
  (
    SELECT
      DISTINCT "subject" AS "reference"
    FROM
      "Observation"
    WHERE
      (
        "Observation"."subject" IN ($1)
        AND "Observation"."deleted" = $2
        AND (
          "Observation"."compartments" IS NOT NULL
          AND "Observation"."compartments" & & ARRAY [$3,$4]::UUID []
        )
      )
  ) AS "references"
  JOIN LATERAL (
    SELECT
      "ranked"."id",
      "ranked"."content"
    FROM
      (
        SELECT
          "Observation"."id",
          "Observation"."content",
          ROW_NUMBER() OVER () AS rn
        FROM
          "Observation"
        WHERE
          (
            "Observation"."subject" = "references"."reference"
            AND "Observation"."deleted" = $5
            AND (
              "Observation"."compartments" IS NOT NULL
              AND "Observation"."compartments" & & ARRAY [$6,$7]::UUID []
            )
          )
      ) AS "ranked"
    WHERE
      "ranked"."rn" <= $8
  ) AS "results" ON true;
```

I'm still digesting all of the complexities of `"references"`, `"ranked"`, `ROW_NUMBER()`, and `JOIN LATERAL`.

Apologies if this is an obvious/dumb question, but could we just use the existing search code except pass in an array of values for `subject`?

* You could continue using your new `parseSearchArgsWithReference`, `searchDataLoaders`, `buildResolveBySearchDataLoader`, etc - that all looks good
* Then call the normal `repo.search()` with multiple subjects
* In any existing search, there is already guaranteed to be only one entry per resource, so you don't need to worry about the `DISTINCT` on subject
* You get the results, separate them out by subject, and then add them to output
* Hard parts:
   * How does `LIMIT` work there? 🤔 😱   Is it possible to add a logical `LIMIT` per subject?

My overall feeling is that this feels like a great optimization, but i'm pretty nervous about the complexity of an entirely new search code path, and all of the long tail edge cases of access controls and whatnot.  If we can find a way to leverage the existing search code, I would feel much better about it.

codyebberson:
Would this work here?

image

The inner query could be the exact same as our existing code + the `row_number()` term.  Then we could use all of the existing search code.

codyebberson:
The `LATERAL JOIN` should work with a really minimal first/outer query:

```sql
SELECT sub.id, sub.subject, sub.value
FROM (VALUES ($1), ($2), ($3)) AS s(subject)
CROSS JOIN LATERAL (
    SELECT id, subject, value
    FROM Observation
    WHERE subject = s.subject
    ORDER BY id  -- or any other column you might want to order by
    LIMIT N       -- Replace N with the number of rows you want per subject
) AS sub;
```

mattlong:
Yes, after some offline discussion, the inner SELECT query shown below can be generated using the same code path as `repo.search` --> `searchImpl` --> `getSearchEntries` on everything in the `SearchRequest` except the reference filter. Then, the rest of the reference filter can be injected and and the rest of the lateral join query shell can built around it.

Based on our understanding of `INNER JOIN LATERAL`, this will effectively run the inner select query once per value in the style of a for loop as confirmed by the query plan below.

```sql
medplum=# EXPLAIN ANALYZE SELECT
        "results"."id",
        "results"."content",
        "references"."ref"
FROM (
        VALUES('Patient/ec22ea52-2b61-4abc-a86b-3c755cfe6457'),
                ('Patient/262bda3e-eeeb-4db6-8afe-61b4dc1f95bb'),
                ('Patient/c738afee-74b2-45bf-8a30-d356173315aa'),
                ('Patient/fake')) AS "references" ("ref")
         INNER JOIN LATERAL (
                SELECT
                        "Observation"."id",
                        "Observation"."content"
                FROM
                        "Observation"
                WHERE
                        "Observation"."subject" = "references"."ref"
                        AND "Observation"."deleted" = FALSE
                ORDER BY "Observation"."lastUpdated"
                LIMIT 20) AS "results" ON TRUE;
                                                                     QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=31.75..127.67 rows=28 width=711) (actual time=0.135..0.433 rows=41 loops=1)
   ->  Values Scan on "*VALUES*"  (cost=0.00..0.05 rows=4 width=32) (actual time=0.003..0.006 rows=4 loops=1)
   ->  Limit  (cost=31.75..31.76 rows=7 width=687) (actual time=0.101..0.102 rows=10 loops=4)
         ->  Sort  (cost=31.75..31.76 rows=7 width=687) (actual time=0.100..0.101 rows=10 loops=4)
               Sort Key: "Observation"."lastUpdated"
               Sort Method: quicksort  Memory: 25kB
               ->  Bitmap Heap Scan on "Observation"  (cost=4.47..31.65 rows=7 width=687) (actual time=0.033..0.077 rows=11 loops=4)
                     Recheck Cond: (subject = "*VALUES*".column1)
                     Filter: (NOT deleted)
                     Heap Blocks: exact=44
                     ->  Bitmap Index Scan on "Observation_subject_idx"  (cost=0.00..4.46 rows=7 width=0) (actual time=0.025..0.025 rows=11 loops=4)
                           Index Cond: (subject = "*VALUES*".column1)
 Planning Time: 2.183 ms
 Execution Time: 0.522 ms
```

packages/server/src/fhir/search.ts
```diff
@@ -344,8 +442,11 @@ async function getSearchIncludeEntries(
             value: canonicalReferences.join(','),
           },
         ],
-      })
-    );
+        count: DEFAULT_MAX_SEARCH_COUNT,
```
mattlong, line 445: Before this PR, since this search didn't specify count, `SelectQuery.limit_` had a value of `NaN` which resulted in no LIMIT clause being included. Perhaps that was the intention here, but it certainly feels like relying on `NaN` to achieve that is not ideal.

packages/server/src/fhir/search.ts
```diff
@@ -389,18 +482,20 @@ async function getSearchRevIncludeEntries(
...
+  const searchRequest = {
+    resourceType: revInclude.resourceType as ResourceType,
+    filters: [
+      {
+        code: revInclude.searchParam,
+        operator: Operator.EQUALS,
+        value: value,
+      },
+    ],
+    count: DEFAULT_MAX_SEARCH_COUNT,
```
mattlong, line 494: Same as above; previous this search didn't have a count which caused the resulting select query to not have a limit specified anywhere.

codyebberson (commented):
This is awesome.  Thanks for doing this.  I know it's been a few revs to get here, and I appreciate your dilligence on it 🙏 

1. Can we please add one test in `packages/server/src/fhir/operations/graphql.test.ts` that exercises this + project boundaries or access policies?  either one, just something that ensures the necessary security filters are added.  It looks like that is true in this PR, but I worry about things getting lost in future refactors.
2. We should deploy to staging and schedule a mini bug bash.

This is great, super excited.

codyebberson (approved):
lgtm 👍 

Thanks for all the iterations on this, feels like this is in a good place.

packages/server/src/fhir/search.ts
```diff
@@ -114,6 +110,75 @@ export async function searchImpl<T extends Resource>(
...
+    content: string;
+    ref: string;
+  }[] = await builder.execute(repo.getDatabaseClient(DatabaseMode.READER));
+
+  const results: Record<string, T[]> = {};
+  for (const ref of referenceValues) {
+    results[ref] = [];
+  }
+  for (const row of rows) {
+    const resource = JSON.parse(row.content) as T;
```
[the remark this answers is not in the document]
mattlong: I don't think there's much that can be done about it; this is one of the main serial bits that the batch size setting counteracts.

packages/fhir-router/src/repo.test.ts
```diff
@@ -139,4 +141,102 @@ describe('MemoryRepository', () => {
...
+            resourceType: 'Observation',
+            subject: createReference(patient),
+            valueString: i.toString(),
+          } as Observation)
+        );
+      }
+      return resources;
+    }
+
+    function zip<A, B>(a: A[], b: B[]): [A, B][] {
```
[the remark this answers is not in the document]
mattlong: 🐍

codyebberson (approved):

PR #4812, opened 2024-07-07
fixed the medplum aws update-app issue

### LINKED ISSUE #4655: `medplum aws update-app` hard coded to use public npm registry
## Background
In our environment, access to the public Internet is very restrictive, meaning access to the public npm registry is not allowed. Instead, we have a [Nexus](https://www.sonatype.com/products/sonatype-nexus-repository) artifact server that serves as the npm registry. This is a requirement of our security posture that we cannot easily change.

## Issue
It appears as though the public npm registry is hard coded in the file below which causes the `medplum aws update-app` to fail. Can this be changed to be configurable, perhaps by an environment variable or a CLI option?

https://github.com/medplum/medplum/blob/fe7eb441b35f67a6b691ff98218832f2edd59ab7/packages/cli/src/aws/update-app.ts#L69-L73

RajCherukuru:
Fix #4655

Changes:

Added a tarpath variable for user to give the path.
Updated the was index.ts for the updateAppCommand

### DIFF
--- a/packages/cli/src/aws/index.ts
+++ b/packages/cli/src/aws/index.ts
@@ -61,6 +61,7 @@ export function buildAwsCommand(): Command {
       '--dryrun',
       'Displays the operations that would be performed using the specified command without actually running them.'
     )
+    .option('--tar-path [tarPath]', 'Specifies the path to the extracted tarball of the @medplum/app package.') 
     .action(updateAppCommand);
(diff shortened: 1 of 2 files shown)


codyebberson (approved):
lgtm 👍 

Thanks @RajCherukuru !

Issue #4815, opened 2024-07-08
[Feature Request] Pre-signed URLs for google cloud storage


ivank:
We want to use google cloud storage with medplum. It works out quite alright, because [google storage supports s3 API](https://cloud.google.com/storage/docs/interoperability) out of the box. 

There is a problem however. Pre signed URLs do not use the default S3 functionality but rather rely on cloudfront's signer.

https://github.com/medplum/medplum/blob/91418730a7acf7847131d350c155ee005db617d9/packages/server/src/cloud/aws/storage.ts#L2

I understand the reasons for it is to simplify security concerns and place all of them on cloudfront, but if you were to use the native s3 functionality, you'll get the ability to host on GCP. 

From the looks of it, it should be pretty easy to add an option something like:
```ts
{
  signing: "cloudfront" | "native"
}
```

Or some other bool flag, and then switch between `@aws-sdk/s3-request-presigner` and `@aws-sdk/cloudfront-signer`. AWS credentials can be left to be provided via env vars. That is what we currently do in order to upload images to google cloud. This way you would unlock "cloudfront-less" deployment as well as "GCP". 

I'm more than happy to attempt to write a PR for you guys, but wanted to check if there's been a consideration regarding this, otherwise we'll need to fork and implement it on our end, as we need to deploy on GCP.

PR #4817, opened 2024-07-08
Pre-signed urls using native s3 presigner

ivank:
This is implements medplum/medplum#4815

Instead of a flag, we can make `signingKey` and `signingKeyId` optional, thus relying on AWS config directly.

- Since the original presigner doesn't have test, I didn't add any, but could try to figure out a way how to test it properly.
- Docs not yet updated

I can address those points but wanted to get a sense if the general direction is acceptable to the team.

---

I tested it locally and it the code does construct URLs for Binary that are signed correctly, (can be opened in incognito mode)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2105,6 +2105,20 @@
         "node": ">=16.0.0"
       }
     },
+    "node_modules/@aws-sdk/util-format-url": {
+      "version": "3.598.0",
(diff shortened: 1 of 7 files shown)


ivank:
Just wanted to report that we're actively using this particular change in our setup (by maintaining a fork with this PR auto applied to recent releases) and it does work as expected.

PR #4819, opened 2024-07-08
[WIP] Photon Integration Auth and Sync Patient

ksmith94:
Create a new example directory to demo how to integrate Medplum with Photon Health's prescription API. This creates a patient prescription page with a button to sync patients to photon health. Authorization is handled so the API can be accessed.


ksmith94:
Currently, the button does not sync patients, but just makes a call to the Photon API and successfully auths. It just queries for all patients in the org and logs them to the console.

PR #4829, opened 2024-07-09
Precompute ValueSet expansions

mattwiller:


packages/server/src/fhir/operations/expand.ts
```diff
@@ -266,32 +277,70 @@ async function computeExpansion(
...
       await includeInExpansion(include, expansion, codeSystem, params);
     }
-
-    if (expansion.length > (count ?? MAX_EXPANSION_SIZE)) {
-      // Return partial expansion
-      break;
-    }
   }
 
+  await addExpandJob(valueSet, codeSystemCache); // Precompute expansion in the background for faster future searches
```
codyebberson, line 282: Will this create a BullMQ job for every expansion, even if has been expanded before?

mattwiller: This should only run after we needed to generate the expansion: once there are precomputed expansion rows in the DB, it won't queue the expansion job anymore

packages/server/src/fhir/operations/expand.ts
```diff
@@ -266,32 +277,70 @@ async function computeExpansion(
...
   return expansion;
 }
 
+async function isExpansionPrecomputed(valueSet: ValueSet): Promise<boolean> {
+  const result = await new SelectQuery('ValueSet_Membership')
```
codyebberson, line 287: 👍 

As discussed offline, this won't handle a couple of cases:
1. The "zero expanded values" case - where we would end up repeatedly trying the expansion
    * Probably fine, because (1) that's a rare event and (2) those are probably fast operations
    * But still an edge case that we may want to better handle in the future
2. The "expansion in progress" case - where partial results are in the table
    * Probably fine because (1) only applies during small time window and (2) partial results are not catastrophic

We can perhaps file an issue for a future `ValueSet_State` table that tracks all of this explicitly, or whatever other clever solutions you have in mind.

packages/server/src/fhir/operations/expand.ts
```diff
@@ -266,32 +277,70 @@ async function computeExpansion(
...
+  valueSet: ValueSet,
+  params: ValueSetExpandParameters
+): Promise<ValueSetExpansionContains[]> {
+  const query = new SelectQuery('Coding').column('code').column('display');
+  const joinAlias = query.getNextJoinAlias();
+  query
+    .innerJoin(
+      'ValueSet_Membership',
+      joinAlias,
+      new Condition(new Column('Coding', 'id'), '=', new Column(joinAlias, 'coding'))
```
codyebberson, line 305: I'd want to measure it, but including `valueSet` in the `ON` clause might be able to take advantage of the index better?  Or maybe postgres is smart enough to figure that out so they're equivalent?

packages/server/src/workers/expand.ts
```diff
@@ -0,0 +1,156 @@
...
+            .column(new Column('expansion', 'id'));
+          const writeQuery = new InsertQuery('ValueSet_Membership', rowQuery).ignoreOnConflict();
+          await writeQuery.execute(getDatabasePool(DatabaseMode.WRITER));
+        }
+      }
+    }
+
+    ctx.logger.info('ValueSet expanded successfully', { id: valueSet.id });
+  } catch (ex: any) {
+    ctx.logger.info('ValueSet expand job exception', ex);
```
codyebberson, line 153: `logger.error`?

packages/server/src/workers/expand.ts
```diff
@@ -0,0 +1,156 @@
...
+let queue: Queue<ExpandJobData> | undefined = undefined;
+let worker: Worker<ExpandJobData> | undefined = undefined;
+
+/**
+ * Initializes the ValueSet expand worker.
+ * Sets up the BullMQ job queue.
+ * Sets up the BullMQ worker.
+ * @param config - The Medplum server config to use.
+ */
+export function initExpandWorker(config: MedplumServerConfig): void {
```
codyebberson, line 35: Thinking about this more... What is the motivation for using a BullMQ job for this?

Do we expect it to fail and retry?

packages/server/src/workers/expand.ts
```diff
@@ -0,0 +1,156 @@
...
+        const mappings = mapFilter(validCodings, (c) => (c?.id ? { valueSet: valueSet.id, coding: c.id } : undefined));
+        await new InsertQuery('ValueSet_Membership', mappings)
+          .ignoreOnConflict()
+          .execute(getDatabasePool(DatabaseMode.WRITER));
+      } else {
+        const query = expansionQuery(include, codeSystem);
+        if (query) {
+          // Construct outer INSERT query
+          const rowQuery = new SelectQuery('expansion', query)
+            .column(new Literal(valueSet.id ?? ''))
```
codyebberson, line 143: If `id` is empty or undefined, we should probably throw rather than default to `''`, yeah?

packages/server/src/fhir/sql.ts
```diff
@@ -552,7 +556,11 @@ export class SelectQuery extends BaseQuery implements Expression {
       if (!first) {
         sql.append(', ');
       }
-      sql.appendColumn(column);
+      if (column instanceof Literal) {
+        sql.appendParameters(column.value, false);
```
codyebberson, line 560: Nice, ty for using parameters 🙏

packages/server/src/fhir/operations/expand.ts
```diff
@@ -266,32 +287,72 @@ async function computeExpansion(
...
       await includeInExpansion(include, expansion, codeSystem, params);
     }
-
-    if (expansion.length > (count ?? MAX_EXPANSION_SIZE)) {
-      // Return partial expansion
-      break;
-    }
   }
 
+  if (params._precompute && requireSuperAdmin()) {
```
codyebberson, line 292: 👍

codyebberson (approved):

PR #4834, opened 2024-07-10
fixed the issue of Create project, no alert when create is wrong

### LINKED ISSUE #3583: Create project, no alert when create is wrong
I have identified a user interface issue during the project creation phase. Currently, when a request encounters an error, no information is displayed to the user. While the error is explicitly present in the request, presenting it in a simplified manner could be beneficial to enhance user understanding.

image
image
image

I will start to work on it

RajCherukuru:
Fixes #3583

Before:
Screenshot 2024-07-10 at 12 36 32 AM

After:
Screenshot 2024-07-10 at 12 33 40 AM


codyebberson (approved):
Eek!  Good catch @RajCherukuru !

Issue #4825, opened 2024-07-09
[Photon Health] Set up UX


ksmith94:
Set up a simple app with non-functional prescription UX on the patient page.

PR #4836, opened 2024-07-10
[Photon Health] Set up prescription UX

ksmith94:
This sets up a simple one-page app and adds a prescription section to a patient's page. It uses [Photon Health's elements](https://docs.photon.health/docs/elements) but currently does not implement authentication so they are not usable. This is just to get the elements installed and lay out the UX of the prescription page. Related to #4825


rahul1 (approved):

rahul1:
@SocketSecurity ignore

PR #4846, opened 2024-07-11
Add covering compound index to token lookup tables

mattwiller:


packages/generator/src/migrate.ts
```diff
@@ -226,9 +227,7 @@ function buildCreateTables(result: SchemaDefinition, resourceType: string, fhirT
     ],
     indexes: [
       { columns: ['resourceId'], indexType: 'btree' },
-      { columns: ['code'], indexType: 'btree' },
-      { columns: ['system'], indexType: 'btree' },
-      { columns: ['value'], indexType: 'btree' },
+      { columns: ['code', 'value', 'system'], indexType: 'btree', includeColumns: ['resourceId'] },
```
codyebberson, line 230: Would love to get perf numbers on the `INCLUDE` in particular.

codyebberson (commented):
Code looks good to me 👍  Let's get some benchmarks before pushing though.

Thank you! 🙏

PR #4851, opened 2024-07-11
[Patient Intake Sample] Creates all expected resources based on intake form response

filipeximenes:
# Description
Still missing:
- Fix consent codes
- Add bot setup scripts


examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +210,123 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
+      return currentItem;
+    } else if (currentItem.item) {
+      return findQuestionnaireItem(currentItem.item, linkId);
+    }
+
+    return undefined;
+  }, undefined);
+}
+
+export function getGroupRepeatedAnswers(
```
filipeximenes, line 290: This turned out to be a handy helper to deal with the values of multiple coverage values, perhaps there should be something similar in the core lib.

examples/medplum-patient-intake-demo/src/bots/core/intake-form.ts
```diff
@@ -75,5 +83,52 @@ export async function handler(event: BotEvent<QuestionnaireResponse>, medplum: M
     answers['education-level'].valueCoding
   );
 
+  const insuranceProviders = getGroupRepeatedAnswers(questionnaire, response, 'coverage-information');
+
+  for (const provider of insuranceProviders) {
+    await addCoverage(medplum, patient, provider);
+  }
+
+  // Handle consents
```
filipeximenes, line 92: There are a few `FIXME` comments in this section that we will need to review

filipeximenes (commented):
@rahul1 this ended a bit larger than ideal but many of the changes are in fixture files.

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +210,123 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
 }
+
+export async function addCoverage(
+  medplum: MedplumClient,
+  patient: Patient,
+  answers: Record<string, QuestionnaireResponseItemAnswer>
+): Promise<void> {
+  const payor = await medplum.createResource({
+    resourceType: 'Organization',
+    name: answers['insurance-provider']?.valueString,
```
rahul1, line 221: We should update the questionnaire and response to use `valueReference`. I will update your project with our PayerDirectory, which has a link to all kinds of existing `Oganization` resources fo rinsurance companies. You should not have to create this one

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +210,123 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
+  consetGiven: boolean,
+  scope: CodeableConcept,
+  category: CodeableConcept,
+  policyRule: CodeableConcept,
+  date: Consent['dateTime']
+): Promise<void> {
+  await medplum.createResource({
+    resourceType: 'Consent',
+    patient: createReference(patient),
+    status: consetGiven ? 'active' : 'rejected',
```
rahul1, line 257:
```suggestion
    status: consentGiven ? 'active' : 'rejected',
```

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
rahul1: Yeah I think that's a great suggestion. Might need a few tweaks, since non-group item can _also_ have sub-answers. Was a get `getAllQuestionnaireAnswers` useful in this context?

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes: Thanks!

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes: The problem with `getAllQuestionnaireAnswers` is that it returns a list of all answers as a list to a given key. 
```
{
    linkId1: [answer11, answer12, answer13],
    linkId2: [answer21, answer22, answer23],
}
```
In the scenario where all answers are required (which I was not not my initial expectation) that's fine, it's easy to match answers in groups but if any of the fields is not required than using the order of items to match groups becomes unreliable.

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes, line 1: @rahul1 I added comments to the code here

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
rahul1: Yes I see, that makes sense

examples/medplum-patient-intake-demo/data/core/patient-intake-questionnaire.json
```diff
@@ -72,18 +75,21 @@
               {
                 "linkId": "insurance-provider",
                 "text": "Insurance Provider",
-                "type": "string"
+                "type": "string",
```
rahul1, line 78: As discussed onlinem wi will change this to reference type

examples/medplum-patient-intake-demo/data/core/patient-intake-questionnaire.json
```diff
@@ -186,40 +192,9 @@
...
-                  {
-                    "question": "acknowledgement-for-advance-directives-choice",
-                    "operator": "=",
-                    "answerCoding": {
-                      "system": "http://terminology.hl7.org/CodeSystem/v2-0532",
-                      "code": "Y"
-                    }
-                  }
-                ]
+                "linkId": "acknowledgement-for-advance-directives-help",
```
rahul1, line 195: Why this change?

examples/medplum-patient-intake-demo/src/pages/IntakeResponsePage.tsx
```diff
@@ -79,7 +79,11 @@ export function IntakeResponsePage(): JSX.Element {
 
   const questionnaireResponse = medplum.readResource('QuestionnaireResponse', responseId).read();
 
-  const omittedItems = ['agreement-to-pay-for-treatment-help', 'notice-of-privacy-practices-help'];
+  const omittedItems = [
```
rahul1, line 82: Why are these items omitted?

examples/medplum-patient-intake-demo/src/bots/core/test-data/intake-form-test-data.ts
```diff
@@ -12,9 +12,208 @@ export const intakePatient: Patient = {
   ],
 };
 
+export const intakeQuestionnaire: Questionnaire = {
+  resourceType: 'Questionnaire',
+  id: '58b7094f-4b04-46d3-9753-0272b1681a0c',
```
rahul1, line 17: remove id

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +233,159 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
+type FindQuestionnaireItemType = QuestionnaireItem | QuestionnaireResponseItem | undefined;
+
+/**
+ * Finds a QuestionnaireItem or QuestionnaireResponseItem with the given linkId
+ *
+ * @param items - The array of objects present in the `.item` attribute of a Questionnaire or QuestionnaireResponse
+ * @param linkId - The id to be found
+ * @returns - The found item or undefined in case it's not found
+ */
+export function findQuestionnaireItem(
```
rahul1, line 305: Use generics here: 
```ts
findQuestionnaireItem<T extends QuestionnaireItem | QuestionnaireResponseItem>
```

examples/medplum-patient-intake-demo/src/bots/core/test-data/intake-form-test-data.ts
```diff
@@ -12,9 +12,208 @@ export const intakePatient: Patient = {
   ],
 };
 
+export const intakeQuestionnaire: Questionnaire = {
```
rahul1, line 15: You can actually import the `data/core/patientinakequestionnair` into this test file, to avoid having a to keep them in sync

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes: Yes, that's better, thanks

examples/medplum-patient-intake-demo/src/bots/core/test-data/intake-form-test-data.ts
filipeximenes: Cool, doing this

examples/medplum-patient-intake-demo/src/bots/core/test-data/intake-form-test-data.ts
filipeximenes: Ids are actually important here because they are being used to make references between test fixtures. Is there a better way to do it without ids?

examples/medplum-patient-intake-demo/data/core/patient-intake-questionnaire.json
pamella: We made those changes to simplify the modeling. As we couldn't find how to model those consent-related information, we'd need to dive deeper into that.

examples/medplum-patient-intake-demo/src/pages/IntakeResponsePage.tsx
pamella: To simplify the display of the form response. 

Here is the diff:

| with omit | without omit |
| --- | --- |
| [Screencast from 15-07-2024 11:07:45.webm](https://github.com/user-attachments/assets/f432f81a-825a-41f8-8611-edc40c9bb108) | [Screencast from 15-07-2024 11:08:29.webm](https://github.com/user-attachments/assets/1085c12e-1bd2-4dad-9f00-e2bf89899783) |

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +235,151 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
+      status: 'active',
+      beneficiary: createReference(patient),
+      subscriberId: subscriberId,
+      relationship: { coding: [answers['relationship-to-subscriber'].valueCoding as Coding] },
+      payor: [payor],
+    },
+    {
+      beneficiary: getReferenceString(patient),
+      payor: getReferenceString(payor),
+      // subscriberId: subscriberId,
```
filipeximenes, line 267: This is unexpected to me but including `subscriberId` in the query makes the test fail. Not sure if a problem with the test client or this is expected behavior

rahul1: What is the error message?

filipeximenes: Not an error, [the test just breaks](https://github.com/medplum/medplum/pull/4851/files#diff-fe5ca34e17f32d220c942739b3d9c6ecc4250bc22cc37d8db74dc94616723bf4R241), the upsert doesn't work and the object is created twice when it shouldn't.

examples/medplum-patient-intake-demo/src/pages/IntakeResponsePage.tsx
rahul1: Thank you!

examples/medplum-patient-intake-demo/src/bots/core/intake-form.ts
```diff
@@ -75,5 +84,53 @@ export async function handler(event: BotEvent<QuestionnaireResponse>, medplum: M
...
+
+  await addConsent(
+    medplum,
+    patient,
+    !!answers['consent-for-treatment-signature']?.valueBoolean,
+    consentScopeMapping.treatment,
+    consentCategoryMapping.med,
+    undefined,
```
filipeximenes, line 101: @rahul1 Not a blocker for this PR, while passing `undefined` to the policy rule works in our tests, I noticed that it breaks when running the bot in production `Error: Constraint ppc-1 not met: Either a Policy or PolicyRule ({"fhirpath":"policy.exists() or policyRule.exists()"}) (Consent)`.

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -42,47 +57,153 @@ export const observationCategoryMapping: Record<string, CodeableConcept> = {
...
+ * @param valueXAttribute - The value[x] field where the answer should be stored
+ * @param value - The value to be stored in the observation
+ */
 export async function upsertObservation(
   medplum: MedplumClient,
   patient: Patient,
   code: CodeableConcept,
   category: CodeableConcept,
-  valueCoding: QuestionnaireResponseItemAnswer | undefined
+  valueXAttribute: 'valueCodeableConcept' | 'valueDateTime',
```
rahul1, line 166: This parameter needs a better name and a better type

```suggestion
  answerType: QuestionnaireItemType,
```

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -42,47 +57,153 @@ export const observationCategoryMapping: Record<string, CodeableConcept> = {
...
+  value: QuestionnaireResponseItemAnswer | undefined
 ): Promise<void> {
   const coding = code.coding?.[0];
```
rahul1, line 169: There's no need to extract the `coding` from the code.

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
rahul1: Can you file a separate issue for this? Let's debug in a separate PR

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +271,158 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
+  groupLinkId: string
+): Record<string, QuestionnaireResponseItemAnswer>[] {
+  const questionnaireItem = findQuestionnaireItem(questionnaire.item, groupLinkId) as QuestionnaireItem;
+  const responseItem = findQuestionnaireItem(response.item, groupLinkId) as QuestionnaireResponseItem;
+
+  if (questionnaireItem.type !== 'group' || !questionnaireItem?.item) {
+    return [];
+  }
+
+  const responses = responseItem.item || [];
```
rahul1, line 380:
```suggestion
  const responses = responseItem.item ?? [];
```

Prefer `??` where possible

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -136,3 +271,158 @@ export function addLanguage(patient: Patient, valueCoding: Coding, preferred: bo
...
+ */
+export function getGroupRepeatedAnswers(
+  questionnaire: Questionnaire,
+  response: QuestionnaireResponse,
+  groupLinkId: string
+): Record<string, QuestionnaireResponseItemAnswer>[] {
+  const questionnaireItem = findQuestionnaireItem(questionnaire.item, groupLinkId) as QuestionnaireItem;
+  const responseItem = findQuestionnaireItem(response.item, groupLinkId) as QuestionnaireResponseItem;
+
+  if (questionnaireItem.type !== 'group' || !questionnaireItem?.item) {
```
rahul1, line 376:
```suggestion
  if (questionnaireItem.type !== 'group' ?? !questionnaireItem?.item) {
```

Prefer `??`

rahul1 (approved):
@filipeximenes , I've made a few suggestions, but you can merge after making those changes. 

Please also file an issue for the issue with duplicate resources for `subscriberID`

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes: @rahul1 I agree `??` is more adequate in `const responses = responseItem.item ?? [];` as it's a coalescing operator and semantically that's what we are doing in that expression. But in this case here I'd argue the "or" logical operator is more semantically correct. That said, I understand the end result is the same and I'm all in for following conventions over arguing over small details, LMK your final decision and I'll make the change here and in all the other places over this file.

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes: @rahul1 the main reason this is done is to be used in the query parameter of the `upsertResource` call. Is there a better way to extract the token that I'm no aware of?

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
filipeximenes: Here is the link to the issue: https://github.com/medplum/medplum/issues/4957

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
rahul1: I'm sorry, you're right. I was reviewing a bit too quickly 😅. The `||` operator was correct

examples/medplum-patient-intake-demo/src/bots/core/intake-form.ts
rahul1: Thanks for flagging @filipeximenes . I think the best course of action is to find the right policy rules. 

For this one, I think this is the right code `cric`: 
https://hl7.org/fhir/R4/codesystem-consent-policy.html#consent-policy-cric

examples/medplum-patient-intake-demo/src/bots/core/intake-form.ts
```diff
@@ -72,8 +83,75 @@ export async function handler(event: BotEvent<QuestionnaireResponse>, medplum: M
...
+    convertDateToDateTime(answers['notice-of-privacy-practices-date'].valueDate)
+  );
+
+  await addConsent(
+    medplum,
+    patient,
+    !!answers['acknowledgement-for-advance-directives-signature']?.valueBoolean,
+    consentScopeMapping.adr,
+    consentCategoryMapping.acd,
+    undefined,
```
rahul1, line 152: We're just going to have to make up a code for this. can we just make the system `http://medplum.com` and the code `BasicADR`

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
```diff
@@ -42,56 +57,166 @@ export const observationCategoryMapping: Record<string, CodeableConcept> = {
...
-      coding: [valueCoding],
-    },
+    ...(answerType === 'valueCodeableConcept'
+      ? {
+          valueCodeableConcept: {
+            coding: [value],
+          },
+        }
+      : {}),
+    ...(answerType === 'valueDateTime' ? { valueDateTime: value.valueDateTime } : {}),
```
rahul1, line 190: For readability, you should use an if statement _after_ this initial construction, rather than using the big ternary operator

examples/medplum-patient-intake-demo/src/bots/core/intake-utils.ts
rahul1: This is the right way to do it. Maybe just move this closer to line 192, since it is where it is now used

rahul1 (approved):

PR #4853, opened 2024-07-11
refactor(core/subscriptions): use `ReconnectingWebSocket` for `SubscriptionManager`

### LINKED ISSUE #4586: Auto-reconnecting in client when `WebSocket` not intentionally closed
WebSockets are often flaky under circumstances where connectivity is spotty or inconsistent. We can make working with `WebSocket Subscriptions` and `WebSockets` in general easier for application authors using `MedplumClient` by auto-reconnecting behind the scenes and in the case of `WebSocket Subscriptions` recreating subscriptions with the same criteria automatically if the connection reconnects

ThatOneBro:
Closes #4586


packages/server/src/subscriptions/websockets.ts
```diff
@@ -151,6 +151,8 @@ export async function handleR4SubscriptionConnection(socket: ws.WebSocket): Prom
     }
     subscribeWsToSubscription(socket, subscriptionId);
     ensureHeartbeatHandler();
+    // Send a handshake to notify client that this subscription is active for this connection
+    socket.send(JSON.stringify(createHandshakeBundle(subscriptionId)));
```
codyebberson, line 155: Does this require that clients and server are on same version?

In other words, will this break clients on older versions?

codyebberson (approved):
Nice.  This looks good to me 👍  Looks like a lot of yak shaving, so thank you for persevering through that 🙏 

See one comment on backwards compatibility with clients on older version.

Thoughts on real world testing?  Should we do a bug bash on staging?

packages/server/src/subscriptions/websockets.ts
ThatOneBro: Technically it will still work but it won't be filtered out, so users would get it as a bundle in their callback and would have to filter it out

This could be seen as a breaking change in that regard, though this is actually compliant with the FHIR spec and we should probably do it for more consistent behavior with disconnecting / reconnecting logic 🤔 
https://build.fhir.org/ig/HL7/fhir-subscription-backport-ig/channels.html#:~:text=single%20websocket%20connection.-,Server%20sends%20one%20or%20more%20handshake%20messages%20via%20websockets%20(one%20per,last%20%E2%80%98n%E2%80%99%20messages%2C%20etc.).%20Clients%20are%20expected%20to%20handle%20either%20flow.,-Server%20may%20send

We might need to check with some of the clients who are already using it in prod?

ThatOneBro (commented):
Yeah I think we can do a bug bash on staging. Should we schedule it for Tuesday or what are you thinking?

codyebberson (approved):
boom 💥 👍 🚀

codyebberson:
8kb size increase to core.  Not the end of the world.

main:

```
$ ls -laR packages/core/dist/esm/
packages/core/dist/esm/:
total 1373
drwxr-xr-x 1 cody cody      0 Aug  7 09:10 .
drwxr-xr-x 1 cody cody      0 Aug  7 09:10 ..
-rw-r--r-- 1 cody cody 224491 Aug  7 09:10 index.d.ts
-rw-r--r-- 1 cody cody 184718 Aug  7 09:10 index.mjs
-rw-r--r-- 1 cody cody 984875 Aug  7 09:10 index.mjs.map
-rw-r--r-- 1 cody cody     18 Aug  7 09:10 package.json
```

after:

```
$ ls -laR packages/core/dist/esm/
packages/core/dist/esm/:
total 1409
drwxr-xr-x 1 cody cody       0 Aug  7 09:13 .
drwxr-xr-x 1 cody cody       0 Aug  7 09:13 ..
-rw-r--r-- 1 cody cody  229710 Aug  7 09:13 index.d.ts
-rw-r--r-- 1 cody cody  192021 Aug  7 09:13 index.mjs
-rw-r--r-- 1 cody cody 1009070 Aug  7 09:13 index.mjs.map
-rw-r--r-- 1 cody cody      18 Aug  7 09:13 package.json
```

codyebberson (approved):

codyebberson (approved):

PR #4856, opened 2024-07-12
Revert "Revert "fixed the matchesStringValue doesn't match exactly on token...""

codyebberson:
Reverts medplum/medplum#4855

This is a "revert of the revert".  The [original PR](https://github.com/medplum/medplum/commit/ba1cdfd95eebcb56d764954a846f0b06ae912fc0) did not handle `SearchParameter.type='token'` on `Coding` values.  That led to a production issue of `Subscription` webhooks not firing that depended on `token` and `Coding` (such as `Task?_tag=x`.  

This PR "fixes" the issue and adds tests.  If we want to merge, we can.

However, the `matchesStringValue` implementation overall is brittle, and probably needs an overhaul.  It relies on a series of heuristics and special cases rather than a concrete system of handling all possible combinations of `SearchParameter.type` and `ElementDefinition.type`.

### DIFF
--- a/packages/core/src/search/match.test.ts
+++ b/packages/core/src/search/match.test.ts
@@ -9,6 +9,7 @@ import {
   QuestionnaireResponse,
   SearchParameter,
   ServiceRequest,
+  Task,
 } from '@medplum/fhirtypes';
 import { indexSearchParameterBundle } from '../types';
 import { indexStructureDefinitionBundle } from '../typeschema/types';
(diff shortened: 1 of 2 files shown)


rahul1:
I would prefer doing this via the `evalFhirpathTyped` method, to give us a bit more consistency between how search params are handled in the in-memory vs. sql implementations. Given the issue we just had, I'm not sure the quick fix is worth it.

codyebberson:
Abandoning this, will replace with deeper fix.

PR #4858, opened 2024-07-12
Fix bug when selecting multiple values in AsyncAutocomplete

mattlong:
The `onChange` events emitted from `AsyncAutocomplete` were incorrect when multiple values are selected; only the most recently selected value wound up being included in the call to the `onChange` handler.

### DIFF
--- a/packages/mock/src/mocks/valueset.ts
+++ b/packages/mock/src/mocks/valueset.ts
@@ -8,8 +8,8 @@ export const exampleValueSet: ValueSet = {
     contains: [
       {
         system: 'x',
-        code: 'test-code-1',
-        display: 'Test Display 1',
+        code: 'test-code',
+        display: 'Test Display',
       },
       {
         system: 'x',
--- a/packages/mock/src/mocks/valueset.ts
+++ b/packages/mock/src/mocks/valueset.ts
@@ -8,8 +8,18 @@ export const exampleValueSet: ValueSet = {
     contains: [
       {
         system: 'x',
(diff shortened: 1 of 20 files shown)


packages/mock/src/mocks/valueset.ts
```diff
@@ -8,8 +8,18 @@ export const exampleValueSet: ValueSet = {
     contains: [
       {
         system: 'x',
-        code: 'test-code',
-        display: 'Test Display',
+        code: 'test-code-1',
```
codyebberson, line 11: I *think* this is fine, but there's minor risk that a random customer test is relying on this `code` or `display`.  If we wanted to be really safe, we'd leave these as-is, and add the new codes.

I'm probably fine with how you've done it though.

codyebberson (approved):

packages/mock/src/mocks/valueset.ts
mattlong: Ah, good point. No reason for the churn. I'll switch this one back.

dillonstreator:
Thoughts on including some of the additional fixes in this other PR? https://github.com/medplum/medplum/pull/4393

codyebberson (approved):

codyebberson:
@mattlong - can you please take a look at @dillonstreator 's related PR? https://github.com/medplum/medplum/pull/4393

mattlong:
@dillonstreator Thanks so much and apologies for missing your PR. I integrated most of the changes from your PR into this one.

PR #4869, opened 2024-07-15
Fixes #4343 - ONC service base url

### LINKED ISSUE #4343: Medplum's capability statement not matching what's expected by ONC

codyebberson:
Replaces https://github.com/medplum/medplum/pull/4352

The old PR was stuck on auth refactoring.  Serving this content from the API server would require some pretty deep changes to how auth works for the FHIR base URL (i.e., `/fhir/R4`) because the ONC service base URL must be public.

I was reading through the requirements, and realized that the ONC service base URL does not need to be the same as the FHIR base URL.  So we can do the really simple thing, and just host it on www.medplum.com 🙃 

Basic FHIR validator: https://inferno.healthit.gov/validator/

Service Base URL test kit: https://inferno.healthit.gov/test-kits/service-base-url/

### DIFF
--- a/packages/docs/static/onc-service-base-url.json
+++ b/packages/docs/static/onc-service-base-url.json
@@ -1,0 +1,90 @@
+{
+  "resourceType": "Bundle",
+  "type": "searchset",
+  "entry": [
+    {
+      "resource": {
+        "resourceType": "Endpoint",
+        "id": "medplum",
+        "status": "active",
+        "connectionType": {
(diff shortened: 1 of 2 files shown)


ThatOneBro (approved):

ThatOneBro:
Nice, glad we don't have to hack around with any of that mess.

PR #4887, opened 2024-07-15
Fixes #4882 - inclue email claim in ID token JWT

### LINKED ISSUE #4882: Return email claim in JWT
User Driven: https://medplum.slack.com/archives/C057GEVQCRE/p1721064330820739

Currently, our JWT doesn't return the `email` claim, even when the `email` scope is provided

codyebberson:


packages/server/src/oauth/utils.ts
```diff
@@ -501,15 +501,12 @@ export async function setLoginScope(login: Login, scope: string): Promise<Login>
 }
 
 export async function getAuthTokens(
+  user: User | ClientApplication,
   login: Login,
   profile: Reference<ProfileResource>,
   refreshLifetime?: string
 ): Promise<TokenResult> {
   const clientId = login.client && resolveId(login.client);
-  const userId = resolveId(login.user);
```
ThatOneBro, line 509: Should we assert that the `user` and `login.user` match here before we continue?

codyebberson: `getAuthTokens` is only called in 2 places, and that is guaranteed to be true.

I suppose we could do that for defensive futureproofing, but I consider that low pri in this case.

codyebberson: There are 2 call sites:
1. `register.ts` - on registering a new user, relatively rare event
2. `token.ts` - when sending a token response, relatively common event

In `token.ts`, we immediately read `user` and `profile` right before this.  If we're ok with double reading in the `register.ts` case, we could move that into `getAuthTokens`, then you don't need to worry about the assumptions.

codyebberson: Hm, k, I tried that, and things get messy, because you need the `user` and `membership` outside of `getAuthTokens` as well, so that would be a bigger refactor.

Bottom line, I guess my preference would be to leave as-is.

codyebberson: Will do

ThatOneBro (approved):

rahul1 (approved):

PR #4902, opened 2024-07-17
[Chart demo] Fixes bots handler function signature

filipeximenes:
# Description
While working at he Patient Intake demo bots I noticed that the handler parameters from the Charts demo were reversed, `handler(event, medplum)` instead of `handler(medplum, event)`, this PR fixes this issue.


rahul1 (approved):
Thank you!

PR #4918, opened 2024-07-18
Resend subscriptions UI

codyebberson:
image

image

image

Also added a "Patch" button which will use FHIR `PATCH` rather than `PUT`:

image

### DIFF
--- a/packages/app/src/resource/ResendSubscriptionsModal.tsx
+++ b/packages/app/src/resource/ResendSubscriptionsModal.tsx
@@ -1,0 +1,61 @@
+import { Button, Checkbox, Group, Modal, Stack } from '@mantine/core';
+import { showNotification } from '@mantine/notifications';
+import { getReferenceString, normalizeErrorString } from '@medplum/core';
+import { Resource } from '@medplum/fhirtypes';
+import { Form, ResourceInput, useMedplum } from '@medplum/react';
+import { useCallback, useState } from 'react';
+
+export interface ResendSubscriptionsModalProps {
+  readonly resource: Resource | undefined;
+  readonly opened: boolean;
+  readonly onClose: () => void;
+}
+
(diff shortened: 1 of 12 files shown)


packages/app/src/resource/ResendSubscriptionsModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+      await medplum.post(medplum.fhirUrl(resource.resourceType, resource.id, '$resend'), {
+        subscription: choose && subscription ? getReferenceString(subscription) : undefined,
+        verbose,
+      });
+      showNotification({ color: 'green', message: 'Done' });
+      onClose();
+    } catch (err) {
+      showNotification({ color: 'red', message: normalizeErrorString(err), autoClose: false });
+    }
+  }, [medplum, resource, onClose, choose, subscription, verbose]);
```
ThatOneBro, line 35: This callback is fine because we are not passing any handlers to `useDisclosure` and therefore `resendSubscriptiosnDisclosure[1].close` doesn't change identities due to `const { onClose } = {};` always being `undefined`. Otherwise if we give any un-memoized  callbacks to `ResendSubscriptionsModal` this always will recompute on every render.

In general, if using a callback from a prop in a `useCallback`, I prefer to use a `ref` and leave the `props.CALLBACK_NAME_HERE` out of the deps array. This prevents the constant re-computation when a callback changes identity.

ThatOneBro (approved):

packages/app/src/resource/ResendSubscriptionsModal.tsx
ThatOneBro: Reference: https://github.com/mantinedev/mantine/blob/master/packages/@mantine/hooks/src/use-disclosure/use-disclosure.ts

PR #4922, opened 2024-07-18
[Patient Intake Sample] Fixes fetching of questionnaire on app loading

filipeximenes:
#4909 fixed the loop problem on the welcome page but broke the app once the user was logged, this PR addresses fixes the issue.

### DIFF
--- a/examples/medplum-patient-intake-demo/src/App.tsx
+++ b/examples/medplum-patient-intake-demo/src/App.tsx
@@ -23,12 +23,13 @@ export function App(): JSX.Element | null {
     if (medplum.isLoading() || !profile) {
       return;
     }
-    try {
-      const intakeQuestionnaire = medplum.searchOne('Questionnaire', { name: 'patient-intake' }).read();
-      setIntakeQuestionnaire(intakeQuestionnaire);
-    } catch (err) {
-      console.error(err);
-    }
+
+    medplum
+      .searchOne('Questionnaire', { name: 'patient-intake' })
+      .then((intakeQuestionnaire) => {
+        setIntakeQuestionnaire(intakeQuestionnaire);
+      })
+      .catch(() => {});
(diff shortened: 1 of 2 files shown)


examples/medplum-patient-intake-demo/src/App.tsx
```diff
@@ -23,8 +23,12 @@ export function App(): JSX.Element | null {
     if (medplum.isLoading() || !profile) {
       return;
     }
-    const intakeQuestionnaire = medplum.searchOne('Questionnaire', { name: 'patient-intake' }).read();
-    setIntakeQuestionnaire(intakeQuestionnaire);
+    try {
+      const intakeQuestionnaire = medplum.searchOne('Questionnaire', { name: 'patient-intake' }).read();
+      setIntakeQuestionnaire(intakeQuestionnaire);
+    } catch (err) {
+      console.error(err);
```
rahul1, line 30: Why is this the fix? If there's an error being thrown, we should try to resolve it

filipeximenes: Sure, I was going to send you a message about this. My initial approach to load the questionnaire was using `useSearchOne` instead of `medplum.searchOne` which would work fine after user is logged in but would make the welcome page to stay in an infinite loop refreshing the page (this was the initial issue addressed in #4909).
Unfortunatelly I didn't notice that this broke the application after the user was logged in. The issue seems to be related to the promise in this request which is will raise an error signaling it is `Pending` and breaks the application. The `try/catch` fixes this issue. 
I actually spend about an hour trying to debug what is making `useSearchOne` to behave in such a way but couldn't find out so I decided to move on to another approach. My hunch is that it's some incompatibility it has when it's called in the same component where `const medplum = useMedplum()` is loaded.

rahul1 (approved):

Issue #4914, opened 2024-07-18
Report user AccessPolicies in the Medplum Client


rahul1:
User driven: https://medplum.slack.com/archives/C0720R0KVTQ/p1721025520098469

## Proposal:
 AccessPolicy.basedOn which would be an array of references to AccessPolicy resources.  then, when we "resolve" all of the nested and parameterized access policies, we can build that basedOn array.

rahul1:
One suggestion, maybe instead of an array of "AccessPolicy", it can be an array of `ProjectMembershipAccess` - this would allow the client to see the resolved variables as well

codyebberson:
Fixed in https://github.com/medplum/medplum/pull/4925

PR #4925, opened 2024-07-18
Add and set AccessPolicy.basedOn

codyebberson:
Context: Users can optionally have one or more `AccessPolicy` resources attached to their `ProjectMembership`.  Those `AccessPolicy` resources can be parameterized.  At runtime, Medplum server gathers everything and "resolves" them into a single ready-to-use `AccessPolicy`, which is used for actual enforcement.

A user requested having some kind of linkage in that "derived" or "resolved" `AccessPolicy` back to the originals.

This PR adds new `AccessPolicy.basedOn` element, and populates it during the resolution step.

User requested: https://medplum.slack.com/archives/C0720R0KVTQ/p1721025520098469

### DIFF
--- a/packages/definitions/dist/fhir/r4/fhir.schema.json
+++ b/packages/definitions/dist/fhir/r4/fhir.schema.json
@@ -60253,6 +60253,13 @@
           "description": "A name associated with the AccessPolicy.",
           "$ref": "#/definitions/string"
         },
+        "basedOn": {
+          "description": "Other access policies used to derive this access policy.",
+          "items": {
+            "$ref": "#/definitions/Reference"
+          },
+          "type": "array"
+        },
         "compartment": {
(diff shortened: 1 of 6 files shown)


rahul1:
@codyebberson I made one comment [here](https://github.com/medplum/medplum/issues/4914#issuecomment-2236984110) about setting the type of `basedOn` to be `ProjectMembershipAccess[]`, so that we can see the values of the resolved variables. What do you think about that!

codyebberson:
> @codyebberson I made one comment [here](https://github.com/medplum/medplum/issues/4914#issuecomment-2236984110) about setting the type of `basedOn` to be `ProjectMembershipAccess[]`, so that we can see the values of the resolved variables. What do you think about that?

`ProjectMembershipAccess` is a `BackboneElement` within `ProjectMembership`, so it cannot be referenced outside of `ProjectMembership`

rahul1 (approved):

Issue #4926, opened 2024-07-19
Medplum Web App Build Error - @Medplum/react-hooks


ha-ccoon:
Hi there,

 - Medplum version: 3.2.2
 -  Install Medplum by the  following documents:
https://www.medplum.com/docs/self-hosting/install-on-ubuntu
https://www.medplum.com/docs/contributing/run-the-stack#run-the-medplum-web-app

## Problem

I found the build error for the Medplum web app while trying the self-hosted version.

```sh
:~/medplum/packages/app$ npm run dev

> @medplum/app@3.2.2 dev
> vite

Re-optimizing dependencies because vite config has changed

  VITE v5.3.3  ready in 749 ms

  ➜  Local:   http://localhost:3000/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
Error:   Failed to scan for dependencies from entries:
  /home/test/medplum/packages/app/index.html

  ✘ [ERROR] Failed to resolve entry for package "@medplum/react-hooks". The package may have incorrect main/module/exports specified in its package.json. [plugin vite:dep-scan]

    ../../node_modules/esbuild/lib/main.js:1225:21:
      1225 │         let result = await callback({
           ╵                      ^

    at packageEntryFailure (file:///home/test/medplum/node_modules/vite/dist/node/chunks/dep-CzJTQ5q7.js:47164:15)
```

## Solved

Please see the linked PR.

## References
https://github.com/medplum/medplum/issues/3362

PR #4847, opened 2024-07-11
fix(react): make `@medplum/react-hooks` peer, mark as `external` in esbuild

ThatOneBro:

PR #4927, opened 2024-07-19
fix: web build error - vite config

ha-ccoon:
- [ ] https://github.com/medplum/medplum/issues/4926


codyebberson (approved):
attn @ThatOneBro 

1. I don't think this should be necessary today, because `app` only depends on `react`, not `react-hooks` (and `react-hooks` is re-exported by `react`)
2. This will be necessary after we merge https://github.com/medplum/medplum/pull/4847 because then `react-hooks` is a dependency via peer deps
3. I don't think there is any harm in merging this now

PR #4945, opened 2024-07-19
Project feature flag for reference lookup table

mattwiller:


packages/server/src/fhir/search.ts
```diff
@@ -1010,13 +1011,20 @@ function buildChainedSearch(selectQuery: SelectQuery, resourceType: string, para
...
     return buildChainedSearchUsingReferenceTable(selectQuery, resourceType, param);
   } else {
     return buildChainedSearchUsingReferenceStrings(selectQuery, resourceType, param);
   }
 }
 
+function usesReferenceLookupTable(): boolean {
+  return !!(
+    getConfig().chainedSearchWithReferenceTables ||
+    getAuthenticatedContext().project.features?.includes('reference-lookups')
```
codyebberson, line 1024: I don't think we can use `getAuthenticatedContext()` here.  It works for user `Repository` but would not be valid for `getSystemRepo()`.

Probably needs to be part of `RepositoryContext`?

packages/server/src/fhir/search.ts
```diff
@@ -79,1146 +79,1157 @@ export interface ChainedSearchParameter {
...
-  validateSearchResourceTypes(repo, searchRequest);
-
-  // Ensure that "count" is set.
-  // Count is an optional field.  From this point on, it is safe to assume it is a number.
-  if (searchRequest.count === undefined) {
-    searchRequest.count = DEFAULT_SEARCH_COUNT;
-  } else if (searchRequest.count > maxSearchResults) {
-    searchRequest.count = maxSearchResults;
-  }
+export class RepositorySearch {
```
mattwiller, line 82: This is easiest to review with "ignore whitespace", since most of these changes are just indentation

codyebberson (approved):
ty 🙏 

I filed a ticket to drop the old reference lookups for v4: https://github.com/medplum/medplum/issues/4969

PR #4949, opened 2024-07-19
Fix GraphQL max depth validation

mattwiller:
Fragments are now correctly included in the depth, but do not themselves count towards the query depth


packages/fhir-router/src/graphql/graphql.test.ts
```diff
@@ -552,7 +552,7 @@ describe('GraphQL', () => {
 
   test('Max depth', async () => {
     // The definition of "depth" is a little abstract in GraphQL
-    // We use "selection", which, in a well formatted query, is the level of indentation
+    // We use field depth, where fragment expansion does not contribute to depth
```
codyebberson, line 555: `does not contribute` / `does contribute`?

packages/fhir-router/src/graphql/graphql.test.ts
```diff
@@ -743,6 +762,13 @@ describe('GraphQL', () => {
                     resource {
                       ...on ServiceRequest {
                         id
+                        asNeededCodeableConcept {
+                          coding {
+                            extension {
+                              url
```
codyebberson, line 768: Out of curiosity, do we count these simple object traversals as "depth"?  I.e., navigating into a `BackboneElement`?

I'm realizing that we've been talking about "depth" in general, but there's a big difference between getting properties off of a nested `BackboneElement` vs resolving by searching.

packages/fhir-router/src/graphql/graphql.ts
```diff
@@ -442,25 +444,76 @@ const DEFAULT_MAX_DEPTH = 12;
...
-          new GraphQLError(`Field "${fieldName}" exceeds max depth (depth=${depth}, max=${maxDepth})`, {
-            nodes: node,
-          })
-        );
+  (context: ValidationContext): ASTVisitor =>
+    new MaxDepthVisitor(context, maxDepth);
+
+type DepthRecord = { depth: number; node?: FieldNode };
+
+class MaxDepthVisitor {
```
codyebberson, line 452: Should we add `implements ASTVisitor`?

packages/server/src/fhir/operations/graphql.test.ts
```diff
@@ -670,7 +675,7 @@ describe('GraphQL', () => {
     `,
       });
     expect(res2.status).toBe(400);
-    expect(res2.body.issue[0].details.text).toEqual('Field "id" exceeds max depth (depth=14, max=12)');
+    expect(res2.body.issue[0].details.text).toEqual('Field "id" exceeds max depth (depth=13, max=12)');
```
codyebberson, line 678: I assume this means our definition of "depth" changed, and before it was `14` and now it is `13`?

Does that mean we're implicitly allowing more depth?

codyebberson: Or is the `13` just due to short circuiting?

```
// Short circuit, no need to keep computing
```

packages/fhir-router/src/graphql/graphql.ts
```diff
@@ -442,25 +444,76 @@ const DEFAULT_MAX_DEPTH = 12;
...
+    }
+  }
+
+  /**
+   * Returns the depth of the GraphQL node in a query.
+   * We use field depth as the representation of depth: the number of concrete fields (not counting fragment expansions)
+   * @param nodes - The AST nodes.
+   * @returns The maximum "depth" of the nodes.
+   */
+  private getDepth(...nodes: ASTNode[]): DepthRecord {
```
codyebberson, line 483: I think this method makes sense.  Thanks for digging into the graphql AST to figure this out.  Oof.

Now that we're breaking the seal on AST parsing, I'm wondering if we should add in some intelligence here for the specifics of FHIR GraphQL.

My interpretation of this function is that it's treating all "fields" the same (same as before).

Maybe we should keep that as-is for now.

I assume that in the near future, we would want to add in checks for `isCapital(node.name[0])` or `node.name.endsWith('List')` to detect "resolve by ID", "resolve by search", "resolve by connection API", etc.

Or, even better, if it's possible to get the resolver from the AST.  I don't know if that's possible though.

Is that correct?  Too much for this PR?  I'd be fine either way, just trying to mind meld.

packages/fhir-router/src/graphql/graphql.test.ts
```diff
@@ -686,6 +673,38 @@ describe('GraphQL', () => {
...
+      query: {},
+      params: {},
+      body: {
+        query: `
+        {
+          CareTeamList {
+            ...CareTeamFields
+          }
+
+          fragment CareTeamFields {
```
codyebberson, line 690: Just so I understand this test...

Inside our `MaxDepthRule` AST code, this fragment is already substituted into `CareTeamList`?  So we would be getting the "correct" depth?

packages/fhir-router/src/graphql/graphql.test.ts
mattwiller: Ah, I see how this is ambiguous: the expansion itself (which does have a selection set) doesn't count, but the fields inside the expansion do count

packages/fhir-router/src/graphql/graphql.ts
mattwiller: That's basically how it's handled in the [query complexity calculation](https://github.com/medplum/medplum/pull/4924/files#diff-c401c28cb20cf21ea2918c9ba3d7bf86cbabb0152b53aa0e8be327e7ad80b56fR518-R526) — I'm inclined to have a simple and straightforward max depth that just counts fields (since even backbone elements increase branching factor and can contribute to query complexity), but we can also raise the limit to compensate?

packages/fhir-router/src/graphql/graphql.ts
mattwiller: I actually tried that: it doesn't work because of the way the `ASTVisitor` type is defined in `graphql-js`

packages/fhir-router/src/graphql/graphql.test.ts
mattwiller: The substitution isn't done in the AST itself: there are corresponding `FragmentDefinition` and `FragmentSpread` nodes.  But out complexity calculation code handles it: https://github.com/medplum/medplum/pull/4924/files#diff-c401c28cb20cf21ea2918c9ba3d7bf86cbabb0152b53aa0e8be327e7ad80b56fR533-R546

packages/fhir-router/src/graphql/graphql.test.ts
mattwiller: Yep, it's currently counted simply as each nested field counting as an additional layer of depth

rahul1 (approved):

PR #4951, opened 2024-07-20
fix(fhircast): associate topic with separate random connection endpoint

ThatOneBro:
This makes `FHIRcast` connections slightly more secure since knowing the secret `topic` will no longer be enough to connect to the topic WebSocket endpoint URL, and all users will need to POST a subscription request in order to receive the proper endpoint URL, which requires authentication to a protected route first.

This also makes the implementation more in line with the spec: https://fhircast.org/specification/STU2/#:~:text=When%20using%20WebSockets%2C%20the%20HTTP%20body%20of%20the%20response%20SHALL%20consist%20of%20a%20JSON%20object%20containing%20an%20element%20name%20of%20hub.channel.endpoint%20and%20a%20value%20of%20the%20WSS%20URL.%20The%20WebSocket%20WSS%20URL%20SHALL%20be%20cryptographically%20random%2C%20unique%2C%20and%20unguessable.

### DIFF
--- a/packages/server/src/fhircast/routes.ts
+++ b/packages/server/src/fhircast/routes.ts
@@ -151,9 +151,13 @@ async function handleSubscriptionRequest(req: Request, res: Response): Promise<v
   let subscriptionEndpoint: string;
   try {
     const results = await getRedis()
+      // Multi allows for multiple commands to be executed in a transaction
       .multi()
+      // Sets the endpoint key for this topic if it doesn't exist
       .setnx(`medplum:fhircast:topic:${topic}:endpoint`, generateId())
+      // Gets the endpoint, either previously generated endpoint secret or the newly generated key if a previous one did not exist
       .get(`medplum:fhircast:topic:${topic}:endpoint`)
+      // Executes the transaction
       .exec();
 
     if (!results) {
--- a/packages/server/src/fhircast/routes.ts
+++ b/packages/server/src/fhircast/routes.ts
@@ -5,9 +5,11 @@ import {
   generateId,
   getWebSocketUrl,
   normalizeErrorString,
+  serverError,
 } from '@medplum/core';
 import { Request, Response, Router } from 'express';
 import { body, oneOf, validationResult } from 'express-validator';
+import assert from 'node:assert';
 import { asyncWrap } from '../async';
 import { getConfig } from '../config';
 import { getLogger } from '../context';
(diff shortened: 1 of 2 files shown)


packages/server/src/fhircast/routes.ts
```diff
@@ -99,9 +101,39 @@ protectedCommonRoutes.post(
...
+
+      assert(results.length === 2, 'Redis did not return 2 command results for FHIRcast endpoint retrieval');
 
+      const [err, result] = results[1];
+      if (err) {
+        throw err;
+      }
+      subscriptionEndpoint = result as string;
+    } catch (err) {
+      sendOutcome(res, serverError(new Error('Failed to get endpoint for topic')));
```
codyebberson, line 128: Using `serverError` will return HTTP 500.  Maybe `badRequest` for HTTP 400 would be better?

packages/server/src/fhircast/routes.ts
```diff
@@ -99,9 +101,39 @@ protectedCommonRoutes.post(
...
     const topic = req.body['hub.topic'];
-    const subscriptionEndpoint = topic; // TODO: Create separate subscription endpoint for topic
-    const config = getConfig();
+    let subscriptionEndpoint: string;
+    try {
+      const results = await getRedis()
+        // Multi allows for multiple commands to be executed in a transaction
+        .multi()
+        // Sets the endpoint key for this topic if it doesn't exist
+        .setnx(`medplum:fhircast:topic:${topic}:endpoint`, generateId())
```
codyebberson, line 110: I think `topic` is an unvalidated string at this point... Is there any validation we can do?  Will it always be a UUID?

Not sure what the risk is here, but feels kinda SQL-injection-ish.

packages/server/src/fhircast/routes.ts
ThatOneBro: In this case it would actually be good if it returned a 500 since the only time this case should be hit is if something happens with Redis

packages/server/src/fhircast/routes.ts
ThatOneBro: Topic can be any string as far as I know, but the scheme implemented should be "secure". In our case for SMART launch it is a UUID, but I don't know if we can enforce that.

https://fhircast.org/specification/STU2/#:~:text=MAY%20be%20a%20Universally%20Unique%20Identifier%20(UUID). 
> MAY be a Universally Unique Identifier ([UUID](https://tools.ietf.org/html/rfc4122)).

I think we should just recommend against using easily guessable topics

codyebberson: Maybe a simple regex check for alphanumeric + dashes + underscores?  My primary concern here is a bad actor crafting a value that can muck with redis.

packages/server/src/fhircast/routes.ts
codyebberson: >  it would actually be good if it returned a 500 since the only time this case should be hit is if something catastrophic happens to Redis

Hm, can you check my logic here?  My interpretation is that a well crafted malicious HTTP request with an unknown topic ID would trigger a 500, which would wake people up in the middle of the night.  If that's correct, I think we should at least handle that one case and throw a 404.

packages/server/src/fhircast/routes.ts
codyebberson: Also should probably namespace this on project ID.  Especially if this is a user-submittable value, and we can probably assume users will do silly things like use very guessable `topic` like `abc` or whatever.

ThatOneBro: > Maybe a simple regex check for alphanumeric + dashes + underscores? My primary concern here is a bad actor crafting a value that can muck with redis.

I'm ok with enforcing UUIDs or at least alpha + dashes + underscores. And maybe a minimum length

> Also should probably namespace this on project ID. Especially if this is a user-submittable value, and we can probably assume users will do silly things like use very guessable topic like abc or whatever.

We could make it connected to project, but then it would kill the use case of cross-project communication with a general purpose hub. Not sure how much we care about that but I did think that was one of the use cases

packages/server/src/fhircast/routes.ts
ThatOneBro: This route is like an upsert kinda thing,
That's why we call `setnx` before `get` in this transaction. 

The logic is: 
1) We conditionally set the topic's url to be a new random UUID if it doesn't exist. If it does, skip this step.
2) We get the value of the topic's url

If this fails for some reason beyond what is normal Redis operation (thinking timeouts or something like that, not actually sure what cases would trigger this but they are outside of healthy)

codyebberson (approved):

PR #4961, opened 2024-07-23
Expand nested ValueSet references

### LINKED ISSUE #4839: Expand Valuesets that include other Valuesets
User Driven Example: https://medplum.slack.com/archives/C057GEVQCRE/p1720138395455249

mattwiller:
Resolves #4839

### DIFF
--- a/packages/server/src/fhir/sql.ts
+++ b/packages/server/src/fhir/sql.ts
@@ -150,7 +150,7 @@ export class Column {
 }
 
 export class Literal {
-  constructor(readonly value: any) {}
+  constructor(readonly value: string) {}
 }
 
 export interface Expression {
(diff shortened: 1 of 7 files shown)


packages/server/src/fhir/sql.ts
```diff
@@ -150,7 +150,7 @@ export class Column {
 }
 
 export class Literal {
-  constructor(readonly value: string) {}
+  constructor(readonly value: any) {}
```
codyebberson, line 153: I don't see any cases where this was necessary?

```suggestion
  constructor(readonly value: string) {}
```

codyebberson (approved):
lgtm 👍 

See comment on the `any` in the `Literal` constructor.  Your call.

PR #4974, opened 2024-07-26
Faster JOIN path for ValueSet hierarchy expansion

mattwiller:
Adding an index on `"Coding_Property" (target, property, coding)` and preloading the `"CodeSystem_Property"."id"` makes expanding a hierarchy significantly faster.  Additionally, when a text `filter` is present and likely to be relatively selective, the query starts with that filter and then checks whether those codes have the required ancestor.  Otherwise, it's faster is still recursively expand the hierarchy.


packages/server/src/fhir/operations/expand.ts
```diff
@@ -397,10 +403,22 @@ export function expansionQuery(
...
-          break;
         case 'descendent-of':
-          query = addDescendants(query, codeSystem, condition.value).where('code', '!=', condition.value);
+          if (params?.filter) {
+            const base = new SelectQuery('Coding', undefined, 'origin')
+              .column('id')
+              .column('code')
+              .column('display')
+              .where('system', '=', codeSystem.id)
+              .where(new Column('origin', 'code'), '=', 'code');
```
codyebberson, line 413: Is the `alias` required?  Would `new Column('Coding', 'code')` have worked?

mattwiller: My bad, this was a bug.  The updated implementation should be a major improvement over the current typeahead performance

packages/server/medplum.config.json
```diff
@@ -42,5 +42,6 @@
     "removeOnFail": { "count": 1 },
     "removeOnComplete": { "count": 1 }
   },
-  "shutdownTimeoutMilliseconds": 30000
+  "shutdownTimeoutMilliseconds": 30000,
+  "chainedSearchWithReferenceTables": true
```
codyebberson, line 46: 👍  - this makes sense.  we should all make mental note of this if/when developers experience weird behavior on localhost

PR #4983, opened 2024-07-29
Improve clamp on estimated search count

codyebberson:


packages/server/src/fhir/search.ts
```diff
@@ -518,10 +518,31 @@ async function getEstimateCount(
...
+  if (searchRequest.count === 0 || rowCount === undefined) {
+    // If "count only" or rowCount is undefined, then the estimate is the best we can do
+    return estimateCount;
+  }
+
+  const pageSize = searchRequest.count ?? DEFAULT_SEARCH_COUNT;
+  const startIndex = searchRequest.offset ?? 0;
+  const minCount = rowCount > 0 ? startIndex + rowCount : 0;
+  const maxCount = startIndex + Math.min(pageSize, rowCount);
+  return Math.max(minCount, Math.min(maxCount, estimateCount));
```
rahul1, line 545: Won't this always return just the end of the current page? 

Eg. 
```
estimateCount: 120,000
offset: 1
pageSize: 20
rowCount: 20
```

I would expect the clamped estimated count to be `120,000`, because there's no reason to throw it out. 

But tracing through, we would get
```
startIndex: 1
minCount: 21
maxCount: 21
retval: max(21, min(21, 120,000)) = 21
```

codyebberson: Doh

rahul1 (approved):

PR #4987, opened 2024-07-29
Remove ValueSet expand precomputation

mattwiller:
After shipping #4974 the precomputed expansions are no longer necessary to provide expected performance


codyebberson (approved):
Amazing 🎉

PR #4993, opened 2024-07-30
Use StructureDefinition.type 

codyebberson:
### Update 2024-07-31

First, let's consider the problem.  Here is a snippet of a `StructureDefinition` from the [HL7 C-CDA project](https://github.com/HL7/CDA-core-sd):

```js
{
  resourceType: 'StructureDefinition',
  url: 'http://hl7.org/cda/stds/core/StructureDefinition/Patient',
  name: 'Patient',
  status: 'active',
  kind: 'logical',
  abstract: false,
  type: 'http://hl7.org/cda/stds/core/StructureDefinition/Patient',
}
```

Expected behavior: we should be able to index this structure definition, and everything should work

Actual behavior: when we index this structure definition, it overwrites the FHIR `Patient`

Observations:
* `name` is `Patient` - when "indexing" this structure definition, we override the main FHIR `Patient`, which is bad
* `type` is a URL - in most FHIR structure definitions, `type === name`, but here we have something totally different

So, here is my new mental model for how to think about these various properties.

First, let's review some definitions from [StructureDefinition](https://hl7.org/fhir/r4/structuredefinition.html):

|  element  |  definition |
| --- | --- |
| `url` | The identifier that is used to identify this structure when it is referenced in a specification, model, design or an instance. |
| `name` | A Computer-ready name (e.g. a token) that identifies the structure - suitable for code generation. |
| `type` | The type the structure describes. type `uri`, [FHIRDefinedType](https://hl7.org/fhir/r4/valueset-defined-types.html) (Extensible) |

So here is my attempt at plain and simple explanation:
* `url` is the official unique identifier for the `StructureDefinition`
    * In the future, we should use this value more
    * We currently still rely on some assumptions that certain `type` implies certain `url`
    * In the future, we should instead use `getDataTypeByUrl()`
* `name` is a "computer-ready name"
    * This is what we previously used for generating type names
    * Due to the risk of conflicts, this value is not actually useful
* `type` is more important
    * We previously did not use this value
    * After this PR, this is the primary source of truth for the type name

And that gets to the main point of this PR:
* Before this PR, we used `name` to define a lot of types
* After this PR, we're using `type` to define types

Let's look at some examples:

<table>
<tr><td>name</td><td>type</td><td>file</td><td>notes</td></tr>
<tr><td colspan="4">http://hl7.org/fhir/StructureDefinition/Patient</td></tr>
<tr><td>Patient</td><td>Patient</td><td>profiles-resources.json</td><td>Clean and simple case of a FHIR defined type</td></tr>
<tr><td colspan="4">http://hl7.org/fhir/StructureDefinition/Observation</td></tr>
<tr><td>Observation</td><td>Observation</td><td>profiles-resources.json</td><td>Another clean and simple case of a FHIR defined type</td></tr>
<tr><td colspan="4">http://hl7.org/fhir/us/core/StructureDefinition/us-core-blood-pressure</td></tr>
<tr><td>USCoreBloodPressureProfile</td><td>Observation</td><td>uscore-v5.0.1-structuredefinitions.json</td><td>A FHIR Profile on Observation</td></tr>
<tr><td colspan="4">http://hl7.org/fhir/StructureDefinition/Quantity</td></tr>
<tr><td>Quantity</td><td>Quantity</td><td>profiles-types.json</td><td>FHIR defined type, but "complex-type", so not a resource</td></tr>
<tr><td colspan="4">http://hl7.org/fhir/StructureDefinition/SimpleQuantity</td></tr>
<tr><td>SimpleQuantity</td><td>Quantity</td><td>profiles-types.json</td><td>FHIR defined profile</td></tr>
<tr><td colspan="4">https://medplum.com/fhir/StructureDefinition/User</td></tr>
<tr><td>User</td><td>User</td><td>profiles-medplum.json</td><td>Medplum defined resource</td></tr>
<tr><td colspan="4">http://hl7.org/cda/stds/core/StructureDefinition/Patient</td></tr>
<tr><td>Patient</td><td>http://hl7.org/cda/.../Patient</td><td>profiles-ccda.json</td><td>HL7 C-CDA Patient type, note that "type" is a URL, meaning this is completely separate</td></tr>
</table>

--------

### Original Comments

Context: https://medplum.slack.com/archives/C02AMUU6SBH/p1720453993583959

This came up while doing more experiments with FHIR Mapping Language, and the HL7 `StructureDefinition` resources for C-CDA.

I suspect this is something similar to what @mattlong ran into with US-Core StructureDefinitions, as all of my prototyping keeps breaking these tests 😆😢

The immediate issue that I ran into is another `StructureDefinition` with `name === 'Patient'` which is completely disconnected from a FHIR `Patient` - it's not a profile, it's a totally different type.  We currently naively index types by name, so that simply overwrote the R4 `Patient` type, and everything fell apart

That sent me down the rabbit hole on the nitty gritty details of `name` vs `type` vs `url`, and I think we may need to revisit some pretty fundamental assumptions and implementation details.

Some options I'm considering:

1. Replace `StructureDefinition.name` with `StructureDefinition.type` in terms of the main identifier that we use for types and type definitions and profiles
    a. This may be the most technically correct model, but will also be the most invasive
2. Only index by `name` when the `url` is one of the "official" or whitelisted URL's
3. We could just have multiple indexes: `dataTypesByName`, `dataTypesByType`, `dataTypesByUrl` - and corresponding `getDataTypeByName`, etc.  That might end up being the most surgical change.
4. Probably a combination of 2 and 3
5. Something else?

There are a few interesting wrinkles with that:

1. We currently have TypeScript type definitions for `SimpleQuantity` and `MoneyQuantity`.  Technically, these are "profiles" on top of `Quantity`.  so we might need to figure out a more concrete philosophy on "which `StructureDefinitions` become types?"
2. The SQL-on-FHIR `ViewDefinition` type doesn't meet the definition of any of this, so we may need to just embrace some degree of special cases for types
3. There are a whole bunch of places where we're making the implicit assumption that the `targetProfile` URL can be parsed, and the last token becomes a type name.  Those probably need to all be re-opened 🤒🤮

### DIFF
--- a/packages/core/src/typeschema/types.ts
+++ b/packages/core/src/typeschema/types.ts
@@ -98,6 +98,10 @@ const PROFILE_DATA_TYPES: { [profileUrl: string]: DataTypesMap } = Object.create
 
 // Special case names for StructureDefinitions that are technically "profiles", but are used as base types.
 // This is for backwards compatibility with R4 StructureDefinitions that are used as base types.
+// MoneyQuantity and SimpleQuantity are technically "profiles" on Quantity, but we allow them to be used as base types.
+// ViewDefinition is a special case for SQL-on-FHIR.
+// We can add more types here in the future as necessary, when we want them to be used as base types.
+// For example, exporting new types in "@medplum/fhirtypes".
 const TYPE_SPECIAL_CASES: { [url: string]: string } = {
   'http://hl7.org/fhir/uv/sql-on-fhir/StructureDefinition/ViewDefinition': 'ViewDefinition',
   'http://hl7.org/fhir/StructureDefinition/Quantity': 'Quantity',
@@ -105,37 +109,28 @@ const TYPE_SPECIAL_CASES: { [url: string]: string } = {
   'http://hl7.org/fhir/StructureDefinition/SimpleQuantity': 'SimpleQuantity',
 };
 
-function getDataTypesMap(profileUrl?: string): DataTypesMap {
+function getDataTypesMap(profileUrl: string): DataTypesMap {
   let dataTypes: DataTypesMap;
-
-  if (profileUrl) {
-    dataTypes = PROFILE_DATA_TYPES[profileUrl];
-    if (!dataTypes) {
-      dataTypes = PROFILE_DATA_TYPES[profileUrl] = Object.create(null);
-    }
-  } else {
-    dataTypes = DATA_TYPES;
+  dataTypes = PROFILE_DATA_TYPES[profileUrl];
+  if (!dataTypes) {
+    dataTypes = PROFILE_DATA_TYPES[profileUrl] = Object.create(null);
   }
-
   return dataTypes;
 }
 
 /**
  * Parses and indexes structure definitions
  * @param bundle - Bundle or array of structure definitions to be parsed and indexed
- * @param profileUrl - (optional) URL of the profile the SDs are related to
  */
-export function indexStructureDefinitionBundle(
-  bundle: StructureDefinition[] | Bundle,
-  profileUrl?: string | undefined
-): void {
+export function indexStructureDefinitionBundle(bundle: StructureDefinition[] | Bundle): void {
   const sds = Array.isArray(bundle) ? bundle : (bundle.entry?.map((e) => e.resource as StructureDefinition) ?? []);
   for (const sd of sds) {
-    loadDataType(sd, profileUrl);
+    loadDataType(sd);
   }
 }
 
-export function loadDataType(sd: StructureDefinition, profileUrl?: string | undefined): void {
+export function loadDataType(sd: StructureDefinition): void {
+  const profileUrl = sd.url;
   if (!sd?.name) {
     throw new Error(`Failed loading StructureDefinition from bundle`);
   }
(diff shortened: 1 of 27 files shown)


packages/react/src/ExtensionDisplay/ExtensionDisplay.tsx
```diff
@@ -75,7 +75,7 @@ export function ExtensionDisplay(props: ExtensionDisplayProps): JSX.Element | nu
   return (
     <BackboneElementDisplay
       path={props.path}
-      value={{ type: typeSchema.name, value: props.value }}
+      value={{ type: typeSchema.type as string, value: props.value }}
```
mattlong, line 78: Can we make `InternalTypeSchema.type` and perhaps `InternalTypeSchema.url` required now to avoid having to cast them to string?

packages/core/src/typeschema/types.ts
```diff
@@ -138,7 +146,26 @@ export function loadDataType(sd: StructureDefinition, profileUrl?: string | unde
...
+    sd.type?.startsWith('http://') ||
+    sd.type?.startsWith('https://') ||
+    profileUrl
+  ) {
+    dataTypes[sd.type] = schema;
+  }
+
+  // Special cases by "name"
+  // These are StructureDefinitions that are technically "profiles", but are used as base types
+  if (!profileUrl) {
```
mattlong, line 163: It seems like it would be possible for a StructureDefinition to never wind up being added to `dataTypes` and just silently pass through `loadDataType`, i.e. if neither this if statement or the one before it passes. Compared to before where all SDs would definitely wind up being added via `dataTypes[sd.name] = schema`.

Should we be throwing if that happens?

packages/core/src/typeschema/types.ts
```diff
@@ -138,7 +146,26 @@ export function loadDataType(sd: StructureDefinition, profileUrl?: string | unde
...
+    const specialCase = TYPE_SPECIAL_CASES[sd.url];
+    if (specialCase) {
+      dataTypes[specialCase] = schema;
+    }
+  }
```
mattlong, line 168: Does the same logic changes looking for special cases and using type vs name need to be applied in the `for (const inner of schema.innerTypes)` loop down below?

packages/core/src/typeschema/types.ts
```diff
@@ -97,6 +96,15 @@ const PROFILE_SCHEMAS_BY_URL: { [profileUrl: string]: InternalTypeSchema } = Obj
 // is maintained per profile URL
 const PROFILE_DATA_TYPES: { [profileUrl: string]: DataTypesMap } = Object.create(null);
 
+// Special case names for StructureDefinitions that are technically "profiles", but are used as base types.
```
mattlong, line 99: What's the algorithm for determining this set of special cases? Could you add that as a comment?

mattlong (changes requested):

packages/react/src/ExtensionDisplay/ExtensionDisplay.tsx
mattlong: Those get assigned from their SD equivalents which don't allow `undefined`
The same applies to `InternalTypeSchema.kind`

packages/core/src/typeschema/types.ts
codyebberson: Right now it's 100% manual.  Yes, will add comments here on the philosophy.

packages/core/src/typeschema/types.ts
codyebberson: Summarizing offline discussion and changes:

It turns out the `profileUrl` param is 100% of the time the same as `StructureDefinition.url`.  I removed the param, and went through all of the cleanup.

packages/react/src/ExtensionDisplay/ExtensionDisplay.tsx
codyebberson: I looked into this, and the reason they are optional is due to "inner" types.

For example, the FHIR `Patient` type ends up being multiple Medplum types:

1. `Patient`
2. `PatientCommunication`
3. `PatientContact`
4. `PatientLink`

Those inner types do not have a FHIR `type`, `url`, or `kind`.

packages/core/src/typeschema/types.ts
codyebberson: Fixed ✅

packages/core/src/typeschema/types.ts
codyebberson: Fixed ✅

packages/core/src/typeschema/types.ts
codyebberson: Fixed ✅

mattlong (approved):
As discussed offline, we need to either add `url` to the fields being requested in `client.requestSchema` or preferably just request the SD through the FHIR endpoint rather than using GQL at all. It would be nice to request the subset of fields we actually need, but that might not be any better in terms of fibbing about the return type of the GQL query.

At least then any issues would be more likely to be caught at test time since GQL queries aren't supported in mock client.

codyebberson:
Deployed and tested on staging ✅ 

https://app.staging.medplum.dev/Patient?_count=20&_fields=id,_lastUpdated,name,birthDate,gender&_offset=0&_sort=-_lastUpdated

PR #5001, opened 2024-07-31
[Photon Health] Add props to photon elements

ksmith94:
This adds the necessary props so that the Photon Elements are now functional.


examples/medplum-photon-integration/src/config.ts
```diff
@@ -0,0 +1,3 @@
+// Default values to enable Photon Elements. Replace with your own photon credentials for use. https://docs.photon.health/docs/elements
+export const PHOTON_CLIENT_ID = 'NzATQk3GC5qoCZHWtLH1WHNV3pcZesPJ';
```
rahul1, line 2: Let's remove Medplum's client ID from this PR

rahul1 (changes requested):

rahul1 (approved):

PR #5014, opened 2024-08-02
Use x-amzn-trace-id header if available

codyebberson:


packages/server/src/context.ts
```diff
@@ -140,6 +140,27 @@ export function getTraceId(req: Request): string | undefined {
...
+  // Example header: X-Amzn-Trace-Id: Root=1-67891233-abcdef012345678912345678
+  // Example header: X-Amzn-Trace-Id: Self=1-67891233-12456789abcdef012345678;Root=1-67891233-abcdef012345678912345678
+  // Example in Athena: "TID_e0fbe3c75b3c5a45ab84fb156906649b"
+  const parts = amznTraceId.split(';');
+  for (const part of parts) {
+    const [key, value] = part.split('=');
+    if (key === 'Root' || key === 'Self') {
+      return value;
+    }
+  }
```
mattwiller, line 163: This might be more efficient with a regex, to avoid all the array creation from splitting

mattwiller (approved):

PR #5023, opened 2024-08-02
fix(SearchControl): prevent rerender triggering `loadResults`

ThatOneBro:


packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -550,7 +558,7 @@ export function SearchControl(props: SearchControlProps): JSX.Element {
   );
 }
 
-export const MemoizedSearchControl = memo(SearchControl);
+export const MemoizedSearchControl = SearchControl;
```
codyebberson, line 561: Should we add a comment marking this as deprecated?

codyebberson (approved):
Nice, this looks good.  Let's deploy to staging for some extra testing.

Should we check for all existing use of `MemoizedSearchControl` in examples?

codyebberson:
also, can we add a test that exercises the new `deepEquals` check?

image

PR #5028, opened 2024-08-05
Allow conditional references in Resource.meta

mattwiller:


packages/server/src/fhir/repo.test.ts
```diff
@@ -1010,11 +1010,13 @@ describe('FHIR Repo', () => {
 
       const patient = await systemRepo.createResource<Patient>({
         resourceType: 'Patient',
+        meta: { account: createReference(practitioner) },
```
rahul1, line 1013: Might be worth using a conditional reference here to test the resolution

rahul1 (approved):
One suggestion on the tests, then g2g for me

PR #5037, opened 2024-08-06
Add sampling config for slow query log

mattwiller:


packages/server/src/config.ts
```diff
@@ -55,6 +55,7 @@ export interface MedplumServerConfig {
   heartbeatEnabled?: boolean;
   accurateCountThreshold: number;
   slowQueryThresholdMilliseconds?: number;
+  slowQuerySampleRate?: number;
```
codyebberson, line 58: Need to add numeric param handling in `packages/server/src/cloud/aws/config.ts`

codyebberson: Apologies in advance, because that is some unfortuante tech debt.  Someday we should figure out a proper solution for that.

codyebberson (approved):

PR #5038, opened 2024-08-06
Add request ID to Postgres transaction error logs

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -2029,6 +2029,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
     this.transactionDepth++;
     const conn = await this.getConnection(DatabaseMode.WRITER);
     if (this.transactionDepth === 1) {
+      await conn.query(`SET application_name TO '${tryGetRequestContext()?.requestId ?? ''}'`);
```
codyebberson, line 2032: are there any issues if `application_name` is empty string?

codyebberson (approved):

packages/server/src/fhir/repo.ts
codyebberson: Looked into this, and the answer is "no", there are no risks.

From a readability perspective, we may want to default to `'system'`, but that's minor and low priority.

PR #5060, opened 2024-08-08
Update function signatures for crawlResource

mattwiller:
Per a conversation with @codyebberson, creating separate sync and async versions of the `crawlResource` utility function and deprecating the legacy function signatures.  The old versions of the function can be removed later, in the next major version of `@medplum/core`


codyebberson (approved):
Nice, clean, ty 🙏

PR #5068, opened 2024-08-09
Batch writes to resource tables in reindex

mattwiller:
Batching writes to the primary resource tables when reindexing significantly reduces the time needed to complete the job.  In local testing on my laptop, I observed the following performance for a reindex over three resource types with around 25k + 3k + 2k = 30k total test resources in them:

| | Observation | Condition | DiagnosticReport | Total Wall Time |
| -------- | ----------- | --------- | ---------------- | --------------- |
| **Baseline** | 214 s | 26.6 s | 17.6 s | 4:19 |
| **Batched writes** | 120 s (-44%) | 17.2 s (-35%) | 9.71 s (-45%) | 2:39 (-39%) |

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -903,9 +903,9 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
       resource = resources[i];
       (resource.meta as Meta).compartment = this.getCompartments(resource);
 
-      await this.writeResource(conn, resource);
       await this.writeLookupTables(conn, resource, false);
     }
+    await this.batchWriteResources(conn, ...resources);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -903,9 +903,9 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
       resource = resources[i];
       (resource.meta as Meta).compartment = this.getCompartments(resource);
 
-      await this.writeResource(conn, resource);
       await this.writeLookupTables(conn, resource, false);
     }
+    await this.batchWriteResources(conn, ...resources);
```
ThatOneBro, line 908: Only nit is that this creating a new array with a lot of new indices, though it's probably negligible on perf

ThatOneBro (approved):

PR #5070, opened 2024-08-10
null-check-added@readResources

### LINKED ISSUE #4873: Client throws error with missing id in `readResource`
Consider the following SDK call:
`medplum.readResource('Patient', undefined)`

This translates to the HTTP call: 
`GET Patient/`
which returns a bundle. 

This is confusing for client side code, that is returning a single `Patient` resource, and lead to confusing errors. 

## Proposal
the SDK should throw an error is the `id` parameter is nullish or `''`

SURAJ-SHARMA27:
Resolves #4873 

I have added a null check so that if the ID is null or an empty string, the readResource function will throw an error(I have also tested it locally).
```
if (id == null || id.trim() === '') {
      throw new Error('The "id" parameter cannot be null, undefined, or an empty string.');
    }
```

[screen-capture (17).webm](https://github.com/user-attachments/assets/0dcb7e13-22bc-4bb8-ac5c-2f33f2b04a91)

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -1662,7 +1662,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    */
   readResource<K extends ResourceType>(
     resourceType: K,
-    id?: string,
+    id: string,
     options?: MedplumRequestOptions
   ): ReadablePromise<ExtractResource<K>> {
     if (!id) {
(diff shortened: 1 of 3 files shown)


packages/core/src/client.ts
```diff
@@ -1665,8 +1665,13 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
     id: string,
     options?: MedplumRequestOptions
   ): ReadablePromise<ExtractResource<K>> {
+    if (id == null || id.trim() === '') {
```
codyebberson, line 1668:
```suggestion
    if (!id) {
```

codyebberson (commented):
Thanks @SURAJ-SHARMA27 !  Can you please add a test as well?  Thank you!

SURAJ-SHARMA27:
@CodyEbberson Sir, I have added the test cases for the changes I made. Could you please review them?

codyebberson (approved):

Issue #3007, opened 2023-10-08
Certify and test Application access – all data request § 170.315(g)(9)


reshmakh:
* [Official documentation](https://www.healthit.gov/test-method/application-access-all-data-request)
* [Proctor sheet](https://docs.google.com/document/d/1i9W0U3BlGUuvY_Dm5Phni8GIIeTL2ls7jFxfR57-0eE/edit)

reshmakh:
Related test https://github.com/medplum/medplum/issues/4345

codyebberson:
We did it! 🎉

PR #5072, opened 2024-08-11
Handle non-resource types in crawler and validator

codyebberson:
In service of https://github.com/medplum/medplum/issues/3005 and https://github.com/medplum/medplum/issues/4726

I've been experimenting with the HL7 FHIR Mapping Language maps for C-CDA to/from FHIR: https://github.com/hl7ch/cda-fhir-maps

One of the challenging parts is reading the XML files to the necessary JSON structure (i.e., when to treat XML elements as singletons vs arrays, when to parse attributes as numbers vs strings, etc).

Thankfully, HL7 publishes `StructureDefinition` resources for the C-CDA types, so we can validate the input: https://github.com/HL7/CDA-core-sd

This uncovered a couple bugs/limitations of our resource validation and "crawler" utilities, because they relied on assumptions of conventions in the FHIR R4 `StructureDefinition` resources.

The main two fixes here:

1. We cannot assume any implicit relationship between `StructureDefinition.name`, `StructureDefinition.type`, and `StructureDefinition.snapshot.element[0].path`.
    * We fixed most of the `name` vs `type` issues in https://github.com/medplum/medplum/pull/4993
    * The jump from `path` back to `type` was still lingering in `crawler`, which is used by `validate`
    * This is fixed by using `TypedValue` everywhere, and not relying on `toTypedValue(myObj)`
2. We cannot assume that choice-of-type elements have the `[x]` path suffix.
    * This assumption was really only in `getTypedPropertyValueWithSchema`
    * To be clear, this isn't 100% fixed, because we now cannot know the type of the property value
    * For example, with `Observation.value[x]`, you know `valueString` is a `string` and `valueInteger` is an `integer`, etc
    * But for these C-CDA types, the property names don't give any information, so currently we just take the first match
    * You could try to be clever and look for the first value that validates (gross)
    * For now, I'm doing the simple thing, which is good enough for the C-CDA types, and does not affect the R4 types at all

Then two additional drive-by fixes:

3. I renamed `crawlResource`/`crawlResourceAsync` to `crawlTypedValue`/`crawlTypedValueAsync`
    * This is perhaps a minor point, but it is more technically correct
    * We need to preserve the `TypedValue.type` information
    * Rather than adding a whole new set of methods and overloads, I went ahead and marked the old ones as deprecated
    * So now `crawlTypedValue`/`crawlTypedValueAsync` are the final methods to rule them all
4. Replaced `isLowerCase(value.type.charAt(0))` with a proper `isPrimitiveType(value.type)`
    * There appears to be a convention of using full HTTP URL's for `StructureDefinition.type` values
    * Obviously `http://...` starts with a lower case letter, and those values are not primitives
    * So now, we check against the known list of primitive types, which we already have in the code

### DIFF
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -36,7 +36,7 @@ import { HTTP_HL7_ORG, LOINC, RXNORM, SNOMED, UCUM } from '../constants';
 import { ContentType } from '../contenttype';
 import { OperationOutcomeError } from '../outcomes';
 import { createReference, deepClone } from '../utils';
-import { indexStructureDefinitionBundle } from './types';
+import { indexStructureDefinitionBundle, loadDataType } from './types';
 import { validateResource, validateTypedValue } from './validation';
 
 describe('FHIR resource validation', () => {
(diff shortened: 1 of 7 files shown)


reshmakh:
This is work toward #3005 and #3007 adding here for visibility.

packages/core/src/typeschema/validation.test.ts
```diff
@@ -1535,6 +1535,115 @@ describe('FHIR resource validation', () => {
...
+              max: '*',
+            },
+            type: [{ code: 'dateTime' }, { code: 'time' }],
+          },
+        ],
+      },
+    };
+    loadDataType(sd);
+
+    const value = { effectiveTime: ['2022-02-02T12:00:00Z'] };
```
mattwiller, line 1643: Would it be possible to also add a case with a valid `time`, and an invalid type, just to cover the different permutations?

mattwiller (approved):

PR #5076, opened 2024-08-12
FHIRMapper cast, cc, and import conceptmap

codyebberson:


packages/core/src/fhirmapper/transform.ts
```diff
@@ -560,6 +593,57 @@ function evalAppend(ctx: TransformContext, target: StructureMapGroupRuleTarget):
...
+ * @returns The evaluated target values.
+ * @internal
+ */
+function evalCast(ctx: TransformContext, target: StructureMapGroupRuleTarget): TypedValue[] {
+  const arg1 = resolveParameter(ctx, target.parameter?.[0])?.[0];
+  const arg2 = resolveParameter(ctx, target.parameter?.[1])?.[0]?.value;
+  if (arg2 === 'string') {
+    return [{ type: 'string', value: arg1?.value?.toString() }];
+  }
+  return [arg1];
```
mattwiller, line 614: Should this be more defensive?  I'm not too familiar with how this is used in practice

mattwiller (approved):

packages/core/src/fhirmapper/transform.ts
codyebberson: In short term for the C-CDA maps, this covers 100% of our needs.

In longer term, yeah, there's a lot of work here.  The FHIRPath implementation already includes a lot of casting and conversion logic.  It would be nice to tap into that and dispatch to the correct conversion code.  That's a project though.

PR #5077, opened 2024-08-12
Dependency upgrades 2024-08-12

medplumbot:
Dependency upgrades


codyebberson (changes requested):
Not working on staging, needs investigation

codyebberson:
Staging was actually working, my phone autocompleted to https://app.staging.medplum.com not https://app.staging.medplum.dev 🤦 

I'm looking into why `package-lock.json` jumped up by 2000 lines...

I noticed that we're including some `@types/` dependencies that are no longer needed, notably `@types/jest` and `@types/express-rate-limit`.

The remainder are required:

```bash
$ find ./packages -name "package.json" -not -path "*/node_modules/*" -exec grep -oP '"@types/[^"]*' {} + | sed 's/^.*:"@types\///' | sort | uniq
async-eventemitter
base-64
bcryptjs
body-parser
bytes
compression
cookie-parser
cors
express
jest
json-schema
mailparser
node
node-fetch
nodemailer
pdfmake
pg
qrcode
react
react-dom
set-cookie-parser
supertest
text-encoding
ua-parser-js
unzipper
uuid
validator
ws
```

codyebberson:
The reason for the extra 2000 lines in `package-lock.json` is that the latest version of NPM started adding `license` to all dependencies 🙄 😢

codyebberson (approved):

PR #5082, opened 2024-08-12
fix(expo-polyfills): split keys larger than 2000 chars

### LINKED ISSUE #5055: Medplum Client/ExpoClientStorage size limit issue
Using `ExpoClientStorage` for the storage function of `MedplumClient` in React Native is causing this warning: 
"Value being stored in SecureStore is larger than 2048bytes and it may not be stored successfully. In a future SDK version, this call may throw an error".

My team has been receiving reports of behavior that appears to be caused by storage failures described in this warning. 
As a workaround, we have switched to `MemoryStorage` from @medplum/core rather than `ExpoClientStorage`.

ThatOneBro:
Fixes #5055

### DIFF
--- a/packages/expo-polyfills/src/storage.ts
+++ b/packages/expo-polyfills/src/storage.ts
@@ -2,6 +2,8 @@ import { ClientStorage, IClientStorage, OperationOutcomeError } from '@medplum/c
 import * as SecureStore from 'expo-secure-store';
 import { Platform } from 'react-native';
 
+const MAX_VALUE_LENGTH = 2000;
+
 export interface IExpoClientStorage extends IClientStorage {
   getInitPromise(): Promise<void>;
   length: number;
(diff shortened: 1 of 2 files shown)


packages/expo-polyfills/src/storage.ts
```diff
@@ -139,6 +141,56 @@ export class SyncSecureStorage implements Storage {
...
+    // Append ___continued___ if this isn't the only chunk
+    // Otherwise just put the full chunk in
+    const promises = [];
+    promises.push(SecureStore.setItemAsync(key, chunks.length === 1 ? chunks[0] : `${chunks[0]}___continued___`));
+    for (let i = 1; i < chunks.length; i++) {
+      promises.push(
+        SecureStore.setItemAsync(
+          `${key}___chunk${i}___`,
+          // If this isn't the last chunk append ___continued___ to it
+          i === chunks.length - 1 ? chunks[i] : `${chunks[i]}___continued___`
```
codyebberson, line 187: If `chunks[i]` is 2000 characters long, will appending `__continued__` exceed the limit?

ThatOneBro: The limit is 2048, so 2000 + 13 shouldn't exceed. Good question though

codyebberson (approved):

PR #5085, opened 2024-08-13
Added server config setting for max search offset

codyebberson:


packages/server/src/fhir/search.ts
```diff
@@ -177,7 +177,18 @@ function applyCountAndOffsetLimits<T extends Resource>(
     searchRequest.count = maxSearchResults;
   }
 
-  searchRequest.offset ??= 0;
+  if (searchRequest.offset === undefined) {
+    searchRequest.offset = 0;
+  } else {
+    const maxOffset = getConfig().maxSearchOffset;
+    if (maxOffset !== undefined) {
+      if (searchRequest.offset > maxOffset) {
```
ThatOneBro, line 185:
```suggestion
    if (maxOffset !== undefined && searchRequest.offset > maxOffset) {
```
Can we collapse these?

ThatOneBro (approved):

packages/server/src/fhir/search.ts
codyebberson: Yeeep, good catch 👍

PR #5088, opened 2024-08-13
Enable setting DB parameters via CDK

mattwiller:


packages/core/src/config.ts
```diff
@@ -49,6 +49,7 @@ export interface MedplumSourceInfraConfig {
   rdsSecretsArn?: ValueOrExternalSecret<string>;
   rdsReaderInstanceType?: ValueOrExternalSecret<string>;
   rdsProxyEnabled?: ValueOrExternalSecret<boolean>;
+  rdsDBParameters?: StringMap;
```
codyebberson, line 52: Any objections to simply `rdsParameters`?  Or maybe `rdsClusterParameters` to be more specific?

`DB` is a tricky abbreviation.

I suppose we could always go `rdsDatabaseParameters`?

Naming things is hard.

codyebberson (approved):
ty ty 👍

PR #5094, opened 2024-08-14
chore(core/client): #5090 exported `UnauthenticatedError`

### LINKED ISSUE #5090: Typesafe Unauthenticated Error
User Driven: https://medplum.slack.com/archives/C057GEVQCRE/p1723429159091039

Proposal is to type the unauthenticated error as `MedplumUnauthenticatedError` to allow the use of `error instanceof `MedplumUnauthenticatedError`

dillonstreator:
Fixes #5090

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -33,6 +33,8 @@ import {
   notFound,
   serverError,
   unauthorized,
+  unauthorizedTokenAudience,
+  unauthorizedTokenExpired,
 } from './outcomes';
 import { MockAsyncClientStorage } from './storage';
 import { getDataType, isDataTypeLoaded, isProfileLoaded } from './typeschema/types';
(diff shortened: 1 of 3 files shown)


dillonstreator:
@codyebberson Thank you for the detailed write-up and no worries on the response time (this seemed relatively low priority)!
I noticed most of the use of `OperationOutcomeError` in the client was used _after_ a request to the external FHIR server. In re-review, I see that fhircast subscribing/unsubscribing has client-side validation that uses the `OperationOutcomeError` pre-request. Agreed that with this being part of the public interface, it makes sense to consider the long-term strategic approach that interops with the existing primitives and patterns and aligns closer with the FHIR-spec rather than this quick & dirty solution. Also agreed that these are just different flavors of 'unauthenticated' errors and can extend the existing `unauthenticated` error to provide additional issue context.

codyebberson (approved):

codyebberson (approved):

PR #5098, opened 2024-08-14
Search pagination with _cursor and lastUpdated

codyebberson:
What is this?
* Before:
    * Pagination via `Bundle.link` and `next` was done using the `_offset` parameter
    * The `_offset` parameter translates to a SQL `LIMIT`, which is very slow with values >1m
    * This is leading to HTTP 504 timeouts for various requests
* After:
    * When possible, the `next` link will now use a `_cursor` parameter
    * "When possible" right now is narrowly defined as "exactly one sort parameter of `_lastUpdated` ascending"
    * The `_cursor` parameter is essentially just a `_lastUpdated` timestamp value that is managed by the server

What does this *not* do?
* It leaves the existing `_offset` parameter alone, everything still works as expected
* It does not alter the `<SearchControl>` react component at all

Future work:
1. Update Bulk Export to use this behavior
2. Use a smarter cursor for both `next` and `previous` links?


packages/server/src/fhir/search.test.ts
```diff
@@ -4027,6 +4028,41 @@ describe('FHIR Search', () => {
...
+              resourceType: 'Task',
+              status: 'accepted',
+              intent: 'order',
+              code: { text: 'cursor_test' },
+            });
+            expect(task.meta?.lastUpdated).toBeDefined();
+            expect(task.meta?.lastUpdated).not.toEqual(prevLastUpdated);
+            tasks.push(task);
+            prevLastUpdated = task.meta?.lastUpdated as string;
+            await sleep(1);
```
mattlong, line 4047: Can you use `jest.useFakeTimers` and `jest.advanceTimersByTime ` instead?

packages/server/src/fhir/search.ts
```diff
@@ -218,14 +228,25 @@ interface GetSelectQueryForSearchOptions extends GetBaseSelectQueryOptions {
...
-  builder.offset(searchRequest.offset ?? 0);
+  if (searchRequest.offset > 0) {
+    builder.offset(searchRequest.offset);
+  }
+  if (searchRequest.cursor) {
+    const cursor = parseCursor(searchRequest.cursor);
+    if (cursor) {
+      builder.orderBy(new Column(searchRequest.resourceType, 'lastUpdated', false));
+      if (cursor.instant) {
+        builder.where(new Column(searchRequest.resourceType, 'lastUpdated'), '>', cursor.instant);
```
mattlong, line 246: Using `>` seems like it has an edge case when two resources have the same `lastUpdated` but only one of them is included in the previous page/batch. The other resource would be skipped.

packages/server/src/fhir/search.ts
```diff
@@ -526,42 +549,165 @@ async function getSearchRevIncludeEntries(
...
-      result.push({
-        relation: 'previous',
-        url: getSearchUrl({ ...searchRequest, offset: offset - count }),
-      });
+/**
+ * Returns the maximum lastUpdated values from a set of search bundle entries.
+ * @param entries - The search bundle entries.
+ * @returns The maximum lastUpdated values.`
+ */
+function getMaxLastUpdated(entries: BundleEntry[]): string {
```
mattlong, line 665: Another edge case: if there are more than `count` records with the same `lastUpdated` that happen to line up with a given page/batch, the cursor-based `next` links will point back to the same set of records, no?

mattlong: Actually, I think this would just be a similar situation as the other edge case I called out. I don't think there's an infinite loop possibility with `>`

packages/server/src/fhir/search.ts
```diff
@@ -218,14 +228,25 @@ interface GetSelectQueryForSearchOptions extends GetBaseSelectQueryOptions {
...
   const builder = getBaseSelectQuery(repo, searchRequest, options);
   addSortRules(builder, searchRequest);
-  const count = searchRequest.count as number;
+  const count = searchRequest.count;
   builder.limit(count + (options?.limitModifier ?? 1)); // Request one extra to test if there are more results
-  builder.offset(searchRequest.offset ?? 0);
+  if (searchRequest.offset > 0) {
+    builder.offset(searchRequest.offset);
+  }
+  if (searchRequest.cursor) {
```
mattwiller, line 241: Should this be an `else if`?  Would we ever want someone combining these?

codyebberson: Yeah... I couldn't decide which one should have priority.

I suppose if both are set, that should be an error?

packages/server/src/fhir/search.ts
```diff
@@ -526,42 +549,165 @@ async function getSearchRevIncludeEntries(
...
+  return { version: parts[0], instant: parts[1] };
+}
+
+/**
+ * Formats a cursor object into a cursor string.
+ * @param cursor - The cursor object.
+ * @returns The cursor string.
+ */
+function formatCursor(cursor: Cursor): string {
+  return `${cursor.version}-${cursor.instant}`;
```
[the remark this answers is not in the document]
codyebberson: Yeah, you mentioned that yesterday, and I looked into it.  Perhaps I am rationalizing laziness, but here's my reasons for not doing it:

1. General YAGNI
2. Base64 will add friction, but ultimately nothing will truly prevent people if they want to reverse engineer
3. In this case, there really isn't any advantage to reverse engineering, because this is just a convenience around `_lastUpdated` filters
4. There's also a performance consideration that we use `repo.search()` all over the place internally, and often do not need the `Bundle.link` values.  I know base64 encoding will be fast and cheap relative to the actual db query, but it still feels wrong.  That could probably be avoided with some kind of `_nolinks=true` option, but that's additional work.

packages/server/src/fhir/search.test.ts
codyebberson: After using `lastUpdated` and `id` together, the sleep and timers are unnecessary.

packages/server/src/fhir/search.ts
codyebberson: New model:
* Sort by `lastUpdated` ascending and `id` asc
* Filter on `(lastUpdated=x and id>y) or lastUpdated>x`

After that, no edge cases with multiple resources having the same `lastUpdated`

✅

packages/server/src/fhir/search.ts
codyebberson: Same as above, edge cases should be fixed.

✅

packages/server/src/fhir/search.ts
codyebberson: Made it an error to use offset>0 and cursor.

✅

mattlong (approved):

codyebberson:
Hm, deployed to staging, investigating:

image

PR #5100, opened 2024-08-15
Use turbo for linting

mattlong:
Previously we ran eslint against the entire repo which seems to be hitting some scaling issues and surfacing errors unrelated to the diffs of PRs/commits such as [this build](https://github.com/medplum/medplum/actions/runs/10391113203/job/28774368521?pr=5093).

### DIFF
--- a/examples/foomedical/tsconfig.json
+++ b/examples/foomedical/tsconfig.json
@@ -1,5 +1,4 @@
 {
-  "extends": "../../tsconfig.json",
   "compilerOptions": {
     "target": "ESNext",
     "useDefineForClassFields": true,
@@ -15,5 +14,6 @@
     "resolveJsonModule": true,
     "noEmit": true,
     "jsx": "react-jsx"
-  }
+  },
+  "exclude": ["dist"]
 }
--- a/examples/foomedical/tsconfig.json
+++ b/examples/foomedical/tsconfig.json
@@ -1,4 +1,5 @@
 {
+  "extends": "../../tsconfig.json",
   "compilerOptions": {
     "target": "ESNext",
     "useDefineForClassFields": true,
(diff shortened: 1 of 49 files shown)


examples/foomedical/tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "extends": "../../tsconfig.json",
```
codyebberson, line 2: Sorry, can't do this, because `foomedical`, and all packages in `examples` get copied to separate repos.  These packages cannot reference anything outside of their respective folders.

turbo.json
```diff
@@ -20,6 +20,7 @@
       "dependsOn": ["build"],
       "outputs": ["coverage/**"],
       "inputs": ["src/**/*.tsx", "src/**/*.ts"]
-    }
+    },
+    "lint": {}
```
codyebberson, line 24: Does `lint` depend on `build` now?

mattlong: I thought it should, but since we exclude `dist` everywhere, it shouldn't need to depend on `build`.

mattlong: FWIW, the official [eslint example](https://turbo.build/repo/docs/guides/tools/eslint) also doesn't have it depending on anything

examples/foomedical/tsconfig.json
mattlong: Fixed

turbo.json
codyebberson: Hm.  We exclude `dist` from list of files to be linted, but some ESLint rules depend on `.d.ts` files.  

In particular, the `no-floating-promises` rule might depend on build output.

turbo.json
```diff
@@ -20,6 +20,9 @@
       "dependsOn": ["build"],
       "outputs": ["coverage/**"],
       "inputs": ["src/**/*.tsx", "src/**/*.ts"]
+    },
+    "lint": {
+      "dependsOn": ["build"]
```
codyebberson, line 25: 👍

codyebberson (approved):
let's go

PR #5101, opened 2024-08-16
Enhance reindex job for full database reindex

### LINKED ISSUE #5097: Polish reindex job for full DB reindex
- Enable the reindex job to automatically run over all resource types
- Better error handling and reporting
- Restart or cancel reindex jobs

mattwiller:
Resolves #5097


packages/server/src/workers/reindex.ts
```diff
@@ -219,24 +236,34 @@ function formatResults(results: ReindexJobData['results']): Parameters {
     resourceType: 'Parameters',
     parameter: Object.keys(results).map((resourceType) => {
       const result = results[resourceType];
-      if (typeof result === 'string') {
+      if ('err' in result && result.err) {
```
codyebberson, line 239: is the `'err' in result` check necessary?  seems redundant?  might be necessary to keep TS happy?

packages/server/src/workers/reindex.ts
```diff
@@ -174,39 +179,51 @@ async function processPage(searchRequest: SearchRequest, job: Job<ReindexJobData
...
 
-async function updateStatus(job: Job<ReindexJobData>): Promise<void> {
+function searchRequestForNextPage(job: Job<ReindexJobData>): SearchRequest {
+  const { resourceTypes, currentTimestamp, endTimestamp, searchFilter } = job.data;
+  const resourceType = resourceTypes[0];
+  const searchRequest: SearchRequest = {
+    resourceType,
+    count: batchSize,
+    sortRules: [{ code: '_lastUpdated', descending: false }],
+    filters: [
```
codyebberson, line 197: For consideration:  With the new `_cursor` feature, it will automatically handle this filtering and pagination.  It would have the added benefit of being more robust with resources with identical `meta.lastUpdated`, and would avoid double indexing the resources at page boundaries.

It should be a matter of replacing `currentTimestamp` with `cursor`.

See the bulk data PR for a reference: https://github.com/medplum/medplum/pull/5102

packages/server/src/workers/reindex.ts
```diff
@@ -219,24 +236,34 @@ function formatResults(results: ReindexJobData['results']): Parameters {
...
       const result = results[resourceType];
-      if (typeof result === 'string') {
+      if ('err' in result && result.err) {
+        // Resource types that encountered an error report the error,
+        // and a timestamp to allow restarting from previous checkpoint
+        return {
+          name: 'result',
+          part: [
+            { name: 'resourceType', valueCode: resourceType },
+            { name: 'error', valueString: result.err.message },
```
codyebberson, line 246: Consider using `normalizeErrorString(result.err)`

packages/server/src/workers/reindex.ts
```diff
@@ -115,56 +140,36 @@ export async function execReindexJob(job: Job<ReindexJobData>): Promise<void> {
...
+        newCount += resources.length;
+      }
 
-    // Enqueue job to handle next page of the current resource type
-    await addReindexJobData({
-      ...job.data,
-      currentTimestamp: nextTimestamp,
-      count: newCount,
+      if (bundle.link?.some((link) => link.relation === 'next')) {
+        nextTimestamp = bundle.entry?.[bundle.entry?.length - 1].resource?.meta?.lastUpdated;
```
codyebberson, line 164: If we switch to `cursor`, you could instead use `nextCursor = parseSearchRequest(url).cursor`

packages/server/src/fhir/repo.ts
```diff
@@ -895,7 +895,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
    * @param conn - Database client to use for reindex operations.
    * @param resources - The resource(s) to reindex.
    */
-  async reindexResources<T extends Resource>(conn: PoolClient, ...resources: T[]): Promise<void> {
+  async reindexResources<T extends Resource>(conn: PoolClient, resources: T[]): Promise<void> {
```
codyebberson, line 898: 👍

codyebberson (approved):
I added some comments on using `_cursor` rather than `_lastUpdated`.

We may want to hold off on that, as `_cursor` is relatively new, and the whole point of this PR is stability.  Up to you

packages/server/src/workers/reindex.ts
mattwiller: Yeah, this was to placate tsc

PR #5105, opened 2024-08-16
Use trigram index for Coding display text

mattwiller:
Using trigrams over the `tsvector` index allows for matching in the middle of words, and gives an easy way to order on similarity; this is expected to increase the quality of matches when using the `filter` parameter in `ValueSet/$expand`


packages/server/src/fhir/operations/expand.ts
```diff
@@ -410,7 +410,13 @@ export function expansionQuery(
 
 function addExpansionFilters(query: SelectQuery, params: ValueSetExpandParameters): SelectQuery {
   if (params.filter) {
-    query.where('display', '!=', null).where('display', 'TSVECTOR_ENGLISH', filterToTsvectorQuery(params.filter));
+    query
+      .whereExpr(
+        new Conjunction(params.filter.split(/\s+/g).map((filter) => new Condition('display', 'LIKE', `%${filter}%`)))
```
codyebberson, line 415: What happens if `params.filter` includes a `%` ?

Are there any other special characters that need to be handled?

codyebberson (approved):
💥

PR #5112, opened 2024-08-19
config(server): Add PG pool max connection config.

charlesbthomas:


packages/server/src/config.ts
```diff
@@ -112,6 +112,8 @@ export interface MedplumDatabaseConfig {
   ssl?: MedplumDatabaseSslConfig;
   queryTimeout?: number;
   runMigrations?: boolean;
+  maxThreads?: number;
```
codyebberson, line 115: How would you feel about `maxClients` or `maxConnections`?  

Sorry to be picky about this, but config setting names live forever, and naming things is hard 🙃

packages/server/src/database.ts
```diff
@@ -47,7 +47,8 @@ async function initPool(config: MedplumDatabaseConfig, proxyEndpoint: string | u
     user: config.username,
     password: config.password,
     ssl: config.ssl,
-    max: 100,
+    max: config.maxThreads ?? 100,
+    min: config.minThreads,
```
codyebberson, line 51: Hm, this is interesting.

`min` is not in the docs: https://node-postgres.com/apis/pool

`min` is not in the pool source: https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/index.js

But `min` is in the types: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg/index.d.ts#L46

I don't have any objections to `min`, I just want to make sure we understand what it does.

packages/server/src/config.ts
charlesbthomas: I am good with maxConnections!

packages/server/src/database.ts
charlesbthomas: Wow, that's kind of shocking, min connections is a pretty common concept for connection thread pools IME. But I am totally good to exclude it from this PR; generally speaking max connections is more critical.

packages/server/src/config.ts
charlesbthomas: Done!

packages/server/src/database.ts
charlesbthomas: Removed!

codyebberson (approved):

packages/server/src/database.ts
codyebberson: Agree that it is a common feature of connection pools.

The `pg` library has been good to us, but it is showing signs of neglect and decay.

We have run some internal experiments with [`postgres.js`](https://github.com/porsager/postgres), which is modernized, has better Postgres pipelining support, and is generally much faster.

image

(lower is better)

Source: https://github.com/porsager/postgres-benchmarks

It is not as widely used though, and is not as actively maintained as we would like.
image

https://npmtrends.com/pg-vs-postgres

PR #5121, opened 2024-08-20
Enable Projects to set default profiles for resources

mattwiller:


packages/definitions/dist/fhir/r4/fhir.schema.json
```diff
@@ -59399,6 +59399,13 @@
             "$ref": "#/definitions/ProjectLink"
           },
           "type": "array"
+        },
+        "defaultProfile": {
+          "description": "Default profiles to apply to resources in this project that do not individually specify profiles",
+          "items": {
+            "$ref": "#/definitions/ProjectDefaultProfile"
```
codyebberson, line 59406: Hm, these appears to be an invalid reference.  I'm surprised that the generator tool didn't create a `ProjectDefaultProfile` type :think

codyebberson (approved):

PR #5123, opened 2024-08-21
[Scheduling Sample] Adjust appointment cancellation flow

pamella:
## Description

- Adjust appointment cancellation flow
  - The Schedule page does not show cancelled appointments to prevent them from overlapping free slots during rendering. That's due to the calendar lib.
  - New bot to handle the cancellation behavior

## Steps to test

- [ ] Build the bots: `npm run build:bots`
- [ ] Upload core data
- [ ] Upload example bots
- [ ] Cancel an appointment
- [ ] Check if it no longer show in the Schedule page
- [ ] Check if an (new) available slot show in its place in the Schedule page

[Screencast from 21-08-2024 15:16:31.webm](https://github.com/user-attachments/assets/e115b3ef-ede3-4e0b-8ab1-f6ede1a48388)


examples/medplum-scheduling-demo/src/bots/core/cancel-appointment.ts
```diff
@@ -0,0 +1,30 @@
+import { BotEvent, MedplumClient, resolveId } from '@medplum/core';
+import { Appointment, Slot } from '@medplum/fhirtypes';
+
+export async function handler(medplum: MedplumClient, event: BotEvent<Appointment>): Promise<Appointment> {
```
pamella, line 4: The bot implements a simplified logic. Let me know if that's enough for the demo app.

rahul1: One small change: Can you implement both updates inside a `Bundle` with type transaction. We'll very soon have better transactional semantics, meaning that the writes will succeed or fail together

rahul1 (approved):

examples/medplum-scheduling-demo/src/bots/core/cancel-appointment.ts
pamella: Sure, done

Issue #5124, opened 2024-08-21
User Defined Compartments


rahul1:
Right now, the `meta.account` field functions as a user defined compartment, and is propagated from patients to linked resources.

As this can be a reference to anything, users often make use of this to field to define compartments for "tenants" within their system.

Currently, there is a limitation that a Patient can only be part of a single user-defined compartment via meta.account.

## Proposal
Allow users to write directly to `meta.compartments`, to specify up to `k` user-defined compartments

The server would then be responsible for making sure that the `Project` and `Patient` compartments are always present if applicable.

PR #5125, opened 2024-08-21
Allow adding to Meta.compartment

mattwiller:


rahul1:
Fixes: https://github.com/medplum/medplum/issues/5124

packages/server/src/fhir/repo.ts
```diff
@@ -1282,24 +1282,38 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
...
       if (patientId && validator.isUUID(patientId)) {
-        result.push(patient);
+        compartments.add(patient.reference);
       }
     }
 
-    return result;
+    // Carry forward anything added to the resource compartments array
+    if (resource.meta?.compartment?.length) {
+      for (const compartment of resource.meta.compartment) {
```
mattlong, line 1304: We had discussed having a max compartments length. What should the behavior be if the user specifies more than that max?

codyebberson: Good point.  I'd vote for HTTP 400 if greater than N, where N should probably be a server config setting with a default of ~10.

codyebberson: > What should the behavior be if the user specifies more than that max?

It would be a performance hit, loading up the `compartments` array.  Not great, but not the end of the world.

reshmakh:
@rahul1 @ksmith94 this is high priority for documentation.

codyebberson (approved):

codyebberson:
🎉

PR #5136, opened 2024-08-26
Dependency upgrades 2024-08-26

medplumbot:
Dependency upgrades


codyebberson:
Mantine `<Badge>` regression still open: https://github.com/mantinedev/mantine/issues/6698

No major version upgrades in this PR, and no significant upgrades.  Going to skip this week, and hope the Mantine regression fix is released by then.

PR #5139, opened 2024-08-26
Relax new transaction inside low-level repo operations

### LINKED ISSUE #5138: Remove nested transactions for Conditional Update
## Background
Currently a conditional update creates a nest of two transactions: 
- the first transaction starts before the search
- the second transaction happens at the time of write to database

This hampers write performance by slowing down the performance, and increases the size of the WAL, which can be a constraint during large data ingests (e.g. large bundles of Conditional Update operations)

We saw this issue in real world perf testing with a customer. Multiple "savepoint" operations indicated that nested transactions were running

## Proposal
Logically, this operation only need a single transaction. At the time of the write to database. We only need to confirm that we are within a transaction. We do not need to start a new one.

This ticket proposes modifying the second transaction creation to be become a check for transactionality.

mattwiller:
Resolves #5138

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -808,7 +808,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
    * @param create - If true, then the resource is being created.
    */
   private async writeToDatabase<T extends Resource>(resource: T, create: boolean): Promise<void> {
-    await this.withTransaction(async (client) => {
+    await this.ensureInTransaction(async (client) => {
       await this.writeResource(client, resource);
       await this.writeResourceVersion(client, resource);
       await this.writeLookupTables(client, resource, create);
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -808,7 +808,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
    * @param create - If true, then the resource is being created.
    */
   private async writeToDatabase<T extends Resource>(resource: T, create: boolean): Promise<void> {
-    await this.withTransaction(async (client) => {
+    await this.ensureInTransaction(async (client) => {
```
codyebberson, line 811: 👍

packages/server/src/fhir/repo.ts
```diff
@@ -966,7 +966,7 @@ export class Repository extends BaseRepository implements FhirRepository<PoolCli
 
       await deleteCacheEntry(resourceType, id);
 
-      await this.withTransaction(async (conn) => {
+      await this.ensureInTransaction(async (conn) => {
```
codyebberson, line 969: 👍

codyebberson (approved):
Nice, super clean

https://www.youtube.com/watch?v=-JfEJq56IwI

Issue #1241, opened 2022-12-09
Mock client doesn't handle `system|value` search syntax


reshmakh:
This returns undefined:

replicaClient.searchOne(
      'ServiceRequest',
      `identifier=customer/productionUUID|${id}`
    );

However, this returns the appropriate value: 
replicaClient.searchOne(
      'ServiceRequest',
      `identifier=${id}`
    );

codyebberson:
See https://github.com/medplum/medplum/blob/main/packages/core/src/search/match.ts

PR #5148, opened 2024-08-27
Improve matching on identifier and codeablceconcept token searches

mattlong:
Updated the logic based on https://hl7.org/fhir/R4/search.html#token


codyebberson (approved):

mattlong:
Fixes #1241

PR #5164, opened 2024-09-02
Fix parsing of a profile with 2 subsequent sliced properties

deklungel:
Greetings,

We're running into an issue while parsing the BEPractitionerRole profile (https://www.ehealth.fgov.be/standards/fhir/core/StructureDefinition-be-practitionerrole.html).

The error message is `Invalid slice start before discriminator: SNOMED-CT (PractitionerRole.specialty:SNOMED-CT)`.

Diving deeper into packages/core/src/typeschema/types.ts where the error is thrown, it seems that the slicingContext of the previous property 'PractitionerRole.code' is only closed after handling the first 'PractitionerRole.specialty' element (in checkFieldExit). Because of this, no new slicingContext is opened for 'PractitionerRole.specialty' in checkFieldEnter.

I believe following changes fix this issue.

Kind regards

### DIFF
--- a/packages/core/src/typeschema/__test__/be-practitionerrole.json
+++ b/packages/core/src/typeschema/__test__/be-practitionerrole.json
@@ -1,0 +1,2930 @@
+{
+  "resourceType" : "StructureDefinition",
+  "id" : "be-practitionerrole",
+  "text" : {
+    "status" : "extensions",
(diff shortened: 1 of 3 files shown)


codyebberson:
Hi @deklungel - thanks for submitting.  The fix looks good to me.  Would it be possible to create a more surgical test case / smaller `StructureDefinition` example?  The Medplum team can assist with that if it's ok with you.

deklungel:
Hey @codyebberson , I'll make an attempt, but honestly, I have no clue what the rules are for a StructureDefinition. Feel free to make changes :)

deklungel:
I've removed many elements that I believe are not relevant for the test case. 
Feel free to update.

Kind regards

mattwiller (approved):

PR #5189, opened 2024-09-05
Include the URL for OAUTH on localhost.

mgolden:
The instructions indicate how to get an OAUTH token on the medplum server, but not self-hosted on localhost.


codyebberson (approved):

codyebberson:
Thanks @mgolden !

PR #5191, opened 2024-09-06
Use default max search results for _(rev)include circuit breaker

nate-watkins:
Hey team- ran into this in our self-hosted instance. We had increased [that value](https://github.com/medplum/medplum/blob/cd10c14c048b16077f92c7081f8df5ca34c1e524/packages/core/src/search/search.ts#L8) for our use case but hit this on a request using `_include`. Is this circuit breaker fundamentally different from the max search results?


mattwiller (approved):
It makes sense to me that these would be the same limit 👍

PR #5193, opened 2024-09-06
curl command is incorrect

mgolden:
Should be -X not -x


mattwiller (approved):
Great catch!  Thanks @mgolden

PR #5194, opened 2024-09-06
tweak(core): align `.isLoading()` to what was previously perceived in React

ThatOneBro:
`MedplumClient.isLoading()` is currently used in examples to prevent render before being authed, but with our new events that keep the `MedplumProvider` more in sync with `MedplumClient`, we read `.isLoading()` more frequently, and in many cases it is now evaluating to true, ie. when a profile is just being refreshed. This causes unmounting in many cases and is probably not the behavior that most users would expect, and in many cases it is an unintentional downstream breaking change.

This PR would bring `isLoading()` in line with the behavior previously observed from the perspective of users of the `useMedplum` hook. It is technically a breaking change from the previous explicit contract, though it fixes a "regression" from the perspective of React.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -2771,12 +2771,12 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
   }
 
   /**
-   * Returns true if the client is waiting for authentication.
-   * @returns True if the client is waiting for authentication.
+   * Returns true if the client is waiting for initial authentication.
+   * @returns True if the client is waiting for initial authentication.
    * @category Authentication
    */
   isLoading(): boolean {
-    return !this.isInitialized || !!this.profilePromise;
+    return !this.isInitialized || (!!this.profilePromise && !this.sessionDetails?.profile);
   }
 
   /**


packages/core/src/client.ts
```diff
@@ -2771,12 +2771,12 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
   /**
-   * Returns true if the client is waiting for authentication.
-   * @returns True if the client is waiting for authentication.
+   * Returns true if the client is waiting for initial authentication.
+   * @returns True if the client is waiting for initial authentication.
    * @category Authentication
    */
   isLoading(): boolean {
-    return !this.isInitialized || !!this.profilePromise;
+    return !this.isInitialized || (!!this.profilePromise && !this.sessionDetails?.profile);
```
mattwiller, line 2779: The `!!` mixed in with the `!`s threw me a bit when scanning this line — minor suggestion to clarify with e.g.:
```suggestion
    return !this.isInitialized || (Boolean(this.profilePromise) && !this.sessionDetails?.profile);
```

mattwiller (approved):

codyebberson (approved):

codyebberson (approved):
| isInitialized | profilePromise | profile | result |
| --- | --- | --- | --- |
| `false` | `false` | `false` | `true` - not initialized yet |
| `true` | `false` | `false` | `false` - initialized and not loading profile |
| `false` | `true` | `false` | `true` - not initialized yet |
| `true` | `true` | `false` | `true` - new case, loading profile, but no profile yet |
| `false` | `false` | `true` | `true` - not initialized yet |
| `true` | `false` | `true` | `false` - initialized and not loading profile |
| `false` | `true` | `true` | `true` - not initialized yet |
| `true` | `true` | `true` | `false` - new case, refreshing profile, have existing profile |

ThatOneBro:
I'm adding a test for this really quickly, just to prove that it's achieving the intended result from the perspective of React 😉

PR #5203, opened 2024-09-09
Update graphiql introspection url default

codyebberson:
None of these changes should be required, so this is more of an experiment.

The `MEDPLUM_INTROSPECTION_URL` value should come from Github Actions Secrets, which is defined:

image

The `MEDPLUM_INTROSPECTION_URL` environment variable should be passed along:
* `deploy.yml`: https://github.com/medplum/medplum/blob/main/.github/workflows/deploy.yml#L65
* `cicd-deploy.sh`: uses `source deploy-graphiql.sh`, which should pass env vars
* `deploy-graphiql.sh`

### DIFF
--- a/scripts/deploy-graphiql.sh
+++ b/scripts/deploy-graphiql.sh
@@ -2,17 +2,11 @@
 
 pushd packages/graphiql
 
-# No cache
-
-aws s3 cp dist/ s3://graphiql.medplum.com/ \
-  --region us-east-1 \
-  --recursive \
-  --content-type "text/html" \
-  --cache-control "no-cache" \
-  --exclude "*" \
-  --include "*.html"
-
-# Cache forever
+# First deploy hashed files that are cached forever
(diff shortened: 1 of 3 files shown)


scripts/deploy-graphiql.sh
```diff
@@ -45,4 +39,17 @@ aws s3 cp dist/ s3://graphiql.medplum.com/ \
   --exclude "*" \
   --include "*.txt"
 
+# Now deploy named files that are not cached.
+# These are small lightweight files that are not hashed.
+# It is important to deploy these files last,
+# because they reference the previously uploaded hashed files.
```
mattwiller, line 45: 🙏

ThatOneBro: Is the rhyming of the first 3 lines here intentional? Felt like I was reading a clever rhyme written by ChatGPT or something

ThatOneBro (approved):

PR #5204, opened 2024-09-09
Update server shutdown order

mattwiller:
Attempting to align with the [order from server startup](https://github.com/medplum/medplum/blob/01d0e96a930ea3cab55c00bb424e9957a6a84210/packages/server/src/app.ts#L135-L137), applying it in reverse

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -204,6 +204,7 @@ export function initAppServices(config: MedplumServerConfig): Promise<void> {
 }
 
 export async function shutdownApp(): Promise<void> {
+  cleanupHeartbeat();
   await closeWebSockets();
   if (server) {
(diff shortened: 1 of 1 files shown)


packages/server/src/app.ts
```diff
@@ -204,20 +204,20 @@ export function initAppServices(config: MedplumServerConfig): Promise<void> {
...
-  closeRateLimiter();
-
   if (server) {
     await new Promise((resolve) => {
       (server as http.Server).close(resolve);
     });
     server = undefined;
   }
 
+  cleanupHeartbeat();
```
ThatOneBro, line 215: I think this one might need to still be done before `closeWebSockets()` since the heartbeat action is usually dependent on websockets being open.

codyebberson (approved):

PR #5210, opened 2024-09-10
[Photon Health] Sync formulary to Photon catalog

ksmith94:
This adds a page to the demo to view and manage your formulary. Additionally, it provides functionality through a bot to sync your formulary to your project's catalog in Photon Health


examples/medplum-photon-integration/data/example-data.json
rahul1, line 1: So things have changed on the Medplum Server, and 'transactions` are now much heavier weight operation in Medplum. These can all be Bundles of type `batch`

rahul1 (approved):

PR #5211, opened 2024-09-10
System bots and executing bots from linked projects

codyebberson:


packages/server/src/fhir/operations/execute.ts
```diff
@@ -487,11 +487,44 @@ async function getBotAccessToken(runAs: ProjectMembership): Promise<string> {
...
 
-async function getBotSecrets(bot: Bot): Promise<Record<string, ProjectSetting>> {
+/**
+ * Returns a collection of secrets for the bot.
+ *
+ * Secrets can come from 1-4 different sources:
+ * 1. Bot project secrets
+ * 2. Bot project system secrets (if bot.system is true)
+ * 3. RunAs project secrets (if running in a different linked project)
+ * 4. RunAs project system secrets (if bot.system is true and running in a different linked project)
```
rahul1, line 497: We should codify the precedence order for these secrets, if they conflict in name. This can happen, for example, for customers who have their own `HEALTH_GORILLA_CLIENT_ID`, and are not using the system version. 

I think the principal of "most specific beats more general" should apply here: 


I think the order above works, but we should make sure that we note that this is an intentional order

codyebberson: This makes sense.

"Run as project" beats "bot project" - makes sense.

I don't have any intuition on "secret" vs "system secret".  I suppose you could say that project admins can set "secrets", which would override "system secrets" set by super admins?

In that case, the official order would be:

1. Bot project system secrets
2. Bot project secrets
3. RunAs project system secrets
4. RunAs project secrets

?

rahul1 (approved):

ThatOneBro (approved):

PR #5213, opened 2024-09-10
Correct handling of multiple query parameters

mattwiller:
- Fixes `FhirRequest` interface type to have a more correct `Record<string, string | string[]>` type
- Unifies parsing of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) in the router for both regular and batch processing, to handle multiple instances of the same search parameter
- Cleans up usages of `FhirRequest.query` in server to honor the new type


packages/server/src/fhir/routes.ts
```diff
@@ -302,9 +302,9 @@ protectedRoutes.use(
 
     const request: FhirRequest = {
       method: req.method as HttpMethod,
-      pathname: req.originalUrl.replace('/fhir/R4', '').split('?').shift() as string,
+      pathname: req.originalUrl.replace('/fhir/R4', ''),
```
codyebberson, line 305: Hmm.  This is unfortunate.  `pathname` is no longer "pathname" 🤔

packages/fhir-router/src/batch.ts
```diff
@@ -145,16 +145,13 @@ class BatchProcessor {
...
-
     const result = await this.router.handleRequest(
       {
         method: request.method as HttpMethod,
-        pathname: url.pathname,
+        pathname: request.url,
+        // Params and query string are parsed from the `pathname` by the router
         params: Object.create(null),
-        query: Object.fromEntries(url.searchParams),
+        query: Object.create(null),
```
codyebberson, line 154: I see what you're going for here.  I feel like this is now more loose and ambiguous than before though 😞 

How nasty would it be to remove `pathname` and `query`, and replace them with `url` for consistency?

Or is the idea that `FhirRequest` is now a 2-phase object?  There's the input value, where `pathname === url`, and then there's the pre-processed value that gets passed to the handlers, where `query` is populated?

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -265,17 +263,43 @@ export class FhirRouter extends EventTarget {
...
     if (!result) {
       return [notFound];
     }
-    const { handler, params } = result;
+    const { handler, params, query } = result;
+
+    // Populate request object with parsed URL components from router
     req.params = params;
+    if (query) {
+      req.pathname = url.slice(0, url.indexOf('?'));
```
mattlong, line 298: Could the logic of separating `pathname` from the query string be centalized? Similar versions of splitting `url` on  `url.indexOf('?')` happen both here and in `find`.

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -265,17 +263,43 @@ export class FhirRouter extends EventTarget {
...
+    }
+    return result;
   }
 
   async handleRequest(req: FhirRequest, repo: FhirRepository): Promise<FhirResponse> {
-    const result = this.router.find(req.method, req.pathname);
+    const url = req.url;
+    // User-specified URL is parsed into component parts, which are populated back onto the request
+    if (req.pathname) {
+      throw new OperationOutcomeError(badRequest('FhirRequest must specify url instead of pathname'));
```
mattlong, line 287: Would rejecting requests with `pathname` specified cause backwards compatibility issues? e.g. if someone were using `fhirrouter` directly without `@medplum/server`

packages/server/src/fhir/operations/agentupgrade.ts
```diff
@@ -40,15 +40,17 @@ export async function agentUpgradeHandler(req: FhirRequest): Promise<FhirRespons
 
   let timeout: number | undefined;
   if (_timeout) {
-    timeout = Number.parseInt(_timeout, 10);
+    timeout = Number.parseInt(singularize(_timeout) as string, 10);
```
mattlong, line 43: `singularlize` is "first wins" in the case of an array whereas the old query parsing logic was "last wins". Any concerns about that?

packages/fhir-router/src/fhirrouter.ts
mattwiller: Possibly, but I think we may want to offer different compatibility guarantees on `fhir-router` as compared to `server`, `core`, and `mock`

packages/server/src/fhir/operations/agentupgrade.ts
mattwiller: True — I think in this case that's probably okay, though I'd also be open to changing `singularize` if we wanted to preserve the old behavior exactly

packages/fhir-router/src/batch.ts
```diff
@@ -18,7 +18,7 @@ import { IncomingHttpHeaders } from 'node:http';
 const maxUpdates = 50;
 const maxSerializableTransactionEntries = 8;
 
-const localBundleReference = /urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
+const localBundleReference = /urn(:|%3A)uuid(:|%3A)[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
```
codyebberson, line 21: Is this actually valid?  Or is this a backwards compatibility thing?

codyebberson (approved):
Nice, ty 🙏

packages/fhir-router/src/batch.ts
mattwiller: I believe this isn't explicitly specified between FHIR and HTTP specifications, but I gathered a few data points:
- The FHIR spec doesn't mention anything in either the [Batch HTTP](https://www.hl7.org/fhir/http.html#transaction) section or the [description of `Bundle.entry.request.url`]((http://hl7.org/fhir/R4/bundle-definitions.html#Bundle.entry.request.url)
- The [`uri` data type](http://hl7.org/fhir/R4/datatypes.html#uri) is specified to contain "A Uniform Resource Identifier Reference ([RFC 3986](http://tools.ietf.org/html/rfc3986))"
- The aforementioned RFC says about [reserved characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) that:
   > URI producing applications should percent-encode data octets that
   correspond to characters in the reserved set *unless these characters
   are specifically allowed by the URI scheme to represent data in that
   component*.  If a reserved character is found in a URI component and
   no delimiting role is known for that character, then it must be
   interpreted as representing the data octet corresponding to that
   character's encoding in US-ASCII. [emphasis mine]
- Specifically for [query params](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4), the syntax is given as `query = *( pchar / "/" / "?" )`, which expands to the following set of characters: `ALPHA`, `DIGIT`, `-`, `.`, `_`, `~`, `!`, `$`, `&`, `'`, `(`, `)`, `*`, `+`, `,`, `;`, `=`, `:`, `@`, `/`, and `?`.  These characters do not need to be percent-encoded in the query string, but there's nothing saying they can't be

PR #5219, opened 2024-09-11
Add default database settings

mattwiller:
Added core database settings to CDK and local development config, in preparation for moving these settings [from connection time](https://github.com/medplum/medplum/blob/e7a15721dfd362b127a948b05389106629121a14/packages/server/src/database.ts#L66-L71) in the next major version


codyebberson (approved):
Nice, lgtm 👍 

The one thing we still need to figure out is how to communicate this to self-hosters.  That's a good topic for offline discussion.

PR #5223, opened 2024-09-12
feat(react): extend `AsyncAutocomplete` to allow pill customization

dillonstreator:
This PR allows the pill component to be customized. The use case I have is to display a condensed version of the label in the pill with a tooltip on hover.


ThatOneBro (approved):
Nice, thank you Dillon 👍

PR #5229, opened 2024-09-12
Remove ID condition on cursor search

mattwiller:
The dual sort on `lastUpdated, id` was causing queries using the cursor to be a sequential scan in the database.  Using just `lastUpdated` allows the search to hit an index, resulting in a speedup of ~30x


packages/server/src/fhir/search.ts
```diff
@@ -280,10 +270,10 @@ async function getSearchEntries<T extends Resource>(
   repo: Repository,
   searchRequest: SearchRequestWithCountAndOffset<T>,
   builder: SelectQuery
-): Promise<{ entry: BundleEntry<T>[]; rowCount: number; hasMore: boolean }> {
+): Promise<{ entry: BundleEntry<T>[]; rowCount: number; nextResource?: T }> {
   const rows = await builder.execute(repo.getDatabaseClient(DatabaseMode.READER));
   const rowCount = rows.length;
-  const resources = rows.slice(0, searchRequest.count).map((row) => JSON.parse(row.content as string)) as T[];
+  const resources = rows.slice(0, searchRequest.count + 1).map((row) => JSON.parse(row.content as string)) as T[];
```
codyebberson, line 276: Why `slice()` here?  We request `+ 1` to know if there are more.  I think the `slice` is a no-op?

packages/server/src/fhir/search.ts
```diff
@@ -77,8 +77,8 @@ type SearchRequestWithCountAndOffset<T extends Resource = Resource> = SearchRequ
 
 interface Cursor {
   version: string;
-  lastInstant: string;
-  lastId: string;
+  nextInstant: string;
```
codyebberson, line 80: potayto, potahto

packages/server/src/fhir/search.ts
```diff
@@ -568,11 +562,17 @@ function getSearchLinks(
     },
   ];
 
-  if (searchRequest.count > 0 && entries && entries.length > 0) {
+  if (searchRequest.count > 0 && entries?.length) {
     if (canUseCursorLinks(searchRequest)) {
-      buildSearchLinksWithCursor(searchRequest, entries, hasMore, result);
+      console.assert(
```
codyebberson, line 567: We only want to log?  Or do we want to `throw`?

codyebberson (approved):
lgtm 👍 

See the comment about the unnecessary `slice()` call.

PR #5231, opened 2024-09-12
Add health gorilla demo app

mattlong:
Part 3 of #5205


examples/medplum-health-gorilla-demo/src/index.css
codyebberson, line 1: This file looks unused?  Can be deleted?

codyebberson (approved):
Nice, ty 👍 🙏 

Looks like there might be a couple leftover files from Vite's default project?

Otherwise lgtm 👍

PR #5256, opened 2024-09-17
fix(core): fix hard reloading other tabs on token refresh

### LINKED ISSUE #5255: Fix hard reload of other tabs when refreshing profile
Currently due to [this line](https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts#L3822) we hard reload the page of other open tabs on every token refresh. This was originally done in order to prevent being logged in with different profiles in multiple tabs at the same time for security reasons. 

This reasoning makes sense, but we should not reload whenever the profile is the same as before the storage event was fired.

ThatOneBro:
Fixes #5255

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -2385,13 +2385,61 @@ describe('Client', () => {
     callback({ key: 'randomKey' });
     expect(mockReload).not.toHaveBeenCalled();
 
+    const practitioner1 = randomUUID();
+    const practitioner2 = randomUUID();
+
+    // Should NOT refresh when we have the same profile
     mockReload.mockReset();
-    callback({ key: 'activeLogin' });
+    callback({
+      key: 'activeLogin',
+      oldValue: JSON.stringify({ profile: { resourceType: 'Practitioner', id: practitioner1 } }),
+      newValue: JSON.stringify({ profile: { resourceType: 'Practitioner', id: practitioner1 } }),
+    } as StorageEvent);
+    expect(mockReload).not.toHaveBeenCalled();
+
+    // Should refresh when we change to a new profile
+    mockReload.mockReset();
+    callback({
+      key: 'activeLogin',
+      oldValue: JSON.stringify({ profile: { resourceType: 'Practitioner', id: practitioner1 } }),
+      newValue: JSON.stringify({ profile: { resourceType: 'Practitioner', id: practitioner2 } }),
+    } as StorageEvent);
     expect(mockReload).toHaveBeenCalled();
+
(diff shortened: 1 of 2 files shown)


packages/core/src/client.test.ts
```diff
@@ -2385,13 +2385,61 @@ describe('Client', () => {
...
+      key: 'activeLogin',
+      oldValue: "{ profile: 'bla' ", // Invalid JSON
+      newValue: JSON.stringify({ profile: { resourceType: 'Practitioner', id: null } }),
+    } as StorageEvent);
+    expect(mockReload).not.toHaveBeenCalled();
+    expect(console.error).toHaveBeenCalled();
+    console.error = originalConsoleError;
+
+    mockReload.mockReset();
+    callback({ key: null });
```
ThatOneBro, line 2441: @codyebberson is there some specific reason why changing the behavior at this line might be bad? Sending `null` currently will no longer trigger the reload. I can change the logic if we still need it to trigger it, I just wasn't sure when `null` might be passed as the key to a storage event in our case

ThatOneBro: Talked offline: we want it to trigger reload since `null` key means storage is being cleared.

codyebberson: As discussed offline, `key: null` is the event for `localStorage.clear()`, and yeah, I think we want to preserve that case.

packages/core/src/client.ts
```diff
@@ -3820,11 +3820,21 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
-          // Storage events fire when different tabs make changes.
-          // On storage clear (key === null) or activeLogin change (key === 'activeLogin')
-          // Refresh the page to ensure the active login is up to date.
+        // Storage events fire when different tabs make changes.
+        // On storage clear (key === null) or profile change (key === 'activeLogin', and profile in 'activeLogin' is different)
+        // Refresh the page to ensure the active login is up to date.
+        if (e.key === null) {
           window.location.reload();
+        } else if (e.key === 'activeLogin') {
+          try {
```
codyebberson, line 3829: Do we need this `try`/`catch`? should only happen on malformed JSON (which generally shouldn't happen).  I think (?) that will result in same behavior, an unhandled error in the console.

codyebberson (approved):
One comment about whether we need the new `try`/`catch`.  Your call.  Otherwise lgtm 👍

packages/core/src/client.ts
ThatOneBro: Good point, I removed it 👍

Issue #3668, opened 2024-01-05
Possibility to add more than one sub-compartment


nikzen:
Hi and thanks again for your great work!

I am evaluating to build some "shared task list" usecase and i am trying to figure out the correct concept around access policies as this is crucial.

In our usecase, we have multiple organizations with different users inside a project. The idea is to allow to share a task between organizations. Reading through the docs i thought that combining compartment & parameterized access policies might work.

**Access Policy**
```
{
  "resourceType": "AccessPolicy",
  "name": "Practitioner Access Policy",
  "resource": [
    {
      "resourceType": "Task",
      "criteria": "Task?_compartment=%organization"
    }
  ]
}
```

**ProjectMembership User of Organization 1234**

```
{
  "resourceType": "ProjectMembership",
  "profile": {
    "reference": "Practitioner/1234",
    "display": "Name"
  },
  "admin": false,
  "access": [
    {
      "parameter": [
        {
          "name": "organization",
          "valueReference": {
            "reference": "Organization/1234",
            "display": "Organization 1234"
          }
        }
      ],
      "policy": {
        "reference": "AccessPolicy/uuid",
        "display": "Practitioner Access Policy"
      }
    }
  ]
}
```

**ProjectMembership User of Organization 5678**

```
{
  "resourceType": "ProjectMembership",
  "profile": {
    "reference": "Practitioner/5678",
    "display": "Name"
  },
  "admin": false,
  "access": [
    {
      "parameter": [
        {
          "name": "organization",
          "valueReference": {
            "reference": "Organization/5678",
            "display": "Organization 5678"
          }
        }
      ],
      "policy": {
        "reference": "AccessPolicy/uuid",
        "display": "Practitioner Access Policy"
      }
    }
  ]
}
```

### Usecase
The user of organization a creates a task and there is a bot which adds the compartment of `organization 1234` using the `meta.account` field:

```
{
  "resourceType": "Task",
  "status": "accepted",
  "intent": "order",
  "code": {
    "coding": [
      {
        "code": "This is a new task",
        "display": "This is a new task"
      }
    ]
  },
  "meta": {
    "compartment": [
      {
        "reference": "Project/42093f51-8d08-4cd1-a7af-a2b92596ecee"
      },
      {
        "reference": "Practitioner/43aca9fd-b813-414c-a07b-c89711c05407",
        "display": "Practitoner"
      }
    ],
    "account": {
      "reference": "Organization/1234",
      "display": "Organization 1234"
    }
  }
[...]

PR #5264, opened 2024-09-18
Allow setting multiple accounts per resource

mattwiller:
This PR tackles the first stage of #3668 by adding a new `Meta.accounts` array property


packages/core/src/base-schema.json
```diff
@@ -2194,6 +2194,14 @@
           }
         ]
       },
+      "accounts": {
+        "max": 9007199254740991,
```
codyebberson, line 2198: Let's hope nobody tries to use 9,007,199,254,740,991 accounts 😱

packages/server/src/fhir/repo.ts
```diff
@@ -601,9 +602,10 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     if (project) {
       resultMeta.project = project;
     }
-    const account = await this.getAccount(existing, updated, create);
-    if (account) {
-      resultMeta.account = account;
+    const accounts = await this.getAccounts(existing, updated, create);
+    if (accounts) {
+      resultMeta.account = accounts[0];
```
codyebberson, line 607: 🤔 

This is interesting.  I'm ok with it, but I also wonder if `accounts.length > 1` then we know they're using the new feature, and we could avoid setting legacy `account`?

Minor idle thought, definitely non-blocking

packages/server/src/fhir/repo.ts
```diff
@@ -1688,41 +1697,52 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    if (updatedAccounts && this.canWriteAccount()) {
       // If the user specifies an account, allow it if they have permission.
-      return account;
+      return updatedAccounts;
     }
 
     if (create && this.context.accessPolicy?.compartment) {
       // If the user access policy specifies a compartment, then use it as the account.
-      return this.context.accessPolicy.compartment;
+      return [this.context.accessPolicy.compartment];
```
codyebberson, line 1713: This leads to an interesting question about whether `accessPolicy.compartment` should also be an array.

Deferring that to future discussions about `AccessPolicy` v2

packages/server/src/fhir/repo.ts
```diff
@@ -1688,41 +1697,52 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
-        // If the resource is in a patient compartment, then lookup the patient.
         try {
-          const systemRepo = getSystemRepo();
-          const patient = await systemRepo.readReference(patientRef);
-          if (patient.meta?.account) {
-            // If the patient has an account, then use it as the resource account.
-            return patient.meta.account;
-          }
+          // If the resource is in a patient compartment, then lookup the patient.
+          const patient = await getSystemRepo().readReference(patientRef);
```
codyebberson, line 1720: Now that `accounts` is an array, should this be the union of all patients' accounts?

For example, imagine a `Group` resource, with multiple `Patient` members.  At present, the `Group` would only inherit the account of the first patient.  Should that be the set of all accounts for all patients?

attn @rahul1

packages/server/src/fhir/repo.ts
```diff
@@ -1688,41 +1697,52 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
         }
       }
     }
 
     // Otherwise, default to the existing value.
-    return existing?.meta?.account ?? updated.meta?.account;
+    return this.extractAccountReferences(existing?.meta);
+  }
+
+  private extractAccountReferences(meta: Meta | undefined): Reference[] | undefined {
```
codyebberson, line 1733: Thoughts on using a `Set` here?

It seems like it might be possible for someone to include duplicates of the same value?

packages/server/src/fhir/repo.ts
```diff
@@ -1688,41 +1697,52 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    if (!meta) {
+      return undefined;
+    }
+    if (meta.accounts && meta.account) {
```
codyebberson, line 1737: This is worth highlighting from a customer service perspective.

We're not always using both.

Is it still possible to remove an account?

codyebberson (approved):
Overall looks good to me 👍 

There are 2 general questions I have with the overall "multiple" strategy:
1. Do we allow duplicates?
2. Is it still possible to remove accounts?

There's also one very wonky question about accounts for resources with multiple patients.  That would be a net new feature, so I'm ok to punt that to future PR.

PR #5296, opened 2024-09-23
fix(docs): fix broken `auth/methods/client-credentials` page

### LINKED ISSUE #5295: https://www.medplum.com/docs/auth/methods/client-credentials is broken
Screenshot 2024-09-23 at 1 22 59 PM

ThatOneBro:
Fixes #5295

The page was broken because all children of `Tabs` have to be `TabItem`s and the free-floating text becomes a `p` element.


ianplunkett (approved):
Looks good to me.

PR #5297, opened 2024-09-24
Only delete Addresses for known ResourceType

mattlong:


packages/server/src/fhir/lookups/address.ts
```diff
@@ -85,7 +95,7 @@ export class AddressTable extends LookupTable {
    * @returns Promise on completion.
    */
   async indexResource(client: PoolClient, resource: Resource, create: boolean): Promise<void> {
-    if (!create) {
+    if (!create && AddressTable.resourceTypes.has(resource.resourceType)) {
```
codyebberson, line 98: 👍

codyebberson (approved):

PR #5302, opened 2024-09-24
Document Project settings

mattwiller:


packages/docs/docs/access/projects.md
```diff
@@ -72,14 +72,15 @@ To switch to the SuperAdmin project or check if you are already in it, you can u
...
+| Setting                      | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             | Default |
```
codyebberson, line 79: I think the subset here might be more confusion than just linking to the full page.

codyebberson (approved):
Nice, ty 👍 🙏 

Minor comment about removing the subset table from `projects.md`, and just linking to `project-settings.md`.

PR #5305, opened 2024-09-25
feat(ToolsPage): add confirm on `Upgrade`

### LINKED ISSUE #5276: Add confirmation dialog to `Upgrade` button on Agent `ToolsPage`

ThatOneBro:
Closes #5276


packages/core/src/version-utils.ts
```diff
@@ -0,0 +1,106 @@
+import { normalizeErrorString } from './outcomes';
+
+export const GITHUB_RELEASES_URL = 'https://api.github.com/repos/medplum/medplum/releases';
```
codyebberson, line 3: Mental note: We should use this in `packages/cli/src/aws/utils.ts` as well.

packages/mock/src/client.ts
```diff
@@ -97,6 +98,7 @@ export interface MockClientOptions
    * Override the `MockFetchClient` used by this `MockClient`.
    */
   readonly mockFetchOverride?: MockFetchOverrideOptions;
+  readonly fetch?: FetchLike;
```
codyebberson, line 101: Can we add this to `Pick<>` on line 90 instead?  I think that should work?

packages/mock/src/client.ts
```diff
@@ -154,9 +156,11 @@ export class MockClient extends MedplumClient {
...
       ) => client.mockCreatePdf(docDefinition, tableLayouts, fonts),
-      fetch: (url: string, options: any) => {
-        return client.mockFetch(url, options);
-      },
+      fetch: clientOptions?.fetch
+        ? clientOptions.fetch
+        : (url: string, options: any) => {
+            return client.mockFetch(url, options);
+          },
     });
```
codyebberson, line 164: Super minor, but I think you should be able to use `clientOptions?.fetch ?? ...`

packages/app/src/resource/ToolsPage.test.tsx
```diff
@@ -213,60 +245,347 @@ describe('ToolsPage', () => {
...
     });
 
-    await expect(screen.findByText('Error')).resolves.toBeInTheDocument();
+    // This sleep is load bearing
+    // Basically there is some strange behavior around the Mantine Portal implementation where it is initially rendered as `null`
+    // The theory is that the above `act` is unable to track the Modal children since they are initially not rendered and therefore their useEffects
+    // Are not queued before the end of the `act` block
+    // See: https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Portal/Portal.tsx
+    await act(async () => {
+      await sleep(150);
```
codyebberson, line 573: Oof.  If it was just one `sleep()`, I might be able to swallow it, but repeating it for every test feels wrong.

Do we need to use Mantine Portal?

packages/core/src/version-utils.ts
```diff
@@ -0,0 +1,106 @@
...
+  return manifest;
+}
+
+/**
+ * Tests that a given version string follows the basic semver pattern of `<int>.<int>.<int>`, which is used for Medplum versions.
+ *
+ * @param version - A version string that should be tested for valid semver semantics.
+ * @returns `true` if `version` is a valid semver version that conforms to the Medplum versioning system, otherwise `false`.
+ */
+export function isValidMedplumSemver(version: string): boolean {
```
codyebberson, line 76: Why `isValidMedplumSemver` vs `isValidSemver`?

packages/core/src/version-utils.test.ts
```diff
@@ -0,0 +1,268 @@
...
+      tag_name: 'canary',
+      assets: [
+        {
+          name: 'medplum-agent-canary-linux',
+          browser_download_url: 'https://example.com',
+        },
+      ],
+    } as ReleaseManifest;
+    const fetchSpy = jest.spyOn(globalThis, 'fetch').mockImplementation(
+      // @ts-expect-error Not perfect match to fetch
```
codyebberson, line 253: Can we cast rather than use `@ts-expect-error`? 

`@ts-expect-error` should be used for exceptional cases, because it can hide other valid errors.

packages/core/src/version-utils.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+export function clearReleaseCache(): void {
+  releaseManifests.clear();
+}
+
+/**
+ * Asserts that a given candidate is a `ReleaseManifest`.
+ * @param candidate - An object assumed to be a `ReleaseManifest`.
+ */
+export function assertReleaseManifest(candidate: unknown): asserts candidate is ReleaseManifest {
```
codyebberson, line 20: Do we need all of this validation code?  It seems like we should be able to get the values we need, and either they are present or not?

packages/app/src/resource/ToolsPage.test.tsx
codyebberson: Did you try using `await act(async () ... `?

packages/core/src/version-utils.ts
ThatOneBro: SemVer technically includes allowances of a bunch of extra symbols:
`^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
https://semver.org/#:~:text=Additional%20labels%20for%20pre%2Drelease%20and%20build%20metadata%20are%20available%20as%20extensions%20to%20the%20MAJOR.MINOR.PATCH%20format.

This is just a check for the simple `MAJOR.MINOR.PATCH` we use

packages/mock/src/client.ts
ThatOneBro: Nope, I tried that. You can't do nullish-coallescing if the right side is an arrow function. Looks like it creates a syntax error:
`Uncaught SyntaxError: Malformed arrow function parameter list`

packages/core/src/version-utils.ts
ThatOneBro: What's wrong with asserting that a particular object has all the appropriate fields? It makes it so that we don't have to check in multiple places. Also some of our releases coming from Github didn't have some of these fields but had others, this was a simple abstraction that ensures we have a valid release manifest past a certain point, and then you get type-time info if something matches this shape.

packages/app/src/resource/ToolsPage.test.tsx
ThatOneBro: Not sure how but removing those `act`s works now....

It truly is a mystery because I spent probably 4 hours trying to get that to work on Wednesday

packages/core/src/version-utils.ts
ThatOneBro: Since we are exporting it from `@medplum/core` I don't want users to believe something is NOT semver because our function returns false.

Alternatively we could provide the full regex implementation provided by the `SemVer` site, I'm ok with doing that

packages/app/src/resource/ToolsPage.test.tsx
ThatOneBro: Okay, so now the tests pass locally but still fail in CI:
https://github.com/medplum/medplum/actions/runs/11116183748/job/30886228678?pr=5305

Going to try adding back the `await act(async () => {});`

packages/core/src/version-utils.test.ts
ThatOneBro: I kinda agree but also like that there is a lint rule for explaining why this `expect-error` is necessary (kinda reminds me of Rust unsafe), though might not be a good enough reason to use it

packages/app/src/resource/ToolsPage.test.tsx
ThatOneBro: Yep not even that works. Load-bearing sleeps seem to be necessary.

I don't fully understand why except I noticed that some of the effects that change the rendering of the modal children are not able to run, no matter how long you sleep in the first `act`, it seems like the updates in the first render have to be flushed before you can continue and trigger the effects in `UpgradeConfirmContent`.

codyebberson (approved):

PR #5308, opened 2024-09-25
Enable async batch processing

mattwiller:
Allowing users to send a batch request with the `Prefer: respond-async` header, which will enqueue processing of the batch as an async job and return immediately.  This allows processing larger batches, since the processing is decoupled from the HTTP request timeout; however, the batch JSON must still be small enough to fit in memory and be parsed as a whole.  The batch response `Bundle` is stored as a `Binary` resource, and can be retrieved asynchronously after the batch has finished processing.


packages/server/src/app.ts
```diff
@@ -145,6 +149,35 @@ export async function initApp(app: Express, config: MedplumServerConfig): Promis
   app.use(attachRequestContext);
   app.use(getRateLimiter(config));
   app.use('/fhir/R4/Binary', binaryRouter);
+
+  // Handle async batch by enqueueing job
+  app.post(
```
codyebberson, line 154: Can we move this handler to separate file?  I feel like `app.ts` is already too big, and individual handlers are a pretty clean way to keep things organized.

packages/server/src/fhir/binary.ts
```diff
@@ -41,70 +44,44 @@ async function handleBinaryWriteRequest(req: Request, res: Response): Promise<vo
     return;
   }
 
-  let binary: Binary | undefined = undefined;
   let binarySource: BinarySource = stream;
-
-  // From the spec: https://hl7.org/fhir/R4/binary.html#rest
```
codyebberson, line 47: Please try to avoid unnecessary code movement.

packages/server/src/fhir/binary.ts
```diff
@@ -41,70 +44,44 @@ async function handleBinaryWriteRequest(req: Request, res: Response): Promise<vo
...
-    ...binary,
-    url: getBinaryStorage().getPresignedUrl(binary),
+  const binary = await uploadBinaryData(repo, binarySource, {
+    id,
+    contentType,
+    filename: req.query['_filename'] as string | undefined,
+    securityContext: req.get('X-Security-Context'),
   });
+
+  await sendResponse(req, res, create ? created : allOk, binary);
```
codyebberson, line 84: I'm having a hard time following all of the refactoring.  It seems like the main objective was to refactor out a `uploadBinaryData` helper?

packages/server/src/fhir/binary.ts
```diff
@@ -41,70 +44,44 @@ async function handleBinaryWriteRequest(req: Request, res: Response): Promise<vo
...
+        // """
+        //   When binary data is written to the server (create/update - POST or PUT),
+        //   the data is accepted as is and treated as the content of a Binary,
+        //   including when the content type is "application/fhir+xml" or "application/fhir+json",
+        //   except for the special case where the content is actually a Binary resource.
+        // """
+        const resource = body as Binary;
+        const binary = await (create ? repo.createResource<Binary>(resource) : repo.updateResource<Binary>(resource));
+        await sendResponse(req, res, create ? created : allOk, binary);
+        return;
```
codyebberson, line 69: If we're going to make this refactor, then I think this whole try/catch block needs to be inside the special case handler.  For example, as written, if `createResource` fails for some reason, we would jump to the `catch` block, and then proceed to upload the body as a binary.

packages/server/src/app.ts
```diff
@@ -154,6 +156,10 @@ export async function initApp(app: Express, config: MedplumServerConfig): Promis
   app.use(attachRequestContext);
   app.use(getRateLimiter(config));
   app.use('/fhir/R4/Binary', binaryRouter);
+
+  // Handle async batch by enqueueing job
+  app.post('/fhir/R4/', authenticateRequest, asyncWrap(asyncBatchHandler(config)));
```
codyebberson, line 161: I don't think we use trailing slashed on any other routes... It may not matter, but preference for consistency.

Does that mean that this route handler will intercept all other `POST /fhir/R4` requests?

I'm trying to think if there are any others....  Maybe mult-type search by POST?

https://hl7.org/fhir/R4/search.html#Introduction

image

packages/server/src/fhir/binary.test.ts
```diff
@@ -259,23 +259,16 @@ describe('Binary', () => {
       .set('Content-Type', ContentType.FHIR_JSON)
       .send({
         ...binary,
-        securityContext: { reference: 'Patient/123' },
```
codyebberson, line 262: Why did we remove this test?

Can we instead add new test?

codyebberson (commented):
2 open items, and then I'm good to go on this:

1. Can we confirm that "POST to search with multi-type search" still works?
2. Can we add back the Binary + securityContext test?

packages/server/src/fhir/binary.test.ts
mattwiller: If you expand the diff above, you'll see that this test case was a mis-named copy-paste and not really testing `securityContext` per se — I replaced it with a test of the changes I made.  The `securityContext` test case still exists and is being run

codyebberson (approved):

reshmakh:
@ianplunkett - let's confirm/check if rate limit will affect this behavior

PR #5309, opened 2024-09-26
Actually fail esbuild in catch

mattlong:
@ianplunkett discovered that `@medplum/cli` was failing to build locally. That build failure was due to needing to change the esbuild config a bit. More importantly, this uncovered that our pattern of `.catch(console.error)` in esbuild scripts was effectively suppressing build failures.


ThatOneBro (approved):

ThatOneBro:
Good catch 👍

PR #5321, opened 2024-09-27
Make stream.not.readable a warning

codyebberson:


packages/server/src/app.ts
```diff
@@ -125,6 +125,15 @@ function errorHandler(err: any, req: Request, res: Response, next: NextFunction)
     sendOutcome(res, badRequest('File too large'));
     return;
   }
+  if (err.type === 'stream.not.readable') {
+    // This is a common error when the client disconnects
+    // See: https://expressjs.com/en/resources/middleware/body-parser.html
+    // It is commonly associated with an AWS ALB disconnect status code 460
+    // See: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-460-issues
+    getLogger().warn('Stream not readable', err);
+    sendOutcome(res, badRequest('Stream not readable'));
```
ThatOneBro, line 134: Just want to note this error may not always be tied to this particular case, but as long as we don't introduce any async weirdness accidentally in the future then hopefully we won't be seeing it pop up in another case

ThatOneBro (approved):

PR #5327, opened 2024-09-30
Fixes #5145 - set meta.account via query param

### LINKED ISSUE #5145: Alternatives to setting meta.account.reference
Running into an issue when posting FHIR resources using [firely](https://github.com/FirelyTeam/firely-net-sdk). I am unable to set  `meta.account.reference` (Get a forbidden error when i POST without this). Is there a way I can re-use the same client app (this means I can't hardcode the compartment onto the access policy) inorder to create resources between different organizations?

```
{
  "resourceType": "Observation",
    "meta": {
      "account": {
        "reference": "Organization/c12340f2-eaf3-4017-9c24-da5fbe05fe96"
      }
    }, 
```

codyebberson:


ThatOneBro (approved):
LGTM, seems like a sensible workaround that is easy to grok

mattwiller:
@codyebberson Did we consider using an extension on the `Meta` type instead of or in addition to the query param?  That might be a more future-proof method towards potentially getting rid of "extended mode"

PR #5328, opened 2024-10-01
feat(server): add `SearchParameter` for `{Appointment, Slot}.end`

### LINKED ISSUE #5310: Search parameters for Slot.end and Appointment.end
User Driven: https://medplum.slack.com/archives/C078NUJP6B1/p1727303627541449

The base FHIR spec doesn't include a search parameter for `Slot.end` or `Appointment.end`. This makes it very clunky to find out which slots/appointments overlap a given point in time. 

Currently the workaround is to: 
1. Search for all slots that start in some  large time range before the current instant
2. Filter in memory to those that also end after the current instant

ThatOneBro:
Fixes #5310


codyebberson (approved):
Deployed on staging, rebuilt search params, everything looks good 👍

PR #5331, opened 2024-10-01
feat(server): add `DiagnosticReport.study` as `SearchParameter`

ThatOneBro:


packages/generator/src/migrate.ts
```diff
@@ -226,9 +226,7 @@ function buildCreateTables(result: SchemaDefinition, resourceType: string, fhirT
     ],
     indexes: [
       { columns: ['resourceId'], indexType: 'btree' },
-      { columns: ['code'], indexType: 'btree' },
```
ThatOneBro, line 229: We merged these indexes into composite ones so they are no longer needed

mattwiller (approved):

PR #5333, opened 2024-10-02
Index default search parameters for all resources

deklungel:
Greetings,

We're running into an issue while using the UI to create a MedicinalProductPacked. We've linked our own profile to this resource, tying the packageItem.manufacturedItem to our (other) own profile.

This causes the UI to execute following call to the server, searching for viable MedicinalProductManufactured items:
`https://{fhir server url}/fhir/R4/MedicinalProductManufactured?_id=&_count=10&_profile={profile url}`
This call fails with the error message `Unknown search parameter: _profile`.

It seems that the default search parameters are only populated if there also exists another, custom, search parameter. Which is not the case for MedicinalProductManufactured: https://www.medplum.com/docs/api/fhir/resources/medicinalproductmanufactured#search-parameters

I believe the following changes fix this issue. I've written a unit test which failed before these changes "Can search MedicinalProductManufactured on _profile"

Thoughts and remarks are welcome.

Kind regards

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -156,7 +156,7 @@ export function indexSearchParameterBundle(bundle: Bundle<SearchParameter>): voi
 export function indexDefaultSearchParameters(bundle: StructureDefinition[] | Bundle): void {
   const sds = Array.isArray(bundle) ? bundle : (bundle.entry?.map((e) => e.resource as StructureDefinition) ?? []);
   for (const sd of sds) {
-    if(sd.resourceType === 'StructureDefinition'){
+    if (sd.resourceType === 'StructureDefinition') {
       getOrInitTypeSchema(sd.type);
     }
   }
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/structure.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    test('Can search Organization on _profile', async () =>{
+        const param = getSearchParameter('Organization', '_profile');
+        expect(param).toBeDefined();
+    });
+
+    test('Can search MedicinalProductManufactured on _profile', async () =>{
+        const param = getSearchParameter('MedicinalProductManufactured', '_profile');
+        expect(param).toBeDefined();
+    });
+});
```
mattwiller: Looks like this file is missing a newline at the end
```suggestion
});

```

packages/core/src/types.ts
```diff
@@ -152,15 +153,17 @@ export function indexSearchParameterBundle(bundle: Bundle<SearchParameter>): voi
...
-  for (const resourceType of searchParam.base ?? []) {
-    let typeSchema = globalSchema.types[resourceType];
+export function indexDefaultSearchParameters(bundle: StructureDefinition[] | Bundle): void {
+  const sds = Array.isArray(bundle) ? bundle : (bundle.entry?.map((e) => e.resource as StructureDefinition) ?? []);
+  for (const sd of sds) {
+    if(sd.resourceType === 'StructureDefinition'){
+      getOrInitTypeSchema(sd.type);
+    }
+  }
+}
```
mattwiller: It looks like the signature for this function is broader than it needs to be, and it could be simplified using one of our helper functions:
```suggestion
export function indexDefaultSearchParameters(bundle: Bundle): void {
  const sds = flatMapFilter(bundle.entry, (e) => e.resource?.resourceType === 'StructureDefinition' ? e.resource : undefined) ?? [];
  for (const sd of sds) {
    getOrInitTypeSchema(sd.type);
  }
}
```

packages/core/src/types.ts
```diff
@@ -152,15 +153,17 @@ export function indexSearchParameterBundle(bundle: Bundle<SearchParameter>): voi
...
+
+export function getOrInitTypeSchema(resourceType: string): TypeInfo {
```
mattwiller, line 165: Does this function need to be exported?
```suggestion
function getOrInitTypeSchema(resourceType: string): TypeInfo {
```

packages/server/src/fhir/structure.test.ts
deklungel: Seems to be resolved by autofix.ci

mattwiller (approved):

PR #5337, opened 2024-10-03
refactor(server): make `code` for `Bot/$deploy` optional

### LINKED ISSUE #4353: Make `code` parameter from Bot/$deploy optional
Long ago, the `Bot` resource only stored the source code, and the compiled code was set during `$deploy`

Now we store both the source and compiled code as Attachments on the `Bot` resource, so the `code` parameter on the $deploy operation is unncessary. We can make it optional and then deprecate in v4

ThatOneBro:
Closes #4353

### DIFF
--- a/packages/app/src/resource/BotEditor.test.tsx
+++ b/packages/app/src/resource/BotEditor.test.tsx
@@ -103,20 +103,11 @@ describe('BotEditor', () => {
     await setup('/Bot/123/editor');
     expect(await screen.findByText('Deploy')).toBeInTheDocument();
 
-    // Mock the code frame
-    (screen.getByTestId<HTMLIFrameElement>('code-frame').contentWindow as Window).postMessage = (
-      _message: any,
-      _targetOrigin: any,
-      transfer?: Transferable[]
-    ) => {
-      (transfer?.[0] as MessagePort).postMessage({ result: 'console.log("foo");' });
-    };
-
     await act(async () => {
       fireEvent.click(screen.getByText('Deploy'));
     });
 
-    expect(await screen.findByText('Deployed')).toBeInTheDocument();
+    expect(screen.getByText('Deployed')).toBeInTheDocument();
   });
 
   test('Deploy error', async () => {
@@ -130,7 +121,7 @@ describe('BotEditor', () => {
       fireEvent.click(screen.getByText('Deploy'));
     });
 
-    expect(await screen.findByText('Something went wrong')).toBeInTheDocument();
+    expect(screen.getByText('Something went wrong')).toBeInTheDocument();
   });
 
   test('Execute success', async () => {
--- a/packages/app/src/resource/BotEditor.test.tsx
+++ b/packages/app/src/resource/BotEditor.test.tsx
@@ -13,7 +13,7 @@ describe('BotEditor', () => {
     jest.spyOn(medplum, 'download').mockImplementation(async () => ({ text: async () => 'test' }) as unknown as Blob);
 
     // Mock bot operations
-    medplum.router.router.add('POST', 'Bot/:id/$deploy', async () => [allOk]);
+    medplum.router.router.add('GET', 'Bot/:id/$deploy', async () => [allOk]);
     medplum.router.router.add('POST', 'Bot/:id/$execute', async () => [allOk]);
 
     await act(async () => {
@@ -120,23 +120,17 @@ describe('BotEditor', () => {
   });
 
   test('Deploy error', async () => {
-    await setup('/Bot/123/editor');
-    expect(await screen.findByText('Deploy')).toBeInTheDocument();
+    const medplum = new MockClient();
+    medplum.router.router.add('GET', 'Bot/:id/$deploy', async () => [badRequest('Something went wrong')]);
 
-    // Mock the code frame
-    (screen.getByTestId<HTMLIFrameElement>('code-frame').contentWindow as Window).postMessage = (
-      _message: any,
-      _targetOrigin: any,
-      transfer?: Transferable[]
-    ) => {
-      (transfer?.[0] as MessagePort).postMessage({ error: badRequest('Error') });
-    };
+    await setup('/Bot/123/editor', medplum);
+    expect(await screen.findByText('Deploy')).toBeInTheDocument();
 
     await act(async () => {
       fireEvent.click(screen.getByText('Deploy'));
     });
 
-    expect(screen.getByText('Error')).toBeInTheDocument();
+    expect(screen.getByText('Something went wrong')).toBeInTheDocument();
   });
 
   test('Execute success', async () => {
--- a/packages/app/src/resource/BotEditor.test.tsx
+++ b/packages/app/src/resource/BotEditor.test.tsx
@@ -13,7 +13,7 @@ describe('BotEditor', () => {
     jest.spyOn(medplum, 'download').mockImplementation(async () => ({ text: async () => 'test' }) as unknown as Blob);
 
     // Mock bot operations
-    medplum.router.router.add('GET', 'Bot/:id/$deploy', async () => [allOk]);
+    medplum.router.router.add('POST', 'Bot/:id/$deploy', async () => [allOk]);
     medplum.router.router.add('POST', 'Bot/:id/$execute', async () => [allOk]);
 
     await act(async () => {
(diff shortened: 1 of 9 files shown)


packages/server/src/fhir/routes.ts
```diff
@@ -207,6 +207,7 @@ function initInternalFhirRouter(): FhirRouter {
 
   // Bot $deploy operation
   router.add('POST', '/Bot/:id/$deploy', deployHandler);
+  router.add('GET', '/Bot/:id/$deploy', deployHandler);
```
codyebberson, line 210: Hm, we try to avoid using `GET` requests for non-idempotent actions.  I suppose you could argue that repeatedly deploying is somewhat idempotent, but imho it kinda goes against the spirit.

It seems like it should be possible to just make `code` optional within the post handler, yeah?

packages/app/src/resource/BotEditor.tsx
```diff
@@ -106,16 +100,15 @@ export function BotEditor(): JSX.Element | null {
       e.stopPropagation();
       setLoading(true);
       try {
-        const code = await getCodeOutput();
-        await medplum.post(medplum.fhirUrl('Bot', id, '$deploy'), { code });
+        await medplum.get(medplum.fhirUrl('Bot', id, '$deploy'));
```
codyebberson, line 103: Prefer to keep this a `POST` request.

packages/app/src/resource/BotEditor.tsx
```diff
@@ -75,20 +71,18 @@ export function BotEditor(): JSX.Element | null {
...
             path: '/sourceCode',
             value: sourceCode,
-          });
-        } else {
-          operations.push({
+          },
+          {
             op: 'add',
-            path: '/sourceCode',
+            path: '/executableCode',
```
codyebberson, line 82: This won't work, will it?  `executableCode` has to be CJS, it won't work with TypeScript?

packages/app/src/resource/BotEditor.tsx
```diff
@@ -55,10 +55,6 @@ export function BotEditor(): JSX.Element | null {
     return sendCommand(codeFrameRef.current as HTMLIFrameElement, { command: 'getValue' });
   }, []);
 
-  const getCodeOutput = useCallback(() => {
```
codyebberson, line 58: Why did we remove this?

codyebberson (changes requested):

packages/app/src/resource/BotEditor.tsx
ThatOneBro: I guess I don't understand how it is currently working? Since all I did was set the `executableCode` to what gets passed into the `$deploy` operation previously

packages/app/src/resource/BotEditor.tsx
ThatOneBro: We can't do post without a body in `MedplumClient`. Should I make that possible instead?

packages/app/src/resource/BotEditor.tsx
ThatOneBro: Ah I see, `getCodeOutput` is compiling the code...

packages/app/src/resource/BotEditor.tsx
ThatOneBro: Yeah my bad. I put it back in

packages/app/src/resource/BotEditor.tsx
ThatOneBro: I just passed in `undefined` instead

packages/app/src/resource/BotEditor.test.tsx
```diff
@@ -120,23 +120,17 @@ describe('BotEditor', () => {
...
-    };
+    await setup('/Bot/123/editor', medplum);
+    expect(await screen.findByText('Deploy')).toBeInTheDocument();
 
     await act(async () => {
       fireEvent.click(screen.getByText('Deploy'));
     });
 
-    expect(screen.getByText('Error')).toBeInTheDocument();
+    expect(screen.getByText('Something went wrong')).toBeInTheDocument();
```
ThatOneBro, line 133: Changed this because common words like `Error` often appear in the document and give false positives on tests

packages/app/src/resource/BotEditor.tsx
```diff
@@ -74,21 +76,21 @@ export function BotEditor(): JSX.Element | null {
       setLoading(true);
       try {
         const code = await getCode();
+        const codeOutput = await getCodeOutput();
         const sourceCode = await medplum.createAttachment(code, 'index.ts', 'text/typescript');
-        const operations: PatchOperation[] = [];
-        if (bot?.sourceCode) {
```
ThatOneBro, line 79: Not necessary to check if field is populated and conditionally do a `replace`, since `add` means "add or replace if exists":
https://www.npmjs.com/package/rfc6902#:~:text=add%3A%20Insert%20the%20given%20value%20at%20path.%20Or%20replace%20it%2C%20if%20it%20already%20exists.

packages/server/src/fhir/operations/deploy.ts
```diff
@@ -12,47 +13,62 @@ export async function deployHandler(req: FhirRequest): Promise<FhirResponse> {
...
   // First read the bot as the user to verify access
   await ctx.repo.readResource<Bot>('Bot', id);
 
   // Then read the bot as system user to load extended metadata
   const systemRepo = getSystemRepo();
   const bot = await systemRepo.readResource<Bot>('Bot', id);
 
+  // Validate that the request body has a code property
+  // Or that the Bot already has executable code attached
+  const code = req.body.code as string | undefined;
```
ThatOneBro, line 25: Moved this block because we need to get the `bot` before doing this check now, which requires the `readResource` to come first

packages/core/src/client.ts
```diff
@@ -3349,9 +3349,9 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
   private setRequestBody(options: MedplumRequestOptions, data: any): void {
     if (
       typeof data === 'string' ||
-      (typeof Blob !== 'undefined' && (data instanceof Blob || data.constructor.name === 'Blob')) ||
-      (typeof File !== 'undefined' && (data instanceof File || data.constructor.name === 'File')) ||
-      (typeof Uint8Array !== 'undefined' && (data instanceof Uint8Array || data.constructor.name === 'Uint8Array'))
+      (typeof Blob !== 'undefined' && (data instanceof Blob || data?.constructor.name === 'Blob')) ||
```
ThatOneBro, line 3352: If `body` is `undefined`, need this check here before comparing the constructor name

packages/server/src/fhir/operations/deploy.test.ts
```diff
@@ -22,6 +42,128 @@ describe('Deploy', () => {
...
+      expect.objectContaining({
+        ...bot,
+        executableCode: expect.objectContaining({ url: expect.any(String) }),
+        meta: expect.objectContaining({ ...bot.meta, lastUpdated: expect.any(String), versionId: expect.any(String) }),
+      }),
+      code
+    );
+  });
+
+  test('Deprecated: Deploy bot with code parameter', async () => {
```
codyebberson, line 123: Is this actually deprecated?  I thought we were just making it optional?

ThatOneBro: Good question, I was just kinda going off what Rahul mentioned about deprecating it but I guess that discussion was never conclusive per se.

Philosophically I'm more for deprecating it (reducing amount of ways to do one thing, there's no good reason to keep both ways forever I think), mostly just to homogenize how code is attached to a Bot at some point in the future. But I'm definitely open to be persuaded for it not to be deprecated.

cc: @rahul1

packages/app/src/resource/BotEditor.tsx
ThatOneBro: Update 2:
Just passing `undefined` kinda broke the `MedplumClient.post` functionality, I ended up having to add support for `POST` without a body. Works now 👍

codyebberson (approved):
overall lgtm

I'd probably remove the word "deprecated", but that's a minor point

For context, we definitely need to support this for existing customers, and likely will for the foreseeable future.

In the earliest version of bots, we didn't store the compiled code anywhere on the `Bot` or a `Binary` or anything, it just went direct into the lambda.  Several customers invested in Bot automation based on that paradigm, so it will take a fair amount of migration effort before we can ever remove that code path.

I agree with your in principle on the idea of "there should be one recommended way to do things".

rahul1:
I've seen a lot of users work around the existing behavior in the wild. Think we should mark as deprecate in v4 to signal, and remove in r5

PR #5339, opened 2024-10-03
Deduplicate search results

### LINKED ISSUE #5322: Duplicate results in _has query
In the query to retrieve Patients using _has:ChargeItem, for example:

`Patient?_has:ChargeItem:patient:occurrence=ge2024/09/01&_total=accurate`

The result is duplicated as many times as the patient has a ChargeItem on the specified date. If you add another _has with an end date, for example:

`Patient?_has:ChargeItem:patient:occurrence=gt2024/09/01&_has:ChargeItem:patient:occurrence=le2024/09/30&_total=accurate`

The results are repeated even more. However, the total inside the bundle is correct and corresponds to the number of patients meeting the condition.

mattwiller:
Resolves #5322

### DIFF
--- a/packages/server/src/fhir/search.ts
+++ b/packages/server/src/fhir/search.ts
@@ -355,6 +355,8 @@ function getBaseSelectQueryForResourceType(
   if (opts?.resourceTypeQueryCallback) {
     opts.resourceTypeQueryCallback(resourceType, builder);
   }
+
+  builder.distinctOn(new Column(resourceType, 'id'));
   return builder;
 }
 
--- a/packages/server/src/fhir/search.ts
+++ b/packages/server/src/fhir/search.ts
@@ -344,10 +344,9 @@ function getBaseSelectQueryForResourceType(
   opts?: GetBaseSelectQueryOptions
 ): SelectQuery {
   const builder = new SelectQuery(resourceType);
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/sql.ts
```diff
@@ -145,13 +145,17 @@ export function escapeLikeString(str: string): string {
   return str.replaceAll(/[\\_%]/g, (c) => '\\' + c);
 }
 
-export class Column {
+export class Column implements Expression {
```
ThatOneBro, line 148: Nice

packages/server/src/fhir/search.ts
```diff
@@ -344,17 +344,20 @@ function getBaseSelectQueryForResourceType(
...
+    builder.column(idColumn).column(new Column(resourceType, 'content'));
   }
   repo.addDeletedFilter(builder);
   repo.addSecurityFilters(builder, resourceType);
   addSearchFilters(repo, builder, resourceType, searchRequest);
   if (opts?.resourceTypeQueryCallback) {
     opts.resourceTypeQueryCallback(resourceType, builder);
   }
+
+  if (builder.joins.length > 0) {
```
ThatOneBro, line 358: Cool, makes sense

ThatOneBro (approved):

PR #5342, opened 2024-10-04
fix(BaseChat): only call `onMessageReceived` on new message

### LINKED ISSUE #4876: BaseChat shouldn't call `onMessageReceived` twice when setting received timestamp
Consider this scenario for the `BaseChat` component:

- `onMessageReceived` currently triggers on any change made to a communication that:
- Matches the search query
- The sender is not the current user
- So `onMessageReceived` gets called:
- received is null
- Client sets `Communication.received`
- The `Communication` is upserted
- **`onMessageReceived called _again_, with the new received time**  (this should change, since the message is new)

https://github.com/medplum/medplum/blob/b1f86f8c1a0a0999d9ec91e203662b15b3fb4a33/packages/react/src/chat/BaseChat/BaseChat.tsx#L20-L39

ThatOneBro:
Fixes #4876

### DIFF
--- a/packages/react/src/chat/BaseChat/BaseChat.test.tsx
+++ b/packages/react/src/chat/BaseChat/BaseChat.test.tsx
@@ -217,6 +217,7 @@ describe('BaseChat', () => {
 
     expect(await screen.findByText("Doc, I can't feel my legs")).toBeInTheDocument();
     expect(onMessageReceived).toHaveBeenCalledWith(incomingMessage);
+    expect(onMessageReceived).toHaveBeenCalledTimes(1);
   });
 
   test('`onMessageReceived` not called on outgoing message', async () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/chat/BaseChat/BaseChat.test.tsx
```diff
@@ -78,11 +78,8 @@ describe('BaseChat', () => {
   let defaultMedplum: MockClient;
   let defaultSubManager: MockSubscriptionManager;
 
-  beforeAll(() => {
-    defaultMedplum = new MockClient({ profile: DrAliceSmith });
-  });
-
   beforeEach(() => {
+    defaultMedplum = new MockClient({ profile: DrAliceSmith });
```
ThatOneBro, line 82: Moved this so that tests aren't affecting each other's state

codyebberson (approved):

PR #5346, opened 2024-10-07
Dependency upgrades 2024-10-07

medplumbot:
Dependency upgrades


codyebberson:
Looks like some kind of Mantine change or regression:

https://github.com/medplum/medplum/actions/runs/11212326405/job/31163511491?pr=5346

image

PR #5364, opened 2024-10-09
Add ProjectMembership.identifier

codyebberson:


packages/docs/static/data/datatypeDefinitions/meta.json
```diff
@@ -218,10 +218,28 @@
       "min": 0,
       "max": "1",
       "short": "Optional account reference that can be used for sub-project compartments.",
-      "definition": "@deprecated Use Meta.compartment instead",
+      "definition": "@deprecated Use Meta.accounts instead",
```
codyebberson, line 221: Changes in this file appear to be missing changes from the `compartment`/`account`/`accounts` cleanup.

packages/server/src/auth/me.ts
```diff
@@ -61,6 +61,7 @@ export async function meHandler(req: Request, res: Response): Promise<void> {
     membership: {
       resourceType: 'ProjectMembership',
       id: membership.id,
+      identifier: membership.identifier,
```
codyebberson, line 64: Including the new identifiers in `/auth/me`

ThatOneBro (approved):

PR #5366, opened 2024-10-09
fix(fhircast): align context change request with STU3

### LINKED ISSUE #5356: Allow using STU3 URLs for FHIRcast
We are trying to utilize the medplum client to interact with a third party fhircast hub which implements the STU3 spec where the hub topic can be defined in the body instead of being required on the url itself: https://hl7.org/fhir/uv/fhircast/2024May/2-6-RequestContextChange.html#request-context-change-example. Currently having to manually override the url:

```
  const { fetch: originalFetch } = window;

  window.fetch = async (...args) => {
    let [resource] = args;

    return await originalFetch(
      resource.toString().replace("fhircast/STU3", fhircastPrefix)
    );
  };
```

This should be supported if Medplum plans on supporting FHIR cast STU3: https://www.medplum.com/docs/fhircast. If this is already on the road map, apologies!

ThatOneBro:
Fixes #5356

### DIFF
--- a/packages/server/src/fhir/outcomes.ts
+++ b/packages/server/src/fhir/outcomes.ts
@@ -1,20 +1,28 @@
 import { getStatus, isAccepted } from '@medplum/core';
 import { OperationOutcome } from '@medplum/fhirtypes';
-import { randomUUID } from 'crypto';
 import { Response } from 'express';
 import { Result, ValidationError } from 'express-validator';
+import { randomUUID } from 'node:crypto';
 import { buildTracingExtension } from '../context';
 
 export function invalidRequest(errors: Result<ValidationError>): OperationOutcome {
   return {
     resourceType: 'OperationOutcome',
     id: randomUUID(),
-    issue: errors.array().map((error) => ({
-      severity: 'error',
-      code: 'invalid',
-      expression: getValidationErrorExpression(error),
-      details: { text: error.msg },
-    })),
+    issue: errors
+      .array()
+      .flatMap((error) => {
+        if (error.type === 'alternative') {
+          return error.nestedErrors;
+        }
+        return error;
+      })
+      .map((error) => ({
+        severity: 'error',
+        code: 'invalid',
+        expression: getValidationErrorExpression(error),
+        details: { text: error.msg },
+      })),
   };
 }
 
--- a/packages/server/src/fhircast/routes.ts
+++ b/packages/server/src/fhircast/routes.ts
@@ -7,7 +7,7 @@ import {
   normalizeErrorString,
 } from '@medplum/core';
 import { Request, Response, Router } from 'express';
-import { body, validationResult } from 'express-validator';
+import { body, oneOf, validationResult } from 'express-validator';
 import { asyncWrap } from '../async';
 import { getConfig } from '../config';
 import { getLogger } from '../context';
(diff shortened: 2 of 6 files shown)


packages/server/src/fhircast/routes.ts
```diff
@@ -71,65 +71,42 @@ publicSTU3Routes.get('/.well-known/fhircast-configuration', (_req: Request, res:
...
+// Or publish an event depending on payload
 protectedCommonRoutes.post(
   '/',
-  [
-    body('hub.channel.type').notEmpty().withMessage('Missing hub.channel.type'),
-    body('hub.mode').notEmpty().withMessage('Missing hub.mode'),
-    body('hub.topic').notEmpty().withMessage('Missing hub.topic'),
-    body('hub.events').notEmpty().withMessage('Missing hub.events'),
-  ],
+  oneOf(
```
ThatOneBro, line 78: This is a way to validation of two different types of payloads. https://express-validator.github.io/docs/api/one-of/

One of them is the "subscription request"
The other is the "context change request"

packages/server/src/fhircast/routes.ts
```diff
@@ -71,65 +71,42 @@ publicSTU3Routes.get('/.well-known/fhircast-configuration', (_req: Request, res:
...
-    if (type !== 'websocket') {
-      sendOutcome(res, badRequest('Invalid hub.channel.type'));
-      return;
-    }
-
-    const mode = req.body['hub.mode'] as 'subscribe' | 'unsubscribe';
-    if (!(['subscribe', 'unsubscribe'] as const).includes(mode)) {
-      sendOutcome(res, badRequest('Invalid hub.mode'));
+    // If there is an ID, this is a context change request
+    if (req.body.id) {
```
ThatOneBro, line 105: We check for `id` to determine whether its a context change request or subscription request

packages/server/src/fhircast/routes.ts
```diff
@@ -140,40 +117,92 @@ protectedCommonRoutes.post(
...
+    await getRedis().del(`medplum:fhircast:topic:${event['hub.topic']}:latest`);
+  } else if (event['hub.event'] === 'DiagnosticReport-update') {
+    // See: https://build.fhir.org/ig/HL7/fhircast-docs/3-6-3-DiagnosticReport-update.html#:~:text=The%20Hub%20SHALL,the%20new%20updates.
+    event['context.priorVersionId'] = event['context.versionId'];
+    event['context.versionId'] = generateId();
+    stringifiedBody = JSON.stringify(req.body);
+    // TODO: Make sure this is actually supposed to be stored / overwrite open context? (ambiguous from docs, see: https://build.fhir.org/ig/HL7/fhircast-docs/2-9-GetCurrentContext.html)
+    await getRedis().set(`medplum:fhircast:topic:${event['hub.topic']}:latest`, stringifiedBody);
+  }
+  await getRedis().publish(event['hub.topic'], stringifiedBody);
```
ThatOneBro, line 204: The main change here is using `event['hub.topic']` instead of the `req.params.topic`

packages/server/src/fhircast/routes.ts
```diff
@@ -71,65 +71,43 @@ publicSTU3Routes.get('/.well-known/fhircast-configuration', (_req: Request, res:
...
+        body('event.context').notEmpty().withMessage('Missing event.context'),
+      ],
+      [
+        body('hub.channel.type').notEmpty().withMessage('Missing hub.channel.type'),
+        body('hub.mode').notEmpty().withMessage('Missing hub.mode'),
+        body('hub.topic').notEmpty().withMessage('Missing hub.topic'),
+        body('hub.events').notEmpty().withMessage('Missing hub.events'),
+      ],
+    ],
+    { errorType: 'least_errored' }
```
ThatOneBro, line 95: `least_errored` error type means of the different groups, the one with the least errors is selected as our "choice", ie. the one for which the validation feedback is given back to the user in the case that neither group passes validation

packages/server/src/fhir/outcomes.ts
```diff
@@ -1,20 +1,28 @@
...
-    issue: errors.array().map((error) => ({
-      severity: 'error',
-      code: 'invalid',
-      expression: getValidationErrorExpression(error),
-      details: { text: error.msg },
-    })),
+    issue: errors
+      .array()
+      .flatMap((error) => {
+        if (error.type === 'alternative') {
```
ThatOneBro, line 15: Because of the `oneOf` syntax, we need to flat map all the errors out of the `nestedErrors` field in the case of `ValidationError` of type `alternative`, since the errors for each group are nested under one error.

packages/server/src/fhircast/routes.ts
```diff
@@ -140,40 +118,92 @@ protectedCommonRoutes.post(
...
       stringifiedBody = JSON.stringify(req.body);
-      // TODO: Make sure this is actually supposed to be stored / overwrite open context? (ambiguous from docs, see: https://build.fhir.org/ig/HL7/fhircast-docs/2-9-GetCurrentContext.html)
-      await getRedis().set(`medplum:fhircast:topic:${req.params.topic}:latest`, stringifiedBody);
     }
-    await getRedis().publish(req.params.topic as string, stringifiedBody);
-    res.status(201).json({ success: true, event: body });
-  })
-);
+    // TODO: we need to get topic from event and not route param since per spec, the topic shouldn't be the slug like we have it
+    await getRedis().set(`medplum:fhircast:topic:${event['hub.topic']}:latest`, stringifiedBody);
```
codyebberson, line 192: While we're making these changes, should we move forward w/ namespacing these with project ID?

This is still open to potential topic ID collision, yeah?

codyebberson (approved):

packages/server/src/fhircast/routes.ts
ThatOneBro: Yeah for sure. We can do that in the other `FHIRcast` PR, I think I'll have to rebase that PR once this PR gets merged:
https://github.com/medplum/medplum/pull/4951

PR #5372, opened 2024-10-10
chore(actions): auto-assign reviewers from list

ThatOneBro:
This PR adds [this action](https://github.com/kentaro-m/auto-assign-action), which enables auto-assigning reviewers from a pre-defined set of reviewers, as well as does the "assign author as assignee" functionality our existing auto-assign action did.

The thought behind this is to reduce siloing and spread the burden of reviewing all PRs among the core dev team. We're just going to try it out and see if it works for now

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,15 @@
+# Set to true to add reviewers to pull requests
+addReviewers: true
+
+# Set addAssignees to 'author' to set the PR creator as the assignee.
+addAssignees: author
+
+# A list of reviewers to be added to pull requests (GitHub user name)
+reviewers:
+  - mattlong
+  - mattwiller
+  - ThatOneBro
+
+# A number of reviewers added to the pull request
+# Set 0 to add all the reviewers (default: 0)
+numberOfReviewers: 1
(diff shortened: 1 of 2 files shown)


.github/workflows/assign-pull-request.yml
```diff
@@ -1,8 +1,8 @@
 name: Assign pull request
 
 on:
-  pull_request:
-    types: [opened]
+  pull_request_target:
+    types: [opened, ready_for_review]
```
ThatOneBro, line 5: We want this to now run not only when a PR is opened, but run it again when PR marked ready for review, since by default it won't assign a reviewer to a draft PR

.github/auto_assign.yml
```diff
@@ -0,0 +1,15 @@
+# Set to true to add reviewers to pull requests
+addReviewers: true
+
+# Set addAssignees to 'author' to set the PR creator as the assignee.
+addAssignees: author
+
+# A list of reviewers to be added to pull requests (GitHub user name)
+reviewers:
```
ThatOneBro, line 8: This list can be expanded, but for now I thought we could start with us 3

codyebberson (approved):

PR #5389, opened 2024-10-14
Fixes #5385 - optional source param on QuestionnaireForm

### LINKED ISSUE #5385: QuestionnaireForm component errors if medplum profile not present
Currently, the QuestionnaireForm component assumes that `profile` is non-null when saving the resulting `QuestionnaireResponse`. This was fine assumption in the world before `on-behalf-of` auth. 

With `on-behalf-of`, auth is handled by the proxy server so will be undefined in the `MedplumClient`.

[source: createReference(source as ProfileResource),](https://github.com/medplum/medplum/blob/8965299a17ad87d5dcedafdccd5edf561876c8c6/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx#L101).

## Proposal
- [ ] Remove the hard requirement on `profile`
- [ ] Add an optional prop to allow the caller to specify the `source` element of the QuestionnaireResponse

Question: Should we enforce that _either_ profile _or_ the input prop `source` are defined

codyebberson:


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -67,6 +67,26 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+      props.onSubmit({
+        ...response,
+        questionnaire: getReferenceString(questionnaire as Questionnaire),
+        subject: props.subject,
+        source,
+        authored: new Date().toISOString(),
+        status: 'completed',
+      });
+    }
+  }, [medplum, props, questionnaire, response]);
```
ThatOneBro, line 88: This `useCallback` is going to change on every re-render since `props` will always be a different object.

You probably want to pull out those things from `props` you want at the top and add them individually to the deps array OR optionally keep the function unmemoized

```suggestion
  }, [medplum, onSubmit, source, subject, questionnaire, response]);
```

ThatOneBro (changes requested):

rahul1 (approved):

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
codyebberson: good catch

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -28,7 +29,6 @@ export interface QuestionnaireFormProps {
 
 export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element | null {
   const medplum = useMedplum();
```
ThatOneBro, line 31:
```suggestion
  const { onSubmit, subject, source: sourceFromProps } = props;
  const medplum = useMedplum();
```

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -67,6 +67,26 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+      if (!source) {
+        const profile = medplum.getProfile();
+        if (profile) {
+          source = createReference(profile);
+        }
+      }
+      props.onSubmit({
+        ...response,
+        questionnaire: getReferenceString(questionnaire as Questionnaire),
+        subject: props.subject,
```
ThatOneBro, line 82:
```suggestion
        subject,
```

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -67,6 +67,26 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
     [onChange]
   );
 
+  const handleSubmit = useCallback(() => {
+    if (props.onSubmit && response) {
```
ThatOneBro, line 71:
```suggestion
    if (onSubmit && response) {
```

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -67,6 +67,26 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+      let source = props.source;
+      if (!source) {
+        const profile = medplum.getProfile();
+        if (profile) {
+          source = createReference(profile);
+        }
+      }
+      props.onSubmit({
```
ThatOneBro, line 79:
```suggestion
      onSubmit({
```

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -67,6 +67,26 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
     [onChange]
   );
 
+  const handleSubmit = useCallback(() => {
+    if (props.onSubmit && response) {
+      let source = props.source;
```
ThatOneBro, line 72:
```suggestion
      let source = sourceFromProps;
```

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -28,44 +29,68 @@ export interface QuestionnaireFormProps {
...
-
-        return newResponse;
+      }
+
+      return newResponse;
+    });
+  }, []);
+
+  const handleSubmit = useCallback(() => {
+    const onSubmit = onSubmitRef.current;
```
ThatOneBro, line 75: Nice, I like that

ThatOneBro (approved):

PR #5391, opened 2024-10-15
DoseSpot UI ready for evaluation

codyebberson:
1. Can place an order
2. Checks for notifications from DoseSpot on a timer
3. Shows notifications in "Rx" icon in the toolbar
4. Clicking on "Rx" icon opens DoseSpot notifications page

image

Corresponding bot PR: https://github.com/medplum/medplum-ee/pull/257

### DIFF
--- a/examples/medplum-provider/src/components/DoseSpotIcon.tsx
+++ b/examples/medplum-provider/src/components/DoseSpotIcon.tsx
@@ -1,0 +1,101 @@
+import { ActionIcon, Indicator, Tooltip } from '@mantine/core';
+import { showNotification } from '@mantine/notifications';
+import { normalizeErrorString } from '@medplum/core';
+import { Identifier } from '@medplum/fhirtypes';
+import { useMedplumNavigate } from '@medplum/react';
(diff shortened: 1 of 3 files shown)


examples/medplum-provider/src/components/DoseSpotIcon.tsx
```diff
@@ -0,0 +1,101 @@
...
+      setUnreadCount(newCount);
+    } catch (err: unknown) {
+      showNotification({ color: 'red', title: 'Error', message: normalizeErrorString(err) });
+      stopTimer();
+    }
+  }, [medplum, stopTimer]);
+
+  const startTimer = useCallback(() => {
+    timerRef.current = setInterval(() => {
+      updateCount().catch(console.error);
```
ThatOneBro, line 72: Looks like we only run `updateCount` after the first timer has elapsed, do we wanna run it immediately on page load in case there are already notifications we can show?

ThatOneBro (approved):
Overall looks good, one comment about potentially updating count on mount

PR #5398, opened 2024-10-16
Fix condition for GraphQL requests to use reader DB when available

mattwiller:


packages/server/src/fhir/operations/graphql.test.ts
```diff
@@ -1062,6 +1063,19 @@ describe('GraphQL', () => {
...
+    const pool = getDatabasePool(DatabaseMode.READER);
+    const readerSpy = jest.spyOn(pool, 'query');
+
+    const res = await request(app)
+      .post('/fhir/R4/$graphql')
+      .set('Authorization', 'Bearer ' + accessToken)
+      .set('Content-Type', ContentType.JSON)
+      .send({ query: `{ PatientList(_id: "${patient.id}") { id } }` });
+    expect(res.status).toBe(200);
+    expect(readerSpy).toHaveBeenCalled();
```
mattlong, line 1077: Would it be reasonable to also check that the writer pool is only called for the expected operations if any? e.g. creating audit events?

mattlong (approved):

PR #5400, opened 2024-10-18
feat(db): add data migration to startup

### LINKED ISSUE #5566: Enable data migrations
Currently we don't have a process for doing data migrations. Before v4, we should add a protocol for doing data migrations in a way which prevents running a version where a previous data migration is required.

Our main strategy for preventing running a version without the required data migration would be to link a data migration to a specific Medplum version, the version that the data migration was introduced in, which will be at least 1 version before the code to use that migration is introduced. That means cutting a release as soon as a data migration is added.

Using the data migrations, which depend on a particular version of the server, as a gate on startup will prevent running a version that assumes that a data migration has already been applied, because if the data migration exists in the folder, it has to have been run, and since the server will error if not on the correct version to run a data migration, this should prevent getting into a bad state.

The procedure is relatively simple:

1. Check if this is first startup of the server. This step already happens in main, except we just add an entry to the DatabaseMigration table setting version and dataVersion both to 0. Now, we will set the dataVersion to the latest we know about in order to no-op data migration and skip data version checks since we don't have any data.
2. If not the first startup, check to see if there is a data migration for our current version + 1. If so, check the data-migration-manifest.json to see what is the required serverVersion. If we're on the wrong server version, bail out with an error.
3. Do the normal schema migration process, applying all schema migrations for the current version
4. After migrating schema, finally apply our data migration. This is currently sync but it might be better for large migrations that this is done in an AsyncJob, and allow the server to run as usual.
5. If all these steps are successful, continue running as usual. If you were applying this as part of a migration strategy to a future version, then you can redeploy with the next version at this point.

ThatOneBro:
Closes #5566 

## Background
After working through this process with @mattwiller, I drafted this PR in order to validate our theorized data migration strategy. Our goal was to make minimal changes to the current startup flow, keep the process dead simple, and also ensure that self-hosted customers will not end up in bad states by skipping versions where data migrations were required.

This PR adds a simple data migration strategy to Medplum server. Currently we don't really have a data migration strategy that works across the board, everything is very ad hoc.

## Basic Strategy
Our main strategy for preventing running versions that require a certain data migration before the code for the new schema can be used is to link a data migration to a specific Medplum version, the version that the data migration was introduced in, which will be at least 1 version before the code to use that migration is introduced. That means cutting a new minor release as soon as a data migration is added.

Using the data migrations, which depend on a particular version of the server, as a gate on startup will prevent running a version that assumes that a data migration has already been applied, since if we have a pending migration and the version required for that pending migration is lower than the current version, then we know we cannot run the current version safely.

## Protocol
The protocol that this PR introduces is pretty simple:

1. Check if this is first startup of the server. This step already happens in main, except we just add an entry to the `DatabaseMigration` table setting `version` and `dataVersion` both to 0. Now, we will set the `dataVersion` to the latest we know about in order to no-op data migration and skip data version checks since we don't have any data.
2. If not the first startup, check to see if there is a data migration for our current version + 1. If so, check the `data-migration-manifest.json` to see what is the required `serverVersion`. If we're on the server version higher than the required*, bail out with an error.
3. Do the normal schema migration process, applying all schema migrations for the current version
4. After migrating schema, clients can issue a post to `/admin/super/migrate`, kicking off the data migration job which runs in the background. The job can be polled via an `AsyncJob` returned from the endpoint.
5. Continue running as usual. If you were applying this as part of a migration strategy to a future version, then you can redeploy with the next version after the migration `AsyncJob` reported `completed` status.

*For now as this is the first data migration required and customers are not used to running data migrations, which can be very long-running jobs, we have decided that self-hosters can run their data migration in any version before v4. This is allowed via the check in the `migrate` startup path for any version less than the next minor version after the required version currently. Going forward we will want to tighten this check to any version less than or equal to the `serverVersion` in the manifest since we need to ensure that we can prune dead paths in the next patch after a minor version containing a data migration.

## CLI `update-server`
While changes to `update-server` are not included in this PR, implications regarding usage of the new protocol in tandem with `update-server` have been considered thoroughly and informed this PR overall.

A few things to consider:
1. When using CDK, there are transition periods where nodes in the cluster can be on previous versions after a recent new version deploy after before schema migrations complete. Nodes in previous versions may "lie" about a migration being complete if you naively ask about the "current migration" being applied or not without asserting which version you are talking about (hence the inclusion of a `dataVersion` assertion)
2. Because multiple clients may be admins that could run the data migrations or redeploy the stack at a given time. This is what necessitates the inclusion of an `upgradelock` which lives in Redis and is taken out by the client before attempting to step through an upgrade path version by version. With the CLI making a request for this lock before attempting to run through the upgrade path, it should prevent clients from trying to upgrade simultaneously.
3. Since this is the first migration, the CLI is not necessarily required for this first upgrade, but getting the means of supporting a CLI upgrade from the current version to any future version would be nice, hence why the `upgradelock` and similar have already been preemptively added.


packages/server/src/database-utils.ts
```diff
@@ -0,0 +1,67 @@
+import { EventEmitter } from 'node:events';
+import { Duplex, Readable, Writable } from 'node:stream';
+import { Pool, PoolClient, QueryArrayResult } from 'pg';
+
+export class MockPoolClient extends Duplex implements PoolClient {
```
ThatOneBro, line 5: Moved this out to cleanup the `database-config.test.ts` file

ThatOneBro:
Closed in favor of #5820

PR #5404, opened 2024-10-18
Validate date search values

mattwiller:


ThatOneBro (approved):
Nice, I found this pain point the other day when a date search silently failed. This should be a nice little DX boost

PR #5412, opened 2024-10-21
Validate access policy criteria against stated resource type

mattwiller:
Ensure that `AccessPolicy` criteria are consistent: they must refer to the same resource type as the `AccessPolicy.resource.resourceType`, e.g.
```
{
  "resourceType": "Practitioner",
  "criteria": "Practitioner?identifier=http://hl7.org/fhir/sid/us-npi|9999999999"
}
```

This is checked in two places:
- When the `AccessPolicy` is modified, invalid `criteria` will trigger a 400 error
- When the access policy is applied, invalid `criteria` will be ignored

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1178,7 +1178,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
           // Deprecated - to be removed
           // Add compartment restriction for the access policy.
           expressions.push(new Condition('compartments', 'ARRAY_CONTAINS', policyCompartmentId, 'UUID[]'));
-        } else if (policy.criteria) {
+        } else if (policy.criteria?.startsWith(policy.resourceType + '?')) {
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -1178,7 +1178,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
           // Deprecated - to be removed
           // Add compartment restriction for the access policy.
           expressions.push(new Condition('compartments', 'ARRAY_CONTAINS', policyCompartmentId, 'UUID[]'));
-        } else if (policy.criteria) {
+        } else if (policy.criteria?.startsWith(policy.resourceType + '?')) {
```
ThatOneBro, line 1181: Should we add a case for logging when an AccessPolicy has an invalid criteria so we can take stock on who's AccessPolicy's are currently broken?

ThatOneBro (approved):

PR #5413, opened 2024-10-21
feat(useSearch): add debounce to `useSearch`

### LINKED ISSUE #5402: Add debounce to `useSearch` and related search hooks
Currently changing FHIR search query input to these hooks constantly retriggers search without debounce, so the re-render velocity of the search determines how fast searches are sent out

ThatOneBro:
Fixes #5402

### DIFF
--- a/packages/react-hooks/src/useDebouncedValue/useDebouncedValue.ts
+++ b/packages/react-hooks/src/useDebouncedValue/useDebouncedValue.ts
@@ -1,5 +1,5 @@
 /*
-  This hook was forked from: https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/hooks/src/use-debounced-value/use-debounced-value.ts
+  This hook was forked from: https://github.com/mantinedev/mantine/blob/fbcee929e0b11782092f48c1e7af2a1d1c878823/packages/%40mantine/hooks/src/use-debounced-value/use-debounced-value.ts
   and has the following license:
 
   MIT License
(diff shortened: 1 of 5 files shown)


mattwiller (approved):

packages/react-hooks/src/useSearch/useSearch.ts
```diff
@@ -44,34 +52,41 @@ export function useSearchOne<K extends ResourceType>(
...
   const medplum = useMedplum();
-  const [searchKey, setSearchKey] = useState<string>();
+  const [lastSearchKey, setLastSearchKey] = useState<string>();
   const [loading, setLoading] = useState<boolean>(true);
-  const [result, setResult] = useState<ReturnType>();
+  const [result, setResult] = useState<SearchReturnType>();
   const [outcome, setOutcome] = useState<OperationOutcome>();
 
+  const searchKey = medplum.fhirSearchUrl(resourceType, query).toString();
+  const [debouncedSearchKey] = useDebouncedValue(searchKey, options?.debounceMs ?? DEFAULT_DEBOUNCE_MS, {
```
mattlong, line 79: Adding debouncing by default feels like too big of a breaking change. I think we should have it disabled by default and make a v4 issue to enable debouncing by default.

mattwiller: I don't totally agree that this is breaking: I would argue that instant search queries for all inputs isn't really a part of the contract of the hook

packages/react-hooks/src/useDebouncedValue/useDebouncedValue.ts
```diff
@@ -0,0 +1,59 @@
...
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+*/
+
+import { useCallback, useEffect, useRef, useState } from 'react';
+
```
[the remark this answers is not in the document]
ThatOneBro: Good feedback

packages/react-hooks/src/useSearch/useSearch.ts
ThatOneBro: As discussed offline, concerns about breaking customer functionality are fair. We decided to drop the default debounce time to 250ms, since it's much less likely to cause perceivable differences in functionality while still retaining some inherent performance wins for the default settings.

Also worth pointing out that we are using the `leading` option for the `useDebouncedValue` hook from `@mantine/hooks` which applies the first change in the value in a string of updates immediately, so that means for only one change in the search, the search is still immediate and it's only when the query parameters change rapidly that they are debounced by the 250ms.

mattlong (approved):

codyebberson (approved):

PR #5419, opened 2024-10-24
docs(client): add docs about `MedplumClient` lifecycle

### LINKED ISSUE #5155: Document "state machine" around `MedplumClient.isLoading()`, `useMedplumProfile()`, and auth lifecycle events
We should also consider fleshing out semantics around `signedIn` vs. `loading` and enumerate use cases for each.

Separately (I'll create a discussion about this later):
`MedplumClient.isLoading()` is currently used in examples to prevent render before being authed, but with our new events that keep the `MedplumProvider` more in sync with `MedplumClient`, we read `isLoading()` more frequently, and in many cases it is now evaluating to `true`, ie. when a profile is just being refreshed. This causes unmounting in many cases and is probably not the behavior that most users would expect, and in many cases it is an unintentional downstream breaking change.

We should have mechanic in place for making sure that the client is "ready" and later on making sure that we have a valid user logged in possibly separate from if we have a valid profile we can use (in the case of a pending refresh).

ThatOneBro:
Closes #5155

### DIFF
--- a/packages/docs/docs/react/lifecycle-events.mdx
+++ b/packages/docs/docs/react/lifecycle-events.mdx
@@ -21,7 +21,7 @@ These are usually used for triggering rerenders within the context of React, but
 
 ## Working with the `MedplumClient` Lifecycle in React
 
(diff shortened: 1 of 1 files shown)


codyebberson (approved):
ty 🙏 

Everyone loves docs 😛

PR #5421, opened 2024-10-24
Support string values in batch PATCH requests

sam-silv:
Running into an issue where we're trying to update the `status` of a `Communication` in a batch request, but the handler at present is only accepting objects/arrays due to the `JSON.stringify` always being applied

## Example request

```json
{
  "resourceType": "Bundle",
  "type": "transaction",
  "entry": [
    {
      "resource": {
        "resourceType": "Parameters",
        "parameter": [
          {
            "name": "operation",
            "part": [
              { "name": "op", "valueCode": "replace" },
              { "name": "path", "valueString": "/status" },
              { "name": "value", "valueString": "completed" }
            ]
          }
        ]
      },
      "fullUrl": "Communication/d33411f7-3370-4efa-a7ad-a8a7d1ecb3a5",
      "request": {
        "method": "PATCH",
        "url": "Communication/d33411f7-3370-4efa-a7ad-a8a7d1ecb3a5"
      }
    },
    // ... other operations
  ]
}
```

## Expected

200 on the PATCH

## Actual

```json
{
    "resourceType": "Bundle",
    "type": "transaction-response",
    "entry": [
        {
            "response": {
                "outcome": {
                    "issue": [
                        {
                            "severity": "error",
                            "code": "invalid",
                            "details": {
                                "text": "Unexpected token 'c', \"completed\" is not valid JSON"
                            }
                        }
                    ]
                },
                "status": "400"
            }
        }
    ]
}
```

### DIFF
--- a/packages/fhir-router/src/batch.test.ts
+++ b/packages/fhir-router/src/batch.test.ts
@@ -615,7 +615,7 @@ describe('Batch', () => {
                 name: 'operation',
                 part: [
                   { name: 'op', valueCode: 'replace' },
-                  { name: 'path', valueString: '/gender', },
+                  { name: 'path', valueString: '/gender' },
                   { name: 'value', valueString: 'female' },
                 ],
               },
--- a/packages/fhir-router/src/batch.test.ts
+++ b/packages/fhir-router/src/batch.test.ts
@@ -578,6 +578,7 @@ describe('Batch', () => {
   test('Process batch patch Parameters', async () => {
     const patient = await repo.createResource<Patient>({
       resourceType: 'Patient',
+      birthDate: '2000-01-01',
       gender: 'unknown',
     });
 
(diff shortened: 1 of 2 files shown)


codyebberson (commented):
@mattwiller - any comments on correctness here?

In particular, whether we believe there is a contract for JSON vs string in `valueString`.

The PR, as written, seems better than assuming JSON, but it does create some ambiguity around the parsing behavior.

If we don't like this, we could just say that to use `valueString` with an actual string value requires including the surrounding quotes (i.e., `valueString: '"female"'`)

That is somewhat counterintuitive though.

The heavyweight option would be to try to infer if the `path` resolves to a string element or an object element, but that seems fraught....

mattwiller:
The intended contract is that all values in `valueString` are stringified JSON, which isn't [as well-documented as it could be](https://www.medplum.com/docs/fhir-datastore/fhir-batch-requests#patch-actions-in-a-batch).  This is necessary to avoid parsing ambiguities around values like `true` or `false`, so I would advocate for sticking with the current behavior.  I put together a prospective PR to [improve the documentation](https://github.com/medplum/medplum/pull/5425)

sam-silv:
makes sense! totally missed that you could stringify a string – I'll give that a shot

PR #5422, opened 2024-10-24
Fix documentation links

darthdie:
Fixes a number of dead links.

| Before | After |
|--------|--------|
| `https://docs.medplum.com/` | `https://www.medplum.com/docs/` |
| `https://github.com/medplum/medplum-demo-bots/blob/main/src/examples/*` | `https://github.com/medplum/medplum-demo-bots/blob/main/src/*` |
| `https://docs.medplum.com/storybook/index.html?` | `https://storybook.medplum.com/` | 

I also changed the header of the Bots index page from `Intro` to `Bots`.
This was to align it with the other index pages, and I also personally felt that it read odd to have the first sentence of an Intro section talking about reading the intro on another page 🙂

### DIFF
--- a/examples/foomedical/README.md
+++ b/examples/foomedical/README.md
@@ -78,7 +78,7 @@ Contact the medplum team ([support@medplum.com](mailto:support@medplum.com) or [
 
 ### Data Setup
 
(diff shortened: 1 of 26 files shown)


codyebberson (approved):
Awesome, thank you @darthdie !

PR #5437, opened 2024-10-28
Use FHIR stringify in server responses

mattlong:
Previously, there were cases in which empty text/array fields were not being properly removed from FHIR server responses due to the particulars of the caching logic. This is a more robust fix that ensures all FHIR responses use the appropriate JSON-serialization method.

To revert to the legacy `JSON.stringify`, a super admin can include the following entry in `Project.systemSetting`:

```js
{ name: 'legacyFhirJsonResponseFormat', valueBoolean: true }
```

### DIFF
--- a/packages/server/src/fhir/routes.ts
+++ b/packages/server/src/fhir/routes.ts
@@ -1,4 +1,4 @@
-import { allOk, ContentType, isOk, OperationOutcomeError } from '@medplum/core';
+import { allOk, ContentType, isOk, OperationOutcomeError, stringify } from '@medplum/core';
 import { BatchEvent, FhirRequest, FhirRouter, HttpMethod, RepositoryMode } from '@medplum/fhir-router';
 import { ResourceType } from '@medplum/fhirtypes';
 import { NextFunction, Request, Response, Router } from 'express';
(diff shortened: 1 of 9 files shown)


codyebberson (approved):

mattwiller (approved):

ThatOneBro (approved):

packages/server/src/fhir/routes.ts
```diff
@@ -71,7 +71,23 @@ fhirRouter.use((req: Request, res: Response, next: NextFunction) => {
...
     }
 
-    return res.json(data);
+    let legacyFhirJsonResponseFormat: boolean | undefined;
+    try {
+      const ctx = getAuthenticatedContext();
+      legacyFhirJsonResponseFormat = ctx.project.systemSetting?.find(
+        (s) => s.name === 'legacyFhirJsonResponseFormat'
+      )?.valueBoolean;
+    } catch (_err) {
```
ThatOneBro, line 80: Why do we ignore errors here? This is just because we are not guaranteed to have an authenticated context here?

ThatOneBro (approved):

packages/server/src/fhir/routes.ts
mattlong: Yep, there were a couple tests that pass through this while unauthenticated. I'll update the comment to clairify.

PR #5439, opened 2024-10-29
Combine resource account references from multiple sources

mattwiller:


ianplunkett (approved):
I've tested this branch locally against the scenario described in this issue - https://github.com/medplum/medplum-customer-tasks/issues/227 and observed the desired results.  

This looks good to me.

packages/server/src/fhir/repo.ts
```diff
@@ -1702,42 +1702,60 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    if (!existing && this.context.accessPolicy?.compartment?.reference) {
+      // If the creator's access policy specifies a compartment, then use it as the account.
+      // The writer's access policy is only applied at resource creation: simply editing a
+      // resource does NOT pull it into the user's account.
+      accounts.add(this.context.accessPolicy.compartment.reference);
     }
 
-    if (updated.resourceType !== 'Patient') {
+    if (updated.resourceType === 'Patient') {
+      // Otherwise, default to the existing value.
```
codyebberson, line 1725: Minor nit, I think this comment is a copy/paste artifact.

```suggestion
      // When examining a Patient resource, we only look at the individual patient
      // We should not call `getPatients` and `readReference`
```

packages/server/src/fhir/repo.ts
```diff
@@ -1702,42 +1702,60 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+      if (existingAccounts?.length) {
+        for (const account of existingAccounts) {
+          accounts.add(account.reference as string);
+        }
+      }
+    } else {
       for (const patientRef of getPatients(updated)) {
         try {
           // If the resource is in a patient compartment, then lookup the patient.
           const patient = await getSystemRepo().readReference(patientRef);
```
codyebberson, line 1736: Minor nit, we could now use `readReferences` to make this a batch call.  Low priority, that can be a future optimization.

codyebberson (approved):
Some optional minor nits

PR #5442, opened 2024-10-29
docs(useSubscription): add docs page for `useSubscription`

### LINKED ISSUE #5239: Add thorough docs for `useSubscription` hook

ThatOneBro:
Fixes #5239


packages/docs/docs/react/use-subscription.md
```diff
@@ -0,0 +1,235 @@
+---
+sidebar_position: 3
```
mattwiller, line 2: Not sure what's up with the sidebar order, but this page currently shows up first in the preview — we may want to check the numbering for all pages in this section and make sure they're sensible

packages/docs/docs/react/use-subscription.md
```diff
@@ -0,0 +1,235 @@
...
+tags:
+  - subscription
+---
+
+# `useSubscription` Hook
+
+`useSubscription` creates an in-memory `Subscription` resource with the given criteria on the Medplum server and calls the 
+given callback when an event notification is triggered by a resource interaction over a WebSocket connection.
```
mattwiller, line 10: Minor suggestion for clarity/reducing run-on sentence
```suggestion
`useSubscription` creates an in-memory `Subscription` resource on the Medplum server with the given criteria and calls the 
given callback when an event notification is triggered.
```

packages/docs/docs/react/use-subscription.md
```diff
@@ -0,0 +1,235 @@
...
+## Use Cases
+
+The `useSubscription` hook is a powerful tool for creating applications that require the client to wait for events asynchronously from the server.
+
+Some examples include:
+- Live chat
+- Notification badges for tasks
+- Realtime analytics dashboards
+
+The semantics of the hook make it extremely simple to subscribe to events for resource interactions that satisfy a specified criteria and act
```
mattwiller, line 24:
```suggestion
The hook makes it extremely simple to listen for resource interactions that satisfy a specified criteria and act
```

packages/docs/docs/react/use-subscription.md
```diff
@@ -0,0 +1,235 @@
...
+on them reactively, rather than having to poll via expensive search requests on a timer.
+
+## Usage
+
+The `useSubscription` hook takes a [FHIR search criteria](https://hl7.org/fhir/R4/search.html) and a callback function to call when a resource interaction
```
mattwiller, line 29: I would suggest linking to our own page instead (`/docs/search/basic-search`), which itself links to the FHIR spec

packages/docs/docs/react/use-subscription.md
```diff
@@ -0,0 +1,235 @@
...
+  );
+
+  return <div>Notifications received: {notificationCount}</div>;
+}
+```
+
+### Parsing the Subscription `Bundle`
+
+The callback receives a `Bundle` which contains a `SubscriptionStatus` as its first entry (`bundle.entry[0].resource`) 
+and the resource for which the `Subscription` fired for in its second entry (`bundle.entry[1].resource`).
```
mattwiller, line 52: It might be more useful to give a full example of the JSON received by the callback, showing what both Bundle entries actually look like

mattwiller (approved):

packages/docs/docs/react/use-subscription.md
ThatOneBro: The numbering is kind of arbitrary indicator of roughly what order the pages should appear in. 3 is technically first right now but adding an actual 1 would demote its position since 1 is obviously lower. 

I haven't seen a clear way for deciding the numbers, to me it's kinda like hinting "this should probably go last even if there are more pages in the future". 🤷 Maybe should do some research on what other projects do

codyebberson (approved):

PR #5445, opened 2024-10-30
S3 Native Presigned URLs

ivank:
Adding pre-signed URLs implementation using s3 native functionality so that alternative (non AWS) deployments can be achieved

Medplum is nodejs server + postgres + redis plus s3, with a CDN in front. 
But you can deploy all of that not just on AWS but on other clouds.
Lots of clouds offer postgres/redis/s3-compatible offerings, notably GCP.
With this small change we allow the usage of native pre-signed s3 functionality so that it can be deployed in many more arrangements than just AWS

This is the implementation we envision for the issue #4815

We're using this change in production with our deployment in GCP (we've forked medplum with only this specific change) and we keep it in sync with upstream.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2706,6 +2706,25 @@
         "node": ">=16.0.0"
       }
     },
+    "node_modules/@aws-sdk/s3-request-presigner": {
+      "version": "3.705.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/s3-request-presigner/-/s3-request-presigner-3.705.0.tgz",
+      "integrity": "sha512-dAQiXv/TqjEUCoEeiKqQGI8LJ3g8Xv+XJL4W9CwsB6ZHHDq0Q05ulpDSkhhCf52ySXf5dJ33e1o/VeUDY3q0pw==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-sdk/signature-v4-multi-region": "3.696.0",
+        "@aws-sdk/types": "3.696.0",
+        "@aws-sdk/util-format-url": "3.696.0",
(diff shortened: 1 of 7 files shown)


codyebberson:
Thanks @ivank !

I personally have limited GCP experience, so my initial reaction to this was to propose a GCP Cloud CDN.  However, we spoke with some of our partners with deep experience in GCP, and they agreed with your PR, that this is the correct approach.

In AWS, serving protected content typically uses S3 + CloudFront with presigned URLs because:
1. CloudFront provides edge caching
2. S3 isn't optimized for direct user access

However, GCP's architecture is different:
1. GCS is designed for direct user access and has built-in global load balancing
2. For protected, user-specific content (like PDFs, documents), edge caching provides minimal benefit since:
   - Each user typically downloads their files once
   - Files are usually accessed by single users, not shared
   - Content changes frequently as users upload new versions

Your implementation using direct GCS signed URLs is the recommended approach here. It's simpler to maintain and provides good performance for this use case. If we later discover specific performance needs, we can add Cloud CDN, but current GCS performance characteristics should be sufficient.

LGTM 👍

ivank:
Amazing! Thanks for spending time checking this out. I'm sure you have a busy schedule.

We've been using this for awhile (the time since the issue was raised) and have been satisfied with the performance so far, though our scale is not too big for now.

I've not done any changes in the documentation to describe this change as I'm not sure how you guys would like to approach this, and I don't want to pollute the docs with this edge case, but would be happy to modify it if you deem it appropriate. 

Or we can always merge this and you guys can document it before release.

codyebberson (approved):

ivank:
This PR would help #5622 Though it might be good idea to add it the terraform config as well

PR #5447, opened 2024-10-30
Note that referential integrity is not supported for deletes

### LINKED ISSUE #5446: Document that referential integrity is not supported for delete operations
User generated: https://discord.com/channels/905144809105260605/1273741912737579069/1273741912737579069
Add a note to the [deleting data docs](https://www.medplum.com/docs/fhir-datastore/deleting-data#delete-operation) that referential integrity is not supported due to performance issues

ksmith94:
Fixes #5446 Adds a note to the deleting data docs that referential integrity is not supported when deleting data.

### DIFF
--- a/packages/docs/docs/fhir-datastore/deleting-data.md
+++ b/packages/docs/docs/fhir-datastore/deleting-data.md
@@ -29,7 +29,7 @@ However, the original content of the resource is not destroyed. It can still be 
 
 :::caution Referential Integrity on Deletes
 
-Referential integrity is **not** supported for deletes at this time as it would require checking every resource in a project for inbound references to the deleted resource. Unfortunately, performance tradeoffs forbid Medplum from implementing this.
+Referential integrity is **not** supported for deletes at this time.
 
 :::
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/fhir-datastore/deleting-data.md
```diff
@@ -27,6 +27,12 @@ However, the original content of the resource is not destroyed. It can still be
 - Reading resource history: `GET Patient/123/_history`
 - Reading a resource version: `GET Patient/123/_history/1`
 
+:::caution Referential Integrity on Deletes
+
+Referential integrity is **not** supported for deletes at this time as it would require checking every resource in a project for inbound references to the deleted resource. Unfortunately, performance tradeoffs forbid Medplum from implementing this.
```
codyebberson, line 32:
```suggestion
Referential integrity is **not** supported for deletes at this time.
```

Once upon a time, we did not support it for performance reasons.  Our database architecture has changed a lot since we first added basic support for referential integrity, and it would now be more feasible for us to implement.  It is still fair to call out that it is unsupported, but I don't think we should point to performance issues .

ThatOneBro (approved):

PR #5449, opened 2024-10-30
feat: add Questionnaire Bot interaction dropdown

### LINKED ISSUE #5294: Allow users to configure `create` only `Subscription` for `Questionnaire` bots from `@medplum/app`
Currently the default for bots associated to `Questionnaire` via the automatic `Subscription` creation in the `Questionnaire`-related UI will create a `Subscription` that will trigger not only on creation of a `QuestionnaireResponse` associated with the given `Questionnaire` but also on any update to those `QuestionnaireResponse`s that may happen at a later time.

This may not be the assumed behavior of most users authoring Questionnaires and it should be easier to get the assumed behavior, and as a separate question, it may be better if `create only` trigger was the default.

The proposed UI change would be a dropdown where you can select the `Trigger Event` for this `Subscription`; something like `All Events`, `Update Only`, `Create Only`, etc. would be the options. The selection of this value would then add the appropriate extension (or not add any in the case of `All Events` trigger being selected) without the user having to go in and manually update the `Subscription` resource.

ThatOneBro:
Fixes #5294

### DIFF
--- a/packages/app/src/resource/QuestionnaireBotsPage.tsx
+++ b/packages/app/src/resource/QuestionnaireBotsPage.tsx
@@ -89,7 +89,7 @@ export function QuestionnaireBotsPage(): JSX.Element {
         <Group>
           <NativeSelect
             name="subscription-trigger-event"
-            defaultValue={'Create Only'}
+            defaultValue="Create Only"
             label="Subscription Trigger Event"
             data={SUBSCRIPTION_INTERACTION_KEYS}
             onChange={(event) =>
--- a/packages/app/src/resource/QuestionnaireBotsPage.tsx
+++ b/packages/app/src/resource/QuestionnaireBotsPage.tsx
@@ -1,4 +1,4 @@
-import { Button, Group, Title } from '@mantine/core';
+import { Button, Divider, Group, InputLabel, NativeSelect, Stack, Title } from '@mantine/core';
 import { showNotification } from '@mantine/notifications';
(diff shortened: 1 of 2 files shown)


packages/app/src/resource/ResourcePage.test.tsx
```diff
@@ -155,6 +157,109 @@ describe('ResourcePage', () => {
...
+      jest.advanceTimersByTime(1000);
+    });
+
+    // Wait for the drop down
+    await act(async () => {
+      jest.advanceTimersByTime(1000);
+    });
+  });
+
+  test.only('Questionnaire bots -- all interactions', async () => {
```
codyebberson, line 190: Remove `only`

packages/app/src/resource/QuestionnaireBotsPage.tsx
```diff
@@ -42,21 +66,38 @@ export function QuestionnaireBotsPage(): JSX.Element {
...
+          Connect to bot
+        </InputLabel>
+        <Group>
+          <ResourceInput name="bot" resourceType="Bot" onChange={(r) => setConnectBot(r as Bot)} />
+          <Button onClick={connectToBot}>Connect</Button>
+        </Group>
+        <Group>
+          <NativeSelect
+            name="subscription-trigger-event"
+            defaultValue={'Create Only'}
```
codyebberson, line 92:
```suggestion
            defaultValue="Create Only"
```

codyebberson (approved):
lgtm.  let's remove the `only` though

PR #5455, opened 2024-10-30
Case-insensitive search on ContactPoint

### LINKED ISSUE #5424: Practitioners with Uppercase characters in their email won't map to Users with the same
If you create a Practitioner with an email that contains uppercase letters like

```
{
    "resourceType": "Practitioner",
    "meta": {
        "accounts": [
            {
                "reference": "Organization/A"
            }
        ]
    },
    "name": [
        {
            "given": [
                "Practitioner"
            ],
            "family": "A.1"
        }
    ],
    "telecom": [
        {
            "system": "email",
            "use": "work",
            "value": "Practitioner.A.1@example.com"
        }
    ]
}
```

Then create a user also using uppercase letters in the email 

```
{
    "resourceType": "Practitioner",
    "firstName": "Practitioner",
    "lastName": "A.1",
    "email": "Practitioner.A.1@example.com",
    "sendEmail": "false",
    "password": "foobar"
}
```

These accounts won't be matched together. 

I think the logic to create a user converts the email to lowercase whereas the logic to create a Practitioner does not.

mattlong:
`ContactPoint` token search parameters are now treated as case-insensitive so that searching for `upper@example.com` matches against an email address entered as `UPPER@example.com` in a resource.

Reindexing all objects is required to take full advantage of this change, but it is backwards compatible such that the previous/current case-sensitive search is still supported.

Fixes #5424

### DIFF
--- a/packages/server/src/fhir/lookups/token.ts
+++ b/packages/server/src/fhir/lookups/token.ts
@@ -109,10 +109,17 @@ export class TokenTable extends LookupTable {
    * @param _selectQuery - The select query builder.
    * @param resourceType - The resource type.
    * @param table - The resource table.
+   * @param param - The search parameter.
    * @param filter - The search filter details.
    * @returns The select query where expression.
    */
-  buildWhere(_selectQuery: SelectQuery, resourceType: ResourceType, table: string, filter: Filter): Expression {
+  buildWhere(
+    _selectQuery: SelectQuery,
+    resourceType: ResourceType,
+    table: string,
+    param: SearchParameter,
+    filter: Filter
+  ): Expression {
     const lookupTableName = this.getTableName(resourceType);
 
     const conjunction = new Conjunction([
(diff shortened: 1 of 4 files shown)


packages/server/src/fhir/lookups/token.ts
```diff
@@ -465,27 +504,38 @@ function buildWhereCondition(tableName: string, operator: FhirOperator, query: s
...
     ]);
   } else if (operator === FhirOperator.CONTAINS) {
     getLogger().warn('Potentially expensive token lookup query', { operator });
-    return new Condition(column, 'LIKE', escapeLikeString(value.trim()) + '%');
+    return new Condition(column, 'LIKE', escapeLikeString(value) + '%');
+  } else if (caseSensitive) {
+    return new Condition(column, '=', value);
   } else {
-    return new Condition(column, '=', value.trim());
+    // In Medplum v4, or when there is a guarantee all resources have been reindexed, the IN (...) can be
```
ThatOneBro, line 536: 👍 sounds good to me

packages/server/src/fhir/search.test.ts
```diff
@@ -3397,6 +3397,27 @@ describe('FHIR Search', () => {
         expect(result.entry?.length).toBe(0);
       }));
 
+    test('Practitioner by email case insensitive', () =>
```
ThatOneBro, line 3400: Do we have a similar test for the case sensitive case matching exact casing and not the same word with different casing?

ThatOneBro (approved):
LGTM, one comment about potentially adding a test for a case if it doesn't already exist

packages/server/src/fhir/search.test.ts
mattlong: Added one just incase

PR #5457, opened 2024-10-30
Enable chained search over canonical references

mattwiller:
Canonical references differ from regular `Reference` values: they are bare `string` values that must correspond with the `url` field of the target (referenced) resource.  Chained search over canonical references is handled in the database by a direct `JOIN` to the next resource table, using string equality on the `url` column

### DIFF
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -24,6 +24,7 @@ import {
   Condition,
   DiagnosticReport,
   Encounter,
+  EvidenceVariable,
   Goal,
   HealthcareService,
   Location,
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/search.ts
```diff
@@ -1367,55 +1367,83 @@ function buildChainedSearchUsingReferenceTable(
...
+    join = new Condition(new Column(currentTable, 'url'), eq, new Column(nextTableAlias, link.details.columnName));
+  } else {
+    join = new Condition(new Column(nextTableAlias, 'url'), eq, new Column(currentTable, link.details.columnName));
+  }
+
+  selectQuery.join('INNER JOIN', link.resourceType, nextTableAlias, join);
+  return nextTableAlias;
+}
+
+function linkLiteralReference(
```
ThatOneBro, line 1405: Looks like we just extracted previous implementation into this function 👍

packages/server/src/fhir/search.test.ts
```diff
@@ -1866,6 +1867,41 @@ describe('FHIR Search', () => {
...
+        const ev = randomUUID();
+        const evidenceVariable = await repo.createResource<EvidenceVariable>({
+          resourceType: 'EvidenceVariable',
+          status: 'unknown',
+          relatedArtifact: [{ type: 'derived-from', resource: url }],
+          identifier: [{ value: ev }],
+        });
+
+        const result = await repo.search(
+          parseSearchRequest(`Questionnaire?_has:EvidenceVariable:derived-from:identifier=${ev}`)
```
ThatOneBro, line 1892: 👍 Definitely using canonical reference in the chain

packages/server/src/fhir/search.ts
```diff
@@ -1367,55 +1367,83 @@ function buildChainedSearchUsingReferenceTable(
...
+        currentResourceType as ResourceType,
+        currentTable,
         link.filter
       );
     }
   }
   throw new OperationOutcomeError(badRequest('Unterminated chained search'));
 }
 
+function linkCanonicalReference(selectQuery: SelectQuery, currentTable: string, link: ChainedSearchLink): string {
```
ThatOneBro, line 1390: Surprisingly tiny function for handling canonical reference case, nice

ThatOneBro (approved):

PR #5460, opened 2024-10-31
Ignore conflicts in reference lookup table

mattwiller:


packages/server/src/fhir/lookups/reference.ts
```diff
@@ -33,6 +34,28 @@ export class ReferenceTable extends LookupTable {
...
+    values: Record<string, any>[]
+  ): Promise<void> {
+    if (values.length === 0) {
+      return;
+    }
+    const tableName = this.getTableName(resourceType);
+
+    // Reference lookup tables have a covering primary key, so a conflict means
+    // that the exact desired row already exists in the database
+    const insert = new InsertQuery(tableName, values).ignoreOnConflict();
```
ThatOneBro, line 56: Nice, very clean

Do you mind adding a test to check that `pg` doesn't actually emit an error in this case? (I think that's what was happening previously right?)

ThatOneBro (approved):

packages/server/src/fhir/lookups/reference.ts
mattwiller: The conflict is relatively difficult to reproduce reliably in a test, since it's effectively a race condition in the database

PR #5461, opened 2024-10-31
Add opt-in `Project.systemSetting` to perform searches on readers when available

mattlong:
Previously, only GraphQL queries (except mutations) would be sent to readers. With this, searches can also be sent to readers if configured that way by a super-admin.

### DIFF
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -72,8 +72,13 @@ async function batch(req: FhirRequest, repo: FhirRepository, router: FhirRouter)
 }
 
 // Search
-async function search(req: FhirRequest, repo: FhirRepository): Promise<FhirResponse> {
-  setSearchRepositoryMode(req, repo);
+async function search(
+  req: FhirRequest,
+  repo: FhirRepository,
+  _router: FhirRouter,
+  options?: FhirRouteOptions
+): Promise<FhirResponse> {
+  setSearchRepositoryMode(req, repo, options);
 
   const { resourceType } = req.params;
   const bundle = await repo.search(parseSearchRequest(resourceType as ResourceType, req.query));
@@ -81,8 +86,13 @@ async function search(req: FhirRequest, repo: FhirRepository): Promise<FhirRespo
 }
 
 // Search multiple types
-async function searchMultipleTypes(req: FhirRequest, repo: FhirRepository): Promise<FhirResponse> {
-  setSearchRepositoryMode(req, repo);
+async function searchMultipleTypes(
+  req: FhirRequest,
+  repo: FhirRepository,
+  _router: FhirRouter,
+  options?: FhirRouteOptions
+): Promise<FhirResponse> {
+  setSearchRepositoryMode(req, repo, options);
 
   const searchRequest = parseSearchRequest('MultipleTypes' as ResourceType, req.query);
   if (!searchRequest.types || searchRequest.types.length === 0) {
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/routes.test.ts
```diff
@@ -441,11 +452,23 @@ describe('FHIR Routes', () => {
     expect(res.status).toBe(200);
   });
 
-  test('Search', async () => {
+  test.each<['writer' | 'reader']>([['writer'], ['reader']])('Search on %s', async (repoMode) => {
```
mattwiller, line 457: Rather than repeating this `.each()` on every test, I'd recommend wrapping the tests in a `describe.each()`

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -89,12 +94,21 @@ async function searchMultipleTypes(req: FhirRequest, repo: FhirRepository): Prom
...
+
   const { resourceType } = req.params;
   const query = req.body as Record<string, string[] | string | undefined>;
   const bundle = await repo.search(parseSearchRequest(resourceType as ResourceType, query));
   return [allOk, bundle];
 }
 
+function setSearchRepositoryMode(req: FhirRequest, repo: FhirRepository): void {
+  if (req.config?.searchOnReader) {
+    console.log('Setting search mode to READER');
```
mattwiller: Remove?

packages/fhir-router/src/graphql/graphql.ts
```diff
@@ -117,6 +117,8 @@ export async function graphqlHandler(
 
   if (includesMutations(query)) {
     repo.setMode(RepositoryMode.WRITER);
+  } else {
+    repo.setMode(RepositoryMode.READER);
```
codyebberson, line 121: 🤔 

I'm trying to remember why we used the original implementation...

I think the idea was "if we're already in `writer` mode, then we need to stay in `writer` mode"

In practice, that's probably not something that happens (?)  Maybe if someone tries to run a GraphQL request inside of a batch request?  I'm not sure if we actually even support that?

codyebberson (approved):
Overall looks good.  See comment on moving the setMode call into graphql.ts

mattwiller (approved):
Just curious, why the change to opt-in for this?

packages/server/src/fhir/routes.ts
```diff
@@ -314,17 +314,11 @@ protectedRoutes.use(
         graphqlMaxDepth: ctx.project.systemSetting?.find((s) => s.name === 'graphqlMaxDepth')?.valueInteger,
         graphqlMaxPageSize: ctx.project.systemSetting?.find((s) => s.name === 'graphqlMaxPageSize')?.valueInteger,
         graphqlMaxSearches: ctx.project.systemSetting?.find((s) => s.name === 'graphqlMaxSearches')?.valueInteger,
+        searchOnReader: ctx.project.systemSetting?.find((s) => s.name === 'searchOnReader')?.valueBoolean,
         transactions: ctx.project.features?.includes('transaction-bundles'),
       },
     };
 
-    if (request.url.includes('$graphql')) {
```
ThatOneBro, line 321: Does this mean that GraphQL searches will no longer go to the reader unless configured to at the project level?

ThatOneBro: Oh I see, it just happens in `graphql.ts` now. Nevermind

ThatOneBro:
> Just curious, why the change to opt-in for this?

Yeah I understand the desire to be able to control projects going to the readers for search maybe (in the case of saturating the connection pool with lots of FHIR searches and graphql requests) but feel like maybe it should be opt-out?

Not a strongly held opinion though

mattlong:
@mattwiller @ThatOneBro Regarding opt-in vs opt-out, the main concern is that replication-lag between writer and readers can be non-trivial (e.g. ~20 ms) even in the steady state and can spike higher for various reasons so we want to limit the blast radius of reading inconsistent data as much as possible for now.

PR #5463, opened 2024-10-31
Update 'Install on Ubuntu' docs

codyebberson:
* Update to Postgres 16
* Update to Node 22
* Added optional section for nginx and certbot

Setup a full working test cluster here: https://app.nginx.medplum.dev/

### DIFF
--- a/packages/docs/docs/self-hosting/install-on-ubuntu.md
+++ b/packages/docs/docs/self-hosting/install-on-ubuntu.md
@@ -4,7 +4,7 @@ sidebar_position: 2
 
 # Install on Ubuntu
 
-This guide was written for Ubuntu 20.04, although it should work for most recent versions.
+This guide was last updated for Ubuntu 24.04, although it should work for most recent versions.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/self-hosting/install-on-ubuntu.md
```diff
@@ -126,12 +126,167 @@ Install dependencies
...
+
+```nginx
+# /etc/nginx/sites-available/api.example.com
+server {
+    listen 80;
+    listen [::]:80;
+    server_name api.example.com;
+
+    # Redirect HTTP to HTTPS
+    location / {
```
ThatOneBro, line 230: Nice, the auto redirect 👍

packages/docs/docs/self-hosting/install-on-ubuntu.md
```diff
@@ -126,12 +126,167 @@ Install dependencies
...
+        return 301 https://$server_name$request_uri;
+    }
+}
+
+server {
```
ThatOneBro, line 235: Did we make sure this works for WebSockets? When I get back to my laptop I can check but I remember last time I set up nginx with WebSockets I had to do something special

ThatOneBro: Actually I think this is fine, I think I might be thinking about tuning buffers which is a separate issue. LGTM

ThatOneBro (approved):

PR #5469, opened 2024-11-01
Default to legacy FHIR JSON response format

mattlong:


mattlong:
The team decided not to revert the new default

PR #5471, opened 2024-11-01
More updates to 'Install on Ubuntu'

codyebberson:
1. Add more details and some "info" and "warning" blocks
2. Changed `app` from dev server to preview server
   * Originally my plan was to go all the way and let nginx serve the files directly from disk, which would be faster and generally better
   * However, the instructions got unwieldy due to all kinds of extra requirements regarding file permissions, directory structure, SPA fallback, etc
   * So instead I settled on using the preview server, which is probably fine, even in "production", because the app server is only hit on initial page load, and is pretty tiny

If we're going to push this forward any further, then I think we're at the point where we should just ship a `.deb` file and maybe an APT repository.

It would be neat to just say:

```bash
sudo apt install medplum
```

### DIFF
--- a/packages/docs/docs/self-hosting/install-on-ubuntu.md
+++ b/packages/docs/docs/self-hosting/install-on-ubuntu.md
@@ -132,7 +132,7 @@ Build the server, app, and necessary dependencies
 npm run build:fast
 ```
 
-## Start Medplum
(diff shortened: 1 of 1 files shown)


ThatOneBro (approved):
Nice, LGTM

PR #5472, opened 2024-11-01
Clean up a lot of as BundleEntry casts

mattlong:


packages/core/src/bundle.ts
```diff
@@ -85,7 +85,7 @@ export function reorderBundle(bundle: Bundle): Bundle {
 
   const entryMap: Record<string, BundleEntry> = {};
 
-  for (const entry of bundle.entry || []) {
+  for (const entry of bundle.entry ?? []) {
```
ThatOneBro, line 88: Nice

packages/core/src/client.ts
```diff
@@ -3261,10 +3261,14 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    */
   private async executeAutoBatch(): Promise<void> {
     // Get the current queue
-    const entries = [...(this.autoBatchQueue as AutoBatchEntry[])];
+    if (this.autoBatchQueue === undefined) {
```
ThatOneBro, line 3264: Niceee

packages/fhir-router/src/batch.ts
```diff
@@ -375,7 +375,7 @@ class BatchProcessor {
     let errors = 0;
 
     for (const entryIndex of bundleInfo.ordering) {
-      const entry = this.bundle.entry?.[entryIndex] as BundleEntry;
+      const entry = entries[entryIndex];
```
ThatOneBro, line 378: Good catch

ThatOneBro (approved):

PR #5473, opened 2024-11-01
fix(mock): make `MockClient` strip empty values for FHIR compliance

ThatOneBro:


packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -459,7 +459,7 @@ export function SearchControl(props: SearchControlProps): JSX.Element {
           )}
         </Table.Tbody>
       </Table>
-      {resources?.length === 0 && (
+      {(resources?.length ?? 0) === 0 && (
```
mattwiller, line 462:
```suggestion
      {!resources?.length && (
```

mattwiller (approved):

PR #5476, opened 2024-11-02
fix(agent): harden reconnect logic

### LINKED ISSUE #5329: Harden Agent reconnect logic
We have some reconnect logic for Medplum Agent and we have a heartbeat protocol for the Agent WebSocket connection but currently the heartbeat being missed does not trigger a reconnect, only a full close of the WebSocket does. We could stitch those two mechanics together to harden the Agent reconnect logic

ThatOneBro:
In this PR, we swap in the `ReconnectingWebSocket` implementation to the Agent's client-side WebSocket, and also add a force reconnect when we detect more than `MAX_MISSED_HEARTBEATS` have been missed. 

Fixes #5329

### DIFF
--- a/packages/core/src/websockets/reconnecting-websocket.ts
+++ b/packages/core/src/websockets/reconnecting-websocket.ts
@@ -477,7 +477,7 @@ export class ReconnectingWebSocket<WS extends IWebSocket = WebSocket>
         this._debug('connect', { url: this._url, protocols: this._protocols });
         this._ws = this._protocols ? new WS(this._url, this._protocols) : new WS(this._url);
 
-        this._ws.binaryType = this._binaryType as BinaryType;
+        this._ws.binaryType = this._binaryType;
         this._connectLock = false;
         this._addListeners();
 
@@ -524,7 +524,7 @@ export class ReconnectingWebSocket<WS extends IWebSocket = WebSocket>
 
     assert(this._ws, 'WebSocket is not defined');
 
-    this._ws.binaryType = this._binaryType as BinaryType;
+    this._ws.binaryType = this._binaryType;
 
     // send enqueued messages (messages sent before websocket open event)
     this._messageQueue.forEach((message) => this._ws?.send(message));
--- a/packages/core/src/websockets/reconnecting-websocket.ts
+++ b/packages/core/src/websockets/reconnecting-websocket.ts
@@ -89,8 +89,9 @@ function cloneEvent(e: Event): Event {
   return new (e as any).constructor(e.type, e);
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/websockets/reconnecting-websocket.ts
```diff
@@ -121,24 +122,27 @@ export type Message = string | ArrayBuffer | Blob | ArrayBufferView;
 
 let didWarnAboutMissingWebSocket = false;
 
-export class ReconnectingWebSocket extends TypedEventTarget<WebSocketEventMap> implements IReconnectingWebSocket {
+export class ReconnectingWebSocket<BType extends string = BinaryType>
```
mattlong, line 125: I'm not very familiar with `WebSocket.BinaryType`, but the fact that it is both 1) used a type generic that gets narrowed in the constructor and 2) is mutable within this class feels a bit off which I think is evident by all the type-casting that is being done around `BType`. Would it be reasonable to get rid of the generic and just have the binaryType getter and setter functions just work with a type `string` or `BinaryType` instead?

Related, looking at the other places in our code that mention binaryType, it's unclear which BinaryType(s) should be settable. Is it really just `"arraybuffer" | "blob"` or should `"nodebuffer" | "fragments"` also be acceptable?

ThatOneBro: It depends on the environment. In the browser `nodebuffer` and `fragments` don't exist

ThatOneBro: The reason the typecasting is done is because the default for `BinaryType` comes from builtin typing but it assumes the browser context

ThatOneBro: Ideally we would be able to pass the `WebSocket` class type as a generic and pull the `binaryType` from there, but the `WebSocket` global type from the built-in types does not include that information until you construct the WebSocket itself (it's just an interface)

ThatOneBro: Okay so I was able to achieve the (almost) ideal solution by creating a generic interface with the default being the global `WebSocket` interface and reading the type of `binaryType` off of the passed in implementation. Let me know if this is better and/or if it makes more sense now what is happening

codyebberson (approved):

packages/core/src/websockets/reconnecting-websocket.ts
mattlong: Nice! Thank you for the explanation; looks great!

mattlong (approved):

PR #5492, opened 2024-11-04
Fix chained search join

mattwiller:


packages/server/src/fhir/search.ts
```diff
@@ -1397,7 +1397,7 @@ function linkCanonicalReference(selectQuery: SelectQuery, currentTable: string,
     join = new Condition(new Column(nextTableAlias, 'url'), eq, new Column(currentTable, link.details.columnName));
   }
 
-  selectQuery.join('INNER JOIN', link.resourceType, nextTableAlias, join);
+  selectQuery.join('LEFT JOIN', link.resourceType, nextTableAlias, join);
```
ThatOneBro, line 1400: Ah... yeah I think that makes sense. Nice catch

ThatOneBro (approved):

PR #5506, opened 2024-11-07
Tighten up JSON vs FHIR JSON response formatting

### LINKED ISSUE #5502: Fix `HGAutocompleteBotResponse` and related HealthGorilla search results when `result` is `undefined`
When the HealthGorilla hooks / components were initially written, the server always returned `result` on the response from the HealthGorilla bot even when the `result` array was empty, but now we are always filtering out empty values and so the components expecting `result` to be there are hitting unhandled cases when there are no results for a search.

mattlong:
The issue that prompted this further cleanup was that Bot execution responses had been getting the FHIR JSON treatment even when they do not return a FHIR resource.

For backwards compatibility, we still assume a FHIR response if the *input* to the bot execution was FHIR JSON. For the future, we should consider having a more explicit way to specify the content types.

Fixes #5502

### DIFF
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -43,7 +43,13 @@ export type FhirRequestConfig = {
   transactions?: boolean;
 };
 
-export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
+export type ContentTypeOverride = string;
+
+// For v4, consider changing FhirResponse from an Array of variable length to an object with optional fields
+export type FhirResponse =
+  | [OperationOutcome]
+  | [OperationOutcome, Resource]
+  | [OperationOutcome, Resource, ContentTypeOverride];
 
 export type FhirRouteOptions = {
   batch?: boolean;
--- a/packages/fhir-router/src/fhirrouter.ts
+++ b/packages/fhir-router/src/fhirrouter.ts
@@ -43,13 +43,14 @@ export type FhirRequestConfig = {
   transactions?: boolean;
 };
 
-export type ContentTypeOverride = string;
+export type FhirResponseOptions = {
+  contentType?: string;
+};
 
-// For v4, consider changing FhirResponse from an Array of variable length to an object with optional fields
 export type FhirResponse =
   | [OperationOutcome]
   | [OperationOutcome, Resource]
-  | [OperationOutcome, Resource, ContentTypeOverride];
+  | [OperationOutcome, Resource, FhirResponseOptions];
 
 export type FhirRouteOptions = {
   batch?: boolean;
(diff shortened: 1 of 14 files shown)


packages/server/src/fhir/operations/execute.ts
```diff
@@ -534,14 +532,16 @@ async function addBotSecrets(
   }
 }
 
+const MIRRORED_CONTENT_TYPES: string[] = [ContentType.FHIR_JSON, ContentType.TEXT, ContentType.HL7_V2];
```
mattlong, line 535: I would be in favor or removing FHIR_JSON from this list so that we never automatically use FHIR stringify on Bot responses; especially since FHIR stringify wouldn't have even been used up until about a week ago.

ThatOneBro: I think this makes sense...

What do we think @mattwiller @codyebberson ?

mattwiller: That would cause some Bots' output to change from `application/fhir+json` to `application/json`, right?  I think that should be a reasonable change, since both MIME types are compatible

ThatOneBro (approved):
So just for readers from the future, here we stopped mirroring `FHIR_JSON` content type in the `Bot/$execute` response by default, since when content type defaults to `FHIR_JSON` in the client which would trigger that, and subsequently after the change in #5437, would strip out empty arrays which for non-FHIR responses is not expected.

I think we have most of the bases covered now. LGTM

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -43,7 +43,13 @@ export type FhirRequestConfig = {
...
 };
 
-export type FhirResponse = [OperationOutcome] | [OperationOutcome, Resource];
+export type ContentTypeOverride = string;
+
+// For v4, consider changing FhirResponse from an Array of variable length to an object with optional fields
+export type FhirResponse =
+  | [OperationOutcome]
+  | [OperationOutcome, Resource]
+  | [OperationOutcome, Resource, ContentTypeOverride];
```
codyebberson, line 52: For future compatibility, how would you feel about making the 3rd arg into an object such as `ExtendedOptions`?  It seems likely that we will want to add more stuff here over time (http headers, cache control, etc)

ThatOneBro: I think we were considering refactoring `FhirResponse` completely for v4 (making it an object)

codyebberson (approved):

packages/fhir-router/src/fhirrouter.ts
codyebberson: Ah, that makes even more sense.

PR #5514, opened 2024-11-09
feat(AsyncJob): add `$cancel` operation for `AsyncJob`s

ThatOneBro:
EDIT:

This PR adds a `$cancel` operation to the `AsyncJob` resource, which allows for cancellation of a job by setting the `status` to `cancelled`, which should trigger the job worker to stop the processing of long-running jobs.

 The current prospective case is for the future data migration jobs, which will often be very long running jobs. In particular, reindex will be a job we want to be able to cancel in an emergency, which is probably the type of job most of our data migrations will be.

### DIFF
--- a/packages/server/src/fhir/operations/asyncjobcancel.ts
+++ b/packages/server/src/fhir/operations/asyncjobcancel.ts
@@ -1,0 +1,41 @@
+import { allOk, assert, badRequest, OperationOutcomeError } from '@medplum/core';
+import { OperationDefinition } from '@medplum/fhirtypes';
+import { Request, Response } from 'express';
+import { asyncWrap } from '../../async';
+import { getAuthenticatedContext } from '../../context';
+import { sendOutcome } from '../outcomes';
+
+export const operation: OperationDefinition = {
+  id: 'AsyncJob-cancel',
+  resourceType: 'OperationDefinition',
+  name: 'asyncjob-cancel',
+  status: 'active',
+  kind: 'operation',
+  code: 'cancel',
+  experimental: true,
+  resource: ['AsyncJob'],
+  system: false,
+  type: false,
+  instance: true,
+  parameter: [{ use: 'out', name: 'return', type: 'OperationOutcome', min: 1, max: '1' }],
+};
+
+/**
+ * Handles HTTP requests for the AsyncJob $cancel operation.
+ * Sets the status of the `AsyncJob` to `cancelled`.
+ */
+export const asyncJobCancelHandler = asyncWrap(async (req: Request, res: Response) => {
+  assert(req.params.id, 'This operation can only be executed on an instance');
+  // Update status of async job
+  try {
+    const { repo } = getAuthenticatedContext();
+    await repo.patchResource('AsyncJob', req.params.id, [{ op: 'add', path: '/status', value: 'cancelled' }]);
+    sendOutcome(res, allOk);
+  } catch (err) {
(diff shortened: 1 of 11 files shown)


mattwiller (approved):

ThatOneBro:
@mattlong The current prospective case is for the future data migration case, which will often be very long running jobs. In particular, reindex will be a job we want to cancel (@mattwiller plans to implement this by checking the `AsyncJob.status` whenever the job in `bullmq` is re-queued and before writing to the `AsyncJob.output`.

Good and valid questions about attempting to cancel jobs, I think it will be up to the implementer of the job to respect the cancel status, but I think it should be done declaratively, having the job itself being responsible for periodically reading the status of the `AsyncJob` itself rather than imperatively done by the operation.

Down the line, we might want to use the `Subscription` mechanic somehow... @mattwiller and I tried to evaluate what going down this path looked like, maybe using channelType = `system`, using `bullmq` queue for `Subscription`s of different types, but no silver bullet yet.

Good question about letting `completed` or `errored` jobs be canceled, probably doesn't make sense. I'll add a check that we are not in one of those "completed" states already 👍

packages/server/src/fhir/operations/asyncjobcancel.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Handles HTTP requests for the AsyncJob $cancel operation.
+ * Sets the status of the `AsyncJob` to `cancelled`.
+ */
+export const asyncJobCancelHandler = asyncWrap(async (req: Request, res: Response) => {
+  assert(req.params.id, 'This operation can only be executed on an instance');
+  // Update status of async job
+  try {
+    const { repo } = getAuthenticatedContext();
+    await repo.patchResource('AsyncJob', req.params.id, [
+      { op: 'test', path: '/status', value: 'accepted' },
```
mattwiller, line 33: I would argue that if the status if already `cancelled`, this operation should be a no-op but return successfully

mattwiller (approved):

packages/server/src/fhir/operations/asyncjobcancel.ts
```diff
@@ -0,0 +1,56 @@
...
+        break;
+      default:
+        sendOutcome(
+          res,
+          badRequest(`AsyncJob cannot be cancelled if status is not 'accepted', job had status '${job.status}'`)
+        );
+    }
+    sendOutcome(res, allOk);
+  } catch (err) {
+    if (err instanceof OperationOutcomeError) {
```
codyebberson, line 50:
```suggestion
    sendOutcome(res, normalizeOperationOutcome(err));
```

packages/server/src/fhir/operations/asyncjobcancel.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+/**
+ * Handles HTTP requests for the AsyncJob $cancel operation.
+ * Sets the status of the `AsyncJob` to `cancelled`.
+ */
+export const asyncJobCancelHandler = asyncWrap(async (req: Request, res: Response) => {
+  assert(req.params.id, 'This operation can only be executed on an instance');
+  // Update status of async job
+  try {
```
codyebberson, line 30: Is this `try`/`catch` necessary?  I believe we're just rethrowing the same errors, correct?

ThatOneBro: Good point

PR #5517, opened 2024-11-09
Revert "feat(useSearch): add debounce to `useSearch`"

codyebberson:
Reverts medplum/medplum#5413

This broke several of the demos.  The issue is this:

1. Something happens on the page, which triggers a search
    * Search happens
    * Search gets cached
    * `lastSearchKey` gets set
2. Something creates a resource, such as a `Slot` in the scheduling demo
    * Resource is created
    * Cached search is invalidated in `MedplumClient`
3. Page rerenders, expecting a new search
    * However, with the new debounc feature, `searchKey` has not changed, so no search is executed

I'm ok with the idea of the debounce, but we need to rethink it, or at least figure out what changes would be necessary for the example apps.


ThatOneBro (approved):

ThatOneBro:
Hmm interesting... so looking at the scenario you described, how is this different with debounce or not?
The search should only be performed if the `key` is different than `searchKey`, even without the debounce, right?

ThatOneBro:
Yeah I'm actually still seeing the same behavior on the scheduling demo, even on this branch. Are we sure reverting this actually fixes the demo?

From what I understand, this search even before debounce will only fire when `fhirSearchUrl(query).toString()` changes, which wouldn't happen unless the search query params themselves changed.

ThatOneBro (changes requested):
Let's make sure that debounce is what caused these problems, it seems like this might be a separate issue

PR #5523, opened 2024-11-12
Store copy of resource from time of call in cache post-commit hook

### LINKED ISSUE #5416: Odd caching bug around `projectId`s
I'm running into a bunch of 404s on GraphQL after PATCHing resources. 

The cause of this "404" is that frequently in the redis cache, we're putting in resources with missing projectIds, so when we query overy GQL, the API 404s the object, since it can't verify which project it came from.
image

I dropped a debugger in the code, and it seems to be stemming from the postCommit callbacks. On my specific transaction, I see that there are 3 stacked postCommit callBacks, and on the final callBack, project becomes undefined and then we store the broken object in Redis, which ends up causing the 404. (On the first two callBacks, there is a `projectId` present)

I'm not exactly sure what's causing this final callback to be missing the project, but I believe it's related to the changes in PR #4358. I reverted to version `3.2.10` and the behavior returns to what we expect.

mattwiller:
Resolves #5416


packages/server/src/fhir/repo.test.ts
```diff
@@ -1225,4 +1225,28 @@ describe('FHIR Repo', () => {
       });
       await expect(repo.createResource(patientJson)).resolves.toBeDefined();
     }));
+
+  test('Patch post-commit does stores full resource in cache', async () =>
```
ThatOneBro, line 1229:
```suggestion
  test('Patch post-commit stores full resource in cache', async () =>
```

packages/server/src/fhir/repo.ts
```diff
@@ -360,6 +362,13 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     if (this.isSuperAdmin()) {
       return true;
     }
+    console.log(
```
ThatOneBro, line 365: Should we use `globalLogger` here? Or was this intended for debugging?

packages/server/src/fhir/repo.ts
```diff
@@ -2234,7 +2243,10 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
   private async setCacheEntry(resource: Resource): Promise<void> {
     // No cache access allowed mid-transaction
     if (this.transactionDepth) {
-      await this.postCommit(() => this.setCacheEntry(resource));
+      const cachedResource = deepClone(resource);
```
ThatOneBro, line 2246: Ouch, I wish we didn't need to clone before caching...

Is this because of the potential for destructive operations during the processing of a `Bundle`? Or something similar?

packages/server/src/fhir/repo.ts
mattwiller: Just for debugging, thanks for flagging!

packages/server/src/fhir/repo.ts
mattwiller: Yeah, the core issue here is that, in context of a PATCH, without the `X-Medplum: extended` header (i.e. `extendedMode` in the `Repository`), several fields are removed from `Resource.meta` at the end of `updateResourceImpl()` — before the post-commit hooks run.  At least the cloning should only ever happen in cases where we're inside of a transaction: for simple writes the existing object should not be cloned, and would be stored in the cache immediately

ThatOneBro (approved):

PR #5526, opened 2024-11-12
Fix link to newline-delimited JSON spec

### LINKED ISSUE #4605: updating the url referencing ndjson in one of the readme's
https://github.com/medplum/medplum/blob/main/packages/docs/docs/api/fhir/operations/bulk-fhir.mdx#bulk-fhir-api

In this Readme, the ndjson url's domain has expired and based on this hackernews post (https://news.ycombinator.com/item?id=39626735), might be an un-safe url 

this seems like the new url that should be linked in that Readme https://jsonlines.org/

mattwiller:
Resolves #4605


ThatOneBro:
This is definitely an improvement over the current broken link, though `NDJSON` and `jsonl` are technically different content types. `jsonl` supports `\r\n` or `\n` line terminators, while `NDJSON` does not. I wonder if FHIR will decide to switch over to `jsonl` since looks like `NDJSON` will die out, though neither actually has an RFC yet...

Just a little context:
https://www.hypertesto.me/en/blog/jsonl-and-ndjson/
https://github.com/wardi/jsonlines/issues/19

ThatOneBro (approved):

PR #5527, opened 2024-11-13
Make QuestionnaireForm change event handling pure

mattlong:
Also made a small change in behavior to `QuestionnaireForm` such that receiving a new `questionnaire` does not erase all existing answers.

It might arguably be better to make `QuestionnaireForm` more of a controlled component that takes `response` as a property; right now it's halfway-controlled ; it fires on changes with responses but has no way to specify the response.

### DIFF
--- a/examples/medplum-health-gorilla-demo/src/components/TestMetadataCardInput.tsx
+++ b/examples/medplum-health-gorilla-demo/src/components/TestMetadataCardInput.tsx
@@ -53,7 +53,7 @@ export function TestMetadataCardInput({ test, metadata, error }: TestMetadataCar
 
             <TextInput
               label="Notes"
-              value={metadata.notes}
+              value={metadata.notes ?? ''}
               onChange={(event) => {
                 updateTestMetadata(test, { notes: event.currentTarget.value });
               }}
--- a/packages/core/src/fhirpath/utils.ts
+++ b/packages/core/src/fhirpath/utils.ts
@@ -223,8 +223,9 @@ export function getTypedPropertyValueWithoutSchema(
     // id + identifier = not ok, because "entifier" is not a valid type
     // resource + resourceType = not ok, because "type" is not a valid type
     //eslint-disable-next-line guard-for-in
-    for (const propertyType in PropertyType) {
-      const propertyName = path + capitalize(propertyType);
+    const trimmedPath = path.endsWith('[x]') ? path.substring(0, path.length - 3) : path;
+    for (const propertyType of Object.values(PropertyType)) {
+      const propertyName = trimmedPath + capitalize(propertyType);
       if (propertyName in input) {
         const propertyValue = (input as { [key: string]: unknown })[propertyName];
         if (Array.isArray(propertyValue)) {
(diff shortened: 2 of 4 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -41,9 +41,24 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
   onSubmitRef.current = props.onSubmit;
 
   useEffect(() => {
-    setResponse(questionnaire ? buildInitialResponse(questionnaire) : undefined);
+    setResponse((prevResponse) => {
```
ThatOneBro, line 44: If the questionnaire changes, shouldn't we start with a fresh questionnaire instead of `prevResponse`?

packages/core/src/fhirpath/utils.ts
```diff
@@ -223,8 +223,9 @@ export function getTypedPropertyValueWithoutSchema(
     // id + identifier = not ok, because "entifier" is not a valid type
     // resource + resourceType = not ok, because "type" is not a valid type
     //eslint-disable-next-line guard-for-in
-    for (const propertyType in PropertyType) {
-      const propertyName = path + capitalize(propertyType);
+    const trimmedPath = path.endsWith('[x]') ? path.substring(0, path.length - 3) : path;
```
ThatOneBro, line 226: Nice

examples/medplum-health-gorilla-demo/src/components/TestMetadataCardInput.tsx
```diff
@@ -53,7 +53,7 @@ export function TestMetadataCardInput({ test, metadata, error }: TestMetadataCar
 
             <TextInput
               label="Notes"
-              value={metadata.notes}
+              value={metadata.notes ?? ''}
```
ThatOneBro, line 56: 👍

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -41,9 +41,24 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+      if (prevResponse) {
+        return prevResponse;
+      }
+      return questionnaire ? buildInitialResponse(questionnaire) : undefined;
+    });
   }, [questionnaire]);
 
+  useEffect(() => {
```
ThatOneBro, line 52: I think this should work for pretty much every case we care about 👍 

If it proves to cause `onChange` to fire excessively, we can investigate using `usePrevious`

ThatOneBro:
Looks good, just one question about the `prevResponse` always being kept if `questionnaire` changes

codyebberson (approved):

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
mattlong: Yep, that is very valid. Putting backwards incompatibility aside for a bit, the answer would be that you should change the `key` property if you want it to totally start over from a clean slate for a new Questionnnaire.

Let me see how it feels to make it (optionally) fully controlled...

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -168,3 +175,7 @@ function mergeItems(
 
   return result;
 }
+
+function getQuestionnaireIdenity(questionnaire: Questionnaire | undefined): Questionnaire | string | undefined {
```
ThatOneBro, line 179:
```suggestion
function getQuestionnaireIdentity(questionnaire: Questionnaire | undefined): Questionnaire | string | undefined {
```

ThatOneBro (approved):
Looks good, just minor spelling nit

PR #5528, opened 2024-11-13
chore(core): `createReference` narrow `reference` to `string`

dillonstreator:
`reference` is guaranteed to be set on the returned `Reference` object from `createReference`. We should narrow it to prevent unnecessary type assertions in calling code.


ThatOneBro (approved):

mattwiller:
@dillonstreator I noticed there was a build failure in an example Bot unrelated to your change; I went ahead and fixed that for you, but now you may need to pull this branch again before adding any more commits

PR #5530, opened 2024-11-13
feat(fhirtypes): adding note to a diagnostic report resource

airinterface:
According to 

https://build.fhir.org/diagnosticreport.html

Diagnostic Report has note section. 

This is is to add note field to the diagnostic report.

### DIFF
--- a/packages/fhirtypes/dist/DiagnosticReport.d.ts
+++ b/packages/fhirtypes/dist/DiagnosticReport.d.ts
@@ -3,6 +3,7 @@
  * Do not edit manually.
  */
 
+import { Annotation } from './Annotation';
 import { Attachment } from './Attachment';
 import { CarePlan } from './CarePlan';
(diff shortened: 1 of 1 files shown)


mattwiller:
Hi @airinterface — thanks for the PR!  This is getting a bit into the weeds of FHIR, but the URL you linked points to `build.fhir.org`, which is the _not-yet released, in-progress version_ of the FHIR specification.  The released versions of the specification have URLs starting with `hl7.org/fhir/`, e.g. http://hl7.org/fhir/R4/diagnosticreport.html for the R4 version of the specification that Medplum uses.  This version does not have the `DiagnosticReport.note` field, which was [added later in the R5 version](https://hl7.org/fhir/R5/diagnosticreport.html).  Unfortunately, this change will likely need to wait until Medplum (and the healthcare industry more broadly) migrates to a newer version of the specification sometime in the future

PR #5537, opened 2024-11-14
Fix ResearchStudy type by removing Citation reference

codyebberson:
Context:
* We backported some R5 features of the `ResearchStudy` type to support https://clinicaltrials.gov/
   * See https://github.com/medplum/medplum/pull/5394 and https://github.com/medplum/medplum/pull/5428
* Backporting R5 is always risky, due to the sprawling nature of new data types
* Despite knowing that, we missed this one, where `ResearchStudy` can refer to `Citation`, which does not exist in R4 at all
* Our automation did not catch this
   * The `@medplum/fhirtypes` package is autogenerated by some custom code, not `api-generator`
   * All of our projects use `"skipLibCheck": true` in `tsconfig.json`, which silently ignores this type of error

In the future, we will explore using `api-generator`, or having at least one test project which uses `"skipLibCheck": false`.

### DIFF
--- a/packages/docs/static/data/resourceDefinitions/researchstudy.json
+++ b/packages/docs/static/data/resourceDefinitions/researchstudy.json
@@ -1788,9 +1788,6 @@
           "datatype": "EvidenceReport"
         },
         {
-          "datatype": "Citation"
-        },
-        {
           "datatype": "DiagnosticReport",
           "documentLocation": "resource"
         }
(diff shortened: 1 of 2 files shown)


ThatOneBro (changes requested):
I think you need to remove `EvidenceReport` too

ThatOneBro (approved):

ThatOneBro (approved):

PR #5538, opened 2024-11-14
Use distroless base Docker image

mattwiller:
Two changes to the Docker build process to improve build times and harden/optimize the resulting image:
- Instead of running `npm ci` a second time inside the Docker container, include the `node_modules` from the original build step in the `medplum-server.tar.gz` artifact
- Use Google's [distroless](https://github.com/GoogleContainerTools/distroless) base image to shrink the final image, reduce security surface area, and automatically use a nonroot user to run the service

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,19 @@
 # Supported architectures:
 # linux/amd64, linux/arm64, linux/arm/v7
 # https://github.com/docker-library/official-images#architectures-other-than-amd64
-FROM gcr.io/distroless/nodejs20-debian12:nonroot
-WORKDIR /app
+FROM node:20-slim AS build
+WORKDIR /build
 
 # Add the application files: Docker handles extracting the tarball automatically
 ADD ./medplum-server.tar.gz ./
+# Install dependencies inside Docker container to handle multi-arch builds
+RUN npm ci --maxsockets 1
+
+
+FROM gcr.io/distroless/nodejs20-debian12:nonroot
+WORKDIR /app
+
+COPY --from=build /build /app
 
 ENV PATH=/nodejs/bin:$PATH
 ENV NODE_ENV=production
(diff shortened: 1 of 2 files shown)


scripts/build-docker.sh
```diff
@@ -24,14 +24,17 @@ tar \
   -czf medplum-server.tar.gz \
   package.json \
   package-lock.json \
+  node_modules \
```
codyebberson, line 27: I'm not sure this works with multi arch builds?

codyebberson (commented):
Neat.  TIL about Google's distroless

scripts/build-docker.sh
ThatOneBro: Most of the packages include the native bindings for all architectures. Not a guarantee but there's a chance it works

ThatOneBro (approved):

Dockerfile
```diff
@@ -10,25 +10,21 @@
...
+RUN npm ci --maxsockets 1
 
-# Install dependencies, create non-root user, and set permissions in one layer
-RUN npm ci --maxsockets 1 && \
-  groupadd -r medplum && \
-  useradd -r -g medplum medplum && \
-  chown -R medplum:medplum /usr/src/medplum
 
-EXPOSE 5000 8103
+FROM gcr.io/distroless/nodejs20-debian12:nonroot
```
codyebberson, line 22: Hm.  Why did we need 2 phase build?  Does the distroless image not include `npm`?

mattwiller: Correct: the distroless image is stripped down to the bare essentials to run the application, and specifically does not include the package manager

codyebberson:
Trying this on staging: https://github.com/medplum/medplum/actions/runs/11904586334

codyebberson:
Staging deploy succeeded 🎉 

image

Comparing staging action for [`distroless-image`](https://github.com/medplum/medplum/actions/runs/11904586334) vs [the previous run](https://github.com/medplum/medplum/actions/runs/11827405516)...

"Build Docker" step went from 2 min 15 sec to 3 min 44 sec, which makes sense, because of the 2 phase build, and downloading more base images.

Docker image size actually went up 136.5 MB to 142.9 MB.

https://hub.docker.com/repository/docker/medplum/medplum-server-staging/tags

image

PR #5540, opened 2024-11-14
Add build script to @medplum/fhirtypes

mattlong:
As a follow up to #5537, adding `tsc` as the build step to `@medplum/fhirtypes`.

Big shoutout to @ThatOneBro for identifying `typeRoots` to avoid checking everything in various `node_modules/@types`!

### DIFF
--- a/packages/fhirtypes/tsconfig.json
+++ b/packages/fhirtypes/tsconfig.json
@@ -3,5 +3,6 @@
     "noEmit": true,
     "skipLibCheck": false,
     "typeRoots": ["dist"]
-  }
+  },
+  "include": ["dist"]
 }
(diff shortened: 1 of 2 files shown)


packages/fhirtypes/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "skipLibCheck": false,
+    "typeRoots": ["dist"]
+  }
```
ThatOneBro, line 6: We might wanna be explicit about include, otherwise it defaults to `**/*`

ThatOneBro (approved):

PR #5545, opened 2024-11-15
First draft of DoseSpot react components

codyebberson:


examples/medplum-provider/src/components/DoseSpotIcon.tsx
```diff
@@ -91,7 +20,7 @@ export function DoseSpotIcon(): JSX.Element {
     </Tooltip>
   );
 
-  return unreadCount > 0 ? (
+  return unreadCount !== undefined && unreadCount > 0 ? (
```
ThatOneBro, line 23:
```suggestion
  return unreadCount ? (
```
Think we can simplify this safely, right?

packages/dosespot-react/esbuild.mjs
```diff
@@ -0,0 +1,46 @@
...
+  loader: { '.ts': 'ts', '.tsx': 'tsx' },
+  logLevel: 'info',
+  resolveExtensions: ['.ts', '.tsx'],
+  target: 'es2021',
+  tsconfig: 'tsconfig.build.json',
+  minifyWhitespace: true,
+  minifyIdentifiers: false,
+  minifySyntax: true,
+  sourcemap: true,
+  external: ['@medplum/core', '@medplum/mock', '@medplum/react', 'react'],
```
ThatOneBro, line 21: May need to add `@medplum/react-hooks` here?

This may be something we need to check in other things importing `@medplum/react` too

packages/dosespot-react/src/index.test.ts
```diff
@@ -0,0 +1,7 @@
+import { useDoseSpotNotifications } from '.';
+
+describe('Index', () => {
+  test('Expected exports', () => {
+    expect(useDoseSpotNotifications).toBeDefined();
```
ThatOneBro, line 5: Should we test `useDoseSpotIframe` as well?

packages/dosespot-react/src/useDoseSpotNotifications.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+      if (result.RefillRequestsCount !== undefined) {
+        newCount += result.RefillRequestsCount;
+      }
+      if (result.TransactionErrorsCount !== undefined) {
+        newCount += result.TransactionErrorsCount;
+      }
+      if (newCount !== unreadCount) {
+        setUnreadCount(newCount);
+        onChange?.(newCount);
```
ThatOneBro, line 50: Interesting, why have both `unreadCount` and also accept a function for `onChange`?

Not super icked by it, just seems like these both do the same thing, ie.:

```tsx
  // This
  const unreadCount = useDoseSpotNotifications();
  useEffect(() => {
     doSomething();
  }, [unreadCount]);
  
  // is the same thing as this
 useDoseSpotNotifications({ onChange: () => doSomething() });
```

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -0,0 +1,70 @@
...
+      initializingRef.current = true;
+      if (patient) {
+        await syncPatient();
+      }
+      await openIframe();
+    }
+  }, [patient, syncPatient, openIframe]);
+
+  useEffect(() => {
+    initPage().catch(console.error);
```
ThatOneBro, line 66: This will be triggered any time the patient resource is updated, not sure if that is intentional

ThatOneBro (commented):
Looks good for the most part! One major concern and a few minor questions

packages/dosespot-react/src/useDoseSpotIFrame.ts
codyebberson: Good question... It might be better to use "patient ID" for everything, and let the Bot fetch the Patient resource at execute time, rather than trying to juggle that state here.

packages/dosespot-react/src/useDoseSpotNotifications.ts
codyebberson: Yeah.... I debated that.   First one probably better?

packages/dosespot-react/src/index.test.ts
codyebberson: Yeah, I'm writing tests now.

I was copying ML's approach with `@medplum/health-gorilla-react`, where he just verifies that at least one export is available.  My assumption was that the intent was just to verify the barrel is working as expected.  In this case, where we only have 2 exports, maybe it's worth it to be exhaustive.

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function useDoseSpotIFrame(options: DoseSpotIFrameOptions): string | undefined {
+  const medplum = useMedplum();
+  const { patientId, onPatientSyncSuccess, onPatientSyncError, onIframeSuccess, onIframeError } = options;
+  const patient = useResource<Patient>(patientId ? { reference: `Patient/${patientId}` } : undefined);
+  const initializingRef = useRef<boolean>(false);
+  const [iframeUrl, setIframeUrl] = useState<string | undefined>(undefined);
+
+  const onPatientSyncSuccessRef = useRef(onPatientSyncSuccess);
```
ThatOneBro, line 21: I like this pattern a lot 👍

packages/dosespot-react/esbuild.mjs
codyebberson: Good call

packages/dosespot-react/src/useDoseSpotNotifications.ts
ThatOneBro: Yeah I could probably go either way... maybe it's actually fine to have both?

I do prefer the first one. But we already have `onError` either way, so for symmetry I could see `onChange` also existing...

But from the perspective of DX / simplicity I do like the idea of having one way to do everything...

I vote for 1 but if we keep both I wouldn't revolt

packages/dosespot-react/src/useDoseSpotIFrame.ts
ThatOneBro: Yeah I think that's probably the right way to do it, if we can get away with only using ID

codyebberson: I remember why we did it this way... One of the bots needs the Medplum patient ID, and one of the bots needs the DoseSpot patient ID.  I'll need to do a little more refactoring so that both bots operate on Medplum patient ID.

examples/medplum-provider/src/components/DoseSpotIcon.tsx
codyebberson: ✅

packages/dosespot-react/src/index.test.ts
codyebberson: ✅

packages/dosespot-react/src/useDoseSpotNotifications.ts
codyebberson: Leaving both for now, happy to revisit if we don't like it.

packages/dosespot-react/src/useDoseSpotIFrame.ts
codyebberson: 🎉  I definitely think of this as a Derrick-ism 😎

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -0,0 +1,53 @@
...
+  const initPage = useCallback(async () => {
+    if (initializingRef.current) {
+      return;
+    }
+
+    initializingRef.current = true;
+    try {
+      if (patientId) {
+        await medplum.executeBot(DOSESPOT_PATIENT_SYNC_BOT, { patientId });
+        onPatientSyncSuccessRef.current?.();
```
mattlong, line 36: It could be nice to add `patientId` as a parameter of `onPatientSyncSuccess` to echo back to users which patient was synced but maybe that premature.

mattlong (approved):

ThatOneBro (approved):

packages/dosespot-react/src/common.ts
```diff
@@ -0,0 +1,21 @@
+import { Identifier } from '@medplum/fhirtypes';
+
+export const MEDPLUM_BOT_SYSTEM = 'https://www.medplum.com/bots';
```
rahul1, line 3: nit: in the health gorilla integration, I tried to use a system string that was more specificy that this is a medplum integration: `https://www.medplum.com/integrations/bot-identifier`

Might be overkill, but might be nice to be consistent across our integration

https://github.com/medplum/medplum-ee/blob/19c61e15a21f646d957cf2c2c3a1a404212bc64c/packages/health-gorilla/src/common/constants.ts#L5

PR #5548, opened 2024-11-15
Defensive check in async resource crawler

mattwiller:
Noticed a `TypeError: Cannot convert undefined or null to object` coming from this part of the code, adding a defensive check here to prevent trying to recurse on an empty value

### DIFF
--- a/packages/core/src/typeschema/crawler.ts
+++ b/packages/core/src/typeschema/crawler.ts
@@ -263,7 +263,7 @@ class AsyncCrawler {
       await this.visitor.onEnterObject(path, obj, schema);
     }
 
-    if (this.excludeMissingProperties) {
+    if (this.excludeMissingProperties && obj.value) {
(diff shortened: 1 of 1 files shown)


mattlong (approved):

packages/core/src/typeschema/crawler.ts
```diff
@@ -263,7 +263,7 @@ class AsyncCrawler {
       await this.visitor.onEnterObject(path, obj, schema);
     }
 
-    if (this.excludeMissingProperties) {
+    if (this.excludeMissingProperties && obj.value) {
```
ThatOneBro, line 266: Does this account for falsy values (0, false)?

ThatOneBro: Ah I see, it can only be an object. `TypedValue.value` typing being `any` was throwing me off

ThatOneBro (approved):

PR #5553, opened 2024-11-18
Add automatic retry of transaction serialization errors after rollback

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -2168,18 +2171,29 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    for (let i = 0; i < transactionAttempts; i++) {
+      try {
+        const client = await this.beginTransaction(options?.serializable ? 'SERIALIZABLE' : undefined);
+        const result = await callback(client);
+        await this.commitTransaction();
+        return result;
+      } catch (err) {
+        const operationOutcomeError = normalizeDatabaseError(err);
+        // Assigning here and throwing below is necessary to satisfy TypeScript
+        error = err;
```
ThatOneBro, line 2184: Small nit, could cast to `OperationOutcomeError | Error`:
```ts
let error:  OperationOutcomeError | Error;
```

...

```ts
error = err as OperationOutcomeError | Error;
```

Then pass it into `this.isRetryableTransactionError` and make signature:

```ts
  private isRetryableTransactionError(err: OperationOutcomeError | Error): boolean {
```

ThatOneBro (approved):
LGTM, small nit

packages/server/src/fhir/repo.ts
mattwiller: Since `isRetryableTransactionError()` tries accessing `err.code`, it requires `any` or perhaps an interface type: that property doesn't exist on either `OperationOutcomeError` or generic `Error`

ThatOneBro: Oh good point

packages/server/src/fhir/repo.ts
```diff
@@ -2168,18 +2171,31 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
-      const result = await callback(client);
-      await this.commitTransaction();
-      return result;
-    } catch (err) {
-      const operationOutcomeError = normalizeDatabaseError(err);
-      await this.rollbackTransaction(operationOutcomeError);
-      throw operationOutcomeError;
-    } finally {
-      this.endTransaction();
+    let error: OperationOutcomeError | undefined;
```
codyebberson, line 2174: ooo.  Clever.  This is a deeper insertion point than I was expecting...  But I like it, I think it makes a lot of sense.

I took a quick look over all of our uses of `withTransaction()` and asked the question "would we be ok if this retried 3 times?"

In all cases, I think it's good, and it will absolutely work.

The one case that I was somewhat uncertain about was the reindex worker, when processing a full page of 500 resources.  Even in that case, I think it's fine?  Might actually make the reindex job more robust?

codyebberson (approved):
Very clever

packages/server/src/fhir/repo.ts
mattwiller: I struggled a bit going back and forth with handling this at the FHIR interaction level vs. here, but ultimately I think that having this handled consistently in one spot is the best outcome in terms of ensuring that the retry is performed correctly

PR #5559, opened 2024-11-20
feat(server): add `Bot.timeout`, sync timeout on `Bot/$deploy`

### LINKED ISSUE #5558: Be able to set lambda timeout via Bot resource
The default timeout for AWS lambda is 15s, which can be too short for some applications of Bots. 
Right now the only way to set a Bot timeout is to log into the AWS console and manually set it on the Lambda resource

## Proposal

- [ ] Add a new element on the `Bot` resource called `timeout`, that specifies the timeout (in seconds)
- [ ] When [Bot deploy](https://www.medplum.com/docs/api/fhir/operations/bot-deploy#invoke-deploy-operation) is called, update the underlying runner (e.g. AWS Lambda) with the appropriate timeout configuration

ThatOneBro:
Fixes #5558

### DIFF
--- a/packages/server/src/cloud/aws/deploy.ts
+++ b/packages/server/src/cloud/aws/deploy.ts
@@ -17,7 +17,6 @@ import { ConfiguredRetryStrategy } from '@smithy/util-retry';
 import JSZip from 'jszip';
 import { getConfig } from '../../config';
 import { getLogger } from '../../context';
-import { globalLogger } from '../../logger';
 
 export const LAMBDA_RUNTIME = 'nodejs18.x';
 export const LAMBDA_HANDLER = 'index.handler';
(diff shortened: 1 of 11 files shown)


packages/server/src/cloud/aws/deploy.ts
```diff
@@ -141,18 +180,22 @@ async function lambdaExists(client: LambdaClient, name: string): Promise<boolean
     const command = new GetFunctionCommand({ FunctionName: name });
     const response = await client.send(command);
     return response.Configuration?.FunctionName === name;
-  } catch (_err) {
-    return false;
+  } catch (err) {
+    if (err instanceof ResourceNotFoundException) {
```
ThatOneBro, line 184: Before we swallowed all errors and returned false, we now catch `ResourceNotFoundException` which is returned when the Lambda doesn't already exist, and rethrow all other errors

codyebberson (approved):
nice 👍

PR #5565, opened 2024-11-21
Allow cancellation of reindex job

mattwiller:
Refactors the Reindex job into an instance of an abstract class encapsulating the re-entrant job pattern for long-running background processes, and adds checks to abort the reindex if the associated `AsyncJob` resource has been updated to `cancelled` status:
- Once at the beginning of every iteration of the job
- Updates to the `AsyncJob` resource are now done conditionally to ensure the update is applied to the latest version; if the status was updated to `cancelled` mid-execution it will be caught at this point before an update would clobber the `cancelled` status

### DIFF
--- a/packages/server/src/workers/longterm-job.ts
+++ b/packages/server/src/workers/longterm-job.ts
@@ -2,6 +2,7 @@ import { AsyncJob, Parameters } from '@medplum/fhirtypes';
 import { Job } from 'bullmq';
 import { AsyncJobExecutor } from '../fhir/operations/utils/asyncjobexecutor';
 import { getSystemRepo, Repository } from '../fhir/repo';
+import { getStatus, OperationOutcomeError } from '@medplum/core';
 
 export interface LongTermJobData {
   asyncJob: AsyncJob;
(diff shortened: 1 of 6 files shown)


packages/server/src/workers/longterm-job.ts
```diff
@@ -0,0 +1,136 @@
...
+      // happens less frequently than every iteration
+      const output = this.formatResults(result, job);
+      if (output) {
+        try {
+          await this.updateStatus(job, output);
+        } catch (err) {
+          if (err instanceof OperationOutcomeError && getStatus(err.outcome) === 412) {
+            // Conflict: AsyncJob was updated by another party between when the job started and now!
+            // Check status to see if job was cancelled
+            const canContinue = await this.checkJobStatus(job);
```
ThatOneBro, line 74: Clever, pretty clean

packages/server/src/workers/longterm-job.ts
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+          await this.failJob(job);
+        }
+      } else {
+        // Enqueue job for the specified next iteration
+        // NOTE: We do not check the AsyncJob status before enqueuing: it will be checked
+        // at the beginning of the next iteration
+        await this.enqueueJob(nextIteration);
+      }
+    } catch (err: any) {
```
ThatOneBro, line 100: Small nit:

Prefer `err: unknown`

With case to a specific error type in the function arguments

packages/server/src/workers/reindex.ts
```diff
@@ -45,7 +45,7 @@ export function initReindexWorker(config: MedplumServerConfig): void {
   queue = new Queue<ReindexJobData>(queueName, {
     ...defaultOptions,
     defaultJobOptions: {
-      attempts: 3,
+      attempts: 1,
```
ThatOneBro, line 48: In the case that a job fails due to some connection error with Redis or similar, don't we maybe want to still retry?

A potential way to think about it:
From the perspective of `bullmq` job only truly fails if errors other than a cancellation occur...

Not sure if that makes sense but something to consider

ThatOneBro (commented):
Nice, think this came out really clean.

A few things I wanted to flag before approving

packages/server/src/workers/reindex.ts
mattwiller: The way the job is written, all errors that would be thrown are caught by the worker and result in the `AsyncJob` being updated to `error` status.  There's no possibility for the job to be retried anyway, so I was just aligning this to that reality

ThatOneBro:
Nice, do we want to deploy to staging for a test run?

ThatOneBro:
One thing I noticed while running the reindex from UI on staging is that after the job is cancelled, we continue to poll the job and never indicate that we are no longer indexing (spinner)

We can address that in a separate PR but looks like the cancel worked as expected for the most part. The only weird thing I noticed is the `AsyncJob` did not end up with a `transactionTime` set, which might be useful just to know how long the whole job ran before being cancelled

ThatOneBro (approved):
LGTM,

I'll create an issue for the UI change

PR #5574, opened 2024-11-21
Text overflows in ResourceDiffTable

techdavidy:
Example A:

Example B:

### DIFF
--- a/packages/react/src/ResourceDiffTable/ResourceDiffTable.module.css
+++ b/packages/react/src/ResourceDiffTable/ResourceDiffTable.module.css
@@ -10,6 +10,10 @@
   & td {
     padding: var(--mantine-spacing-sm) var(--mantine-spacing-sm);
     vertical-align: top;
+    word-wrap: break-word;
+    overflow-wrap: anywhere;
+    white-space: normal; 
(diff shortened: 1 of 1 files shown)


packages/react/src/ResourceDiffTable/ResourceDiffTable.module.css
```diff
@@ -10,6 +10,10 @@
   & td {
     padding: var(--mantine-spacing-sm) var(--mantine-spacing-sm);
     vertical-align: top;
+    word-wrap: break-word;
+    overflow-wrap: anywhere;
```
techdavidy, line 14: Makes sure it overlaps with strings like urls that have no spaces.

packages/react/src/ResourceDiffTable/ResourceDiffTable.module.css
```diff
@@ -10,6 +10,10 @@
...
+    white-space: normal; 
+    min-width: 100px;
```
techdavidy, line 16: if a column is too wide. it makes sure the other do not wraps its content. Like in example B.

codyebberson (approved):
Nice, ty! 🙏

PR #5580, opened 2024-11-22
[Photon Health] Sync patients from Photon

ksmith94:
This PR creates a bot that syncs patients from Photon into Medplum. 
This is still a work in progress as the UI needs to be updated to include a way to execute the bot. Scripts also need to be updated to include this bot.

### DIFF
--- a/examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
+++ b/examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
@@ -155,7 +155,7 @@ export async function handler(medplum: MedplumClient, event: BotEvent): Promise<
       // Add the patient resource to a bundle
       const patientEntry: BundleEntry = {
         fullUrl: patientUrl,
-        request: { method: 'PUT', url: `Patient?identifier=https://neutron.health|${photonPatient.id}` },
+        request: { method: 'PUT', url: `Patient?identifier=${NEUTRON_HEALTH_PATIENTS}|${photonPatient.id}` },
         resource: medplumPatient,
       };
       batch.entry?.push(patientEntry);
--- a/examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
+++ b/examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
@@ -1,8 +1,35 @@
-import { BotEvent, MedplumClient } from '@medplum/core';
-import { PhotonPatient } from '../photon-types';
-import { handlePhotonAuth, photonGraphqlFetch } from './utils';
+import {
+  BotEvent,
+  createReference,
+  getDisplayString,
+  MedplumClient,
+  normalizeErrorString,
+  RXNORM,
+} from '@medplum/core';
+import {
+  Address,
+  AllergyIntolerance,
+  Bundle,
+  BundleEntry,
+  ContactPoint,
+  HumanName,
+  MedicationRequest,
+  Patient,
+  Practitioner,
+  Reference,
+} from '@medplum/fhirtypes';
+import { randomUUID } from 'crypto';
+import {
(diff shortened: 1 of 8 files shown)


examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -0,0 +1,389 @@
...
+  const photonPatients = result.data.patients as PhotonPatient[];
+  const batch: Bundle = {
+    resourceType: 'Bundle',
+    type: 'batch',
+    entry: [],
+  };
+
+  for (const photonPatient of photonPatients) {
+    // Make sure the patient does not already exist
+    if (!(await checkForExistingPatient(photonPatient, medplum))) {
```
rahul1, line 150: I think we should still sync the patient's allergies and medications even if the patient already exists. We just don't want to update the `Patient` resource

examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -0,0 +1,389 @@
...
+  }
+  const allergies: AllergyIntolerance[] = [];
+  for (const photonAllergy of photonAllergies) {
+    const { allergen, comment, onset } = photonAllergy;
+    if (!allergen.rxcui) {
+      continue;
+    }
+
+    const allergy: AllergyIntolerance = {
+      resourceType: 'AllergyIntolerance',
```
rahul1, line 266: We should also add `meta.source=<PHOTON_URL>` to the `AllergyIntolerance` and `MedicationRequest` . this is a searchable parameter that can be used for upsert

rahul1 (changes requested):

examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -0,0 +1,389 @@
...
+  const address = photonPatient.address && formatAddress(photonPatient.address);
+
+  if (address) {
+    patient.address = address;
+  }
+
+  return patient;
+}
+
+export function createAllergies(
```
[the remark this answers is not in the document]
ksmith94: Photon tracks a patient's allergies using a [PatientAllergy](https://reference.photon.health/#definition-PatientAllergy), which doesn't have an id, so there's unfortunately no way to add an identifier for allergies.

examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -0,0 +1,446 @@
...
+    batch.entry?.push(patientEntry);
+
+    // If there are allergies, create entries and add them to the bundle
+    if (allergies) {
+      const allergyEntries: BundleEntry[] = allergies.map((allergy) => {
+        return {
+          fullUrl: 'urn:uuid:' + randomUUID(),
+          request: {
+            method: 'PUT',
+            url: `AllergyIntolerance?_source=${NEUTRON_HEALTH}&code=${allergy.code}&patient=${allergy.patient.reference}`,
```
rahul1, line 179: I got some guidance that we shouldn't be using `urn:uuid` references in the `url` during an upsert. 

Instead, we should be using chained search like this: 
`patient.identifier=${NEUTRON_HEALTH_PATIENTS}|${photonPatient.id}`

rahul1 (approved):
requested one change re: upserting Allergies

ThatOneBro:
Hi @ksmith94 , do you mind rebasing this branch before attempting to merge again. 
Looks like some recent change in `main` is causing your branch to fail merge checks for `eslint` even though CI passed on this branch before merge. 

Screenshot 2024-12-19 at 12 56 54 PM

PR #5581, opened 2024-11-22
fix(server,client): stop polling async request when cancelled

### LINKED ISSUE #5575: Stop polling `AsyncJob`s after they are cancelled
Currently whenever an `AsyncJob` is cancelled, we continue polling indefinitely because the client is not aware of the `cancelled` status. We should stop polling and handle the cancelled status specifically in the cancelled reindex job case.

Screenshot 2024-11-21 at 3 35 37 PM

ThatOneBro:
Fixes #5575

### DIFF
--- a/packages/app/src/admin/SuperAdminPage.tsx
+++ b/packages/app/src/admin/SuperAdminPage.tsx
@@ -38,7 +38,7 @@ export function SuperAdminPage(): JSX.Element {
   }
 
   function reindexResourceType(formData: Record<string, string>): void {
-    startAsyncJob(medplum, 'Reindexing Resources', 'admin/super/reindex', formData);
+    startAsyncJob(medplum, 'Reindexing Resources', 'admin/super/reindex', { body: formData, cancellable: true });
   }
 
   function removeBotIdJobsFromQueue(formData: Record<string, string>): void {
@@ -207,7 +207,12 @@ export function SuperAdminPage(): JSX.Element {
   );
 }
 
-function startAsyncJob(medplum: MedplumClient, title: string, url: string, body?: Record<string, string>): void {
+type AsyncJobOptions = {
+  body?: Record<string, string>;
+  cancellable?: boolean;
+};
+
+function startAsyncJob(medplum: MedplumClient, title: string, url: string, asyncJobOptions?: AsyncJobOptions): void {
   const controller = new AbortController();
   notifications.show({
     id: url,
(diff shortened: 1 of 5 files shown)


packages/core/src/client.ts
```diff
@@ -360,6 +360,10 @@ export interface MedplumRequestOptions extends RequestInit {
    */
   pollStatusPeriod?: number;
 
+  /**
+   * Optional AbortSignal for cancelling an async request.
+   */
+  asyncReqCancelSignal?: AbortSignal;
```
codyebberson, line 366: Can we just use the normal `options.signal`?

packages/core/src/client.ts
```diff
@@ -3344,6 +3348,10 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
 
   private async pollStatus<T>(statusUrl: string, options: MedplumRequestOptions, state: RequestState): Promise<T> {
     const statusOptions: MedplumRequestOptions = { ...options, method: 'GET', body: undefined, redirect: 'follow' };
+    if (options?.asyncReqCancelSignal?.aborted) {
+      await this.request('DELETE', statusUrl, { ...options, method: 'DELETE', body: undefined });
```
codyebberson, line 3352: The `delete()` and `request()` methods will correctly set the `method`

```suggestion
      await this.delete(statusUrl, options);
```

packages/core/src/client.ts
```diff
@@ -3344,6 +3348,10 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
+      return this.request('GET', statusUrl, statusOptions);
```
codyebberson, line 3353:
```suggestion
      return this.get(statusUrl, statusOptions);
```

packages/app/src/admin/SuperAdminPage.tsx
```diff
@@ -207,34 +207,86 @@ export function SuperAdminPage(): JSX.Element {
...
+            title,
+            message: 'Job cancelled',
+            icon: <IconX size="1rem" />,
+            loading: false,
+            autoClose: false,
+            withCloseButton: true,
+          };
+          // If the signal has been aborted, it means we made the change via our X button
+          // That removes the notification so we have to show it again
+          if (controller.signal.aborted) {
```
codyebberson, line 267: Hm.. I thought `update` would bring back the notification if closed, no?

ThatOneBro: Nope, it does not. It stays closed sadly

packages/core/src/client.ts
ThatOneBro: Nope because that will abort the fetch request itself as well. I originally tried that but it was hard to untangle the options since you have to, in the middle of a bunch of tied together requests, remove the signal from the options or else risk other fetch requests made in the lifecycle of the parent request being pre-aborted. 

Logically these are two separate signals as well, so it kind of makes sense not to overload them. Maybe there's a complicated way but it seemed to preserve the intent of the signal better to separate it out.

packages/core/src/client.ts
ThatOneBro: Still need to explicitly remove the `body`, but otherwise can remove the explicit method

ThatOneBro:
Deploying to staging

packages/core/src/client.ts
ThatOneBro: Ok I tried to make it work again and I remember why it didn't work last time. 

We have to pass through the signal to each subsequent call to `pollStatus`, which means if you pass the signal through to `request`, it will be passed to `fetch` and if abort the signal while that request is still in flight.

You don't get the next call to `pollStatus` where it would have picked up that it needs to cancel and so it gets stuck and never cancels, just stops polling.

ThatOneBro: Ok, as discussed offline did a big refactor that now only touches `packages/app` for this part. Cancelling done via `DELETE` on status URL and then just going down the same server side path letting the cancel pop up naturally while polling

codyebberson (approved):

Issue #5568, opened 2024-11-21
Duplicate offline notifications


codyebberson:
Medplum app detects if the user is offline, and shows an offline error message.  The bug is that the error notification can show up more than one time (see screenshot). 

These notifications use the Mantine notification system: https://mantine.dev/x/notifications/

Mantine notifications supports re-using and updating an existing notification.  See this particular example: https://mantine.dev/x/notifications/#update-notification

The code that shows the offline error message is here: `packages/react/src/AppShell/AppShell.tsx`

As a stretch goal, we could try to hide that notification if the user goes back online.

<img width="927" alt="image" src="https://github.com/user-attachments/assets/bb45f6df-648d-4baf-b131-cf26cd580c43">

PR #5583, opened 2024-11-22
Show only one offline notification at a time

techdavidy:
Fixes https://github.com/medplum/medplum/issues/5568
Display one notification at a time.

https://github.com/user-attachments/assets/ed4db365-7ab4-47a9-94fe-55fd466b9cc8

### DIFF
--- a/packages/react/src/AppShell/AppShell.tsx
+++ b/packages/react/src/AppShell/AppShell.tsx
@@ -23,26 +23,22 @@ export interface AppShellProps {
 
 export function AppShell(props: AppShellProps): JSX.Element {
   const [navbarOpen, setNavbarOpen] = useState(localStorage['navbarOpen'] === 'true');
-  const [isNotificationShown, setIsNotificationShown] = useState(false);
   const medplum = useMedplum();
   const profile = useMedplumProfile();
 
   useEffect(() => {
     function eventListener(): void {
-      if (!isNotificationShown) {
         showNotification({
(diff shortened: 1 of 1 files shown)


packages/react/src/AppShell/AppShell.tsx
```diff
@@ -23,16 +23,26 @@ export interface AppShellProps {
...
   const [navbarOpen, setNavbarOpen] = useState(localStorage['navbarOpen'] === 'true');
+  const [isNotificationShown, setIsNotificationShown] = useState(false);
   const medplum = useMedplum();
   const profile = useMedplumProfile();
 
   useEffect(() => {
     function eventListener(): void {
-      showNotification({ color: 'red', message: 'No connection to server', autoClose: false });
+      if (!isNotificationShown) {
+        showNotification({
```
codyebberson, line 33: I think in this case, we can simply add an `id` property to the `showNotification` argument, and then the Mantine notification system will automatically handle the de-duplication.

That has the added benefit of (in the future) giving the notification a handle that we can use to update status to "back online".  We don't have an event for "online" right now, so that is something in the future.

```suggestion
        showNotification({ id: 'offline', color: 'red', message: 'No connection to server', autoClose: false });
```

codyebberson (commented):
Thanks @techdavidy  - I think there might be a simpler way to do this using the Mantine notifications system.  See the inline suggestion.

PR #5587, opened 2024-11-22
feat(AsyncJob): support `AsyncJob` cancel via DELETE on `Content-Location` 

### LINKED ISSUE #5584: Support AsyncJob cancel via DELETE request on `Content-Location`
As part of the Asynchronous Request Pattern spec:
https://www.hl7.org/fhir/R4/async.html#3.1.6.3.0.2

We missed this on our initial pass working towards data migrations but I found we had already stubbed this route out here:
https://github.com/medplum/medplum/blob/585e3cb3b1ac18acb2bdf0694d790c8899830278/packages/server/src/fhir/job.ts#L36

ThatOneBro:
Closes #5584

### DIFF
--- a/packages/server/src/fhir/job.ts
+++ b/packages/server/src/fhir/job.ts
@@ -1,4 +1,4 @@
-import { accepted, allOk, isOk, normalizeOperationOutcome } from '@medplum/core';
+import { accepted, allOk, isOk } from '@medplum/core';
 import { FhirRequest, HttpMethod } from '@medplum/fhir-router';
 import { AsyncJob, OperationOutcome } from '@medplum/fhirtypes';
 import { Request, Response, Router } from 'express';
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/operations/asyncjobcancel.ts
```diff
@@ -24,8 +22,11 @@ export const operation: OperationDefinition = {
 /**
  * Handles HTTP requests for the AsyncJob $cancel operation.
  * Sets the status of the `AsyncJob` to `cancelled`.
+ *
+ * @param req - A request to cancel a particular async job.
+ * @returns A FhirResponse.
  */
-export const asyncJobCancelHandler = asyncWrap(async (req: Request, res: Response) => {
+export async function asyncJobCancelHandler(req: FhirRequest): Promise<FhirResponse> {
```
ThatOneBro, line 29: Converted previous handler to proper FHIRRouter handler rather than bare express handler

packages/server/src/fhir/job.ts
```diff
@@ -33,6 +36,48 @@ jobRouter.get(
...
+      const request: FhirRequest = {
+        method: req.method as HttpMethod,
+        url: req.originalUrl.replace('/fhir/R4', ''),
+        pathname: '',
+        params: req.params,
+        query: {},
+        body: req.body,
+        headers: req.headers,
+      };
+      const [outcome] = await asyncJobCancelHandler(request);
```
ThatOneBro, line 53: We call `asyncJobCancelHandler` for this request and look at the outcome. The only special case is that we want `allOk` to translate to a 202 and an "accepted" `OperationOutcome`

mattwiller (approved):
Looks good to me, just one minor nit to address before merging

PR #5588, opened 2024-11-25
Dependency upgrades 2024-11-25

medplumbot:
Dependency upgrades


lien-dkseo:
for react-router v7,
According to the migration guide on the React Router website,

https://reactrouter.com/upgrading/v6#upgrade-to-v7

`> Replace react-router-dom with react-router`
```
npm uninstall react-router-dom
npm install react-router@latest
```

`> Update imports`
```
-import { useLocation } from "react-router-dom";
+import { useLocation } from "react-router";
```

codyebberson:
Thanks @lien-dkseo !

packages/core/src/utils.test.ts
```diff
@@ -308,13 +308,13 @@ describe('Core Utils', () => {
...
     const input = new Uint32Array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
-    const output = arrayBufferToHex(input.buffer);
+    const output = arrayBufferToHex(input);
     expect(output).toBe('000102030405060708090a');
   });
 
   test('Convert ArrayBuffer to base-64 encoded string', () => {
     const input = new Uint32Array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
-    const output = arrayBufferToBase64(input.buffer);
+    const output = arrayBufferToBase64(input);
```
ThatOneBro, line 317: Is this correct?

`new Uint8Array(input.buffer)` is not the same as `new Uint8Array(input)`
 
Screenshot 2024-11-25 at 10 49 28 PM

`new Uint8Array(input)` goes value by value and truncates them to fit into a byte, `new Uint8Array(input.buffer)` reinterprets the raw bytes as is, just applying a different "view" to it

ThatOneBro: See #5599

codyebberson: Yeah... I don't know what the correct behavior should be.

Changing from `input.buffer` to `input` is actually a revert back to the originally behavior.

Here is the finalized diff now, which just fixes the type error:

image

tbh, I don't know why TypeScript didn't catch the original errors a long time ago, but so it is

My intent with this PR was simply to upgrade dependencies without affecting the current behavior.  My preference would be to get this merged, and then have a conversation about what the "right" behavior should be.

#5599 looks reasonable to me.  I don't know if we actually need that in the wild though, or do we only need it for those 2 tests?  If it's just for the tests, maybe we should just simplify the behavior?

ThatOneBro: I think this is the simple one because `TypedArray` extends `ArrayBuffer`, so the type will always accept both even though one is a view on top of the other. Either way, if you pass in an a `TypedArray` and not its underlying `.buffer`, you are truncating the output and it leads to an incorrect result.

The test expected an incorrect result for the `Uint32Array` we were passing in. In my opinion, this check for if the input is a view makes the implementation simpler. It's worth noting this is exported in core so I think we should make the function easy to use rather than dead simple but with a footgun as it currently is, given the test itself was even wrong.

ThatOneBro: https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#support-for---target-es2024-and---lib-es2024

I think if we are planning to accept both `TypedArray` and `ArrayBuffer`, this should be `ArrayBufferLike` and not just `ArrayLike` to be more specific.

I think that covers both

ThatOneBro: Actually, what happens if we have to use backwards incompatible types? That means users need to be on 5.7 too

codyebberson (approved):

codyebberson (approved):

PR #5594, opened 2024-11-25
[Photon Health] Create `MedicationDispense` from Fills when syncing from Photon

ksmith94:
This adds functionality to the sync from photon bot to handle fills that are included in a patient's medical history or current prescriptions, by creating `MedicationDispense` resources from them.


examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -267,57 +266,107 @@ export function createAllergies(
...
+      request: { method: 'POST', url: 'MedicationRequest' },
+      resource: prescription,
+    });
+
+    if (photonPrescription.fills) {
+      for (const fill of photonPrescription.fills) {
+        const dispense = await createDispenseResource(fill, medplum, prescriptionReference, patientReference);
+        entries.push({
+          fullUrl: 'urn:uuid:' + randomUUID(),
+          request: { method: 'POST', url: 'MedicationDispense' },
```
rahul1, line 298: We should make these upserts as well

examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -267,57 +266,107 @@ export function createAllergies(
...
+
+export async function createDispenseResource(
+  fill: Fill,
+  medplum: MedplumClient,
+  authorizingPrescription: Reference<MedicationRequest>,
+  patientReference: Reference<Patient>
+): Promise<MedicationDispense> {
+  const { codes, name } = fill.treatment;
+  const medicationElement = await getMedicationElement(medplum, codes.rxcui, name);
+  const medicationDispense: MedicationDispense = {
```
rahul1, line 361: add `meta.source` here

rahul1 (changes requested):

examples/medplum-photon-integration/src/bots/sync-patient-from-photon.ts
```diff
@@ -309,83 +307,140 @@ export function createAllergies(
...
-): Promise<MedicationRequest[] | undefined> {
+): Promise<BundleEntry<MedicationDispense | MedicationRequest>[] | undefined> {
   if (!photonPrescriptions || photonPrescriptions.length === 0) {
     return undefined;
   }
 
-  const prescriptions: MedicationRequest[] = [];
+  const entries: BundleEntry<MedicationDispense | MedicationRequest>[] = [];
   for (const photonPrescription of photonPrescriptions) {
-    if (await checkForExistingPrescription(medplum, photonPrescription)) {
```
mattlong, line 323: I'm not familiar with Photon, so bear with me here, but it seems like the new version doesn't have any way of checking for existing resources or upserting existing `MedicationRequest` resources. Is there something that prevents duplicates from being made after repeated calls of this bot?

rahul1: I think you're correct @mattlong . @ksmith94 Is there a reason you're not upserting the `MedicationRequest` resource?

ksmith94: @rahul1 @mattlong Agreed this should be upserted. I think I may have removed this when I was skipping any existing patients, so there was no need to dedupe then. I'll add it back

rahul1 (approved):

PR #5599, opened 2024-11-26
fix(core): allow raw buf and views in arrayBuffer fns

ThatOneBro:
Somehow our tests have been passing for a long time with incorrect assertions.

This is because ArrayBuffer accepts both TypedArray (eg. Uint32Array), ie. views on ArrayBuffer, as well as raw bytes ArrayBuffer.

Suddenly in the last dependency upgrade, the tests actually failed, which is actually correct. The strings we have been producing are actually truncated.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -726,6 +726,10 @@ for (let n = 0; n < 256; n++) {
  * @returns The resulting hex string.
  */
 export function arrayBufferToHex(arrayBuffer: ArrayBuffer): string {
+  // We want to allow both views and raw ArrayBuffers, so we conditionally "unwrap" the underlying ArrayBuffer
+  // If `arrayBuffer` was a view
+  // Otherwise, each of the views "elements" (ie. a uint32) will be truncated to a uint8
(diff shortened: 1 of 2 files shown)


packages/core/src/utils.ts
```diff
@@ -726,7 +726,12 @@ for (let n = 0; n < 256; n++) {
  * @returns The resulting hex string.
  */
 export function arrayBufferToHex(arrayBuffer: ArrayBuffer): string {
-  const bytes = new Uint8Array(arrayBuffer);
+  // We want to allow both views and raw ArrayBuffers, so we conditionally "unwrap" the underlying ArrayBuffer
+  // If `arrayBuffer` was a view
+  // Otherwise, each of the views "elements" (ie. a uint32) will be truncated to a uint8
+  // And you end up with an incorrect shortened result of the same number of bytes as there were elements in the original TypedArray
+  const buffer = ArrayBuffer.isView(arrayBuffer) ? arrayBuffer.buffer : arrayBuffer;
+  const bytes = new Uint8Array(buffer);
```
mattwiller, line 734: Could this logic be extracted into a utility function called in both exported functions?

mattwiller (approved):

PR #5601, opened 2024-11-27
OAuthPage UI customizations

techdavidy:
fixes https://github.com/orgs/medplum/projects/1/views/17?pane=issue&itemId=69935428&issue=medplum%7Cmedplum%7C4805

- Added welcome string and logo as `Attachment` to ClientApplication.
- New endpoint /auth/clientinfo/:clientId returns welcome string and logo attachment.
- Modified OAuthPage to use new endpoint and display custom UI.

Screenshot 2024-11-27 at 5 00 45 PM


packages/app/src/OAuthPage.test.tsx
```diff
@@ -101,4 +101,29 @@ describe('OAuthPage', () => {
...
+
+    await setup('/oauth?client_id=123');
+    await waitFor(() => expect(medplum.get).toHaveBeenCalledWith('/auth/clientinfo/123'));
+    expect(screen.getByText('Test Client')).toBeInTheDocument();
+    const logo = screen.getByAltText('Test Client logo');
+    expect(logo).toBeInTheDocument();
+    expect(logo).toHaveAttribute('src', 'https://example.com/logo.png');
+  });
+
+  test('Fetch and render default info', async () => {
```
techdavidy, line 120: Mock an empty response and validate default info is rendered.

packages/server/src/auth/clientinfo.ts
```diff
@@ -0,0 +1,24 @@
...
+import { asyncWrap } from '../async';
+import { sendOutcome } from '../fhir/outcomes';
+import { getSystemRepo } from '../fhir/repo';
+import { rewriteAttachments, RewriteMode } from '../fhir/rewrite';
+
+export const clientInfoHandler: RequestHandler = asyncWrap(async (req: Request, res: Response) => {
+  const { clientId } = req.params;
+  const systemRepo = getSystemRepo();
+
+  try {
```
codyebberson, line 13: You can safely skip this `try`/`catch`.  We have built-in express middleware that will automatically catch the error, and return the correct 404 error response.

packages/server/src/auth/clientinfo.ts
```diff
@@ -0,0 +1,24 @@
...
+    const client = await systemRepo.readResource<ClientApplication>('ClientApplication', clientId);
+    res.status(200).json(
+      await rewriteAttachments(RewriteMode.PRESIGNED_URL, systemRepo, {
+        welcomeString: client.signInForm?.welcomeString,
+        logo: client.signInForm?.logo,
+      })
+    );
+  } catch (_err) {
+    sendOutcome(res, badRequest('Client not found'));
```
codyebberson, line 22: As mentioned above, you can skip this entire `catch` block.

But, for future reference, the `sendOutcome()` method will set the HTTP status code based on the `OperationOutcome` value.  When using `badRequest`, the HTTP status code will be a generic 400.  If you want a 404 status code, then you need to use the special `notFound` `OperationOutcome`, such as:

```ts
import { notFound } from '@medplum/core`;
```

That's just an FYI.  You can safely delete this line and the `try`/`catch`.

packages/server/src/auth/clientinfo.ts
```diff
@@ -0,0 +1,24 @@
...
+import { rewriteAttachments, RewriteMode } from '../fhir/rewrite';
+
+export const clientInfoHandler: RequestHandler = asyncWrap(async (req: Request, res: Response) => {
+  const { clientId } = req.params;
+  const systemRepo = getSystemRepo();
+
+  try {
+    const client = await systemRepo.readResource<ClientApplication>('ClientApplication', clientId);
+    res.status(200).json(
+      await rewriteAttachments(RewriteMode.PRESIGNED_URL, systemRepo, {
```
codyebberson, line 16: Nice 👍

packages/app/src/OAuthPage.tsx
```diff
@@ -1,20 +1,58 @@
...
   const navigate = useNavigate();
   const [params] = useSearchParams();
-
+  const medplum = useMedplum();
+  const scope = params.get('scope') || 'openid';
+  const [welcomeString, setWelcomeString] = useState<string | null>('Medplum');
+  const [logoUrl, setLogoUrl] = useState<string | null>(null);
   const clientId = params.get('client_id');
+
+  useEffect(() => {
```
codyebberson, line 29: (Correct me if I'm wrong here) I think this will show a flash of the Medplum logo while the effect is running, yeah?

I think we should probably block rendering of the logo and the welcome string until the client info is loaded.

(Please correct me if that's incorrect).

packages/app/src/OAuthPage.tsx
```diff
@@ -1,20 +1,58 @@
...
+import { useEffect, useState } from 'react';
+import { showNotification } from '@mantine/notifications';
+
+interface LogoInfo {
+  contentType: string;
+  url: string;
+  title: string;
+}
+
+interface ClientInfo {
```
codyebberson, line 15: I think we can replace these types with the new auto-generated types in `ClientApplication.d.ts`.  This is basically the same as `ClientApplicationSignInForm`

```ts
import { ClientApplicationSignInForm } from '@medplum/fhirtypes';
```

That will be less code, and will also strongly enforce that they stay in sync with each other.

packages/app/src/OAuthPage.tsx
```diff
@@ -1,20 +1,58 @@
...
+  welcomeString: string | null;
+}
 
 export function OAuthPage(): JSX.Element | null {
   const navigate = useNavigate();
   const [params] = useSearchParams();
-
+  const medplum = useMedplum();
+  const scope = params.get('scope') || 'openid';
+  const [welcomeString, setWelcomeString] = useState<string | null>('Medplum');
```
codyebberson, line 25: We can avoid 2 calls to `setState` if we flatten these into a single client info state:

```ts
const [clientInfo, setClientInfo] = useState<ClientApplicationSignInForm>();
```

packages/app/src/OAuthPage.tsx
techdavidy: The flash would happen.
Blocking rendering logo and welcome screen.

codyebberson:
Deploying to staging: https://github.com/medplum/medplum/actions/runs/12098813196

PR #5602, opened 2024-11-29
findResourceById

AbhinavMir:
test123


codyebberson:
Thanks @AbhinavMir !  Would you be interested in adding a test case as well?

AbhinavMir:
@codyebberson all good if I re-open? I was swamped @ work and forgot about this haha, but I'm good to add tests too. Unless there's already a better PR candidate! 

I'll continue work here: https://github.com/medplum/medplum/pull/7156

PR #5613, opened 2024-12-02
Chained search sort fix

### LINKED ISSUE #5499: Using chained search or _filter in query with _sort and _count breaks expected sorting.
Hello again,

## Background

I have been doing a lot of testing recently and came across this issue while learning chained search and _filter in the standard.
All testing here was done against a Synthea dataset.

I initially had a query with the following parameters.
```
   {
        subject: ref,
        _count: 20,
        category: 'BG,CG,CH,CP,GE,HM,IMM,LAB,MB,MCB,MYC,PAR,PAT,PF,SR,TX,VR,URN',
        status: 'final',
        _offset: 0,
        _sort: -date,
        _total: accurate
    }
```
This would yield a list of `DiagnosticReport` resources presorted by the date time in descending order.

I then tried filtering with a chained search like so

```
   {
        subject: ref,
        _count: 20,
        category: 'BG,CG,CH,CP,GE,HM,IMM,LAB,MB,MCB,MYC,PAR,PAT,PF,SR,TX,VR,URN',
        status: 'final',
        'result:Observation.category': 'http://terminology.hl7.org/CodeSystem/observation-category|laboratory',
        _offset: 0,
        _sort: -date,
        _total: accurate
    }
```

Now, each block of 20 resource instances are sorted by the date time in descending order (possibly because of a sorting function after the query), but the global sorting is no longer valid.
Meaning, if I set the _offset to 20, the next page of results does not appear to match the expected order.

No difference was spotted with the usage of _filter

```
   {
        subject: ref,
        _count: 20,
        category: 'BG,CG,CH,CP,GE,HM,IMM,LAB,MB,MCB,MYC,PAR,PAT,PF,SR,TX,VR,URN',
        status: 'final',
        _filter: 'result.category eq http://terminology.hl7.org/CodeSystem/observation-category|laboratory',
        _offset: 0,
        _sort: -date,
        _total: accurate
    }
```

In fact, the results and sorting are comparable.

Meaning, 

1. Imaging you have a some list L [ R1(2008/10/11), R2(2006/10/11), R3(2007/10/11), R4(2008/10/11)]
2. It sorted like L1 [ R1(2008/10/11), R4(2008/10/11)] and L2 [R3(2007/10/11), R2(2006/10/11)]
3. With chained search or _filter, the new sorting is more like L1 [ R1(2008/10/11), R3(2007/10/11)] and L2 [R4(2008/10/11), R2(2006/10/11)] during pagination.

## Raw FHIR Query
`subject=Patient%2F4b7f03ae-6696-4e73-9ca7-ba6b3a201fce&_count=20&category=BG%2CCG%2CCH%2CCP%2CGE%2CHM%2CIMM%2CLAB%2CMB%2CMCB%2CMYC%2CPAR%2CPAT%2CPF%2CSR%2CTX%2CVR%2CURN&status=final&result%3AObservation.category%3Anot=http%3A%2F%2Fterminology.hl7.org%2FCodeSystem%2Fobservation-category%7Claboratory&_offset=0&_sort=-date&_total=accurate`

image

## Sample Synthea case

[Rozella39_Mertz280_ef167059-cef0-12c4-49db-993ca3a20c01.zip](https://github.com/user-attachments/files/17637929/Rozella39_Mertz280_ef167059-cef0-12c4-49db-993ca3a20c01.zip)

### Without chained search
image
image
image

### With chained search

image
image
image

I expected the survey entries to drop but the order to remain the same.

## Possible Scenarios

- Sorting is done after isolating n count of entries from unsorted match pool.
- Sorting is applied globally but perhaps the chained search returns the final match pool unsorted.
- Other?

I will need to find out the exact SQL query pushed to the database to get an idea of how the FHIR query is getting transformed. Maybe it is as simple as a subtle order of operations issue.
I will try to report back what I find later in the week, but reporting issue now in case someone with experience recognizes the issue right away.

## Notes

- I tried using _query and _has for the filtering. _has may be wrong for what I was trying to do, but both of these special parameters are not recognized by the server.

luissantosHCIT:
Resolves #5499 by making sure that the distinct clause includes the sort fields. It also makes sure the sort fields are added before the id field to ensure we sort the unique rows by the requested criteria.

Opened PR for review by @mattwiller.


mattwiller:
I tried copying the [test cases from my PR](https://github.com/medplum/medplum/pull/5513/files#diff-20e89f55d78c99c46876e4fb2dcd3447ac051e4c8e9f91e5f89829e71cef6a53) over, and this one currently fails:

```ts
test('Chained search with negated filter', () => withTestContext(async () => {
  config.chainedSearchWithReferenceTables = true;

  // Create linked resources
  const link1 = await repo.createResource<Patient>({
    resourceType: 'Patient',
    identifier: [{ value: randomUUID() }],
  });
  const link2 = await repo.createResource<Patient>({
    resourceType: 'Patient',
    identifier: [{ value: randomUUID() }],
  });

  const patientIds: string[] = [];
  for (let i = 1; i < 10; i++) {
    const middlePatient = await repo.createResource<Patient>({
      resourceType: 'Patient',
      link: [
        { type: 'seealso', other: createReference(link1) },
        { type: 'seealso', other: createReference(link2) },
      ],
    });
    const patient = await repo.createResource<Patient>({
      resourceType: 'Patient',
      birthDate: '1994-11-0' + i,
      link: [{ type: 'seealso', other: createReference(middlePatient) }],
    });
    patientIds.unshift(patient.id as string);
  }

  const result = await repo.search(
    parseSearchRequest(`Patient?link:Patient.link:Patient.identifier:not=${randomUUID()}&_sort=-birthdate`)
  );
  expect(result.entry?.map((e) => (e.resource as Patient).birthDate)).toEqual([
    '1994-11-09',
    '1994-11-08',
    '1994-11-07',
    '1994-11-06',
    '1994-11-05',
    '1994-11-04',
    '1994-11-03',
    '1994-11-02',
    '1994-11-01',
  ]);
  expect(result.entry?.map((e) => e.resource?.id)).toEqual(patientIds);
}));
```

The generated SQL from that test case is:
```sql
SELECT 
  DISTINCT ON (
    "Patient"."birthdate", "Patient"."id"
  )
  "Patient"."id", "Patient"."content" 
FROM "Patient" 
  LEFT JOIN "Patient_References" AS "T1" ON (
    "T1"."resourceId" = "Patient"."id" 
    AND "T1"."code" = $1
  ) 
  LEFT JOIN "Patient" AS "T2" ON "T2"."id" = "T1"."targetId" 
  LEFT JOIN "Patient_References" AS "T3" ON (
    "T3"."resourceId" = "T2"."id" 
    AND "T3"."code" = $2
  ) 
  LEFT JOIN "Patient" AS "T4" ON "T4"."id" = "T3"."targetId" 
WHERE (
  "Patient"."deleted" = $3 
  AND (
    "Patient"."compartments" IS NOT NULL 
    AND "Patient"."compartments" && ARRAY[$4,$5]::UUID[]
  ) 
  AND NOT EXISTS(
    SELECT 
      "Patient_Token"."resourceId" 
    FROM 
      "Patient_Token" 
    WHERE 
      (
        "T4"."id" = "Patient_Token"."resourceId" 
        AND "Patient_Token"."code" = $6 
        AND "Patient_Token"."value" = $7
      )
  )
) 
ORDER BY 
  "Patient"."birthdate" DESC, 
  "Patient"."id", 
  "Patient"."birthdate" DESC 
LIMIT 21;
```

It appears that the generated SQL is missing a `WHERE` clause like `"T4"."id" IS NOT NULL` to ensure that the reference chain is valid even though the `EXISTS()` is negated.  Manually adding that clause seemed to give the correct query results for me.  I'm going to load up a few hundred thousand Synthea patients into my local machine and check out what the query plan looks like at that scale

luissantosHCIT:
@mattwiller Is that a new test case and is it addressing another issue with the search functionality I was not considering (outside my ticket)? 

My original queries were not about negation, but now I think I understand why you were looking at a negated version of my queries before.

Since my changes focus on the distinct portion of the query, I am guessing I would need to look at your PR and consolidate changes you may have that deal with negation issues not covered here. Am I understanding this correctly?

mattwiller:
@luissantosHCIT I did some investigation on how the query produced by this PR performs compared to the other PR, with my findings written up [in the associated issue thread](https://github.com/medplum/medplum/issues/5499#issuecomment-2513583588), but the __tl;dr__ is that I think this PR actually produces a slightly more complex query plan than the other.  For now, I think I may prefer to go with [the other solution](https://github.com/medplum/medplum/pull/5513); but I'm open to arguments for alternatives

mattwiller:
Closing in favor of #5513 (see [discussion](https://github.com/medplum/medplum/issues/5499#issuecomment-2513583588))

PR #5618, opened 2024-12-03
Use exponential backoff for bot deploy retries

codyebberson:
Still occasionally running into these errors:

```
[Error]: The operation cannot be performed at this time. An update is in progress for resource: arn:aws:lambda:us-east-1:647991932601:function:medplum-bot-lambda-...
```

Before, we used constant 1 second retry interval.

After, using exponential backoff of 1 sec, 2 sec, 4 sec, etc.

### DIFF
--- a/packages/server/src/fhir/operations/deploy.ts
+++ b/packages/server/src/fhir/operations/deploy.ts
@@ -32,7 +32,6 @@ export async function deployHandler(req: FhirRequest): Promise<FhirResponse> {
   }
 
   try {
-    const code = req.body.code as string | undefined;
     let updatedBot: Bot | undefined;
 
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/operations/deploy.ts
```diff
@@ -32,7 +32,6 @@ export async function deployHandler(req: FhirRequest): Promise<FhirResponse> {
   }
 
   try {
-    const code = req.body.code as string | undefined;
```
codyebberson, line 35: Drive-by fix.  This is the exact same as line 25 above, so this line is redundant and confusing.

ThatOneBro (approved):

PR #5620, opened 2024-12-04
Fixes #5616 - replace add/save buttons on search filter editor

### LINKED ISSUE #5616: Applying filters via the `Filters` modal is confusing (UX issue)
~~Data points: `PractitionerRole.organization`, `PractitionerRole.practitioner`, and `Patient.organization`~~

Screenshot 2024-12-02 at 3 01 51 PM

Filling out this modal and then pressing ok results in the following URL string:
https://app.medplum.com/Patient?_count=20&_fields=id,_lastUpdated,name,birthDate,gender&_sort=-_lastUpdated

Which is the same as before filling out the modal.

It's worth noting that applying the same filters via the "Quick Filters" menu above the column does work:
Screenshot 2024-12-02 at 3 08 21 PM

EDIT: Looks like it's more of a UX "bug", the modal does function properly if you press `Add` before pressing `OK`.

Changing the name of this issue to reflect that this is more of a design thing

codyebberson:
Before:
* The search filter editor dialog showed an "in progress" filter row
* But that row was only added if the user hit "Add" first
* That led to a common source of customer confusion

image

After
* No default "in progress" filter row
* You have to hit "Add Filter" add a new filter
* All rows are in "edit" mode

image

### DIFF
--- a/packages/react/src/SearchFilterEditor/SearchFilterEditor.test.tsx
+++ b/packages/react/src/SearchFilterEditor/SearchFilterEditor.test.tsx
@@ -51,6 +51,10 @@ describe('SearchFilterEditor', () => {
       />
     );
 
+    await act(async () => {
+      fireEvent.click(screen.getByText('Add Filter'));
+    });
+
     const fieldInput = screen.getByTestId('filter-field');
     expect(fieldInput).toBeInTheDocument();
     expect(fieldInput).toHaveValue('');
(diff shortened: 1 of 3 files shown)


ThatOneBro (approved):
Nice, LGTM

codyebberson:
Hold - there's some busted behavior when working with multiple filters on the same search parameter.

ThatOneBro (approved):

PR #5624, opened 2024-12-04
Add support for enableWhenExpression

techdavidy:
Several lines modified because isQuestionEnabled now has QuestionnaireResponse as parameter. The new parameter required to change tests.

In the example, date is presented when 
q1= 'female' and q2=true

https://github.com/user-attachments/assets/9f3db591-fd62-4700-9ed4-eda2a0c7626d

### DIFF
--- a/packages/react/src/utils/questionnaire.test.ts
+++ b/packages/react/src/utils/questionnaire.test.ts
@@ -31,7 +31,7 @@ test('isQuestionEnabled', () => {
       },
       {
         resourceType: 'QuestionnaireResponse',
-        status: "completed",
+        status: 'completed',
         item: [
           {
             linkId: 'q1',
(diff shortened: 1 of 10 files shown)


packages/react/src/utils/questionnaire.test.ts
```diff
@@ -955,12 +1102,84 @@ describe('isQuestionEnabled', () => {
...
+              linkId: 'q1',
+              answer: [{ valueCoding: { code: 'MEDPLUM123' } }],
+            },
+          ],
+        }
+      )
+    ).toBe(false);
+  });
+
+  test('expression evaluation', () => {
```
techdavidy, line 1119: Actual tests, other changes are just refactoring.

codyebberson (approved):
Nice, clean, thanks @techdavidy ! 🙏

mattlong (approved):

PR #5625, opened 2024-12-04
Return undefined reference string when resource has no ID

### LINKED ISSUE #4378: Consider constraining allowed types of input to `getReferenceString`
Currently `getReferenceString` allows an input of `Resource | Reference`. However because all the properties on `Reference` are optional, this function allows any object to be passed in.

For example, the following TypeScript code compiles without error:

```ts
const string = getReferenceString({});
console.log('reference string is', string);
```

and prints the following:

```ts
reference string is undefined/undefined
```

this tripped us up in production recently.

---

would you consider potentially changing the signature of this function to enforce the string value being present on the passed-in input? for example something like so?

```ts
export function getReferenceString(
  input:
    | (Reference & { reference: string })
    | (Resource & { resourceType: ResourceType; id: string }),
): string {
  if (isReference(input)) {
    return input.reference;
  }
  return `${input.resourceType}/${input.id}`;
}
```

this would result in the above code sample no longer compiling successfully:

```ts
const string = getReferenceString({});
```

```ts
error TS2345: Argument of type '{}' is not assignable to parameter of type '(Reference<Resource> & { reference: string; }) | (Resource & { resourceType: "Appointment" | "AccessPolicy" | "Account" | "ActivityDefinition" | ... 157 more ... | "VisionPrescription"; id: string; })'.
````

### PR DESCRIPTION
**Pull request by mattwiller**
Resolves #4378

### CODE REVIEW

**Review by codyebberson — COMMENTED** [review]
Nice, this turned out really clean.

File: packages/server/src/fhir/repo.ts [hunk 1]
```diff
@@ -573,7 +574,11 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     }
   }
 
-  private async updateResourceImpl<T extends Resource>(resource: T, create: boolean, versionId?: string): Promise<T> {
+  private async updateResourceImpl<T extends Resource>(
+    resource: T,
+    create: boolean,
+    versionId?: string
+  ): Promise<WithId<T>> {
     const { resourceType, id } = resource;
```
mattlong, line 582: This is one place where the `isResourceWithId` narrowing function could be used:

```suggestion
    if (!isResourceWithId(resource)) {
      throw new OperationOutcomeError(badRequest('Missing id'));
    }
    if (!validator.isUUID(resource.id)) {
      throw new OperationOutcomeError(badRequest('Invalid id'));
    }

    const { resourceType, id } = resource;
```

packages/server/src/fhir/repo.ts
```diff
@@ -606,8 +611,8 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       }
     }
 
-    let updated = await rewriteAttachments<T>(RewriteMode.REFERENCE, this, {
-      ...this.restoreReadonlyFields(resource, existing),
+    let updated = await rewriteAttachments(RewriteMode.REFERENCE, this, {
+      ...this.restoreReadonlyFields(resource as WithId<T>, existing),
```
mattlong, line 615: Then this cast wouldn't be needed.

packages/server/src/fhir/repo.ts
ThatOneBro: +1, this was a suggestion I made in passing yesterday

packages/core/src/utils.ts
```diff
@@ -60,19 +60,7 @@ export function isResourceWithId<T extends Resource>(
...
-  // Must contain `resourceType` and `id` keys
-  if (!('resourceType' in resource && 'id' in resource)) {
-    return false;
-  }
-  // Optionally check resource type
-  if (resourceType && resource.resourceType !== resourceType) {
-    return false;
-  }
-  return true;
+  return isResource(resource, resourceType) && 'id' in resource;
```
ThatOneBro, line 63:
```suggestion
  return isResource(resource, resourceType) && 'id' in resource && typeof resource.id === 'string';
```
Probably want to check typeof on `resource.id` since `in` only checks that the field exists on the object and not that it is defined
Screenshot 2024-12-06 at 10 26 23 AM

PR #5630, opened 2024-12-05
Don't allow control chars in strings

### LINKED ISSUE #5629: Reject strings with control characters
From the R4 spec for the [string primitive type](https://hl7.org/fhir/r4/datatypes.html#primitive):

> Strings SHOULD not contain Unicode character points below 32, except for u0009 (horizontal tab), u0010 (carriage return) and u0013 (line feed).

mattlong:
From the R4 spec for the [string primitive type](https://hl7.org/fhir/r4/datatypes.html#primitive):

> Strings SHOULD not contain Unicode character points below 32, except for u0009 (horizontal tab), u0010 (carriage return) and u0013 (line feed).

Fixes #5629


ThatOneBro (approved):

mattwiller:
Should this apply to all other string-based types, like `uri`?

Issue #1938, opened 2023-04-28
Support multiple selections in QuestionnaireForm drop-downs


codyebberson:

Aquilaafuadajo:
Hi @codyebberson, please can you assign this to me? I've worked on questionnaire builder like this in a previous application and I'd love to help add this.

PR #5643, opened 2024-12-09
Support multi select

techdavidy:
fixes https://github.com/medplum/medplum/issues/1938

Screenshot 2024-12-09 at 11 48 39 AM

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
@@ -5,6 +5,7 @@ import {
   formatCodeableConcept,
   formatCoding,
   getElementDefinition,
+  HTTP_HL7_ORG,
   stringify,
   TypedValue,
 } from '@medplum/core';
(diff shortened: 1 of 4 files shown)


packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -436,6 +472,14 @@ function isDropDownChoice(item: QuestionnaireItem): boolean {
   );
 }
 
+function isMultiSelectChoice(item: QuestionnaireItem): boolean {
+  return !!item.extension?.some(
```
mattlong, line 476: The `getExtensionValue` helper could be useful here.

mattlong (approved):
LGTM, one minor suggestion

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
techdavidy: was doing some testing but left as is to keep it consistent with the rest of the file

Issue #1245, opened 2022-12-10
Questionnaire Builder strips out whitespace in question description


reshmakh:
When constructing a questionnaire, this can produce unexpected behavior.  

@rahul1 if you could provide repro details, that would be appreciated.

rahul1:
![Uploading erasing-whitespace.gif…]()

rahul1:
![erasing-whitespace](https://github.com/user-attachments/assets/c0cb5413-b4c5-40be-b108-e1258dc27296)

PR #5648, opened 2024-12-10
Questionnaire Builder preserve  whitespace in question

techdavidy:
Fixes https://github.com/medplum/medplum/issues/1245

https://github.com/user-attachments/assets/70cabfdb-d758-43c5-b1be-213005281213


codyebberson (commented):
Thanks @techdavidy !

One small comment on using `cx`/`clsx` rather than string concatenation.

I tested it on the Vercel preview, and everything feels good.

codyebberson (approved):

PR #5651, opened 2024-12-10
Update server README for v4

mattwiller:


mattlong (approved):
@codyebberson can clarify, but I believe we're officially going to support Postgres 13-17 in v4.

mattlong (commented):
To finish the thought, so I think this should be `13` instead.

PR #5657, opened 2024-12-11
feat(react): add C-CDA viewer to `AttachmentDisplay`

ThatOneBro:
Addresses #4726 and builds towards ONC attestation.

Screenshot 2024-12-10 at 11 16 06 PM


packages/react/src/utils/dom.ts
```diff
@@ -35,3 +35,35 @@ export function isCheckboxCell(el: Element): boolean {
...
+ * Sends a structured command to the iframe using postMessage.
+ *
+ * Normally postMessage implies global event listeners. This method uses
+ * MessageChannel to create a message channel between the iframe and the parent.
+ * @param frame - The receiving IFrame.
+ * @param command - The command to send.
+ * @returns Promise to the response from the IFrame.
+ * @see https://advancedweb.hu/how-to-use-async-await-with-postmessage/
+ */
+export async function sendCommand<T = string, R = unknown>(frame: HTMLIFrameElement, command: Command<T>): Promise<R> {
```
ThatOneBro, line 54: Factored this function out since it was used in both `app` and now in `CcdaDisplay`; this also made it easier to mock

codyebberson (approved):

PR #5662, opened 2024-12-12
feat(server): log more stats per table, allow tables to be specified

ThatOneBro:
This enables logging stats about the per-table autovacuum settings, as well as number of live and dead tuples in a table, when the last auto-vacuum and auto-analyzes ran, as well as the ability to specify which tables to include in the results.

### DIFF
--- a/packages/server/src/fhir/operations/dbstats.ts
+++ b/packages/server/src/fhir/operations/dbstats.ts
@@ -39,6 +39,9 @@ export async function dbStatsHandler(req: FhirRequest): Promise<FhirResponse> {
   const params = parseInputParameters<{ tableNames?: string }>(operation, req);
 
   const client = getDatabasePool(DatabaseMode.WRITER);
+
+  const tableNames = params.tableNames?.split(',').map((name) => name.trim());
+
   const sql = `SELECT * FROM (
     SELECT
       i.relname AS table_name,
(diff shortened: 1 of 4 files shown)


packages/server/src/fhir/operations/dbstats.ts
```diff
@@ -15,27 +15,57 @@ const operation: OperationDefinition = {
...
+      i.idx_tup_read AS index_entries_read,
+      i.idx_tup_fetch AS index_rows_fetched,
+      c.reloptions AS table_level_overrides,
+      n_live_tup AS live_tuples,
+      n_dead_tup AS dead_tuples,
+      last_autovacuum,
+      last_autoanalyze
+    FROM pg_stat_user_indexes i JOIN pg_class c ON i.relid = c.oid JOIN pg_stat_user_tables u ON i.relname = u.relname
+    ${
+      params.tableNames
```
codyebberson, line 62: Can we move this to SQL params rather than string concatenation?  I realize this is a super admin endpoint, so the risk is relatively low, but more of a good hygiene thing.

Or, at the very least, some basic string sanitization

ThatOneBro: FYI you can use `WHERE <column> = ANY($1::text[])` to compare against parameterized arrays

mattwiller (approved):

Issue #2726, opened 2023-08-25
Implement Questionnaire Extension: calculatedExpression


jamestouri:
This calculated value is based on an expression that gets evaluated and can allow for a deeper analysis for next questions in the questionnaire flow. 

http://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-calculatedExpression.html

PR #5668, opened 2024-12-12
Implement Questionnaire Extension: calculatedExpression

techdavidy:
Solves https://github.com/medplum/medplum/issues/2726

https://github.com/user-attachments/assets/f7ac2929-ec55-437b-977d-8c27abfd7161


packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -69,14 +69,29 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
       );
 
-      const newResponse: QuestionnaireResponse = {
+      const tempResponse: QuestionnaireResponse = {
         resourceType: 'QuestionnaireResponse',
         status: 'in-progress',
         item: mergedItems,
       };
+
+      const updatedItems = evaluateCalculatedExpressions(questionnaire?.item ?? [], tempResponse);
```
techdavidy, line 78: Once responses are added to QuestionnaireResponse. Then go through all questionnaire.item checking whether they have an expression to be evaluated and it is does use tempResponse to evaluate.

packages/react/src/utils/questionnaire.ts
```diff
@@ -95,6 +96,96 @@ export function isQuestionEnabled(
...
+
+        return {
+          id: item?.id,
+          linkId: item?.linkId,
+          text: item.text,
+          answer: [answer],
+        };
+      }
+    })
+    .filter((item) => item !== null) as QuestionnaireResponseItem[];
```
mattlong, line 129: I have a bit of a vendetta against type casts. I think you could avoid it here by making the filter function a guard function: `(item): item is QuestionnaireResponseItem => item !== null`
and similarly on your map function above: `(item): QuestionnaireResponseItem | null =>`

packages/react/src/utils/questionnaire.ts
```diff
@@ -95,6 +96,96 @@ export function isQuestionEnabled(
...
+    item,
+    HTTP_HL7_ORG + '/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-calculatedExpression'
+  );
+
+  if (extension) {
+    const expression = extension.valueExpression?.expression;
+    if (expression) {
+      const value = toTypedValue(response);
+      const result = evalFhirPathTyped(expression, [value], { '%resource': value });
+      return result;
```
mattlong, line 183: Could this instead return `result[0]` so that calling functions wouldn't have to assume that this array returned only has one entry?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -61,22 +66,38 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+        const tempResponse: QuestionnaireResponse = {
+          resourceType: 'QuestionnaireResponse',
+          status: 'in-progress',
+          item: mergedItems,
+        };
+
+        const updatedItems = evaluateCalculatedExpressions(questionnaire?.item ?? [], tempResponse);
+        const mergedItemsWithUpdates = mergedItems.map((mergedItem) => {
+          const updatedItem = updatedItems.find((updated) => updated.linkId === mergedItem.linkId);
+          return updatedItem || mergedItem;
```
mattlong, line 87: I think there was some discussion the other day about how to resolve conflicts when a calculated response on the same item that the user answers. It'd be good to add some comments above this code explaining that we decided the calculated response always wins. If there's any excerpts you could add from the FHIR spec to support that decision, that would be helpful.

mattlong (commented):
This directionally seems reasonable to me! I know it's still in progress, but tests around this would be really appreciated since `QuestionnaireForm` is a complicated part of the code.

packages/react/src/utils/questionnaire.ts
techdavidy: good point for sure

packages/react/src/utils/questionnaire.ts
```diff
@@ -95,6 +96,119 @@ export function isQuestionEnabled(
...
+      return value.type === PropertyType.string ? { valueString: value.value } : undefined;
+    case QuestionnaireItemType.attachment:
+      return value.type === PropertyType.Attachment ? { valueAttachment: value.value } : undefined;
+    case QuestionnaireItemType.reference:
+      return value.type === PropertyType.Reference ? { valueReference: value.value } : undefined;
+    case QuestionnaireItemType.quantity:
+    case QuestionnaireItemType.decimal:
+    case QuestionnaireItemType.integer:
+    case QuestionnaireItemType.string:
+      return { valueQuantity: value.value };
```
codyebberson, line 162: Interesting that TypeScript didn't catch this... 😬   It's because `TypedValue.value` is type `any` 🤦 

`valueQuantity` should only be a `Quantity`.  See: https://hl7.org/fhir/R4/questionnaireresponse.html

So it would be something like:

```ts
    case QuestionnaireItemType.quantity:
      return { valueQuantity: value.value };
    case QuestionnaireItemType.decimal:
      return { valueDecimal: value.value };
    case QuestionnaireItemType.integer:
      return { valueInteger: value.value };
    case QuestionnaireItemType.string:
      return { valueString: value.value };
```

Kinda gross and verbose.  There might be a more clever way to do this, because it should be systematic, but an exhaustive `switch` statement is also okay.

packages/react/src/QuestionnaireForm/QuestionnaireFormComponents/QuestionnaireFormGroup.tsx
```diff
@@ -14,6 +14,10 @@ interface QuestionnaireRepeatableGroupProps {
 export function QuestionnaireRepeatedGroup(props: QuestionnaireRepeatableGroupProps): JSX.Element | null {
   const [responses, setResponses] = useState(props.response);
 
+  useEffect(() => {
+    setResponses(props.response);
```
codyebberson, line 18: We should check with @ThatOneBro  on this - It might be fine, but I'm not sure what our philosophy is here in this component, whether `responses` is intended to be a "controlled" component or not.

techdavidy: I had an issue with nested items. 
When questions belonged to the same group/ .item. It would work. 
But I nested several questions in different groups, the calculated values would not refresh in the UI.
useEffect would solve the issue in the QuestionnaireRepeatedGroup would solve the issue.

ThatOneBro: I do remember when this was originally being worked on that the intent was to propagate all changes to the root via `props.onChange` and essentially enforce the hierarchy all being controlled components... whether that is the case today or not, I'm not sure. Should probably be something we review in the near future.

This change seems sane to me at first glance though, seems like it might have been a bug that has been there

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -61,22 +67,35 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+        const newResponse: QuestionnaireResponse = {
+          resourceType: 'QuestionnaireResponse',
+          status: 'in-progress',
+          item: mergedItemsWithUpdates,
+        };
+
+        return newResponse;
+      });
+    },
+    [questionnaire]
```
ThatOneBro, line 97: I think we talked about this before right @mattlong ? Do you remember anything about the conversation we had about memoizing based on the passed in questionnaire?  Not sure if it's relevant here but wanted to flag it

techdavidy: Adding questionnaire as dependency because we need to compare against responses, but given the use case here `questionnaire` wouldn't change causing the hook to re execute, right ?

ThatOneBro: There is a case where the questionnaire's identity is different on every rerender of the top level component:

```ts
<QuestionnaireForm questionnaire={{ resourceType: 'Questionnaire', id: '123'  }} />
```

What will happen if that `useEffect` triggers on every rerender of the parent component?

ThatOneBro (approved):

PR #5672, opened 2024-12-12
Extract server implementation details from SearchParameterDetails

mattlong:
Removes the server-specific `columnName` from `SearchParameterDetails` into a new `SerchParameterImplementation` construct that lives in `packages/server`:

```ts
export interface SearchParameterImplementation extends SearchParameterDetails {
  readonly columnName: string;
  readonly storagePattern: 'column' | 'lookup-table';
}
```

The other field, `storagePattern` is used to be the single source of truth on how a search parameter is implemented. Right now there are only two options, `column` or `lookup-table`, but that will very likely be changing in the near future as more sophisticated search parameter implementations are added.

I'm not very fond of `storagePattern` the more I've looked at it. I think a better name might be `strategy`. Also very open to feedback on naming.

Moves the `migrate.ts` script from `packages/generator` to `packages/server` since all the logic it needs to generate migrations is now available within `server` and writes its output to `server`, collocate it.

One sore spot of this is that `FileBuilder` was moved from `packages/generator` to `packages/core` so it is commonly importable by both `generator` and `server`. Alternatively, it could be left in `generator` and imported in `migrate.ts` with something like the following, but that feels a bit gross. Very open to feedback on this.

```js
import { FileBuilder } from '../../../generator/src/filebuilder';
```


packages/server/src/fhir/lookups/token.ts
```diff
@@ -515,9 +515,11 @@ function buildWhereCondition(
   const parts = splitN(query, '|', 2);
   // Handle the case where the query value is a system|value pair (e.g. token or identifier search)
   if (parts.length === 2) {
-    const systemCondition = new Condition(new Column(tableName, 'system'), '=', parts[0]);
-    return parts[1]
-      ? new Conjunction([systemCondition, buildValueCondition(tableName, operator, caseSensitive, parts[1])])
+    const [system, value] = parts;
+    // TODO is `system || null` correct here?
```
mattwiller: What was the resolution for this?

packages/server/src/fhir/search.ts
```diff
@@ -934,9 +933,14 @@ function buildSearchFilterExpression(
...
-  if (lookupTable) {
-    return lookupTable.buildWhere(selectQuery, resourceType, table, param, filter);
+  const impl = getSearchParameterImplementation(resourceType, param);
+  if (impl.storagePattern === 'lookup-table') {
+    const lookupTable = getLookupTable(resourceType, param);
+    if (lookupTable) {
+      return lookupTable.buildWhere(selectQuery, resourceType, table, param, filter);
+    }
+
+    //TODO: error handling
```
mattwiller, line 943: Is there anything critical to handle here?

mattwiller (approved):

packages/generator/src/index.ts
```diff
@@ -32,14 +33,25 @@ export function main(): void {
   }
 }
 
+class GeneratorFileBuilder extends FileBuilder {
```
codyebberson, line 36: I'd be ok to cut this, especially if the only difference is the "Generated by ..." message.

packages/server/src/fhir/searchparameter.ts
```diff
@@ -0,0 +1,107 @@
...
+): SearchParameterImplementation {
+  const code = searchParam.code;
+  const builder = getSearchParameterDetails(resourceType, searchParam) as ImplementationBuilder;
+
+  const columnName = convertCodeToColumnName(code);
+  builder.columnName = columnName;
+
+  let storagePattern: ImplementationBuilder['storagePattern'] = 'column';
+  if (!searchParam.base?.includes(resourceType as ResourceType)) {
+    // TODO is ignoring this really the right behavior? When does this happen in practice?
```
codyebberson, line 68: It would be straightforward to write a little script "for all resource types, for all search parameters, does this ever happen?"

We do something similar in the "Everything" test in `packages/core/src/search/details.test.ts`.  It's not a lot of data, so it will run fast.

codyebberson (approved):
Overall it looks good to me.

It seems like it's mostly moving code around and cleaning up names, correct?  Are there any functional changes that you're aware of (other than removing `columnName`)?

> I'm not very fond of storagePattern the more I've looked at it. I think a better name might be strategy. Also very open to feedback on naming.

It's an internal variable, so I'm not too worried about it.  `storagePattern` seems fine.  `strategy` might be too generic.

* `storageMode`
* `storageStrategy`
* `searchMode`
* `columnMode`
* `searchStrategy`

> One sore spot of this is that FileBuilder was moved from packages/generator to packages/core so it is commonly importable by both generator and server. Alternatively, it could be left in generator and imported in migrate.ts with something like the following, but that feels a bit gross. Very open to feedback on this.

I'd rather have it in `core` than reaching across normal folder structure boundaries.  

It's only 80 lines, and doesn't have any weird dependencies, so it seems fine.

There's a separate bigger / longer term project to make sure tree-shaking works as expected, which would make this decision totally moot.

See the one comment on `SearchParameter.base`.  I think it's fine, but it should be straightforward to verify.

PR #5683, opened 2024-12-15
Assorted fixes to build-deb action

codyebberson:
Several minor fixes found while working on https://github.com/medplum/medplum/pull/5682

1. Publish deb on release, not just manual `workflow_dispatch`
2. Removed uploading as artifacts
    * Action was broken because using unset `env.MEDPLUM_VERSION`
    * Already uploading to `apt.medplum.com`, which is better anyway
3. Don't disable the systemd service on upgrade, only on remove
4. Automatically start the service on install and upgrade

### DIFF
--- a/scripts/build-deb.sh
+++ b/scripts/build-deb.sh
@@ -302,16 +302,35 @@ addgroup --system $SERVICE_NAME
 adduser --system --ingroup $SERVICE_NAME $SERVICE_NAME
 chown $SERVICE_NAME:$SERVICE_NAME "/var/lib/$SERVICE_NAME"
 
-# Start the service
-systemctl daemon-reload
-systemctl enable $SERVICE_NAME.service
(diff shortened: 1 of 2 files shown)


scripts/build-deb.sh
```diff
@@ -302,16 +302,35 @@ addgroup --system $SERVICE_NAME
...
-systemctl disable $SERVICE_NAME.service
+set -e
+
+# Only fully stop/disable on package removal
+if [ "\$1" = "remove" ]; then
+    systemctl stop $SERVICE_NAME.service
+    systemctl disable $SERVICE_NAME.service
+fi
+
+# For upgrades, we only need to stop the service temporarily
```
ThatOneBro, line 330: 👍

ThatOneBro (approved):

PR #5687, opened 2024-12-17
Specify "authorizedBy" location in Health Gorilla Hook

rahul1:


packages/health-gorilla-react/src/useHealthGorillaLabOrder.test.tsx
```diff
@@ -214,6 +231,33 @@ describe('useHealthGorilla', () => {
...
+      // Make sure the order is valid
+      await act(async () => {
+        result.current.addTest(RWS_AOE_TEST);
+        result.current.setPerformingLab({ resourceType: 'Organization', id: 'Lab-123' });
+        result.current.updateBillingInformation({ billTo: 'patient' });
+      });
+
+      const { serviceRequest } = await result.current.createOrderBundle();
+
+      const expectedValue = isReference(requestingLocation) ? requestingLocation : createReference(requestingLocation);
```
mattlong, line 254: I would suggest moving the expected value as a second field in the `test.each` array, but that might just be a personal preference of mine; not sure which is "better".

```typescript
test.each([
  [{ resourceType: 'Location', id: 'L-123' }, { reference: 'Location/L-123'}],
  ...
])('Requesting Location set', async (requestingLocation, expectedRef) => {
  ...
```

mattlong (approved):
LGTM, just one minor suggestion

PR #5690, opened 2024-12-17
Fixes #5679 - support SCIM PATCH requests

### LINKED ISSUE #5679: Support SCIM PATCH requests
Okta requires certain the following user changes to be made via PATCH requests

* Activating a User
* Deactivating a User
* Syncing the User password

https://developer.okta.com/docs/api/openapi/okta-scim/guides/scim-20/#update-a-specific-user-patch

codyebberson:


packages/server/src/scim/utils.ts
```diff
@@ -244,3 +299,49 @@ export function convertToScimListResponse<T>(Resources: T[]): ScimListResponse<T
...
+ * See: https://www.rfc-editor.org/rfc/rfc7644#section-3.5.2
+ *
+ * @param scimPatch - The original SCIM patch request.
+ * @returns The converted JSONPatch request.
+ */
+export function convertScimToJsonPatch(scimPatch: ScimPatchRequest): Operation[] {
+  return scimPatch.Operations.flatMap((inputOperation) => {
+    const { op, path, value } = inputOperation;
+
+    if (op !== 'add' && op !== 'remove' && op !== 'replace' && op !== 'copy' && op !== 'move' && op !== 'test') {
```
mattlong, line 318: Aren't the only allowed operations "add", "remove", or "replace"?

packages/server/src/scim/utils.ts
```diff
@@ -244,3 +299,49 @@ export function convertToScimListResponse<T>(Resources: T[]): ScimListResponse<T
...
+      throw new Error('Invalid SCIM patch: unsupported operation');
+    }
+
+    if (path) {
+      // This is a normal JSONPatch operation
+      if (!path.startsWith('/')) {
+        return { ...inputOperation, path: `/${path}` } as Operation;
+      }
+      return inputOperation as Operation;
```
mattlong, line 327: Shouldn't this be an error since a path starting with a `/` is not a valid SCIM patch. I suspect you're trying to not be totally strict to the SCIM spec, but then we're in hybrid middle ground which seems like it'd be harder to reason about.

mattlong: Similarly, we aren't validating that the `"schemas":
      ["urn:ietf:params:scim:api:messages:2.0:PatchOp"]` is included.

mattlong (changes requested):
Most of it looks good, mainly just questions about how strict we should be when accepting SCIM patch requests.

packages/server/src/scim/utils.ts
codyebberson: Hm.  Let me go back and check.  I thought SCIM allowed both.

packages/server/src/scim/utils.ts
codyebberson: Hm, good call.  I thought it supported all.  I'll double check the SCIM spec.

packages/server/src/auth/login.test.ts
```diff
@@ -490,4 +490,39 @@ describe('Login', () => {
...
+      scope: 'openid offline',
+      codeChallenge: 'xyz',
+      codeChallengeMethod: 'plain',
+      projectId: project.id,
+    });
+    expect(res.status).toBe(400);
+    expect(res.body.login).toBeUndefined();
+    expect(res.body.code).toBeUndefined();
+    expect(res.body.memberships).toBeUndefined();
+    expect(res.body.issue[0].details.text).toBe('Profile not active');
```
ThatOneBro, line 526: 👍

packages/definitions/dist/fhir/r4/profiles-medplum.json
```diff
@@ -1434,6 +1434,23 @@
               "max" : "*"
             }
           },
+          {
+            "id" : "ProjectMembership.active",
```
ThatOneBro, line 1438: Cool, it's finally here

ThatOneBro (approved):

PR #5695, opened 2024-12-18
feat(client): add `isAuthenticated()` method

### LINKED ISSUE #5660: Add helper to Medplum client for "is authenticated"
User requested: https://medplum.slack.com/archives/C078NUJP6B1/p1733855563449909

# Add isAuthenticated() method to MedplumClient

## Background
Currently, there's no way to check authentication status without potentially triggering the `onUnauthenticated` callback during MedplumClient initialization. Users want to be able to check auth status client-side without making API requests that might result in 401 errors.

## Current Behavior
- Authentication status can only be determined by making requests and handling 401 errors
- Users must rely on `onUnauthenticated` callback during client initialization
- No direct method to check current auth status client-side

## Requested Feature
Add an `isAuthenticated()` method to MedplumClient that:
- Performs a client-side check of the current access token and expiration
- Does not make any API requests
- Returns a boolean indicating auth status

## Implementation Notes
- Client-side check would be sufficient for most use cases
- Should leverage existing utilities for checking access token status
- Will not account for server-side token revocation, but this is an acceptable limitation

ThatOneBro:
Fixes #5660

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -2799,7 +2799,10 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    * **NOTE: Does not check whether the auth token has been revoked server-side.**
    */
   isAuthenticated(gracePeriod?: number): boolean {
-    if (this.accessTokenExpires && Date.now() > this.accessTokenExpires - (gracePeriod ?? this.refreshGracePeriod)) {
+    if (
+      this.accessTokenExpires !== undefined &&
+      Date.now() > this.accessTokenExpires - (gracePeriod ?? this.refreshGracePeriod)
+    ) {
       return true;
     }
     return false;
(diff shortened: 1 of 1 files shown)


packages/core/src/client.ts
```diff
@@ -2791,6 +2791,23 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
+   * **NOTE: Does not check whether the auth token has been revoked server-side.**
+   */
+  isAuthenticated(gracePeriod?: number): boolean {
+    if (
+      this.accessTokenExpires !== undefined &&
+      Date.now() < this.accessTokenExpires - (gracePeriod ?? this.refreshGracePeriod)
+    ) {
+      return true;
+    }
+    return false;
```
codyebberson, line 2808:
```suggestion
    return this.accessTokenExpires !== undefined &&
      Date.now() < this.accessTokenExpires - (gracePeriod ?? this.refreshGracePeriod);
```

codyebberson (approved):
Minor style thing, otherwise lgtm 👍   ty 🙏

packages/core/src/client.ts
techdavidy: Nit: agree with this

PR #5696, opened 2024-12-18
Truncate string search columns to accommodate index size limits

mattlong:
This is using the most conservative possible length of 675. If anyone balks at 675 being too low, we could be smarter about the limit and see how many bytes the string is and iteratively shorten it, but I'm hesitant to over-engineer this.

### DIFF
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -385,7 +385,7 @@ describe('FHIR Repo', () => {
       expect(patient.meta?.lastUpdated).toStrictEqual(lastUpdated);
     }));
 
-  test.only('Create ResearchDefinition with very long description', () =>
+  test('Create ResearchDefinition with very long description', () =>
     withTestContext(async () => {
       const author = 'Practitioner/' + randomUUID();
 
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -24,6 +24,7 @@ import {
   Project,
   ProjectMembership,
   Questionnaire,
+  ResearchDefinition,
   ResourceType,
   ServiceRequest,
   StructureDefinition,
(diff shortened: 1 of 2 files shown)


codyebberson (approved):

packages/server/src/fhir/repo.test.ts
```diff
@@ -384,6 +385,28 @@ describe('FHIR Repo', () => {
       expect(patient.meta?.lastUpdated).toStrictEqual(lastUpdated);
     }));
 
+  test.only('Create ResearchDefinition with very long description', () =>
```
mattwiller, line 388: Remove `.only()`

packages/server/src/fhir/repo.ts
```diff
@@ -1462,19 +1462,26 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
     }
 
-    if (searchParam.type === 'quantity') {
-      return this.buildQuantityColumn(value);
+    if (!stringValue) {
+      return undefined;
     }
 
-    return typeof value === 'string' ? value : stringify(value);
+    // return stringValue;
```
mattwiller, line 1483: Remove?

packages/server/src/fhir/repo.ts
```diff
@@ -2564,3 +2571,15 @@ function patchObject(obj: any, patch: Operation[]): void {
...
+/**
+ * Apply a maximum string length to ensure the value can accommodate the maximum
+ * size for a btree index entry: 2704 bytes. Be as conservative as possible to avoid
+ * write errors; so limit to 675 characters to accommodate the entire string being
+ * random 4-byte UTF-8 characters.
+ * @param value - The column value to truncate.
+ * @returns The possibly truncated column value.
+ */
+function truncateTextColumn(value: string): string {
+  return value.substring(0, 675);
```
mattwiller, line 2584: Is this not a breaking change for folks who currently have resources with strings that are longer than this, but under the byte limit?

mattlong: Yes, there would be breaking changes at search time for those resources. The most likely would be a `:contains` with a query in the latter part of the string.

codyebberson: Discussed offline.  Recommend using `Blob(str).size` rather than simple string length.  (Although, we still need to figure out how to truncate in a safe way, so we don't cut into a multibyte code point or whatever)

For strings that are too long for the btree index, then it won't be a breaking change, because the initial write would have been a database error in the first place.

For strings that do fit in the btree index, this shouldn't be a change.

ThatOneBro (approved):

codyebberson (approved):

PR #5711, opened 2024-12-23
feat(react,server): add ability to disable MFA

### LINKED ISSUE #5627: Missing Endpoint /auth/mfa/disable
The "Disable MFA" endpoint on our MFA page in `app.` seems to be pointing to a missing endpoint. 

https://github.com/medplum/medplum/blob/eab9c11a443cb000cf37373c855561cd33c1d01e/packages/app/src/MfaPage.tsx#L36

ThatOneBro:
https://github.com/user-attachments/assets/5b25d47e-8f04-45c8-a8dc-dc04abfa2c50

Fixes #5627 

* Adds a `/auth/mfa/disable` endpoint for disabling MFA for the authed user
* Also changes secret on disable of MFA so next enrollment uses a new secret
* Refactors `MfaForm` to take only a `onSubmit` handler so it can be reused for entering MFA before disabling
* Adds tests for all of these things

### DIFF
--- a/packages/react/src/auth/SignInForm.tsx
+++ b/packages/react/src/auth/SignInForm.tsx
@@ -129,7 +129,17 @@ export function SignInForm(props: SignInFormProps): JSX.Element {
             </AuthenticationForm>
           );
         } else if (mfaRequired) {
-          return <MfaForm login={login} handleAuthResponse={handleAuthResponse} />;
+          return (
+            <MfaForm
+              onSubmit={async (fields) => {
+                const res = await medplum.post('auth/mfa/verify', {
+                  login: login,
+                  token: fields.token,
+                });
+                handleAuthResponse(res);
+              }}
+            />
+          );
         } else if (memberships) {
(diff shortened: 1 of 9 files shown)


packages/react/src/auth/SignInForm.tsx
```diff
@@ -129,7 +129,18 @@ export function SignInForm(props: SignInFormProps): JSX.Element {
             </AuthenticationForm>
           );
         } else if (mfaRequired) {
-          return <MfaForm login={login} handleAuthResponse={handleAuthResponse} />;
+          return (
+            <MfaForm
```
ThatOneBro, line 133: I did change the API for `MfaForm`, but I do think this one is more generic and could be more useful to consumers other than `SignInForm`. Perhaps we can wait until v4 if we think this is an unacceptable breaking change since this component is actually exported?

codyebberson (approved):
Cool, ty 🙏

PR #5712, opened 2024-12-23
Dependency upgrades 2024-12-23

medplumbot:
Dependency upgrades


scripts/upgrade.sh
```diff
@@ -35,7 +35,8 @@ EXCLUDE="@types/express @types/react @types/react-dom eslint node-fetch react re
 # --reject - Exclude packages matching the given string
 # --target - Determines the version to upgrade to
 # "minor" - Upgrade to the highest minor version without bumping the major version
-npx npm-check-updates --workspaces --root --upgrade --reject "$EXCLUDE" --target minor
+# `enginesNode` makes sure that packages can be run against the node requirement specified in the monorepo "engines.node"
+npx npm-check-updates --workspaces --root --upgrade --reject "$EXCLUDE" --target minor --enginesNode
```
ThatOneBro, line 39: Added `--enginesNode` option to `npm-check-updates`, which makes sure that package versions match the requirement in `engines.node` in `package.json`

This was required so that we didn't increment to `npm` version 11, which does not work on lower versions of Node20 or Node18 at all (which we still support until v4)

We may want to wait a few weeks to let the dust settle on npm 11, but either way we should probably always have this `--enginesNode` check to prevent upgrades that would not work for our desired Node version

package.json
```diff
@@ -66,10 +66,11 @@
   },
   "overrides": {
     "es5-ext": "0.10.53",
-    "esbuild": "0.24.0",
+    "esbuild": "0.24.2",
     "formidable": "3.5.2",
     "got": "11.8.6",
     "react": "18.2.0",
+    "react-compiler-runtime": "latest",
```
ThatOneBro, line 73: Explicit override to latest for `babel-preset-expo` and `graphiql` conflict on this dependency. `react-compiler-runtime` is currently using beta versions with commit hashes and both packages declare explicit versions of those beta versions which conflict. We should just use whatever is latest until it's out of beta.

codyebberson (approved):

codyebberson:
Deploying to staging...

codyebberson:
Staging looks good 👍

PR #5717, opened 2024-12-24
fix(client): don't try to parse when `content-length: 0`

### LINKED ISSUE #5713: MedplumClient 'Error parsing response 201 SyntaxError: Unexpected end of JSON input' occurs when body is empty
Description:
When making a create request, I received a 201 response with an empty body (not empty json, fully empty), the client then raised the following

```
Error parsing response 201 SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:5472:19)
    at successSteps (node:internal/deps/undici/undici:5454:27)
    at fullyReadBody (node:internal/deps/undici/undici:4381:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async consumeBody (node:internal/deps/undici/undici:5463:7)
    at It.parseBody (/Users/elaye/project/node_modules/@medplum/core/src/client.ts:3229:15)
    at It.request (/Users/elaye/project/node_modules/@medplum/core/src/client.ts:3198:17)
```

After some digging, what went wrong was as follows:
1.  The response body was EMPTY (not empty json).
2. Client correctly parsed the content-type header of the response and identified it as a json (i.e., the vendor I was polling returned an empty body with a json content-type).
3. The client did not check the response content-length header (which is 0).
4. The client then attempts to parse the body, see it should be a json, and attempts request.json() (which fails on empty body).

Suggestions:
Either parse content-length to check if body is empty before calling the parser in line 3198 of the client, or just directly check if the raw body is empty before attempting to parse it as a json.

ThatOneBro:
Fixes #5713

Testing this was a little challenging since our test utils for mocking fetch prevent this error from happening by always returning the body, even when the body is not parseable JSON. We should probably polyfill the fetch API in `core`, `react`, and `app` instead of trying to mock them to prevent these kinds of hidden bugs. 

I did verify and attempting to call `.json` on a `Response` when the body is `null` does cause it to throw the described error:
Screenshot 2024-12-24 at 12 20 00 AM

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1116,12 +1116,14 @@ describe('Client', () => {
         switch (type) {
           case 'json': {
             // Simulate parsing non-JSON
+            // This will always throw
             JSON.parse('EMPTY');
             break;
           }
           default:
             return '';
         }
+        throw new Error('UNREACHABLE');
       };
       return Promise.resolve({
         ok: true,
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1104,6 +1104,44 @@ describe('Client', () => {
     expect(result).toStrictEqual({});
   });
 
+  test('HTTP CREATE with Content-Length 0', async () => {
+    const consoleErrorSpy = jest.spyOn(console, 'error');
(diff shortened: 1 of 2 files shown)


codyebberson (approved):
Nice, good find 👍

PR #5718, opened 2024-12-25
Replace ua-parser-js dependency with bowser

### LINKED ISSUE #5552: Remove ua-parser-js dependency
The `ua-parser-js` dependency changed license from MIT to AGPL, which is incompatible with our Apache 2 license.

We have temporarily handled this by pinning to an older MIT licensed version, but that is not a long term strategy.

While we are always open to paying for dependencies, I don't think that is appropriate in this particular case.  As a server dependency, it would create a bunch of complexity to introduce a commercial dependency.  And our usage of this dependency is pretty minor.

The only place we use the `ua-parser-js` library is as a convenience feature for the security page to show each access token's browser and operating system.  See here for more details: https://github.com/medplum/medplum/pull/1166

Broad options:
* Continue to using outdated/unsupported MIT licensed version
* Fork or vendor the MIT licensed version
   * It's only one ~1000 line file: https://github.com/faisalman/ua-parser-js/blob/1.0.x/src/ua-parser.js
   * Could probably be slimmed down substantially by only supporting modern browsers and operating systems
   * Would still be a fair amount of work
* Change to a difference license
   * Quick search did not yield any obvious alternatives, could use more research
* Try a quick minimalist strategy that just extracts operating system, does not try to extract browser
   * This aligns with other modern platforms such as Gmail and Netflix
   * Just show "Windows", "mac", "iPhone", "Android", etc, don't get bogged down with exact browser details

There is not a lot of competition in the user-agent parsing space (either very outdated or relatively low usage):

image

adam-mak:
Fixes #5552 

This pull request focuses on replacing the `ua-parser-js` library with `bowser` due to licensing incompatibilities, and updating the related code and dependencies accordingly.

High-level comparison between the two libraries: Similar in popularity (as shown in the figure presented in #5552), `bowser` has a smaller unpacked size of 217 kB vs. `ua-parser-js`' 1.13 MB. This is due to `ua-parser-js` comprehensive features not required in determining browser and OS, hence suggesting `bowser` to be the better choice regardless of licensing issues.

Ran a quick test with my local environment:

With `ua-parser-js`:
```
{"level":"INFO","timestamp":"2024-12-24T23:59:59.498Z","msg":"Login success","email":"admin@example.com","requestId":"9ed6cebc-739c-472f-862d-d361e048f0f4","traceId":"9a11bb4a-642a-47e3-9f81-4cc21b16963d"}
Browser: Edge
OS: Windows
```

With `bowser`:
```
{"level":"INFO","timestamp":"2024-12-24T23:41:23.592Z","msg":"Login success","email":"admin@example.com","requestId":"42fba0f5-3a9e-4f8a-8051-af9cab64a2c3","traceId":"b54a8bde-bf36-4495-adab-fe34b4e2ffe1"}
Browser: Microsoft Edge
OS: Windows
```

Note, this is my first PR to Medplum :)


packages/server/package.json
```diff
@@ -61,6 +61,7 @@
     "@smithy/util-stream": "3.3.2",
     "bcryptjs": "2.4.3",
     "body-parser": "1.20.3",
+    "bowser": "^2.11.0",
```
ThatOneBro, line 64: Can we make this an exact version?
```suggestion
    "bowser": "2.11.0",
```

ThatOneBro:
Hi @adam-mak, this is great! Thank you for making your first contribution!

I haven't heard of `bowser` before but looks like exactly what we want.

It looks like the build might have failed on CI due to `package-lock.json` being out of sync. Do you mind running `npm install` in the root of the repo and recommiting `package-lock.json`?

Overall LGTM but I'll wait for comments from some other maintainers first

packages/server/src/auth/me.ts
```diff
@@ -1,12 +1,13 @@
 import { getReferenceString, Operator, ProfileResource } from '@medplum/core';
 import { Login, Project, ProjectMembership, Reference, User, UserConfiguration } from '@medplum/fhirtypes';
 import { Request, Response } from 'express';
-import { UAParser } from 'ua-parser-js';
 import { getAuthenticatedContext } from '../context';
 import { getAccessPolicyForLogin } from '../fhir/accesspolicy';
 import { getSystemRepo, Repository } from '../fhir/repo';
 import { rewriteAttachments, RewriteMode } from '../fhir/rewrite';
```
ThatOneBro, line 8:
```suggestion
```
Do you mind deleting this space too? Small nit

ThatOneBro (commented):
Awesome, looks like this should pass now probably. One final nit and then I think this is pretty much good

codyebberson:
Thanks @adam-mak !  I rebased to fix merge conflicts with recent changes to `package-lock.json`.  Everything looks good to me.  Thank you! 🙏

ThatOneBro (approved):

PR #5719, opened 2024-12-25
Adding findResourceInBundle to the official SDK

### LINKED ISSUE #4397: Util suggestion: findResourceById
Hi,

while building with the Medplum SDK we've developed a few helper methods to help us work with the API. It would be fantastic if some version of these could be added to the official SDK. Making separate issues for more focused discussion :)

We often find ourselves searching a Bundle for a specific resource by ID.

```typescript
function findResourceById<T extends Resource = Resource>(
    bundle: Bundle<T>,
    id: string
): T | undefined {
    return bundle.entry?.find(
        ({ resource }) => `${resource?.resourceType}/${resource?.id}` === id
    )?.resource as T | undefined
}
```

adam-mak:
Fixes #4397 

Looking for an initial review.

The initial request was for a string such as "Patient/123" where the resourceType and id delimited by a forward slash is used as input. I wasn't able to find any FHIR resource documentation stating that this is a conventional input, whereas the [location URL](https://build.fhir.org/resource.html) is stated to be used to identify a unique resource.

I've added a second method that takes two arguments: resourceType and id (representing the Logical ID). I'm also suggesting that requesting a location URL as an argument could be possible here too.

### DIFF
--- a/packages/core/src/utils.test.ts
+++ b/packages/core/src/utils.test.ts
@@ -29,6 +29,7 @@ import {
   findObservationReferenceRange,
   findResourceByCode,
   findResourceById,
+  findResourceByLogicalId,
   flatMapFilter,
   getAllQuestionnaireAnswers,
   getCodeBySystem,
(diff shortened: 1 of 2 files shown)


mattlong:
Hi, @adam-mak

Thanks so much for the PR! To get this function more inline with our other function signatures could you please update the implementation to just be single function instead with the following signature?

```ts
function findResourceInBundle<K extends ResourceType>(
  bundle: Bundle,
  resourceType: K,
  id: string
): ExtractResource<K> | undefined
```

Thanks so much!

mattlong (changes requested):
Hi, @adam-mak

Thanks so much for the PR! To get this function more inline with our other function signatures could you please update the implementation to just be single function instead with the following signature?

```ts
function findResourceInBundle<K extends ResourceType>(
  bundle: Bundle,
  resourceType: K,
  id: string
): ExtractResource<K> | undefined
```

Thanks so much!

PR #5730, opened 2025-01-03
fix: add `storageDomainName` to `content-src` in CDK security headers

### LINKED ISSUE #5729: "Refused to connect to 'https://storage.medplum.com/binary/{...}' because it violates the document's Content Security Policy."
With this error we also get a more specific error:
`client.ts:3263 Refused to connect to 'https://storage.medplum.com/binary/{...}' because it violates the following Content Security Policy directive: "connect-src 'self' api.medplum.com *.google.com".`

This seems to be because we have never really needed `storage.medplum.com` to be directly fetchable within appdot, since we typically use some kind of i-frame for most content types. Now after adding the C-CDA viewer in https://github.com/medplum/medplum/pull/5657, I think we need to add `config.storageDomainName` to the `content-src` string in our CDK config.

ThatOneBro:
Fixes #5729

### DIFF
--- a/packages/cdk/src/frontend.ts
+++ b/packages/cdk/src/frontend.ts
@@ -72,7 +72,7 @@ export class FrontEnd extends Construct {
               `default-src 'none'`,
               `base-uri 'self'`,
               `child-src 'self'`,
-              `connect-src 'self' ${config.apiDomainName} *.google.com`,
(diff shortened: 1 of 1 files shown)


mattlong (approved):

ThatOneBro:
Closed in favor of #5753

Issue #5742, opened 2025-01-06
[DRAFT] - Subscriptions not firing when the `if-match` header is set


ianplunkett:
A customer noticed weird behavior with Subscriptions not firing when the `if-match` header was set while performing `PUT` operations on a resource.

reshmakh:
This is completed

PR #5744, opened 2025-01-07
Do not modify resources used by post-commit hook

mattwiller:
In order to prevent issues where an access policy could remove fields that might affect processing of subscriptions or other post-commit background jobs, the fields are instead stripped from a copy of the resource


mattwiller:
Pending full confirmation of the root cause of the issue, this may be related to fixing #5742

codyebberson (approved):

PR #5745, opened 2025-01-07
Use uuid v7 for resourceIds

mattlong:
We've been discussing this change for quite a while. The benefit being less churn/computation when Postgres is updating the btree indexes covering resource IDs...in other words, faster writes.

### DIFF
--- a/packages/server/src/fhir/operations/projectclone.ts
+++ b/packages/server/src/fhir/operations/projectclone.ts
@@ -1,7 +1,6 @@
 import { created, forbidden, getResourceTypes, isResourceType, Operator } from '@medplum/core';
 import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
 import { Binary, Project, Resource, ResourceType } from '@medplum/fhirtypes';
-import { randomUUID } from 'node:crypto';
 import { getAuthenticatedContext } from '../../context';
 import { Repository } from '../repo';
 import { getBinaryStorage } from '../storage';
@@ -65,7 +64,7 @@ class ProjectCloner {
         if (!entry.resource || !this.isAllowedResourceId(entry.resource.id as string)) {
           continue;
         }
-        this.idMap.set(entry.resource.id as string, randomUUID());
+        this.idMap.set(entry.resource.id as string, repo.generateId());
         buildBinaryIds(entry.resource, binaryIds);
         if (entry.resource.resourceType !== 'Project') {
           allResources.push(entry.resource);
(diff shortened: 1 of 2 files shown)


codyebberson (commented):
As discussed offline yesterday...
* This would theoretically be a big win for net-new deployments, that don't already have a bunch of v4 UUIDs
* The benefits are less clear for deployments that already have v4 UUIDs, because the sort order could still cause btree thrashing
* There's also uncertainty regarding future plans for client assigned IDs, and potentially using v5 UUIDs

Do we have any benchmarks or predictions on performance impact?

Part of me is tempted to go ahead and approve, because "it won't be any worse".  But I guess even that is just an assumption?

mattlong:
Claude's take on it actually seems pretty reasonable/accurate and seems to match our hypotheses.

My question:
```
Let's assume I have a Postgres 16 database that has been using v4 UUIDs as the primary keys of most tables for quite a while. I know that using v7 UUIDs would theoretically make inserts faster since they are roughly monotonically increasing versus being fully random like v4 UUIDs. Should I expect a similar speedup if I were to switch new records being added to this database from using v4 to v7 UUIDs?
```

Response:
```
This is an interesting database optimization question that requires thinking through several factors:

1. The primary bottleneck with random UUIDs (v4) in B-tree indexes is that new inserts can happen anywhere in the tree, causing page splits and fragmentation. This leads to:
   - More random I/O operations
   - Index fragmentation over time
   - Higher maintenance overhead

2. However, in your case with millions of existing v4 UUIDs, switching to v7 UUIDs for new records won't fully solve these issues because:
   - The existing random v4 UUIDs have already created a fragmented index structure
   - New v7 UUIDs will still need to be interspersed between existing random v4 values
   - The monotonic properties of v7 only help when they're used exclusively

3. You might see some localized benefits where recent inserts cluster together due to v7's timestamp component, but this would be limited since these new records still need to fit between existing random v4 values.

4. To get the full performance benefit of monotonic UUIDs, you would need to:
   - Create new tables with v7 UUIDs
   - Migrate existing data with new v7 IDs
   - Rebuild indexes

Given this analysis, I would expect only minimal performance improvements from switching to v7 UUIDs for new records in your existing tables. The benefit would be much smaller than what you'd see in a fresh table using only v7 UUIDs.

If insert performance is a critical concern, you might want to consider:
1. Regularly reindexing to combat fragmentation
2. Using table partitioning if your access patterns support it
3. Planning a gradual migration to new tables if the performance gains would justify the effort

Would you like me to elaborate on any of these alternative approaches?
```

mattlong:
Two runs of serially inserting 5,000 Observations on a production-sized database with existing records having v4 uuids. Times in milliseconds.

#### v4 UUIDs
Min: 25, 25
Max: 142, 80
Mean: 31.7, 32.0

#### v7 UUIDs
Min: 24, 22
Max: 755, 103
Mean: 30.5, 29.0

Not terribly scientific, but a slight advantage to v7 that is probably within the margin of error.

codyebberson (approved):

codyebberson:
Thanks for running the experiment 🙏

PR #5746, opened 2025-01-08
Preserve transaction when using system repo

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -1791,7 +1791,9 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         }
       }
     } else {
-      const patients = await getSystemRepo().readReferences(getPatients(updated));
+      const systemRepo = getSystemRepo();
+      systemRepo.conn = this.conn; // Re-use DB connection to preserve transaction state
```
codyebberson, line 1795: Oof, this seems risky.  This is a pretty big departure from the model where 1 repo === 1 db connection.

In this particular case, it might be fine, but I feel like we're breaking a seal on something that has been quite sacred.

I.e., what happens here when `[Symbol.dispose]` gets called for the `systemRepo`?  Wouldn't that call `systemRepo.close()`, and close the connection?

I understand the need to re-use the transaction scope, so we have access to the `Patient` resource created in the bundle...  I think we need to add some more machinery here.

For example:
* Have an alternative `readReferences` method that bypasses access policies for special cases like this
* Or, perhaps a more general, "temporarily disable access policies" mode for the user's Repository (but that might have other risks elsewhere)
* Have an explicit flag on `Repository` that "this is a borrowed" connection, to avoid incorrect closing logic

ThatOneBro (approved):

packages/server/src/fhir/repo.ts
```diff
@@ -2446,23 +2453,21 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+        this.releaseConnection(new Error('Closing Repository with active transaction'));
+      } else {
+        // Good state, return healthy connection to pool
+        this.releaseConnection();
+      }
     }
     this.closed = true;
   }
 
-  [Symbol.dispose](): void {
```
ThatOneBro, line 2462: Context (from offline conversation): `[Symbol.dispose]` is not automatically called unless using the `using` keyword, so although we are not currently taking advantage of that, we can standardize on `[Symbol.dispose]` as the "closer" for our various resources with explicit close functions.

PR #5755, opened 2025-01-09
Terraform for Azure

diegosucaria:


terraform/azure/terraform.tfvars
```diff
@@ -0,0 +1,9 @@
+
+# Azure 
+resource_group_name = "medplum"
+location            = "eastus2"
+tags = {
+  app = "medplum"
+}
+app_domain                = "app.azure.medplum.dev"
+app_certificate_secret_id = "https://medplum-certs-12hh.vault.azure.net/certificates/medplum-appb0836994-69f1-445f-b0a9-ebadf4fb7ad6/b542c39e93694747b8f26d50ff284f95"
```
codyebberson, line 9: Should this secret be in the public terraform scripts?

terraform/azure/.terraform.lock.hcl
```diff
@@ -0,0 +1,120 @@
+# This file is maintained automatically by "terraform init".
```
codyebberson, line 1: I assume this file should probably be stored in a different repository, as it is specific to our own azure deployment?

codyebberson (commented):
Thanks @diegosucaria !  I left a couple questions for you

terraform/azure/terraform.tfvars
diegosucaria: Hi Cody! good catch. Yes, the certificate secret ID is not the secret itself, but I've added a placeholder instead

terraform/azure/.terraform.lock.hcl
diegosucaria: The terraform lock is just the dependencies lock, and it should be checked in the repo, but it is currently for Linux, as that is the platform I've used to deploy.

For simplicity, I just removed it.
 
https://www.terraform.io/docs/language/dependency-lock.html

codyebberson (approved):

PR #5757, opened 2025-01-09
Preserve transaction when using system repo

mattwiller:
Retry of #5746 after it was reverted in #5754 

The line that was previously causing issues in the logging middleware has been removed


codyebberson (approved):
lgtm.  Let's test on staging before merge

ty! 🙏

packages/server/src/fhir/repo.ts
```diff
@@ -218,21 +218,27 @@ export interface ResendSubscriptionsOptions extends InteractionOptions {
...
+  private readonly disposable: boolean = true;
   private transactionDepth = 0;
   private closed = false;
   mode: RepositoryMode;
 
   private preCommitCallbacks: (() => Promise<void>)[] = [];
   private postCommitCallbacks: (() => Promise<void>)[] = [];
 
-  constructor(context: RepositoryContext) {
+  constructor(context: RepositoryContext, conn?: PoolClient) {
```
mattlong, line 229: Does anything need to change with the `clone` method below to account for a passed in `conn` to future proof it? I think I'd expect `conn` to be reused as well in the clone, but that's just my first take on it. Perhaps `clone` should throw if `this.disposable` is false?

mattwiller:
Tested on Staging, working as expected

PR #5756, opened 2025-01-09
Add `config.logRequest` test that uses database

mattlong:

PR #5764, opened 2025-01-10
test: add basic smoke tests / e2e tests using `playwright`

### LINKED ISSUE #3955: Add smoke tests for basic flows
We can start simple for now and slowly add more tests over time.

Tests we should add:
- [ ] Email login flow
- [ ] Google login flow
- [ ] Creating resource
- [ ] Reading created resource
- [ ] Search
- [ ] Uploading a binary

ThatOneBro:
Fixes #3955

The primary intent of this PR is to setup smoke tests to gate a merge on running some sanity check e2e tests

Follow-up: 

Run the same tests against prod right after a deployment in prod. The tricky part will be making sure that the tests run against the new nodes and not the nodes rotating out, but the point would be to run it against prod to make sure everything is still in working order, and create an incident if any of the smoke tests fail


packages/e2e/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "bugs": {
+    "url": "https://github.com/medplum/medplum/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/medplum/medplum.git",
+    "directory": "packages/e2e"
+  },
+  "scripts": {
+    "test:smoke": "playwright test smoke/medplum.smoke.test.ts --reporter=list,html"
```
ThatOneBro, line 21: Probably want to leave this as separate from `test` since Turbo will try to autorun this unless we apply a specific filter which seems like unneeded complexity.

packages/server/medplum.config.json
```diff
@@ -38,5 +38,6 @@
     "removeOnComplete": { "count": 1 }
   },
   "shutdownTimeoutMilliseconds": 30000,
-  "chainedSearchWithReferenceTables": true
+  "chainedSearchWithReferenceTables": true,
+  "logRequests": true
```
ThatOneBro, line 42: Adding `"logRequests": true` to the default config. This does make logs slightly noisier but we already have this enabled in prod, and having this off made us miss a bug in the logging middleware before.

Having this on will make sure we catch any bugs that would happen in our new e2e tests

packages/e2e/files/frodo_baggins.png
ThatOneBro, line 1: Generated by Midjourney and pixelated and optimized down to 4KB, needed for testing upload capability

packages/server/medplum.config.json
codyebberson: @mattlong  added a test specifically for that case, so i think we can leave this off: https://github.com/medplum/medplum/pull/5756

Or i suppose we could use Jest's silent mode: https://jestjs.io/docs/cli#--silent

ThatOneBro: Alright we can remove it. I guess that test should at least help us to detect most kinds of catastrophic failures

packages/e2e/smoke/medplum.smoke.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    await page.getByRole('button', { name: 'Add', exact: true }).click();
+    await page.getByRole('button', { name: 'Create' }).click();
+    await expect(page).toHaveURL(/^http:\/\/localhost:3000\/Patient\/[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*$/);
+    await expect(page.getByTestId('timeline-item').getByText('Frodo Baggins')).toBeVisible();
+  });
+
+  test('Search for patient via searchbar', async ({ page }) => {
+    await signIn(page, 'admin@example.com', 'medplum_admin');
+
+    await page.getByPlaceholder('Search').fill('Frodo Baggins');
```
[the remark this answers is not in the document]
ThatOneBro: I am 50/50 in agreement with you, I also think that our UI is a little hard to get good selectors for at the moment, at least using the visual codegen tool that Playwright recommends using. That's probably on us but I think to make it less brittle we might need to do some refactoring first

PR #5766, opened 2025-01-10
metrics(bullmq): add subscription queue length metric

### LINKED ISSUE #4860: Add metrics for subscription BullMQ queue length

ThatOneBro:
Fixes #4860

### DIFF
--- a/packages/server/src/healthcheck.ts
+++ b/packages/server/src/healthcheck.ts
@@ -76,7 +76,7 @@ export async function healthcheckHandler(_req: Request, res: Response): Promise<
 
   const subscriptionQueue = getSubscriptionQueue();
   if (subscriptionQueue) {
-    setGauge('medplum.bullmq.subscriptionQueueLength', await subscriptionQueue.getWaitingCount());
+    setGauge('medplum.subscription.waitingCount', await subscriptionQueue.getWaitingCount());
   }
 
   res.json({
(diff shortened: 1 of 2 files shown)


packages/server/src/healthcheck.ts
```diff
@@ -73,6 +74,11 @@ export async function healthcheckHandler(_req: Request, res: Response): Promise<
     BASE_METRIC_OPTIONS
   );
 
+  const subscriptionQueue = getSubscriptionQueue();
+  if (subscriptionQueue) {
+    setGauge('medplum.bullmq.subscriptionQueueLength', await subscriptionQueue.getWaitingCount());
```
mattwiller, line 79: I'll bikeshed a more general/descriptive name for the metric, e.g`medplum.subscriptions.enqueuedCount` or `medplum.queue.subscription.enqueuedCount`

mattwiller (approved):

packages/server/src/healthcheck.ts
ThatOneBro: Went with `medplum.subscription.waitingCount` since we have `medplum.subscription` namespace already and `waiting` is the name of the status, which only includes jobs currently waiting to be picked up and doesn't include `delayed` jobs.

Actually now that I'm thinking about it should we have both of those metrics?

mattwiller: @ThatOneBro 👍 to both!

Issue #5759, opened 2025-01-10
Bot timeline improvements


codyebberson:
1. Don't show code snippets by default, add a button or link for "Expand..." for any Binary/Attachment diffs
2. Only show the first 3 diffs by default, add a button or link for "Show..." to all diffs after the first 3

PR #5791, opened 2025-01-14
Bot timeline improvements

techdavidy:
resolves https://github.com/medplum/medplum/issues/5759


packages/react/src/ResourceDiffTable/ResourceDiffTable.tsx
```diff
@@ -24,6 +24,7 @@ export function ResourceDiffTable(props: ResourceDiffTableProps): JSX.Element |
   const medplum = useMedplum();
   const { original, revised } = props;
   const [schemaLoaded, setSchemaLoaded] = useState(false);
+  const shouldToggleDisplay = (original as Bot).sourceCode?.contentType === 'text/typescript';
```
techdavidy, line 27: Is this the correct way to find out better to present the button ?

package-lock.json
codyebberson, line 1: Can you revert this change?  `package-lock.json` changes should only be added when making changes to `package.json` files.

packages/react/src/ResourceDiffTable/ResourceDiffTable.tsx
codyebberson: Hmm... good question.  I actually wonder if we should use this logic for all attachments, not just Typescript files?

packages/react/src/ResourceDiffRow/ResourceDiffRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { name, path, property, originalValue, revisedValue, shouldToggleDisplay } = props;
+  const [isCollapsed, setIsCollapsed] = useState(true);
+  const toggleCollapse = (): void => setIsCollapsed((prev) => !prev);
+
+  return (
+    <>
+      <Table.Tr>
+        <Table.Td>
+          <Stack>
+            <span>{name}</span>
```
codyebberson, line 26: I think we should be able to remove the `<Stack>` and `<span>` wrappers... maybe?

techdavidy: good point

packages/react/src/ResourceDiffTable/ResourceDiffTable.tsx
techdavidy: maybe let's leave it for now and expand to other cases as we see necessary ?

codyebberson:
Nice, this is good progress, thanks @techdavidy 

I was playing with the preview, and I have a couple requests:
1. Let's only use the expand/collapse feature on "Attachment" properties.  Other properties (strings, numbers, etc) can use the normal behavior.
2. For the "Expand" button, can we move it inline into the empty table cells?  Maybe `colspan=2` so it spans both "Before" and "After"?

image

techdavidy:
Due to cell content (check before and after getting moved in line), button would be moved.
Right aligned button and keep it inline when no content is displayed.
https://github.com/user-attachments/assets/efbea40b-08f2-4e12-8684-4aedfd8efbe1

packages/react/src/ResourceDiffRow/ResourceDiffRow.tsx
```diff
@@ -0,0 +1,73 @@
...
+  path: string;
+  property: InternalSchemaElement | undefined;
+  originalValue: TypedValue | undefined;
+  revisedValue: TypedValue | undefined;
+  shouldToggleDisplay: boolean;
+}
+
+export function ResourceDiffRow(props: ResourceDiffRowProps): JSX.Element {
+  const { name, path, property, originalValue, revisedValue, shouldToggleDisplay } = props;
+  const [isCollapsed, setIsCollapsed] = useState(true);
```
codyebberson, line 18: The default value here should be `true` if the `property` is an `Attachment` type, and `false` otherwise.

We can check if the `property` is an `Attachment` like this:

```ts
property?.type?.find((t) => t.code === 'Attachment')
```

So:
```suggestion
  const isAttachmentType = property?.type?.find((t) => t.code === 'Attachment');
  const [isCollapsed, setIsCollapsed] = useState(isAttachmentType);
```

packages/react/src/ResourceDiffRow/ResourceDiffRow.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const toggleCollapse = (): void => setIsCollapsed((prev) => !prev);
+
+  return (
+    <>
+      {(shouldToggleDisplay && !isCollapsed) || !shouldToggleDisplay ? (
```
codyebberson, line 23: Then we can simplify this logic:

```suggestion
      {!isCollapsed ? (
```

codyebberson (commented):
After playing with the preview, I have some recommendations on how we can simplify things.  See inline comments

codyebberson (approved):

PR #5795, opened 2025-01-15
Always verify PKCE on token grant if present

mattwiller:


codyebberson (approved):
Looks good to me after tests passing 👍

PR #5801, opened 2025-01-16
Inferno testing fixes for Group Export

mattwiller:
A couple fixes related to passing the Inferno tests related to Group Export:
- Implementing some export cancellation logic
- Returning 404 Not Found `OperationOutcome`  for cancelled exports


packages/server/src/fhir/bulkdata.ts
```diff
@@ -46,7 +53,16 @@ bulkDataRouter.get(
   })
 );
 
-bulkDataRouter.delete('/export/:id', (req: Request, res: Response) => {
+bulkDataRouter.delete('/export/:id', async (req: Request, res: Response) => {
+  const { id } = req.params;
+  const bulkDataExport = await getExportResource(id);
```
ThatOneBro, line 58: At some point in the near future, this should probably call the `AsyncJob/$cancel` handler like the analogous DELETE route for the `AsyncJob` route

packages/server/src/fhir/operations/groupexport.ts
```diff
@@ -30,7 +30,7 @@ export async function groupExportHandler(req: FhirRequest): Promise<FhirResponse
 
   // Start the exporter
   const exporter = new BulkExporter(ctx.repo, since, types);
-  const bulkDataExport = await exporter.start(concatUrls(baseUrl, 'fhir/R4' + req.pathname));
+  const bulkDataExport = await exporter.start(concatUrls(baseUrl, 'fhir/R4/' + req.pathname));
```
ThatOneBro, line 33:
```suggestion
  const bulkDataExport = await exporter.start(concatUrls(baseUrl, 'fhir/R4', req.pathname));
```
Does this work?

ThatOneBro (approved):

packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx
```diff
@@ -25,7 +25,7 @@ export function SmartAppLaunchLink(props: SmartAppLaunchLinkProps): JSX.Element
       })
       .then((result) => {
         const url = new URL(client.launchUri as string);
-        url.searchParams.set('iss', medplum.getBaseUrl() + 'fhir/R4');
+        url.searchParams.set('iss', medplum.getBaseUrl() + 'fhir/R4/');
```
codyebberson, line 28:
```ts
import { ensureTrailingSlash } from '@medplum/core';
```

```suggestion
        url.searchParams.set('iss', ensureTrailingSlash(medplum.fhirUrl()));
```

packages/server/src/fhir/bulkdata.ts
```diff
@@ -46,7 +53,16 @@ bulkDataRouter.get(
...
+
+  if (bulkDataExport.status !== 'cancelled') {
+    bulkDataExport.status = 'cancelled';
```
codyebberson, line 61: Should we do this if the job is already completed?

Anything is better than the current non-implementation, so I'm pretty open minded here.

packages/server/src/fhir/operations/groupexport.ts
codyebberson, line 33: 👍

packages/server/src/fhir/operations/patienteverything.ts
```diff
@@ -80,5 +80,9 @@ export async function getPatientEverything(
     filters,
     count: params?._count ?? defaultMaxResults,
     offset: params?._offset,
+    include: [{ resourceType: 'Patient', searchParam: 'organization' }],
+    revInclude: [
+      { resourceType: 'Location', searchParam: 'organization', targetType: 'Organization', modifier: Operator.ITERATE },
```
codyebberson, line 85: If not too annoying, can we add a test for `Organization` and/or `Location`?

I'm not concerned about this not working.  I'm thinking of it more as a matter of capturing the requirement when future engineers inevitably touch this code.

codyebberson (approved):
Some minor feedback, at your discretion.

packages/server/src/fhir/operations/groupexport.ts
codyebberson: @ThatOneBro  - yeah, i had the same thought...  Our current `concatUrls()` method does not support this.  we could though.  might be out of scope for this PR.

packages/server/src/fhir/bulkdata.ts
ThatOneBro: In the `asyncJobCancelHandler` we do a test for `accepted` status in the patch operation before setting status to `cancelled` FWIW

mattwiller: This was a requirement of the test case: Inferno expects the cancelled export to return a 404 status, even if it was completed before the cancellation occurred

PR #5802, opened 2025-01-16
Return only R4 resources in CapabilityStatement

mattwiller:


packages/server/src/fhir/metadata.ts
```diff
@@ -238,32 +238,38 @@ function buildResourceTypes(): CapabilityStatementRestResource[] {
   return Object.entries(getAllDataTypes())
     .filter(
       ([resourceType, typeSchema]) =>
-        isResourceType(resourceType) && typeSchema.url?.startsWith('http://hl7.org/fhir/StructureDefinition/')
+        isResourceType(resourceType) &&
+        typeSchema.url?.startsWith('http://hl7.org/fhir/StructureDefinition/') &&
+        typeSchema.version === '4.0.1'
```
codyebberson, line 243: 👍

packages/core/src/typeschema/types.ts
```diff
@@ -260,6 +261,7 @@ class StructureDefinitionParser {
       title: sd.title,
       type: sd.type,
       url: sd.url as string,
+      version: sd.version,
```
codyebberson, line 264: Do we need to re-run the base-schema generator after this change?

It actually seems like we would *not* want `version` to be in the base-schema, because (1) I don't think it includes any `Resource` types that would appear in the capability statement, and (2) it would be a lot of unnecessary data.

codyebberson (approved):

packages/core/src/typeschema/types.ts
mattwiller: That seems correct to me — I don't think we need to add `version` to base schema

PR #5805, opened 2025-01-16
`getLogger` no longer causes circular imports

mattlong:
Since `getLogger` was in `context.ts` which imports `fhir/repo.ts` (directly and via `fhir/accesspolicy.js`),
server code importing `getLogger` would tend to cause circular dependencies, e.g.:

```
context.js > fhir/accesspolicy.js > fhir/repo.js > workers/index.js > workers/subscription.js
```

By separating parts of `context.ts`, we avoid said circular imports.

This was originally detected after edits to code being used by the `src/migrations/migrate.ts` script; the circular imports don't seem to affect the dev server; ostensibly since `NodeNext` is a more robust module resolution system.


ThatOneBro (approved):
Yup, LGTM! As discussed offline

mattwiller (approved):

mattlong:
Thank you for the idea of using `IRequestContext`, @ThatOneBro! 🙏

PR #5807, opened 2025-01-16
fix(fhircast): make sure only one heartbeat listener for all of FHIRcast

ThatOneBro:
This PR fixes a bug I noticed when subscribing to the same topic with multiple subscribers where you would get multiple heartbeats per tick. This was because we were creating one listener per subscriber on connect to the FHIRcast WebSocket endpoint.

This PR aligns our FHIRcast heartbeat implementation with the one we use in FHIR WebSocket subscriptions where there is only one driver of heartbeats for all WebSockets / subscriptions.

I also took the opportunity to add some metrics about FHIRcast WebSocket count and number of active topics for FHIRcast.

### DIFF
--- a/packages/server/src/fhircast/websocket.test.ts
+++ b/packages/server/src/fhircast/websocket.test.ts
@@ -235,7 +235,7 @@ describe('FHIRcast WebSocket', () => {
                 const endTime = Date.now();
 
                 // setInterval doesn't guarantee a minimum time between executions, so we give a little leniency for the 100ms
-                expect(endTime - startTime).toBeGreaterThanOrEqual(95);
+                expect(endTime - startTime).toBeGreaterThanOrEqual(90);
               })
               // We're just expecting the two calls we already caught in the above exec
               .expectJson((obj) => {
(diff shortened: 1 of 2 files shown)


packages/server/src/fhircast/websocket.test.ts
```diff
@@ -166,16 +167,21 @@ describe('FHIRcast WebSocket', () => {
...
+              },
+            });
           })
           .sendJson({ ok: true })
           .close()
           .expectClosed();
       }));
 
-    test('Check that timer and promises are cleaned up after no topics active', () =>
+    test('Make sure that we only get one heartbeat per tick for a given topic', () =>
```
ThatOneBro, line 184: I repurposed this test since I realized it was the exact same test as the one above with a different name 😬 So gave it a little more purpose now

packages/server/src/fhircast/websocket.ts
```diff
@@ -1,11 +1,56 @@
...
+            projectAndTopic as string,
+            JSON.stringify({
+              ...baseHeartbeatPayload,
+              event: { ...baseHeartbeatPayload.event, 'hub.topic': projectAndTopic.split(':')[1] },
+            })
+          )
+          .catch(console.error);
+      }
+
+      setGauge('medplum.fhircast.websocketCount', websocketMap.size, METRIC_OPTIONS);
```
ThatOneBro, line 43: New metrics for FHIRcast `websocketCount` and `topicCount`, just for future reference

mattlong (approved):

PR #5808, opened 2025-01-17
Helm chart cleanup

codyebberson:


charts/Chart.yaml
```diff
@@ -1,24 +1,47 @@
...
-# A chart can be either an 'application' or a 'library' chart.
-#
-# Application charts are a collection of templates that can be packaged into versioned archives
-# to be deployed.
-#
-# Library charts provide useful utilities or functions for the chart developer. They're included as
-# a dependency of application charts to inject those utilities and functions into the rendering
-# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+description: |
+  Helm chart for deploying Medplum Server - an open source healthcare development platform.
```
codyebberson, line 4: These values will appear in Artifact Hub (like npm for Helm charts)

charts/templates/deployment.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     {{- include "medplum.labels" . | nindent 4 }}
 spec:
   {{- if not .Values.deployment.autoscaling.enabled }}
-  replicas: {{ default 1 .Values.replicaCount }}
+  replicas: {{ default 1 .Values.deployment.replicaCount }}
```
codyebberson, line 10: This is a bug fix

charts/templates/deployment.yaml
```diff
@@ -40,17 +40,8 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.deployment.image.repository }}:{{ .Values.deployment.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: IfNotPresent
-          command:
```
codyebberson, line 43: This command already exists inside [Dockerfile](https://github.com/medplum/medplum/blob/main/Dockerfile#L34), so it is preferred to not override it.

Medplum server, and therefore the Docker image, take one argument, which is "where to load config from".  Rather than trying to use YAML logic to construct that string, I think it will be better to treat that as a single opaque value that the system administrator can enter directly.

reshmakh (approved):

PR #5815, opened 2025-01-17
Always populate default search parameters

mattlong:
previously, `indexDefaultSearchParameters` had to be explicitly called for a bundle containing resource StructureDefinitions. Now that happens implicitly in `indexStructureDefinitionBundle`.

Note this is chained off of another open PR, https://github.com/medplum/medplum/pull/5812, so I'll need to merge the other one first.

### DIFF
--- a/packages/core/src/search/details.ts
+++ b/packages/core/src/search/details.ts
@@ -104,7 +104,7 @@ function buildSearchParameterDetails(resourceType: string, searchParam: SearchPa
       builder.array = true;
       builder.propertyTypes.add('code');
     } else {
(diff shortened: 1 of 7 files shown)


codyebberson (approved):

codyebberson:
Should we rename `indexStructureDefinitionBundle` to something more general, like `indexDefinitionBundle`?

mattlong:
> Should we rename `indexStructureDefinitionBundle` to something more general, like `indexDefinitionBundle`?

@codyebberson I'm indifferent to renaming; if it had had "Resource" in the name, I'd say so, but I think `indexStructureDefinitionBundle` is still accurate enough.

PR #5820, opened 2025-01-18
feat(server): data migration rework

### LINKED ISSUE #5566: Enable data migrations

ThatOneBro:
Fixes #5566
Replaces #5400


packages/server/src/database.ts
```diff
@@ -149,11 +178,58 @@ async function migrate(client: PoolClient): Promise<void> {
...
+    pendingDataMigration = dataVersion + 1;
+
+    const manifest = JSON.parse(
+      readFileSync(resolve(__dirname, 'migrations/data/data-version-manifest.json'), { encoding: 'utf-8' })
+    ) as Record<string, { serverVersion: string }>;
+    const requiredServerVersion = manifest['v' + pendingDataMigration].serverVersion;
+
+    const serverVersion = getServerVersion();
+
+    globalLogger.error('required', { requiredServerVersion });
```
ThatOneBro, line 216: Remove these debug logs

packages/server/src/database.ts
```diff
@@ -11,6 +19,24 @@ export enum DatabaseMode {
...
+export function getPendingDataMigration(): number {
+  return pendingDataMigration;
+}
+
+export async function markPendingDataMigrationCompleted(job: AsyncJob): Promise<void> {
+  assert(job.dataVersion);
+  await getDatabasePool(DatabaseMode.WRITER).query('UPDATE "DatabaseMigration" SET "dataVersion" = $1', [
+    job.dataVersion,
+  ]);
+  pendingDataMigration = 0;
```
mattwiller, line 38: Why 0 here and not -1?

mattwiller: Making specific `const` values for -1 and 0 might be useful, at least

ThatOneBro: Oh good call

packages/server/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "module": "NodeNext",
     "moduleResolution": "NodeNext"
   },
-  "include": ["src/**/*.ts"]
+  "include": ["src/**/*.ts", "src/**/*.json"]
```
ThatOneBro, line 7: Needed to add `.json` so that `data-version-manifest.json` is included in the `dist`

mattwiller (approved):

packages/server/src/database.ts
```diff
@@ -1,16 +1,45 @@
...
 import * as migrations from './migrations/schema';
-import { sleep } from '@medplum/core';
+import { getServerVersion } from './util/version';
 
 export enum DatabaseMode {
   READER = 'reader',
   WRITER = 'writer',
 }
 
+const UNKNOWN = -1;
```
codyebberson, line 20: Minor nit:  Let's either give these a prefix (i.e., `DATA_VERSION_UNKNOWN`) or wrap them in a const object

packages/server/src/util/version.ts
```diff
@@ -0,0 +1,13 @@
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+let serverVersion: string | undefined;
+
+export function getServerVersion(): string {
```
codyebberson, line 6: We should not need this.  It should always be the same as `MEDPLUM_VERSION`

```ts
import { MEDPLUM_VERSION } from '@medplum/core';
```

packages/server/src/database.ts
```diff
@@ -149,11 +181,53 @@ async function migrate(client: PoolClient): Promise<void> {
...
+
+    const serverVersion = getServerVersion();
+
+    // TODO(ThatOneBro 16 Dec 2024): Make this version strict after v4 (exact version only)
+    // ----  We made this requirement looser so that self-hosters can run first migration on any version within the minor version before v4
+
+    // We allow any version where the data migration is present and it less than the required version
+    // To run the migration, as we can assume that after the data migration is added, any version between the migration being present
+    // And the minor release that it is supposed to be run in will have the capacity to safely run the migration
+    if (!semver.satisfies(serverVersion, `<${semver.inc(requiredServerVersion, 'minor')}`)) {
```
codyebberson, line 223: Is there a `<=` syntax?

ThatOneBro: We want it to be less than the next minor version, not equal to it. It's expected to be applied before that version and we don't want the server to run unless it has been applied by that minor version

packages/server/src/util/version.ts
ThatOneBro: `MEDPLUM_VERSION` is not exactly the same as the version present in `package.json`. Also it makes it a lot harder to change the version for local testing purposes if you only refer to `MEDPLUM_VERSION` because it requires the version to be changed in `package.json` in `@medplum/core` and then build it in order to go back and forth between version numbers.

We can remove it but that means everywhere we have to in-line process `MEDPLUM_VERSION` to account for the fact that it could have a hash or maybe not

packages/server/src/database.ts
ThatOneBro: For example:

| serverVersion  | requiredServerVersion | Comparison | Result |
| ------------- | ------------- | --------- | ------------ |
| 3.3.0  | 3.3.0  | 3.3.0 < 3.4.0 | Passes |
| 3.3.25 | 3.3.0 | 3.3.25 < 3.4.0 | Passes |
| 3.4.0  | 3.3.0  | 3.4.0 < 3.4.0 | Fails, as desired |

The point is to allow the first data migration to be applied on any version before 4.0.0.

Then after v4 is released, we can do a direct comparison, so we can ensure that the patch release after a data migration can use the changes

```ts
if (serverVersion === requiredServerVersion) {
```

packages/server/src/util/version.ts
ThatOneBro: Would it be ok to add `getServerVersion` which strips the hash from the version?

ThatOneBro:
Deployed to staging with a `v2` migration, looks like it's working 👍 
Screenshot 2025-01-27 at 5 58 40 PM

packages/server/src/database.ts
ThatOneBro: After talking online, we potentially want to just encode the version that the data migration has to be ran before being able to start up on as a separate entry in `data-version-manifest.json`, potentially as `requiredBefore` or similar. For example:
```ts
{
  "v1": {
    "serverVersion": "3.3.0" // the version to skip to for upgrader
    "requiredBefore": "4.0.0" // Needs to be applied BEFORE this version
  },
  "v2": {
    "serverVersion": "4.1.0", // again, version to skip to when updating from the upgrade-server CLI command
    "requiredBefore": "4.1.1" // Required before next patch version like the proposed convention, but we encode it explicitly
  }
}
```

This would allow us to account for data migrations which are to be applied any time before a major version, but also allow for flexibility to require a migration to be applied on the same version it was introduced in by setting the `requiredBefore` version to the subsequent version.

What do we think about that?

ThatOneBro:
Tested on staging one final time with a fake v3 migration on this branch:
https://github.com/medplum/medplum/tree/derrick-reworked-migration-test

When deployed with `requiredVersion` lower than the current version, stack refused to start up (as expected):
Screenshot 2025-02-10 at 2 18 25 PM

When deployed with `requiredVersion` higher than the current version, was able to successfully migrate:
Screenshot 2025-02-10 at 3 00 31 PM
Screenshot 2025-02-10 at 3 00 40 PM

PR #5832, opened 2025-01-20
Add additional token search tests

mattlong:
This also corrects our implementation of search for token of the form `[parameter]=|[code]` by falling back to `NULL` rather than looking for rows in the token table with a system value of the empty string.

From the spec:
> [parameter]=|[code]: the value of [code] matches a Coding.code or Identifier.value, and the Coding/Identifier has no system property

### DIFF
--- a/packages/server/src/fhir/lookups/token.ts
+++ b/packages/server/src/fhir/lookups/token.ts
@@ -97,6 +97,7 @@ export class TokenTable extends LookupTable {
     const values = tokens.map((token) => ({
       resourceId,
       code: token.code,
+      // logical OR coalesce to ensure that empty strings are inserted as NULL
       system: token.system?.trim?.() || undefined,
       value: token.value?.trim?.() || undefined,
     }));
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/lookups/token.ts
```diff
@@ -516,7 +516,7 @@ function buildWhereCondition(
   // Handle the case where the query value is a system|value pair (e.g. token or identifier search)
   if (parts.length === 2) {
     const [system, value] = parts;
-    const systemCondition = new Condition(new Column(tableName, 'system'), '=', system);
+    const systemCondition = new Condition(new Column(tableName, 'system'), '=', system || null);
```
codyebberson, line 519: This makes sense

https://hl7.org/fhir/r4/search.html#token

```
[parameter]=|[code]: the value of [code] matches a Coding.code or Identifier.value, and the Coding/Identifier has no system property
```

codyebberson (approved):
lgtm 👍

PR #5833, opened 2025-01-20
Improve cron job change detection

codyebberson:
Before:

1. We were checking for `cronString === ''`, which will never happen, because that's not valid FHIR
2. We weren't using `previousVersion`, presumably because it had not been introduced yet at this point
3. Bottom line, "disabling" a cron a very confusing experience
4. Also, we were using a bunch of deprecated features

After:

1. Use `previousVersion` for concrete check on whether the cron string changed or not
2. Use the recommended methods from bullmq
3. Should be much more deterministic regarding updating the bullmq job

### DIFF
--- a/packages/server/src/workers/cron.ts
+++ b/packages/server/src/workers/cron.ts
@@ -22,7 +22,6 @@ export interface CronJobData {
 }
 
 const queueName = 'CronQueue';
-const jobName = 'CronJobData';
 
 let queue: Queue<CronJobData> | undefined = undefined;
 let worker: Worker<CronJobData> | undefined = undefined;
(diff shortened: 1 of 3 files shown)


packages/server/src/workers/cron.ts
```diff
@@ -112,51 +111,52 @@ export async function addCronJobs(resource: Resource): Promise<void> {
...
-    {
-      resourceType: bot.resourceType,
-      botId: bot.id as string,
-    },
-    jobOptions
-  );
+  if (newCronStr) {
+    logger.info('Adding cron job for bot', { botId: bot.id });
+    if (queue) {
+      await queue.add(
```
ThatOneBro, line 131: Should we be using `upsertJobScheduler`?
https://docs.bullmq.io/guide/job-schedulers

codyebberson: Oo, nice find. Yes, will update.  Thanks!

rahul1 (approved):

ThatOneBro (approved):

PR #5841, opened 2025-01-21
feat(super): add `reloadcron` super admin endpoint

ThatOneBro:
Serves two purposes:
1. Clears out potentially orphaned jobs in the `cron` queue
2. Makes sure cron queue job schedulers are synced with the current cron strings of all `Bot` resources

Tested locally:
Screenshot 2025-01-21 at 9 43 20 AM

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1163,7 +1163,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
 
   async processAllResources<T extends Resource>(
     initialSearchRequest: SearchRequest<T>,
-    processPage: (resources: T[]) => Promise<void>,
+    process: (resource: T) => Promise<void>,
     options?: ProcessAllResourcesOptions
   ): Promise<void> {
     let searchRequest: SearchRequest<T> | undefined = initialSearchRequest;
@@ -1172,13 +1172,11 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       if (!bundle.entry?.length) {
         break;
       }
-      const resources: T[] = [];
       for (const entry of bundle.entry) {
         if (entry.resource?.id) {
-          resources.push(entry.resource);
+          await process(entry.resource);
         }
       }
-      await processPage(resources);
       const nextLink = bundle.link?.find((b) => b.relation === 'next');
       if (nextLink) {
         searchRequest = parseSearchRequest<T>(nextLink.url);
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1156,6 +1156,35 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     }
   }
 
+  async processAllResources<T extends Resource>(
+    initialSearchRequest: SearchRequest<T>,
+    processPage: (resources: T[]) => Promise<void>
+  ): Promise<void> {
+    let searchRequest: SearchRequest<T> | undefined = initialSearchRequest;
+    while (searchRequest) {
+      const bundle: Bundle<T> = await this.search<T>(searchRequest);
+      console.log({ bundle });
+      if (!bundle.entry?.length) {
+        break;
+      }
+      const resources: T[] = [];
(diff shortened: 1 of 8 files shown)


packages/server/src/workers/cron.ts
```diff
@@ -216,3 +216,19 @@ export async function removeBullMQJobByKey(botId: string): Promise<void> {
     await queue.removeJobScheduler(botId);
   }
 }
+
+export async function reloadCronBots(): Promise<void> {
+  if (queue) {
+    // Clears all jobs from the cron queue, including active ones
+    await queue.obliterate({ force: true });
+
+    const allBots = await getSystemRepo().searchResources<Bot>({ resourceType: 'Bot' });
```
codyebberson, line 225: This will only get the first page, with page size 20.  We will need to add a pagination wrapper.

The total number of bots in most prod systems is reasonably (~1000), so this shouldn't be a major performance issue, but we should probably still follow best practices for pagination.

See `exportResourceType` in `packages/server/src/fhir/operations/export.ts` for an example.

packages/server/src/workers/cron.ts
```diff
@@ -216,3 +216,19 @@ export async function removeBullMQJobByKey(botId: string): Promise<void> {
     await queue.removeJobScheduler(botId);
   }
 }
+
+export async function reloadCronBots(): Promise<void> {
+  if (queue) {
+    // Clears all jobs from the cron queue, including active ones
+    await queue.obliterate({ force: true });
```
codyebberson, line 223: lol at `obliterate`.

obliterate, purge, expunge... we're collecting quite the vocabulary for "kill all the things" 🙃

codyebberson (changes requested):
See comment on paginating bots

packages/server/src/workers/cron.ts
ThatOneBro: Added `processAllResources` method to `Repository` as an abstraction of this pattern. Funny enough the pattern didn't quite work for `expunge`, which the tests caught, since following the `next` link doesn't work after deleting the resources behind the cursor since it affects the offset.

Added a note about that and move cron to use this new abstraction. While there also added a delay between pages, so hopefully this will scale better. Also made the `cron reload` async, though it should be pretty snappy for now since most of the bots will not be cron bots and max page size is 500

packages/server/src/fhir/repo.ts
```diff
@@ -1156,6 +1161,36 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     }
   }
 
+  async processAllResources<T extends Resource>(
```
codyebberson, line 1164: Can we add a comment to this method?

packages/server/src/fhir/repo.ts
```diff
@@ -1156,6 +1161,36 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    initialSearchRequest: SearchRequest<T>,
+    processPage: (resources: T[]) => Promise<void>,
```
codyebberson, line 1166: Should the callback be `processPage` or `processResource`?  

It looks like all of the call sites just iterate over `resources`, so maybe we can just make that the API contract?

ThatOneBro: Initially when trying to get `expunge` to fit into the shape, I noticed there was a pattern of collecting the IDs for a page and then calling `expungeResources` on the whole page at once. After removing that callsite it appears that we only iterate but could see a future caller that wants to bulk read references or insert or something...

But if we think that's premature we can move to the `processResource` model

ThatOneBro: Also for reference we have `processPage` as part of the `LongJob` framework, which operates in a similar way for the reindex job (though using the "cursor"):
https://github.com/medplum/medplum/blob/derrick-add-cron-bot-reload-superadmin/packages/server/src/workers/reindex.ts#L179

ThatOneBro: Okay I changed it to operate on `T` instead of `T[]` now

codyebberson (approved):

PR #5846, opened 2025-01-22
Fix FHIRPath is() check

mattwiller:


codyebberson (approved):

packages/core/src/fhirpath/utils.ts
```diff
@@ -446,7 +446,7 @@ export function fhirPathIs(typedValue: TypedValue, desiredType: string): boolean
     case 'Quantity':
       return isQuantity(value);
     default:
-      return typeof value === 'object' && value?.resourceType === desiredType;
+      return typeof value === 'object' && (value?.resourceType === desiredType || typedValue.type === desiredType);
```
codyebberson, line 449: Actually, super minor, but:
1. We may want to invert the order of the `||` because
2. Checking `typedValue.type` will be faster than `typeof`
3. It maybe sends the correct signal of "this is the default case, and here's an oddball legacy case"

```suggestion
      return typedValue.type === desiredType || (typeof value === 'object' && value?.resourceType === desiredType);
```

ThatOneBro (approved):

PR #5851, opened 2025-01-22
Enable lint rule for returning Promise from try catch

mattwiller:
See https://typescript-eslint.io/rules/return-await/


ThatOneBro (approved):
Good find 👍

PR #5857, opened 2025-01-24
Fetch related non-compartment resources for Patient/$everything

mattwiller:
Extending #5825 and getting tests passing:
- Selecting only a specific subset of referenced resource types to resolve
- Resolving references in batches for each tranch of resources
- Updating test case to validate what's returned
- Maintaining correct pagination, marking included referenced resources via `Bundle.entry.search.mode`
  - Note that these may be duplicated across multiple pages of operation results, and should not affect paging
- Optimizing by minimizing intermediate allocations

This passes Inferno test `Multi-Patient Authorization and API STU2` for the required included resource types, since Group Export [uses `Patient/$everything` under the hood](https://github.com/medplum/medplum/blob/433be709a018bb022f3284c0ac02bcc24d1d486b/packages/server/src/fhir/operations/groupexport.ts#L59)


packages/server/src/fhir/operations/patienteverything.ts
```diff
@@ -74,11 +90,82 @@ export async function getPatientEverything(
...
+async function addResolvedReferences(repo: Repository, entries: BundleEntry[] | undefined): Promise<void> {
+  const processedRefs = new Set<string>();
+  let page = entries;
+  while (page?.length) {
+    const references = processReferencesFromResources(page, processedRefs);
+    const resolved = await repo.readReferences(references);
+    page = flatMapFilter(resolved, (resource) =>
+      isResource(resource) ? { resource, search: { mode: 'include' } } : undefined
+    );
+    entries?.push(...page);
```
ThatOneBro, line 121: Any reason to make a shallow copy here? I think since all these functions are exported and only used locally we know that the page is read only

ThatOneBro (approved):
LGTM, some small nits

packages/server/src/fhir/operations/patienteverything.ts
mattwiller: Page is iterated over each loop, but needs to be copied into `entries` as well; what's the alternative?

ThatOneBro: Oh actually I misread what this block is doing, sorry. I thought we were making a copy of `page` in order to avoid potential mutation in `processReferencesFromResources` but we're just copying the result in `entries` before the next iteration

ThatOneBro (approved):

Issue #4140, opened 2024-03-09
Add `WWW-Authenticate` header to unauthorized response


codyebberson:
Apparently this is necessary as some callers cannot use preemptive basic auth, and instead need to be prompted.

codyebberson:
Fixed in https://github.com/medplum/medplum/pull/5863

PR #5863, opened 2025-01-25
Conditionally include WWW-Authenticate header if query param included in request

mattlong:
We've tried [including WWW-Authenticate](https://github.com/medplum/medplum/pull/4167), but doing so unconditionally unfortunately leads to [a very bad experience](https://github.com/medplum/medplum/pull/4200) in browsers, e.g. when using `@medplum/app`, since it causes a basic authentication dialog to be shown to the user.


ThatOneBro (approved):
Nice, clean

rahul1:
Fixes #4140

PR #5876, opened 2025-01-29
Remove post-filtering of exports for efficiency

mattwiller:
The internal `BulkExporter` utility was applying filtering to the results written, which is inefficient: these resources would be loaded from the database only to be thrown away.  This refactor moves three classes of filtering from `BulkExporter` into the upstream export logic, where they can be applied at the time the database is accessed:
- Removing `AuditEvent` resources from Bulk Export results
- General type filtering (i.e. `_type` in `Patient/$everything` and Group Export)
- `_since` parameter in `Patient/$everything` and Group Export

Incidentally, this also adds support for the `_type` parameter to the `Patient/$everything` operation

### DIFF
--- a/packages/server/src/fhir/operations/export.ts
+++ b/packages/server/src/fhir/operations/export.ts
@@ -50,10 +50,10 @@ async function startExport(req: FhirRequest, exportType: string): Promise<FhirRe
   const since = singularize(req.query._since);
   const types = singularize(req.query._type)?.split(',');
 
-  const exporter = new BulkExporter(ctx.repo, since);
+  const exporter = new BulkExporter(ctx.repo);
   const bulkDataExport = await exporter.start(concatUrls(baseUrl, 'fhir/R4' + req.pathname));
 
-  exportResources(exporter, ctx.project, types, exportType)
+  exportResources(exporter, ctx.project, types, exportType, since)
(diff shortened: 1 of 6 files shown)


packages/server/src/fhir/operations/utils/bulkexporter.ts
```diff
@@ -77,15 +73,6 @@ export class BulkExporter {
   }
 
   async writeResource(resource: Resource): Promise<void> {
-    if (this.types.length > 0 && !this.types.includes(resource.resourceType)) {
```
ThatOneBro, line 80: Nice 👍

packages/server/src/fhir/operations/export.ts
```diff
@@ -119,6 +119,7 @@ const unexportedResourceTypes = [
   'ValueSet',
   'BulkDataExport',
   'AsyncJob',
+  'AuditEvent',
```
ThatOneBro, line 122: 👍

ThatOneBro (commented):
LGTM, just want to deploy to staging first to verify everything is good

ThatOneBro (approved):
Was able to bulk export on staging via Medplum CLI 👍 LGTM

PR #5572, opened 2024-11-21
Remove deprecated functionality for v4

mattwiller:

PR #5880, opened 2025-01-29
Replace deprecated usage of createAttachment

nate-watkins:
Ran across a few instances of `createAttachment` using the deprecated options. I think this cleans it up?


mattwiller:
Hey @nate-watkins — thank you for the PR!  We're preparing for our v4 release soon, which will officially remove the deprecated functionality.   I think one of the existing PRs associated with that release — #5572 — might already include the related cleanup

nate-watkins:
Good call- missed that one. Looking forward to v4

PR #5893, opened 2025-02-01
Adding Code System to Eligibility Demo App

### LINKED ISSUE #5892: medplum-eligibility-demo missing example CodeSystem in resource bundle
Not getting any service types to populate, which means you can't Submit, to initiate eligibility request
Image

DoseSpot is only set up to use the code "30", representing Plan Coverage and General Benefits.

finnbergquist:
In order to select an X-12 service type, there's no code system to populate the drop down in the eligibility demo app.

Added a Code System in the Bundle resource which is added when the user clicks "Upload Core Data". The code system only contains one x12-service-type-code, which is "30"

closes #5892

### DIFF
--- a/examples/medplum-eligibility-demo/data/core/core-data.json
+++ b/examples/medplum-eligibility-demo/data/core/core-data.json
@@ -841,6 +841,24 @@
           ]
         }
       }
+    },
+    {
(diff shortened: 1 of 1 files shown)


ianplunkett (approved):
This looks good to me.

PR #5899, opened 2025-02-03
Update hl7-into-fhir.md

ajwwong:
I don't this is currently present in medplum -- at least I couldn't find it in my version of it.  Perhaps I am mistaken, but I think it is dated.   The docs say you can set access policies for ClientApplications in the Admin UI, but I looked around and AFAIK that functionality doesn't seem to exist.


reshmakh:
Hi @ajwwong!  This is present in Medplum App, if you cannot see it in your Project that means that you are not an Admin of the project.  If you create a new Project as Admin,  or are promoted to Admin as an administrator, you should be good to go.

reshmakh:
Closing this for now.

PR #5903, opened 2025-02-03
feat: Support POST on Bulk Patient Export and Bulk Group Export

jdjkelly:
## Why
* The System-level export supports both POST and GET method
* Patient and Group-level export only supports GET
* The IG supports both methods: https://build.fhir.org/ig/HL7/bulk-data/export.html#bulk-data-kick-off-request
* However, the Medplum CLI / @medplum/core only implements the POST operation, so Patient and Group exports in the CLI are currently broken

## What
* Adds POST method routes for the Patient and Group, the handler is the same for all of them

### DIFF
--- a/packages/server/src/fhir/routes.ts
+++ b/packages/server/src/fhir/routes.ts
@@ -233,9 +233,11 @@ function initInternalFhirRouter(): FhirRouter {
 
   // Group $export operation
   router.add('GET', '/Group/:id/$export', groupExportHandler);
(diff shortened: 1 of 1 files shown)


mattwiller:
@jdjkelly The handler will actually need to be updated, so it can read the operation parameters from either the POST body or GET query parameters.  If you don't have time to update the PR, I can take care of that later!

The-Alchemist:
Merged as part of #7673

PR #5907, opened 2025-02-03
Return User email in userinfo endpoint

### LINKED ISSUE #5894: Endpoint oauth2/userinfo returning Practitioner's email, not User's
I got a question about what email the `oauth2/userinfo` will return (assuming it is given email in its scope).

After creating a new account to login with(creating both a Practitioner and User), I found that `oauth2/userinfo` it returns the first email on the Practitioner's telecom [], even if it no longer matches the User resource's email.

`oauth2/userinfo` should return the email from the User resource, even if the email is changed on the Practitioner.

mattwiller:
Resolves #5894


packages/server/src/oauth/userinfo.ts
```diff
@@ -70,10 +70,15 @@ function buildProfile(userInfo: Record<string, any>, profile: ProfileResource):
...
-    userInfo.email = contactPoint.value;
-    userInfo.email_verified = !!(userInfo.email === user.email && user.emailVerified);
+  if (user.email) {
+    userInfo.email = user.email;
+    userInfo.email_verified = user.emailVerified;
+  } else {
+    const contactPoint = profile.telecom?.find((cp) => cp.system === 'email');
+    if (contactPoint) {
+      userInfo.email = contactPoint.value;
+      userInfo.email_verified = !!(userInfo.email === user.email && user.emailVerified);
```
codyebberson, line 80: Will this expression ever be true?  `user.email` is falsy at this point.

mattlong (approved):

PR #5918, opened 2025-02-05
Allow configuring access token lifetime

### LINKED ISSUE #5154: Config Value for Access Token TTL
We have a field on `ClientApplication` that allows admins to set the length of the refreshToken. 

We should a field to shorten the Access token TTL (max 1h), to make it easier to developers to debug token refresh issues

mattwiller:
Resolves #5154

### DIFF
--- a/packages/server/src/oauth/token.test.ts
+++ b/packages/server/src/oauth/token.test.ts
@@ -1139,6 +1139,76 @@ describe('OAuth2 Token', () => {
     expect(res5.body).toMatchObject({ error: 'invalid_request', error_description: 'Invalid token' });
   });
 
+  test('accessTokenLifetime -- Valid duration', async () => {
+    // Create a new client application with external auth
+    const validLifetimeClient = await createClient(systemRepo, {
+      project,
+      name: 'accessTokenLifetime - Valid Client',
+      accessTokenLifetime: '60s',
+    });
+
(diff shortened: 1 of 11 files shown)


packages/server/src/oauth/token.test.ts
```diff
@@ -1472,7 +1542,7 @@ describe('OAuth2 Token', () => {
       client_assertion: jwt,
     });
     expect(res.status).toBe(200);
-    expect(jwtVerify).toHaveBeenCalledTimes(3);
+    expect(jwtVerify).toHaveBeenCalledTimes(4);
```
codyebberson, line 1545: Out of curiosity, why did this change from 3 to 4?

codyebberson (approved):
Overall code looks good

One question on the `expect(jwtVerify).toHaveBeenCalledTimes(4);` - if there's a good reason, great.  Cursory review suggests that we may not be clearing our mocks properly?

packages/server/src/oauth/token.test.ts
mattwiller: Due to [this additional call](https://github.com/medplum/medplum/pull/5918/files#diff-8684369c3fa991aac52183ced98bbf3cf1b849be67da8c3e4772d9db132f098aR609) to `jwtVerify()`

PR #5919, opened 2025-02-05
Fixed Bundle-composition search

codyebberson:


mattwiller (approved):

packages/server/src/fhir/lookups/reference.ts
```diff
@@ -87,22 +87,41 @@ function getSearchReferences(resource: Resource): ReferenceRow[] {
 
     const typedValues = evalFhirPathTyped(searchParam.expression as string, typedResource);
     for (const value of typedValues) {
-      if (value.type !== PropertyType.Reference || !value.value.reference) {
-        continue;
+      if (value.type === PropertyType.Reference && value.value.reference) {
+        const [_targetType, targetId] = value.value.reference.split('/', 2);
```
ThatOneBro, line 91: We also have `resolveId()` that should work here yeah?

codyebberson: Oo, good call. This was existing code, but I'll touch it up.

codyebberson: ✅

PR #5924, opened 2025-02-05
Include projectId in new token logging

mattlong:
Followup to #5922


mattwiller (approved):

mattlong:
@mattwiller The implementation of this has completely changed in case you want to take another look.

mattwiller (approved):

PR #5939, opened 2025-02-10
Optionally pass client credentials in HTTP header

### LINKED ISSUE #5897: Authentication issue with OAuth 2.0  servers that do not support request-body token authentication
We've been experiencing an authentication issue when using machine-to-machine authentication (although the problem is not limited to this use case) with one of our providers.
When trying to authenticate with the provider, whether by direct use of startClientLogin, or via refresh (which used this function), we would receive a "Failed to fetch tokens".

After some investigation, we found the root cause to be this:
You seem to send the client_id and client_secret in the body of the client credentials request.
https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts#L3647

However, this causes an error when we try to authenticate with our provider since they don't support this method, but only support header authentication (since the Auth2 specs specify that header authentication MUST be supported but that request-body MAY be supported).
https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

I am somewhat weary of offering a PR for that myself since authentication changes could undoubtably impact a variety of things in your framework but I was hoping there could be some resolution to this, perhaps an optional flag to choose between the two if no better way?

mattwiller:
Resolves #5897


packages/core/src/client.ts
```diff
@@ -3875,22 +3885,25 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
+  private async fetchTokens(params: Record<string, string>): Promise<ProfileResource> {
+    const formBody = new URLSearchParams(params);
+    const headers: HeadersInit = { 'Content-Type': ContentType.FORM_URL_ENCODED, ...this.defaultHeaders };
+    if (this.basicAuth) {
+      headers['Authorization'] = `Basic ${this.basicAuth}`;
+    } else if (this.credentialsInHeader && params.client_id && params.client_secret) {
+      headers['Authorization'] = `Basic ${encodeBase64(params.client_id + ':' + params.client_secret)}`;
+      formBody.delete('client_id');
+      formBody.delete('client_secret');
+    }
```
elaye-canopy, line 3900: What if basicAuth is defined, but also credentialsInHeader is true?

Based on the subsequent logic of the else, you might want to delete the client_id and client_secret from the body in that situation.

Perhaps something along the lines of

```suggestion
if (this.basicAuth) {
  headers['Authorization'] = `Basic ${this.basicAuth}`;
} 

if (this.credentialsInHeader) {
  formBody.delete('client_id');
  formBody.delete('client_secret');

  if ((!this.basicAuth) && (params.client_id && params.client_secret)) {
    headers['Authorization'] = `Basic ${encodeBase64(params.client_id + ':' + params.client_secret)}`;
  }
}
```

packages/core/src/client.ts
```diff
@@ -3875,22 +3885,30 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
    * Makes a POST request to the tokens endpoint.
    * See: https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint
-   * @param formBody - Token parameters in URL encoded format.
+   * @param params - Token parameters.
    * @returns The user profile resource.
    */
-  private async fetchTokens(formBody: URLSearchParams): Promise<ProfileResource> {
+  private async fetchTokens(params: Record<string, string>): Promise<ProfileResource> {
+    const formBody = new URLSearchParams(params);
+    const headers: HeadersInit = { 'Content-Type': ContentType.FORM_URL_ENCODED, ...this.defaultHeaders };
```
ThatOneBro, line 3893:
```suggestion
    const headers: HeadersInit = { ...this.defaultHeaders, 'Content-Type': ContentType.FORM_URL_ENCODED };
```
We probably don't want default headers to clobber anything we are setting in-line, though this is probably pedantic

ThatOneBro (commented):
LGTM on first pass, small nit and want to try deploying to staging first

packages/core/src/client.ts
mattwiller: You're right, it makes more sense this way :+1:

mattwiller:
@ThatOneBro These are client-side only changes, nothing to deploy to Staging

codyebberson (commented):
👍

codyebberson (approved):

PR #5941, opened 2025-02-10
Implement SMART launch operation endpoint

mattwiller:
Adding a bespoke `ClientApplication/:id/$launch` operation endpoint to handle redirecting to the `launchUri`, plus additional logic for validation and creating the `SmartAppLaunch` resource.  This largely replaces the client-side logic in [the `SmartAppLaunchLink` component](https://github.com/medplum/medplum/blob/89791c928fcded92050af7b1f9ab0a37398a301d/packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx) and enables simply making a client-side GET request to the operation endpoint from the user's browser and letting the server handle the rest

### DIFF
--- a/packages/core/src/outcomes.ts
+++ b/packages/core/src/outcomes.ts
@@ -5,6 +5,7 @@ const OK_ID = 'ok';
 const CREATED_ID = 'created';
 const GONE_ID = 'gone';
 const NOT_MODIFIED_ID = 'not-modified';
+const FOUND_ID = 'found';
 const NOT_FOUND_ID = 'not-found';
 const CONFLICT_ID = 'conflict';
 const UNAUTHORIZED_ID = 'unauthorized';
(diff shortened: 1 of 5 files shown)


mattwiller:
@rahul1 Sure thing: the plan was to update the `SmartAppLaunchLink` component after this PR is merged, so if `ClientApplication.id` is present it just generates e.g. `<a href="BASE_URL/fhir/R4/ClientApplication/ID/$launch"> App Name </a>`

packages/core/src/outcomes.ts
```diff
@@ -280,6 +281,23 @@ export function serverTimeout(msg?: string): OperationOutcome {
   };
 }
 
+export function redirect(url: string): OperationOutcome {
```
ThatOneBro, line 284: I wonder if we want to change the signature of this to:
```suggestion
export function redirect(url: URL): OperationOutcome {
```
That way we make sure any caller has to assert this is a valid URL

Just food for thought

ThatOneBro (approved):
Overall LGTM, some small nits

Like the overall vision of having just a single endpoint that takes care of everything for the client, SMART launch is complicated enough as it is 👍

ThatOneBro:
Actually for discoverability, do we want to call this `ClientApplication/$smart-launch`?

This clues people into the fact this is the same thing as a SMART launch which might not be obvious

packages/core/src/outcomes.ts
```diff
@@ -280,6 +281,24 @@ export function serverTimeout(msg?: string): OperationOutcome {
...
+  const urlStr = url.toString();
+  return {
+    resourceType: 'OperationOutcome',
+    id: FOUND_ID,
+    issue: [
+      {
+        severity: 'information',
+        code: 'informational',
+        details: {
+          coding: [{ system: 'urn:ietf:rfc:3986', code: urlStr }],
```
codyebberson, line 294: Nice

codyebberson (approved):
🚀

ThatOneBro (approved):

PR #5945, opened 2025-02-11
Filter encounters for the corresponding user and sort by lastUpdated

techdavidy:


examples/medplum-provider/src/pages/patient/PatientPage.utils.ts
```diff
@@ -34,7 +34,11 @@ export function getPatientPageTabOrThrow(tabId: string): PatientPageTabInfo {
 export const PatientPageTabs: PatientPageTabInfo[] = [
   { id: 'timeline', url: '', label: 'Timeline' },
   { id: 'edit', url: 'edit', label: 'Edit' },
-  { id: 'encounter', url: 'Encounter?_count=20&_fields=_lastUpdated,period,status,type', label: 'Encounter' },
+  {
+    id: 'encounter',
+    url: 'Encounter?_count=20&_fields=_lastUpdated,period,status,type&_sort=-_lastUpdated&patient=%patient.id',
```
ThatOneBro, line 39: Not sure if this is right but might also want to think about sorting by period, but not sure if providers expect to see things sorted by reverse chronological order or lastUpdated

ThatOneBro (approved):

examples/medplum-provider/src/pages/patient/PatientPage.utils.ts
techdavidy: Good point. But since this is the default sorting, the resource page allows to changes in sorting.

PR #5948, opened 2025-02-11
Log potential validation failures in non-strict mode

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -755,30 +755,45 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    }
+  }
 
-      const profileUrls = resource.meta?.profile;
-      if (profileUrls) {
-        await this.validateProfiles(resource, profileUrls);
-      }
+  private async validateResourceStrictly(resource: Resource): Promise<void> {
+    const logger = getLogger();
+    const start = process.hrtime.bigint();
```
ThatOneBro, line 777: Is there any reason why we use hrtime here if we end up converting to millis anyways? Can we get away with `Date.now()`?

PR #5949, opened 2025-02-11
Filter sent:missing=false communications in BaseChat

techdavidy:
slack thread https://medplum.slack.com/archives/C02AMUU6SBH/p1738779271086559

Screenshot 2025-02-11 at 3 04 41 PM

### DIFF
--- a/packages/react/src/chat/BaseChat/BaseChat.tsx
+++ b/packages/react/src/chat/BaseChat/BaseChat.tsx
@@ -78,7 +78,7 @@ export interface BaseChatProps extends PaperProps {
  *
  * **NOTE: The component automatically filters `Communication` resources where the `sent` property is `undefined`.**
  *
- * @param props - TheBaseChat React props.
+ * @param props - The BaseChat React props.
  * @returns The BaseChat React node.
  */
 export function BaseChat(props: BaseChatProps): JSX.Element | null {
--- a/packages/react/src/chat/BaseChat/BaseChat.tsx
+++ b/packages/react/src/chat/BaseChat/BaseChat.tsx
@@ -75,7 +75,8 @@ export interface BaseChatProps extends PaperProps {
 
 /**
  * BaseChat component for displaying and managing communications.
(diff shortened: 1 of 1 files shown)


mattwiller (approved):

ThatOneBro (approved):
This is a tricky one, I think filtering out messages without a sent time kind of makes sense for now, we probably just need to document this behavior somewhere so that authors are aware of it, otherwise it could be confusing that some Communications just disappear

But it is probably the right thing to do

packages/react/src/chat/BaseChat/BaseChat.tsx
```diff
@@ -73,6 +73,13 @@ export interface BaseChatProps extends PaperProps {
   readonly onError?: (err: Error) => void;
 }
 
+/**
+ * BaseChat component for displaying and managing communications.
+ * Component filters communcations where property sent is undefined.
+ *
+ * @param props - TheBaseChat React props.
```
ThatOneBro, line 80:
```suggestion
 * @param props - The BaseChat React props.
```

packages/react/src/chat/BaseChat/BaseChat.tsx
```diff
@@ -73,6 +73,13 @@ export interface BaseChatProps extends PaperProps {
   readonly onError?: (err: Error) => void;
 }
 
+/**
+ * BaseChat component for displaying and managing communications.
+ * Component filters communcations where property sent is undefined.
```
ThatOneBro, line 78:
```suggestion
 *
 * **NOTE: The component automatically filters `Communication` resources where the `sent` property is `undefined`.**
```

ThatOneBro: I like to bold notes like these to draw attention when hovering over them in the IDE

ThatOneBro (approved):
Minor nit on the style of the JSDoc comments

Sorry I'm super particular about using the backticks for readability 😅

PR #5953, opened 2025-02-12
Log accounts writes

### LINKED ISSUE #5952: Add logging for non-patient PUT calls with meta.accounts
To better understand whether anyone is updating the _compartment_ on any non-patient resources by writing a new _meta.account_ or _meta.accounts_  as an admin, add logging of the _meta.account_ or _meta.accounts_ diffs on any update.

mattwiller:
Resolves #5952

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1834,7 +1834,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         // Log intentional writes to accounts on non-Patient resources
         if (
           !existing ||
-          new Set(this.extractAccountReferences(existing.meta)).difference(new Set(updatedAccounts)).size > 0
+          new Set(this.extractAccountReferences(existing.meta)).symmetricDifference(new Set(updatedAccounts)).size > 0
         ) {
           getLogger().warn('Write to accounts within compartment', { resource: getReferenceString(updated) });
         }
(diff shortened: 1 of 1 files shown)


ThatOneBro (approved):

rahul1 (approved):

packages/server/src/fhir/repo.ts
```diff
@@ -1830,6 +1830,15 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
   private async getAccounts(existing: Resource | undefined, updated: Resource): Promise<Reference[] | undefined> {
     const updatedAccounts = this.extractAccountReferences(updated.meta);
     if (updatedAccounts && this.canWriteAccount()) {
+      if (updated.resourceType !== 'Patient') {
+        // Log intentional writes to accounts on non-Patient resources
+        if (
+          !existing ||
+          new Set(this.extractAccountReferences(existing.meta)).difference(new Set(updatedAccounts)).size > 0
```
ThatOneBro, line 1837: Actually now that I'm checking this, this only fires when the first set contains an element that the second set does not, and not the other way around

There is also `.symmetricDifference` if we want to know if there is a missing element on either side of the comparison https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/symmetricDifference

mattwiller: @ThatOneBro Great catch!

ThatOneBro: I would have expected the default behavior to be what `.symmetricDifference` is doing, but seems these folks are big set theory people

PR #5959, opened 2025-02-12
Log account writes

mattwiller:
Second attempt at #5953 (which was reverted in #5957) after manually implementing difference between arrays of references

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1834,7 +1834,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         // Log intentional writes to accounts on non-Patient resources
         if (
           !existing ||
-          new Set(this.extractAccountReferences(existing.meta)).difference(new Set(updatedAccounts)).size > 0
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -1898,6 +1904,36 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     }
   }
 
+  private sameReferences(a: Reference[] | undefined, b: Reference[] | undefined): boolean {
```
ThatOneBro, line 1907: We might be able to do something closer to the original code, like:

`a.difference(b).size === 0 && b.difference(a).size === 0`

mattwiller: True, I was just trying to avoid allocating the arrays from `.difference()`

ThatOneBro: Fair,

If we can at least get a test case that represents what we are trying to log then I think we can get enough coverage for this

I guess scenarios could be:
* Create non-Patient resource with 1 account
* Have a project admin remove that account

* Create non-Patient resource with 1 account
* Have a project admin add another account

Something like that?

ThatOneBro (approved):
👍

PR #5960, opened 2025-02-12
QuestionnaireBuilder override item.requiered

techdavidy:
QuestionnaireFormItem is being reused by `QuestionnaireForm` and `QuestionnaireBuilder`. 
This introduced an issue that submitting a `QuestionnaireBuilder` that contains an item where `required: true`, the form won't submit.
Added an override to `QuestionnaireFormItem` so  `QuestionnaireBuilder` can set required to false.

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
@@ -37,7 +37,7 @@ import { QuestionnaireFormContext } from '../QuestionnaireForm.context';
 export interface QuestionnaireFormItemProps {
   readonly item: QuestionnaireItem;
   readonly index: number;
-  readonly isRequired?: boolean;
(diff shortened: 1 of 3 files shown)


mattwiller (approved):

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -37,6 +37,7 @@ import { QuestionnaireFormContext } from '../QuestionnaireForm.context';
 export interface QuestionnaireFormItemProps {
   readonly item: QuestionnaireItem;
   readonly index: number;
+  readonly isRequired?: boolean;
```
codyebberson, line 40: Can we change this to just `required` ?

We try to reserve the `is` prefix for getter functions.  Boolean variables should just be the word itself.

codyebberson (changes requested):

codyebberson (approved):

PR #5961, opened 2025-02-12
Execute Bot respecting On-Behalf-Of header

### LINKED ISSUE #5956: Bot $execute with X-Medplum-On-Behalf-Of header
### Current Behavior

Currently, when bots with the _runAsUser_ field set to true are executed, the Bot will run using the credentials specified by the ProjectMembership of the ClientApplication or Practitioner.
It does not take into account the `X-Medplum-On-Behalf-Of` header which can be added as a header to any HTTP request from a server side app to Medplum server, authenticated with client credentials (client id and client secret).  This is because the user is technically set to the ClientApplication.

### Desired Behavior

Even when _runAsUser_ is set to true for a bot, if the `X-Medplum-On-Behalf-Of` header is added to HTTP request `/Bot/[id]/$execute` the Bot should use the credentials from the User referenced in the `X-Medplum-On-Behalf-Of` header.

mattwiller:
Resolves #5956

### DIFF
--- a/packages/server/src/fhir/operations/execute.test.ts
+++ b/packages/server/src/fhir/operations/execute.test.ts
@@ -1,4 +1,4 @@
-import { ContentType, Operator, badRequest, createReference, getReferenceString } from '@medplum/core';
+import { ContentType, Operator, badRequest, createReference, getReferenceString, parseJWTPayload } from '@medplum/core';
 import {
   AsyncJob,
   AuditEvent,
@@ -537,7 +537,7 @@ describe('Execute', () => {
       .set('Authorization', 'Bearer ' + accessToken1)
       .send({});
     expect(res6.status).toBe(200);
-    const selfToken = parseJwt(res6.body.token);
+    const selfToken = parseJWTPayload(res6.body.token);
     expect(selfToken.profile).toMatch(/^ClientApplication\//);
 
     // Execute the bot with ProjectMembership ID
@@ -548,7 +548,7 @@ describe('Execute', () => {
       .set('X-Medplum-On-Behalf-Of', getReferenceString(membership))
       .send({});
     expect(res7.status).toBe(200);
-    const membershipToken = parseJwt(res7.body.token);
+    const membershipToken = parseJWTPayload(res7.body.token);
     expect(membershipToken.profile).toEqual(getReferenceString(profile));
 
     // Execute the bot with profile resource ID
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/operations/execute.test.ts
```diff
@@ -768,3 +844,19 @@ describe('Execute', () => {
     });
   });
 });
+
+function parseJwt(token: string): Record<string, string> {
```
ThatOneBro, line 848: Surprised we don't have a util for this in core. Might be a good followup?

packages/server/src/fhir/operations/execute.test.ts
```diff
@@ -491,6 +493,80 @@ describe('Execute', () => {
...
+    const bot = res1.body as Bot;
+
+    // Deploy the bot
+    const res5 = await request(app)
+      .post(`/fhir/R4/Bot/${bot.id}/$deploy`)
+      .set('Content-Type', ContentType.FHIR_JSON)
+      .set('Authorization', 'Bearer ' + accessToken1)
+      .send({
+        code: `
+          const { getReferenceString } = require("@medplum/core");
```
ThatOneBro, line 524: Do we need this line?

packages/server/src/fhir/operations/execute.test.ts
```diff
@@ -768,3 +844,19 @@ describe('Execute', () => {
     });
   });
 });
+
+function parseJwt(token: string): Record<string, string> {
+  const base64Url = token.split('.')[1];
+  const base64 = base64Url.replace(/-/g, '+').replace(/_/g, '/');
+  const jsonPayload = decodeURIComponent(
+    global
```
ThatOneBro, line 852: `globalThis` I believe is preferred now

ThatOneBro (approved):
LGTM, some small nits

packages/server/src/fhir/operations/execute.test.ts
codyebberson: We do have a util for this in core:  `parseJWTPayload` in `packages/core/src/jwt.ts`

codyebberson (changes requested):

packages/server/src/fhir/operations/execute.test.ts
codyebberson: I apologize for inconsistent capitalization... That function is from early 2021 before we got OCD about such things.

ThatOneBro: Ah yeah I thought that was a thing

codyebberson (approved):

PR #5964, opened 2025-02-13
Scroll to top of viewport on route change in Medplum web ui

sam-silv:
Running into a bit of pain, I've been spending a lot of time flicking between resources in the web ui to debug various things and finding myself doing a lot of scrolling.

When navigating between a sufficiently long `Patient` (or any other entity) and an entity of the same type, the scroll position is maintained - this PR fixes this behavior. Usually I hit it when searching, but also happens when manually throwing a reference in the url.

Let me know if you'd rather this be a hook.


codyebberson:
Continuing here https://github.com/medplum/medplum/pull/9358

Moving into the `medplum` repo grants full access to all privileged actions and automations

Thanks @sam-silv

PR #5971, opened 2025-02-13
Explicitly support SMART v1 scopes

mattwiller:


codyebberson (approved):
lgtm 👍 

Just to confirm... this should silently ignore an invalid scope, such as `patient/*.search`, correct?

mattwiller:
Correct, [scopes not matching the regex are ignored](https://github.com/medplum/medplum/pull/5971/files#diff-224c505fc594bc905f555b9af8a05f2e1916f9b81871a65257e21b4018099887R118-R127)

PR #5973, opened 2025-02-13
Log requests made on behalf of another user

mattwiller:
When a request is made with the `X-Medplum-On-Behalf-Of` header, ensure that server logs accurately reflect which actors are involved

### DIFF
--- a/packages/server/src/context.ts
+++ b/packages/server/src/context.ts
@@ -63,10 +63,6 @@ export class AuthenticatedRequestContext extends RequestContext {
     return this.membership.profile;
   }
 
-  get realProfile(): Reference<ProfileResource | Bot | ClientApplication> {
-    return this.authState.membership.profile;
-  }
-
   [Symbol.dispose](): void {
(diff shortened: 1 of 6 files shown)


packages/server/src/context.ts
```diff
@@ -59,8 +59,12 @@ export class AuthenticatedRequestContext extends RequestContext {
     return this.authState.login;
   }
 
-  get profile(): Reference<ProfileResource> {
-    return this.membership.profile as Reference<ProfileResource>;
+  get profile(): Reference<ProfileResource | Bot | ClientApplication> {
+    return this.membership.profile;
+  }
+
+  get realProfile(): Reference<ProfileResource | Bot | ClientApplication> {
```
codyebberson, line 66: I don't love this name, especially without comments or documentation.

How would you feel about removing this getter, and just using `.authState.membership.profile` at the call site?

The convention being:  if you know what you're doing, and you know what you need, then you should use `.authState` directly.

codyebberson (approved):

PR #5987, opened 2025-02-16
Fix server config settings circular dependencies

codyebberson:
Chipping away at removing all circular dependencies identified by `npm run circular`

This should be a total no-op, other than removing circular dependencies.  Most of the diff is updating `import` statements.

Before:
 * `config.ts` included 3 things: type definitions, loader functions, and util functions

After:
 * `config/types.ts` - type definitions
 * `config/utils.ts` - util functions (default values, type utils, etc)
 * `config/loader.ts` - loader functions (load, get, etc)


ThatOneBro (approved):
LGTM, I got a finger workout clicking through all the files that changed to make sure it was just the import path changing 🙃

PR #5992, opened 2025-02-17
fix: update package.json

avasisht23:
Without this package, I was getting:
```
[Error] Cannot find package 'postcss-simple-vars' imported from /Users/ajayvasisht/Desktop/apps/medplum-hello-world/postcss.config.mjs
```

### DIFF
--- a/examples/medplum-hello-world/package.json
+++ b/examples/medplum-hello-world/package.json
@@ -34,6 +34,7 @@
     "@vitejs/plugin-react": "4.3.4",
     "postcss": "8.5.1",
     "postcss-preset-mantine": "1.17.0",
+    "postcss-simple-vars": "^7.0.1",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-router-dom": "6.28.0",
--- a/examples/medplum-hello-world/package.json
+++ b/examples/medplum-hello-world/package.json
@@ -34,7 +34,7 @@
     "@vitejs/plugin-react": "4.3.4",
     "postcss": "8.5.1",
     "postcss-preset-mantine": "1.17.0",
-    "postcss-simple-vars": "^7.0.1",
+    "postcss-simple-vars": "7.0.1",
     "react": "18.2.0",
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Hi @avasisht23, I'm not able to repro what you're seeing. Are you sure you are on latest and have run `npm install` before attempting to build or run `npm run dev`?
Screenshot 2025-02-17 at 11 31 02 AM

avasisht23:
> Hi @avasisht23, I'm not able to repro what you're seeing. Are you sure you are on latest and have run `npm install` before attempting to build or run `npm run dev`? Screenshot 2025-02-17 at 11 31 02 AM

Hi! Yes, I'm on the latest and was using `pnpm install` then `pnpm dev`.

ThatOneBro:
Ah ok this is valid, I thought this was on `pnpm` but it's actually kind of doing the right thing and `npm` is just allowing us to be sloppier. Thank you for finding this!

examples/medplum-hello-world/package.json
```diff
@@ -34,6 +34,7 @@
     "@vitejs/plugin-react": "4.3.4",
     "postcss": "8.5.1",
     "postcss-preset-mantine": "1.17.0",
+    "postcss-simple-vars": "^7.0.1",
```
ThatOneBro, line 37:
```suggestion
    "postcss-simple-vars": "7.0.1",
```

ThatOneBro: We always use explicit versions, so removing the caret here

ThatOneBro (approved):

codyebberson (changes requested):
Any change to any `package.json` requires updating `package-lock.json`, otherwise `npm ci` won't work, because the files are out of sync.

@avasisht23 - do you want to update `package-lock.json`?  Or let us know, and we can handle that too. 

Thanks for your contribution!

avasisht23:
Sorry I didn't get to this! Thanks for resolving :)

PR #6010, opened 2025-02-19
feat(server): data migration rework

ThatOneBro:
This reverts commit 8eff076cd54b38db4e3df63aeff38a38c33e0123.

Retry of #5820 after having to previously revert to cut a hotfix before releasing `3.3.0`

### DIFF
--- a/packages/server/src/database.ts
+++ b/packages/server/src/database.ts
@@ -317,42 +317,45 @@ export async function maybeStartDataMigration(assertedDataVersion?: number): Pro
   const systemRepo = getSystemRepo();
   let dataMigrationJob: AsyncJob | undefined;
 
-  await systemRepo.withTransaction(async () => {
-    // Check if there is already a migration job in progress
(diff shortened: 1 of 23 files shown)


packages/server/src/database.ts
```diff
@@ -167,3 +255,107 @@ async function migrate(client: PoolClient): Promise<void> {
...
+    return undefined;
+  }
+
+  const systemRepo = getSystemRepo();
+  let dataMigrationJob: AsyncJob | undefined;
+
+  await systemRepo.withTransaction(
+    async () => {
+      // Check if there is already a migration job in progress
+      const existingJob = await systemRepo.searchOne<AsyncJob>(
```
ThatOneBro, line 323: Need to use `searchResources` with count 2 and check if jobs > 1 too

mattwiller (approved):

codyebberson (approved):

PR #6011, opened 2025-02-19
Update SMART launch context on subsequent authorization

mattwiller:
Follow up to #6005 to add more `Login` parameters that should be updated

### DIFF
--- a/packages/server/src/oauth/authorize.ts
+++ b/packages/server/src/oauth/authorize.ts
@@ -64,52 +64,38 @@ async function validateAuthorizeRequest(req: Request, res: Response, params: Rec
     res.status(400).send('Client has no redirect URI');
     return false;
   }
-
   if (!URL.canParse(redirectUri)) {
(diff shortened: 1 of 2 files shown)


packages/server/src/oauth/authorize.ts
```diff
@@ -128,6 +114,7 @@ async function validateAuthorizeRequest(req: Request, res: Response, params: Rec
       codeChallenge: params.code_challenge ?? existingLogin.codeChallenge,
       codeChallengeMethod: params.code_challenge_method ?? existingLogin.codeChallengeMethod,
       code: generateSecret(16),
+      launch: params.launch ? { reference: `SmartAppLaunch/${params.launch}` } : existingLogin.launch,
```
codyebberson, line 117: Unfortunately I think we need to validate that `params.launch` is actually a valid ID of a `SmartAppLaunch` resource

codyebberson (approved):

PR #6012, opened 2025-02-19
Patient $set-accounts operation

### LINKED ISSUE #1928: $set-accounts fhir operation
Right now, if you update `Patient.meta.accounts`, it does not automatically update the `accounts` element for all resources in the compartment. However, any time another resource in the compartment is updated, the `accounts` field is updated at that time. 

This issue proposes a new  operation, `Patient/[id]/$set-accounts`, which would update the `accounts` for _all_ resources within a compartment. Because this could potentially take some time, this should be considered an asynchronous background job.

finnbergquist:
resolves #1928 

- new custom FHIR operation $set-accounts for Patient resources that propagates account associations to all resources in the patient's compartment

- tests verifying the operation updates related resources (ex. Observations) with the correct account references from the Patient

- documentation for the $set-accounts operation

- operation endpoint at /fhir/R4/Patient/:id/$set-accounts in the FHIR router


packages/server/src/fhir/operations/patientupdateaccounts.ts
```diff
@@ -0,0 +1,53 @@
...
+    const patient = await ctx.repo.readResource<Patient>('Patient', id);
+    const bundle = await getPatientEverything(ctx.repo, patient);
+
+    // Update each resource to trigger meta.accounts refresh
+    let count = 0;
+    for (const entry of bundle.entry ?? []) {
+      if (entry.search?.mode === 'match') {
+        const resource = entry.resource;
+        if (resource && resource.id !== patient.id) {
+          if (patient.meta?.accounts !== resource.meta?.accounts) {
```
finnbergquist, line 34: Any cases here where someone might want a resource within the compartment of a Patient to persist it's existing _meta.accounts_?

packages/server/src/fhir/operations/patientupdateaccounts.ts
```diff
@@ -0,0 +1,53 @@
...
+import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
+import { Patient } from '@medplum/fhirtypes';
+import { getAuthenticatedContext } from '../../context';
+import { getPatientEverything } from './patienteverything';
+
+/**
+ * Handles the Patient $update-account operation.
+ * This operation updates the account reference for all resources in the patient compartment.
+ * The operation only updates resources from the getPatientEverything bundle that are not
+ * included in the patient's compartment, but are linked to the patient through a reference.
```
mattwiller, line 11: I'm not quite sure what this means: the code looks like it __does__ operate on the resources from the Patient compartment

packages/server/src/fhir/operations/patientupdateaccounts.ts
```diff
@@ -0,0 +1,53 @@
...
+            delete resource.meta?.accounts;
+            delete resource.meta?.account;
+          }
+          await ctx.repo.updateResource(resource);
+          count++;
+        }
+      }
+    }
+
+    return [allOk, { resourceType: 'Parameters', parameter: [{ name: 'resourcesUpdated', valueInteger: count }] }];
```
mattwiller, line 45: We should draft an `OperationDefinition` ([example](https://github.com/medplum/medplum/blob/7fe3a24ab552c849caa258c2ab8bd9bcd0d66e4a/packages/server/src/fhir/operations/codesystemimport.ts#L10-L37)) to codify the expected input and output parameters, and then use the `buildOutputParameters()` helper to format and validate the output

packages/server/src/fhir/operations/patientupdateaccounts.ts
```diff
@@ -0,0 +1,53 @@
...
+  } catch (error) {
+    return [
+      badRequest(
```
mattwiller, line 48: I don't think we want to swallow and convert all errors to a 400 here: we should probably just let errors get thrown from this function and be handled as normal

packages/server/src/fhir/operations/patientupdateaccounts.ts
```diff
@@ -0,0 +1,53 @@
...
+  try {
+    const patient = await ctx.repo.readResource<Patient>('Patient', id);
+    const bundle = await getPatientEverything(ctx.repo, patient);
+
+    // Update each resource to trigger meta.accounts refresh
+    let count = 0;
+    for (const entry of bundle.entry ?? []) {
+      if (entry.search?.mode === 'match') {
+        const resource = entry.resource;
+        if (resource && resource.id !== patient.id) {
```
mattwiller, line 33: This could be more strict:
```suggestion
        if (resource && (resource.resourceType !== 'Patient' || resource.id !== patient.id)) {
```

packages/server/src/fhir/routes.ts
```diff
@@ -256,6 +257,9 @@ function initInternalFhirRouter(): FhirRouter {
   router.add('GET', '/Patient/:id/$summary', patientSummaryHandler);
   router.add('POST', '/Patient/:id/$summary', patientSummaryHandler);
 
+  // Patient $update-account operation
+  router.add('POST', '/Patient/:id/$update-accounts', patientUpdateAccountHandler);
```
mattwiller, line 261: There's inconsistent use in the docs between `$update-account` and `$update-accounts`: should standardize presumably on the second one.  Also this name makes it sound like the Patient is being updated; I'd suggest perhaps e.g. `$propagate-accounts` for clarity

packages/docs/docs/api/fhir/operations/patient-update-account.md
```diff
@@ -0,0 +1,57 @@
+---
+sidebar_position: 5.1
+---
+
+# Patient $update-account Operation
+
+Medplum implements a custom `$update-account` operation for Patient resources to manage account associations.
+
+This operation updates the account references for all resources in the patient compartment to match the patient's accounts. This is useful when you need to ensure consistent account access across all resources related to a patient.
```
mattwiller, line 9: We may wish to note that "all" isn't totally accurate here, and only the first 1,000 resources in the compartment will be updated

packages/server/src/fhir/operations/patientsetaccounts.ts
```diff
@@ -0,0 +1,99 @@
...
+  const bundle = await getPatientEverything(ctx.repo, patient);
+  const accounts = params.accounts ?? [];
+
+  // step 1: update the patient resource with the new accounts
+  patient.meta = {
+    ...patient.meta,
+    accounts: accounts,
+    account: accounts[0] ?? undefined,
+  };
+
```
finnbergquist: Looking for feedback here on how to let users remove all the references in a patient's accounts. Passing a blank array or deleting meta.accounts and then updating doesn't remove accounts for the patient.

packages/server/src/fhir/operations/patientsetaccounts.ts
```diff
@@ -0,0 +1,99 @@
...
+  await ctx.repo.updateResource(patient);
+
+  // step 2: update the resources in the patient compartment to trigger meta.accounts refresh
+  let count = 0;
```
mattwiller, line 79: Should this be 1, since the Patient was already updated?

packages/server/src/fhir/operations/patientsetaccounts.ts
```diff
@@ -0,0 +1,99 @@
...
+  const patient = await ctx.repo.readResource<Patient>('Patient', id);
+  const bundle = await getPatientEverything(ctx.repo, patient);
+  const accounts = params.accounts ?? [];
+
+  // step 1: update the patient resource with the new accounts
+  patient.meta = {
+    ...patient.meta,
+    accounts: accounts,
+    account: accounts[0] ?? undefined,
+  };
```
mattwiller, line 74: Empty array isn't a valid value in FHIR, so this should probably be:
```suggestion
  const accounts = params.accounts;

  // step 1: update the patient resource with the new accounts
  patient.meta = {
    ...patient.meta,
    accounts: accounts,
    account: accounts?.[0],
  };
```

packages/server/src/fhir/operations/patientsetaccounts.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export async function patientSetAccountsHandler(req: FhirRequest): Promise<FhirResponse> {
+  const { id } = req.params;
+  if (!id) {
+    return [badRequest('Must specify Patient ID')];
+  }
+
+  const params = parseInputParameters<PatientSetAccountsParameters>(operation, req);
+
+  const ctx = getAuthenticatedContext();
```
mattwiller, line 64: We should probably check that the current user is a Project Admin; if not this operation should probably return `[forbidden]`

packages/docs/docs/api/fhir/operations/patient-set-accounts.md
```diff
@@ -0,0 +1,74 @@
+---
+sidebar_position: 5.1
+---
+
+# Patient $set-accounts Operation
+
+Medplum implements a custom `$set-accounts` operation for Patient resources to manage account references. __This is the recommended way to manage account references for patients__.
+
+This operation sets the Patient's _meta.accounts_ references and propogates them to the resources in that patient's compartment. This is useful when you need to ensure consistent accounts access across all resources related to a patient.
```
mattwiller, line 9: I'd recommend using \`meta.accounts\` consistently, since this is a direct code reference

packages/server/src/fhir/operations/patientsetaccounts.test.ts
```diff
@@ -0,0 +1,176 @@
...
+      params: { id: '' },
+      method: 'POST',
+      url: '/fhir/R4/Patient/$set-accounts',
+      pathname: '/fhir/R4/Patient/$set-accounts',
+      body: {},
+      query: {},
+    });
+    expect(res[0].issue?.[0]?.details?.text).toBe('Must specify Patient ID');
+  });
+});
```
mattwiller, line 176: Could you also add a test showing that resources returned by `Patient/$everything` but not in the patient's compartment, or other Patients in the compartment, are not modified?

packages/server/src/fhir/operations/patientsetaccounts.ts
finnbergquist: Yea, that definitely makes more sense

packages/server/src/fhir/operations/patientsetaccounts.test.ts
finnbergquist: done

packages/docs/docs/api/fhir/operations/patient-set-accounts.md
finnbergquist: Yea, good call

packages/server/src/fhir/operations/patientsetaccounts.test.ts
```diff
@@ -146,6 +146,19 @@ describe('Patient Set Accounts Operation', () => {
...
+  test("Make sure resources returned in $patient-everything that are NOT in the patient's compartment are not updated", async () => {
+    const res = await request(app)
+      .get(`/fhir/R4/Patient/${patient.id}/$everything`)
+      .set('Authorization', 'Bearer ' + accessToken);
+    expect(res.status).toBe(200);
+    const everything = res.body as Bundle;
+    const resourcesNotInCompartment = everything.entry?.filter((entry) => entry?.search?.mode !== 'match');
+    expect(resourcesNotInCompartment?.length).toBeGreaterThan(0);
+    resourcesNotInCompartment?.forEach((entry) => {
+      expect(entry?.resource?.meta?.accounts).toBeUndefined();
```
mattwiller, line 158: This is a little tricky: since the request above didn't set the `X-Medplum: extended` header, I think `meta.accounts` will always be undefined — even for `match` resources

finnbergquist: @mattwiller I addressed this in a new test now added. It uses the _resourcesUpdated_ output field, which keeps track of how many updates. The test confirms that the number of resources returned by $set-accounts in _resourcesUpdated_ is the number of resources returned by $patient-everything minus the resources in $patient-everything that are not in patient compartment. In other words, it  verifies that only resources that are actually in the patient's compartment were updated.

mattwiller (approved):

PR #6014, opened 2025-02-20
feat(AsyncJob): add `AsyncJob/$complete` operation

ThatOneBro:
This allows us mainly to complete a `data-migration`-type `AsyncJob` via `AsyncJobExecutor.completeJob`. This triggers the post-complete logic for the `data-migration` job type which updates the `DatabaseMigration` table's `dataVersion` to the `dataVersion` in the `AsyncJob`.

We will be using this specifically to skip a data migration we know we ourselves do not need to apply but most of self-hosted users probably need to or should apply.

Depends on #6010

### DIFF
--- a/packages/server/src/fhir/operations/asyncjobcomplete.ts
+++ b/packages/server/src/fhir/operations/asyncjobcomplete.ts
@@ -1,0 +1,44 @@
+import { allOk, assert, badRequest } from '@medplum/core';
+import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
+import { AsyncJob, OperationDefinition } from '@medplum/fhirtypes';
+import { requireSuperAdmin } from '../../admin/super';
+import { getAuthenticatedContext } from '../../context';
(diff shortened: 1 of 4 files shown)


packages/server/src/fhir/operations/asyncjobcomplete.ts
```diff
@@ -0,0 +1,44 @@
...
+  name: 'asyncjob-complete',
+  status: 'active',
+  kind: 'operation',
+  code: 'complete',
+  experimental: true,
+  resource: ['AsyncJob'],
+  system: false,
+  type: false,
+  instance: true,
+  parameter: [{ use: 'out', name: 'return', type: 'OperationOutcome', min: 1, max: '1' }],
```
mattwiller, line 21: I think this should return the updated `AsyncJob`

codyebberson (approved):
clean 👍

ThatOneBro:
Closed in favor of #6024

PR #6021, opened 2025-02-20
Fixes #5844 - stringify removes nested empty values

### LINKED ISSUE #5844: Medplum FHIR `stringify` does not recursively strip out empty values in certain scenarios
In certain scenarios, `stringify` does not correctly strip out an empty object

Take for example a `Patient` resource with the following form:
```tsx
{
  "resourceType": "Patient",
  "name": [
    { "text": "" }
  ]
}
```

The text property is an empty string, which should be stripped, which leaves the name object as an empty object which should be stripped, which leaves name as an empty array which should also be stripped, so that the resulting resource should look like:
```tsx
{
  "resourceType": "Patient"
}
```

However, Medplum's `stringify` currently yields:
```tsx
{
  "resourceType": "Patient",
  "name": [{ }]
}
```

This is because of how the property replacer evaluates `isEmpty` on the way down the JSON syntax tree, but not on the way back up, so because of the `isEmpty` evaluation changing at the bottom of the tree, we end up with an invalid FHIR resource with an empty but defined name property

codyebberson:
### Performance results

Methodology:

- 1,000,000 iterations of JSON.stringify on an object with a nested array.
- Run 3 times
- Take the best of 3

| Method                            | Time   | Relative to JSON.stringify | Relative to old stringify |
| --------------------------------- | ------ | -------------------------- | ------------------------- |
| JSON.stringify                    | 275 ms | 1.0x                       | 0.3x                      |
| Old stringify                     | 850 ms | 3.1x                       | 1.0x                      |
| New stringify                     | 597 ms | 2.2x                       | 0.7x                      |
| New stringify with ownKey guard   | 716 ms | 2.6x                       | 0.8x                      |
| New stringify with Object.entries | 910 ms | 3.3x                       | 1.1x                      |

### DIFF
--- a/packages/ccda/src/roundtrip.test.ts
+++ b/packages/ccda/src/roundtrip.test.ts
@@ -60,21 +60,7 @@ describe('convertFhirToCcda', () => {
 function normalizeFhir<T extends Resource>(resource: T): T {
   // We need to remove all "empty" elements, such as empty strings and empty objects
   // This is because FHIR is very strict about what is allowed in the JSON
-  // Medplum "stringify" function does this, althought it is not perfect
-  // So, we need to use a hack and iteratively do this until we get no more changes
-  let currResource = resource;
-  let currString = stringify(currResource);
-
-  for (let i = 0; i < 100; i++) {
-    currResource = JSON.parse(stringify(currResource)) as T;
-    const nextString = stringify(currResource);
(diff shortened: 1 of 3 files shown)


codyebberson:
Oo, this is a sneaky edge case.  Glad we caught it, even if indirectly.

image

PR #6024, opened 2025-02-20
feat(super): add `setdataversion` superadmin route

ThatOneBro:
This route will allow us to set the data version of the database without having to run a migration we know we don't need to run.

### DIFF
--- a/packages/server/src/admin/super.ts
+++ b/packages/server/src/admin/super.ts
@@ -10,6 +10,7 @@ import {
   validateResourceType,
 } from '@medplum/core';
 import { ResourceType } from '@medplum/fhirtypes';
+import { assert } from 'console';
 import { Request, Response, Router } from 'express';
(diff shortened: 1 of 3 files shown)


codyebberson (approved):

packages/server/src/admin/super.ts
```diff
@@ -239,6 +240,31 @@ superAdminRouter.post(
...
+    requireSuperAdmin();
+
+    const errors = validationResult(req);
+    if (!errors.isEmpty()) {
+      sendOutcome(res, invalidRequest(errors));
+      return;
+    }
+
+    assert(req.body.dataVersion !== undefined);
+    await getDatabasePool(DatabaseMode.WRITER).query('UPDATE "DatabaseMigration" SET "dataVersion" = $1', [
```
mattwiller, line 260: Do we also want to create the corresponding `AsyncJob` for record keeping?

PR #6022, opened 2025-02-20
QuestionnaireForm support for QuestionnaireResponse

techdavidy:
Allow QuestionnaireForm to prepopulate fields/anwsers using a QuestionnaireResponse.

PR #6028, opened 2025-02-21
Release Version 4.0.0

### LINKED ISSUE #5844: Medplum FHIR `stringify` does not recursively strip out empty values in certain scenarios

### LINKED ISSUE #5997: Add explicit dependency to all examples using `postcss-simple-vars` in `postcss.config.mjs`
In light of #5992, looks like we have a lot of examples with this same issue. Just need to add the dependency to all of their `package.json` and run `npm install` in the root to make sure `package-lock.json` is updated

### LINKED ISSUE #6008: Change TURBO_REMOTE_ONLY to TURBO_CACHE
WARNING  TURBO_REMOTE_ONLY is deprecated and will be removed in a future major version. Use TURBO_CACHE=remote:rw

medplumbot:
## What's Changed

* Remove terminology feature flag (#4995)
* Remove deprecated functionality for v4 (#5572)
* Update server README for v4 (#5651)
* Update build and test matrices for v4 (#5674)
* _filter eq should result in exact matching (#5631)
* Don't allow control chars in strings (#5630)
* Remove legacy chained search implementation (#5493)
* Return undefined reference string when resource has no ID (#5625)
* Update docs to deprecate databaseProxyEndpoint (#5675)
* Added docs to C-CDA readme (#6017)
* Fixes #5997 - add missing postcss dep to nextauth-demo (#6019)
* Fixes #6008 - replace TURBO_REMOTE_ONLY with TURBO_CACHE (#6018)
* feat(server): data migration rework (#6010)
* QuestionnaireForm support for QuestionnaireResponse (#6022)
* Fixes #5844 - stringify removes nested empty values (#6021)
* feat(super): add `setdataversion` superadmin route (#6024)

**Full Changelog**: https://github.com/medplum/medplum/compare/v3.2.33...v4.0.0


ThatOneBro:
We'll prepare another release off of main after we merge #6013 just to be safe

PR #6033, opened 2025-02-22
Add `__version` column to all resources

### LINKED ISSUE #6002: Adding a version column to resources
## Background

To fully take advantage of almost all data migrations, certain schema migrations, and various other server logic changes (e.g. a new SearchParameter), it is necessary to "reindex" all resources on a Medplum server deployment. Reindexing is the process of iterating over all non-deleted resources to recompute search parameter related columns, replacing lookup table entries, etc. to ensure they have the most up to date representation in the database based on the latest Medplum code version, defined SearchParameters, etc.

Reindexing all resource types on a large Medplum deployment can take a very long time. In some scenarios, the server change that a super admin wants to ensure all resources reflect could have been already been live for quite a while (e.g. populating the `projectId` column) meaning that a decent percentage of resources already reflect the change. In these cases, it is wasteful and needlessly time consuming to reindex all resources when only a small fraction of them really needed to be reindexed.

Resource rows do not currently include a definitive way to determine the state of the server when they were written. The closest analog is the `lastUpdated` column with the big caveat that `lastUpdated` has FHIR semantics and can be manually set to any timestamp by a super admin. In other words, it is not a reliable, server-controlled "when was this record saved to the database?" timestamp.

## Adding a version column

To more reliably know which rows reflect which state of the server, a new `_version` column (actual column name TBD) is added to all resource tables. During all resource creates and updates, it is set to a constant integer value `RESOURCE_VERSION` (this is just a placeholder name) included in Medplum server code. This constant value will be incremented as necessary when code changes include changes in how a resource is represented in the database; generally all schema migrations as well as bug fixes or improvements in search parameter implementations/representations.

With the `_version` column, Medplum could support a few variants of reindexing:

- **minimum version reindex**: Reindex resources with `_version` older than a `minimumVersion` parameter specified when a reindex job is created. The most selective variant; used when some specific feature/functionality is desired ASAP.
- **current version reindex**: Reindex resources with `_version` less than `RESOURCE_VERSION`. The standard variant of reindexing to get all resources shored up with the latest available functionality.
- **full reindex**: Reindex all resources; ignores current value of `_version`. Equivalent to how reindexing works today. This becomes the "break glass" variant of reindexing when something is suspected to have gone wrong that should only be needed sparingly.

### Potential extension

Distilling the version of a resource down to a single integer as described above is a huge improvement over having nothing, but perhaps that is too simplistic. An alternative idea discussed by the team is introducing the concept of a "server deployment". Each time a server starts, an entry is added to the new server deployment table that captures more information about the state of the server:

- `RESOURCE_VERSION` as described above
- A timestamp of when the server started
- The Medplum server version, e.g. `3.2.32`
- Additional server metadata TBD?

Now, instead of (or in addition to since denormalizing the resource version would likely be desirable) each resource having a `_version` column, it has a `_deployment` column referencing the server deployment that wrote the resource. This additional information would be a boon to the debugability of Medplum and allow for more fine grained operations in the future.

mattlong:
Fixes #6002


codyebberson:
@mattlong  - do we know how long this will take on tables with 100m+ resources?  Any concerns about clogging up deployment pipeline, or will this be ok?

mattlong:
> @mattlong - do we know how long this will take on tables with 100m+ resources? Any concerns about clogging up deployment pipeline, or will this be ok?

@codyebberson Since it's a nullable column, the migration happens nearly instantaneously.

packages/server/src/workers/reindex.ts
```diff
@@ -310,7 +311,8 @@ async function addReindexJobData(job: ReindexJobData): Promise<Job<ReindexJobDat
 export async function addReindexJob(
   resourceTypes: ResourceType[],
   job: AsyncJob,
-  searchFilter?: SearchRequest
+  searchFilter?: SearchRequest,
+  maxResourceVersion?: number
```
mattwiller, line 315: How should this be used by the job?

mattlong: Good catch! 😅  Updated.

mattwiller (approved):

PR #6037, opened 2025-02-22
Fixed sonar warnings in agent dockerfile

codyebberson:


packages/agent/Dockerfile
```diff
@@ -7,18 +7,17 @@ ENV GIT_SHA ${GIT_SHA}
 ENV MEDPLUM_VERSION ${MEDPLUM_VERSION}
 ENV MEDPLUM_LOG_LEVEL=${MEDPLUM_LOG_LEVEL:-"INFO"}
 
+COPY bin/medplum-agent-${MEDPLUM_VERSION}-linux /srv/medplum-agent
```
ThatOneBro, line 10: hm I don't think this path is valid anymore for our current agent builds?

This might be an artifact of when we were still using `pkg` instead of `node sea`

codyebberson: Mmm... interesting.  This may have been broken for a while...

Good catch, will investigate more.

ThatOneBro: Yeah after building the linux agent, I got the binary in `packages/agent` itself 🤔 

We could fix that and just set all binaries to go into `bin` but we probably need to update that in a few places

ThatOneBro: Yeah that's the more direct fix, was just saying might be cleaner to put them in `bin`, I'm good with either

ThatOneBro (approved):

PR #6038, opened 2025-02-22
Fixed a bunch of sonar issues

codyebberson:
Mostly small minor stuff:

* Add `alt` prop to `` tags
* Add `title` prop to `<iframe>` tags
* Add `readonly` whenever possible
* Remove unused props
* Remove unnecessary casts
* Prefer `??` over `||`

### DIFF
--- a/packages/health-gorilla-core/src/aoe.ts
+++ b/packages/health-gorilla-core/src/aoe.ts
@@ -62,6 +62,7 @@ export function getMissingRequiredQuestionnaireItems(
             continue;
           }
         }
+
         if (hasValue) {
           continue;
         }
--- a/packages/health-gorilla-core/src/aoe.ts
+++ b/packages/health-gorilla-core/src/aoe.ts
@@ -52,13 +52,18 @@ export function getMissingRequiredQuestionnaireItems(
     ) {
       const answers = answersByLinkId?.[item.linkId];
       if (answers && answers.length > 0) {
+        let hasValue = false;
         for (const answer of answers) {
           const values = arrayify(
             getTypedPropertyValue({ type: 'QuestionnaireResponseItemAnswer', value: answer }, 'value[x]')
           );
           if (values?.some((v) => v.value)) {
+            hasValue = true;
             continue;
           }
+        }
+        if (hasValue) {
+          continue;
         }
       }
 
--- a/packages/health-gorilla-core/src/aoe.ts
+++ b/packages/health-gorilla-core/src/aoe.ts
@@ -35,7 +35,6 @@ export function getMissingRequiredQuestionnaireItems(
   qr: QuestionnaireResponse | undefined,
   includeRequiredWhenSpecimenCollected: boolean
 ): string[] {
-  const required: string[] = [];
   const missing: string[] = [];
   const answersByLinkId: Record<string, QuestionnaireResponseItemAnswer[] | undefined> | undefined =
     qr && getAllQuestionnaireAnswers(qr);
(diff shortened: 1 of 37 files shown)


packages/health-gorilla-core/src/aoe.ts
```diff
@@ -51,24 +50,16 @@ export function getMissingRequiredQuestionnaireItems(
...
           const values = arrayify(
             getTypedPropertyValue({ type: 'QuestionnaireResponseItemAnswer', value: answer }, 'value[x]')
           );
           if (values?.some((v) => v.value)) {
-            hasValue = true;
             continue;
           }
         }
-
-        if (hasValue) {
```
ThatOneBro, line 69: This is changing behavior no?

packages/health-gorilla-core/src/aoe.ts
```diff
@@ -51,24 +50,16 @@ export function getMissingRequiredQuestionnaireItems(
...
       const answers = answersByLinkId?.[item.linkId];
       if (answers && answers.length > 0) {
-        let hasValue = false;
         for (const answer of answers) {
           const values = arrayify(
             getTypedPropertyValue({ type: 'QuestionnaireResponseItemAnswer', value: answer }, 'value[x]')
           );
           if (values?.some((v) => v.value)) {
-            hasValue = true;
             continue;
```
ThatOneBro, line 60: This is the unnecessary `continue` according to Biome:
Screenshot 2025-02-22 at 3 30 58 PM

ThatOneBro (commented):
Things look good for the most part, one comment about changing behavior

packages/health-gorilla-core/src/aoe.ts
ThatOneBro: The earlier `continue` continues the inner loop, the `hasValue` is trying to detect if there is a value in the inner loop and the check later skips over adding the current outer iteration to the `missing` array

packages/health-gorilla-core/src/lab-order.ts
```diff
@@ -264,7 +264,7 @@ export function createLabOrderBundle(inputs: PartialLabOrderInputs): Bundle {
       performer: [performerReference],
       category: [{ coding: [{ system: SNOMED, code: '103693007', display: 'Diagnostic procedure' }] }],
       code: { coding: [test] },
-      priority: metadata?.priority || 'routine',
+      priority: metadata?.priority ?? 'routine',
```
ThatOneBro, line 267: This also changes behavior if `priority` is an empty string. Probably not a big deal for typescript, but does allow users to pass empty string and it will not assign default value of `routine`

packages/health-gorilla-core/src/aoe.ts
codyebberson: Hm.  Yes, you're right.  Sonar didn't think so, but you're right.

I will revert.

packages/health-gorilla-core/src/lab-order.ts
codyebberson: Eh....  That comes from the `testMetadata` lookup table, so someone would have had to construct a full table and specifically set `priority` to empty string.

We can revert.  I think Sonar is too aggressive with this rule, and it does lead to pushing us to less desirable behavior.  I wish it didn't ding us for that.  I think this case is probably fine though.

ThatOneBro: yeah now that I look at this you would have to be in some pretty weird territory

ThatOneBro (approved):

PR #6042, opened 2025-02-23
Remove extra jwtVerify call

codyebberson:


packages/server/src/oauth/token.ts
```diff
@@ -607,8 +607,7 @@ async function sendTokenResponse(res: Response, login: Login, client?: ClientApp
     fhircastProps['hub.topic'] = topic;
   }
 
-  const decodedAccessToken = await verifyJwt(tokens.accessToken);
```
codyebberson, line 610: This is a JWT that we ourselves just generated, so we do not need to re-verify it.

ThatOneBro (approved):

packages/core/src/jwt.ts
```diff
@@ -1,12 +1,63 @@
 import { decodeBase64 } from './base64';
 
+/**
+ * Recognized JWT Claims Set members, any other members may also be present.
+ * @see {@link https://github.com/panva/jose/blob/main/src/types.d.ts#L532}
+ */
+export interface JWTPayload {
```
ThatOneBro, line 7: We could add a generic here that takes an object and `&`s the base type and the generic type, so that we can properly type other claims and also prevent unknown props in certain scenarios

PR #6062, opened 2025-02-25
cleanup: use canonical ref to questionnaire in `QuestionnaireResponse`

ThatOneBro:
A lot of cleanup after fixing invalid values for `QuestionnaireResponse.questionnaire` in `QuestionnaireForm`. See: #5625 

It turns out we were doing this in a lot more places and we did not update those.

These include: 
* Some example apps
* Some tests
* `QuestionnaireBotsPage`
* `QuestionnaireResponsePage`

### DIFF
--- a/packages/app/src/resource/QuestionnaireBotsPage.tsx
+++ b/packages/app/src/resource/QuestionnaireBotsPage.tsx
@@ -1,6 +1,6 @@
 import { Button, Divider, Group, InputLabel, NativeSelect, Stack, Title } from '@mantine/core';
 import { showNotification } from '@mantine/notifications';
-import { getReferenceString, normalizeErrorString } from '@medplum/core';
+import { getReferenceString, normalizeErrorString, WithId } from '@medplum/core';
 import { Bot, Questionnaire, Subscription } from '@medplum/fhirtypes';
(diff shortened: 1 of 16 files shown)


packages/app/src/resource/QuestionnaireBotsPage.tsx
```diff
@@ -82,31 +83,47 @@ export function QuestionnaireBotsPage(): JSX.Element {
...
-            defaultValue="Create Only"
-            label="Subscription Trigger Event"
-            data={SUBSCRIPTION_INTERACTION_KEYS}
-            onChange={(event) =>
-              setSupportedInteraction(SUBSCRIPTION_INTERACTION_MAP[event.target.value as SubscriptionInteractionKey])
-            }
-          />
-        </Group>
+        {!questionnaire?.url ? (
+          <p>Cannot create new bot subscriptions until a canonical URL is added to the questionnaire.</p>
```
codyebberson, line 87: Do we want to be this assertive?  We could easily support the existing behavior.

I'm a little bit nervous about the added friction of requiring people to come up with `Questionnaire.url` strings for this.

codyebberson (approved):

PR #6076, opened 2025-02-26
Display OAuth2 scope consent screen on subsequent authorizations

mattwiller:
After authorizing a SMART app launch, if the app tries to re-authorize the user (e.g. to update the scopes requested) the server would redirect straight back to the application without giving the user another chance to consent to scopes.  This ensures that when `prompt=none` is not set, the user will have the opportunity to change the granted scopes on subsequent re-authorization

### DIFF
--- a/packages/server/src/fhir/routes.ts
+++ b/packages/server/src/fhir/routes.ts
@@ -339,7 +339,7 @@ protectedRoutes.use(
 
     const request: FhirRequest = {
       method: req.method as HttpMethod,
-      url: req.originalUrl.replace('/fhir/R4', ''),
+      url: stripPrefix(req.originalUrl, '/fhir/R4'),
       pathname: '',
       params: req.params,
       query: Object.create(null), // Defer query param parsing to router for consistency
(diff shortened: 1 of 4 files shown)


packages/server/src/fhir/routes.ts
```diff
@@ -339,7 +339,7 @@ protectedRoutes.use(
 
     const request: FhirRequest = {
       method: req.method as HttpMethod,
-      url: req.originalUrl.replace('/fhir/R4', ''),
+      url: stripPrefix(req.originalUrl, '/fhir/R4'),
```
codyebberson, line 342: Do you remember what input required this change?

codyebberson (approved):

packages/server/src/fhir/routes.ts
mattwiller: In my local testing environment with ngrok, I had the API running on a sub-path, so `req.originalUrl` was e.g. `/api/fhir/R4/$graphql`

PR #6078, opened 2025-02-27
Provider app - checkin, chart and overview screen

techdavidy:
Initial PR of multi step encounter screens.

Screenshot 2025-02-26 at 7 19 15 PM


examples/medplum-provider/src/App.tsx
```diff
@@ -165,3 +170,13 @@ export function App(): JSX.Element | null {
     </AppShell>
   );
 }
+
+function EncounterRedirect(): JSX.Element {
```
codyebberson, line 174: Can we use the built-in `<Navigate>` component from `react-router` ?

```tsx
    <Route
        path="*"
        element={<Navigate to="/" replace />}
    />
```

codyebberson (changes requested):
Nice, lgtm 👍 

One small request on the redirect implementation

codyebberson (approved):

PR #6090, opened 2025-02-28
Remove unnecessary IS NOT NULL

mattlong:
On its own, it wasn't a big deal, but when it was being negated, it is problematic since a NULL value passes negation when it shouldn't.

In particular, `old_negated` being `true` is problematic.
```
postgres=# SELECT
    (NULL IS NOT NULL AND NULL && ARRAY['1']::text[]) as old_normal,
    NOT(NULL IS NOT NULL AND NULL && ARRAY['1']::text[]) as old_negated,
    (NULL && ARRAY['1']::text[]) as new_normal,
    NOT(NULL && ARRAY['1']::text[]) as new_negated;
 old_normal | old_negated | new_normal | new_negated
------------+-------------+------------+-------------
 f          | t           |            |
```

For non-null values, everything still works as expected:

```
postgres=# SELECT
    ('{1}' IS NOT NULL AND '{1}' && ARRAY['1']::text[]) as old_normal,
    NOT('{1}' IS NOT NULL AND '{1}' && ARRAY['1']::text[]) as old_negated,
    ('{1}' && ARRAY['1']::text[]) as new_normal,
    NOT('{1}' && ARRAY['1']::text[]) as new_negated;
 old_normal | old_negated | new_normal | new_negated
------------+-------------+------------+-------------
 t          | f           | t          | f
```

```
postgres=# SELECT
    ('{2}' IS NOT NULL AND '{2}' && ARRAY['1']::text[]) as old_normal,
    NOT('{2}' IS NOT NULL AND '{2}' && ARRAY['1']::text[]) as old_negated,
    ('{2}' && ARRAY['1']::text[]) as new_normal,
    NOT('{2}' && ARRAY['1']::text[]) as new_negated;
 old_normal | old_negated | new_normal | new_negated
------------+-------------+------------+-------------
 f          | t           | f          | t
```

### DIFF
--- a/packages/server/src/fhir/sql.test.ts
+++ b/packages/server/src/fhir/sql.test.ts
@@ -49,16 +49,14 @@ describe('SqlBuilder', () => {
     test('Select where array contains', () => {
       const sql = new SqlBuilder();
       new SelectQuery('MyTable').column('id').where('name', 'ARRAY_CONTAINS', 'x', 'TEXT[]').buildSql(sql);
-      expect(sql.toString()).toBe(
-        'SELECT "MyTable"."id" FROM "MyTable" WHERE ("MyTable"."name" IS NOT NULL AND "MyTable"."name" && ARRAY[$1]::TEXT[])'
-      );
+      expect(sql.toString()).toBe('SELECT "MyTable"."id" FROM "MyTable" WHERE "MyTable"."name" && ARRAY[$1]::TEXT[]');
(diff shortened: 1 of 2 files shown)


mattlong:
It turns out that this behavior is intentional right now. Putting this PR on hold.

mattlong:
This is by design so that NULL and empty arrays are treated the same.

PR #6096, opened 2025-03-03
Upgrade to react-router v7

codyebberson:
Upgrades `react-router-dom` v6 to `react-router` v7

See: https://reactrouter.com/upgrading/v6

Most of the PR falls into the following:
* Big `package-lock.json` change
* Changing imports from `'react-router-dom'` to `'react-router'`
* Dealing with the new function signature of `navigate()` (see below)

--------

### Fix floating promises in navigate()

As part of our React Router upgrade from v6 to v7, we encountered an issue where the `navigate` function returned by `useNavigate()` now returns a Promise. This caused 237 violations of our ESLint rule `@typescript-eslint/no-floating-promises`, which is critical for preventing unhandled promise rejections in our codebase.

~~After evaluating several approaches, I've decided to use the `void` operator to explicitly indicate that we're intentionally ignoring these promises:~~

After evaluating several approaches, I decided to use `?.catch(console.error)` as needed.

```typescript
// Before
navigate('/dashboard');

// After
navigate('/dashboard')?.catch(console.error);
```

## Alternatives Considered

1. **`useMedplumNavigate`**: Would require big changes to test setup code
2. **Creating a wrapper hook**: Would require either a new shared package or duplicating code across all examples
3. **Adding `void` prefix**: Too many code tools forbid `void`, making this annoying to deal with
4. **Full promise handling with awaits**: Overkill for our use case as we don't need routing completion feedback
6. **Disabling ESLint selectively**: Would result in hundreds of ESLint suppressions, degrading code quality


codyebberson:
This must be the most annoying return type I have ever seen.

image

ThatOneBro:
> This must be the most annoying return type I have ever seen. 
Why is there an or? is it because it can return sync or async depending on transitions or something? I don't really understand the new React 19 stuff yet

codyebberson:
> Why is there an or? is it because it can return sync or async depending on transitions or something?

Yeah, that's my best guess too.  There's nothing in the comments of the code of the type definitions, and I didn't want to wade through the implementation.

> I don't really understand the new React 19 stuff yet

100%.  Real opportunity for a SPA-only fork of this mess.

examples/foomedical/src/test.setup.ts
```diff
@@ -1,7 +1,11 @@
 import '@testing-library/jest-dom';
+import { TextDecoder, TextEncoder } from 'node:util';
 
 (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
 
+Object.defineProperty(globalThis.window, 'TextDecoder', { value: TextDecoder });
```
mattlong, line 6: Curious why these changes were needed.

examples/foomedical/src/components/Header.tsx
```diff
@@ -26,7 +26,7 @@ export function Header(): JSX.Element {
     <AppShell.Header>
       <Container>
         <div className={classes.inner}>
-          <UnstyledButton className={classes.logoButton} onClick={() => navigate('/')}>
+          <UnstyledButton className={classes.logoButton} onClick={() => navigate('/')?.catch(console.error)}>
```
ThatOneBro, line 29: Lol clever but crazy this is necessary now 😂 🥲

ThatOneBro (approved):

mattlong (commented):
Huge diff, so I looked at it commit by commit.

Reading over the feature flags part of the upgrade guide, it seems like [v7_relativeSplatPath](https://reactrouter.com/upgrading/v6#v7_relativesplatpath) is the only one that _might_ affect us since we do indeed have some routes in the format mentioned. Anything to do there?

examples/foomedical/src/test.setup.ts
codyebberson: `<BrowserRouter>` requires it, not sure why? 🤷‍♂️

examples/foomedical/src/components/Header.tsx
codyebberson: Crazy / aggravating

codyebberson:
> Huge diff, so I looked at it commit by commit and breezed over what I assumed were the mechanical changes.

Yes, this is basically a giant find/replace with regex magic to preserve `navigate()` semantics.

> Reading over the feature flags part of the upgrade guide, it seems like [v7_relativeSplatPath](https://reactrouter.com/upgrading/v6#v7_relativesplatpath) is the only one that _might_ affect us since we do indeed have some routes in the format mentioned. Anything to do there?

Good question....

> If you have any routes with a path + a splat like `<Route path="dashboard/*">` that have relative links like `<Link to="relative">` or `<Link to="../relative">` beneath them, you will need to update your code.

Good catch, we do in some of the example apps:

image


Will fix.

mattlong (approved):

examples/foomedical/src/components/Header.tsx
techdavidy: 🥲
was the test setup code change a big overkill in favor of useMedplumNavigate ?

PR #6100, opened 2025-03-03
test(app): Migrate from fireEvent to user-events

### LINKED ISSUE #2637: Migrate from `fireEvent` to `user-events`
While trying to write tests for `<CopyButton>` (https://github.com/medplum/medplum/pull/2634), I found it quite difficult to properly test the hover events and delayed effects.  I looked into how Mantine tests those things, and it turns out internally they use `user-event`:

https://testing-library.com/docs/user-event/intro/#difference-to-fireevent

[When to use userEvent.click and when to use fireEvent](https://stackoverflow.com/questions/64006345/react-testing-library-when-to-use-userevent-click-and-when-to-use-fireevent)

> Behind the scenes, userEvent uses the fireEvent. You can consider fireEvent being the low-level api, while userEvent sets a flow of actions.
> 
> [Here is the code for userEvent.click](https://github.com/testing-library/user-event/blob/5feaa942f46bb37d96c2f2fbeb4b33e8beff75ad/src/click.js#L87-L103)
> 
> You can see that depending of which element you are trying to click, userEvent will do a set of different actions (e.g. if it's a label or a checkbox).

There are probably many cases in `react` and `app` that could benefit from this.

Aquilaafuadajo:
should resolve #2637


Aquilaafuadajo:
Hi @codyebberson, I decided to take on #2637, but it requires changes in 110 files. Would you prefer that I create separate PRs for these changes, or should I include them all in a single PR?

ThatOneBro:
hi @Aquilaafuadajo , thanks for starting this branch! If you want to do a few files at a time it might make it easier to review. This PR looks good so far so if you are ready you can mark it as ready for review and we'll try to get that merged. Thank you!

Aquilaafuadajo:
Alright, I'll do few pr's at a time then. Thank you 👍🏼

Aquilaafuadajo:
Hi @ThatOneBro please can you help take a look at this pr?

ThatOneBro (approved):
LGTM, thank you!

PR #6102, opened 2025-03-03
Local reference resolution with duplicate entries

dylman123:
# Description
This PR adds a test case that demonstrates a potential regression in the FHIR transaction bundle processing when handling duplicate conditional creates with local references.

⚠️ This test is only observed to fail on `medplum/medplum-server:3.2.11` and later ⚠️ 

## The Issue
When a transaction bundle contains multiple entries that:
1. Use conditional create (`ifNoneExist`)
2. Point to the same resource (i.e., would resolve to the same resource)
3. Are referenced by other resources in the bundle using their `fullUrl`

The local reference resolution doesn't correctly handle the fact that both entries resolve to the same resource, leading to incorrect reference resolution.

## Test Case Added
Added test "Local reference resolution with duplicate entries" which demonstrates this scenario:
- Two Organization resources with the same `ifNoneExist` criteria
- A Patient resource that references both Organizations using their respective `fullUrl`s
- The test expects both references in the Patient to resolve to the same Organization ID

## Expected Behavior
When multiple entries in a transaction bundle resolve to the same existing resource via conditional create:
1. All local references using the `fullUrl` of any of these entries should resolve to the same final resource
2. The reference resolution should be consistent regardless of which `fullUrl` was used

## Current Behavior
On v3.2.11 and later, the test fails:
```
 ● Batch › Local reference resolution with duplicate entries

    expect(received).toStrictEqual(expected) // deep equality

    Expected: "Organization/5f101e80-be43-43ba-a8e2-e057a2c84df0"
    Received: "Organization/54066184-9f70-4ab6-af39-a5264937b935"

      1439 |     expect(org1.id).toStrictEqual(org2.id);
      1440 |     expect(patient.generalPractitioner?.[0].reference).toStrictEqual(`Organization/${org1.id}`);
    > 1441 |     expect(patient.managingOrganization?.reference).toStrictEqual(`Organization/${org1.id}`);
           |                                                     ^
      1442 |   });
      1443 | });
      1444 |

      at Object.toStrictEqual (src/batch.test.ts:1441:53)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 37 passed, 38 total
Snapshots:   0 total
Time:        1.492 s, estimated 2 s
```

## Previous Behavior
On v3.2.10, the same test passes:
```
 PASS  src/batch.test.ts
  Batch
    ✓ Local reference resolution with duplicate entries (8 ms)
    ○ skipped Process batch with missing bundle type
    ○ skipped Process batch with invalid bundle type
    ○ skipped Process batch with missing entries
    ○ skipped Process batch success
    ○ skipped Process batch create success
    ○ skipped Process batch create missing resource
    ○ skipped Process batch create missing resourceType
    ○ skipped Process batch create missing required properties
    ○ skipped Process batch create ignore http fullUrl
    ○ skipped Process batch create does not rewrite identifier
    ○ skipped Process batch create ifNoneExist success
    ○ skipped Process batch create ifNoneExist invalid resource type
    ○ skipped Process batch create ifNoneExist multiple matches
    ○ skipped Use ifNoneExist result in other reference
    ○ skipped Process batch update
    ○ skipped Process batch update missing resource
    ○ skipped Process batch patch
    ○ skipped JSONPath error messages
    ○ skipped JSONPatch error messages
    ○ skipped Process batch patch invalid url
    ○ skipped Process batch patch missing resource
    ○ skipped Process batch patch wrong pach type
    ○ skipped Process batch patch wrong pach type
    ○ skipped Process batch delete
    ○ skipped Process batch delete invalid URL
    ○ skipped Process batch missing request
    ○ skipped Process batch missing request.method
    ○ skipped Process batch unsupported request.method
    ○ skipped Process batch missing request.url
    ○ skipped Process batch not found
    ○ skipped Process batch read history
    ○ skipped Valid null
    Process Transactions
      ○ skipped Embedded urn:uuid
      ○ skipped Transaction update after create
      ○ skipped urn:uuid in PATCH

Test Suites: 1 passed, 1 total
Tests:       35 skipped, 1 passed, 36 total
Snapshots:   0 total
Time:        1.647 s
Ran all test suites matching /packages\/fhir-router\/src\/batch.test.ts/i.
```

## Related Issues
- None currently filed, but this test case demonstrates the need for a fix

## Type of Change
- [x] Bug report (non-breaking change which demonstrates a bug)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing
The new test case is added to `packages/fhir-router/src/batch.test.ts`

### DIFF
--- a/packages/fhir-router/src/batch.test.ts
+++ b/packages/fhir-router/src/batch.test.ts
@@ -18,6 +18,7 @@ import {
   DiagnosticReport,
   Observation,
   OperationOutcome,
+  Organization,
   Patient,
   Practitioner,
   SearchParameter,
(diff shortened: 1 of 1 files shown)


mattwiller:
Hi @dylman123 — unfortunately, I don't believe this use case is considered valid per the [FHIR spec][tx-spec], which says:
> A resource can only appear in a transaction once (by identity)

It further reiterates:
> If any resource identities (including resolved identities from conditional update/delete) overlap in steps 1-3, then the transaction SHALL fail.

Since these resource identities are necessarily resolved before the transaction is processed, both conditional creates will succeed, creating two different new resources when no matching resource exists on the server before the transaction is processed.  We could be more explicit here and return an error when multiple identical conditional creates/updates/deletes are passed in, but that wouldn't protect against similar situations where the conditions/resources between entries overlap but are not identical.

[tx-spec]: http://hl7.org/fhir/R4/http.html#trules

dylman123:
> Hi @dylman123 — unfortunately, I don't believe this use case is considered valid per the [FHIR spec](http://hl7.org/fhir/R4/http.html#trules), which says:
> 
> > A resource can only appear in a transaction once (by identity)
> 
> It further reiterates:
> 
> > If any resource identities (including resolved identities from conditional update/delete) overlap in steps 1-3, then the transaction SHALL fail.
> 
> Since these resource identities are necessarily resolved before the transaction is processed, both conditional creates will succeed, creating two different new resources when no matching resource exists on the server before the transaction is processed. We could be more explicit here and return an error when multiple identical conditional creates/updates/deletes are passed in, but that wouldn't protect against similar situations where the conditions/resources between entries overlap but are not identical.

Thanks @mattwiller for the feedback. That is a good catch. Okay nevermind, I will close this issue

PR #6104, opened 2025-03-03
Fix warning about missing key in ChooseScopeForm

### LINKED ISSUE #6098: [ChooseScopeForm] Warning: Each child in a list should have a unique "key" prop
Image

mattwiller:
Resolves #6098

### DIFF
--- a/packages/react/src/auth/ChooseScopeForm.tsx
+++ b/packages/react/src/auth/ChooseScopeForm.tsx
@@ -62,10 +62,10 @@ export function ChooseScopeForm(props: ChooseScopeFormProps): JSX.Element {
               ];
             }
             return (
-              <>
+              <div key={scopeName + '_group'}>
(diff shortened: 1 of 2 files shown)


packages/react/src/auth/ChooseScopeForm.tsx
```diff
@@ -62,10 +62,10 @@ export function ChooseScopeForm(props: ChooseScopeFormProps): JSX.Element {
               ];
             }
             return (
-              <>
+              <div key={scopeName + '_group'}>
```
codyebberson, line 65: If you don't actually need a `<div>`, you can use `Fragment` (imported `from 'react';`)

That's the syntactic way to add a `key` prop to a fragment.

```suggestion
              <Fragment key={scopeName + '_group'}>
```

ThatOneBro (approved):

codyebberson (approved):

Issue #1563, opened 2023-02-19
Replace node-fetch with Node.js 18 built-in fetch


codyebberson:
Node.js 18 includes built-in `fetch` which means we should be able to stop using the `node-fetch` library

We already require Node.js 18 in `server` and bots, so no issues there.

We already allow "bring your own `fetch`" in `MedplumClient`, so client users can continue to use whatever `fetch` implementation they want.

The blocker right now is that TypeScript Node.js type definitions do not include `fetch`: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924

It is possible to workaround like this:

```ts
import { FetchLike } from '@medplum/core';

declare global {
  var fetch: FetchLike;
}
```

But I am inclined to wait until https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924 is resolved, and do it right.

codyebberson:
Switching from `node-fetch` to built-in fetch will be straightforward.

Switching `fetch` in tests will take a little more work: https://stackoverflow.com/questions/74656122/how-do-i-mock-node-js-fetch-http-requests-responses-in-node-18

codyebberson:
This may need to wait until Node.js 20

https://twitter.com/ebey_jacob/status/1709975146939973909?t=qd3X0yPEDBuxSEu6zQLAHA&s=19

codyebberson:
Confirmed that this needs to wait for Node.js 20: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924#issuecomment-1749274902

PR #6105, opened 2025-03-03
Fix excluded dep upgrades: rimraf, supertest

codyebberson:
`rimraf` - we were waiting for Node 20 🎉 

`supertest` - we were waiting for `superwstest` to be fixed, which it looks like it is 🎉 

--------

As an aside, I tried upgrading ESLint to 9.  Unfortunately, that is a much bigger endeavor.  The syntax has changed dramatically.  It would be one thing if we were just updating a local ESLint config within a project.  The fact that we're publishing `@medplum/eslint-config` as a dependency available to others makes this quite a bit more complicated though

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -92,11 +92,8 @@ fi
 # commander - v13 has backwards-incompatible changes which require a decent amount of refactoring to get our current code to work. We are considering migrating off of commander but for now we should just freeze it
 # eslint - version 9+ conflicts with Next.js plugins, holding back until fixed
 # jose - version 6+ requires ESM (depending on the precise NodeJS version), holding back until server supports ESM
-# react-router-dom - version 7+ has breaking changes, will fix separately
 # node-fetch - version 3+ requires ESM, holding back until server supports ESM
-# rimraf - version 6+ requires Node 20+, holding back until Medplum v4
-# supertest - version 7+ incompatible with superwstest, waiting for fix
-MAJOR_EXCLUDE="@types/express @types/node @types/react @types/react-dom commander eslint jose react-router-dom node-fetch rimraf supertest"
+MAJOR_EXCLUDE="@types/express @types/node @types/react @types/react-dom commander eslint jose node-fetch"
 
 if [ "$LAST_STEP" -lt 1 ]; then
     # First, only upgrade patch and minor versions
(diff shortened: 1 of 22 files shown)


mattwiller (approved):

scripts/upgrade.sh
```diff
@@ -92,11 +92,8 @@ fi
 # commander - v13 has backwards-incompatible changes which require a decent amount of refactoring to get our current code to work. We are considering migrating off of commander but for now we should just freeze it
 # eslint - version 9+ conflicts with Next.js plugins, holding back until fixed
 # jose - version 6+ requires ESM (depending on the precise NodeJS version), holding back until server supports ESM
-# react-router-dom - version 7+ has breaking changes, will fix separately
 # node-fetch - version 3+ requires ESM, holding back until server supports ESM
-# rimraf - version 6+ requires Node 20+, holding back until Medplum v4
-# supertest - version 7+ incompatible with superwstest, waiting for fix
-MAJOR_EXCLUDE="@types/express @types/node @types/react @types/react-dom commander eslint jose react-router-dom node-fetch rimraf supertest"
+MAJOR_EXCLUDE="@types/express @types/node @types/react @types/react-dom commander eslint jose node-fetch"
```
ThatOneBro, line 96: Can probably remove `@types/react-dom` too?

ThatOneBro: Also did we investigate removing `node-fetch` now that `fetch` has stabilized for `Node20`? or I guess we may have to support it because of bots...

ThatOneBro (approved):

scripts/upgrade.sh
codyebberson: > Can probably remove `@types/react-dom` too?

I figured we should treat that the same as `react`?  Might be worth a try.

> Also did we investigate removing node-fetch now that fetch has stabilized for Node20? 

I've been keeping my notes here:  https://github.com/medplum/medplum/issues/1563

Node 20 does seem like the last blocker.

I think that testing is quite a bit different with built-in fetch, and you need to use the undici `MockAgent` or something.  Maybe?  I assume that's a non-trivial change.

codyebberson:
Not sure what's going on here.  This doesn't repro for me on local machine, with Node 20 or 22.  It looks like it might be an ipv6 thing, so I'm going to try nudging it to use ipv4

image

ThatOneBro:
Yeah I think we have seen network weirdness on Github runners. Probably that.

PR #6107, opened 2025-03-04
Enhance token not equals test

mattlong:
On reading the description of `ne` again, I'm not entirely sure this is the correct behavior, but it is indeed the current behavior of Medplum.


mattwiller:
This does appear to be the intended behavior per [the FHIR spec](http://hl7.org/fhir/R4/search.html#token) for `token` search with the `:not` modifier:
> Reverse the code matching described in the paragraph above: return all resources that do not have a matching item. Note that this includes resources that have no value for the parameter - e.g. `?gender:not=male` includes all patients that do not have `gender = male`, including patients that do not have a `gender` at all

mattwiller (approved):

PR #6108, opened 2025-03-04
Fixed flaky test in date chained search

codyebberson:


packages/server/src/fhir/search.test.ts
```diff
@@ -4213,7 +4215,7 @@ describe('FHIR Search', () => {
           resourceType: 'Composition',
           status: 'final',
           type: { text: 'test' },
-          date: '2025-01-01',
```
codyebberson, line 4216: `Composition.date` is a `dateTime` field, so it should have a date+time component.

packages/server/src/fhir/search.test.ts
```diff
@@ -4234,7 +4236,9 @@ describe('FHIR Search', () => {
         expect(searchResult1.entry?.[0]?.resource?.id).toBe(bundle.id);
 
         // Chained search on the Composition
-        const searchResult2 = await repo.search(parseSearchRequest(`Bundle?composition.date=2025-01-01`));
```
codyebberson, line 4237: At present, searching on a `dateTime` field with a `date` string can lead to some funky results.

In the future, when we move to proper intervals for everything, then this would (insert time/zone caveats) work as expected.

For now, best to be explicit for deterministic results.

packages/server/src/fhir/search.test.ts
```diff
@@ -4339,7 +4343,7 @@ describe('FHIR Search', () => {
         withTestContext(async () => {
           const parentObservations = await createObservations(repo, 3, {});
           const childObservations = [
-            await createObservations(repo, 2, { hasMember: [parentObservations[0]] }),
+            await createObservations(repo, 1, { hasMember: [parentObservations[0]] }),
```
codyebberson, line 4346: I'm not 100% sure, but it seems like this test depends on `childObservations[0][0]`, and when we created 2 observations, it was not always guaranteed that we would get the right one.

This test is 7 months old, and I don't think I've ever seen a failure here before today, but it failed 2 times in rapid succession today:
1. https://github.com/medplum/medplum/actions/runs/13643910874/job/38139380804?pr=6105
2. https://github.com/medplum/medplum/actions/runs/13643696777/job/38138680228?pr=6107

codyebberson:
Another one: https://github.com/medplum/medplum/actions/runs/13647518413/job/38149112048?pr=6109

image

mattwiller (approved):

codyebberson:
@mattlong  implemented an alternate fix to the `searchByReference` test in https://github.com/medplum/medplum/pull/6111

I'm going to revert that part, but keep the Composition-date fix

PR #6110, opened 2025-03-04
Provide a default chart nore with/out a plan definition to encounter

techdavidy:
- After an encounter is created and new `Chart Note` is created as well for the corresponding Encounter.
(ChartNote is added to Encounter with or without a PlanDefinition.)
- Create useEncounterHook
- Reuse usePatientHook
- Create a component `EncounterStatus` to encapsulate its logic.

Screenshot 2025-03-03 at 10 58 10 PM


examples/medplum-provider/src/pages/components/Encounter/EncounterStatus.tsx
```diff
@@ -0,0 +1,29 @@
...
+    case 'in-progress':
+      return classes.green;
+    case 'finished':
+      return classes.blue;
+    default:
+      return classes.gray;
+  }
+};
+
+export const EncounterStatus = ({ encounter }: EncounterStatusProps): JSX.Element => {
```
codyebberson, line 23: We have a `<StatusBadge>` component in `@medplum/react`, which would be nice to use here.  We should be able to eliminate this entire component.

The colors don't exactly line up, but I think we can change the colors in `<StatusBadge>` to align with the requirements here.

See: [`packages/react/src/StatusBadge/StatusBadge.tsx`](https://github.com/medplum/medplum/blob/main/packages/react/src/StatusBadge/StatusBadge.tsx)

examples/medplum-provider/src/hooks/useEncounter.ts
```diff
@@ -0,0 +1,16 @@
+import { OperationOutcome, Encounter } from '@medplum/fhirtypes';
+import { useResource } from '@medplum/react';
+import { useParams } from 'react-router';
+
+type Options = {
+  ignoreMissingEncounterId?: boolean;
+  setOutcome?: (outcome: OperationOutcome) => void;
+};
+
+export function useEncounter(options?: Options): Encounter | undefined {
```
codyebberson, line 10: It looks like we don't ever use the `options` argument at any of the call sites, so maybe we can remove it?

codyebberson (changes requested):

codyebberson (approved):

PR #6114, opened 2025-03-04
Ability to create sql functions in migrations

mattlong:
This is in preparation for the new token implementation that will require at least one SQL function.


mattlong:
A different version of this will be part in an upcoming PR.

PR #6118, opened 2025-03-05
Add `setDefaultResultOrder('ipv4first')` in test setup

mattlong:
After https://github.com/medplum/medplum/pull/6105, websocket tests in `packages/server` were hanging for me locally (MacOS Sequoia) unless I also add `--dns-result-order=ipv4first` to `NODE_OPTIONS`. Trying this as an alternative.

### DIFF
--- a/packages/server/src/test.setup.ts
+++ b/packages/server/src/test.setup.ts
@@ -11,6 +11,7 @@ import {
   Resource,
 } from '@medplum/fhirtypes';
 import { randomUUID } from 'crypto';
+import { setDefaultResultOrder } from 'dns';
 import { Express } from 'express';
 import internal from 'stream';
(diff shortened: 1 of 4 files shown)


codyebberson (approved):

packages/server/src/test.setup.ts
```diff
@@ -21,6 +22,9 @@ import { generateAccessToken } from './oauth/keys';
 import { tryLogin } from './oauth/utils';
 import { requestContextStore } from './request-context-store';
 
+// supertest v7 can cause websocket tests to hang without this
+setDefaultResultOrder('ipv4first');
```
ThatOneBro, line 26: Do we need to do this in agent too?

PR #6136, opened 2025-03-10
feat(agent): add PID file system

### LINKED ISSUE #6084: Gracefully Handle two agents running on the same machine
## Background
example situation: 

- Agent is already running as a Windows service
- User _also_ clicks on the agent.exe file, not knowing that the service s running
- Now you have two agents, with the same config. 

## Observed Behavior
- Each agent "steals" the connection from the other, causing the agent connectivity status to continually "bounce"
- intermittent failures on `Agent/$push`

## Desired Behavior
- The 2nd agent fails to start and exits, as there is already an active agent running

ThatOneBro:
Fixes #6084

The agent service creates a PID file and checks for its existence before continuing startup.

An important part of the repro was making sure there were no channels that bound to ports before launching the second process, as attempting to bind to the same port twice would crash the second agent already. This is still a problem if the second agent was started and never stopped before adding the channels, which would then produce a weird case where both instances fight for the same port and potentially the wrong process binds to the port first (the free-standing process versus the service-controlled process).


packages/agent/src/pid.ts
```diff
@@ -0,0 +1,139 @@
...
+    // Sending signal 0 doesn't actually send a signal but checks if process exists
+    process.kill(pid, 0);
+    // If we didn't throw an error above, the process is still running, and PID is not stale
+    return true;
+  } catch (err) {
+    // If the error is "ESRCH", the process doesn't exist
+    if ((err as Error & { code: string }).code === 'ESRCH') {
+      return false;
+    }
+    // If the error is "EPERM", the process exists but we don't have permission
```
ThatOneBro, line 59: This is necessary for Windows in the common case of double clicking the exe in `C:\ProgramFiles\MedplumAgent` since it by default will run as a user without permissions to check the service process which is running as an administrator.

packages/agent/src/pid.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    // If we make it here, PID file exists but it's stale
+    pidLogger.info('Stale PID file found. Overwriting...');
+  }
+
+  // We need to make sure the directory exists first since we aren't just using a system-created dir like /var/run
+  ensureDirectoryExists(path.dirname(pidFilePath));
+
+  // Write the PID file atomically using a temporary file
```
mattlong, line 94: Can you expand on why the PID file needs to be written atomically and how using the `.tmp` suffix ensures that? It seems like it'd need to be a random extension if this is geared at competing medplum agent processes attempting to write the PID file? In other words, doesn't this just shift the potential for conflict from `pidFilePath` to `${pidFilePath}.tmp`?

ThatOneBro: That's a good callout, yeah competing agents was the thought but you're right, technically the second process could clobber the PID in the tmp file but it wouldn't be able to rename the file since it won't exist after the first process renamed it. So it could be in a bad state.

I looked at the Node docs again and remembered that using the `wx` flags should give us the atomic write to prevent the PID file from being overwritten, so I'll change it to use that 👍

mattlong (approved):

Issue #5966, opened 2025-02-13
SNOMED synonym search


codyebberson:
I tried entering data for the ONC (b)(1) / (g)(9) certification.

One of the entries is an allergic reaction to Penicillin, and reaction of "Hives".

I tried searching for "Hives", but no responses came up.

It turns out the technical term is "Wheal":

https://bioportal.bioontology.org/ontologies/SNOMEDCT?p=classes&conceptid=247472004

![Image](https://github.com/user-attachments/assets/923eb6ca-db9e-4418-bac1-ce15c8b27502)

It looks like "Hives" is a known synonym of "Wheal".  How bad would it to be add synonyms to code search?

reshmakh:
Related: https://github.com/medplum/medplum/pull/6842

PR #6137, opened 2025-03-10
Add column for Coding synonyms

mattwiller:
This is the first portion of work to support searching on synonyms in `ValueSet/$expand` (#5966): adding a column to the internal `Coding` lookup table to support marking some display text as a synonym.  The expected database representation is e.g.:

| id | system | code | display | isSynonym |
| -- | ------ | ---- | ------- | --------- |
| ...
| 123 | http://snomed.info/sct | 247472004 | Wheal (finding) | \<NULL\> |
| 124 | http://snomed.info/sct | 247472004 | Welt | true |
| 125 | http://snomed.info/sct | 247472004 | Weal | true |
| 126 | http://snomed.info/sct | 247472004 | Uticarial rash | true |
| 127 | http://snomed.info/sct | 247472004 | Hives | true |
| 128 | http://snomed.info/sct | 247472004 | Nettle rash | true |

The next step will be updating the `UNIQUE` index on the `Coding` table from `(system, code)` to `(system, code) WHERE "isSynonym" IS NULL` to enable writing the synonym rows.


packages/server/src/migrations/schema/v87.ts
```diff
@@ -0,0 +1,12 @@
...
+ * This is a generated file
+ * Do not edit manually.
+ */
+
+import { PoolClient } from 'pg';
+
+export async function run(client: PoolClient): Promise<void> {
+  await client.query('ALTER TABLE IF EXISTS "Coding" ADD COLUMN IF NOT EXISTS "isSynonym" BOOLEAN');
+  await client.query('DROP INDEX CONCURRENTLY "Coding_display_idx"');
+  await client.query('ALTER INDEX IF EXISTS "Coding_display_trgm_idx" RENAME TO "Coding_display_idx"');
```
mattwiller, line 11: This migration was tested against a `Coding` table containing a full import of UMLS on my laptop, and ran in 10 ms

packages/server/src/migrations/migrate.ts
```diff
@@ -696,15 +696,11 @@ function buildCodingTable(result: SchemaDefinition): void {
...
+      { name: 'isSynonym', type: 'BOOLEAN' },
     ],
     indexes: [
       { columns: ['id'], indexType: 'btree', unique: true },
       { columns: ['system', 'code'], indexType: 'btree', unique: true, include: ['id'] },
-      {
-        columns: ['system', { expression: "to_tsvector('english'::regconfig, display)", name: 'display' }],
-        indexType: 'gin',
-        where: 'display IS NOT NULL',
-      },
```
mattwiller, line 707: This index is no longer in use, and was replaced by the one below it in the definition

codyebberson (approved):

PR #6142, opened 2025-03-11
Added SubmitButton component to prevent double submit

codyebberson:


packages/react/src/Form/Form.tsx
```diff
@@ -1,27 +1,41 @@
-import { CSSProperties, ReactNode, SyntheticEvent } from 'react';
+import { CSSProperties, ReactNode, SyntheticEvent, useState } from 'react';
+import { FormContext } from './Form.context';
 import { parseForm } from './FormUtils';
 
 export interface FormProps {
-  readonly onSubmit?: (formData: Record<string, string>) => void;
+  readonly onSubmit?: (formData: Record<string, string>) => Promise<void> | void;
```
codyebberson, line 6: 😢

packages/react/src/Form/Form.tsx
```diff
@@ -1,27 +1,41 @@
...
+          const formData = parseForm(e.target as HTMLFormElement);
+          if (props.onSubmit) {
+            setSubmitting(true);
+            const result = props.onSubmit(formData);
+            if (result?.then) {
+              result
+                .then(() => {
+                  setSubmitting(false);
+                })
+                .catch(console.error);
```
mattwiller, line 30: If there's an error, should the button be returned to the normal state so the user can attempt to re-submit?

mattwiller (approved):

packages/react/src/Form/Form.tsx
codyebberson: Yes, good catch, will change to 'finally'

PR #6143, opened 2025-03-11
Add guard to canonical chained search types

mattwiller:
The `RelatedArtifact` data type contains an element with type `canonical(Any)`, which is referenced in some search parameters (e.g. `EvidenceVariable:derived-from`).  This can cause issues with chained search, since the canonical reference expects the target type to have `url` search parameter, which many resource types do not have.  Adding an explicit guard will make this more obvious to end users

### DIFF
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -4636,6 +4636,17 @@ describe('FHIR Search', () => {
           expect(result.entry?.find((e) => e.resource?.valueQuantity?.value === 85)).toBeDefined();
         }));
     });
+
+    test('Invalid canonical chained search link', async () =>
+      withTestContext(async () => {
+        await expect(
+          repo.search(parseSearchRequest('EvidenceVariable?derived-from:ResearchStudy.status=active'))
+        ).rejects.toThrow('ResearchStudy cannot be chained via canonical reference (EvidenceVariable:derived-from)');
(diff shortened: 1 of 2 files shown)


ThatOneBro (commented):
Is this something that we should look at getting working in the future? Also is there any mention of reverse chaining with canonical references in the FHIR spec?

ThatOneBro:
@mattwiller Ah yeah I think I misread the test, this makes sense. Thank you

ThatOneBro (approved):

pamella:
Hey folks @mattwiller @ThatOneBro

It looks like this recent change affected a filter I was using. Would you happen to have any suggestions for a workaround? I’d appreciate your insights.

For reference, the filter is:
`ResearchStudy?_fields=title,status,_lastUpdated&_sort=-_lastUpdated&outcome-measure-reference:EvidenceVariable.derived-from:Questionnaire.title:contains=adl`

To provide some context on the data modeling:

- `ResearchStudy.outcomeMeasure` points to an `EvidenceVariable`
- `EvidenceVariable.relatedArtifact` points to a URL of a `Questionnaire`
- The filter is designed to retrieve research studies based on the `title` of a `Questionnaire`

mattwiller:
@pamella Sorry about that, we have [a fix in review](https://github.com/medplum/medplum/pull/6162)

PR #6147, opened 2025-03-11
Import correct `getLogger`

mattlong:
I checked the source of the [nodemailer getLogger](https://github.com/nodemailer/nodemailer/blob/v6.10.0/lib/shared/index.js#L373-L401) to see where those log lines were going; it was a noop implementation, i.e. `() => false`, since we were not configuring nodemailer logging at all.


ThatOneBro (approved):
Thank you 🙏

PR #6170, opened 2025-03-14
Use erasableSyntaxOnly

codyebberson:
This PR explores using TypeScript 5.8's new "erasableSyntaxOnly" flag to move us closer to Node.js's built-in TypeScript runners. The changes primarily involve converting `enum`s to `const` objects with corresponding `type` definitions, and replacing constructor property declarations with standard class property declarations.

## Changes
- Implemented TypeScript's "erasableSyntaxOnly" flag for compilation
- Converted enums to const objects with type definitions
- Changed constructor property declarations to standard class property declarations

## Motivation
- Enable future ability to run TypeScript natively without transpilation tools like `ts-node` and `ts-node-dev`
- Set the groundwork for adopting Node.js's built-in test runner, eliminating dependencies on `jest` and `vitest`
- Take advantage of Node.js's improving TypeScript support (as referenced in [running TypeScript natively](https://nodejs.org/en/learn/typescript/run-natively))

## Compatibility Notes
- This approach is not 100% backwards compatible (enums will lose reverse-lookup capability by default)
- However, the impact should be minimal enough to avoid requiring major version bumps

## Current Limitations
- We cannot yet include "erasableSyntaxOnly" in tsconfig.json as it's not supported by `api-extractor`
- A fix for this has been merged in `api-extractor` (https://github.com/microsoft/rushstack/pull/5151) but hasn't been released yet

## Feedback Requested

This PR is exploratory rather than definitive. I'm looking for team feedback on this approach before we commit to this direction.

### DIFF
--- a/packages/core/src/access.ts
+++ b/packages/core/src/access.ts
@@ -29,27 +29,29 @@ export const projectAdminResourceTypes = [
  *
  * Codes taken from http://hl7.org/fhir/codesystem-restful-interaction.html
  */
-export enum AccessPolicyInteraction {
-  READ = 'read',
-  VREAD = 'vread',
-  UPDATE = 'update',
-  PATCH = 'patch',
-  DELETE = 'delete',
-  HISTORY = 'history',
-  HISTORY_INSTANCE = 'history-instance',
-  HISTORY_TYPE = 'history-type',
-  HISTORY_SYSTEM = 'history-system',
-  CREATE = 'create',
-  SEARCH = 'search',
-  SEARCH_TYPE = 'search-type',
-  SEARCH_SYSTEM = 'search-system',
-  SEARCH_COMPARTMENT = 'search-compartment',
-  CAPABILITIES = 'capabilities',
-  TRANSACTION = 'transaction',
-  BATCH = 'batch',
-  OPERATION = 'operation',
-}
-const resourceReadInteractions = [
+export const AccessPolicyInteraction = {
+  READ: 'read',
+  VREAD: 'vread',
+  UPDATE: 'update',
+  PATCH: 'patch',
+  DELETE: 'delete',
+  HISTORY: 'history',
+  HISTORY_INSTANCE: 'history-instance',
+  HISTORY_TYPE: 'history-type',
+  HISTORY_SYSTEM: 'history-system',
+  CREATE: 'create',
+  SEARCH: 'search',
+  SEARCH_TYPE: 'search-type',
+  SEARCH_SYSTEM: 'search-system',
+  SEARCH_COMPARTMENT: 'search-compartment',
+  CAPABILITIES: 'capabilities',
+  TRANSACTION: 'transaction',
+  BATCH: 'batch',
+  OPERATION: 'operation',
+} as const;
+export type AccessPolicyInteraction = (typeof AccessPolicyInteraction)[keyof typeof AccessPolicyInteraction];
+
+const resourceReadInteractions: AccessPolicyInteraction[] = [
   AccessPolicyInteraction.READ,
   AccessPolicyInteraction.VREAD,
   AccessPolicyInteraction.HISTORY,
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -114,10 +114,10 @@ export function reorderBundle(bundle: Bundle): Bundle {
 type AdjacencyList = Record<string, string[]>;
 
 const VertexState = {
-  NotVisited: 'NotVisited' as const,
-  Visiting: 'Visiting' as const,
-  Visited: 'Visited' as const,
-};
+  NotVisited: 'NotVisited',
+  Visiting: 'Visiting',
+  Visited: 'Visited',
+} as const;
 type VertexState = (typeof VertexState)[keyof typeof VertexState];
 
 function topologicalSortWithCycles(graph: AdjacencyList): { sorted: string[]; cycles: string[][] } {
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -113,11 +113,12 @@ export function reorderBundle(bundle: Bundle): Bundle {
 
 type AdjacencyList = Record<string, string[]>;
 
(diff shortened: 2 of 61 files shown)


packages/core/src/bundle.ts
```diff
@@ -113,11 +113,12 @@ export function reorderBundle(bundle: Bundle): Bundle {
...
-enum VertexState {
-  NotVisited,
-  Visiting,
-  Visited,
-}
+const VertexState = {
+  NotVisited: 'NotVisited' as const,
+  Visiting: 'Visiting' as const,
+  Visited: 'Visited' as const,
+};
```
mattwiller, line 120: Why this difference in syntax from the file above?  In `access.ts`, the whole object is `as const`, where here it's each property — I think I'd prefer the former if they're more or less equivalent

packages/core/src/access.ts
```diff
@@ -29,27 +29,29 @@ export const projectAdminResourceTypes = [
...
+  SEARCH: 'search',
+  SEARCH_TYPE: 'search-type',
+  SEARCH_SYSTEM: 'search-system',
+  SEARCH_COMPARTMENT: 'search-compartment',
+  CAPABILITIES: 'capabilities',
+  TRANSACTION: 'transaction',
+  BATCH: 'batch',
+  OPERATION: 'operation',
+} as const;
+export type AccessPolicyInteraction = (typeof AccessPolicyInteraction)[keyof typeof AccessPolicyInteraction];
```
mattwiller, line 52: Just curious, how does this work, having two exports with the same name but one is a type?

packages/core/src/fhirlexer/parse.ts
```diff
@@ -12,10 +12,13 @@ export interface Atom {
...
-    public readonly operator: string,
-    public readonly child: Atom
-  ) {}
+  readonly operator: string;
+  readonly child: Atom;
+
+  constructor(operator: string, child: Atom) {
+    this.operator = operator;
+    this.child = child;
+  }
```
mattwiller, line 21: 👍

packages/core/src/search/match.test.ts
```diff
@@ -135,7 +135,7 @@ describe('Search matching', () => {
     const resource: Observation = { resourceType: 'Observation', subject: { reference: 'Patient/123' } } as Observation;
     const search = {
       resourceType: 'Observation',
-      filters: [{ code: 'subject', operator: Operator.EQUALS, value: 'Patient/123' }],
+      filters: [{ code: 'subject', operator: Operator.EQUALS as Operator, value: 'Patient/123' }],
```
mattwiller, line 138: This feels pretty icky: shouldn't using the constant directly mean it has the right type already?

packages/core/src/bundle.ts
ThatOneBro: Yeah I think we should make the whole object `as const`

packages/core/src/search/search.ts
```diff
@@ -49,45 +49,46 @@ export interface IncludeTarget {
...
 
   // Reference
-  IDENTIFIER = 'identifier',
+  IDENTIFIER: 'identifier' as const,
 
   // _include and _revinclude
-  ITERATE = 'iterate',
-}
+  ITERATE: 'iterate' as const,
+};
```
ThatOneBro, line 90: We should probably make the whole object as const, I think that will fix the cast issue

ThatOneBro (commented):
Yeah I think this is not too bad, I already prefer the const objects to enums anyways.

I think the Node Typescript runtime is worth exploring, though we need to do side-by-side comparisons and maybe do some research on how it works under the hood. 

There are going to be some optimizations that compilers do that we will miss out on if we just erase types which could have some significant perf impacts as scale, or they could be negligible. But especially for prod, we will want to consider if we still want to keep building via esbuild rather than running TypeScript directly. For dev, it definitely seems like it would be preferable if we can run the TypeScript directly though as that definitely simplifies our tooling

packages/core/src/access.ts
codyebberson: See "Alternative 2: Object Literals" here: https://maxheiber.medium.com/alternatives-to-typescript-enums-50e4c16600b1

In TypeScript, values and types exist in two separate "namespaces" or scopes:

1. **Value space**: Contains variables, functions, classes, enums, and other entities that exist at runtime in JavaScript.

2. **Type space**: Contains interfaces, type aliases, and other type constructs that exist only during compilation and are erased in the JavaScript output.

TypeScript determines from the context whether you're referring to the value or the type:

```typescript
// Using the value
const action = AccessPolicyInteraction.READ;

// Using the type
function doSomething(interaction: AccessPolicyInteraction) { ... }
```

packages/core/src/bundle.ts
codyebberson: Yeah, good catch.  These `as const` weren't required.  I think Github Copilot may have added them in.  I'll fix.

packages/core/src/search/details.ts
```diff
@@ -14,18 +14,19 @@ import { PropertyType, getElementDefinition, globalSchema } from '../types';
...
+  NUMBER: 'NUMBER',
+  QUANTITY: 'QUANTITY',
+  TEXT: 'TEXT',
+  REFERENCE: 'REFERENCE',
+  CANONICAL: 'CANONICAL',
+  DATE: 'DATE',
+  DATETIME: 'DATETIME',
+  PERIOD: 'PERIOD',
+  UUID: 'UUID',
+};
```
ThatOneBro, line 28: Think you need `as const` here

ThatOneBro (approved):

codyebberson:
Holding on this until after Monday dependency upgrade for `api-extractor` version 7.52.1: https://github.com/microsoft/rushstack/blob/main/apps/api-extractor/CHANGELOG.md

Then we can add `"erasableSyntaxOnly"` to `tsconfig.json`, and remove `--erasableSyntaxOnly` from all `package.json` files

mattwiller (approved):

PR #6172, opened 2025-03-14
Support search on _project :missing/:present

mattlong:
This will effectively only work for super admins since `addSecurityFilters` will negate the effects.


mattwiller:
> This will effectively only work for super admins since `addSecurityFilters` will negate the effects

Would you mind adding a test to that effect?

codyebberson (approved):

codyebberson (approved):

PR #6176, opened 2025-03-17
perf(bulkstatus): use MGET to get all statuses at once

### LINKED ISSUE #6015: Bulk Agent status does not include all resources as part of the collection
Related to [4194](https://github.com/medplum/medplum/issues/4194) the response from the bulk-status operation is set to a `type: collection` but it is not returning all results. To get more than 20 results, the `_count=` query param must be used. The Bundle.type with `collection` does not need to support a `_count` as that would entail some pagination. Could the collection type return all results or move to a different Bundle type?

URL example: `Agent/$bulk-status?status=active` only returns a subset
To get additional resources: `Agent/$bulk-status?status=active&_count=50`

```
{
	"resourceType": "Bundle",
	"type": "collection",
	"entry": [
		{
			"resource": {
				"resourceType": "Parameters",
				"parameter": [
					{
						"name": "agent",
						"resource": {
							"resourceType": "Agent",
```

ThatOneBro:
Fixes #6015

### DIFF
--- a/packages/server/src/fhir/operations/agentbulkstatus.test.ts
+++ b/packages/server/src/fhir/operations/agentbulkstatus.test.ts
@@ -240,7 +240,13 @@ describe('Agent/$bulk-status', () => {
     expect(bundle.entry).toHaveLength(1);
 
     expectBundleToContainOutcome(bundle, agents[1], {
-      issue: [expect.objectContaining({ severity: 'error', code: 'exception' })],
+      issue: [
+        expect.objectContaining({
+          severity: 'error',
+          code: 'invalid',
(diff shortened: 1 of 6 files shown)


codyebberson (commented):
How would you feel about the following:

1. In `agentbulkstatus.ts`, refactor the implementation out into a standalone method (not a request handler, just a function that you can call with an array of agents).
2. In `agentstatus.ts`, use that method, rather than 

The thinking is:
1. Better to have a single implementation of "get agent status(es) from redis"
2. Negligible overhead of using `mget` with one value vs a normal `get`.

codyebberson (approved):

codyebberson:
Nice, clean 🤌

PR #6178, opened 2025-03-17
Dependency upgrades 2025-03-17

medplumbot:
Dependency upgrades


codyebberson:
I reviewed the dependency license issues.  Most of them are due to formatting issues (vite, rolloup) or size issues (aws-cdk). 

`utf8-byte-length` uses "WTFPL", which is not technically "OSI Approved", but only because they viewed it as ["redundant"](https://github.com/openjournals/joss/issues/248#issuecomment-302613744), which I'm ok to ignore.

codyebberson:
Deployed and tested on staging ✅

codyebberson (approved):

PR #6181, opened 2025-03-17
Fixing compile issues on SMART app

reshmakh:


codyebberson (changes requested):
`react-router-dom` is v6.  we upgraded to v7.

This is likely a `npm install` issue.  I'll ping you separately.

PR #6183, opened 2025-03-17
Align crawler function signatures

codyebberson:
This is technically a breaking API change, although this is pretty deep in the internals, such that i'm 99% sure this is ok.  Open to feedback if there are any objections.

The key goal here is to align the function signatures of `crawlTypedValue` and `crawlTypedValueAsync`


mattlong (approved):

codyebberson:
This is potentially a breaking change.  Maybe for v5.

PR #6186, opened 2025-03-18
Added C-CDA narratives and display text

codyebberson:
This PR enhances our C-CDA and International Patient Summary (IPS) exports by implementing narrative text sections that were previously minimal or blank. While our prior focus was on structured data export, this update addresses the critical need for human-readable narrative content that C-CDA and IPS viewers heavily rely on.

Key changes:
- Added HTML tables in each section's narrative block based on ONC consultant samples
- Implemented consistent formatting across all narrative sections
- Ensured narrative content properly reflects the structured data being exported
- Maintained parallel development across both C-CDA and IPS features due to their significant overlap

These improvements significantly enhance document readability in clinical viewers while maintaining compliance with the expected format. The narrative content follows validated patterns from our ONC consultant samples, giving us confidence that the implementation meets industry requirements.

For example: https://www.ccdaviewer.com/viewer/

image

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1404,3 +1404,21 @@ export function flatMapFilter<T, U>(arr: T[] | undefined, fn: (value: T, idx: nu
   }
   return result;
 }
+
+/**
+ * Returns the escaped HTML string of the input string.
+ * @param unsafe - The unsafe HTML string to escape.
+ * @returns The escaped HTML string.
+ */
+export function escapeHtml(unsafe: string): string {
+  return unsafe
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/“/g, '&ldquo;')
+    .replace(/”/g, '&rdquo;')
+    .replace(/‘/g, '&lsquo;')
+    .replace(/’/g, '&rsquo;')
+    .replace(/…/g, '&hellip;');
+}
--- a/packages/server/src/fhir/operations/patientsummary.ts
+++ b/packages/server/src/fhir/operations/patientsummary.ts
@@ -153,6 +153,9 @@ export async function getPatientSummary(
   return builder.build();
 }
 
+export type ResultResourceType = DiagnosticReport | Observation;
+export type PlanResourceType = CarePlan | Goal | ServiceRequest;
+
 /**
  * Builder for the Patient Summary.
  *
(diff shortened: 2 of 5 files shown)


packages/server/src/fhir/operations/patientsummary.ts
```diff
@@ -287,41 +298,12 @@ export class PatientSummaryBuilder {
       case 'vital-signs':
         this.vitalSigns.push(obs);
         break;
-      case 'imaging':
```
codyebberson, line 290: Before, we were trying to be clever about putting `Observation` resources into different sections.  It turns out that wasn't correct, and it would make the summary tables a lot more complicated, so I'm pulling back on that.

packages/server/src/fhir/operations/patientsummary.ts
```diff
@@ -267,9 +281,6 @@ export class PatientSummaryBuilder {
       case 'Observation':
         this.chooseSectionForObservation(resource);
         break;
-      case 'Task':
```
codyebberson, line 270: My original plan was to use `Task` resources for a lot of the "Plan of Treatment" data.  I changed course on that, and settled on `CarePlan` + `ServiceRequest`, which appears to be more aligned with the IPS Patient Summary recommendations.

packages/server/src/fhir/operations/patientsummary.ts
```diff
@@ -342,17 +324,17 @@ export class PatientSummaryBuilder {
...
-        createSection(LOINC_MEDICATIONS_SECTION, 'Medications', this.medications),
-        createSection(LOINC_PROBLEMS_SECTION, 'Problem List', this.problemList),
-        createSection(LOINC_RESULTS_SECTION, 'Results', this.results),
-        createSection(LOINC_SOCIAL_HISTORY_SECTION, 'Social History', this.socialHistory),
-        createSection(LOINC_VITAL_SIGNS_SECTION, 'Vital Signs', this.vitalSigns),
-        createSection(LOINC_PROCEDURES_SECTION, 'Procedures', this.procedures),
-        createSection(LOINC_DEVICES_SECTION, 'Devices', this.devices),
-        createSection(LOINC_ASSESSMENTS_SECTION, 'Assessments', this.assessments),
-        createSection(LOINC_PLAN_OF_TREATMENT_SECTION, 'Plan of Treatment', this.planOfTreatment),
+        this.createAllergiesSection(),
```
codyebberson, line 327: Each section now has its own logic for building the "summary" table in the narrative section.

packages/core/src/utils.ts
```diff
@@ -1404,3 +1404,21 @@ export function flatMapFilter<T, U>(arr: T[] | undefined, fn: (value: T, idx: nu
   }
   return result;
 }
+
+/**
+ * Returns the escaped HTML string of the input string.
+ * @param unsafe - The unsafe HTML string to escape.
+ * @returns The escaped HTML string.
+ */
+export function escapeHtml(unsafe: string): string {
```
codyebberson, line 1413: This function already existed in the `generator` package.  Here I'm moving it from `generator` to `core` so we can access it from `server`.

If we're concerned about bloating `core`, then we could also just duplicate this in `server`.

ThatOneBro: It's fine to put it in core I think 👍

ThatOneBro (approved):

rahul1 (approved):

Issue #6161, opened 2025-03-12
PlanDefinition $apply for ServiceRequest and Procedure


codyebberson:
Sample ActivityDefinition for ServiceRequest:

```json
{
  "resourceType": "ActivityDefinition",
  "id": "example-lab-order",
  "status": "active",
  "kind": "ServiceRequest",
  "name": "CompleteBloodCountOrder",
  "title": "Complete Blood Count Order",
  "description": "Order for a complete blood count",
  "code": {
    "coding": [
      {
        "system": "http://snomed.info/sct",
        "code": "26604007",
        "display": "Complete blood count"
      },
      {
        "system": "http://www.ama-assn.org/go/cpt",
        "code": "85025",
        "display": "Complete CBC with automated differential WBC"
      }
    ]
  },
  "intent": "order",
  "priority": "routine",
  "participant": [
    {
      "type": "practitioner",
      "role": {
        "coding": [
          {
            "system": "http://terminology.hl7.org/CodeSystem/practitioner-role",
            "code": "doctor",
            "display": "Doctor"
          }
        ]
      }
    }
  ]
}
```

Sample output ServiceRequest:

```json
{
  "resourceType": "ServiceRequest",
  "id": "generated-from-activity-definition",
  "status": "draft",
  "intent": "order",
  "priority": "routine",
  "code": {
    "coding": [
      {
        "system": "http://snomed.info/sct",
        "code": "26604007",
        "display": "Complete blood count"
      },
      {
        "system": "http://www.ama-assn.org/go/cpt",
        "code": "85025",
        "display": "Complete CBC with automated differential WBC"
      }
    ]
  },
  "subject": {
    "reference": "Patient/example",
    "display": "John Smith"
  },
  "requester": {
    "reference": "Practitioner/example",
    "display": "Dr. Jane Doe"
  },
  "performer": [
    {
      "reference": "Organization/lab",
      "display": "Clinical Laboratory"
    }
  ],
  "authoredOn": "2025-03-12T14:30:00Z",
  "occurrenceDateTime": "2025-03-12T14:30:00Z",
  "reasonCode": [
    {
      "coding": [
        {
          "system": "http://snomed.info/sct",
          "code": "386661006",
          "display": "Fever"
        }
      ]
    }
  ],
  "definitionUri": "ActivityDefinition/example-lab-order"
}
```

Sample ActivityDefinition for Procedure:

```json
{
  "resourceType": "ActivityDefinition",
[...]

PR #6187, opened 2025-03-18
$apply PlanDefinition for ActivityDefinition ServiceRequest

techdavidy:
handles https://github.com/medplum/medplum/issues/6161
Creates a `ServiceRequest` from a `ActivityDefinition` during /PlanDefinition/[id]/$apply.

Procedure was not handled as ActivityDefinition.kind does not support it.
https://hl7.org/fhir/valueset-request-resource-types.html

### DIFF
--- a/packages/server/src/fhir/operations/plandefinitionapply.ts
+++ b/packages/server/src/fhir/operations/plandefinitionapply.ts
@@ -149,7 +149,7 @@ async function createActivityDefinitionTask(
         subject: subject,
         requester: requester as ServiceRequest['requester'],
         encounter: encounter,
-        code: activityDefinition.code
+        code: activityDefinition.code,
       });
 
       return createTask(repo, requester, subject, action, encounter, [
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/operations/plandefinitionapply.ts
```diff
@@ -119,6 +123,53 @@ async function createQuestionnaireTask(
...
+        {
+          type: {
+            text: 'ServiceRequest',
+          },
+          valueReference: {
+            display: action.title,
+            reference: getReferenceString(serviceRequest),
+          },
+        },
+      ]);
```
codyebberson, line 165: I don't think we need to create a `Task` for the `ServiceRequest` - the `ServiceRequest` should be viable on its own.

```suggestion
      return {
        title: action.title,
        resource: createReference(serviceRequest),
      };
```

techdavidy: The idea behind it is to track status similar to other tasks, add notes, due date, etc.
Screenshot 2025-03-17 at 6 17 05 PM

techdavidy: and it can record the output. e.g  "procedure"

codyebberson: Got it.  Claude says this is ok ✅

codyebberson (approved):

PR #6193, opened 2025-03-19
pre-deploy and post-deploy migrations

mattlong:
Based on https://github.com/medplum/medplum/discussions/6130

There's obviously a lot going on here; a few things to call out that should help in reviewing this:

We attempt to isolate dependence of BullMQ features and functionality as much as was reasonable from the logic of post-deploy migration jobs. Besides generally being good for separation of concerns, my hand was forced on this since BullMQ gets mocked (see `packages/server/src/__mocks__/bullmq.ts`) during tests.

There is documentation for the `FIRST_BOOT` special handling in `packages/server/src/migrations/migration-versions.ts` that you won't want to miss. Eventually another pass at centralizing/expanding the documentation for the new migration system will be merited.

The Queue Registry is still not fully being taken advantage of (e.g. the `get<Some>Queue` functions are still being used which would ideally be replaced by use of the registry). Adding the registry wasn't just for the sake of wanting to write something with the word "registry" in it to sound cool; it was to facilitate the handling of Reindex jobs on both the reindex and the post-deploy-migration queue, particularly when needing to requeue a job for various reasons.

During testing, the `DatabaseMigration.dataVersion` column is a very important shared resource that should NOT be changed around willy-nilly since it can cause `initAppServices` to fail depending on configuration. As such, all tests that attempt to change the `dataVersion` column were colocated to `packages/server/src/database-migration.test.ts` where the `markPostDeployMigrationCompleted` function is mocked out to prevent the `dataVersion` column from actually being changed. That's why that test module has such a wide range of tests.


packages/server/src/migrations/data/v2.ts
```diff
@@ -0,0 +1,20 @@
+import { getResourceTypes, WithId } from '@medplum/core';
+import { prepareReindexJobData, ReindexJob, ReindexPostDeployMigration } from '../../workers/reindex';
+import { AsyncJob } from '@medplum/fhirtypes';
+
+export const migration: ReindexPostDeployMigration = {
```
mattlong, line 5: Oops, this migration should not be here yet. Removing...

codyebberson (approved):
Whew 

image

ThatOneBro (approved):
LGTM to me for the most part, good work

One minor-ish concern about swallowing errors but everything else is low-pri

PR #6194, opened 2025-03-19
Add operation chargeitemdefinition $apply

techdavidy:
First pass of the implementation of `ChargeItemDefinition` `$apply` operation. 
The request gets a reference to an existing `ChargeItem` and applies price modifiers. 
Currently just adding `base` price.
The response is the update `ChargeItem` resource.

### DIFF
--- a/packages/server/src/fhir/operations/chargeitemdefinitionapply.test.ts
+++ b/packages/server/src/fhir/operations/chargeitemdefinitionapply.test.ts
@@ -1,0 +1,153 @@
+import { ContentType } from '@medplum/core';
+import { ChargeItem } from '@medplum/fhirtypes';
+import express from 'express';
+import request from 'supertest';
+import { initApp, shutdownApp } from '../../app';
+import { initTestAuth } from '../../test.setup';
+import { loadTestConfig } from '../../config/loader';
+
+const app = express();
+let accessToken: string;
+
+describe('ChargeItemDefinition Apply', () => {
+  beforeAll(async () => {
+    const config = await loadTestConfig();
+    await initApp(app, config);
+    accessToken = await initTestAuth();
+  });
+
+  afterAll(async () => {
+    await shutdownApp();
+  });
+
+  test('Happy path - Apply ChargeItemDefinition', async () => {
+    // 1. Create a patient
+    const patient = await request(app)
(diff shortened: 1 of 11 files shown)


packages/server/src/fhir/operations/chargeitemdefinitionapply.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    const config = await loadTestConfig();
+    await initApp(app, config);
+    accessToken = await initTestAuth();
+  });
+
+  afterAll(async () => {
+    await shutdownApp();
+  });
+
+  test('Happy path - Apply ChargeItemDefinition', async () => {
```
codyebberson, line 23: Nice, this is a good test! 👍 🙏

packages/server/src/fhir/operations/chargeitemdefinitionapply.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param req - The FHIR request.
+ * @returns The FHIR response.
+ */
+export async function chargeItemDefinitionApplyHandler(req: FhirRequest): Promise<FhirResponse> {
+  const ctx = getAuthenticatedContext();
+  const { id } = req.params;
+  const chargeItemDefinition = await ctx.repo.readResource<ChargeItemDefinition>('ChargeItemDefinition', id);
+  const params = parseInputParameters<ChargeItemDefinitionParameters>(operation, req);
+  const inputChargeItemRef = params.chargeItem;
+  const inputChargeItem = await ctx.repo.readReference<ChargeItem>(inputChargeItemRef);
```
codyebberson, line 29: This is interesting/surprising...   I agree with your implementation, this is what the spec says to do.  It's just not what I would have expected.  Create the `ChargeItem`, then apply the `ChargeItemDefinition` to the `ChargeItem`.... 🤔 

Odd

codyebberson (approved):

packages/server/src/fhir/operations/chargeitemdefinitionapply.ts
techdavidy: Yeah we had a convo with @ThatOneBro too. It is surprising.

PR #6197, opened 2025-03-20
fix(subscription): rewrite attachments for subscriptions

ThatOneBro:
Currently subscriptions send back resources that have not had their attachment URLs rewritten to presigned URLs. This is because we currently only rewrite attachments just before we send back a FHIR response when it comes to your average case. Because we take a different path for subscriptions, we send back the Binary reference version of the attachment URLs. This PR makes it so that we rewrite resource attachment URLs just before sending the subscription notification for both WebSocket subscription and rest hook subscription cases.

### DIFF
--- a/packages/server/src/subscriptions/websockets.ts
+++ b/packages/server/src/subscriptions/websockets.ts
@@ -5,12 +5,15 @@ import { JWTPayload } from 'jose';
 import crypto from 'node:crypto';
 import os from 'node:os';
 import ws from 'ws';
+import { getRepoForLogin } from '../fhir/accesspolicy';
 import { AdditionalWsBindingClaims } from '../fhir/operations/getwsbindingtoken';
 import { CacheEntry } from '../fhir/repo';
 import { getFullUrl } from '../fhir/response';
+import { rewriteAttachments, RewriteMode } from '../fhir/rewrite';
 import { heartbeat } from '../heartbeat';
 import { globalLogger } from '../logger';
-import { verifyJwt } from '../oauth/keys';
+import { MedplumBaseClaims, verifyJwt } from '../oauth/keys';
+import { getLoginForAccessToken } from '../oauth/utils';
 import { setGauge } from '../otel/otel';
 import { getRedis, getRedisSubscriber } from '../redis';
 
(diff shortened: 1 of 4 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -248,6 +255,11 @@ export async function addSubscriptionJobs(
   logFn(`Evaluate ${subscriptions.length} subscription(s)`);
 
   const wsEvents = [] as [Resource, string, SubEventsOptions][];
+  const rewrittenResource = await rewriteAttachments(
```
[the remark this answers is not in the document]
ThatOneBro: Good catch. Do we need to do a "on behalf of" for the author of the subscription in the websocket case?

And then that also brings a good question of what context should they be rewritten for rest hook subscriptions 😅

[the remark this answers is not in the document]
codyebberson: I'm trying to remember how we set that up... For websocket subscriptions, we use `meta.author`, correct?

Hm.  I wonder if we need to maintain some additional user/session context in `packages/server/src/subscriptions/websockets.ts` and do the rewrite there 🤔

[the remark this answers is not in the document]
ThatOneBro: Yeah ok I tried to thread the needle through, using the tokens we generate for each subscription, on binding to a token we can associate that token to the websocket + subscription pair and upon iterating over websockets for a given subscription, we can get the token and get the repo for that token, and then use that to rewrite attachments. Seems like that is working now

I added code for rewriting rest hook subs last night too, though it seems like it might be insecure to use system repo to rewrite them 😬, even for rest hooks... should I just remove that and we can come back to that later?

packages/server/src/subscriptions/websockets.ts
```diff
@@ -51,7 +61,35 @@ async function setupSubscriptionHandler(): Promise<void> {
     for (const [resource, subscriptionId, options] of subEventArgsArr) {
       const bundle = createSubEventNotification(resource, subscriptionId, options);
       for (const socket of subToWsLookup.get(subscriptionId) ?? []) {
-        socket.send(JSON.stringify(bundle), { binary: false });
+        // Get the repo for this socket in the context of the subscription
+        const subMetadataMap = wsToSubLookup.get(socket);
+        if (!subMetadataMap) {
+          // We should really never hit this log, this is a true error
+          globalLogger.error('[WS] Unable to find sub metadata map for WebSocket');
```
codyebberson, line 68: Hm.  should we kill the subscription in this case?

packages/server/src/subscriptions/websockets.ts
```diff
@@ -51,7 +61,35 @@ async function setupSubscriptionHandler(): Promise<void> {
...
+          continue;
+        }
+        const subMetadata = subMetadataMap.get(subscriptionId);
+        if (!subMetadata) {
+          // We should really never hit this log, this is a true error
+          globalLogger.error('[WS] Unable to find sub metadata entry for WebSocket', { subscriptionId });
```
codyebberson, line 74: Same here?

packages/server/src/subscriptions/websockets.ts
```diff
@@ -210,30 +270,20 @@ export async function handleR4SubscriptionConnection(socket: ws.WebSocket): Prom
         return;
       }
 
-      let tokenPayload: JWTPayload;
-      try {
-        const { payload } = await verifyJwt(token);
```
codyebberson, line 215: Is there a reason we removed this?

codyebberson (approved):
Broad strokes look good to me.  Some questions about whether we should kill the subscription if it ever gets into a weird state, but maybe that's just defensive coding?

Non-websocket case looks good 👍 

Good to go after test coverage passes ✅

packages/server/src/subscriptions/websockets.ts
ThatOneBro: It moved to the `verifyWsToken` function since we have to call it for `onBind` or `onUnbind` to clean up the function signatures (we needed to pass both the raw token and the token payload, or invoke it twice if we did it there)

packages/server/src/subscriptions/websockets.ts
ThatOneBro: I don't think we should kill the subscription, it could just be that the socket disconnected and for some reason this didn't get cleaned up. We don't support multiple WebSockets binding to the same subscription yet but in the future we could. I would like to see if we even hit this case in the logs first 👀

packages/server/src/subscriptions/websockets.ts
ThatOneBro: Same as above I think, I think this one is maybe even less likely

PR #6199, opened 2025-03-21
CCDA support for more social history observations

codyebberson:


packages/ccda/src/fhir-to-ccda.ts
```diff
@@ -1576,36 +1578,35 @@ class FhirToCcdaConverter {
     const category = observation.category?.[0]?.coding?.[0]?.code;
 
     if (code === LOINC_TOBACCO_SMOKING_STATUS) {
-      // Smoking status
```
codyebberson, line 1579: This was minor cleanup.  Before, we didn't have constants for all of the LOINC codes, so the comments were more necessary.  After the LOINC constants, this is superfluous.

ThatOneBro (approved):

PR #6203, opened 2025-03-21
feat(react): add CCDA validation to `CcdaDisplay`

ThatOneBro:
Example validation within `@medplum/app`:
Screenshot 2025-03-21 at 10 23 56 AM

### DIFF
--- a/packages/react/src/utils/dom.ts
+++ b/packages/react/src/utils/dom.ts
@@ -1,3 +1,4 @@
+import { ContentType } from '@medplum/core';
 import { MouseEvent, SyntheticEvent } from 'react';
 
 /**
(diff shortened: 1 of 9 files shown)


packages/react/src/utils/dom.ts
```diff
@@ -76,3 +77,25 @@ export async function sendCommand<T = string, R = unknown>(frame: HTMLIFrameElem
     frame.contentWindow?.postMessage(command, new URL(frame.src).origin, [channel.port2]);
   });
 }
+
+/**
+ * Creates a Blob object from the JSON object given and downloads the object.
+ * @param jsonString - The JSON string.
+ * @param fileName - Optional file name. Default is based on current timestamp.
+ */
+export function exportJsonFile(jsonString: string, fileName?: string): void {
```
codyebberson, line 86: Huh, didn't know we had this!

There's a very similar method in `packages/react/src/PatientExportForm/PatientExportForm.tsx` - mental note that we can de-dupe.

codyebberson (approved):
Nice, super clean 👍 😎 🙏

PR #6206, opened 2025-03-21
Removed 'any' from fhir operation param utils

codyebberson:
I was doing some work inside `parameters.ts` and got a little lost with `any` params, so I tried to tighten it up by changing `any` to `unknown` everywhere.

It turns out that I was on a dead end (I was going to add `Reference` to stringified GET param parsing, but that's not valid).  So this is all unnecessary.

But maybe the cleanup is nice?  

Should not be any changes to functionality.

### DIFF
--- a/packages/server/src/fhir/operations/utils/parameters.ts
+++ b/packages/server/src/fhir/operations/utils/parameters.ts
@@ -5,6 +5,7 @@ import {
   flatMapFilter,
   isEmpty,
   isResource,
+  isTypedValue,
   validateResource,
 } from '@medplum/core';
 import { FhirRequest } from '@medplum/fhir-router';
@@ -36,7 +37,7 @@ export function parseParameters<T>(input: T | Parameters): T {
  */
 export function parseInputParameters<T>(operation: OperationDefinition, req: Request | FhirRequest): T {
   if (!operation.parameter) {
-    return {} as any;
+    return {} as T;
   }
   const inputParameters = operation.parameter.filter((p) => p.use === 'in');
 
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/operations/utils/parameters.ts
```diff
@@ -145,26 +153,26 @@ function validateInputParam(param: OperationDefinitionParameter, value: any): an
...
 
-    parsed[param.name as string] = validateInputParam(param, value);
+    parsed[param.name] = validateInputParam(param, value);
   }
 
   return parsed;
 }
 
-export function buildOutputParameters(operation: OperationDefinition, output: any): Parameters {
+export function buildOutputParameters(operation: OperationDefinition, output: object | undefined): Parameters {
```
mattwiller, line 175: Should this be `Record<string, unknown>` here instead of `object`, to avoid the cast below?

mattwiller (approved):

packages/server/src/fhir/operations/utils/parameters.ts
codyebberson: So, yes, I would have loved that.  Unfortunately TypeScript doesn't allow you to pass normal `interface` objects as `Record<string, unknown>`, because:

```
Conversion of type 'ConceptMapTranslateOutput' to type 'Record<string, unknown>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Index signature for type 'string' is missing in type 'ConceptMapTranslateOutput'.
```
  
I find that frustrating and silly, but it is what it is?

If we change it to `Record<string, unknown>`, there would be ~50 casts required at various call sites for `as Record<string, unknown>`.

Definitely open to feedback or alternatives.

PR #6208, opened 2025-03-21
Dependency upgrades 2025-03-21

medplumbot:
Dependency upgrades


examples/medplum-nextauth-demo/package.json
```diff
@@ -25,7 +25,7 @@
     "@mantine/notifications": "7.17.2",
     "@medplum/core": "4.0.3",
     "@medplum/eslint-config": "4.0.3",
-    "next": "15.2.2",
+    "next": "15.2.3",
```
ThatOneBro, line 28: The important update which addresses: https://github.com/medplum/medplum/security/dependabot/225

PR #6211, opened 2025-03-22
Update encounter header and status changes

techdavidy:
Screenshot 2025-03-21 at 5 07 49 PM

Initial updates to encounter header.
Include period in/out dates.


codyebberson (approved):
Nice 👍

Issue #4177, opened 2024-03-17
Should FHIRPath support non-standard properties


codyebberson:
A long time ago, we explicitly enabled reading non-standard properties via FHIRPath.  IIRC, the motivating use case was using FHIRPath to read GraphQL properties.

Original PR: https://github.com/medplum/medplum/pull/708

Example:

```ts
  test('GraphQL embedded queries', () => {
    const observations: Observation[] = [
      {
        resourceType: 'Observation',
        code: { coding: [{ code: 'ALB' }] },
        valueQuantity: { value: 120, unit: 'ng/dL' },
      },
      {
        resourceType: 'Observation',
        code: { coding: [{ code: 'HBA1C' }] },
        valueQuantity: { value: 5, unit: '%' },
      },
    ];

    // This is an example of how FHIR GraphQL returns embedded searches.
    // The "ObservationList" is not a real property, but a search result.
    const serviceRequest: ServiceRequest = {
      resourceType: 'ServiceRequest',
      ObservationList: observations,
    } as ServiceRequest;

    const query = "ObservationList.where(code.coding[0].code='HBA1C').value";
    const result = evalFhirPathTyped(query, [toTypedValue(serviceRequest)]);
    expect(result).toEqual([
      {
        type: PropertyType.Quantity,
        value: { value: 5, unit: '%' },
      },
    ]);
  });
```

That works today.

The reason I ask this question is that I recently ran a server profiling exercise, and noticed one of the highest CPU uses is `getTypedPropertyValueWithoutSchema`.  The thing is, `getTypedPropertyValueWithoutSchema` isn't supposed to be called from the server at all, because the server will always have the schema loaded.  `getTypedPropertyValueWithoutSchema` is an escape hatch for client-side code which may not have all `StructureDefinition` resources loaded.

It turns out that we're hitting the `getTypedPropertyValueWithoutSchema` due to the `individual-birthdate` `SearchParameter`.

Consider the input: `{ "resourceType": "Patient", "birthDate": "1990-01-01" }`

Consider the FHIRPath expression: `Patient.birthDate | Person.birthDate | RelatedPerson.birthDate`

Our FHIRPath engine evaluates each branch of the union:

1. `Patient.birthDate`
    1. Evaluates `Patient` - matches the current `resourceType`, and carries on
    2. Evaluates `birthDate` - reads the value
2. `Person.birthDate`
    1. Evaluates `Person` - does not match `resourceType`, so tries to read `Person` as a property (this could be skipped)
[...]

PR #6229, opened 2025-03-24
Use optimized expressions for search parameter FHIRPath evaluation

mattlong:
For more context see https://github.com/medplum/medplum/issues/4177#issuecomment-2004014410

While profiling `ReindexJob`, this optimization was found to yield significant performance improvements for reindexing resources. e.g. locally, reindexing 1,333 `ValueSet` records went from 5.6 seconds to 1.3 seconds.


packages/core/src/search/details.ts
```diff
@@ -251,12 +254,20 @@ export function getExpressionsForResourceType(resourceType: string, expression:
...
-    return undefined;
+    return fhirPathExpression;
+  }
+
+  let result: Atom = atoms[0];
+  for (let i = 1; i < atoms.length; i++) {
+    result = new UnionAtom(result, atoms[i]);
   }
-  return atoms.map((atom) => atom.toString()).join(' | ');
+  return new FhirPathAtom('<original-not-available>', result);
```
mattlong, line 270: We could use `result.toString()` instead of a placeholder value, but that feels a little too self-referential

mattwiller (approved):

codyebberson (approved):
Nice

PR #6230, opened 2025-03-24
Perform batched lookup table operations during reindex

### LINKED ISSUE #6225: Batched version of `LookupTable.indexResource`
Currently lookup table entries are computed and inserted resource-by-resource. A batched code-path could be a nice perf win since it'd reduce a lot of back and forth with the DB.

mattlong:
Fixes #6225


packages/server/src/fhir/lookups/lookuptable.ts
```diff
@@ -47,13 +51,64 @@ export abstract class LookupTable {
...
+
+    const resourceType = resources[0].resourceType;
+
+    if (!create) {
+      await this.batchDeleteValuesForResources(client, resources);
+    }
+
+    const resourceBatchSize = 500;
+    for (let i = 0; i < resources.length; i += resourceBatchSize) {
+      const batch = resources.slice(i, i + resourceBatchSize);
```
ThatOneBro, line 95: Do we need a slice here if we're just going to iterate over a segment of the resources array and create new rows from that segment?

packages/server/src/fhir/lookups/reference.ts
```diff
@@ -99,16 +94,16 @@ function getSearchReferences(resource: WithId<Resource>): ReferenceRow[] {
...
       }
 
       if (isResource(value.value) && value.value.id && isUUID(value.value.id)) {
-        addSearchReferenceResult(result, resource, searchParam, value.value.id);
+        addSearchReferenceResult(resultMap, resource, searchParam, value.value.id);
       }
     }
   }
-  return Array.from(result.values());
+  result.push(...Array.from(resultMap.values()));
```
ThatOneBro, line 106:
```suggestion
  result.push(...resultMap.values());
```
This should avoid an intermediate array

packages/server/src/fhir/lookups/lookuptable.ts
mattlong: The slicing at the resource level was intended to avoid tying up the event loop for too long with synchronous work without any breaks between the calls to `await this.insertValuesForResource(client, resourceType, newRows)` below. 500 was an arbitrary choice and seems a bit aggressive now that I think about it.

ThatOneBro: The conceptual "slicing" is fine but we don't need to create an intermediate array do we?

Instead of `for (const resource of batch) {`

We could do
```
  for (let j = i; j < i + resourceBatchSize; j++) {
    resource = resources[j];
  }
```

Minor optimization but we would create a lot less intermediate arrays which have to get GC'd in quick succession

mattlong: Ah, sorry, I misunderstood your comment. Looking at it again

packages/server/src/fhir/lookups/lookuptable.ts
```diff
@@ -47,13 +51,65 @@ export abstract class LookupTable {
...
+    if (!create) {
+      await this.batchDeleteValuesForResources(client, resources);
+    }
+
+    // Batch at the resource level to avoid tying up the event loop for too long
+    // with synchronous work without any async breaks between DB calls.
+    const resourceBatchSize = 200;
+    for (let i = 0; i < resources.length; i += resourceBatchSize) {
+      const newRows: LookupTableRow[] = [];
+      for (let j = i; j < i + resourceBatchSize && j < resources.length; j++) {
```
ThatOneBro, line 98: 👍 nice

ThatOneBro (approved):
I think we're good to go, let's see the results

PR #6244, opened 2025-03-26
Deploying in digital ocean

dvidsilva:
When loading ENVIRONMENT VARIABLES, it was missing some nested values for `database.ssl` 

If the keys are present, it will load the configuration correctly 

Additionally added docs information about using digital ocean

### DIFF
--- a/packages/docs/docs/self-hosting/digital-ocean.md
+++ b/packages/docs/docs/self-hosting/digital-ocean.md
@@ -1,0 +1,123 @@
+---
+sidebar_position: 9
+---
+
+# Setting Up Medplum on DigitalOcean
+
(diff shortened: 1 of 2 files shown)


dvidsilva:
Thanks! 

Apologies just seeing this. 

Yes please feel free. I'm going to attempt
A new deployment from scratch this weekend because I left it stall too long and having odd problems

dvidsilva:
hi @codyebberson just looked at the tests

realized there was a different format expected, but this should work nonetheless to allow both options

edited the tests to reflect that

codyebberson:
Continuing here: https://github.com/medplum/medplum/pull/9348

PR #6247, opened 2025-03-27
Log requests on close event

mattwiller:
Currently, the logger might not trigger when e.g. a request times out at the load balancer: the client disconnects, so the response is never fully sent to them.  Per the [Node.js `stream.Writeable` docs](https://nodejs.org/docs/latest-v20.x/api/stream.html#class-streamwritable):
> The `'finish'` event is emitted after the [stream.end()](https://nodejs.org/docs/latest-v20.x/api/stream.html#writableendchunk-encoding-callback) method has been called, and all data has been flushed to the underlying system

This seemingly only triggers when the response is sent successfully.  In contrast:

> The `'close'` event is emitted when the stream and any of its underlying resources (a file descriptor, for example) have been closed. The event indicates that no more events will be emitted, and no further computation will occur
This event should trigger the logging in either success or failure modes.


packages/server/src/app.ts
```diff
@@ -251,7 +251,7 @@ const loggingMiddleware = (req: Request, res: Response, next: NextFunction): voi
   const ctx = getRequestContext();
   const start = Date.now();
 
-  res.on('finish', () => {
+  res.on('close', () => {
```
ThatOneBro, line 254: Close is actually probably what we want but it is good to note that it is slightly different than finish in exactly what we're measuring in case we see any increases in our logged request durations

ThatOneBro (approved):

ThatOneBro (approved):

PR #6254, opened 2025-03-28
wip(agent): zero downtime upgrade

### LINKED ISSUE #6171: Investigate (near)-zero downtime agent upgrade
We found that on average there is about ~10 seconds of downtime on HL7 servers hosted by the agent during upgrade. We would like to provide a way for users to have 0 downtime or close to 0 downtime across all agents during upgrades

### TODO:
- [ ] Write-up regarding potential strategies to reduce downtime

ThatOneBro:
Fixes #6171


packages/core/src/version-utils.ts
```diff
@@ -1,7 +1,8 @@
 import { MEDPLUM_VERSION } from './client';
 import { normalizeErrorString } from './outcomes';
 
-export const MEDPLUM_RELEASES_URL = 'https://meta.medplum.com/releases';
+// export const MEDPLUM_RELEASES_URL = 'https://meta.medplum.com/releases';
```
ThatOneBro, line 4: TODO: Fix this after testing

packages/agent/src/app.ts
```diff
@@ -665,10 +734,6 @@ export class App {
     }
 
     try {
-      // Stop the agent to prepare for service being restarted
```
ThatOneBro, line 668: We no longer want to stop the service here since it happens inside the installer itself now

packages/agent/src/app.ts
```diff
@@ -627,17 +676,36 @@ export class App {
       this.log.info(`Forcing upgrade from ${MEDPLUM_VERSION} to ${targetVersion}`);
     }
 
+    if (upgradeInProgress && message.force) {
+      // If running, just cleanup the file since it could be that the cleanup failed for whatever reason
+      if (isAppRunning('medplum-upgrading-agent')) {
+        removePidFile(getPidFilePath('medplum-upgrading-agent'));
```
ThatOneBro, line 682: Need to change this and check all references to `removePidFile` since we simplified the signature

packages/agent/src/main.ts
```diff
@@ -1,12 +1,80 @@
...
     await upgraderMain(argv);
+  } else if (argv[2] === '--stop-old-services') {
+    const logFileFd = openSync(TEMP_LOG_FILE, 'a');
+
+    let allAgentServices: string[] = [];
+    const currentServiceName = `MedplumAgent_${MEDPLUM_VERSION}`;
+
+    while (!allAgentServices.includes(currentServiceName)) {
+      const output = execSync('cmd.exe /c sc query type= service state= all | findstr /i "SERVICE_NAME.*MedplumAgent"');
+      appendFileSync(logFileFd, `${output}\r\n`, { encoding: 'utf-8' });
```
ThatOneBro, line 28: We can clean this up in the `Agent/$fetch-logs` PR

.github/workflows/publish.yml
```diff
@@ -149,7 +149,8 @@ jobs:
         run: npm ci --maxsockets 1
 
       - name: Build
-        run: npm run build -- --filter=@medplum/agent
+        # We need to force so MEDPLUM_VERSION matches, since using cached @medplum/core results in mismatch sometimes
```
codyebberson: 🤔   Why would that be true?

Do we need to add the git has to the turbo.json environment details or something?

packages/agent/installer.nsi
```diff
@@ -42,10 +47,19 @@ Var clientId
...
+Var LineLen
+Var TempLen
+Var CurrentLen
+Var CurrentServiceName
+
 # The onInit handler is called when the installer is nearly finished initializing.
 # See: https://nsis.sourceforge.io/Reference/.onInit
 Function .onInit
-    ReadRegStr $0 HKLM "SYSTEM\CurrentControlSet\Services\${SERVICE_NAME}" "ImagePath"
+    ReadRegStr $0 HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${BASE_SERVICE_NAME}" "DisplayName"
```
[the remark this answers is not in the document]
codyebberson: Hmm...   So the idea here is that we're using the presence of the uninstaller to determine if another version is installed?  Is that reliable?

Should we bite the bullet and just write our own registry key using `BASE_SERVICE_NAME`?

packages/agent/installer.nsi
```diff
@@ -36,22 +37,39 @@ RequestExecutionLevel admin
...
     ${If} $0 != ""
         StrCpy $alreadyInstalled 1
     ${Else}
         StrCpy $alreadyInstalled 0
     ${EndIf}
 
+    ReadRegStr $0 HKLM "SYSTEM\CurrentControlSet\Services\${SERVICE_NAME}" "ImagePath"
+    ${If} $0 != ""
+        StrCpy $alreadyInstalledThisVersion 1
```
codyebberson, line 68: 👍

packages/agent/installer.nsi
```diff
@@ -141,7 +159,10 @@ Section
     DetailPrint "${APP_NAME}"
     SetOutPath "$INSTDIR"
 
-    ${If} $alreadyInstalled == 1
+    ${If} $alreadyInstalledThisVersion == 1
+        DetailPrint "This version is already installed on this system."
```
codyebberson, line 163: 👍

packages/agent/installer.nsi
```diff
@@ -153,27 +174,16 @@ SectionEnd
...
-
-    # Deleting the service
-    DetailPrint "Deleting service..."
-    ExecWait "sc.exe delete ${SERVICE_NAME}" $1
-    DetailPrint "Exit code $1"
-
     # Copy the new files to the installation directory
+    # We temporarily set overwrite to "ifdiff" so that we don't try to overwrite files that may already exist from a previous installation
+    # This should prevent us from trying to overwrite shawl which could still be running, but shouldn't be different if it's the same version
+    SetOverwrite ifdiff
```
codyebberson, line 180: TIL 👍

packages/agent/src/pid.ts
```diff
@@ -82,6 +99,52 @@ export function checkProcessExists(pid: number): boolean {
...
+ */
+export function getAppPid(appName: string): number | undefined {
+  const pidFilePath = getPidFilePath(appName);
+  if (!fs.existsSync(pidFilePath)) {
+    return undefined;
+  }
+  const existingPidStr = fs.readFileSync(pidFilePath, 'utf8').trim();
+  const existingPid = Number.parseInt(existingPidStr, 10);
+  if (Number.isNaN(existingPid)) {
+    throw new Error('PID file does not contain a valid numeric PID');
```
codyebberson: Should we throw or ignore here?

packages/agent/src/pid.ts
```diff
@@ -134,12 +193,23 @@ export function ensureDirectoryExists(directoryPath: string): void {
...
 }
 
+/**
+ * Waits for a given app's PID file to be present in the filesystem.
+ * @param appName - The name of the app associated with the PID file you want to wait for.
+ */
+export async function waitForPidFile(appName: string): Promise<void> {
+  // Wait for agent PID file to exist
+  while (!existsSync(getPidFilePath(appName))) {
+    await sleep(0);
```
codyebberson, line 209: How long do we expect this to spin?

If under 1 sec, then maybe this is ok.

If this could run for multiple seconds, then I think we should consider a 10-100 ms delay, otherwise it could thrash the file system.

packages/core/src/version-utils.ts
codyebberson: Bump this

codyebberson (commented):
Overall, structurally looks good.  Mostly minor comments and questions.

Thanks for fighting through this 👍

packages/agent/src/pid.ts
ThatOneBro: It's usually spinning for on the order of ms, just there's some cases where if you read too fast, the FS hasn't updated the entry for this file yet.

packages/agent/installer.nsi
```diff
@@ -200,11 +210,42 @@ Function UpgradeApp
...
+            DetailPrint "upgrade.json removed, continuing..."
+            ${Break}
+        ${EndIf}
+    ${Loop}
+
+    DetailPrint "Stopping and deleting old Medplum Agent services..."
+    ExecWait "$\"$INSTDIR\${SERVICE_FILE_NAME}$\" --remove-old-services" $1
+    DetailPrint "Exit code $1"
+
+    DetailPrint "Writing installer for upgraded version..."
```
mattlong:
```suggestion
    DetailPrint "Writing uninstaller for upgraded version..."
```

packages/agent/src/dicom.ts
```diff
@@ -154,8 +161,16 @@ export class AgentDicomChannel extends BaseChannel {
     this.started = true;
     const address = new URL(this.getEndpoint().address as string);
     this.log.info(`Channel starting on ${address}`);
-    this.server.on('networkError', (e) => console.log('Network error: ', e));
-    this.server.listen(Number.parseInt(address.port, 10));
+    const port = Number.parseInt(address.port, 10);
+    this.server.on('networkError', async (err) => {
```
mattlong, line 165: If `port` remains in use, it seems like this would loop indefinitely? Is that okay or should there be some max iteration count?

packages/agent/src/pid.ts
mattlong: Should there be some max number of iterations here or duration before this gives up?

packages/agent/installer.nsi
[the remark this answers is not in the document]
ThatOneBro: We could do that but it would not be backwards compatible with versions currently installed in the wild.

One option would be to do a one time check for the uninstall key and then add our own key when the uninstall key is detected but our key is not present, but I think just checking for the uninstaller should be good enough? 🤔 🤷 

What do you think?

packages/agent/src/pid.ts
ThatOneBro: Not sure, in our case these are things that we don't really want to fail per se, because once this spin begins we want it to keep spinning until the other service shuts down. If someone goes to turn off the service manually, I think we would want the second service to be spinning even 5-10 mins later after upgrade to be waiting ready to rebind to the port. Could be convinced otherwise but I ran into that scenario a lot while mucking about.

ThatOneBro: The main case where this would happen would be when you are going to do the first upgrade from non-zero downtime -> zero downtime agent by manually running the installer. If for some reason the second agent fails to stop, then you would want the zero downtime agent to keep trying to bind until the previous agent is stopped

ThatOneBro: I ended up adding a default timeout of 3000 ms, think that should be more than enough for how we use this 👍

packages/agent/src/dicom.ts
ThatOneBro: In our case I think we might want to let it keep looping, since once the channels have been started we are assuming that they will eventually get control...

It seems kinda yucky but I do think this is more robust because even if the old process gets stuck and the service eventually stops 10 minutes later these channels will still be waiting and will immediately bind to the port when available

packages/agent/installer.nsi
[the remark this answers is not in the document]
ThatOneBro: Talked about this offline, we decided preserving this backwards compat is important, especially in the context of downgrading agent versions

PR #6263, opened 2025-03-29
Fixes #6261 - allow React 19 as peer dep

### LINKED ISSUE #6261: Add React 19 to peer dep versions
Both "react" and "react-dom"

All react packages

codyebberson:
Some notes for potential discussion:
* We still had React 17 as a valid peer dep version.  I don't know if that's correct or not.  I'm leaving it alone for now.
* `@medplum/react-hooks` declared `react-dom` as a peerDep, which I think is wrong because it's a pure hooks lib, so I removed it.

For reference

`@mantine/hooks` uses:

```json
      "peerDependencies": {
        "react": "^18.x || ^19.x"
      }
```

`@mantine/core` uses:

```json
      "peerDependencies": {
        "react": "^18.x || ^19.x",
        "react-dom": "^18.x || ^19.x"
      }
```

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -115,8 +115,8 @@
     "@mantine/notifications": "^7.0.0",
     "@medplum/core": "4.0.4",
     "@medplum/react-hooks": "4.0.4",
-    "react": "^17.0.2 || ^18.0.0",
-    "react-dom": "^17.0.2 || ^18.0.0",
+    "react": "^17.0.2 || ^18.0.0 || ^19.0.0",
+    "react-dom": "^17.0.2 || ^18.0.0 || ^19.0.0",
     "rfc6902": "^5.0.1"
   },
   "peerDependenciesMeta": {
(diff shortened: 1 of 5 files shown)


packages/react/package.json
```diff
@@ -115,8 +115,8 @@
     "@mantine/notifications": "^7.0.0",
     "@medplum/core": "4.0.4",
     "@medplum/react-hooks": "4.0.4",
-    "react": "^17.0.2 || ^18.0.0",
-    "react-dom": "^17.0.2 || ^18.0.0",
+    "react": "^17.0.2 || ^18.0.0 || ^19.0.0",
```
ThatOneBro, line 118: Are we sure 17 still works with Mantine? If they requiring 18 or 19? 🤔

ThatOneBro: Oh I just read description. We can leave it for now I guess 👍

ThatOneBro (approved):

packages/react/package.json
codyebberson: Yeah.... We should probably treat that as a separate issue.

In a perfect world, we would have cut React 17 support in Medplum v4.  Oops.

PR #6275, opened 2025-03-31
Log correct status when request times out

mattwiller:


ThatOneBro (approved):
Nice 👍

packages/server/src/app.ts
```diff
@@ -274,7 +274,7 @@ const loggingMiddleware = (req: Request, res: Response, next: NextFunction): voi
       profile: userProfile,
       projectId,
       receivedAt: start,
-      status: res.statusCode,
+      status: res.writableFinished ? res.statusCode : 504,
```
ThatOneBro, line 277: Actually we might want to put a comment above this to explain why if the writable hasn't finished at this point, that means the load balancer has timed out

ThatOneBro (approved):

PR #6276, opened 2025-03-31
Don't try to resolve nonlocal references

### LINKED ISSUE #6272: Regression in 4.0.4 on use of _include with display references: reference missing reference property
This example test will succeed in 4.0.3 but fail in 4.0.4 (see below).

Basically, we are creating a FHIR Resource containing `display` only references and then executing a query with `_include` options on those fields.

In 4.0.3 this correctly (to my understanding), resolves 1 resource. In 4.0.4 this throws and returns an OperationOutcome with `Reference missing reference property`

```
      const eobId = randomUUID();
      const batch = await medplum.executeBatch({
        resourceType: 'Bundle',
        type: 'batch',
        entry: [
          {
            fullUrl: `ExplanationOfBenefit/${eobId}`,
            request: {
              method: 'PUT',
              url: `ExplanationOfBenefit/${eobId}`,
            },
            resource: {
              resourceType: 'ExplanationOfBenefit',
              id: eobId,
              status: 'active',
              type: {
                coding: [{ code: 'claim' }],
              },

              use: 'claim',
              created: '2021-01-01',
              insurer: {
                display: 'Test Insurer',
              },
              patient: {
                display: 'Test Patient',
              },
              provider: {
                display: 'Test Provider',
              },
              outcome: 'complete',
              insurance: [
                {
                  focal: true,
                  coverage: {
                    display: 'Test Insurance',
                  },
                },
              ],
            },
          },
        ],
      });


      expect(batch.entry[0].response?.status).toBe('200');

      const response = await medplum.get(
        '/fhir/R4/ExplanationOfBenefit?_include=ExplanationOfBenefit:encounter&_include=ExplanationOfBenefit:facility&_include=ExplanationOfBenefit:claim&_include=ExplanationOfBenefit:care-team&_include=ExplanationOfBenefit:coverage',
      );

      expect(response.entry.length).toBe(1);
```

mattwiller:
Resolves #6272

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -415,6 +415,11 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     reference: Reference,
     cacheEntry: CacheEntry | undefined
   ): Promise<Resource | Error> {
+    if (!reference.reference?.match(/^[A-Z][a-zA-Z]+\//)) {
+      // Non-local references cannot be resolved
+      return new OperationOutcomeError(notFound);
+    }
+
     try {
       const [resourceType, id] = parseReference(reference);
       validateResourceType(resourceType);
--- a/packages/server/src/fhir/batch.test.ts
+++ b/packages/server/src/fhir/batch.test.ts
@@ -1193,7 +1193,6 @@ describe('Batch and Transaction processing', () => {
       .set('Content-Type', ContentType.FHIR_JSON)
       .send();
 
-    console.log(JSON.stringify(query.body, null, 2));
     expect(query.status).toBe(200);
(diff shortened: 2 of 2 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -415,6 +415,11 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     reference: Reference,
     cacheEntry: CacheEntry | undefined
   ): Promise<Resource | Error> {
+    if (!reference.reference?.match(/^[A-Z][a-zA-Z]+\//)) {
+      // Non-local references cannot be resolved
+      return new OperationOutcomeError(notFound);
+    }
```
jdjkelly, line 421: @mattwiller this makes sense to me

PR #6278, opened 2025-04-01
Dependency upgrades 2025-04-01

medplumbot:
Dependency upgrades


packages/server/package.json
```diff
@@ -62,20 +62,20 @@
...
+    "bullmq": "5.45.2",
     "bytes": "3.1.2",
     "compression": "1.8.0",
     "cookie-parser": "1.4.7",
     "cors": "2.8.5",
     "cron-validator": "1.3.1",
     "dataloader": "2.2.3",
     "dotenv": "16.4.7",
-    "express": "4.21.2",
+    "express": "5.1.0",
```
ThatOneBro, line 78: Looks like express tried to upgrade to v5. I think we'll want to hold this back for now

codyebberson (approved):

Issue #5009, opened 2024-08-01
Add more granular log levels for MedplumClient


jeffrey-peterson-vanna:
When I instantiate the SDK with `MedplumClient({ verbose: true })`, I get chatty logs that include the bearer token. If I do `verbose: false` I get no logs. Additional intermediate log levels would be very useful, especially in contexts where logs are aggregated and persisted. A log level that simply output the following would be great:
```
> GET https://api.medplum.com/fhir/R4/Patient
< 200 OK
```

dillonstreator:
@jeffrey-peterson-vanna this should be possible by overriding the `fetch` method on the `MedplumClient` to hook into outgoing requests:
```ts
const medplum = new MedplumClient({
  fetch: async (input: RequestInfo, init?: RequestInit): Promise<Response> => {
    const response = await fetch(input, init)
    console.log(`${init?.method ?? 'GET'} ${input}`, response.status)
    return response
  },
})
```

AldrinSeanPereira:
Can I get assigned to this issue?
If yes, could you also guide me where I can find the `MedplumClient` code. 
Thanks

reshmakh:
Is this one ok to close?

PR #6288, opened 2025-04-02
Consumer control over logger

dillonstreator:


packages/core/src/websockets/reconnecting-websocket.ts
```diff
@@ -416,9 +414,7 @@ export class ReconnectingWebSocket<WS extends IWebSocket = WebSocket>
   }
 
   private _debug(...args: unknown[]): void {
-    if (this._options.debug) {
```
ThatOneBro, line 419: Without this guard, the debug logs can get extremely chatty. I don't think we want them on by default. 

I think we should keep the debug option

packages/core/src/logger.ts
```diff
@@ -107,7 +116,7 @@ export class Logger {
         level: LogLevelNames[level],
         timestamp: new Date().toISOString(),
         msg: this.prefix ? `${this.prefix}${msg}` : msg,
-        ...data,
+        ...(data && typeof data === 'object' ? data : { data }),
```
ThatOneBro, line 119: If data is an array this will still merge it into the main object:
Screenshot 2025-04-02 at 9 11 06 AM

packages/core/src/logger.ts
```diff
@@ -76,23 +85,23 @@ export class Logger {
     return { write, metadata, level, options };
   }
 
-  error(msg: string, data?: Record<string, any> | Error): void {
+  error(msg: string, data?: Record<string, any> | unknown[] | unknown | Error): void {
```
ThatOneBro, line 88: Any reason we changed the type signature here? What use cases do you need to pass an array?

packages/core/src/client.ts
```diff
@@ -896,7 +903,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
 
     if (this.storage.getInitPromise === undefined) {
       if (!options?.accessToken) {
-        this.attemptResumeActiveLogin().catch(console.error);
+        this.attemptResumeActiveLogin().catch(this.logger.error);
```
ThatOneBro, line 906: This is slightly different, as errors in console will no longer show up as console errors but just regular log statements. If we can find away to preserve the mapping to `console.{log, warn, error}` that would be ideal

packages/core/src/client.ts
```diff
@@ -3317,7 +3324,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
       try {
         body = await response.json();
       } catch (err) {
-        console.error('Error parsing response', response.status, err);
+        this.logger.error(`Error parsing response ${response.status}`, err instanceof Error ? err : undefined);
```
ThatOneBro, line 3327: I think this is also slightly different with `Logger`, since the properties of error will attempt to be splatted out into the object that is logged

packages/core/src/websockets/reconnecting-websocket.ts
```diff
@@ -164,8 +165,7 @@ export type Options<WS extends IWebSocket = WebSocket> = {
   maxRetries?: number;
   maxEnqueuedMessages?: number;
   startClosed?: boolean;
-  debug?: boolean;
-  debugLogger?: (...args: any[]) => void;
+  logger?: ILogger;
```
ThatOneBro, line 168: Can we avoid this breaking change?

Core is not the only consumer of this class

ThatOneBro (commented):
I definitely like the spirit of the PR!

Some minor feedback about breaking changes

And also some open questions about how we want to handle preserving how errors and warning are displayed in the browser console

Thank you @dillonstreator for putting this together, we definitely need to allow users to configure logging more and I think your general approach is probably correct

packages/core/src/websockets/reconnecting-websocket.ts
ThatOneBro: We might want to support both `logger` and set `debugLogger` to `logger` if `logger` exists but `debugLogger` wasn't passed in

packages/core/src/websockets/reconnecting-websocket.ts
ThatOneBro: If the issue is that it's not exposed to `MedplumClient` we might want to add a `debug` option to `MedplumClient` too. I know we have an open issue for adding more verbose logging. Maybe the flag could be `verbose` in `MedplumClient` constructor?

https://github.com/medplum/medplum/issues/5009

PR #6280, opened 2025-04-01
feat(server): add `qualification-code` search param for `Practitioner` 

ThatOneBro:
Fixes #6277

PR #6155, opened 2025-03-12
Add Group:characteristic-reference search parameter

mattwiller:
Resolves #5837

PR #6291, opened 2025-04-02
Tokens in main resource tables

### LINKED ISSUE #5676: Handle token search parameters in main resource tables
For query performance and minimizing transactional serialization errors reasons, we intend on moving away from/deprecating the `<Resource>_Token` tables and instead adding columns to the corresponding main `<Resource>` tables to handle searching/filtering by token search parameters.

### Related projects (completed)
To facilitate this project, improvements were made to the `migrate.ts` schema migration generation script. The most of important of which are being able to declaratively define expression-based indexes, declare multiple indexes/columns per `SearchParameter`, and handle derived search parameters. This work has already been shipped mostly in #5686.

Some other closely related changes are to allow more easily defining and using various search parameter implementation strategies and moving code around such that there is a single source of truth for which implementation is being used. This was done in #5672.

### Writing resources and indexing strategy

The proposed new strategy for handling token search parameters consists of adding an array column per search parameter to the `<Resource>` table of type `TEXT[] NOT NULL DEFAULT ARRAY[]::TEXT[]`. Several derived strings per token are added to the array to facilitate the various query patterns possible on tokens.

```ts
interface Token {
  readonly code: string;
  readonly system: string | undefined;
  readonly value: string | undefined;
}

// These code points are reserved and cannot appear in token `system` or `value` values.
const DELIM = '\x01';
const NULL_SYSTEM = '\x02';

// Simplified example code to be added in the `buildColumn()`
// method of `packages/server/src/fhir/repo.ts`
const tokenSet = new Set<string>();
for (const token of buildTokensForSearchParameter(allTokens, resource, searchParam)) {
  const { code, system, value } = token;
  if (system) {
    // [parameter]=[system]|
    tokenSet.add(system);

    if (value) {
      // [parameter]=[system]|[code]
      tokenSet.add(system + DELIM + value);
    }
  }

  if (value) {
    // [parameter]=[code]
    tokenSet.add(DELIM + value);
  
    if (!system) {
      // [parameter]=|[code]
      tokenSet.add(NULL_SYSTEM + DELIM + value);
    }
  }
}

// populated the token's array column
columns[details.columnName] = Array.from(rowTokens);
```

To handle queries efficiently, two indexes are created per token column where `token_array_to_text` is defined as below using another reserved code point as a delimiter:

```sql
CREATE OR REPLACE FUNCTION token_array_to_text(text[])
 RETURNS text
 LANGUAGE sql
 IMMUTABLE
AS $function$SELECT e'\x03'||array_to_string($1, e'\x03')||e'\x03'$function$
```

`"<ResourceType>_<columnName>_idx" gin (<columnName>)`

A GIN index facilitates simple inclusion queries for the various string combos of `system` and `value` shown above in the example code.

`"<ResourceType>_<columnName>Trgm_idx" gin (token_array_to_text(<columnName>) gin_trgm_ops)`

A trigram GIN index facilitates `:text` and `:contains` partial match search operators using `LIKE`, `ILIKE`, and regex Postgres operations.

### Sorting by token values

[R4 spec behavior](https://www.hl7.org/fhir/r4/search.html#_sort):
> A search parameter can refer to an element that repeats, and therefore there can be
multiple values for a given search parameter for a single resource. In this case,
the sort is based on the item in the set of multiple parameters that comes earliest in
the specified sort order when ordering the returned resources.

In [R5](https://www.hl7.org/fhir/r5/search.html#_sort) and beyond, that language is replaced with:
> Servers have discretion on the implementation of sorting for both repeated elements and complex
elements. For example, if requesting a sort on Patient.name, servers might search by family name
then given, given name then family, or prefix, family, and then given. Similarly, when sorting with
multiple given names, the sort might be based on the 'earliest' name in sort order or the first name
in the instance.

With the implementation as described above, sorting would be done over the first value found in the token array column using a Postgres expression like `substring(token_array_to_text(<columnName>), <needle>)` which can result in incorrect result ordering when a resource has multiple token values for the same code. We could consider sorting the order of entries in the array column, split, reverse, and join the array entries back together to correctly handle ascending and descending sorts, but that would be very computationally, very likely prohibitively, expensive at query time.

Alternatively, to correctly handle ascending and descending sorts performantly, two additional `text` columns could be added per search parameter, e.g. `<code>Asc` and `<code>Desc` for each token search parameter that contain the lexically minimum and maximum values respectively. 

### Supporting the `:contains` search modifier

Since trigram indexes are somewhat expensive relative to the amount data being indexed, we should ideally only include tokens that require partial matching in a trigram index. By [the spec](https://hl7.org/fhir/r4/search.html#token), only the `:text` modifier needs to be supported which would cover a subset of token values. However, Medplum [also supports](https://github.com/medplum/medplum/blob/d4b408d7dfa3770453f1afc50c7540a2245bb3c5/packages/server/src/fhir/lookups/token.ts#L554-L556) the `:contains` search modifier on token values even though that is not part of the FHIR spec. To continue supporting `:contains` searches on tokens in a performant way, *all* tokens need to be included in the trigram index.

If we choose NOT to optimize support for `:contains` on arbitrary tokens, we should alter the plan described above by creating an additional column that only contains tokens that would be considered by `:text` searches and ONLY create a trigram index on that column.

### LINKED ISSUE #6271: Implement double-writing for search parameters affected by FHIRpath toString() change
## Background

PR #6266 fixed parentheses handling in FHIRpath's toString() methods for infix operators. This revealed that we use Atom.toString() in getSearchParameterDetails() when finding ElementDefinitions, which affects how repo.ts determines search strategies.

As a result, 41 search parameters are switching from "column" strategy to "lookup-table" strategy. While no DB migration is required, reindexing is necessary, and some critical parameters like "MedicationRequest-code" and "Observation-value-concept" are affected.

## Proposed Solution

In PR https://github.com/medplum/medplum/pull/6266 we did the following:
1. Fixed the Atom.toString() bug
2. Added a special case for the affected search parameters

This issue addresses step 3 of our plan:
- Implement double-writing for the affected search parameters
- This will allow both "column" and "lookup-table" strategies to work during the transition period

## Implementation Details
- Modify the indexing process to write to both column and lookup table for the 41 affected parameters
- Update search query execution to check both locations
- Add metrics/monitoring to track usage of both paths
- Add tests to verify that both strategies return identical results

## Future Work
Once double-writing is in place and validated, we'll address this conclusively during the token-table cleanup project (see issue https://github.com/medplum/medplum/issues/5676), where we'll:
1. Complete the transition to the lookup-table strategy
2. Remove the special case handling
3. Stop the double-writing

## Affected Search Parameters

https://docs.google.com/spreadsheets/d/1OkA5MbJQkFwdEl8MHE04FFEr5IjEWIbq-BmbatZai18/edit?gid=302653367#gid=302653367

Image

mattlong:
Fixes #5676. See that issue for more context on what's going on here.

Some things to callout that may help the review process:

Right now, the usage of the new token columns in controlled by `TokenColumnsFeature.{write,read}` in `packages/server/src/fhir/tokens.ts`. Those values are setup to be changeable in code to facilitate testing both code paths. `TokenColumnsFeature` is NOT intended to be a long-lived "feature flag"; following our newly adopted post-deploy migration policy/guidelines, writing to the new columns is intended to remain on indefinitely following the merging of this PR. `TokenColumnsFeature.read` will be set to `true` once the eventual `4.2.1` release PR is put together which will be the `requiredVersion` for the two post-deploy migrations added in this PR.

Related to feature flagging, we'll likely want to add a way to set `TokenColumnsFeature.read` on a project-by-project basis to facilitate smoke testing BEFORE `4.2.1` gets published.

This PR should NOT be merged until:
- [x] https://github.com/medplum/medplum/pull/6266 is merged
- [x] Step 3 of #6271 is accounted for in this PR.

More blockers from offline discussion:
- [x] Add `__` prefix to all new token columns
- [x] Resolve the two `//TODO{mattlong}` comments
- [x] Merge #6280 
- [x] Merge #6155 
- [x] Adjust `Repository.VERSION` in this PR based on increments to it in #6280 and #6155

Fixes #6271


packages/server/src/fhir/token-column.ts
```diff
@@ -0,0 +1,318 @@
...
+    'ARRAY_CONTAINS_SUBQUERY',
+    aggregatedQ,
+    'TEXT[]'
+  );
+
+  return cond;
+}
+
+export function escapeRegexString(str: string): string {
+  // TODO{mattlong} - validate this is correct; it is from an LLM
```
codyebberson, line 302: Do we need to fill this in?

packages/server/src/migrations/migrate-functions.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type MigrationAction = { name: string; durationMs: number };
+
+export async function query(client: PoolClient, actions: MigrationAction[], queryStr: string): Promise<void> {
+  const start = Date.now();
+  await client.query(queryStr);
+  actions.push({ name: queryStr, durationMs: Date.now() - start });
+}
+
+export async function idempotentCreateIndex(
```
ThatOneBro, line 11: Do you mind mentioning why this is necessary too?

Just for future readers

packages/server/src/migrations/data/v2.ts
```diff
@@ -0,0 +1,20 @@
...
+export const migration: ReindexPostDeployMigration = {
+  type: 'reindex',
+  prepareJobData(asyncJob: WithId<AsyncJob>) {
+    return prepareReindexJobData(
+      getResourceTypes().filter((rt) => rt !== 'Binary'),
+      asyncJob.id,
+      undefined,
+      // Repository.VERSION was bumped to 2 for token-column search parameters,
+      // so reindex all resources with a lower version.
+      1
```
ThatOneBro, line 14: I think we might need to tweak this based on the other PRs now

packages/server/src/fhir/tokens.ts
```diff
@@ -0,0 +1,338 @@
...
+ * - Observation-value-concept
+ *
+ * Most others are rarely-used "usageContext" parameters.
+ *
+ * DO NOT MODIFY THIS LIST without coordinating with the team responsible for search parameter
+ * implementation. Any changes may require database reindexing.
+ *
+ * See follow-up issue: https://github.com/medplum/medplum/issues/6271
+ */
+const legacyTokenColumnSearchParamResourceTypeAndCodes = new Set([
```
ThatOneBro, line 292: 👀 👍

packages/server/src/fhir/repo.ts
```diff
@@ -1485,30 +1492,49 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+        columnImpl = getSearchParameterImplementation(resource.resourceType, searchParam, true);
+      }
+    } else {
+      impl satisfies ColumnSearchParameterImplementation;
+      columnImpl = impl;
     }
 
-    columns[impl.columnName] = columnValue;
+    if (columnImpl) {
+      let columnValue = null;
```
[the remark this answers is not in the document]
ThatOneBro: Small nit

ThatOneBro (commented):
Looks good, some small requests for clarification but overall seems sound

codyebberson (approved):

PR #6300, opened 2025-04-03
Calculate total price of all charge items

techdavidy:
Screenshot 2025-04-03 at 3 31 03 PM


examples/medplum-provider/src/pages/components/ChargeItem/ChageItemPanel.tsx
```diff
@@ -61,7 +55,7 @@ export default function ChargeItemPanel(props: ChargeItemPanelProps): JSX.Elemen
           </ActionIcon>
         </Flex>
         <Grid columns={12}>
-          <Grid.Col span={10}>
+          <Grid.Col span={12}>
```
codyebberson, line 58: Looks like this is now just a `<Grid>` with one full width `<Grid.Col>`?  Can probably be replaced with `<Box>` or something more appropriate.

PR #6330, opened 2025-04-09
Broken routes from react router v7 upgrade

finnbergquist:


ThatOneBro:
Can we investigate if this is potentially a solution? Looks like the reason that the `Patient/:id` route doesn't match anything is because you still need an index separately from the `*` route

https://github.com/remix-run/react-router/issues/12460#issuecomment-2518973962

finnbergquist:
> Can we investigate if this is potentially a solution? Looks like the reason that the `Patient/:id` route doesn't match anything is because you still need an index separately from the `*` route
> 
> [remix-run/react-router#12460 (comment)](https://github.com/remix-run/react-router/issues/12460#issuecomment-2518973962)

Yea, that works too. Do you think that the argument for doing it this way would be to make the routes more flexible such that you don't need to continue adding new sub-routes in App.tsx? 

Ultimately, I think this is a problem in a bunch of the other example apps as well.

ThatOneBro:
Yeah I think it's mostly for the case where routing is being handled within the component itself (for example, `PatientPage`) so if I add a new tab I shouldn't need to also add a new route mapping back to the same component. I think the migration was a find and replace, we probably just need to go back and make sure they all have an index where the `*` path occurs

PR #6336, opened 2025-04-10
Handle missing migration files

mattlong:
As would happen when a new post-deploy migration is first deployed but old workers from the previous deploy are still running and consuming jobs

### DIFF
--- a/packages/server/src/workers/post-deploy-migration.ts
+++ b/packages/server/src/workers/post-deploy-migration.ts
@@ -1,6 +1,6 @@
 import { getReferenceString, WithId } from '@medplum/core';
 import { AsyncJob, Parameters } from '@medplum/fhirtypes';
-import { Job, JobsOptions, Queue, QueueBaseOptions, Worker } from 'bullmq';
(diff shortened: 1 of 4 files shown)


packages/server/src/workers/post-deploy-migration.ts
```diff
@@ -78,7 +79,27 @@ export async function jobProcessor(job: Job<PostDeployJobData>): Promise<void> {
...
+        `${PostDeployMigrationQueueName} processor delaying job since migration definition was not found on this worker`,
+        {
+          jobId: job.id,
+          delayMs,
+          ...getJobDataLoggingFields(job),
+          version: migrationNumber,
+        }
+      );
+      await job.moveToDelayed(Date.now() + delayMs, job.token);
+      throw new DelayedError('Post-deploy migration delayed since migration definition was not found on this worker');
```
ThatOneBro, line 98: Interesting model for delaying jobs, calling both a method on the job and throwing an error. TIL

https://docs.bullmq.io/patterns/process-step-jobs#delaying

ThatOneBro (approved):
👍

PR #6338, opened 2025-04-10
feat(docker): add `@medplum/app` Dockerfile

### LINKED ISSUE #6332: Add Dockerfile for `@medplum/app`

ThatOneBro:
Also adds publishing of Medplum app Dockerfile on publish and deploy, similar to Medplum server.

Closes #6332


packages/app/docker-entrypoint.sh
```diff
@@ -0,0 +1,71 @@
...
+# Defaults taken from .env.defaults in Medplum monorepo packages/app
+DEFAULT_MEDPLUM_BASE_URL=http://localhost:8103/
+DEFAULT_MEDPLUM_CLIENT_ID=
+DEFAULT_GOOGLE_CLIENT_ID=
+DEFAULT_RECAPTCHA_SITE_KEY=6LfHdsYdAAAAAC0uLnnRrDrhcXnziiUwKd8VtLNq
+DEFAULT_MEDPLUM_REGISTER_ENABLED=true
+DEFAULT_MEDPLUM_AWS_TEXTRACT_ENABLED=true
+
+# Find all JS files in the assets directory
+for file in $(find /usr/share/nginx/html/assets -name "*.js"); do
```
codyebberson, line 15: This can be done in a single command, which will be a little faster and cleaner

See: https://github.com/medplum/medplum/blob/main/scripts/build-deb.sh#L270-L278

And you can avoid the `if`/`then`/`else` statements by just setting the default value:

```sh
${ENV_VAR:=default_value}
```

.github/workflows/deploy.yml
```diff
@@ -58,7 +58,8 @@ jobs:
         env:
           APP_BUCKET: ${{ secrets.APP_BUCKET }}
           AWS_REGION: ${{ secrets.AWS_REGION }}
-          DOCKERHUB_REPOSITORY: ${{ secrets.DOCKERHUB_REPOSITORY }}
+          SERVER_DOCKERHUB_REPOSITORY: ${{ secrets.SERVER_DOCKERHUB_REPOSITORY }}
```
codyebberson, line 61: 🤔 

I understand the push for clarity, but slight preference to keep the original `DOCKERHUB_REPOSITORY` name as-is to reduce regression risk.

packages/app/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM nginxinc/nginx-unprivileged:alpine
```
codyebberson, line 1: Hm.  Trying to understand which one is the right base image.  I would have assumed it's this: https://hub.docker.com/_/nginx

ThatOneBro: Yeah the problem is that image requires to run as root which is flagged by Sonar and probably other security tools

ThatOneBro: I originally tried with that image. This is the official unprivileged nginx Docker image

ThatOneBro:
Need to add the `APP_DOCKERHUB_REPOSITORY` secret before merging: `medplum/medplum-app`

PR #6342, opened 2025-04-10
Fixing broken routes in example apps from react-router v7 upgrade

finnbergquist:
### Change required for example apps routing to work with v7 react-router

Everything that follows this pattern
```
<Route path="/:someResource/:id">
  <Route path="*" element={<SomePage />} />
</Route>
```
throws a `/<someResource>/<id> does not have an element or Component in the console and the page never mounts.`
This can be fixed by adding the index route
```
<Route path="/:someResource/:id">
  <Route index element={<TaskTab />} />
  <Route path="*" element={<SomePage />} />
</Route>
```

Reference to discussion about this in react-router/issues [here](https://github.com/remix-run/react-router/issues/12460#issuecomment-2518973962)


ThatOneBro (approved):
Nice 👍

PR #6361, opened 2025-04-14
Dependency upgrades 2025-04-14

medplumbot:
Dependency upgrades


packages/agent/src/dicom.test.ts
```diff
@@ -13,7 +13,7 @@ let endpoint: Endpoint;
 describe('DICOM', () => {
   beforeAll(async () => {
     console.log = jest.fn();
-    dimse.log.transports.forEach((t) => (t.silent = true));
+    dimse.log.disableAll(false);
```
codyebberson, line 16: Nice 👍 

I saw the build error and got scared that it was going to be much more complicated than this

codyebberson (approved):
sweet, ty @ThatOneBro 👍 🙏

PR #6366, opened 2025-04-15
perf(seed): batch writes to database during seed

ThatOneBro:
Total (in-process) startup time went from ~40 seconds to ~15 seconds

Screenshot 2025-04-14 at 11 29 28 PM
Screenshot 2025-04-14 at 11 29 38 PM


ThatOneBro:
Closing due to bitrot, still might be worth looking into it again later

PR #6372, opened 2025-04-15
Claim $export CMS 1500 pdf 

techdavidy:
Expose  `/fhir/R4/Claim/${claim.id}/$export` to generate  CMS 1500 pdf


packages/server/src/fhir/operations/claimexport.ts
```diff
@@ -0,0 +1,85 @@
...
+      issued: new Date().toISOString(),
+      content: {
+        contentType: 'application/pdf',
+        url: getReferenceString(binary),
+        title: 'cms-1500.pdf',
+      },
+    };
+    return [allOk, media];
+  } catch (error) {
+    return [badRequest(`Error exporting claim: ${(error as Error).message}`)];
```
ThatOneBro, line 83: We have the util `normalizeErrorString` which would be good here too 👍

ThatOneBro: Also can we get coverage for that branch?

ThatOneBro (commented):
Looks good for the most part, some small comments

packages/server/src/fhir/operations/utils/cms1500pdf.ts
```diff
@@ -0,0 +1,406 @@
...
+  const docDefinition: TDocumentDefinitions = {
+    defaultStyle: {
+      font: 'Helvetica',
+      fontSize: 8,
+    },
+    pageSize: { width: PAGE_WIDTH, height: PAGE_HEIGHT },
+    pageMargins: 0,
+    content: [
+      {
+        image: 'data:image/png;base64,' + imageToBase64(__dirname + '/cms1500.png'),
```
ThatOneBro, line 107: I think we can cache this in the module scope since this doesn't change for the lifetime of the server, ie.
```ts
const CMS_1500_IMG_DATA_URL = 'data:image/png;base64,' + imageToBase64(__dirname + '/cms1500.png');
```
```suggestion
        image: CMS_1500_IMG_DATA_URL,
```

ThatOneBro (commented):
Think this optimization is worth making, other than that I think everything is good

packages/server/package.json
```diff
@@ -14,7 +14,7 @@
   "license": "Apache-2.0",
   "author": "Medplum <hello@medplum.com>",
   "scripts": {
-    "build": "npm run clean && tsc && tsc --project tsconfig.build.json",
+    "build": "npm run clean && tsc && tsc --project tsconfig.build.json && node -e \"require('node:fs').cpSync('src/img', 'dist/img', { recursive: true });\"",
```
ThatOneBro, line 17: Added an `img` dir in `src` that we will always copy over. Using `node` to be platform agnostic

codyebberson: Hmm... I don't think we want to copy this into `dist`

We should just add a `public` or `static` directory, and make sure that ends up in the Docker image.

I'd recommend handling this in `scripts/build-docker.sh`

packages/server/src/img/cms1500.png
codyebberson, line 1: I recommend moving this to a new directory such as `/package/server/static`

codyebberson (changes requested):

packages/server/src/fhir/operations/utils/cms1500pdf.ts
```diff
@@ -0,0 +1,410 @@
...
+  'MEDICARE',
+  'MEDICAID',
+  'TRICARE',
+  'CHAMPVA',
+  'GROUP HEALTH PLAN',
+  'FECA BLK LUNG',
+  'OTHER',
+]);
+
+const CMS_1500_IMG_DATA_URL =
```
codyebberson, line 28: We can avoid this.

See: https://pdfmake.github.io/docs/0.1/document-definition-object/images/

We can refer to the static file inline

packages/server/src/pdf/pdf.test.ts
codyebberson, line 1: Recommend moving this to `packages/server/src/util` rather than creating new `pdf` directory.

packages/server/src/pdf/pdf.ts
codyebberson, line 1: Recommend moving this to `packages/server/src/util` rather than creating new `pdf` directory.

packages/server/src/fhir/operations/utils/cms1500pdf.ts
```diff
@@ -0,0 +1,393 @@
...
+  const docDefinition: TDocumentDefinitions = {
+    defaultStyle: {
+      font: 'Helvetica',
+      fontSize: 8,
+    },
+    pageSize: { width: PAGE_WIDTH, height: PAGE_HEIGHT },
+    pageMargins: 0,
+    content: [
+      {
+        image: 'static/cms1500.png',
```
codyebberson, line 105: Actually, I was wrong before, we're going to need to make this a relative path.

When we run on local dev machines, we are typically running from the `server` directory, so `static/cms1500.png` is correct.

When we run on the Docker image, we are running in the project root directory, so it needs to be `packages/server/static/cms1500.png`.

We can deal with that using `join(__dirname, '../../../static/cms1500.png')` or whatever

packages/server/src/fhir/operations/claimexport.ts
```diff
@@ -0,0 +1,85 @@
+import { allOk, badRequest, getReferenceString, normalizeErrorString } from '@medplum/core';
+import { FhirRequest, FhirResponse } from '@medplum/fhir-router';
+import { Binary, Claim, Media, OperationDefinition } from '@medplum/fhirtypes';
+import { getAuthenticatedContext } from '../../context';
+import { createPdf } from '../../util/pdf';
```
codyebberson, line 5: Looks like this file didn't get re-added

image

codyebberson (approved):
Nice, thanks @techdavidy 💪

PR #6386, opened 2025-04-16
Global redis-based feature flag for new token implementation

mattlong:
This feels like I might have gone a little bit overboard and added the starts of a minimal little redis-based feature flag system for Medplum, but having the ability to immediately toggle higher-risk features feels like something we want.


mattlong:
Closing this for now since we don't want to commit to a Redis-based feature flag system just yet. In place of this will be a new server config value to toggle the default read strategy.

PR #6390, opened 2025-04-17
Enable more useful OTel traces

mattwiller:


packages/server/src/async.ts
```diff
@@ -10,7 +10,9 @@ import { NextFunction, Request, RequestHandler, Response } from 'express';
  * @returns Async wrapped handler.
  */
 export function asyncWrap(callback: (req: Request, res: Response, next: NextFunction) => Promise<any>): RequestHandler {
-  return function (req: Request, res: Response, next: NextFunction): void {
+  const fn = function (req: Request, res: Response, next: NextFunction): void {
     callback(req, res, next).catch(next);
   };
+  Object.defineProperty(fn, 'name', { writable: false, value: callback.name });
```
mattlong, line 16: Could you add a comment explaining what's going on here? Namely that this is for the sake of traces so that the name of the `callback` parameter will be used?

packages/server/src/otel/instrumentation.ts
```diff
@@ -60,32 +55,47 @@ export function initOpenTelemetry(): void {
...
-    // new ExpressInstrumentation(),
-    // new GraphQLInstrumentation({
-    //   ignoreTrivialResolveSpans: true, // Don't record simple object property lookups
-    // }),
-    // new DataloaderInstrumentation(),
-
-    // new AwsInstrumentation(),
-    // new AwsLambdaInstrumentation(),
+    new ExpressInstrumentation({
+      ignoreLayers: [
```
mattlong, line 77: Could you add some explanation on why we ignore these layers?

mattlong (approved):
LGTM, just a couple requests for more documentation/comments.

PR #6396, opened 2025-04-17
Add possibly surprising stringify test

mattlong:


codyebberson (approved):
Yeah....

I suppose one thing that is not explicitly stated is that `stringify()` does not do validation.  This behavior is "correct" insofar as that it's doing the right thing assuming that it was valid FHIR.

codyebberson:
> I suppose one thing that is not explicitly stated is that `stringify()` does not do validation. This behavior is "correct" insofar as that it's doing the right thing assuming that it was valid FHIR.

I wonder if we should go further and add a comment on `stringify()` such as "Output behavior is undefined for invalid FHIR inputs"

PR #6399, opened 2025-04-18
fix(agent): no crash on startup when unable to connect

### LINKED ISSUE #6388: Agent crashes at startup if network is down before agent calls `MedplumClient.startClientLogin`

ThatOneBro:
Fixes #6388

### DIFF
--- a/packages/agent/src/agent-main.ts
+++ b/packages/agent/src/agent-main.ts
@@ -1,6 +1,7 @@
-import { MedplumClient, parseLogLevel } from '@medplum/core';
+import { MedplumClient, normalizeErrorString, parseLogLevel, sleep } from '@medplum/core';
 import { existsSync, readFileSync } from 'node:fs';
 import { App } from './app';
+import { RETRY_WAIT_DURATION_MS } from './constants';
 
 interface Args {
   baseUrl: string;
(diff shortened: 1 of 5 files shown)


packages/agent/src/app.ts
```diff
@@ -269,13 +267,10 @@ export class App {
...
-    return new Promise<void>((resolve, reject) => {
-      const connectTimeout = setTimeout(
-        () => reject(new Error('Timeout when attempting to connect to server WebSocket')),
-        10000
-      );
+    return new Promise<void>((resolve) => {
+      const connectInterval = setInterval(() => this.webSocket?.reconnect(), RETRY_WAIT_DURATION_MS);
       this.webSocket?.addEventListener('open', () => {
-        clearTimeout(connectTimeout);
+        clearInterval(connectInterval);
```
codyebberson, line 273: Nice, this makes sense.

packages/agent/src/agent-main.ts
```diff
@@ -46,10 +47,32 @@ export async function agentMain(argv: string[]): Promise<App> {
   const { baseUrl, clientId, clientSecret, agentId } = args;
 
   const medplum = new MedplumClient({ baseUrl, clientId });
-  await medplum.startClientLogin(clientId, clientSecret);
+
+  let loggedIn = false;
+  while (!loggedIn) {
+    try {
+      await medplum.startClientLogin(clientId, clientSecret);
```
codyebberson, line 54: Hm, this might be too broad.  For example, if the client credentials are wrong, do we want to keep retrying?

packages/agent/src/agent-main.ts
```diff
@@ -46,10 +47,32 @@ export async function agentMain(argv: string[]): Promise<App> {
...
+    }
+  }
 
   const app = new App(medplum, agentId, parseLogLevel(args.logLevel ?? 'INFO'));
-  await app.start();
+
+  let started = false;
+  while (!started) {
+    try {
+      await app.start();
```
codyebberson, line 68: Do we know what the conditions are that can fail at this point?

packages/agent/src/agent-main.ts
ThatOneBro: Good point, we might want to only catch the fetch failed error?

packages/agent/src/agent-main.ts
ThatOneBro: Yeah maybe it's actually not a good idea to catch here because there could be real errors we want to throw and stop the service. With restarts it should be more robust to that now anyways

ThatOneBro: In rare cases fetch could fail when attempting to fetch the config from Medplum server, but that should be pretty rare since we basically make sure we are connected in two spots before we get to that line (the call to `App.startWebSocket` just before and also the `startClientLogin` loop we have

packages/agent/src/agent-main.ts
ThatOneBro: Ok after talking offline we decided to keep this since it will likely spin anyways after adding auto-restart of the service... in most cases this will fail due to network being disconnected so this is probably still fine

PR #6411, opened 2025-04-18
Various Reindex job fixes

### LINKED ISSUE #6408: Support worker version to prevent old workers from taking jobs they aren't qualified for
Currently, when a fix goes out for a bullmq job that was failing, workers from the previous deploy tend to pick them up during the transition to workers from the new deploy, destining the job to fail again.

Proposal is to

- add a `VERSION` field to our various queue worker's that gets included in `job.data` when the job is enqueued. 
- when the worker takes a job, if its `VERSION` is lower than the version specified in `job.data`, it puts the job back on the queue

### LINKED ISSUE #6409: ReindexJob needs a higher postgres statement_timeout
This would be largely obviated by starting to use indexes on `__version` column, but until then...

When most or all resources of a large table have already been reindexed, a query will wind up scanning most or all of the table to look for resources needing reindexes; this can take longer than the default `statement_timeout` of 60 seconds.

mattlong:
Fixes #6408 
Fixes #6409

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2570,7 +2570,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     return this.context;
   }
 
-  [Symbol.dispose](): void {
+  [Symbol.dispose](removeConnection?: boolean): void {
     this.assertNotClosed();
     if (this.disposable) {
       if (this.transactionDepth > 0) {
@@ -2579,7 +2579,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         this.releaseConnection(new Error('Closing Repository with active transaction'));
       } else {
         // Good state, return healthy connection to pool
-        this.releaseConnection();
+        this.releaseConnection(removeConnection);
       }
     }
     this.closed = true;
--- a/packages/server/src/workers/reindex.ts
+++ b/packages/server/src/workers/reindex.ts
@@ -9,7 +9,7 @@ import {
   WithId,
 } from '@medplum/core';
 import { AsyncJob, Parameters, ParametersParameter, Resource, ResourceType } from '@medplum/fhirtypes';
-import { DelayedError, Job, Queue, QueueBaseOptions, Worker } from 'bullmq';
+import { Job, Queue, QueueBaseOptions, Worker } from 'bullmq';
 import { getRequestContext, tryRunInRequestContext } from '../context';
 import { DatabaseMode, getDatabasePool } from '../database';
 import { AsyncJobExecutor } from '../fhir/operations/utils/asyncjobexecutor';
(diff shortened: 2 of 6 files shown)


packages/server/src/workers/reindex.ts
```diff
@@ -182,6 +194,20 @@ export class ReindexJob {
       return 'finished';
     }
 
+    const client = await this.systemRepo.getDatabaseClient(DatabaseMode.WRITER);
+    try {
+      await client.query(`SET statement_timeout TO 7200000`); // 2 hours
```
mattwiller, line 199: This seems pretty long for the type of queries I would expect to see in the reindex worker; what is it related to?

mattlong: This if for #6409 for the case when the job is scanning a very large table for resources to reindex, but most/all have already been reindex, so the search winds up scanning/filtering the entire table to find no results with a query like the following:

```sql
SELECT "Task"."id", "Task"."lastUpdated", "Task"."content"
FROM "Task"
WHERE (
  ("Task"."__version" <= 2 OR "Task"."__version" IS NULL)
  AND "Task"."deleted" = false
  AND "Task"."lastUpdated" < '2025-04-19'
ORDER BY "Task"."lastUpdated" LIMIT 501
```

You're right; 2 hours is probably overkill. I'll tone it down to 1 hour for now since I want to make sure the reindex we're trying to get through right now completes without incurring additional timeouts. I've filed an issue (#6412) to further revise down this timeout after seeing how long the queries take for the reindex job at hand.

packages/server/src/fhir/repo.ts
```diff
@@ -2570,7 +2570,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     return this.context;
   }
 
-  [Symbol.dispose](): void {
+  [Symbol.dispose](removeConnection?: boolean): void {
```
ThatOneBro, line 2573: We should note that normally your `[Symbol.dispose]()` should not contain args since the symbol is intended to be used by automatic resource cleanup via [`using` keyword](https://simonkoeck.medium.com/typescript-5-2s-using-keyword-managing-resources-with-ease-e0a31efa3eec)

Eventually we should clean this up

ThatOneBro (approved):
Let's ship it but one comment about cleanup down the line

packages/server/src/fhir/repo.ts
mattlong: Yeah...I didn't love adding that, but I wasn't sure of a better way to achieve it for now. Very open to suggestions!

mattlong: We could add a new method on `Repository` to set the state to remove the connection on disposal I suppose, but I'm not sure how much better that is.

PR #6425, opened 2025-04-22
Let `/admin/super/vacuum` run `ANALYZE` queries

mattlong:
Rounding out the functionality of the endpoint to allow running only `ANALYZE` in addition to `VACUUM` and `VACUUM ANALYZE`.

Added the `--prefer-async` flag to faciliate using the vacuum endpoint via the Medplum CLI.


ThatOneBro (approved):
LGTM

PR #6436, opened 2025-04-23
Stop using token lookup tables

mattlong:
This will be a part of `v4.2.x`


mattlong:
Closing this PR since it needs to be broken up into at two steps:
* `4.2.0` - Stop reads from token tables, but keep writing to them
* `4.3.0` - Stop writing to token tables, delayed post-deploy migration to drop.

PR #6500, opened 2025-04-29
Add Flag.status search parameter

### LINKED ISSUE #6488: Backport R5 search parameters for Flag resource (status and category)
R5 introduced two new search parameters for the Flag resource: status and category. Their utility is self-evident and their absence makes using the Flag resource painful (ex. instead of using the data from our FHIR API hook, we have to build a custom wrapper to client-side filter for Flags to exclude inactive and entered-in-error Flags).

Is it possible to backport these to the Medplum R4 API?

mattwiller:
Backporting from FHIR R5 to support customer use cases

Resolves #6488

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1365,9 +1365,9 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    * 2. 04/09/25 - Added qualification-code search param for `Practitioner` (https://github.com/medplum/medplum/pull/6280)
    * 3. 04/09/25 - Added columns for `token-column` search strategy (https://github.com/medplum/medplum/pull/6291)
    * 4. 04/25/25 - Consider `resource.id` in lookup table batch reindex (https://github.com/medplum/medplum/pull/6479)
-   *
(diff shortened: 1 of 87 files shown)


mattlong:
Could you please bump `Repository.VERSION`? 🙏

packages/server/src/fhir/repo.ts
```diff
@@ -1365,9 +1365,9 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    * 2. 04/09/25 - Added qualification-code search param for `Practitioner` (https://github.com/medplum/medplum/pull/6280)
    * 3. 04/09/25 - Added columns for `token-column` search strategy (https://github.com/medplum/medplum/pull/6291)
    * 4. 04/25/25 - Consider `resource.id` in lookup table batch reindex (https://github.com/medplum/medplum/pull/6479)
-   *
+   * 5. 04/29/25 - Added `status` param for `Flag` resources (https://github.com/medplum/medplum/pull/6500)
```
ThatOneBro, line 1368: 👍

ThatOneBro (approved):
LGTM, thank you @mattwiller !

ThatOneBro:
Oh I just thought about this, do we want to also create a targeted reindex for `Flag`, that one should always be a pretty small index job I would think?

PR #6508, opened 2025-04-29
Use I/O-optimized storage for RDS with SSD available

mattwiller:
AWS RDS Aurora allows using database instance types [with attached NVMe SSD](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.DBInstanceClass.Types.html#Concepts.DBInstanceClass.Types.optimized-reads) storage.  By default, these instances will use the SSD only for temporary storage while executing a query plan (e.g. for sorting or merging data).  However, with [I/O-optimized storage](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Overview.StorageReliability.html#aurora-storage-type) enabled, it will also use the SSD as a tiered cache; that should [further reduce latency](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraPostgreSQL.optimized.reads.html) at a slightly higher overall cost.  When opting into the more expensive SSD-backed instances, it makes sense to ensure they provide the maximum performance relative to price

### DIFF
--- a/packages/cdk/src/backend.ts
+++ b/packages/cdk/src/backend.ts
@@ -17,7 +17,7 @@ import {
   aws_wafv2 as wafv2,
 } from 'aws-cdk-lib';
 import { Repository } from 'aws-cdk-lib/aws-ecr';
-import { ClusterInstance, ParameterGroup } from 'aws-cdk-lib/aws-rds';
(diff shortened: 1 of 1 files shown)


ThatOneBro (approved):
Excited to see how this might boost perf 🤞

PR #6522, opened 2025-05-01
Fixes #6520 - MockClient period search

### LINKED ISSUE #6520: [MockClient] due-date filters not working as expected for Task resources
**Description:**
When using the MockClient from `@medplum/mock`, we observed that applying filters on the dueDate field for Task resources—via .search or `.searchResources` —does not yield the expected results, regardless of the operator used (e.g., ge, le, eb, etc.).

**Steps to Reproduce:**

Create a Task resource with a restriction.period.end using MockClient.createResource.

Attempt to search using `.search` or `.searchResources` with a filter such as due-date=eb2024-01-0112:00:00:000Z.

Observe that the expected results are not returned.

**What We Tried:**

Verified the syntax and structure of the search parameters.

Used the same Task resources created in the test suite and manually created them via the Medplum Admin UI.

Applied equivalent dueDate filters via URL parameters in the Admin UI.

In the Admin UI, the filtering behavior worked as expected and returned the correct resources.

**Expected Behavior:**
Search results using the MockClient should match the behavior observed in the Medplum Admin UI when using equivalent filter criteria.

**Actual Behavior:**
The MockClient fails to return matching resources when filtering by dueDate.

**Environment:**

@medplum/mock version: 4.1.0

Testing framework: Jest

**Additional Context:**
This may point to an inconsistency or lack of support for date filtering in the MockClient implementation.

codyebberson:
Before: MockClient date/dateTime filters only worked on date/dateTime property values

After: MockClient date/dateTime filters also work on Period values

Implementation notes:
* This is yet another example of drift between Postgres search and Memory search
* We have been talking about the "search IR" framework, which likely would have caught and fixed this bug

Context: https://medplum.slack.com/archives/C06H7LWCA2U/p1746127075556339?thread_ts=1746111315.434809&cid=C06H7LWCA2U

### DIFF
--- a/packages/core/src/search/match.ts
+++ b/packages/core/src/search/match.ts
@@ -274,8 +274,8 @@ function buildDateTimeColumn(value: unknown): string | undefined {
     try {
       const date = new Date(value);
       return date.toISOString();
-    } catch (_err) {
-      // Silent ignore
+    } catch (err) {
+      console.debug('Failed to parse date', value, err);
     }
   } else if (isPeriod(value)) {
     // Can be a Period
--- a/packages/core/src/search/match.ts
+++ b/packages/core/src/search/match.ts
@@ -222,7 +222,11 @@ function matchesDateFilter(resource: Resource, filter: Filter, searchParam: Sear
   const negated = isNegated(filter.operator);
   for (const resourceValue of resourceValues) {
     for (const filterValue of filterValues) {
-      const match = matchesDateValue(resourceValue as string, filter.operator, filterValue);
+      const match = matchesDateValue(
+        buildDateTimeColumn(resourceValue),
+        filter.operator,
+        buildDateTimeColumn(filterValue) as string
+      );
       if (match) {
         return !negated;
       }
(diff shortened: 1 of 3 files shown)


packages/core/src/search/match.ts
```diff
@@ -253,6 +256,36 @@ function matchesDateValue(resourceValue: string, operator: Operator, filterValue
...
+ * @returns The date/time string if parsed; undefined otherwise.
+ */
+function buildDateTimeColumn(value: unknown): string | undefined {
+  if (!value) {
+    return undefined;
+  }
+  if (typeof value === 'string') {
+    try {
+      const date = new Date(value);
+      return date.toISOString();
```
mattlong, line 273: The tests that are failing are because we're now comparing date.toISOString(), e.g. `1990-01-01T00:00:00.000Z` to the raw filter string, e.g. `1990-01-01`. It seems like this would work out okay if we also pass `filterValue` through `buildDateTimeColumn`?

mattlong (changes requested):

packages/core/src/search/match.ts
codyebberson: Yeah...

I started down the path of doing a more proper "search IR" idea that we've talked about, but that broke even more tests.

For now, I'm going to just use our "search param details" to differentiate between "date" and "datetime".

mattlong (approved):

PR #6523, opened 2025-05-01
New super admin page to manage column statistics targets

mattlong:


packages/server/src/fhir/operations/utils/parameters.ts
```diff
@@ -233,6 +233,16 @@ export function buildOutputParameters(operation: OperationDefinition, output: ob
...
 function makeParameter(param: OperationDefinitionParameter, value: unknown): ParametersParameter | undefined {
+  if (param.part && value && Array.isArray(value)) {
+    const parts: ParametersParameter[] = [];
+    for (const val of value.map((v) => makeParameter(param, v))) {
+      if (val) {
+        parts.push(val);
+      }
+    }
+    return { name: param.name, part: parts };
+  }
```
mattwiller, line 244: Would you mind adding a specific unit test case for this in `parameters.test.ts`?

packages/server/src/fhir/routes.ts
```diff
@@ -335,6 +337,8 @@ function initInternalFhirRouter(): FhirRouter {
   router.add('POST', '/$db-stats', dbStatsHandler);
   router.add('POST', '/$db-schema-diff', dbSchemaDiffHandler);
   router.add('POST', '/$db-invalid-indexes', dbInvalidIndexesHandler);
+  router.add('GET', '/$db-column-statistics', lookupDbColumnStatisticsHandler);
+  router.add('POST', '/$db-column-statistics', updateDbColumnStatisticsHandler);
```
mattwiller, line 341: This doesn't really comport with how FHIIR models Operations: `$db-column-statistics` has two conflicting OperationDefinitions.  These should be split into two different operation routes, each with their own OperationDefinition

packages/server/src/fhir/operations/dbcolumnstatisticslookup.ts
```diff
@@ -0,0 +1,276 @@
...
+      name: 'defaultStatisticsTarget',
+      type: 'integer',
+      min: 1,
+      max: '1',
+    },
+    {
+      use: 'out',
+      name: 'table',
+      part: [
+        { use: 'out', name: 'tableName', type: 'string', min: 1, max: '1' },
```
mattwiller, line 40: You shouldn't need to specify `Parameters.parameter.part.use`, since it's implicitly the same as its parent parameter's `use`

mattlong:
This has dropped in priority quite a bit. Just going to close it for now.

PR #6524, opened 2025-05-02
Revised inlined token strategy of dedicated columns per search parameter

mattlong:
The current new token implementation that has been enabled for some projects on prod adds a single `__tokens` array column to each resource table. Every code/system/value tuple for token-based search parameters that was previously written to the separate `<Resource>_Token` table are now also added to the `__tokens` column in several different formats to facilitate the various ways tokens can be searched:
code+system
code+system+value
code+value

In theory, putting all search params into a single column was the most efficient since it means only two new indexes would be needed per table (one for exact match searches, one for `:text/:contains` searches). However, this approach can lead to poor query performance on tables with very many rows and lots of token-based search parameters, e.g. AuditEvent, Task, Observation.

It turns out that putting so many disparate values into a single array column makes indexes on that column difficult to work with for the query planner since there are so many disparate values being stored in it.

To address that, we now use more than one array column to store token values in each resource table. The theory being that by distributing all these token array elements into smaller, more targeted arrays that would each have fewer distinct values and more homogenous contents, the query planner should have more precise statistics and make better estimates for row counts.

We had considered (which can be seen in the individual commits of this PR) going even further and using four array columns per search parameter: system, system+value, value , and one for :text searches, but that would have required significantly more indexes per column which we weren't comfortable with.


mattlong:
Replaced by #6727

PR #6527, opened 2025-05-02
feat(server): add `transactionAttempts` and `transactionExpBackoffBaseDelayMs` config

### LINKED ISSUE #6533: Add config options for transaction retries and exponential backoff
Currently there is no way to configure Medplum server to retry transactions that failed due to conflicts from table contention more than once or add exponential backoff, we could add both a config option for number of attempts as well as the base retry wait time for exponential backoff so that self-hosters can tweak these parameters based on how high volume or contentious their traffic is.

ThatOneBro:
Fixes #6533

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2375,14 +2375,16 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         this.endTransaction();
       }
 
-      const baseDelayMs = config.transactionExpBackoffBaseDelayMs ?? defaultExpBackoffBaseDelayMs;
-      // Attempts are 0-indexed, so first wait after first attempt will be somewhere between 75% and 100% of baseDelayMs
-      // This calculation results in something like this for the default values:
-      // Attempt 1: 50 * (2^0) = 50 * [0.75, 1] = **[37.5, 50] ms**
-      // Attempt 2: 50 * (2^1) = 100 * [0.75, 1] = **[75, 100] ms**
-      // etc...
-      const delayMs = Math.ceil(baseDelayMs * 2 ** attempt * (0.75 + Math.random() * 0.25));
-      await sleep(delayMs);
+      if (attempt + 1 < transactionAttempts) {
+        const baseDelayMs = config.transactionExpBackoffBaseDelayMs ?? defaultExpBackoffBaseDelayMs;
+        // Attempts are 0-indexed, so first wait after first attempt will be somewhere between 75% and 100% of baseDelayMs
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -2344,6 +2347,15 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       } finally {
         this.endTransaction();
       }
+
+      const baseDelayMs = config.transactionExpBackoffBaseDelayMs ?? defaultExpBackoffBaseDelayMs;
+      // Attempts are 0-indexed, so first wait after first attempt will be somewhere between baseDelayMs 75% and 100%
```
mattwiller: Is delaying the first retry here intentional?  Previously, an immediate retry was generally quite effective in getting a successful outcome

packages/server/src/fhir/repo.ts
```diff
@@ -2344,6 +2347,15 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
       }
+
+      const baseDelayMs = config.transactionExpBackoffBaseDelayMs ?? defaultExpBackoffBaseDelayMs;
+      // Attempts are 0-indexed, so first wait after first attempt will be somewhere between 75% and 100% of baseDelayMs
+      // This calculation results in something like this for the default values:
+      // Attempt 1: 50 * (2^0) = 50 * [0.75, 1] = **[37.5, 50] ms**
+      // Attempt 2: 50 * (2^1) = 100 * [0.75, 1] = **[75, 100] ms**
+      // etc...
+      const delayMs = Math.ceil(baseDelayMs * 2 ** attempt * (0.75 + Math.random() * 0.25));
+      await sleep(delayMs);
```
codyebberson, line 2358: We should guard here, otherwise we will perform an unnecessary `sleep` on the final iteration.  It's not the end-of-the-world, but might be nice to add a guard.

packages/server/src/fhir/repo.ts
ThatOneBro: Yeah I think as we scale we may notice the immediate retry might not be as effective, especially in cases where there is a large volume of individual requests that conflict with each other (eg. hundreds per second). I think in the case where we already have contention this small penalty is not much overhead and should help the overall health of the system since we get some jitter baked in which may help to smooth bursty, contentious traffic. We can always downtune the base delay if that seems too high too.

packages/server/src/fhir/repo.ts
ThatOneBro: Done

PR #6536, opened 2025-05-02
Reindex AsyncJob includes input parameters

mattlong:
We should do this in a more elegant way, but this will give us at least something to start with.

Screenshot 2025-05-02 at 3 03 50 PM


packages/definitions/dist/fhir/r4/search-parameters-medplum.json
```diff
@@ -1326,7 +1326,7 @@
         "id": "AsyncJob-status",
         "url": "https://medplum.com/fhir/SearchParameter/AsyncJob-status",
         "version": "4.0.1",
-        "name": "type",
+        "name": "status",
```
ThatOneBro, line 1329: Was this intentional to commit? We should probably also run generator to put that into the docs

ThatOneBro:
Ty for running generator 💯

PR #6540, opened 2025-05-04
Added error_description param to authorize error redirects

codyebberson:


packages/server/src/oauth/authorize.ts
```diff
@@ -78,35 +78,35 @@ async function validateAuthorizeRequest(req: Request, res: Response, params: Rec
...
     return false;
   }
   if (params.response_type !== 'code') {
-    sendErrorRedirect(res, redirectUri, 'unsupported_response_type', state);
+    sendErrorRedirect(res, redirectUri, 'unsupported_response_type', 'Invalid response type', state);
     return false;
   }
   if (params.request) {
-    sendErrorRedirect(res, redirectUri, 'request_not_supported', state);
+    sendErrorRedirect(res, redirectUri, 'request_not_supported', 'Missing request', state);
```
ThatOneBro, line 89: This seems inverted, request is not missing in this case but seems like it's just not allowed?

packages/server/src/oauth/authorize.ts
```diff
@@ -78,35 +78,35 @@ async function validateAuthorizeRequest(req: Request, res: Response, params: Rec
...
     return false;
   }
   if (!isValidAudience(params.aud)) {
-    sendErrorRedirect(res, redirectUri, 'invalid_request', state);
+    sendErrorRedirect(res, redirectUri, 'invalid_request', 'Invalid audience', state);
     return false;
   }
   if (params.code_challenge && !params.code_challenge_method) {
-    sendErrorRedirect(res, redirectUri, 'invalid_request', state);
+    sendErrorRedirect(res, redirectUri, 'invalid_request', 'Invalid code challenge', state);
```
ThatOneBro, line 97: Could we be more specific here?
```suggestion
    sendErrorRedirect(res, redirectUri, 'invalid_request', 'Missing code_challenge_method', state);
```

codyebberson:
Thanks @ThatOneBro  - good catches.  Fixed

ThatOneBro (approved):

PR #6543, opened 2025-05-05
Dependency upgrades 2025-05-05

medplumbot:
Dependency upgrades


codyebberson:
Looks like the main issue is GraphiQL major version upgrade:

image

Migration guide:

https://github.com/graphql/graphiql/blob/main/docs/migration/graphiql-4.0.0.md

image

codyebberson:
New failure in `react-router`.  IIRC, we saw this 1 or 2 weeks ago as well and held back `react-router`.  I don't fully understand what's going on in this stack trace.  Why `TypeError: Cannot read properties of undefined (reading 'toUpperCase')` inside `isMutationMethod (../../node_modules/react-router/dist/development/index.js:4699:42)` ??? 🤔 

image

codyebberson:
ChatGPT analysis:

## Root cause

`createBrowserRouter()` builds a **Request** object every time you call `navigate()`.
In real browsers (and in Node 20 +) `new Request('/') .method` is `"GET"`, but JSDOM ships its own stubbed `window.Request` where `.method` is **`undefined`**.
When React Router later does `method.toUpperCase()` you get

```
TypeError: Cannot read properties of undefined (reading 'toUpperCase')
```

`<MemoryRouter>` never touches that code path, so those tests keep passing.

## Fix

You just need a full WHATWG Fetch implementation in the test runtime.

**Option A – patch the globals once in your Jest/Vitest setup file**

```ts
// setupTests.ts  (runs before the tests)
if (globalThis.fetch)   { window.fetch   = globalThis.fetch; }
if (globalThis.Request) { window.Request = globalThis.Request; }
if (globalThis.Response){ window.Response= globalThis.Response; }
if (globalThis.Headers) { window.Headers = globalThis.Headers; }
```

Node 20+ already provides these classes; the snippet simply re-exports them on JSDOM’s `window`.

**Option B – use a test environment that already fixes JSDOM**

```bash
npm i -D jest-fixed-jsdom
```

```js
// jest.config.js
module.exports = {
  testEnvironment: 'jest-fixed-jsdom',
};
```

Either way

```ts
console.log(new Request('/').method); // → "GET"
```

and the React Router crash disappears.

## Why this isn’t a version problem

React Router **v7** is the current stable line. All DOM helpers now live directly in `react-router`; `react-router-dom@7` is just a re-export shim.
Your single `react-router@7.x` install is fine—the only missing piece in the failing tests was a real Fetch implementation.

codyebberson:
Bunch of Jest failures in `expo-polyfills`...

image

I'm guessing it's this?

```ts
  const originalWindow: Window = window;

  beforeEach(() => {
    cleanupMedplumWebAPIs();
    Object.defineProperty(globalThis, 'window', {
      value: { ...originalWindow },
    });
  });
```

codyebberson:
Both `babel-preset-expo` and `jest-expo` upgraded `react-native` to 0.78.  Going to try that...

codyebberson:
Fixed about ~600 `JSX` imports using this script:

```js
const fs = require('fs');
const path = require('path');

const startDir = process.argv[2];

function walk(dir) {
  for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
    const full = path.join(dir, entry.name);
    if (entry.isDirectory()) {
      walk(full);
    } else if (entry.isFile() && full.endsWith('.tsx')) {
      handleTsx(full);
    }
  }
}

function handleTsx(filePath) {
  const src = fs.readFileSync(filePath, 'utf8');

  // 1. Does it reference JSX.Element anywhere?
  if (!/JSX\.Element\b/.test(src)) return;

  // 2. Does it already import JSX?  (Assumes "import" and "JSX" share a line.)
  const hasJsxImport = /import[^\n;]*\bJSX\b[^\n;]*from[^\n;]*['"]react['"]/.test(src);
  if (hasJsxImport) return;

  // 3. Prepend the missing import.
  const updated = `import { JSX } from 'react';\n${src}`;
  fs.writeFileSync(filePath, updated, 'utf8');
  console.log(`Added JSX import ➜ ${filePath}`);
}

walk(startDir);
```

codyebberson:
* `react` and `react-dom` v19
   * Now need to explicitly import `JSX`
   * Now need to provide default value when using `useRef()`
* `react-native` v0.79.2

When fixing the `JSX` imports, I ran `prettier` with `prettier-plugin-organize-imports` on `examples`, which led to a huge cascade of changes.  I'm ok to pay the one-time price for this.

ThatOneBro (approved):
LGTM

PR #6547, opened 2025-05-05
Make transaction callbacks idempototent

mattwiller:


ThatOneBro (approved):

ThatOneBro:
Can we add a regression test for the retry case we were seeing? If it's not too complex to create such a test

codyebberson (approved):

PR #6558, opened 2025-05-07
Allow more x-trace-id formats

codyebberson:


mattwiller:
To prevent issues with downstream observability services, we should ensure that the hex part of the trace ID can be packed into 128 bits.  I'd recommend normalizing to a 32-character hex string, with all punctuation removed, left-filling with `0` as needed for shorter values, and maybe truncating for longer values

ianplunkett:
Is this compatible with x-ray? Should we just close this out and update our docs to make sure external trace ids work with x-ray? @codyebberson & @mattwiller

PR #6570, opened 2025-05-08
ci(agent): finally fix windows cache path

ThatOneBro:
See: https://github.com/cypress-io/github-actions-cache/blob/4fc8abd580f3084f31c4bcddae97fd15aefe8339/examples.md?plain=1#L73C59-L73C78

Not much docs about Windows npm cache on GitHub Actions runners anywhere else

EDIT: After looking on the official `actions/cache` repo, I saw the recommended way is to get the cache directory directly from npm by running `npm config get cache`

https://github.com/actions/cache/blob/5a3ec84eff668545956fd18022155c47e93e2684/examples.md#node---npm

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -34,12 +34,18 @@ jobs:
           node-version: 20
           registry-url: 'https://registry.npmjs.org'
 
+      # See: https://github.com/actions/cache/blob/5a3ec84eff668545956fd18022155c47e93e2684/examples.md#node---npm
+      - name: Get npm cache directory
+        id: npm-cache-dir
+        shell: pwsh
+        run: echo "dir=$(npm config get cache)" >> ${env:GITHUB_OUTPUT}
+
       - name: Cache node modules
         uses: actions/cache@v4
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-agent.yml
```diff
@@ -34,12 +34,18 @@ jobs:
           node-version: 20
           registry-url: 'https://registry.npmjs.org'
 
+      # See: https://github.com/actions/cache/blob/5a3ec84eff668545956fd18022155c47e93e2684/examples.md#node---npm
+      - name: Get npm cache directory
+        id: npm-cache-dir
+        shell: pwsh
+        run: echo "dir=$(npm config get cache)" >> ${env:GITHUB_OUTPUT}
```
codyebberson, line 41: Neat.  I didn't think this would work on Windows, but I guess PowerShell ftw?  👍

codyebberson (approved):

.github/workflows/build-agent.yml
ThatOneBro: Yeah powershell makes Windows a lot less bad, sometimes 😛

PR #6572, opened 2025-05-08
QuestionnaireForm: Add Item Format for Multi-select via Checkbox

maddyli:
Adds functionality to choice questionnaire items: 

- Allows dropdown component to be both multi and single select for both explicit element items and valueset items. 

- Extends radio buttons (single select only) to be rendered for valueset items along with explicit element items. 

- Creates checkboxes (multi select only), that can render both explicit element items and valueset items.

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
@@ -1034,7 +1034,28 @@ export const Choices = (): JSX.Element => (
           },
           {
             linkId: 'q4',
-            text: 'Question 4 - Value Set MultiselectDropdown',
+            type: 'choice',
+            text: 'Question 4 - Value Set Checkbox Choice',
+            answerValueSet: 'http://loinc.org/vs/LL4436-3',
+            extension: [
+              {
+                url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl',
+                valueCodeableConcept: {
+                  coding: [
+                    {
+                      system: 'http://hl7.org/fhir/questionnaire-item-control',
+                      code: 'check-box',
+                      display: 'Check box',
+                    },
+                  ],
+                  text: 'Check box',
+                },
+              },
+            ],
+          },
+          {
+            linkId: 'q5',
+            text: 'Question 5 - Value Set Multiselect Dropdown',
             type: 'choice',
             answerValueSet: 'http://loinc.org/vs/LL4436-3',
             extension: [
@@ -1062,27 +1083,6 @@ export const Choices = (): JSX.Element => (
                     },
                   ],
                   text: 'Multi select',
-                },
-              },
-            ],
-          },
-          {
-            linkId: 'q5',
-            type: 'choice',
-            text: 'Question 5 - Value Set Checkbox Choice > 30 items',
-            answerValueSet: 'http://terminology.hl7.org/ValueSet/v3-Ethnicity',
-            extension: [
-              {
-                url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl',
-                valueCodeableConcept: {
-                  coding: [
-                    {
-                      system: 'http://hl7.org/fhir/questionnaire-item-control',
-                      code: 'check-box',
-                      display: 'Check box',
-                    },
-                  ],
-                  text: 'Check box',
                 },
               },
             ],
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
@@ -985,21 +985,21 @@ export const MultipleChoice = (): JSX.Element => (
                 valueString: 'Yellow',
               },
             ],
-            "extension": [
+            extension: [
               {
-                url: "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
+                url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl',
                 valueCodeableConcept: {
                   coding: [
                     {
-                      system: "http://hl7.org/fhir/questionnaire-item-control",
-                      code: "check-box",
-                      display: "Check box",
-                    }
+                      system: 'http://hl7.org/fhir/questionnaire-item-control',
+                      code: 'check-box',
+                      display: 'Check box',
+                    },
                   ],
-                  text: "Check box",
-                }
-              }
-            ]
+                  text: 'Check box',
+                },
+              },
+            ],
           },
         ],
       }}
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
@@ -1073,19 +1073,19 @@ export const Choices = (): JSX.Element => (
             answerValueSet: 'http://terminology.hl7.org/ValueSet/v3-Ethnicity',
             extension: [
               {
-                url: "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
+                url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl',
                 valueCodeableConcept: {
                   coding: [
                     {
-                      system: "http://hl7.org/fhir/questionnaire-item-control",
-                      code: "check-box",
-                      display: "Check box",
-                    }
+                      system: 'http://hl7.org/fhir/questionnaire-item-control',
+                      code: 'check-box',
+                      display: 'Check box',
+                    },
                   ],
-                  text: "Check box",
-                }
-              }
-            ]
+                  text: 'Check box',
+                },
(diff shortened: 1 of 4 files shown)


maddyli:
@codyebberson I can add a quick truncation to only render the first 50 items from a set as checkboxes. That being said, I noticed that the rendering options seem to restrict multi-select dropdown rendering to non-valueset items. Makes sense since valueset items can be queried for autocomplete, but it does mean that the QuestionnaireForm will not allow a valueset from making multiple items. Mind if I add that feature? I see it in some of our example apps that don't use this form 

https://github.com/medplum/medplum/blob/main/packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx#L252-L282

packages/react/src/QuestionnaireForm/QuestionnaireForm.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { getQuestionnaireAnswers } from '@medplum/core';
+import { getAllQuestionnaireAnswers, getQuestionnaireAnswers } from '@medplum/core';
```
maddyli, line 1: Is there a reason why we have both getAllQuestionnaireAnswers and getQuestionnaireAnswers?

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -249,19 +253,20 @@ export function QuestionnaireFormItem(props: QuestionnaireFormItemProps): JSX.El
       break;
     case QuestionnaireItemType.choice:
     case QuestionnaireItemType.openChoice:
-      if (isDropDownChoice(item) && !item.answerValueSet) {
+      if (isCheckboxChoice(item)) {
         formComponent = (
-          <QuestionnaireChoiceDropDownInput
-            name={name}
+          <QuestionnaireCheckboxInput
+            name={response?.id ?? name}
```
techdavidy, line 259: do we want the response?.id here ?

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -307,28 +312,61 @@ interface QuestionnaireChoiceInputProps {
...
 
-  if (item.repeats) {
-    const { propertyName, data } = formatSelectData(props.item);
-    const currentAnswer = getCurrentMultiSelectAnswer(response);
+  if (item.answerValueSet) {
+    return (
+      <AsyncAutocomplete<Coding>
+        name={name}
+        placeholder="Select items"
+        loadOptions={async (inputValue: string, _signal: AbortSignal) => {
```
techdavidy, line 333: lets move create functions before the return
see https://github.com/medplum/medplum/blob/8e783a1f24ebd701b55ea2085e433ff0f6870af6/packages/react/src/ValueSetAutocomplete/ValueSetAutocomplete.tsx#L57

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -307,28 +312,61 @@ interface QuestionnaireChoiceInputProps {
...
+          const valueSet = await medplum.valueSetExpand({
+            url: item.answerValueSet as string,
+            filter: inputValue,
+            count: 50,
```
techdavidy, line 337: I think our default is 10

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -337,96 +375,121 @@ function QuestionnaireChoiceDropDownInput(props: QuestionnaireChoiceInputProps):
...
+function useValueSetOptions(valueSetUrl: string | undefined): [ValueSetExpansionContains[], boolean] {
+  const medplum = useMedplum();
+  const [valueSetOptions, setValueSetOptions] = useState<ValueSetExpansionContains[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
 
-  if (!item.answerOption?.length) {
-    return <NoAnswerDisplay />;
-  }
+  useEffect(() => {
+    let mounted = true;
```
techdavidy, line 434: do we need this ?

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -446,35 +509,142 @@ function QuestionnaireChoiceRadioInput(props: QuestionnaireChoiceInputProps): JS
...
 
+  if (isLoading) {
+    return <Text>Loading options...</Text>;
+  }
+
+  if (options.length === 0) {
+    return <NoAnswerDisplay />;
+  }
+
+  const limitedOptions = options.slice(0, 30);
```
techdavidy, line 520: I think we should create a const for 30. Magic number.

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -446,35 +509,142 @@ function QuestionnaireChoiceRadioInput(props: QuestionnaireChoiceInputProps): JS
...
+  const valueElementDefinition = getElementDefinition('QuestionnaireItemAnswerOption', 'value[x]');
+  const currentAnswers = getCurrentMultiSelectAnswer(response);
+  const [valueSetOptions, isLoading] = useValueSetOptions(item.answerValueSet);
+
+  const options: [string, TypedValue][] = [];
+
+  if (item.answerValueSet) {
+    options.push(...getOptionsFromValueSet(valueSetOptions, name));
+  } else if (item.answerOption) {
+    for (let i = 0; i < item.answerOption.length && i < 50; i++) {
```
techdavidy, line 572: I think we can use high order functions here

```
const options = item.answerOption
  .slice(0, 50) // Limit to 50 items
  .filter(option => {
    const optionValue = getItemAnswerOptionValue(option);
    return !!optionValue?.value;
  })
  .map((option, i) => {
    const optionName = `${name}-option-${i}`;
    const optionValue = getItemAnswerOptionValue(option);
    return [optionName, optionValue] as [string, typeof optionValue];
  });
```

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
```diff
@@ -500,20 +670,30 @@ function getCurrentRadioAnswer(options: [string, TypedValue][], defaultAnswer: T
...
-  return !!item.extension?.some(
+type ChoiceType = 'check-box' | 'drop-down' | 'radio-button' | 'multi-select' | undefined;
+
+function getChoiceType(item: QuestionnaireItem): ChoiceType {
+  return item.extension?.find(
     (e) =>
-      e.url === 'http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl' &&
-      e.valueCodeableConcept?.coding?.[0]?.code === 'drop-down'
-  );
+      e.url === HTTP_HL7_ORG + '/fhir/StructureDefinition/questionnaire-itemControl' &&
```
techdavidy, line 678: 👍

packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
```diff
@@ -969,6 +1016,76 @@ export const MultipleChoice = (): JSX.Element => (
...
+                    },
+                  ],
+                  text: 'Check box',
+                },
+              },
+            ],
+          },
+          {
+            linkId: 'q5',
+            text: 'Question 5 - Value Set Multiselect Dropdown',
```
techdavidy, line 1058: when I tried this, it didn't allow for multiple selection.

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
maddyli: According to cursor, yes 

Screenshot 2025-05-15 at 11 33 59 AM

packages/react/src/QuestionnaireForm/QuestionnaireFormItem/QuestionnaireFormItem.tsx
maddyli: Hm since these value set items are being used to render the checkboxes and radiobuttons, I have set it to my new const for the limit on number to render

packages/react/src/QuestionnaireForm/QuestionnaireForm.stories.tsx
maddyli: fixed

techdavidy (approved):

PR #6575, opened 2025-05-09
wip(agent): add `$fetch-logs` operation

ThatOneBro:


ThatOneBro:
Closing this PR for now, the files this PR touches have changed quite a bit since this was opened and I think we will end up going with a completely different approach, using something like Winston to accomplish both log rotation instead of relying on shawl, and also allowing for the configuration of logs of separate categories (main logs, channel message logs) to be written to separate files.

PR #6580, opened 2025-05-09
Support dynamic post-deploy migrations based on schema drift

mattlong:
As Medplum's database migration system has evolved quite a bit. In older iterations, it was possible for some migration actions to not have actually been executed successfully but still be marked as complete. The declarative nature of our migration generator script made it straightforward to calculate and view the diff between the expected database schema and reality; what we refer to as "schema drift".

This PR goes the next step of being able to actually execute the migration actions that had previously been missed. It utilizes the post-deploy migration functionality when possible. Getting here required separating the migration action calculation/generation logic from the previously hardcoded output path of writing them to a typescript file. This was achieved by introducing an intermediate, JSON-serializable representation `MigrationAction`. The `migrate.ts` script now generates an ordered list of this actions that can either be written to a typescript migration file as before, or enqueued as a new AsyncJob/BullMQ job.


packages/server/src/migrations/data/types.ts
```diff
@@ -31,14 +32,24 @@ export interface PostDeployMigration<T extends PostDeployJobData = PostDeployJob
   run(repo: Repository, job: Job<T> | undefined, data: T): Promise<PostDeployJobRunResult>;
 }
 
-// Custom Jobs
+export type MigrationActionResult = { name: string; durationMs: number };
```
codyebberson, line 35: What was the motivation for renaming `CustomMigrationAction` to `MigrationActionResult`?

It seems that we're now being somewhat loose with "action" vs "result", which seems like distinct concepts?

Sorry, I hate to be pedantic, but it seems like this is a big part of the diff, and it's not totally clear.

Naming things is hard.

packages/server/src/workers/utils.ts
```diff
@@ -324,7 +324,7 @@ export function addVerboseQueueLogging<TDataType>(
   });
 }
 
-export async function moveToDelayed(job: Job, reason: string): Promise<void> {
+export async function moveToDelayedAndThrow(job: Job, reason: string): Promise<void> {
```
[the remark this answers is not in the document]
mattlong: Good call, from [the typescript handbook](https://www.typescriptlang.org/docs/handbook/2/functions.html#never):

> The never type represents values which are never observed. In a return type, this means that the function throws an exception or terminates execution of the program.

packages/server/src/migrations/data/types.ts
mattlong: Yeah...I think normalizing on "action" for the things to be done in a migration and "action result" or "results" as the outcomes from those things.

Pushing a lot more diff to align with that.

codyebberson (approved):

PR #6586, opened 2025-05-11
Fixes #6424 - allow hl7 client socket to be forcefully closed

### LINKED ISSUE #6424: Socket cannot be closed until connection is established
Problem:
The current implementation of hl7 client does not allow to add a custom timeout on the connecting. This is because we cannot destroy the socket while it's still in the process of connecting. As a result, repeated connection attempts can lead to multiple parallel connection attempts being initiated.

Proposed Solutions:
Allow the connection to be forcefully closed even if it has not yet been successfully established

Сode with the current problem:

`hl7/src/client.ts`
```
    return new Promise((resolve, reject) => {
      const socket = connect({ host: this.host, port: this.port, keepAlive: this.keepAlive }, () => {
        let connection: Hl7Connection;
        this.connection = connection = new Hl7Connection(socket, this.encoding);
        socket.off('error', reject);
        connection.addEventListener('close', () => this.dispatchEvent(new Hl7CloseEvent()));
        connection.addEventListener('error', (event) => this.dispatchEvent(new Hl7ErrorEvent(event.error)));
        resolve(this.connection);
      });
```

``` 
 close(): void {
    if (this.connection) {
      this.connection.close();
      delete this.connection;
    }
  }
```

codyebberson:


packages/hl7/src/client.ts
```diff
@@ -26,24 +29,79 @@ export class Hl7Client extends Hl7Base {
     this.port = this.options.port;
     this.encoding = this.options.encoding;
     this.keepAlive = this.options.keepAlive ?? false;
+    this.connectTimeout = this.options.connectTimeout ?? 30000; // Default 30 seconds
```
ThatOneBro, line 32: Should this default to 0 when `keepAlive` is true?

packages/hl7/src/client.ts
```diff
@@ -26,24 +29,79 @@ export class Hl7Client extends Hl7Base {
...
+        // Create the HL7 connection
         let connection: Hl7Connection;
-        this.connection = connection = new Hl7Connection(socket, this.encoding);
-        socket.off('error', reject);
-        connection.addEventListener('close', () => this.dispatchEvent(new Hl7CloseEvent()));
-        connection.addEventListener('error', (event) => this.dispatchEvent(new Hl7ErrorEvent(event.error)));
+        this.connection = connection = new Hl7Connection(this.socket, this.encoding);
+
+        // Remove the timeout listener as we're now connected
+        this.socket.setTimeout(0);
```
ThatOneBro, line 82: Do we only wanna timeout whenever connecting?

It might be useful to be able to detect when a connection is broken, which by default won't happen, even when keepalive probes are not responded to for long periods of time 🤔

ThatOneBro (approved):

packages/hl7/src/client.ts
codyebberson: https://nodejs.org/docs/latest-v22.x/api/net.html#socketsettimeouttimeout-callback

> Sets the socket to timeout after timeout milliseconds of inactivity on the socket. By default net.Socket do not have a timeout.

I don't think we want to timeout on inactivity though 🤔

ThatOneBro: I mean if keepalive = true, then even if keepalive messages are not acked, we never do anything about fixing that broken connection. Typically keepalive is paired with a timeout in case of flaky connections. Otherwise you could be sending stuff into the void and have no knowledge of it until you send the next message

ThatOneBro: I thought we set a timeout automatically when setting keepalive, otherwise it doesn't fix itself if the other side stops responding to keepalive messages

codyebberson: Mm, i see, you mean the ping/pongs - sorry, I was misinterpreting "keepalives".

Maybe?  I don't have as much conviction on that point.

ThatOneBro: Yeah I mean if we do that it can always be a separate PR, I think we mostly added keepalive to keep LB from timing out connections for the moment. So it serves that purpose as is

PR #6605, opened 2025-05-12
Explanation for rate limit calculations

finnbergquist:
Also included directions for self hosters to set the project systemSettings


ianplunkett (approved):
Looks good. Can you add a short note that these limits are subject to change?

codyebberson (approved):

PR #6610, opened 2025-05-13
[Docs] Update Agent Setup Docs

maddyli:
- Add instructions for spinning up agent via CLI 
- Add a way to send a specific hl7 message via local file


ianplunkett (approved):
looks good to me

PR #6625, opened 2025-05-14
Report rate limit status from client

mattwiller:
Adding a `MedplumClient.rateLimitStatus()` method to report the most recent rate limit information received from the server

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -846,6 +846,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
   private refreshPromise?: Promise<any>;
   private profilePromise?: Promise<any>;
   private sessionDetails?: SessionDetails;
+  private currentRateLimits?: string;
   private basicAuth?: string;
   private initPromise: Promise<void>;
   private initComplete = true;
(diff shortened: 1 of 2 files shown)


packages/core/src/client.ts
```diff
@@ -3389,6 +3393,38 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
   }
 
+  private setCurrentRateLimit(res: Response): void {
+    const rateLimitHeader = res.headers.get('ratelimit');
+    if (rateLimitHeader) {
+      this.currentRateLimits = rateLimitHeader;
+    }
+  }
+
+  rateLimitStatus(): { name: string; remainingUnits: number; secondsUntilReset: number }[] {
```
ianplunkett, line 3403: Can we add annotations to get picked up by the doc generator?

packages/agent/src/agent-main.test.ts
```diff
@@ -150,6 +151,7 @@ describe('Main', () => {
     fetchSpy.mockImplementation(async () => {
       return {
         ok: true,
+        headers: new Headers(),
```
codyebberson, line 154: Interesting.  I assume this is because any mock request will throw if `headers` is `undefined`?  

Any concerns about this impacting other customers in their unit tests?  I guess it will only happen at dev-time, and this is technically correct / technically required?

codyebberson (approved):

PR #6632, opened 2025-05-15
Fix quote escaping issue in `migrate.ts`

mattlong:


packages/server/src/migrations/migrate.ts
```diff
@@ -961,7 +954,7 @@ function writeActionsToBuilder(b: FileBuilder, actions: MigrationAction[]): void
       }
       case 'CREATE_INDEX': {
         b.appendNoWrap(
-          `await fns.idempotentCreateIndex(client, actions, '${action.indexName}', '${action.createIndexSql}');`
+          `await fns.idempotentCreateIndex(client, actions, '${action.indexName}', \`${action.createIndexSql}\`);`
```
mattlong, line 957: This was the actual bug being fixed; the other changes were cleanups.

mattwiller (approved):

PR #6635, opened 2025-05-16
Patient Summary V2

techdavidy:


codyebberson (commented):
Nice, ty @kevinwadeshaw  and @techdavidy !  Overall, this looks really good.  Big improvement all around 🎉 🙏 

@techdavidy  Can you take a quick pass at trying to use some of the built-in utilities from `@medplum/core` such as `formatAddress()`, `getExtension()`, etc?

Stretch goal to avoid using `getByTestId()`, as we're trying to eliminate that for accessibility improvements, but understood if that's too invasive.  We have a backlog of those, so that might be a bigger future project.

codyebberson (approved):
Nice 👍 🙏 🚀

PR #6636, opened 2025-05-16
Allows Twilio request header in unauthenticated webhooks

filipeximenes:
Adds the Twilio `X-Twilio-Signature` header to the list of allowed webhook signatures (as documented [here](https://www.twilio.com/docs/usage/webhooks/webhooks-security)). Since we already allowed the Twilio email signature I believe this is a simple and reasonable addition but please LMK if this requires more thought before it's merged, I'll be happy to open a discussion.

### DIFF
--- a/packages/server/src/webhook/routes.ts
+++ b/packages/server/src/webhook/routes.ts
@@ -1,11 +1,10 @@
-import { allOk, badRequest, getStatus, isOperationOutcome, isResource } from '@medplum/core';
+import { allOk, badRequest, getStatus, isOperationOutcome } from '@medplum/core';
 import { Bot, ProjectMembership, Reference } from '@medplum/fhirtypes';
 import { Request, Response, Router } from 'express';
 import { asyncWrap } from '../async';
 import { executeBot, getResponseBodyFromResult, getResponseContentType } from '../fhir/operations/execute';
 import { sendOutcome } from '../fhir/outcomes';
 import { getSystemRepo } from '../fhir/repo';
-import { sendFhirResponse } from '../fhir/response';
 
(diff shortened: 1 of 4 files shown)


mattwiller (approved):

filipeximenes:
One issue I just noticed, even though this change would allow the bot to run it wouldn't serve its purpose because it seems like webhook bots can't respond with a payload:
https://github.com/medplum/medplum/blob/7a93b67a01d44aeb2a7ce280ce22598d9ea8b40e/packages/server/src/webhook/routes.ts#L61-L71

For this context I need to respond to Twilio with a [TwiML formated XML](https://www.twilio.com/docs/voice/twiml).

packages/server/src/webhook/routes.ts
```diff
@@ -58,17 +63,21 @@ export const webhookHandler = asyncWrap(async (req: Request, res: Response) => {
...
+  const outcome = result.success ? allOk : badRequest(result.logResult);
+
+  if (isResource(responseBody, 'Binary')) {
+    await sendFhirResponse(req, res, outcome, responseBody);
+    return;
   }
-  res.sendStatus(getStatus(outcome));
+
+  // Send the response
+  res.status(getStatus(outcome)).type(getResponseContentType(req)).send(responseBody);
```
finnbergquist, line 80: @filipeximenes , can you add a test to check this scenario where you send a response?

filipeximenes: @finnbergquist Done

codyebberson (approved):
lgtm.  Thanks @filipeximenes !

PR #6637, opened 2025-05-16
Use rate limit header to calculate client backoff

mattwiller:


packages/core/src/client.test.ts
```diff
@@ -2968,6 +2968,33 @@ describe('Client', () => {
...
+        headers: new Headers({ 'content-type': ContentType.FHIR_JSON }),
+        json: async () => ({ resourceType: 'Patient' }),
+      };
+    });
+
+    const client = new MedplumClient({ fetch });
+    const patientPromise = client.readResource('Patient', '123');
+
+    // Promise should resolve after one second retry delay
+    await sleep(800);
```
mattlong, line 2990: Would it be possible to use fake timers here or some other way to avoid this test taking the full second?

mattlong (approved):

PR #6642, opened 2025-05-17
Setting `fhirQuota` on UserConfiguration docs

finnbergquist:
Also explaining:
- How to setup a UserConfiguration
- UserConfiguration.menu
- How to view your project's totalFhirQuota & userFhirQuota


packages/docs/docs/rate-limits/index.md
```diff
@@ -47,6 +47,24 @@ with the sum of each user's interactions in a given minute compared to the confi
 a limit on the total interactions for all users within a Project as a whole, which defaults to ten times the per-user
 limit.
 
+### How to view your project's FHIR quota rate limits
+
+The limit that will be enforced on each User in a Project can be viewed in the [Project's](/docs/api/fhir/medplum/project) system settings:
+
+- `Project.systemSettings.userFhirQuota` - the limit that will be enforced on each User in the Project
```
mattlong, line 54: Should these both specify that they expect/require an integer value?

mattlong (approved):

packages/docs/docs/rate-limits/index.md
finnbergquist: Sure. I also left a note that they may not be set. If this is the case, then it will use the default values

PR #6668, opened 2025-05-23
Add utility for summarizing Observation data

mattwiller:


codyebberson (commented):
👍 

What is the argument for using `component.valueSampledData` rather than using top level `valueSampledData`?

What is the recommendation for someone who is working with the output?  Any objections to adding a  utility that unpacks that summary `Observation` into either an array of values or an array of synthetic `Observation` resources?

Should `Observation.effectiveDateTime` be replaced with `Observation.effectivePeriod`?

finnbergquist:
> What is the recommendation for someone who is working with the output? Any objections to adding a utility that unpacks that summary `Observation` into either an array of values or an array of synthetic `Observation` resources?
> 
I agree that this would be super useful, specifically for outputting "an array of synthetic `Observation` resources". 

For people graphing Observation resources in their FE and depending on the current format of data, it would minimize the number of code changes required to migrate to this model. Also for people exporting Observations from Medplum -> data warehouses that just want the normal format for the Observation resources.

codyebberson (approved):

PR #6676, opened 2025-05-26
Fixes #6663 - option to ignore unrecognized C-CDA sections

### LINKED ISSUE #6663: Allow for a way to fail silently on CCDA to FHIR converter for missing templateIds
When calling the `convertCcdaToFhir(ccda)` function, there is an error that gets thrown pretty consistently across many of our CCDA files:

`'Unhandled act templateId: 2.16.840.1.113883.10.20.22.2.6` The templateId can vary depending on the document.

Would love a way to still use the CCDA to FHIR for what it can grab even if it's missing certain templateIds.

Looks like this one is `2.16.840.1.113883.10.20.22.2.6` https://confluence.hl7.org/spaces/SD/pages/51227465/C-CDA+OIDs is Allergy Section (entries optional)

codyebberson:
Also added the missing sections described in #6663


ThatOneBro (approved):
👍 nice, seems very reasonable

PR #6685, opened 2025-05-28
Webhook handler processing of Binary Bot response

finnbergquist:
If the Bot returns a Binary resource in the unauthenticated webhook handler, then the response's content type will be the Binary resource's contentType field

### DIFF
--- a/packages/server/src/webhook/routes.ts
+++ b/packages/server/src/webhook/routes.ts
@@ -1,5 +1,5 @@
 import { allOk, badRequest, getStatus, isOperationOutcome } from '@medplum/core';
-import { Bot, ProjectMembership, Reference, Binary } from '@medplum/fhirtypes';
+import { Binary, Bot, ProjectMembership, Reference } from '@medplum/fhirtypes';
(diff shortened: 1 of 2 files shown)


packages/server/src/webhook/routes.ts
```diff
@@ -70,7 +71,11 @@ export const webhookHandler = asyncWrap(async (req: Request, res: Response) => {
   const responseBody = getResponseBodyFromResult(result);
   const outcome = result.success ? allOk : badRequest(result.logResult);
 
-  res.status(getStatus(outcome)).type(getResponseContentType(req)).send(responseBody);
+  if (result.returnValue?.resourceType === 'Binary') {
```
finnbergquist, line 74: I looked into calling the whole sendFhirResponse() method, but in the case that the Bot response is not Binary, it calls getAuthenticatedContext(), which fails because this is an unauthenticated webhook

filipeximenes (approved):

ThatOneBro (approved):
LGTM 👍 good work Finn

codyebberson (approved):

PR #6700, opened 2025-06-03
Fix SMTP Configuration via ENV vars

xcskier56:
Currently, if you are using ENV vars to configure the medplum server, you are unable to properly setup `SMTP` as it is required to be an object but the ENV loader does not permit that. This PR fixes this issue by mimicking how `env.database` and `env.redis` are configured. 

With this fix, you can use 

- `MEDPLUM_SMTP_HOST=`
- `MEDPLUM_SMTP_PORT=`
- `MEDPLUM_SMTP_PASSWORD=`
- `MEDPLUM_SMTP_USERNAME=`

to configure the smtp settings via ENV vars


xcskier56:
You're welcome @codyebberson! Any timeline on when this will be released? Thanks

PR #6701, opened 2025-06-03
[Docs] Add Example Script to Create Transaction Bundle

maddyli:
Adds to transaction bundle documentation. Includes an example script that pulls resources from a project, maintains their internal relationships, and creates a transaction bundle that can be loaded to a new project.

### DIFF
--- a/packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
+++ b/packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
@@ -352,3 +352,76 @@ Instead you should create the queue of `Promise` requests and then use `Promise.
     {ExampleCode}
   </MedplumCodeBlock>
 </details>
+
+## Example Resource Creation Bundle
+
+Below is an example script that creates transaction bundle that creates resources and maintains relationships between resources. 
+
+Please note that the below script creates/`POST`s resources, and that there is a limit of 50 update/`PUT` resources per transaction bundle. Also, filters may be used to ensure that the bundle only includes resources that are part of the relationships. 
+
+```ts
+import { MedplumClient } from '@medplum/core';
+import { Bundle, MedicationKnowledge, Organization, Substance } from '@medplum/fhirtypes';
+import { convertToTransactionBundle } from '@medplum/core';
+
+async function createMedicationKnowledgeBundle(medplum: MedplumClient): Promise<Bundle> {
+  // First, search for all the resources we need
+  const timestamp = '2025-01-01T00:00:00Z'
+  const [medicationKnowledge, substances, organizations] = await Promise.all([
+    medplum.search('MedicationKnowledge', {_count: 1000}), // filters may be used to keep the transaction bundle size reasonable, such as the _lastUpdated below
+    medplum.search('Substance', {_count: 1000}),
+    medplum.search('Organization', {'_lastUpdated': `ge${timestamp}`, _count: 1000}),
+  ]);
+
+  // Create a bundle with all resources
+  const bundle: Bundle = {
+    resourceType: 'Bundle',
+    type: 'transaction',
+    entry: [],
+  };
+
+  // Add all resources to the bundle
+  // We'll use the existing IDs to maintain references
+  for (const entry of medicationKnowledge.entry || []) {
+    if (entry.resource) {
+      bundle.entry?.push({
+        fullUrl: `urn:uuid:${entry.resource.id}`,
+          request: {
+          method: 'POST',
+          url: 'MedicationKnowledge',
+        },
+        resource: entry.resource,
+      });
+    }
+  }
+
+  for (const entry of substances.entry || []) {
+    if (entry.resource) {
+      bundle.entry?.push({
+        fullUrl: `urn:uuid:${entry.resource.id}`,
+        request: {
+          method: 'POST',
+          url: 'Substance',
+        },
+        resource: entry.resource,
+      });
+    }
+  }
+
+  for (const entry of organizations.entry || []) {
+    if (entry.resource) {
+      bundle.entry?.push({
+        fullUrl: `urn:uuid:${entry.resource.id}`,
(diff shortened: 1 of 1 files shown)


packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
```diff
@@ -352,3 +352,76 @@ Instead you should create the queue of `Promise` requests and then use `Promise.
...
   </MedplumCodeBlock>
 </details>
+
+## Example Resource Creation Bundle
+
+Below is an example script that creates transaction bundle that creates resources and maintains relationships between resources. 
+
+Please note that the below script creates/`POST`s resources, and that there is a limit of 50 update/`PUT` resources per transaction bundle. Also, filters may be used to ensure that the bundle only includes resources that are part of the relationships. 
+
+```ts
```
mattwiller, line 362: Coulld we move this into the corresponding `examples/**/*.ts` file, so it gets type-checked and linted and stays up to date?

packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
```diff
@@ -352,3 +352,76 @@ Instead you should create the queue of `Promise` requests and then use `Promise.
     {ExampleCode}
   </MedplumCodeBlock>
 </details>
+
+## Example Resource Creation Bundle
+
+Below is an example script that creates transaction bundle that creates resources and maintains relationships between resources. 
+
+Please note that the below script creates/`POST`s resources, and that there is a limit of 50 update/`PUT` resources per transaction bundle. Also, filters may be used to ensure that the bundle only includes resources that are part of the relationships.
```
mattwiller, line 360: It might be worth pointing out that this could result in duplicate resource creations if the Bundle is submitted multiple times

packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
```diff
@@ -352,3 +352,76 @@ Instead you should create the queue of `Promise` requests and then use `Promise.
...
+    type: 'transaction',
+    entry: [],
+  };
+
+  // Add all resources to the bundle
+  // We'll use the existing IDs to maintain references
+  for (const entry of medicationKnowledge.entry || []) {
+    if (entry.resource) {
+      bundle.entry?.push({
+        fullUrl: `urn:uuid:${entry.resource.id}`,
```
mattwiller, line 388: This is handled by `convertToTransactionBundle()`, no need to duplicate the logic for setting `fullUrl` here

packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
```diff
@@ -352,3 +352,76 @@ Instead you should create the queue of `Promise` requests and then use `Promise.
...
+import { Bundle, MedicationKnowledge, Organization, Substance } from '@medplum/fhirtypes';
+import { convertToTransactionBundle } from '@medplum/core';
+
+async function createMedicationKnowledgeBundle(medplum: MedplumClient): Promise<Bundle> {
+  // First, search for all the resources we need
+  const timestamp = '2025-01-01T00:00:00Z'
+  const [medicationKnowledge, substances, organizations] = await Promise.all([
+    medplum.search('MedicationKnowledge', {_count: 1000}), // filters may be used to keep the transaction bundle size reasonable, such as the _lastUpdated below
+    medplum.search('Substance', {_count: 1000}),
+    medplum.search('Organization', {'_lastUpdated': `ge${timestamp}`, _count: 1000}),
```
mattwiller, line 373: This pattern could create very large transaction Bundles, which puts increased load on the database.  I think this example should demonstrate how to chunk the resources up into multiple well-sized transactions.  Maybe consider having an array of `SearchRequest` objects to be done, and for each one loop through and construct the Bundle?

packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
```diff
@@ -352,3 +352,76 @@ Instead you should create the queue of `Promise` requests and then use `Promise.
...
+  for (const entry of substances.entry || []) {
+    if (entry.resource) {
+      bundle.entry?.push({
+        fullUrl: `urn:uuid:${entry.resource.id}`,
+        request: {
+          method: 'POST',
+          url: 'Substance',
+        },
+        resource: entry.resource,
+      });
```
mattwiller, line 407: This logic could be simplified and re-used between resource types, e.g.:
```suggestion
      bundle.entry?.push({
        request: { method: 'POST', url: entry.resource.resourceType },
        resource: entry.resource,
      });
```

packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
maddyli: Hm I think the underlying logic here is that MedicationKnowledges can have many Substances, and Substances can belong to many MedicationKnowledges, so the line for breaking up the bundle isn't super clear. We could also consider just limiting the count? Or we could consider showing a different example, though this is a pretty good representation of some of the more complicated relations that need to be maintained

packages/docs/docs/fhir-datastore/fhir-batch-requests.mdx
rahul1, line 1: @maddyli I think we should re-think this example to make it a better teaching example. I have a few issues with it: 
- Most of the "heavy lifting" in terms of ordering and reference resolution is hidden inside the call to the `convertToTransactionBundle`. If we want to teach users about this function, we should just describe it explicitly
- As @mattwiller mentioned, the resource counts here are too big for a single transaction, which doesn't make it a good example for users to copy


I think a simpler example that maintains reference relationships makes sense here. 


For these large batches of resources, `batch` requests coupled with conditional references might make more sense

rahul1 (changes requested):

PR #6713, opened 2025-06-04
Provider - Include HG labs request

techdavidy:


packages/react/src/PatientSummary/Labs.tsx
```diff
@@ -112,7 +145,7 @@ function hasLaboratoryCategory(category: CodeableConcept): boolean {
   }
 
   for (const coding of category.coding) {
-    if (coding.code === 'laboratory') {
+    if (coding.code === 'LAB') {
```
codyebberson, line 148: Nice catch 👍 

https://hl7.org/fhir/R4/valueset-diagnostic-service-sections.html

PR #6719, opened 2025-06-05
Add diagnostic information to rate limit response

mattwiller:


packages/server/src/ratelimit.ts
```diff
@@ -32,7 +34,10 @@ export function rateLimitHandler(config: MedplumServerConfig): Handler {
 
           const result = err as RateLimiterRes;
           addRateLimitHeader(result, res);
-          res.status(429).json(tooManyRequests).end();
+
+          const outcome: OperationOutcome = { ...tooManyRequests };
```
ThatOneBro: Might want to use `deepClone` instead, since `outcome.issue` will be the reference to the original `outcome.issue` still in this case

ThatOneBro (changes requested):

packages/server/src/ratelimit.ts
mattwiller: Ack, good call

ThatOneBro (approved):
LGTM 👍

PR #6727, opened 2025-06-05
Hashed token array columns per search parameter

mattlong:


packages/server/src/fhir/token-column.ts
```diff
@@ -159,11 +224,38 @@ export function buildTokenColumnsSearchFilter(
...
+          return new Negation(cond);
+        }
+        return cond;
+      }
+
+      if (!impl.hasDedicatedColumns) {
+        const cond = new TypedCondition(
+          new Column(tableName, impl.tokenColumnName),
+          'ARRAY_CONTAINS',
+          // hashTokenColumnValue(filter.code),
```
mattlong: This indicates a gap in the tests that needs to be addressed; searching for an un-hashed value is invalid.

packages/server/src/migrations/data/data-version-manifest.json
```diff
@@ -20,5 +20,11 @@
   },
   "v7": {
     "serverVersion": "4.1.3"
+  },
+  "v8": {
+    "serverVersion": "4.1.4"
```
mattlong: These versions need to be updated to the latest release.

PR #6730, opened 2025-06-06
feat(agent): add `bytestream` channel type

### LINKED ISSUE #7454: [Agent] Add `bytestream` channel type to support ASTM devices over serial
Many medical lab instruments use protocols (like ASTM) over serial ports. These protocols necessitate a channel type which will support a setup consisting of a serial-to-Ethernet device (such as the ones Lantronix, Digi, etc. produce), which will then forward a bytestream over TCP to the agent.

ThatOneBro:
Fixes #7454

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -24,7 +24,7 @@ import {
   normalizeErrorString,
   sleep,
 } from '@medplum/core';
-import { Agent, AgentChannel, Endpoint, Reference } from '@medplum/fhirtypes';
+import { Agent, AgentChannel, Endpoint, OperationOutcomeIssue, Reference } from '@medplum/fhirtypes';
 import { Hl7Client } from '@medplum/hl7';
 import { ChildProcess, ExecException, ExecOptionsWithStringEncoding, exec, spawn } from 'node:child_process';
 import { existsSync, openSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs';
@@ -467,8 +467,15 @@ export class App {
             details: {
               text: `${errors.length} error(s) occurred while reloading channels`,
             },
-            diagnostics: errors.map((err) => normalizeErrorString(err)).join('\n'),
           },
+          ...errors.map(
+            (err) =>
+              ({
+                severity: 'error',
+                code: 'invalid',
+                details: { text: normalizeErrorString(err) },
+              }) satisfies OperationOutcomeIssue
+          ),
         ],
       });
     }
--- a/packages/agent/src/bytestream.ts
+++ b/packages/agent/src/bytestream.ts
@@ -113,7 +113,7 @@ export class AgentByteStreamChannel extends BaseChannel {
   }
 
   sendToRemote(msg: AgentTransmitResponse): void {
-    const connection = this.connections.get(msg.remote as string);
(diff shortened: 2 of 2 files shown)


ThatOneBro:
Removing this from the milestone for now since we haven't determined the priority on this just yet

packages/agent/src/app.ts
```diff
@@ -447,9 +451,27 @@ export class App {
...
+      throw new OperationOutcomeError({
+        resourceType: 'OperationOutcome',
+        issue: [
+          {
+            severity: 'error',
+            code: 'invalid',
+            details: {
+              text: `${errors.length} error(s) occurred while reloading channels`,
+            },
+            diagnostics: errors.map((err) => normalizeErrorString(err)).join('\n'),
```
mattlong: Curious why putting these all into one `issue` rather than mapping at that level?

packages/agent/src/bytestream.ts
```diff
@@ -0,0 +1,210 @@
...
+  start(): void {
+    if (this.started) {
+      return;
+    }
+    this.started = true;
+
+    const address = new URL(this.getEndpoint().address);
+    this.log.info(`Channel starting on ${address}...`);
+    this.configureTcpServerAndConnections();
+    this.server.listen(Number.parseInt(address.port, 10));
```
mattlong, line 43: Is validation on the endpoint or port in particular needed? Or will this fail in an obvious way if it was configured incorrectly

packages/agent/src/bytestream.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.definition = definition;
+    this.endpoint = endpoint;
+
+    this.log.info('Reloading config... Evaluating if channel needs to change address...');
+
+    if (this.needToRebindToPort(previousEndpoint, endpoint)) {
+      await this.stop();
+      this.start();
+      this.log.info(`Address changed: ${previousEndpoint.address} => ${endpoint.address}`);
+    } else if (previousEndpoint.address !== endpoint.address) {
```
mattlong, line 79: The `if` and `else if` blocks never need to both be run?

ThatOneBro: Yeah these are mutually exclusive, we check if we need to rebind to a different port, else we just reconfigure. Calling stop and start again will rebind and also call the same configure function so the `if` block is a special case and superset of the `else if`

packages/agent/src/bytestream.ts
ThatOneBro: https://github.com/medplum/medplum/blob/57a0407d4677e36a70f2d3f51fde4e2c1f5aa2da/packages/agent/src/app.ts#L463

We do validate at a higher level before we instantiate the channels

packages/agent/src/app.ts
ThatOneBro: Good call 👍

packages/agent/src/bytestream.ts
ThatOneBro: Specifically https://github.com/medplum/medplum/blob/57a0407d4677e36a70f2d3f51fde4e2c1f5aa2da/packages/agent/src/app.ts#L476

The `URL` constructor will throw on an invalid port

PR #6742, opened 2025-06-09
Add client helper function to automatically add securityContext to DocumentReference resources on creation

maddyli:
Mimics the functionality of createMedia helper function, which takes any content in the Media resource and adds a securityContext to it.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -31,6 +31,7 @@ import {
   Subscription,
   UserConfiguration,
   ValueSet,
+  DocumentReference,
 } from '@medplum/fhirtypes';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 /** @ts-ignore */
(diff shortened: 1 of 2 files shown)


packages/core/src/client.ts
```diff
@@ -575,6 +576,13 @@ export interface CreateMediaOptions extends CreateBinaryOptions {
   readonly additionalFields?: Partial<Media>;
 }
 
+export interface CreateDocumentReferenceOptions extends CreateBinaryOptions {
+  /**
+   * Optional additional fields for the DocumentReference resource.
+   */
+  readonly additionalFields?: Partial<DocumentReference>;
```
mattwiller: Should this be `Omit<Partial<DocumentReference>, 'content'>` to prevent overwriting the attachment?

maddyli: fixed!

PR #6746, opened 2025-06-10
Add Community tab to website

maddyli:


rahul1:
@maddyli can you add an icon next to word "Community" to indicate their going to a different domain. https://tabler.io/icons/icon/external-link

maddyli:
@rahul1 my only pushback to that is that the links on the right hand side of the page are also external. Maybe I could move the community to the right side for consistency?

PR #6755, opened 2025-06-11
[Docs] Medplum CLI Agent Commands

maddyli:
Document the commands implemented [here](https://github.com/medplum/medplum/pull/4750)

### DIFF
--- a/packages/docs/docs/agent/agent-cli-commands.md
+++ b/packages/docs/docs/agent/agent-cli-commands.md
@@ -15,7 +15,7 @@ medplum agent status [agentIds...] [options]
 ```
 
 ### Arguments
-- `[agentIds...]`: Optional list of agent IDs to get status for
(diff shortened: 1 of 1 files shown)


packages/docs/docs/agent/agent-cli-commands.md
```diff
@@ -0,0 +1,165 @@
...
+## Status Command
+
+Get the status of one or more agents.
+
+```bash
+medplum agent status [agentIds...] [options]
+```
+
+### Arguments
+- `[agentIds...]`: Optional list of agent IDs to get status for
```
ThatOneBro: We should note that the `agentIds...` syntax and the `--criteria` selector are mutually exclusive for all of these commands that have both

maddyli: Sounds good, there is a note on the criteria about them being mutual exclusive now

PR #6758, opened 2025-06-11
Axp interactions

mattwiller:


packages/core/src/access.ts
```diff
@@ -33,46 +33,40 @@ export const AccessPolicyInteraction = {
   READ: 'read',
   VREAD: 'vread',
   UPDATE: 'update',
-  PATCH: 'patch',
+  // PATCH: 'patch',
```
codyebberson: Cleanup

packages/definitions/dist/fhir/r4/fhir.schema.json
```diff
@@ -61899,9 +61899,16 @@
           "$ref": "#/definitions/string"
         },
         "readonly": {
-          "description": "Optional flag to indicate that the resource type is read-only.",
+          "description": "@deprecated Use AccessPolicy.resource.interaction \u003d [\u0027search\u0027, \u0027read\u0027, \u0027vread\u0027, \u0027history\u0027]",
```
codyebberson: 👍

packages/core/src/access.ts
```diff
@@ -82,19 +65,32 @@ export function canReadResourceType(accessPolicy: AccessPolicy, resourceType: Re
  * @param accessPolicy - The access policy.
  * @param resourceType - The resource type.
  * @returns True if the current user can write the specified resource type.
+ * @deprecated Use accessPolicySupportsInteraction() instead.
  */
 export function canWriteResourceType(accessPolicy: AccessPolicy, resourceType: ResourceType): boolean {
   if (protectedResourceTypes.includes(resourceType)) {
```
mattlong, line 71: Since `accessPolicySupportsInteraction` doesen't consider `protectedResourceTypes`, it seems like this isn't quite a drop-in replacement for `canWriteResourceType`.  Should one of the other methods be recommended to replace it instead?

packages/core/src/access.ts
```diff
@@ -82,19 +65,32 @@ export function canReadResourceType(accessPolicy: AccessPolicy, resourceType: Re
...
-    }
-  }
-  return false;
+  return accessPolicySupportsInteraction(accessPolicy, AccessPolicyInteraction.UPDATE, resourceType);
+}
+
+/**
+ * Shallow check that an interaction is permitted by the AccessPolicy on a given resource type,
+ * at least for some resources.  A more in-depth check for the specific resource(s) being accessed
+ * is required in addition to this one.
```
mattlong, line 80: Could the more in-depth checks that need to be performed be referenced here?

packages/core/src/access.ts
mattwiller: The new logic is correct: Super Admins can access these resource types, so they shouldn't be blocked at the `@medplum/core` level

PR #6761, opened 2025-06-12
feat(agent): support 'enhanced mode' acks

### LINKED ISSUE #6757: Medplum Agent "Fast-ACK" via CA/AE/AA
Implement "enhanced-mode" ACK messages, so that Agent can ACK and allow single-threaded devices to send next message without having to wait for full Agent->Server->Agent turnaround. 

Enhanced-mode statuses documented here: https://hl7-definition.caristix.com/v2/HL7v2.3/Tables/0008

Corepoint support for enhanced-mode statuses: https://rhapsody.health/resources/hl7-acknowledgement-ack/#:~:text=Enhanced%20Mode%20Acknowledgement%C2%A0%E2%80%93%20an%20%E2%80%9CApplication%E2%80%9D%20ACK%20that%20is%20a%20resultant%20status%20return%20rather%20than%20a%20communication%20response%20(i.e.%20query%20results%2C%20order%20response%2C%20etc.)

ThatOneBro:
Fixes #6757

### DIFF
--- a/packages/core/src/hl7.ts
+++ b/packages/core/src/hl7.ts
@@ -1,5 +1,26 @@
 import { isStringArray } from './utils';
 
+export const AckCode = {
+  /** AA - Application Accept */
+  AA: 'AA',
+  /** AE - Application Error */
+  AE: 'AE',
+  /** AR - Application Reject */
+  AR: 'AR',
+  /** CA - Commit Accept */
+  CA: 'CA',
+  /** CE - Commit Error */
+  CE: 'CE',
+  /** CR - Commit Reject */
+  CR: 'CR',
+} as const;
+export type AckCode = keyof typeof AckCode;
+
+export interface Hl7AckOptions {
+  ackCode: AckCode;
+  errSegment?: Hl7Segment | boolean;
+}
+
 /**
  * The Hl7Context class represents the parsing context for an HL7 message.
  *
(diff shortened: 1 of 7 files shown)


packages/hl7/src/connection.ts
```diff
@@ -17,14 +17,16 @@ export type Hl7MessageQueueItem = {
 export class Hl7Connection extends Hl7Base {
   readonly socket: net.Socket;
   readonly encoding: string;
+  readonly enhancedMode: boolean;
```
mattwiller, line 20: Is this property used anywhere?

mattwiller (approved):

packages/hl7/src/connection.ts
ThatOneBro: Not currently, but the intent would be for introspection (ie. for debugging, being able to see what a given connection's `enhancedMode` is set to)

packages/core/src/hl7.ts
```diff
@@ -160,10 +192,96 @@ export class Hl7Message {
...
       ),
-      new Hl7Segment(['MSA', 'AA', controlId, 'OK'], this.context),
+      new Hl7Segment(['MSA', ackCode, controlId, TEXT_MSG_FOR_ACK_CODE[ackCode]], this.context),
+      ...(options?.errSegment
+        ? [options?.errSegment instanceof Hl7Segment ? options.errSegment : this.buildDefaultErrSegment(options)]
+        : []),
     ]);
   }
 
+  private buildDefaultErrSegment(options?: Hl7AckOptions): Hl7Segment {
```
codyebberson: Hey, nice work on this. Good call on adding the `AE` and `CE` support, that's a solid improvement.

On the error-building helpers: I'm thinking we should avoid baking business logic (like default error messages) into the library itself. The library should provide the tools, but the calling application should decide what the error content is.

What do you think about changing the API to just accept a pre-built `HL7Segment` for the error part? That would give the user full control and keep our library's responsibility clean.

Also, I'd prefer to remove the optional "Information" segment from the standard "AA" to keep the success path minimal.

Happy to chat through it if you'd like.

ThatOneBro: No I think that makes sense, removed the default ERR segment and made it so that if you want an errSegment you have to pass it in 👍

codyebberson (approved):
🥳

PR #6766, opened 2025-06-13
Fixes #6355 - implement OAuth RFC 7662 for external auth

### LINKED ISSUE #6355: Implement Token Introspection Authentication
The IUA specification describes how a FHIR server can accept an authorization token from an external auth provider, _on every resource request_

https://infoscribe.infoway-inforoute.ca/pages/viewpage.action?pageId=194838753

This generally follows OAuth2.1 standards. However, one step, labeled [ITI-102 Token Introspection](https://profiles.ihe.net/ITI/IUA/index.html#3102-introspect-token-iti-102), seems to rely on a IHE defined introspection standard.

However, the docs are a bit unclear. This might be completely equivalent to [OAuth RFC7662](https://datatracker.ietf.org/doc/html/rfc7662)

One major design question is how this would work with Medplum's Project setup, and whether this is possible

codyebberson:


packages/core/src/base64.ts
```diff
@@ -36,3 +36,27 @@ export function encodeBase64(data: string): string {
...
   throw new Error('Unable to encode base64');
 }
+
+/**
+ * Encodes a string into Base64URL format.
+ * This is the encoding required for JWT parts.
+ * @param data - The unencoded input string.
+ * @returns The Base64URL encoded string.
+ */
+export function encodeBase64Url(data: string): string {
```
codyebberson, line 46: This is not directly related to the external auth implementation, but it came up and it was bugging me.

There are a number of places where we generate fake JWT's in tests, and they need to be able to be decoded properly. 

In the past, we incorrectly used "base 64 encoding" rather than "base 64 URL encoding", and just got lucky that it didn't matter.

On this PR, some of the fake JWT's were hitting encoding/decoding errors, so I went ahead and added proper utilities to do this correctly.

packages/eslint-config/index.cjs
```diff
@@ -32,7 +32,6 @@ module.exports = {
     'default-param-last': 'error',
     'guard-for-in': 'error',
     'no-array-constructor': 'error',
-    'no-div-regex': 'error',
```
codyebberson: This lint error was triggered on this line:

```ts
.replace(/={1,2}$/, ''); // Remove trailing =
```

Which I guess doesn't bother me.  We could keep it in our ESLint config, and add a single line override if anyone cares.

packages/server/src/oauth/utils.ts
```diff
@@ -777,18 +785,18 @@ function includeRefreshToken(request: LoginRequest): boolean {
 /**
  * Returns the external identity provider user info for an access token.
  * This can be used to verify the access token and get the user's email address.
- * @param idp - The identity provider configuration.
+ * @param userInfoUrl - The user info URL from the identity provider configuration.
  * @param externalAccessToken - The external identity provider access token.
  * @returns The user info claims.
  */
 export async function getExternalUserInfo(
```
codyebberson, line 794: This came together nicely.  We're using the same external userinfo flow for both RFC 7662 (this PR) and RFC 8693 (OAuth 2.0 Token Exchange).

In the future, this should probably be extended to support both `/introspect` and `/userinfo`.  Token introspection is the better option, but it is not as widely supported as user info.

packages/server/src/oauth/utils.ts
```diff
@@ -980,3 +993,145 @@ async function tryAddOnBehalfOf(req: IncomingMessage | undefined, authState: Aut
...
+async function tryExternalAuth(req: Request | undefined, accessToken: string): Promise<AuthState | undefined> {
+  const externalAuthProviders = getConfig().externalAuthProviders;
+  if (!externalAuthProviders) {
+    // No external auth providers configured
+    return undefined;
+  }
+
+  if (!isJwt(accessToken)) {
+    // Not a JWT, so we cannot verify it
+    return undefined;
```
codyebberson, line 1020: Currently requiring JWT's, and not allowing opaque tokens.

In the future, we can revisit this decision.

In theory, we could have a "catch all" auth server for unrecognized access tokens.

I checked with the customer who is sponsoring this work, and they are ok with "JWT only" for now.

packages/server/src/oauth/token.ts
```diff
@@ -673,21 +674,3 @@ function verifyCode(challenge: string, method: string, verifier: string): boolea
...
-
-/**
- * Returns the base64-url-encoded SHA256 hash of the code.
- * The details around '+', '/', and '=' are important for compatibility.
- * See: https://auth0.com/docs/flows/call-your-api-using-the-authorization-code-flow-with-pkce
- * See: packages/client/src/crypto.ts
- * @param code - The input code.
- * @returns The base64-url-encoded SHA256 hash.
- */
-export function hashCode(code: string): string {
```
codyebberson, line 685: I moved this from `oauth/token.ts` to `oauth/utils.ts` because it was causing circular import issues.

packages/eslint-config/index.cjs
mattwiller: If it's a one-off, might as well just fix it to comply with the rule, e.g. `.replace(/[=]{1,2}$/, '');`

packages/core/src/base64.ts
```diff
@@ -36,3 +36,27 @@ export function encodeBase64(data: string): string {
...
+  return encodeBase64(data)
+    .replace(/\+/g, '-') // Replace + with -
+    .replace(/\//g, '_') // Replace / with _
+    .replace(/={1,2}$/, ''); // Remove trailing =
```
mattwiller: Node.js has a built-in `base64url` encoding for `Buffer`, so ideally we would just pass the encoding through to the `Buffer.toString()` in `encodeBase64()`, as an optional parameter

packages/core/src/base64.ts
```diff
@@ -36,3 +36,27 @@ export function encodeBase64(data: string): string {
...
+
+/**
+ * Decodes a string from Base64URL format.
+ * @param data - The Base64URL encoded input string.
+ * @returns The decoded string.
+ */
+export function decodeBase64Url(data: string): string {
+  data = data.padEnd(data.length + ((4 - (data.length % 4)) % 4), '=');
+  const base64 = data.replace(/-/g, '+').replace(/_/g, '/');
+  return decodeBase64(base64);
```
mattwiller, line 67: Similarly, this could just pass the `base64url` encoding through to the `Buffer.from()` in `decodeBase64()`

packages/server/src/oauth/utils.ts
```diff
@@ -980,3 +993,159 @@ async function tryAddOnBehalfOf(req: IncomingMessage | undefined, authState: Aut
...
+ * @returns The base64-url-encoded SHA256 hash.
+ */
+export function hashCode(code: string): string {
+  return createHash('sha256')
+    .update(code)
+    .digest()
+    .toString('base64')
+    .replaceAll('+', '-')
+    .replaceAll('/', '_')
+    .replaceAll('=', '');
```
mattwiller, line 1153:
```suggestion
    .toString('base64url');
```

packages/server/src/oauth/utils.ts
```diff
@@ -801,7 +809,7 @@ export async function getExternalUserInfo(
   }
 
   if (response.status === 429) {
-    log.warn('Auth rate limit exceeded', { url: idp.userInfoUrl, clientId: idp.clientId });
+    log.warn('Auth rate limit exceeded', { url: userInfoUrl });
```
mattwiller: Not having the client ID in this log might make it more difficult to troubleshoot; is there anything else we cold log here to make it easy for on-call to diagnose issues?

PR #6767, opened 2025-06-13
Include resources with missing values when sorting

mattlong:
Before merging, needs to be perf tested. UPDATE: Performance seemed to be largely unchanged.

[Slack thread](https://medplum.slack.com/archives/C051SKJKZ50/p1749841080728269)

Previous PR that introduced the INNER JOIN: https://github.com/medplum/medplum/pull/3823

### DIFF
--- a/packages/server/src/fhir/lookups/humanname.test.ts
+++ b/packages/server/src/fhir/lookups/humanname.test.ts
@@ -241,7 +241,7 @@ describe('HumanName Lookup Table', () => {
       expect(searchResult.entry?.[0]?.resource?.id).toStrictEqual(patient.id);
     }));
 
-  test('Sort by name, missing always last', () =>
+  test('Sort by name with some missing', () =>
     withTestContext(async () => {
       const name = randomUUID();
       const identifier = randomUUID();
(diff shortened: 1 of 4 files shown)


mattwiller (approved):

mattlong:
I was wanting to include a change where FHIR search always includes `NULLS LAST` when sorting, but there's a bit of code in the terminology service that relies on null values being first:

```typescript
    sortRules: [
      // Select highest version (by lexical sort -- no version is assumed to be "current")
      { code: 'version', descending: true },
      // Break ties by selecting more recently-updated resource (lexically -- no date is assumed to be current)
      { code: 'date', descending: true },
    ],
```

I feel like it's better to have consistency across all search parameter types and therefore never specify `NULLS LAST` for now, but that doesn't feel great either... 🤔

mattlong:
> I was wanting to include a change where FHIR search always includes `NULLS LAST` when sorting, but there's a bit of code in the terminology service that relies on null values being first:
> 
> ```ts
>     sortRules: [
>       // Select highest version (by lexical sort -- no version is assumed to be "current")
>       { code: 'version', descending: true },
>       // Break ties by selecting more recently-updated resource (lexically -- no date is assumed to be current)
>       { code: 'date', descending: true },
>     ],
> ```
> 
> I feel like it's better to have consistency across all search parameter types and therefore never specify `NULLS LAST` for now, but that doesn't feel great either... 🤔

Update: Controlling the position of NULLS impacts the ability of the query planner to utilize indexes, so that part was removed from this PR.

PR #6777, opened 2025-06-14
Stop writing to `Binary` search parameter columns

mattlong:
Even though Medplum does not allow search on `Binary`, the default set of search parameters inherited from `Resource` were being populated when writing to the database.

This is step one: stop writing to the search parameter columns for `Binary`.
Step two (removing the columns from the `Binary` table) needs to wait for a minor version bump.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -170,7 +170,8 @@ function getOrInitTypeSchema(resourceType: string): TypeInfo {
     globalSchema.types[resourceType] = typeSchema;
   }
 
-  if (!typeSchema.searchParams) {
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -170,7 +170,8 @@ function getOrInitTypeSchema(resourceType: string): TypeInfo {
     globalSchema.types[resourceType] = typeSchema;
   }
 
-  if (!typeSchema.searchParams) {
+  // Binary has no search parameters; not even those inherited from Resource
+  if (!typeSchema.searchParams && resourceType !== 'Binary') {
```
codyebberson, line 174: 👍

PR #6778, opened 2025-06-14
Fixed bot execute with gcp storage

codyebberson:


packages/server/src/cloud/gcp/storage.ts
```diff
@@ -15,18 +15,37 @@ export class GoogleCloudStorage extends BaseBinaryStorage {
...
+    if (isString(data)) {
+      // For strings, the file.save() method is the most direct and simplest way.
+      // It handles encoding and buffering automatically.
+      await file.save(data, options);
+    } else {
+      // For streams, we pipe the data to a write stream provided by the client library.
+      // We wrap this callback-based stream logic in a Promise to work seamlessly with async/await.
+      await new Promise<void>((resolve, reject) => {
+        const writeStream = file.createWriteStream(options);
+        (data as Readable)
```
ThatOneBro: Should we handle the case where `data` is a `Uint8Array` before the else?

packages/server/src/cloud/gcp/storage.ts
```diff
@@ -15,18 +15,37 @@ export class GoogleCloudStorage extends BaseBinaryStorage {
...
 
-  async writeFile(key: string, contentType: string | undefined, stream: BinarySource): Promise<void> {
+  /**
+   * Writes a file to Google Cloud Storage.
+   * This method now correctly handles both Readable streams and strings as input.
+   * @param key - The key (path) for the file in the bucket.
+   * @param contentType - The MIME type of the content.
+   * @param data - The content to write, either as a Readable stream or a string.
+   */
+  async writeFile(key: string, contentType: string | undefined, data: BinarySource): Promise<void> {
```
ThatOneBro, line 25: Wait I think this interface is wrong isn't it?

Ohh are we going to have to make the `Storage` interface more generic? 🤔

packages/server/src/cloud/gcp/storage.ts
ThatOneBro: Oh I think this is a workaround for the strict typing of the interface, yeah?

codyebberson: Ahh, my bad.  This cast is unnecessary.

> Should we handle the case where data is a Uint8Array before the else?

No, that shouldn't be necessary here, because we are in `server`:

```ts
export type BinarySource = Readable | string;
```

Which is confusing, because it's different in `core`:

```ts
export type BinarySource = string | File | Blob | Uint8Array;
```

🙃

packages/server/src/cloud/gcp/storage.ts
codyebberson: I think it's ok, yeah?

Interface definition in `packages/server/src/storage/types.ts`:

```ts
  writeFile(key: string, contentType: string | undefined, stream: BinarySource): Promise<void>;
```

I think it gets back to how `BinarySource` is different in `core` and `server`

ThatOneBro: Yeah I was looking at the wrong definition, I thought we were importing from `@medplum/core`. LGTM

PR #6784, opened 2025-06-16
Add `Project.exportedResourceType` to limit which linked projects are searched 

mattlong:
Linked projects can now declare which resource types they export when linked to another project. When `Project.exportedResourceType` is specified, the linked project will only be included when searching for those resource types.

Previously, Medplum included every linked project when searching for resources. Due to the ["Ugly OR" problem](https://www.cybertec-postgresql.com/en/avoid-or-for-better-performance/), this makes it impossible to take advantage of multi-column indexes prefix with `projectId`, a pattern that we imminently want to start using.


packages/server/src/fhir/accesspolicy.ts
```diff
@@ -27,13 +34,24 @@ export async function getRepoForLogin(authState: AuthState, extendedMode?: boole
...
+      }
+    }
+
+    const linkedProjectsOrError = await getSystemRepo().readReferences<Project>(linkedProjectRefs);
+    for (const linkedProjectOrError of linkedProjectsOrError) {
+      if (isResource(linkedProjectOrError)) {
+        allowedProjects.push(linkedProjectOrError);
+      } else {
+        //TODO{mattlong}: Log error? Do something? Fail? Ignore it?
+        // Ignore non-existent linked projects
```
codyebberson: Good question.  "Ignore" would be the correct equivalent of current behavior, because if a super admin creates a project link to a non-existent project, it would just be a no-op.

Maybe `getLogger().debug('Linked project not found...')`?

packages/server/src/fhir/repo.ts
```diff
@@ -407,7 +407,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
   }
 
   private canReadCacheEntry(cacheEntry: CacheEntry): boolean {
-    if (!this.isSuperAdmin() && !this.context.projects?.includes(cacheEntry.projectId)) {
+    if (!this.isSuperAdmin() && !this.context.projects?.map((p) => p.id).includes(cacheEntry.projectId)) {
```
codyebberson: Micro optimization:

```suggestion
    if (!this.isSuperAdmin() && !this.context.projects?.some((p) => p.id === cacheEntry.projectId)) {
```

packages/server/src/fhir/repo.ts
```diff
@@ -416,10 +416,10 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     return true;
   }
 
-  async readReferences<T extends Resource>(references: Reference<T>[]): Promise<(T | Error)[]> {
+  async readReferences<T extends Resource>(references: Reference<T>[]): Promise<(WithId<T> | Error)[]> {
```
codyebberson, line 419: 👍

codyebberson (approved):
Nice, lgtm 👍 

Some minor feedback, at your discretion

PR #6787, opened 2025-06-17
MCP server

codyebberson:
Adds optional MCP server support.  Must opt-in using `mcpEnabled: true` server config setting.

We expose 3 MCP "tools", 1 real tool and 2 dummy tools:
- Real tool
   - `fhir-request` - Make a Medplum FHIR request
      - `method` - the HTTP method (i.e., `GET`, `POST`, `DELETE`, etc)
      - `path` - the FHIR path (i.e., `/Patient`, `/Patient/123`, `/Patient/123/$everything`, etc)
      - `body` - optional FHIR request body, can be string or JSON
- Dummy tools - these are required by some major LLMs, but are not applicable to Medplum
   - `search` - This is supposed to be for free text search for the purpose of finding related articles
   - `fetch` - This is supposed to be for retrieving articles for citations

This is largely just a thin adapter/wrapper around our existing API.

The one tricky part was implementing the SSE transport, which requires a persistent connection similar to WebSockets.  Technically, the SSE transport is deprecated, but at the time of this writing it is the only transport supported by Claude.

To work around that limitation, we used the Redis pub/sub model, exactly how we handle WebSocket subscriptions and Agent WebSocket connections.

TODO:
- [x] Add optional "default scopes" to `ClientApplication`
- [x] Add tests for `/oauth2/register`
- [x] Add MCP tests
- [x] Figure out how to do SSE with multiple boxes behind load balancer
- [x] Add SSE heartbeat

Proof it works:

image

### DIFF
--- a/packages/server/medplum.config.json
+++ b/packages/server/medplum.config.json
@@ -21,7 +21,6 @@
   "defaultBotRuntimeVersion": "vmcontext",
   "allowedOrigins": "*",
   "introspectionEnabled": true,
-  "mcpEnabled": true,
   "database": {
     "host": "localhost",
     "port": 5432,
--- a/packages/server/medplum.config.json
+++ b/packages/server/medplum.config.json
@@ -21,6 +21,7 @@
   "defaultBotRuntimeVersion": "vmcontext",
   "allowedOrigins": "*",
   "introspectionEnabled": true,
+  "mcpEnabled": true,
   "database": {
     "host": "localhost",
     "port": 5432,
--- a/packages/server/src/mcp/server.ts
+++ b/packages/server/src/mcp/server.ts
@@ -4,6 +4,7 @@ import fetch from 'node-fetch';
 import { z } from 'zod';
 import { getConfig } from '../config/loader';
 import { getAuthenticatedContext } from '../context';
+import { getLogger } from '../logger';
 
 const server = new McpServer({
   name: 'medplum',
@@ -25,7 +26,7 @@ const dummyDocument = {
 } as const;
 
 server.tool('search', { query: z.string() }, async ({ query }) => {
-  console.log(`Performing search for: "${query}"`);
+  getLogger().debug(`Performing search for: "${query}"`);
   return { content: [dummyDocument] };
 });
 
@@ -35,7 +36,7 @@ server.tool(
     id: z.string().describe('The ID of the resource to fetch, obtained from a search result.'),
   },
   async ({ id }) => {
-    console.log(`Performing fetch for ID: "${id}"`);
+    getLogger().debug(`Performing fetch for ID: "${id}"`);
     return { content: [dummyDocument] };
   }
 );
@@ -64,7 +65,7 @@ server.tool(
       body = JSON.parse(body);
     }
 
-    let response: any;
+    let response: unknown;
     switch (method.toLowerCase()) {
       case 'get':
         response = await proxy.get(fhirUrl);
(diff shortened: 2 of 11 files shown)


codyebberson:
Ping

image

packages/server/src/mcp/server.ts
```diff
@@ -0,0 +1,102 @@
+import { concatUrls, isString, MEDPLUM_VERSION, MedplumClient } from '@medplum/core';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
```
mattlong, line 2: Curious why the `.js` are necessary on all the imports from `@modelcontextprotocol/sdk`.

packages/server/src/mcp/server.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+const dummyDocument = {
+  type: 'text',
+  id: 'dummy-doc',
+  text: 'This is a dummy document used for testing purposes.',
+  uri: 'https://example.com/dummy-doc',
+} as const;
+
+server.tool('search', { query: z.string() }, async ({ query }) => {
+  console.log(`Performing search for: "${query}"`);
```
mattlong:
```suggestion
```

packages/server/src/mcp/server.ts
```diff
@@ -0,0 +1,102 @@
...
+  return { content: [dummyDocument] };
+});
+
+server.tool(
+  'fetch',
+  {
+    id: z.string().describe('The ID of the resource to fetch, obtained from a search result.'),
+  },
+  async ({ id }) => {
+    console.log(`Performing fetch for ID: "${id}"`);
```
mattlong:
```suggestion
```

packages/server/src/mcp/server.ts
```diff
@@ -0,0 +1,102 @@
...
+    const fhirUrl = concatUrls(baseFhirUrl, path);
+    const accessToken = ctx.authState.accessToken;
+    const proxy = new MedplumClient({ baseUrl, accessToken, fetch });
+
+    // MCP allows sending JSON, but some clients (like Claude) send the body as a string
+    if (isString(body)) {
+      body = JSON.parse(body);
+    }
+
+    let response: any;
```
mattlong: Can this be something besides `any`?

packages/server/medplum.config.json
```diff
@@ -21,6 +21,7 @@
   "defaultBotRuntimeVersion": "vmcontext",
   "allowedOrigins": "*",
   "introspectionEnabled": true,
+  "mcpEnabled": true,
```
mattlong: Just checking that this is intentionally being enabled by default for the dev config file.

packages/server/src/mcp/server.ts
codyebberson: That's how it is in their docs: https://github.com/modelcontextprotocol/typescript-sdk

Let me try removing it.

The SDK has an interesting way of doing exports, which I guess is the new recommended pattern because it's better for tree shaking: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/package.json#L21-L26

packages/server/medplum.config.json
codyebberson: I debated that.  I'm ok to leave it out for now.

packages/server/src/mcp/server.ts
codyebberson: image

mattlong (approved):

PR #6790, opened 2025-06-17
OAuth updates for MCP

codyebberson:
Extracting just the OAuth changes from the MCP PR: https://github.com/medplum/medplum/pull/6787

IMHO, the actual MCP implementation is too early for a production server, and I do not feel comfortable adding it to `server` at this time.  We should be able to achieve more/less the same outcome using a separate standalone service that proxies API calls.  My plan is to setup something separately on `https://mcp.medplum.ai`)

The MCP investigation did reveal a few gaps in our OAuth/OIDC implementation, which are addressed by this PR:

1. There are some subtle differences in the `/authorize` endpoint between OAuth2 and OIDC, notably where the `scope` param is required or not.  To address that, this PR adds `defaultScope` to the `ClientApplication` type, which allows you to determine the behavior you want.
2. MCP requires OAuth 2.0 "Dynamic Client Registration Protocol" (DCR).  We're not going to allow full public access to DCR.  Instead, this whitelists some client applications by `redirect_url`.  We already had a similar flow for the Medplum CLI auth step (i.e., what happens when you run `npx medplum login`).  This PR formalizes those changes.
3. Added `/oauth-authorization-server` and `/oauth-protected-resource` to `.well-known`.  Required by MCP, and generally seems like good hygeine.

### DIFF
--- a/packages/server/src/oauth/token.test.ts
+++ b/packages/server/src/oauth/token.test.ts
@@ -71,6 +71,7 @@ describe('OAuth2 Token', () => {
   let client: WithId<ClientApplication>;
   let pkceOptionalClient: ClientApplication;
   let externalAuthClient: ClientApplication;
+  let invalidAuthClient: ClientApplication;
 
   beforeAll(async () => {
     config = await loadTestConfig();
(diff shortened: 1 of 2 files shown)


packages/server/src/oauth/clients.ts
```diff
@@ -0,0 +1,57 @@
...
+import { randomUUID } from 'crypto';
+
+/**
+ * Standard clients embeded in the Medplum server.
+ *
+ * Medplum CLI is represented as a special client with "medplum-cli" as its ID.
+ *
+ * Claude and ChatGPT are also included as standard clients as required by the MCP protocol.
+ */
+const standardClients: ClientApplication[] = [
```
mattlong: Should all/some of these entries live in server config instead of being hardcoded?

codyebberson: That's a good idea 👍

codyebberson:
```json
[
  {
    "resourceType": "ClientApplication",
    "id": "mcp-inspector-1",
    "name": "MCP Inspector 1",
    "redirectUri": "http://127.0.0.1:6274/oauth/callback",
    "pkceOptional": true
  },
  {
    "resourceType": "ClientApplication",
    "id": "mcp-inspector-2",
    "name": "MCP Inspector 2",
    "redirectUri": "http://localhost:6274/oauth/callback",
    "pkceOptional": true
  },
  {
    "resourceType": "ClientApplication",
    "id": "claude",
    "name": "Claude",
    "redirectUri": "https://claude.ai/api/mcp/auth_callback",
    "pkceOptional": true
  },
  {
    "resourceType": "ClientApplication",
    "id": "chatgpt",
    "name": "ChatGPT",
    "secret": "foo",
    "redirectUri": "https://chatgpt.com/connector_platform_oauth_redirect",
    "pkceOptional": true
  }
]
```

mattlong (approved):

PR #6794, opened 2025-06-17
Enable limiting number of resources per Project

mattwiller:


packages/server/src/fhir/operations/expunge.ts
```diff
@@ -16,7 +16,7 @@ import { buildBinaryIds } from './utils/binary';
  */
 export async function expungeHandler(req: FhirRequest): Promise<FhirResponse> {
   const ctx = getAuthenticatedContext();
-  if (!ctx.project.superAdmin) {
+  if (!ctx.project.superAdmin && !ctx.membership.admin) {
```
mattlong, line 21: The docs will need updating for this; they mention that expunge is only available to super admins right now.

https://www.medplum.com/docs/fhir-datastore/deleting-data#expunge-operation

packages/server/src/fhir/resource-cap.ts
```diff
@@ -0,0 +1,114 @@
...
+      // If user is already over the limit, just block them
+      throw new OperationOutcomeError(businessRule('resource-cap', 'Resource cap exceeded'));
+    }
+
+    try {
+      const result = await this.limiter.consume(this.projectKey, points);
+      this.setState(result);
+    } catch (err: unknown) {
+      // Give back the point consumed when attempting to create a resource, since it didn't get created
+      await this.limiter.reward(this.projectKey, 1);
```
mattlong: If the error happens while attempting to consume `points`, i.e. before the consume has made it to Redis, wouldn't this be giving back points that were never actually consumed?

Also, why is this `1` instead of `points`?

packages/server/src/fhir/resource-cap.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      if (err instanceof Error && this.enabled) {
+        throw err;
+      }
+      const result = err as RateLimiterRes;
```
mattlong: Is there some way to validate/check this cast?

packages/server/src/fhir/resource-cap.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.initPromise = undefined;
+  }
+
+  /**
+   * Consume from resource cap in Redis store.
+   * @param points - Number of resources consumed.
+   */
+  async consume(points: number): Promise<void> {
+    if (!this.current) {
+      this.initPromise ??= this.init();
```
mattlong, line 73: Could `this.init()` be made idempotent on its own so that `consume()` and `created()` could do `await this.init()` instead?

packages/server/src/fhir/resource-cap.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    let currentStatus = await this.limiter.get(this.projectKey);
+    if (!currentStatus) {
+      const subqueries = countedResourceTypes.map((rt) =>
+        new SelectQuery(rt).raw(`COUNT(*)::int as "count"`).where('projectId', '=', this.projectKey)
+      );
+      const query = new SelectQuery('combined', new Union(...subqueries)).column('count');
+      const tableCounts = await query.execute(getDatabasePool(DatabaseMode.READER));
+      const totalCount = tableCounts.reduce((sum, row) => sum + row.count, 0);
+      currentStatus = await this.limiter.set(this.projectKey, totalCount, ONE_DAY);
```
mattlong, line 57: Doing the `UNION` of counts across all resource types could of course be fairly slow. Once per day for doing this of course isn't horrible, but is there someway we could generally have this done out of band instead of during a request? It seems like having a recurring async job that will refresh/update all rate limiter's data could do the trick. I know that's future work from v1 of this, but curious if there are any plans for how that might work.

mattwiller: Agreed, the `UNION` isn't ideal to run in band, but each subquery should at least be relatively cheap — essentially an index scan on `projectId`.  I think in the future we could convert this to a cron job that loops through all Projects on the server and updates their rate limit buckets, but this might actually be worse.  Since many projects are unused on any given day, the `UNION` queries to count their resources, even if those counts are very low, would be purely wasted work.  The existing solution at least minimizes this to calculating only for Projects that are in active use

codyebberson:
@claude please review

mattlong (approved):

PR #6795, opened 2025-06-18
Ianplunkett/docker compose pre commit

ianplunkett:
The diff is here a bit large b/c `cody-pre-commit-bot-proof-of-concept` needs rebased against main.

The main things here are
 1. a bot that demonstrates the pre-commit hook with both `UPDATE` and `DELETE` operations
 2. a refactoring of the pre-commit hook logic so that it can apply to updates and deletes
 3. a hapi server in the docker-compose to test against.


packages/server/medplum.config.json
```diff
@@ -34,8 +34,18 @@
...
+    }
   },
-  "shutdownTimeoutMilliseconds": 30000
+  "shutdownTimeoutMilliseconds": 30000,
+  "externalAuthProviders": [
+    {
+      "issuer": "http://127.0.0.1:4444",
+      "userInfoUrl": "http://127.0.0.1:4444/userinfo"
+    }
+  ]
```
ianplunkett: This config shouldn't make it's way into main, but is useful for the external auth POC

docker-compose.full-stack.yml
```diff
@@ -121,3 +121,9 @@ services:
       interval: 10s
       timeout: 5s
       retries: 5
+  # Hapi container
+  hapi-server:
+    image: hapiproject/hapi:latest
+    restart: always
+    ports:
+      - "8080:8080"
```
ianplunkett: This config shouldn't make its way into main, but is useful for the pre-commit POC

codyebberson (approved):

PR #6801, opened 2025-06-18
Dependency upgrades 2025-06-18

medplumbot:
Dependency upgrades


codyebberson:
The reason the `package-lock.json` diff is so big is because `jest-expo`  still requires Jest 29, so we now have both Jest 29 and Jest 30.

I'm going to revert Jest back to 29.

codyebberson (approved):

PR #6809, opened 2025-06-19
Remove Buffer base64url optimization

codyebberson:
Context: https://medplum.slack.com/archives/C091LPUQBHP/p1750358453288879

Next.js injects a `Buffer` polyfill, and it does not support the `base64url` encoding.


mattwiller (approved):
🪦

PR #6823, opened 2025-06-21
Provider  - use real ICD codes

techdavidy:
Screenshot 2025-06-20 at 5 22 25 PM

### DIFF
--- a/examples/medplum-provider/src/components/Conditions/ConditionModal.tsx
+++ b/examples/medplum-provider/src/components/Conditions/ConditionModal.tsx
@@ -58,7 +58,7 @@ export default function ConditionModal(props: ConditionDialogProps): JSX.Element
         />
 
         <CodeableConceptInput
-          binding="http://hl7.org/fhir/ValueSet/icd-10"
+          binding="http://medplum.com/fhir/ValueSet/icd-codes"
           label="ICD-10 Code"
           name="diagnosis"
           path="Condition.code"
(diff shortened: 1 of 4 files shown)


examples/medplum-provider/src/components/Conditions/ConditionModal.tsx
```diff
@@ -58,7 +58,7 @@ export default function ConditionModal(props: ConditionDialogProps): JSX.Element
         />
 
         <CodeableConceptInput
-          binding="http://hl7.org/fhir/ValueSet/icd-10"
+          binding="http://medplum.com/fhir/ValueSet/icd-codes"
```
codyebberson: Where does this ValueSet exist?

(i.e., did we create it in a test project?  should we add it to the R4 or the UMLS projects?)

codyebberson: As discussed offline:

1. Let's move the `ValueSet` to `packages/definitions/dist/fhir/r4/valuesets-medplum.json`
   * Then it will be added to the `R4` project after running "Rebuild Value Sets"
   * Then it will be automatically available to all projects
2. Let's use `https://` rather than `http://`

techdavidy: Using this URL "http://hl7.org/fhir/sid/icd-10-cm/vs" available from UMLS implementation.

techdavidy:
> show the appropriate warning message.

https://github.com/medplum/medplum/issues/6869 got it

PR #6829, opened 2025-06-21
Add a Redis-based server registry

mattlong:
To facilitate destructive migrations, e.g. columns or tables being deleted, by adding a mechanism to ascertain when all server instances of a previous version have stopped after a rolling deploy.

As an example `4.1.10` currently writes to token lookup tables and reads from them (depending on server configuration). In version `4.2.x`, it will not be possible to read from token lookup tables; they have been deprecated. Ideally, a post-deploy migration added in `v4.2.0` would be able to delete the token lookup tables, but for any deployment of Medplum utilizing rolling deploys as is advised, this would lead to errors during the rolling deploy window since the old server instance would still be attempting to write to the tables deleted in the `v4.2.0` post-deploy migration. If server registry information were available, `4.2.0` could delay running the post-deploy migration until it know with certainty that all old server instances had stopped. 

[Demo video](https://www.loom.com/share/1605f26b6f2444f590d7d45393b462e6?sid=dfa8b1f5-0a38-4728-8ebb-7320ec84a3aa) of a prototype of this system in action.

While this isn't being used directly right now, start having the data collected now so that it can be relied on in future minor versions.

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -15,7 +15,7 @@ import { getConfig } from './config/loader';
 import { MedplumServerConfig } from './config/types';
 import { attachRequestContext, AuthenticatedRequestContext, closeRequestContext, getRequestContext } from './context';
 import { corsOptions } from './cors';
-import { closeDatabase, initDatabase } from './database';
+import { closeDatabase, initDatabase, maybeAutoRunPendingPostDeployMigration } from './database';
 import { dicomRouter } from './dicom/routes';
 import { emailRouter } from './email/routes';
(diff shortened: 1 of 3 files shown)


reshmakh:
Video was very interesting.  Looks like safety and observability are the major drivers - am I right?

codyebberson (approved):

mattlong:
> Video was very interesting. Looks like safety and observability are the major drivers - am I right?

Yep! Safety was the primary motivator, but this will indeed be a very nice addition for observability.

PR #6838, opened 2025-06-23
Dependency upgrades 2025-06-23

medplumbot:
Dependency upgrades


codyebberson:
1. `jest-expo` still not compatible with Jest 30, so reverting all Jest changes
2. Storybook 9 seems stable, but some plugins don't have updates.  I suspect those plugins are deprecated.  Going to investigate and possibly remove them.

codyebberson:
[`storybook-addon-mantine`](https://github.com/josiahayres/storybook-addon-mantine) - support for Storybook 9 only works with Mantine 8.

So... effectively that means we're going to hold back Mantine 8, Storybook 9, and `storybook-addon-mantine` untile Medplum 5 in the fall.

https://github.com/josiahayres/storybook-addon-mantine/releases/tag/v6.0.1

codyebberson (approved):

PR #6842, opened 2025-06-23
Write Coding.isSynonym in support of Synonyms in terminology

mattwiller:
Started writing to the new `"Coding"."isSynonym"` column in the `CodeSystem/$import` operation, and added a data migration to backfill the column in preparation for making it `NOT NULL` in the future.  This migration will simplify the table and associated indexes down the road


codyebberson (approved):
👍 

For my own understanding, we're not setting `isSynonym` to `true` anywhere yet, and that's coming in future PR?

packages/server/src/migrations/data/v16.ts
```diff
@@ -0,0 +1,38 @@
...
+
+async function run(client: PoolClient, results: MigrationActionResult[]): Promise<void> {
+  await fns.query(client, results, `UPDATE "Coding" SET "isSynonym" = false WHERE "isSynonym" IS NULL`);
+  await fns.query(client, results, `ALTER TABLE IF EXISTS "Coding" ALTER COLUMN "isSynonym" SET NOT NULL`);
+
+  await fns.idempotentCreateIndex(
+    client,
+    results,
+    'Coding_system_code_preferred_idx',
+    `CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "Coding_system_code_preferred" ON "Coding" ("system", "code") INCLUDE ("id") WHERE (NOT "isSynonym")`
```
mattlong: The index name here should match the one on the line above but it's missing the `_idx`.

packages/server/src/fhir/operations/utils/terminology.ts
```diff
@@ -80,6 +80,16 @@ function sameTerminologyResourceVersion(a: TerminologyResource, b: TerminologyRe
...
 }
 
+export function selectCoding(systemId: string, ...code: string[]): SelectQuery {
+  return new SelectQuery('Coding')
+    .column('id')
+    .column('code')
+    .column('display')
+    .where('system', '=', systemId)
+    .where('code', 'IN', code)
+    .where('isSynonym', '=', false);
```
mattlong, line 90: Is this meant to be used in conjunction with the new `Coding_system_code_preferred_idx` index? We should confirm that the query planner is smart enough to map`isSynonym = false` to `NOT isSynonym` or switch the conditional part of the index definition to be `isSynonym = false`.

PR #6856, opened 2025-06-24
Add external identifiers to pre-commit hooks

### LINKED ISSUE #6858: Pre-commit Bots return modified resource
Context: https://medplum.slack.com/archives/C07BTRE5BQC/p1750725213733749

ianplunkett:
fixes #6858

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -664,8 +664,8 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     versionId?: string
   ): Promise<WithId<T>> {
     const interaction = create ? AccessPolicyInteraction.CREATE : AccessPolicyInteraction.UPDATE;
-    const { resourceType, id } = resource;
     let validatedResource = this.checkResourcePermissions(resource, interaction);
+    const { resourceType, id } = validatedResource;
 
     const preCommitResult = await preCommitValidation(
       this.context.author,
@@ -681,7 +681,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       validatedResource = this.checkResourcePermissions(preCommitResult, interaction);
     }
 
-    const existing = create ? undefined : await this.checkExistingResource<T>(resourceType, id as string);
+    const existing = create ? undefined : await this.checkExistingResource<T>(resourceType, id);
     if (existing) {
       (existing.meta as Meta).compartment = this.getCompartments(existing); // Update compartments with latest rules
       if (!this.canPerformInteraction(interaction, existing)) {


packages/server/src/fhir/repo.ts
```diff
@@ -661,7 +661,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       throw new OperationOutcomeError(forbidden);
     }
 
-    await preCommitValidation(this.context.projects?.[0], resource, 'update');
+    resource = ((await preCommitValidation(this.context.projects?.[0], resource, 'update')) as WithId<T>) ?? resource;
```
codyebberson: Cool, yes, this makes sense.

packages/server/src/fhir/precommit.ts
```diff
@@ -25,22 +25,28 @@ export function isPreCommitSubscription(subscription: WithId<Subscription>): boo
...
  * @param interaction - The interaction type (e.g., 'create', 'update', 'delete').
+ * @returns The validated resource if a pre-commit bot returns one, otherwise undefined.
  */
 export async function preCommitValidation(
+  author: Reference,
   project: WithId<Project> | undefined,
   resource: WithId<Resource>,
   interaction: BackgroundJobInteraction
-): Promise<void> {
+): Promise<Resource | boolean | undefined> {
```
rahul1: is returning `undefined` useful here? Should we just return `true` in these exit cases?

packages/server/src/fhir/repo.ts
```diff
@@ -654,14 +649,33 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+  }
+
+  private async updateResourceImpl<T extends Resource>(
+    resource: T,
+    create: boolean,
+    versionId?: string
+  ): Promise<WithId<T>> {
+    const interaction = create ? AccessPolicyInteraction.CREATE : AccessPolicyInteraction.UPDATE;
+    const { resourceType, id } = resource;
+    resource = this.checkResourcePermissions(resource, interaction) as T;
```
codyebberson: Hm... we try to avoid casts like this, especially deep in the bowels of the update code.

I suspect the reason you needed `as T` here is because `checkResourcePermissions` returns `WithId<T>` but `resource` is `T`.

I wonder if the better fix would be to change `resource` from `T` to `WithId<T>`

packages/server/src/fhir/repo.ts
```diff
@@ -654,14 +649,33 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    const preCommitResult = await preCommitValidation(
+      this.context.author,
+      this.context.projects?.[0],
+      resource as WithId<Resource>,
+      'update'
+    );
 
-    const existing = create ? undefined : await this.checkExistingResource<T>(resourceType, id);
+    if (isResourceWithId(preCommitResult, resource.resourceType) && preCommitResult.id === resource.id) {
+      resource = this.checkResourcePermissions(preCommitResult as T, interaction) as T;
```
codyebberson: Both of these casts should be removed.

We should make sure that `isResourceWithId` asserts correctly (I think it does).

I think this will be a matter of changing the `resource` parameter type from `T` to `WithId<T>` in both `checkResourcePermissions` and `updateResourceImpl`

packages/server/src/fhir/repo.ts
```diff
@@ -673,9 +687,9 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       }
     }
 
-    let updated = await rewriteAttachments(RewriteMode.REFERENCE, this, {
+    let updated = (await rewriteAttachments(RewriteMode.REFERENCE, this, {
       ...this.restoreReadonlyFields(resource, existing),
-    });
+    })) as WithId<T>;
```
codyebberson: This cast should not be necessary.

codyebberson:
@claude please review

codyebberson (approved):

PR #6872, opened 2025-06-26
Adds information about the deployment of bots to the Twilio demo README

filipeximenes:


examples/medplum-demo-bots/esbuild-script.mjs
```diff
@@ -22,7 +24,7 @@ const esbuildOptions = {
     '.ts': 'ts', // Load TypeScript files
   },
   resolveExtensions: ['.ts', '.js'],
-  external: botLayerDeps,
+  external: [...botLayerDeps, ...additionalExternals],
```
codyebberson, line 27: Do you recall why this was necessary?  Isn't `twilio` now in `botLayer.dependencies` ?

Does this only apply to older versions before `twilio` was added?

examples/medplum-demo-bots/src/twilio/README.md
```diff
@@ -72,6 +72,32 @@ This bot provides a foundation for voice call handling. You can extend it to:
...
+const botLayerDeps = Object.keys(botLayer.dependencies);
+const additionalExternals = ['twilio'];
+
+esbuild.build({
+  ...
+  external: [...botLayerDeps, ...additionalExternals],
+});
+```
+
+For this to work as expected you also need to be careful with the way you import from `twilio`. Use the following pattern to access internal modules of the `twilio` package.
```
codyebberson, line 92: Hmm... Interesting and unfortunate.

Do you know why this is the case?

I know that AWS Lambda Layers has some weird subtle differences between CommonJS and ESM.  Is that related?

It would be nice if we could add an explanation why this change is necessary.

examples/medplum-demo-bots/esbuild-script.mjs
filipeximenes: Actually, you are right, this was based on my setup when I was testing the bundle before we added twilio. Upgrading to the latest Medplum version shouldn't require this. Should I remove these changes or add some kind o warning about older versions?

examples/medplum-demo-bots/src/twilio/README.md
filipeximenes: @codyebberson I was just doing some tests in order to give you a better answer and It seems like I got this all wrong. I was having some issues with my Twilio bots and I think in the process of fixing it I mistakenly attributed the issue to the way I was doing imports. I think we can close this PR. I'll keep testing and will reopen it in case I notice any inconsistencies, but as of now it looks like it can be imported either way.

filipeximenes:
It seems like I got this wrong (see https://github.com/medplum/medplum/pull/6872#discussion_r2175243867 for more context), closing this for now.

PR #6881, opened 2025-06-26
Fix mocked `dataVersion` tests

mattlong:
The previous hard-coded `1` was a ticking time bomb for the `4.2.0` release. The mocked value wasn't relevant to the tests it was causing to fail: https://github.com/medplum/medplum/actions/runs/15911694790/job/44880758363?pr=6877


mattlong:
This was included in #6877

PR #6884, opened 2025-06-27
Fixed ProjectMembership-profile search param types

codyebberson:


ThatOneBro (approved):

codyebberson:
> Should we also update the `ProfileResource` type in `@medplum/core`?

That's a good call - unfortunately that's going to be a bit more involved.  Many of the React components use `ProfileResource` for "current user", which ends up setting properties such as `Communication.sender`, `Media.operator`, `Questionnaire.source`, etc.

In practice, I don't think `Bot` or `ClientApplication` will be logging into a UI and using our React components, so this feels OK.

The real solution is probably to have some notion of:
* `ProfileResource` - any resource type that can be logged in
* `ReactUserProfileResource` (naming is hard) - any resource that can actually be a logged in user with the React components

PR #6890, opened 2025-06-27
fix(ci): fix the paths for published agent artifacts

ThatOneBro:
Also adds official published release versions for ARM64 Linux agent

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -236,14 +236,14 @@ jobs:
               repo: context.repo.repo,
               release_id: release.data.id,
               name: "medplum-agent-installer-" + process.env.MEDPLUM_VERSION + ".exe",
-              data: await fs.readFileSync("packages/agent/medplum-agent-installer-" + process.env.MEDPLUM_VERSION + ".exe")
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -313,13 +311,89 @@ jobs:
...
               owner: context.repo.owner,
               repo: context.repo.repo,
               release_id: release.data.id,
-              name: "medplum-agent-" + packageJson.version + "-linux.sha256",
-              data: await fs.readFileSync("packages/agent/medplum-agent-" + packageJson.version + "-linux.sha256")
+              name: "medplum-agent-" + process.env.MEDPLUM_VERSION + "-linux-x64.sha256",
+              data: await fs.readFileSync(`packages/agent/medplum-agent-${process.env.MEDPLUM_VERSION}-${process.env.MEDPLUM_GIT_SHORTHASH}-linux-x64.sha256`)
+            });
+
+  build_agent_linux-arm64:
```
codyebberson: Would it be possible to use a matrix strategy rather than duplicating the block?

codyebberson (approved):
Makes sense.

Preference to somehow "parameterize" and reuse existing block if that's possible, but I understand if it's not.

.github/workflows/publish.yml
ThatOneBro: Yeah I found a way to make it happen 🚀

PR #6906, opened 2025-06-29
Stop writing and drop old token implementations

mattlong:
This PR must be a part of `4.3.0`

TODOs
- [ ] https://github.com/medplum/medplum/pull/6905 merged


packages/server/src/workers/post-deploy-migration.ts
```diff
@@ -61,6 +65,17 @@ export const initPostDeployMigrationWorker: WorkerInitializer = (config) => {
...
 
+export async function isClusterCompatible(migrationNumber: number): Promise<boolean> {
+  if (!enforceStrictMigrationVersionChecks()) {
+    return true;
+  }
+
+  const servers = await getRegisteredServers(true);
+  const entry = getPostDeployManifestEntry(migrationNumber);
+  const requiredVersion = entry.serverVersion;
+  return servers.every((server) => semver.gte(server.version, requiredVersion));
```
mattlong, line 76: For those following `main` rather than waiting for versioned releases like api.medplum.com, the expected behavior from this check is that the cluster will be considered incompatible until the official `entry.serverVersion` release, e.g. `4.3.0`, is deployed. The result will be the post-deploy migration repeatedly delayed in the interim.

While not ideal, this shouldn't necessarily be problematic. We may want to come up with an alternative that isn't strictly based on numbered releases. Hmmm...

mattlong:
@claude please review

packages/server/src/migrations/data/v15.ts
```diff
@@ -0,0 +1,188 @@
...
+        await run(client, results);
+        return results;
+      });
+    });
+  },
+};
+
+// prettier-ignore
+async function run(client: PoolClient, results: MigrationActionResult[]): Promise<void> {
+  await fns.query(client, results, `DROP TABLE IF EXISTS "ValueSetElement"`);
```
mattlong, line 24: Usage of `ValueSetElement` stopped in #4995

PR #6911, opened 2025-06-30
feat(server/ws): add WebSocket subscriptions per-user limit

### LINKED ISSUE #6341: Set max concurrent websocket Subscriptions for a given user

ThatOneBro:
Fixes #6341

TODO:
- [x] Cleanup the user keys on delete / closing of WebSocket connection
- [x] Add tests
~- [ ] Consider logging first to see how many users are exceeding this limit~

### DIFF
--- a/packages/server/src/subscriptions/websockets.ts
+++ b/packages/server/src/subscriptions/websockets.ts
@@ -539,6 +539,16 @@ export async function markInMemorySubscriptionsInactive(
     refStrs.push(...refs);
   }
 
+  // Remove from project-level active hashes
+  for (const [resourceType, refs] of refsByResourceType) {
+    try {
+      await removeActiveSubscriptions(projectId, resourceType, refs);
+    } catch {
+      globalLogger.debug('Attempted to mark subscriptions as inactive when Redis is closed');
+      return;
+    }
+  }
+
   // Read cache entries upfront to collect author refs for per-user set cleanup
   // (must happen before del so the data is still available)
   let authorToRefs: Map<string, string[]> | undefined;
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -932,9 +942,11 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     // Handle special cases for resource caching
     if (resource.resourceType === 'Subscription' && resource.channel?.type === 'websocket') {
       const project = resource?.meta?.project;
-      if (!project) {
-        throw new OperationOutcomeError(serverError(new Error('No project connected to the specified Subscription.')));
+      const author = resource?.meta?.author?.reference;
+      if (!(project && author)) {
```
mattwiller, line 946: Pushing down the negation also matches the error message a little better here
```suggestion
      if (!project || !author) {
```

packages/server/src/fhir/repo.ts
```diff
@@ -952,6 +964,28 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    const project = await this.getProjectById(projectId);
+    if (!project) {
+      throw new OperationOutcomeError(badRequest('No valid project connected to the specified Subscription.'));
+    }
+    const maxUserWsSubs =
+      project.systemSetting?.find((setting) => setting.name === 'maxUserWebSocketSubscriptions')?.valueInteger ??
+      // We know this is defined in defaults so this is safe to cast
+      (getConfig().defaultMaxUserWebSocketSubscriptions as number);
+    const userSubCount = await getUserActiveWebSocketSubscriptionCount(author);
+    if (userSubCount >= maxUserWsSubs) {
```
mattwiller, line 982: We might want to relax this to `>`, since the user could theoretically be modifying an existing subscription while at (but not over) the limit

ThatOneBro: Agreed. I'll fix that in a quick follow-up PR

PR #6938, opened 2025-07-01
Add Patch support to GraphQL

maddyli:


codyebberson:
https://build.fhir.org/graphql.html

image

😢

codyebberson (commented):
Code makes sense to me 👍 

It's a bummer that the spec authors didn't add this.  I think this seems like a reasonable approximation.  We'll have to accept that if patch is specified in the future, we may need to support both this and the spec version.

Or maybe we should submit this for spec review??

PR #6940, opened 2025-07-02
Avoid querying on non-existant Binary compartments column

mattlong:


codyebberson (approved):
👍 

Summary of offline conversations:  This is instead handled by `Binary.securityContext`, which is the only reliable way to secure FHIR `Binary` resources.

See: https://www.medplum.com/docs/access/binary-security-context

mattwiller (approved):

PR #6947, opened 2025-07-02
BaseChat exclude header

techdavidy:
Screenshot 2025-07-02 at 2 04 51 PM


techdavidy:
@claude feedback please

packages/react/src/chat/BaseChat/BaseChat.tsx
```diff
@@ -92,6 +93,7 @@ export function BaseChat(props: BaseChatProps): JSX.Element | null {
     onMessageUpdated,
     inputDisabled,
     onError,
+    excludeHeader = false,
```
codyebberson, line 96: Does this need default value?  If `excludeHeader` is undefined, everything would still work, yeah?

codyebberson (approved):
lgtm 👍

packages/react/src/chat/BaseChat/BaseChat.tsx
techdavidy: yeah should work  just making it explicit

PR #6948, opened 2025-07-02
Allow Super Admin to use On-Behalf-Of

mattwiller:


packages/server/src/fhir/accesspolicy.ts
```diff
@@ -32,9 +32,12 @@ export type PopulatedAccessPolicy = AccessPolicy & { resource: AccessPolicyResou
  * @returns A repository configured for the login details.
  */
 export async function getRepoForLogin(authState: AuthState, extendedMode?: boolean): Promise<Repository> {
-  const { project, login, membership, onBehalfOfMembership } = authState;
+  const { login, membership: realMembership, onBehalfOfMembership } = authState;
+  const membership = onBehalfOfMembership ?? realMembership;
+  const systemRepo = getSystemRepo();
   const accessPolicy = await getAccessPolicyForLogin(authState);
 
+  const project = await systemRepo.readReference(membership.project);
```
codyebberson, line 40: Is there a reason that we need to re-read the project here?  It's already available in `authState`.

packages/server/src/fhir/accesspolicy.ts
```diff
@@ -32,9 +32,12 @@ export type PopulatedAccessPolicy = AccessPolicy & { resource: AccessPolicyResou
  * @returns A repository configured for the login details.
  */
 export async function getRepoForLogin(authState: AuthState, extendedMode?: boolean): Promise<Repository> {
```
codyebberson, line 34: I'm trying to parse these changes....  Did anything change in this function, or is this all refactoring?

packages/server/src/oauth/utils.ts
```diff
@@ -965,7 +965,7 @@ async function tryAddOnBehalfOf(req: IncomingMessage | undefined, authState: Aut
     return;
   }
 
-  if (!authState.membership.admin) {
+  if (!authState.membership.admin && !authState.project.superAdmin) {
```
codyebberson, line 968: 👍  - I think this was the main change?

packages/server/src/fhir/accesspolicy.ts
mattwiller: Yes, because the Project in `authState` is the "real" one (i.e. the Super Admin project) in this case

packages/server/src/fhir/accesspolicy.ts
mattwiller: Other than the change to read the correct project, the rest is minor refactoring

mattlong (approved):

PR #6955, opened 2025-07-03
fix(cli/agent): `agent upgrade` - change flag to `--agentVersion`, add `--force`

### LINKED ISSUE #6953: Global `--version` clobbers CLI `medplum agent upgrade --version` flag
Related to #6815

The main immediate reason downgrade doesn't work is that the global `--version` flag which prints the version of the CLI clobbers the `--version` option of the `agent upgrade` command, so it shortcircuits and just prints the version.

A workaround is to change the `agent upgrade` flag to `--agentVersion`

ThatOneBro:
Fixes #6953 

Also enables using the `force` parameter on the `Agent/$upgrade` operation from the CLI

### DIFF
--- a/packages/cli/src/agent.ts
+++ b/packages/cli/src/agent.ts
@@ -23,6 +23,7 @@ export type CallAgentBulkOperationArgs<T extends Record<string, string>, R exten
   operation: string;
   agentIds: string[];
   options: MedplumClientOptions & { criteria: string; output?: 'json' };
+  params?: Record<string, string | boolean | number>;
   parseSuccessfulResponse: (response: AgentBulkOpResponse<R>) => T;
 };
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/agent.ts
```diff
@@ -197,24 +198,37 @@ agentUpgradeCommand
...
   )
   .option(
-    '--version <version>',
-    'An optional version to upgrade to. Defaults to the latest version if flag not included'
+    '--agentVersion <version>',
+    'An optional agent version to upgrade to. Defaults to the latest version if flag not included'
+  )
+  .option(
+    '--force',
+    'Used to force an upgrade in a scenario where a pending upgrade may have gotten into a bad state. USE AT YOUR OWN RISK!'
```
codyebberson: Dialing down the temperature 😄 

```suggestion
    'Forces an upgrade when a pending upgrade is in an inconsistent state. Use with caution.'
```

codyebberson (approved):
Let's update the help text, then lgtm 👍

PR #6956, opened 2025-07-03
fix(agent): spawn upgrader with version specified in `Agent/$upgrade` operation

### LINKED ISSUE #6954: Agent upgrader always upgrades to latest version
It looks like we haven't been passing the version into the upgrader process from the main agent process, which means that the agent will always upgrade to the latest.

This means downgrade from the agent itself is currently not possible either.

ThatOneBro:
Fixes #6954

### DIFF
--- a/packages/agent/src/app.test.ts
+++ b/packages/agent/src/app.test.ts
@@ -2101,7 +2101,7 @@ describe('App', () => {
       }
       clearTimeout(timeout);
 
-      expect(spawnSpy).toHaveBeenLastCalledWith(resolve(__dirname, 'app.ts'), ['--upgrade'], {
+      expect(spawnSpy).toHaveBeenLastCalledWith(resolve(__dirname, 'app.ts'), ['--upgrade', '4.3.1'], {
         detached: true,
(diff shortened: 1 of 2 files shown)


packages/agent/src/app.test.ts
```diff
@@ -1766,7 +1766,7 @@ describe('App', () => {
       }
       clearTimeout(timeout);
 
-      expect(spawnSpy).toHaveBeenLastCalledWith(resolve(__dirname, 'app.ts'), ['--upgrade'], {
+      expect(spawnSpy).toHaveBeenLastCalledWith(resolve(__dirname, 'app.ts'), ['--upgrade', '3.1.6'], {
```
ThatOneBro, line 1769: We had a test for this case but the test was not asserting the right thing for the 'version specified' case 😬

codyebberson (approved):

ThatOneBro:
I think I want to make sure we address #6958 somehow before we merge this, since it could result in users getting stuck in a weird limbo if they downgrade to a pre-zero-downtime version

PR #6967, opened 2025-07-03
Signature pad react component

codyebberson:


packages/react/src/SignatureInput/SignatureInput.tsx
```diff
@@ -0,0 +1,71 @@
+import { CloseButton, Paper } from '@mantine/core';
+import { createReference, ProfileResource } from '@medplum/core';
+import { Signature } from '@medplum/fhirtypes';
+import { useMedplum } from '@medplum/react-hooks';
+import { JSX, useEffect, useRef } from 'react';
+import SignaturePad from 'signature_pad';
+
+export interface SignatureInputProps {
```
techdavidy: we could `extends PaperProps `
and

```
<Paper withBorder p={0} w={width} h={height} pos="relative" {...paperProps}>
```

packages/react/src/SignatureInput/SignatureInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+  readonly width?: number;
+  readonly height?: number;
```
techdavidy, line 10: and remove w and h

packages/react/src/SignatureInput/SignatureInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+    if (signaturePadRef.current) {
+      signaturePadRef.current.clear();
+    }
+  };
+
+  const width = props.width ?? 500;
+  const height = props.height ?? 200;
+
+  return (
+    <Paper withBorder p={0} w={width} h={height} pos="relative">
```
techdavidy:
```
<Paper withBorder p={0} w={500} h={200} pos="relative" {...paperProps}>
```

packages/react/src/SignatureInput/SignatureInput.tsx
codyebberson: I like the idea.  It turns out `PaperProps` width and height are more complex than `<canvas>`'s width and height, so i kept them as numbers.  

+++ for the idea of extending `PaperProps` overall though 👍

packages/react/package.json
```diff
@@ -103,6 +103,7 @@
     "react-dom": "19.1.0",
     "react-router": "7.6.3",
     "rfc6902": "5.1.2",
+    "signature_pad": "5.0.10",
```
techdavidy, line 106: 🚀

packages/react/src/SignatureInput/SignatureInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }, [medplum, defaultValue]);
+
+  const clearSignature = (): void => {
+    if (signaturePadRef.current) {
+      signaturePadRef.current.clear();
+    }
+  };
+
+  return (
+    <Paper withBorder p={0} w={width} h={height} pos="relative" {...rest}>
```
techdavidy, line 64: you can kill width and height, if dev sets `h` or `w`  in `SignatureInput` component. It will be overridden by `...rest`

packages/react/src/SignatureInput/SignatureInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+      onChangeRef.current?.({
+        type: [
+          {
+            system: HTTP_HL7_ORG + '/fhir/signature-type',
+            code: 'ProofOfOrigin',
+            display: 'Proof of Origin',
+          },
+        ],
+        when: new Date().toISOString(),
+        who: createReference(medplum.getProfile() as ProfileResource),
```
rahul1: Recommend adding a prop for this. We have customers where a caregiver / practitioner is filling out a Questionnaire on behalf of a patient and then hands the device to them for signature. So the logged-in user may not be the same as the person actually doing the signing.

techdavidy (approved):

PR #6975, opened 2025-07-04
Add Coding synonym reference column

mattwiller:
In order to correctly expand ValueSets, a reference is needed from the synonyms to the "main" entry for a code in the `Coding` table.  This PR updates the table with the necessary column and indexes

### DIFF
--- a/packages/server/src/fhir/operations/utils/terminology.ts
+++ b/packages/server/src/fhir/operations/utils/terminology.ts
@@ -87,7 +87,7 @@ export function selectCoding(systemId: string, ...code: string[]): SelectQuery {
     .column('display')
     .where('system', '=', systemId)
     .where('code', 'IN', code)
-    .where('isSynonym', '=', false);
+    .where('synonymOf', '=', null);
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/operations/utils/terminology.ts
```diff
@@ -87,7 +87,7 @@ export function selectCoding(systemId: string, ...code: string[]): SelectQuery {
     .column('display')
     .where('system', '=', systemId)
     .where('code', 'IN', code)
-    .where('isSynonym', '=', false);
```
codyebberson, line 90: Does this mean we will eventually drop the `isSynonym` column?

codyebberson (approved):

packages/server/src/fhir/operations/utils/terminology.ts
mattwiller: Yeah, that would be the end goal: to replace `isSynonym` with `synonymOf`

PR #6981, opened 2025-07-07
Dependency upgrades 2025-07-07

medplumbot:
Dependency upgrades


codyebberson:
Fixed the `react-native` override.

It had been a while since we audited the overrides section.... Cleaned out a bunch that are not referenced in `package-lock.json` at all:

image

codyebberson (approved):

PR #7013, opened 2025-07-10
feat(agent): add stats logging

ThatOneBro:


packages/agent/src/hl7.ts
```diff
@@ -56,9 +57,11 @@ export class AgentHl7Channel extends BaseChannel {
 
   private handleNewConnection(connection: Hl7Connection): void {
     const c = new AgentHl7ChannelConnection(this, connection);
+    updateStat('hl7ConnectionsOpen', getCurrentStats().hl7ConnectionsOpen++);
     c.hl7Connection.addEventListener('close', () => {
       this.log.info(`Closing connection: ${c.remote}`);
       this.connections.delete(c.remote);
+      updateStat('hl7ConnectionsOpen', getCurrentStats().hl7ConnectionsOpen--);
```
sachin-sunny-rad: I lack context on this code but should this be something like `updateStat('hl7ConnectionsOpen', this.connections.size);`? Just to avoid these stats from drifting incase something else mutates connections?

ThatOneBro: Hi @sachin-sunny-rad , good question. I actually had to double check myself.

In this case we are keeping a global count of connections which is not tracked in `this.connections`, as that is only tracking the connections for the one channel in this context.

It is possible that `updateStat` gets called somewhere else but in this case we know that the only places it gets called currently are within the context of creating a new HL7 connection and when this particular connection gets destroyed, so it should be the case that as long as there are no connections tracked in a different context there should always be parity with the true number of total outstanding active HL7 connections

sachin-sunny-rad: ahh that makes sense, thank you!

codyebberson:
@claude please review

PR #7015, opened 2025-07-10
Change Project Admin AccessPolicy so admins can change project-scoped user emails

maddyli:


codyebberson (approved):

packages/server/src/fhir/accesspolicy.ts
```diff
@@ -258,7 +258,7 @@ function applyProjectAdminAccessPolicy(
     accessPolicy.resource.push({
       resourceType: 'User',
       hiddenFields: ['passwordHash', 'mfaSecret'],
-      readonlyFields: ['email', 'emailVerified', 'mfaEnrolled', 'project'],
+      readonlyFields: ['emailVerified', 'mfaEnrolled', 'project'],
```
mattwiller, line 261: This kind of breaks the guarantee of `emailVerified`, no?  If someone can manually change the email without the system sending a verification link to that address, then `emailVerified` could accidentally be incorrect

PR #7017, opened 2025-07-10
feat(agent): add hl7 connection throttling for channels

### LINKED ISSUE #7016: [Agent] Add throttle for HL7 channels
After enabling enhanced ACK mode in #6761, it's possible that sending devices will DOS Medplum server by dumping large amounts of messages to be processed as fast as they can when they are replaying message history at a Medplum agent channel. 

Because of that, we might want to throttle HL7 connections by emitting messages to the processor at a particular predefined message rate, expressed as messages per second. This should be interpreted as a maximum message rate, which translates to making sure it's been at least `1000 / messagesPerSec` milliseconds before emitting the next message.

ThatOneBro:
Fixes #7016


codyebberson (approved):

packages/hl7/src/connection.ts
```diff
@@ -81,15 +98,6 @@ export class Hl7Connection extends Hl7Base {
       if (this.enhancedMode) {
         this.send(event.message.buildAck({ ackCode: 'CA' }));
       }
-      // Check if we have messages pending a response in the pending messages map
-      if (!this.pendingMessages.size) {
```
ThatOneBro, line 85: Removed this because this actually causes issues when receiving messages on a server-side connection since server-side connections don't have pending messages

PR #7031, opened 2025-07-12
fix the MCP server url in ReadMe for streamable http

jalanubha:
Was testing the MCP server capabilities in Medplum and noticed the url in the documentation for Streamable HTTP should be `/mcp/stream` instead of `/mcp`.

Verified this using MCP Inspector.


codyebberson:
Thanks @jalanubha !

PR #7047, opened 2025-07-14
Fixes #7014 - remove signature header requirement

### LINKED ISSUE #7014: Consider removing the requirement for certain request headers when calling webhooks
Before #6912 all bots could be called as a webhook as long as the request carried one of the vetted allowed headers (https://github.com/medplum/medplum/blob/de44a5d93599c3198f0f228e8effa4cd2790d627/packages/server/src/webhook/routes.ts#L24). Given that we now need to explicitly tag the bots that are allowed as webhook I believe it's safe to relax the header verification logic. This would enable integrations with other systems without the need to change the Medplum code, and in practice the header verification logic is not adding much as it doesn't check the signature it just requires the existence of the headers regardless of the value in it.

codyebberson:
As discussed in #7014 now that we have a formal opt-in for `Bot.publicWebhook`, we can remove this arbitrary signature requirement.

### DIFF
--- a/packages/server/src/webhook/routes.test.ts
+++ b/packages/server/src/webhook/routes.test.ts
@@ -156,15 +156,6 @@ describe('Anonymous webhooks', () => {
     await shutdownApp();
   });
 
-  test('Missing signature header', async () => {
-    const res = await request(app)
-      .post(`/webhook/${botMembership.id}`)
-      .set('Content-Type', ContentType.TEXT)
-      .send('input');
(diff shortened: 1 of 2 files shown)


ThatOneBro:
We may want to add documentation around creating webhook bots with signature verification heavily emphasized for security, and also probably add some kind of messaging in the UI when opting into public webhook through the UI (when the box is checked, display a warning to make sure to include signature verification, with a link to the docs), just because it seems like an easy footgun.

PR #7052, opened 2025-07-14
Transaction retry logging

mattlong:


packages/server/src/fhir/repo.ts
```diff
@@ -2455,7 +2459,20 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+        getLogger().info('Retrying transaction after delay', {
+          attempt,
+          attemptDurationMs,
+          transactionAttempts,
+          delayMs,
+          baseDelayMs,
+        });
         await sleep(delayMs);
+      } else {
+        getLogger().info('Transaction failed after final <attempt></attempt>', {
```
mattwiller: What's the HTML-esque tag doing?

packages/server/src/fhir/sql.ts
```diff
@@ -552,6 +552,13 @@ export class SqlBuilder {
   }
 }
 
+export const PostgresError = {
```
mattwiller, line 555: 👍

packages/server/src/fhir/repo.ts
mattlong: That was an errant VSCode extension that wanted to auto-expand "attempt" to a tag for some reason. 🤷

ThatOneBro (approved):

PR #7081, opened 2025-07-17
Added CDK support for AWS FireLens

codyebberson:


packages/cdk/src/index.ts
```diff
@@ -23,20 +23,12 @@ export function main(context?: Record<string, string>): void {
...
-    });
+  const normalizedConfig = await normalizeInfraConfig(config);
+  const stack = new MedplumStack(app, normalizedConfig);
+  console.log('Stack', stack.primaryStack.stackId);
+  app.synth();
 }
 
 if (require.main === module) {
-  main();
+  main().catch(console.error);
```
mattlong, line 33: Not sure if it really matters here, but we should probably keep the `process.exit(1)` in `catch`

PR #7084, opened 2025-07-17
fix(scripts/deploy-app): add empty defaults for google and recaptcha

ThatOneBro:


scripts/deploy-app.sh
```diff
@@ -16,17 +16,9 @@ if [[ -z "${MEDPLUM_BASE_URL}" ]]; then
...
-  exit 1
-fi
-
-if [[ -z "${RECAPTCHA_SITE_KEY}" ]]; then
-  echo "RECAPTCHA_SITE_KEY is missing"
-  exit 1
-fi
-
 # Setting defaults for the replacements of placeholders which should have defaults
+: ${GOOGLE_CLIENT_ID:=""}
```
mattlong, line 20: I understand where this is coming from; neither of these are required for app to function correctly, but do we benefit from having a sanity check that they are indeed populated before errantly deploying app.medplum.com? In other words, is this script intended to be used by self-hosters or just for our hosted offering?

codyebberson: This is a good question.  Self-hosters are encouraged to use Medplum CLI (i.e., `medplum aws deploy-app`).  But we know some self-hosters found this script and use it themselves 🤷

codyebberson (approved):

PR #7090, opened 2025-07-18
Write synonyms to Coding table from code properties

mattwiller:


mattwiller:
Related to #5966

packages/server/src/migrations/migrate.ts
```diff
@@ -1196,6 +1213,7 @@ function generateIndexesActions(
 
   for (const startIndex of startTable.indexes) {
     if (!matchedIndexes.has(startIndex)) {
+      console.log('EXTRA:', startIndex);
```
codyebberson: Cleanup

packages/server/src/migrations/migrate.ts
```diff
@@ -1182,6 +1198,7 @@ function generateIndexesActions(
 
     const startIndex = startTable.indexes.find((i) => indexDefinitionsEqual(i, targetIndex));
     if (!startIndex) {
+      console.log('MISSING:', targetIndex);
```
codyebberson: Cleanup

packages/server/src/migrations/migrate.ts
```diff
@@ -1257,6 +1275,10 @@ function buildIndexSql(tableName: string, indexName: string, index: IndexDefinit
     result += ')';
   }
 
+  if (index.nulls) {
+    result += ' NULLS ' + index.nulls;
```
codyebberson: This is the main blocker on Postgres 13, yeah?

codyebberson:
@claude please review

packages/server/src/migrations/migrate.ts
mattwiller: Correct: this is the way we would like to construct our indexes/queries, but it was only added in Postgres 15

packages/server/src/migrations/data/data-version-manifest.json
```diff
@@ -70,5 +70,8 @@
   },
   "v20": {
     "serverVersion": "4.3.6"
+  },
+  "v21": {
+    "serverVersion": "4.3.7"
```
mattlong: Needs to be updated to `4.3.9`

packages/server/src/migrations/migrate.ts
mattlong: Should this if block throw an error until we support it so that no one tries to use it? Granted, it'd be caught at test time, but it would be nice to give faster feedback than that.

mattwiller: I went ahead and removed all the `NULLS (NOT) DISTINCT` stuff for now, since it'll require dropping support for lower Postgres versions

PR #7097, opened 2025-07-21
fix(agent): precisely diff channel config when reloading

### LINKED ISSUE #7083: [Agent] Changing the endpoint query params does not trigger a reload for a channel
Currently only address or port changes for an endpoint trigger a "full" channel reload, we need it to also reload on certain query parameter changes, like `enhanced=true`(enhanced mode ACKs)

ThatOneBro:
Fixes #7083

### DIFF
--- a/packages/agent/src/channel.ts
+++ b/packages/agent/src/channel.ts
@@ -14,8 +14,8 @@ export interface Channel {
 
 export abstract class BaseChannel implements Channel {
   readonly app: App;
-  private definition: AgentChannel;
-  private endpoint: Endpoint;
+  protected definition: AgentChannel;
+  protected endpoint: Endpoint;
 
   constructor(app: App, definition: AgentChannel, endpoint: Endpoint) {
     this.app = app;
@@ -27,22 +27,7 @@ export abstract class BaseChannel implements Channel {
   abstract start(): void;
   abstract stop(): Promise<void>;
   abstract sendToRemote(message: AgentTransmitResponse): void;
-
-  async reloadConfig(definition: AgentChannel, endpoint: Endpoint): Promise<void> {
-    const previousEndpoint = this.endpoint;
-    this.definition = definition;
-    this.endpoint = endpoint;
-
-    this.log.info('Reloading config... Evaluating if channel needs to change address...');
-
-    if (needToRebindToPort(previousEndpoint, endpoint)) {
-      await this.stop();
-      this.start();
-      this.log.info(`Address changed: ${previousEndpoint.address} => ${endpoint.address}`);
-    } else {
-      this.log.info(`No address change needed. Listening at ${endpoint.address}`);
-    }
-  }
+  abstract reloadConfig(definition: AgentChannel, endpoint: Endpoint): Promise<void>;
 
   getDefinition(): AgentChannel {
     return this.definition;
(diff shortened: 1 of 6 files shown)


packages/agent/src/channel.ts
```diff
@@ -14,8 +14,8 @@ export interface Channel {
 
 export abstract class BaseChannel implements Channel {
   readonly app: App;
-  private definition: AgentChannel;
-  private endpoint: Endpoint;
+  protected definition: AgentChannel;
```
ThatOneBro, line 17: Changed to protected so that subclass can access / change these fields

packages/agent/src/channel.ts
```diff
@@ -27,22 +27,7 @@ export abstract class BaseChannel implements Channel {
...
-
-    if (needToRebindToPort(previousEndpoint, endpoint)) {
-      await this.stop();
-      this.start();
-      this.log.info(`Address changed: ${previousEndpoint.address} => ${endpoint.address}`);
-    } else {
-      this.log.info(`No address change needed. Listening at ${endpoint.address}`);
-    }
-  }
+  abstract reloadConfig(definition: AgentChannel, endpoint: Endpoint): Promise<void>;
```
ThatOneBro, line 30: Made this method abstract; there are slight divergences / implementation-specific details between HL7 and DICOM channels already. 

The superclass method seemed to be a pre-mature factoring out of this method. Pushing the implementation back down to the subclass.

packages/agent/src/hl7.ts
```diff
@@ -27,11 +27,11 @@ export class AgentHl7Channel extends BaseChannel {
...
     }
     this.started = true;
-    const address = new URL(this.getEndpoint().address as string);
-    const encoding = address.searchParams.get('encoding') ?? undefined;
-    const enhancedMode = address.searchParams.get('enhanced')?.toLowerCase() === 'true';
+
+    const address = new URL(this.getEndpoint().address);
     this.log.info(`Channel starting on ${address}...`);
-    this.server.start(Number.parseInt(address.port, 10), encoding, enhancedMode);
+    this.configureHl7ServerAndConnections();
```
ThatOneBro, line 33: Centralizing all configuration into one method that we can call both on initialization and reloading of config

packages/agent/src/hl7.ts
```diff
@@ -55,6 +55,49 @@ export class AgentHl7Channel extends BaseChannel {
...
   }
 
+  async reloadConfig(definition: AgentChannel, endpoint: Endpoint): Promise<void> {
+    const previousEndpoint = this.endpoint;
+    this.definition = definition;
+    this.endpoint = endpoint;
+
+    this.log.info('Reloading config... Evaluating if channel needs to change address...');
+
+    if (this.needToRebindToPort(previousEndpoint, endpoint)) {
```
ThatOneBro, line 65: Only do a full restart of a channel if we need to rebind to a different interface / different port...
Otherwise we can maintain the original connections and just tweak their configurations

packages/hl7/src/connection.ts
```diff
@@ -14,14 +14,16 @@ export type Hl7MessageQueueItem = {
   reject?: (err: Error) => void;
 };
 
+export const DEFAULT_ENCODING = 'utf-8';
+
 export class Hl7Connection extends Hl7Base {
   readonly socket: net.Socket;
-  readonly encoding: string;
-  readonly enhancedMode: boolean;
+  encoding: string;
```
ThatOneBro, line 21: Removed the `readonly` here so that we can set them later via setters, didn't make these fields private though to preserve backwards compat on the public interface

PR #7098, opened 2025-07-21
Adding timestamp to deployment diagram

reshmakh:


packages/docs/docusaurus.config.ts
```diff
@@ -18,9 +18,11 @@ const config: Config = {
 
   // Set this to true to enable the faster experimental build mode.
   // https://github.com/facebook/docusaurus/issues/10556
-  future: {
-    v4: true,
-    experimental_faster: true,
+  customFields: {
```
ThatOneBro: Hmmm I think this shouldn't be necessary, this seems to undo our opting into `@docusaurus/faster`. 

Might have to run `npm ci` on latest in order to get these fields to validate properly? This was a pretty recent update to our dependencies

ThatOneBro (approved):

PR #7107, opened 2025-07-22
Update versions of Postgres in CI config

mattwiller:


codyebberson:
https://github.com/medplum/medplum/blob/main/.github/workflows/build.yml#L239-L244

PR #7113, opened 2025-07-23
Update multi-tenant-access-policy.md

finnbergquist:
%current_organization should already be a full ref here

We found that Claude was giving errant example AccessPolicies based on this example


maddyli (approved):
lgtm

PR #7117, opened 2025-07-23
[Docs] Update Project Settings docs around enableFhirQuota

maddyli:


codyebberson (approved):
Nice catch.

We should keep an eye on this.  I'm not sure we should have actually enabled it by default, as this will be a surprise for users, and might create some weird circumstances for first time users.

For example, one of the first things people do with a new install is blast it with new resources, and try performance benchmarking.  This effectively handicaps that new user experience.

We definitely want this enabled on prod, and it is strongly recommended for self-hosters.

rahul1 (approved):

PR #7127, opened 2025-07-25
Add securityContext to Binary resources uploaded from the Edit page 

maddyli:
Files uploaded through the Edit page of app.medplum do not include a securityContext. The uploader is the green button in the screenshot below. This PR aims to mimic the behavior of the upload button on the Timeline page of app.medplum, which does include a securityContext

Screenshot 2025-07-25 at 9 46 53 AM

### DIFF
--- a/packages/react/src/ResourceForm/ResourceForm.tsx
+++ b/packages/react/src/ResourceForm/ResourceForm.tsx
@@ -3,6 +3,7 @@ import {
   AccessPolicyInteraction,
   applyDefaultValuesToResource,
   canWriteResourceType,
+  createReference,
   isPopulated,
   satisfiedAccessPolicy,
   tryGetProfile,
@@ -13,6 +14,7 @@ import { IconAlertCircle, IconChevronDown, IconEdit, IconTrash } from '@tabler/i
 import cx from 'clsx';
 import { FormEvent, JSX, useEffect, useMemo, useState } from 'react';
 import { BackboneElementInput } from '../BackboneElementInput/BackboneElementInput';
+import { ElementsContext } from '../ElementsInput/ElementsInput.utils';
 import { FormSection } from '../FormSection/FormSection';
 import classes from './ResourceForm.module.css';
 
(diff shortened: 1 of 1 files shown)


codyebberson:
We should fix the Sonar issue regarding React re-renders: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=7127&id=medplum_medplum&open=AZhDmRBYEH-OrWYO9JMc

Feel free to ping [#eng](https://medplum.slack.com/archives/C051SKJKZ50) if you'd like any pointers.

packages/react/src/ResourcePropertyInput/ResourcePropertyInput.utils.ts
```diff
@@ -7,13 +7,17 @@ export interface BaseInputProps {
   readonly valuePath?: string;
   /** (optional) OperationOutcome from the last attempted system action*/
   readonly outcome?: OperationOutcome;
+  /** (optional) Security context reference for file uploads */
+  readonly securityContext?: any;
```
techdavidy, line 11: isn't this a `Reference` ?

packages/react/src/ResourceForm/ResourceForm.tsx
```diff
@@ -118,16 +120,30 @@ export function ResourceForm(props: ResourceFormProps): JSX.Element {
...
-        path={value.resourceType}
-        valuePath={value.resourceType}
-        typeName={resourceType}
-        defaultValue={value}
-        outcome={outcome}
-        onChange={setValue}
-        profileUrl={props.profileUrl}
-        accessPolicyResource={accessPolicyResource}
-      />
+      <ElementsContext.Provider
```
techdavidy, line 123: do we need `ElementsContext.Provider` here ?

maddyli: Hm I believe this is passed as an alternative way to get the securityContext resource in the next component, though I am definitely not a UI expert

packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
```diff
@@ -56,6 +56,10 @@ export function ResourcePropertyInput(props: ResourcePropertyInputProps): JSX.El
       : property.type[0].code;
   const propertyTypes = property.type as ElementDefinitionType[];
 
+  // Get securityContext from props or from context
+  const elementsContext = useContext(ElementsContext);
+  const securityContext = props.securityContext ?? (elementsContext as any).securityContext;
```
techdavidy, line 61: it seems we don't need it to add an ElementContext since props.securityContex would have the same value.

packages/react/src/ResourceForm/ResourceForm.tsx
codyebberson: Yeah, the `<ElementsContext.Provider>` should already be added automatically higher up in the DOM tree.

For example, see `packages/react/src/BackboneElementInput/BackboneElementInput.tsx` and the call to `maybeWrapWithContext()`.

techdavidy:
@claude please check this PR

techdavidy:
@claude ElementsContext.Provider and BackboneElementInput get the exact same `accessPolicyResource={accessPolicyResource}` so it would need the fallback as props and elementsContext would provide the same value.
  ```
  const securityContext = props.securityContext ?? (elementsContext as any).securityContext;
  ```

techdavidy:
@claude For sure I mean keep elementsContext props but remove <ElementsContext.Provider> as it is created in ElementsInput.utils.ts

techdavidy:
🤦🏻‍♂️ It wen't ahead and did it.

techdavidy:
@claude revert the changes please. it was a comment about the ElementsContext.Provider, not a request.

PR #7129, opened 2025-07-25
fix(hl7): use msg ctrl id to correlate resp for `Hl7Client.sendAndWait`

### LINKED ISSUE #7115: `Hl7Client.sendAndWait` does not handle enhanced-mode ACKs properly
Currently the `Hl7Client` naively returns the next message it receives from the server for the `sendAndWait` method, which with the enhanced mode ACK protocol, does not function as intended.

Whenever we return a CA (Commit Accept) ACK in enhanced mode, we expect an AA (Application Accept) ACK to be generated later, which the client does not currently take into account.

We should check that the  control ID matches the original message before returning an ACK for `sendAndWait`

ThatOneBro:
Fixes #7115

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -872,7 +872,10 @@ export class App {
           this.log.info(`Persistent connection to remote '${message.remote}' closed`);
         });
         client.addEventListener('error', (event) => {
-          this.hl7Clients.delete(message.remote);
+          // If the current client for this remote is this client, make sure to clean it up
+          if (this.hl7Clients.get(message.remote) === client) {
+            this.hl7Clients.delete(message.remote);
+          }
           this.log.error(
             `Persistent connection to remote '${message.remote}' encountered error: '${normalizeErrorString(event.error)}' - Closing connection...`
           );


codyebberson:
@claude please review

codyebberson (approved):
Structurally looks good to me.  Good to go after test coverage clears.

packages/hl7/src/connection.ts
```diff
@@ -59,24 +76,59 @@ export class Hl7Connection extends Hl7Base {
...
+        return;
+      }
+
+      // Two modes:
+      // Application-level or first ACK
+
+      // First should always return on any ACK message, this is the default
+      // The exception is APPLICATION, which should not resolve when the ACK is a CA, but should resolve on all other ACK types
+      // On CA, we return early
+      if (queueItem.returnAck === ReturnAckCategory.APPLICATION && ackCode === 'CA') {
```
ThatOneBro, line 131: The only case we really want to skip resolving the Promise when receiving in ACK is when:
1. The user opts into the APPLICATION-level ACK
2. The ACK code returned is a CA (all other C-level ACKs indicate early failure and probably that you won't get an applciation-level ACK)

codyebberson (approved):

PR #7131, opened 2025-07-25
Search IR

codyebberson:
This unifies portions of the 2 search implementations (in-memory and database).

Historically, we have considered the database implementation to be the "main" one, and the in-memory implementation to be an "extra" for supplementary use cases.  However, over time, we have used the in-memory implementation more and more.  For example, it is now used when evaluating subscription criteria or access policy criteria.  Under those circumstances, it is now more important than ever that these search implementations be aligned for maximum correctness.

### DIFF
--- a/packages/core/src/format.ts
+++ b/packages/core/src/format.ts
@@ -35,7 +35,7 @@ export interface HumanNameFormatOptions {
  * @returns The string representation of the typed value.
  */
 export function typedValueToString(typedValue: TypedValue | undefined): string {
-  if (!typedValue?.value) {
+  if (!typedValue) {
     return '';
   }
   switch (typedValue.type) {
@@ -54,7 +54,7 @@ export function typedValueToString(typedValue: TypedValue | undefined): string {
     case 'Reference':
       return formatReferenceString(typedValue.value);
     default:
-      return typedValue.value.toString();
+      return typedValue.value?.toString() ?? '';
   }
 }
 
--- a/packages/core/src/format.ts
+++ b/packages/core/src/format.ts
@@ -35,7 +35,7 @@ export interface HumanNameFormatOptions {
  * @returns The string representation of the typed value.
  */
 export function typedValueToString(typedValue: TypedValue | undefined): string {
-  if (!typedValue) {
+  if (!typedValue?.value) {
     return '';
   }
   switch (typedValue.type) {
--- a/packages/core/src/search/ir.ts
+++ b/packages/core/src/search/ir.ts
@@ -73,6 +73,9 @@ export function convertToReferenceSearchIR(typedValues: TypedValue[]): Reference
   const result: ReferenceSearchIR[] = [];
   for (const typedValue of typedValues) {
     const { value } = typedValue;
+    if (!value) {
+      continue;
+    }
     if (isString(value)) {
       // Handle "canonical" properties such as QuestionnaireResponse.questionnaire
       // This is a reference string that is not a FHIR reference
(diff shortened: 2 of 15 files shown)


codyebberson:
@claude please summarize and review

codyebberson:
@claude - minor issues are resolved, can you please review again?

codyebberson:
@claude - Please review again.  A couple notes:

1. `toLocaleLowerCase()` is a method, and is preferred over `toLowerCase()`.  See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase
2. The test coverage report only includes `core` tests.  The `TokensContext.caseInsensitive` is tested extensively in the `server` package.

packages/core/src/format.ts
```diff
@@ -35,7 +35,7 @@ export interface HumanNameFormatOptions {
  * @returns The string representation of the typed value.
  */
 export function typedValueToString(typedValue: TypedValue | undefined): string {
-  if (!typedValue) {
+  if (!typedValue?.value) {
```
mattwiller: This wouldn't be the correct output for Boolean `false`, right?

packages/core/src/search/ir.ts
```diff
@@ -0,0 +1,244 @@
...
+import { typedValueToString } from '../format';
+import { isReference, PropertyType, TypedValue } from '../types';
+import { getReferenceString, isResourceWithId, isString } from '../utils';
+
+export interface SearchableToken {
+  readonly system: string | undefined;
+  readonly value: string | undefined;
+}
+
+export function convertToSearchableNumbers(typedValues: TypedValue[]): number[] {
```
mattwiller: Numbers also have implicit precision similar to dates; since this is going to be a public interface, should we account for that now with e.g. `number[][]`?

packages/server/src/fhir/repo.ts
```diff
@@ -1668,90 +1671,68 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+      // 2. The underlying FHIR ElementDefinition referred to by the search parameter has a type of "date".
+      return convertToSearchableDates(typedValues)
+        .map((p) => (p.start ?? p.end)?.substring(0, 10))
+        .filter(Boolean) as string[];
     }
 
     if (details.type === SearchParameterType.DATETIME) {
-      return this.buildDateTimeColumn(value);
+      return convertToSearchableDates(typedValues)
+        .map((p) => p.start ?? p.end)
```
mattwiller: Would you mind leaving a comment here, to the effect that we should eventually write the whole period to the DB once we figure the `tstzrange` stuff out?

packages/server/src/fhir/repo.ts
```diff
@@ -1668,90 +1671,68 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+
+    if (searchParam.type === 'number') {
+      return convertToSearchableNumbers(typedValues).filter((n) => n !== undefined);
     }
 
     if (searchParam.type === 'quantity') {
-      return this.buildQuantityColumn(value);
+      return convertToSearchableQuantities(typedValues)
+        .map((q) => q.value)
+        .filter((q) => q !== undefined);
```
mattwiller: Probably not a big deal, but these could use our `flatMapFilter()` helper to perform a single loop over the initial array and create only a single output array

packages/core/src/format.ts
codyebberson: Excellent catch

mattwiller (approved):

PR #7136, opened 2025-07-25
DoseSpot Clinic Setting of Favorite Medications

finnbergquist:
- React components and hooks to use the `DoseSpot Add Favorite Medication Bot` and `DoseSpot Search Medication Bot`
- Provider app implementation

https://github.com/user-attachments/assets/0cffe044-992d-4394-b771-cce5b9a91ef7


packages/dosespot-react/src/useDoseSpotFormularyCenter.ts
```diff
@@ -0,0 +1,57 @@
+import { Bundle, MedicationKnowledge } from '@medplum/fhirtypes';
+import { useMedplum } from '@medplum/react-hooks';
+import { useCallback, useState } from 'react';
+import { DOSESPOT_ADD_FAVORITE_MEDICATION_BOT, DOSESPOT_SEARCH_MEDICATIONS_BOT } from './common';
+
+export interface DoseSpotClinicFormularyReturn {
```
rahul1, line 8: some comments on each element of this return bundle would be valuable

packages/dosespot-react/src/useDoseSpotFormularyCenter.ts
```diff
@@ -0,0 +1,57 @@
...
+  readonly searchResults: Bundle<MedicationKnowledge> | undefined;
+  readonly searchLoading: boolean;
+  readonly searchMedications: (searchTerm: string) => Promise<Bundle<MedicationKnowledge> | undefined>;
+  readonly addFavoriteMedication: (medication: MedicationKnowledge) => Promise<MedicationKnowledge>;
+  readonly addFavoriteMedicationLoading: boolean;
```
rahul1: I'm not a react hooks expert, but my preference would be that we expose as stateless an API from an API perspective, we should keep the hook as stateless as possible. Especially for these loading states, these can be something that clients of the hook can manage for themselves, as long as `addFavoriteMedications` or `searchMedications` return promises. 

Persisting the state inside the hook gives clients less control on how rendering gets triggered

packages/dosespot-react/src/useDoseSpotFormularyCenter.ts
```diff
@@ -0,0 +1,57 @@
+import { Bundle, MedicationKnowledge } from '@medplum/fhirtypes';
+import { useMedplum } from '@medplum/react-hooks';
+import { useCallback, useState } from 'react';
+import { DOSESPOT_ADD_FAVORITE_MEDICATION_BOT, DOSESPOT_SEARCH_MEDICATIONS_BOT } from './common';
+
+export interface DoseSpotClinicFormularyReturn {
+  readonly searchResults: Bundle<MedicationKnowledge> | undefined;
+  readonly searchLoading: boolean;
+  readonly searchMedications: (searchTerm: string) => Promise<Bundle<MedicationKnowledge> | undefined>;
```
rahul1: Unless there's a good reason I need the `Bundle`

```suggestion
  readonly searchMedications: (searchTerm: string) => Promise<MedicationKnowledge[] >;
```

packages/dosespot-react/src/DoseSpotMedicationSelect.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+      const results = await (searchMedications as (term: string) => Promise<Bundle<MedicationKnowledge> | undefined>)(
+        input.trim()
+      );
+
+      if (!results) {
+        return [];
+      }
+
+      return results.entry?.map((entry) => entry.resource as MedicationKnowledge).filter(Boolean) || [];
```
rahul1: unless there's a strong reason not to, I'd suggest rolling all of this unbundling logic into the implementation of the hook (or the bot). Pretty much all clients are going to have to do this. 

That should basically eliminate the need for this wrapper component entirely, allowing you to plug directly into AsyncAutocomplete, as we do here.

https://github.com/medplum/medplum/blob/053b7d09a39edd9026cefac536a40d4462a900b6/examples/medplum-health-gorilla-demo/src/pages/HomePage.tsx#L105-L114

packages/dosespot-react/src/DoseSpotNewMedicationForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+      amount: {
+        value: formData.quantity,
+        unit: 'units',
+      },
+      extension: [
+        {
+          url: DOSESPOT_REFILLS_SYSTEM,
+          valueString: formData.refills.toString(),
+        },
+      ],
```
rahul1: I don't want to introduce an extensions for now, as we talked about updating this to use `ActivityDefinition` in the future. 

once we introduce this, it becomes part of our public API

packages/dosespot-react/src/DoseSpotNewMedicationForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+          ],
+        },
+      ],
+    };
+
+    await addFavoriteMedication(completedMedication);
+    setMedication(completedMedication);
+  };
+
+  const getCodingValue = (medication: MedicationKnowledge, system: string): string | undefined => {
```
rahul1: We have a utility function for this: [`getCodeBySystem`](https://www.medplum.com/docs/sdk/core.getcodebysystem)

packages/dosespot-react/src/DoseSpotNewMedicationForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+    return medication.code?.coding?.find((coding) => coding.system === system)?.code;
+  };
+
+  const ndcCode = medication ? getCodingValue(medication, NDC) : undefined;
```
rahul1:
```suggestion
  const ndcCode = medication?code && getCodeBySystem(medication.code, NDC)
```

packages/dosespot-react/src/DoseSpotNewMedicationForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+export function DoseSpotNewMedicationForm(props: DoseSpotNewMedicationFormProps): React.JSX.Element {
+  const { searchMedications, addFavoriteMedication, loading = false } = props;
+  const [formData, setFormData] = useState<FormData>({
+    quantity: 1,
+    refills: 0,
+    directions: '',
+  });
+  const [medication, setMedication] = useState<MedicationKnowledge | null>(null);
+
+  const handleAddFavorite = async (): Promise<void> => {
```
rahul1: @techdavidy could you comment on this - something feels off to me for a React callback to both depend on `medication` and to call `setMedication`, but I'm not a React expert. 

At the very least, seems like `useCallback` should be used uer

packages/dosespot-react/src/DoseSpotNewMedicationForm.tsx
rahul1, line 1: @finnbergquist  with our Health Gorilla react package, we didn't actually ship an order form as part of the react package, just the hooks. The order form was left as an example exercise. 

I think we could justify exporting this as a component, as users often want a smart default, rather than copying example code. However, we should work on prop design here. I would imagine the main advantage of this component would be actually to invoke the `useDosepotFormulary` hook on your behalf, so you wouldn't need to pass in the `search**` and `addFavorite` props. 

Instead, we might want to accept hooks on more lifecycle events, rather than making this component stateful. Eg. accept an `onMedicationSaved` callback for when the medication is saved. 

let's talk more offline about the use case for these components. I think you get away with this PR just including the hooks, and pushing the componetns in teo the Provider App

packages/dosespot-react/src/useDoseSpotFormularyCenter.ts
finnbergquist: Yea, good point

finnbergquist:
Closing this PR. Following up this work with all comments addressed in https://github.com/medplum/medplum/pull/7223

PR #7141, opened 2025-07-27
Support environments without crypto.subtle

codyebberson:


packages/core/src/client.ts
```diff
@@ -3712,11 +3712,17 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
-    return { codeChallengeMethod: 'S256', codeChallenge };
+    try {
+      const arrayHash = await encryptSHA256(codeVerifier);
+      const codeChallenge = arrayBufferToBase64(arrayHash)
+        .replaceAll('+', '-')
+        .replaceAll('/', '_')
+        .replaceAll('=', '');
+      return { codeChallengeMethod: 'S256', codeChallenge };
+    } catch (err) {
+      console.warn('Failed to generate code challenge', err);
```
ThatOneBro: A more accurate message might be "Failed to hash code verifier. Falling back to 'plain' code challenge method" or something similar. But minor nit

ThatOneBro (approved):

packages/core/src/client.test.ts
```diff
@@ -533,6 +533,40 @@ describe('Client', () => {
...
+        clientId: 'medplum-client-123',
+      },
+      false
+    );
+    expect(result).toBeUndefined();
+    expect(assign).not.toHaveBeenCalledWith(expect.stringContaining('code_challenge'));
+    expect(assign).not.toHaveBeenCalledWith(expect.stringContaining('code_challenge_method'));
+
+    Object.defineProperty(crypto, 'subtle', {
+      value: originalSubtle,
```
ThatOneBro, line 566: We might want to restore the original `location.assign` as well? 🤔

PR #7143, opened 2025-07-28
Dependency upgrades 2025-07-28

medplumbot:
Dependency upgrades


packages/server/package.json
```diff
@@ -87,7 +87,7 @@
     "pdfmake": "0.2.20",
     "pg": "8.16.3",
     "qrcode": "1.5.4",
-    "rate-limiter-flexible": "7.1.1",
+    "rate-limiter-flexible": "7.2.0",
```
ThatOneBro, line 90: This is the only "big change" I can find, and looks like they are following strict semver so, only thing added in the minor bump was a new rate limiter strategy/ type: Drizzle

mattwiller: I checked the `bullmq` changes as well, but they seem like minor bug fixes unlikely to impact us

PR #7145, opened 2025-07-28
Publish helm charts in github actions

codyebberson:


.github/workflows/build-helm-charts.yml
```diff
@@ -0,0 +1,60 @@
...
+
+concurrency:
+  group: ci-build-helm-${{ github.sha }}
+  cancel-in-progress: true
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
```
ThatOneBro, line 11: Makes sense 👍

PR #7175, opened 2025-08-01
fix(hl7): async `close`, cleanup close logic, resolve deferred connect promise

ThatOneBro:
This fixes a number of bugs that we had in the client logic around connecting, emitting close events from `Hl7Connection`, calling close while connecting, etc.

The main objective of the original PR was to fix a few broken tests that were making it difficult to test other flaky `Hl7Client` tests,  but in the process I discovered quite a few bugs in the lifecycle of the `Hl7Client` and `Hl7Connection`, especially around all the corner cases of ordering of successive calls to `connect` and then accounting for the other side closing the connection.


packages/hl7/src/client.ts
```diff
@@ -114,18 +135,24 @@ export class Hl7Client extends Hl7Base {
     return (await this.connect()).sendAndWait(msg);
   }
 
-  close(): void {
+  async close(): Promise<void> {
```
mattwiller, line 199: Is this supposed to take a callback?  I thought the code in `packages/agent/src/app.ts` passed one to this method

ThatOneBro: We call `.catch` on the resulting promise and provide a callback it, maybe that's what you are thinking about?

packages/hl7/src/connection.test.ts
```diff
@@ -1,67 +1,74 @@
 import { Hl7Message } from '@medplum/core';
 import iconv from 'iconv-lite';
+import { Duplex } from 'node:stream';
 import { Hl7Connection } from './connection';
 import { CR, FS, VT } from './constants';
 import { Hl7MessageEvent } from './events';
 
+class MockSocket extends Duplex {
```
ThatOneBro: Implemented this `MockSocket` class to handle calls to `socket.end` and `socket.destroy`

packages/agent/src/app.ts
```diff
@@ -856,15 +863,20 @@ export class App {
...
+            this.hl7Clients.delete(message.remote);
+          }
           this.log.info(`Persistent connection to remote '${message.remote}' closed`);
         });
-        client.addEventListener('error', () => {
+        client.addEventListener('error', (event) => {
           this.hl7Clients.delete(message.remote);
-          this.log.info(
-            `Persistent connection to remote '${message.remote}' encountered an error... Closing connection...`
+          this.log.error(
```
ThatOneBro, line 874: This was only logging at info level before, this should definitely be error level

packages/agent/src/app.ts
```diff
@@ -316,10 +316,15 @@ export class App {
...
     const logStatsFreqSecs = agent?.setting?.find((setting) => setting.name === 'logStatsFreqSecs')?.valueInteger;
 
+    // If keepAlive is off and we have clients currently connected, we should stop them and remove them from the clients
     if (!keepAlive && this.hl7Clients.size !== 0) {
-      for (const client of this.hl7Clients.values()) {
-        client.close();
+      const results = await Promise.allSettled(Array.from(this.hl7Clients.values()).map((client) => client.close()));
+      for (const result of results) {
+        if (result.status === 'rejected') {
+          this.log.error(normalizeErrorString(result.reason));
```
mattwiller, line 324: This error message could probably use some context (i.e. that this is occurring while closing connections)

mattwiller (approved):

PR #7179, opened 2025-08-04
Dependency upgrades 2025-08-04

medplumbot:
Dependency upgrades


packages/server/src/otel/instrumentation.ts
```diff
@@ -108,7 +108,9 @@ export function httpResponseHook(
 }
 
 export function pgResponseHook(span: Span, { data }: PgResponseHookInformation): void {
-  span.setAttribute('medplum.db.rowCount', data.rowCount);
+  if (data.rowCount !== null) {
```
ThatOneBro, line 111: Had to add this to prevent attempting to set attribute to a null value, which is explicitly not allowed in the OTel spec.
> Attribute values of null are not valid and attempting to set a null value is undefined behavior.

https://opentelemetry.io/docs/specs/otel/common/#:~:text=Attribute%20values%20of%20null%20are%20not%20valid%20and%20attempting%20to%20set%20a%20null%20value%20is%20undefined%20behavior.

packages/server/src/otel/instrumentation.ts
```diff
@@ -36,8 +36,8 @@ export function initOpenTelemetry(): void {
     return;
   }
 
-  const resource = Resource.default().merge(
-    new Resource({
+  const resource = defaultResource().merge(
```
ThatOneBro, line 39: The latest version of the otel package changed the interface for creating a default resource and how you instantiate resource objects in general

ThatOneBro:
Some of the changes I made in the process of this upgrade:

1. I reverted the `typescript` upgrade from 3.9.x back to 3.8.3. This is mostly because of a type breakage in the call to `xhr.send` which TypeScript 3.9 currently says that our passing in of a `Uint8Array` is invalid since it doesn't have the proper type signature, however the MDN spec says that we can pass in a `TypedArray` which `Uint8Array` is... I'm guessing this is a regression and we should check on this next week.
2. OTel packages upgrade - there were some minor breaking changes in the API but they were easy enough to resolve
3. Had to add `override` for `vite@7.0.6` since some transitive dependencies depend on older versions of Vite which was causing vite to be localized instead of hoisted to root `node_modules`, which was causing the following error:
Screenshot 2025-08-04 at 1 36 00 PM
Screenshot 2025-08-04 at 1 36 26 PM

Looks like everything is still working, the main package I saw which was depending on older versions of Vite was storybook's React docgen plugin

ThatOneBro:
Due to version bump before this PR was done I'm going to abandon this one and do a fresh run of the deps upgrade job and just surgically apply the changes since the diff is pretty gnarly

PR #7182, opened 2025-08-04
Normalize userInfoUrl before validating protocol

mattlong:
e.g. to strip any leading/trailing whitespace

### DIFF
--- a/packages/server/src/oauth/utils.test.ts
+++ b/packages/server/src/oauth/utils.test.ts
@@ -8,6 +8,7 @@ import {
   getAuthTokens,
   getClientApplication,
   getMembershipsForLogin,
+  normalizeUserInfoUrl,
   tryLogin,
   validateLoginRequest,
   validatePkce,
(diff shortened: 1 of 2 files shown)


packages/server/src/oauth/utils.test.ts
```diff
@@ -432,6 +433,28 @@ describe('OAuth utils', () => {
...
+
+  describe('normalizeUserInfoUrl', () => {
+    test.each([
+      ['http://example.com/oauth2/userinfo', false],
+      [' http://example.com/oauth2/userinfo ', false],
+      ['https://example.com/oauth2/userinfo', false],
+      [' https://example.com/oauth2/userinfo ', false],
+      ['file://example.com/oauth2/userinfo', true],
+      [' file://example.com/oauth2/userinfo ', true],
+    ])('with URL [%s]', async (userInfoUrl, expectError) => {
```
ThatOneBro: I think you might want to remove `async` here?

ThatOneBro (approved):

PR #7188, opened 2025-08-05
Dependency upgrades 2025-08-05

medplumbot:
Dependency upgrades

### DIFF
--- a/packages/server/src/otel/instrumentation.test.ts
+++ b/packages/server/src/otel/instrumentation.test.ts
@@ -71,4 +71,14 @@ describe('Instrumentation', () => {
 
     expect(span.setAttribute).toHaveBeenCalledWith('medplum.db.rowCount', 21);
   });
+
+  test('Postgres response hook -- null rowCount', async () => {
+    const span = {
+      setAttribute: jest.fn(),
+    } as unknown as Span;
+
+    pgResponseHook(span, { data: { rowCount: null } } as unknown as PgResponseHookInformation);
+
+    expect(span.setAttribute).not.toHaveBeenCalled();
+  });
 });


packages/server/src/otel/instrumentation.ts
```diff
@@ -36,8 +36,8 @@ export function initOpenTelemetry(): void {
     return;
   }
 
-  const resource = Resource.default().merge(
-    new Resource({
+  const resource = defaultResource().merge(
```
ThatOneBro, line 39: New OTel lib version changed how resources are instantiated

packages/server/src/otel/instrumentation.ts
```diff
@@ -108,7 +108,9 @@ export function httpResponseHook(
 }
 
 export function pgResponseHook(span: Span, { data }: PgResponseHookInformation): void {
-  span.setAttribute('medplum.db.rowCount', data.rowCount);
+  if (data.rowCount !== null) {
```
ThatOneBro, line 111: The `rowCount` prop is now nullable according to the latest pg-instrumentation package, so we have to check before setting the attribute, since attributes cannot be null in OTel

ThatOneBro:
This is a replacement for #7179 since there was a release cut between the generation of the PR and before it was ready for review.

Basically I ran the upgrade with `typescript` and `hibp` excluded, because both of these were causing breakages which didn't seem fixable at the moment.

* Typescript 3.9 didn't like our type passed into a `xhr.send` call but according to the MDN spec our `Uint8Array` should be accepted, and we know that it works at runtime but seems like a regression in the new types which hopefully will get fixed by next week.

* `hibp` was pinned since latest version is ESM-only which we can't do yet.

* I had to fix a few things with the latest OTel versions, but changes were really minor and mostly just instantiating resources slightly differently.

* I added a forced override for `vite` to `7.0.6` since it seemed some minor transitive dependencies (namely storybook React docgen plugin) don't have Vite 7 listed as a compatible version yet. 

* Finally the last thing that was breaking tests was `react` and `react-dom` were being overridden to a mismatched version, which led to a bunch of weird errors in test where the React module being imported for tests was a null object. That led to test failures like this: https://github.com/medplum/medplum/actions/runs/16738470123/job/47382118942#step:10:31627
 So I fixed the override to match the intended versions

codyebberson:
Staging deploy failed: https://github.com/medplum/medplum/actions/runs/16750678075/job/47419823292

codyebberson:
Oh, the last release wasn't published.  Publishing now: https://github.com/medplum/medplum/releases/tag/v4.3.8

codyebberson:
`hibp` changes look good

image

codyebberson (approved):

PR #7189, opened 2025-08-05
Fixed cli-wrapper double execution

codyebberson:
This broke the login command:

image

Drive-by fix, call `process.exit(0);` to speed up the local web server shutdown process.


ThatOneBro:
Might need to rebase

PR #7190, opened 2025-08-05
Add license header to all source files

codyebberson:
This is the last remaining item for OpenSSF "Gold" badge: https://www.bestpractices.dev/en/projects/10900?criteria_level=2

This PR adds the `eslint-plugin-header` to automatically add the Copyright and License statement.

This PR was 99% automated via `eslint . --fix`.  The one case that ESLint auto-fix does not correctly handle is if the file starts with a comment.  If the file starts with a comment, then the plugin replaces the comment, which requires manual touchup.


ThatOneBro:
Can we add this to autofix.ci?

codyebberson:
> Can we add this to autofix.ci?

It's implemented as an ESLint rule, so it should get fixed automatically, yeah?

ThatOneBro:
Ah yeah. I guess that's a good place to start, but would be nice if CI automatically fixed it instead of it requiring a manual run of eslint locally

Currently we don't autofix eslint in CI because most rules are nuanced and half the time errors are not auto-fixable, this is one I think it's pretty straightforward and might be worth breaking out into autofix

codyebberson:
I haven't played with autofix much.  Is that just a matter of adding `npm run lint:fix` to `.github/workflows/autofix-ci.yml` ?

If we replace ESLint+Prettier with Biome in v5, does that solve this?

ThatOneBro:
Yeah we could add `npm run lint:fix` to autofix, it might fix a lot of issues potentially

Moving to Biome is a separate story and not sure this is fixed by that because not sure if they have this same rule for headers 🤔 https://github.com/biomejs/biome/discussions/206

mattlong (approved):
LGTM. I spot checked a lot of the files listed in `git show --stat` that differed from the common case of adding 2 or 3 lines and they all seemed reasonable.

PR #7195, opened 2025-08-05
$set-accounts only updating meta.accounts and meta.account

finnbergquist:
### Current behavior
When you call `Patient/<id>/$set-accounts`, and a resource in that Patient's compartment has a `meta.security`, the `meta.security` field gets unintentionally deleted. 

### Desired behavior
The `Patient/<id>/$set-accounts` does not touch any of the other `meta` fields

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -628,9 +628,9 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       let result: WithId<T>;
       if (options?.ifMatch) {
         // Conditional update requires transaction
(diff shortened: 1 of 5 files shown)


mattwiller (approved):

packages/server/src/fhir/repo.ts
```diff
@@ -669,7 +671,8 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
   private async updateResourceImpl<T extends Resource>(
     resource: T,
     create: boolean,
-    versionId?: string
+    versionId?: string,
+    inheritAccounts?: boolean
```
mattwiller: Do we want to start a new `options` Object here, to accommodate any future parameters as well?

PR #7199, opened 2025-08-06
MCP OAuth updates

codyebberson:
1. The `/.well-known/oauth-protected-resource` endpoint now requires a `resource` property
2. The `/oauth2/register` endpoint response now requires a `redirect_uris` property
3. `ClientApplication.redirectUri` needs to be an array
   1. I don't know if any of the major providers necessarily need this, but MCP Inspector does, and it seems aligned with the OAuth spec.  This was a miss on our original `ClientApplication` implementation 😢 
   2. Deprecating the original `ClientApplication.redirectUri`, but still fully supported for backwards compatibility
   4. Added a new `ClientApplication.redierctUris` which is an array
   5. Added `server` internal utilities to smooth this over
3. Fixed bug in `defaultScopes` - need to join by space rather than default join by comma

Aside: In most places, Medplum requires an exact match with the redirect URI.  There is one place in `/auth/external` where we allow a partial match as long as the requested redirect URI starts with one of the predefined values.  While many OAuth servers use similar logic, this is not actually spec compliant, and has some long tail security considerations.  I added a "warn" logging statement in the `external` handler for cases where a client is depending on this partial matching behavior.  I recommend we let it run for a period of time, migrate any customers who are currently depending on this behavior, and then close that hole.

### DIFF
--- a/packages/server/src/oauth/clients.test.ts
+++ b/packages/server/src/oauth/clients.test.ts
@@ -20,11 +20,8 @@ describe('OAuth clients', () => {
     expect(getClientRedirectUri(client, 'https://example.com/callback?param=value', true)).toBe(
       'https://example.com/callback?param=value'
     );
-    expect(getClientRedirectUri(client, 'https://example.com/callback?param=value', false)).toBe(
-      'https://example.com/callback?param=value'
-    );
-    expect(getClientRedirectUri(client, 'https://example.com/other/path', true)).toBeUndefined();
     expect(getClientRedirectUri(client, 'https://example.com/callback?param=value', false)).toBeUndefined();
+    expect(getClientRedirectUri(client, 'https://example.com/other/path', true)).toBeUndefined();
   });
 
   test('getClientRedirectUri with deprecated redirectUri', () => {


reshmakh (approved):

packages/server/medplum.config.json
```diff
@@ -21,6 +21,41 @@
   "defaultBotRuntimeVersion": "vmcontext",
   "allowedOrigins": "*",
   "introspectionEnabled": true,
+  "logRequests": true,
```
mattlong: Clean up/remove all this?

packages/server/src/oauth/clients.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    expect(getClientRedirectUri(client, 'https://example.com/callback')).toBe('https://example.com/callback');
+    expect(getClientRedirectUri(client, 'https://example.com/redirect')).toBe('https://example.com/redirect');
+    expect(getClientRedirectUri(client, 'https://example.com/other')).toBeUndefined();
+    expect(getClientRedirectUri(client, 'https://malicious.com/callback')).toBeUndefined();
+
+    // Partial matching is not recommended, but supported with the allowPartial flag
+    expect(getClientRedirectUri(client, 'https://example.com/callback?param=value', true)).toBe(
+      'https://example.com/callback?param=value'
+    );
```
mattlong, line 22:
```suggestion
    expect(getClientRedirectUri(client, 'https://example.com/callback?param=value', true)).toBe(
      'https://example.com/callback?param=value'
    );
    expect(getClientRedirectUri(client, 'https://example.com/callback?param=value', false)).toBeUndefined();
```

mattlong (approved):
couple suggestions

reshmakh:
@claude please review

codyebberson:
For reference:

```json
"mcpEnabled": true,
  "defaultOAuthClients": [
    {
      "resourceType": "ClientApplication",
      "id": "mcp-inspector-1",
      "name": "MCP Inspector 1",
      "redirectUris": ["http://127.0.0.1:6274/oauth/callback", "http://127.0.0.1:6274/oauth/callback/debug"],
      "defaultScope": ["openid", "profile", "email", "phone", "address"],
      "pkceOptional": true
    },
    {
      "resourceType": "ClientApplication",
      "id": "mcp-inspector-2",
      "name": "MCP Inspector 2",
      "redirectUris": ["http://localhost:6274/oauth/callback", "http://localhost:6274/oauth/callback/debug"],
      "defaultScope": ["openid", "profile", "email", "phone", "address"],
      "pkceOptional": true
    },
    {
      "resourceType": "ClientApplication",
      "id": "claude",
      "name": "Claude",
      "redirectUris": ["https://claude.ai/api/mcp/auth_callback"],
      "pkceOptional": true
    },
    {
      "resourceType": "ClientApplication",
      "id": "chatgpt",
      "name": "ChatGPT",
      "secret": "foo",
      "redirectUris": ["https://chatgpt.com/connector_platform_oauth_redirect"],
      "pkceOptional": true
    }
  ],
```

PR #7206, opened 2025-08-07
Update azure storage to handle string data in writeFile method

### LINKED ISSUE #7196: AzureBlobStorage.writeFile - TypeError: this.readable.on is not a function
Bug reported in our Fission+Azure set up - https://discord.com/channels/905144809105260605/1402374880304824430/1402374880304824430

> I'm excited to be trying out the new Fission bot support on Azure, but I've run into an issue in my dev environment. I'm hoping someone can help me determine if it's a configuration problem.

> While Medplum and Fission seem to be working, I get an exception when I click 'Run' in the bot editor. The bot saves to Azure Blob Storage and deploys to Fission correctly. I can execute the bot manually via the Fission CLI and direct HTTP requests.

> However, running it from the editor page fails. The Medplum logs show an excemption: 'TypeError: this.readable.on is not a function', which is thrown by AzureBlobStorage.writeFile. I suspect this is triggered by the writeBotInputToStorage function in the Medplum server.

> This seems like a storage configuration issue, but I'm not sure why Medplum can write the bot's code to storage but not the bot's input. I'm planning to debug the server locally, but I wanted to check if anyone has seen this before.

> Any help would be much appreciated!

dougjmac:
## Description

Updates the Azure Blob Storage `writeFile` method to handle both string and stream data types.

## Problem

The `writeFile` method in `AzureBlobStorage` was only handling `BinarySource` of type `Readable` using `uploadStream()`. When string data was passed (e.g., from JSON uploads via `writeBotInputToStorage`), it would fail with: 'TypeError: this.readable.on is not a function'. This is because the `writeBotInputToStorage` method calls `JSON.stringify()` on the data before passing it to the `writeFile` method  and azure/storage-blob library only accepts a `Readable` stream as a parameter in the `uploadStream()` function. To upload strings, the `upload` can be used. 

## Solution 

I followed the pattern in the GoogleCloudStorage class:

- use the `isString()` function from `medplum/core` to check the type of the `BinarySource` data. 
- use the `azure/storage-blob` `upload` method for uploading string data and continue to use `uploadStream` for all other data. 

[Microsoft docs for uploading blob from a string](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-upload-javascript?tabs=typescript#upload-a-block-blob-from-a-string)

I think this is the simplest solution to the problem but let me know if you'd be happier with a different approach. I haven't added tests because I'm not sure if they would be useful to test these changes - it might be possible to check that the expected method is called on a mocked `AzureBlobStorage`?

## Related Issues

Fixes #7196


mattwiller:
@dougjmac Thanks for the PR — the change looks good to me 👍 I do think a test would be useful, but also that it would probably just look like asserting that a mock gets called with the string as input — would you mind adding a test case like for that?

Also, just wanted to give you a heads up that it looks like you'll need to [signoff your commit](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository) to satisfy the DCO check in CI

dougjmac:
> @dougjmac Thanks for the PR — the change looks good to me 👍 I do think a test would be useful, but also that it would probably just look like asserting that a mock gets called with the string as input — would you mind adding a test case like for that?
> 
> Also, just wanted to give you a heads up that it looks like you'll need to [signoff your commit](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository) to satisfy the DCO check in CI

@mattwiller Thanks for the review. I've added some test cases to cover the changes. The tests mock out the azure storage layer and the tests verify that the expected methods on the mock object are called. Let me know if you'd like any more changes. 

I've attempted to sign-off my commits but let me know if there are any issues with that.

codyebberson (approved):

PR #7207, opened 2025-08-07
Docs binary-security-context typo/bug

finnbergquist:


packages/docs/docs/access/binary-security-context.md
```diff
@@ -96,7 +96,9 @@ const media = await medplum.createMedia({
   data: fileData,
   filename: 'patient-report.pdf',
   contentType: 'application/pdf',
-  subject: { reference: `Patient/${patientId}` },
+  additionalFields: {
+    subject: { reference: `Patient/${patientId}` },
+  }
```
mattwiller: This example can/should be moved into the `examples` directory, so errors like this can be caught and constantly checked against the actual SDK code

mattwiller (approved):

packages/docs/docs/access/binary-security-context.md
rahul1: +1 . @finnbergquist let me know if you need help with the `MedplumCodeBlock`

PR #7208, opened 2025-08-07
Add Helm chart support for initContainers and envFrom

WELLaaronaston:
- Add support for initContainers in deployment spec
- Add support for envFrom to inject environment variables from ConfigMaps/Secrets
- Both features are optional and controlled via values.yaml


codyebberson:
Moved here: https://github.com/medplum/medplum/pull/7233

PR #7210, opened 2025-08-07
`projectId` column NOT NULL

### LINKED ISSUE #7138: Make `projectId` column `NOT NULL`
The `projectId` column in resource tables is not actually NOT NULL since some resources are considered “system” resources that do not belong to a specific project. For safety we should strongly consider formally making the column not nullable before attempting sharding. A well-known value can be specified for “system” resources to meet the constraint.

mattlong:
Fixes #7138


packages/server/src/fhir/repo.ts
```diff
@@ -434,6 +435,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     }
 
     const resource = JSON.parse(rows[0].content as string) as WithId<T>;
+    //TODO{mattlong} replace meta.project of UUID_NIL with undefined
```
mattwiller: Does this case still need to be handled?

packages/server/src/fhir/sql.ts
```diff
@@ -20,6 +20,11 @@ export type OperatorFunc = (sql: SqlBuilder, column: Column, parameter: any, par
 
 export type TransactionIsolationLevel = 'READ COMMITTED' | 'REPEATABLE READ' | 'SERIALIZABLE';
 
+/**
+ * The UUID used to populate NON NULL "projectId" columns for resources not associated with a specific project.
+ */
+export const SYSTEM_PROJECT_ID = 'ffffffff-ffff-ffff-ffff-ffffffffffff';
```
mattwiller: Why not just use the zero UUID here?  That would effectively be the default value for the DB  `uuid` field anyway

packages/server/src/fhir/repo.ts
mattlong: Thanks. No, it does not since we don't use the `projectId` column when returning a read resource; just `content`.

packages/server/src/fhir/sql.ts
mattlong: Yeah, I'm not quite sure what would be the best value to use here. Any hard-coded value runs the risk of colliding with an existing project's ID on a self-hosted instance of Medplum. Is it better to use a clearly artificial value like the zero UUID or would it be safer to choose a `v5` UUID like we did for the `R4` project? I lean towards an artificial one to aid human debugability.

Either way, for defense in depth, we should adda constraint on the `Project.id` column to prevent a row with the placeholder UUID from being inserted. If we do use the zero or the `ffff...` UUID, we could more easily add a range of reserved UUIDS. e.g.:

```SQL
ALTER TABLE "Project"
    ADD CONSTRAINT project_reserved_ids
    CHECK(id > '00000000-0000-0000-0000-0000000000ff');
```

codyebberson:
@claude please review

PR #7222, opened 2025-08-08
Fixed lint error in provider

codyebberson:


codyebberson:
handled in https://github.com/medplum/medplum/pull/7218

Issue #6819, opened 2025-06-20
DoseSpot - Syncing Medication Favorites


finnbergquist:
### Goal
Reduce time required to put together a prescription in DoseSpot iframe. A lot of people only end up prescribing a few different medications, so we should give people a way to have the favorites programmatically synced.

### Get Favorite Prescriptions Bot

_Retrieve Existing Favorites - DoseSpot docs Section 3.19.1_

### Add new Favorite Prescription Bot
There seem to be a few different ways to do this

**on a per clinician basis:**
- Section 3.19.2 - Add Clinic Favorite Medication (coded)
- Section 3.19.4 - Add Clinic Favorite Compiled Compound
- Section 3.19.6 - Add Clinic Favorite Supply

**on a per clinic wide basis:**
- Section 3.20.2 - Add Clinic Favorite Medication (coded)
- Section 3.20.4 - Add Clinic Favorite Compiled Compound
- Section 3.20.6 - Add Clinic Favorite Supply
note: this may require some admin credentialing, which we need to figure out

finnbergquist:
Section 3.20.2 - Add Clinic Favorite Medication (coded)  seems to be the best way to do this

PR #7223, opened 2025-08-08
DoseSpotFormularyCenter react hook and components in provider

finnbergquist:
- `UseDoseSpotFormularyCenter` react hook allows people to search for medications and add them to their favorites without interacting with the nested `MedicationRequest` objects that are returned by the search
- There are two new bots to service this feature. [Here](https://github.com/medplum/medplum-ee/pull/403) is the PR for those
- Example components in Provider app for clinic wide page to set favorites

Addresses [#6819](https://github.com/medplum/medplum/issues/6819)


techdavidy:
@claude please review

examples/medplum-provider/src/pages/integrations/FavoriteMedicationsTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+        };
+        const queryString = formatSearchQuery(searchRequest);
+        const result = await medplum.search('MedicationKnowledge', queryString);
+        setMedications(result.entry?.map((entry) => entry.resource as MedicationKnowledge) || []);
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    loadMedications().catch(console.error);
```
techdavidy: show error notification please

examples/medplum-provider/src/pages/integrations/DoseSpotFavoritesPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * This is a demo component for how you could display your favorite Medications
+ * from DoseSpot.
+ *
+ * @returns A React component that displays the favorite medications and allows you to add new favorites.
+ */
+export function DoseSpotFavoritesPage(): React.JSX.Element {
+  const [modalOpened, setModalOpened] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [refreshKey, setRefreshKey] = useState(0);
```
techdavidy: is the refreshKey used at all ?

examples/medplum-provider/src/pages/integrations/DoseSpotFavoritesPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+      setLoading(true);
+      await addFavoriteMedication(medication);
+      setModalOpened(false);
+      showNotification({
+        title: 'Medication added to favorites',
+        message: 'The medication has been added to your favorites',
+        color: 'green',
+      });
+    } catch (error) {
+      showNotification({
```
techdavidy: provider has a `showErrorNotification`

examples/medplum-provider/src/pages/integrations/DoseSpotFavoritesPage.tsx
finnbergquist: Yea, it is part of the useEffect dependency array in FavoriteMedicationsTable.tsx to reload medications. Would you suggest a different way to do that @techdavidy ?

techdavidy:
@claude why do need the delay ?

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const setDirections = (directions: string | undefined): void => {
+    privateSetDirections(directions);
+  };
+
+  const setSelectedMedication = (medication: MedicationKnowledge | undefined): void => {
+    privateSetSelectedMedication(medication);
+  };
+
+  const getMedicationName = (medication: MedicationKnowledge | undefined): string => {
```
rahul1: because this doesn't use any state from the hook, i recommend exporting this as a separate utility function (either from `core` or this package) so clients don't need to instantiate the hook to use it.

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
```diff
@@ -0,0 +1,104 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import { MedicationKnowledge } from '@medplum/fhirtypes';
+import { useMedplum } from '@medplum/react-hooks';
+import { useCallback, useState } from 'react';
+import { DOSESPOT_ADD_FAVORITE_MEDICATION_BOT, DOSESPOT_SEARCH_MEDICATIONS_BOT } from './common';
+
+export interface DoseSpotClinicFormularyReturn {
+  state: DoseSpotClinicFormularyState;
+  /**
+   * Search for DoseSpot Medications and returns array of synthetic MedicationKnowledge objects that are not yet saved to the FHIR server
```
rahul1:
```suggestion
   * Search for DoseSpot Medications and returns array of temporary MedicationKnowledge objects that are not yet saved to the FHIR server
```

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  readonly setSelectedMedication: (medication: MedicationKnowledge | undefined) => void;
+  /**
+   * Set the directions for the currently selected medication
+   */
+  readonly setDirections: (directions: string | undefined) => void;
+  /**
+   * Add a DoseSpot Medication to the Clinic's favorites and returns the MedicationKnowledge object that was added
+   */
+  readonly addFavoriteMedication: () => Promise<MedicationKnowledge>;
```
rahul1: to make it clearer that this persists to the server, a clearer name might be `saveFavoriteMedication`

examples/medplum-provider/src/pages/integrations/DoseSpotFavoritesPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+      });
+    } catch (error) {
+      showErrorNotification({
+        title: 'Error adding medication to favorites',
+        message: normalizeErrorString(error as Error),
+        color: 'red',
+      });
+    } finally {
+      setSelectedMedication(undefined);
+      setDirections(undefined);
```
rahul1: People will forget to do this. I recommend adding a `clear()` method as well - we probably should have that for the health gorilla hook as well

rahul1 (approved):

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
```diff
@@ -0,0 +1,104 @@
...
+import { useMedplum } from '@medplum/react-hooks';
+import { useCallback, useState } from 'react';
+import { DOSESPOT_ADD_FAVORITE_MEDICATION_BOT, DOSESPOT_SEARCH_MEDICATIONS_BOT } from './common';
+
+export interface DoseSpotClinicFormularyReturn {
+  state: DoseSpotClinicFormularyState;
+  /**
+   * Search for DoseSpot Medications and returns array of synthetic MedicationKnowledge objects that are not yet saved to the FHIR server
+   */
+  readonly searchMedications: (searchTerm: string) => Promise<MedicationKnowledge[]>;
```
rahul1: One API note here, since the Bots are also WIP. I would recommend that the `searchMedications` and `setSelectedMedication` to use the minimal set of information, which would be a `CodeableConcept`

```ts
readonly searchMedications: (searchTerm: string) => Promise<CodeableConcept[]>;
readonly setSelectedMedication: (medication: CodeableConcept | Coding | undefined) => void;
```

The reasons being: 
- this is closer to our "builder pattern", where we are inputing the individual fields, and the hook is responsible for constructing the resource. this is the pattern we follow in the HG hook, where we just select `TestCodings`, not `ServiceRequests`
- It makes the hook more modular. 
  - For example, say the user not using the `searchMedications` function, but selecting from a valueset (using our `CodingInput`). 
  - Then they could call `setSelectedMedication` without having to wrap it in a `MedicationKnowledge` resource first
 - It sets us up nicely for when we start using `ActivityDefinitions`. For example:
   -  Imagine in a future PR we add `setNumberRefills`
   - Then, with the state of (codeable concept, directions, refills) the hook can sync to dosespot and generate an `ActivityDefinition` and `MedicationKnowledge` output

rahul1 (commented):
This has come a really long way. I have one last suggestion for the API of the Bots / hook to use `CodeableConcepts` vs. `MedicationKnowledge` as the intermediate format

rahul1 (changes requested):
One last API suggestion re: `searchMedications` output

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
finnbergquist: Yea, this all makes sense. 

Definitely agree that this adds flexibility if we want to provide a ValueSet with Medispan(dispensable drug) ids in the future

`readonly setSelectedMedication: (medication: CodeableConcept | Coding | undefined) => void;
`

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
```diff
@@ -1,54 +1,72 @@
...
 export interface DoseSpotClinicFormularyState {
-  selectedMedication: MedicationKnowledge | undefined;
+  selectedMedication: CodeableConcept | Coding | undefined;
   directions: string | undefined;
 }
 
 export function useDoseSpotClinicFormulary(): DoseSpotClinicFormularyReturn {
   const [directions, privateSetDirections] = useState<string | undefined>(undefined);
-  const [selectedMedication, privateSetSelectedMedication] = useState<MedicationKnowledge | undefined>(undefined);
+  const [selectedMedication, privateSetSelectedMedication] = useState<CodeableConcept | Coding | undefined>(undefined);
```
rahul1: nit: to simplify the state for the end user, we can make this `CodeableConcept |  undefined`

We can accept `CodeableConcept | Coding | undefined`, but if it's a coding we can just wrap it in a `CodeableConcept`. 

That way the client doesn't have to do many checks to see if we've saved a `Coding` or `CodeableConcept` on their behalf

packages/dosespot-react/src/useDoseSpotClinicFormulary.ts
```diff
@@ -1,54 +1,72 @@
...
 
+    // Create the code property based on the type of selectedMedication
+    let code: CodeableConcept;
+
+    if (isCodeableConcept(selectedMedication)) {
+      // selectedMedication is already a CodeableConcept
+      code = {
+        text: selectedMedication.text,
+        coding: selectedMedication.coding,
+      };
```
rahul1:
```suggestion
      code = {...selectedMedication};
```

examples/medplum-provider/src/pages/integrations/DoseSpotFavoritesPage.tsx
```diff
@@ -122,26 +125,25 @@ export function DoseSpotFavoritesPage(): React.JSX.Element {
...
         size="lg"
         withCloseButton
       >
         <Box>
-          <AsyncAutocomplete<MedicationKnowledge>
+          <AsyncAutocomplete<CodeableConcept>
             placeholder="Search medications..."
             loadOptions={searchMedications}
             toOption={toOption}
             itemComponent={({ resource }) => (
```
rahul1: nit 
```suggestion
            itemComponent={({ concept }) => (
```

PR #7228, opened 2025-08-09
Limit history items shown

mattlong:
We may want to refine if/when we limit the number of items shown; e.g. only bother doing this if `items.length > 20`, but this seems like a reasonable starting point.


mattwiller:
Could this be paginated instead?

codyebberson (approved):

PR #7230, opened 2025-08-11
Dependency upgrades 2025-08-11

medplumbot:
Dependency upgrades


codyebberson:
Hey @techdavidy -

Here is the build error: 
image


Here's what I think happened:
* The upgrade job ran and completed successfully: https://github.com/medplum/medplum/actions/runs/16875717806/job/47799731760
* However, now `npm ci` is failing: https://github.com/medplum/medplum/actions/runs/16875831771/job/47800101297?pr=7230
* The upgrade job updated `vite` to `7.1.1`
* But it did not upgrade the root `overrides` in `package.json`, which is still at `7.0.6`
* That mismatch causes `npm ci` to fail
* Quick fix is to change the `vite` version in `overrides` from `7.0.6` to `7.1.1`
* But why didn't our upgrade script upgrade the overrides?
* I don't know for sure, but what I assume happened is that the `npm-check-updates` tool only upgrades `overrides` if the dependency is in the same `package.json`.  So maybe we should also add `vite` to the root `package.json`?
* After making those 2 changes, then run `npm i --strict-peer-deps` from the root

For reference, here is our general guide on working with `package.json`: https://www.medplum.com/docs/contributing/package-json

ThatOneBro:
@techdavidy I reverted typescript upgrade last week too, seems like we are still seeing a type breakage where there shouldn't be one (`BinarySource` in `@medplum/core`)

techdavidy:
- Moved vite to 7.1.1 and included vite to root.
- Kept typescript to 5.8.3
- Added `moduleResolution": "bundler` to app, examples and graphiql

examples/medplum-demo-bots/package.json
```diff
@@ -54,7 +54,7 @@
     "ssh2-sftp-client": "12.0.1",
     "stripe": "18.4.0",
     "twilio": "^5.8.0",
-    "typescript": "5.8.3",
+    "typescript": "5.9.2",
```
ThatOneBro:
```suggestion
    "typescript": "5.8.3",
```

examples/medplum-client-external-idp-demo/package.json
```diff
@@ -23,20 +23,20 @@
...
     "postcss": "8.5.6",
     "postcss-preset-mantine": "1.18.0",
     "react": "19.1.1",
     "react-chartjs-2": "5.3.0",
     "react-dom": "19.1.1",
-    "react-router": "7.7.1",
-    "typescript": "5.8.3",
-    "vite": "7.0.6"
+    "react-router": "7.8.0",
+    "typescript": "5.9.2",
```
ThatOneBro:
```suggestion
    "typescript": "5.8.3",
```

package.json
```diff
@@ -72,7 +72,8 @@
     "ts-node": "10.9.2",
     "tslib": "2.8.1",
     "turbo": "2.5.5",
-    "typescript": "5.8.3"
+    "typescript": "5.8.3",
```
codyebberson, line 75: It looks like there might be 2 different typescript versions?  Here it says `5.8.3`.  In several other package.json files it says `"typescript": "5.9.2"`.  We should keep those all consistent.

PR #7235, opened 2025-08-12
Update $set-accounts operation

mattwiller:
A few related changes to make the `$set-accounts` operation more flexible:
- Enabled the operation on instances of all resource types
- Made the compartment propagation behavior opt-in with a `propagate` parameter
- Computed the updated accounts for each compartment resource and updated with system repo, instead of always inheriting from the parent resource
- Refactored the internal logic for getting the patient compartment, avoiding wasted work resolving resources outside the compartment for inclusion in `Patient/$everything`


packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -0,0 +1,134 @@
...
+    ...target.meta,
+    accounts: accounts,
+    account: accounts?.[0],
+  };
+  await repo.updateResource(target);
+  let count = 1; // Target resource is updated already
+
+  if (params.propagate && target.resourceType === 'Patient') {
+    // Calculate the difference between the previous accounts array and new one, in order to
+    // propagate only those changes to compartment resources
```
finnbergquist, line 104: Nice

packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -0,0 +1,134 @@
...
+    const additions = accounts.filter((a) => !oldAccounts?.find((o) => o.reference === a.reference));
+    const removals = oldAccounts?.filter((o) => !accounts.find((a) => a.reference === o.reference)) ?? [];
+
+    // Update the resources in the target compartment to trigger meta.accounts refresh
+    // const bundle = await getPatientEverything(repo, target);
+    const bundle = await searchPatientCompartment(repo, target);
+    for (const entry of bundle.entry ?? []) {
+      const resource = entry.resource;
+      if (resource && resource.resourceType !== 'Patient') {
+        let accountList = resource.meta?.accounts;
```
finnbergquist, line 113: Do we need to check if the resource is not a Project reference as well? 

Does updating resource now with system repo change any intial assumptions about this where Project refernces could not be added to meta.compartment

packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -0,0 +1,134 @@
...
+  const params = parseInputParameters<SetAccountsParameters>(operation, req);
+  const { repo } = getAuthenticatedContext();
+
+  const isSuperAdmin = repo.isSuperAdmin();
+  if (!repo.isProjectAdmin() && !isSuperAdmin) {
+    return [forbidden];
+  }
+
+  const target = await repo.readResource(resourceType, id);
+  const accounts = params.accounts;
```
finnbergquist: Do you think that we should force people to have 'x-medplum'='extended' header so that they make sure to get their existing accounts here. 

Might be confusing/unexpected behavior depending on which mode they are in. Or at least we could recommend to have it on in documentation. Thoughts?

packages/server/src/fhir/operations/set-accounts.ts
mattwiller: This is adding to `meta.account`, not `meta.compartment`, so I don't think there should generally be any Project resources?

packages/server/src/fhir/operations/set-accounts.ts
mattwiller: That's a good idea — I'll see if I can just force it on in a sensible way

mattwiller:
Documentation preview: https://medplum-www-git-set-accounts-all-medplum.vercel.app/docs/api/fhir/operations/patient-set-accounts

packages/docs/docs/api/fhir/operations/patient-set-accounts.md
```diff
@@ -2,37 +2,45 @@
...
-POST [base]/R4/Patient/<id>/$set-accounts
+POST [base]/R4/<ResourceType>/<id>/$set-accounts
 ```
 
 ### Input
+
 The input is a [FHIR Parameters](/docs/api/fhir/resources/parameters) resource containing:
+
 - `accounts` a reference to set in each resource's _meta.accounts_
+- `propagate` an optional boolean, which instruct the operation to also update resources in the target compartment
```
mattlong:
```suggestion
- `propagate` an optional boolean, which instructs the operation to also update resources in the target compartment
```

packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -0,0 +1,140 @@
...
+  let count = 1; // Target resource is updated already
+
+  if (params.propagate && target.resourceType === 'Patient') {
+    // Calculate the difference between the previous accounts array and new one, in order to
+    // propagate only those changes to compartment resources
+    const additions = accounts.filter((a) => !oldAccounts?.find((o) => o.reference === a.reference));
+    const removals = oldAccounts?.filter((o) => !accounts.find((a) => a.reference === o.reference)) ?? [];
+
+    // Update the resources in the target compartment to trigger meta.accounts refresh
+    // const bundle = await getPatientEverything(repo, target);
```
mattlong: cleanup

packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -0,0 +1,140 @@
...
+          }
+        }
+
+        resource.meta = {
+          ...resource.meta,
+          accounts: accountList,
+          account: accountList?.[0],
+        };
+        // Use system repo to force update meta.accounts
+        await systemRepo.updateResource(resource);
```
mattlong, line 132: Since these updates are happening serially, is this at risk of hitting the request timeout if there are the 1,000 resources? Would it be preferable to fire off the updates in batches that get awaited with `Promise.all` or parallelize it some other way?

packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -0,0 +1,140 @@
...
+  accounts: Reference[];
+  propagate?: boolean;
+}
+
+/**
+ * Handles the $set-accounts operation.
+ * This operation updates the account reference for  a resource, and optionally
+ * all resources in the target compartment as well.
+ *
+ * NOTE: the operation is currently capped at 1000 resources in the patient compartment.
```
mattlong, line 61: The 1,000 resource cap feels like a footgun waiting to happen. Maybe there have been conversations about this and its an acceptable risk for now, but I wonder if it would be preferable to throw an error if we detect there are more than 1,000 resources that need updating?

mattwiller:
@mattlong Agreed that the page size limit is problematic: I'm planning to address that in a followup PR to add an async mode

mattlong (approved):

PR #7240, opened 2025-08-12
[Self-hosting Docs] Monitoring Recommendations

maddyli:


packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+**Memory Usage** tracks container memory consumption, useful for detecting leaks or capacity planning.
+
+**Network I/O** provides supporting information about data transfer volumes, helpful for understanding traffic patterns and response sizes.
+
+## Alerting Strategy
+
+### High-Priority Alerts
+Set up immediate alerts for metrics that directly impact users:
+- Load balancer error rates (especially 504s)
+- Database writer CPU > 60-70%
```
mattlong: I'd suggest this be a bit lower of a range, maybe 50%-60%, but sustained for a few minutes. Thoughts, @codyebberson?

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+Monitoring metrics are organized by layer in your stack, from highest-level user impact down to infrastructure details. The higher-level metrics are generally more useful for spotting issues, while lower-level metrics provide supporting information for diagnosis.
+
+**Metric Sources:**
+- **Medplum server metrics**: Available across all deployment environments
+- **Load balancer metrics**: Available from any load balancer (AWS ALB, nginx, etc.)
+- **Database metrics**: Standard PostgreSQL metrics available from any Postgres installation
+- **Cache metrics**: Standard Redis metrics available from any Redis installation
+- **Container metrics**: Available from any container orchestration platform
+
+## Server-Level Metrics
```
mattlong, line 20: I believe (all?) the metrics in this section are our otel-based metrics. That should be mentioned somewhere along with a link to configuring otel.

It's probably also worth specifying that all our otel metrics are prefixed with `medplum.`

Also, for each metric, should we mention the full path, e.g. `medplum.subscription.waitingCount`? Or would that be too busy?

mattlong (approved):

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+When running a self-hosted Medplum instance, proper monitoring is essential for maintaining system health and performance. This guide covers the key metrics Medplum recommends self-hosting users to track and what information they surface about your deployment. 
+
+## Overview
+
+Monitoring metrics are organized by layer in your stack, from highest-level user impact down to infrastructure details. The higher-level metrics are generally more useful for spotting issues, while lower-level metrics provide supporting information for diagnosis.
+
+**Metric Sources:**
+- **Medplum server metrics**: Available across all deployment environments
+- **Load balancer metrics**: Available from any load balancer (AWS ALB, nginx, etc.)
```
mattwiller: I'd recommend softening this to something like "generally available from cloud load balancers, e.g. AWS ALB".  If someone is running their own HAProxy or something, it's not going to come with turnkey metrics that would be available in a dashboard.  Also, I'm not sure how we might want to convey this, but "available" in this case might mean "the service has an HTTP endpoint that returns stats", not necessarily "has a pipeline for transforming those stats into time-series metrics and feeding them into a dashboard".  In cloud environments the cloud provider probably has it all wired up, but it might vary a bit between clouds and if you're on bare metal you're kind of on your own for that piece of the observability backend

packages/docs/docs/self-hosting/monitoring.md
mattwiller: Agreed that it's worth pointing out that these are specifically OpenTelemetry metrics emitted by the server, with a link to `/docs/self-hosting/opentelemetry`

I'd also recommend spelling out the full `medplum.foo.bar` metric name for each one, since that's what users would be searching for in their observability UI

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+- **Database metrics**: Standard PostgreSQL metrics available from any Postgres installation
+- **Cache metrics**: Standard Redis metrics available from any Redis installation
+- **Container metrics**: Available from any container orchestration platform
+
+## Server-Level Metrics
+
+These metrics come directly from the Medplum server and provide the richest view of application behavior.
```
mattwiller:
```suggestion
These metrics come directly from the Medplum server and provide the most comprehensive view of application behavior.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Connection Health
+**Round-Trip Time (RTT)** tracks network latency between the server and both PostgreSQL and Redis. Spikes indicate connectivity issues or performance bottlenecks that can slow down the entire application.
```
mattwiller: Probably worth being specific that there are two distinct metrics here, measuring latency to each of our data stores

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+**Pending Subscriptions** shows subscriptions waiting to be processed. High queue depths correlate with database performance issues since subscriptions generate audit events that require database writes.
+
+## Load Balancer Metrics
+
+These metrics measure what users actually experience and provide the most accurate view of external performance.
+
+### Traffic and Performance
+**Request Count** tracks total incoming requests. Compare with server-level metrics to identify discrepancies that might indicate routing issues.
+
+**Response Time** measured at the load balancer level reflects actual user experience better than server-side measurements. Track both median (P50) and 99th percentile (P99) response times.
```
mattwiller:
```suggestion
**Response Time** measured at the load balancer level reflects actual user experience better than server-side measurements. We recommend tracking both median (P50) and 99th percentile (P99) response times, instead of just the average, to better identify performance issues.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+**FHIR Interaction Counts** break down API activity by operation type (reads, searches, writes). This helps you understand usage patterns and identify which operations are consuming the most resources.
+
+**FHIR Error Rates** track failed operations by type. Some fluctuation is normal, especially for writes at low volume, but sustained increases warrant investigation.
+
+### Subscription Processing
+**Pending Subscriptions** shows subscriptions waiting to be processed. High queue depths correlate with database performance issues since subscriptions generate audit events that require database writes.
+
+## Load Balancer Metrics
+
+These metrics measure what users actually experience and provide the most accurate view of external performance.
```
mattwiller:
```suggestion
These metrics measure traffic closer to the end user and provide a more accurate view of external performance.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Traffic and Performance
+**Request Count** tracks total incoming requests. Compare with server-level metrics to identify discrepancies that might indicate routing issues.
+
+**Response Time** measured at the load balancer level reflects actual user experience better than server-side measurements. Track both median (P50) and 99th percentile (P99) response times.
+
+### Error Tracking
+**HTTP Error Responses** (4xx, 5xx) measured at the load balancer are the definitive source for user-facing errors. Pay special attention to 504 timeouts, which occur when the server takes too long to respond even if the operation eventually succeeds.
```
mattwiller:
```suggestion
**HTTP Error Responses** (4xx, 5xx) measured at the load balancer are the definitive source for user-facing errors. Pay special attention to 504 timeout errors, which occur when the server takes too long to respond and the load balancer returns an error to the user — even if the operation eventually succeeds on the server.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+PostgreSQL metrics provide detailed insight into data layer performance.
+
+### Resource Usage
+**CPU Utilization** is critical, especially for the writer instance. Since there's only one writer (single point of failure), keep writer CPU below 60-70%. Reader instances can run hotter since you can scale horizontally by adding more readers.
+
+**Connection Count** tracks active database connections. PostgreSQL creates roughly one process per connection, so monitor to avoid resource exhaustion.
+
+### Query Efficiency
+**Tuples Read/Written** measure database activity at the row level. A "tuple" is essentially a row, though it can also exist in indexes.
```
mattwiller:
```suggestion
**Tuples Read/Written** measure database activity at the row level. A "tuple" is essentially a single row in a database table, but they are also used to build database indexes.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Tuples Scanned vs. Fetched** reveals query efficiency. When scanned significantly exceeds fetched, the database is loading data it ultimately discards, indicating inefficient queries or missing indexes.
```
mattwiller:
```suggestion
**Tuples Scanned vs. Fetched** reveals query efficiency. When scanned significantly exceeds fetched, the database is reading data it ultimately discards, indicating inefficient queries or missing indexes.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Transactions Per Second** shows overall database activity. The metric splits between commits (successful) and rollbacks (failed). Spikes in rollbacks usually indicate problems.
+
+## Cache Metrics (Redis)
+
+Redis serves as a fast in-memory cache to reduce database load.
```
mattwiller:
```suggestion
Redis serves as a fast in-memory cache to reduce database load and store ephemeral data.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Performance
+**CPU and Memory Usage** track Redis resource consumption. Memory usage directly correlates with stored data since Redis is an in-memory store.
+
+**Cache Hit/Miss Ratio** shows cache effectiveness. Some misses are normal, but unusual spikes might indicate application issues or invalid reference patterns.
+
+### Capacity Management
+**Evictions** occur when Redis runs out of memory and must remove least-recently-used items. Ideally this should be zero, but occasional evictions aren't necessarily problematic if your working set still fits in cache.
```
mattwiller:
```suggestion
**Evictions** occur when Redis runs out of memory and must remove least-recently-used items. Ideally this should be zero, but occasional evictions aren't necessarily problematic if all the data being operated on at one time still fits in cache.
```

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Alert Recommendation**: Monitor for evictions > 0 as an early warning of capacity issues.
+
+## Container Metrics
+
+If using containerized deployment (Docker, Kubernetes, ECS, etc.), these metrics track resource usage at the container level.
```
mattwiller: These are more accurately "host metrics", since they could be coming from a bare metal system, a VM, or a container

packages/docs/docs/self-hosting/monitoring.md
mattwiller: That sounds good to me 👍

packages/docs/docs/self-hosting/monitoring.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Memory Usage** tracks container memory consumption, useful for detecting leaks or capacity planning.
+
+**Network I/O** provides supporting information about data transfer volumes, helpful for understanding traffic patterns and response sizes.
+
+## Alerting Strategy
+
+### High-Priority Alerts
+Set up immediate alerts for metrics that directly impact users:
+- Load balancer error rates (especially 504s)
```
mattwiller: 500s are also pretty important to capture

packages/docs/docs/self-hosting/monitoring.md
maddyli:
```suggestion
- **Load balancer metrics**: Generally availability from cloud load balancers, e.g. AWS ALB
```

packages/docs/docs/self-hosting/monitoring.md
maddyli:
```suggestion
- 500 error rates, especially load balancer error rates (504s)
```

packages/docs/docs/self-hosting/monitoring.md
maddyli: Sounds fair, added metric names and linked the otel docs

packages/docs/docs/self-hosting/monitoring.md
maddyli:
```suggestion
## Container Metrics

These metrics track resource usage at the system level, whether running on bare metal servers, virtual machines, or containerized deployments (Docker, Kubernetes, ECS, etc.).
```

PR #7250, opened 2025-08-13
Add super admin UI for $explain operation

mattlong:
**The form**
Screenshot 2025-08-13 at 3 54 27 PM

**The results modal**
Screenshot 2025-08-13 at 3 54 48 PM

### DIFF
--- a/packages/app/src/admin/SuperAdminPage.tsx
+++ b/packages/app/src/admin/SuperAdminPage.tsx
@@ -63,46 +63,6 @@ export function SuperAdminPage(): JSX.Element {
     startAsyncJob(medplum, 'Reload Cron Resources', 'admin/super/reloadcron');
   }
 
-  function explainSearch(formData: Record<string, any>): void {
-    if (!formData.query) {
-      showNotification({ color: 'red', message: 'Query is required', autoClose: false });
-      return;
-    }
-    console.log(formData);
-    formData.analyze = formData.analyze ?? false;
-    const onBehalfOfHeader: string | undefined = formData['onBehalfOfProjectMembership'];
-    delete formData['onBehalfOfProjectMembership'];
-
-    const toSubmit = {
-      query: formData.query,
-      analyze: formData.analyze,
-      format: 'text',
-    };
-
-    const headers: HeadersInit = {};
-    if (onBehalfOfHeader) {
-      headers['x-medplum-on-behalf-of'] = onBehalfOfHeader;
-    }
-    console.log(headers);
-    console.log(toSubmit);
-    medplum
-      .post('fhir/R4/$explain', toSubmit, undefined, { headers })
-      .then((params: Parameters) => {
-        setModalTitle('Database Explain');
-        const explainLine = params.parameter?.find((p) => p.name === 'explain')?.valueString;
-        const queryLine = params.parameter?.find((p) => p.name === 'query')?.valueString;
-        const parametersLine = params.parameter?.find((p) => p.name === 'parameters')?.valueString;
-        const lines = [queryLine, parametersLine, '\n', explainLine].join('\n');
-        setModalContent(
-          <Code block maw={'100%'}>
-            {lines}
-          </Code>
-        );
-        open();
-      })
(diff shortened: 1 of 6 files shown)


packages/server/src/fhir/operations/explain.ts
```diff
@@ -26,6 +28,34 @@ const operation: OperationDefinition = {
       min: 1,
       max: '1',
     },
+    {
+      use: 'in',
+      name: 'analyze',
+      type: 'boolean',
+      min: 0,
+      max: '1',
+    },
```
mattwiller, line 38: Since just the plan without `ANALYZE` isn't very useful, should we consider leaving this out?

packages/server/src/fhir/operations/explain.ts
```diff
@@ -38,18 +68,62 @@ const operation: OperationDefinition = {
...
-  const output = buildOutputParameters(operation, {
-    explain: JSON.stringify(explain, (key, value) => (key.endsWith('Blocks') && value === 0 ? undefined : value), 0),
-  });
+  const result = await selectQuery.execute(client);
+
+  let output: Parameters;
+  if (params.format === 'json') {
+    const explain = result[0]['QUERY PLAN'][0];
+    output = buildOutputParameters(operation, {
+      explain: JSON.stringify(explain, (key, value) => (key.endsWith('Blocks') && value === 0 ? undefined : value), 0),
```
mattwiller: Don't we need to return the required `query` and `parameters` here?

packages/app/src/admin/SuperAdminPage.tsx
```diff
@@ -135,6 +180,68 @@ export function SuperAdminPage(): JSX.Element {
...
+  // function onBehalfOfChange(ref: Reference<Resource> | undefined): void {
+  //   console.log(ref);
+  //   setProjectMembershipProfile(ref);
+  // }
+
+  // const projectMembershipProfileRef: string | undefined =
+  //   explainProject?.reference?.startsWith('Practitioner') ||
+  //   explainProject?.reference?.startsWith('Patient')
+  //     ? explainProject?.reference
+  //     : undefined;
```
mattwiller: Remove?

packages/app/src/admin/SuperAdminPage.tsx
```diff
@@ -135,6 +180,68 @@ export function SuperAdminPage(): JSX.Element {
...
+    console.log('onOnBehalfOfProfileChange', ref);
+    setExplainProfile(ref);
+  }
+
+  const explainSearchCriteria: Record<string, string> | undefined = useMemo(() => {
+    if (!explainProject?.reference) {
+      return undefined;
+    }
+
+    // return undefined;
```
mattwiller: Cleanup

packages/app/src/admin/SuperAdminPage.tsx
```diff
@@ -135,6 +180,68 @@ export function SuperAdminPage(): JSX.Element {
     startAsyncJob(medplum, 'Reconcile Schema Diff', 'admin/super/reconcile-db-schema-drift');
   }
 
+  const [explainProject, setExplainProject] = useState<Reference<Resource> | undefined>();
+  const [explainProfile, setExplainProfile] = useState<Reference<Resource> | undefined>();
+  const [explainMemberships, setExplainMemberships] = useState<WithId<ProjectMembership>[] | undefined>();
+
+  function onBehalfOfProjectChange(ref: Reference<Resource> | undefined): void {
+    console.log('onBehalfOfProjectChange', ref);
```
mattwiller: Cleanup

packages/server/src/fhir/operations/explain.ts
mattlong: We have a case where a particular search takes a very long time to execute and generally times out, so seeing the query plan without execution would be insightful for investigating.

mattwiller: Should we consider manually changing/eliminating the timeout for the DB connection here?  Since it's a Super Admin endpoint, I think it would be good to enable getting the actual executed plan, even for super long queries like that

mattwiller (approved):

PR #7256, opened 2025-08-14
Require MFA in invite flow

codyebberson:


codyebberson:
@claude please review

codyebberson:
@claude please review.  please note that the Github security alert for "missing rate limits" is a false positive, because the express rate limit middleware is enabled at a parent level.

PR #7266, opened 2025-08-18
Fix datetime normalization for partial seconds

mattwiller:


mattlong:
For the record, previously, an input of the form `YYYY-MM-DDThh:mm:ssZ`, e.g. `1985-11-30T05:05:55Z` would be parsed incorrectly into:
```typescript
{
  "end": "1985-11-30T05:05:55Z999Z",
  "start": "1985-11-30T05:05:55Z000Z",
}
```

PR #7271, opened 2025-08-19
Add ServiceRequest.reason-code search parameter

### LINKED ISSUE #7247: Add Search support for ServiceRequest#reasonCode
A customer has requested we add search support for `ServiceRequest#reasonCode`

See: https://medplum.slack.com/archives/C078NUJP6B1/p1755104100757369

mattwiller:
Adding a non-standard search parameter for `ServiceRequest.reasonCode`.  This underlying field will change in FHIR R5, being replaced by a new `ServiceRequest.reason` field of type `CodeableReference`

Resolves #7247

### DIFF
--- a/packages/definitions/dist/fhir/r4/search-parameters-medplum.json
+++ b/packages/definitions/dist/fhir/r4/search-parameters-medplum.json
@@ -1650,6 +1650,23 @@
           "VisionPrescription"
         ]
       }
+    },
+    {
+      "fullUrl": "https://medplum.com/fhir/SearchParameter/ServiceRequest-reasonCode",
+      "resource": {
+        "resourceType": "SearchParameter",
+        "id": "ServiceRequest-reasonCode",
+        "url": "https://medplum.com/fhir/SearchParameter/ServiceRequest-reasonCode",
+        "version": "4.0.1",
+        "name": "reasonCode",
+        "status": "draft",
+        "publisher": "Medplum",
+        "description": "Coded value representing the reason for this ServiceRequest.  NOTE: The underlying field will change in FHIR R5 to ServiceRequest.reason.where(concept.exists())",
+        "code": "reason",
+        "base": ["ServiceRequest"],
+        "type": "token",
+        "expression": "ServiceRequest.reasonCode"
+      }
     }
   ]
 }
--- a/packages/definitions/dist/fhir/r4/search-parameters-medplum.json
+++ b/packages/definitions/dist/fhir/r4/search-parameters-medplum.json
@@ -1662,7 +1662,7 @@
         "status": "draft",
         "publisher": "Medplum",
         "description": "Coded value representing the reason for this ServiceRequest.  NOTE: The underlying field will change in FHIR R5 to ServiceRequest.reason.where(concept.exists())",
-        "code": "reason",
+        "code": "reasonCode",
         "base": ["ServiceRequest"],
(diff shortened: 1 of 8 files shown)


packages/definitions/dist/fhir/r4/search-parameters-medplum.json
```diff
@@ -1650,6 +1650,23 @@
...
+      "fullUrl": "https://medplum.com/fhir/SearchParameter/ServiceRequest-reasonCode",
+      "resource": {
+        "resourceType": "SearchParameter",
+        "id": "ServiceRequest-reasonCode",
+        "url": "https://medplum.com/fhir/SearchParameter/ServiceRequest-reasonCode",
+        "version": "4.0.1",
+        "name": "reasonCode",
+        "status": "draft",
+        "publisher": "Medplum",
+        "description": "Coded value representing the reason for this ServiceRequest.  NOTE: The underlying field will change in FHIR R5 to ServiceRequest.reason.where(concept.exists())",
```
mattlong: I believe we're concerned more with R6 than R5 for the next version that Medplum will eventually support, so maybe change the description to reference R6 instead?

ThatOneBro (approved):

packages/server/src/migrations/schema/v96.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Do not edit manually.
+ */
+
+import { PoolClient } from 'pg';
+import * as fns from '../migrate-functions';
+
+// prettier-ignore
+export async function run(client: PoolClient): Promise<void> {
+  const results: { name: string; durationMs: number }[] = []
+  await fns.query(client, results, `ALTER TABLE IF EXISTS "ServiceRequest" ADD COLUMN IF NOT EXISTS "__reasonCode" UUID[]`);
```
mattlong, line 14: Should we consider adding this search param to `DedicatedTokenColumnsOverridesByResourceTypeAndCode` in `packages/server/src/fhir/searchparameter.ts` with a value of `false` so it'll be served by the shared tokens columns rather than getting a dedicated column?

mattlong: I suspect we should since this is likely going to get niche usage.

mattwiller: Ah nice, I didn't know that was a thing.  Makes sense to me, I'll change it

mattlong: Discussed offline; disregard my suggestions. 15.6% of `ServiceRequest` resources have a `reasonCode`. That's more than enough to justify a dedicated column.

packages/server/src/migrations/data/data-version-manifest.json
```diff
@@ -73,5 +73,8 @@
   },
   "v21": {
     "serverVersion": "4.3.9"
+  },
+  "v22": {
+    "serverVersion": "4.3.11"
```
mattlong, line 78: Can we add a second post-deploy to reindex `ServiceRequest`?

mattlong (changes requested):

packages/server/src/migrations/data/data-version-manifest.json
mattwiller: Sure thing!  I can also have it reindex `Task`, since that was previously updated to be in the Patient compartment

PR #7277, opened 2025-08-20
Enable async response for $set-accounts operation

mattwiller:


codyebberson:
@claude please review

PR #7284, opened 2025-08-22
feat(agent): add configurable logging to agent

### LINKED ISSUE #7243: Rework Agent logging for log fetching and configurable log rotation
## Problem

While attempting to implement #6319 (`Agent/$fetch-logs`) we noted a few problems with the current logging system:
1. The Agent itself cannot be aware of the log file location since it is set up by the installer and exists outside of the agent itself. This especially would get tricky if we allowed for a configurable log file location as described in #4497
2. The Agent puts all logs of various types into one stream, meaning they all go to the same file. This includes messages logged by channels, which include PHI, which can be concerning for some end-users who want to avoid turning over logs that contain PHI to customers who are using the agent and are asking for logs to help in debugging problems with the agent.
3. A tangential but somewhat related problem is that log retention and rotation are not configurable (see #4498)

## Solution

* We can use a logging system like `winston`(https://www.npmjs.com/package/winston) or `pino` (https://www.npmjs.com/package/pino) to allow for different configurable logging channels, each with their own logging retention policy, configurable log level (`info`, `warn`, etc.), file location, etc. [Heavily leaning towards `pino` as it seems to be performance focused]
* We can have different transports for main system/ error logs and HL7 channel message logs that contain PHI. This means that `Agent/$fetch-logs` can just return the logs from the main logs and not have access to any PHI

ThatOneBro:
Fixes #7243

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -24,10 +24,8 @@ import { Agent, AgentChannel, Endpoint, Reference } from '@medplum/fhirtypes';
 import { Hl7Client } from '@medplum/hl7';
 import { ChildProcess, ExecException, ExecOptionsWithStringEncoding, exec, spawn } from 'node:child_process';
 import { existsSync, openSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs';
-import { writeFile } from 'node:fs/promises';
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Closing in favor of #7366

PR #7298, opened 2025-08-26
App Navigation UI changes

techdavidy:


codyebberson:
@claude please review

packages/react/src/AppShell/Navbar.tsx
```diff
@@ -190,9 +194,44 @@ function getActiveLink(
  */
 function getLinkScore(currentPathname: string, currentSearchParams: URLSearchParams, linkHref: string): number {
   const linkUrl = new URL(linkHref, 'https://example.com');
+
+  if (
```
codyebberson: Thanks for tackling this, but I'd like to chat about the approach here. As you know, we try to keep our core components, like `getLinkScore`, as general-purpose as possible.  The original function was designed to be a universal utility, not a place for specific business logic. These changes, while functional, introduce special cases for specific URLs like `/admin/project` and `/integrations/dosespot`. This makes the function harder to maintain, less reusable, and more prone to breaking as we add new features or change URL structures.

A better way to solve this would be to introduce a more flexible configuration for the navigation links themselves. We could add an optional property, let's say `matchType`, to the `NavbarLink` type. The `getLinkScore` function could then simply read this `matchType` and apply the appropriate logic, like `'exact'` or `'startsWith'`. This keeps the scoring function generic and pushes the business logic out to where it belongs: the configuration of our navigation menu.

Something like this:

```ts
type MatchType = 'exact' | 'startsWith';

/**
 * Calculates a score for a link based on its URL and specified match behavior.
 * @param currentPathname - The current pathname.
 * @param currentSearchParams - The current search parameters.
 * @param linkHref - A candidate link href.
 * @param matchType - The matching behavior ('exact' or 'startsWith').
 * @returns The link score.
 */
function getLinkScore(
  currentPathname: string,
  currentSearchParams: URLSearchParams,
  linkHref: string,
  matchType: MatchType = 'exact'
): number {
  const linkUrl = new URL(linkHref, 'https://example.com');

  switch (matchType) {
    case 'startsWith':
      if (!currentPathname.startsWith(linkUrl.pathname)) {
        return 0;
      }
      break;
    case 'exact':
    default:
      if (currentPathname !== linkUrl.pathname) {
        return 0;
      }
      break;
  }

  // Existing logic for matching search parameters
  const ignoredParams = ['_count', '_offset'];
  for (const [key, value] of linkUrl.searchParams.entries()) {
    if (ignoredParams.includes(key)) {
      continue;
    }
    if (currentSearchParams.get(key) !== value) {
      return 0;
    }
  }

  let count = 1;
  for (const [key, value] of currentSearchParams.entries()) {
    if (ignoredParams.includes(key)) {
      continue;
    }
    if (linkUrl.searchParams.get(key) === value) {
      count++;
    }
  }
  return count;
}
```

I'm happy to pair on it to get it right. It's a great opportunity to improve the overall design of this part of the application.

PR #7304, opened 2025-08-27
Add sort columns for HumanName-based search parameters

mattlong:


The-Alchemist:
Is it just me, or is there a missing index on `Practitioner.deleted`?  Maybe it makes sense to have a composite index for `id, deleted` instead of just `id` if the two columns are usually queried together?

mattlong:
> Is it just me, or is there a missing index on `Practitioner.deleted`? Maybe it makes sense to have a composite index for `id, deleted` instead of just `id` if the two columns are usually queried together?

Moving deleted records to a separate table, e.g. `Deleted_Patient`, seems like it would be a win all around. Since deleted records aren't searchable, it wouldn't degrade functionality at all and would simplify queries by getting rid of that `deleted = false` filter which would be a boon for the query planner. Thoughts?

Curious if @codyebberson has any thoughts on this.

mattwiller:
Since deleted records aren't generally searchable, we could consider switching to using partial indexes with `WHERE deleted = false` as well, which would also result in smaller indexes.  This migration could happen simultaneously with one prefixing most search params with `projectId`, if we wanted to feed two birds with one scone

PR #7327, opened 2025-09-03
Allow async batch processing to wait for additional FHIR quota

### LINKED ISSUE #6750: Async Batch Operations are processed as async jobs
As customer rate limits are rolled out, users will need ways to perform big batch uploads for ingesting data into the system without triggering rate limits

mattwiller:
Resolves #6750


packages/server/src/fhir/fhirquota.ts
```diff
@@ -106,7 +117,19 @@ export class FhirRateLimiter {
         msToReset: result.msBeforeNext,
         enabled: this.enabled,
       });
-      if (this.enabled) {
+      await this.block(points, result);
```
mattlong, line 120: Are there any concerns with `consumeImpl` recursively calling itself through `block`? Any risk of hitting the stack size limit in a particuarly long or unlucky operation that repeatedly gets rate limited?

packages/server/src/workers/batch.ts
```diff
@@ -83,30 +85,23 @@ async function addBatchJobData(job: BatchJobData): Promise<Job<BatchJobData>> {
...
 export async function execBatchJob(job: Job<BatchJobData>): Promise<void> {
-  const { bundle, login, project, membership } = job.data;
+  const bundle = job.data.bundle;
+  const { login, project, membership } = job.data.authState;
   const logger = getLogger();
 
   // Prepare the original submitting user's repo
   const userConfig = await getUserConfiguration(getSystemRepo(), project, membership);
-  const repo = await getRepoForLogin({ login, project, membership, userConfig });
+  const repo = await getRepoForLogin({ login, project, membership, userConfig }, true);
```
mattlong, line 104: What necessitates extended mode here?

packages/server/src/async-batch.ts
```diff
@@ -26,7 +27,12 @@ export function asyncBatchHandler(
     }
     const bundle = req.body as Bundle;
 
-    const { repo } = getAuthenticatedContext();
+    if (bundle.type === 'transaction' && project.features?.includes('transaction-bundles')) {
```
mattlong, line 30: Why does this need the feature check? Shouldn't it throw regardless? It seems like not having the `transaction-bundles` feature enabled should throw a different error?

packages/server/src/fhir/fhirquota.ts
mattwiller: It's possible, but I figured that such a case should be pretty rare.  If a batch runs for that long, it's also likely be be interrupted by a server restart or other external cause

packages/server/src/workers/batch.ts
mattwiller: Nothing especially, our code just usually implicitly assumes that repositories being used within the system are in extended mode and thought it would be better to have it than not

packages/server/src/async-batch.ts
mattwiller: If we wouldn't actually be processing the Bundle within a Postgres transaction, it would be safe to proceed

mattlong (approved):

PR #7331, opened 2025-09-04
Fix bug of forceNewMembership working only in one case (Practitioner …

thomabigapneal:
Based on this discussion here : https://discord.com/channels/905144809105260605/1412071423278710814

I've made the change to cover a maximum of cases. 
The test that was already implemented worked because it was specified for servers scoped Practitioner only. 
The test implemented here covers a whole range of use cases.


codyebberson:
@claude please review

codyebberson (approved):

codyebberson:
Thanks @thomabigapneal !

thomabigapneal:
Hi @codyebberson thank you ! 
Do you know what release the fix will be in ?

PR #7336, opened 2025-09-05
Optimized identifiers proof-of-concept

codyebberson:
Proof-of-concept for "optimized identifiers" idea we have been kicking around.

1. New optional "optimizedIdentifiers" server config setting
2. Each optimized identifier has 3 values:
   1. `resourceType`
   2. `system`
   3. `status` (either `backfill` or `active`)
3. Server automatically creates column and index if they do not exist
4. Server writes to column when available
5. Server uses column for search when `status === 'active'`

### DIFF
--- a/packages/server/src/fhir/search.ts
+++ b/packages/server/src/fhir/search.ts
@@ -981,7 +981,7 @@ function buildSearchFilterExpression(
     };
   }
 
-  const impl = getSearchParameterImplementation(resourceType, param);
+  const impl = getSearchParameterImplementation(resourceType, param, filter);
 
   if (impl.searchStrategy === 'token-column') {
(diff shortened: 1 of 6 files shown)


packages/server/src/fhir/search.ts
```diff
@@ -1257,6 +1257,9 @@ function buildTokenSearchFilter(
   operator: Operator,
   values: string[]
 ): Expression {
+  if (impl.stripSystem) {
+    values = values.map((v) => (v.includes('|') ? v.split('|')[1] : v));
```
mattlong, line 1261: This needs to consider escaped `|` characters.

codyebberson:
Closing this for now, may revisit in the future.

PR #7337, opened 2025-09-05
cleanup: removing dead biome.json

The-Alchemist:
someone correct me if i'm wrong, but i believe `biome.json` isn't used anywhere anymore


ThatOneBro:
Not exactly dead; we recommend installing Biome in VSCode for quick feedback on linting errors (namely Cody and I have done that) and so we have kept it in the repo for now. @codyebberson and I have also discussed moving to Biome fully for linting once a few rules stabilize (namely `no-floating-promises` and a few others) but the jury is still out on that one.

PR #7340, opened 2025-09-08
Dependency upgrades 2025-09-08

medplumbot:
Dependency upgrades


mattwiller:
I checked for the compromised npm packages from [today's supply chain attack](https://jdstaerk.substack.com/p/we-just-found-malicious-code-in-the); it doesn't appear that we reference any of the affected packages/versions

PR #7156, opened 2025-07-30
Add findResourceById utility function to Bundle utilities

AbhinavMir:
Adds a new utility function `findResourceById` to the core Bundle utilities that allows developers to efficiently search for a specific resource within a Bundle by its ID in the standard FHIR format (ResourceType/id).

Candidate for https://github.com/medplum/medplum/issues/4397, Previous PR closed due to inactivity. Re-opening to add tests.

PR #7346, opened 2025-09-09
feat(core): added findResourceById (#4397)

The-Alchemist:
As suggested by @Swizec .

@Swizec , could you check to see if this meets your needs/expectations?

Based on #7156 and #5719

### DIFF
--- a/packages/core/src/bundle.test.ts
+++ b/packages/core/src/bundle.test.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import { Bundle, BundleEntry, DiagnosticReport, Patient, Resource, Specimen } from '@medplum/fhirtypes';
-import { convertContainedResourcesToBundle, convertToTransactionBundle } from './bundle';
+import { convertContainedResourcesToBundle, convertToTransactionBundle, findResourceInBundle } from './bundle';
 import { getDataType } from './typeschema/types';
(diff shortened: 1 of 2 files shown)


packages/core/src/bundle.test.ts
```diff
@@ -406,4 +406,23 @@ describe('Bundle tests', () => {
...
+        name: [{ given: ['John'], family: 'Doe' }],
+      };
+      const bundle: Bundle = {
+        resourceType: 'Bundle',
+        type: 'transaction',
+        entry: [{ resource: input }],
+      };
+      const result = findResourceInBundle(bundle, 'Patient', '123');
+      expect(result).toMatchObject({
+        resourceType: 'Patient',
```
mattlong, line 494:
```suggestion
        resourceType: 'Patient',
        id: '123',
```

packages/core/src/bundle.test.ts
```diff
@@ -406,4 +406,23 @@ describe('Bundle tests', () => {
...
+    test('Simple resource', () => {
+      const input: Patient = {
+        resourceType: 'Patient',
+        id: '123',
+        name: [{ given: ['John'], family: 'Doe' }],
+      };
+      const bundle: Bundle = {
+        resourceType: 'Bundle',
+        type: 'transaction',
+        entry: [{ resource: input }],
```
mattlong: Could we add another couple items to `entry`? One with a different resource type but same `id`, and one with the same resource type by different `id`

The-Alchemist: good catch.  done.  added a few more tests for a few more edge cases as well

PR #7358, opened 2025-09-10
[v5] Node 22 and ESM everything

codyebberson:
The long `await`ed migration to ESM is finally here.

We have tried this migration multiple times over the past three years, always getting stuck behind some build tool incompatibility, dependency incompatibility, or testing incompatibility.  This year, the stars have aligned, and everything fell into place.

This is part of the Medplum v5 major version upgrade.  This also includes our annual Node.js major version bump, upgrading the minimum required version from v20 to v22.  To be specific, we are changing the minimum required version to `"engines": { "node": "^22.18.0 || >=24.2.0" }` for access to `import.meta.main`, as described below.

We will continue to dual-publish all SDK packages as both CommonJS and ESM for backwards compatibility.  There is still at least one hard requirement for CommonJS, which is Node.js [SEA (Single Executable Applications)](https://nodejs.org/api/single-executable-applications.html).

This PR also replaces `ts-node` and `ts-node-dev` with `tsx`.  This is should be a pretty seamless transition which only affects Medplum developers at dev time.  The `tsx` tool has significantly better support for ESM and also has some performance wins.

This PR also switches the `server` package build process from simply using `tsc` to using `esbuild`.  Using `tsc` alone worked ok, but it meant many small files, and more importantly it didn't work with ESM unless we changed all `import` statements to use `.js` file extensions.  More on that inline.  This also resolved all of our circular import issues, so `madge` now returns a clean bill of health.

Bots now work with both CommonJS and ESM:
* The `app` BotEditor now defaults to ESM
* If the user uploads code with `.cjs` file extension, treat as CommonJS
* If the user uploads code with `.mjs` file extension, treat as ESM
* If the code uses `export` and not `module.exports`, treat as ESM
* Otherwise default to CommonJS

-------

* All `package.json` files should use:
   * `"type": "module"` to default to ESM
* All `tsconfig.json` files should use:
   * `"target": "esnext"` to default to ESM output
   * `"module": "esnext"` to default to ESM module system
   * `"moduleResolution": "bundler"` to default to "bundler" module resolution - this one is somewhat controversial, but I think it makes the most sense for us.  The alternative would be to use `"nodenext"` which requires `.js` file extensions on all `import` statements
* Need to re-evaluate all uses of `__dirname`
   * We have 2 common cases of using `__dirname`:
      1. JSON files which should be JSON `import` statements
      2. Data files which should be a in a `static` directory
   * This may require shifting some things around, such as db migration directories
* Replace `ts-node-dev` with `tsx`
   * This was surprising to me
   * I had assumed that the order of popularity was (1) `nodemon`, (2) `ts-node-dev`, (3) `tsx`
   * But that is outdated. The real order is (1) `tsx`, (2) `nodemon`, (3) `ts-node-dev`
   * This should yield faster startup time and better future-proofing compatibility
* Start using `esbuild` to bundle `server`
   * Before, we just used `tsc` to create production output
   * That works fine, but it means many small files
   * That doesn't work directly with ESM unless we change all `import` statements to use `.js` file extensions
   * This also resolves all of our circular imports, `madge` now returns clean report
   * So, using `esbuild` to bundle the server kills 3 birds with 1 stone

--------

Comparing `ts-node-dev` with `tsx`:

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,7 +158,7 @@ jobs:
     needs: [build]
     strategy:
       matrix:
-        node-version: [22, 22]
+        node-version: [22, 24]
         pg-version: [14, 18]
         redis-version: [7]
     env:
@@ -260,7 +260,7 @@ jobs:
     needs: [build]
     strategy:
       matrix:
-        node-version: [22, 22]
+        node-version: [22, 24]
         pg-version: [14, 18]
         redis-version: [7]
     env:


packages/server/src/migrations/migrate-main.ts
```diff
@@ -0,0 +1,100 @@
...
+import { FileBuilder, isMain } from '@medplum/core';
+import { readdirSync, writeFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { Client } from 'pg';
+import type { BuildMigrationOptions } from './migrate';
+import { buildMigration, buildSchema, indexStructureDefinitionsAndSearchParameters } from './migrate';
+
+export const SCHEMA_DIR = resolve('./src/migrations/schema');
+
+export async function main(): Promise<void> {
```
codyebberson, line 12: Explanation why this moved to a separate file:

1. For `server` to be ESM compatible, we had to make a choice
   a. Use ESM compatible `import` statements, which would mean adding `.js` suffixes to all imports.
   b. Start using a bundler like `esbuild` to package everything into a single bundled `.js` file
   c. In this PR, I opted for the `esbuild` bundling option
2. For ESM compatible code, we swapped out our checks for `require.main === module` with the new `isMain(import.meta)`
   a. However, the behavior of that function doesn't work if it gets bundled, because `esbuild` doesn't have any awareness of `main` semantics, so instances of `isMain()` will get bundled into the output

Because we import `migrate.ts` in server, the old `main()` method and `isMain()` checks were included in the bundled output.

By moving the `main` entry point to a separate file, we can keep the clear separation.

packages/server/src/migrations/migration-utils.ts
```diff
@@ -88,7 +88,7 @@ export function getPostDeployManifestEntry(migrationNumber: number): {
   requiredBefore: string | undefined;
 } {
   const manifest = JSON.parse(
-    readFileSync(resolve(__dirname, 'data/data-version-manifest.json'), { encoding: 'utf-8' })
+    readFileSync(resolve('./src/migrations/data/data-version-manifest.json'), { encoding: 'utf-8' })
```
codyebberson: TODO: This needs to move to a `data` directory outside of `src` and `dist`

codyebberson:
@claude - please review this draft PR. please ignore the open todo items, at this point I'm more interested in architecture and strategy feedback. after we have consensus on the strategy, then we can focus on cleanup and polish.

codyebberson:
Deployed and running on staging: https://app.staging.medplum.dev/

packages/server/package.json
```diff
@@ -13,15 +13,16 @@
   },
   "license": "Apache-2.0",
   "author": "Medplum <hello@medplum.com>",
+  "type": "module",
```
mattlong, line 16: 🔥

packages/server/src/otel/instrumentation.ts
```diff
@@ -130,7 +130,7 @@ export async function shutdownOpenTelemetry(): Promise<void> {
   }
 }
 
-if (require.main === undefined) {
+if (!('main' in import.meta)) {
```
mattlong: Why not `!import.meta.main`?

packages/server/package.json
```diff
@@ -13,15 +13,16 @@
...
   "scripts": {
-    "build": "npm run clean && tsc && tsc --project tsconfig.build.json",
+    "build": "npm run clean && tsc && node build.mjs",
     "circular": "npx madge --circular dist/index.js",
     "clean": "rimraf dist",
-    "dev": "ts-node-dev --poll --respawn --transpile-only --require ./src/otel/instrumentation.ts src/index.ts",
+    "dev": "tsx watch src/index.ts",
```
mattlong: Nothing needed for `otel/instrumentation` here?

codyebberson: Good catch - will fix this.

packages/server/package.json
codyebberson: its-happening

PR #7365, opened 2025-09-10
Docs update: Fhir quota weights no longer in testing phase

finnbergquist:


mattwiller:
Should the `warning` box also be removed now?

PR #7366, opened 2025-09-10
feat(agent): add configurable logging system using `winston`

### LINKED ISSUE #4497: Agent allow for configurable log directory
The Agent logs are currently stored within the application directory. As these logs roll over after 2MB, they can clutter this directory. It has been requested to allow for the log (current and archived) to go to a specific location on the host. This destination could be another folder inside the application directory (which is not removed if the Agent is removed), a different absolute directory on the same host environment or a network share (using a Windows mapped drive). This should be either supportable during the installer process (maybe as a selective checkbox that launches a screen after clicking Next?), allowing to select a location or configurable post-installation via the agent.config. Not selecting a different location (or even showing that screen?) should use the default application installation directory for the logs.

Any upgrade to a current Agent installation, should only allow for a change to the agent.config and not prompt for a different destination.

### LINKED ISSUE #4498: Agent log retention default is too low
The Agent installation is using a default log retention of 2, this value is far too small of most heavy interfaces. This value should be configurable within the agent.properties to allow a larger value by the system admins running the host operating system. By default I would bump this value to at least 20, but that change should be coupled with #4497

### LINKED ISSUE #7243: Rework Agent logging for log fetching and configurable log rotation

ThatOneBro:
Before we relied on `shawl` to read logs from stdout and manage log files, however there is a long-standing request to make log file locations configurable (see: #4497), be able to tweak how often they rotate, how large the files can be, and how many log files to retain (see: #4498).

For log fetching, we also needed to address having separate log files for system and channel message logs so that system logs can be retrieved by an admin without risk of touching HL7 messages themselves. This adds a layer of protection around PHI stored in logs embedded in the relevant HL7 messages.

* Closes #4497
* Closes #4498
* Closes #7243 

### TODO:
-  [ ] Test rotate frequency


packages/agent/src/agent-main.test.ts
```diff
@@ -163,4 +164,127 @@ describe('Main', () => {
...
+
+    const mockChannelLogger = {
+      warn: jest.fn(),
+      debug: jest.fn(),
+      info: jest.fn(),
+      error: jest.fn(),
+      log: jest.fn(),
+      clone: jest.fn(),
+      level: LogLevel.INFO,
+    };
```
mattwiller: This happens a few times in this file, might be worth extracting into a little helper function, e.g.:
```ts
function mockLogger(level: LogLevel = LogLevel.INFO): ILogger {
  return {
    warn: jest.fn(),
    debug: jest.fn(),
    info: jest.fn(),
    error: jest.fn(),
    log: jest.fn(),
    clone: jest.fn(),
    level,
  };
}
```

packages/agent/src/dicom.ts
```diff
@@ -139,7 +141,8 @@ export class AgentDicomChannel extends BaseChannel {
           });
           response.setStatus(dimse.constants.Status.Success);
         } catch (err) {
-          DcmjsDimseScp.channel.log.error(`DICOM error: ${normalizeErrorString(err)}`);
+          DcmjsDimseScp.channel.log.error(`DICOM error - check channel logs`);
+          DcmjsDimseScp.channel.channelLog.error(`DICOM error: ${normalizeErrorString(err)}`);
```
mattwiller, line 145: 👍

packages/agent/src/logger.ts
```diff
@@ -0,0 +1,332 @@
...
+import { AgentArgs } from './types';
+
+export const LoggerType = {
+  MAIN: 'main',
+  CHANNEL: 'channel',
+} as const;
+export type LoggerType = (typeof LoggerType)[keyof typeof LoggerType];
+
+export const DEFAULT_LOGGER_CONFIG = {
+  logDir: __dirname,
```
mattwiller, line 26: Is this going to play nicely with the ESM stuff happening contemporaneously?

packages/agent/src/logger.ts
```diff
@@ -0,0 +1,332 @@
...
+  filesToKeep: number;
+  logLevel: LogLevel;
+}
+
+export interface FullAgentLoggerConfig {
+  main: AgentLoggerConfig;
+  channel: AgentLoggerConfig;
+}
+
+export interface PartialFullAgentLoggerConfig {
```
mattwiller: This name 🙃

packages/agent/src/logger.ts
ThatOneBro: I think we may have to hold off on ESM for the agent for some time:
https://github.com/nodejs/node/issues/53565 😓

packages/agent/src/logger.ts
ThatOneBro: Yeah it's been irking me too... Going to change ontology to:
`FullAgentLoggerConfig` -> `AgentLoggerConfig`
`AgentLoggerConfig` -> `AgentLoggerConfigEntry`

ThatOneBro: Ended up going with `AgentLoggerConfig` and `AgentMultiLoggerConfig`

PR #7370, opened 2025-09-11
fix: CCDA to FHIR converter attribute mapping

jdjkelly:
Remove incorrect `@_use` attribute in address and telecom mappings - should just be `use`

### DIFF
--- a/packages/ccda/src/ccda-to-fhir.ts
+++ b/packages/ccda/src/ccda-to-fhir.ts
@@ -276,7 +276,7 @@ class CcdaToFhirConverter {
       return undefined;
     }
     return addresses?.map((addr) => ({
-      '@_use': addr['@_use'] ? ADDRESS_USE_MAPPER.mapCcdaToFhir(addr['@_use']) : undefined,
+      use: addr['@_use'] ? ADDRESS_USE_MAPPER.mapCcdaToFhir(addr['@_use']) : undefined,
       line: addr.streetAddressLine,
       city: addr.city,
       state: addr.state,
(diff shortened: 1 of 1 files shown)


codyebberson:
Hey @jdjkelly  - Thanks for looking into this.

Are you using `fast-xml-parser`?

The `@_` prefix is the `fast-xml-parser` syntax for converting a JavaScript property to an XML attribute (versus a nested XML element).

See docs here: https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/3.XMLBuilder.md

You can see example usage here: https://github.com/medplum/medplum/blob/main/packages/ccda/src/xml.ts#L82-L93

I believe the current code is correct as-is for use with `fast-xml-parser`.  If you're using a different parser, then we may need to make this configurable, so you can use a different prefix, or omit the prefix entirely.

jdjkelly:
> Hey @jdjkelly - Thanks for looking into this.
> 
> Are you using `fast-xml-parser`?
> 
> The `@_` prefix is the `fast-xml-parser` syntax for converting a JavaScript property to an XML attribute (versus a nested XML element).
> 
> See docs here: https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/3.XMLBuilder.md
> 
> You can see example usage here: https://github.com/medplum/medplum/blob/main/packages/ccda/src/xml.ts#L82-L93
> 
> I believe the current code is correct as-is for use with `fast-xml-parser`. If you're using a different parser, then we may need to make this configurable, so you can use a different prefix, or omit the prefix entirely.

@codyebberson  This impacts `convertCcdaToFhir` - the Bundle has resources with a field of `@_use` in non-valid ways, for example:

```json
{
      "resource": {
        "resourceType": "Organization",
        "id": "5134412c-d18a-41a8-a813-a82fc85f91fe",
        "telecom": [
          {
            "@_use": "work",
            "system": "phone",
            "value": " 555-555-1002"
          }
        ]
      }
    }
```

codyebberson:
Eek!  My bad, I was reversed.  I thought I was looking at FHIR-to-CCDA.  You're absolutely right.

PR #7374, opened 2025-09-11
Update `getResponseItemByContext` to support repeatable group answers in `QuestionnaireForm`

### LINKED ISSUE #7373: `QuestionnaireForm` only saves last answer for repeatable groups
## Description

There is a bug in the `QuestionnaireForm` when handling repeatable groups. When a `Questionnaire` with repeatable groups is submitted, the generated `QuestionnaireResponse` does not correctly capture multiple answers. Although multiple group instances are created in the response, only the last submitted answer contains data, while the others remain empty.

### Root cause

The `getResponseItemByContext` function in `useQuestionnaireForm.ts `only uses linkId to locate response items. For repeatable groups, multiple instances share the same `linkId`, which causes overwriting.

### Proposed solution

Update `getResponseItemByContext` to use the unique `id` field when available, falling back to `linkId` otherwise.

## Steps to test

1. Create a `Questionnaire` with a repeatable group.
2. Submit the `Questionnaire` with at least two answers in the repeatable group.
3. Inspect the `QuestionnaireResponse` → multiple entries are created, but only one contains data, and the others remain empty.

https://github.com/user-attachments/assets/396536a2-dd40-4e7a-80df-d2db987c3a43

pamella:
## Description

- Resolves #7373

## Steps to test

- Added a regression test to confirm repeated group responses are saved correctly and to prevent future regressions.

[Screencast from 11-09-2025 18:12:38.webm](https://github.com/user-attachments/assets/6d373b2a-b914-42d3-8443-618ca3ec92e0)

### DIFF
--- a/packages/react-hooks/src/useQuestionnaireForm/useQuestionnaireForm.ts
+++ b/packages/react-hooks/src/useQuestionnaireForm/useQuestionnaireForm.ts
@@ -175,7 +175,9 @@ export function useQuestionnaireForm(props: UseQuestionnaireFormProps): Readonly
     let currentItem: QuestionnaireResponse | QuestionnaireResponseItem | undefined =
       state.current.questionnaireResponse;
     for (const contextElement of context) {
-      currentItem = currentItem?.item?.find((i) => i.linkId === contextElement.linkId);
+      currentItem = currentItem?.item?.find((i) =>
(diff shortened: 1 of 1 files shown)


techdavidy:
@claude please review

techdavidy:
@claude please review

PR #7377, opened 2025-09-11
Add GIN index stats page

mattlong:
This is a precursor to adding functionality to dynamically enable/disable `fastupdate` for GIN indexes.


packages/server/src/fhir/operations/dbgetginindexes.ts
```diff
@@ -0,0 +1,185 @@
...
+      min: 1,
+      max: '1',
+    },
+    {
+      use: 'out',
+      name: 'index',
+      min: 0,
+      max: '*',
+      part: [
+        { use: 'out', name: 'tableName', type: 'string', min: 1, max: '1' },
```
mattlong: It was mentioned in a [previous PR](https://github.com/medplum/medplum/pull/6523) that`Parameters.parameter.part.use` is not necessary since it inherits from the outermost parent, but excluding it causes a Typescript error. Looking at https://hl7.org/fhir/R4/operationdefinition.html, my read is that it is actually required here to be compliant since it's a recursive definition.

packages/server/src/migrations/schema/v98.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Do not edit manually.
+ */
+
+import { PoolClient } from 'pg';
+import * as fns from '../migrate-functions';
+
+// prettier-ignore
+export async function run(client: PoolClient): Promise<void> {
+  const results: { name: string; durationMs: number }[] = []
+  await fns.query(client, results, `CREATE EXTENSION IF NOT EXISTS pageinspect`);
```
mattlong: `pageinspect` is available in [AWS RDS](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html), [GCP](https://cloud.google.com/sql/docs/postgres/extensions#pageinspect), and [Azure](https://learn.microsoft.com/en-us/azure/postgresql/extensions/concepts-extensions-versions#pageinspect).

mattlong: Ugh, but [not available on Aurora](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraPostgreSQLReleaseNotes/AuroraPostgreSQL.Extensions.html)!

PR #7378, opened 2025-09-12
feat(agent): add `$fetch-logs` operation

### LINKED ISSUE #6319: Add `Agent/$fetch-logs` operation
This feature will allow agent users to retrieve the last N log lines in the agent's log file, which will be useful for debugging problems which otherwise require on-site presence or remote access of the machine.

Considerations:
- PHI in HL7v2 messages should probably be redacted. Meaning we scrub most segments except maybe the MSH  segment?

ThatOneBro:
Follow-up from #7366 

Fixes #6319


packages/server/src/fhir/operations/agentfetchlogs.test.ts
```diff
@@ -0,0 +1,137 @@
...
+import {
+  MockAgentResponseHandle,
+  cleanupMockAgents,
+  configMockAgents,
+  mockAgentResponse,
+} from './utils/agenttestutils';
+
+const NUM_DEFAULT_AGENTS = 2;
+
+describe('Agent/$fetch-logs', () => {
```
mattlong, line 22: Might be worth adding a test for the `Agent/[id]/$fetch-logs` endpoint as well

packages/agent/src/app.ts
```diff
@@ -829,6 +834,36 @@ export class App {
...
+      await this.sendToWebSocket({
+        type: 'agent:error',
+        body: errMsg,
+        callback: command.callback,
+      });
+      return;
+    }
+
+    try {
+      const logs = await this.log.fetchLogs({ limit: command.limit });
```
mattlong, line 850: Does any min/max bounds checking need to happen on the `limit` we send to Winston? Or any error code/message translation for expected failure modes? e.g. sending a negative value, non-integer `limit`, or something else I'm not thinking of?

codyebberson:
@claude please review

codyebberson (approved):

PR #7380, opened 2025-09-12
Update default Aurora PostgreSQL version to 16.9

paddyrehill:
The [self host install guide for AWS ](https://www.medplum.com/docs/self-hosting/install-on-aws) don't include a guide or steps to include the Postgres engine version key in CDK config file and so the default value of 12.9 is used.

[Version 12.9 is now deprecated](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-versions.html#postgresql-versions-version12) and incurs significant Amazon RDS Extended Support costs which are typically 100% of the hosting cost itself.


codyebberson:
Thanks @paddyrehill !

PR #7390, opened 2025-09-15
Dependency upgrades 2025-09-15

medplumbot:
Dependency upgrades


ThatOneBro:
Looks like there was a major version bump on `eslint-plugin-jsdoc`, which broke just two small things for us on two new included rules:

1. [require-throws-type](https://github.com/gajus/eslint-plugin-jsdoc/blob/HEAD/docs/rules/require-throws-type.md#readme)
2. [require-yields-type](https://github.com/gajus/eslint-plugin-jsdoc/blob/HEAD/docs/rules/require-yields-type.md#readme)

I just turned both of these rules off since we are already using TS and have so few instances of both of these JSDoc annotations

techdavidy (approved):

PR #7395, opened 2025-09-16
Fix validation regex for StructureDefinition.differential

### LINKED ISSUE #7391: Cannot add StructureDefinition. Problem with FHIRPath replaceMatches()
Related: https://github.com/medplum/medplum/pull/6760

## Problem
When I try to add [this StructureDefinition](http://fhir.de/StructureDefinition/informationrecipient) to medplum, I get an error. The same error happens with many StructureDefinitions inside the [de.basisprofil.r4](https://simplifier.net/packages/de.basisprofil.r4/1.5.4) package that use differentials.

`Constraint sdf-8a not met: In any differential, all the elements must start with the StructureDefinition's specified type for non-logical models, or with the same type name for logical models ({"fhirpath":"(%resource.kind = 'logical' or element.first().path.startsWith(%resource.type)) and (element.tail().empty() or element.tail().all(path.startsWith(%resource.differential.element.first().path.replaceMatches('\\\\..*','')&'.')))"}) (StructureDefinition.differential)`

Image

## Deep Dive
After doing some research, I found the following: The FHIRPath in [sdf-8a](https://hl7.org/fhir/R4/structuredefinition-definitions.html) uses `replaceMatches('\\..*','')` to turn e.g. `"Extension.url"` into `"Extension"`. This transformation however fails in the current @medplum/core, because it internally creates the regex `/\\..*/` instead of `/\..*/`.

## Try it out
```ts
test('Test replaceMatches()', () => {
  expect(evalFhirPath(`foo.replaceMatches('\\\\..*','')`, { foo: 'Extension.url' })).toStrictEqual(['Extension']); // ❌ Fails currently
  //                                         ^ Why 4 backslashes?
  // Because the FHIRPath expression for sdf-8a is literally \\..*
  // That's why we need to escape them in the javascript string
  // Instead of '\\\\..*' we could also write String.raw`\\..*`
});
```

## Solution(?)
My first thought would be to manually remove the "over-escaped" backslashes. But I'm not sure if that's the right place to make the change. It could also be a bug upstream, and the [source file](https://simplifier.net/packages/hl7.fhir.r4.core/4.0.1/files/2831303/~xml) should actually say `\..*` instead of `\\..*`. But someone else should be the judge of that.

https://github.com/medplum/medplum/blob/5db6a9472dd897cd8c5fe3d9e355577042dba314/packages/core/src/fhirpath/functions.ts#L1301

```diff
-      (str, pattern, substition) => str.replaceAll(new RegExp(pattern, 'g'), substition)
+      (str, pattern, substition) => str.replaceAll(new RegExp(pattern.replaceAll('\\\\', '\\'), 'g'), substition)
```

mattwiller:
Resolves #7391

### DIFF
--- a/packages/core/src/fhirpath/functions.ts
+++ b/packages/core/src/fhirpath/functions.ts
@@ -1887,7 +1887,9 @@ function applyStringFunc<T>(
   if (typeof value !== 'string') {
     throw new Error('String function cannot be called with non-string');
   }
-  const result = func(value, ...argsAtoms.map((atom) => atom?.eval(context, input)[0]?.value));
+
+  const args = argsAtoms.map((atom) => atom?.eval(context, input)[0]?.value);
+  const result = func(value, ...args);
   if (result === undefined) {
     return [];
   }
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -316,6 +316,81 @@ describe('FHIR resource validation', () => {
     expect(() => validateResource(medplumBundle)).not.toThrow();
   });
 
+  test('StructureDefinition with differential', () => {
+    const sd: StructureDefinition = {
+      resourceType: 'StructureDefinition',
+      url: 'http://fhir.de/StructureDefinition/informationrecipient',
+      name: 'ExtensionInformationRecipient',
+      status: 'active',
+      kind: 'complex-type',
+      abstract: false,
+      context: [
+        {
+          type: 'element',
+          expression: 'Composition',
+        },
+      ],
+      type: 'Extension',
+      baseDefinition: 'http://hl7.org/fhir/StructureDefinition/Extension',
+      derivation: 'constraint',
+      differential: {
+        element: [
+          {
+            id: 'Extension.extension',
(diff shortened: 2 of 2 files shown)


packages/definitions/dist/fhir/r4/profiles-resources.json
```diff
@@ -337822,7 +337822,7 @@
             "key" : "sdf-8a",
             "severity" : "error",
             "human" : "In any differential, all the elements must start with the StructureDefinition's specified type for non-logical models, or with the same type name for logical models",
-            "expression" : "(%resource.kind = 'logical' or element.first().path.startsWith(%resource.type)) and (element.tail().empty() or element.tail().all(path.startsWith(%resource.differential.element.first().path.replaceMatches('\\\\..*','')&'.')))",
+            "expression" : "(%resource.kind = 'logical' or element.first().path.startsWith(%resource.type)) and (element.tail().empty() or element.tail().all(path.startsWith(%resource.differential.element.first().path.replaceMatches('\\..*','')&'.')))",
```
ThatOneBro, line 337825: 👍

ThatOneBro (approved):

PR #7413, opened 2025-09-19
Refactor FHIRPath string handling

fardolieri:
See issue: https://github.com/medplum/medplum/issues/7391

Initially, I thought the fix would only require adjusting the string-to-regex behavior in the `replaceMatches` function. However, I discovered that the FHIRPath string parsing itself may not handle escape sequences correctly.

This PR highlights those issues and provides a possible solution. Consider it a starting point for discussion rather than a final implementation.

@codyebberson I had to adjust some tokenizer tests. Could you please check if I might have broken something there?

FYI, if I'm slow to respond, it's because I just left for vacation. I'll be back in a week.

### DIFF
--- a/packages/core/src/fhirlexer/tokenize.ts
+++ b/packages/core/src/fhirlexer/tokenize.ts
@@ -177,7 +177,7 @@ export class Tokenizer {
       }
       if (char2 === 'u') {
         this.advance();
-        const hex = this.str.substring(this.pos.index, this.pos.index + 4).match(/^[0-9a-fA-F]{4}$/)?.[0];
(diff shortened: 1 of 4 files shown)


fardolieri:
I force-pushed to add the `--signoff` to my commits

codyebberson:
Hi @fardolieri  - this looks good to me.  You can see Sonar and Claude analysis here: https://github.com/medplum/medplum/pull/7414

@mattwiller  is going to take one extra look, just to double check.

mattwiller (approved):
Looks great, thanks for the deep investigation and follow-up PR @fardolieri!  We should also revert the other changes I made in #7397 alongside this change to the escape code handling

fardolieri:
Awesome, thanks to you both for your hard work!

PR #7418, opened 2025-09-19
$db-configure-indexes uses AsyncJob

mattlong:
Since it can call `VACUUM` which can easily take more than a few seconds.

### DIFF
--- a/packages/server/src/fhir/operations/db-configure-indexes.ts
+++ b/packages/server/src/fhir/operations/db-configure-indexes.ts
@@ -1,12 +1,16 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { allOk, badRequest, OperationOutcomeError } from '@medplum/core';
+import { accepted, badRequest, concatUrls, OperationOutcomeError } from '@medplum/core';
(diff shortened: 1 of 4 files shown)


packages/server/src/fhir/operations/db-configure-indexes.ts
```diff
@@ -83,6 +87,9 @@ type OutputAction = {
 
 export async function dbConfigureIndexesHandler(req: FhirRequest): Promise<FhirResponse> {
   requireSuperAdmin();
+  if (req.headers?.['prefer'] !== 'respond-async') {
```
ThatOneBro, line 90: Small nit, do we also want to export `requireAsync` from `super.ts` and use it here like `requireSuperAdmin`?

mattlong: Yeah, I wanted to do that, but `requireAsync` takes a `req: Request` whereas this operation is a `FhirRequest`. I could always just make that function accept either? It would change the implementation slightly, but that's probably okay?

```ts
function requireAsync(req: Request | FhirRequest): void {
  if (req.headers?.['prefer'] !== 'respond-async') {
    throw new OperationOutcomeError(badRequest('Operation requires "Prefer: respond-async"'));
  }
}
```

ThatOneBro: Ah yeah not a big deal then 👍

PR #7432, opened 2025-09-20
Fixes #7428 - added 'next' cursor to GraphQL Connection API

### LINKED ISSUE #7428: Add "next" links to GraphQL Connection API
When searching for many resources via GraphQL Connection API, there should be `next`, `last`, `first`, `previous`, etc links in the `Connection` object.

Those links are currently missing.

Spec: https://hl7.org/fhir/R4/graphql.html#searching

Implementation: https://github.com/medplum/medplum/blob/main/packages/fhir-router/src/graphql/graphql.ts#L309-L346

Note that we're not setting `next`.

The good news is that the `Bundle` already has all of the necessary details:

```json
    "link": [
        {
            "relation": "self",
            "url": "https://api.medplum.com/fhir/R4/Observation?_count=20&_sort=_lastUpdated&_total=accurate"
        },
        {
            "relation": "first",
            "url": "https://api.medplum.com/fhir/R4/Observation?_count=20&_sort=_lastUpdated&_total=accurate"
        },
        {
            "relation": "next",
            "url": "https://api.medplum.com/fhir/R4/Observation?_count=20&_cursor=2-1742910550089&_sort=_lastUpdated&_total=accurate"
        }
    ]
```

The task is to extract the `_cursor` param and return it as the `next` value.

codyebberson:


codyebberson:
@claude please review

One thing to note is that this PR only adds `next` cursor values, not `first`, `prev`, and `last`.  The reason is that Medplum's cursor-based search only supports forward pagination under narrow conditions of sorting by `_lastUpdated` ascending.  This is intended for "big" pagination (i.e., going past the first 10,000 results).  For "small" pagination, you can pass in `_offset` for simple index based pagination.

ThatOneBro (approved):

PR #7441, opened 2025-09-22
Add fhirpath fuzz testing

codyebberson:
This PR introduces fuzz testing as a new defensive layer to our testing suite. It adds the [`@fast-check/jest`](https://www.npmjs.com/package/@fast-check/jest) library and several preliminary property-based tests for our FHIRPath parser.

#### Why Fuzz Testing?

Our current unit tests are excellent for verifying that the parser works as expected for specific, pre-defined inputs (e.g., `'3 + 4'`, `'Patient.name.given'`). This is a critical first step. However, they can only test the scenarios we have explicitly thought of.

Fuzz testing takes a different approach. Instead of testing for a specific example, we test for a property that should hold true for all inputs. For instance, we can test the property that "our parser should never crash for a well-formed arithmetic expression, no matter how complex."

The `@fast-check/jest` library automates this by intelligently generating thousands of diverse and often unexpected inputs, including edge cases like deeply nested parentheses, empty strings, or unusual number combinations. This is especially effective at finding bugs we didn't know existed.

#### Goal of this PR

This PR's primary goal is not to achieve 100% fuzz test coverage but to integrate the tool and establish a new pattern for our testing philosophy. It's about putting this powerful tool into our toolbox so we can build on it in the future.

Think of it as our first step into a new way of thinking about quality:

* Proactive Defense: Instead of just fixing bugs as they appear, we can use fuzz tests to find potential issues before they become problems in production.
* Regression Prevention: Going forward, when we encounter a bug caused by an unexpected input (e.g., a strange string format), the new best practice will be to add a fuzz test that covers this scenario. This ensures the bug is fixed and prevented from ever happening again.

The included example tests demonstrate how to:

1. Generate random, valid arithmetic expressions.
2. Generate diverse FHIR-like resources.
3. Combine these arbitraries to test complex interactions between FHIRPath expressions and FHIR data.

This is a step forward in making our FHIRPath library more robust, resilient, and reliable. I believe this will become an invaluable part of our development process.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@babel/preset-env": "7.28.3",
         "@babel/preset-react": "7.27.1",
         "@babel/preset-typescript": "7.27.1",
+        "@fast-check/jest": "2.1.1",
         "@microsoft/api-documenter": "7.26.34",
         "@microsoft/api-extractor": "7.52.13",
         "@types/node": "20.19.14",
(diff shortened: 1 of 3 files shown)


codyebberson:
@claude please review

mattwiller (approved):

PR #7449, opened 2025-09-23
Replace flushdb with a less destructive alternative

### LINKED ISSUE #7446: Consider alternatives to `flushdb` in server tests
Right now there are three test suites that intentionally use `flushdb` to get back to a clean cache state between tests. They are segregated by all other test suites by manually specifying `config.redis.db = 6` while the default test config uses `config.redis.db = 7`. This generally work well enough, but due to a subtle configuration bug, one of those test files was unintentionally flushing db 7 instead of 6.

A more maintainable alternative could be using an alternative to `flushdb` like the following. It would work well with the `config.redis.keyPrefix` provided by `ioredis`.

```
// Instead of: await getRedis().flushdb()
// Use pattern-based deletion:
const redis = getRedis();
const keys = await redis.keys('test-specific-prefix:*');
if (keys.length > 0) {
  await redis.del(...keys);
}
```

mattlong:
I don't love the introduction of `TestRedisConfig` to allow the usage of `keyPrefix`, but I more strongly don't think it should be added to `MedplumRedisConfig`...

Fixes #7446

### DIFF
--- a/packages/server/src/redis.test.ts
+++ b/packages/server/src/redis.test.ts
@@ -1,9 +1,11 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
+import { randomUUID } from 'crypto';
 import { Redis } from 'ioredis';
 import { loadTestConfig } from './config/loader';
 import type { MedplumServerConfig } from './config/types';
 import { closeRedis, getRedis, getRedisSubscriber, getRedisSubscriberCount, initRedis } from './redis';
+import { deleteRedisKeys } from './test.setup';
 
 describe('Redis', () => {
   let config: MedplumServerConfig;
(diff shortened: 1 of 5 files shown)


packages/server/src/redis.test.ts
```diff
@@ -88,4 +90,56 @@ describe('Redis', () => {
       await closeRedis();
     });
   });
+
+  // `deleteRedisKeys` is for tests only (i.e. in test.setup.ts), so testing it is a bit meta,
+  // but the function is just complicated enough that it testing is merited
```
The-Alchemist: minor grammar nitpick: `complicated enough that testing it is merited`

The-Alchemist (changes requested):

ThatOneBro (approved):

The-Alchemist (approved):

PR #7456, opened 2025-09-23
Improved `CREATE TABLE` migration generation

mattlong:
This also submarines in some exploratory code for generating the entire Medplum schema as a single .sql file that is disabled by default.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -26,8 +26,7 @@ sonar.exclusions=**/node_modules/**,\
   packages/mock/src/**/*.json,\
   packages/mock/src/test-resources/*.ts,\
   packages/react/src/stories/*,\
-  packages/server/src/migrations/data/**/*.ts,\
-  packages/server/src/migrations/schema/**/*.ts
+  packages/server/src/migrations/**/*.ts
 sonar.test.inclusions=**/*.test.ts,**/*.test.tsx
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
 sonar.links.homepage=https://www.medplum.com/


packages/server/src/migrations/migrate.ts
```diff
@@ -1145,11 +1234,15 @@ function generateAlterColumnActions(
   return actions;
 }
 
-function getCreateTableQueries(tableDef: TableDefinition): string[] {
+function getCreateTableQueries(tableDef: TableDefinition, options: { includeIfExists: boolean }): string[] {
   const queries: string[] = [];
   const createTableLines = [];
   for (const column of tableDef.columns) {
-    createTableLines.push(`  "${column.name}" ${column.type}`);
+    const typePart = column.serial ? column.type.replace('INT', 'SERIAL') : column.type;
```
mattwiller: Is this correct for all types/would it be better to have an explicit mapping here?

mattlong: Good question! The docs come to the rescue! Here's the [relevant excerpt](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL):

The data types `smallserial`, `serial` and `bigserial` are not true types, but merely a notational convenience for creating unique identifier columns (similar to the `AUTO_INCREMENT` property supported by some other databases). In the current implementation, specifying:

```sql
CREATE TABLE tablename (
    colname SERIAL
);
```

is equivalent to specifying:

```sql
CREATE SEQUENCE tablename_colname_seq AS integer;
CREATE TABLE tablename (
    colname integer NOT NULL DEFAULT nextval('tablename_colname_seq')
);
ALTER SEQUENCE tablename_colname_seq OWNED BY tablename.colname;
```

_End of excerpt_

A couple more tests:

```
postgres=# create table foo (id1 int, id2 serial);
CREATE TABLE
postgres=# alter table foo add column id3 serial;
ALTER TABLE
postgres=# \d+ foo
                                                       Table "public.foo"
 Column |  Type   | Collation | Nullable |             Default              | Storage | Compression | Stats target | Description
--------+---------+-----------+----------+----------------------------------+---------+-------------+--------------+-------------
 id1    | integer |           |          |                                  | plain   |             |              |
 id2    | integer |           | not null | nextval('foo_id2_seq'::regclass) | plain   |             |              |
 id3    | integer |           | not null | nextval('foo_id3_seq'::regclass) | plain   |             |              |
```

So yes, we could treat them as first class citizens, but when we introspect the state of the current database, we'd have to de-sugar the serial type definitions anyway to be able to compare. I'll take a stab at that...

PR #7469, opened 2025-09-24
Implement ConceptMap/$import

### LINKED ISSUE #7467: Import ConceptMap data that does not fit within resource JSON
To facilitate use of large concept mapping data sets that are available (e.g. SNOMED-to-ICD or the constituents of RxNorm), we will need to import large amounts of data — on the order of hundreds of thousands of mappings — that cannot fit in a single `ConceptMap` resource due to JSON size constraints.  To accommodate this, we will implement a bespoke `ConceptMap/$import` operation to populate lookup tables in the database, similar to the [existing solution](https://github.com/medplum/medplum/blob/57a0407d4677e36a70f2d3f51fde4e2c1f5aa2da/packages/server/src/fhir/operations/codesystemimport.ts) for `CodeSystem`

For more information see #7174

mattwiller:
Resolves #7467


packages/server/src/migrations/schema/v98.ts
```diff
@@ -0,0 +1,36 @@
...
+    "targetCode" TEXT NOT NULL,
+    "relationship" TEXT,
+    "sourceDisplay" TEXT,
+    "targetDisplay" TEXT,
+    "comment" TEXT
+  )`);
+  await fns.query(client, results, `CREATE UNIQUE INDEX IF NOT EXISTS "ConceptMapping_map_source_target_idx" ON "ConceptMapping" ("conceptMap", "sourceSystem", "sourceCode", "targetSystem", "targetCode")`);
+  await fns.query(client, results, `CREATE INDEX IF NOT EXISTS "ConceptMapping_map_reverse_idx" ON "ConceptMapping" ("conceptMap", "targetSystem", "targetCode", "sourceSystem")`);
+  await fns.query(client, results, `CREATE TABLE IF NOT EXISTS "ConceptMapping_Attribute" (
+    "mapping" BIGINT NOT NULL,
```
The-Alchemist, line 29: should this be a FK to `ConceptMapping.id`?

packages/server/src/migrations/schema/v98.ts
```diff
@@ -0,0 +1,36 @@
...
+import { PoolClient } from 'pg';
+import * as fns from '../migrate-functions';
+
+// prettier-ignore
+export async function run(client: PoolClient): Promise<void> {
+  const results: { name: string; durationMs: number }[] = []
+  await fns.query(client, results, `CREATE TABLE IF NOT EXISTS "ConceptMapping" (
+    "id" BIGSERIAL PRIMARY KEY,
+    "conceptMap" UUID NOT NULL,
+    "sourceSystem" TEXT NOT NULL,
```
The-Alchemist: I think it makes sense to create a separate `ConceptSystem` table or something and put a FK (e.g., `sourceCodeId`).  this way we can avoid having millions/billions entries of "http://snomed.info/sct").  

Much lower disk space, smaller indices, faster query performance, etc.

packages/server/src/migrations/schema/v98.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+import { PoolClient } from 'pg';
+import * as fns from '../migrate-functions';
+
+// prettier-ignore
+export async function run(client: PoolClient): Promise<void> {
+  const results: { name: string; durationMs: number }[] = []
+  await fns.query(client, results, `CREATE TABLE IF NOT EXISTS "ConceptMapping" (
+    "id" BIGSERIAL PRIMARY KEY,
```
The-Alchemist: `BIGSERIAL/SERIAL` are basically legacy.  `IDENTITY` is the way go.

Should be: `"id" int8 GENERATED ALWAYS AS IDENTITY PRIMARY KEY` 

* https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_serial
* https://www.enterprisedb.com/blog/postgresql-10-identity-columns-explained
* https://github.com/kristiandupont/schemalint/tree/main/src/rules#prefer-identity-to-serial

packages/server/src/migrations/migrate.ts
```diff
@@ -916,6 +918,58 @@ function buildCodeSystemPropertyTable(result: SchemaDefinition): void {
...
+  result.tables.push({
+    name: 'ConceptMapping_Attribute',
+    columns: [
+      { name: 'mapping', type: 'BIGINT', notNull: true },
+      { name: 'uri', type: 'TEXT', notNull: true },
+      { name: 'type', type: 'TEXT', notNull: true },
+      { name: 'value', type: 'TEXT', notNull: true },
+      { name: 'kind', type: 'TEXT', notNull: true },
+    ],
+    indexes: [
```
mattlong: You can directly declare the composite PK and avoid needing to manually define this index:

```suggestion
    compositePrimaryKey: ['mapping', 'uri', 'type', 'value', 'kind'],
```

packages/server/src/migrations/schema/v98.ts
The-Alchemist: @mattlong just added `IDENTITY` support at https://github.com/medplum/medplum/pull/7481!  Thank you, ml 👍

packages/server/src/fhir/lookups/conceptmapping.ts
```diff
@@ -0,0 +1,94 @@
...
+   * @param resourceType - The FHIR resource type.
+   * @param before - The date before which resources should be purged.
+   */
+  async purgeValuesBefore(client: Pool | PoolClient, resourceType: ResourceType, before: string): Promise<void> {
+    if (resourceType !== 'ConceptMap') {
+      return;
+    }
+
+    // Delete ConceptMapping_Attribute entries with a double join
+    await new DeleteQuery('ConceptMapping_Attribute')
```
mattlong: Should the two deletes go in a transaction?

packages/server/src/migrations/data/v29.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import { WithId } from '@medplum/core';
+import { AsyncJob, ResourceType } from '@medplum/fhirtypes';
+import { prepareReindexJobData, ReindexJob, ReindexPostDeployMigration } from '../../workers/reindex';
+
+// Repository.VERSION was bumped to 11 for the ConceptMapping lookup table,
+// so reindex all resources with a lower version.
+const maxResourceVersion = 1;
```
mattlong:
```suggestion
const maxResourceVersion = 10;
```

packages/server/src/fhir/repo.ts
```diff
@@ -278,7 +278,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    *  8. 08/06/25 - Added Task to Patient compartment (https://github.com/medplum/medplum/pull/7194)
    *  9. 08/19/25 - Added search parameter `ServiceRequest-reason-code` (https://github.com/medplum/medplum/pull/7271)
    * 10. 08/27/25 - Added HumanName sort columns (https://github.com/medplum/medplum/pull/7304)
-   *
+   * 11. 09/25/25 - Added ConceptMapping lookup table (https://github.com/medplum/medplum/pull/7469)
    */
   static readonly VERSION: number = 10;
```
mattlong:
```suggestion
  static readonly VERSION: number = 11;
```

packages/server/src/fhir/operations/conceptmapimport.ts
```diff
@@ -0,0 +1,301 @@
...
+  if (!repo.isProjectAdmin() && !isSuperAdmin) {
+    return [forbidden];
+  }
+
+  const params = parseInputParameters<ConceptMapImportParameters>(operation, req);
+
+  let conceptMap: WithId<ConceptMap>;
+  if (req.params.id) {
+    conceptMap = await repo.readResource('ConceptMap', req.params.id);
+  } else if (params.url) {
```
mattlong, line 126: Should an error be thrown if both `req.params.id` and `params.url` are specified but `params.url` does not match the url of the `req.params.id` resource?

packages/server/src/fhir/lookups/conceptmapping.ts
mattwiller: We've considered it in the past, but the feeling was that a) the operation is idempotent anyway, so just retrying on failure should be fine; and b) such a transaction could touch hundreds of thousands of rows in both tables, potentially being vulnerable to conflicts.  I'm open to adding it though, if we think the benefit would be worth it

packages/server/src/fhir/operations/conceptmapimport.ts
mattwiller: I think the parameters are effectively mutually exclusive; if anything maybe we should throw if both are specified?

PR #7470, opened 2025-09-24
feat(agent): enable fetch logs via `agent:transmit:request` message

ThatOneBro:
Enables use of `$fetch-logs` functionality on older versions of Medplum server before the `$fetch-logs` PR by overloading the already deprecated `push` message as a stopgap


ThatOneBro:
Closing this since we don't need to do this one-off backdoor shenanigans after all

PR #7474, opened 2025-09-25
Added Type parameter to `isReference`

rahul1:
Add an optional type parameter to assert that the input reference is of a certain type. 

This helps avoid downstream casting by clients  isReference is called

### DIFF
--- a/packages/core/src/types.test.ts
+++ b/packages/core/src/types.test.ts
@@ -110,6 +110,7 @@ describe('Type Utils', () => {
   });
 
   test('isReference', () => {
+    // Basic reference validation (no resourceType parameter)
     expect(isReference(undefined)).toBe(false);
     expect(isReference(null)).toBe(false);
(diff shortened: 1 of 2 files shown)


mattlong (approved):

mattlong:
> This is basically just doing the `as` cast, but hiding it behind the assume safety of this type check function. I'd prefer to actually check the type if this function is going to assert that it matches

That's a good point. This function exists in the Health Gorilla library, maybe it should be moved alongside `isReference` in core?

```typescript
export function isReferenceOfType<T extends ResourceType>(
  resourceType: T,
  value: unknown
): value is Reference<ExtractResource<T>> & { reference: string } {
  return isReference(value) && value.reference.startsWith(resourceType + '/');
}
```

mattlong:
Or keep it as one function and add `resourceType: T` as a second optional parameter of `isResource`?

Issue #7468, opened 2025-09-24
$set-accounts target resource update requires extended mode


finnbergquist:
### Overview

In, the `$set-accounts ` operation, the target resource is updated using repo instead of systemRepo. See [this line](https://github.com/medplum/medplum/blob/e43c875df23cd562374c2bd6325a31885dbf30cf/packages/server/src/fhir/operations/set-accounts.ts#L144).

This means that there is different behavior depending on whether or not you pass the `x-medplum : extended` header. 

### Example scenario - removing a reference from Patient's compartment

1. Create a Patient
2. Now call `$set-accounts` with 
```
{
    "resourceType": "Parameters",
    "parameter": [
        {
            "name": "accounts",
            "valueReference": {
                "reference": "Organization/123"
            }
        },
 {
            "name": "accounts",
            "valueReference": {
                "reference": "Organization/456"
            }
        },
        {
            "name": "propagate",
            "valueBoolean": false
        }
    ]
}
```
3. Now that `meta.accounts` is populated on the Patient, there is different behavior depending on whether or not `x-medplum : extended` is added on the next `$set-accounts` call:

call `$set-accounts` with to remove the first Organization reference
```
{
    "resourceType": "Parameters",
    "parameter": [
      {
            "name": "accounts",
            "valueReference": {
                "reference": "Organization/456"
            }
        },
        {
            "name": "propagate",
            "valueBoolean": false
        }
    ]
}
```

Observed behavior without `x-medplum : extended` - no updates are made to the Patient

Observed behavior with `x-medplum : extended` - the Organization reference is removed [Intended behavior]

### Fix

The update to the target resource should be called with systemRepo, after applying the appropriate permission check for write access to that resource type

syp1xd:
@mattwiller Can I pick this up?

PR #7475, opened 2025-09-25
fix(server): $set-accounts uses systemRepo with write permission check

psyp1x:
## Fix 
Issue: https://github.com/medplum/medplum/issues/7468

## Problem
Removing accounts via `$set-accounts` depended on caller extended mode
Target updates were performed using the caller’s repo, which meant meta fields weren’t handled consistently.

## Root Cause
The target resource was updated through the caller’s repo instead of a privileged, consistent path.

## Solution
- Read the target resource via `systemRepo`.
- Enforce caller **READ** and **UPDATE** permission checks against the caller repo.
- Perform the actual update using `systemRepo.updateResource`.
- Continue using `systemRepo` for compartment propagation.

This ensures consistent handling of `meta.accounts` without relying on extended mode.

### Permissions
- Verify caller has **READ** on target (caller repo).
- Verify caller has **UPDATE** on target (caller repo).
- Perform updates via `systemRepo` (avoids extended-mode dependency).

### Behavior
- Removing/adding `meta.accounts` no longer requires the extended header.
- Patient-compartment propagation remains unchanged.

### Files Changed
- **set-accounts.ts**  
  - Use `systemRepo` for target updates  
  - Add explicit write-permission check  
- **set-accounts.test.ts**  
  - Added regression: *“Removes account without extended header”*  
  - Kept existing coverage intact  
- **seed.test.ts**  
  - Test-only: initialize Redis before `initAppServices` to avoid local *“Redis not initialized”* errors during post-commit cache writes  

## Testing
- Focused: `set-accounts.test.ts` — **PASS** (9/9), including the new regression.  
- Full `@medplum/server` suite (local): **177/178 suites PASS**  
  - One suite showed unrelated flakiness (`src/fhir/repo.test.ts`).  
  - First run: timeout in *“many entries stored in lookup table”*.  
  - Re-run: passed; different legacy UUID test failed with *“Assigned ID is already in use”* (stateful/local flake).  

No schema changes; no migrations needed.

## Risk and Compatibility
- **Risk:** Low — uses `systemRepo` for target updates after permission checks.  
- **Compatibility:** Backward compatible; API unchanged. Behavior is more consistent (extended header no longer required).  

## Verification Steps
1. Create a Patient and two Organization accounts.  
2. `POST Patient/{id}/$set-accounts` with both accounts and `propagate=true`.  
3. `POST` again with only one account (no extended header).  
4. `GET Patient` → confirm `meta.accounts` reflects removal.  
   - If `propagate=true`, related compartment resources reflect updates.  

## Release Notes
- `$set-accounts` no longer requires extended mode to update or remove `meta.accounts`.  
- Target updates now use `systemRepo` with proper permission checks.


packages/server/src/fhir/operations/set-accounts.ts
```diff
@@ -141,7 +141,10 @@ export async function setResourceAccounts(
     accounts: accounts,
     account: accounts?.[0],
   };
-  await repo.updateResource(target);
+  if (!repo.canPerformInteraction(AccessPolicyInteraction.UPDATE, target)) {
+    throw new OperationOutcomeError(forbidden);
+  }
+  await systemRepo.updateResource(target);
```
mattlong, line 147: Right now `updateResource` eventually calls `getAccounts` which has differing behavior for extended mode which I believe is intentional? Paging @mattwiller

mattlong: Oops, disregard my comment; I missed #7468.

PR #7478, opened 2025-09-25
Server config to discourage sequential scans

### LINKED ISSUE #7476: Discourage sequential scans in FHIR searches
The Postgres query planner's selectivity estimates for filters on array columns can vastly overestimate how many rows are expected to match the filter. This can cause Postgres to choose a suboptimal query plan, even sequential scans, rather than utilizing the available indexes on the column.  This leads to significantly slower and more expensive queries and system instability. See #7310 for a more in-depth technical analysis.

To address this in the short term, an opt-in `@medplum/server` feature is added. When enabled, the shape of the SQL queries generated to fulfill certain FHIR searches will we changed to encourage the query planner to utilize indexes rather than resorting to expensive sequential scans. The results of the FHIR search will remain unchanged.

Based on current research, this feature will have two effects.

First, it will strongly discourage sequential scans by setting `enable_seqscan = off` for the FHIR search's Postgres query. Despite the name, this change doesn't truly remove the possibiilty of a sequential scan being used; it just massively inflates the cost of a sequential scan to the planner.

Second, enforce a artificially higher floor in the `LIMIT N` clause included in the generated SQL queries. Why? When they query planner vastly over estimates how many matching rows exist and we are only asking for a couple rows to be returned, i.e. `LIMIT 2`, it chooses sequential scan since it thinks it will surely find those two rows without needing to scan much of the table. Increasing the limit has a dramatic impact on the planner's estimated cost of the sequential scan which makes it more likely for an alternative, index-based plan to be used.

mattlong:
Fixes #7476


packages/server/src/fhir/search.ts
```diff
@@ -312,10 +312,32 @@ async function getSearchEntries<T extends Resource>(
   searchRequest: SearchRequestWithCountAndOffset<T>,
   builder: SelectQuery
 ): Promise<{ entry: BundleEntry<WithId<T>>[]; rowCount: number; nextResource?: T }> {
-  const rows = await builder.execute(repo.getDatabaseClient(DatabaseMode.READER));
-  const rowCount = rows.length;
+  const config = getConfig();
+  const originalLimit = builder.limit_;
+
+  if (config.fhirSearchMinLimit !== undefined && config.fhirSearchMinLimit > builder.limit_) {
+    builder.limit(Math.max(builder.limit_, config.fhirSearchMinLimit));
```
codyebberson: Nice, this works because the SQL builder `limit()` method will only add one limit, last call wins 👍

codyebberson (approved):

PR #7480, opened 2025-09-26
[Certification Provider App] Fix `negationInd` usage for procedure entry

pamella:
## Description

- Fix `negationInd` usage for procedure entry

## Steps to test

- C1 flow
  - Generate the QRDA file


pamella:
CC @techdavidy

PR #7481, opened 2025-09-26
Support identity columns

mattlong:
As mentioned in https://github.com/medplum/medplum/pull/7469#pullrequestreview-3268843966


The-Alchemist:
LGTM, glad we can use `IDENTITY` now 🙇

PR #7482, opened 2025-09-26
Avoid nulls when removing 2nd address line

codyebberson:


packages/core/src/utils.ts
```diff
@@ -646,6 +646,25 @@ export function isPopulated<T extends { length: number } | object>(arg: CanBePop
...
+ */
+export function trimTrailingEmptyElements<T>(arr: T[] | undefined): T[] | undefined {
+  if (!arr) {
+    return undefined;
+  }
+  let i = arr.length - 1;
+  while (i >= 0 && isEmpty(arr[i])) {
+    i--;
+  }
+  return i > 0 ? arr.slice(0, i + 1) : undefined;
```
mattwiller: Here `i` is the index of the last non-null element in the array: if it's zero we still want to return that element
```suggestion
  return i >= 0 ? arr.slice(0, i + 1) : undefined;
```

codyebberson: Eek, my last minute cleanup added a bug.  Ty 🙏

PR #7488, opened 2025-09-26
[Certification Provider App] Refactor code to use `string` instead of `dateTime`

pamella:
## Description

Refactor questionnaires and bots to use `string` instead of `dateTime` due to timezone-aware datetimes created in questionnaire responses due to QuestionnaireForm submit behavior

> [!WARNING]
> **PLEASE NOTE:** After this change, on resource detail pages in app.medplum, the `JSON` tab will display datetimes in UTC (matching the values shown in patient record HTML files and QRDA XML files), while the `Details` tab will display timezone-aware values. 
>
> I attached a video below to demonstrate this (in the example, the user’s timezone is UTC-3), which is why the value shifts from 8 AM UTC to 5 AM timezone-aware.

[Screencast from 26-09-2025 17:24:41.webm](https://github.com/user-attachments/assets/58f0edbd-1260-4919-b261-779bbccf15e2)

## Steps to test

- C1 flow
  - Create a new Patient (Intake form)
  - Generate the QRDA file

### DIFF
--- a/examples/medplum-provider/data/core/c1/questionnaires.json
+++ b/examples/medplum-provider/data/core/c1/questionnaires.json
@@ -39,7 +39,7 @@
               },
               {
                 "linkId": "dob",
-                "text": "Date of Birth",
+                "text": "Date of Birth (YYYY-MM-DDTHH:MM:SS+00:00)",
                 "type": "string"
               },
               {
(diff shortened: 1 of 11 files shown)


pamella:
CC @techdavidy

Issue #7434, opened 2025-09-20
Internal server error when uploading a Binary resource


dselans:
Hello,

My service is uploading `Binary` resources via `POST /fhir/r4/Binary` endpoint.

I am defining the binary as follows (using google's fhir protobuf lib):

```
	// Create Binary resource
	binary := &bcr_gp.ContainedResource{
		OneofResource: &bcr_gp.ContainedResource_Binary{
			Binary: &binary_go_proto.Binary{
				ContentType: &binary_go_proto.Binary_ContentTypeCode{Value: "application/octet-stream"},
				Data:        &dt_gp.Base64Binary{Value: data},
			},
		},
	}
```

I initially thought it may have been because the Binary had `application/json` content-type and that caused the extra validation - but I am seeing the same behavior with the octet-stream content-type as well.

---

This is the stack trace I am seeing from the server:

```
Maximum call stack size exceeded
RangeError: Maximum call stack size exceeded
                                                                                                                                              at RegExp.exec (<anonymous>)
at Mr.validateString (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:50188)
at Mr.validatePrimitiveType (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:49960)
at Mr.checkPropertyValue (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:47740)
at Mr.visitProperty (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:47392)
at Vr.crawlProperty (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:43819)
at Vr.crawlObject (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:43567)
at Vr.crawl (/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:43268)
at Lt(/usr/src/medplum/packages/core/dist/cjs/index.cjs:1:43078)
```

---

The binary contains a FHIR JSON blob of various size (4K-20MB) that I do not have control over. It also might be invalid JSON which is why I am storing it for later processing.

TYIA!

codyebberson:
Hi @dselans  - thanks for reporting, apologies for the slow reply.

Is the `Binary` a contained resource?  If the outer-most content is `octet-stream`, then it should not attempt to do any validation.  However, if the outer-most is a FHIR resource, and the `Binary` is a contained resource, then it may still go through the validation path.

PR #7494, opened 2025-09-28
[Fix] validate base64Binary as a first-class primitive; remove Binary.data special-case

syp1xd:
# Validate `base64Binary` as a Primitive

## Related Issue:
https://github.com/medplum/medplum/issues/7434

## Summary
This change fixes validation issues for large Binary uploads and other `base64Binary` fields by:

- Treating `base64Binary` as a first-class primitive during validation.  
- Removing the `Binary.data` field-specific special-case in traversal.  
- Adding a decoded-size guard (~50MB) to avoid pathological payloads while keeping the primitive opaque to generic string checks.  

This ensures consistent behavior across resources (e.g., `Binary.data`, `Attachment.data`) and prevents string-level checks (whitespace trimming, 1MB string limit) from incorrectly applying to `base64Binary` values.

---

## Problem
Previously:
- `base64Binary` values were validated indirectly via generic FHIR string rules.  
- A `Binary.data` special-case existed in traversal.  
- Large uploads could trigger:
  - 1MB FHIR string limit  
  - whitespace trimming rules (intended for human text)  
- This caused **internal server errors** for large Binary uploads.

---

### `validation.ts`
- Route `base64Binary` through `validateBase64Binary` in `validatePrimitiveType`.  
- Enforce only:
  - base64 alphabet/padding via `validationRegexes.base64Binary`  
  - approximate decoded-size limit of 50MB  
- Remove resource/field-specific logic in traversal (no `Binary.data` special-casing).  
- Keep generic string rules for FHIR `string`-like types unchanged.  

### `validation.test.ts`
- Add/update tests for:
  - Valid/invalid base64 shapes  
  - Large payloads within decoded-size cap  
  - Non-recursion and error surfacing for invalid additional properties  

---

## Rationale
- `base64Binary` is not human-readable text; string-level rules like trimming and 1MB limits are inappropriate.  
- Uniform primitive-level validation improves consistency and removes brittle special-casing.  
- A decoded-size guard protects against unbounded payloads while remaining practical for real-world uploads.  

---

## Tests
- **@medplum/core**: 62/62 passing (`typeschema/validation.test.ts` updated).  
- **@medplum/server**: 13/13 passing (Binary create/read/update, encodings, JSON fallbacks, disallowed file types)

---

## Backwards Compatibility & Risk
- Behavior change limited to `base64Binary` validation:
  - Previously: subject to FHIR string rules (trimming, 1MB).  
  - Now: regex + 50MB decoded-size guard.  
- Applies uniformly across resources (`Binary`, `Attachment`, etc.).  
- **Low risk**:
  - Targeted tests fully cover new behavior.  
  - Binary endpoint integration tests pass end-to-end.  

---

## Follow-ups (Optional)
- Add dedicated test for `Attachment.data` to demonstrate uniform handling.  
- Consider making the 50MB limit configurable if product requirements vary.  

---

## Changelog
- **core**: Validate `base64Binary` as primitive with regex + decoded-size guard; remove `Binary.data` traversal special-case; keep generic string rules unchanged.  
- **tests(core)**: Expand `base64Binary` test coverage; adjust invalid object expectations.  

---


packages/core/src/typeschema/validation.ts
```diff
@@ -500,7 +500,11 @@ class ResourceValidator implements CrawlerVisitor {
       }
       // Then, perform additional checks for specialty types
       if (expectedType === 'string') {
-        this.validateString(value as string, type, path);
+        if (type === 'base64Binary') {
```
mattwiller: I'd recommend unnesting this into e.g.
```ts
if (type === 'base64binary') {
  // ...
} else if (expectedType === 'string') {
  // ...
} else if (expectedType === 'number') {
  // ...
}
```

packages/core/src/typeschema/validation.ts
```diff
@@ -510,6 +514,41 @@ class ResourceValidator implements CrawlerVisitor {
...
+   * - Apply an approximate decoded-size limit to guard against pathological payloads.
+   * @param str - The base64-encoded string.
+   * @param path - The FHIR element path for issue reporting.
+   */
+  private validateBase64Binary(str: string, path: string): void {
+    // Must be a non-empty string; allow empty only if schema constraints permit it (handled elsewhere)
+    if (typeof str !== 'string') {
+      this.issues.push(createStructureIssue(path, 'Invalid JSON type: expected string, but got ' + typeof str));
+      return;
+    }
```
mattwiller: This type check shouldn't be necessary, given the function signature

packages/core/src/typeschema/validation.ts
```diff
@@ -510,6 +514,41 @@ class ResourceValidator implements CrawlerVisitor {
...
+
+    const regex = validationRegexes.base64Binary;
+    if (regex && !regex.exec(str)) {
```
mattwiller: Is it possible for the regex to be falsy here?
```suggestion
    if (!validationRegexes.base64Binary.exec(str)) {
```

packages/core/src/typeschema/validation.ts
```diff
@@ -510,6 +514,41 @@ class ResourceValidator implements CrawlerVisitor {
...
+
+    // Approximate decoded size: 3/4 of length minus padding
+    let padding = 0;
+    if (str.endsWith('==')) {
+      padding = 2;
+    } else if (str.endsWith('=')) {
+      padding = 1;
+    }
+    const approxBytes = Math.max(0, Math.floor((str.length * 3) / 4) - padding);
+    const MAX_BYTES = 50 * 1024 * 1024; // 50 MB
```
mattwiller: Ideally any value like this should be configurable, not static.  In any case, there's already a cap on the JSON size that can be sent to the server: when would that limit be insufficient?

packages/core/src/typeschema/validation.ts
syp1xd: done

packages/core/src/typeschema/validation.ts
syp1xd: It intentionally does not allow URL‑safe characters (-/_) since FHIR uses standard base64.
It matches full strings only; invalid tail bytes/padding are rejected.
base64Binary with only whitespace  -> fails with “only whitespace”.
base64Binary with newlines/spaces mixed in a valid payload -> passes.
Boundary: valid payload exactly at cap after whitespace removed -> passes; +1 byte -> fails.

packages/core/src/typeschema/validation.ts
```diff
@@ -139,6 +141,7 @@ class ResourceValidator implements CrawlerVisitor {
     } else {
       this.schema = parseStructureDefinition(options.profile);
     }
+    this.base64BinaryMaxBytes = options?.base64BinaryMaxBytes ?? 50 * 1024 * 1024;
```
codyebberson: After reviewing the PR, we've decided to keep the default maximum string length at 1 MB. Our priority is to keep the platform as lightweight and performant as possible, especially given the potential impact on database size and application memory for our users.

That said, we noticed you made the max length a configurable value in the validator, which is a great idea. While we'll keep our hosted platform's default at 1 MB, we would be open to a follow-up pull request that adds a server-side configuration setting for this value. This would give self-hosting customers the flexibility to set a higher limit if they choose, while ensuring our core defaults remain optimized.

```suggestion
    this.base64BinaryMaxBytes = options?.base64BinaryMaxBytes ?? 1 * 1024 * 1024;
```

codyebberson (changes requested):

packages/core/src/typeschema/validation.ts
syp1xd: Thanks for the review, @codyebberson. I’ve updated the default base64BinaryMaxBytes to 1 MB. I’ll open a follow-up PR to add a server-side configuration option.

PR #7504, opened 2025-09-29
Added npm-check-updates cooldown

codyebberson:


codyebberson:
@claude - please review.  Note that the default cooldown value is "3 days".  Is that too long or too short?

PR #7527, opened 2025-10-01
test(server): run tests against PostgreSQL 18 as well

The-Alchemist:
PostgreSQL 18 just became available on Docker Hub, so introducing it to our test matrix increases our effective test coverage.

https://hub.docker.com/layers/library/postgres/18

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -159,7 +159,7 @@ jobs:
     strategy:
       matrix:
         node-version: [20, 22]
-        pg-version: [13, 17, 18]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -159,7 +159,7 @@ jobs:
     strategy:
       matrix:
         node-version: [20, 22]
-        pg-version: [13, 17]
+        pg-version: [13, 17, 18]
```
mattlong: Great! Let's remove 17 from the list

PR #7528, opened 2025-10-01
DRAFT (DO NOT MERGE): feat(db): rename tables to lowercase and remove the quotes from the SQL builder

The-Alchemist:


codyebberson (changes requested):
Yeah, this is a big "thing i would change if we could start again".  My understanding is that there isn't a simple way to undo that mistake without some pretty heavy duty effort (duplicate tables, triggers, etc).

The-Alchemist:
> Yeah, this is a big "thing i would change if we could start again". My understanding is that there isn't a simple way to undo that mistake without some pretty heavy duty effort (duplicate tables, triggers, etc).

I'd like to hear more about it, because I've had good luck with just `ALTER TABLE ... RENAME TO`.  It's an instant operation and doesn't require re-building indices or anything.

codyebberson:
> I'd like to hear more about it, because I've had good luck with just `ALTER TABLE ... RENAME TO`. It's an instant operation and doesn't require re-building indices or anything.

The trick is how to synchronize that with running servers.  Current SQL expects there to be double quotes and case-sensitive names.  So you have servers running with that model.  You need to execute `ALTER TABLE` and switch all of those servers at the exact same moment in time.

You could potentially add server logic to catch the error, and flip into a different mode?  Kinda thorny.

For what it's worth, I 100% accept that the current model is not ideal, and I take full responsibility for the suboptimal naming strategy.  I would happily go back in time and kick myself.

codyebberson:
Aspirational, but I fear more risk than it's worth.

I take full responsibility for the bad naming 😢

PR #7530, opened 2025-10-02
[Enhancement] core configurable base64 caps

psyp1x:
## Summary
This follow-up PR completes the **base64Binary hardening** https://github.com/medplum/medplum/pull/7494  by making the decoded-size cap configurable at the server level, threading that policy through the validator, and adding a defense-in-depth exact decoded-size check in the repository when persisting `Binary.data`.

---

### Core validator (`@medplum/core`)
- Accepts `base64BinaryMaxBytes` via `ValidatorOptions` and enforces it for any base64Binary field (`Binary.data`, `Attachment.data`, etc.) using an **approximate decoded-size check** (3/4 length minus padding).

### Server (`@medplum/server`)
- New server config: `base64BinaryMaxBytes` (default set in config utils; configurable via env).
- Repository now passes `ValidatorOptions` built from `getConfig()` into `validateResource(...)` for both:
  - Base validation pass
  - Per-profile validations
- **Defense-in-depth:** `handleBinaryData()` decodes base64 and rejects payloads when the exact decoded size exceeds the configured cap, before writing to storage.

---

## Why
- Prevents **stack overflow / internal server errors** from pathological base64 payloads by treating `Binary.data` as opaque and bounding decoded size.  
- Centralizes policy ownership in the server while keeping the core reusable and environment-agnostic.  
- Aligns with existing policy patterns (e.g., `maxSearchOffset`): server provides policy, core enforces via options.  

---

## Behavior and API
- **No REST API changes.**  
- Validation:
  - Core enforces base64Binary cap if provided via `ValidatorOptions`.
  - Server always supplies the cap from `getConfig()` for strict validation passes.
- Binary writes are additionally guarded by an **exact decoded-size check** prior to storage.


mattlong:
@claude please review and be extra diligent in the changes in validation-related code

mattlong (approved):
Thanks so much, LGTM!

Thoughts, @codyebberson?

PR #7531, opened 2025-10-02
feat(server): adding pg_partman to PostgreSQL setup

The-Alchemist:
## Summary

This PR installs the `pg_partman` extension in our PostgreSQL Docker container to match the behavior of AWS RDS PostgreSQL instances.

**Why this is useful:**
- AWS RDS PostgreSQL automatically includes the `pg_partman` extension, so this change ensures our local development environment matches production
- `pg_partman` is a powerful tool for managing partitioned tables, which can be useful for performance optimization and data management
- Having it pre-installed in our Docker container eliminates the need to manually install it when needed for development or testing

**NOTE**: This PR only installs the extension; it does not actually use or configure `pg_partman` anywhere in the codebase. The extension will be available for future use when needed.

**Changes:**
- Updated `Dockerfile.postgres` to install `postgresql-16-partman` package
- Added proper apt cleanup to keep the Docker image lean and mean

This require devs to run `docker compose build` to build the image and take advantage of the changes, which takes <1 minute.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,10 @@
 # You can start all services by running "docker-compose up"
 services:
   postgres:
-    image: postgres:16
+    shm_size: 2g
+    build:
+      context: .
+      dockerfile: Dockerfile.postgres
     restart: always
     environment:
       - POSTGRES_USER=medplum
--- a/Dockerfile.postgres
+++ b/Dockerfile.postgres
@@ -1,0 +1,7 @@
+FROM postgres:16
+
+# Install pg_partman extension
+RUN apt-get update && \
+    apt-get install -y postgresql-16-partman && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
--- a/docker-compose.full-stack.yml
+++ b/docker-compose.full-stack.yml
@@ -1,6 +1,9 @@
 services:
   postgres:
-    image: postgres:16
+    shm_size: 2g
+    build:
+      context: .
+      dockerfile: Dockerfile.postgres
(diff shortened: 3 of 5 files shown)


docker-compose.yml
```diff
@@ -5,7 +5,9 @@
 # You can start all services by running "docker-compose up"
 services:
   postgres:
-    image: postgres:16
+    build:
```
mattlong, line 9: Instead of the build step, there's also https://github.com/dbsystel/postgresql-partman-container. Images available on ghcr.io: https://github.com/dbsystel/postgresql-partman-container/pkgs/container/postgresql-partman

The-Alchemist: I admire community-drived work like https://github.com/dbsystel/postgresql-partman-container.

From a supply chain vulnerability perspective, I'd prefer not to use external Docker images without a clear cost/benefit analysis.  `Dockerfile.postgres` is only 7 lines total, so the maintenance burden is very low and our supply chain vulnerability is basically zero.  Upgrading or testing with different Postgres images is a very simple as well.

As a case in point, I'd like to point to the https://www.akamai.com/blog/security-research/critical-linux-backdoor-xz-utils-discovered-what-to-know vulnerability.

Hope that makes sense...

codyebberson:
@claude please review

codyebberson:
Thanks @The-Alchemist  -  This is a great starting point for a deeper investigation.

As we discussed offline, let's hold off on merging this PR until the data from the benchmarks clearly justifies adding the extension to our dev setup. I look forward to reviewing the results.

PR #7487, opened 2025-09-26
Jest 30

codyebberson:
* Jest 30: https://jestjs.io/blog/2025/06/04/jest-30
* Migration guide: https://jestjs.io/docs/upgrading-to-jest30
* Mostly smooth, two unfortunate overrides:
   * Docusaurus uses `css-minimizer-webpack-plugin`, which uses `jest-worker` 29
   * `jest-websocket-mock` uses `jest-diff` 29
   
--------

Update 2025-09-27:
* Jk, not smooth.  Jest 30 breaks a lot.  The big one for us will be that you can no longer use `Object.defineProperty` on `window` 😱 

--------

Update 2025-10-03:
* Everything builds and passes
* Unfortunately some tests are timing out
* Performance regressions are apparently known issue with Jest 30: https://github.com/jestjs/jest/issues/15718
* Some other devs have reports that using `isolatedModules` can speed things up
* So now focusing on the `isolatedModules` PR: https://github.com/medplum/medplum/pull/7540

--------

Update 2025-10-04:
* After https://github.com/medplum/medplum/pull/7542 all tests passing
* Still recommend we first merge the `isolatedModules` PR: https://github.com/medplum/medplum/pull/7540

--------

Update 2025-10-08
* Enabled new Jest `globalsCleanup` feature for better perf
* Changed all of the websocket tests to use different port numbers to avoid port contention

PR #7533, opened 2025-10-02
Jest 30

codyebberson:
Extracting `locationUtils` from the [Jest 30](https://github.com/medplum/medplum/pull/7487) pull request.

Jest 30 upgrades JSDOM from v21 to v26.  JSDOM became a lot more strict over those versions.  In particular, you no longer can modify `window` or `window.location`, which we used in a bunch of tests to emulate various browser transitions, or trying to detect when running in browser vs Node.js environments.


mattlong (approved):

codyebberson:
@claude - please review

PR #7537, opened 2025-10-03
fix(agent): parse optional `logLevel` from CLI args

ThatOneBro:
In the [Agent Dockerfile](https://github.com/medplum/medplum/blob/d7dc85640b00d36495ffc6860ca643038460ccc6/packages/agent/Dockerfile#L23), we pass in an optional `logLevel` arg, which we apparently haven't been parsing. This PR fixes that and fixes a test that should have caught that

### DIFF
--- a/packages/agent/src/agent-main.test.ts
+++ b/packages/agent/src/agent-main.test.ts
@@ -78,6 +78,30 @@ describe('Main', () => {
     expect(process.exit).not.toHaveBeenCalled();
   });
 
+  test('Command line arguments with optional logLevel', async () => {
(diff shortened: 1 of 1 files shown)


packages/agent/src/agent-main.test.ts
```diff
@@ -242,18 +266,8 @@ describe('Main', () => {
       return mockLogger(config.logLevel);
     });
 
-    // Mock parseLoggerConfigFromArgs to return empty warnings
```
ThatOneBro, line 245: This was hiding a log level not matching what was expected by returning `LogLevel.INFO` no matter what 🙃

The-Alchemist (approved):
great find!

PR #7540, opened 2025-10-03
Type imports for `isolatedModules` and `verbatimModuleSyntax`

codyebberson:
This PR enables the `isolatedModules` compiler option. This flag enforce a stricter, more predictable compilation process and are a best practice for modern TypeScript projects.  This lays the groundwork for future use of `verbatimModuleSyntax`, after we go full ESM.

#### Why This Is Important

* **`verbatimModuleSyntax`**: This setting ensures that the TypeScript compiler's output directly reflects the module syntax we write. Specifically, it guarantees that `import type` and `export type` statements are fully removed from the compiled JavaScript, while regular imports are always preserved. This removes ambiguity and makes our compilation more robust, particularly when using bundlers or transpilers that process files individually.
* **`isolatedModules`**: This flag ensures that every file in our codebase can be compiled independently. This is a critical requirement for tools like Babel, esbuild, and swc, which transpile TypeScript without full type-system knowledge. Because this flag is a prerequisite for using these faster transpilers, enabling it prepares our codebase for future performance improvements in our build process, including potentially faster Jest test runs.

#### What's in the PR

The primary changes in this PR are the addition of the `type` keyword to many import and export statements. These changes were automatically applied using a configured ESLint rule.

* `tsconfig.json` changes:
    * `"isolatedModules": true`

(We can't actually adopt `"verbatimModuleSyntax": true` until after we're full ESM)

* `packages/eslint-config/index.cjs` changes to enforce the correct syntax:
    * `'no-duplicate-imports': 'off'`
    * `'import/consistent-type-specifier-style': ['error', 'prefer-inline']`
    * `'import/no-duplicates': 'error'`
    * `'@typescript-eslint/consistent-type-imports': ['error', { prefer: 'type-imports', fixStyle: 'inline-type-imports' }]`
    * `'@typescript-eslint/consistent-type-exports': 'error'`

The `eslint --fix` command was run to apply these changes across the entire codebase, which is why a large number of files are affected.


examples/medplum-healthie-importer/src/healthie/allergy.ts
```diff
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { AllergyIntolerance, Patient, Reference } from '@medplum/fhirtypes';
-import { HealthieClient } from './client';
+import { type AllergyIntolerance, type Patient, type Reference } from '@medplum/fhirtypes';
```
ThatOneBro, line 3: Personal preference for:
```suggestion
import type { AllergyIntolerance, Patient, Reference } from '@medplum/fhirtypes';
```
But maybe that's just me.
Fully separating out all type imports from non-type imports into separate import statements feels a little cleaner

codyebberson: Yeah.... I went back and forth on that.  There are other cases where having the inline felt much cleaner.  I fear this is going  to be a tabs-vs-spaces thing, and that the worst option is "mixed", but that means people will be unhappy 😢

codyebberson: For example, in cases with only 2 imports:

Inline:

```ts
import { A, type B } from 'foo';
```

Top:

```ts
import { A } from 'foo';
import type { B } from 'foo`';
```

codyebberson:
@claude please review

codyebberson:
Closing in preference of https://github.com/medplum/medplum/pull/7546

PR #7541, opened 2025-10-03
[Do not merge] Type imports and Jest 30

codyebberson:
Testing the combination of:
* https://github.com/medplum/medplum/pull/7540
* https://github.com/medplum/medplum/pull/7487

To see if it fixes the Jest 30 perf issues


codyebberson:
Ok, this proves it's possible.  Now let's see if we can get the Jest 30 PR to pass independently.

PR #7543, opened 2025-10-06
Dependency upgrades 2025-10-06

medplumbot:
Dependency upgrades

Manual interventions:

Updated `react` and `react-dom` overrides to match the newly available `19.2.0`

Reverted docusaurus packages to 3.8.1 due to [a build bug](https://github.com/facebook/docusaurus/issues/11430#issuecomment-3340616193) that should be fixed in the next release.


codyebberson (approved):
`LicenseRef-W3C-Community-Final-Specification-Agreement` was false positive.  I updated it in Socket.dev.  lgtm 👍

PR #7546, opened 2025-10-06
Type imports for `isolatedModules`, using `separate-type-imports`

codyebberson:
This PR enables the `isolatedModules` compiler option. This flag enforce a stricter, more predictable compilation process and are a best practice for modern TypeScript projects.  This lays the groundwork for future use of `verbatimModuleSyntax`, after we go full ESM.

#### Why This Is Important

* **`verbatimModuleSyntax`**: This setting ensures that the TypeScript compiler's output directly reflects the module syntax we write. Specifically, it guarantees that `import type` and `export type` statements are fully removed from the compiled JavaScript, while regular imports are always preserved. This removes ambiguity and makes our compilation more robust, particularly when using bundlers or transpilers that process files individually.
* **`isolatedModules`**: This flag ensures that every file in our codebase can be compiled independently. This is a critical requirement for tools like Babel, esbuild, and swc, which transpile TypeScript without full type-system knowledge. Because this flag is a prerequisite for using these faster transpilers, enabling it prepares our codebase for future performance improvements in our build process, including potentially faster Jest test runs.

#### What's in the PR

The primary changes in this PR are the addition of the `type` keyword to many import and export statements. These changes were automatically applied using a configured ESLint rule.

* `tsconfig.json` changes:
    * `"isolatedModules": true`

(We can't actually adopt `"verbatimModuleSyntax": true` until after we're full ESM)

* `packages/eslint-config/index.cjs` changes to enforce the correct syntax:
    * `'no-duplicate-imports': 'off'`
    * `'import/consistent-type-specifier-style': ['error', 'prefer-top-level']`
    * `'import/no-duplicates': 'error'`
    * `'@typescript-eslint/consistent-type-imports': ['error', { prefer: 'type-imports', fixStyle: 'separate-type-imports' }]`
    * `'@typescript-eslint/consistent-type-exports': 'error'`

The `eslint --fix` command was run to apply these changes across the entire codebase, which is why a large number of files are affected.

--------

This PR is derived from https://github.com/medplum/medplum/pull/7540

We debated using `prefer-inline` vs `prefer-top-level`.  In the end, we let the data decide.  Using `prefer-inline` led to a diff of `+6,167 −5,266`.  Using `prefer-top-level` led to a diff of `+4,877 −4,364`, so we're going with that.


codyebberson:
@claude please review

PR #7551, opened 2025-10-07
Use PassThrough stream in S3 Upload

mattlong:
A series of recent changes in `@aws-sdk/lib-storage` have come together to cause multi-part uploads where an Express `Request` object being passed in as a a stream causes failures. A `PassThrough` stream supports both `Readable` and `string` inputs and avoids the [issues with Request objects](https://github.com/aws/aws-sdk-js-v3/issues/7408).


mattwiller (approved):

packages/server/src/cloud/aws/storage.ts
```diff
@@ -58,7 +58,8 @@ export class S3Storage extends BaseBinaryStorage {
         Key: key,
         CacheControl: 'max-age=3600, s-maxage=86400',
         ContentType: contentType ?? 'application/octet-stream',
-        Body: stream,
+        // Not wrapping in test env so we can know when PassThrough is not necessary anymore
```
codyebberson, line 64:
```suggestion
        // TODO: Remove after `@aws-sdk/lib-storage` bug is fixed
        // See: https://github.com/aws/aws-sdk-js-v3/issues/7408
        // Not wrapping in test env so we can know when PassThrough is not necessary anymore
```

PR #7505, opened 2025-09-29
HealthChain blog post

codyebberson:
Preview: https://medplum-www-git-cody-healthchain-blog-post-medplum.vercel.app/blog/healthchain

PR #7495, opened 2025-09-29
Dependency upgrades 2025-09-29

medplumbot:
Dependency upgrades

PR #7556, opened 2025-10-07
feat(server): use LATERAL JOINs for lookup tables

The-Alchemist:
- **examples and tests of LATERAL JOINs**
- **Dependency upgrades 2025-09-29 (#7495)**
- **Add npm dedupe to upgrade process (#7498)**
- **Added NOTICE file (#7499)**
- **Revert "Dependency upgrades 2025-09-29 (#7495)" (#7503)**
- **Added npm-check-updates cooldown (#7504)**
- **HealthChain blog post (#7505)**
- **Handle empty args in upgrade.sh (#7506)**
- **Reconnect on Redis readonly error (#7511)**
- **Fix conflict between account propagation and default profiles (#7489)**
- **feat(server): use LATERAL JOINs for lookup tables**


packages/server/src/fhir/lookups/lookuptable.ts
```diff
@@ -215,14 +216,21 @@ export abstract class LookupTable {
...
+
     selectQuery.join(
-      'LEFT JOIN',
-      new SelectQuery(lookupTableName).distinctOn('resourceId').column('resourceId').column(columnName),
+      'LEFT JOIN LATERAL',
+      new SelectQuery(lookupTableName)
+        .column('resourceId')
+        .column(columnName)
+        .whereExpr(whereExpression)
+        .orderBy('resourceId', false)
```
mattlong: Could this order by `columnName`? I believe every possible `columnName` across all lookup tables have a btree index on them, so it should be okay performance wise

packages/server/src/fhir/sql.ts
```diff
@@ -247,6 +247,18 @@ export class Column implements Expression {
   }
 }
 
+export class Constant implements Expression {
```
mattlong, line 250: Heh, there's been a history of classes like this being [added](https://github.com/medplum/medplum/pull/4829), [changed](https://github.com/medplum/medplum/issues/4811), and [removed](https://github.com/medplum/medplum/pull/5451) and then [added again](https://github.com/medplum/medplum/pull/5444).

I agree though; having a non-`Parameter` one makes sense!

packages/server/src/fhir/lookups/lookuptable.ts
The-Alchemist: done!

packages/server/src/fhir/sql.ts
The-Alchemist: Thank you for pointing out the history there.  I couldn't use `Parameter` because it introduces `$1` into the query, whereas I just need `ON true`.

Does the word `Literal` make more sense?  I see a `LiteralAtom` in `packages/core`, but I don't think / hope it would cause confusion.

Issue #7560, opened 2025-10-08
WebSocket subscriptions not firing when Subscription's `meta.account` does not match the resource `meta.account`


ThatOneBro:

ThatOneBro:
Fixed in #7643

PR #7561, opened 2025-10-08
tweak(core/subscriptions): log when acc mismatch prevents sub firing

ThatOneBro:
Solution for #7560 has been found but we can't immediately implement it without potentially affecting customers. For now we will log when a subscription would have fired but the exact match for `meta.account` caused it not to fire when it should have.

Preference for using `_compartment` search parameter in the `Subscription` criteria can be used to more accurately get the desired behavior of filtering resources by the account compartment.

### DIFF
--- a/packages/core/src/subscriptions/index.ts
+++ b/packages/core/src/subscriptions/index.ts
@@ -555,7 +555,7 @@ export async function resourceMatchesSubscriptionCriteria({
 
   // Should eventually remove this once we've seen who is relying on this behavior
   if (subscription.meta?.account && resource.meta?.account?.reference !== subscription.meta.account.reference) {
-    logger?.warn('This subscription would have fired for this resource but the account fields do not match', {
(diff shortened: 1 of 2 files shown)


packages/core/src/subscriptions/index.ts
```diff
@@ -554,7 +549,21 @@ export async function resourceMatchesSubscriptionCriteria({
...
 
-  return matchesSearchRequest(resource, searchRequest);
+  const matches = matchesSearchRequest(resource, searchRequest);
+  if (!matches) {
+    return false;
+  }
+
+  // Should eventually remove this once we've seen who is relying on this behavior
+  if (subscription.meta?.account && resource.meta?.account?.reference !== subscription.meta.account.reference) {
+    logger?.warn('This subscription would have fired for this resource but the account fields does not match', {
```
mattwiller: Tiny suggestion to make the message a little more succinct:
```suggestion
    logger?.warn('Subscription suppressed due to mismatched meta.account', {
```

PR #7573, opened 2025-10-09
Preload all CodeSystemProperty IDs for terminology operation queries

mattwiller:
Follow-on work after #7571 to also preload database IDs for simple attribute properties, in addition to hierarchical relationship properties.  This should significantly improve the performance of `ValueSet/$expand` and related operations on many common types of value sets.


mattwiller:
@The-Alchemist The latter: this is refactoring that shouldn't change existing behavior.  All the tests in `packages/server/src/fhir/operations/expand.test.ts` etc. should still pass and exercise this new code effectively

Issue #7539, opened 2025-10-03
Pad "too unique" array columns to ensure better stats


mattlong:
The idea is to add extra, well-known elements to array columns that fall victim to the quirk in Postgres array column statistics discussed in https://github.com/medplum/medplum/issues/7310 as well as on a [pg mailing list thread](https://www.postgresql.org/message-id/flat/PH3PPF1C905D6E6F24A5C1A1A1D8345B593E16FA%40PH3PPF1C905D6E6.namprd15.prod.outlook.com).

Some initial thoughts and math on what this would look like.

The formula for minimum frequency for consideration into the MCE list:

```c
/*
 * We want statistics_target * 10 elements in the MCELEM array. This
 * multiplier is pretty arbitrary, but is meant to reflect the fact that
 * the number of individual elements tracked in pg_statistic ought to be
 * more than the number of values for a simple scalar column.
 */
num_mcelem = stats->attr->attstattarget * 10;

/*
 * We set bucket width equal to num_mcelem / 0.007 as per the comment
 * above.
 */
bucket_width = num_mcelem * 1000 / 7;

/*
 * Construct an array of the interesting hashtable items, that is,
 * those meeting the cutoff frequency (s - epsilon)*N.  Also identify
 * the minimum and maximum frequencies among these items.
 *
 * Since epsilon = s/10 and bucket_width = 1/epsilon, the cutoff
 * frequency is 9*N / bucket_width.
 */
cutoff_freq = 9 * element_no / bucket_width;
```

For the identifier scenario where each row has an array with a common element and a unique element.

```
element_no = 2 * row_count
bucket_width = stats_target * 10 * 1000 / 7
cutoff_freq = ⌈9 * (2 * row_count) / (stats_target * 10 * 1000 / 7)⌉
```

With a statistics target of 100, the `ANALYZE` command samples `R` = 100 * 300 = 30,000 rows. In the “too unique” identifiers scenario, each row of the array column contains a common array element and a unique element. The array column sample contains 60,000 total elements `n` across the sampled `R` rows; 2 elements per row. For an element to be included in most_common_elements (MCE), it must appear at least `k` times where:

`k = ⌈9 * 2 * 30000 /142,857⌉ ≈ ⌈3.78⌉ = 4`

[...]

PR #7575, opened 2025-10-10
Array column padding

mattlong:
See https://github.com/medplum/medplum/issues/7539 for more context.

A key insight is that cutoff frequency is independent from the statistics target; so using higher statistics target yields lower minfreqs.

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2607,7 +2607,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     const issue = err.outcome.issue[0];
     return Boolean(
       issue.code === 'conflict' &&
-        issue.details?.coding?.some((c) => retryableTransactionErrorCodes.includes(c.code as string))
+      issue.details?.coding?.some((c) => retryableTransactionErrorCodes.includes(c.code as string))
     );
   }
 


packages/server/src/fhir/token-column.test.ts
```diff
@@ -792,3 +797,30 @@ describe('buildTokenColumnsSearchFilter', () => {
     });
   });
 });
+describe('getPaddingElement', () => {
+  test('Math.random is 0.99999999', () => {
+    const randomMock = jest.spyOn(Math, 'random').mockReturnValue(0.99999999);
+    expect(getPaddingElement({ m: 1, lambda: 150, statisticsTarget: 1 })).toStrictEqual(undefined);
+    // once to decide (not to) return a padding element
+    expect(randomMock).toHaveBeenCalledTimes(1);
+    randomMock.mockRestore();
```
noahsilas: This seems like something that, on breaking, could become a pain to track down. If something calls `Math.random` more than once, I think that the expectation here would throw and prevent this cleanup code from being executed. This mock could then leak into other tests and cause confusing failures.

Shifting this into before/after setup blocks should mean it always runs even if the unit under test fails.

packages/server/src/fhir/token-column.ts
```diff
@@ -378,3 +389,21 @@ export function escapeRegexString(str: string): string {
   // | - alternation (OR operator)
   return str.replaceAll(/[.^$*+?()[\]{}\\|]/g, '\\$&');
 }
+
+// Not a v5 UUID, so will not collide with hashed token values
+const UUID_TEMPLATE = '00000000-0000-0000-0000-000000000000';
+
+export function getPaddingElement({ m, lambda, statisticsTarget }: ArrayColumnPaddingConfig): string | undefined {
+  if (Math.random() < (m * lambda) / (statisticsTarget * 300)) {
```
noahsilas: Checking my understanding here:

- `m` is the number of distinct padding elements to be included
- `statisticsTarget` is a value that is set on the column we are considering. It defaults to 100, but part of this overall mitigation strategy would be to tune that on specific columns, and when doing so we would adjust this parameter to match
- `lambda` is controlling the frequency with which we will insert a padding element, which we intend to tune such that we expect the padding elements to appear with some minimum frequency when Postgres samples the values for query planning

I think the last bit that evades my understanding here is the division by 300 - where did that constant factor come from?

packages/server/src/fhir/repo.test.ts
```diff
@@ -1131,6 +1131,27 @@ describe('FHIR Repo', () => {
       await expect(systemRepo.deleteResource(patient.resourceType, patient.id)).resolves.toBeUndefined();
     }));
 
+  test('deleted rows include array padding', async () =>
```
mattwiller: 👍

mattlong:
> Do we need to add any code that prevents these padding values from being read back? Or is it the case that search these columns are only used to test memberships of the real values and so having this extra data that will never match anything is going to be OK?

@noahsilas  Correct! These array columns (and pretty much all other columns in resource tables) are only used at search time; the contents of the resources returned are pretty much just the JSON object stored in the `content` column verbatim.

packages/server/src/fhir/token-column.ts
mattlong: Correct on `m`, `statisticsTarget`. It would be more accurate to dynamically query for `statisticsTarget` from Postgres to have a single source of truth instead of having to redundantly specify it here, but wanted to keep this as simple as possible for now since it's already an advanced feature. That dynamism would be a good v2 feature.

`lambda` has an inverse relation to the frequency we want elements to appear (`f = λ/30,000`)

The 300 is what Postgres internally multiplies the statistics target by when deciding how many rows to sample that is very poorly if at all documented; I've internalized it at this point. I'm failing to find it in the postgres docs at the moment, but here it is [in the source](https://github.com/postgres/postgres/blob/820343bab30852142ddc50db3aa81ef58d7cb676/src/backend/commands/analyze.c#L1939) and here's [a blog post](https://www.crunchydata.com/blog/indexes-selectivity-and-statistics) that discusses it.

mattlong: I'm going to move a lot of the write up from https://github.com/medplum/medplum/issues/7539 into code comments

nate-watkins:
fascinating- thank you for digging into this! Happen to have a recommendation for a good starting point of what these config options should be set to when experiencing this issue?

mattlong:
> fascinating- thank you for digging into this! Happen to have a recommendation for a good starting point of what these config options should be set to when experiencing this issue?

Hi @nate-watkins, apologies; I missed your comment since my inbox is a flood of Github notifications.

The settings to choose depends on a few variables. All the math I'm about to do is discussed in https://github.com/medplum/medplum/issues/7539.

First off, the shape of your data. Particularly, the average number of array elements per row. For the unique identifier case where each resource has one identifier, e.g. `my-system|abc123`, that results in three array elements per row (in Medplum, the elements are UUIDs instead of text, but text is more useful here):

```sql
['my-system', 'system|abc123', 'abc123']
```

Second off, your statistics target. Our default recommendation for high-cardinality resource types is 1,000 is a good middle ground between the default 100 and the maximum 10,000. Higher values allows array padding to yield increasingly accurate, more selective frequencies.

With these two inputs, `elems_per_row` and `stats_target`, we calculate the Postgres array element cutoff frequency **k**:

```
elems_per_row = 3
stats_target = 1000

sampled_rows = 300 * stats_target = 300000

k = ⌈9 * elems_per_row * sampled_rows / (stats_target * 10 * 1000 / 7)⌉
k = ⌈9 * 3 * 300000 / (1000 * 10 * 1000 / 7)⌉ ≈ ⌈5.67⌉ = 6
```

So our cutoff frequency is 6; which is used as `X` in the Poisson CDF calculation.

We still need to decide on `m` and `lambda`. Before that, we decide how certain we want to be we'll get enough rows with array padding in the sampled rows. We want to be very confident, so we choose `p=0.999999` (six nines).

From here let's just fast-forward a bit and look at a table of approximate minimal `lambda` needed to satisfy `P(X ≤ (k - 1)) ≤ (1-p)^(1/m)` for various `m` where `P(X ≤ (k - 1))` is the cumulative Poisson distribution function.

p | m | P(X ≤ (k - 1)) ≤ (1-p)^(1/m) | lambda | f = lambda / sampled_rows | m * f
-- | -- | -- | -- | -- | --
0.999999 | 5 | 0.0631 | 10.2 | 0.00003400 | 0.00017
0.999999 | 10 | 0.2512 | 7.5 | 0.00002500 | 0.00025
0.999999 | 20 | 0.5012 | 5.7 | 0.00001900 | 0.00038
0.999999 | 40 | 0.7079 | 4.5 | 0.00001500 | 0.00060
0.999999 | 60 | 0.7943 | 4 | 0.00001333 | 0.00080
0.999999 | 80 | 0.8414 | 3.6 | 0.00001200 | 0.00096

How do you decide from here? Let's start by looking at the `f` column, the expected `minfreq` from Postgres array stats calculations. Remember that k, our cutoff frequency is the lowest `f` the Postgres stats system will accept:

```
k / sampled_rows = 6 / 300000 = 0.00002
```

So if we want the stats to be as accurate as possible, we minimize `m` while keeping `f` less than `.00002` which  in this table is `m=20 lambda=5.7`. A selectivity of `.00002` is quite a bit lower than the Postgres default `.005` and in my testing of tables with 200million rows, is sufficient (probably more than sufficient?) to get good, index-based query plans instead of sequential scans. The `m * f` column is the fraction of rows that will have a padding element added. It's useful for estimating how much extra data we're adding to the table. For our chosen row, `m*f = .00038` or 380 extra array elements per million rows which seems like a very reasonable tradeoff to me.

So putting it all together to answer your original question, I'd recommend `m=20, lambda=5.7, statisticsTarget=1000` as a good starting point for the unique identifier scenario. Formatted for Medplum server config:

```json
"arrayColumnPadding": {
  "identifier": {
    "resourceType": [
      "Observation",
      "Patient"
    ],
    "config": {
      "m": 20,
      "lambda": 5.7,
      "statisticsTarget": 1000
    }
  }
}
```

Note that this feature does NOT update the statistics target for you. You'll need to set the statistics target to 1000 (or whatever value you wind up choosing) for all relevant columns yourself. A Medplum super admin UI is available in `@medplum/app`  at `/admin/super/db` to change statistics targets.

mattlong:
For posterity, Postgres queries that can be used to recreate array column padding:

```sql
--DANGER: DO NOT disable autovacuum and reload conf on prod! We do it here to speed up testing
--ALTER SYSTEM SET autovacuum = 0; 
--SELECT pg_reload_conf();

SET statement_timeout = 0;
SHOW statement_timeout;
SHOW autovacuum;

CREATE TABLE public.test(
    id integer,
    tags text[]
);
ALTER TABLE public.test ALTER COLUMN tags SET STATISTICS 1000;

DO $$
DECLARE
  m integer := 20;
  lambda decimal := 5.7;
  stats_target integer := 1000;
BEGIN
  DROP INDEX IF EXISTS test_tags;
  TRUNCATE TABLE public.test;
  INSERT INTO public.test (id, tags)
  SELECT n,
	CASE
    WHEN random() <= (m*lambda::decimal/(300 * stats_target)) THEN  ARRAY['system', 'system|' || TO_CHAR(n,'fm0000000000'), TO_CHAR(n,'fm0000000000'), 'wellknown_' || floor(random()*m)::text]
    ELSE ARRAY['system', 'system|' || TO_CHAR(n,'fm0000000000'), TO_CHAR(n,'fm0000000000')]
	END FROM ( SELECT generate_series(1,1_000_000) AS n );
END $$;

SELECT count(*) from public.test;

SELECT * FROm public.test LIMIT 5;

ANALYZE public.test;
SELECT
  most_common_elem_freqs[array_upper(most_common_elem_freqs,1)-2] as minfreq,
  most_common_elem_freqs[array_upper(most_common_elem_freqs,1)-1] as maxfreq,
  most_common_elem_freqs[array_upper(most_common_elem_freqs,1)-0] as nullfreq,
  array_length(most_common_elems, 1) as mce_count
FROM pg_stats
WHERE tablename = 'test' AND attname in ('tags');

CREATE INDEX test_tags ON public.test USING GIN (tags);
EXPLAIN SELECT * FROM public.test WHERE tags @> '{0000000001}' LIMIT 1;
```

nate-watkins:
>Hi @nate-watkins, apologies; I missed your comment since my inbox is a flood of Github notifications.

@mattwiller no worries at all. Immense appreciation for you and the Medplum team. Thank you much for the comprehensive discovery and solution. Please reach out if I can provide any perspective or information on behalf of other self-hosters in the future.

mattlong:
> > Hi @nate-watkins, apologies; I missed your comment since my inbox is a flood of Github notifications.
> 
> @mattwiller no worries at all. Immense appreciation for you and the Medplum team. Thank you much for the comprehensive discovery and solution. Please reach out if I can provide any perspective or information on behalf of other self-hosters in the future.

Thanks for the kind words, @nate-watkins! If you end up utilizing the array column approach, I'd be super curious to hear about how it goes! I'm at ml@medplum.com.

PR #7576, opened 2025-10-10
tweak(hl7): force connection close on `Hl7Server#stop` after timeout

### LINKED ISSUE #7214: `Hl7Server.stop` hangs forever if client holds connection open

ThatOneBro:
Fixes #7214

### DIFF
--- a/packages/hl7/src/server.ts
+++ b/packages/hl7/src/server.ts
@@ -27,7 +27,7 @@ export class Hl7Server {
   private encoding: string | undefined = undefined;
   private enhancedMode = false;
   private messagesPerMin: number | undefined = undefined;
-  private connections = new Set<Hl7Connection>();
+  private readonly connections = new Set<Hl7Connection>();
 
   constructor(handler: (connection: Hl7Connection) => void) {
     this.handler = handler;


packages/hl7/src/server.ts
```diff
@@ -5,12 +5,19 @@ import net from 'node:net';
 import type { Hl7ConnectionOptions } from './connection';
 import { Hl7Connection } from './connection';
 
+export interface Hl7ServerStopOptions {
+  forceDrainMs?: number;
```
mattlong:
```suggestion
  forceDrainTimeoutMs?: number;
```

packages/hl7/src/server.ts
```diff
@@ -52,20 +63,34 @@ export class Hl7Server {
     this.server = server;
   }
 
-  async stop(): Promise<void> {
+  async stop(options?: Hl7ServerStopOptions): Promise<void> {
```
mattlong, line 89: Add docstring to `stop` and `Hl7ServerStopOptions`?

packages/hl7/src/server.ts
```diff
@@ -52,20 +63,34 @@ export class Hl7Server {
...
     return new Promise<void>((resolve, reject) => {
       if (!this.server) {
         reject(new Error('Stop was called but there is no server running'));
         return;
       }
+      let forceDrainTimeout: NodeJS.Timeout | undefined;
+      if (options?.forceDrainMs !== -1) {
+        forceDrainTimeout = setTimeout(() => {
+          for (const connection of this.connections) {
+            // TODO: Handle this better
```
mattlong: What would be better than calling close?

ThatOneBro: Oops yeah I forgot to unstash a change here 😅  meant that for myself when thinking if there was a better way to handle error than console.logging it. 

But I just added a new clarifying comment now after remembering `Hl7Connection#close` already handles all errors internally by emitting them as events on the `Hl7Connection`. So this `.catch` should theoretically not get hit anyways.

Issue #7544, opened 2025-10-06
Template Extract operation does not handle arrays


MacDaysix:
# Template Extract operation does not handle arrays

We are attempting to extract a blood pressure Observation from a QuestionnaireResponse using the $extract operation.

### Given this Questionnaire:

<details>

```data

{
  "resourceType": "Questionnaire",
  "name": "test-questionnaire",
  "title": "Test Questionnaire",
  "url": "Questionnaire/8ec05fdb-3e5f-4c08-b483-6ceabe3caabc",
  "id": "8ec05fdb-3e5f-4c08-b483-6ceabe3caabc",
  "contained": [
    {
      "resourceType": "Observation",
      "id": "bloodPressureTemplate",
      "status": "final",
      "category": [
        {
          "coding": [
            {
              "system": "http://terminology.hl7.org/CodeSystem/observation-category",
              "code": "vital-signs",
              "display": "Vital Signs"
            }
          ]
        }
      ],
      "code": {
        "text": "Blood Pressure",
        "coding": [
          {
            "system": "http://loinc.org",
            "code": "85354-9",
            "display": "Blood pressure panel with all children optional"
          },
          {
            "system": "http://snomed.info/sct",
            "code": "75367002",
            "display": "Blood pressure (observable entity)"
          }
        ]
      },
      "component": [
        {
          "code": {
            "text": "Diastolic blood pressure",
            "coding": [
              {
                "system": "http://loinc.org",
                "code": "8462-4",
                "display": "Diastolic blood pressure"
              },
              {
                "system": "http://snomed.info/sct",
                "code": "1091811000000102",
                "display": "Diastolic arterial pressure (observable entity)"
              }
            ]
          },
          "valueQuantity": {
            "_value": {
              "extension": [
                {
                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                  "valueString": "%resource.item.where(linkId='blood-pressure-group').item.where(linkId='diastolic-bp').answer.valueInteger"
                }
              ]
            },
            "unit": "mmHg",
            "system": "http://unitsofmeasure.org",
            "code": "mm[Hg]"
          }
        },
        {
          "code": {
            "text": "Systolic blood pressure",
            "coding": [
              {
                "system": "http://loinc.org",
                "code": "8480-6",
                "display": "Systolic blood pressure"
              },
              {
                "system": "http://snomed.info/sct",
                "code": "72313002",
[...]

PR #7584, opened 2025-10-11
[Fix]: make $extract array-aware (preserve multi-valued FHIRPath results)

syp1xd:
# Summary
Make `$extract` array-aware and add tests. Fixes incorrect handling of multi-valued FHIRPath results and context duplication in **SDC Template Extract**.

**Fixes:** https://github.com/medplum/medplum/issues/7544

---

## Problem
`$extract` was effectively treating multi-valued expressions as singletons (implicit `[0]`), causing:

- Only the first value to be written when the path is non-indexed (e.g., `name.given`).
- Incorrect behavior for nested arrays (e.g., `Observation.component[].coding[]`).
- Context extension not duplicating array elements per match.
- Occasional stray primitive underscore fields when expressions yield no results.

---

### TemplateExtract Behavior
- **Preserve multi-valued FHIRPath results.**
- **Non-indexed target path:**
  - 0 results -> remove field  
  - 1 result -> write scalar  
  - more than 1 results -> write array
- **Indexed target path:**
  - Replace/merge array elements safely (no implicit `[0]`).
- **Context extension (`sdc-questionnaire-templateExtractContext`):**
  - Duplicate array elements for each context match and evaluate values per duplicate.
- **Primitive/underscore handling:**
  - Remove corresponding underscore fields when values are set or fields are removed.
  - Avoid invalid `"…value"` paths.

### Tests (packages/server)
- Removes field when expression yields empty.
- Context duplicates array elements when there are multiple matches.
- Existing tests still pass; added assertions improve coverage for arrays and primitives.

### Minor
- FHIRPath parse tests updated where helpful to reflect collection semantics.

---

## Implementation Notes
Core logic remains in `TemplateExtractor/makePatch` with careful handling of:

- Context/value extensions  
- `asJsonPath` / `replacePathIndex` helpers  
- JSON Patch operations being idempotent and consistent

---

### DIFF
--- a/packages/server/src/fhir/operations/extract.test.ts
+++ b/packages/server/src/fhir/operations/extract.test.ts
@@ -762,9 +762,11 @@ describe('QuestionnaireResponse/$extract', () => {
     expect(patient.gender).toBe('unknown');
   });
 
-  test('Value extension yields array for primitive array element', async () => {
-    // Place a value extension on Patient.name._given[0] (primitive array element). The expression yields
-    // multiple strings; server should replace the entire given array with those values.
+  test('Context-driven duplication for primitive array element', async () => {
+    // Place a context extension on HumanName to iterate over each "given" answer and
+    // a value extension on Patient.name._given[0] (primitive array element). The expression
+    // should be evaluated per-context (per given item), and the server should replace the
+    // entire given array with those values.
     const questionnaire: Questionnaire = {
       resourceType: 'Questionnaire',
       status: 'unknown',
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/operations/extract.test.ts
```diff
@@ -761,4 +761,197 @@ describe('QuestionnaireResponse/$extract', () => {
...
+          name: [
+            {
+              _given: [
+                {
+                  extension: [
+                    { url: valueExtension, valueString: "item.where(linkId = 'given').answer.value" },
+                  ],
+                } as any,
+              ],
+            },
```
mattwiller, line 785: This test case isn't quite correct: it's missing a `contextExtension` on the `HumanName` to direct the creation of the top-level data element.  This is likely the cause of the erroneous behavior you're observing.  Something like this should work:
```suggestion
            {
            extension: [{ url: contextExtension, valueString: `item.where(linkId = 'given')` }],
              _given: [
                {
                  extension: [{ url: valueExtension, valueString: 'answer.value' }],
                } as any,
              ],
            },
```

mattwiller:
Hey @psyp1x — it looks like we're [working in parallel](https://github.com/medplum/medplum/pull/7599) on the same problem, which I think is going to be a bit more complex than initially anticipated.  If you don't mind, I'll try to close this out in that PR

PR #7586, opened 2025-10-13
Dependency upgrades 2025-10-13

medplumbot:
Dependency upgrades


mattwiller:
The alleged [license policy violation](https://github.com/medplum/medplum/pull/7586#issuecomment-3396556548) is a false positive: that library is dual licensed under MIT and the invalid license: https://github.com/zone-eu/mailsplit?tab=readme-ov-file#license

PR #7622, opened 2025-10-17
fix(ci): add NSIS to PATH when building Windows agent

### LINKED ISSUE #7621: Fix Agent build for Windows
The last two versions of the agent have not been published at all for Windows.
We need to either republish those versions or cut a new release after fixing the build.

We might want to somehow alert on failure of any of the agent builds. This is not the first time that there have been a few versions that published without a Windows agent version.

ThatOneBro:
Fixes #7621 

Basically Chocolatey's NSIS package doesn't actually add `makensis` to PATH automatically...
We've been relying on the `makensis` in the PATH on the default `windows-latest` image, which previously pointed to the `windows-2022` image until very recently. After the `windows-latest` tag changed to `windows-2025`, NSIS is no longer in PATH by default on the base image, hence why we are seeing breakages.

The fix is to just manually add NSIS to PATH since we are already installing it ourselves.

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -14,10 +14,10 @@ jobs:
     runs-on: windows-latest
     timeout-minutes: 45
     env:
-      NODE_VERSION: "20"
+      NODE_VERSION: '20'
       TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review.  Is there a difference between single quotes and double quotes in YAML?

PR #7635, opened 2025-10-18
S3 Deploy util

codyebberson:
As our team has grown and our application footprint has expanded, the existing Vercel preview strategy has become a bottleneck due to concurrency limits and increasing cost. Furthermore, consolidating deployment infrastructure into AWS simplifies our compliance and documentation burden.

This change achieves two main goals:

1. Migrates all PR Previews to a unified AWS S3/CloudFront solution.
2. Replaces Bash/AWS CLI deployment scripts with a faster Node.js utility.

### Unified AWS Preview Architecture

This PR uses a new preview deployment target within our AWS "dev" environment, moving away from Vercel's scaling constraints.

#### Implementation Details:

* **Target:** All previews now deploy to the new `preview.medplum.dev` domain, served by a single S3 bucket and a single CloudFront distribution.  
* **Intelligent Routing with Lambda@Edge:** A new Lambda@Edge function is configured to intercept requests and rewrite the URL path. This allows us to use a single CloudFront distribution for all preview subdomains.  
  * *Example:* A request to `my-pull-request.preview.medplum.dev` is internally rewritten to hit the S3 object path `/previews/my-pull-request/index.html`.  
* **Automatic Cleanup:** An **S3 Lifecycle Policy** has been applied to the bucket, ensuring that all preview artifacts are automatically expired and deleted after **90 days**, preventing storage creep.  
* **CI/CD Updates:** Github Actions workflows have been updated to target this new S3 bucket for all static app builds.

### Optimized S3 Deployment Utility

The previous deployment method, which relied on Bash scripts executing multiple sequential `aws s3 cp` commands, was inefficient and non-reusable.

#### Implementation Details:

* **Node.js / AWS SDK (`lib-storage`):** We now use a new centralized Node.js script that leverages the AWS SDK, specifically the @aws-sdk/lib-storage package.  
* **Fast Parallelization:** The new script enables rapid, multi-threaded parallel uploads, resulting in **significantly faster deployment times** for all S3-deployed apps.  
* **Code Reusability:** The new utility is designed to be easily incorporated into all of our static app build processes, eliminating copy/paste of complex Bash logic and ensuring consistent, efficient deployment across the platform.

### Benefits

| Category | Impact |
| :---- | :---- |
| **Scalability** | Removes the Vercel concurrency bottleneck, supporting our growing team and increasing PR volume without interruption. |
| **Cost & Efficiency** | Reduces Vercel expenditure and improves computational efficiency by performing faster, optimized S3 uploads and consolidating CloudFront resources. |
| **Compliance** | Simplifies our vendor stack and makes documentation easier by consolidating pre-production infrastructure entirely within AWS. |
| **Deployment Speed** | The new Node.js utility with parallelization drastically cuts deployment time for the final S3 sync step. |

### Rollout Plan

The existing Vercel preview deployments are currently running **in parallel** with the new AWS deployments. This allows us to test the new system thoroughly and ensure stability before fully decommissioning the Vercel deployment jobs.


scripts/cicd-deploy.sh
```diff
@@ -125,32 +133,22 @@ curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$SLACK_WEBHO
 
 if [[ "$DEPLOY_APP" = true ]]; then
   echo "Deploy app"
-
-  # We create a subshell for the build since we need to set the env vars to our placeholders for the Docker build
```
ThatOneBro, line 138: There was a very deliberate reason why this was done here, I would like to remember what the implications are of doing it above in a separate step

I think it had to do with the fact if you are going to deploy app, you want to force a rebuild of all of its dependencies so that core has a matching `MEDPLUM_VERSION`...

Before this change, `@medplum/core` and hence `MEDPLUM_VERSION` would be reused with an old SHA hash because nothing within `@medplum/core` might have changed but the hash of the commit would still be different if something in `@medplum/app` changed, for example. 

So the hash in the version wouldn't match what was expected unless something also changed in core without the `--force` here.

This was particularly problematic for comparing agent versions, which would often evaluate to being different than the current version even when on the latest, because the hash in `MEDPLUM_VERSION` was stale.

scripts/cicd-deploy.sh
```diff
@@ -125,32 +133,22 @@ curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$SLACK_WEBHO
...
+
 if [[ "$DEPLOY_GRAPHIQL" = true ]]; then
   echo "Deploy GraphiQL"
-  npm run build -- --force --filter=@medplum/graphiql
   source ./scripts/deploy-graphiql.sh
 fi
 
 if [[ "$DEPLOY_SERVER" = true ]]; then
   echo "Deploy server"
-  npm run build -- --force --filter=@medplum/server
```
ThatOneBro, line 162: These forces were also done for similar reasons, you wouldn't want to have a stale version built into your server Docker image I think?

scripts/cicd-deploy.sh
```diff
@@ -125,32 +133,22 @@ curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$SLACK_WEBHO
...
 fi
 
+if [[ "$DEPLOY_DOCS" = true ]]; then
+  echo "Deploy docs"
+  source ./scripts/deploy-docs.sh
+fi
+
 if [[ "$DEPLOY_GRAPHIQL" = true ]]; then
   echo "Deploy GraphiQL"
-  npm run build -- --force --filter=@medplum/graphiql
```
ThatOneBro, line 156: `graphiql` might be less defensible for the force? 🤷

ThatOneBro:
Arguably this probably just means we need to add the git commit hash to `@medplum/core`'s `turbo.json` env  ...  🤔

scripts/cicd-deploy.sh
codyebberson: > There was a very deliberate reason why this was done here, I would like to remember what the implications are of doing it above in a separate step

Yeah, I remember you adding it, I went through this quite a few times.  

> I think it had to do with the fact if you are going to deploy app, you want to force a rebuild of all of its dependencies so that core has a matching MEDPLUM_VERSION...

Hm, ok, good call out.  I feel like that should be handled in `turbo.json` as an environment change.  But then every commit would invalidate the Turborepo cache 🤔 

Or, if we're always going to do a `--force`, should we move it to the `build` step in `build.yml` ?  Because we're going to do a force build anyway?

Do old SHA hashes matter outside of published releases?

My goal here is simply to keep the build process tight, and avoid running builds more times than we have to.  If we know that we're going to need to do a force build on every CI/CD run, then we might as well just do it once up front.

codyebberson:
> need to add the git commit hash to `@medplum/core`'s `turbo.json` env

Agree that would fix it, but that would effectively kill the majority of Turborepo cache opportunities.  Each commit would invalidate `core` and anything that depends on `core` (which is almost everything).

Maybe each package should have its own `MEDPLUM_VERSION` logic in each build script?  Then you would get the good caching behavior and accurate version strings?

For example, `app` should have `MEDPLUM_APP_VERSION`, `server` should have `MEDPLUM_SERVER_VERSION`, etc?

codyebberson:
Maybe the simplest option to preserve your behavior would be to change this line:

https://github.com/medplum/medplum/blob/main/.github/workflows/deploy.yml#L50

From:

```yaml
        run: npm run build:fast
```

To:

```yaml
        run: npm run build:fast -- --force
```

??

I added that change to the PR.

scripts/cicd-deploy.sh
ThatOneBro: It matters for agent version checks in `@medplum/app` since we publish on every commit to main right?

ThatOneBro:
> ```yaml
>         run: npm run build:fast -- --force
> ```

I think that works? It doesnt mean a cache bust on every deploy but I think that was already happening in a majority of cases

Does `build:fast` actually build `graphiql` though?

codyebberson:
> Does `build:fast` actually build `graphiql` though?

Good point.  No, it does not.

```json
"build:fast": "turbo run build --filter=@medplum/app --filter=@medplum/server",
```

We could change it to simply `build`:

```yaml
run: npm run build -- --force
```

Or if we wanted to be surgical we could do:

```yaml
run: npm run build -- --force --filter=@medplum/app --filter=@medplum/graphql --filter=@medplum/server
```

That's getting brittle though 🤔   Maybe `npm run build -- --force` is the simple option

Or we could remove the `build` step from YAML entirely and do it all in `cicd-deploy.sh`?

ThatOneBro:
> Or if we wanted to be surgical we could do:
> 
> ```yaml
> run: npm run build -- --force --filter=@medplum/app --filter=@medplum/graphql --filter=@medplum/server
> ```
I kinda like this one... especially since it excludes docs. If we need to build docs we can do so JIT in the `DEPLOY_DOCS` block

>
> That's getting brittle though 🤔 Maybe `npm run build -- --force` is the simple option
> 
That's true, but building docs every time will definitely slow things down for a deploy. Docs should definitely only be built on deploy for docs

> Or we could remove the `build` step from YAML entirely and do it all in `cicd-deploy.sh`?

We could but I think we would need to approach it similar to my solution with the subshell at least for the app deploy. I think there was something funky about env var clobbering or something like that. It's hard to remember why that was happening. We could just try it and see I guess

scripts/cicd-deploy.sh
```diff
@@ -125,32 +133,22 @@ curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$SLACK_WEBHO
...
-    export GOOGLE_CLIENT_ID="__GOOGLE_CLIENT_ID__"
-    export RECAPTCHA_SITE_KEY="__RECAPTCHA_SITE_KEY__"
-    npm run build -- --force --filter=@medplum/app
-  )
-
   source ./scripts/build-docker-app.sh
   source ./scripts/deploy-app.sh
 fi
 
+if [[ "$DEPLOY_DOCS" = true ]]; then
```
ThatOneBro: I think I would recommend we do this at the very bottom

That way docs doesn't block getting something deployed on prod

scripts/cicd-deploy.sh
```diff
@@ -125,32 +133,22 @@ curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$SLACK_WEBHO
...
+  echo "Deploy docs"
+  source ./scripts/deploy-docs.sh
```
ThatOneBro:
```suggestion
  npm run build --filter=@medplum/docs
  source ./scripts/deploy-docs.sh
```
Think we might need to add this here

codyebberson:
@claude please review.  is this a good idea or horrible idea?

scripts/cicd-deploy.sh
```diff
@@ -125,32 +133,23 @@ curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$SLACK_WEBHO
...
   echo "Deploy server"
-  npm run build -- --force --filter=@medplum/server
   source ./scripts/build-docker-server.sh
   source ./scripts/deploy-server.sh
 fi
+
+# Deploy docs last since it is the slowest
+if [[ "$DEPLOY_DOCS" = true ]]; then
+  echo "Deploy docs"
+  source ./scripts/deploy-docs.sh
```
ThatOneBro, line 177: I think we might still need to add `npm run build -- --filter=@medplum/docs` (probably no need for force) here for the deploy case for a docs change

ThatOneBro (approved):

PR #7637, opened 2025-10-19
[v5] Upgrade to eslint 9

codyebberson:
Upgrades ESLint from 8 to 9.  ESLint 9 had a pretty major set of changes and includes a pretty big overhaul of how config changes work.  We have tried this upgrade a few times over the past year, and each time got stuck on plugin incompatibilities.  This time the upgrade appears to have worked successfully.  This does impact how projects consume `@medplum/eslint-config`.  Using `eslintConfig` from inside `package.json` is no longer supported.  Packages now must have a `eslint.config.js` or `eslint.config.mjs` in the package.  Other than that, most of the actual ESLint rules are still the same.

--------

Some fun discoveries from this upgrade:

```ts
    const mockDispatchEvent = jest.spyOn(ReconnectingWebSocket.prototype, 'dispatchEvent').mockImplementation(function (
      this: ReconnectingWebSocket,
      event: Event
    ) {
      // ...
      // eslint-disable-next-line no-invalid-this
      originalDispatchEvent.call(this, event);
    });
```

This disable comment is no longer necessary because ESLint is now smart enough to figure out that `this` is actually valid in this case.

```ts
  // eslint-disable-next-line no-constant-condition
  while (true) {
    // ...
    if (validationFunc(answer)) {
      return answer;
    }
  }
```

This disable comment is no longer necessary because the [new default](https://eslint.org/docs/latest/rules/no-constant-condition) is "Disallow constant expressions in all loops except while loops with expression true."

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,0 +1,5 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import { medplumEslintConfig } from '@medplum/eslint-config';
+import { defineConfig } from 'eslint/config';
+export default defineConfig(medplumEslintConfig);
--- a/examples/foomedical/package.json
+++ b/examples/foomedical/package.json
@@ -15,11 +15,6 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "eslintConfig": {
-    "extends": [
-      "@medplum/eslint-config"
-    ]
-  },
   "devDependencies": {
     "@babel/core": "7.28.5",
     "@babel/preset-env": "7.28.5",
--- a/examples/medplum-chart-demo/package.json
+++ b/examples/medplum-chart-demo/package.json
@@ -19,11 +19,6 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "eslintConfig": {
-    "extends": [
-      "@medplum/eslint-config"
-    ]
-  },
   "devDependencies": {
     "@mantine/core": "7.17.8",
     "@mantine/hooks": "7.17.8",
--- a/examples/medplum-chat-demo/package.json
+++ b/examples/medplum-chat-demo/package.json
@@ -15,11 +15,6 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "eslintConfig": {
-    "extends": [
-      "@medplum/eslint-config"
-    ]
-  },
   "devDependencies": {
     "@mantine/core": "7.17.8",
     "@mantine/hooks": "7.17.8",
--- a/examples/medplum-demo-bots/package.json
+++ b/examples/medplum-demo-bots/package.json
@@ -21,15 +21,6 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "eslintConfig": {
-    "parserOptions": {
-      "project": "./tsconfig.json"
-    },
-    "extends": [
-      "@medplum/eslint-config"
-    ],
-    "root": true
-  },
   "devDependencies": {
     "@medplum/bot-layer": "4.5.2",
     "@medplum/cli": "4.5.2",
--- a/examples/medplum-eligibility-demo/package.json
+++ b/examples/medplum-eligibility-demo/package.json
@@ -19,11 +19,6 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "eslintConfig": {
-    "extends": [
-      "@medplum/eslint-config"
-    ]
-  },
   "devDependencies": {
     "@mantine/core": "7.17.8",
     "@mantine/hooks": "7.17.8",
--- a/examples/medplum-fhircast-demo/package.json
+++ b/examples/medplum-fhircast-demo/package.json
@@ -10,11 +10,6 @@
     "lint:fix": "eslint . --fix",
     "preview": "vite preview"
   },
-  "eslintConfig": {
-    "extends": [
-      "@medplum/eslint-config"
-    ]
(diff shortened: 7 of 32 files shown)


examples/medplum-provider/src/pages/SchedulePage.tsx
```diff
@@ -347,7 +347,7 @@ export function SchedulePage(): JSX.Element | null {
 
 // This function collapses contiguous or overlapping slots of the same status into single events
 function slotsToEvents(slots: Slot[]): SlotEvent[] {
-  if (!slots || slots.length === 0) {
+  if (slots?.length === 0) {
```
mattlong: Hmm, isn't the original version correct if slots can be `undefined`? In which case, the function signature should be changed to `slots?: Slot[]`. If it cannot be undefined, this should just be `slots.length === 0`.

packages/core/src/fhirmapper/transform.ts
```diff
@@ -388,7 +388,7 @@ function evalSource(ctx: TransformContext, source: StructureMapGroupRuleSource):
   }
 
   let sourceValue = evalFhirPathTyped(sourceElement, [sourceContext]);
-  if (!sourceValue || sourceValue.length === 0) {
+  if (sourceValue?.length === 0) {
```
mattlong: Same comment as above, `sourceValue` being undefined no longer passes this if statement.

packages/core/src/fhirpath/utils.ts
```diff
@@ -134,7 +134,7 @@ export function getTypedPropertyValueWithSchema(
 
   const value = typedValue.value;
   const types = element.type;
-  if (!types || types.length === 0) {
+  if (types?.length === 0) {
```
mattlong: Same here regarding the undefined case.

packages/server/src/cloud/aws/textract.ts
```diff
@@ -45,7 +45,7 @@ export async function awsTextractHandler(req: FhirRequest): Promise<FhirResponse
   } else {
     // DocumentReference
     const inputDocRef = await repo.readResource<DocumentReference>('DocumentReference', id);
-    if (!inputDocRef.content || inputDocRef.content.length === 0) {
+    if (inputDocRef.content?.length === 0) {
```
mattlong: Same here for `undefined`

mattlong (changes requested):

packages/core/src/fhirpath/utils.ts
techdavidy: mm true

const types = undefined;
//1 case: TRUE
//2 case: FALSE

examples/medplum-provider/src/pages/SchedulePage.tsx
codyebberson: Huh, interesting. You're totally correct.... this appears to be a bug in ESLint's auto-fix?  Because I didn't make this change 😬

codyebberson: I think this could work:

```ts
if (!slots?.length) {
```

Still concerning that auto-fix changed the behavior.

codyebberson: Hm.  I wasn't able to repro that in the ESLint Playground: https://typescript-eslint.io/play/#ts=5.9.2&fileType=.ts&code=KYDwDg9gTgLgBAMwK4DsDGMCWEWIhACgEMooiBPALjgGcYpMUBzAbQF04AfOVAE2ASNgvAJTUARvgA2wIrgDeAKDhxMCOAQCEJMuS7cdFAHQzmMABZwAvDbgAGEXCUqVUYDCRRc9JMADcynAAvoFuHl6IRFI0-opBQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJGUQAusoC%2BIfUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

There were a few minor patch upgrades since I originally set this up.  I'm going to revert all `.ts` and `.tsx` changes.  

Good catch @mattlong

codyebberson:
@claude - please review.  Please note @mattlong's observations about how some of the ESLint fixes changed the logical behavior.  After cleaning out some unused dependencies, upgrading all ESLint dependencies to the latest version, and reinstalling all dependencies, those issues appear to have been fixed.  However, I feel like we should be extra vigilant about all changes in `.ts` and `.tsx` files to ensure that the changes are purely semantic, and that they did not change the behavior.

codyebberson:
@claude please double check the logic

You said:

> Example 1 - packages/agent/src/app.ts:639
> 
> ```
> - if (msg && msg.type === 'agent:transmit:response' && msg.channel) {
> + if (msg?.type === 'agent:transmit:response' && msg.channel) {
> ```
>
> Problem: Original checks for msg truthiness, new version allows falsy msg values through if they have the right properties.

I believe these are logically equivalent.  If `msg` is `undefined`, then this check will correctly fail, because `undefined !== 'agent:transmit:response'`, right?

You said:

> Example 2 - packages/core/src/typeschema/crawler.ts:260
> 
> ```
> - } else if (options?.withPath && current && current.value !== undefined) {
> + } else if (options?.withPath && current?.value !== undefined) {
> ```
> 
> Problem: Original ensures current is truthy before checking properties, new version only checks the property.

If `current` is `undefined`, then this check will correctly fail, because `undefined === undefined`, right?

PR #7643, opened 2025-10-21
fix(core/subscriptions): eval both `meta.accounts` and `meta.account`

### LINKED ISSUE #7616: Support both `meta.accounts` and `meta.account` filtering during Subscription evaluation
In the past, account was a single reference and the check for agreement between the `meta.account` for the `Subscription` and the evaluated resource was a valid way to enforce multi-tenant isolation within a project. However, we now support the plural `meta.accounts` which is an array and `meta.account` is populated by the first entry in `meta.accounts`.

On top of that, we also support using the `_compartment` search parameter, which includes all `meta.accounts` references. This is now the canonical way to do tenant isolation.

This is a breaking change, but we should use the upcoming v5 release to remove this functionality, as it is already only used by 1 or 2 customers that we know of based on the logs we added in #7561.

EDIT: Originally we were planning to remove `meta.account` filtering altogether, however we decided that it may be easier to  normalize the `meta.account` and `meta.accounts` first, using an existing util `extractAccountReferences` from `@medplum/server` that we move to `@medplum/core`.

ThatOneBro:
Fixes #7616 and #7560

### DIFF
--- a/packages/core/src/subscriptions/index.test.ts
+++ b/packages/core/src/subscriptions/index.test.ts
@@ -1213,105 +1213,110 @@ describe('resourceMatchesSubscriptionCriteria', () => {
   });
 
   describe('Account matching logic', () => {
+    const ORGANIZATION_ONE = { reference: 'Organization/123' };
+    const ORGANIZATION_TWO = { reference: 'Organization/456' };
+    const ORGANIZATION_THREE = { reference: 'Organization/789' };
+    const ORGANIZATION_FOUR = { reference: 'Organization/999' };
+
     test.each([
       {
         description:
           'should return true when subscription has meta.account and resource has matching account in meta.accounts even if meta.account differs',
-        subscriptionMeta: { account: { reference: 'Organization/123' } },
+        subscriptionMeta: { account: ORGANIZATION_ONE },
         resourceMeta: {
-          account: { reference: 'Organization/456' },
-          accounts: [{ reference: 'Organization/456' }, { reference: 'Organization/123' }],
+          account: ORGANIZATION_TWO,
+          accounts: [ORGANIZATION_TWO, ORGANIZATION_ONE],
         },
         shouldMatch: true,
       },
       {
         description:
           'should return true when subscription has meta.accounts and resource has matching account in meta.accounts',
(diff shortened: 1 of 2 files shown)


packages/core/src/subscriptions/index.test.ts
```diff
@@ -1246,7 +1246,691 @@ describe('resourceMatchesSubscriptionCriteria', () => {
       logger: new Logger((msg) => log(msg)),
     });
 
-    expect(matches).toStrictEqual(false);
-    expect(log).toHaveBeenCalledWith(expect.stringContaining('Subscription suppressed due to mismatched meta.account'));
+    expect(matches).toStrictEqual(true);
```
ThatOneBro: This is our case that was previously failing in the repro

packages/core/src/subscriptions/index.ts
```diff
@@ -553,13 +553,22 @@ export async function resourceMatchesSubscriptionCriteria({
...
+  const resourceAccounts = extractAccountReferences(resource.meta) ?? [];
+
+  if (subscriptionAccounts.length) {
+    // Check if there is any common account between the subscription and the resource
+    if (
+      !subscriptionAccounts.some((subAccount) =>
+        resourceAccounts.some((resAccount) => resAccount.reference === subAccount.reference)
+      )
+    ) {
+      logger?.debug('Subscription suppressed due to mismatched accounts', {
```
mattwiller, line 566: Do we not want to see this error by default?  Otherwise it should be INFO

packages/core/src/subscriptions/index.ts
```diff
@@ -553,13 +553,22 @@ export async function resourceMatchesSubscriptionCriteria({
...
+        subscriptionId: subscription.id,
+        resourceId: resource.id,
```
mattwiller:
```suggestion
        resource: getReferenceString(resource),
```

packages/core/src/subscriptions/index.ts
ThatOneBro: Yeah I think at this point if we get this error it is probably correct that it was suppressed, but it may be useful for debugging locally

I was mainly going with the logic of the other messages in this function which are already debug level

PR #7652, opened 2025-10-21
Fixes #7608 - google profile pictures

### LINKED ISSUE #7608: Google Auth - auto populate profile resource photo
### Description
When a User logs in with google auth for the first time after being invited, the User's profile resource photo should be populated by their gmail photo.

This should be an opt-in feature, as part of the way that the User is invited.

codyebberson:


packages/server/src/oauth/utils.ts
```diff
@@ -429,6 +432,23 @@ export async function setLoginMembership(login: Login, membershipId: string): Pr
     throw new OperationOutcomeError(badRequest('Google authentication is required'));
   }
 
+  // Optionally update the profile picture from Google
+  if (
+    login.authMethod === 'google' &&
+    login.pictureUrl &&
+    project.setting?.find((s) => s.name === 'google-auth-profile-pictures' && s.valueBoolean)
```
ThatOneBro: Should we camel case this rather than kebab-case it, since it's a setting and not a feature?

codyebberson: Yep, my bad, I started with it as a "feature", but if it's a setting, it should be kebab.  Will fix.

ThatOneBro:
```suggestion
    project.setting?.find((s) => s.name === 'googleAuthProfilePictures' && s.valueBoolean)
```

packages/server/src/auth/google.test.ts
```diff
@@ -234,6 +234,51 @@ describe('Google Auth', () => {
...
+        email,
+        password,
+      });
+      state.profile = profile as Practitioner;
+
+      // As a super admin, update the project to require Google auth
+      const systemRepo = getSystemRepo();
+      await systemRepo.updateResource({
+        ...project,
+        setting: [{ name: 'google-auth-profile-pictures', valueBoolean: true }],
```
ThatOneBro:
```suggestion
        setting: [{ name: 'googleAuthProfilePictures', valueBoolean: true }],
```

ThatOneBro (commented):
I added the suggested changes 👀

ThatOneBro (approved):
👍

PR #7655, opened 2025-10-21
Use project references in server and related packages tsconfig.json

mattlong:
Previously, jump to definition across package lines was often very slow; due to the TS/JS lang server needing to repeatedly load a different context (i.e. tsconfig) for the package being switched to. After a bit of digging, I _think_ this is the correct fix? So far, this seems to be addressing the slowness/reloading/re-initializaiton for me.

https://www.typescriptlang.org/docs/handbook/project-references.html

### DIFF
--- a/packages/definitions/package.json
+++ b/packages/definitions/package.json
@@ -35,7 +35,7 @@
   ],
   "scripts": {
     "build": "npm run clean && tsc --project tsconfig.build.json",
-    "clean": "rimraf dist/index.js dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


mattlong:
@claude please review. Explain the implications of introduction of `references`, `composite`, etc. in the various tsconfig files.

mattlong:
Getting "Go to Source Definition" to work as desired is gated on https://github.com/microsoft/rushstack/issues/1886

PR #7660, opened 2025-10-22
Local testing signing key cleanup

codyebberson:
This PR addresses a security concern by **removing all hardcoded example signing keys** from sample configuration files (e.g., `docker-compose.full-stack.yml`) and implementing a new, secure default behavior for local development.

This change ensures that no secret keys are exposed in our public repository, while maintaining a smooth, out-of-the-box experience for developers.

### Key Changes

1.  **Removed Hardcoded Keys:** The `MEDPLUM_SIGNING_KEY` and related environment variables have been removed from all sample configurations. This eliminates the risk of these keys being accidentally used in a production environment and resolves the publicly reported security exposure.

2.  **Implemented Auto-Generated Default Key:** When a signing key is required but not explicitly configured, the Medplum server will now **generate a temporary, in-memory signing key on startup**.

3.  **Clear Logging and Warnings:** When the temporary key is used, the server logs a clear message detailing the **limitations** of this approach:
    * It is **not persistent** (keys won't survive restarts).
    * It **will not work in a clustered environment** (keys are not shared).
    * It is intended **only for local development**.

4.  **Updated Public Documentation:** The official documentation for [Presigned URLs] has been completely updated to:
    * Explain the security context of signing keys.
    * Detail the new auto-generation behavior for local development.
    * Provide clear instructions and best practices for configuring a **secure, persistent key** for production environments.

### Security and Usability Rationale

This strategy provides the best balance of security and usability:

* **Security:** No secrets are stored in the repository.
* **Usability:** Features like FHIR Binary pre-signed URLs continue to work immediately after cloning and starting the repo, without requiring developers to manually generate keys just to test locally.
* **Best Practice:** The logged warnings actively guide developers toward the correct, persistent key management setup required for production or clustered environments.


codyebberson:
@claude please review

ThatOneBro:
cc @ianplunkett do you remember when we had to switch to double quotes for some reason in the docker compose file?

codyebberson:
> cc @ianplunkett do you remember when we had to switch to double quotes for some reason in the docker compose file?

Ah, k, interesting. We should update the prettier config if there's a good reason.

ThatOneBro:
@codyebberson actually I think it's fine, it was specifically related to the escaping in the signing key:
https://github.com/medplum/medplum/pull/6759
Which we remove in this PR

PR #7668, opened 2025-10-23
Support token exchange with multiple memberships in same project

codyebberson:
The OAuth token exchange flow (`exchangeExternalAccessToken`) fails when a user has multiple ProjectMemberships in the same project. This is because:
1. When multiple memberships exist, `login.membership` is left undefined (by design, to prompt selection)
2. `sendTokenResponse()` assumes `membership` is always defined
3. There's no way for OAuth clients to discover or switch between available memberships

### Why we can't reuse the normal `/auth/login` flow
Our standard `/auth/login` flow handles multiple memberships by returning a list of available profiles and prompting the user to select one before completing authentication. However, the token exchange flow is based on [RFC 8693 OAuth 2.0 Token Exchange](https://datatracker.ietf.org/doc/rfc8693/), which imposes specific constraints:

- Token exchange is a direct, single request/response flow to the token endpoint
- It's designed for programmatic, server-to-server token exchanges without user interaction
- There's no provision for multi-step flows, redirects, or user agent involvement
- The response must conform to the OAuth 2.0 token response format

These constraints mean we cannot pause mid-flow for user input like we do with the traditional login flow.

### Solution
This PR implements a pragmatic solution for OAuth clients that need to handle multiple ProjectMemberships:

1. **Fix token exchange with multiple memberships**: When multiple memberships exist in the same project, default to the first one instead of failing. This ensures the OAuth flow completes successfully and remains RFC 8693 compliant.

2. **Add new property to `/auth/me` endpoint**: New property includes all ProjectMemberships for the authenticated user in the current project. This allows OAuth clients to:
   - Discover available memberships after authentication
   - Build their own organization/profile switcher UI
   - Initiate new auth flows for specific memberships as needed

### DIFF
--- a/packages/server/src/auth/me.ts
+++ b/packages/server/src/auth/me.ts
@@ -69,7 +69,7 @@ export async function meHandler(req: Request, res: Response): Promise<void> {
       memberships: memberships.map((membership) => ({
         resourceType: 'ProjectMembership',
         id: membership.id,
-        project: membership.project,
+        identifier: membership.identifier,
         profile: membership.profile,
         admin: membership.admin,
       })),


codyebberson:
@claude please review

packages/server/src/auth/memberships.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export async function membershipsHandler(req: Request, res: Response): Promise<void> {
+  const systemRepo = getSystemRepo();
+  const { authState } = getAuthenticatedContext();
+  const { login } = authState;
+  const userRef = login.user as Reference<User>;
+
+  const memberships = await systemRepo.searchResources<ProjectMembership>({
+    resourceType: 'ProjectMembership',
+    filters: [
```
mattlong: Claude's suggestion to filter by project seems valid since `authState.project` is set.

packages/server/src/auth/memberships.ts
```diff
@@ -0,0 +1,39 @@
...
+      {
+        code: 'user',
+        operator: Operator.EQUALS,
+        value: userRef.reference as string,
+      },
+    ],
+  });
+
+  const result = {
+    memberships: memberships.map((membership) => ({
```
mattlong: Assuming we don't limit to `authState.project`, this seems like a lot of info to be returning about memberships to other projects?

codyebberson:
@claude please review

packages/server/src/auth/me.ts
```diff
@@ -47,9 +48,31 @@ export async function meHandler(req: Request, res: Response): Promise<void> {
...
     security = {
       mfaEnrolled: !!user.mfaEnrolled,
       sessions,
+      memberships: memberships.map((membership) => ({
+        resourceType: 'ProjectMembership',
+        id: membership.id,
+        project: membership.project,
+        profile: membership.profile,
+        admin: membership.admin,
+      })),
```
jeffrey-peterson-vanna, line 75: Is it possible to include the `identifiers` here as well?

For ProjectMemberships that share a profile within a project, there’s no way to differentiate them based on this data alone. If the identifiers were included, I could label the ProjectMemberships with what they are used for in the project like “org A” vs “org B”. Thanks in advance.

PR #7670, opened 2025-10-23
Allow specifying autoMinorVersionUpgrade on RDS cluster

mattlong:


packages/core/src/config.ts
```diff
@@ -52,6 +52,7 @@ export interface MedplumSourceInfraConfig {
   rdsReaderInstanceType?: ValueOrExternalSecret<string>;
   rdsProxyEnabled?: ValueOrExternalSecret<boolean>;
   rdsClusterParameters?: StringMap;
+  rdsAutoMinorVersionUpgrade?: boolean;
```
codyebberson:
```suggestion
  rdsAutoMinorVersionUpgrade?: ValueOrExternalSecret<boolean>;
```

codyebberson: Sorry to be annoying on this.  Some customers load 100% of their config settings from external secrets.

This is something that should get massively simplified and cleaned up if/when we move to Terraform, k8s, and Helm charts for everything, because they have more built-in primitives for these things.

PR #7671, opened 2025-10-24
feat(agent): add `Hl7ClientPool`; enable `maxClientsPerRemote`

### LINKED ISSUE #7615: [Agent] Add a `maxConnectionsPerTarget` setting
This will limit the max connections per target when pushing messages to a remote target (Agent -> Device).

ThatOneBro:
Fixes #7615

## TODO
- [x] Make sure that reloaded changes are propagated to the clients in an existing pool, especially when keepAlive is not changed between reloads

### DIFF
--- a/packages/agent/src/app.test.ts
+++ b/packages/agent/src/app.test.ts
@@ -3635,8 +3635,8 @@ describe('App', () => {
       expect(client?.stats).toBeDefined();
       expect(client?.stats?.getSampleCount()).toBe(1);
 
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Actually now that I'm considering the high throughput scenario with a single `keepAlive` client, the current implementation here will reduce outstanding messages on a single connection (pipelining) to `1`.

We should probably have the default pipelining be higher than one and also make that configurable so that clients can put out more than one message at a time.

Going to tweak this before merging.

PR #7680, opened 2025-10-25
fix(agent): actually close hl7 clients on stop

ThatOneBro:
Looks like there was a bug introduced by some copy-pasta that is not well-exercised in the wild since `App#stop` is mostly used in tests

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -576,8 +576,8 @@ export class App {
 
     if (this.hl7Clients.size !== 0) {
       const clientClosePromises = [];
-      for (const channel of this.channels.values()) {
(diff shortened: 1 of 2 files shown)


The-Alchemist (approved):

packages/agent/src/app.ts
```diff
@@ -576,8 +576,8 @@ export class App {
 
     if (this.hl7Clients.size !== 0) {
       const clientClosePromises = [];
-      for (const channel of this.channels.values()) {
```
codyebberson, line 579: Don't we still need to close the channels?

codyebberson (approved):

PR #7681, opened 2025-10-25
fix(deps): re-pin `@swc/core` to `1.13.5`

ThatOneBro:
Looks like `latest` tag is now pointing back at `1.13.5` for now. We should revert to match the new `latest`.

https://github.com/swc-project/swc/issues/11129

This will enable agent build for arm64 to succeed in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   },
   "overrides": {
     "@jest/types": "30.2.0",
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Adding an `override` for now until the upstream is fixed

PR #7695, opened 2025-10-27
Upgrade dependencies previously blocked on ESM

codyebberson:


codyebberson:
The server side works, but unfortunately Jest does not.

You can follow the instructions here and use `node --experimental-vm-modules node_modules/jest/bin/jest.js` rather than `jest`

But you still get unfortunate errors

PR #7697, opened 2025-10-27
Fix server relative paths

codyebberson:
Fixes regression in CMS1500 base PNG loader due to `__dirname` and ESM related changes.

Original line:

```ts
image: path.relative(process.cwd(), path.resolve(__dirname, '../../../../static/cms1500.png')),
```

Then the broken line when porting to ESM, which assumes that the server will always be run in the proper current working directory.

```ts
image: path.relative(process.cwd(), path.resolve('./static/cms1500.png')),
```

And now our updated line:

```ts
image: relative(process.cwd(), resolve(getDirName(), '../../../../static/cms1500.png')),
```

With this helper:

```
/**
 * Returns the directory name of the current module.
 * Works with both CommonJS and ES modules.
 * @returns The directory name of the current module.
 */
function getDirName(): string {
  return typeof __dirname !== 'undefined' ? __dirname : dirname(fileURLToPath(import.meta.url));
}
```

We use that same helper in `@medplu/definitions`.  Probably an opportunity for abstraction or shared utility.

### DIFF
--- a/packages/server/src/config/loader.ts
+++ b/packages/server/src/config/loader.ts
@@ -4,7 +4,6 @@ import { splitN } from '@medplum/core';
 import { mkdtempSync, readFileSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
-import { fileURLToPath } from 'url';
 import { loadAwsConfig } from '../cloud/aws/config';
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

PR #7699, opened 2025-10-28
Embed the CMS 1500 base png as a data url

codyebberson:
Context:  our CMS 1500 PDF generator uses a "base PNG" image as the background to provide borders and label text.  

In https://github.com/medplum/medplum/pull/7358 the file loading behavior changed so that it was relative to the current working directory, which broke some Docker images and the Debian APT images.

In https://github.com/medplum/medplum/pull/7697 we tried to restore the relative path logic, but it was brittle and appears to depend on some build steps.

So, in this PR, we're just baking the PNG as a data URL string directly into the minified output.  Esbuild supports this out of the box.  The only tricky part was keeping Jest happy.

### DIFF
--- a/packages/server/jest.config.json
+++ b/packages/server/jest.config.json
@@ -8,7 +8,10 @@
   "transform": {
     "^.+\\.(js|jsx|ts|tsx)$": "babel-jest"
   },
-  "moduleFileExtensions": ["ts", "js", "json", "node"],
+  "moduleFileExtensions": ["ts", "js", "json", "node", "png"],
+  "moduleNameMapper": {
+    "\\.(png|jpg|jpeg|gif|svg|webp)$": "<rootDir>/src/__mocks__/image.ts"
+  },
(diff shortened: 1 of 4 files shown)


codyebberson:
@claude please review

packages/server/src/fhir/operations/utils/cms1500pdf.ts
```diff
@@ -9,10 +9,9 @@ import {
   HTTP_HL7_ORG,
 } from '@medplum/core';
 import type { Address, Claim, HumanName, Practitioner, RelatedPerson } from '@medplum/fhirtypes';
-import { dirname, relative, resolve } from 'node:path';
-import { fileURLToPath } from 'node:url';
 import type { Content, TDocumentDefinitions } from 'pdfmake/interfaces';
 import { getAuthenticatedContext } from '../../../context';
+import imageData from './cms1500.png';
```
techdavidy, line 14: 🎉

PR #7710, opened 2025-10-29
Fix uses of __dirname in generator

mattwiller:


packages/generator/src/index.ts
```diff
@@ -366,6 +367,15 @@ function getTypeScriptTypeForProperty(
   return baseType;
 }
 
+/**
+ * Returns the directory name of the current module.
+ * Works with both CommonJS and ES modules.
+ * @returns The directory name of the current module.
+ */
+export function getDirName(): string {
```
The-Alchemist: i heard you can use:

```
const __dirname = import.meta.dirname;
```

https://stackoverflow.com/a/50052194

perhaps that would be cleaner?

tested on my Node.js v24

The-Alchemist (approved):
🥳

mattwiller:
Thanks for the crucial assist @The-Alchemist!

PR #7713, opened 2025-10-29
Fix URL for v2-0360 CodeSystem

mattwiller:


packages/server/src/migrations/data/v31.ts
```diff
@@ -0,0 +1,20 @@
...
+export const migration: CustomPostDeployMigration = {
+  type: 'custom',
+  prepareJobData: (asyncJob) => prepareCustomMigrationJobData(asyncJob),
+  run: async (repo, job, jobData) => runCustomMigration(repo, job, jobData, callback),
+};
+
+// prettier-ignore
+async function callback(client: PoolClient, results: MigrationActionResult[]): Promise<void> {
+  const repo = getSystemRepo(client);
+  return rebuildR4ValueSets(repo);
```
mattlong: How about time and push something like `{ name: 'rebuildR4ValueSets', durationMs}` to `results` for the rebuild?

PR #7718, opened 2025-10-30
Fix bot editor and vmcontext bots

codyebberson:


packages/app/src/resource/BotEditor.tsx
```diff
@@ -49,6 +50,7 @@ export function BotEditor(): JSX.Element | null {
       .readResource('Bot', id)
       .then(async (newBot: Bot) => {
         setBot(newBot);
+        setModule(newBot.runtimeVersion === 'vmcontext' ? 'commonjs' : 'esnext');
```
ThatOneBro, line 53: Do we know if fission will also work with `esnext`?

Or should we invert this check?
```suggestion
        setModule(newBot.runtimeVersion === 'awslambda' ? 'esnext' : 'commonjs');
```

ThatOneBro (approved):

packages/app/src/resource/BotEditor.tsx
codyebberson: Good call - we should test.

PR #7412, opened 2025-09-18
Template-based questionnaire extraction (QuestionnaireResponse/$extract)

mattwiller:
Implements an initial version of the `QuestionnaireResponse/$extract` operation endpoint, as specified by the current draft version of the [Structured Data Capture IG](https://build.fhir.org/ig/HL7/sdc/extraction.html#template-extract).

Users provide a QuestionnaireResponse (by ID, or directly in the request body) and the server gets the associated Questionnaire (by reference from the QuestionnaireResponse, or directly from the request body) that defines that extraction rules and template resources.  The operation scans the Questionnaire for various Extensions defined by the SDC IG (primarily the [templateExtract](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-templateExtract.html) extension), and uses them to extract values from the QuestionnaireResponse into the template resources.  Template resources are placed in `Questionnaire.contained` and should contain [context](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-templateExtractContext.html) and [value](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-templateExtractValue.html) extensions; these are recursively scanned to generate a set of patches which will be applied to the template in order to fill it with values and remove the template metadata.  The operation returns a transaction Bundle containing the resources filled with values from the QuestionnaireResponse.

PR #7725, opened 2025-10-30
Document Structured Data Capture feature

mattwiller:
See #7412

### DIFF
--- a/packages/docs/docs/questionnaires/structured-data-capture.md
+++ b/packages/docs/docs/questionnaires/structured-data-capture.md
@@ -53,7 +53,10 @@ the extraction rules for that response item if it can be extracted in isolation.
 Within the template resources, [`templateExtractValue`][value-ext] and [`templateExtractContext`][context-ext]
 extensions work together to defined the rules for extracting data from the `QuestionnaireResponse` into the template.
 
-The value extension is placed on the field of the template resource into which a value should be placed:
+The value extension is placed on the field of the template resource into which a value should be placed. Since the
+fields containing these values often have "primitive" data types (e.g. `string`), they do not have a straightforward place
+to attach an extension. In these cases, a [FHIR primitive extension][fhir-primitive-ext] uses a special
+underscore-prefixed version of the field to contain the extension(s):
 
 ```json
 {
@@ -65,11 +68,12 @@ The value extension is placed on the field of the template resource into which a
       "id": "patientTemplate",
       "name": [
         {
-          // Since name is required, this is safe
+          // Primitive extension on the `text` field contains the value extension
           "_text": {
             "extension": [
               {
                 "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
+                // Since name is required, this is safe
                 "valueString": "item.where(linkId = 'name').answer.value.first()"
               }
             ]
@@ -88,8 +92,7 @@ The value extension is placed on the field of the template resource into which a
     {
       "linkId": "name",
       "type": "string",
-      "required": true,
-      "text": "fullName"
+      "required": true
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


packages/docs/docs/api/fhir/operations/extract.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Output
+
+The operation returns a `Bundle` resource containing the resources populated from templates in the Questionnaire, derived from values in the QuestionnaireResponse.
+
+**Request**:
+
+Given a `QuestionnaireResponse` with an associated `Questionnaire`, e.g.`
+
+```json
```
rahul1, line 35: nit: this removes all the warnings re: comments

```suggestion
```js
```

packages/docs/docs/questionnaires/structured-data-capture.md
```diff
@@ -0,0 +1,468 @@
...
+  "resourceType": "Questionnaire",
+  "status": "draft",
+  "contained": [
+    {
+      "resourceType": "Patient",
+      "id": "patientTemplate",
+      "name": [
+        {
+          // Since name is required, this is safe
+          "_text": {
```
rahul1, line 72: A lot of readers are probably not familiar with primitive extensions, so an inline comment for why `_text` is being used here would be helpful.

Alternatively, modifying the example to populate `given` name might be a better introduction

packages/docs/docs/questionnaires/structured-data-capture.md
```diff
@@ -0,0 +1,468 @@
...
+              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext",
+              "valueString": "item.where(linkId = 'phone')"
+            }
+          ],
+          "use": "home", // Default value for field specified
+          "_use": {
+            "extension": [
+              {
+                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
+                "valueString": "item.where(linkId = 'use').answer.value.first().code"
```
rahul1: does this override the default?

packages/docs/docs/questionnaires/structured-data-capture.md
```diff
@@ -0,0 +1,468 @@
...
+      // NOTE: This ID refers to the template itself,
+      // not the generated ID for intra-Bundle linking
+      "id": "patient",
+      "name": [
+        {
+          "_text": {
+            "extension": [
+              {
+                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
+                "valueString": "answer.value.first()"
```
rahul1, line 511: Doesn't there need to be a context here?

packages/docs/docs/questionnaires/structured-data-capture.md
```diff
@@ -0,0 +1,468 @@
...
+      "code": { "text": "Astrological sign" },
+      "valueCodeableConcept": {
+        "coding": [
+          {
+            "extension": [
+              {
+                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
+                // Simplified expressions because the extraction context is
+                // the relevant QuestionnaireResponse.item
+                "valueString": "answer.value"
```
rahul1, line 529: Doesn't there need to be a context here?

packages/docs/docs/questionnaires/structured-data-capture.md
```diff
@@ -0,0 +1,468 @@
...
+
+## Parsing Multiple Resources
+
+Many questionnaires gather information that must be processed into multiple separate FHIR resources, often ones that
+link together via references. To [link the resources together][bundle-intern-refs] in the results `Bundle`, place one
+or more [`extractAllocateId`][allocate-id] extensions at the root of the `Questionnaire` to pregenerate an internal
+reference string that can be assigned to a template during extraction via the associated `templateExtract` extension
+and then referenced in other resource templates.
+
+```json
```
rahul1, line 488: I think I understand what's going on here, but it introduces 2 concepts at the same time:

1. Setting the `sdc-questionnaire-templateExtract` at a non-root item
2. Allocating a reference

This is a complicated topic, so having these split into 2 separate examples would help follow along, even if they are a bit contrived.

For example: 
1. Take patient name for multiple patients (e.g. siblings) - no dependencies
2. An example like this that creates 2 resources that are linked, liek this

packages/docs/docs/questionnaires/structured-data-capture.md
```diff
@@ -0,0 +1,468 @@
...
+      "text": "Full Name",
+      "required": true,
+      // Starting the extraction in the context of the relevant item simplifies extract rules
+      "extension": [
+        {
+          "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtract",
+          "extension": [
+            { "url": "template", "valueReference": { "reference": "#patient" } },
+            // Assign the generated UUID to this generated resource
+            { "url": "fullUrl", "valueString": "%patientRef" }
```
rahul1, line 559: This comment is helpful to understand how `patientRef` is allocated to the fullURL is assigned to the patient, but it's kind of buried down in the depths of the example. 

A couple of things that would make this easier to understand: 
- Use the docusaurus [line highlighting feature](https://docusaurus.io/docs/markdown-features/code-blocks#line-highlighting) to call attention to these lines
- Having a bulleted list above the code block that provides a high level outline of what's going on? 
  -  Line ##: We allocate a ref (u
  - Line ##: we use the ref 
  - Line ##: We assign the ref to the fullURL 
  - Line ###: ....

packages/docs/docs/questionnaires/structured-data-capture.md
mattwiller: Makes sense — I split it into two examples

packages/docs/docs/questionnaires/structured-data-capture.md
mattwiller: I didn't end up using the line highlighting, but took the suggestion of writing out an explicit list above the code block and then referred to those items in the example

packages/docs/docs/questionnaires/structured-data-capture.md
mattwiller: There doesn't, actually, since the item is required and there will always be a value.  I added a section about the context stuff to try and clarify

packages/docs/docs/questionnaires/structured-data-capture.md
mattwiller: Yep, added a comment to make it more clear

packages/docs/docs/questionnaires/structured-data-capture.md
mattwiller: Got it — I added a note about primitive extensions with a link to the FHIR docs for more info.  Since `given` also has type `string`, it would need a primitive extension too (as would everything in `HumanName` other than `period`)

packages/docs/docs/api/fhir/operations/extract.md
mattwiller: Those warnings only show up in the GitHub diff, I didn't see them while developing or testing so probably fine?

PR #7728, opened 2025-10-31
fix(agent): fix running via tsx (revert to cjs)

ThatOneBro:
Was unable to run `npm run agent` since `tsx` respects `package.json`, which says the agent is a module. This means `__dirname` is not present in the runtime, which causes `tsx` to fail.

### DIFF
--- a/packages/agent/esbuild.mjs
+++ b/packages/agent/esbuild.mjs
@@ -15,9 +15,6 @@ const options = {
   resolveExtensions: ['.js', '.ts'],
   target: 'es2021',
   tsconfig: 'tsconfig.json',
-  define: {
-    'import.meta.main': 'true',
-  },
 };
 
 // The single executable application feature only supports running a single embedded CommonJS file.
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -13,7 +13,7 @@
   },
   "license": "Apache-2.0",
   "author": "Medplum <hello@medplum.com>",
-  "type": "module",
+  "type": "commonjs",
   "files": [
     "dist"
   ],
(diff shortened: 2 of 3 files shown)


codyebberson:
Which part didn't work?

Is it just `__dirname`?  If that's it, can we add a shim?

At this point, we're so close to not having any exceptions, and if at all possible, I'd like to hold the line and push forward.

ThatOneBro:
> Which part didn't work?

Running with `npm run agent` invokes `tsx` which always respects the `type` of the module in `package.json`.
The agent has to built as CommonJS anyways, so I think it makes sense to keep the type as `commonjs`.

`__dirname` is not available whenever you run it as esm in `tsx`.

codyebberson:
Most of the calls to `__dirname` should probably be changed to current working directory anyway, no?

ThatOneBro:
We could set it CWD but we will need to see what that evaluates to in the Node SEA binary first.

Separately, running in `tsx` with type `module`, the program also exits immediately. This is because it's not the built output, and we are relying on esbuild to find and replace `import.meta.main` to `true`. I don't think there is a work around for this

My preference is to label this as a CommonJS module to keep the runtime of the dev environment, testing, and built output the same and avoid running into divergent behavior because "it worked on my machine"(TM).

codyebberson (approved):
K.  I see the wisdom.  It makes me sad though 😢

ThatOneBro:
Yeah I'm bummed about the lack of ESM support for Node SEA too, hoping to see some progress there soon...

For tracking purposes there is this issue for ESM support in SEA:
https://github.com/nodejs/node/issues/53565

PR #7731, opened 2025-10-31
Filter out inactive project memberships from /auth/me

codyebberson:
Context: https://medplum.slack.com/archives/C06U3LCJQUS/p1761925068844779

### DIFF
--- a/packages/server/src/auth/me.ts
+++ b/packages/server/src/auth/me.ts
@@ -66,13 +66,15 @@ export async function meHandler(req: Request, res: Response): Promise<void> {
     security = {
       mfaEnrolled: !!user.mfaEnrolled,
       sessions,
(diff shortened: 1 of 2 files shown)


packages/server/src/auth/me.ts
```diff
@@ -66,13 +66,15 @@ export async function meHandler(req: Request, res: Response): Promise<void> {
...
       sessions,
-      memberships: memberships.map((membership) => ({
-        resourceType: 'ProjectMembership',
-        id: membership.id,
-        identifier: membership.identifier,
-        profile: membership.profile,
-        admin: membership.admin,
-      })),
+      memberships: memberships
+        .filter((m) => m.active !== false)
```
codyebberson, line 70: `ProjectMembership.active` is default `true`, so need to explicitly check `!== false`

PR #7732, opened 2025-10-31
feat(agent): add RTT metrics to stats

### LINKED ISSUE #7727: [Agent] Add additional logging for request/response correlation, add RTT to stats
Currently there is not any logging around when a response (an ACK) is sent back on the channel that the message came in on. It would be useful for tracking how long a given message took to be processed through the whole system in setups where messages come in on a channel and then are pushed to another HL7 server remotely. RTT for a message could be also included in the log message itself.

Additionally, an RTT metric like average RTT for both inbound (messages coming in from a channel) and outbound (pushed messages) would be useful to have.

Inbound RTT would be per channel, and outbound RTT could be per remote address.

ThatOneBro:
Fixes #7727

### DIFF
--- a/packages/agent/src/hl7.test.ts
+++ b/packages/agent/src/hl7.test.ts
@@ -1116,6 +1116,7 @@ describe('HL7', () => {
       const mockServer = new Server('wss://example.com/ws/agent');
       const state = {
         transmitRequests: [] as AgentTransmitRequest[],
+        shouldSendAck: true,
       };
 
       mockServer.on('connection', (socket) => {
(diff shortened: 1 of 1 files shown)


packages/agent/src/channel-stats-tracker.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Adds an RTT sample to the collection, maintaining max size.
+   * @param rtt - RTT value in milliseconds.
+   */
+  private addRttSample(rtt: number): void {
+    this.completedRtts.push(rtt);
+    // Keep only the most recent samples
+    if (this.completedRtts.length > this.maxRttSamples) {
+      this.completedRtts.shift();
```
mattwiller, line 114: It may be worthwhile in the future to use a ring buffer instead, since `shift()` likely has to update the whole array

packages/agent/src/channel-stats-tracker.ts
```diff
@@ -0,0 +1,181 @@
...
+   * Gets the number of completed RTT samples.
+   */
+  getCompletedCount(): number {
+    return this.completedRtts.length;
+  }
+
+  /**
+   * Cleans up the ChannelStats instance.
+   */
+  cleanup(): void {
```
mattwiller, line 224: Should this get called by a `Symbol.dispose` method or something?

mattwiller (approved):

packages/agent/src/channel-stats-tracker.ts
ThatOneBro: Yeah I was thinking about this too. However, I remember recent research about `shift()` proving that it is implemented for non-primitives in a way which operates in O(1) time but I'll double check that.

ThatOneBro: Apparently this is not the case for v8 though 😬 

https://issues.chromium.org/issues/42202676

Also see: https://github.com/nodejs/node/issues/42449

Might be worth using a `RingBuffer` after all

PR #7735, opened 2025-10-31
Proof-of-concept: migrate core to vitest

codyebberson:
Spent about an hour on this, just to see how hard it would be.  Intended for a conversation starter.

Not ready for merge.  `@fast-check/vitest` has a peer dependency version mismatch.  Jammed it through with `--legacy-peer-deps`, but that will likely fail where we use `--strict-peer-deps`.  Will hopefully be fixed soon: https://github.com/dubzzz/fast-check/pull/6236

This also seems to have exacerbated some flaky tests in `packages/core/src/client-subscriptions.test.ts`


dubzzz:
Normally the peer dependency mismatch should be fixed now 🤞

The-Alchemist:
https://github.com/dubzzz/fast-check/pull/6236 has been merged!

The-Alchemist:
replaced by #8916

PR #7739, opened 2025-10-31
Fixes #7729 - default super admin email, password, client

### LINKED ISSUE #7729: Custom default admin user and ClientApplication
Hi Medplum team,

I have a question regarding self-hosting.

Is there a way to seed a Medplum server instance with a custom admin user (instead of the default admin@example.com / medplum_admin user)?

Additionally, is it possible to pre-seed a Project and a ClientApplication with specific client_id and client_secret values?

If this isn’t currently supported, would it be reasonable to use the /admin/project/* endpoints to programmatically create these default entities during setup?

Thanks in advance for your help!

codyebberson:
Before:  When first launching Medplum server, the user had to login with default credentials and then change password.

After: Medplum server config option includes options for `defaultSuperAdminEmail` and `defaultSuperAdminPassword`.

Also: Server config includes options for `defaultSuperAdminClientId` and `defaultSuperAdminClientPassword`, which will create a new default `ClientApplication`

### DIFF
--- a/packages/server/src/seed.ts
+++ b/packages/server/src/seed.ts
@@ -1,8 +1,9 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import { createReference } from '@medplum/core';
-import type { Practitioner, Project, ProjectMembership, User } from '@medplum/fhirtypes';
+import type { ClientApplication, Practitioner, Project, ProjectMembership, User } from '@medplum/fhirtypes';
 import { bcryptHashPassword } from './auth/utils';
+import type { MedplumServerConfig } from './config/types';
 import { r4ProjectId } from './constants';
 import type { Repository } from './fhir/repo';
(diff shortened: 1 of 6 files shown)


codyebberson:
@claude please review

packages/server/src/seed.ts
```diff
@@ -90,6 +94,29 @@ async function createSuperAdmin(systemRepo: Repository): Promise<void> {
     profile: createReference(practitioner),
     admin: true,
   });
+
+  if (config.defaultSuperAdminClientId && config.defaultSuperAdminClientSecret) {
+    // Use specified client ID and secret
+    const client = await systemRepo.updateResource<ClientApplication>({
```
ThatOneBro, line 100: Should this be a `createResource` rather than `updateResource`? 🤔 Or does it even matter with `systemRepo`?

codyebberson: To use client-assigned ID, you have to use `updateResource`.  Using `createResource` will ignore the provided `id` and always generate a new one.

PR #7747, opened 2025-11-03
Dependency upgrades 2025-11-03

medplumbot:
Dependency upgrades


codyebberson:
Storybook 10 blocked on https://github.com/josiahayres/storybook-addon-mantine/issues/21

Next.js 16 upgrade is low priority, we can punt https://nextjs.org/docs/app/guides/upgrading/version-16

PR #7749, opened 2025-11-03
Update data manifest entry for v29 migration

mattwiller:


packages/server/src/migrations/data/data-version-manifest.json
```diff
@@ -112,7 +112,8 @@
     "requiredBefore": "4.4.0"
   },
   "v29": {
-    "serverVersion": "4.4.3"
+    "serverVersion": "4.5.2",
```
ThatOneBro, line 115: I think it's ok to keep the serverVersion the same, no? Since it could still run in any version from 4.4.3 -> now?

mattwiller: The fixes that made it run correctly [landed in `4.5.2`](https://github.com/medplum/medplum/commit/90bf828c5ad899532be03c427ed6100ded18ca36), so I thought that made more sense as the version to use

ThatOneBro: Hmm yeah that makes sense actually 👍

PR #7751, opened 2025-11-03
Fix version commit hash to length 7

noahsilas:
Git adjusts the length of this commit hash based on the uniqueness of the commits in the repository, but we would like to set this to a fixed length. We're choosing 7 to match Github UI elements.

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -67,7 +67,7 @@ jobs:
         # This forces git shorthash to match between @medplum/agent and @medplum/core
         run: |
           set -e
-          echo "MEDPLUM_GIT_SHORTHASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


ThatOneBro (changes requested):
I think we may need to make this change elsewhere, such as in the CI job for generating agent builds. Otherwise this might lead to hash mismatches between filenames and what is reported by the runtime 👀 

Try doing a search for `MEDPLUM_GIT_SHORTHASH` for all the different instances where we generate this in CI

For example:
https://github.com/medplum/medplum/blob/main/.github/workflows/publish.yml#L174

Looks like there is another instance of this in the `app` package too:
https://github.com/medplum/medplum/blob/main/packages/app/vite.config.ts#L17

noahsilas:
Thanks @ThatOneBro - great context, I didn't realize this was used beyond the UI. I think I found the related usage in 093f244d, let me know if you have other pointers I should follow.

PR #7752, opened 2025-11-03
Improve bulk export memory efficiency

nate-watkins:
#### Feel free to edit or cherrypick any part you like or ignore entirely

Hey team- we found our larger `$export` operations were causing containers to get `OOMKilled`. I believe this is primarily due to the continuously growing `resourceSet` throughout the entire operation, even after an entire `resourceType` is completed. This PR attempts to free memory incrementally throughout the operation, increasing the size of an export the server is capable of under the same memory constraint. This will have a non-zero impact on how long this operation takes to complete, but I don't have a good benchmark for it.

- New method in `BulkExporter`:
  - `closeWriter(resourceType)` - removes exported resource references from tracking set and closes file writer immediately after a resource type completes
- Stream backpressure handling in BulkFileWriter.write():
  - If `stream.write` returns false, indicating the internal buffer is greater than the `highWaterMark`, it will wait for the stream's `drain` event

### DIFF
--- a/packages/server/src/fhir/operations/utils/bulkexporter.ts
+++ b/packages/server/src/fhir/operations/utils/bulkexporter.ts
@@ -113,7 +113,7 @@ export class BulkExporter {
 
   async close(project: Project): Promise<AsyncJob> {
     if (!this.resource) {
-      throw new Error('Export muse be started before calling close()');
+      throw new Error('Export must be started before calling close()');
     }
 
     for (const writer of Object.values(this.writers)) {


packages/server/src/fhir/operations/utils/bulkexporter.ts
```diff
@@ -21,8 +21,14 @@ class BulkFileWriter {
     this.writerPromise = getBinaryStorage().writeBinary(binary, filename, NDJSON_CONTENT_TYPE, this.stream);
   }
 
-  write(resource: Resource): void {
-    this.stream.write(JSON.stringify(resource) + '\n');
+  async write(resource: Resource): Promise<void> {
+    const data = JSON.stringify(resource) + '\n';
+    // Handle backpressure - if write buffer is full, wait for drain
+    if (!this.stream.write(data)) {
```
ThatOneBro: 👍 nice

mattwiller:
Thanks @nate-watkins!  I'll take a look, but right away it looks like you'll need to take care of the [DCO signoff](https://github.com/medplum/medplum/pull/7752/checks?check_run_id=54411985791) before merging.  Would you mind signing your commit and force-pushing the branch?

packages/server/src/fhir/operations/export.ts
```diff
@@ -82,6 +82,10 @@ export async function exportResources(
       continue;
     }
     await exportResourceType(exporter, resourceType, pageSize, since);
+    
+    // Close writer and free memory for this resource type immediately
+    await exporter.closeWriter(resourceType);
+    exporter.clearTrackingForResourceType(resourceType);
```
mattwiller: I think this cleanup logic should go inside `exportResourceType()`, to ensure it happens at the correct time

packages/server/src/fhir/operations/utils/bulkexporter.ts
```diff
@@ -21,8 +21,14 @@ class BulkFileWriter {
...
+      await new Promise<void>((resolve) => {
+        this.stream.once('drain', () => resolve());
+      });
```
mattwiller:
```suggestion
      await new Promise<void>((resolve, reject) => {
        this.stream.once('drain', () => resolve());
        this.stream.once('error', (err) => reject(err));
      });
```

packages/server/src/fhir/operations/utils/bulkexporter.ts
```diff
@@ -64,6 +70,24 @@ export class BulkExporter {
...
+    const writer = this.writers[resourceType];
+    if (writer) {
+      await writer.close();
+      // Keep reference for formatOutput(), but free the stream resources
+    }
+  }
+
+  clearTrackingForResourceType(resourceType: string): void {
+    // Remove all tracked resources of this type to free memory
+    // This is safe to call after closeWriter() for a resource type
```
mattwiller: Should this just be included as part of `closeWriter()`, instead of being its own method?

nate-watkins: Probably. Closing the writer itself is probably a trivial amount of memory improvement compared to removing the items from the `resourceSet`, so I considered not doing it at all. I'll adjust it to be part of `closeWriter`.

mattwiller (approved):

nate-watkins:
The other thing I considered was to make `resourceSet` a `Map<string, Set<string>>`, so clearing a resourceType would be `resourceSet.delete(resourceType)` instead of iterating over the values. Can get that in if that's preferable.

mattwiller:
@nate-watkins That would be great if you''re willing to add that in: I think making the cleanup operation more efficient would be a good low-hanging-fruit performance optimization

PR #7757, opened 2025-11-04
Removing chart demo as functionality in `medplum-provider` supercedes

reshmakh:


techdavidy (approved):

techdavidy:
🚀

reshmakh:
Package Lock has a reference to this - is this ok? 
image

codyebberson:
> Package Lock has a reference to this - is this ok? 

This will work just fine.  NPM ignores extra packages in `package-lock.json` if they don't match a `package.json` on disk.  We actually intentionally use that behavior inside docker images (i.e., we don't want front-end dependencies in Docker, but we want to be able to use `package-lock.json`).

This will all get cleaned out on the next dependency upgrade, which will rewrite `package-lock.json` entirely.

Or, if you want to clean it up, you can run `npm i` from the root.

PR #7759, opened 2025-11-04
Prefer `String#replaceAll()` over `String#replace()`

noahsilas:
- Shows more explicit intention to perform multiple replacements
- Allows using string literals as the search pattern when not needing Regex power

See: https://sonarcloud.io/organizations/medplum/rules?open=typescript%3AS7781&rule_key=typescript%3AS7781

### DIFF
--- a/examples/medplum-demo-bots/src/epic/epic-query-patient.ts
+++ b/examples/medplum-demo-bots/src/epic/epic-query-patient.ts
@@ -22,7 +22,7 @@ export async function handler(medplum: MedplumClient, event: BotEvent<Patient>):
   }
 
   // Handles unknown issue with newlines in private key
-  const privateKeyString = event.secrets['EPIC_PRIVATE_KEY']?.valueString?.replace(/\\n/g, '\n');
+  const privateKeyString = event.secrets['EPIC_PRIVATE_KEY']?.valueString?.replaceAll('\\n', '\n');
   if (!privateKeyString) {
     throw new Error('Missing EPIC_PRIVATE_KEY');
   }
--- a/packages/core/src/crypto.ts
+++ b/packages/core/src/crypto.ts
@@ -29,7 +29,7 @@ export async function encryptSHA256(str: string): Promise<ArrayBuffer> {
  * @returns A random UUID.
  */
 export function generateId(): string {
-  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replaceAll(/[xy]/g, (c) => {
     const r = (Math.random() * 16) | 0;
     const v = c === 'x' ? r : (r & 0x3) | 0x8;
     return v.toString(16);
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -352,7 +352,7 @@ export function getPropertyDisplayName(propertyName: string): string {
   // Then normalize whitespace to single space character
   // For example, for property name "birthDate",
   // the display name is "Birth Date".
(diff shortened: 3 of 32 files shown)


examples/medplum-demo-bots/src/epic/epic-query-patient.ts
```diff
@@ -22,7 +22,7 @@ export async function handler(medplum: MedplumClient, event: BotEvent<Patient>):
   }
 
   // Handles unknown issue with newlines in private key
-  const privateKeyString = event.secrets['EPIC_PRIVATE_KEY']?.valueString?.replace(/\\n/g, '\n');
+  const privateKeyString = event.secrets['EPIC_PRIVATE_KEY']?.valueString?.replaceAll(/\\n/g, '\n');
```
mattwiller: Could this one also use a string literal?
```suggestion
  const privateKeyString = event.secrets['EPIC_PRIVATE_KEY']?.valueString?.replaceAll('\\n', '\n');
```

mattwiller (commented):
Looks like there's an extraneous `test .md` file; would you mind removing that before merging?

packages/core/src/types.ts
```diff
@@ -352,7 +352,7 @@ export function getPropertyDisplayName(propertyName: string): string {
   // Then normalize whitespace to single space character
   // For example, for property name "birthDate",
   // the display name is "Birth Date".
-  return words.map(capitalizeDisplayWord).join(' ').replace('_', ' ').replace(/\s+/g, ' ');
+  return words.map(capitalizeDisplayWord).join(' ').replace('_', ' ').replaceAll(/\s+/g, ' ');
```
noahsilas, line 355: Not the same flavor of mechanical change as the rest of this commit, but should `replace('_', ' ')` also be doing a `replaceAll` operation here?

examples/medplum-demo-bots/src/epic/epic-query-patient.ts
noahsilas: Absolutely, thanks! Updated in d25a1dfb.

packages/core/src/crypto.ts
```diff
@@ -29,7 +29,7 @@ export async function encryptSHA256(str: string): Promise<ArrayBuffer> {
  * @returns A random UUID.
  */
 export function generateId(): string {
-  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replaceAll(/[xy]/g, (c) => {
```
noahsilas, line 32: Potential follow up: the above comment suggests we can use `crypto.randomUUID()` here; that has pretty wide support now, maybe we can replace this implementation?

noahsilas:
> Looks like there's an extraneous `test .md` file; would you mind removing that before merging?

Arg, oops. Thanks for catching that!

packages/core/src/crypto.ts
ThatOneBro: Good question, there's actually one case where `crypto.randomUUID` is not available: non-localhost cleartext HTTP traffic

I think that's the main use-case for the usage of `generateId` here 🥲

packages/core/src/types.ts
ThatOneBro: I think `replaceAll` should be fine here 👍

mattwiller (approved):

techdavidy:
mind checking the new issues flagged by Sonar. They are Low but seem to be relevant to the changes in this PR.

techdavidy (approved):

noahsilas:
> mind checking the new issues flagged by Sonar. They are Low but seem to be relevant to the changes in this PR.

Thanks, good fixes, and TIL about `String.raw`. I can't believe that's been around so long and I had never bumped into it!

Issue #7651, opened 2025-10-21
Autodownloader race condition when resource contains multiple presigned URLs


maddyli:
The autodownloader feature scans resources for external URL's, and adds them to a queue that attempts to download the content of these URL's into Medplum binary storage. 

If `n` external URL's are present in the same resource, each of those URL's are added to the job queue separately. Upon the content being downloaded, the autodownloader updates the whole resource. This causes a race condition where each of the jobs overwrites the work of other jobs, and the resulting resources are queued into the autodownloader again, until eventually `n` jobs write non-external URLs. 

Relevant file with updateResource line: https://github.com/medplum/medplum/blob/main/packages/server/src/workers/download.ts#L242

Example resource where we observed this behavior: https://app.medplum.com/DocumentReference/fb6a6d35-ff7f-4628-bf41-3a92f02d7178/history
<img width="572" height="722" alt="Image" src="https://github.com/user-attachments/assets/04a20a2c-4779-4cba-9bb3-fdb6ea4bf6ff" />

Relevant cursor chat: 
[cursor_is_autodownload_enabled_for_pres.md](https://github.com/user-attachments/files/23031368/cursor_is_autodownload_enabled_for_pres.md)

PR #7764, opened 2025-11-04
sketch of autodownloader job race conditions

maddyli:
https://github.com/medplum/medplum/issues/7651

### DIFF
--- a/packages/server/src/workers/download.ts
+++ b/packages/server/src/workers/download.ts
@@ -11,6 +11,7 @@ import { getConfig } from '../config/loader';
 import { tryGetRequestContext, tryRunInRequestContext } from '../context';
 import { getSystemRepo } from '../fhir/repo';
 import { getLogger, globalLogger } from '../logger';
+import { getRedis } from '../redis';
 import { getBinaryStorage } from '../storage/loader';
 import { parseTraceparent } from '../traceparent';
 import type { WorkerInitializer } from './utils';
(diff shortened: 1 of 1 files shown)


maddyli:
@medplum/dev happy to add tests if this is considered a good solution, can close otherwise

packages/server/src/workers/download.ts
```diff
@@ -253,12 +343,45 @@ export async function execDownloadJob<T extends Resource = Resource>(job: Job<Do
...
+    // This ensures we only replace the specific URL we're processing
+    const updated = replaceAttachmentUrl(latestResource, url, binary.id);
+    if (!updated) {
+      // URL was removed between check and update, unlock and exit
+      await unlockUrl(resourceType, id, url);
+      return;
+    }
+
     (updated.meta as Meta).author = { reference: 'system' };
     await systemRepo.updateResource(updated);
```
noahsilas, line 378: Using Redis to lock the resource is an interesting idea, but adds some complexity around concurrency and unlocking that seems likely to cause future issues. The other approach in here of creating a much more targeted update seems like it has the potential to resolve the issue more elegantly.

I think that the trick to making that work as expected will be to switch this line from an `update` to a `patch` to avoid clobbering fields other than the one we are explicitly trying to update. I'm going to take a quick stab at adapting your `replaceAttachmentUrl` function to make it work nicely with that style of partial update, but I'm still brand new and learning about the system, and I am very open to feedback on this approach!

maddyli: Definitely! I think the only reason I didn't go with a patch is that the patch results in a resource update, which means that the "remaining" external URL's will likely get re-queued, which will result in less of the same behavior as before, but still some unnecessary jobs being added to the autodownloader. Maybe there is a clean way to get around that?

noahsilas: Ah, interesting! So it seems like there are two ways that locks can provide an enhancement:

- Preventing two writers from racing to mutate the row and clobbering each others changes
- Preventing two jobs from spawning to both fetch the same URL and update the references

I think I had assumed that the second one there wouldn't be a problem in a Patch operation because we would filter down to only looking at updated fields, but that doesn't seem to be the case. Maybe we can address that as a root cause as well?

ThatOneBro: Hey y'all, I do agree we probably need a patch here 🤔 
I'm thinking a slow download could cause an update to overwrite newer data since we read the file before the download starts and update after the download / stream finishes.
We should probably add a test that the field is what we expect and then patch it I think?

Separately, another "locking" mechanism would be to use the `AsyncJob` framework, although how exactly that would be implemented is not exactly clear immediately. But generally the job could be a single "source of truth" for a given job based on the URL as a key.

noahsilas: @ThatOneBro I think @maddyli had addressed what you are describing above by refetching after the download (https://github.com/medplum/medplum/pull/7764/files#diff-0816f84a8cf359fb7b27d26c84f0787e741d6cfbe5e6a39e7a4387134e4c2e0cR350). Since we go on to replace exact URL matches that should be pretty safe, although since it's not actually gated on an exact patch update it still could be subject to racing probably?

I've put up a draft that builds on the direction described above, keeping that bit where we refetch but replacing the locking with more fine-grained enqueuing: https://github.com/medplum/medplum/pull/7781/commits/88a140f1fa746d49d61a2ce5e234c5889e5ab17c

I'd love any early feedback on that. It seems to work pretty well in my local testing, but there are some awkward bits that I'm hoping to improve on still, mostly around the interface between the crawler (which seems to operate on FHIRPath) and our patch tool (which only takes JSON Patch). Let me know what you all think!

maddyli:
@noahsilas cool stuff! referencing the other PR here: https://github.com/medplum/medplum/pull/7781

PR #7769, opened 2025-11-05
Enable mTLS passthrough on ALB

codyebberson:
Implements [OAuth RFC 8705](https://datatracker.ietf.org/doc/html/rfc8705) (mTLS client auth) with AWS ALB

mTLS is an optional part of the HTI-4 / electronic prior auth spec (https://hl7.org/fhir/us/davinci-pas/privacy.html), but it appears that at least some payors will require it.

There are 4 new optional settings in the CDK config JSON:

- `mtlsDomainName?: string` - Domain name for mTLS load balancer
- `mtlsSslCertArn?: string` - SSL certificate ARN for mTLS load balancer
- `mtlsInternetFacing?: boolean` - Flag to enable or disable whether the mTLS load balancer is internet facing (default true)
- `mtlsWafIpSetArn?: string` - Optional WAF IP set ARN for the mTLS load balancer

There is a new property on the `ClientApplicaiton` type:

```typescript
  /**
   * Optional PEM-formatted certificates that are allowed to authenticate
   * to this service via mutual TLS. Supports both Certificate Authorities
   * (CAs) and self-signed certificates. Multiple certificates can be
   * included.
   */
  certificateTrustStore?: string;
```

Once configured, here is an example of how to use mTLS for auth:

```typescript
  test('mTLS with self-signed certificate success', async () => {
    const { cert } = generateSelfSignedCert('CN=Test Client');
    const result = validateClientCert(cert, cert);
    expect(result).toBeInstanceOf(X509Certificate);
    expect(result.subject).toBe('CN=Test Client');

    const { client } = await createTestProject({
      withClient: true,
      client: {
        certificateTrustStore: cert,
      },
    });

      .post('/oauth2/token')
      .type('form')
      .set('x-mtls-cert', encodeURIComponent(cert))
      .send({
        grant_type: 'client_credentials',
        client_id: client.id,
      });
    expect(res.status).toBe(200);
    expect(res.body.error).toBeUndefined();
    expect(res.body.access_token).toBeDefined();
  });
```

This has been deployed and validated on staging:

```
https://mtls.api.staging.medplum.dev/
```

### DIFF
--- a/packages/server/src/oauth/cert.ts
+++ b/packages/server/src/oauth/cert.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { normalizeErrorString } from '@medplum/core';
 import { X509Certificate } from 'node:crypto';
+import { getLogger } from '../logger';
 
 /**
  * Validates the client certificate received via ALB passthrough.
@@ -128,5 +129,6 @@ function validateCaSignedCert(clientCert: X509Certificate, trustedCerts: X509Cer
       lastError = normalizeErrorString(err);
     }
   }
-  throw new Error(`Certificate validation failed: ${lastError ?? 'No matching CA found'}`);
+  getLogger().warn(`Certificate validation failed for ${clientCert.subject}. Last error: ${lastError}`);
+  throw new Error('Certificate validation failed: No matching CA found');
 }


packages/server/src/logger.ts
```diff
@@ -9,15 +9,9 @@ export const globalLogger = new Logger(
   process.env.NODE_ENV === 'test' ? LogLevel.ERROR : LogLevel.INFO
 );
 
-function write(msg: string): void {
```
ThatOneBro, line 12: +1 on standardization, but do we want to standardize on the `process.stdout.write` version? 

`console.log` implicitly flushes which in high throughput scenarios adds a lot of unnecessary overhead.

Some quick benchmarks showed as we use them, `systemLogger` is twice as fast when logging 50,000 messages with mixed message sizes.

Screenshot 2025-11-05 at 11 52 11 AM
Screenshot 2025-11-05 at 11 52 20 AM

codyebberson: Yes, although I'm a little concerned about how many tests that will break that depended on spying on `console.log`.  let me try it

codyebberson: Actually - I'm going to revert these changes, and let's treat that as a separate PR.  I only touched this because I was confused by `getLogger()` behavior, but that's a separate issue.

packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
```diff
@@ -240,7 +240,7 @@ export function ElementDefinitionTypeInput(props: ElementDefinitionTypeInputProp
     case PropertyType.time:
     case PropertyType.uri:
     case PropertyType.url:
-      if (props.path === 'Project.secret.value[x]') {
+      if (props.path === 'Project.secret.value[x]' || props.path === 'ClientApplication.certificateTrustStore') {
```
ThatOneBro, line 244: 👍

codyebberson:
Update:  The current strategy unfortunately affects non-mTLS usage, leading to dialogs like this:

image

We will need to use one of these alternate solutions:
1. Add another listener with the same ALB on a different port (i.e., https://api.medplum.com:8443/)
   * Technically secure
   * Minor risks of customers not supporting non-standard ports
   * Somewhat "ugly" - perhaps looks unprofessional
2. Add an entirely separate ALB with different domain name (i.e., https://mtls.api.medplum.com/)
   * Equally secure
   * Uses standard ports, so no risk of port problems
   * Looks professional
   * Extra $100-$200 per month

codyebberson:
@claude please review

packages/server/src/oauth/cert.ts
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      // Check if this CA issued the certificate
+      if (clientCert.checkIssued(caCert)) {
+        // Verify the signature using the CA's public key
+        if (clientCert.verify(caCert.publicKey)) {
+          return;
+        }
+      }
+    } catch (err) {
+      lastError = normalizeErrorString(err);
```
ThatOneBro, line 129: Are these errors always safe to return to the client or is possible that they might leak info about the certs in the trust store?

codyebberson: Good catch. Let me think through this:

The risk is real but limited here. The error message `Certificate validation failed: ${lastError ?? 'No matching CA found'}` could leak implementation details.

In practice, the Node.js `X509Certificate` methods typically throw errors like:
- `"error:0900006e:PEM routines:..."` (OpenSSL error strings)
- `"Unsupported key type"` etc.

These don't directly expose CA cert contents, but they **do** reveal:
- Which OpenSSL operations are being attempted
- Internal key type/algorithm details
- Potentially *which* CA matched `checkIssued` but failed `verify` — which is the most sensitive leak

That last point is the real issue: if `checkIssued` passes but `verify` throws, the error effectively tells the client "we have a CA cert matching your issuer DN, but your signature is wrong" — which confirms information about the trust store's contents.

I will update to log the full error, and return a generic error.

Thanks @ThatOneBro 🙏

PR #7776, opened 2025-11-05
ProjectMembership differentiator in ChooseProfileForm

finnbergquist:
### Problem
When a User has multiple ProjectMemberships in the same project (only doable with [.../invite](https://www.medplum.com/docs/api/project-admin/invite) when `forceNewMembership=true`)  , it is impossible to differentiate between ProjectMemberships when you are logging in. 

### Solution

- ChooseProfileForm will optionally display Medplum sanctioned ProjectMembership identifier label (`https://medplum.com/identifier/label`) so that the User can differentiate between multiple ProjectMemberships that both point to the same Project and same Profile resource
- Requires small server change to pass the `ProjectMembership.identifier` as part of the login response to be displayed in ChooseProfileForm

### DIFF
--- a/packages/react/src/auth/ChooseProfileForm.test.tsx
+++ b/packages/react/src/auth/ChooseProfileForm.test.tsx
@@ -88,7 +88,6 @@ describe('ChooseProfileForm', () => {
     );
 
     expect(screen.getByText('Choose profile')).toBeInTheDocument();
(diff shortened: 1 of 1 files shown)


finnbergquist:
Screenshot 2025-11-05 at 3 57 38 PM

techdavidy:
@claude please review

techdavidy:
mind adding a test in the react files please

PR #7778, opened 2025-11-06
Send HTTP headers to bot subscriptions

codyebberson:


techdavidy:
80.0% coverage exactly

PR #7780, opened 2025-11-06
feat(bot): Add bot name to lambda description

xcskier56:
When you start to have more than a few lambda bots, it gets very difficult to tell them apart. By using the unused `Description` field on lambdas, we can add human readable names to our lambdas which makes tasks like finding cloudwatch logs much easier.

### DIFF
--- a/examples/medplum-provider/src/pages/spaces/SpacesPage.test.tsx
+++ b/examples/medplum-provider/src/pages/spaces/SpacesPage.test.tsx
@@ -190,7 +190,6 @@ describe('SpacesPage', () => {
 
       expect(screen.getByText('AI Assistant')).toBeInTheDocument();
     });
-
   });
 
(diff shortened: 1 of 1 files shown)


codyebberson:
Thanks @xcskier56 - this is a great idea.

Code change looks good to me 👍 

See here for how to fix the DCO check: https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot

tl;dr, run:

```bash
git commit --amend --signoff
git push -f origin feat/Lambda-Bot-Description
```

xcskier56:
DCO signoff added! Now I'll have to do the work to update to the most recent version to get this. I already started going through our lambdas and adding the description by hand. I just wish I could do this for the log groups but based on some cursory research that's just not possible.

PR #7781, opened 2025-11-06
Fix autodownload race

noahsilas:
When adding a resource with multiple attachments subject to the Autodownloader, each attachment enqueues a Download job. These jobs can run in parallel, pulling the remote asset into Storage and then trying to update the reference to the asset in the resource.

There are two flavors of race happening here:

- A given worker would read the Resource early in its lifetime, import  the file into storage, and then try to update and write back the  entire resource. This could cause it to overwrite other updates already applied by the related workers.

  This has been resolved by switching from `updateResource` to a more highly targeted `patchResource` operation. Further, we refresh the resource before generating the patch so that we minimize the chance that we are patching over a mutated copy.

- When a worker is committing a change, the autodownload scanner sees the update and runs again. Unprocessed files are spotted and matched again, enqueuing duplicate jobs for them.

  To prevent this I have added a check to only enqueue the autodownloader job for a URL when it is in a field that was added or changed from the previous version.

Github-Issue: https://github.com/medplum/medplum/issues/7651

### DIFF
--- a/packages/server/src/workers/download.ts
+++ b/packages/server/src/workers/download.ts
@@ -311,10 +311,12 @@ export async function execDownloadJob<T extends Resource = Resource>(job: Job<Do
       return;
     }
 
-    await systemRepo.patchResource(resourceType, id, [
-      ...patches,
-      { op: 'replace', path: '/meta/author', value: { reference: 'system' } },
-    ]);
+    await systemRepo.patchResource(
+      resourceType,
+      id,
+      [...patches, { op: 'replace', path: '/meta/author', value: { reference: 'system' } }],
+      { ifMatch: resource.meta?.versionId }
+    );
   } catch (ex: any) {
     log.info('Download exception: ' + ex, ex);
     throw ex;


packages/server/src/workers/download.ts
```diff
@@ -253,28 +268,54 @@ export async function execDownloadJob<T extends Resource = Resource>(job: Job<Do
...
 }
 
-function getAttachments(resource: Resource): Attachment[] {
-  const attachments: Attachment[] = [];
+// Translate basic FHIRPath into JSON Patch pointer
+export function pathToPointer(path: string): string {
+  return (path.startsWith('.') ? path : `.${path}`)
+    .replaceAll('.', '/')
+    .replaceAll(/\[(\d+)]/g, (m, g1) => `/${g1}`) // convert `[0]` => `/0`
+}
```
noahsilas: This feels like something that should already have a more robust implementation somewhere, though I haven't found it yet. I'd appreciate references to existing work I can reuse.

Some alternative approaches:
- Teach the crawler to natively emit paths as JSON pointers
- Apply the FHIRPath directly instead of going through a JSON pointer at all. It looks like We could evaluate a path via (`evalFhirPath`)[https://github.com/medplum/medplum/blob/main/packages/core/src/fhirpath/parse.ts#L245], but it's not quite so obvious how to use this to patch the resource - we seem to only support JSON Patch in the API today.

packages/server/src/workers/download.ts
```diff
@@ -253,28 +268,54 @@ export async function execDownloadJob<T extends Resource = Resource>(job: Job<Do
...
     log.info('Download exception: ' + ex, ex);
     throw ex;
   }
 }
 
-function getAttachments(resource: Resource): Attachment[] {
-  const attachments: Attachment[] = [];
+// Translate basic FHIRPath into JSON Patch pointer
+export function pathToPointer(path: string): string {
+  return (path.startsWith('.') ? path : `.${path}`)
```
noahsilas: I _think_ it's the case that a FHIRPath never starts with a separator but JSON Pointer must be empty or start with `/`. So we probably can remove this conditional and always apply it as a prefix?

packages/server/src/workers/download.ts
```diff
@@ -104,13 +110,22 @@ export async function addDownloadJobs(resource: WithId<Resource>, context: Backg
...
-    if (!isExternalUrl(attachment.url) || !isUrlAllowedByProject(project, attachment.url)) {
+    // Only process allowed external URLs
+    const url = attachment.value.url;
+    if (!isExternalUrl(url) || !isUrlAllowedByProject(project, url)) {
+      continue
+    }
+
+    // Skip if this mutation didn't adjust the URL in question
+    const pointer = Pointer.fromJSON(`${pathToPointer(attachment.path)}/url`);
+    if (pointer.get(resource) === pointer.get(previousVersion)) {
```
noahsilas, line 136: There are some instances where this doesn't work as expected. For example, if you delete the first entry in an array, that's going to look to this code like all the remaining objects were mutated.

I _think_ that the outcome in that kind of scenario is mild enough not to warrant spending additional time fixing. It would mean that jobs for each remaining URL could be double-enqueued, which could result in creating extraneous Binary resources that have no references to them. This is probably worth calling out in a comment as a tradeoff we're explicitly making.

packages/server/src/workers/download.ts
```diff
@@ -253,28 +268,54 @@ export async function execDownloadJob<T extends Resource = Resource>(job: Job<Do
...
+    resource = await systemRepo.readResource<T>(resourceType, id);
+
+    const attachments = getAttachments(resource);
+    const patches = attachments
+      .filter(attachment => attachment.value.url === url)
+      .map(value => ({
+        op: 'replace' as const,
+        path: `${pathToPointer(value.path)}/url`,
+        value: `Binary/${binary.id}`,
+      }));
```
noahsilas, line 299: I think we should bail out if `patches.length === 0`. This would mean that since we started this job someone else has already removed the URL we fetched from the resource. We might even want to destroy the item we just wrote into storage since it won't have any pointers to it.

packages/server/src/workers/download.ts
```diff
@@ -253,28 +268,54 @@ export async function execDownloadJob<T extends Resource = Resource>(job: Job<Do
...
-    (updated.meta as Meta).author = { reference: 'system' };
-    await systemRepo.updateResource(updated);
+    // re-fetch resource so we are mutating as recent a copy as possible
+    // (there may have been other mutations applied while we were writing the
+    // object into storage)
+    resource = await systemRepo.readResource<T>(resourceType, id);
+
+    const attachments = getAttachments(resource);
+    const patches = attachments
+      .filter(attachment => attachment.value.url === url)
```
noahsilas: One slightly bothersome thing here is that `attachment.value` here is typed as an `any`, so we lost a little bit of type safety compared to before when we were explicitly casting that to an `Attachment` type.

It sort of feels like `TypedValue` should accept a type parameter (`TypedValue<Attachment>`), or maybe be defined more exactly as a discriminated union?

```
type TypedValue =
  | { type: 'Address', value: Address }
  | { type: 'Age', value: Age }
  ...
```

If folks have thoughts on this let me know, but if I do work on this I'll open a separate PR - it seems likely to balloon out a bit.

techdavidy:
@claude please check

packages/server/src/workers/download.ts
noahsilas: I've shifted this to be co-located with the crawler that emits these paths, and added some test coverage. @mattwiller, you mentioned that there may be more places where we have very similar code? My search didn't turn up similar implementations but I might be using poor search terms. If you have any pointers for anything you would like to see consolidated please let me know!

Commit here: https://github.com/medplum/medplum/pull/7781/commits/51e80fd0f46ba9f0824f7b24e4d6275d6b8148ab

mattwiller (approved):

PR #7782, opened 2025-11-06
Support streaming in $ai

techdavidy:


techdavidy:
@claude please review

packages/server/src/fhir/operations/ai.ts
```diff
@@ -51,6 +52,14 @@ const operation: OperationDefinition = {
       type: 'string',
       documentation: 'JSON string containing the tools array (optional)',
     },
+    {
+      name: 'stream',
```
ThatOneBro: This is a good idea but maybe we can just rely on the `Accept` header value (of `text/event-stream`) to trigger the streaming behavior?

packages/server/src/fhir/operations/ai.ts
```diff
@@ -75,79 +84,209 @@ type AIOperationParameters = {
...
-
-      parameters.push({
-        name: 'tool_calls',
-        valueString: JSON.stringify(toolCallsWithParsedArgs),
-      });
+    // Set SSE headers
+    res.setHeader('Content-Type', 'text/event-stream');
+    res.setHeader('Cache-Control', 'no-cache');
+    res.setHeader('Connection', 'keep-alive');
+    res.flushHeaders();
```
ThatOneBro, line 161: 👍 This looks right

packages/server/src/fhir/operations/ai.ts
```diff
@@ -75,79 +84,209 @@ type AIOperationParameters = {
...
+  res: ExpressResponse
+): Promise<void> {
+  const response = (await callAI(messages, apiKey, model, tools, true)) as Response;
+  if (!response.body) {
+    throw new Error('No response body available for streaming');
+  }
+
+  // Stream OpenAI response directly to client
+  const reader = response.body.getReader();
+  const decoder = new TextDecoder();
```
ThatOneBro: We might be able to use [TextDecoderStream](https://nodejs.org/api/webstreams.html#class-textdecoderstream) here

techdavidy:
Thanks for the feedback @ThatOneBro
@claude mind checking again please

packages/server/src/fhir/routes.ts
```diff
@@ -157,6 +157,38 @@ fhirRouter.use(protectedRoutes);
 // CSV Export (cannot use FhirRouter due to CSV output)
 protectedRoutes.get(['/:resourceType/$csv', '/:resourceType/%24csv'], csvHandler);
 
+// AI $ai operation - handled directly in Express routes to support streaming
+// protectedRoutes.post(['/$ai', '/%24ai'], aiOperation);
+protectedRoutes.post(['/$ai', '/%24ai'], async (req: Request, res: Response) => {
```
ThatOneBro: Can we pull this into the operation definition like `agentPushHandler`? That way all the implementation details are in one spot

So something like exporting an `aiOperationHandler` maybe?

packages/server/src/fhir/operations/ai.ts
```diff
@@ -161,11 +295,19 @@ export async function callAI(
...
+  // For streaming, return raw response
+  if (stream) {
+    return response;
+  }
+
+  // For non-streaming, parse and return structured data
   if (!response.ok) {
     const errorData = await response.json().catch(() => ({}));
-    throw new Error(
+    const error: any = new Error(
```
ThatOneBro: Instead of casting to any here, can we instead at line 309 do a type assertion? Something like:
`(error as Error & { statusCode: number }).statusCode = response.status`

packages/server/src/fhir/operations/ai.ts
```diff
@@ -79,75 +80,208 @@ type AIOperationParameters = {
...
-    const parameters: ParametersParameter[] = [];
-
-    if (result.content) {
-      parameters.push({
-        name: 'content',
-        valueString: result.content,
-      });
+  // Check Accept header for streaming
+  const acceptHeader = req.headers?.accept || req.headers?.Accept;
+  const wantsStreaming = acceptHeader?.includes('text/event-stream');
```
ThatOneBro: I think this can be:
```suggestion
  const wantsStreaming = req.header('Accept')?.includes('text/event-stream');
```

packages/server/src/fhir/operations/ai.ts
```diff
@@ -79,75 +80,208 @@ type AIOperationParameters = {
...
+
+            if (!delta?.content) {
+              continue;
+            }
+
+            // Forward content chunks to client
+            res.write(`data: ${JSON.stringify({ content: delta.content })}\n\n`);
+          } catch (e) {
+            // Skip malformed JSON
+            console.error('Error parsing SSE data:', e);
```
ThatOneBro: Oh actually two things here: 
1. We typically try to use `RequestContext.logger` or `globalLogger` in server. I think here context logger would be appropriate. Maybe:
`const ctx = getRequestContext();`
Some where at the top of this function and then replace:

```suggestion
            ctx.logger.error('Error parsing SSE data:', e);
```

ThatOneBro: Another thing is, I think that malformed JSON should probably result in an error here instead of continuing?

Would that be a critical error?

If so, I think you can call `res.end` and it should trigger the `onerror` callback for the client-side `EventSource`.

packages/server/src/fhir/operations/ai.ts
```diff
@@ -79,75 +80,208 @@ type AIOperationParameters = {
...
+          }
+        }
+      }
     }
+  } finally {
```
ThatOneBro, line 247: I'm thinking any error thrown in this `try` block might put the stream into a weird state 🤔 

Might want to add a catch here where we can at least log the error like above and then call `res.end` to trigger the error handling logic on the client

What do you think?

techdavidy: Claude mentioned we can potentially have a counter and after a max then end the connection. 
Potentially one bad json doesn't mean the rest would be the same.
Let's keep as is and we can monitor if we need to adjust.

ThatOneBro (approved):

PR #7795, opened 2025-11-07
Spaces - Display DiagnosisReport and ServiceRequest

techdavidy:
- Fetch DiagnosisReport for corresponding ServiceRequest 
- Display in spaces LabOrderDetails for ServiceRequest 
- Display in LabResultDetails for ServiceRequest

### DIFF
--- a/examples/medplum-provider/src/components/labs/LabOrderDetails.test.tsx
+++ b/examples/medplum-provider/src/components/labs/LabOrderDetails.test.tsx
@@ -612,7 +612,6 @@ describe('LabOrderDetails', () => {
       });
     });
 
-
     it('fetches and displays specimen label documents', async () => {
       const user = userEvent.setup();
       medplum.searchResources = vi.fn().mockImplementation(async (resourceType: string, params: any) => {
@@ -891,4 +890,3 @@ describe('LabOrderDetails', () => {
     });
   });
 });
-


techdavidy:
@claude please review

examples/medplum-provider/src/components/labs/LabOrderDetails.tsx
```diff
@@ -48,41 +47,34 @@ interface ProgressStep {
...
-      setActiveDetailTab('report');
-    }
-  }, [activeTab]);
+    const fetchPrimaryReport = async (): Promise<void> => {
+      const primaryReport = await medplum.searchResources('DiagnosticReport', {
+        'based-on': `ServiceRequest/${order.id}`,
+        _sort: '-_lastUpdated',
+        _count: 1,
+      });
+      setDiagnosticReports(primaryReport);
```
noahsilas, line 73: It looks like the only place this is invoked today uses the order ID as the `key`, so this should be well behaved today (a change to the order ID will instantiate an entirely new component instance and not adjust the props coming to an existing one). If we want to make this more future proof, we could cancel this `set` if the order id changes. Something like:

```ts
useEffect(() => {
  let cancelled = false;
  const primaryReport = await medplum.searchResources(...);
  if (!cancelled) { setDiagnosticReports(primaryReport); }
  return () => { cancelled = true; }
}, [medplum, order.id]);
```

examples/medplum-provider/src/components/labs/LabOrderDetails.tsx
```diff
@@ -755,7 +747,7 @@ export function LabOrderDetails({
             )}
 
             {/* Progress Tracker Tab Content - for open items */}
-            {activeDetailTab === 'progress' && activeTab === 'open' && (
+            {activeDetailTab === 'progress' && order.status !== 'completed' && (
```
noahsilas: On my first read through it seems like we shouldn't ever be in the state where we are on the "progress" tab with a "completed" order. So this might be more easily understood as exclusively switching on the `activeDetailTab` state without the extra condition here?

If we do get into that mismatched state, would there be a better way to recover than rendering nothing in this position?

examples/medplum-provider/src/pages/patient/LabsPage.tsx
```diff
@@ -73,38 +72,14 @@ export function LabsPage(): JSX.Element {
...
   const fetchData = useCallback(async (): Promise<void> => {
     setLoading(true);
     try {
-      await Promise.all([fetchOrders(), fetchDiagnosticReports()]);
+      await fetchOrders();
     } finally {
       setLoading(false);
     }
-  }, [fetchOrders, fetchDiagnosticReports]);
+  }, [fetchOrders]);
```
noahsilas, line 82: This wrapper feels a little extra now that it is only surrounding the single data operation. There could be potential for cleanup here. Maybe something like switching to the `useSearchResources` hook?

examples/medplum-provider/src/pages/spaces/ConversationList.tsx
```diff
@@ -33,13 +33,12 @@ export function ConversationList({ currentTopicId, onSelectTopic }: Conversation
 
   return (
     <Paper h="100%">
-      <Stack p="md">
-        <Text size="sm" fw={600} c="dimmed" mt="md">
-          Recent Conversations
-        </Text>
-
+      <Stack p={0} h="100%" style={{ display: 'flex', flexDirection: 'column' }}>
```
noahsilas: I would have expected that `<Stack>` sets CSS attributes like `display` and `flexDirection` automatically, is this doing some kind of clever override I can learn more about?

Actually, digging deeper, this stack only seems to have one child, maybe we can remove it entirely?

examples/medplum-provider/src/pages/spaces/SpacesPage.tsx
```diff
@@ -241,68 +241,111 @@ export function SpacesPage(): JSX.Element {
...
+            style={{
+              ...styles,
+              position: 'absolute',
+              left: 0,
+              top: 0,
+              zIndex: 10,
+              display: 'flex',
+              flexDirection: 'column',
+            }}
+          >
```
noahsilas, line 333: It seems like there might be an opportunity to extract an `<OverlaySidebar>` component that can be reused with a bunch of this shared styling.

noahsilas (commented):
Overall this looks good to me - the strategy seems sound and the code looks reasonable. 

I have a couple minor thoughts/questions, mostly cleanup opportunities, nothing that I think would be blocking for this PR.

It looks like this branch had a merge from `main` pulled in, causing the commit list to look somewhat awkward and making it a bit harder to isolate your changes for review in context. Would you be interested in spending a couple minutes together to rebase this change and get those other commits out of this history?

examples/medplum-provider/src/pages/patient/LabsPage.tsx
techdavidy: fetchData is all called when creating a new lab order to refresh the entries.

examples/medplum-provider/src/pages/spaces/SpacesPage.tsx
techdavidy: Actually there is Mantine Drawer but this is a placeholder for now.

examples/medplum-provider/src/components/labs/LabOrderDetails.tsx
techdavidy: yeah removed the  `order.status !== 'completed'`

examples/medplum-provider/src/pages/spaces/ConversationList.tsx
techdavidy: Thanks used <Flex> instead

examples/medplum-provider/src/pages/patient/LabsPage.tsx
noahsilas: Ah, I see. Thanks!

ThatOneBro (approved):
LGTM 👍 

Thanks to @noahsilas for initial feedback

PR #7799, opened 2025-11-10
feat: Implement Group $everything operation

jdjkelly:
## Summary

Implements the FHIR R4 `Group $everything` operation as specified in https://hl7.org/fhir/R4/group-operation-everything.html

This operation returns all resources related to all patients in a group, enabling bulk data retrieval for groups of patients.

### DIFF
--- a/packages/server/src/fhir/operations/groupeverything.test.ts
+++ b/packages/server/src/fhir/operations/groupeverything.test.ts
@@ -270,4 +270,90 @@ describe('Group Everything Operation', () => {
     expect(resourceRefs).toContain(getReferenceString(group));
     expect(resourceRefs).toContain(getReferenceString(practitioner));
   });
+
+  test('Pagination with _count and _offset', async () => {
+    // Create patients
+    const patient1Res = await request(app)
+      .post(`/fhir/R4/Patient`)
+      .set('Authorization', 'Bearer ' + accessToken)
+      .set('Content-Type', ContentType.FHIR_JSON)
+      .send({
+        resourceType: 'Patient',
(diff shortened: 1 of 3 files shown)


jdjkelly:
Would love to get this merged @ThatOneBro :)

ThatOneBro:
Hey @jdjkelly! Sorry for the delayed feedback on your PR, and thanks for the bump 👍 

How would you feel if we added automatic pagination to the Bundle like we do for Patient $everything by default, otherwise if you send the `Prefer: respond-async` header, we return the full unpaginated Bundle?

I talked with the team and we speculated that the operation might be prone to eating a lot of memory without pagination and then still timing out before the operation can complete and throwing all the work away at the end

We recently discussed the shape of the SQL queries for Patient $everything being suboptimal right now and so issuing N of those queries for this one operation might not be ideal for large deployments

jdjkelly:
> How would you feel if we added automatic pagination to the Bundle like we do for Patient $everything by default, otherwise if you send the `Prefer: respond-async` header, we return the full unpaginated Bundle?

@ThatOneBro This makes sense to me (on both points)

I agree with the other notes

One thing that's interesting too to consider is that Bulk Exports by Group ID are possible now... I think in the case of large population usage / context you'd want the Bulk Export path anyways. 

The use case for Group $everything as a REST operation is closer to "there's 2+ Patients (ie. parent / child) whose data we want to access together" (this is our specific use case - where we actually no control over the number of Patient resources defined by some upstream data source. For a given patient access query, we can get several Patients representing people subscribed to a particular health plan for example.

codyebberson (approved):

PR #7800, opened 2025-11-10
Dependency upgrades 2025-11-10

medplumbot:
Dependency upgrades


ThatOneBro:
Closing this one out since not much change (although there was license policy warning 👀 ) and we cut a release which created a lot of conflicts. 

Also next 16 seems like it still needs some work. We can revisit next week when more people are back online.

PR #7814, opened 2025-11-11
Move set-accounts async flow into BullMQ job

mattwiller:


ThatOneBro (approved):
Nice, LGTM

Aligned with the existing batch implementation 👍

PR #7815, opened 2025-11-11
fix(definitions): include FHIR JSON definitions in NPM package

The-Alchemist:
it appears they were missed, as they are missing from the published modules on npmjs


ThatOneBro:
Thanks @The-Alchemist 👍

Issue #7641, opened 2025-10-20
Access policy discrepancy in GraphQL and REST


ianplunkett:
# AccessPolicy with chained search criteria not applied during GraphQL reference resolution

## Summary

AccessPolicies using chained search criteria (e.g., `subject:Patient.general-practitioner=%profile`) appear to work for REST API searches but are not applied during GraphQL reference resolution. However, [Medplum's documentation](https://www.medplum.com/docs/access/access-policies#criteria-based-access-control) states that chained searches are not supported in AccessPolicy criteria, creating confusion about expected behavior.

## Issue Description

This issue highlights an inconsistency in how AccessPolicy criteria are handled:

1. **REST API behavior**: Chained search criteria in AccessPolicies appear to work in some cases for REST searches
2. **GraphQL behavior**: The same criteria are not applied during GraphQL reference resolution  
3. **Documentation**: States that chained searches are not supported in AccessPolicy criteria

This creates confusion about whether chained search criteria should work at all, and if they do work in REST, why they don't work consistently in GraphQL.

## Expected Behavior

AccessPolicy criteria should be applied consistently across both REST API and GraphQL API, including when resolving nested references in GraphQL queries.

## Actual Behavior

- REST API searches correctly apply AccessPolicy criteria with chained parameters
- GraphQL reference resolution bypasses AccessPolicy criteria and attempts direct resource access
- This results in "Not found" errors in GraphQL for resources that are accessible via REST

## Environment

- Medplum version: 4.5.1
- API: Both REST and GraphQL

### 1. Test Data Setup

Create the test data using conditional references across multiple bundles:

**Step 1: Create Practitioners**
```json
{
  "resourceType": "Bundle",
  "type": "batch",
  "entry": [
    {
      "request": {
        "method": "POST",
        "url": "Practitioner",
        "ifNoneExist": "identifier=http://example.org/practitioner-ids|PRAC001"
      },
      "resource": {
        "resourceType": "Practitioner",
        "identifier": [
          {
            "system": "http://example.org/practitioner-ids",
            "value": "PRAC001"
          }
        ],
        "name": [
          {
            "family": "Smith",
            "given": ["John"]
          }
        ]
      }
    },
    {
      "request": {
        "method": "POST",
        "url": "Practitioner",
        "ifNoneExist": "identifier=http://example.org/practitioner-ids|PRAC002"
      },
      "resource": {
        "resourceType": "Practitioner",
        "identifier": [
          {
[...]

PR #7816, opened 2025-11-11
Explicitly disallow chained search in AccessPolicy

mattwiller:
Partially addresses #7641

### DIFF
--- a/packages/definitions/src/fhir/r4/profiles-medplum.json
+++ b/packages/definitions/src/fhir/r4/profiles-medplum.json
@@ -3669,7 +3669,7 @@
                 "key": "axp-3",
                 "severity": "error",
                 "human": "Criteria must be a valid FHIR search string on the correct resource type, e.g. Patient?identifier=123",
-                "expression": "criteria.exists() implies criteria.matches('^' & %context.resourceType & '\\?([a-zA-Z_-]+=[^&]+&?)+')"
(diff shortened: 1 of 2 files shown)


packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -3669,7 +3669,7 @@
                 "key": "axp-3",
                 "severity": "error",
                 "human": "Criteria must be a valid FHIR search string on the correct resource type, e.g. Patient?identifier=123",
-                "expression": "criteria.exists() implies criteria.startsWith(%context.resourceType & '?')"
+                "expression": "criteria.exists() implies criteria.matches('^\\\\' & %context.resourceType & '\\\\?([a-zA-Z_:-]+=[^&]+&?)+') and criteria.contains('_has:').not()"
```
noahsilas, line 3672: Leaving some notes from an in-person conversation with @mattwiller covering my understanding:
- This string is intentionally double escaped, as it is an expression encoded in a JSON file
- `context.resourceType` might be the wildcard `*`, which we would want to match with an expression like `^\*`. In the case where it isn't a wildcard, escaping the first character of a resource name (`^\Patient`) is going to have no effect
- The next regex component matches query parts, and intentionally omits `.` as a character in the query key matcher, preventing chaining
- Finally we don't allow reverse chaining by disallowing `_has:` expressions

PR #7825, opened 2025-11-12
Only apply defaultPatientAccessPolicy when userType is "Patient"

### LINKED ISSUE #7761: [Bug] SCIM requires defaulPatientAccessPolicy for Practitioner Users
User Driven: https://discord.com/channels/905144809105260605/1435213007859880087/1435213007859880087

noahsilas:
When creating a user of type "Practitioner", you should not need to have a `defaultPatientAccessPolicy` defined, and if it is defined it should not be applied to the new user.

When creating users of type "Patient" we do continue to enforce this policy exists and automatically attach it.

Resolves #7761. Addresses some cleanup opportunities spotted while researching the problem in the first commits in the chain.

### DIFF
--- a/packages/server/src/scim/routes.ts
+++ b/packages/server/src/scim/routes.ts
@@ -1,6 +1,13 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { badRequest, getStatus, isReference, normalizeErrorString, normalizeOperationOutcome, OperationOutcomeError } from '@medplum/core';
+import {
+  badRequest,
+  getStatus,
+  isReference,
+  normalizeErrorString,
+  normalizeOperationOutcome,
+  OperationOutcomeError,
+} from '@medplum/core';
 import type { User } from '@medplum/fhirtypes';
 import type { NextFunction, Request, RequestHandler, Response } from 'express';
 import { Router } from 'express';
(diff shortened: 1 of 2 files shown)


mattwiller (approved):

packages/server/src/scim/routes.test.ts
```diff
@@ -166,7 +166,7 @@ describe('SCIM Routes', () => {
     expect(patchResponse.body.active).toBe(false);
   });
 
-  test.skip('Create missing medplum user type', async () => {
+  test('Create, missing medplum user type, creates a Practitioner', async () => {
```
rahul1, line 169: is there a reason for this change in behavior? My preference would be to leave the existing behavior for missing user type as is. 

Because we don't yet _require_ Practitioners to have access policies, we want to avoid a misconfiguration leading to a user being given "Practitioner" status

rahul1 (commented):
See comment about default user tyep

packages/server/src/scim/routes.test.ts
noahsilas: This actually already is the default behavior, I'm updating this test to match it. It changed in #4134 (see https://github.com/medplum/medplum/pull/4134/files#diff-e40593c8cfa0f88e9466ecbdde25367e42fde68e5dbf33fa4ab57e276881f420R199-R200).

That said, up until now if you called this endpoint without a user type, it would have created a Practitioner that had the default patient access policy applied (if one was defined). So there is potentially a regression if someone is depending on that behavior 🤔 

Any suggestions on the best thing to do here?

rahul1: Good question. Think best would be to throw the 400 as the original (skipped) test had it written. For this kind of stuff related to auth access, generally prefer everything to be super explicit vs. allowing implicit behavior

PR #7827, opened 2025-11-13
Return result value paths from FHIRPath eval

mattwiller:


packages/core/src/fhirpath/functions.ts
```diff
@@ -314,7 +314,7 @@ export const functions: Record<string, FhirPathFunction> = {
    * @returns A collection containing only those elements in the input collection for which the stated criteria expression evaluates to true.
    */
   select: (context: AtomContext, input: TypedValue[], criteria: Atom): TypedValue[] => {
-    return input.map((e) => criteria.eval({ parent: context, variables: { $this: e } }, [e])).flat();
+    return input.flatMap((e) => criteria.eval({ parent: context, variables: { $this: e } }, [e]));
```
ThatOneBro, line 317: 👍

packages/core/src/fhirpath/parse.ts
```diff
@@ -279,8 +280,14 @@ export function evalFhirPathTyped(
...
-  return ast.eval({ variables }, input).map((v) => ({
-    type: v.type,
-    value: v.value?.valueOf(),
-  }));
+  return ast.eval({ variables }, input).map((v) => {
+    const result: TypedValue & { path?: string } = {
+      type: v.type,
+      value: v.value?.valueOf(),
+    };
+    if ('path' in v && typeof v.path === 'string') {
```
ThatOneBro: Is it possible for `v.path` to not be a string if `path` is defined? Or can we just remove the typeof check and assert it as a string?

mattwiller: Good call, I replaced the check with an `as string` assertion

ThatOneBro (approved):

packages/core/src/fhirpath/parse.ts
ThatOneBro: Nice, thank you 🙏

PR #7830, opened 2025-11-13
$update-email documentation

finnbergquist:


finnbergquist:
Relevant docs [here](https://medplum-www-git-finn-update-email-operation-docs-medplum.vercel.app/docs/api/fhir/operations/user-update-email)

rahul1 (approved):

packages/docs/docs/api/fhir/operations/user-update-email.md
```diff
@@ -0,0 +1,125 @@
...
+
+```typescript
+await medplum.post(`fhir/R4/User/${userId}/$update-email`, {
+  resourceType: 'Parameters',
+  parameter: [
+    { name: 'email', valueString: 'newemail@example.com' },
+    { name: 'updateProfileTelecom', valueBoolean: true }
+  ]
+});
+```
```
mattwiller, line 59: Ideally this example would live in the `examples/` directory so it can be type-checked automatically, but I think this small case is fine

PR #7837, opened 2025-11-13
Display email in use on password form

noahsilas:
I mistyped my own email address, and was surprised when my login didn't work. It took me an embarrassingly long time to figure out what had happened.

- Added a small notice about what email address is being used, which helps someone like me notice that they typo'd on the previous screen

- Added a "change" button into that notice so that if you realize you made a mistake there is an obvious way to fix it.

## UI Screenshots
login email notice in light mode
login email notice in dark mode

## Storybook color palette
@techdavidy correctly pointed out that I hadn't considered how this looks in dark mode. To make supporting that easier, I've added a color scheme toggle to the storybook as well as a color palette page.

Storybook Color Palette

### DIFF
--- a/packages/react/src/auth/AuthenticationForm.tsx
+++ b/packages/react/src/auth/AuthenticationForm.tsx
@@ -11,7 +11,7 @@ import {
   PasswordInput,
   Stack,
   Text,
-  TextInput
+  TextInput,
 } from '@mantine/core';
 import type {
   BaseLoginRequest,


mattwiller (approved):

techdavidy:
mind adding a test please

packages/react/src/auth/AuthenticationForm.tsx
```diff
@@ -186,6 +187,10 @@ export function PasswordForm(props: PasswordFormProps): JSX.Element {
           autoFocus={true}
           error={getErrorsForInput(outcome, 'password')}
         />
+        <Flex justify="space-between" align="center" px="xs" bg="gray.0" bdrs="sm">
```
techdavidy: let's add support for dark mode too

techdavidy (changes requested):
Image

we can add a module css for dark mode support

packages/react/src/auth/AuthenticationForm.tsx
```diff
@@ -186,6 +187,10 @@ export function PasswordForm(props: PasswordFormProps): JSX.Element {
...
+          <small>Logging in as <em>{props.email}</em></small>
```
techdavidy: maybe 
```<Text size="xs">``` ?

File: packages/react/src/auth/AuthenticationForm.tsx (hunk 1 above)
Comment (noahsilas) [reply packages/react/src/auth/AuthenticationForm.tsx]: I took a quick stab at this in https://github.com/medplum/medplum/pull/7837/commits/e8a0ff000cd3b7d45952242b2531766bfa5b0cf0.

This first attempt seems fine, but not great. In another comment you mentioned that this could be don with a CSS module - do you have a pointer to somewhere doing that I can crib from?

I wonder if we should extract some kind of "<SubtleNotice>" component or similar that will always choose an appropriate background color? Seems overkill if this is the only place we're doing it though.

File: packages/react/src/auth/AuthenticationForm.tsx [hunk 3]
```diff
@@ -173,19 +188,36 @@ export function PasswordForm(props: PasswordFormProps): JSX.Element {
...
           required={true}
           autoFocus={true}
           error={getErrorsForInput(outcome, 'password')}
+          data-testid="auth.password"
         />
+        <Flex
+          justify="space-between"
+          align="center"
+          px="xs"
+          bg={colorScheme === 'light' ? 'gray.0' : 'gray.8'}
```
techdavidy: the css with Mantine way to achieve this would be 
```
.panel {
  background : light-dark(var(--mantine-color-gray-0), var(--mantine-color-gray-8));
}
```

techdavidy:
when I test the storybook preview goes directly to dark mode. Can we default to light mode please?

packages/react/src/auth/AuthenticationForm.tsx
noahsilas: Ah, of course! Thank you, updated in 2e152efc3

noahsilas:
> when I test the storybook preview goes directly to dark mode. Can we default to light mode please?

This should be remembering your last chosen theme, and using the OS default if you haven't chosen one (https://github.com/vuelessjs/storybook-dark-mode?tab=readme-ov-file#default-theme). Did you experience the default dark theme on your initial load? Is your OS preference in light or dark mode at the time it loaded?

PR #7848, opened 2025-11-17
ci(test): lean into turbo, set turbo concurrency to 1

ThatOneBro:
This is just a test to see how this setup performs in CI. Theoretically this simplifies the concurrency model and allows Jest to handle concurrency on a per task basis which should theoretically help prevent resource exhaustion on the smallish GitHub CI boxes


ThatOneBro:
Looks like this is not working as intended. Closing for now

Issue #7572, opened 2025-10-09
Additional invite validation if User already exists in project


finnbergquist:
### Current behavior

If a project scoped User is invited to a project and they already have an existing project scoped User in the project with the same email, they will get an error message, unless they specify `forceNewMembership = true` on the `.../invite`.

If a project scoped User is invited to a project, and an existing server scoped User is in the project with the same email, then no validation is triggered. After that, when the User goes to login, it is there is no way to designate which User & ProjectMembership they want to login with.

There isn't a way for this User to specify whether they want to login with their porject scoped User or server scoped User. It is possible to add a client side login switcher, using `setActiveLogin()` Example [here](https://github.com/medplum/medplum/blob/b132fea7f9c732d541e05e58e1f1db4396aacffd/packages/react/src/AppShell/HeaderDropdown.tsx#L44-L49), but this is a somewhat advanced feature for people to add to their app.

### Desired Behavior

I think that it _should_ be ok for a User to have two ProjectMemberships in the same project, but we should make sure that people are doing it **intentionally**.

This scenario **where a project scoped User is invited to a project, and an existing server scoped User is invited to the project with the same email** should also require `forceNewMembership = true` so that people do not accidentally end up in this state. The same is true for the reverse of this order where a server scoped User is in the project, and a project scoped User now gets invited.

noahsilas:
Closed via #7852

PR #7852, opened 2025-11-17
Validate invited users don't exist at different scope

noahsilas:
There is potential for confusion when someone has two User records that are both tied to the same email address and connected to the same project.

While this is a valid state to be in, we want to ensure that folks are reaching it intentionally. To this end, we add a new validation:

- When creating a project-scoped user, ensure that there does not already exist a membership connected to a server-scoped user with the same email address

- When creating a server-scoped user, ensure that there does not already exist a membership connected to a project-scoped user with the same email address

Github-Issue: #7572

### DIFF
--- a/packages/server/src/admin/invite.ts
+++ b/packages/server/src/admin/invite.ts
@@ -81,52 +81,55 @@ export async function inviteUser(request: ServerInviteRequest): Promise<ServerIn
   let existingUser = false;
   let passwordResetUrl: string | undefined;
 
-  // Instantiate unsaved user record from the request
-  const userResource = await makeUserResource(request);
-
-  // If inviting with an email address, check for existing memberships tied to
-  // this project/email combination that are at a different scope than the one
-  // we would create. This avoids confusion of someone having separate
-  // server-scoped and project-scoped user records.
-  //
-  // This check is bypassed if the caller explicitly passes `forceNewMembership: true`
-  if (!request.forceNewMembership && userResource.email) {
-    const projectFilter = userResource.project
-      ? { code: 'user:User.project', operator: Operator.MISSING, value: 'true' }
-      : { code: 'user:User.project', operator: Operator.EXACT, value: `Project/${project.id}` };
-
-    const existingMemberships = await systemRepo.searchResources<ProjectMembership>({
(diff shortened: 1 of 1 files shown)


packages/server/src/admin/invite.ts
```diff
@@ -81,8 +81,35 @@ export async function inviteUser(request: ServerInviteRequest): Promise<ServerIn
   let existingUser = false;
   let passwordResetUrl: string | undefined;
 
-  // Upsert User resource
+  // Instantiate unsaved user record from the request
   const userResource = await makeUserResource(request);
+
+  // If inviting with an email address, check for existing memberships tied to
+  // this project/email combination that are at a different scope than the one
+  // we would create. This avoids confusion of someone havnig separate
```
mattwiller:
```suggestion
  // we would create. This avoids confusion of someone having separate
```

packages/server/src/admin/invite.ts
```diff
@@ -81,8 +81,35 @@ export async function inviteUser(request: ServerInviteRequest): Promise<ServerIn
...
+      : { code: 'user:User.project', operator: Operator.EXACT, value: `Project/${project.id}` };
+
+    const existingMemberships = await systemRepo.searchResources<ProjectMembership>({
+      resourceType: 'ProjectMembership',
+      filters: [
+        { code: 'user:User.email', operator: Operator.EXACT, value: userResource.email },
+        { code: 'project', operator: Operator.EXACT, value: `Project/${project.id}` },
+        projectFilter,
+      ],
+    });
```
mattwiller: Does this need to be wrapped in a serializable transaction with the upsert below, to guarantee safety against race conditions?

noahsilas: Good question. That would prevent a race where two users with the same email but different scopes are created at almost the same time. My intuition is that this is unlikely, but if it's easy to add this safety in that seems like a nice win.

It looks like the repo supports nesting `withTransaction()`  calls, I'll take a quick pass at this adjustment.

noahsilas: Done in commit 4cc4d7bf44, thanks!

PR #7863, opened 2025-11-19
Do not trigger subscriptions for websocket subscriptions

mattlong:


packages/server/src/workers/subscription.test.ts
```diff
@@ -1826,7 +1827,54 @@ describe('Subscription Worker', () => {
...
+          expect(subscription).toBeDefined();
+          expect(subscription.id).toBeDefined();
+
+          const queue = getSubscriptionQueue() as any;
+          queue.add.mockClear();
+
+          const sub = await wsSubRepo.createResource<Subscription>({
+            resourceType: 'Subscription',
+            status: 'active',
+            reason: "raison d'être",
```
ThatOneBro, line 1863: 🇫🇷 🥖

packages/server/src/workers/subscription.test.ts
```diff
@@ -1826,7 +1827,54 @@ describe('Subscription Worker', () => {
         ]);
       }));
 
-    test('WebSocket Subscription -- Feature Flag Not Enabled', () =>
+    test.each([
+      [{ type: 'websocket' }, false],
```
ThatOneBro: It's the exclusion of WebSocket that it is important since it's the only one that won't be stored in the DB 👍 

Might want to put a comment about that?

PR #7866, opened 2025-11-19
feat(agent): add `assignSeqNo` for channels

### LINKED ISSUE #7785: [Agent] Add `serializedMode` to channel endpoints for end-to-end serialized processing
When using [`enhanced` ACK mode](https://www.medplum.com/docs/agent/acknowledgement-modes#enhanced-acknowledgement-mode-fast-ack), which is part of the HL7v2 spec, it's possible in flows that push messages to other remote systems on receipt of an inbound HL7 message ( **Remote System 1** -> **Agent 1** -> **Bot** -> **Agent 2** -> **Remote System 2** ) that messages at the final destination can arrive out of order, especially when there are variable processing times within bots. 

This can make it hard to process messages in order since one logical inbound channel from remote system 1 can produce messages that arrive out of order from the perspective of **Remote System 2**, due to variance in network connectivity, processing speed of the **Bot** (earlier bot executions can finish after subsequent executions).

However, there is a mechanism in place in the HL7v2 spec which allows for explicit serial processing via a sequence number in `MSH.13`. See: https://hl7-definition.caristix.com/v2/HL7v2.2/Fields/MSH.13

This is a two part solution:
1. **[THIS ISSUE]** - Make it so that a channel can automatically add (or replace) an incoming sequence number on a received message in the order that they are received on the channel. This means that even with the `enhanced` ACK mode turned on, messages will still be numerically sequenced in order from the beginning. This is not the default so if sequence numbers are already being provided by **Remote System 1**, this is not necessary as long as **Agent 2** is configured properly.
2.  #7784 - Make is so that an outbound client (for `pushToAgent` functionality) honors a `Device`'s configuration for `serializedMode` and waits to send a message when a message with a sequence number higher than the next expected sequence number is received. This makes sure that `Remote System 2` will always receive the messages in the order they were received from `Remote System 1` and hence will be processed by the system serially.

ThatOneBro:
Fixes #7785

### DIFF
--- a/packages/agent/src/hl7.ts
+++ b/packages/agent/src/hl7.ts
@@ -38,7 +38,7 @@ export class AgentHl7Channel extends BaseChannel {
   stats?: ChannelStatsTracker;
   private appLevelAckMode: AppLevelAckMode = 'AL'; // Default app level ack mode is AL (Always)
   private assignSeqNo: boolean = false;
-  private lastSequenceNo = -1;
+  private lastSeqNo = -1;
 
   constructor(app: App, definition: AgentChannel, endpoint: Endpoint) {
     super(app, definition, endpoint);
@@ -82,8 +82,8 @@ export class AgentHl7Channel extends BaseChannel {
     return this.assignSeqNo;
   }
 
-  takeNextSequenceNo(): number {
-    return ++this.lastSequenceNo;
+  takeNextSeqNo(): number {
+    return ++this.lastSeqNo;
   }
 
   sendToRemote(msg: AgentTransmitResponse): void {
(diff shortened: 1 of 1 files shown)


packages/agent/src/hl7.ts
```diff
@@ -173,6 +184,12 @@ export class AgentHl7Channel extends BaseChannel {
     }
 
     this.appLevelAckMode = parseAppLevelAckMode(appLevelAckRaw, this.log);
+    this.assignSeqNo = assignSeqNo;
+
+    // If not in serialized mode, set lastSerialNo to 0
```
mattwiller:
```suggestion
    // If not in serialized mode, set lastSerialNo to -1
```

PR #7874, opened 2025-11-20
Sonar fixups

noahsilas:
Best reviewed commit-by-commit. A handful of mechanical transformations to kill some issues that Sonar has flagged.


packages/hl7/src/connection.ts
```diff
@@ -290,7 +290,7 @@ export class Hl7Connection extends Hl7Base {
   async close(): Promise<void> {
     // If we have already received the close event, then we can just return immediately
     if (this.isClosed()) {
-      return Promise.resolve();
```
ThatOneBro, line 293: This is there for our "consistent return" lint rule I believe. Technically returning undefined is slightly different than returning a resolved void promise

noahsilas: Inside an async function it shouldn't make a difference, but the lint rule doesn't seem to be quite context aware enough to catch this. I'm going to change the end of this function from `return Promise(resolve => ...)` to `await Promise(resolve => ...)`, which makes this function internally consistent, and avoids leaking any of the details from socket.close through the void promise.

techdavidy:
nice thanks

PR #7901, opened 2025-11-24
Fixes #7896 - remove Next.js examples from monorepo

### LINKED ISSUE #7896: Remove Next.js examples from the monorepo
Moving the Next.js examples to dedicated repos here:
* https://github.com/medplum/medplum-nextjs-demo
* https://github.com/medplum/medplum-nextauth-demo

While Medplum does work with Next.js and SSR, it is an advanced and complicated setup, which we do not recommend lightly.  It is not a priority for us to upgrade the Next.js examples to the most recent version of Next.js, so these are accumulating some debt.

Next.js also represents ~300 MB out of 1.6 GB in `node_modules` (~19%):

Image

(Image generated by [node-modules-inspector](https://github.com/antfu/node-modules-inspector))

Context:

* Dependency upgrade challenges: https://medplum.slack.com/archives/C051SKJKZ50/p1763404551939459
* Impact on `node_modules`: https://medplum.slack.com/archives/C051SKJKZ50/p1752685695707319
* Bloating turborepo cache: https://medplum.slack.com/archives/C051SKJKZ50/p1741741159371169

codyebberson:
Moving the Next.js examples to dedicated repos here:
* https://github.com/medplum/medplum-nextjs-demo
* https://github.com/medplum/medplum-nextauth-demo

After:

image


codyebberson:
@noahsilas - yes, exactly!

> If that's right, this all seems good! We probably should follow up with a note in those repo README files to let folks finding them know that they are not getting the same level of support / updates anymore.

That's a great suggestion, good call.

PR #7911, opened 2025-11-25
[Documentation] Add undocumented search parameters

rahul1:


packages/examples/src/search/advanced-search-parameters.ts
```diff
@@ -444,3 +444,55 @@ curl 'https://api.medplum.com/fhir/R4/Patient?_source=https://foomedical.com' \
...
+
+/*
+// start-block typeCli
+medplum get 'fhir/R4?_type=Patient,Observation&name=Smith'
+// end-block typeCli
+
+// start-block typeCurl
+curl 'https://api.medplum.com/fhir/R4?_type=Patient,Observation&name=Smith' \
+	-H 'authorization: Bearer $ACCESS_TOKEN' \
+  -H 'content-type: application/fhir+json' \
```
noahsilas: This question is purely for my education: is there some tricky reason that in these examples one of these lines uses a hard tab and one uses two spaces for indentation? Or is this just something that started that way once and has propagated through the examples via copy/paste?

rahul1: it might be a cursor artifact :P

packages/docs/docs/search/advanced-search-parameters.mdx
```diff
@@ -302,3 +306,82 @@ The `_filter` parameter can be used to filter for more complex queries. For more
 ## \_sort
 
 The `_sort` parameter allows you to sort the results of your search based on different parameters. For details on how to use the `_sort` parameter, see the [Sorting the Results docs](/docs/search/basic-search#sorting-the-results).
+
+## \_deleted
+
+The `_deleted` parameter allows you to search for resources that have been soft-deleted. By default, Medplum filters out deleted resources.
```
mattlong, line 312: Probably worth noting that mixing `_deleted` with most other search parameters (besides `_project` and maybe one or two other special search params) will not work as expected since the DB columns backing other search parameters are not populated in deleted resources.

codyebberson (approved):

PR #7913, opened 2025-11-25
Ensure resources references by ProjectMembership can't be deleted

### LINKED ISSUE #7649: Prevent Deletion of ProfileResources with Active ProjectMemberships
Currently, it is possible to delete a `ProfileResource` (Patient, Practitioner, or RelatedPerson) even if there are active `ProjectMembership` resources associated with it. This can lead  inconsistencies within the system that affect login.

**Proposal**
We would like to enhance the deletion logic to prevent the deletion of a `ProfileResource` if there is at least one `ProjectMembership` resource that references it. When a user attempts to delete such a `ProfileResource`, the system should return an error (e.g., an OperationOutcome) indicating that the resource cannot be deleted due to existing `ProjectMembership` references.

To avoid breaking existing customers, this functionality should be controlled by a new **project-level setting**, for example, `preventProfileDeletionWithMembership`.

This can be seen as a special case of a broader future feature to prevent the deletion of *any* resource that has a reference pointing to it, thus enforcing referential integrity across the FHIR data store.

*   **Default Value:** `false` for existing projects and `true` for newly created projects. This allows existing customers to opt-in at their convenience, while new projects benefit from enhanced data integrity from the start.

**Impact on `auth/me`:**

The `/auth/me` endpoint, retrieves the user's ProjectMembership and then tries to read the associated profile resource:
If the ProfileResource has been deleted, systemRepo.readReference will fail, causing a server error and preventing the user from accessing the application.

*   **Current Deletion Logic and Referential Integrity:**
    The core `deleteResource` method is located in the `Repository` class.
    [https://github.com/medplum/medplum/blob/main/packages/server/src/fhir/repo.ts#L1173](https://github.com/medplum/medplum/blob/main/packages/server/src/fhir/repo.ts#L1173)

    Currently, referential integrity is explicitly *not* supported for deletes, as noted in the documentation:
    [https://github.com/medplum/medplum/blob/main/packages/docs/docs/fhir-datastore/deleting-data.md#L30-L34](https://github.com/medplum/medplum/blob/main/packages/docs/docs/fhir-datastore/deleting-data.md#L30-L34)

mattwiller:
Resolves #7649

### DIFF
--- a/packages/server/src/fhir/precommit.ts
+++ b/packages/server/src/fhir/precommit.ts
@@ -178,7 +178,8 @@ async function checkReferencesForDelete(resource: WithId<Resource>): Promise<voi
   const checkForCriticalRefs = new SelectQuery('ProjectMembership_References')
     .column('resourceId')
     .where('targetId', '=', resource.id)
-    .where('code', 'IN', criticalProjectMembershipReferences);
+    .where('code', 'IN', criticalProjectMembershipReferences)
+    .limit(1);
 
   const results = await checkForCriticalRefs.execute(db);
   if (results.length) {


packages/server/src/fhir/precommit.ts
```diff
@@ -35,9 +39,13 @@ export function isPreCommitSubscription(subscription: WithId<Subscription>): boo
 export async function preCommitValidation<T extends Resource>(
   author: Reference,
   project: WithId<Project> | undefined,
-  resource: T,
+  resource: WithId<T>,
   interaction: BackgroundJobInteraction
 ): Promise<T | boolean | undefined> {
+  if (interaction === 'delete' && !project?.superAdmin) {
```
mattlong: The description of `project` is `The project to which the resource belongs. Must be passed separately because this is called before the resource meta is assigned.` whereas I think the intent here is to allow super admins to bypass this check, right? Should the signature of `preCommitValidation` be changed/amended? Maybe something like:

```typescript
export async function preCommitValidation<T extends Resource>(
  repo: Repository,
  resource: T,
  interaction: BackgroundJobInteraction
): Promise<T | boolean | undefined> {
```

It seems like that would tighten up the abstraction and would let this line be `if (interaction === 'delete' && !repo.isSuperAdmin()) {`

mattwiller: Great suggestion, I think passing the whole `Repository` is much cleaner and didn't require much in the way of additional changes

packages/server/src/fhir/precommit.ts
```diff
@@ -124,3 +138,26 @@ export async function preCommitValidation<T extends Resource>(
...
 }
+
+/**
+ * Ensures that critical references are not left dangling when a resource is deleted.
+ * Specifically, resources referenced by a ProjectMembership should not be deleted until all memberships
+ * that refer to them are deleted.
+ * @param resource - The resource to be deleted.
+ * @throws {OperationOutcomeError} When the resource cannot be deleted because of a critical reference.
+ */
+async function checkReferencesForDelete(resource: WithId<Resource>): Promise<void> {
```
codyebberson, line 176: We could be more surgical and only check `ProjectMembership.user` and `ProjectMembership.profile` types (i.e., `User`, `Practitioner`, `Patient`, etc).

That would save us a bunch of unnecessary queries when deleting `Communication`, `Observation`, and `Task` resources, which are the more common case for deletes.

mattwiller: I opted for all resource types because any resource could be referenced by `ProjectMembership.access.parameter.valueReference`, but I suppose that would generally just result in lost access instead of some kind of harder-to-track down issue with auth.  I'm cool with special-casing the resource types @codyebberson

codyebberson: Ah, that's an interesting call out.

I think our main objective here is to prevent people from getting into completely broken states, which is mostly about stopping people from deleting their `Practitioner` or `AccessPolicy`.

If it's ok with you, let's take a lighter touch on this first pass.

packages/server/src/fhir/precommit.ts
```diff
@@ -124,3 +144,48 @@ export async function preCommitValidation<T extends Resource>(
 
   return undefined;
 }
+
+const criticalProjectMembershipReferences = ['profile', 'user', 'access-policy', 'accessPolicy'];
```
codyebberson, line 148:
```suggestion
const projectMembershipReferenceTypes = ['AccessPolicy', 'Bot', 'ClientApplication', 'Patient', 'Practitioner', 'RelatedPerson', 'User'];
```

packages/server/src/fhir/precommit.ts
```diff
@@ -26,18 +33,32 @@ export function isPreCommitSubscription(subscription: WithId<Subscription>): boo
...
+  repo: Repository,
+  resource: WithId<T>,
   interaction: BackgroundJobInteraction
 ): Promise<T | boolean | undefined> {
+  const logger = getLogger();
+
+  if (
+    interaction === 'delete' &&
+    !repo.isSuperAdmin() &&
+    getCriticalReferenceTargets().includes(resource.resourceType)
```
codyebberson, line 51:
```suggestion
    projectMembershipReferenceTypes.includes(resource.resourceType)
```

packages/server/src/fhir/precommit.ts
mattwiller: I had gone with listing the reference fields that we care about, since we could e.g. add new profile types like `Device` and I would want those to automatically be reflected here

packages/server/src/fhir/precommit.ts
```diff
@@ -124,3 +144,48 @@ export async function preCommitValidation<T extends Resource>(
...
+ * that refer to them are deleted.
+ * @param resource - The resource to be deleted.
+ * @throws {OperationOutcomeError} When the resource cannot be deleted because of a critical reference.
+ */
+async function checkReferencesForDelete(resource: WithId<Resource>): Promise<void> {
+  const db = getDatabasePool(DatabaseMode.WRITER);
+  const checkForCriticalRefs = new SelectQuery('ProjectMembership_References')
+    .column('resourceId')
+    .where('targetId', '=', resource.id)
+    .where('code', 'IN', criticalProjectMembershipReferences);
```
mattlong: Might as well add a `.limit(1)` if that's all we're looking at.

mattlong (approved):

PR #7920, opened 2025-11-26
wip(agent): auto-failover mode

ThatOneBro:


packages/agent/src/app.ts
```diff
@@ -1118,4 +1161,157 @@ export class App {
...
+   * If primary is disconnected or status is unknown, promotes this agent to primary.
+   */
+  private startStatusPolling(): void {
+    if (this.statusPollTimer) {
+      return;
+    }
+    this.log.info('Starting status polling for secondary agent');
+    this.statusPollTimer = setInterval(async () => {
+      const statusInfo = await this.checkAgentStatus();
+      if (!statusInfo || statusInfo.status !== 'connected') {
```
ThatOneBro: This seems wrong

ThatOneBro:
Previous requirement of this feature not requiring any server-side is looking to be relaxed, as the urgency for this feature has gone down significantly. Looking for cleaner way to do this via coordination with the server in the future. Closing for now

PR #7923, opened 2025-11-26
fix(vmcontext): always pass new `require`

ThatOneBro:
Fixes issue where `vmcontext` bots were getting the error `Dynamic require of "@medplum/core" is not supported`


ianplunkett:
Tested against some previously failing bots.  This change works for me.

codyebberson (approved):
Nice, thanks @ThatOneBro ! 🙏 🚀

PR #7928, opened 2025-11-29
Fixes #7789 - handle C-CDA with single participant

### LINKED ISSUE #7789: C-CDA to FHIR converter fails to process documents with single <participant> element
`"@medplum/ccda": "4.5.1"` (though this also affects the latest 5.0.0 release)

When attempting to use `convertCcdaToFhir` with XML that contains a single `<participant>`, the conversion fails due to:

``` 
 [cause]: TypeError: e is not iterable
      at sn.createRelatedPersons (/app/node_modules/@medplum/ccda/dist/cjs/index.cjs:2:1312)
      at sn.processHeader (/app/node_modules/@medplum/ccda/dist/cjs/index.cjs:2:469)
      at sn.convert (/app/node_modules/@medplum/ccda/dist/cjs/index.cjs:2:176)
      at ua (/app/node_modules/@medplum/ccda/dist/cjs/index.cjs:2:69)
```

The createRelatedPersons() method in ccda-to-fhir.ts assumes this.ccda.participant is always an array, but XML parsers return a single object when there's only one `<participant>` element in the C-CDA document. This causes a TypeError: e is not iterable error.

I believe the fix is to add `'ClinicalDocument.participant'` to the `ARRAY_PATHS` in
https://github.com/medplum/medplum/blob/main/packages/ccda/src/xml.ts#L7

`/packages/ccda/testdata` fixtures all contain 0 or 2+ participants, so the case of (1) participant is missed and will fail to convert.  

I'd be happy to provide a PR... it's a 1 line addition to `xml.ts` to support `ClinicalDocument.participant`, but I wasn't if we wanted to update `roundtrip.test.ts` with a new fixture that has a single participant.

codyebberson:


codyebberson:
@claude - please review

ThatOneBro (approved):

PR #7933, opened 2025-11-30
CDS Hooks bots

codyebberson:
This Pull Request introduces the necessary API endpoints and infrastructure to enable Medplum Bots to function as compliant **HL7 Clinical Decision Support (CDS) Services**, adhering to the CDS Hooks specification.

This feature allows external Electronic Health Records (EHRs) and other CDS Clients to discover and invoke user-defined Medplum Bot logic in real-time during clinical workflows (e.g., `patient-view`, `order-sign`).

The strategy here is to take a very light touch, and delegate most of the work to Medplum Bots.

### Key Changes and Architectural Decisions

1.  **Dedicated CDS Hooks Router:** Introduces a new express router at `/cds-services` separate from the FHIR API (`/fhir/R4`).
2.  **Auth Requirement:** All CDS Hooks endpoints (`/cds-services` and `/cds-services/{id}`) are secured via **JWT authentication**. This ensures all service discovery and invocation is scoped to an authenticated user's project/tenant.
3.  **Discovery Endpoint (`GET /cds-services`):**
    * Dynamically generates the service list by searching for `Bot` resources where the new custom property `Bot.cdsService` is defined.
    * Leverages the new **`cds-hook` SearchParameter** for efficient filtering of Bots.
    * Returns the entire service list in a single, non-paginated response, per specification.
4.  **Service Invocation Endpoint (`POST /cds-services/{id}`):**
    * Directly maps the URL parameter `{id}` to a `Bot` resource ID.
    * Executes the target Bot, passing the entire CDS Hooks request body (including `hook`, `context`, and `fhirAuthorization`) as the Bot's input.
    * The Bot's output must be a valid CDS Hooks response (a JSON object with the `cards` array).
5.  **Data Model Contract:** The CDS service configuration is defined within a new singleton property on the Bot resource: **`Bot.cdsService`**. This avoids overloading FHIR extensions or the `OperationDefinition` resource, ensuring a clean, type-safe contract for our platform.

More details are available in the accompanying `README.md`

See https://cds-hooks.hl7.org/

### DIFF
--- a/packages/react/src/BackboneElementInput/BackboneElementInput.module.css
+++ b/packages/react/src/BackboneElementInput/BackboneElementInput.module.css
@@ -1,0 +1,5 @@
+.nested {
+  padding-left: var(--mantine-spacing-md);
+  padding-top: var(--mantine-spacing-md);
+  padding-bottom: var(--mantine-spacing-sm);
+}


packages/react/src/BackboneElementInput/BackboneElementInput.tsx
```diff
@@ -60,6 +62,15 @@ export function BackboneElementInput(props: BackboneElementInputProps): JSX.Elem
       defaultValue={defaultValue}
       onChange={props.onChange}
       outcome={props.outcome}
+      stackProps={
+        isNested
```
codyebberson: This creates a little indentation for nested elements in the `<ResourceForm>`

packages/server/src/database.test.ts
```diff
@@ -49,6 +50,9 @@ describe('Database config', () => {
           if (sql === 'SELECT pg_try_advisory_lock($1)') {
             result.rows = [{ pg_try_advisory_lock: advisoryLockResponse } as unknown as R];
           }
+          if (sql === mockQueries.GetVersionSql) {
+            result.rows = [{ version: preDeployVersion } as unknown as R];
```
codyebberson, line 54: @mattlong  - can you please double check me on this?

I believe this is the first pre-deploy migration that uses `idempotentCreateIndex`, which has more complex logic than we were handling with this mock postgres client.

By adding this check, we essentially trick the setup code into thinking that all pre-deploy migrations are already run.

codyebberson:
@claude - please review

codyebberson:
@claude please review

packages/server/src/cds/routes.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // The CDS Hooks spec does not define pagination for discovery
+  // Most servers will have under 10 CDS services, so we set a high max count.
+  const maxCount = 100;
+
+  const bots = await repo.searchResources<Bot>({
+    resourceType: 'Bot',
+    filters: [{ code: 'cds-hook', operator: Operator.MISSING, value: 'false' }],
+    count: maxCount,
+  });
```
noahsilas, line 30: If someone does add > 100 CDS bots, it seems like this will find some subset of the bots available to list. Should we fail loudly in that scenario to prevent confusion about missing entries? I'm imagining something roughly like:

```ts
if (bots.length === maxCount) {
  throw new Error('CDS bot limit exceeded`);
}
```

packages/fhirtypes/dist/Bot.d.ts
```diff
@@ -185,6 +185,12 @@ export interface Bot {
    */
   executableCode?: Attachment;
 
+  /**
+   * CDS service definition if the bot is used as a CDS Hooks service. See
+   * https://cds-hooks.hl7.org/ for more details.
+   */
+  cdsService?: BotCdsService;
```
rahul1, line 192: we should consider making this an array. I can see use cases where developers would like to use the same bot code to service multiple hook types

packages/server/src/cds/routes.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  res.status(200).json(response);
+});
+
+// Calling a CDS Service: https://cds-hooks.hl7.org/#calling-a-cds-service
+cdsRouter.post('/:id', async (req: Request, res: Response) => {
+  const { id } = req.params as { id: string };
+
+  const ctx = getAuthenticatedContext();
+  ctx.repo.setExtendedMode(true); // Need extended mode for meta.project
```
noahsilas: Looking around a bit, it appears that other places where we are performing `executeBot` we fetch the resource by using the system repo, which has extendedMode set.

Does it make sense to follow that same pattern here? Or to adjust other places to match this behavior of using a scoped repo with the extended mode flag set? (Not a blocker, just trying to understand the direction the code is moving in so I can match it going forward)

Are there things like middleware that might be using this context repo that we don't want to also receive extended mode behavior? Should we do something like `ctx.repo.clone()` before mutating to isolate this change?

packages/react/src/ElementsInput/ElementsInput.tsx
```diff
@@ -18,6 +19,7 @@ export interface ElementsInputProps extends BaseInputProps {
   readonly defaultValue: any;
   readonly onChange: ((value: any) => void) | undefined;
   readonly testId?: string;
+  readonly stackProps?: StackProps;
```
noahsilas: I think this would let callers introduce a confusing conflict if they specify `data-testid` in both props and `stackProps`. We could explicitly disallow that one:

```suggestion
  readonly stackProps?: Omit<StackProps, 'data-testid'>;
```

We could do the same thing with `style`, or we could merge any style prop passed into the styles that we are trying to add (`flexGrow: 1`). I'd probably lean towards disallowing passing `style`.

packages/fhirtypes/dist/Bot.d.ts
codyebberson: I looked into that.  Hooks are required to have unique IDs, so using the Bot ID as the hook ID would break.  Claude and Gemini both recommended making it a singleton, so I went with that.

packages/react/src/BackboneElementInput/BackboneElementInput.tsx
noahsilas: Another option that might be a little less awkward than threading Stack props down could be to wrap this element to give it spacing at this level we are in control of. Something along these lines maybe?

```tsx
  <Box {...(isNested ? { pl: 'md', pt: 'md', pb: 'sm' } : {})>
    <ElementsInput ... />
  </Box>
```

packages/server/src/cds/routes.ts
codyebberson: Yeah....  The "read as user, and then read again as system" is a bit of a hack.

When purely acting in FHIR-mode, we need to respect the user's wishes about whether to use extended mode or not.

I felt like when we're acting in CDS-mode, it would be valid to essentially say "we're going to force extended mode here".

There may be other places where that is also valid, so, yes, agreed that it is perhaps something for investigation.

packages/server/src/cds/routes.ts
codyebberson: Yeah, there isn't a great answer here.  It's a weird oversight in the spec to not allow pagination.

I need to noodle on throwing an error.  I see the merit in failing loudly, but it also feels risky.

We could just allow for sorting and pagination using the FHIR query params (i.e., `_sort`, `_offset`, etc)

packages/server/src/database.test.ts
mattlong: Generally, index creation should be in a post-deploy migration. Is there a reason this one needed to be in a pre-deploy? I think that would avoid needing to add this?

packages/react/src/BackboneElementInput/BackboneElementInput.tsx
codyebberson: ✅  I switched to the `<Box>` strategy.

packages/server/src/cds/routes.ts
codyebberson: My preference for now is to move forward with a v1, even if imperfect, to unblock things.  We can revisit if/when this becomes an issue.

packages/server/src/cds/routes.ts
codyebberson: I reverted the `setExtendedMode()` change and went back to double-reading.  We can revisit potential optimizations in the future.

packages/react/src/ElementsInput/ElementsInput.tsx
codyebberson: I reverted the `stackProps` idea, and went with your `<Box>` wrapper suggestion ✅

packages/server/src/database.test.ts
mattlong: This probably points to us needing to further improve `migrate.ts` to spit out proper post-deploy migrations instead of having to manually transform a pre-deploy migration into a post-deploy migration 😓

PR #7934, opened 2025-12-01
Dependency upgrades 2025-12-01

medplumbot:
Dependency upgrades

### DIFF
--- a/packages/core/src/eventtarget.ts
+++ b/packages/core/src/eventtarget.ts
@@ -67,7 +67,6 @@ export class EventTarget {
 export class TypedEventTarget<TEvents extends Record<string, Event>> {
   private readonly emitter = new EventTarget();
 
-  // This unions with string
   dispatchEvent<TEventType extends keyof TEvents>(event: TEvents[TEventType]): void {
     this.emitter.dispatchEvent(event);
   }


packages/react/.storybook/preview.tsx
```diff
@@ -45,17 +63,21 @@ function ColorSchemeWrapper({ children }: { children: React.ReactNode }) {
...
     </BrowserRouter>
   ),
-  (Story) => (
+  (Story: any) => (
     <ColorSchemeWrapper>
       <Story />
     </ColorSchemeWrapper>
   ),
+  (Story: any) => (
+    <MantineProvider theme={medplumDefaultTheme}>
```
ThatOneBro, line 79: Need to add Mantine provider since we are no longer using `storybook-addon-mantine`. It was the last thing keeping us from upgrading to Storybook 10, and seemed more like a nice to have rather than a requirement for our Storybook

packages/react/.storybook/main.ts
```diff
@@ -19,16 +17,12 @@ const config: StorybookConfig = {
   async viteFinal(inputConfig, { configType }) {
     let config = inputConfig;
 
-    if (configType === 'PRODUCTION') {
```
ThatOneBro, line 22: No longer need this since turbosnap is included by default now, apparently since Storybook 8:

Image

packages/react/.storybook/main.ts
```diff
@@ -19,16 +17,12 @@ const config: StorybookConfig = {
...
-        plugins: [turbosnap({ rootDir: config.root ?? process.cwd() })],
-      });
-    } else if (configType === 'DEVELOPMENT') {
+    if (configType === 'DEVELOPMENT') {
       config = mergeConfig(config, {
         resolve: {
           alias: {
-            '@medplum/core': path.resolve(__dirname, '../../core/src'),
-            '@medplum/react-hooks': path.resolve(__dirname, '../../react-hooks/src'),
+            '@medplum/core': path.resolve(import.meta.dirname, '../../core/src'),
```
ThatOneBro, line 20: Storybook is ESM only now

packages/react/package.json
```diff
@@ -69,41 +69,40 @@
...
     "react-router": "7.9.6",
     "rfc6902": "5.1.2",
-    "signature_pad": "5.1.1",
+    "signature_pad": "5.1.2",
     "sinon": "21.0.0",
-    "storybook": "9.1.16",
-    "storybook-addon-mantine": "6.0.0",
+    "storybook": "10.1.2",
     "typescript": "5.9.3",
     "vite-plugin-turbosnap": "1.0.3"
```
noahsilas: I think you have removed the only consumer of this dependency, so I think we can drop this line now too 😁

ThatOneBro: Good catch

PR #7950, opened 2025-12-03
Provider - use Communication url for messages

techdavidy:
- In URLs use Communication rather than Message
- Handle empty inbox.

Screenshot 2025-12-03 at 11 18 31 AM


techdavidy:
@claude please review

codyebberson (approved):

PR #7953, opened 2025-12-04
feat: Patient $merge operation MVP

rahul1:
This PR adds full FHIR R5 compliance to the Patient $merge operation, implementing all missing features from the specification.

https://fhir.hl7.org/fhir/patient-operation-merge.html

**Note:** Pagination and Async support to come in future PRs

### 1. Output Format 
- Updated to match FHIR spec with `input`, `outcome`, and `result` parameters
- Maintains backward compatibility with `return` parameter
- True echo of original input parameters (including identifiers, preview, result-patient)

### 2. Identifier-Based Patient Resolution 
- Added `source-patient-identifier` parameter (0..*)
- Added `target-patient-identifier` parameter (0..*)
- Validates exactly one patient match (rejects 0 or multiple matches)
- Supports both system|value and value-only identifier searches

### 3. Custom Merge with result-patient Parameter 
- Added `result-patient` parameter for custom merge specifications
- Validates result-patient ID matches target-patient
- Validates required 'replaces' link is present
- Allows precise control over merged patient structure

### 4. Preview Mode 
- Added `preview` parameter (boolean)
- Returns merge result without committing changes
- Useful for user review before committing merges
- Does not create Provenance resources in preview mode

### 5. Provenance Generation 
- Automatically creates Provenance resource after each merge
- Tracks who performed the merge, when it occurred, and which patient versions were involved
- Uses standard FHIR codings (activity: merge, reason: PATADMIN)
- Links to authenticated user who performed the merge

### 6. Machine-Readable OperationOutcome 
- Added extensions to OperationOutcome for machine-readable data:
  - `resourcesUpdated` count as `valueInteger`
  - `preview` flag as `valueBoolean`
- Maintains human-readable text messages

### DIFF
--- a/packages/core/src/patientmerge.test.ts
+++ b/packages/core/src/patientmerge.test.ts
@@ -1,16 +1,16 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 // Ported from examples/medplum-demo-bots/src/deduplication/merge-matching-patients.test.ts
-import type { WithId } from '@medplum/core';
-import { createReference } from '@medplum/core';
-import type { Patient, ServiceRequest } from '@medplum/fhirtypes';
+import type { Identifier, Patient, ServiceRequest } from '@medplum/fhirtypes';
 import {
   linkPatientRecords,
   mergePatientRecords,
   patientsAlreadyMerged,
   replaceReferences,
   unlinkPatientRecords,
-} from '@medplum/core';
+} from './patientmerge';
+import type { WithId } from './utils';
+import { createReference, getIdentifier } from './utils';
 
 describe('Patient Merge Utils', () => {
   test('should link two patient records correctly', () => {
@@ -201,9 +201,12 @@ describe('Patient Merge Utils', () => {
     const result = mergePatientRecords(srcPatient, targetPatient);
 
     expect(result.target.identifier).toHaveLength(3);
-    expect(result.target.identifier?.find((id) => id.system === 'http://foo.org')?.use).toBe('old');
-    expect(result.target.identifier?.find((id) => id.system === 'http://bar.org')?.use).toBe('old');
-    expect(result.target.identifier?.find((id) => id.system === 'http://baz.org')?.use).toBe('official');
+    expect(getIdentifier(result.target, 'http://foo.org')).toBe('123');
+    expect(getIdentifier(result.target, 'http://bar.org')).toBe('456');
+    expect(getIdentifier(result.target, 'http://baz.org')).toBe('789');
+    expect(result.target.identifier?.find((id: Identifier) => id.system === 'http://foo.org')?.use).toBe('old');
+    expect(result.target.identifier?.find((id: Identifier) => id.system === 'http://bar.org')?.use).toBe('old');
+    expect(result.target.identifier?.find((id: Identifier) => id.system === 'http://baz.org')?.use).toBe('official');
   });
 
   test('should handle patients without identifiers in source', () => {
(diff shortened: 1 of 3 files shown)


rahul1:
@claude

packages/server/src/fhir/operations/patientmerge.ts
```diff
@@ -0,0 +1,241 @@
...
+  // Merge identifiers and contact info
+  const mergedPatients = mergePatientRecords(linkedPatients.src, linkedPatients.target);
+
+  // Rewrite all clinical resource references from source to target
+  // Note: This PR only processes the first page of results (no pagination)
+  const resourcesUpdated = await rewriteClinicalReferences(repo, mergedPatients.src, mergedPatients.target);
+
+  // Update both patients
+  await repo.updateResource(mergedPatients.src);
+  await repo.updateResource(mergedPatients.target);
```
noahsilas: This feels like a good candidate for being wrapped in a transaction.

packages/server/src/fhir/operations/utils/patientmergeutils.ts
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Recursively searches for all references to the source resource and translates them to the target resource.
+ * This function mutates the input object in place.
+ *
+ * @param obj - A FHIR resource or element (will be mutated)
+ * @param srcReference - The reference string referring to the source resource (e.g., "Patient/123")
+ * @param targetReference - The reference string referring to the target resource (e.g., "Patient/456")
+ */
+export function replaceReferences(obj: any, srcReference: string, targetReference: string): void {
```
noahsilas: `any` basically entirely opts this function out of the type checker. It would be nice to be more specific with something like `Resource | Element`, although I think that there are some tricky bits below due to Typescript's choices around how to treat the soundness of "`for (key in obj)`" style iteration. But doing a more specific cast in a couple places might be a good way to improve safety overall, as well as flag at compile time if we are ever calling this with an unexpected type.

As an entirely alternative strategy, we could try to build the paths that resolve to the identifier that need to be updated, which is a non-mutating operation that should be nicely type-safe. We could feed those into a `patchResource` call which would be more semantically descriptive of the operation we want to achieve.

packages/server/src/fhir/operations/utils/patientmergeutils.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (typeof obj[key] === 'object' && obj[key] !== null) {
+      // Check if this is a Reference object
+      if (isReference(obj[key])) {
+        // If the reference matches the source, replace it with the target
+        if (obj[key].reference === srcReference) {
+          obj[key].reference = targetReference;
+        }
+      } else {
+        // Recursively process nested objects
+        replaceReferences(obj[key], srcReference, targetReference);
```
noahsilas: I had an initial thought that this wouldn't work for Arrays, but in practice I think it actually does? Nice use of "`for (key in obj)`" style iteration to to handle Objects and Arrays :tada:

packages/server/src/fhir/operations/patientmerge.ts
```diff
@@ -51,28 +64,135 @@ const operation: OperationDefinition = {
...
+  if (reference) {
+    return reference;
+  }
+
+  // If identifiers provided, resolve them
+  if (identifiers) {
+    const idArray = Array.isArray(identifiers) ? identifiers : [identifiers];
+
+    // Build search query for identifiers
+    const searchParams = new URLSearchParams();
```
noahsilas, line 163: I'm a bit surprised to see URLSearchParams used in this way. It looks like you are using it like a hash that can hold many values for the same key? This might be more cleanly expressed as something like a `[string, string][]`. Or maybe building up the Filter entries directly as you iterate?
```ts
const filters: Filter[] = identifiers
  .filter((identifier) => Boolean(identifier.value))
  .map((identifier) => ({
    code: 'identifier
    operator: 'eq' as const,
    value: identifier.system ? `${identifier.system}|${identifier.value}` : identifier.value
  })
const bundle = await repo.search({ resourceType: 'Patient', filters });
```

PR #7954, opened 2025-12-04
CDS Hooks client

codyebberson:
This PR adds utility methods to make it easier to call CSD Hooks/Services.

* Added TypeScript type definitions for all of the CDS types
* Added `getCdsServices()` to `MedplumClient`, which will return a list of available CDS services/hooks
* Added `buildCdsRequest()` helper utility, which can be used to assemble a CDS request, including parsing and populating prefetch requirements
* Added `callCdsService()` to `MedplumClient` to actually call the CDS service
* I intentionally kept `buildCdsRequest()` separate from `callCdsService()` for a few reasons:
  * The spec does not require clients to fill in all prefetch fields, so there may be some judgement involved here
  * In many cases, the FHIR server can be separate from the CDS hooks server, so you may want to call these separately

Example Medplum bot that demonstrates how to call a CDS Hook with prefetch:

```ts
import { BotEvent, MedplumClient, buildCdsRequest } from '@medplum/core';

export async function handler(medplum: MedplumClient, event: BotEvent): Promise<any> {  
  const services = (await medplum.getCdsServices()).services[0];
  const context = { patientId: '123' };
  const request = await buildCdsRequest(medplum, event.requester, service, context);
  const response = await medplum.callCdsService(service.id, request);
  return response;
}
```

### DIFF
--- a/packages/core/src/cds.ts
+++ b/packages/core/src/cds.ts
@@ -203,7 +203,7 @@ function evaluatePrefetch(
     return Promise.resolve(null);
   }
 
-  const referenceMatch = query.match(/^(\w+)\/([0-9a-zA-Z-_]+)$/);
+  const referenceMatch = /^(\w+)\/([0-9a-zA-Z-_]+)$/.exec(query);
   if (referenceMatch) {
     // Read request
     const [resourceType, id] = referenceMatch.slice(1);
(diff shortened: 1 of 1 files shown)


codyebberson:
@claude please review

PR #7955, opened 2025-12-04
Fixed AuditEvent validation warning on bots and client apps

codyebberson:


packages/definitions/build.mjs
```diff
@@ -37,7 +37,6 @@ esbuild
   })
   .then(() => {
     writeFileSync('./dist/cjs/package.json', '{"type": "commonjs"}');
-    copyDataFiles();
```
codyebberson, line 40: I noticed we're double-copying all of the FHIR R4 files from `src` to `dist`.  We only need to do that once.

PR #7956, opened 2025-12-04
Fixed CodeQL regex warnings

codyebberson:
CodeQL was complaining about these regexes.  I don't think it's actually a big deal, but making the changes to keep the tools happy.


codyebberson:
@claude please review

PR #7957, opened 2025-12-04
Fix client parsing of RateLimit header

mattwiller:
Fixes `MedplumClient` to use the [correct format](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-10.html#name-ratelimit-field) when parsing `RateLimit` headers from the server

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -3763,8 +3763,8 @@ describe('Client', () => {
     expect(result).toBeDefined();
     expect(client.rateLimitStatus()).toStrictEqual(
       expect.arrayContaining([
-        { name: 'requests', remainingUnits: 15, secondsUntilReset: 59 },
-        { name: 'fhirInteractions', remainingUnits: 255, secondsUntilReset: 59 },
+        { name: 'requests', remainingUnits: 15, secondsUntilReset: 59, resetsAfter: expect.any(Number) },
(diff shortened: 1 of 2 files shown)


packages/core/src/client.test.ts
```diff
@@ -3769,6 +3746,32 @@ describe('Client', () => {
...
+    });
+
+    const client = new MedplumClient({ fetch });
+    expect(client.rateLimitStatus()).toStrictEqual([]);
+    const result = await client.readResource('Patient', '123');
+    expect(result).toBeDefined();
+    expect(client.rateLimitStatus()).toStrictEqual(
+      expect.arrayContaining([
+        { name: 'requests', remainingUnits: 59539, secondsUntilReset: 4 },
+        { name: 'fhirInteractions', remainingUnits: 0, secondsUntilReset: 3 },
```
noahsilas: Tangentially related improvement idea: Having "secondsUntilReset" is helpful if you know you just received a rate limit response, but becomes a bit challenging if you are keeping a client in memory longer than a few seconds. What do you think about including a timestamp? Something like `resetsAt: Date.now() + secondsUntilReset * 1000` would let consumers more easily check if the reset has happened.

mattwiller: Good idea — let me add that

PR #7960, opened 2025-12-04
Provider - graphql single last message query per topic

techdavidy:
Fix issue that graphql would return all messages instead of the last message per topic/thread.

### DIFF
--- a/examples/medplum-provider/src/hooks/useThreadInbox.ts
+++ b/examples/medplum-provider/src/hooks/useThreadInbox.ts
@@ -39,23 +39,22 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
 
   useEffect(() => {
     const fetchAllCommunications = async (): Promise<void> => {
-
       const searchParams = new URLSearchParams(query);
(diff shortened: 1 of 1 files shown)


techdavidy:
@claude please review

examples/medplum-provider/src/hooks/useThreadInbox.ts
```diff
@@ -42,58 +43,70 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
...
-      const allCommunications = await medplum.graphql(`
-        query {
-          CommunicationList(
+      if (parents.length === 0) {
+        setThreadMessages([]);
+        return;
+      }
+
+      const queryParts = parents.map((parent) => {
+        const safeId = parent.id?.replace(/-/g, '') || '';
```
codyebberson, line 54: out of curiosity, why was this required?

codyebberson (approved):

examples/medplum-provider/src/hooks/useThreadInbox.ts
techdavidy: graphql syntax error

PR #7965, opened 2025-12-04
More efficiently update reference lookup tables

mattlong:
Should result in less churn on reference tables (and fewer auto vacuums and auto analyzes) during reindexing and updating resources. We could use a similar approach for the HumanName and Address tables, but decided to keep it focused on references for now.

Locally, seeing about a **10% speedup** when reindexing Observations.

Previous strategy:
1. Delete all existing rows for resources being indexed
2. Insert all desired rows for said resources

New strategy:
1. Get existing rows for resources being indexed
2. Calculate which resources have a diff between existing and desired rows
3. Delete all existing rows for resources with diffs
4. Insert desired rows for resources with diffs

Includes optimizations for resources being created or with no existing reference table rows.

### DIFF
--- a/packages/server/src/fhir/lookups/reference.ts
+++ b/packages/server/src/fhir/lookups/reference.ts
@@ -29,7 +29,7 @@ export interface ReferenceTableRow extends LookupTableRow {
  * Each reference is represented as a separate row in the "<ResourceType>_References" table.
  */
 export class ReferenceTable extends LookupTable {
-  static allColumnNames = ['resourceId', 'targetId', 'code'] as const;
+  static readonly allColumnNames = ['resourceId', 'targetId', 'code'] as const;
 
   getTableName(resourceType: ResourceType): string {
     return resourceType + '_References';
@@ -48,16 +48,11 @@ export class ReferenceTable extends LookupTable {
     return false;
   }
 
-  /**
-   * Indexes the resource in the lookup table.
-   * @param client - The database client.
-   * @param resources - The resources to index.
-   * @param create - True if the resource should be created (vs updated).
-   */
   async batchIndexResources<T extends Resource>(
-    client: PoolClient,
+    client: PoolClient | Pool,
     resources: WithId<T>[],
-    create: boolean
+    create: boolean,
+    resourceBatchSize: number = 200
   ): Promise<void> {
     if (resources.length === 0) {
       return;
@@ -68,7 +63,7 @@ export class ReferenceTable extends LookupTable {
     const existingRows = create ? undefined : await this.getExistingRows(client, resources);
     if (existingRows === undefined || existingRows.length === 0) {
       const newRows: ReferenceTableRow[] = [];
-      await this.extractAllValues(newRows, resources);
+      await this.extractAllValues(newRows, resources, resourceBatchSize);
 
       // nothing to delete since no existing rows
 
@@ -89,7 +84,7 @@ export class ReferenceTable extends LookupTable {
     }
 
     const newRowsByResource = new Map<string, ReferenceTableRow[]>();
-    await this.extractAllValues(newRowsByResource, resources);
+    await this.extractAllValues(newRowsByResource, resources, resourceBatchSize);
 
     const resourcesWithChangedReferences: Resource[] = [];
     const rowsToInsert: LookupTableRow[] = [];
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/lookups/reference.ts
```diff
@@ -75,6 +224,15 @@ export class ReferenceTable extends LookupTable {
   }
 }
 
+/**
+ * Creates a hash string for a reference row for efficient comparison.
+ * @param row - The reference table row.
+ * @returns A hash string combining resourceId, targetId, and code.
+ */
+function hashRow(row: ReferenceTableRow): string {
+  return `${row.resourceId}|${row.targetId}|${row.code}`;
```
mattwiller: Could this refer to `allColumnNames`, to ensure that if new columns are added they automatically become part of the hash?

packages/server/src/fhir/lookups/reference.ts
```diff
@@ -65,17 +65,83 @@ export class ReferenceTable extends LookupTable {
...
+    const rowsToInsert: LookupTableRow[] = [];
+    for (const resource of resources) {
+      const existingHashes = existingHashesByResource.get(resource.id) ?? new Set<string>();
+      const newRowsForResource = newRowsByResource.get(resource.id) ?? [];
+      const newHashes = new Set(newRowsForResource.map(hashRow));
+
+      const identical = existingHashes.size === newHashes.size && [...existingHashes].every((h) => newHashes.has(h));
+      if (!identical) {
+        resourcesToDelete.push(resource);
+        rowsToInsert.push(...newRowsForResource);
```
noahsilas: Double checking my understanding here: if there is any change in the lookup rows, `resourcesToDelete` gets used to flag that _all_ existing lookup rows related to `resource` should be removed first, and then all the entries we generated should be added. We expect this to happen inside a transaction, so there shouldn't be an observable state for any other readers where some row is missing in between those steps.

Is that correct? If so I think the only feedback I would offer is that `resourcesToDelete` is maybe a slightly misleading name, as the resource being pushed into it won't be deleted itself. 

One thing I briefly considered was if it would be worthwhile to try to do a more fine-grained destroy+create. I thought we could potentially have the DB do less work by having the app compute which rows specifically need to be destroyed and created. I think the bookkeeping is likely to add a lot more code overhead for little practical gain (especially since the rows don't have an `id` primary key but use the whole row as a composite key).

Nice :tada:

packages/server/src/fhir/lookups/reference.ts
```diff
@@ -65,17 +65,83 @@ export class ReferenceTable extends LookupTable {
...
+    await this.extractAllValues(newRowsByResource, resources);
+
+    const resourcesToDelete: Resource[] = [];
+    const rowsToInsert: LookupTableRow[] = [];
+    for (const resource of resources) {
+      const existingHashes = existingHashesByResource.get(resource.id) ?? new Set<string>();
+      const newRowsForResource = newRowsByResource.get(resource.id) ?? [];
+      const newHashes = new Set(newRowsForResource.map(hashRow));
+
+      const identical = existingHashes.size === newHashes.size && [...existingHashes].every((h) => newHashes.has(h));
```
noahsilas: I think that this is unimportant at the scale involved here, but if you replace `[...existingHashes]` with `existingHashes.values()` you save an array allocation. Iterator#every is supported in Node v22+, so this should be allowed now that we're on Medplum 5.x.

Like I said, not important for this case, just pointing it out for consideration in future work 😁

packages/server/src/fhir/lookups/reference.ts
```diff
@@ -45,17 +48,163 @@ export class ReferenceTable extends LookupTable {
...
+          });
+          throw err;
+        }
+
+        // Yield to event loop between batches
+        if (i + resourceBatchSize < resources.length) {
+          await new Promise<void>((resolve) => {
+            setImmediate(() => resolve());
+          });
+        }
```
noahsilas: This looks to me like it is inside the nested `j` loop and yielding for every record instead of each batch - should this be hoisted up one level into the `i` loop?

mattlong:
@claude please review

packages/server/src/fhir/lookups/reference.ts
mattlong: Yes, doing only the exact rows needing to be deleted would be nice, but constructing the DELETE queries to do so would be kind of gross since these tables don't have a PK, but maybe it's worth it? We could add a PK column to facilitate this pattern, but perhaps we leave that as a potential v2. That would align with my general preference for every table having a PK (preferably an auto-incrementing one) to facilitate maintenance operations like this.

It would have to be either a lot of individual DELETE which is probably a non starter considering perf:
```sql
DELETE FROM "Patient_References" WHERE "resourceId" = $1 AND code = $2 AND "targetId" = $3;
```

Or bigger composite ones like:
```sql
DELETE FROM "Patient_References" WHERE (
("resourceId" = $1 AND code = $2 AND "targetId" = $3) OR
("resourceId" = $4 AND code = $5 AND "targetId" = $6) OR
("resourceId" = $7 AND code = $8 AND "targetId" = $9) OR
...
)
```

mattlong:
@claude take a look at the changes since your previous review.

PR #7966, opened 2025-12-04
Add database column for Coding display language

mattwiller:


codyebberson (approved):
Lgtm 👍 

I assume that `TEXT` with 2-5 characters is comparable in perf to a 32-bit integer?

PR #7899, opened 2025-11-24
Add MedplumBaseSubscription FHIR profile with FSH tooling

rahul1:
## Summary

This PR adds a new FHIR profile `MedplumBaseSubscription` that simplifies the usage of Medplum subscription extensions by bundling all common extensions into a single profile.

## Changes

- **Added FSH definitions** for `MedplumBaseSubscription` profile and related extensions
- **Refactored build process** to properly generate FSH profiles with snapshots and merge them into `profiles-medplum.json`
- **Updated documentation** to inform users about the new profile and its availability in the Medplum App
- **Removed unused script** (`prepare-fsh.mjs`) that was no longer needed
- **Standardized extension URLs** to use `medplum.com` (no www) for consistency
- **Added comprehensive test coverage** for build-profiles script (69% coverage, 11 tests)
- **Improved build error handling** to fail immediately on FSH build errors

## Profile Usage

The `MedplumBaseSubscription` profile can be selected when creating a new Subscription in the Medplum App. Users can find it in the "Profiles" tab of the Subscription resource page, as shown in the screenshot below:

MedplumBaseSubscription Profile in App

The profile includes definitions for all common Medplum subscription extensions:
- Max Attempts
- Supported Interaction  
- Secret
- Success Codes
- FHIRPath Criteria Expression

## Technical Details

- FSH profiles are built using SUSHI with the `--snapshot` flag to generate complete snapshots
- Build process ensures profiles are merged correctly without overwriting existing profiles
- All generated StructureDefinitions include snapshots only (no differentials)
- Extension URLs standardized to `https://medplum.com/fhir/StructureDefinition/...` (no www)
- Server code maintains backwards compatibility with old www URLs

## Code Review Feedback Addressed

✅ **URL Consistency**: Standardized all extension URLs to use `medplum.com` (no www)  
✅ **FSH Configuration**: Clarified comment about snapshot generation  
✅ **Test Coverage**: Added comprehensive test suite for build-profiles script  
✅ **Build Resilience**: Improved error handling to fail immediately in CI/CD  
✅ **Error Context**: Enhanced error messages with specific guidance

## Deployment Note

⚠️ **Important**: After this PR is deployed, it will require a super admin to rebuild structureDefinitions to make the new profile available in the system.

## Testing

- ✅ Build process generates all 23 profiles correctly (17 original + 6 FSH-generated)
- ✅ All profiles include snapshots only
- ✅ Lint and tests pass (13 tests total, 69% coverage for build-profiles)
[...]

Issue #7686, opened 2025-10-27
$find Operation for Appointment Scheduling


rahul1:
## Overview

Implement the FHIR `$find` operation for searching available appointment slots, based on the [[IHE ITI.Scheduling specification](https://build.fhir.org/ig/IHE/ITI.Scheduling/OperationDefinition-appointment-find.html)](https://build.fhir.org/ig/IHE/ITI.Scheduling/OperationDefinition-appointment-find.html) with Medplum-specific enhancements.

## Endpoints

The operation is available on three resource types, returning different response formats:

- **`[base]/Appointment/$find`** - Returns `Bundle` of proposed `Appointment` resources

All three share the same input parameters and processing logic, differing only in response format.

## MVP Scope

**Version 1 includes:**
- `start`, `end` - date range (required)
- `schedule` - explicit Schedule reference(s) (required for MVP)
- `service-type` - filter by service type
- `patient-reference` - patient context
- Support for multi-schedule intersection queries
- Buffer time calculations
- Booking limit enforcement
- Alignment interval/offset handling

**Out of scope for Version 1** (future enhancements):
- Schedule inference from `practitioner`, `organization`, `location-reference`, `location-string`
- `specialty` filtering
- `reason` codes
- `service-request` references  
- `coverage` information
- `preferred-times` optimization
- `patient-resource` for new patients

For MVP, clients must explicitly provide `schedule` parameter(s). Schedule discovery from other parameters will be added in follow-up PRs.

### Required Parameters

**`start`** (dateTime, 1..1) - *IHE Standard*
- Start of availability search period
- Must be in ISO 8601 format with timezone

**`end`** (dateTime, 1..1) - *IHE Standard*
- End of availability search period
- Must be in ISO 8601 format with timezone

**`schedule`** (string, 1..*) - *Medplum Extension*
- Direct Schedule resource reference(s)
- Format: `Schedule/{id}`
- Multiple schedules calculate availability intersection (all must be available)
- Example: `schedule=Schedule/dr-smith` or `schedule=Schedule/surgeon&schedule=Schedule/or-3`

### Optional Parameters

**`service-type`** (string, 0..*) - *IHE Standard*
- Service type codes in format `[system]|[code]`
- Used to select which scheduling-parameters extension applies
- Multiple values treated as OR (any match)

**`patient-reference`** (string, 0..1) - *IHE Standard*
- Patient reference for context
- Format: `Patient/{id}`

### 1. Schedule Validation

[...]

PR #7967, opened 2025-12-04
`$find` FHIR operation for a single Schedule resource

noahsilas:
To begin the Scheduling line of work, this introduces a `$find` handler that operates on a single schedule. 

- Starting with a handler for a single schedule lets us test all the fundamental features in a simpler environment than one where we handle schedule intersections. #7686 requests that we expose this for individual Schedule resources as well, so this is a useful chunk of work even as-is.

- I am temporarily punting on embedding the FHIR definition of the extension. You can find the one I have been developing against in both FSH and JSON forms in this gist: https://gist.github.com/noahsilas/06d6b2ae9191c2ba876b8a605e72de86. Moving forward on FSH versus JSON is a larger question than we will address in this PR. (See #7899, #8111 for some possible strategies). This operation will still work without the definition posted anywhere, as Medplum server does not currently validate against an extension schema. For this reason I've tried to be fairly defensive in how we parse the extension data from a resource.

- I added `temporal-polyfill` to the server bundle. This is used to transform availability rules like "Monday at 10am" into `Date` objects representing the correct real time in a given timezone. This was chosen over libraries like `date-fns`, as it is where the ecosystem is trending towards (Temporal is a stage 3 proposal and is available now in V8). The intention here is to eventually drop this dependency when all our supported node versions include it natively. This is used fairly conservatively (Temporal usage is only in `resolveAvailability` and `eachDayOfInterval`), so if we find problems with this approach we can adjust it without too much intervention.

- If we can't load all the Slot resources that could affect availability, we bail out entirely and ask the viewer to search a narrower range. Finding all relevant slots is a bit ugly - I'm interested in following up down the line to consider how we can improve the query interface, but that is out of scope for this PR.

- I've added a limitation that the search range can be at most 31 days. I've also added an optimization to stop finding slots after a page has been filled, so that someone requesting availability across such a large window won't waste so much time and memory processing results that won't be returned. In practice, I suspect folks are more likely to be constrained by the need to load all existing Slot resources affecting the window, and so might be forced to use a narrower search interval than this anyways.

https://github.com/medplum/medplum/issues/7686

### DIFF
--- a/packages/server/src/fhir/operations/utils/find.ts
+++ b/packages/server/src/fhir/operations/utils/find.ts
@@ -47,10 +47,8 @@ function hasMatchingServiceType(slot: Slot, inputCoding: Coding | undefined): bo
 
   // Check if there any of the Slot's service type codes match the input code
   for (const codeableConcept of serviceType) {
-    for (const coding of codeableConcept.coding ?? []) {
-      if (coding.system === inputCoding.system && coding.code === inputCoding.code) {
-        return true;
-      }
+    if (codeableConcept.coding?.some((c) => c.system === inputCoding.system && c.code === inputCoding.code)) {
+      return true;
     }
   }
   return false;


examples/medplum-fsh-profiles/src/profiles/input/fsh/scheduling-parameters.fsh
```diff
@@ -0,0 +1,62 @@
...
+    bufferAfter 0..1 and
+    alignmentInterval 0..1 and
+    alignmentOffset 0..1 and
+    duration 1..1
+* extension ^slicing.discriminator.type = #value
+* extension ^slicing.discriminator.path = "url"
+
+* extension[availability].value[x] only Timing
+* extension[availability].valueTiming.repeat 1..1
+* extension[availability].valueTiming.event 0..0
```
noahsilas: These constraints let us define our extension as a subset of the more general field types. In this case, our "availability" field is a `Timing` with the cardinality of `repeat` constrained to exactly one, and `event` constrained to none.

I don't think that this will be strictly enforced in our data store yet, so I probably should handle parsing of these pretty defensively.

I do like how the FSH style syntax here lets us avoid a bunch of the gnarly/verbose JSON blobs, but I could use some support from a team member on making this fit into our broader definitions structure.

packages/server/src/fhir/operations/find.ts
```diff
@@ -0,0 +1,169 @@
...
+        },
+
+        {
+          code: '_filter',
+          operator: Operator.EQUALS,
+          // Slot starts sometime in interval, OR
+          // Slot ends sometime in interval, OR
+          // Slot time fully contains interval
+          value: `((start ge "${start}" and start le "${end}") or (end ge "${start}" and end le "${end}") or (start lt "${start}" and end gt "${end}"))`,
+        },
```
noahsilas, line 132: This feels sort of clunky; is there a better way to express this query than building this `_filter` expression?

packages/server/src/fhir/operations/find.ts
```diff
@@ -0,0 +1,169 @@
...
+      ],
+    }),
+  ]);
+
+  // If we filled a full search page of slots, then there may be slots we
+  // didn't fetch that would impact availability. Fail loudly here.
+  if (slots.length === DEFAULT_MAX_SEARCH_COUNT) {
+    throw new OperationOutcomeError(badRequest('Too many slots found in range; try searching with smaller bounds'));
+  }
```
noahsilas, line 147: This also is a little awkward. I sort of want a helper like `searchResourcesExhaustive()` that could directly throw if there are rows beyond the search limit? Or maybe there's another query method to use that would let us guarantee that we loaded all relevant records?

packages/server/src/fhir/operations/find.ts
```diff
@@ -111,7 +111,15 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
     throw new OperationOutcomeError(badRequest('Too many slots found in range; try searching with smaller bounds'));
   }
 
-  const timezone = getTimezone(schedule);
+  if (schedule.actor.length !== 1) {
+    throw new OperationOutcomeError(badRequest('$find only supported on schedules with exactly one actor'));
+  }
+
+  const [actor] = await ctx.repo.readReferences(schedule.actor);
```
noahsilas: It would be nice if we could have loaded this concurrently with the other resources we need, but it is tricky due to the polymorphic nature of schedule.actor - this could be a Practitioner, Location, or Device resource. Waiting to get the fetched Schedule to read its references seems like the best option.

It does seem common that the Schedule fetch will be faster than the Slot query, so we potentially could speed this up with some clever chaining:

```ts
  const [[schedule, actor], slots] = await Promise.all([
    ctx.repo.readResource<Schedule>('Schedule', req.params.id).then(
    (schedule) => ctx.repo.readReferences(schedule.actor).then((actor) => [schedule, actor])
    ),
    ctx.repo.searchResources<Slot>(...),
  ]);
```

This feels way less readable though for likely minimal benefit, so for now I'm leaving it in this more direct waterfall style.

packages/server/src/fhir/operations/find.test.ts
```diff
@@ -405,4 +424,192 @@ describe('Schedule/:id/$find', () => {
...
+            end: formatNY(new Date('2025-12-04T13:50:00.000-05:00')),
+            status: 'free',
+            schedule: createReference(schedule),
+          },
+        },
+      ],
+    });
+  });
+
+  test('returns appropriate slots for each serviceType passed', async () => {
```
noahsilas, line 573: @rahul1 Does this seem like the appropriate behavior? The question that came up was what to do if multiple service-type options are passed that have differing availability requirements. My answer was to surface all the options as their parameter groups specify, but it means that your results can have different codes and durations (and alignment/buffer/...) amongst all the results. You also can have overlapping slots in the result set if they came from different scheduling parameter groups.

packages/server/src/fhir/operations/find.ts
noahsilas: This seems to be the best way to do this today. I'm starting to think about an interface sort of akin to a query builder that could generate this in a more type-safe way, but any improvements there are out of scope for this PR.

packages/server/src/fhir/operations/find.ts
```diff
@@ -0,0 +1,188 @@
...
+          // Slot starts sometime in range, OR
+          // Slot ends sometime in range, OR
+          // Slot time fully contains range
+          value: `((start ge "${start}" and start le "${end}") or (end ge "${start}" and end le "${end}") or (start lt "${start}" and end gt "${end}"))`,
+        },
+
+        {
+          code: 'status',
+          operator: Operator.EQUALS,
+          value: 'busy,free,busy-unavailable',
```
mattlong: Maybe this is more a question for the spec, but it's okay to ignore `busy-tentative | entered-in-error`? In particular `busy-tentative` seems like it could be problematic to ignore?

packages/server/src/fhir/operations/find.ts
mattlong: This seems reasonable to me; at least for now. We've erred on the side of not baking in specialized helper functions or options when possible.

If we did want to add this, instead of a separate method, probably add options to `searchResources`:

```typescript
type SearchOptions = {
  exhaustive?: boolean
}

async searchResources<T extends Resource>(searchRequest: SearchRequest<T>, options?: SearchOptions): Promise<WithId<T>[]>
```

packages/server/src/fhir/operations/find.ts
```diff
@@ -0,0 +1,188 @@
...
+  if (slots.length === DEFAULT_MAX_SEARCH_COUNT) {
+    throw new OperationOutcomeError(badRequest('Too many slots found in range; try searching with smaller bounds'));
+  }
+
+  if (schedule.actor.length !== 1) {
+    throw new OperationOutcomeError(badRequest('$find only supported on schedules with exactly one actor'));
+  }
+
+  const [actor] = await ctx.repo.readReferences(schedule.actor);
+  if (actor instanceof Error) {
```
mattlong: You could get implicit not-found handling with `readReference` instead of `readReferences`, e.g. `ctx.repo.readReference(schedule.actor[0]);` with tradeoff of a generic "not-found" error being returned instead of this custom message. Perhaps the custom error message is preferable though.

packages/server/src/fhir/operations/find.ts
```diff
@@ -0,0 +1,188 @@
...
+    throw new OperationOutcomeError(badRequest('Loading actor for schedule failed'), { cause: actor });
+  }
+  const timeZone = getTimeZone(actor);
+  if (!timeZone) {
+    throw new OperationOutcomeError(badRequest('No timezone specified'));
```
mattlong: You should confirm that this expression format is accurate...

```suggestion
    throw new OperationOutcomeError(badRequest('No timezone specified', `Schedule.actor[0].extension.where(url = '${TimezoneExtensionURI}')`));
```

packages/server/src/fhir/operations/utils/find.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  // If we didn't get a specific code to test for, we should only match wildcard slots,
+  // which we ruled out above.
+  if (!inputCoding) {
+    return false;
+  }
+
+  // Check if there any of the Slot's service type codes match the input code
+  for (const codeableConcept of serviceType) {
+    for (const coding of codeableConcept.coding ?? []) {
```
mattlong:
```suggestion
    if (codeableConcept.coding && codeableConcept.coding.any((c) => c.system === inputCoding.system && c.code === inputCoding.code)) { return true; }
```

packages/server/src/fhir/operations/utils/find.ts
```diff
@@ -0,0 +1,358 @@
...
+  if (intervals.length < 2) {
+    return intervals;
+  }
+  const sorted = intervals.concat().sort((a, b) => a.start.valueOf() - b.start.valueOf());
+  return sorted.reduce<Interval[]>((acc, interval) => {
+    const last = acc.at(-1);
+
+    // base case: push the first interval onto the accumulator.
+    if (!last) {
+      return [interval];
```
mattlong, line 140: Huh, I guess there's no harm in throwing away the original `acc` parameter

mattlong (approved):
A lot of interval math and operations going on here, but it all seemed reasonable to me. Just a handful of minor suggestions for your consideration.

mattlong:
@claude Please review and be extra diligent around the functions comparing and operating on intervals.

packages/server/src/fhir/operations/find.ts
noahsilas: The semantics of this are pretty unclear to me. I think that it probably should be up to the specific caller what to do here. If this is being used for patients self-service scheduling, then you are probably right (I would expect that type of use case to result in `busy-tentative` slots being generated). If this is more back of house, you might want to be able to schedule over something that is as-yet unconfirmed.

I think you're probably right though - for this first implementation the best thing to do in my mind is to stay overly conservative, and respecting "tentative" blocks on a calendar is probably the more conservative approach.

packages/server/src/fhir/operations/find.ts
noahsilas: Makes sense. I'll leave as-is for now, and reconsider as I gain more context in the ways folks are interacting with search. Thanks for considering!

packages/server/src/fhir/operations/utils/find.ts
noahsilas: Nice! I think I had started going down a more complex matching path here that I abandoned for initial implementation (in favor of just marking "system" and "code" as both required) and left behind this shell that you have expressed more cleanly. Thanks!

packages/server/src/fhir/operations/utils/find.ts
noahsilas: Ah, yeah, interesting point. You could say something like `acc.append(interval); return acc;`. In this base case the result is nearly identical. You potentially save an array allocation (although it's not fully obvious; I wouldn't be overly surprised if an empty array is specialized in V8 and replaced with a more full array object at the time the first element is inserted).

Semantically though, yeah, reduce just cares about the return value from each iteration. Appending to an existing array and returning it is just one way to continuously build a result.

packages/server/src/fhir/operations/find.ts
noahsilas: Nice, I didn't realize that this was how that worked. Thank you!

noahsilas:
> **1. `intersectIntervals` function logic error (find.ts:67-73)**
> 
> ```ts
> function intersectIntervals(left: Interval, right: Interval): Interval | undefined {
>   if (!areIntervalsOverlapping(left, right)) {
>     return undefined;
>   }
>   return { start: clamp(left.start, right), end: clamp(left.end, right) };
> }
> ```
> 
> **Issue**: This doesn't compute intersection correctly. `clamp(left.start, right)` clamps `left.start` to the `right` interval, but intersection should be:
> 
> * `start = Math.max(left.start, right.start)`
> * `end = Math.min(left.end, right.end)`
> 
> The current implementation could return incorrect intervals when the intervals have complex overlaps.

This method also looks like a valid means to compute the intersection. I think I wrote this in terms of `clamp` because it sort of matched my intuition about the problem: force the `left.start` and `left.end` both into the range of the `right` interval. This should work because we already bailed out if there is no overlap (otherwise this could have returned an intersection starting and ending at the endpoint in `right` closest to `left`).

I don't know what a "complex overlap" is, but I have not yet found any scenario where this does not return the correct intersection.

I think this comes down to a matter of preference. If other folks find this style easier to reason about I am happy to adjust!

> **2. Inconsistent interval boundary semantics**
> 
> * `areIntervalsOverlapping` uses `<` (exclusive) by default, `<=` (inclusive) with option
> * `mergeIntervals` uses `areIntervalsOverlapping(..., { inclusive: true })`
> * `pairWithOverlaps` uses mixed comparisons (`<=` for end checks, `<` for start checks)
> 
> This inconsistency could lead to gaps or double-counting in edge cases where intervals are adjacent.

`pairWithOverlaps` could be written more consistently by calling out to `areIntervalsOverlapping`, but I made an explicit choice to try to write that with speed in mind. It's made an opinionated choice that we only care about non-empty overlaps.

I'm going to end up reusing this helper in later work (intersecting availability from multiple schedules will look quite similar), so this will also continue seeing increasing test coverage as usage expands.

> ### 🔍 Potential Issues
> **3. `pairWithOverlaps` algorithm correctness (find.ts:167-194)** The algorithm looks correct but has a subtle complexity:
> 
> * Line 182: `if (listB[i].end <= a.end)` assumes we can skip this interval for future iterations
> * But if `listB[i].end > a.end`, we might still need it for the next `a`
> * The `nextStartIndex` tracking handles this, but it's easy to get wrong

This complexity seems necessary to me in order to do this in a single linear pass through each list. 

> **4. Modular arithmetic in `findAlignedSlotTimes` (find.ts:275-277, 303-304)**
> 
> ```ts
> function mod(n: number, d: number): number {
>   return ((n % d) + d) % d;
> }
> // ...
> const remainder = mod(firstMinuteStart.getMinutes() - options.offsetMinutes, options.alignment);
> ```
> 
> This looks correct for handling negative offsets, but there's no validation that `options.alignment` is positive and non-zero, which could cause division by zero.

Fair call out. This is coming from the parsed scheduling parameters, which should have coerced `0` to `60`, but it's a pretty cheap defense to add a helpful error message if we get a zero passed in.

> **5. DST transition edge case** The tests cover DST well, but there's a potential issue in `eachDayOfInterval` (find.ts:18-32):
> 
> * It uses `t.epochMilliseconds < interval.end.valueOf()`
> * During DST "fall back", the same wall clock time occurs twice
> * This could theoretically cause infinite loops or skip days if the interval spans the exact transition

This is testing the start of the day, which should be safe from being inside DST transition windows. We are using the Temporal API to advance whole days, which will automatically adjust for DST (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime/add#description). This is safe (until some TZ decides to get tricky around midnight DST transitions).

packages/server/src/fhir/operations/find.ts
noahsilas: In the long term we may want to support schedules with multiple actors. In that case though we probably will require that the timezone be well defined in some other way and we probably won't need to load this at all...

I think I'll follow your suggestion here. I think that the generic error message should be fine, maybe I can follow up by letting the caller pass the reference path in as an option to make it very clear in the error case what reference was causing the issue.

packages/server/src/fhir/operations/find.ts
noahsilas: Your expression looks correct, but it appears that you can also use the nicely tidy format of `Schedule.actor[0].extension('${TimezoneExtensionURI}')` 😁

PR #7869, opened 2025-11-19
New sidebar proof-of-concept

codyebberson:
Builds on https://github.com/medplum/medplum/pull/7264 from @kevinwadeshaw  and @techdavidy 

* Removes header
* Side nav now collapses to icon-only rather than hiding entirely
* Enables "Notification icons" (icons with red indicator bubbles) on the side nav

https://github.com/user-attachments/assets/bda5924e-3970-4a5f-9a6b-ff49bfa66b2b

PR #7986, opened 2025-12-06
Fixed navbar links open in new tab

codyebberson:
Regression introduced in https://github.com/medplum/medplum/pull/7869

Navbar links should be `<a href="...">` to enable right-click, middle-click, ctrl+click, etc.


packages/react/src/AppShell/AppShell.tsx
```diff
@@ -86,6 +86,7 @@ export function AppShell(props: AppShellProps): JSX.Element {
         opened={navbarOpen}
         spotlightEnabled={true}
         userMenuEnabled={true}
+        version={props.version}
```
codyebberson, line 89: Drive-by fix: the popup menu was also missing the Medplum version string.

techdavidy (approved):

Issue #7992, opened 2025-12-07
[Bug]: Docker Compose Full Stack Error when using `MEDPLUM_CONFIG_PATH` option


Drew-Daniels:
I noticed an issue when trying to use a `medplum.config.json` file alongside the full stack Docker Compose stack:

```bash
❯ MEDPLUM_CONFIG_PATH=./medplum.config.json docker compose up
[+] Running 3/3
 ✔ Container medplum-full-stack-redis-1           Running                                                                                                                                                   0.0s
 ✔ Container medplum-full-stack-postgres-1        Running                                                                                                                                                   0.0s
 ✔ Container medplum-full-stack-medplum-server-1  Recreated                                                                                                                                                 0.0s
WARN[0000] could not start menu, an error occurred while starting: Error while reading terminfo data:termbox: unsupported terminal
Attaching to medplum-app-1, medplum-server-1, postgres-1, redis-1
medplum-server-1  | Config file found, running with custom config
medplum-server-1  | {"level":"INFO","timestamp":"2025-12-07T20:28:57.625Z","msg":"Starting Medplum Server...","configName":"file:./medplum.config.json","version":"5.0.0"}
medplum-server-1  | Error: ENOENT: no such file or directory, open './medplum.config.json'
medplum-server-1  |     at readFileSync (node:fs:440:20)
medplum-server-1  |     at loadFileConfig (file:///usr/src/medplum/packages/server/dist/index.js:431:21)
medplum-server-1  |     at loadConfig (file:///usr/src/medplum/packages/server/dist/index.js:374:22)
medplum-server-1  |     at main (file:///usr/src/medplum/packages/server/dist/index.js:50656:24)
medplum-server-1  |     at file:///usr/src/medplum/packages/server/dist/index.js:50677:3
medplum-server-1  |     at ModuleJob.run (node:internal/modules/esm/module_job:377:25)
medplum-server-1  |     at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:691:26)
medplum-server-1  |     at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
medplum-server-1  |   errno: -2,
medplum-server-1  |   code: 'ENOENT',
medplum-server-1  |   syscall: 'open',
medplum-server-1  |   path: './medplum.config.json'
medplum-server-1  | }
dependency failed to start: container medplum-full-stack-medplum-server-1 is unhealthy
```

[...]

PR #7993, opened 2025-12-07
fix medplum.config.json path lookup in docker compose full stack

Drew-Daniels:
Fixes https://github.com/medplum/medplum/issues/7992


ThatOneBro:
Thanks @Drew-Daniels ! Good catch

ThatOneBro:
It looks like this PR is failing on the DCO check. Do you mind signing off on your commit so we can merge this?

You should be able to do that via the following git command:
```bash
git rebase HEAD~1 --signoff
```

Drew-Daniels:
Hey @ThatOneBro just updated! Thanks for taking a look

PR #7994, opened 2025-12-08
Dependency upgrades 2025-12-08

medplumbot:
Dependency upgrades


techdavidy:
Remove Tinybench dependency and packages/generator/src/benchmarks as it hasn't been used.
Upgrade to react 19.2.1

PR #7999, opened 2025-12-08
Fixes #7170 - Handle NULL values in token :not search

### LINKED ISSUE #7170: `:not` modifier should return missing values
## Current Behavior
`GET Patient?gender:not=male` only returns `Patients` that have a gender _present_ and whose gender does not equal `male`

## Expected Behavior
According to the [FHIR search spec](https://hl7.org/fhir/R4/search.html#token)
> Reverse the code matching described in the paragraph above: return all resources that do not have a matching item. Note that this includes resources that have no value for the parameter - e.g. ?gender:not=male includes all patients that do not have gender = male, **including patients that do not have a gender at all**

codyebberson:
### Problem

The FHIR `:not` search modifier was incorrectly excluding rows with NULL values. Per the FHIR specification, the `:not` modifier should match all resources that don't have the specified value, **including those where the field is missing/null**.

**Example**: `Patient?gender:not=male` should return:
- ✅ Patients with gender='female'  
- ✅ Patients with gender='other'
- ✅ Patients with gender=NULL *(but we were excluding these)*

### Root Cause
Our SQL generation was using `NOT(column = value)` which, due to SQL's three-valued logic, evaluates to NULL (not TRUE) when the column is NULL:
```sql
-- This excludes NULL values (incorrect for FHIR)
WHERE NOT("gender" = 'male')
```

### Solution
Special-case the `:not` modifier to use PostgreSQL's `IS DISTINCT FROM` operator:
```sql
-- This includes NULL values (correct for FHIR)
WHERE "gender" IS DISTINCT FROM 'male'
```

### Performance Impact

 `IS DISTINCT FROM` can use standard B-tree indexes efficiently, unlike the alternative `(column <> value OR column IS NULL)` pattern which requires OR logic that often prevents optimal index usage.

### DIFF
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -1441,6 +1441,27 @@ describe('project-scoped Repository', () => {
       expect(bundleContains(bundle1, serviceRequest2)).toBeDefined();
     }));
 
+  test('Token not equals matches missing value', () =>
+    withTestContext(async () => {
+      const family = randomUUID();
+
+      const patient1 = await repo.createResource<Patient>({
+        resourceType: 'Patient',
+        name: [{ family }],
+        gender: 'male',
+      });
+
+      const patient2 = await repo.createResource<Patient>({
+        resourceType: 'Patient',
+        name: [{ family }],
+      });
+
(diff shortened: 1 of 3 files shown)


codyebberson:
@claude please review

PR #8002, opened 2025-12-08
Provider - Intake form show unavailable valuesets

techdavidy:
Screenshot 2025-12-08 at 3 29 50 PM

### DIFF
--- a/examples/medplum-provider/src/pages/patient/IntakeFormPage.tsx
+++ b/examples/medplum-provider/src/pages/patient/IntakeFormPage.tsx
@@ -44,13 +44,12 @@ export function IntakeFormPage(): JSX.Element {
         })
       );
 
-  
       setUnavailableValueSets(unavailable);
     }
 
     checkValueSets()
-    .catch((error) => showErrorNotification(error))
-    .finally(() => setCheckingValueSets(false));
+      .catch((error) => showErrorNotification(error))
+      .finally(() => setCheckingValueSets(false));
   }, [medplum]);
 
   const handleOnSubmit = useCallback(


techdavidy:
@claude please review

techdavidy:
@claude please check again

examples/medplum-provider/src/pages/patient/IntakeFormPage.tsx
```diff
@@ -39,11 +77,79 @@ export function IntakeFormPage(): JSX.Element {
...
+ */
+async function checkValueSetAvailability(
+  valueSetUrl: string,
+  medplum: ReturnType<typeof useMedplum>
+): Promise<boolean> {
+  try {
+    await medplum.valueSetExpand({
+      url: valueSetUrl,
+      count: 1,
+    });
```
noahsilas, line 146: It looks like this is going to do a lot of network calls at the top level to show this error message, and then mount the children which will attempt to do these same operations with a larger `count` (https://github.com/medplum/medplum/blob/v5.0.7/packages/react/src/QuestionnaireForm/QuestionnaireFormItem.tsx#L398-L403).

I wonder if we can handle this at that nested level and either display the error closer to the problem (which might help if there is a specific question that is at issue) or bubble that error up to this component to show this same style of aggregate error message?

Another option would be to introduce a caching layer (maybe something like [@tanstack/react-query](https://github.com/tanstack/query) ?) and then we could emit the same queries from both layers and have that system de-duplicate. That's a much more architectural change to support this pattern though, and introduces some awkwardness in what aspects of the application query patterns live in `packages/react/*`  as a library vs. in `examples/medplum-provider` or `packages/app` as applications.

codyebberson: @noahsilas  - this is a good catch, however `MedplumClient` has some tricks up its sleeve.

1. `MedplumClient` has "auto batching", which will automatically detect batch-able requests, and combine them into a single batch request.
2. `MedplumClient` has intelligent caching of `GET` requests

So, with some slight tweaks, this code can be made to automatically do what you suggest.

The way auto batching works is by adding a small (i.e. 50ms) delay to batch-able requests. 

The current implementation calls each `ValueSet` separately, with an `await` behind each request.  If instead it used `Promise.all()` or `Promise.allSettled()`, then everything would be batched and cached nicely.

codyebberson: Oops!  You're already using `Promise.allsettled()`!

codyebberson (approved):

PR #8009, opened 2025-12-09
Handle NULL values in boolean :not search

nate-watkins:
Hey team- let me know your thoughts here. Deeper in the internals than I'm used to. This is similar to #7999 but for boolean fields.

It's my understanding that `Practitioner?active:not=false` should return resources where `active` is not defined in addition to `true`. The `repo.ts` change is so that these values values are stored as `NULL` in pg instead of `false`.

If I'm understanding correctly, this would also require reindexing `HealthCareService`, `Organization`, `OrganizationAffiliation`, `Patient`, `Practitioner`, `PractitionerRole`, `RelatedPerson`, and `Schedule`.

### DIFF
--- a/packages/server/src/fhir/search.ts
+++ b/packages/server/src/fhir/search.ts
@@ -1298,11 +1298,7 @@ function buildBooleanSearchFilter(
 
   const negated = filter.operator === Operator.NOT_EQUALS || filter.operator === Operator.NOT;
 
-  return new Condition(
-    new Column(table, impl.columnName),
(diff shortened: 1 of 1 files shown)


codyebberson:
Thanks @nate-watkins !

Agreed that this will require re-indexing your listed resource types...  We will explore adding a post-deploy migration to automate that.  We're having an internal debate about whether that's required, or if instead we can publish it as guidance.

Thanks again, we appreciate you 🙏

PR #8030, opened 2025-12-11
Include projectId in auto-download error messages

codyebberson:
Context:
1. "Auto download" is a Medplum feature that automatically fetches remote binary attachments by URL (such as documents, images, movies, etc), and slurps them into Medplum as FHIR `Binary`/`Media`/`DocumentReference` resources.
2. Enterprise customers have access to "log streaming", where we forward select log messages from our hosted environments to their customer managed logging solution.

This PR adds `projectId` to log messages when the auto downloader fails.  This is a prerequisite before we can forward messages to customer logging solutions.


codyebberson:
@claude please review

PR #8036, opened 2025-12-11
Explicitly use npm@11.7.0 in workflows

noahsilas:
NPM v11.6.2 has a bug when dealing with attributes in the dependency graph that was causing it to label more nodes as "peer" than it ought to. This is fixed in v11.6.3 via https://github.com/npm/cli/pull/8645.

For now, the latest NodeJS v24 tagged release is still bundling NPM v11.6.2; the problem will eventually go away as Node's `main` branch has already bumped the bundled `npm` to v11.7.0.

It would be nice if the `setup-node` action we depend on would let us specify an explicit package manager version to use, but that is still a pending feature request (https://github.com/actions/setup-node/issues/529).

In the meantime, we can keep our lockfile in good order by explicitly adding a step to our action that upgrades NPM to a patched version.

I've also rebuilt the lockfile by running `npm install --strict-peer-deps --prefer-dedupe` under npm v11.7.0, which removes the extra "peer" attributes that were introduced in commit e365fa405.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -24,6 +24,12 @@ jobs:
         with:
           node-version: '24'
           registry-url: 'https://registry.npmjs.org'
+      # npm 11.6.2 is currently bundled with Node v24 and has an irksome bug affecting
+      # `peer` attributes in the lockfile (https://github.com/npm/cli/issues/8690).
+      # Upgrade to a patched version; this should be removed after Node has a tagged
+      # release incorporating npm@11.6.3 or later.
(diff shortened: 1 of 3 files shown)


codyebberson (approved):

noahsilas:
I ran the `upgrade-dependencies` workflow on this branch to test this pre-merge. You can see the run reported here: https://github.com/medplum/medplum/actions/runs/20181631785/job/57943208224, and the resulting pull request with the later NPM version at #8060. It looks like it was able to run as expected without adding the `peer` attributes back in the unexpected places. :tada:

PR #8061, opened 2025-12-12
Include _total in search for Message Inbox and Task Board

techdavidy:


techdavidy:
@claude please review

PR #8071, opened 2025-12-15
Fixes #8069 - prefer crypto.randomUUID

### LINKED ISSUE #8069: [Bug] batch code seems to use non-random generateId function
While using the batch API, I noticed that the import chain seems to use the generateId method which is documented as not production ready.

Is this true? If so, is there anything in the pipe, other than the comment, to remedy this? If not, what is the recommended way to go about id generation using the batch API?

https://github.com/medplum/medplum/blob/main/packages/fhir-router/src/batch.ts

https://github.com/medplum/medplum/blob/main/packages/fhir-router/src/repo.ts

https://github.com/medplum/medplum/blob/main/packages/core/src/index.ts

https://github.com/medplum/medplum/blob/main/packages/core/src/crypto.ts

codyebberson:


codyebberson:
@claude please review

PR #8077, opened 2025-12-15
Include tests to resource pages in provider

techdavidy:


techdavidy:
@claude please review

examples/medplum-provider/src/pages/resource/ResourceCreatePage.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter, Routes, Route } from 'react-router';
+import * as reactRouter from 'react-router';
+import { describe, expect, test, vi, beforeEach } from 'vitest';
+import { ResourceCreatePage } from './ResourceCreatePage';
+
+describe('ResourceCreatePage', () => {
+  let medplum: MockClient;
+  let navigateSpy: ReturnType<typeof vi.fn>;
```
noahsilas: I think you can type this as `reactRouter.NavigateFunction` to remove the `any` typecast below.

examples/medplum-provider/src/pages/resource/ResourceCreatePage.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+    // Find the submit button (ResourceForm typically has a "Create" or "OK" button)
+    const submitButton = await screen.findByRole('button', { name: /create|ok|submit/i });
+    expect(submitButton).toBeInTheDocument();
+
+    // Mock createResource to return a result
+    const createdPractitioner = {
+      resourceType: 'Practitioner',
+      id: 'practitioner-123',
+    };
+    vi.spyOn(medplum, 'createResource').mockResolvedValue(createdPractitioner as any);
```
noahsilas, line 75: It would be awesome to remove the `any` typecasts in here. I think you can do this by assigning a specific Resource subtype when you create the object:

```
const createdPractitioner: WithId<Practitioner> = { ... }
```

This probably does create a bit of friction in some cases where more attributes are required to construct a valid object though... for instance below it looks like a Task resource requires attributes `status` and `intent` that you would have to define.

I would say this isn't a blocker, as this is still a huge improvement over the current lack of tests!

examples/medplum-provider/src/pages/resource/ResourceCreatePage.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+describe('ResourceCreatePage', () => {
+  let medplum: MockClient;
+  let navigateSpy: ReturnType<typeof vi.fn>;
+
+  beforeEach(async () => {
+    medplum = new MockClient();
+    vi.clearAllMocks();
+    navigateSpy = vi.fn();
+    vi.spyOn(reactRouter, 'useNavigate').mockReturnValue(navigateSpy as any);
```
noahsilas, line 22: I spent a few minutes trying to figure out a good way to remove this `any` typecast, and found no great results. I think that there might be an inherent tension between the type that this hook emits (which is a function with some specific args) and the Mocked function type that adds some extra spy machinery in...

I would have thought that a mock would be assignable to any function that doesn't specify colliding attributes but that doesn't seem to be the case, and my quick dig into what is happening didn't turn up an obvious answer.

For reference, the type error looks something like this:
```
  1 Argument of type 'MockedFunction<NavigateFunction>' is not assignable to parameter of type 'NavigateFunction'.
  2   Types of parameters 'delta' and 'to' are incompatible.
  3     Type 'To' is not assignable to type 'number'.
  4       Type 'string' is not assignable to type 'number'.
```

I might spend a little more time trying to figure out a good pattern for type safety here later, this seems like something we will want to repeat throughout a lot of tests.

examples/medplum-provider/src/pages/resource/ResourceHistoryPage.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+      resourceType: 'Task',
+      status: 'in-progress',
+      intent: 'order',
+    });
+
+    vi.spyOn(medplum, 'readHistory').mockResolvedValue({
+      resourceType: 'Bundle',
+      type: 'history',
+      entry: [],
+    } as any);
```
noahsilas, line 100: I think you can directly drop most of these `as any` type casts :grin:

examples/medplum-provider/src/pages/resource/ResourceCreatePage.test.tsx
techdavidy: Yeah the idea is to bring tests up to speed and then take a look at some of the issues that we may have.

PR #8085, opened 2025-12-16
Handle display language overrides in ValueSet/$expand and CodeSystem/$validate-code

mattwiller:


packages/server/src/fhir/operations/expand.test.ts
```diff
@@ -1288,24 +1288,66 @@ describe('Expand', () => {
...
+    expect(vsRes.status).toStrictEqual(201);
 
-    const superAdminToken = await initTestAuth({ superAdmin: true });
     const res = await request(app)
-      .get(`/fhir/R4/ValueSet/$expand?url=${url}&filter=clien`)
-      .set('Authorization', 'Bearer ' + superAdminToken);
-    expect(res.status).toBe(200);
-    expect(res.body.expansion.contains[0].display).toStrictEqual('ClientApplication');
+      .get(
+        `/fhir/R4/ValueSet/$expand?url=${encodeURIComponent(valueSet.url as string)}&filter=invalid&displayLanguage=fr`
```
noahsilas: Non blocking: you have this type cast because ValueSet.url is typed as `string | undefined`, but since you are defining an explicit object above with this property set you can do a little better here! Rewriting the definition from a type declaration to a "satisfies" constraint will retain the narrower type and let you use it without a cast, while still validating that the object is assignable to the ValueSet type.

```
const valueSet = {
  url: 'https://example.com/ValueSet/' + randomUUID(),
  ...
} satisfies ValueSet;

encodeURIComponent(valueSet.url)

```

packages/server/src/fhir/operations/codesystemvalidatecode.ts
```diff
@@ -95,7 +98,12 @@ export async function validateCodings(
 
   let result: any[] | undefined;
   if (codesToQuery.size > 0) {
-    const query = selectCoding(codeSystem.id, ...codesToQuery).where('synonymOf', '=', null);
+    const query = selectCoding(codeSystem.id, ...codesToQuery);
+    if (options?.displayLanguage) {
+      query.where('language', '=', options.displayLanguage);
+    } else {
+      query.where('synonymOf', '=', null);
```
noahsilas, line 106: This `synonymOf` filter used to always be applied, but now it is omitted when a displayLanguage is chosen. Is the idea here that these multi-language scenarios will have the localized versions added as "synonyms" of canonical codes? And in that case, we do want to return these synonyms because the caller explicitly asked for them, right?

packages/server/src/fhir/operations/expand.ts
```diff
@@ -96,20 +96,17 @@ export function filterIncludedConcepts(
   system?: string
 ): ValueSetExpansionContains[] {
   const filter = params.filter?.trim().toLowerCase();
-  const codings: Coding[] = flattenConcepts(concepts, { filter, system });
-  if (!filter) {
-    return codings;
-  }
-  return codings.filter((c) => c.display?.toLowerCase().includes(filter));
```
noahsilas, line 105: This filter is safe to drop because it's already being applied inside `flattenConcepts()` above. Nice clean up!

noahsilas (commented):
Small question about a query filter (`synonymOf`) that I suspect makes sense but I don't have the full context to assert confidently. Overall code looks good, with some nice tidy ups!

packages/server/src/fhir/operations/codesystemvalidatecode.ts
mattwiller: Yep, that was the reasoning here!

mattlong (approved):

Issue #8083, opened 2025-12-16
Provider - refresh task list when modifying a task.


techdavidy:
When filtering based on a status, e.g. requested, the view is not refreshed automatically when a task is done.

techdavidy:
solved in https://github.com/medplum/medplum/pull/8090

PR #8090, opened 2025-12-17
Provider - Task update triggers list refresh

techdavidy:
Solves https://github.com/medplum/medplum/issues/8083

### DIFF
--- a/examples/medplum-provider/src/components/tasks/TaskBoard.tsx
+++ b/examples/medplum-provider/src/components/tasks/TaskBoard.tsx
@@ -44,11 +44,12 @@ interface TaskBoardProps {
   query: string;
   selectedTaskId: string | undefined;
   onDeleteTask: (task: Task) => void;
+  onTaskChange: (task: Task) => void;
   onSelectedItem: (task: Task) => string;
 }
 
 export function TaskBoard(props: TaskBoardProps): JSX.Element {
-  const { query, selectedTaskId, onDeleteTask, onSelectedItem } = props;
+  const { query, selectedTaskId, onDeleteTask, onTaskChange, onSelectedItem } = props;
   const medplum = useMedplum();
   const profile = useMedplumProfile();
   const navigate = useNavigate();
@@ -136,7 +137,7 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
 
   const handleNewTaskCreated = (task: Task): void => {
     fetchTasks().catch(showErrorNotification);
-    handleTaskChange(task).catch(showErrorNotification);
+    onTaskChange(task);
   };
 
   const handleTaskChange = async (_task: Task): Promise<void> => {
@@ -274,12 +275,10 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
 
 function EmptyTasksState(): JSX.Element {
   return (
-    <Flex direction="column" h="100%" justify="center" align="center">
-      <Stack align="center" gap="md" pt="xl">
-        <Text c="dimmed" fw={500}>
-          No tasks available.
-        </Text>
-      </Stack>
+    <Flex direction="column" h="100%" justify="center" align="center" pt="xl">
+      <Text c="dimmed" fw={500}>
+        No tasks available.
+      </Text>
     </Flex>
   );
 }
--- a/examples/medplum-provider/src/pages/patient/TasksTab.tsx
+++ b/examples/medplum-provider/src/pages/patient/TasksTab.tsx
@@ -15,6 +15,10 @@ export function TasksTab(): JSX.Element {
     return `/Patient/${patientId}/Task/${task.id}`;
   };
 
+  const onTaskChange = (task: Task): void => {
+    navigate(onSelectedItem(task))?.catch(console.error);
+  };
+
   const handleDeleteTask = (_: Task): void => {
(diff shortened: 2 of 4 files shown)


techdavidy:
@claude can you please take a look

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -279,9 +276,8 @@ function EmptyTasksState(): JSX.Element {
   return (
     <Flex direction="column" h="100%" justify="center" align="center">
       <Stack align="center" gap="md" pt="xl">
```
noahsilas: This stack now contains exactly one item and can probably be dropped as well.

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -140,14 +139,12 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
...
-    setTasks(tasks.map((t) => (t.id === task.id ? task : t)));
-    onTaskChange(task);
-    setSelectedTask(task);
+  const handleTaskChange = async (_task: Task): Promise<void> => {
+    await fetchTasks().catch(showErrorNotification);
   };
 
   const handleDeleteTask = async (task: Task): Promise<void> => {
-    setTasks(tasks.filter((t) => t.id !== task.id));
+    await fetchTasks().catch(showErrorNotification);
```
noahsilas, line 148: It looks like `showErrorNotification` is always used as an error handler for `fetchTasks`, does it make sense to move that into that handler?

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -140,14 +139,12 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
     handleTaskChange(task).catch(showErrorNotification);
   };
 
-  const handleTaskChange = async (task: Task): Promise<void> => {
-    setTasks(tasks.map((t) => (t.id === task.id ? task : t)));
-    onTaskChange(task);
```
noahsilas, line 145: `onTaskChange` looks like it used to trigger a navigation event to take you to the detail page of the updated task, but now we are refetching the list and staying on the current page. It seems like this might have been desirable behavior for a new task? I'm not too familiar with this part of the app, will folks usually want to navigate in to the new task object?

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -140,14 +139,12 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
     handleTaskChange(task).catch(showErrorNotification);
```
noahsilas: It looks like this is duplicative now and will kick off two `fetchTasks` calls. It seems like that function is opting out of the client cache, so I expect this to double up the network call. We should probably remove one of these!

examples/medplum-provider/src/pages/patient/TasksTab.tsx
```diff
@@ -15,10 +15,6 @@ export function TasksTab(): JSX.Element {
     return `/Patient/${patientId}/Task/${task.id}`;
   };
 
-  const handleTaskChange = (task: Task): void => {
-    navigate(onSelectedItem(task))?.catch(console.error);
```
noahsilas, line 19: Unrelated: I got fairly confused by the usage of "onSelectedItem(task)" returning what looks like an in-app URL rather than doing any kind of event handling. But it looks like this gets prop-drilled through a bunch of this component tree so coming up with a better naming scheme or URL helper tool seems like a larger project. Maybe we can revisit this later with an eye towards cleaning this up!

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
techdavidy: if a property if the resource is modified.
It remains in the same resource.

examples/medplum-provider/src/pages/patient/TasksTab.tsx
techdavidy: yeah we let page handle any uri related events rather than the components.

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -137,17 +137,15 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
 
   const handleNewTaskCreated = (task: Task): void => {
     fetchTasks().catch(showErrorNotification);
-    handleTaskChange(task).catch(showErrorNotification);
+    onTaskChange(task);
```
noahsilas, line 140: Now that this is only called from this create handler, it might prevent some confusion down the line if we rename this prop from `onTaskChange` to `onCreateTask`.

UPDATE: Just saw that this gets passed down to child components down below and so can still fire on other updates, so this is probably not actually a good idea. There may be opportunity to improve this overall down the line, but it feels out of scope in this PR.

noahsilas (approved):

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
techdavidy: planning to update that

PR #8095, opened 2025-12-18
feat: configure BullMQ retry attempts and backoff settings

deam65:
## Problem

The BatchQueue worker had hardcoded `attempts: 1`, causing async batch jobs (`Prefer: respond-async`) to permanently fail on transient errors like PostgreSQL serialization conflicts. During concurrent transaction bundle processing, this resulted in ~20-30% job loss even with in-transaction retry logic (`transactionAttempts`).

## Root Cause

When all in-transaction retries fail, the job itself fails permanently because BullMQ wasn't configured to retry at the job level. Other workers (SubscriptionQueue, CronQueue) already have proper retry configuration.

## Solution

Add configurable `defaultAttempts` and `defaultBackoff` settings to the `bullmq` server config, allowing operators to tune job-level retry behavior via config file or AWS Parameter Store.

## Changes

- **`config/types.ts`**: Added `defaultAttempts` and `defaultBackoff` to `MedplumBullmqConfig`
- **`config/utils.ts`**: Added defaults (`defaultAttempts: 3`, `defaultBackoff: { type: 'exponential', delay: 1000 }`) and parsing support
- **`workers/batch.ts`**: Updated BatchQueue to use configurable retry settings
- **`docs/self-hosting/server-config.md`**: Documented new parameters

## Configuration Example

```json
{
  "bullmq": {
    "concurrency": 15,
    "defaultAttempts": 5,
    "defaultBackoff": { "type": "exponential", "delay": 2000 },
    "removeOnComplete": { "count": 1000 },
    "removeOnFail": { "count": 2000 }
  }
}

### DIFF
--- a/packages/server/src/workers/batch.test.ts
+++ b/packages/server/src/workers/batch.test.ts
@@ -211,7 +211,10 @@ describe('Batch Worker', () => {
           {
             response: {
               status: '201',
-              outcome: { resourceType: 'OperationOutcome', issue: [{ severity: 'information', code: 'informational' }] },
+              outcome: {
+                resourceType: 'OperationOutcome',
+                issue: [{ severity: 'information', code: 'informational' }],
+              },
             },
           },
         ],
@@ -226,7 +229,10 @@ describe('Batch Worker', () => {
             {
               response: {
                 status: '201',
-                outcome: { resourceType: 'OperationOutcome', issue: [{ severity: 'information', code: 'informational' }] },
+                outcome: {
+                  resourceType: 'OperationOutcome',
+                  issue: [{ severity: 'information', code: 'informational' }],
+                },
               },
             },
           ],

### CODE REVIEW

**Comment by mattlong** [comment]
I suspect the reasoning is that a non-transaction batch Bundle that e.g. includes creating resources, retrying the bundle could result in duplicate resource creation without a paper trail.

I think that's what [this comment](https://github.com/medplum/medplum/blob/1f774139475f97e17c6cce39e592793693422e8c/packages/server/src/workers/batch.ts#L121-L123) is conveying:
>   // Intentionally swallow all errors thrown during or after execution of the batch request, since we do NOT want to
  // execute part or all of the batch more than once.
  // If this function does not throw an error, the job will be considered "successful" and not requeued

In other words, I believe not retrying on errors is by design. The client should see the failed result of processing the bundle and compute a new Bundle to submit based on the state of the resources involved. That being said, it looks like the partial results of processing a bundle aren't made available in the case of the outcome not being OK.

@mattwiller Curious to get your take on this since you're more familiar with bundle processing and [introduced this worker](https://github.com/medplum/medplum/pull/5308)? How are errors in batch processing meant to be handled by clients?

**Comment by deam65** [comment]
> In other words, I believe not retrying on errors is by design. The client should see the failed result of processing the bundle and compute a new Bundle to submit based on the state of the resources involved. That being said, it looks like the partial results of processing a bundle aren't made available in the case of the outcome not being OK.

@mattlong In my self-hosted Medplum instance, I tried submitting 10 patient bundles, but only 8 made it through - though the AsyncJob shows that all bundles are in `completed` status. Perhaps I tackled the wrong issue, and we should really be looking into the failed async bundle being logged with the incorrect status?

**Comment by codyebberson** [comment]
Hi @deam65  - (as discussed offline, posting here for visibility)

Could you please shared these values?

```ts
  /** Number of attempts for transactions that fail due to retry-able transaction errors */
  transactionAttempts?: number;

  /** Number of milliseconds to use as a base for exponential backoff in transaction retries */
  transactionExpBackoffBaseDelayMs?: number;
```

The defaults are:

```ts
const defaultTransactionAttempts = 2;
const defaultExpBackoffBaseDelayMs = 50;
```

We can try significantly higher values, such as `transactionAttempts` 10+, and `transactionExpBackoffBaseDelayMs` in 100-200.

deam65:
@codyebberson 

My initial tests were ran with `transactionAttempts = 10` and  `transactionExpBackoffBaseDelayMs = 200` - this resulted in bundles missing.

I ran it earlier today with `transactionAttempts = 20` and  `transactionExpBackoffBaseDelayMs = 500` and all bundles went through - this was for ingest loads containing 10, 100, and 1000 Patient bundles! I will try to run with a large number of bundles (10k - 1M) to check if this is still an issue at those levels.

deam65:
closing this as increasing number of `transactionAttempts` will help resolve the issue

PR #8098, opened 2025-12-18
Add email to jwt payload

maddyli:


packages/server/src/oauth/keys.ts
```diff
@@ -49,6 +49,13 @@ export interface MedplumAccessTokenClaims extends MedplumBaseClaims {
    * For example, "Patient/123" or "Practitioner/456".
    */
   profile: string;
+
+  /**
+   * User email address.
+   * Included when the 'email' scope is requested and the user is a User resource.
+   * This is used for integrations with external systems like Onyx.
+   */
+  email?: string;
```
maddyli, line 57: Open to feedback on whether this should be a scope, or just part of the default payload.

maddyli:
@claude please review

packages/server/src/oauth/utils.test.ts
```diff
@@ -430,6 +433,223 @@ describe('OAuth utils', () => {
...
+        const userEmail = `test-${randomUUID()}@example.com`;
+        const user = await systemRepo.createResource<User>({
+          resourceType: 'User',
+          email: userEmail,
+          firstName: 'Test',
+          lastName: 'User',
+        });
+
+        // Create a Patient profile
+        const patient = await systemRepo.createResource<Patient>({
```
mattlong: Seems like this should be made with a repo associated with `project` so that you don't have to specify `meta` at all. You can easily get that repo by modifying up above to `createTestProject({withRepo: true})` and pulling repo out of the result.

packages/server/src/oauth/utils.test.ts
```diff
@@ -430,6 +433,223 @@ describe('OAuth utils', () => {
     }
   });
 
+  describe('getAuthTokens with email scope', () => {
+    test('Access token includes email when email scope is requested for User', async () => {
+      await withTestContext(async () => {
+        const systemRepo = getSystemRepo();
+        const { project } = await createTestProject();
```
mattlong: It looks like all these tests could share the same test project instead of making a project per test? If so, could add something like this instead the `describe` block:

```typescript
let project: Project;
let repo: Repository
beforeAll(async () => {
  const result = await createTestProject({withRepo: true});
  project = result.project;
  repo = result.repo;
});
```

PR #8100, opened 2025-12-19
feat(dosespot-react): add DoseSpot patient pharmacy React hook

oleg-mp:
React hooks for pharmacy search with DoseSpot + ability to save pharmacy in DoseSpot as patient's favorite

### DIFF
--- a/packages/dosespot-react/src/useDoseSpotPatientPharmacy.test.tsx
+++ b/packages/dosespot-react/src/useDoseSpotPatientPharmacy.test.tsx
@@ -6,11 +6,7 @@ import { MedplumProvider } from '@medplum/react-hooks';
 import { act, render, renderHook, screen } from '@testing-library/react';
 import type { JSX } from 'react';
 import { vi } from 'vitest';
-import {
-  DOSESPOT_ADD_PATIENT_PHARMACY_BOT,
-  DOSESPOT_PHARMACY_ID_SYSTEM,
-  DOSESPOT_SEARCH_PHARMACY_BOT,
-} from './common';
+import { DOSESPOT_ADD_PATIENT_PHARMACY_BOT, DOSESPOT_PHARMACY_ID_SYSTEM, DOSESPOT_SEARCH_PHARMACY_BOT } from './common';
 import { useDoseSpotPatientPharmacy } from './useDoseSpotPatientPharmacy';
 
 function TestComponent(): JSX.Element {


packages/dosespot-react/src/useDoseSpotPatientPharmacy.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  const setAsPrimary = (primary: boolean): void => {
+    privateSetAsPrimary(primary);
+  };
+
+  const addFavoritePharmacy = useCallback(
+    async (patientId: string): Promise<AddPatientPharmacyResponse> => {
+      if (!selectedPharmacy) {
+        throw new Error('Must select a pharmacy before adding it as a favorite');
+      }
```
noahsilas, line 144: My intuition here is that it feels strange to make this stateful in this way. As a consumer of the hook, I think I would probably want to write code that looks something like this:

```jsx
function MyComponent({ patient }: MyProps) {
  const { loading, pharmacies, addFavoritePharmacy } = useDoseSpotPatientPharmacy({
    patient,
    query: { zip: "12345" }
  });

  if (loading) { return <Spinner />; }

  return (
    <ol>
      {pharmacies.map((pharmacy) => (
        <li>
          {pharmacy.name}
          <button onClick={() => addFavoritePharmacy(pharmacy)}>Add favorite</button>
        </li>
      ))}
    </ol>
  );
}
```

In an implementation like this the statefulness is pushed into the closures attached as event handlers for each object instead of encoded s directly. One nice thing about this is it makes it easier to handle someone trying to add multiple favorites simultaneously - each call can be independent and run to resolution without worrying about state.

codyebberson:
@claude please review

oleg-mp:
> I'm not sure I really understand the meaning behind "favorite" and "primary" - are those just states that specific pharmacies can be in for a given patient? It sort of sounds like there can be one "primary" and many "favorite" pharmacies? Do the search results reveal if the pharmacies they found are already in those states?

per DoseSpot API doc (3.12.6):
"`POST api/patients/{patientId}/pharmacies` adds a pharmacy to a patient’s preferred pharmacies list. One preferred pharmacy per patient record can be
designated as “primary.” This should be used to indicate a patient’s top preference and will be automatically
linked to new prescriptions for that patient if another pharmacyId is not explicitly specified."

packages/dosespot-react/src/useDoseSpotPatientPharmacy.ts
oleg-mp: A patient in Dosespot can have a list of preferred pharmacies with only one of them at a time designated as 'primary'.
A pharmacy search is separate api call from patient preferred pharmacies query (`api/patients/{patientId}/pharmacies`). When we search for pharmacies to add we just query their API with params like zip code, pharmacy name, etc. That returns a list of ephemeral Organization resources that we can add to a patient by selecting. My thought was that we gonna add one at a time per search result vs multiple at the same time.

PR #8104, opened 2025-12-19
Profile nested type crawling

### LINKED ISSUE #8102: Resource crawler ignores inner types from profiles
`Crawler.crawlPropertyValue` and `AsyncCrawler.crawlPropertyValue` do not properly load types defined in resource profiles when invoking `getDataType`. Two approaches that work; unsure which would be the better fix; but option 2 feels more correct since it doesn't depend on `loadDataType` having been called and instead uses the profile `schema` directly.

### Option 1
```typescript
private crawlPropertyValue(value: TypedValueWithPath, path: string, schema: InternalTypeSchema): void {
  if (!isPrimitiveType(value.type)) {
    // Requires having called `loadDataType(profile);` before the crawler is used, e.g. in Repository.validateProfiles()
    const type = getDataType(value.type, schema.url). 
  }
}
```

### Option 2
```typescript
private crawlPropertyValue(value: TypedValueWithPath, path: string, schema: InternalTypeSchema): void {
  if (!isPrimitiveType(value.type)) {
    const type = schema.innerTypes?.find((t) => t.name === value.type) ?? getDataType(value.type);
  }
}
```

The `Invalid patient with US Core Patient profile (communication.language not in ValueSet)` test in `src/fhir/repo.test.ts` demonstrates the currently incorrect behavior.

mattlong:
Fix #8102

### DIFF
--- a/packages/eslint-config/index.mjs
+++ b/packages/eslint-config/index.mjs
@@ -101,7 +101,6 @@ export const coreConfig = {
       'error',
       {
         definedTags: ['category', 'experimental', 'ts-ignore'],
-        tagNamePreference: { defaultValue: 'default' },
       },
     ],
     'jsdoc/require-hyphen-before-param-description': ['error', 'always'],
@@ -119,6 +118,13 @@ export const coreConfig = {
         ' SPDX-License-Identifier: Apache-2.0',
       ],
     ],
+  },
+  settings: {
+    jsdoc: {
+      tagNamePreference: {
+        default: 'defaultValue',
+      },
+    },
   },
 };
 


packages/core/src/typeschema/crawler.test.ts
```diff
@@ -193,6 +196,126 @@ describe('ResourceCrawler', () => {
...
+    let usCorePatientProfile: InternalTypeSchema;
+
+    beforeAll(() => {
+      usCorePatientSD = JSON.parse(readFileSync(resolve(__dirname, '__test__', 'us-core-patient.json'), 'utf8'));
+      usCorePatientProfile = parseStructureDefinition(usCorePatientSD);
+    });
+
+    test.each([
+      ['no profile', () => undefined, 'http://hl7.org/fhir/ValueSet/languages'],
+      ['US Core patient profile', () => usCorePatientProfile, 'http://hl7.org/fhir/us/core/ValueSet/simple-language'],
```
noahsilas, line 215: Nice use of `test.each` cases!

packages/core/src/typeschema/crawler.ts
```diff
@@ -126,21 +126,15 @@ class Crawler {
...
     if (!isPrimitiveType(value.type)) {
       // Recursively crawl as the expected data type
-      // What is this the correct way to consider schema.innerTypes? e.g. for profiles
-      // See https://github.com/medplum/medplum/issues/8102
-      // Two ways that work in some cases but need further vetting after adding schema: InternalTypeSchema as a parameter to crawlPropertyValue:
-      // 1. const type = getDataType(value.type, schema.url). May require adding call to `loadDataType(profile);` in Repository.validateProfiles()
-      // 2. const type = schema.innerTypes?.find((t) => t.name === value.type) ?? getDataType(value.type);
-      // Similar changes should be made to AsyncCrawler.crawlPropertyValue
-      const type = getDataType(value.type);
+      const type = schema.innerTypes?.find((t) => t.name === value.type) ?? getDataType(value.type);
```
noahsilas, line 137: The schema stuff here is pretty challenging in it's complexity. It feels a little weird to be comparing the `name` and `type` attributes here. It looks to me like in these inner types have a "name" that matches the "type" (based on https://github.com/medplum/medplum/blob/v5.0.10/packages/core/src/typeschema/types.ts#L363-L364). I think this would _feel_ more correct if it used that `.type` instead, but I don't deeply understand the semantics of those fields yet.

No action required here, just thinking out loud.

packages/core/src/client.ts
```diff
@@ -350,7 +350,7 @@ export interface MedplumClientOptions {
    * - 'basic': Log method, URL, and status code only (no sensitive headers)
    * - 'verbose': Log all details including headers (may include sensitive data like tokens)
    *
-   * @default 'none'
```
noahsilas, line 353: Not a blocker: I'm curious what warning you were seeing about this tag? I don't see it when I run `npm run lint` in core.

Unrelated:  I think it's very funny that github automatically links to the user with the handle "default" in the UI when rendering this commit. They probably get pinged a lot from this kind of thing :joy:

PR #8105, opened 2025-12-19
feat(server): add subscription AuditEvent destination extension

rahul1:
Add support for controlling where AuditEvent resources generated by subscription attempts are sent. This extension allows specifying 'resource' (save to database) or 'log' (emit to server logs only), with support for multiple destinations.

## Changes
- Add subscription-audit-event-destination extension support
- Default behavior remains 'resource' (save to DB)
- Support for 'log' destination for high-volume scenarios
- Support for multiple destinations (both resource and log)
- Add comprehensive tests for all destination scenarios
- Update documentation with extension details and examples

## Testing
- Build: ✅ Passes
- Lint: ✅ Passes
- Tests: Infrastructure issues affecting all subscription tests (not specific to this change)

### DIFF
--- a/packages/server/src/workers/subscription.test.ts
+++ b/packages/server/src/workers/subscription.test.ts
@@ -1446,157 +1446,160 @@ describe('Subscription Worker', () => {
       expect(bundle.entry?.length).toStrictEqual(1);
     }));
 
-  test('Subscription AuditEvent destination - log only', () =>
-    withTestContext(async () => {
+  describe('Subscription AuditEvent destination with logging', () => {
+    let originalConsoleLog: typeof console.log;
+
+    beforeEach(async () => {
       const config = await loadTestConfig();
       config.logAuditEvents = true;
-
-      const project = (await createTestProject()).project.id;
-      const originalConsoleLog = console.log;
+      originalConsoleLog = console.log;
       console.log = jest.fn();
+    });
 
-      const subscription = await systemRepo.createResource<Subscription>({
-        resourceType: 'Subscription',
-        reason: 'test',
-        meta: {
-          project,
-        },
-        status: 'active',
-        criteria: 'Patient',
-        channel: {
-          type: 'rest-hook',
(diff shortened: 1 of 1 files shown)


packages/server/src/util/auditevent.ts
```diff
@@ -377,7 +380,32 @@ export async function createSubscriptionAuditEvent(
...
+
+  // Read destination extensions from subscription
+  const destinations: string[] = [];
+  if (subscription?.extension) {
+    for (const ext of subscription.extension) {
+      if (ext.url === SUBSCRIPTION_AUDIT_EVENT_DESTINATION_URL && ext.valueCode) {
+        destinations.push(ext.valueCode);
+      }
+    }
+  }
```
mattwiller, line 393: I'd recommend having a single extension with this `url` and a `valueCodeableConcept` containing the set of destination codes.  That way, you can just use the `getExtension()` helper from `@medplum/core` and get the codes directly from the result

packages/server/src/util/auditevent.ts
```diff
@@ -377,7 +380,32 @@ export async function createSubscriptionAuditEvent(
...
+
+  // Default to 'resource' if no extensions found
+  const finalDestinations = destinations.length > 0 ? destinations : ['resource'];
```
mattwiller, line 396: As a default, should we include `log` as well?

rahul1: that's a change in behavior - i was trying to be conservative for the time being

packages/server/src/util/auditevent.ts
rahul1: I understand that - but I was trying to mirror the behavior in the same feature on Bots, which has it as an array of separate values. 



image

packages/server/src/util/auditevent.ts
rahul1: This would be changing behavior and increase our log volume. To be conservative, I decided to have the default just be the existing behavior.

PR #8108, opened 2025-12-22
Tasks uses URL for filters and search

techdavidy:
Allow TaskBoard to work based on query params. 
e.g
Patient/01971836-657c-70ea-936c-a0fc75ac223b/Task?_count=20&_sort=-_lastUpdated&_total=accurate&owner=Practitioner%2Fc2a740d4-a120-4b78-94f6-6d24c9b46e03&patient=Patient%2F01971836-657c-70ea-936c-a0fc75ac223b

Page containing `TaskBoard` becomes responsible of handling navigation. 
Handled https://github.com/medplum/medplum/issues/8110
Working on a different approach to that test in a separate PR


techdavidy:
@claude please check again 
for navigate(...)?.catch(console.error) unfortunately due to react-router returning a promise, we decided to just log the error

kevinwadeshaw:
This is awesome! A few tangential issues we should address with filtering—not directly related to this PR, so can merge without updating:

- Let's remove "Cancelled" from Status selection in Filters—as well as in Task creation dialogue and Task Properties sidebar tab
  - In Provider app, we globally treat this status as "Delete" (i.e. the trash can icon action button) and we remove items with this status from the front end. Making a Status "Cancelled" via this action button will remove it from the Tasks list—so there is also nothing to Filter for that option.
- Change status badges to use 'light' versions, like we use in the Patient Summary sidebar
- Let's default-select the first item in the list: when the list reloads as a filter is added, let's show details of the first item in the list to the right
  - Let's do this on landing as well, so the only time we are showing an empty state on the right is when there are "No tasks available." in the list

PR #8111, opened 2025-12-22
Convert profiles-medplum.json to FSH

noahsilas:
## Strategy Discussion 

This presents an alternative strategy to #7899. It's more invasive, doing a full replacement of the current `profiles-medplum.json` file with a new FSH based subdirectory. The tradeoff here is that the build step is simplified: we run the "sushi" tool once to generate the file and move it into the right place, without needing to perform any kind of merge step.

I suspect that there is lots of room to improve from the generated FSH (which was automatically decompiled from the previous JSON file), but I'd like to invite early feedback on the approach before I start trying to hand-optimize any of that.

## Implementation notes

Upcoming work on Scheduling operations includes some FHIR extensions that become extremely unwieldy when all the snapshot attributes are bundled into the JSON representations of the FHIR definitions.

To make working on this easier, we start by converting the existing JSON file into FSH files, which lets us use tools to generate much of the verbose JSON. This was accomplished using the [`GoFsh` tool](https://github.com/FHIR/GoFSH):

```
npx gofsh -u 4.0.1 -o src/fsh src/fhir/r4/profiles-medplum.json
```

At build time, we use [`sushi`](https://github.com/FHIR/sushi) to transform from the FSH documents into the JSON representation. This can be done individually with `npm run build:fsh`, which is invoked from the `npm run build` command.

You can verify the correctness by comparing the generated JSON file to the source file as of the latest tag. (Note that to get a nice diff I am piping the version from v5.0.9 through `jq`. This removes some whitespace changes that otherwise are noisy when compared to the generated file.)

```
diff dist/fhir/r4/profiles-medplum.json <(git show v5.0.9:packages/definitions/src/fhir/r4/profiles-medplum.json | jq .)
```

There is one very minor difference remaining, which is that one node (`/entry/16/resource/jurisdiction/0/coding/0`) has its attributes sorted in a different order. This should have no impact on usage.


codyebberson:
Thanks for putting this together and exploring the approach.

After reflecting on this more, I think we should close this PR for now.

The main reason is not the implementation itself, but the scope of the change. This would represent a fairly significant shift in how Medplum authors and maintains FHIR definitions, including introducing FSH/SUSHI into the core development workflow. At the moment we don't have strong enough consensus that this is a direction we want to take, and it isn't a high enough priority relative to other work to justify the migration effort.

We also think that if we were to adopt FSH, we'd want to do so as a more comprehensive strategy rather than ending up with a mix of JSON- and FSH-authored artifacts.

That said, we appreciate the investigation and prototype work. FSH does offer some real advantages in readability and authoring experience, and this PR gives us a useful reference point if we decide to revisit the topic in the future.

Thanks again for taking the time to explore it.

PR #8114, opened 2025-12-23
feat(agent): AA-only enhanced ACK mode

### LINKED ISSUE #8113: [Agent] Add non-standard AA-only enhanced ACK mode
Some configurations don't work well with the HL7v2 standard 2-phase enhanced ACK mode as detailed by the HL7v2 spec. 

In order to get similar behavior at the cost of less interface-level information, we can enable the agent to send AAs immediately upon receiving a message, which will be processed asynchronously.

This enables higher concurrent throughput of messages at the cost of complexity in surfacing errors to the client.

ThatOneBro:
Fixes #8113


techdavidy:
@claude please review

codyebberson (approved):

PR #8115, opened 2025-12-24
Provider 80% test coverage

techdavidy:


techdavidy:
@claude please review

examples/medplum-provider/src/components/ChargeItem/ChargeItemList.test.tsx
```diff
@@ -161,4 +171,179 @@ describe('ChargeItemList', () => {
...
+      const label = input.closest('.mantine-InputWrapper-root')?.querySelector('label');
+      return label?.textContent?.includes('CPT Code');
+    }) as HTMLInputElement;
+
+    expect(cptInput).toBeDefined();
+
+    // Type in CPT code input
+    await act(async () => {
+      await user.type(cptInput, '99214');
+    });
```
noahsilas, line 246: userEvent automatically handles the `act` wrapping, so you should be able to remove this and avoid introducing a new warning in the test output here and in similar places.

examples/medplum-provider/src/components/ChargeItem/ChargeItemList.test.tsx
```diff
@@ -161,4 +171,180 @@ describe('ChargeItemList', () => {
...
+
+    // Wait for CPT code to be selected (check for selected items or code display)
+    await waitFor(
+      () => {
+        // Check if the code appears in selected items or if the input value changed
+        const selectedItems = document.querySelector('[data-testid="selected-items"]');
+        const hasCode =
+          selectedItems?.textContent?.includes('99214') || selectedItems?.textContent?.includes('Office Visit Level 4');
+        return hasCode === true;
```
noahsilas, line 269: returning a falsey value doesn't cause `waitFor` to defer. Maybe this could be an assertion? `expect(hasCode).toBe(true)`?

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -192,25 +245,27 @@ export function TaskBoard(props: TaskBoardProps): JSX.Element {
               <Flex h={64} align="center" justify="space-between" p="md">
                 <Group gap="xs">
                   <Button
-                    className={cx(classes.button, { [classes.selected]: filters.showMyTasks })}
+                    component={Link}
```
noahsilas, line 248: Would it make sense to use the `Link` component directly instead of wrapping it in a `Button`? Does the button wrapper provide some button like styling that you want here?

examples/medplum-provider/src/components/ChargeItem/ChargeItemPanel.tsx
```diff
@@ -98,7 +88,6 @@ export default function ChargeItemPanel(props: ChargeItemPanelProps): JSX.Elemen
               name="cptCode"
               path="cptCode"
               defaultValue={cptCodes}
-              onChange={updateCptCodes}
```
noahsilas, line 101: Double checking my understanding: this removal is safe because this field is always disabled and so this event was never fired. Right?

examples/medplum-provider/src/pages/patient/EditTab.tsx
```diff
@@ -12,12 +12,12 @@ import { ResourceFormWithRequiredProfile } from '../../components/ResourceFormWi
 import { RESOURCE_PROFILE_URLS } from '../resource/utils';
 
 const missingProfileMessage = RESOURCE_PROFILE_URLS.Patient ? (
-  <p>
+  <>
```
noahsilas, line 15: Nice fix for a noisy warning about nested `<p>` tags! Thank you!

examples/medplum-provider/src/pages/patient/IntakeFormPage.tsx
```diff
@@ -37,20 +57,38 @@ export function IntakeFormPage(): JSX.Element {
 
       await Promise.allSettled(
         valueSets.map(async (vs) => {
+          if (abortController.signal.aborted) {
+            return;
+          }
           const isAvailable = await checkValueSetAvailability(vs.url, medplum);
```
noahsilas, line 63: Potential follow up: teaching the medplum client to accept an AbortSignal in request methods so you can use the controller instantiated here to bail out early if this unmounts.

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
techdavidy: Yeah, It has te properties of a button.

Issue #8110, opened 2025-12-22
[Flaky test] IntakeFormPage


mattlong:
I've seen these tests timeout on several PRs over the past week or so.

An example: https://github.com/medplum/medplum/actions/runs/20427360903/job/58690673850?pr=8107

PR #8117, opened 2025-12-24
Cleanup provider test warnings

noahsilas:
I'm hoping to figure out why we have started seeing some flaky tests in here, and noticed that there are a bunch of warnings cluttering up the output when I run them. Let's start by cleaning up some of them. Relates to #8110.

## An update to {} inside a test was not wrapped in act(...)

The first one on my hitlist looks like this:

```
An update to DoseSpotFavoritesPage inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://react.dev/link/wrap-tests-with-act
An update to DoseSpotFavoritesPage inside a test was not wrapped in act(...).
```

What's not obvious at first glance is that one common cause of this is
when rendering a component schedules an asynchronous state change. This
happens because the `render()` helper that `@testing-library/react`
provides is synchronous, and so wraps the render call in a synchronous
`act` handler that doesn't wait for all the async operations.

In common cases this works well, but consider a component like the
following:

```
function MyComponent() {
  const [myState, setMyState] = useState('Work pending');
  useEffect(() => {
    async function doWork() {
      await Promise.resolve(); // Anything that introduces an async boundary
      setMyState('Work done')
    }
    doWork();
  });
  return <p>{myState}</p>;
}
```

Here, the useEffect handler is run synchronously during the render step,
but the function schedules a microtask when awaiting. That's enough to
break it out of the synchronous `act` handler, and cause these warnings
to fire.

There are a few ways to address these warnings:
- Wrap the render function in an async `act()` call
- Use a `waitFor` helper after rendering
- Using `findBy` queries (which implicitly perform a `waitFor` under the
  hood).

For cases where we our components are doing this kind of asynchronous
`setState` during an initial render, I feel that the most expressive
solution is to wrap the original render in an async `act()`, so that's
the approach I've taken in this commit.

There are also a couple places where UI events were not wrapped in
`act()` that I've bundled in to this same commit.

## Remove act wrappers around user-event functions
These functions handle the `act` wrapper internally. Wrapping them with
an `act` call ourselves generates this confusing warning:

> The current testing environment is not configured to support act(...)

My understanding is that while it should be unnecessary, it shouldn't
hurt to have this double wrapped `act`, but I haven't dug as deeply in
to this as the previous one. The warnings do go away when we follow the
published guidelines and let `@testing-library/user-event` handle this
on our behalf.

### DIFF
--- a/examples/medplum-provider/src/components/schedule/CreateVisit.test.tsx
+++ b/examples/medplum-provider/src/components/schedule/CreateVisit.test.tsx
@@ -95,10 +95,8 @@ describe('CreateVisit', () => {
         expect(screen.getByRole('button', { name: /Create Visit/i })).toBeInTheDocument();
       });
 
-      await act(async () => {
-        const submitButton = screen.getByRole('button', { name: /Create Visit/i });
-        await user.click(submitButton);
-      });
+      const submitButton = screen.getByRole('button', { name: /Create Visit/i });
+      await user.click(submitButton);
 
       await waitFor(() => {
         expect(screen.getByText(/Please fill out required fields/i)).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


techdavidy:
@claude please review

PR #8118, opened 2025-12-26
Handle out-of-order query results in `useSearch`

noahsilas:
There are a variety of situations where query results may come in out-of-order; there may be network congestion, one search may have higher complexity, a later request may be served from cache, ...

In such a case, we should not overwrite the results of a request with the slow results from an earlier request.

This method (closing over a variable to track if the effect has been cancelled) is recommended in the official React docs:
https://react.dev/reference/react/useEffect#fetching-data-with-effects

To make this fix more clearly correct, I started by removing a piece of state (`lastSearch`) that made thinking about the effect more difficult.


codyebberson:
@claude please review

PR #8119, opened 2025-12-26
Improve testing coverage in React package

techdavidy:


packages/react/src/NoteDisplay/NoteDisplay.test.tsx
```diff
@@ -48,4 +48,18 @@ describe('NoteDisplay', () => {
...
+    await act(async () => {
+      const { container } = render(
+        <MemoryRouter>
+          <MedplumProvider medplum={medplum}>
+            <NoteDisplay value={undefined} />
+          </MedplumProvider>
+        </MemoryRouter>
+      );
+
+      expect(container.firstElementChild).toBeNull();
```
noahsilas: #nit: I think that assertions are typically done outside of the `act()` wrapper. I think that this should work in this case where we don't need to observe any changes from actions.

I think in this case you can actually drop the `act` wrapper entirely - the `render` helper implicitly wraps with a synchronous `act` wrapper, and so will catch anything that happens without an async deferral. Especially since this assertion is inside the async `act` wrapper you know that it's safe to remove.

PR #8120, opened 2025-12-28
Use native time input for time fields

kvpendergast:
Screenshot 2025-12-28 at 4 30 41 PM
Screenshot 2025-12-28 at 4 30 29 PM
Right now PropertyType.time is handled as a regular text input. This swaps it out for an actual <input type="time" step={1}> so users get a proper time picker instead of typing values manually.

The step={1} lets users pick seconds, which is a requirement for the FHIR time data type (https://hl7.org/fhir/R4/datatypes.html#time)

Server validation still catches bad input, but this makes it less likely users will enter something wrong in the first place.

Side note: This is my first PR into the medplum repo, so please let me know if I'm missing a requirement for contribution.

### DIFF
--- a/packages/react/src/ResourcePropertyInput/ResourcePropertyInput.test.tsx
+++ b/packages/react/src/ResourcePropertyInput/ResourcePropertyInput.test.tsx
@@ -111,6 +111,28 @@ describe('ResourcePropertyInput', () => {
     expect(onChange).toHaveBeenCalledWith(isoString, 'dateTime');
   });
 
+  test('Time property', async () => {
+    const onChange = jest.fn();
+
+    await setup({
+      ...defaultProps,
+      name: 'time',
+      property: { ...baseProperty, type: [{ code: 'time' }] },
+      onChange,
+    });
+
+    const input = screen.getByTestId('time');
+    expect(input).toBeDefined();
+    expect(input).toHaveAttribute('type', 'time');
+    expect(input).toHaveAttribute('step', '1');
+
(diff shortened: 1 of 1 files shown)


packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
```diff
@@ -277,6 +276,19 @@ export function ElementDefinitionTypeInput(props: ElementDefinitionTypeInputProp
...
+    case PropertyType.time:
+      return (
+        <TextInput
+          {...getPrimitiveInputProps()}
+          type="time"
+          step={1}
+          onChange={(e) => {
+            if (onChange) {
+              onChange(e.currentTarget.value);
+            }
```
mattwiller, line 288: I think you could simplify this to `onChange?.(e.currentTarget.value)`

mattwiller (approved):

packages/react/src/ResourcePropertyInput/ResourcePropertyInput.tsx
kvpendergast-unity: definitely. I just followed convention from the other statements.

techdavidy (approved):

techdavidy:
Mind adding a test for this case please ?

kvpendergast-unity:
> Mind adding a test for this case please ?

done!

codyebberson:
Thanks @kvpendergast !

PR #8123, opened 2025-12-29
DoseSpot Validation Update

finnbergquist:
Docs update to reflect the correct validation logic.

### DIFF
--- a/packages/docs/docs/integration/dosespot/getting-started.md
+++ b/packages/docs/docs/integration/dosespot/getting-started.md
@@ -70,7 +70,7 @@ For patients under 18 years of age, you **must** also sync Height and Weight as 
     {
       "system": "phone",
       "use": "home",
-      "value": "6175672093"//Required: cannot have a +1 prefix and must be 9 digits
+      "value": "6175672093"//Required: cannot have a +1 prefix or start with 555, and must be 10 digits
     }
   ],
   "address": [//Required


packages/docs/docs/integration/dosespot/getting-started.md
```diff
@@ -70,7 +70,7 @@ For patients under 18 years of age, you **must** also sync Height and Weight as
     {
       "system": "phone",
       "use": "home",
-      "value": "6175672093"//Required: cannot have a +1 prefix and must be 9 digits
+      "value": "6175672093"//Required: cannot have a +1 prefix or start with 555, and must be 10 digits
```
noahsilas, line 73: Interesting! Does this mean that they disallow the 555 area code (555xxxxxxx) or that they disallow 555 exchanges (xxx555xxxx)?

As a fun side note, within the 555 exchange only the lines 555-0100 through 555-0199 are reserved for fictional use. The remaining lines on that exchange _could_ be assigned, although it seems unlikely that they will be anytime soon because of the misconception that the entire exchange is reserved. 555-1212 is still in use for directory assistance (although it makes sense that folks like dosespot don't allow using it for a patient).

finnbergquist: Very interesting. Not sure exactly what they disallow, but we have run into this validation a few times.

PR #8124, opened 2025-12-30
SSE bot implementation

techdavidy:
### New Streaming Execution Function 

Adds executeBotStreaming() function that mirrors executeBot() but with streaming support
Routes to appropriate runtime executor based on bot.runtimeVersion:

awslambda → runInLambdaStreaming()
vmcontext → runInVmContextStreaming()
fission → Returns error (doesn't support streaming)

### VM Context Streaming

Adds runInVmContextStreaming() function
Expands the sandbox with modern web APIs: ReadableStream, WritableStream, TransformStream, Headers, Request, Response, TextDecoderStream, TextEncoderStream
Switches from node-fetch to globalThis.fetch
Passes onChunk callback to bot handlers for streaming chunks

### AWS Lambda Streaming (deploy.ts)

Converts Lambda handler to use awslambda.streamifyResponse() for response streaming
Bot handlers receive an onChunk callback when in streaming mode
Chunks are written to responseStream with a special format:

{ __medplum_chunk__: chunk } for streamed data
{ __medplum_result__: result } for final result
{ __medplum_error__: message } for errors

## How It Works
Bots can now use streaming by accessing the onChunk callback in their event object:
```
export async function handler(medplum, event) {
  const { onChunk } = event;
  
  // Stream chunks to client
  await onChunk({ type: 'progress', data: 'Processing...' });
  await onChunk({ type: 'data', data: someResult });
  
}
```

### DIFF
--- a/examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
+++ b/examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
@@ -49,11 +49,16 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
   const [refreshKey, setRefreshKey] = useState(0);
   const [selectedResource, setSelectedResource] = useState<string | undefined>();
   const scrollViewportRef = useRef<HTMLDivElement>(null);
+  const isSendingRef = useRef(false);
 
   // Load conversation when topic changes
   useEffect(() => {
     const topicId = topic?.id;
     if (topicId) {
+      // Skip loading if we're actively sending (we just created this topic)
+      if (isSendingRef.current) {
+        return;
+      }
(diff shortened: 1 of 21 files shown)


codyebberson:
@claude - can you please review the proposed architecture here?  In particular, can you please review the use of `__medplum_chunk__`?  Is that necessary when using AWS Lambda streaming?  Or could we do standard "Chunked Transfer Encoding" for everything, and treat it as a simple straight proxy?

codyebberson:
@claude - thank you.  Can you please elaborate on the "Alternative Approach (Simpler)" option?  If the only limitation is "no final return value", then I think that might be worth it.  Can you please provide an estimate for how much that would simplify here?

PR #8125, opened 2025-12-30
Strengthen `MemoryRepository` constraints

noahsilas:
Until now, the MemoryRepository allows creating objects with fields that normally would be managed by the Repository.

- This can allow you to load historical data into the repository without alterations

- This facilitates testing by allowing known fixtures to be loaded.

We used to allow this anywhere, but that can cause some surprising behavior. For instance, creating multiple objects with matching IDs can cause erratic behavior in the resource history.

The following changes have been made in this PR:

- Calling `repo.createResource` will now fail if a resource with that ID is already present. Callers should use `updateResource()` after the first time, which more clearly expresses their intention.

- Passing `meta.versionId` or `meta.lastUpdated` is ignored unless the repository is in "seeding" mode. For example, to seed an old resource, use the following:
  ```
    repo.withSeeding(() => repo.createResource({
      ...
      meta: { lastUpdated: '2020-01-01T00:00:00.000Z' }
    }));
  ```

  Using this wrapper ensures that the seeding state is released after any operations in the block complete.

This alteration revealed a number of small improvements in the test suite. In some places this removes ambiguity, in some it highlights where we can remove duplicative data seeding, and it removes some React duplicate key warnings (as the repo doesn't hold the duplicates anymore).

### DIFF
--- a/packages/fhir-router/src/repo.test.ts
+++ b/packages/fhir-router/src/repo.test.ts
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { OperationOutcomeError, WithId } from '@medplum/core';
+import type { WithId } from '@medplum/core';
 import {
   badRequest,
   createReference,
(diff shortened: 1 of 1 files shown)


codyebberson:
@claude please review

techdavidy (approved):

PR #8126, opened 2025-12-30
Provider - Communication/Message inbox to support URLs 

techdavidy:
- Modified component so it works similar to SearchControl. 
- All items and button use url navigation rather than just internally mounting or updating values.
- Allows for multiple filters in follow up PR. e.g. Filtering by practitioner

### DIFF
--- a/examples/medplum-provider/src/components/messages/ChatList.test.tsx
+++ b/examples/medplum-provider/src/components/messages/ChatList.test.tsx
@@ -67,11 +67,7 @@ describe('ChatList', () => {
       <MemoryRouter>
         <MedplumProvider medplum={medplum}>
           <MantineProvider>
-            <ChatList
-              threads={threads}
-              selectedCommunication={selectedCommunication}
-              getThreadUri={mockGetThreadUri}
-            />
+            <ChatList threads={threads} selectedCommunication={selectedCommunication} getThreadUri={mockGetThreadUri} />
           </MantineProvider>
         </MedplumProvider>
       </MemoryRouter>
--- a/examples/medplum-provider/src/hooks/useThreadInbox.ts
+++ b/examples/medplum-provider/src/hooks/useThreadInbox.ts
@@ -47,8 +47,9 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
 
     const bundle = await medplum.search('Communication', searchParams.toString(), { cache: 'no-cache' });
     const parents =
-      bundle.entry?.map((entry) => entry.resource as Communication).filter((r): r is Communication => r !== undefined) ||
-      [];
+      bundle.entry
+        ?.map((entry) => entry.resource as Communication)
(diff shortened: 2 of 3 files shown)


examples/medplum-provider/src/components/messages/ThreadInbox.tsx
```diff
@@ -19,82 +19,107 @@ import {
...
 
-  const itemsPerPage = 20;
+  const searchParams = useMemo(() => new URLSearchParams(query), [query]);
+  const itemsPerPage = Number.parseInt(searchParams.get('_count') || '20', 10);
+  const currentOffset = Number.parseInt(searchParams.get('_offset') || '0', 10);
+  const currentPage = Math.floor(currentOffset / itemsPerPage) + 1;
+  const status = (searchParams.get('status') as Communication['status']) || 'in-progress';
 
-  const queryWithStatus = useMemo(() => `${query}&status=${status}`, [query, status]);
+  const effectiveQuery = query;
```
noahsilas: It looks like this is creating an additional variable name binding without any real benefit? The code below actually gets a little less readable IMO (as it isn't using the object shorthand like `useThreadInbox({ query, threadId })`).

Is this following some other convention or pattern I should learn more about?

examples/medplum-provider/src/components/messages/ThreadInbox.tsx
```diff
@@ -114,44 +139,50 @@ export function ThreadInbox(props: ThreadInboxProps): JSX.Element {
...
-                          <Flex direction="column" gap="xs">
-                            <Skeleton height={16} width={`${Math.random() * 40 + 60}%`} />
-                            <Skeleton height={14} width={`${Math.random() * 50 + 40}%`} />
-                          </Flex>
-                        </Box>
-                      </Flex>
-                    ))}
+                    {Array.from({ length: 10 }).map((_, index) => {
+                      const titleWidth = skeletonTitleWidths[index % skeletonTitleWidths.length];
+                      const subtitleWidth = skeletonSubtitleWidths[index % skeletonSubtitleWidths.length];
```
noahsilas: Neat way to add some variations in to the loading elements here!

examples/medplum-provider/src/components/messages/ChatListItem.tsx
```diff
@@ -12,11 +12,11 @@ interface ChatListItemProps {
   topic: Communication;
   lastCommunication: Communication | undefined;
   isSelected: boolean;
-  onSelectedItem: (topic: Communication) => string;
+  getThreadUri: (topic: Communication) => string;
```
noahsilas: Refactoring this name looks like it was more in-depth than I would have guessed, but I think it makes the intention much clearer. Thank you for all this work!

techdavidy:
@claude please check pr

codyebberson:
@claude - please review again, I believe all of your previous feedback has been resolved.

PR #8127, opened 2026-01-01
Do not merge: this is a sample of an automated HL7 handler developed …

reshmakh:
…by Claude Code - it is not standard HL7


codyebberson:
@reshmakh  What would you like to do with this one?

PR #8129, opened 2026-01-01
Prefer createReference over getReferenceString

codyebberson:
Mostly minor cleanup

1. Prefer using `createReference()` over `getReferenceString()`
   * A simple FHIR reference looks like this: `{ reference: 'ResourceType/id' }`
   * `@medplum/core` has several utility methods to help work with references
   * `getReferenceString()` takes a resource, and returns the reference string such as `Patient/123`
   * `createReference()` takes a resource, and returns a full reference
   * The main difference is that `createReference()` will also opportunistically populate `Reference.display`
   * Many Medplum React components will opportunistically use that `Reference.display` string to show better display values
   * So, by using `createReference()` as much as possible, it helps keep the UI clean
2. Prefer using `WithId<T>` over `T` when possible
   * One of the big changes in Medplum 4.0 was the `WithId<T>` type
   * This essentially just extends a `Resource` type to declare that the `id` property is populated
   * This can help to eliminate lots of boilerplate `if (id)` checks
3. Prefer resource types over strings for function arguments
   * If a function takes a `patientId: string` argument, that can be error prone, because any string could be passed in
   * Instead, if the function takes a `patient: Patient` argument, then the compiler can help catch many bugs


codyebberson:
@claude please review

mattwiller (approved):

examples/medplum-provider/src/components/ChargeItem/ChargeItemPanel.test.tsx
```diff
@@ -2,15 +2,16 @@
...
 import { MockClient } from '@medplum/mock';
-import { describe, expect, test, beforeEach, vi } from 'vitest';
+import { MedplumProvider } from '@medplum/react';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 import ChargeItemPanel from './ChargeItemPanel';
 
-const mockChargeItem: ChargeItem = {
+const mockChargeItem: WithId<ChargeItem> = {
```
noahsilas, line 15: One other option would be to replace this type annotation with `satisfies ChargeItem`. That would let TypeScript use the concrete object type it infers from the object literal (with the ID field set) instead of the `ChargeItem` type that has ID defined as an optional.

Thinking about the long term here, I wonder if defining this the other way around would be preferable? It seems like most of the time you are dealing with a concrete instance with an ID, and only in creating new instances is the ID optional. It would be a major breaking change (so maybe something not in the cards until Medplum 6), but we could change the type definitions so that `ChargeItem` is defined with a required ID field, and we could define a utility like `Create<ChargeItem>` to use in cases where the ID field is optional.

None of this is blocking, all of this looks great and inline with current patterns. 👍

examples/medplum-provider/src/components/spaces/ResourcePanel.tsx
```diff
@@ -26,13 +26,13 @@ export function ResourcePanel<T extends Resource = Resource>(props: ResourcePane
...
-        return <TaskDetailPanel task={displayResource as Task} />;
+        return <TaskDetailPanel task={displayResource} />;
       case 'DiagnosticReport':
-        return <LabResultDetails result={displayResource as DiagnosticReport} />;
+        return <LabResultDetails result={displayResource} />;
       case 'ServiceRequest':
-        return <LabOrderDetails order={displayResource as ServiceRequest} />;
+        return <LabOrderDetails order={displayResource} />;
       case 'Encounter':
-        return <EncounterChart encounter={displayResource as Encounter} />;
```
noahsilas: Removing many type casts across this PR makes me very happy 😁

rahul1 (approved):

PR #8137, opened 2026-01-02
Add resource count button to project admin page

maddyli:


packages/app/src/admin/ProjectDetailsPage.tsx
```diff
@@ -1,14 +1,105 @@
...
+        entry: resourceTypes.map((resourceType: ResourceType) => ({
+          request: {
+            method: 'GET',
+            url: `${resourceType}?_summary=count`,
+          },
+        })),
+      };
+
+      // Execute the batch request
+      const response = await medplum.executeBatch(request);
```
noahsilas, line 57: Strategy thought: While I like the idea of doing a single batch request, I think that you may be better off making individual calls via the medplum client:
- The individual `get` requests can have their results cached (a batch query doesn't automatically unwrap to inspect the cache for each entry, although maybe it should? Potential future performance improvement)
- The client will auto-batch these requests. It may later be extended to try to be more clever about things like batch size optimizations - for example, it might be preferable to send multiple smaller batches instead of bundling all these into a single request. Using the global client batching means you will get to take advantage of such changes by default.
- In a world where the results are not all in a single batch, you could populate the table progressively as queries return, although that's not quite as appealing in this version where the table is sorted by the results of these queries.

maddyli:
https://github.com/medplum/medplum/pull/8153

PR #8138, opened 2026-01-02
Provider - Tasks auto select when empty

techdavidy:
Handled feedback from comment.
https://github.com/medplum/medplum/pull/8108#issuecomment-3685428134


techdavidy:
@claude please review

PR #8140, opened 2026-01-03
Fixed docker-compose.full-stack.yml after hardened images

codyebberson:
1. Server
   * Cannot use `sh` to conditionally choose config file vs environment variables
   * So, instead, we're just going to default to using env vars
   * Users can still use the config file behavior
2. App
   * Cannot use bash to do environment variable replacement for `MEDPLUM_BASE_URL` and other config settings
   * I went down a deep rabbithole of trying to use the built-in `envsubst` feature
   * But despite the docs saying it's available, it's not actually there in the hardened image
   * Bug filed here: https://github.com/docker-hardened-images/catalog/issues/92
   * Instead, just reverting the hardened image thing entirely for `medplum-app`

--------

### Using file-based config

If you need file-based config, create a `docker-compose.override.yml`:

```yaml
services:
  medplum-server:
    command: ['file:/usr/src/medplum/packages/server/medplum.config.json']
    volumes:
      - ./medplum.config.json:/usr/src/medplum/packages/server/medplum.config.json
```

Then run as normal - Docker Compose automatically merges the override file:

```bash
docker compose up
```

Or explicitly:

```bash
docker compose -f docker-compose.full-stack.yml -f docker-compose.override.yml up
```

### DIFF
--- a/packages/app/Dockerfile
+++ b/packages/app/Dockerfile
@@ -1,7 +1,10 @@
-FROM dhi.io/nginx:1
+FROM nginxinc/nginx-unprivileged:alpine
+
+# Start off as user root so we have permissions to run APK and chown
+USER root
 
 # Copy nginx configuration
-COPY <<EOF /etc/nginx/nginx.conf
+COPY <<EOF /etc/nginx/conf.d/default.conf
 server {
     listen 3000;
     server_name localhost;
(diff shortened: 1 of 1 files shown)


codyebberson:
@claude please review

PR #8143, opened 2026-01-03
Update docker-compose.full-stack.yml to add medplum-provider service

gabriel-virtualizehealth:


docker-compose.full-stack.yml
```diff
@@ -119,3 +119,28 @@ services:
       interval: 10s
       timeout: 5s
       retries: 5
+
+  # Medplum provider container (your fork) - using prebuilt image (Option B)
+  medplum-provider:
+    image: gabriel-virtualizehealth/medplum-provider:latest
```
codyebberson, line 125: We should probably put together an official Docker image for `medplum-provider`

codyebberson:
We will revisit this after migrating `medplum-provider` from `example` to `packages`.

PR #8150, opened 2026-01-05
Provider - include test data for empty projects

techdavidy:
Seeds data automatically from App.tsx.
This checks for Patient or PlanDefinition, if no records are available then it create test data for each case.


techdavidy:
@noahsilas  Yeah definitely code we don't want to run every single time. 
Just demo what we want to seed and move to a separate step.

PR #8152, opened 2026-01-05
Provider - Tasks include priority

techdavidy:
Include priority filter in Tasks. 
Once priority is selected it show selections on filters and adds filter to the query string.

Screenshot 2026-01-05 at 1 39 50 PM


techdavidy:
@claude please review

Issue #6314, opened 2025-04-06
Generator now produces different expressions for search parameters in docs


ThatOneBro:
After #6266 our generator now rewrites expressions in docs for search parameters with additional parenthesis

<img width="1253" alt="Image" src="https://github.com/user-attachments/assets/7a348b94-90bc-4f5a-ac0a-053245146d00" />

We could potentially read the expression directly from JSON or potentially just do the one-time rewrite, although that feels kind of wrong.

codyebberson:
Blerg - another argument for a more surgical approach to handling parentheses in the FHIRPath parser.

Reading directly from the source feels ok-ish to me.

PR #8154, opened 2026-01-05
Regenerate some built files

noahsilas:
- Some of the docs pages have updates when running the generator. Checking in those changes now keeps future PRs tidy.
- The `prettier-plugin-organize-imports` package was rewriting our schema migrations barrel file using a lexical sort. This unintentionally pushes updates out of order. Added a magic comment string to prevent this behavior and regenerated the file.

### DIFF
--- a/packages/server/src/migrations/migrate-main.ts
+++ b/packages/server/src/migrations/migrate-main.ts
@@ -68,6 +68,10 @@ function getNextSchemaVersion(): number {
 
 function rewriteMigrationExports(): void {
   const b = new FileBuilder();
+  b.append(
+    '// organize-imports-ignore - https://github.com/simonhaenisch/prettier-plugin-organize-imports?tab=readme-ov-file#skip-files'
+  );
+  b.newLine();
   const filenamesWithoutExt = getMigrationFilenames()
     .map(getVersionFromFilename)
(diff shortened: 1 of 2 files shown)


ThatOneBro (changes requested):
A lot of these changes are due to a long outstanding regression in the generator, causing it to generate excess parens. Typically we've been working around it by just not committing those changes, though we should probably take the time to fix it now...

See: https://github.com/medplum/medplum/issues/6314

Also FHIRcast is the correct spelling, I wonder how that ended up changing in the docs 🤔

noahsilas:
Thanks for the context @ThatOneBro ! I'll time-box an attempt at answering these questions to see if this is salvageable, or if I'll punt on this and continue the strategy of not committing these changes.

noahsilas:
The FHIRcast -> FHIRCast change is coming from the HL7 spec document (https://hl7.org/fhir/R4/subscription.html#bnr). It looks like that whole chunk is gone in the CI build (https://build.fhir.org/subscription.html#bnr), so that inconsistency should eventually go away.

noahsilas:
Getting the parsing updates handled is more involved than my time-box allows right now, so I'm abandoning this for the time being. 😢

PR #8164, opened 2026-01-06
Disable access token when ProjectMembership set to inactive

### LINKED ISSUE #8078: Validate ProjectMembership.active on all requests
### Overview
If `ProjectMembership.active` gets set to false, any previously generated token for that ProjectMembership will continue to work until expiration.

This can cause confusion when an admin User deactivates another User's ProjectMembership. Their access takes an hour (default token expiration) to be revoked, causing confusion and potentially distrust. 

### Implementation details

 We already check if the membership exists on each request [here](https://github.com/medplum/medplum/blob/0148ca5fd00602cc36b9622b28fda8421c12d4a4/packages/server/src/oauth/utils.ts#L935-L936). 

Add an additional check to confirm that `ProjectMembership.active != false`

mattwiller:
Resolves #8078

### DIFF
--- a/packages/server/src/oauth/utils.ts
+++ b/packages/server/src/oauth/utils.ts
@@ -32,12 +32,12 @@ import type {
 } from '@medplum/fhirtypes';
 import bcrypt from 'bcryptjs';
 import type { Request } from 'express';
-import type { IncomingMessage } from 'http';
 import type { JWTPayload, VerifyOptions } from 'jose';
 import { jwtVerify } from 'jose';
 import fetch from 'node-fetch';
 import assert from 'node:assert/strict';
 import { createHash, timingSafeEqual } from 'node:crypto';
+import type { IncomingMessage } from 'node:http';
 import { authenticator } from 'otplib';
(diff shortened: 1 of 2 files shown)


packages/server/src/oauth/utils.ts
```diff
@@ -935,7 +935,10 @@ export async function getLoginForAccessToken(
   }
 
   const membership = await systemRepo.readReference<ProjectMembership>(login.membership);
-  const project = await systemRepo.readReference<Project>(membership.project as Reference<Project>);
+  if (membership.active === false) {
+    return undefined;
+  }
```
mattwiller, line 940: This is the main fix; other changes are largely lints

noahsilas: Were those things you fixed by hand or is there a linter that can handle these updates? I wonder if it would be good to try and autofix a ton of these in a single patch so that future diffs can be more minimal?

mattwiller: I believe they came from Sonar — not sure how to auto-fix those ones :-\

PR #8166, opened 2026-01-07
Provider - improve testing coverage

techdavidy:


examples/medplum-provider/src/components/encounter/BillingTab.test.tsx
```diff
@@ -844,4 +844,173 @@ describe('BillingTab', () => {
...
+    await waitFor(() => {
+      expect(screen.getByText('CMS 1500 Form')).toBeInTheDocument();
+    });
+
+    await user.click(screen.getByText('CMS 1500 Form'));
+
+    await waitFor(() => {
+      expect(medplum.post).toHaveBeenCalled();
+    });
+  });
```
noahsilas, line 899: The test description (shows error...) sounds like we should probably have some kind of assertion in here about an error message being displayed?

examples/medplum-provider/src/components/schedule/CreateVisit.test.tsx
```diff
@@ -186,4 +186,115 @@ describe('CreateVisit', () => {
...
+    test('updates patient when patient is selected', async () => {
+      await act(async () => {
+        setup(mockSlotInfo);
+      });
+
+      await waitFor(() => {
+        expect(screen.getByLabelText(/Patient/i)).toBeInTheDocument();
+      });
+
+      const patientInput = screen.getByLabelText(/Patient/i);
```
noahsilas: You can remove a little boilerplate here by switching to the `findBy*` queries, which waits up to 1s and rejects if no suitable element is found by then.
```suggestion
      const patientInput = await screen.findByLabelText(/Patient/i);
```

examples/medplum-provider/src/components/spaces/ResourcePanel.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+      resourceType: 'Practitioner',
+      id: practitionerId,
+      name: [{ given: ['Alice'], family: 'Smith' }],
+    };
+    await medplum.createResource(practitioner);
+
+    setup(practitioner);
+
+    await waitFor(() => {
+      expect(screen.getByTestId('resource-panel')).toBeInTheDocument();
```
noahsilas, line 196: Should we be doing some kind of assertion looking deeper to see that a ResourceTable is actually being used? Or is it sufficient to just say that it rendered a Practitioner without errors?

examples/medplum-provider/src/pages/encounter/EncounterModal.test.tsx
```diff
@@ -98,4 +109,226 @@ describe('EncounterModal', () => {
...
+    });
+
+    // Wait for autocomplete options and select
+    await waitFor(
+      () => {
+        const option = screen.queryByText(/ambulatory/i);
+        if (option) {
+          return true;
+        }
+        return false;
```
noahsilas: This looks like it could be simpilfied to `return Boolean(option)`, but my understanding is that `waitFor` won't wait when the function returns falsey, only when it actually throws an error. Switching to `await screen.queryByText(...)` seems likely to be closest to what you want here.

Having said that, I notice that there is a `catch(() => {})` chained on here, so I think I might actually be quite confused about the intention of this block. If you never find this element on the screen is that OK?

examples/medplum-provider/src/components/spaces/ResourcePanel.test.tsx
techdavidy: I think it is fine. Plus we exploring different ways to display resources.

PR #8169, opened 2026-01-07
Reserve database connections for healthcheck

mattwiller:


ianplunkett:
Fixes https://github.com/medplum/medplum-customer-tasks/issues/729

PR #8160, opened 2026-01-06
Add `ActivityDefinition.code`, `Communication.priority`, `Communication.priority-order`, `ProjectMembership.active` search params

noahsilas:
We're bundling these new search parameters into a single update to reduce how often folks will need to go through reindexing steps.

# ActivityDefinition.code (#8057)
Adds the ability to filter and sort ActivityDefinition resources by the "code" field. This will be used to support upcoming [scheduling](https://github.com/medplum/medplum/issues?q=state%3Aopen%20label%3Ascheduling) features.

# Communication.priority, Communication.priority-order (#7760)
Adds the ability to filter and sort Communication resources by the "priority" field. Also added `priority-order`, which allows applying a sort where the resources are scored by the `priority` (roughly: `{ stat: 50, asap: 40, urgent: 30, routine: 20 }[resource.priority] ?? 10`). This lets you sort in the way humans mostly expect ("stat" before "asap" before "urgent" ...) and in line with the documented meaning of the [Request Priority ValueSet](https://www.hl7.org/fhir/valueset-request-priority.html). 

# ProjectMembership.active
Adds the ability to filter ProjectMembership resources by the "active" field. This was requested through the Medplum FDE team.

PR #8170, opened 2026-01-07
Automatically generate separate pre/post deploy migrations

noahsilas:
Previously, when there were a mix of changes that included pre-deploy and post-deploy migrations, the process to prepare the migration was fairly involved. @mattlong provided a detailed write up here: https://github.com/medplum/medplum/pull/8160/changes/BASE..18e75274eaa908c0722742456747501c83dad0e7#r2666672197

In a nutshell, it boiled down to:
- Generate the pre-deploy migration (`npm run migrate -- --skipPostDeploy`)
- Execute that migration
- Generate the post deploy migration (`npm run migrate -- --allowPostDeploy`) and convert the resulting pre-deploy migration into a post-deploy migration

Instead, as we are generating migrations we now build up separate "preDeploy" and "postDeploy" actions lists that are returned to the caller. The `migrate` CLI uses these to build separate migration scripts in the appropriate places. Other places like the `$db-schema-diff` operator can concatenate the results into a single response, retaining behavior matching the earlier form.

As a result, we no longer need some of the flags like `--skipPostDeploy` and `--allowPostDeploy`, and we don't need to thread that optionality through layers of context.

### DIFF
--- a/packages/server/src/migrations/migrate-main.test.ts
+++ b/packages/server/src/migrations/migrate-main.test.ts
@@ -1,0 +1,113 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import fs from 'node:fs';
+import { join } from 'node:path';
+import { addDataMigrationToManifest, DATA_DIR } from './migrate-main';
+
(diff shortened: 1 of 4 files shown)


noahsilas:
Oops, missing appending the result to `src/migrations/data/data-version-manifest.json`, update pending.

PR #8172, opened 2026-01-07
Allow projectmembership user display to be edited

maddyli:


codyebberson (changes requested):
I believe that @mattwiller concluded that this strategy will require deeper changes to how we enforce "readonly" on nested properties, and that instead the more straightforward fix here is a `$rebuild-display-strings` operation.

codyebberson:
Replaced by https://github.com/medplum/medplum/pull/8357

PR #8176, opened 2026-01-08
Avoid re-wrapping too-many-requests error in `fetchTokens`

noahsilas:
We received a report of issues coming from this `fetchTokens` command that turned out to be a rate limiting error. In that scenario, the JSON response is alraedy an OperationOutcome representing the rate limit exhaustion. This object didn't have an `error_description` attribute, so we were creating an error based on `badRequest(undefined)`, which does not help the user understand what happened.

- We now detect if the error object we got back is already an OperationOutcome, and use it as the bais for the OperationOutcomeError if so.

- That block is wrapped in a try-catch that was wrapping the result in another layer of OperationOutcomeError. This is helpful for consistency if the error is of an unexpected type, but when we already have an OperationOutcomeError this could obscure the root cause. In that scenario, we now re-throw the original OperationOutcomeError.

- If we get back any other object without an `error_description` attribute, we use the whole thing as the error message and re-raise it. This will hopefully replace some instances of an unhelpful "Unknown error" message with something that more accurately points to the root cause.

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -1208,7 +1208,7 @@ describe('Client', () => {
       await client.startClientLogin(clientId, clientSecret);
       throw new Error('test');
     } catch (err) {
-      expect((err as Error).message).toBe('Invalid client');
+      expect((err as Error).message).toBe('Failed to fetch tokens: Invalid client');
     }
   });
 
@@ -1246,7 +1246,7 @@ describe('Client', () => {
       await client.startClientLogin(clientId, clientSecret);
       throw new Error('test');
     } catch (err) {
-      expect((err as Error).message).toBe('Too Many Requests');
+      expect((err as Error).message).toBe('Failed to fetch tokens: Too Many Requests');
     }
   });
 
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -4267,6 +4267,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
       throw new Error(JSON.stringify(error));
     } catch (err) {
       if (err instanceof OperationOutcomeError) {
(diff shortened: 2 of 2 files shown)


packages/core/src/client.test.ts
```diff
@@ -1209,6 +1226,30 @@ describe('Client', () => {
...
+        },
+      ],
+    }));
+    const client = new MedplumClient({ fetch });
+    try {
+      client.setBasicAuth(clientId, clientSecret);
+      await client.startClientLogin(clientId, clientSecret);
+      throw new Error('test');
+    } catch (err) {
+      expect((err as Error).message).toBe('Too Many Requests');
```
noahsilas: It's arguable that this should be some composition of the error message returned from the server as well as the "Failed to fetch tokens" message in the client. This could be misleading down the line if fetchTokens is invoked prior to some other query with different rate limiting. Maybe we should continue wrapping the error and just make sure that we are setting the `cause` attribute more correctly in all these cases?

codyebberson: Good question.  

1. `Failed to fetch tokens`
2. `Too many requests`
3. `Failed to fetch tokens: Too many requests`

The last one would be most explicit.

noahsilas: For a quick fix I picked a sort of ugly strategy of mutating the error's `message` attribute by prepending "Failed to fetch tokens: ": https://github.com/medplum/medplum/pull/8176/changes/1b2ce4c26062a95096a7dbb82d27664c8ceebff4#diff-858339789c798051d9c13620bffce8ce860e7ad8ba18dba1c8711eb1f7d51a55R4270

Let me know if you think spending time on a more comprehensive Error handling strategy is better here.

codyebberson (approved):

PR #8177, opened 2026-01-08
Provider - Include Messages participant filter.

techdavidy:
Allow filtering messages by participant.
Once participant is selected, the filter gets applied to the query string. 

Screenshot 2026-01-08 at 3 42 24 PM

### DIFF
--- a/examples/medplum-provider/.claude/settings.local.json
+++ b/examples/medplum-provider/.claude/settings.local.json
@@ -1,5 +1,0 @@
-{
-  "permissions": {
-    "allow": ["Bash(npm test:*)"]
-  }
-}


techdavidy:
@claude please review

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -0,0 +1,273 @@
...
+import classes from './ParticipantFilter.module.css';
+import { showErrorNotification } from '../../utils/notifications';
+
+interface Participant {
+  reference: string;
+  display: string;
+}
+
+interface ParticipantFilterProps {
+  selectedParticipantRefs: string[];
```
noahsilas: I think this has the potential to get confusing, as "ref" is used as a type within React as well as a shorthand for the FHIR "Reference" type. It's a bit wordier, but how would you feel about calling this `selectedParticipantReferences` to help disambiguate?

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -0,0 +1,273 @@
...
+import { IconUsers } from '@tabler/icons-react';
+import { useEffect, useMemo, useState } from 'react';
+import type { JSX } from 'react';
+import classes from './ParticipantFilter.module.css';
+import { showErrorNotification } from '../../utils/notifications';
+
+interface Participant {
+  reference: string;
+  display: string;
+}
```
noahsilas: This looks like it's a slice of the `Reference` type we are pulling in from "@medplum/fhirtypes". I wonder if we can reuse that instead of creating a new interface?

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -0,0 +1,273 @@
...
+        }
+        try {
+          const [resourceType, id] = ref.split('/');
+          if (resourceType && id) {
+            const resource = await medplum.readResource(resourceType as 'Patient' | 'Practitioner', id);
+            const name = 'name' in resource && resource.name?.[0] ? formatHumanName(resource.name[0]) : undefined;
+            resolved.push({
+              reference: ref,
+              display: name ?? ref,
+            });
```
noahsilas: This sort of looks like it is recreating logic that already exists in core's `createReference` helper? In that method, if the resource type is `Patient` or `Practitioner` it will use this same method of formatting the first `name` entry as the display component. So maybe this could look like:


```suggestion
            const resource = await medplum.readResource(resourceType as 'Patient' | 'Practitioner', id);
            resolved.push(createReference(resource));
```

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+      const resolved: Participant[] = [];
+      for (const ref of selectedParticipantRefs) {
+        if (currentUserParticipant?.reference === ref) {
+          continue;
+        }
+        try {
+          const [resourceType, id] = ref.split('/');
+          if (resourceType && id) {
+            const resource = await medplum.readResource(resourceType as 'Patient' | 'Practitioner', id);
```
noahsilas: Doing an `await` inside this for loop will cause this to read each of these in a waterfall style. If you rewrite this as a `Promise.all(selectedParticipantRefs.map((ref) => { ... })` then they can be emitted in parallel (and auto-batched by the medplum client).

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const debouncedSearch = useDebouncedCallback(async (query: string): Promise<void> => {
+    if (!query.trim()) {
+      setSearchResults([]);
+      return;
+    }
+
+    setIsSearching(true);
+    try {
+      const bundle = await medplum.search('Patient', {
+        _type: 'Patient,Practitioner',
```
noahsilas, line 54: Does this work? It seems sort of surprising that you can make a request asking to search `Patient` resources and override the type. At the very least this seems like it will confuse the type system (probably why there is so much casting to `Patient | Practitioner` below).

In the Medplum client it looks like there is one place where we do a multi-type search like this by specifying the empty string as the resource type: https://github.com/medplum/medplum/blob/v5.0.10/packages/core/src/client.ts#L1895. That looks like it would construct a URL without a `:resourceType` url parameter and hit a route that specifically does search multiple types (https://github.com/medplum/medplum/blob/v5.0.10/packages/fhir-router/src/fhirrouter.ts#L311).

It's possible also that the nested route inspects this `_type` field and does the right thing under the hood though - I haven't dug that deep yet.

Aside: It seems like we should consider adding more explicit support for multi-type searching to the client API. That's probably a good follow up for later though.

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+function ParticipantItem(props: ParticipantItemProps): JSX.Element {
+  const { participant, isSelected, isCurrentUser, onToggle, onRemove } = props;
+
+  return (
+    <Group justify="space-between" wrap="nowrap" className={classes.participantItem}>
+      <Group gap="sm" wrap="nowrap" style={{ flex: 1, minWidth: 0 }}>
+        <Checkbox checked={isSelected} onChange={onToggle} />
+        <ResourceAvatar
+          value={{ reference: participant.reference } as Reference<Patient | Practitioner>}
```
noahsilas: It looks to me like your `Participant` type is assignable as a `Reference`, so I think you can simplify this and remove the cast:
```suggestion
          value={participant}
```

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
techdavidy: Will file an issue to have explicit support.

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
```diff
@@ -12,74 +12,35 @@ import type { JSX } from 'react';
...
-        }
-      }
-      setAdditionalParticipants(resolved);
-    };
-
-    resolveParticipants().catch(console.error);
-  }, [selectedParticipantRefs, medplum, currentUserParticipant]);
+    const currentUserRef = currentUserParticipant?.reference;
+    const filtered = selectedParticipants.filter((p) => p.reference !== currentUserRef);
+    setAdditionalParticipants(filtered);
```
noahsilas, line 42: This seems like it could be a piece of derived state instead of in a `useEffect` now that it isn't doing any data fetching 😁

examples/medplum-provider/src/components/messages/ParticipantFilter.tsx
noahsilas: I tested from the CLI and it looks like this is able to return a bundle with mixed results of Patients and Practitioners. So while I do think we should follow up later on making this follow more of a happy path, I don't think that needs to block moving forward here.

To test this I ran `medplum get 'Patient?_type=Patient,Practitioner&name=noah'` and saw that I got both types of resources in the resulting bundle 👍

examples/medplum-provider/src/components/messages/ThreadInbox.tsx
```diff
@@ -105,11 +110,12 @@ export function ThreadInbox(props: ThreadInboxProps): JSX.Element {
   });
 
   const handleParticipantsChange = useCallback(
-    (participantRefs: string[]) => {
+    (participants: Reference<Patient | Practitioner>[]) => {
       // Remove existing recipient filters
       const otherFilters = currentSearch.filters?.filter((f) => f.code !== 'recipient') ?? [];
 
       // Add recipient filter with comma-separated values (OR logic in FHIR)
+      const participantRefs = participants.map((p) => p.reference).filter(Boolean) as string[];
```
noahsilas, line 118: I have a draft of a simple utility for use in this kind of filter expression that can do the type narrowing for you so you don't have to use `as string[]` (https://github.com/medplum/medplum/pull/7967/changes#diff-2a9ad102bf52756209d9f898fe39a60521f678f9cbd71aa2c65a55e74361c65cR9).

I had been planning on putting that into `packages/server`, but maybe it makes sense to pull that into `packages/core` so we can easily use it in places like provider too 🤔 

No blockers here, this is more a note for me to think about the best way to keep moving the TS utils forward.

PR #8178, opened 2026-01-09
Provider messages - include only Communication that are topics/partOf of another Communication

techdavidy:
Using Reverse Chained Search
https://www.medplum.com/docs/search/chained-search

Display Communications that have another Communications referencing the parent in `partOf`.

### DIFF
--- a/examples/medplum-provider/src/hooks/useThreadInbox.test.tsx
+++ b/examples/medplum-provider/src/hooks/useThreadInbox.test.tsx
@@ -376,5 +376,4 @@ describe('useThreadInbox', () => {
       expect(result.current.selectedThread).toBeUndefined();
     });
   });
-
 });


techdavidy:
@claude please review

examples/medplum-provider/src/hooks/useThreadInbox.test.tsx
```diff
@@ -58,11 +49,6 @@ describe('useThreadInbox', () => {
   });
 
   test('fetches thread messages and returns only one message per topic', async () => {
-    await medplum.createResource(mockCommunication1);
-    await medplum.createResource(mockCommunication2);
-    await medplum.createResource(mockCommunication3);
```
noahsilas, line 63: I notice that the strategy for the tests is changing from using the MemoryRepository to mocking search results. I'm guessing that the MemoryRepo doesn't support the reverse chaining searches? Nice pragmatic solution 👍

examples/medplum-provider/src/hooks/useThreadInbox.ts
```diff
@@ -44,6 +44,7 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
     const searchParams = new URLSearchParams(query);
     searchParams.append('identifier:not', 'ai-message-topic');
     searchParams.append('part-of:missing', 'true');
+    searchParams.append('_has:Communication:part-of:_id:not', 'null');
```
noahsilas, line 47: Can you help me understand the intention here a little bit?

It seems like this means that we only find Communication resources that have another Communication pointing to them via the "part-of" field. Do we expect there to never be a top-level message without child nodes? Or is this something like focusing on the "inbox" component here meaning "communications with replies", and we expect replies to have "part-of" pointing back to the message they are replying to?

techdavidy: Correct we are looking for Communication that a have a "child" Communication pointing back to "parent/topic/thread" via "part-of". In the query we are interested in "parent/topic/thread" to populate the inbox.

PR #8191, opened 2026-01-09
Disable MedplumClient caching with on-behalf-of

codyebberson:
This PR disables client-side response caching whenever the `on-behalf-of` header is present.

The MedplumClient caching layer was originally designed around a **single effective principal** per client instance. When `on-behalf-of` is used, that assumption no longer holds, and naive response caching can produce incorrect or surprising behavior if enabled.

While server-side authorization is always enforced correctly, this change makes the SDK more defensive by preventing cached responses from being reused across different effective principals.

**Rationale**

* `on-behalf-of` introduces multiple effective security contexts on a single client
* Client-side caching is not principal-aware
* Even though caching may be disabled by default in some environments, it is easy for applications to enable it without realizing the interaction
* Disabling caching in this case enforces a safer invariant and removes a sharp edge for developers

**Behavior change**

* If an `on-behalf-of` header is present, MedplumClient caching is automatically disabled
* This applies regardless of configured cache time
* No server-side behavior or authorization logic is affected

**Notes**

This is a defensive SDK hardening change intended to improve developer ergonomics and reduce the risk of incorrect client behavior. No server-side security issue or data isolation problem is involved.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -3382,13 +3382,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
   private isCacheEnabled(
     options: MedplumRequestOptions | undefined
   ): this is this & { requestCache: LRUCache<RequestCacheEntry> } {
-    return !!(
-      this.requestCache &&
-      options?.cache !== 'no-cache' &&
-      options?.cache !== 'reload' &&
-      !this.defaultHeaders?.['on-behalf-of'] &&
-      !this.getRequestHeader(options, 'on-behalf-of')
-    );
+    return !!this.requestCache && !this.getRequestHeader(options, 'x-medplum-on-behalf-of');
   }
 
   /**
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

techdavidy (approved):

ThatOneBro (approved):

packages/core/src/client.ts
```diff
@@ -3394,9 +3412,10 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    * Adds a readable promise to the cache.
    * @param key - The cache key to store.
    * @param value - The readable promise to store.
+   * @param options - Optional fetch options for cache settings.
    */
-  private setCacheEntry(key: string, value: ReadablePromise<any>): void {
-    if (this.requestCache) {
+  private setCacheEntry(key: string, value: ReadablePromise<any>, options: MedplumRequestOptions | undefined): void {
+    if (this.isCacheEnabled(options)) {
```
noahsilas, line 3412: In the scenario with `options.cache === 'reload'`, prior to this change that result would have been set into the cache, but now I think it's going to behave the same as `no-cache`. Maybe we should also pass an operation type into `isCacheEnabled` so it can decide to behave differently for gets vs sets? Or maybe we handle `reload` in an alternate way - we could clear the entry from the cache and then drop that option. Or maybe we just deprecate that strategy entirely?

codyebberson: Ahh, good catch!  you're right, that is a regression.  Will fix.  Thanks @noahsilas !

PR #8192, opened 2026-01-12
Dependency upgrades 2026-01-12

medplumbot:
Dependency upgrades


mattwiller:
[Some Express types changed](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74161), which necessitated a few minor edits in `@medplum/server`

PR #8205, opened 2026-01-13
React components for patient preferred pharmacy selection

oleg-mp:
This adds functionality to Provider App that allows searching and selecting preferred pharmacy for a patient. This relies on the DoseSpot integration for the search functionality

### DIFF
--- a/packages/react/src/PatientSummary/PatientSummary.tsx
+++ b/packages/react/src/PatientSummary/PatientSummary.tsx
@@ -59,7 +59,7 @@ export interface PatientSummaryProps {
   readonly patient: Patient | Reference<Patient>;
   readonly onClickResource?: (resource: Resource) => void;
   readonly onRequestLabs?: () => void;
-  readonly PharmacyDialogComponent?: ComponentType<PharmacyDialogBaseProps>;
+  readonly pharmacyDialogComponent?: ComponentType<PharmacyDialogBaseProps>;
 }
 
 interface PatientMedicalData {
@@ -82,7 +82,7 @@ interface PatientMedicalData {
 
 export function PatientSummary(props: PatientSummaryProps): JSX.Element | null {
   const medplum = useMedplum();
-  const { patient: propsPatient, onClickResource, onRequestLabs, PharmacyDialogComponent } = props;
+  const { patient: propsPatient, onClickResource, onRequestLabs, pharmacyDialogComponent } = props;
   const patient = useResource(propsPatient);
   const [medicalData, setMedicalData] = useState<PatientMedicalData>();
   const [createdDate, setCreatedDate] = useState<string | undefined>();
@@ -307,7 +307,7 @@ export function PatientSummary(props: PatientSummaryProps): JSX.Element | null {
             <Pharmacies
               patient={patient}
               onClickResource={onClickResource}
-              PharmacyDialogComponent={PharmacyDialogComponent}
+              pharmacyDialogComponent={pharmacyDialogComponent}
             />
             <Divider />
           </>
--- a/packages/react/src/PatientSummary/Pharmacies.tsx
+++ b/packages/react/src/PatientSummary/Pharmacies.tsx
@@ -27,7 +27,7 @@ export interface PharmaciesProps {
   readonly patient: Patient;
   readonly pharmacies?: Organization[];
   readonly onClickResource?: (resource: Organization) => void;
-  readonly PharmacyDialogComponent?: ComponentType<PharmacyDialogBaseProps>;
+  readonly pharmacyDialogComponent?: ComponentType<PharmacyDialogBaseProps>;
 }
 
 interface PharmacyWithPrimary extends Organization {
@@ -35,7 +35,8 @@ interface PharmacyWithPrimary extends Organization {
 }
 
 export function Pharmacies(props: PharmaciesProps): JSX.Element {
-  const { patient: patientProp, onClickResource, PharmacyDialogComponent } = props;
+  const { patient: patientProp, onClickResource, pharmacyDialogComponent } = props;
+  const PharmacyDialogComponent = pharmacyDialogComponent;
   const medplum = useMedplum();
   const [opened, { open, close }] = useDisclosure(false);
   const [resolvedPharmacies, setResolvedPharmacies] = useState<PharmacyWithPrimary[]>([]);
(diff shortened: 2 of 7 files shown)


techdavidy:
@claude please review

oleg-mp:
@claude review this again

packages/dosespot-react/package.json
```diff
@@ -78,6 +78,7 @@
   "peerDependencies": {
     "@medplum/core": "5.0.12",
     "@medplum/fhirtypes": "5.0.12",
+    "@medplum/react": "5.0.12",
```
codyebberson: Is `@medplum/react` actually a dependency?  I thought it only needed `react-hooks`?

oleg-mp: It was made a dependency to avoid code duplication between `@medplum/react` and `@medplum/dosespot-react`. First there was just some code duplication, but then SonarQube complained wildly and I've refactored the code duplication away, hence the dependency

examples/medplum-scheduling-demo/data/core/example-bots.json
rahul1, line 1: I thought on main we killed this example? Regardless I would expect thsi PR doesn't touch this file

oleg-mp: Removed from the PR. Perhaps it was an artifact of me resolving merge conflicts and accidentally checking it back in

packages/dosespot-react/package.json
oleg-mp: Refactored to remove dependency

packages/react/src/PatientSummary/Pharmacies.tsx
```diff
@@ -0,0 +1,182 @@
...
+        if (!cancelled) {
+          const validResults = results.filter((r): r is PharmacyWithPrimary => r !== null);
+          setResolvedPharmacies(validResults);
+          // If all references failed to resolve, show error state
+          // If some resolved successfully, show loaded state with partial results
+          setLoadState(validResults.length === 0 && pharmacyRefs.length > 0 ? 'error' : 'loaded');
+        }
+      } catch (error) {
+        if (!cancelled) {
+          console.error('Error resolving pharmacies:', error);
```
techdavidy: please remove

techdavidy:
@claude please review

packages/react/src/PatientSummary/PharmacyDialog.tsx
```diff
@@ -0,0 +1,297 @@
...
+      });
+    } finally {
+      setAdding(false);
+    }
+  }, [selectedPharmacy, patient.id, setAsPrimary, medplum, onSubmit]);
+
+  return (
+    <Box>
+      <form
+        onSubmit={(e) => {
```
techdavidy: nit: move code to function above return

packages/react/src/PatientSummary/PharmacyDialog.tsx
```diff
@@ -0,0 +1,297 @@
...
+                const isSelected = selectedPharmacy
+                  ? getPharmacyKey(selectedPharmacy, searchResults.indexOf(selectedPharmacy)) === pharmacyKey
+                  : false;
+
+                return (
+                  <Box
+                    key={pharmacyKey}
+                    p="sm"
+                    style={{
+                      border: isSelected
```
techdavidy: can we move to module css please

packages/react/src/PatientSummary/Pharmacies.tsx
```diff
@@ -0,0 +1,182 @@
...
+  readonly patient: Patient;
+  readonly pharmacies?: Organization[];
+  readonly onClickResource?: (resource: Organization) => void;
+}
+
+interface PharmacyWithPrimary extends Organization {
+  isPrimary?: boolean;
+}
+
+type LoadState = 'loading' | 'loaded' | 'error';
```
techdavidy: maybe we can move this outside ?

packages/hl7/src/client.test.ts
rahul1, line 1: This seems out of scope for dosespot-raect

packages/dosespot-core/src/pharmacy-utils.test.ts
rahul1, line 1: We shouldn't have any vendor-specific consensus or code in Medplum core. We should take the same approach that we took with Health Gorilla, where we have a `health-gorilla-core` package that can be used by the bots and the client-side React components.

packages/dosespot-react/src/useDoseSpotPatientPharmacy.ts
```diff
@@ -0,0 +1,175 @@
...
+  /** The currently selected pharmacy Organization */
+  selectedPharmacy: Organization | undefined;
+  /** Whether to set the selected pharmacy as the patient's primary pharmacy */
+  setAsPrimary: boolean;
+}
+
+/**
+ * Return type for the useDoseSpotPatientPharmacy hook.
+ */
+export interface DoseSpotPatientPharmacyReturn {
```
rahul1, line 59: Design question: from a UI perspective, does it make sense for this one to allow you to select multiple search and then select multiple pharmacies while setting only one to be primary?   

Happy for this to be a different revision, but I think it's worth considering how I could use this hook to add,  a primary, secondary, tertiary pharmacy to a patient

packages/react/src/AppShell/Spotlight.test.tsx
rahul1, line 1: Again, I may be missing something, but this component's tests seem out of scope for the DoseSpot pharmacy selection.

packages/dosespot-react/src/useDoseSpotPatientPharmacy.ts
oleg-mp: The way DoseSpot handles it - only one is "primary" there's no strict ordering of the rest of them. No secondary, tertiary, just primary and other preferred

packages/react/src/PatientSummary/pharmacy-utils.ts
rahul1, line 1: Again, I think that nothing from DoseSpot should ever be imported into the React core library. The dependency should always flow away from the core to the vendor-specific libraries.

packages/core/src/pharmacy-utils.ts
```diff
@@ -0,0 +1,276 @@
...
+
+export interface PreferredPharmacy {
+  organizationRef: Reference<Organization>;
+  isPrimary: boolean;
+}
+
+/**
+ * Parameters for searching pharmacies.
+ */
+export interface PharmacySearchParams {
```
rahul1, line 28: Maybe I'm missing it, but I don't know where these search programs are used. If they're DoseSpot-specific, then they make sense in the DoseSpot utils, not in the core utils.

packages/react/src/PatientSummary/PatientSummary.tsx
```diff
@@ -57,6 +59,7 @@ export interface PatientSummaryProps {
   readonly patient: Patient | Reference<Patient>;
   readonly onClickResource?: (resource: Resource) => void;
   readonly onRequestLabs?: () => void;
+  readonly PharmacyDialogComponent?: ComponentType<PharmacyDialogBaseProps>;
```
rahul1: nit: for consistent. casing, start with lowercase

rahul1 (approved):
a few minor nits, otherwise good

PR #8206, opened 2026-01-13
DoseSpot Documentation: How to enroll User with Enroll Prescriber Bot

finnbergquist:


packages/docs/docs/integration/dosespot/enroll-user.md
```diff
@@ -0,0 +1,169 @@
...
+  ]
+});
+```
+
+## Important Validation Notes
+
+### NPI Validation
+
+- **The NPI must be exactly 10 digits**
+- We recommend using a valid, real NPI number for testing.
```
oleg-mp: for testing we can also recommend an online NPI generator such as `https://jsfiddle.net/alexdresko/cLNB6`
(this is the same one Stedi recommends in their doc: https://www.stedi.com/docs/healthcare/national-provider-identifier)

PR #8208, opened 2026-01-14
Clean up Sonar warnings in server package

mattwiller:


packages/server/src/auth/external.ts
```diff
@@ -143,13 +143,13 @@ export const externalCallbackHandler = async (req: Request, res: Response): Prom
   redirectUrl.searchParams.set('scope', login.scope as string);
   redirectUrl.searchParams.set('nonce', login.nonce as string);
   if (login.codeChallenge) {
-    redirectUrl.searchParams.set('code_challenge', login.codeChallenge as string);
```
noahsilas, line 146: We're going to have some overlap between this and #8210, in which I enable a linter rule to handle this particular warning across the entire codebase and prevent future introductions of such useless casts. I think these should all match pretty precisely and not cause merge conflicts, but I will ensure I wait until this merges and rebase on top of it since nearly all of the changes in my PR are automated so my resolution process will be super quick 👍

PR #8215, opened 2026-01-15
Add Sonar rules to ESLint

mattwiller:


packages/server/src/workers/download.test.ts
```diff
@@ -165,7 +166,7 @@ describe('Download Worker', () => {
       expect(queue.add).toHaveBeenCalled();
 
       (fetch as unknown as jest.Mock).mockImplementation(() => {
-        throw new Error();
+        throw new Error('=(');
```
noahsilas, line 169: ☹️ 😂 

It took me a moment to parse this as an emoticon and not an unintentional error, but I think that's fine. :+1:

noahsilas (approved):
Overall I really like this! The code changes I scanned all seem reasonable, and I'm very much :+1: on pushing more work into the linters.

My only concern is that it appears that this changed the linter runtime from ~40s to ~450s (https://github.com/medplum/medplum/actions/runs/21076137320/job/60618618680?pr=8215). But I think that will happen in parallel with the tests that take even longer (seemingly ~1000s), so this shouldn't slow down overall progress through CI, right?

PR #8218, opened 2026-01-15
Allow padding configuration to be set per resourceType

mattlong:
Previously, all resource types had to share a common configuration per search parameter. That does not allow enough flexibility since the shape (average element count) of the column is quite significant to the configuration.

Related to #7575

### DIFF
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -1291,7 +1291,7 @@ describe('FHIR Repo', () => {
         true,
       ],
       [
-        'array with resourceType in first entry resourceType array',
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/repo.test.ts
```diff
@@ -1268,6 +1268,72 @@ describe('FHIR Repo', () => {
...
+              m: 1,
+              lambda: 300,
+              statisticsTarget: 1,
+            },
+          },
+        ],
+        true,
+      ],
+      [
+        'array with resourceType in first entry resourceType array',
```
mattlong: This should say "second entry". fixing...

PR #8219, opened 2026-01-15
Display `time` properties in resources

noahsilas:
In the Scheduling line of work, we expect folks to use `Timing.repeat.timeOfDay` properties, which are of type `time`. These represent unqualified times of day like "10 AM". Use our "formatWallTime" helper to display them.


techdavidy:
@claude please review

techdavidy (approved):

PR #8222, opened 2026-01-16
VMContext SSE bot implementation 

techdavidy:
Adds Server-Sent Events (SSE) streaming support for Medplum bots running in the VM context runtime.

When a client sends Accept: text/event-stream, the server sets up SSE headers and passes the response stream to the bot handler, enabling real-time streaming responses from bots.

### Auto-Flush on First Write 
The responseStream.write() method checks res.headersSent before each write. On the first write, headers haven't been sent yet, so it sets the SSE headers and flushes them. Subsequent writes skip this since headers are already sent. 

 -> Accept: text/event-stream received                                                                  
->  Bot runs validation           
                                                                               
-> Validation fails?                                                                                           
           Return OperationOutcome                                                                                                    
           Server sends HTTP 400   │ ✅ Proper error           
                                                                                          
-> Validation passes?                                                                                                             
           Bot calls write()                                                                                            
           NOW flush headers (HTTP 200 + SSE) │ ✅ Stream starts  | No return point                                                                   
           Continue streaming

### DIFF
--- a/packages/server/src/bots/types.ts
+++ b/packages/server/src/bots/types.ts
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { WithId } from '@medplum/core';
+import type { BotResponseStream, WithId } from '@medplum/core';
 import type {
   Agent,
   Bot,
@@ -33,7 +33,7 @@ export interface BotExecutionRequest {
   /** Default headers to add to MedplumClient, such as HTTP cookies */
   readonly defaultHeaders?: Record<string, string>;
   /** Optional response stream when invoked with SSE (Server Side Events) */
-  readonly responseStream?: NodeJS.WritableStream;
+  readonly responseStream?: BotResponseStream;
 }
 
 export interface BotExecutionContext extends BotExecutionRequest {
--- a/packages/server/src/fhir/operations/execute.ts
+++ b/packages/server/src/fhir/operations/execute.ts
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { WithId } from '@medplum/core';
+import type { BotResponseStream, WithId } from '@medplum/core';
 import {
   allOk,
   badRequest,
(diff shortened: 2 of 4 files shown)


packages/server/src/bots/types.ts
```diff
@@ -34,13 +34,24 @@ export interface BotExecutionRequest {
   readonly defaultHeaders?: Record<string, string>;
 }
 
+export type StreamingCallback = (chunk: any) => Promise<void>;
```
codyebberson: Can we be more specific than `any` here?

techdavidy:
@claude please check again

codyebberson:
@claude - please review

One open design question:  Should Medplum `server` be responsible for initiating the SSE response or should the actual bot implementation be responsible?

In the current implementation, Medplum server detects the SSE request and immediately sets the response headers for an SSE response:

```ts
  const acceptsStreaming = req.header('Accept')?.includes('text/event-stream');
  let responseStream: NodeJS.WritableStream | undefined = undefined;
  if (acceptsStreaming) {
    res.setHeader('Content-Type', 'text/event-stream');
    res.setHeader('Cache-Control', 'no-cache');
    res.setHeader('Connection', 'keep-alive');
    res.flushHeaders();
    responseStream = res;
  }
```

Pros: Automatic, easier for bot authors.

Cons: Requires SSE requests to have SSE responses, no opportunity for bot authors to do validation and return HTTP 400 (for example, if the current access token does not have access to a feature, or other prerequisite check)

techdavidy:
@claude `validateBotStreamingRequest` would be optional.

techdavidy:
@claude in the latest commit we didn't go the route of providing a validation interface. We delayed flushing headers until we start writing in the stream.

packages/server/src/fhir/operations/execute.ts
```diff
@@ -83,6 +99,28 @@ async function executeOperation(req: Request): Promise<OperationOutcome | BotExe
...
+    // This allows bots to return errors with proper HTTP status codes before streaming starts
+    responseStream = {
+      write: (chunk: string): boolean => {
+        if (!res.headersSent) {
+          res.setHeader('Content-Type', 'text/event-stream');
+          res.setHeader('Cache-Control', 'no-cache');
+          res.setHeader('Connection', 'keep-alive');
+          res.flushHeaders();
+        }
+        return res.write(chunk);
```
codyebberson: Interesting.  I think we have a choice here between "implicit" and "explicit".

In my mind, I assumed we were going to change the `responseStream` type definition to be:

```ts
export interface BotResponseStream extends NodeJS.WritableStream {
  setHeader(key: string, value: string): void;
  flushHeaders(): void;
}
```

That would allow you to pass in the Express `Response` object directly.

That would give Bot authors more control (for example, if they wanted to further customize the headers.

I just took a quick look at how AWS Lambdas handle this streaming case.  Unfortunately, that `BotResponseStream` interface would not work there.

Here's an example of how to do conditional streaming from lambda:

```js
    const metadata = {
      statusCode: 200,
      headers: {
        'Content-Type': 'text/event-stream',
        'Cache-Control': 'no-cache'
      }
    };
    
    responseStream = awslambda.HttpResponseStream.from(responseStream, metadata);
```

So, in that case, there needs to be an explicit "start streaming" call, you cannot spread it across multiple `setHeader` calls.

I think that also means the API contract for VM Context bots won't work with AWS Lambda bots 😬 

What about this:

```ts
export interface BotResponseStream extends NodeJS.WritableStream {
  startStreaming(statusCode: number, headers: Record<string, string>): void;
}
```

techdavidy:
@claude we choose to explicitly flush headers and let bot author have control over headers.

PR #8225, opened 2026-01-16
Use static empty array to streamline iteration

mattwiller:


packages/ccda/src/fhir-to-ccda/entries/condition.ts
```diff
@@ -111,47 +111,43 @@ export function createProblemEntry(converter: FhirToCcdaConverter, problem: Cond
 export function createHealthConcernEntry(converter: FhirToCcdaConverter, healthConcern: Condition): CcdaEntry {
   const entryRelationship: CcdaEntryRelationship[] = [];
 
-  if (healthConcern.evidence) {
-    for (const evidence of healthConcern.evidence) {
-      if (evidence.detail) {
-        for (const detailRef of evidence.detail) {
```
noahsilas, line 117: Feels good to remove a few layers of nesting like this all over the place :grin:

packages/react-hooks/src/useQuestionnaireForm/utils.ts
```diff
@@ -463,30 +460,26 @@ function buildInitialResponseItems(
   items: QuestionnaireItem[] | undefined,
   responseItems: QuestionnaireResponseItem[] | undefined
 ): QuestionnaireResponseItem[] | undefined {
-  if (!items) {
-    return undefined;
-  }
```
noahsilas, line 468: I will say that I'm a fan of this kind of explicit early return. Making it very obvious when this function can return undefined and explicitly handling edge cases up front is kind of nice. That said, I do also like the approach you took of using the `append()` helper to coerce from `undefined` into an array type when actually processing an item.

There is a minor functional change in the places where you do this - before, passing in an empty `items` list would have returned an empty result array, but now it will return `undefined`. I think that should usually be OK, as that was part of the return type before.

mattwiller: Fair enough — this was one I weighed back and forth in my mind before just going with consistency in this case

PR #8233, opened 2026-01-17
Streamlined SSE bot model

codyebberson:
(Note, opening this PR against branch `david-sse-vmcontext-bot`.  If you compare against `main`, it might make more sense)

* Simplified things quite a bit, stripped it down closer to bare Node.js and Express types
* Removed `executeBotStreaming`, `StreamingChunk`, `StreamingCallback`, and `BotStreamingResult`
* Instead, added an optional `responseStream?: NodeJS.WritableStream` property to the `BotEvent`
* If it is an SSE request, then that `responseStream` property will be available for writing
* It is up to the Bot author to call `responseStream.write()` and `responseStream.end()`
* That is the same interface used by AWS Lambda, so this same model should work there as well
* Basically this removes several layers of abstraction, which should keep the surface area much smaller and more manageable.

### DIFF
--- a/packages/server/src/fhir/operations/execute.test.ts
+++ b/packages/server/src/fhir/operations/execute.test.ts
@@ -970,5 +970,17 @@ describe('Execute', () => {
       expect(events[0]).toStrictEqual('data: Hello ');
       expect(events[1]).toStrictEqual('data: World!');
     });
+
+    test('Streaming execution error handling', async () => {
+      const res = await request(app)
+        .post(`/fhir/R4/Bot/${bots.streamingErrorBot.id}/$execute`)
+        .set('Content-Type', ContentType.TEXT)
+        .set('Accept', 'text/event-stream')
+        .set('Authorization', 'Bearer ' + accessToken1)
+        .send('input');
+      expect(res.status).toBe(200);
+      expect(res.headers['content-type']).toBe('text/event-stream');
+      expect(res.text).toStrictEqual('');
+    });
   });
 });


codyebberson:
@claude please review

codyebberson:
@claude - Response to this point:

> Critical: Error handling for streaming bots needs improvement - users should receive error messages, not silent failures

This was an intentional choice, as it is the same as how AWS Lambda handles errors when in streaming mode.  The operating principle is that the response is effectively in an undefined state after the SSE response headers have been sent, so it could easily be an application error if you try to automatically include an error string in the response.

Instead, Bot authors (just like AWS authors) are encouraged to wrap code with a `try`/`catch` so they can return their own error value.

I hope that provides some context.  To the best of my knowledge, there is not an established or accepted pattern for how to do this.

codyebberson:
@claude - Some additional context.  This feature is not yet released.  This PR is open against a feature branch, not against `main`, as we are still iterating on core design.

> Stream lifecycle management - The enhanced documentation above addresses this concern by explicitly documenting the responseStream.end() requirement

Note that the server will close the stream if `responseStream.end()` is not called, so this is technically not a requirement, although perhaps a recommended best practice.

```ts
      if (!res.writableEnded) {
        res.end();
      }
```

> node-fetch import change - In packages/server/src/bots/vmcontext.ts:12, you switched from globalThis.fetch to importing node-fetch. Consider documenting why if it's for Node.js version compatibility.

As noted, this is a PR against a feature branch.  `main` is already using `node-fetch`, so this PR is actually restoring the `main` behavior.

> Test coverage - The removed streaming error tests (lines 996-1089 in execute.test.ts) covered scenarios like missing bot ID, bots not enabled, etc. Confirm these are now handled by non-streaming error paths since errors prevent SSE setup.

With this new implementation strategy, we inherit the existing implementation of all of those error conditions, so the extra tests should not be necessary.

techdavidy (approved):

PR #8239, opened 2026-01-19
Refactor Provider Calendar component

noahsilas:
I'm looking at adding more behaviors to this page for Scheduling workflows. Let's start by cleaning this up some.

- Extract `<Calendar />` display component. This offers an API where it can accept `Slot` and `Appointment` FHIR resources directly as props, and passes the raw resource into the appropriate callback on selection.

- Adds test coverage for that `<Calendar />` component

- Extracted some tricky Slot merging into a tested utility function.

- Move slot filtering into query (reduces useless data coming from server to the browser)

- Discard inactive query results (such as when navigating through pages quickly and queries return out of order)

- Show errors in UI notifications instead of developer console

### DIFF
--- a/examples/medplum-provider/src/components/Calendar.test.tsx
+++ b/examples/medplum-provider/src/components/Calendar.test.tsx
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { render, screen, waitFor, userEvent } from '../test-utils/render';
+import { render, screen, userEvent } from '../test-utils/render';
 import type { Appointment, Slot } from '@medplum/fhirtypes';
 import { Calendar } from './Calendar';
 import type { Range } from '../types/scheduling';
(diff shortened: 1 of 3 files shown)


techdavidy:
@claude please review

reshmakh:
@noahsilas - could I request that this also be added to Storybook?  
https://storybook.medplum.com/?path=/docs/medplum-introduction--docs

noahsilas:
> @noahsilas - could I request that this also be added to Storybook? https://storybook.medplum.com/?path=/docs/medplum-introduction--docs

I think that to add the calendar to the Storybook we would want to pull it out of `examples/medplum-provider` directory and into `packages/react`. It's not super clear to me that this is desirable - the calendar widget comes with some dependencies that would be new to that package: `dayjs` (679 kB unpacked) and `react-big-calendar` (2.6MB unpacked!!). 

It seems to me like this is appropriate for a UI app but not for the globally shared react component library. That said, maybe this would be fine if we make sure that everything is nicely tree-shakable? Or maybe we should create a separate library for this kind of heavier/more opinionated component? Maybe there's a way within Storybook to combine stories from multiple projects - maybe we could register stories from some of the Example apps in a separate category?

Any of these approaches feel out of scope for this particular refactor. Let me know what you think about some of those ideas and I'll spend some time working on a way to move it forward! I agree that it would be nice to have a single great calendar component exposing a FHIR resource based API that we can recycle in more places!

PR #8244, opened 2026-01-20
fix(ccda): Convert timezone offsets to FHIR format

amcgivern:
#### Why
- When testing the C-CDA to FHIR transform on data from [CLEAR's test patients](https://docs.clearme.com/docs/synthetic-test-patients), many resources could not be uploaded to the Medplum server due to invalid FHIR dateTime formats.
- C-CDA timestamps use timezone format `±HHMM` (e.g., `-0500`), but FHIR requires `±HH:MM` (e.g., `-05:00`).

#### What
- Adds timezone parsing to convert C-CDA format `±HHMM` to FHIR format `±HH:MM`
- Converts both `+0000` and `-0000` to `Z` for UTC (previously only handled `+0000`)
- Updates the tests to handle this format

### DIFF
--- a/packages/ccda/testdata/AllergyToEgg.json
+++ b/packages/ccda/testdata/AllergyToEgg.json
@@ -23,7 +23,7 @@
             "display": "Medplum"
           }
         ],
-        "date": "2025-01-01T00:00:00-0500",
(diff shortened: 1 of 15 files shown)


codyebberson:
Thanks @amcgivern !

Issue #7690, opened 2025-10-27
Implement $book Operation for Appointment Scheduling


rahul1:
#
## Overview

Implement the FHIR `Appointment/$book` operation for booking appointments, based on the [[IHE ITI.Scheduling specification](https://build.fhir.org/ig/IHE/ITI.Scheduling/OperationDefinition-appointment-book.html)](https://build.fhir.org/ig/IHE/ITI.Scheduling/OperationDefinition-appointment-book.html).

## Endpoints

- **`[base]/Appointment/$book`** - Book appointments

## MVP Scope

**Version 1 includes:**
- Book from `$hold` results (convert tentative to booked)
- Book from `$find` results (direct booking without hold)
- Multi-schedule atomic booking

**Out of scope for Version 1:**
- Modify existing appointments
- Cancel appointments (separate `$cancel` operation)
- Patient resource parameter
- Comment parameter

## Input Parameters

**`appointment`** (Appointment, 1..1) - *IHE Standard / Medplum Modified*
- From `$hold`: Must have `status=pending` with `slot` references
- From `$find`: Must have `status=proposed` with `participant` actors
- Must include `start`, `end`, `serviceType`, and participant information

### Scenario 1: Book from Hold (status=pending)

**Input validation:**
- Appointment has `status=pending`
- All `slot` references exist with `status=busy-tentative`

**Processing:**
Use FHIR transaction bundle to atomically:
- PATCH all slots: `busy-tentative` → `busy`
- PATCH appointment: `pending` → `booked`

```json
{
  "resourceType": "Bundle",
  "type": "transaction",
  "entry": [
    {
      "request": {
        "method": "PATCH",
        "url": "Slot/tentative-slot-1"
      },
      "resource": {
        "resourceType": "Parameters",
        "parameter": [
          {"name": "operation", "part": [
            {"name": "type", "valueCode": "replace"},
            {"name": "path", "valueString": "Slot.status"},
            {"name": "value", "valueCode": "busy"}
          ]}
        ]
      }
    },
    {
      "request": {
        "method": "PATCH",
        "url": "Appointment/held-appointment-123"
      },
      "resource": {
        "resourceType": "Parameters",
        "parameter": [
          {"name": "operation", "part": [
            {"name": "type", "valueCode": "replace"},
            {"name": "path", "valueString": "Appointment.status"},
            {"name": "value", "valueCode": "booked"}
          ]}
        ]
      }
    }
  ]
}
```

[...]

PR #8254, opened 2026-01-21
Add $find searching to Provider UI

noahsilas:
When the schedule we are viewing has the SchedulingParameters extension present, we add in a sidebar that lets the viewer invoke it to search for available slots.

This is intended to demonstrate the usage of the `$find` operation for a single schedule (https://github.com/medplum/medplum/issues/7686).

- If there are no SchedulingParameters, don't mount the sidebar at all. This keeps the experience consistent for anyone who hasn't yet set up this under-development feature.

- If there is only a single SchedulingParameters extension, invoke $find for it directly. Otherwise, show a list of service types for the viewer to choose from.

- Slots returned are rendered in both a list view in the side bar and visualized on the big calendar.

- For now, choosing a slot simply opens up the existing "Create Visit" drawer with the slot's time selected. In the future, this will instead invoke an operation like `$book` (see https://github.com/medplum/medplum/issues/7690).

- Only shows slots at least 30 minutes in the future. This is implemented in browser, so there is not strong validation here. In the future, operations like $book will be better places to attach strong rules for scheduling.

## Preview

https://github.com/user-attachments/assets/d4279943-1717-4f8c-bd02-d95bdfd58132

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
+++ b/examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
@@ -1,12 +1,12 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Box, Drawer } from '@mantine/core';
+import { Box, Button, Drawer, Group, Stack, Title } from '@mantine/core';
 import { useDisclosure } from '@mantine/hooks';
-import { createReference, getReferenceString } from '@medplum/core';
+import { createReference, isDefined, formatDateTime, getExtensionValue, getReferenceString } from '@medplum/core';
 import type { WithId } from '@medplum/core';
-import type { Appointment, Practitioner, Schedule, Slot } from '@medplum/fhirtypes';
-import { useMedplum, useMedplumProfile } from '@medplum/react';
-import { useCallback, useEffect, useState } from 'react';
+import type { Appointment, Bundle, Coding, Practitioner, Schedule, Slot } from '@medplum/fhirtypes';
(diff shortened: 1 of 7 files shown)


codyebberson:
@claude please review

packages/core/src/utils.ts
```diff
@@ -1507,5 +1507,18 @@ export function escapeHtml(unsafe: string): string {
...
+/**
+ * Helper function to narrow a type by excluding undefined/null values.
+ * @param value - The value to refine
+ * @returns boolean
+ *
+ * Example usage:
+ *   const arr: Array<number | undefined> = [1,undefined];
+ *   const refined: Array<number> = arr.filter(isDefined);
+ */
+export function isDefined<T>(value: T | undefined | null): value is T {
```
codyebberson, line 1519: I'm surprised we didn't already have a util for this in `core`, but I couldn't find anything.  Nice 👍

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -15,6 +15,136 @@ import { showErrorNotification } from '../../utils/notifications';
...
+  const serviceTypes = extensions.map((ext) => getExtensionValue(ext, 'serviceType') as Coding | undefined);
+  return serviceTypes;
+}
+
+// Allows selection of a ServiceType found in the schedule's
+// SchedulingParameters extensions, and runs a `$find` operation to look for
+// upcoming slots that can be used to book an Appointment of that type.
+//
+// See https://www.medplum.com/docs/scheduling/defining-availability for details.
+export function ScheduleFindPane(props: ScheduleFindPaneProps): JSX.Element {
```
techdavidy, line 43: ScheduleFindPanel ?

Issue #8230, opened 2026-01-16
Configurable batch size and statement timeout for ReindexJob


mattlong:
Consider if any other constants should also be configurable.

PR #8256, opened 2026-01-22
Additional ReindexJob setting configuration

mattlong:


packages/server/src/database.ts
```diff
@@ -91,11 +91,13 @@ async function initPool(config: MedplumDatabaseConfig, proxyEndpoint: string | u
   return pool;
 }
 
+const DEFAULT_STATEMENT_TIMEOUT = 60_000;
```
mattwiller, line 94: 👍

packages/server/src/workers/reindex.ts
```diff
@@ -60,8 +69,23 @@ export interface ReindexPostDeployMigration extends PostDeployMigration<ReindexJ
...
+  readonly searchStatementTimeout: number;
+  readonly upsertStatementTimeout: number | 'DEFAULT';
+  readonly delayBetweenBatches: number;
+  readonly maxIterationAttempts: number;
+}
+
+const defaultSettings: ReindexJobSettings = {
+  batchSize: 500,
+  progressLogThreshold: 50_000,
+  searchStatementTimeout: 3_600_000, // 1 hour
```
mattwiller, line 85: Is this a reasonable default?

mattlong: This was what we had been [using previously](https://github.com/medplum/medplum/pull/8256/changes#diff-960989fd4f30be910cf356dd5608abb430e8a06624e481c43eb7b00a6e84c75bL255). That really high value was chosen before adding indexes to optimize the reindex searches, so an hour is probably very much overkill now but I'm inclined to not touch it for now.

mattlong:
Fixed #8230

PR #8274, opened 2026-01-23
Update notifications in Navbar

### LINKED ISSUE #7989: [Provider] Messages should make it clear how to get rid of red bubble
There appears to be a disconnect between the red-bubble logic and the "In progress" list.

* Red bubble: `recipient=${getReferenceString(profile)}&status:not=completed&_summary=count`
* Message list: `${query}&status=in-progress` (no filter on recipient)

My vote would be that the message list should only show "my" messages (or my groups, my services, etc), but open to discussion.

Or maybe messages with "recipient=me" could have different styling (bolding, background shading, etc), so at least it would be clear which ones to read to get rid of the red bubble?

Image

kevinwadeshaw:
For Provider app, this moves the red badge notifications on the navbar from the icon area to the left side of the individual links. For Messages and Tasks, we are treating this as a number count only.

Screenshot 2026-01-23 at 12 59 55 PM

For DoseSpot, we are still showing a red badge on the left side of the navbar link. When navbar is collapsed, a dot indicator is still shown on the icon.

https://github.com/user-attachments/assets/c7412553-e580-4ade-ba72-4567f65ddb14


techdavidy (changes requested):
Thanks,
Will do some clean up and merge !

techdavidy (approved):

mattlong (approved):

techdavidy:
@codyebberson as discussed during 1-1. 
Adding the logic to the Icons in this new layout is not optimal. 
@codyebberson To move this logic to NavbarLink and handle collapsed/expanded state for notifications (red dot) and inbox count.

kevinwadeshaw:
PSA: these changes were included and merged with https://github.com/medplum/medplum/pull/8427!

PR #8279, opened 2026-01-24
fix(agent): `pushToAgent` returns on first ACK by default

### LINKED ISSUE #8278: [Agent] Revert breaking change to `pushToAgent` return ACK behavior
In #7794, we made the `Hl7Client` return on application-level ACK by default since we believed this was more aligned with most use-cases for the `Hl7Client`, and probably for most use cases with `pushToAgent`. However, we realized some users are actually relying on the old behavior as they are sending only `CA`s without ever returning an application-level ACK, which causes `pushToAgent` to hang on newer versions of the agent.

Since this was technically a breaking change that occurred after a major version bump and not with it, and because this change was not broadcasted in advance, it seems reasonable to revert the change in behavior at the agent level, by making the `ReturnAckCategory.FIRST` the default value passed to `Hl7Client#sendAndWait` with the `pushMessage` method in the agent, and allow for overrides at both the `Device` level (via URL params) and also on a per-message basis via a new optional operation parameter, and `MedplumClient#pushToAgent` option.

ThatOneBro:
Fixes #8278

### DIFF
--- a/packages/agent/src/app.test.ts
+++ b/packages/agent/src/app.test.ts
@@ -4551,8 +4551,7 @@ describe('App', () => {
             contentType: ContentType.HL7_V2,
             body: hl7MessageBody,
             callback: 'test-callback',
-            // @ts-ignore This is for testing
-            returnAck: 'invalid_value', // Invalid per-message returnAck
+            returnAck: 'invalid_value' as 'first', // Invalid per-message returnAck
           } satisfies AgentTransmitRequest)
         )
       );


codyebberson:
@claude  - please review

I think the logic and overall structure is sound.  One minor nitpick question is the naming of `ReturnAckCategory`.  This type represents the behavior for a `sendAndWait()` call.  Should it return on "first ACK", or "application ACK"?  Would something like `ReturnAckBehavior` be a more self-explanatory name?  or something else?

ThatOneBro:
The naming already was exported in the existing `@medplum/hl7` package, I just moved where it was exported from. We can change the name but it would be a breaking change for an exported enum 👀 

EDIT for context: Initially I think it made sense since the ACK types are broadly in two "categories" - application and commit. So `APPLICATION` and `COMMIT` were the original choices, but then it made it confusing for cases where there was no commit-level ACK, so I changed it to `APPLICATION` and `ANY` (any category was valid to return on, basically take the first). But then realized that was pretty confusing and changed `ANY` to `FIRST`... and here we are 😄

PR #8283, opened 2026-01-26
Array padding calculator

mattlong:
This is a lower-level calculator that requires a lot of context to use correctly. I have big dreams for this eventually being parts of higher-level abstractions that'll automatically spit out recommended settings, but for now I just wanted to commit the math before I have to build up the mental model in my head again...

Screenshot 2026-01-26 at 5 29 44 PM


techdavidy:
@claude please review

PR #8289, opened 2026-01-27
Rename patient-set-accounts operation to set-accounts

deam65:
Rename operation and update all references

- Rename patient-set-accounts.md to set-accounts.md (create latter and delete former - added some use cases here)
- Update references in multi-tenant-access-policy.md
- Update references in 2025-04-22-multi-tenant-mso.md blog post

reason for this is change is that the `$set-accounts` operation applies to all resources


finnbergquist:
Nice! Originally, you could only all it for a Patient actually :) Good change

PR #8300, opened 2026-01-27
Appointment $book operation

noahsilas:
First implementation for https://github.com/medplum/medplum/issues/7690.

Introduces a new operation, `POST /Appointment/$book`. This can be invoked with a slot from `$find` to book an appointment and add slots to block the calendar.

Not yet implemented:
- accepting an `Appointment` as a resource instead of `Slot`(s).
- accepting slots that have been created from `$hold` style operation

### DIFF
--- a/packages/server/src/fhir/operations/book.ts
+++ b/packages/server/src/fhir/operations/book.ts
@@ -14,6 +14,7 @@ import type { FhirRequest, FhirResponse } from '@medplum/fhir-router';
 import type { Bundle, OperationDefinition, Slot } from '@medplum/fhirtypes';
 import { getAuthenticatedContext } from '../../context';
 import { addMinutes, areIntervalsOverlapping } from '../../util/date';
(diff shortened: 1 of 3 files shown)


codyebberson:
@claude please review

noahsilas:
> There's the one question about potential race condition. I don't have intuition for how risky that is. If we're ok with a calculated risk for now, i'd be good to approve and move forward. If we want to play it safe + can tolerate the DB perf implications, then you could wrap with `repo.withTransaction(() => { ... })`

This is expected to be relatively low volume in usage but with a high expectation for correctness (double booking is bad!). The searches and writes are already wrapped in a serializable transaction (https://github.com/medplum/medplum/pull/8300/changes#diff-3909d8fa03a54fafebd2644e5db58089bb6ff7e4150b38b72e2f89a495983883R187) which should give us some good safety here 😄

PR #8301, opened 2026-01-27
Add nested connection to graphql

maddyli:
[Connection](https://hl7.org/fhir/graphql.html#searching) and [Reverse](https://www.hl7.org/fhir/graphql.html#reverse) are both defined in the FHIR GraphQL spec, but not explicitly defined or excluded in combination (connection on a reverse lookup). This change is somewhat lightweight in that we are using functions and logic already defined in the base spec, but technically this is an extension of the spec as it exists.

### DIFF
--- a/packages/fhir-router/src/graphql/graphql.ts
+++ b/packages/fhir-router/src/graphql/graphql.ts
@@ -293,7 +293,10 @@ function buildUpdateArgs(resourceType: string): GraphQLFieldConfigArgumentMap {
   return args;
 }
 
-export function getConnectionType(resourceType: ResourceType, resourceGraphQLType: GraphQLOutputType): GraphQLOutputType {
+export function getConnectionType(
+  resourceType: ResourceType,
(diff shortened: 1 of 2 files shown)


packages/fhir-router/src/graphql/graphql.ts
```diff
@@ -293,6 +293,19 @@ function buildUpdateArgs(resourceType: string): GraphQLFieldConfigArgumentMap {
...
 
+export function getConnectionType(
+  resourceType: ResourceType,
+  resourceGraphQLType: GraphQLOutputType
+): GraphQLOutputType {
+  const cacheKey = resourceType + 'Connection';
+  let result = outputTypeCache[cacheKey];
+  if (!result) {
+    result = buildConnectionType(resourceType, resourceGraphQLType);
+    outputTypeCache[cacheKey] = result;
```
mattwiller, line 304: 👍

mattwiller (approved):

PR #8304, opened 2026-01-27
Provider Spaces  - streaming and buffered support 

techdavidy:
- Added streaming state management and UI updates for real-time AI responses 
- Added sendToBotStreaming() function to handle SSE responses from bots, plus handling for both streaming and buffered responses (the change we just made)               
- ChatInput.tsx: Minor updates to support streaming flow                          

https://github.com/user-attachments/assets/92268e30-d18f-4375-abee-f62d8f91a562

### DIFF
--- a/examples/medplum-provider/src/utils/spaceMessaging.ts
+++ b/examples/medplum-provider/src/utils/spaceMessaging.ts
@@ -336,7 +336,13 @@ export async function processMessage(params: ProcessMessageParams): Promise<Proc
 
     // Get summary response after tool execution (streaming if callback provided)
     if (onStreamChunk) {
-      content = await sendToBotStreaming(medplum, resourceSummaryBotSseId, currentMessages, selectedModel, onStreamChunk);
+      content = await sendToBotStreaming(
+        medplum,
+        resourceSummaryBotSseId,
+        currentMessages,
+        selectedModel,
+        onStreamChunk
+      );
     } else {
       const summaryResponse = await sendToBot(medplum, resourceSummaryBotId, currentMessages, selectedModel);
       content = summaryResponse.content;


examples/medplum-provider/src/utils/spaceMessaging.ts
```diff
@@ -149,6 +154,103 @@ export async function sendToBot(
...
+  botId: Identifier,
+  messages: Message[],
+  model: string,
+  onChunk: (chunk: string) => void
+): Promise<string> {
+  // Find the bot by identifier
+  const bot = await medplum.searchOne('Bot', { identifier: `${botId.system}|${botId.value}` });
+  if (!bot?.id) {
+    throw new Error(`Bot not found: ${botId.value}`);
+  }
```
noahsilas, line 168: Based on `client.executeBot` It looks like this request can be folded into the next one to save a little bit of network time:

https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts#L2745-L2747

examples/medplum-provider/src/utils/spaceMessaging.ts
```diff
@@ -149,6 +154,103 @@ export async function sendToBot(
...
+
+  const response = await fetch(medplum.fhirUrl('Bot', bot.id, '$execute').toString(), {
+    method: 'POST',
```
noahsilas, line 171: It's slightly bothersome that we need to `POST` here, and so can't use builtin SSE handlers like [`EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource). I don't think it makes sense to try to force this into a `GET` though, so I guess we're left reimplementing ourselves. 😞

examples/medplum-provider/src/utils/spaceMessaging.ts
```diff
@@ -149,6 +154,103 @@ export async function sendToBot(
...
+  if (!response.body) {
+    throw new Error('No response body');
+  }
+
+  const reader = response.body.pipeThrough(new TextDecoderStream()).getReader();
+  let fullContent = '';
+  let buffer = '';
+
+  while (true) {
+    const { done, value } = await reader.read();
```
noahsilas, line 215: Not a blocker: it feels like this could be decomposed into a couple streams that could each be nicely testable as a standalone unit:

```typescript
response.body
  .pipeThrough(new TextDecoderStream())
  .pipeThrough(new LineBufferingStream()) // emit full lines of text as a chunk to the next stream
  .pipeTo(new SSEHandlingStream(onChunk)); // Handle JSON parsing and emitting to onChunk handler
```

examples/medplum-provider/src/utils/spaceMessaging.ts
```diff
@@ -232,9 +336,16 @@ export async function processMessage(params: ProcessMessageParams): Promise<Proc
...
     }
 
-    // Get summary response after tool execution
-    const summaryResponse = await sendToBot(medplum, resourceSummaryBotId, currentMessages, selectedModel);
-    content = summaryResponse.content;
+    // Get summary response after tool execution (streaming if callback provided)
+    if (onStreamChunk) {
+      const uniqueRefs = resourceRefs.length > 0 ? [...new Set(resourceRefs)] : undefined;
+      content = await sendToBotStreaming(medplum, resourceSummaryBotSseId, currentMessages, selectedModel, (chunk) => {
+        onStreamChunk(chunk, uniqueRefs);
```
noahsilas: It looks like nothing is reading the second argument from here, we could potentially simplify by removing it.

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
```diff
@@ -127,13 +144,19 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
         setRefreshKey,
         setCurrentFhirRequest,
         onNewTopic,
+        onStreamChunk: (content) => {
+          setStreamingContent(content);
```
noahsilas: Should this append the new chunk to any streaming content we already got?

```suggestion
          setStreamingContent((existingState) => existingState ?? '' + content);
```

examples/medplum-provider/src/utils/spaceMessaging.test.ts
```diff
@@ -0,0 +1,286 @@
...
+  });
+
+  test('handles SSE with empty data lines', async () => {
+    const encoder = new TextEncoder();
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(encoder.encode('data: {"content":"Hello"}\n\n'));
+        controller.enqueue(encoder.encode('\n\n'));
+        controller.enqueue(encoder.encode('data: \n\n'));
+        controller.enqueue(encoder.encode('data: {"content":" world"}\n\n'));
```
noahsilas, line 243: Does this all work if there is a newline inside the JSON content? Example:

```typescript
        controller.enqueue(encoder.encode('data: {"content":"Stop.\\n Collaborate.\\n Listen.\\n"}\n\n'));
```

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
```diff
@@ -31,13 +31,14 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
   const [messages, setMessages] = useState<Message[]>([]);
   const [input, setInput] = useState('');
   const [loading, setLoading] = useState(false);
-  const [selectedModel, setSelectedModel] = useState('gpt-5');
+  const [selectedModel, setSelectedModel] = useState('gpt-4o-mini');
```
noahsilas, line 34: I think I heard you say that changing to the "mini" model makes a big difference in how fast this can get resolved - is that why you are changing this default state?

examples/medplum-provider/src/utils/spaceMessaging.ts
techdavidy: great idea

examples/medplum-provider/src/utils/spaceMessaging.ts
```diff
@@ -149,6 +154,101 @@ export async function sendToBot(
...
+  let buffer = '';
+
+  while (true) {
+    const { done, value } = await reader.read();
+    if (done) {
+      break;
+    }
+
+    buffer += value;
+    const lines = buffer.split('\n');
```
noahsilas, line 221: It's not fully obvious that this works when there are newlines embedded in the JSON documents, but after some brief testing it looks like it does behave well. It looks like the spec requires that split act on code units, and so the escaped backslash is being treated as a distinct code unit and not matching the way a naive substring find would. Nice 👍

noahsilas (approved):

examples/medplum-provider/src/utils/spaceMessaging.ts
techdavidy: At the moment executeBot is not using fetch for streaming. We will need to add that additional functionality.

PR #8315, opened 2026-01-28
Self-hosting best practices docs

codyebberson:


packages/docs/docs/self-hosting/best-practices.md
```diff
@@ -0,0 +1,194 @@
...
+
+These recommendations come from patterns we've seen across self-hosted deployments, both what works well and what causes pain down the road.
+
+## Infrastructure Organization
+
+### Use a Dedicated Cloud Account
+
+Deploy your Medplum infrastructure in a dedicated cloud account (e.g., an AWS child account under your organization). This provides several benefits:
+
+- **Cleaner blast radius**: Medplum resources are isolated from your other workloads
```
mattwiller:
```suggestion
- **Smaller blast radius**: Medplum resources are isolated from your other workloads
```

packages/docs/docs/self-hosting/best-practices.md
```diff
@@ -0,0 +1,194 @@
...
+
+We strongly recommend treating direct database manipulation as a last resort, not a routine operational tool.
+
+### Plan for PostgreSQL Major Version Upgrades
+
+PostgreSQL major version upgrades (e.g., 14 → 15 → 16) require manual intervention and planning. AWS does not automatically perform major version upgrades for RDS.
+
+- **Schedule regular upgrades**: Don't let your PostgreSQL version fall too far behind. Running an unsupported version limits your options during incidents.
+- **Test in staging first**: Major version upgrades can surface compatibility issues. Always test your upgrade path in a non-production environment.
+- **Budget the time**: A major version upgrade, including preparation, testing, and execution, typically requires several days of focused effort.
```
mattwiller, line 53: Should we add an item about staying relatively close to the current version?  Postgres is very stable software and we often try to leverage the great work they're doing with performance etc., so it might be useful to recommend that folks try and keep up with the latest and greatest

PR #8317, opened 2026-01-28
Add `client.post<T>` generic type annotation

noahsilas:
This lets users of this method specify what they expect the server response type to be in the same way as `client.get<T>`. This helps us remove more `any` types that opt the user out of all type safety.

### DIFF
--- a/examples/medplum-healthie-importer/scripts/deploy-importer.ts
+++ b/examples/medplum-healthie-importer/scripts/deploy-importer.ts
@@ -85,7 +85,7 @@ async function createOrUpdateBot(medplum: MedplumClient): Promise<WithId<Bot>> {
     console.log('No existing connector found. Creating...');
     // Create the bot resource in the project
     existing = await medplum
-      .post<any>('admin/projects/' + projectId + '/bot', {
+      .post('admin/projects/' + projectId + '/bot', {
         name: IMPORTER_BOT.name,
         description: IMPORTER_BOT.description,
         sourceCode,


examples/medplum-healthie-importer/scripts/deploy-importer.ts
```diff
@@ -85,7 +85,7 @@ async function createOrUpdateBot(medplum: MedplumClient): Promise<WithId<Bot>> {
     console.log('No existing connector found. Creating...');
     // Create the bot resource in the project
     existing = await medplum
-      .post('admin/projects/' + projectId + '/bot', {
+      .post<any>('admin/projects/' + projectId + '/bot', {
```
noahsilas: This one is sort of ugly. The response is tested below in what seems to be an attempt to detect an error condition (`response.resourceType !== 'Bot'`), which, to my understanding, _shouldn't_ happen here. Inside that condition, `response` is implicitly coerced to a string, which TypeScript doesn't love if we change from `any` to a type like `Bot`.

For now I'm punting on resolving this weirdness and using an explicit `any` annotation in a place where it used to be implicit.

packages/core/src/client.ts
```diff
@@ -1258,14 +1258,14 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    * @param options - Optional fetch options.
    * @returns Promise to the response content.
    */
-  post(url: URL | string, body?: any, contentType?: string, options: MedplumRequestOptions = {}): Promise<any> {
+  post<T>(url: URL | string, body?: any, contentType?: string, options: MedplumRequestOptions = {}): Promise<T> {
```
noahsilas: @ThatOneBro pointed out that this is a breaking change - consumers currently depending on the response type of `any` would see type errors after this merges.

Our plan is going to be to update this to be `post<T = any>`, matching `get<T = any>` above. This will let folks opt in to stronger types right away, and then when we are preparing for the next major release (v6) we can drop these defaults and force folks who want the weaker type checking to opt-in to it by explicitly specifying an `any` type parameter.

While this PR is in draft mode I'm leaving it as-is for the moment so that the build steps can flag places in our code that we are using these `any` types and clean them up.

noahsilas: Created issue to track this and hopefully remember it at major upgrade time: https://github.com/medplum/medplum/issues/8320

examples/medplum-healthie-importer/scripts/deploy-importer.ts
noahsilas: Funny enough this gets auto-corrected away as a useless cast after introducing the default `<T = any>`. Problems for future-us when we clean that up for Medplum v6?

PR #8322, opened 2026-01-29
Add unique identifiers to slots coming from $find

noahsilas:
This gives us a nice attribute to use in React as a `key` when iterating over these slots (which are not persisted and so don't have an `id` attribute set, and which _can_ have multiple results at the same time when searching with multiple service types).

In future commits we will also use this to determine which slots are coming from `$find` operations and need special treatment.

This is implemented by adding an Identifier in a custom system with `use: 'temp'`. To make that nice, I've updated the core `setIdentifier` utility to accept an optional `options` argument with the `use` parameter inside it.

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
+++ b/examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
@@ -2,7 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 import { Box, Button, Drawer, Group, Stack, Title } from '@mantine/core';
 import { useDisclosure } from '@mantine/hooks';
-import { createReference, isDefined, formatDateTime, getExtensionValue, getReferenceString } from '@medplum/core';
+import {
+  createReference,
+  EMPTY,
+  isDefined,
+  formatDateTime,
+  getExtensionValue,
+  getReferenceString,
+} from '@medplum/core';
 import type { WithId } from '@medplum/core';
 import type { Appointment, Bundle, Coding, Practitioner, Schedule, Slot } from '@medplum/fhirtypes';
 import { CodingDisplay, useMedplum, useMedplumProfile } from '@medplum/react';
@@ -114,9 +121,9 @@ export function ScheduleFindPane(props: ScheduleFindPaneProps): JSX.Element {
             )}
           </Group>
         </Title>
-        {(props.slots ?? []).map((slot) => (
+        {(props.slots ?? EMPTY).map((slot) => (
           <Button
-            key={SchedulingTransientIdentifier.get(slot) ?? slot.id ?? slot.start}
+            key={SchedulingTransientIdentifier.get(slot)}
             variant="outline"
             color="gray.3"
             styles={(theme) => ({ label: { fontWeight: 'normal', color: theme.colors.gray[9] } })}


packages/core/src/utils.ts
```diff
@@ -445,20 +445,35 @@ export function getIdentifier(resource: Resource, system: string): string | unde
...
- */
-export function setIdentifier(resource: Resource & { identifier?: Identifier[] }, system: string, value: string): void {
+ * @param options - Optional attributes to set
+ * @param options.use - IdentifierUse code (https://build.fhir.org/valueset-identifier-use.html)
+ */
+export function setIdentifier(
+  resource: Resource & { identifier?: Identifier[] },
+  system: string,
+  value: string,
+  options?: { use?: Identifier['use'] }
```
mattwiller, line 455: Could these parameters be condensed?
```suggestion
  identifier: Identifier & Required<Pick<Identifier, 'system' | 'value>>,
```

noahsilas: I think if I was designing this API from scratch I probably would do what you describe and require all these as named parameters (which would make the default usage slightly more verbose as `setIdentifier(resource, { system, value })` instead of `setIdentifier(resource, system, value)`).

We could try and get fancy with some overloads to allow calling it in both forms. Very roughly,
```
function setIdentifier(resource: Resource & { identifier?: Identifier[] }, identifier: Identifier & Required<Pick<Identifier, 'system' | 'value'>>): void;
function setIdentifier(resource: Resource & { identifier?: Identifier[] }, system: string, value: string });
function setIdentifier(...args): void {
  identifier = typeof args[1] === 'string' ? { system: args[1], value: args[2] } : args[1];
  ...
}
```

IMO this kind of thing becomes more of a pain to maintain, but it does seem totally doable. Maybe we could introduce this overload and deprecate the old format now, and drop the old style in about a year with v6?

techdavidy:
@claude please review

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -114,7 +116,7 @@ export function ScheduleFindPane(props: ScheduleFindPaneProps): JSX.Element {
         </Title>
         {(props.slots ?? []).map((slot) => (
           <Button
-            key={slot.id ?? slot.start}
+            key={SchedulingTransientIdentifier.get(slot) ?? slot.id ?? slot.start}
```
techdavidy: would SchedulingTransientIdentifier.get(slot) be enough here ?

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -114,7 +116,7 @@ export function ScheduleFindPane(props: ScheduleFindPaneProps): JSX.Element {
         </Title>
         {(props.slots ?? []).map((slot) => (
```
techdavidy: we can use EMPTY here

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
noahsilas: That's probably a good idea! While it's _possible_ that we could pass slots other than $find results in here someday (the state is being mediated by the parent), it seems unlikely to happen. And I think the worst case scenario there would be that the slots wouldn't have the transient ID, and react would get `key={undefined}` and handle it all internally in a slightly less optimal way. Nice 👍 

Fixed up in b93a769f6e50a2dad6ee633d980d95f7f639bfb4

PR #8323, opened 2026-01-29
Ensure seeding releases DB conn

mattlong:
This is a future-proofing cleanup for the seeding system based on sharding R&D where the system repo used for seeding is constructed with a PoolClient passed in the constructor; marking the connection as not being disposable.

### DIFF
--- a/packages/server/src/database.ts
+++ b/packages/server/src/database.ts
@@ -113,23 +113,35 @@ export async function closeDatabase(): Promise<void> {
 }
 
 async function runMigrations(pool: Pool): Promise<void> {
+  return withPoolClient(async (client) => {
+    let hasLock = false;
+    try {
+      hasLock = await acquireAdvisoryLock(client, locks.migration);
+      if (!hasLock) {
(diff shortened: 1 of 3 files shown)


packages/server/src/database.ts
```diff
@@ -113,23 +113,35 @@ export async function closeDatabase(): Promise<void> {
 }
 
 async function runMigrations(pool: Pool): Promise<void> {
+  return withPoolClient(async (client) => {
+    let hasLock = false;
+    try {
+      hasLock = await acquireAdvisoryLock(client, locks.migration);
+      if (!hasLock) {
+        throw new Error('Failed to acquire migration lock');
+      }
```
mattwiller, line 122: This could be slightly simplified:
```suggestion
    try {
      if (!(await acquireAdvisoryLock(client, locks.migration))) {
        throw new Error('Failed to acquire migration lock');
      }
```

Alternatively, `acquireAdvisoryLock()` should perhaps just throw itself when it fails to acquire the lock

PR #8324, opened 2026-01-29
Append tracing extensions to OperationOutcome output

mattwiller:


packages/server/src/fhir/outcomes.ts
```diff
@@ -46,11 +46,16 @@ export function sendOutcome(res: Response, outcome: OperationOutcome): Response
...
     }
   }
+  let extension: Extension[] | undefined;
+  const tracingExt = buildTracingExtension();
+  if (tracingExt) {
+    extension = append(extension, tracingExt);
+  }
+  for (const ext of outcome.extension ?? EMPTY) {
+    extension = append(extension, ext);
+  }
```
noahsilas, line 56: This is a little tricky. Double checking my understanding: We want to merge a potential tracing extension with potential existing extensions. If there are _no_ extensions from either source, we want extension to remain `undefined`. So you are starting with `extension: undefined`, and using `extension = append(extension, x)` to replace it with an array on demand, right?

Not a blocker: What do you think about a slightly more explicit approach?
```typescript
function emptyToUndefined<T>(arr: T[]): T[] | undefined {
  if (arr.length === 0) { return undefined; }
  return arr;
}

const extension = emptyToUndefined([
  ...buildTrackingExtension(),
  ...(outcome.extension ?? EMPTY)
]);
```

noahsilas (approved):

packages/server/src/fhir/outcomes.ts
mattwiller: I don't think it's a big deal either way, but my (slight) preference is to avoid the unnecessary extra allocations with the spread approach

PR #8325, opened 2026-01-29
Add patient identifier support to SMART app launch URLs

maddyli:
Changes primarily documented in the changes on `smart-app-launch.md` 

ClientApplication.launchIdentifierSystems is implemented as an array because the tokens for Smart on FHIR launches have [some proposed changes](https://build.fhir.org/ig/HL7/smart-app-launch/scopes-and-launch-context.html#launch-context-arrives-with-your-access_token). Namely, the most recent proposed version includes `patient` and `encounter` (as did the previous version) along with an option `fhirContext` field, where any FHIR resource can be passed in. As such, this design allows other resource types and identifier systems to be included, if they are relevant/needed

### DIFF
--- a/packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx
+++ b/packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx
@@ -39,7 +39,7 @@ export function SmartAppLaunchLink(props: SmartAppLaunchLinkProps): JSX.Element 
             ...patient,
             identifier: identifier,
           };
-        } 
+        }
       }
 
       // Find encounter identifier system configuration
@@ -53,7 +53,7 @@ export function SmartAppLaunchLink(props: SmartAppLaunchLinkProps): JSX.Element 
             ...encounter,
             identifier: identifier,
           };
-        } 
+        }
       }
     }
 


packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.test.tsx
```diff
@@ -1,17 +1,41 @@
...
 import { MemoryRouter } from 'react-router';
 import { act, fireEvent, render, screen, waitFor } from '../test-utils/render';
-import { SmartAppLaunchLink } from './SmartAppLaunchLink';
+import { SMART_APP_LAUNCH_PATIENT_IDENTIFIER_SYSTEM, SmartAppLaunchLink } from './SmartAppLaunchLink';
 
-const medplum = new MockClient();
+// Mock useResource to return the patient synchronously
+let mockPatientResource: Patient | undefined;
+
+jest.mock('@medplum/react-hooks', () => {
```
codyebberson: Was this necessary for something vs using `MockClient`?

If the mock patient is created in the `MockClient`, then the `useResource()` hook should use it fine.

I'm not strictly opposed to this, but we have found that using `jest.mock()` leads to more brittle tests that are more difficult to maintain.

packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx
```diff
@@ -3,11 +3,26 @@
...
+import { useMedplum, useResource } from '@medplum/react-hooks';
 import type { JSX, ReactNode } from 'react';
 
+/**
+ * Extension URL for specifying which patient identifier system to use in the SMART launch context.
+ * When this extension is present on a ClientApplication, the patient's identifier with the
+ * matching system will be included in the SmartAppLaunch resource's patient reference and
+ * returned to the SMART app in the token response.
+ */
+export const SMART_APP_LAUNCH_PATIENT_IDENTIFIER_SYSTEM =
```
codyebberson: Given the React fast refresh warning, we may want to move this into a separate file, such as `packages/core/src/constants.ts`.

packages/docs/docs/integration/smart-app-launch.md
```diff
@@ -24,6 +24,49 @@ In the example below, this application is configured to launch Inferno, the test
...
+Some SMART apps require a specific external patient identifier rather than the Medplum FHIR resource ID. For example, when integrating with external systems like Health Gorilla, you may need to pass their patient ID instead of your internal Medplum Patient ID.
+
+To configure this, add an extension to your `ClientApplication` that specifies which identifier system to use:
+
+```json
+{
+  "resourceType": "ClientApplication",
+  "name": "Health Gorilla",
+  "launchUri": "https://sandbox.healthgorilla.com/app/patient-chart/launch",
+  "extension": [
```
codyebberson: Given that `ClientApplication` is a custom Medplum type, should we make this a normal property rather than using an extension?  i.e., `launchIdentifierSystem` or whatever?

packages/server/src/oauth/token.ts
```diff
@@ -623,7 +623,9 @@ async function sendTokenResponse(res: Response, login: WithId<Login>, client?: C
 
   if (login.launch) {
     const launch = await systemRepo.readReference(login.launch);
-    patient = resolveId(launch.patient);
+    // Prefer identifier.value if present (for SMART apps that need a specific external identifier),
+    // otherwise fall back to the FHIR resource ID
+    patient = launch.patient?.identifier?.value ?? resolveId(launch.patient);
```
codyebberson, line 628: Hm.  This would apply to any identifier, not just the special case.  Is that correct?  Is that intentional?  It is a change in behavior.

We could preserve original behavior

```suggestion
    patient = launch.patient?.reference ? resolveId(launch.patient) : launch.patient?.identifier?.value;
```

codyebberson (commented):
Thanks Maddy, this looks great!

packages/server/src/oauth/token.ts
maddyli: It is intentional: 
- The patient parameter in the SmartAppLaunch resource needs to construct a token that points to the outside identifier, if it exists 
- However, we don't want to fully replace the patient.reference on SmartAppLaunch since we probably still want a way to internally reference which Medplum resource/Patient this SmartAppLaunch is for 

The codepath that creates the SmartAppLaunch will only include the `patient.identifier` as a field if the ClientApplication doing so has the extension defined above. As such, any ClientApplication without the extension will not create a SmartAppLaunch.patient.identifier, and thus SmartAppLaunch.patient.reference will be the fallback

packages/docs/docs/integration/smart-app-launch.md
maddyli: Hm definitely open to that. I think it would be better to make it tied to the the Patient and Encounter specifically, launchPatientIdentifierSystem plus launchEncounterIdentifierSystem. If no objections, I will make this change

mattwiller:
@maddyli It looks like the tests got stuck on this one: I hit "re-run all" and there appear to be some failures.  Would you mind taking a look?

packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx
```diff
@@ -18,13 +18,57 @@ export interface SmartAppLaunchLinkProps extends AnchorProps {
...
   const { client, patient, encounter, children, ...rest } = props;
+  const patientResource = useResource(patient);
+  const encounterResource = useResource(encounter);
 
   function launchApp(): void {
+    // Build the patient reference, potentially including an identifier
+    let patientRef: Reference<Patient> | undefined = patient;
+    let encounterRef: Reference<Encounter> | undefined = encounter;
+
+    if (client.launchIdentifierSystems && client.launchIdentifierSystems.length > 0) {
```
mattwiller: This form is a bit more succinct and easy to read (to me, at least):
```suggestion
    if (client.launchIdentifierSystems?.length) {
```

packages/react/src/SmartAppLaunchLink/SmartAppLaunchLink.tsx
```diff
@@ -18,13 +18,57 @@ export interface SmartAppLaunchLinkProps extends AnchorProps {
...
+        const encounterIdentifierValue = getIdentifier(encounterResource, encounterIdentifierConfig.system);
+        if (encounterIdentifierValue) {
+          // Include both the reference and the identifier in the encounter reference
+          encounterRef = {
+            ...encounter,
+            identifier: {
+              system: encounterIdentifierConfig.system,
+              value: encounterIdentifierValue,
+            },
+          };
```
mattwiller: Can this be simplified by not using the helper function?  Same for the code above with Patient
```suggestion
        const identifier = encounterResource.identifier?.find((i) => i.system === encounterIdentifierConfig.system);
        if (identifier) {
          encounterRef.identifier = identifier;
```

packages/server/src/oauth/token.test.ts
```diff
@@ -1772,6 +1772,86 @@ describe('OAuth2 Token', () => {
...
+      client_id: client.id,
+      client_secret: client.secret,
+      code_verifier: 'xyz',
+    });
+    expect(res2.status).toBe(200);
+    // When identifier is present, it should be returned instead of the FHIR resource ID
+    expect(res2.body.patient).toBe(externalPatientId);
+  });
+
+  test('Smart App Launch with encounter identifier returns identifier value', async () => {
```
mattwiller: This test and the one above are very similar: they should ideally be combined using `test.each()`: https://jestjs.io/docs/api#testeachtablename-fn-timeout

Also, it would be good to add tests for some of the interesting edge cases here: what if the Encounter has an identifier but not from that system?  What if there are two identifiers from the same system?

maddyli: Combined 

As for those edge cases, I have caught them inside of the SmartAppLaunchLink.test. token.ts already assumes the if the SmartAppLaunch.patient identifier, that identifier must be used in the token, and that SmartAppLaunch already took care of the logic of checking if the system is the correct and valid one. Is there maybe an end to end test we could do instead? I assume that would live somewhere else?

PR #8326, opened 2026-01-29
Provider $book: first UI pass

noahsilas:
This is still fairly rough, but starts demonstrating the full operation of the $find -> $book loop.

- Selecting a Slot returned from $find calls $book with it
- The resulting appointment is displayed on the big calendar
- There is some incompleteness in the full flow here: for rapid development this is focused on the scheduling components, and so currently allows creation of an under-specified appointment (no associated Encounter or other resources created). The appointment and slots are created though, allowing full demonstration of the scheduling flow.

https://github.com/user-attachments/assets/84080d83-8f3e-4a5a-a07c-8c4904568caa

### DIFF
--- a/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
+++ b/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
@@ -9,36 +9,39 @@ import type { JSX } from 'react';
 import { useCallback, useState } from 'react';
 import { showErrorNotification } from '../../utils/notifications';
 
-function UpdateAppointmentForm(props: {
+type UpdateAppointmentFormProps = {
   appointment: Appointment;
   onUpdate: (appointment: Appointment) => void;
-}): JSX.Element {
+};
+
+function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
   const medplum = useMedplum();
   const [patient, setPatient] = useState<Patient | undefined>(undefined);
 
   const { appointment, onUpdate } = props;
   const handleSubmit = useCallback(async () => {
-    if (patient) {
-      const updated = {
-        ...appointment,
-        participant: [
-          ...appointment.participant,
-          {
-            actor: createReference(patient),
-            status: 'tentative',
-          },
-        ],
-      } satisfies Appointment;
-
-      let result: Appointment;
-      try {
-        result = await medplum.updateResource(updated);
-      } catch (error) {
-        showErrorNotification(error);
-        return;
-      }
-      onUpdate?.(result);
+    if (!patient) {
+      return;
     }
+    const updated = {
+      ...appointment,
+      participant: [
+        ...appointment.participant,
+        {
+          actor: createReference(patient),
+          status: 'tentative',
+        },
+      ],
+    } satisfies Appointment;
+
+    let result: Appointment;
+    try {
+      result = await medplum.updateResource(updated);
+    } catch (error) {
+      showErrorNotification(error);
+      return;
+    }
+    onUpdate?.(result);
   }, [medplum, patient, appointment, onUpdate]);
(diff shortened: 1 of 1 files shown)


examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { Button, Group, Stack, Text } from '@mantine/core';
+import { Form, MedplumLink, ResourceAvatar, ResourceInput, useMedplum } from '@medplum/react';
+import { useResource } from '@medplum/react-hooks';
+import { createReference, formatHumanName, formatPeriod } from '@medplum/core';
+import type { Appointment, Patient, Reference } from '@medplum/fhirtypes';
+import type { JSX } from 'react';
+import { useCallback, useState } from 'react';
+import { showErrorNotification } from '../../utils/notifications';
+
+function UpdateAppointmentForm(props: {
```
techdavidy: nit: UpdateAppointmentFormProps

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -0,0 +1,88 @@
...
+  appointment: Appointment;
+  onUpdate: (appointment: Appointment) => void;
+}): JSX.Element {
+  const medplum = useMedplum();
+  const [patient, setPatient] = useState<Patient | undefined>(undefined);
+
+  const { appointment, onUpdate } = props;
+  const handleSubmit = useCallback(async () => {
+    if (patient) {
```
techdavidy: if(!patient) { return }

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -0,0 +1,88 @@
...
+      } satisfies Appointment;
+
+      let result: Appointment;
+      try {
+        result = await medplum.updateResource(updated);
+      } catch (error) {
+        showErrorNotification(error);
+        return;
+      }
+      onUpdate?.(result);
```
techdavidy: move this after the update ?
otherwise you would get an callback when error happens ?

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -245,15 +257,56 @@ export function SchedulePage(): JSX.Element | null {
...
+    },
+    [medplum, appointmentDetailsHandlers]
+  );
+
   const handleSelectSlot = useCallback(
     (slot: Slot) => {
+      // If selecting a slot from "$find", run it through "$book" to create an
+      // appointment and slots
+      if (SchedulingTransientIdentifier.get(slot)) {
+        bookSlot(slot).catch(showErrorNotification);
```
techdavidy: is there any ui feedback that an operation is happening ?

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
noahsilas: There's an early return in the error case, so the callback won't be invoked. 

I sort of like having less inside the try/catch, as an error originating inside the onUpdate callback probably could use different handling? But I don't feel strongly on this, I'd be happy to adjust if colocating this is the preferred style.

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
noahsilas: Not yet, good idea! I'll add a loading state. Done in [601347b] 👍 (https://github.com/medplum/medplum/pull/8326/commits/601347b10bda14091ac6b9bb23a85c01dab266db)

PR #8341, opened 2026-02-02
Replace `getSystemRepo` with project-oriented options

mattlong:
It has become clear that the `Repository` class will be a key player for implementing proper database routing as a Medplum's sharding strategy. To start down that path, replace `getSystemRepo` with alternatives for the various intents that system repositories are used.

Very few logical changes are included in this PR; its a future-proofing effort to facilitate the introduction of sharding.

I did not closely study every `.test.ts` file and bailed out by just changing a lot of them to use `getGlobalSystemRepo`. As the sharding project progresses, more purposeful helpers and functions will be introduced.

The crux of this PR is in `repo.ts`. I'd strongly recommend starting there and then spiraling outwards.

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -344,7 +344,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    * @returns a SystemRepository for the same shard as this repository.
    */
   getSystemRepo(conn?: PoolClient): SystemRepository {
-    return getGlobalSystemRepo(conn);
+    return createSystemRepository(this.shardId, conn);
   }
 
   setMode(mode: RepositoryMode): void {


packages/server/src/fhir/repo.ts
```diff
@@ -2817,9 +2843,18 @@ function getProfileCacheKey(projectId: string, url: string): string {
   return `Project/${projectId}/StructureDefinition/${url}`;
 }
 
-export function getSystemRepo(conn?: PoolClient): Repository {
-  return new Repository(
+export class SystemRepository extends Repository {}
```
codyebberson, line 2864: For the uninitiated, what was the reason for the new `SystemRepository` type?

codyebberson:
@claude please review.  Note that this is mostly pre-work for some upcoming changes for the new sharding architecture.

codyebberson (approved):

PR #8342, opened 2026-02-03
fix(core/subscription): reconnect if `addCriteria` called when WS closed

### LINKED ISSUE #8340: useSubscription - WebSocket does not reconnect after close (when criteria count reaches 0)
If a WebSocket is closed due to `client.unsubscribeFromCriteria` being called (e.g. when a component calling `useSubscription` is unmounted), it will not be reconnected when `client.subscribeToCriteria` is called again in future (e.g. when the component is remounted).

### Steps to reproduce
(These steps reproduce the underlying issue by unmounting and remounting the component, however, it should also be reproducable by changing the criteria or props passed to `useSubscription`)
- mount a component which calls `useSubscription`
- you should see that a subscription has been created in Redis
- trigger the subscription criteria you passed to `useSubscription`
- you should see that the callback passed to `useSubscription` is called
- unmount that component, and wait for the debounce timeout (3 secs; this comes from the useSubscription hook)
- you should see that the subscription is no longer in Redis
- remount that component
- trigger the subscription criteria you passed to `useSubscription`

#### Expected behaviour
There is a new subscription in Redis and the callback you passed to `useSubscription` is called.

#### Actual behaviour
There is no new subscription in Redis and the callback passed to `useSubscription` is not called.

### Explanation
- When `client.unsubscribeFromCriteria` is called, if the criteria count is 0 on the `SubscriptionManager`, `this.subscriptionManager.closeWebSocket()` will be called. Manually closing the WebSocket sets `ws.shouldReconnect` to `false` (reconnecting-websocket.ts `close`).
- When the component calling `useSubscription` is remounted, and `client.subscribeToCriteria` is called, `subscriptionManager.subscribeToCriteria` will return early since `this.ws.readyState` is `WebSocket.CLOSED`, and therefore will not set up the subscription.

### Potential fix
In `client.subscribeToCriteria`, if the WebSocket of the SubscriptionManager is closed, call `reconnectWebSocket` before `addCriteria`.

ThatOneBro:
Fixes #8340

### DIFF
--- a/packages/core/src/subscriptions/index.ts
+++ b/packages/core/src/subscriptions/index.ts
@@ -15,7 +15,7 @@ import type { IReconnectingWebSocket, IReconnectingWebSocketCtor } from '../webs
 import { ReconnectingWebSocket } from '../websockets/reconnecting-websocket';
 
 const DEFAULT_PING_INTERVAL_MS = 5_000;
-const WS_STATES_THAT_NEED_RECONNECT = [WebSocket.CLOSING, WebSocket.CLOSED] as const;
+const WS_STATES_THAT_NEED_RECONNECT = [WebSocket.CLOSING, WebSocket.CLOSED] as readonly number[];
 
 export type SubscriptionEventMap = {
   connect: { type: 'connect'; payload: { subscriptionId: string } };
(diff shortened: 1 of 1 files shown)


codyebberson:
@claude please review

packages/core/src/subscriptions/index.ts
```diff
@@ -489,6 +494,21 @@ export class SubscriptionManager {
     }
     return this.masterSubEmitter;
   }
+
+  async reconnectIfNeeded(): Promise<void> {
+    if (!(WS_STATES_THAT_NEED_RECONNECT as readonly number[]).includes(this.getWebSocket().readyState)) {
```
mattwiller: Should the `readonly number[]` type be hoisted up to the definition above?  Or maybe cast `this.getWebSocket().readyState` instead?

packages/core/src/subscriptions/index.ts
```diff
@@ -489,6 +494,21 @@ export class SubscriptionManager {
...
+      return;
+    }
+
+    await new Promise<void>((resolve) => {
+      const tmpCb = (): void => {
+        resolve();
+        this.getWebSocket().removeEventListener('open', tmpCb);
```
mattwiller: Should the order of these lines be swapped?

PR #8346, opened 2026-02-03
Proposed updates to David's AWS Lambda SSE PR

codyebberson:
Main changes:
* I think we need to do more testing on the non-streaming error handling code, because I think there are some breaking changes.  I tried to fix those, but we need to do real testing on AWS Lambdas to confirm.
* I tried to reorganize the changes so that the diff against `main` is more surgical, so it's more clear what changed, and importantly what didn't change.  The existing bot model is used extensively, and we need to be extremely careful to preserve all of the existing behavior, especially around error handling.

### DIFF
--- a/packages/server/src/cloud/aws/execute.ts
+++ b/packages/server/src/cloud/aws/execute.ts
@@ -43,7 +43,7 @@ export async function runInLambda(request: BotExecutionContext): Promise<BotExec
     traceId,
     headers,
     streaming: !!request.responseStream,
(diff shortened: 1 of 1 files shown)


ThatOneBro:
👀 I promised David a review last night and didn't get to it, I'll review both of these first thing this morning 👍 

Agreed it's super important to be careful with the existing code paths

techdavidy:
@claude please review

PR #8353, opened 2026-02-03
Add explicit type keyword + .d.ts extensions to fhirtypes imports for NodeNext compatibility

### LINKED ISSUE #8348: Add explicit `type` keyword + `.d.ts` extensions to @medplum/fhirtypes type re-exports for strict `NodeNext` module resolution
## Problem

When using `@medplum/fhirtypes` with TypeScript's `moduleResolution: "nodenext"`, imports fail with errors like:

```
error TS2305: Module '"@medplum/fhirtypes"' has no exported member 'ExtractResource'.
error TS2305: Module '"@medplum/fhirtypes"' has no exported member 'Resource'.
```

## Cause

The package has `"type": "module"` in `package.json`, but the type declaration files use re-exports without file extensions:

```typescript
// dist/index.d.ts
export * from './ResourceType';
export * from './Resource';
```

With `moduleResolution: "nodenext"`, TypeScript follows strict Node.js ESM resolution rules which require explicit file extensions.

## Suggested Fix

Add `.js` extensions to the re-exports in the generated `.d.ts` files:

```typescript
export * from './ResourceType.js';
export * from './Resource.js';
```

The `.js` extension works correctly because TypeScript maps `.js` imports to `.d.ts` files during type resolution.

## Workaround

Users can work around this by using `moduleResolution: "bundler"` instead of `"nodenext"`, but this may not be suitable for all projects (e.g., those publishing ESM packages for Node.js).

## Environment

- `@medplum/fhirtypes`: 5.0.6
- TypeScript: 5.4.5
- Node.js: 22.x / 24.x

ianplunkett:
## Summary

The `@medplum/fhirtypes` package uses `"type": "module"` but the generated type declarations use relative imports without file extensions. This breaks TypeScript projects using `moduleResolution: "nodenext"` which enforces strict Node.js ESM resolution rules.

This change updates the generator to add `.js` extensions to all generated imports and re-exports:
- `export * from './ResourceType.js'`
- `import { Resource } from './Resource.js'`

The `.js` extension works correctly because TypeScript maps `.js` imports to `.d.ts` files during type resolution.

## Test plan

- Regenerate fhirtypes and verify imports have `.js` extensions
- Test with a TypeScript project using `moduleResolution: "nodenext"`

Fixes #8348

🤖 Generated with [Claude Code](https://claude.ai/code)


ThatOneBro:
Thanks for starting this PR off @ianplunkett! 

As discussed offline, I found although the `.js` extension worked for whatever reason (these are only TS files and there are no generated JS files 🙃 but it works somehow! 😓 ) it also seems like adding an explicit `type` keyword + the `.d.ts` extension seems to also fix the `NodeNext` module resolution errors, and is a bit clearer on intent to export types rather than JS exports

Let's see if we can get a review on this 👀

PR #8354, opened 2026-02-03
fix(fhirtypes): explicitly export types for `NodeNext`

### LINKED ISSUE #8348: Add explicit `type` keyword + `.d.ts` extensions to @medplum/fhirtypes type re-exports for strict `NodeNext` module resolution

ThatOneBro:
Fixes #8348

### DIFF
--- a/packages/fhirtypes/package.json
+++ b/packages/fhirtypes/package.json
@@ -30,6 +30,11 @@
   "author": "Medplum <hello@medplum.com>",
   "sideEffects": false,
   "type": "module",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts"
+    }
+  },
   "main": "",
   "types": "dist/index.d.ts",
   "files": [


ThatOneBro:
Closing in favor of #8353

PR #8357, opened 2026-02-03
Add operation to refresh reference display strings

mattwiller:


ThatOneBro:
In regards to the fix to the `pathToJSONPointer`, I agree this is probably the right fix but I think we might need to make some tweaks to the auto-download worker to address those changes, since I think we had previously worked around the inclusion of the resource type name 👀 

EDIT: Looks like you already addressed this right before I posted 👍

packages/server/src/fhir/operations/refresh-reference-display.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (resource instanceof Error) {
+      continue;
+    }
+
+    const path = pathToJSONPointer(references[i].path) + '/display';
+    const value = getDisplayString(resource);
+    patch.push({ op: 'add', path, value });
+  }
+
+  const updated = await repo.patchResource(resourceType, id, patch);
```
ThatOneBro: Do we wanna test the version ID in the patch before doing all the adds?

Just because while we are building the patch, the resource could theoretically change from under us and we end up in a weird state 🤔 probably a rare edge case but might be worth protecting against

ThatOneBro (commented):
Looks good for the most part, seems pretty solid overall. One comment about testing the version ID in the patch to make sure we don't clobber updates in our transaction before I'm ready to approve 👍

packages/server/src/fhir/operations/refresh-reference-display.ts
mattwiller: Definitely — I wrapped the whole routine in a transaction to take advantage of automatic retry

ThatOneBro: Even better... I think this looks good 👍

packages/server/src/fhir/operations/refresh-reference-display.ts
```diff
@@ -0,0 +1,71 @@
...
+    const references: TypedValueWithPath[] = [];
+    for (const path of Object.keys(referenceMap)) {
+      references.push(...referenceMap[path]);
+    }
+    const resolved = await repo.readReferences(references.map((r) => r.value));
+
+    const patch: Operation[] = [];
+    for (let i = 0; i < resolved.length; i++) {
+      const resource = resolved[i];
+      if (resource instanceof Error) {
```
ThatOneBro, line 58: In what cases do you get errors here? Would this be for resources that have been deleted / user does not have access to?

I guess skipping does make sense in those cases

mattwiller: Yep, I figured it's fine to just skip given the relatively benign context

PR #8374, opened 2026-02-05
Store heavily rate limited users in memory

mattwiller:
See https://github.com/animir/node-rate-limiter-flexible/wiki/Options#inmemoryblockonconsumed

### DIFF
--- a/packages/server/src/fhir/fhirquota.ts
+++ b/packages/server/src/fhir/fhirquota.ts
@@ -92,10 +92,7 @@ export class FhirRateLimiter {
 
   private async consumeImpl(points: number): Promise<void> {
     this.delta += points;
-    const cachedResult = await this.checkInMemoryBlock(points);
-    if (cachedResult) {
-      await this.block(points, cachedResult);
-    }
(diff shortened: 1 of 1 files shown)


ThatOneBro (commented):
Interesting note:
> In-memory blocking works in current process memory and for `consume` method only.

From: https://github.com/animir/node-rate-limiter-flexible/wiki/Options#inmemoryblockonconsumed
See also: https://github.com/animir/node-rate-limiter-flexible/wiki/In-memory-Block-Strategy

But I believe this is all we are using, correct?

mattwiller:
@ThatOneBro I updated the PR to address what I think was a shortcoming of the previous approach.

> In-memory blocking works in current process memory

This is correct, but I checked the code and found that the in-memory local cache was per instance of the rate limiter, meaning it wouldn't work particularly well here, where we're constructing new instances per request.  Instead, I set up an explicit server-global cache for blocked users to ensure that a number of noisy clients can be blocked without additional Redis lookups once they hit their quota

Issue #8338, opened 2026-02-01
Login endpoint becomes extremely slow (~2 min) after sustained use due to bcrypt performance degradation


doc-krieger:
## Description

After sustained use (running ~100-150 E2E tests), the `/auth/login` endpoint becomes extremely slow, degrading from ~100ms to ~2 minutes response time. Other endpoints (healthcheck, FHIR metadata) continue responding normally.

## Environment

- **Medplum Server Version:** 5.0.13
- **Docker Image:** medplum/medplum-server:5.0.13
- **PostgreSQL:** 16
- **Redis:** 7
- **Node.js (in container):** v24.13.0
- **Platform:** Linux (Docker)

## Steps to Reproduce

1. Start Medplum server via Docker Compose
2. Run a large E2E test suite (~150+ tests) that performs many login operations
3. After ~100-150 tests, attempt to login via API

## Expected Behavior

Login should complete in ~100ms consistently.

## Actual Behavior

Login takes ~2 minutes (120,000ms) after sustained use. The issue is intermittent and sometimes recovers temporarily.

## Investigation Findings

We conducted extensive diagnostics and ruled out common causes:

| Hypothesis | Status | Evidence |
|------------|--------|----------|
| PostgreSQL connection exhaustion | ❌ Ruled out | Only 56 connections (well under limit) |
| OOM kill | ❌ Ruled out | \`OOMKilled=false\`, container using only 241MB RAM |
| Container restarts | ❌ Ruled out | \`restarts=0\`, container stays healthy |
| Redis blocked clients | ❌ Ruled out | 7 blocked clients are normal Bull queue workers |
| Healthcheck flapping | ❌ Ruled out | Healthcheck passes consistently |

### Key Observations

1. **Healthcheck passes** throughout - container reports healthy
2. **Non-auth endpoints work fine** - \`/fhir/R4/metadata\`, \`/healthcheck\` respond in ~5ms
3. **Login with invalid short password** returns instantly (~6ms) - validation fails before bcrypt
4. **Login with valid-length password** hangs (~2 min) - both correct AND incorrect passwords
5. **Server logs show "Login success"** - authentication logic completes, but HTTP response is extremely delayed
6. **Issue occurs even from inside the container** - not a network issue

### Diagnostic Commands

\`\`\`bash
# Healthcheck (always fast)
curl -s http://localhost:8103/healthcheck  # ~5ms

# Login with short password (fast - no bcrypt)
[...]

PR #8377, opened 2026-02-05
perf(server): switch from bcryptjs to native bcrypt for improved perf…

codyebberson:
…ormance

Replace the pure JavaScript bcryptjs package with the native bcrypt package to prevent event loop blocking during password hashing operations.

bcryptjs performs bcrypt operations synchronously on the main thread using setImmediate/nextTick, which blocks the event loop for ~100-300ms per hash (depending on cost factor). Under sustained authentication load, this causes cascading delays as requests queue behind blocked operations.

Native bcrypt offloads hash computation to the libuv thread pool via N-API, keeping the event loop responsive during password verification.

This addresses the performance degradation reported in #8338.

Changes:
- Replace bcryptjs@3.0.3 with bcrypt@6.0.0 in server package
- Update imports in auth/utils.ts, auth/changepassword.ts, oauth/utils.ts

The APIs are identical - this is a drop-in replacement with no code changes required beyond the import statements.

Note: bcrypt v6.0.0 includes prebuilt binaries for common platforms (Windows/Linux/macOS, x64/arm64), minimizing native compilation requirements.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1625,6 +1625,7 @@
       "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.978.0.tgz",
       "integrity": "sha512-2chs05VbfgRNb5ZEYIwooeHCaL+DjwvrW3ElkslI71ltEqVNdeWvB7hbkLWPPKazV3kjY3H90pLDY8mMqsET+A==",
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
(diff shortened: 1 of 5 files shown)


codyebberson:
FYI @doc-krieger 

Moving your branch into the `medplum/medplum` repo so that we can run all checks.  Some checks are restricted on PRs from forks for security reasons.

We've already deployed and completed manual testing on our staging environment.  This is just one last sanity pass with all of the tools and chcks.

Thank you for your contribution! 🙏

PR #8378, opened 2026-02-05
Support multiple Redis servers

mattlong:
Distinct Redis server connection details can be specified for each of four distinct use-cases:
1. Caching
2. PubSub
3. Background Jobs
4. Rate-limiting

### DIFF
--- a/packages/cdk/src/backend.ts
+++ b/packages/cdk/src/backend.ts
@@ -268,7 +268,7 @@ export class BackEnd extends Construct {
     const purposeRedisConfigs = [
       { key: 'cacheRedis' as const, id: 'CacheRedis' },
       { key: 'rateLimitRedis' as const, id: 'RateLimitRedis' },
-      { key: 'pubsubRedis' as const, id: 'PubSubRedis' },
+      { key: 'pubSubRedis' as const, id: 'PubSubRedis' },
       { key: 'backgroundJobsRedis' as const, id: 'BackgroundJobsRedis' },
     ];
 
--- a/packages/cdk/src/index.test.ts
+++ b/packages/cdk/src/index.test.ts
@@ -417,7 +417,7 @@ describe('Infra', () => {
       stackName: 'MedplumPurposeRedisStack',
       cacheRedis: { nodeType: 'cache.r6g.large' },
       rateLimitRedis: { nodeType: 'cache.t3.small' },
-      pubsubRedis: {},
+      pubSubRedis: {},
       backgroundJobsRedis: { securityGroupId: 'sg-1234' },
     });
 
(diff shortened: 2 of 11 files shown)


codyebberson:
@claude please review

codyebberson:
Missing `getBullmqRedis()` function makes sense, because BullMQ will access Redis directly - we would never interface with the Redis instance backing BullMQ, only through BullMQ itself.

packages/server/src/healthcheck.ts
```diff
@@ -37,18 +38,33 @@ export async function healthcheckHandler(_req: Request, res: Response): Promise<
...
-  setGauge('medplum.redis.healthcheckRTT', redisRoundtripMs / 1000, METRIC_IN_SECS_OPTIONS);
+  const redisChecks = getAllRedisInstances();
+  const redisResults = await Promise.all(
+    redisChecks.map(async ({ label, instance }) => {
+      const t0 = Date.now();
+      const ok = await testRedis(instance);
+      const roundtripMs = Date.now() - t0;
+      setGauge('medplum.redis.healthcheckRTT', roundtripMs / 1000, {
+        ...METRIC_IN_SECS_OPTIONS,
+        attributes: { ...METRIC_IN_SECS_OPTIONS.attributes, redisInstanceType: label },
```
ThatOneBro, line 49: 👍

ThatOneBro:
This could be a good opportunity to do non-volatile storage for bullmq Redis instance, ie. via MemoryDB 👀

mattlong:
> This could be a good opportunity to do non-volatile storage for bullmq Redis instance, ie. via MemoryDB 👀

BullMQ will likely be the one usecase that remains on the current Redis server since we don't want to transfer BullMQ state to a new server 😅

mattlong:
@claude review again considering the comment left by [codyebberson](https://github.com/codyebberson)

packages/server/src/subscriptions/websockets.ts
```diff
@@ -461,21 +458,28 @@ export async function markInMemorySubscriptionsInactive(
...
-    await redis.hdel(getActiveSubsKey(projectId, resourceType), ...refs);
+    try {
+      await removeActiveSubscriptions(projectId, resourceType, ...refs);
+    } catch {
+      globalLogger.debug('Attempted to mark subscriptions as inactive when Redis is closed');
+      return;
+    }
+  }
+  if (cacheRedis) {
+    await cacheRedis.del(refStrs);
```
ThatOneBro, line 483: I think we don't need to get cache redis above since we are no longer using it for both steps here


We can probably just change this to:
```suggestion
    await getCacheRedis()?.del(refStrs);
```

mattlong: getCacheRedis can throw though, so we'd still need the try/catch wrapper.

PR #8379, opened 2026-02-05
fix(agent): more feedback to caller, throw on 4xx during upgrade dl

ThatOneBro:
This fixes an issue where we weren't passing back some things that could have been synchronously checked in the upgrader process, passed back to the agent main process, and sent back to Medplum server as an error to provide immediate feedback to callers.


packages/agent/src/app.ts
```diff
@@ -901,12 +901,16 @@ export class App {
           () => reject(new Error('Timed out while waiting for message from child')),
           15000
         );
-        child.on('message', (msg: { type: string }) => {
+        child.on('message', (msg: { type: 'STARTED' } | { type: 'ERROR'; err: string }) => {
           clearTimeout(childTimeout);
+          assert(
```
ThatOneBro: Needs to be a reject still actually

packages/agent/src/upgrader-utils.ts
```diff
@@ -22,12 +22,15 @@ export async function downloadRelease(version: string, path: string): Promise<vo
   const downloadUrl = parseDownloadUrl(release, platform());
 
   // Write file to RELEASE_INSTALLER_FOLDER
-  const { body } = await fetch(downloadUrl);
-  if (!body) {
+  const result = await fetch(downloadUrl);
+  if (!result.ok) {
```
ThatOneBro, line 27: This is the important check that was previously missing. This should result in the upgrade failing early and returning a proper error to the client for the `$upgrade` operation

packages/agent/src/upgrader.ts
```diff
@@ -24,24 +24,35 @@ export async function upgraderMain(argv: string[]): Promise<void> {
...
+  let version: string;
+  let binPath: string;
 
-  // Make sure if version is given, it matches semver
-  if (argv[3] && !isValidMedplumSemver(argv[3])) {
-    throw new Error('Invalid version specified');
-  }
-  const version = argv[3] ?? (await fetchLatestVersionString('agent-upgrader'));
-  const binPath = getReleaseBinPath(version);
+  try {
```
ThatOneBro, line 31: This `try... catch` allows us to communicate with the parent process before the IPC closes which allows for a synchronous error response to the operation

ThatOneBro:
@claude Please review, making to sure to account for the latest changes around the IPC protocol and how we handle errors

The-Alchemist:
some strange errors in CI after merge from main:

```Error: TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Submit claim"```

https://github.com/medplum/medplum/actions/runs/25065080150/job/73430517198?pr=8379#step:7:7566

The-Alchemist (approved):

PR #8385, opened 2026-02-06
Dependency upgrades 2026-02-06

medplumbot:
Dependency upgrades


codyebberson:
Big increase in `package-lock.json` appears to be due to multiple versions of `jsdom` and cascading duplicate dependencies such as `@csstools/css-calc` and friends.

I'm going to try adding `jsdom` to `overrides`

PR #8389, opened 2026-02-06
perf(subscription): only eval subs that match criteria

### LINKED ISSUE #7882: Improve WebSocket subscription eval efficiencies
Currently we pull all the WebSocket subscriptions for a given project into the list that we evaluate against for any resource being evaluated in the `addSubscriptionJobs` function. This could be made more efficient by adding the criteria string itself to the Redis key (moving to a Redis Hash instead of a Redis Set). Which means we would only fetch the whole Subscription resources for Subscriptions that would actually match the given resource.

ThatOneBro:
Fixes #7882

### DIFF
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -14,7 +14,9 @@ import {
   isGone,
   isNotFound,
   isString,
+  matchesSearchRequest,
   normalizeOperationOutcome,
+  parseSearchRequest,
   resourceMatchesSubscriptionCriteria,
   satisfiedAccessPolicy,
   stringify,
@@ -409,8 +411,15 @@ async function getSubscriptions(resource: Resource, project: WithId<Project>): P
   const redisOnlySubRefStrs: string[] = [];
   for (const [ref, criteria] of Object.entries(entries)) {
     const criteriaResourceType = criteria.split('?')[0];
-    if (criteriaResourceType === resource.resourceType) {
-      redisOnlySubRefStrs.push(ref);
+    if (criteriaResourceType !== resource.resourceType) {
+      continue;
+    }
+    try {
+      if (matchesSearchRequest(resource, parseSearchRequest(criteria))) {
+        redisOnlySubRefStrs.push(ref);
+      }
+    } catch (err) {
+      getLogger().warn('[WS] Error while evaluating criteria for subscription', { err, subscription: ref, criteria });
     }
   }
   if (redisOnlySubRefStrs.length) {


packages/server/src/fhir/repo.ts
```diff
@@ -913,8 +913,20 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       if (!project) {
         throw new OperationOutcomeError(serverError(new Error('No project connected to the specified Subscription.')));
       }
+      // Check that the criteria is valid by attempting to parse it
+      // So we only cache subscriptions with a parsable criteria
+      try {
+        parseSearchRequest(resource.criteria);
+      } catch (err) {
+        throw new OperationOutcomeError(badRequest(`Invalid Subscription criteria: ${resource.criteria}`));
```
codyebberson: Hm.  This would be a breaking change.  Is this necessary?

packages/server/src/fhir/repo.ts
```diff
@@ -913,8 +913,20 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+      }
       // WebSocket Subscriptions are also cache-only, but also need to be added to a special cache key
-      await redis.sadd(`medplum:subscriptions:r4:project:${project}:active`, `Subscription/${resource.id}`);
+      // We added v2 to the key because historically this was a Redis Set and then we had to migrate the key to a Redis Hash which would conflict
+      await redis.hset(
```
codyebberson, line 918: I was wondering how you were planning to do this.

One thing we'll need to note is that this is an incompatible change, so any active subscriptions during deployment will silently start failing.

Alternatively, we could support heterogenous values, both legacy reference strings and also JSON blobs of `{reference, criteria}`.  The read-side would need to be aware of both potential formats.

ThatOneBro: This is not possible as previously we used Redis Sets. Redis Hash keys are not compatible with Redis Sets and will throw an error if try to issue hash commands on Sets and vice versa.

packages/server/src/fhir/repo.ts
ThatOneBro: We can avoid throwing but if we put invalid criteria into the hash it will break things no?

ThatOneBro: Ie. it has to pass `parseSearchRequest` at evaluation time anyways

packages/server/src/fhir/repo.ts
ThatOneBro: Also active subscriptions will continue to work as long as the user does not try to create new subscriptions. Upon failing, if the user refreshes, they would create a new WebSocket to a new server with access to the new path and things would work again

packages/server/src/fhir/repo.ts
ThatOneBro: Ok changing it to log and early exit

ThatOneBro: Actually didn't seem to be working as expected, just going to remove that check since we already didn't have it...

packages/server/src/workers/subscription.ts
```diff
@@ -404,8 +404,15 @@ async function getSubscriptions(resource: Resource, project: WithId<Project>): P
     ],
   });
   const redis = getRedis();
-  const setKey = `medplum:subscriptions:r4:project:${projectId}:active`;
-  const redisOnlySubRefStrs = await redis.smembers(setKey);
+  const hashKey = `medplum:subscriptions:r4:project:${projectId}:active:v2`;
```
codyebberson, line 409: 👍

packages/server/src/workers/subscription.ts
```diff
@@ -404,8 +404,15 @@ async function getSubscriptions(resource: Resource, project: WithId<Project>): P
...
+  const entries = await redis.hgetall(hashKey);
+  const redisOnlySubRefStrs: string[] = [];
+  for (const [ref, criteria] of Object.entries(entries)) {
+    const criteriaResourceType = criteria.split('?')[0];
+    if (criteriaResourceType === resource.resourceType) {
```
codyebberson: Can we evaluate full criteria here to avoid the `mget`?

ThatOneBro: I think we can't completely avoid the `mget` since you can also use the FHIRPath criteria? Also things like `meta.accounts` will only be on the full resource 🤔

ThatOneBro: We could evaluate the full criteria to fully narrow before going out to the `mget` though I think...

packages/server/src/workers/subscription.ts
```diff
@@ -404,8 +406,22 @@ async function getSubscriptions(resource: Resource, project: WithId<Project>): P
...
+  const hashKey = `medplum:subscriptions:r4:project:${projectId}:active:v2`;
+  const entries = await redis.hgetall(hashKey);
+  const redisOnlySubRefStrs: string[] = [];
+  for (const [ref, criteria] of Object.entries(entries)) {
+    const criteriaResourceType = criteria.split('?')[0];
+    if (criteriaResourceType !== resource.resourceType) {
+      continue;
+    }
+    try {
+      if (matchesSearchRequest(resource, parseSearchRequest(criteria))) {
```
ThatOneBro, line 418: Checking if this resource matches the criteria before adding it to the refs to `mget`

We could further optimize this later by using an in-memory LRU cache of references -> subscriptions, since WebSocket subscriptions will pretty much never be updated (only be added or deleted) with the current setup using the built in `SubscriptionManager` in `MedplumClient`

codyebberson:
@claude please review

PR #8394, opened 2026-02-10
fix: handle JSON Patch 'add' with '/-' when target array doesn't exist

### LINKED ISSUE #6711: JSON Patch `add` operation fails when appending to non-existent arrays in FHIR resources
## Summary
The JSON Patch `add` operation with path `/member/-` fails when trying to append members to a `Group` resource that has no existing `member` array, throwing `Error: Value required at path: /member/-`.

## Expected Behavior
According to [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902), using the `add` operation with path ending in `/-` should append to an array. When the target array doesn't exist, the implementation should handle this gracefully by creating the array and adding the first element.

## Actual Behavior
When attempting to add a member to a Group with no existing members using `/member/-`, the operation fails with:
```
Error: Value required at path: /member/-
```

## Steps to Reproduce
1. Create a Group resource with no `member` property
2. Attempt to add a member using JSON Patch with:
   ```json
   {
     "op": "add",
     "path": "/member/-",
     "value": {
       "entity": {
         "reference": "Patient/example"
       }
     }
   }
   ```
3. Operation fails with the error above

## Root Cause Analysis
This appears to be related to FHIR's constraint that empty arrays should be omitted entirely rather than represented as `[]`. This creates a conflict between:

1. **RFC 6902 JSON Patch specification** - expects target array to exist for `/-` append operations
2. **FHIR specification** - requires omitting empty arrays entirely

A Group with no members has no `member` property at all, causing the JSON Patch implementation to fail when trying to append to a non-existent array.

## Proposed Solution
Enhance the JSON Patch implementation to be more FHIR-aware by:

1. **Auto-create arrays for `/-` operations**: When using the `/-` syntax to append to a non-existent array property, automatically create the array and add the first element
2. **Maintain RFC 6902 compliance**: Ensure other JSON Patch operations continue to work as expected
3. **Consistent behavior**: Make the operation work the same whether the Group starts empty or already has members

This would eliminate the need for developers to:
- Pre-check if arrays exist before appending
- Use different patch strategies for first vs. subsequent members
- Manually initialize empty arrays (which violates FHIR spec anyway)

### Option 1: Two-step approach
```javascript
// For first member, use replace operation
await medplum.patchResource('Group', groupId, [
  {
    op: 'replace',
    path: '/member',
    value: [firstMember]
  }
]);

// For subsequent members, use add
await medplum.patchResource('Group', groupId, [
  {
    op: 'add',
    path: '/member/-',
    value: additionalMember
  }
]);
```

### Option 2: Pre-check approach
```javascript
const group = await medplum.readResource('Group', groupId);
if (!group.member || group.member.length === 0) {
  // Use replace for first member
} else {
  // Use add for additional members
}
```

## Impact
This issue affects any code that:
- Dynamically adds members to Groups
- Uses JSON Patch operations on FHIR arrays that might be empty
- Expects consistent behavior from RFC 6902 `/-` append operations

## Additional Notes
- The issue occurs specifically when the array property is completely absent (FHIR-compliant empty state)
- Setting an explicit empty array `[]` before patching doesn't resolve the issue
- The operation works correctly when adding to Groups that already have 1+ members
- This is likely to affect other FHIR resource types with similar array properties

themavik:
## Summary

Fixes #6711

**Root cause:** The `rfc6902` library's `applyPatch` throws when an `add` operation with path ending in `/-` targets a non-existent array. Per RFC 6902, `/-` means "append to the end of the array."

## Changes

- `packages/server/src/util/patch.ts`: Pre-processes patch operations to create missing parent arrays for `add` operations with `/-` paths before passing to `applyPatch`.

## Risk Assessment

**Low** - Only affects the specific case of `add` + `/-` when the array is missing. All other patch operations are unaffected.

### DIFF
--- a/packages/server/src/util/patch.ts
+++ b/packages/server/src/util/patch.ts
@@ -12,6 +12,30 @@ import { applyPatch } from 'rfc6902';
  * @param patch - The patch to apply.
  */
 export function patchObject(obj: any, patch: Operation[]): void {
+  // Pre-process: ensure parent arrays exist for 'add' operations with '/-' path suffix.
+  // Per RFC 6902, 'add' with '/-' should append to an array. When the array
+  // doesn't exist yet, we create it so the append operation can succeed.
+  for (const op of patch) {
+    if (op.op === 'add' && op.path.endsWith('/-')) {
+      const parentPath = op.path.slice(0, -2);
+      const segments = parentPath.split('/').filter(Boolean);
+      let current: any = obj;
+      for (let i = 0; i < segments.length; i++) {
+        const segment = decodeURIComponent(segments[i].replace(/~1/g, '/').replace(/~0/g, '~'));
+        if (i === segments.length - 1) {
+          if (current[segment] === undefined || current[segment] === null) {
+            current[segment] = [];
+          }
+        } else {
+          if (current[segment] === undefined || current[segment] === null) {
+            break;
+          }
+          current = current[segment];
+        }
+      }
+    }
+  }
+
   try {
     const patchErrors = applyPatch(obj, patch).filter(Boolean);
     if (patchErrors.length) {


codyebberson:
Thanks for submitting @themavik.  We certainly support this philosophically.

See related discussion on the [rfc6902 repo](https://github.com/chbrown/rfc6902/issues/104).  We had hoped to get this added to the underlying JSONpatch library, but it looks like that conversation has stalled.

We will take a look.  This one will require a lot of tests to cover all of the edge cases here (in particular nested objects, nested arrays, and all combinations of those).

codyebberson:
Hi @themavik 

Follow-up on this.  It looks like this behavior will be fixed upstream in the rfc6902 library itself: https://github.com/chbrown/rfc6902/pull/106

We're going to pursue that option instead.

Thanks again for submitting 🙏

PR #8404, opened 2026-02-11
fix(server/subscription): purge subs w/ invalid token

### LINKED ISSUE #8403: WebSocket subscriptions w/ invalid token evaluated multiple times before cleanup
We see in prod a lot of cases where for one subscription we see the following log lines multiple times in the span of a few minutes before finally disappearing:
`[WS] Unable to get login for the given access token`

It seems that this is related to a token expiring for a WebSocket subscription but have not yet GC'd it. We can clean up all the subscriptions that come back "dead" in that loop directly which should avoid several cycles of trying to evaluate a subscription before it is removed by GC.

ThatOneBro:
Fixes #8403


codyebberson:
@claude please review

PR #8407, opened 2026-02-11
Add token ':text' modifier to SearchControl

codyebberson:
Added "Text contains..." to the popup menu:

image

Added "text" to the filter editor dialog:

image


packages/react/src/SearchPopupMenu/SearchPopupMenu.tsx
```diff
@@ -109,8 +109,9 @@ function SearchParameterSubMenu(props: SearchPopupSubMenuProps): JSX.Element {
     case 'string':
       return <TextFilterSubMenu {...props} />;
     case 'token':
-    case 'uri':
       return <TokenFilterSubMenu {...props} />;
+    case 'uri':
+      return <UriFilterSubMenu {...props} />;
```
mattwiller, line 114: 👍

PR #8415, opened 2026-02-11
Log authentication middleware errors

mattwiller:


packages/server/src/app.ts
codyebberson, line 287: 👍 

```ts
const x = undefined;
class MyClass {}
console.log(x instanceof MyClass); // does not throw, prints false
```

packages/server/src/context.ts
```diff
@@ -133,20 +133,27 @@ export function getAuthenticatedContext(): AuthenticatedRequestContext {
 }
 
 export async function attachRequestContext(req: Request, res: Response, next: NextFunction): Promise<void> {
+  const { requestId, traceId } = requestIds(req);
+  let ctx = new RequestContext(requestId, traceId);
```
codyebberson: Does this mean we will always create a throwaway `RequestContext` instance on authenticated requests?

It doesn't do too much, so perhaps not a big deal, but it is still unnecessary work.

```
  constructor(requestId: string, traceId: string, logger?: Logger, loggerMetadata?: Record<string, any>) {
    this.requestId = requestId;
    this.traceId = traceId;
    this.logger =
      logger ??
      new Logger(write, { ...loggerMetadata, requestId, traceId }, parseLogLevel(getConfig().logLevel ?? 'info'));
  }
```

Could we move this into an `else` block or a `if (!ctx)` block?

packages/server/src/context.ts
```diff
@@ -133,20 +133,27 @@ export function getAuthenticatedContext(): AuthenticatedRequestContext {
...
       ctx = new AuthenticatedRequestContext(requestId, traceId, authState, repo);
-    } else {
-      ctx = new RequestContext(requestId, traceId);
     }
-    requestContextStore.run(ctx, () => next());
-  } catch (err) {
-    next(err);
+  } catch (err: any) {
+    // Ensure next() is called in a request context, so later middleware (e.g. logging) can run correctly
+    requestContextStore.run(ctx, () => {
```
codyebberson, line 155: Why do we need to call `requestContextStore.run()` here?  Versus just calling `next()` directly?

mattwiller: See the comment above: we want to attach the context to ensure that it's available for any subsequent middleware in the stack

packages/server/src/context.ts
mattwiller: Sure, I can refactor a bit

packages/server/src/context.ts
```diff
@@ -141,20 +141,30 @@ export function getAuthenticatedContext(): AuthenticatedRequestContext {
 }
 
 export async function attachRequestContext(req: Request, res: Response, next: NextFunction): Promise<void> {
+  const { requestId, traceId } = requestIds(req);
```
mattlong, line 144: To be extra defensive, should we make it `let requestId...` and `let traceId...` and move the function call inside they `try`?

mattlong (approved):

packages/server/src/context.ts
mattwiller: Those values are used in both the `catch` and after the `try`, so I think the current placement makes the most sense.  I did clean up the double call to `requestIds()` that got introduced in a merge conflict though

PR #8426, opened 2026-02-12
fix(client/subscription): refresh expiring tokens

### LINKED ISSUE #8425: Client does not automatically refresh expiring WebSocket sub tokens
Currently, after one hour if a WebSocket subscription criteria has not changed and the page has not been refreshed, the Subscription stops working since the token expires. We can periodically check all tokens for expiry in the `SubscriptionManager` and get a new token before the expiry time to ensure we don't miss any event notifications.

ThatOneBro:
Fixes #8425

### DIFF
--- a/packages/core/src/subscriptions/index.ts
+++ b/packages/core/src/subscriptions/index.ts
@@ -476,7 +476,9 @@ export class SubscriptionManager {
           criteriaEntry.tokenExpiry - now <= WS_SUB_TOKEN_EXPIRY_GRACE_PERIOD_MS &&
           !this.isReconnecting(criteriaEntry)
         ) {
-          this.rebindCriteriaEntry(criteriaEntry).catch(console.error);
+          this.rebindCriteriaEntry(criteriaEntry).catch((err: Error) => {
+            this.masterSubEmitter?.dispatchEvent({ type: 'error', payload: err });
+          });
         }
       }
     }
--- a/packages/core/src/subscriptions/index.test.ts
+++ b/packages/core/src/subscriptions/index.test.ts
@@ -1388,6 +1388,62 @@ describe('SubscriptionManager', () => {
 
       manager.closeWebSocket();
     });
+
+    test('should emit error on master emitter when rebindCriteriaEntry rejects during token refresh', async () => {
+      const EXPIRING_TOKEN = 'expiring-token-123';
+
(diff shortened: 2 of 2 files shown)


packages/core/src/subscriptions/index.ts
```diff
@@ -398,35 +418,66 @@ export class SubscriptionManager {
...
-        criteriaEntry.clearAttachedSubscription();
-        await this.subscribeToCriteria(criteriaEntry);
+        if (!criteriaEntry.tokenExpiry) {
+          continue;
+        }
+        if (
+          criteriaEntry.tokenExpiry - now <= WS_SUB_TOKEN_EXPIRY_GRACE_PERIOD_MS &&
+          !this.isReconnecting(criteriaEntry)
+        ) {
+          this.rebindCriteriaEntry(criteriaEntry).catch(console.error);
```
mattlong: Burying errors here is okay? If so, could you add a comment explaining why it is okay.

packages/core/src/subscriptions/index.ts
```diff
@@ -305,15 +321,19 @@ export class SubscriptionManager {
...
+    const expiration = parameter?.find((param) => param.name === 'expiration')?.valueDateTime;
 
     if (!token) {
       throw new OperationOutcomeError(validationError('Failed to get token'));
     }
     if (!url) {
       throw new OperationOutcomeError(validationError('Failed to get URL from $get-ws-binding-token'));
     }
+    if (!expiration) {
+      throw new OperationOutcomeError(validationError('Failed to get expiration from $get-ws-binding-token'));
```
mattlong, line 333: Was `expiration` already a required parameter? Without knowing more, this looks like it could be a backwards-incompatible change.

ThatOneBro: Yes this is from the WebSocket subscription spec / the operation definition, expiration was always there we just weren't reading it before on the client side 👍 

https://github.com/medplum/medplum/blob/main/packages/server/src/fhir/operations/getwsbindingtoken.ts#L181

See also: https://build.fhir.org/ig/HL7/fhir-subscription-backport-ig/OperationDefinition-backport-subscription-get-ws-binding-token.html

packages/core/src/subscriptions/index.ts
ThatOneBro: Good call, we can emit an error from the subscription manager

PR #8429, opened 2026-02-13
Fix websocket demo

codyebberson:
While investigating some WebSocket behaviors, I tried using the websocket demo.  I think a couple of things broke due to changes in `MockClient` and `createResourceIfNoneExist`.

### DIFF
--- a/packages/examples/src/react/use-subscription.tsx
+++ b/packages/examples/src/react/use-subscription.tsx
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // start-block lifecycleHooks
+import type { Bundle } from '@medplum/fhirtypes';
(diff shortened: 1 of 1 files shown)


examples/medplum-websocket-subscriptions-demo/src/components/BundleDisplay.tsx
```diff
@@ -35,6 +34,13 @@ export function BundleDisplay(props: BundleDisplayProps): JSX.Element {
     [medplum, communication]
   );
 
+  if (!communication) {
+    return undefined;
+  }
+
+  const [senderType, senderId] = ((communication.sender as Reference).reference as string).split('/');
```
ThatOneBro, line 41: I think this can be `parseReference()`?

PR #8443, opened 2026-02-16
perf(subscriptions): factor out resource from pubsub payload entries

### LINKED ISSUE #8442: Optimization: factor out the resource for the subscription event so it's not in every entry
Currently we needlessly send the resource that a subscription event fired for in every single entry for every single subscription on publish via Redis PubSub. This is extremely inefficient, since if there are even 10 subscriptions that fire for this resource, we copy the resource 10 times for every publish. This would take this from O(N) copies of the resource to O(1) for each resource that triggers multiple subscriptions.

NOTE: Since this does change the payload of the message the subscribing servers will be expecting, we will have to support a "v1" and "v2" payload temporarily.

ThatOneBro:
Fixes #8442 

Follow-up to #8453

### DIFF
--- a/packages/server/src/subscriptions/websockets.test.ts
+++ b/packages/server/src/subscriptions/websockets.test.ts
@@ -719,7 +719,7 @@ describe('WebSocket Subscription', () => {
           if (msg.entry?.[0]?.resource?.resourceType !== 'SubscriptionStatus') {
             return false;
           }
-          const status = msg.entry[0].resource as SubscriptionStatus;
+          const status = msg.entry[0].resource;
           if (status.type !== 'event-notification') {
             return false;
           }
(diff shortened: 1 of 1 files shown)


ThatOneBro:
I think to avoid any regressions during canary deployment, we may need to split this into two parts:
1. Add parsing on the WebSocket subscriber side in one PR
2. Start sending new payload in second PR

EDIT: Actually even that seems problematic for anyone who doesn't stop at this patch version... 🤔 
Two options I see:
1. Cut a minor release after the first PR
2. Make it one PR but put this new payload behind a feature flag until the next minor version

codyebberson (approved):

ThatOneBro:
Thanks @codyebberson 

We should cut a patch release for the step 1 PR to get published for self-hosters then merge this 👍

Issue #8417, opened 2026-02-12
$find & $book time zone issues


Ljeytl:
eachDayOfInterval uses server-local date components instead of actor-local, affects $find and $book for evening queries
eachDayOfInterval in scheduling.ts:15-28 extracts date components via getFullYear()/getMonth()/getDate() (server-local) and passes them to ZonedDateTime.from() with the actor's timezone. On a UTC server with a US-timezone actor, the UTC date can be ahead of the actor's local date during evening hours.
Affected operations: $find and $book — both call resolveAvailability → eachDayOfInterval.
Trigger: interval.start falls between UTC midnight and actor's local midnight (~4pm PST / 7pm EST onward).
Impact: $find returns zero availability for narrow evening ranges. $book can reject valid evening appointments.

Existing tests don't exercise the crossover window.

noahsilas:
Dang, great catch!

I was able to write a quick test case to validate this, but to catch it locally I need to run the test like `TZ=utc npm test` so that node doesn't inherit my local timezone from the environment. I looked briefly into switching the node TZ on the fly for the test, but all the solutions here are fairly heavy-weight...

One option would be to amend our `test` script definition to always pass that env var, but that precludes being able to assert the reverse (that the server behaves in any other TZ). Another option would be to spin up separate server processes in the tests with the different time zones. I'm going to spend a little time considering what that could look like.

In the short term, I'll put up a fix with at least a test case that should catch regressions for this in CI (where I believe the clock is set to UTC).

Thanks!

PR #8447, opened 2026-02-17
Fix scheduling ops during UTC/local date discrepancy

noahsilas:
For operations like `$find` and `$book`, there is an interesting edge case: when the server is run in a different time zone (such as a server in UTC and clients in the US), there is a window where the current date in the server local time is different from the client date.

The original implementation of `eachDayOfInterval` naively used the date component from the interval start without the time components, and so wasn't shifting appropriately to the correct time when the timezone information was added.

I've adjusted the implementation to begin by creating an exact `Temporal.Instant` from the interval start date, and then using `.toZonedDateTime(timezone)` to shift that into the desired timezone. Finally, to keep the previous behavior of emitting start-of-day temporal instances, we explicitly set the time components to zero with `withPlainTime(...)`.

Implementation Note: Testing this is a bit of a chore. The test case as written passes on my machine when run naively with `npm test`, because Node defaults to running with the same time zone as the local system. You can see the test fail by explicitly setting the process to be in a different timezone: `TZ=utc npm test`. Ideally we should be able to test different server/client timezone options, but adjusting the Node timezone after startup seems to be more fraught than expected (see packages like https://www.npmjs.com/package/timezone-mock for some potential solutions).

For now I'm adding a test case that would reveal the problem in CI, where tests are run with the system time set to UTC. In the long term, we probably want a set of integration tests that spin up servers with a variety of configurations, which would allow us to test multiple servers with different timezones and compare their results. I'm deferring building that into the future for now.

Fixes: https://github.com/medplum/medplum/issues/8417

### DIFF
--- a/packages/server/src/fhir/operations/utils/scheduling.test.ts
+++ b/packages/server/src/fhir/operations/utils/scheduling.test.ts
@@ -142,6 +142,35 @@ describe('resolveAvailability', () => {
     ]);
   });
 
+  // regression test for https://github.com/medplum/medplum/issues/8417
+  test('when the request starts early in a UTC day', () => {
+    const schedulingParameters: SchedulingParameters = {
+      availability: [
+        {
+          dayOfWeek: ['tue', 'wed', 'thu'],
+          timeOfDay: ['20:00:00'],
+          duration: 360,
+        },
+      ],
+      duration: 60,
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude  please review

PR #8451, opened 2026-02-17
Remove User along with ProjectMembership if user is project-scoped

maddyli:
A Medplum user ran into an issue with the Smart on FHIR launch, which would have been avoided if project-scoped User resources are also deleted when the ProjectMembership is deleted upon user removal. 

They had added project-scoped users to their project, and then deleted those users using the "Remove user" button in the UI. This resulted in the ProjectMembership being deleted, but the User resources still exists as (as Users in that `project`) without a ProjectMembership attached. 

Then, they re-added these users, with the same emails, as server-scoped users. Upon attempting to oauth in the middle of the Smart on FHIR launch, Medplum server would look up User resources attached to that email, find the User that does not have a ProjectMembership attached (the User that we first provisioned), and it does not allow the users to log in. We were able to resolve this by deleting all the Users without a ProjectMembership attached. 

This PR aims to ensure that we don't have hanging project-scoped User resources without a ProjectMembership resource attached.

### DIFF
--- a/packages/server/src/admin/project.ts
+++ b/packages/server/src/admin/project.ts
@@ -148,7 +148,7 @@ projectAdminRouter.delete('/:projectId/members/:membershipId', async (req: Reque
           {
             code: 'user',
             operator: Operator.EQUALS,
-            value: user.id,
+            value: getReferenceString(user),
           },
         ],
         count: 2,


packages/server/src/admin/project.ts
```diff
@@ -132,6 +133,47 @@ projectAdminRouter.delete('/:projectId/members/:membershipId', async (req: Reque
...
+  let userId: string | undefined;
+  if (membership.user?.reference) {
+    const systemRepo = getSystemRepo();
+    const user = await systemRepo.readReference<User>(membership.user as Reference<User>);
+
+    // Check if the user is project-scoped (has a project field matching the current project)
+    if (user.project?.reference === getReferenceString(ctx.project)) {
+      // Check if there are other ProjectMemberships for this user
+      // (search before deleting to get accurate count)
+      const otherMemberships = await systemRepo.searchResources<ProjectMembership>({
```
mattlong, line 145: The search and the delete(s) should be wrapped in a transaction.

packages/server/src/admin/project.ts
```diff
@@ -132,6 +133,47 @@ projectAdminRouter.delete('/:projectId/members/:membershipId', async (req: Reque
...
+        resourceType: 'ProjectMembership',
+        filters: [
+          {
+            code: 'user',
+            operator: Operator.EQUALS,
+            value: membership.user.reference,
+          },
+        ],
+        count: 10,
```
mattlong: Seems like this could be `count: 2`?

packages/server/src/admin/project.ts
```diff
@@ -132,6 +133,47 @@ projectAdminRouter.delete('/:projectId/members/:membershipId', async (req: Reque
...
+      }
+    }
+  }
+
+  // Delete the ProjectMembership
   await ctx.repo.deleteResource('ProjectMembership', membershipId);
+
+  // Delete the User resource if it's project-scoped and this was their only membership
+  if (shouldDeleteUser && userId) {
+    const systemRepo = getSystemRepo();
```
mattlong: Along with wrapping in a transaction, the same `systemRepo` instance should be used for this delete.

mattlong (approved):

PR #8453, opened 2026-02-17
perf(subscriptions/ws): parse new efficient WS event payload

ThatOneBro:
Split out from #8443

This PR is part of a 3 step process to safely merge this payload change for the WebSocket subscription pubsub system.

The steps are:
1. Merge this PR which only supports parsing the new payload on the WebSocket PubSub subscriber side. This does not change the runtime behavior to the current server, only supports receiving the new payload without ever sending it on the publish side.
2. Publish a minor version so that customers who are migrating as self-hosters will pick up a version that supports both payloads before progressing to the next version
3. Merge #8443 which begins using the new payload, which should significantly reduce the overhead of our publish operations on Redis (at least for WebSocket subscriptions)

### DIFF
--- a/packages/server/src/subscriptions/websockets.test.ts
+++ b/packages/server/src/subscriptions/websockets.test.ts
@@ -537,7 +537,7 @@ describe('WebSocket Subscription', () => {
           if (msg.entry?.[0]?.resource?.resourceType !== 'SubscriptionStatus') {
             return false;
           }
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Deployed to staging: https://github.com/medplum/medplum/actions/runs/22121783078

Was using WebSocket subscriptions during rolling deploy and didn't notice any hiccups or anything. Looks good from that perspective 👍

PR #8455, opened 2026-02-18
Do not visualize $find slots in provider calendar

noahsilas:
This was helpful as a debugging tool for the initial $find/$book pass, but showing the slots on calendar creates confusion about what is already scheduled versus what is available.

Not rendering the available slots from $find on the big calendar should help clear this up.

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
+++ b/examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
@@ -373,7 +373,7 @@ export function SchedulePage(): JSX.Element | null {
           />
         </div>
 
-        {serviceTypes?.length && schedule && range && (
+        {Boolean(serviceTypes?.length) && schedule && range && (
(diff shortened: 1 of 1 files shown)


examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -373,21 +367,19 @@ export function SchedulePage(): JSX.Element | null {
...
             onSelectSlot={handleSelectSlot}
-            slots={[...(slots ?? []), ...(findSlots ?? [])]}
+            slots={slots ?? []}
             appointments={appointments ?? []}
             onRangeChange={setRange}
           />
         </div>
 
-        {serviceTypes?.length && schedule && range && (
+        {Boolean(serviceTypes?.length) && schedule && range && (
```
ThatOneBro, line 376: 👍 so we're not rendering 0

PR #8465, opened 2026-02-18
feat(app): combine Users and Patients into Members tab on admin/project page

maddyli:
Screenshot 2026-02-24 at 6 40 19 PM
Screenshot 2026-02-24 at 6 40 14 PM
Screenshot 2026-02-24 at 6 40 10 PM

### DIFF
--- a/packages/app/src/admin/ResourceMemberTable.test.tsx
+++ b/packages/app/src/admin/ResourceMemberTable.test.tsx
@@ -44,7 +44,9 @@ describe('ResourceMemberTable', () => {
   test('Renders on Clients page with ClientApplication resourceType', async () => {
     await setup('/admin/clients');
     // Page title and create link confirm ResourceMemberTable rendered without crashing
-    expect(await screen.findByRole('heading', { name: 'ProjectMemberships for ClientApplications' })).toBeInTheDocument();
+    expect(
+      await screen.findByRole('heading', { name: 'ProjectMemberships for ClientApplications' })
+    ).toBeInTheDocument();
     expect(screen.getByText('Create new client')).toBeInTheDocument();
   });
 


packages/app/src/AppRoutes.tsx
```diff
@@ -87,18 +85,16 @@ export function AppRoutes(): JSX.Element {
...
         <Route path="/admin" element={<ProjectPage />}>
-          <Route path="patients" element={<PatientsPage />} />
           <Route path="bots/new" element={<CreateBotPage />} />
           <Route path="bots" element={<BotsPage />} />
           <Route path="clients/new" element={<CreateClientPage />} />
           <Route path="clients" element={<ClientsPage />} />
           <Route path="details" element={<ProjectDetailsPage />} />
           <Route path="invite" element={<InvitePage />} />
-          <Route path="users" element={<UsersPage />} />
+          <Route path="members" element={<UsersPage />} />
```
codyebberson: Hm... This will be a breaking change.  Are we concerned about that?

If we want, we could continue to support the old URLs, but just point them to the new unified `<UsersPage>`.

I don't have strong opinions about that.

maddyli: I got some feedback from Kevin, and will likely keep the "Users" language. Will change it back

maddyli:
@claude could you please review this PR?

maddyli:
@claude could you give a quick re-review?

kevinwadeshaw:
@maddyli This is looking good! A few ideas for here for the Users tab:

- I think we could remove the descriptive text here—the title above and tabs below already give you this context 
- If that works, I think we could move the link to the far right as secondary reference 
- Also, since we're already making changes here, what do you think of elevating the invite link as a button? 

Altogether, it would appear like this:

Users-Admin

maddyli:
changed! 
Screenshot 2026-02-25 at 4 49 55 PM

maddyli:
@claude please review

PR #8158, opened 2026-01-06
Defining Availability Time Zones Documentation

finnbergquist:

PR #8468, opened 2026-02-18
Allow specifying timezone only on SchedulingParameters for $find

noahsilas:
In the original implementation, we had expected that timezone would
always be expressed as an attribute of an actor, and so had made that a
requirement to use the scheduling features.

Some use cases exist where folks want to specify the timezone in the
scheduling parameters extension - this likely will be more useful in the
future when this extension can be specified on a shared
ActivityDefinition that could be used for all practitioners in a state,
for example.

Spec update introducing this: commit https://github.com/medplum/medplum/commit/ba37a1b638d4ecb9368f9a987bb0e90724606e2e (https://github.com/medplum/medplum/pull/8158)
Original parameter implementation: commit https://github.com/medplum/medplum/commit/dad7aa0ed2a71b73f1cd6135c0d5961ef554efca (https://github.com/medplum/medplum/pull/8247)

### DIFF
--- a/packages/core/src/outcomes.test.ts
+++ b/packages/core/src/outcomes.test.ts
@@ -89,6 +89,10 @@ describe('Outcomes', () => {
   test('Bad Request', () => {
     expect(isOk(badRequest('bad'))).toBe(false);
     expect(badRequest('bad', 'bad').issue?.[0]?.expression?.[0]).toBe('bad');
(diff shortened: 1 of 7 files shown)


noahsilas:
The `badRequest` helper accepting an array of expressions was implemented in #9088. 

In #8793 we adjusted the timezone handling while working on multi-schedule $find and effectively incorporated the intent here.

If we wanted to ensure no regressions we could pull some of the tests from this PR over to the current `main` branch, but I think for now I'm just going to close this as it is effectively resolved.

PR #8471, opened 2026-02-18
Skip search count queries when possible

mattlong:


mattlong:
@claude take a look; is the short-circuit being applied too broadly?

PR #8474, opened 2026-02-19
fix(deploy): make all deployments share concurrency group

### LINKED ISSUE #8473: Multiple deploys to main can happen concurrently
Due to an edge case where a deploy job happens just late enough to not get collapsed into one deploy by the merge queue, you can actually get 2 deploys running concurrently, which is definitely dangerous.

This is because our concurrency group is based on a key which will always be unique for every commit, which is not what we want. It should probably be based on the branch name (main). We should never see 2 deploy jobs running at the same time.

See:

Image

These two deployments started concurrently and the later deployment actually finished first. I had to manually cancel the earlier deployment to stop it from clobbering the previous one.

ThatOneBro:
Fixes #8473 

This PR makes it so that deployments will be queued if they defeat the merge queue and multiple deploy jobs are attempted concurrently

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,7 +1,7 @@
 name: Deploy
 
 concurrency:
-  group: ci-deploy-${{ github.sha }}
+  group: ci-deploy-main
   cancel-in-progress: false
 
 on:


codyebberson:
@claude - please review.  Is this the right way to use GitHub merge queues?

PR #8477, opened 2026-02-19
Support instance-level custom operations with resource input

rahul1:
## Summary
This change adds support for instance-level custom operations in FHIR, allowing operations to be invoked on specific resource instances (e.g., `/Patient/123/$my-operation`) with the target resource automatically passed as input to the bot.

## Key Changes
- **Modified input resolution logic**: Updated `tryCustomOperation` to detect instance-level operations by checking if the operation is preceded by both a resource type and resource ID in the URL path
- **Automatic resource loading**: For instance-level operations, the target resource is now automatically read from the repository and passed as input to the bot, rather than using the request body/query string
- **Type imports**: Added `ResourceType` import to support dynamic resource type handling
- **Test coverage**: Added comprehensive test case demonstrating instance-level operation invocation, including:
  - Creating a Patient resource
  - Creating a Bot that echoes its input
  - Defining an OperationDefinition with `instance: true`
  - Invoking the operation on a specific Patient instance and verifying the resource is returned

## Implementation Details
The operation detection logic checks if the operation part appears at index 3 or later in the URL path (after resource type and ID), and verifies that both the resource type and ID exist at the expected positions. This allows the system to distinguish between:
- Instance-level: `/fhir/R4/Patient/123/$my-operation` (uses Patient/123 as input)
- Type-level: `/fhir/R4/Patient/$my-operation` (uses request body/query as input)
- System-level: `/fhir/R4/$my-operation` (uses request body/query as input)

https://claude.ai/code/session_01SWTtcyqQ9JvVP6Snuf1Ki1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -793,7 +793,6 @@
       "integrity": "sha512-RB9bKgYTVUiOcEb5bOcZ169jiiVW811dCsJoLT19DcbbFmU4QaK0ghSTssij35QBQ3SCOitXOUrHcGgNVwS7sQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@algolia/client-common": "5.48.0",
         "@algolia/requester-browser-xhr": "5.48.0",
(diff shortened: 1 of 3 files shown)


rahul1:
@claude please review

rahul1:
@claude please review

rahul1:
@claude ready for re review

codyebberson (commented):
We should revert the changes to `package-lock.json`, then good-to-go.

codyebberson (approved):

kapetan:
Hi. I can see this pull-request is already merged, and I like the idea of automatically loading resources in instance level operations. But as far as I can see, with the current implementation there is no way of passing any additional information to the bot handler. E.g. I want to call something like `POST Appointment/:id/$my-operation` with body `{ "test": 1 }`, how do I access the body, which previously was passed as input to the bot handler?

Maybe the `BotEvent` should have property `instance` or `resource` that contains the loaded resource and input contains the body, so it's consistent with system and type level operations. Or add new property, e.g. `body` to the event that contains the body.

What do you think? I could make a pull-request with something like that if it makes sense.

PR #8482, opened 2026-02-19
Move ThreadInbox to react package

techdavidy:
Relocated the ThreadInbox component and its subcomponents from provider into packages/react, and moved the useThreadInbox hook into packages/react-hooks. Both are now exported from their respective package indexes, making them available as reusable library components. 

Updated call sites in provider MessagesPage, CommunicationTab) to import from the packages.

### DIFF
--- a/packages/react-hooks/src/useThreadInbox/useThreadInbox.ts
+++ b/packages/react-hooks/src/useThreadInbox/useThreadInbox.ts
@@ -162,9 +162,7 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
       const updatedThread = await medplum.updateResource({ ...selectedThread, status: newStatus });
       setSelectedThread(updatedThread);
       setThreadMessages((prev) =>
-        prev.map(([parent, lastMsg]) =>
-          parent.id === updatedThread.id ? [updatedThread, lastMsg] : [parent, lastMsg]
-        )
+        prev.map(([parent, lastMsg]) => (parent.id === updatedThread.id ? [updatedThread, lastMsg] : [parent, lastMsg]))
       );
     };
     doUpdate().catch((err: Error) => setError(err));


techdavidy:
@claude please review

packages/react/src/chat/ThreadInbox/ChatList.test.tsx
```diff
@@ -45,16 +43,14 @@ const mockLastCommunication1: Communication = {
...
 
 describe('ChatList', () => {
   let medplum: MockClient;
 
   beforeEach(async () => {
     medplum = new MockClient();
-    vi.clearAllMocks();
-
-    // Create patient resources
+    jest.clearAllMocks();
```
noahsilas, line 53: Asking purely for my own context: are we generally moving towards either jest or vitest? It would be kind of nice to be more consistently in one runner, so maybe I can take a stab at converting more packages in one direction or the other. (My intuition is probably that we want vitest for UI packages and jest for Node things, but that's not necessarily the right choice...)

packages/react/src/chat/ThreadInbox/ChatListItem.module.css
```diff
@@ -10,9 +10,8 @@
 }
 
 .contentContainer {
-  transition: background 0.2s;
```
noahsilas, line 13: Dropping this transition seems fine to me. Was something funky happening with it?

packages/react-hooks/src/useThreadInbox/useThreadInbox.test.tsx
```diff
@@ -1,13 +1,12 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { renderHook, waitFor, act } from '@testing-library/react';
-import { MedplumProvider } from '@medplum/react';
-import type { JSX } from 'react';
+import type { WithId } from '@medplum/core';
```
noahsilas, line 3: There's a bunch of churn here that looks like it's due to the provider app not enforcing the same import ordering rules as other code. I'm thinking about following up by adding that `prettier` plugin to provider to make this kind of movement easier in the future. Any thoughts or objections?

packages/react/src/chat/ThreadInbox/ChatList.test.tsx
```diff
@@ -64,13 +60,12 @@ describe('ChatList', () => {
...
-            <ChatList threads={threads} selectedCommunication={selectedCommunication} getThreadUri={mockGetThreadUri} />
-          </MantineProvider>
-        </MedplumProvider>
-      </MemoryRouter>
+      <ChatList threads={threads} selectedCommunication={selectedCommunication} getThreadUri={mockGetThreadUri} />,
+      ({ children }) => (
+        <MemoryRouter>
+          <MedplumProvider medplum={medplum}>{children}</MedplumProvider>
+        </MemoryRouter>
+      )
```
noahsilas, line 68: #TIL about the `wrapper` option that is being used under the hood here. Very nice that our `render` helper automatically adds the `<MantineProvider>` wrapper!

packages/react/src/chat/ThreadInbox/ChatListItem.module.css
techdavidy: double transition

packages/react/src/chat/ThreadInbox/ChatList.test.tsx
techdavidy: Great question. React package uses jest unfortunately.
We would move forward with vitest in newer projects.

packages/react-hooks/src/useThreadInbox/useThreadInbox.test.tsx
techdavidy: agreed

PR #8485, opened 2026-02-19
fix: set meta.project on User resources created via self-registration

deam65:
When a patient self-registers via `RegisterForm` / `POST /auth/newuser` with a `projectId`, the created `User` resource was missing `meta.project`. The repo layer's `getProjectId` logic for `Users` in a project admin context reads only `meta.project` (`not User.project`), so the project compartment was never stamped, making the resource inaccessible to project admins.

Mirrors the pattern already used in the admin invite flow (invite.ts). Adds regression tests to assert `meta.project` is set (or absent) correctly.

### DIFF
--- a/packages/server/src/auth/newuser.test.ts
+++ b/packages/server/src/auth/newuser.test.ts
@@ -568,17 +568,14 @@ describe('New user', () => {
 
     // Register a new user into that project via the self-registration endpoint
     const newEmail = `patient${randomUUID()}@example.com`;
-    const res = await request(app)
-      .post('/auth/newuser')
-      .type('json')
-      .send({
-        projectId: project.id,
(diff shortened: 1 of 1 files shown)


packages/server/src/auth/newuser.ts
```diff
@@ -108,6 +108,7 @@ export async function createUser(request: Omit<NewUserRequest, 'recaptchaToken'>
   const systemRepo = getGlobalSystemRepo();
   const result = await systemRepo.createResource<User>({
     resourceType: 'User',
+    meta: projectId && projectId !== 'new' ? { project: projectId } : undefined,
```
codyebberson, line 111: 👍

codyebberson (approved):

Issue #8253, opened 2026-01-21
C-CDA to FHIR Transform: `@nullFlavor` handling produces invalid FHIR


amcgivern:
## Summary

When using the C-CDA to FHIR transform function with documents containing `@nullFlavor` attributes, the resulting FHIR is often invalid and gets rejected when uploading to a Medplum FHIR server.
I discovered this while testing with TEFCA test data as part of our effort to transform C-CDA documents received via TEFCA.

### Issue 1: `@nullFlavor` on primitive fields outputs an object instead of a valid value

When `@nullFlavor` exists on a data field with a primitive data type (e.g., string for `postalCode`, `lotNumber`), the transform outputs an object, which is invalid FHIR.

#### Example input (C-CDA)

```xml
<manufacturedMaterial>
  <code codeSystem="2.16.840.1.113883.12.292" codeSystemName="CVX" nullFlavor="NI"/>
  <lotNumberText nullFlavor="NI"/>
</manufacturedMaterial>
```

#### Actual output (invalid FHIR)

```json
{
  "resourceType": "Immunization",
  "id": "af491bce-bf5d-11ef-a464-edfab3944fea",
  "status": "completed",
  "vaccineCode": {
    "coding": [
      {
        "system": "http://nucc.org/cvx"
      }
    ]
  },
  "lotNumber": {
    "@_nullFlavor": "NI"
  }
}
```

#### Expected behavior

`lotNumber` is a string type in FHIR. The field should either be omitted entirely or represented using a Data Absent Reason extension.

### Issue 2: `null` values in CodeableConcept coding arrays

When a code element has `@nullFlavor` but includes a `<translation>` element, the transform outputs a `null` value in the coding array, which is invalid FHIR.

#### Example input (C-CDA)

```xml
<code nullFlavor="UNK">
  <translation code="75320-2" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" displayName="Advance directive"/>
</code>
```

#### Actual output (invalid FHIR)

```json
{
  "code": {
    "coding": [
      null,
      {
        "system": "http://loinc.org",
        "code": "75320-2",
        "display": "Advance directive"
      }
    ]
  }
}
```

#### Expected behavior

The `null` value should be filtered out from the array:

```json
{
  "code": {
    "coding": [
      {
        "system": "http://loinc.org",
        "code": "75320-2",
        "display": "Advance directive"
      }
    ]
  }
}
```

---

[...]

PR #8486, opened 2026-02-19
fix(ccda): Handle nullFlavor values in CCDA to FHIR conversion 

amcgivern:
## Why
- Resolves this issue: #8253

## What
  - Add `sanitizeNullFlavors `that recursively strips `@_nullFlavor` keys from parsed CCDA objects, collapsing objects/arrays that become empty after removal
  - Calls `sanitizeNullFlavors` from `convertCcdaToFhir` so all downstream conversion logic receives clean data
  - Adds defensive guards in `mapCode`, `processVitalsObservation`, and `processObservation` to skip resources that have no valid code after sanitization

2 other small fixes:
  - Fix qualification mapping to correctly nest the CodeableConcept under a code key, matching the FHIR R4 [PractitionerQualification structure](https://hl7.org/fhir/R4/practitioner.html#resource)
  - Handle YYYYMMDDHHMM+TZ format (timezone offset without seconds component)

### DIFF
--- a/packages/ccda/src/systems.test.ts
+++ b/packages/ccda/src/systems.test.ts
@@ -160,7 +160,9 @@ describe('EnumMapper', () => {
       const result = mapCcdaCodeToCodeableConcept({
         '@_nullFlavor': 'OTH',
         '@_codeSystem': OID_SNOMED_CT_CODE_SYSTEM,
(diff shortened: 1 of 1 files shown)


codyebberson:
Thanks @amcgivern !

PR #8487, opened 2026-02-20
Fixes #8454 - CDK option for rdsForceRetain

### LINKED ISSUE #8454: CDK changes for "bring your own database"
- add RemovalPolicy.retain to prevent destruction
- add a retainDatabase boolean config to cleanly manage the transition

codyebberson:
Customers requested the ability to switch to a different database instance (via the `rdsSecretsArn` CDK setting), but still retain the original CDK-managed RDS instance and all surrounding resources (parameter groups, networking resources, etc).

Note that we already use `removalPolicy: RemovalPolicy.RETAIN` for many/most of those resources, however, the CDK diff message can be quite intimidating for infra teams, so this was explicitly requested

To use this, you would:
1. Set `rdsSecretsArn` to the ARN of the AWS Secret with the connection details (as described in docs under "Bring Your Own Database")
2. Set `rdsForceRetain` to `true`

### DIFF
--- a/packages/cdk/src/backend.ts
+++ b/packages/cdk/src/backend.ts
@@ -102,7 +102,7 @@ export class BackEnd extends Construct {
 
     // RDS
     this.rdsSecretsArn = config.rdsSecretsArn;
-    if (!this.rdsSecretsArn) {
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

PR #8488, opened 2026-02-20
Customize appointment creation before $book call

noahsilas:
In #8326 as a shortcut to get a working demo up fast, I called in to `$book` without any patient information. This created appointments on the calendar immediately, which isn't expected, and they would be created with "No patient" displayed and required updating after the fact.

Let's do the work to get this more correct.

- Update the `$book` endpoint to take the `patient-reference` input parameter described in #7690.
- Did some refactoring of components and tests
- Added an intermediate UI step where we collect patient information before calling $book

https://github.com/user-attachments/assets/5046e2f2-15a2-4faf-baa7-1fde04dc0617

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/FindPane.tsx
+++ b/examples/medplum-provider/src/pages/schedule/FindPane.tsx
@@ -15,6 +15,8 @@ import { useSchedulingStartsAt } from '../../hooks/useSchedulingStartsAt';
 import { serviceTypesFromSchedulingParameters, SchedulingTransientIdentifier } from '../../utils/scheduling';
 import { BookAppointmentForm } from '../../components/schedule/BookAppointmentForm';
 
+const ONE_WEEK_MS = 7 * 24 * 60 * 60 * 1000;
+
 type FindPaneProps = {
   schedule: WithId<Schedule>;
   range: Range;
@@ -27,7 +29,7 @@ type FindPaneProps = {
 //
 // See https://www.medplum.com/docs/scheduling/defining-availability for details.
 export function FindPane(props: FindPaneProps): JSX.Element {
-  const [slots, setSlots] = useState<Slot[] | undefined>(undefined);
+  const [slots, setSlots] = useState<readonly Slot[] | undefined>(undefined);
   const [chosenSlot, setChosenSlot] = useState<Slot | undefined>(undefined);
   const { schedule, range, onSuccess } = props;
   const serviceTypes = useMemo(
@@ -60,7 +62,7 @@ export function FindPane(props: FindPaneProps): JSX.Element {
 
     // Compute search range
     const searchStart = range.start < earliestSchedulable ? earliestSchedulable : range.start;
-    const searchEnd = searchStart < range.end ? range.end : new Date(searchStart.getTime() + 1000 * 60 * 60 * 24 * 7);
+    const searchEnd = searchStart < range.end ? range.end : new Date(searchStart.getTime() + ONE_WEEK_MS);
     const start = searchStart.toISOString();
     const end = searchEnd.toISOString();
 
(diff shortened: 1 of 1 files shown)


techdavidy:
@claude please review

techdavidy:
it is a good opportunity to not only show patient resource input but also the rest of the fields to create an encounter and $apply a plan definition.

noahsilas:
> 1. Potential Race Condition in Patient Validation

While this does create a possible race condition, I consider the likelihood to be low, and my intuition is that it is better to keep the transaction tightly scoped. I think that the worst case here would be that an appointment is created with a dangling reference to a deleted patient, which ought to be repairable relatively painlessly from this low-volume endpoint. (I might have to eat these words later if patient deletion turns out to be more common than I expect).

> 2. Generic Error Swallowing (packages/server/src/fhir/operations/book.ts:151-152)

Good point, updated! See 8ba6e2517.

> 3. Missing Form Reset on Error (examples/medplum-provider/src/components/schedule/BookAppointmentForm.tsx:44-46)

This is intentional - if the form submission fails, the viewer should be able to retry it without needing to reconfigure the appointment.

> 4. Potential Memory Leak in FindPane (examples/medplum-provider/src/pages/schedule/FindPane.tsx:61-92)

I can see how this could be a trap for future readers, moving more of this computation inside the effect makes the dependency array more explicit in its intention. Nice! See e046b7b19.

> 5. Missing Accessibility Label (examples/medplum-provider/src/pages/schedule/FindPane.tsx:66-68)

Claude, I think you mean BookAppointmentForm.tsx. This also isn't a missing accessibility label - the ResourceInput uses the `label` prop to provide a `<label>` element tied to the input appropriately.

The field name you are flagging here is a client side concern only, since the native `FormData` isn't used (we provide our own `onSubmit` handler that doesn't use the form attribute names, but constructs the payload from the component state).

I've chosen `Patient-id` to match other places in this app code that use this same name.

> 6. Type Safety Could Be Improved (examples/medplum-provider/src/components/schedule/BookAppointmentForm.tsx:71)

Fair point. As written this had a slight error: the type should more accurately be `Patient | undefined`, as we allow clearing the input. And it is more future proof to explicitly test our runtime expectation and emit an error if something unexpected happens instead of squashing the type checker this way. Done in 20e5166d0

noahsilas:
> it is a good opportunity to not only show patient resource input but also the rest of the fields to create an encounter and $apply a plan definition.

@techdavidy Definitely! I plan on doing that work in this new component a future commit, but I would like to get this in fast to improve the existing demo experience, which has been described as "confusing". 😅

examples/medplum-provider/src/pages/schedule/FindPane.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const earliestSchedulable = useSchedulingStartsAt({ minimumNoticeMinutes: 30 });
+
+  useEffect(() => {
+    if (!schedule || serviceType === null) {
+      return () => {};
+    }
+
+    // Compute search range
+    const searchStart = range.start < earliestSchedulable ? earliestSchedulable : range.start;
+    const searchEnd = searchStart < range.end ? range.end : new Date(searchStart.getTime() + 1000 * 60 * 60 * 24 * 7);
```
techdavidy: magic number let's create a const for this please

examples/medplum-provider/src/pages/schedule/FindPane.tsx
```diff
@@ -0,0 +1,183 @@
...
+    medplum
+      .get<Bundle<Slot>>(`fhir/R4/Schedule/${schedule.id}/$find?${params}`, { signal })
+      .then(
+        (bundle) => {
+          if (!signal.aborted) {
+            if (bundle.entry) {
+              bundle.entry.forEach((entry) => entry.resource && SchedulingTransientIdentifier.set(entry.resource));
+              setSlots(bundle.entry.map((entry) => entry.resource).filter(isDefined));
+            } else {
+              setSlots([]);
```
techdavidy: empty ?

examples/medplum-provider/src/pages/schedule/FindPane.tsx
```diff
@@ -0,0 +1,183 @@
...
+    if (serviceType) {
+      serviceType.coding?.forEach((coding) => {
+        params.append('service-type', `${coding.system}|${coding.code}`);
+      });
+    }
+    medplum
+      .get<Bundle<Slot>>(`fhir/R4/Schedule/${schedule.id}/$find?${params}`, { signal })
+      .then(
+        (bundle) => {
+          if (!signal.aborted) {
```
techdavidy:
```
if (signal.aborted) {
return;
}
```

examples/medplum-provider/src/pages/schedule/FindPane.tsx
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  // tricky: `undefined` means the "wildcard" service type, so we explicitly
+  // test against `null` here.
+  if (serviceType !== null) {
+    return (
+      <Stack gap="sm" justify="flex-start">
+        <Title order={4}>
+          <Group justify="space-between">
+            <span>{serviceType ? <CodeableConceptDisplay value={serviceType} /> : 'Event'}</span>
```
techdavidy, line 146: do we need <span> ?

noahsilas: Yes, `serviceType` can be `undefined`, which means "match any service type" (aka "wildcard"). In that situation we need some fallback for display here.

PR #8510, opened 2026-02-24
feat(superadmin): add dashboard for WS sub stats

ThatOneBro:


packages/server/src/fhir/operations/getwssubstats.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    let criteriaMap = resourceTypeMap.get(resourceType);
+    if (!criteriaMap) {
+      criteriaMap = new Map();
+      resourceTypeMap.set(resourceType, criteriaMap);
+    }
+
+    const entries = await redis.hgetall(key);
+    for (const criteria of Object.values(entries)) {
```
noahsilas: It looks like we're dropping the key names, so you can probably use [`HVALS`](https://redis.io/docs/latest/commands/hvals/) to save a few bytes

packages/server/src/fhir/operations/getwssubstats.ts
```diff
@@ -0,0 +1,164 @@
...
+      resourceTypes,
+    });
+  }
+
+  projects.sort((a, b) => b.subscriptionCount - a.subscriptionCount);
+
+  const { repo } = getAuthenticatedContext();
+  for (const project of projects) {
+    try {
+      const projectResource = await repo.readResource<Project>('Project', project.projectId);
```
noahsilas: Not a blocker: it sort of feels like we should have a better method for bulk reading resources, but on inspection it seems like everything basically ends up in a pattern like this so far. I think for now I need to un-train an intuition I have about this kind of await-in-loop being scary for response times 😓

packages/app/src/admin/WsSubStatsWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+                <Table.Th>Count</Table.Th>
+              </Table.Tr>
+            </Table.Thead>
+            <Table.Tbody>
+              {projects.map((project) => (
+                <Fragment key={project.projectId}>
+                  <Table.Tr onClick={() => toggleProject(project.projectId)} style={{ cursor: 'pointer' }}>
+                    <Table.Td>
+                      <Group gap="xs">
+                        <Text size="xs">{expandedProjects.has(project.projectId) ? '▼' : '▶'}</Text>
```
noahsilas, line 100: There's a part of me that wants to suggest using something like a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) disclosure for this, but I'll admit that I'm not entirely sure how well that works with table layouts like this one.  You could probably do something quite nice with css subgrid support fairly widely available, but honestly it seems like more trouble than it's worth for this.

I might experiment a little with this another day, because it seems like a table with collapsible sections and good accessibility support would be a nice reusable component.

packages/app/src/admin/WsSubStatsWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  return (
+    <>
+      <Button onClick={fetchStats} loading={loading}>
+        Get WS Sub Stats
+      </Button>
+      <Modal opened={opened} onClose={close} title="WebSocket Subscription Stats" size="xl" centered>
+        {!projects?.length && <Text>No active WebSocket subscriptions found.</Text>}
+        {projects && projects.length > 0 && (
+          <Table striped highlightOnHover withTableBorder withColumnBorders>
```
noahsilas: The numbers here are small enough that it probably doesn't make a meaningful difference, but you could consider adding the [`tabularNums`](https://mantine.dev/core/table/#tabular-numbers) prop to get the counts in the table to line up nicely.

packages/server/src/fhir/operations/getwssubstats.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  // Scan for all active subscription hash keys
+  const pattern = 'medplum:subscriptions:r4:project:*:active:*';
+  const keys: string[] = [];
+  let cursor = '0';
+  do {
+    const [nextCursor, foundKeys] = await redis.scan(cursor, 'MATCH', pattern, 'COUNT', '1000');
+    cursor = nextCursor;
+    keys.push(...foundKeys);
+  } while (cursor !== '0');
```
noahsilas, line 84: This might be nice to break out into a standalone function like `const keys = await scanAll(redis,pattern)`;

noahsilas (approved):

packages/server/src/fhir/operations/getwssubstats.ts
codyebberson: Good catch @noahsilas 

We actually do already:

```ts
async readReferences<T extends Resource>(references: Reference<T>[]): Promise<(WithId<T> | Error)[]>
```

That would use Redis `mget` to do a batch operation for the `Project` resources.  They're almost guaranteed to be warm in the cache for this flow.

noahsilas: Ah, nice! I glossed over the caching there and saw that the main body of `readReferences` is doing a similar loop over individually awaited promises, but it makes a ton of sense that with a warm cache those inner awaits will be very quick. Nice!

PR #8511, opened 2026-02-24
New Bot init operation

codyebberson:
Context: https://docs.google.com/document/d/142oF-bCcV8IEyvDR5tfLPUY9FCzWUndwgLhBe4W9UXs/edit?tab=t.0

### DIFF
--- a/packages/server/src/fhir/operations/botinit.ts
+++ b/packages/server/src/fhir/operations/botinit.ts
@@ -146,7 +146,9 @@ export async function createBot(
   });
 
   if (executableCode) {
-    await deployBot(repo, bot);
+    // Need to read bot as system repo to get the full bot resource with meta.project for deployBot
+    const fullBot = await systemRepo.readResource<Bot>('Bot', bot.id);
+    await deployBot(repo, fullBot);
   }
 
   return bot;


packages/server/src/fhir/operations/botinit.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    const sourceCodeCheck = await readBinaryContent(repo, extractBinaryId(res2.body.sourceCode.url));
+    expect(sourceCodeCheck).toBe(sourceCode);
+
+    const executableCodeCheck = await readBinaryContent(repo, extractBinaryId(res2.body.executableCode.url));
+    expect(executableCodeCheck).toBe(executableCode);
+  });
+});
+
+function extractBinaryId(url: string): string {
+  return url.replace('http://localhost:8103/storage/', '').split('/')?.[0]; //?.split('?')?.[0] || '';
```
techdavidy: remove comment

packages/server/src/fhir/operations/botinit.ts
```diff
@@ -0,0 +1,160 @@
...
+export interface BotInitParameters {
+  readonly name: string;
+  readonly description?: string;
+  readonly accessPolicy?: Reference<AccessPolicy>;
+  readonly runtimeVersion?: 'awslambda' | 'vmcontext';
+  readonly sourceCode?: Attachment;
+  readonly executableCode?: Attachment;
+}
+
+const defaultBotCode = `import { BotEvent, MedplumClient } from '@medplum/core';
```
techdavidy, line 67: let's add some documentation to never modify or remove parameters from signature.

techdavidy:
@claude please review

codyebberson:
@claude please review

packages/server/src/fhir/operations/botinit.ts
```diff
@@ -0,0 +1,185 @@
...
+  readonly sourceCode?: Attachment;
+  readonly executableCode?: Attachment;
+}
+
+/**
+ * Default bot code to use when creating a new bot without providing source code.
+ * Note that the function signature must match the expected handler signature for the bot runtime.
+ * This function signature is critical for backwards compatibility.
+ */
+const defaultBotCode = `import { BotEvent, MedplumClient } from '@medplum/core';
```
noahsilas, line 67: Not a blocker, as I see that this is just movement from the old file, so this is a bit #unrelated to this commit. 

How would you feel about moving this code into a file named something like `default_bot.template.ts`? Then it could be protected by the linter / type checker, and we could grab it at startup with something like `const defaultBotCode = await fsPromises.readFile('default_bot.template.ts', { encoding: 'utf8' })`.

packages/server/src/fhir/operations/botinit.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    });
+
+    // Next, Alice creates a bot
+    const res2 = await request(app)
+      .post('/fhir/R4/Bot/$init')
+      .set('Authorization', 'Bearer ' + accessToken)
+      .type('json')
+      .send({
+        name: 'Alice personal bot',
+        description: 'Alice bot description',
```
noahsilas, line 90: I was surprised to see this works! I had thought that these attributes would need to be wrapped in a `Parameters` shaped object, but digging deeper I see that when a JSON object without `resourceType: "Parameters"` is passed then the attributes of that object are used as the input parameters. That does help keep these requests nice and tidy. #TIL!

packages/server/src/fhir/operations/botinit.ts
codyebberson: Yes, that is definitely something we should do.

This one is relatively small and tame.  The much worse culprit is the code in `packages/server/src/cloud/aws/deploy.ts`.

The challenge has been that `readFile` call, and relative paths, and backwards compatibility.  There have been a few occasions where we have read local static files to supplement runtime operations.  For example, the CMS 1500 base PNG, or certain TTF font files.  Those all proved to be quite obnoxious to support the full range of runtime contexts:

* Prod docker images
  * In our official images, we run with current working directory as project root
  * We know that others run derivative images with different current working directories
  * So we need to support some degree of absolute path support
* Local dev server
* Jest unit tests

And, perhaps to make matters worse, that got even more complicated with the CommonJS vs ESM split.

With all that said... I believe it should be possible to do this.  Either we need to make some hard calls on backwards compatibility, or we make the investment in helper utilities to robustly handle all of those different scenarios.

packages/server/src/fhir/operations/botinit.test.ts
codyebberson: Yes.... Good call out.  Technically speaking, this way isn't part of the official FHIR spec, and operations should arguably only accept `Parameters` objects, as you say.

In the early early days, we were not spec compliant on that on a number of FHIR operations.  @mattwiller found a clever solution that accepts `Parameters` but also supports backwards compatibility for the simple/flat objects.

As you can imagine, FHIR afficionados prefer `Parameters`, but plenty of customers prefer the simple/flat objects.

PR #8521, opened 2026-02-24
Explicitly check access policy search permission

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -1628,7 +1628,10 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     const expressions: Expression[] = [];
 
     for (const policy of accessPolicy.resource) {
-      if (policy.resourceType === resourceType || policy.resourceType === '*') {
+      if (
+        (policy.resourceType === resourceType || policy.resourceType === '*') &&
+        (!policy.interaction || policy.interaction.includes(AccessPolicyInteraction.SEARCH))
```
mattlong: `addAccessPolicyFilters` gets called via `readResourceFromDatabase` which itself is used in a few different interaction types. Is it correct to always be checking for `AccessPolicyInteraction.SEARCH`? Or does this function need to be parameterized with the interaction type?

PR #8532, opened 2026-02-26
Provider: Perform encounter creation in a transaction

noahsilas:
I'd like to add behavior here where we create a `Slot` resource at the same time as the `Appointment` is created, to prevent future scheduling operations from colliding with these appointments. ($find and $book, for example, use Slot resources for computing availability and don't inspect Appointment resources.)

Moving these initial resource creations into a transaction gives me a good place to hook in to in the future.

### DIFF
--- a/examples/medplum-provider/src/utils/encounter.ts
+++ b/examples/medplum-provider/src/utils/encounter.ts
@@ -4,6 +4,8 @@ import type { MedplumClient, WithId } from '@medplum/core';
 import { createReference, getExtension, getReferenceString, HTTP_HL7_ORG } from '@medplum/core';
 import type {
   Appointment,
+  Bundle,
+  BundleEntry,
   ChargeItem,
   ClinicalImpression,
   Coding,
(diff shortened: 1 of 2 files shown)


examples/medplum-provider/src/utils/encounter.ts
```diff
@@ -4,16 +4,37 @@ import type { MedplumClient, WithId } from '@medplum/core';
...
+function transact<T extends Resource[]>(
+  medplum: MedplumClient,
+  entry: BundleEntryify<T>
+): Promise<TransactionResponseBundle<T>> {
+  return medplum.executeBatch({
+    resourceType: 'Bundle',
+    type: 'transaction',
+    entry,
+  }) as Promise<TransactionResponseBundle<T>>;
+}
```
noahsilas, line 37: I'd love feedback on this. The idea here is that when we are creating resources of a given tuple of types (example: `transact<[Appointment, Encounter]>(...)`), then we can have the type system know the shape of the resulting response bundle will have an `entry` array that is similarly shaped (`[BundleEntry<Appointment>, BundleEntry<Encounter>]`).

In fact, TS can even infer that generic type, making using this tool nice and clean (see below for an example).

The downside, of course, is that there's some fairly ugly type manipulation going on up above...

examples/medplum-provider/src/utils/encounter.ts
```diff
@@ -23,48 +44,73 @@ export async function createEncounter(
...
+      subject: createReference(patient),
+      encounter: { reference: encounterEntry.fullUrl },
+      date: new Date().toISOString(),
+    },
   };
 
-  await medplum.createResource(clinicalImpressionData);
+  const result = await transact(medplum, [appointmentEntry, encounterEntry, clinicalImpressionEntry]);
+
+  const encounter = result.entry[1].resource;
```
noahsilas, line 110: Here's the payoff: `encounter` here has type `Encounter | undefined`, due to the input types to `transact()`.

noahsilas:
Early feedback request: what do folks think about this kind of helper to make working with transaction bundles nicer? Should we build something like this in to core.client?

noahsilas:
Transaction support is not enabled for all projects (it's gated behind the `transaction-bundles` feature flag). I thought for a bit about switching to a "batch" bundle, but without the atomicity of a transaction I think that we lose expressivity of the code without gaining much real benefit. Closing this draft, thank you @reshmakh for the fast feedback!

For more details see: https://www.medplum.com/docs/fhir-datastore/fhir-batch-requests#batches-vs-transactions

PR #8541, opened 2026-02-27
fix(superadmin): split out stats into two operations

ThatOneBro:
Our prod dashboard was too large to fit into one parameter valueString (over 1MB) so we should remove the criteria and their counts from the all projects operation and have a separate operation for lazily fetching the per resource type stats

### DIFF
--- a/packages/app/src/admin/WsSubStatsWidget.tsx
+++ b/packages/app/src/admin/WsSubStatsWidget.tsx
@@ -41,12 +41,13 @@ export function WsSubStatsWidget(): JSX.Element {
   const [expandedResourceTypes, setExpandedResourceTypes] = useState<Set<string>>(new Set());
   // Cache of per-project detail stats (resource types + criteria), keyed by projectId
   const [projectDetails, setProjectDetails] = useState<Map<string, WsSubResourceTypeDetailStats[]>>(new Map());
-  const [loadingProjectId, setLoadingProjectId] = useState<string | undefined>();
+  // Per-project loading state, keyed by projectId
+  const [loadingProjectDetails, setLoadingProjectDetails] = useState<Map<string, boolean>>(new Map());
 
   function fetchStats(): void {
     setLoading(true);
     medplum
-      .post<Parameters>('fhir/R4/$get-ws-sub-stats')
+      .get<Parameters>('fhir/R4/$get-ws-sub-stats')
       .then((params) => {
         const statsStr = params.parameter?.find((p) => p.name === 'stats')?.valueString;
         if (statsStr) {
@@ -75,9 +76,9 @@ export function WsSubStatsWidget(): JSX.Element {
     const key = `${projectId}:${resourceType}`;
     // Lazy-load criteria for this project if not already fetched
     if (!projectDetails.has(projectId)) {
-      setLoadingProjectId(projectId);
+      setLoadingProjectDetails((prev) => new Map(prev).set(projectId, true));
       medplum
-        .post<Parameters>('fhir/R4/$get-ws-sub-project-stats', { projectId })
+        .get<Parameters>(`fhir/R4/$get-ws-sub-project-stats?projectId=${encodeURIComponent(projectId)}`)
         .then((params) => {
           const statsStr = params.parameter?.find((p) => p.name === 'stats')?.valueString;
           if (statsStr) {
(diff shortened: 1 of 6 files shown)


packages/app/src/admin/WsSubStatsWidget.tsx
```diff
@@ -65,6 +73,21 @@ export function WsSubStatsWidget(): JSX.Element {
 
   function toggleResourceType(projectId: string, resourceType: string): void {
     const key = `${projectId}:${resourceType}`;
+    // Lazy-load criteria for this project if not already fetched
+    if (!projectDetails.has(projectId)) {
+      setLoadingProjectId(projectId);
```
noahsilas: This seems likely to get a bit funky if someone hits several toggles in a row - each of these callbacks will be using the same variable to hold loading state. I think that will probably be OK in this particular scenario - only the latest element you triggered will show a loading indicator, but the others might still be loading in the background. I think the worst thing that could happen here would be an earlier request finishing right after you triggered a fetch and clearing the loading state, leaving it unclear that any fetch was ongoing.

If you want to make this more precise, a couple options:
- you could change the loading state to something like `Map<projectId, boolean>`, similar to how the projectDetails cache works. Then each of these requests could manipulate only its own state. Things still could get funky if a given toggle is triggered multiple times...
- You could push each project into its own React component, and control loading of that project's data with something like `useEffect`. (I've also been thinking about some hooks like `useOperation()` that could abstract over some of the trickiness but getting the general case right in a reusable way is sort of a pain...)

Overall, these ideas might be overkill for this admin widget. I think that this probably works well enough as you currently wrote it to move forward 👍

packages/server/src/fhir/routes.ts
```diff
@@ -392,6 +393,7 @@ function initInternalFhirRouter(): FhirRouter {
   router.add('POST', '/$db-schema-diff', dbSchemaDiffHandler);
   router.add('POST', '/$db-invalid-indexes', dbInvalidIndexesHandler);
   router.add('POST', '/$get-ws-sub-stats', getWsSubStatsHandler);
+  router.add('POST', '/$get-ws-sub-project-stats', getWsSubProjectStatsHandler);
```
noahsilas: Would it make sense for this to be a `GET` so that the client can do automatic caching?

packages/server/src/fhir/operations/getwssubprojectstats.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+    const { resourceType } = parsed;
+
+    let criteriaMap = resourceTypeMap.get(resourceType);
+    if (!criteriaMap) {
+      criteriaMap = new Map();
+      resourceTypeMap.set(resourceType, criteriaMap);
+    }
+
+    const entries = await redis.hvals(key);
```
noahsilas: With these results getting quite large, do you know if it is because there are lots of keys or a smallish number of keys with lots of values?

I'm wondering if it would make sense to pipeline more of these requests; roughly:
```
for (const batch of inBatches(keys, 1000)) {
  const pipeline = redis.pipeline();
  batch.forEach((key) => pipeline.hvals(key));
  const resultBatch = await pipeline.exec()
  for (const [_err, hvals] of resultBatch) {
    for (const criteria of hvals) {
      criteriaMap.set(criteria, (criteriaMap.get(criteria) ?? 0) + 1)
    }
  }
}
```

If the number of keys is small, this probably doesn't make a big difference, but if the number of keys is large then this could be helpful: fewer round trips, and we even could fire these off at the same time as the `scan` is happening (maybe even using the scan loop as our batching mechanism?), avoiding constructing a list with all the keys.

packages/server/src/fhir/operations/getwssubstats.test.ts
```diff
@@ -95,10 +95,8 @@ describe('$get-ws-sub-stats', () => {
       const obType = project?.resourceTypes.find((rt) => rt.resourceType === 'Observation');
       expect(obType).toBeDefined();
       expect(obType?.count).toBe(3);
-      expect(obType?.criteria).toHaveLength(2);
-      // Criteria sorted descending by count
-      expect(obType?.criteria[0]).toEqual({ criteria: 'Observation?code=85354-9', count: 2 });
-      expect(obType?.criteria[1]).toEqual({ criteria: 'Observation?status=final', count: 1 });
+      // Criteria not included in summary response
+      expect((obType as any)?.criteria).toBeUndefined();
```
noahsilas:
```suggestion
  expect(obType).not.toHaveProperty('criteria');
```

codyebberson:
@claude please review

PR #8542, opened 2026-02-27
fix(ws): cleaner WS server closing logic

ThatOneBro:
Unlikely this impacts prod, but could maybe solve some of the intermittent failures we see across WebSockets tests throughout the codebase


The-Alchemist (approved):

packages/server/src/ws/agent.test.ts
```diff
@@ -456,6 +457,35 @@ describe('Agent WebSockets', () => {
     });
   });
 
+  test('Logs when updating agent status fails on disconnect', async () => {
+    const errorSpy = jest.spyOn(globalLogger, 'error').mockImplementation(() => undefined);
```
mattlong, line 461: I'm a broken record, but could we get rid of the sleep and instead do
```suggestion
    const errorSpy = jest.spyOn(globalLogger, 'error').mockImplementation((msg) => {
      if (msg === '[Agent]: Failed to update agent status on disconnect') {
        failedUpdate.resolve(undefined);
      }
    });
```

Could guard it with a timeout race instead of the for loop:
```
await Promise.race([failedUpdate.promise, sleep(200)])
```

mattlong (approved):

packages/server/src/ws/agent.test.ts
ThatOneBro: Yeah I think it might be worth doing a cleanup pass to fix all the tests uniformly

PR #8543, opened 2026-02-27
perf(subscription): cache access policy matches

ThatOneBro:
Caches whether the subscription author's ProjectMembership's synthetic access policy is satisfied by the resource in question so it can be reused throughout the loop against all subscriptions for this resource.

### DIFF
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -298,9 +298,9 @@ export async function addSubscriptionJobs(
   logFn(`Evaluate ${subscriptions.length} subscription(s)`);
 
   const wsSubEvents = [] as [string, SubEventsOptions][];
-  // Cache access policy results per author for the duration of this evaluation.
+  // Cache access policy results per (author, channel type) for the duration of this evaluation.
   // Within one addSubscriptionJobs() call, `resource` and `project` are constant,
-  // so the boolean result is identical for all subscriptions sharing the same author.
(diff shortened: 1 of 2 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -323,7 +327,18 @@ export async function addSubscriptionJobs(
       continue;
     }
     if (matches) {
-      if (!(await satisfiesAccessPolicy(resource, project, subscription))) {
+      const authorRef = getReferenceString(subscription.meta?.author as Reference);
+      // Channel type is part of the key because satisfiesAccessPolicy() is hardcoded to always
+      // return `true` for rest-hook subscriptions (see the TODO comment on that function).
+      // Without it, a cached `true` from a rest-hook sub could bypass the real policy check
```
mattwiller, line 333: 👍

packages/server/src/workers/subscription.ts
```diff
@@ -298,6 +298,10 @@ export async function addSubscriptionJobs(
   logFn(`Evaluate ${subscriptions.length} subscription(s)`);
 
   const wsSubEvents = [] as [string, SubEventsOptions][];
+  // Cache access policy results per (author, channel type) for the duration of this evaluation.
+  // Within one addSubscriptionJobs() call, `resource` and `project` are constant,
+  // so the boolean result is identical for all subscriptions sharing the same author AND channel type.
```
mattwiller, line 303: Good comment, this makes sense to me

PR #8545, opened 2026-02-27
fix(getwssubstats): filter out legacy v2 keys

ThatOneBro:
Very briefly we had keys that match our WebSocket subscription per resource type key pattern, `medplum:subscriptions:r4:project:${projectId}:active:v2`, which was causing the operation to throw. This PR filters out those keys so they do not break the operation.

### DIFF
--- a/packages/server/src/fhir/operations/getwssubstats.ts
+++ b/packages/server/src/fhir/operations/getwssubstats.ts
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { allOk } from '@medplum/core';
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/operations/getwssubstats.ts
```diff
@@ -75,6 +75,10 @@ export function parseActiveSubKey(key: string): { projectId: string; resourceTyp
   }
   const projectId = withoutPrefix.slice(0, activeIdx);
   const resourceType = withoutPrefix.slice(activeIdx + ACTIVE_PART.length);
+  // Filter out legacy pre-release keys that used 'v2' instead of a resource type
+  if (resourceType === 'v2') {
```
mattlong: Could we check `isResourceType(resourceType)` instead?

mattlong (approved):

PR #8546, opened 2026-02-27
Configurable background workers

mattlong:
Adds the ability to selectively enable background job workers or disable all. All BullMQ queue objects are always initialized so that jobs can be queued regardless of worker initialization.


mattlong:
@claude take a look

packages/server/src/config/types.ts
```diff
@@ -240,6 +246,30 @@ export interface MedplumExternalAuthConfig {
...
+   * - undefined/omitted: all workers run (backwards compatible default)
+   * - WorkerName[]: only listed workers run. Empty array = no workers (HTTP-only pool).
+   */
+  enabled?: WorkerName[];
+
+  /**
+   * Per-worker BullMQ overrides, merged on top of global `bullmq` config.
+   * Only takes effect for workers that are enabled.
+   */
+  bullmq?: Partial<Record<WorkerName, Partial<MedplumBullmqConfig>>>;
```
codyebberson, line 272: When would this be different than the top level BullMQ?

Oh, I see, so we could have different Redis instances for different queue types? 🤔

packages/server/src/config/types.ts
```diff
@@ -240,6 +246,30 @@ export interface MedplumExternalAuthConfig {
...
+  | 'post-deploy-migration'
+  | 'set-accounts';
+
+export interface MedplumWorkersConfig {
+  /**
+   * Which workers to run on this server instance.
+   * - undefined/omitted: all workers run (backwards compatible default)
+   * - WorkerName[]: only listed workers run. Empty array = no workers (HTTP-only pool).
+   */
+  enabled?: WorkerName[];
```
codyebberson, line 266: I assume what you're going for here is a future state where different worker types could be run on different hosts?

I'm trying to think through the pros/cons of expressivity vs complexity on that 🤔

codyebberson (approved):
lgtm 👍

packages/server/src/config/types.ts
mattlong: I think only [WorkerOptions](https://api.docs.bullmq.io/interfaces/v1.WorkerOptions.html) like `concurrency` would have any effect.

mattlong: So yes, you could theoretically muck with `connection`, but that would be a strange choice.

packages/server/src/config/types.ts
mattlong: Yeah, this is certainly a bit of future-proofing. But since this an advanced configuration option targeted for at-scale deployments, I feel like it is warranted to provide the flexibility.

PR #8548, opened 2026-02-27
feat(superadmin): clear all WS subs operation

ThatOneBro:
This allows us to clear out all WebSocket subscriptions in case of an emergency


codyebberson:
@claude please review.

In particular, how safe or dangerous is the call to `SCAN` here?

PR #8551, opened 2026-02-27
Provider: visualize scheduling availability on calendar 

noahsilas:
When scheduling, it is helpful to be able to see when the configured availability windows are. In this PR, we dim the background of unavailable times to achieve this.

https://github.com/user-attachments/assets/37d0c78e-2c70-4cc6-98d1-8c3ae9c87f03

## Implementation Notes
There's a bunch of work here that is a bit early. As I was prototyping, I was implementing too many things at once, and it's now a rather large pain to go back and separate it out. Room for me to grow with respect to staying focused; that said, I think that these all will be things we want later so I'm trying to ship fast and not focus too much on it.

As a result, this PR looks rather complex. I recommend reviewing commit-by-commit to see the story of how it evolves.

- I was sharing a bunch of scheduling state across peer components (the Calendar pane and the Find pane) and realizing that we will need access in more places later (particularly for multi-actor scheduling), so I hoisted that state into a new `SchedulingContext` wrapper. This encapsulates some of the awkwardness around "wildcard" values that was starting to spread out a little wide, and means we can do the work of interpreting these complex values once and share the results.

- The SchedulingContext is intended to support future work on scheduling parameters coming from multiple resources. Most specifically, we expect to see shared SchedulingParameter extensions stored on ActivityDefinition resources so they can apply to all of your practitioners. This could have been deferred (we only care about one Schedule today).

- There's a bunch of overlap in scheduling parameter type definitions and utility functions between the Server and Provider packages. Maybe more of this can move into core, or some new `@medplum/scheduling` package?

- Understanding the relevant time zone for the availability is somewhat complex, and may not actually be sufficient (what happens if we are considering two resources with different timezones, like a Practitioner in Arizone and a Location in California?). For now, I'm doing the "simple" version, where we choose one Schedule as the "exemplar" and use their time zone if needed. This feels like something that will have to be smarter later.

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/SchedulePage.test.tsx
+++ b/examples/medplum-provider/src/pages/schedule/SchedulePage.test.tsx
@@ -117,7 +117,6 @@ describe('SchedulePage', () => {
       });
     });
 
-
     test('shows a loading state before the schedule resolves, then renders the calendar', async () => {
       let resolveSearch: (schedule: Schedule | undefined) => void;
       medplum.searchOne = vi.fn().mockReturnValue(


techdavidy:
@claude please review

noahsilas:
Closing for now. There's a lot of movement going on here, including some proposed major changes to the availability format (https://github.com/medplum/medplum/issues/8710). We'll revisit this soon, but for now let's keep iterating on the alpha with less conflict.

PR #8554, opened 2026-02-28
feat(subscription): add logs for WS sub lifecycle

ThatOneBro:
Adds logging for our FHIR WebSocket subscription implementation's lifecycle.

Logs for: 
- WebSocket connected
- WebSocket subscription created (includes current count)
- WebSocket subscription bound to
- WebSocket subscription unbound from
- WebSocket subscription deleted
- WebSocket disconnected

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -964,7 +964,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
           subscriptionId: resource.id,
           criteria: resource.criteria,
           user: author,
-          totalSubscriptions: userSubCount,
+          userSubscriptions: userSubCount,
         });
       }
     }
@@ -1380,7 +1380,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
           getLogger().info('[WS] Subscription deleted', {
             subscriptionId: resource.id,
             user: author,
-            totalSubscriptions: userSubCount,
+            userSubscriptions: userSubCount,
           });
         }
       }


packages/server/src/subscriptions/websockets.ts
```diff
@@ -44,7 +44,7 @@ export interface WebSocketSubMetadata {
   criteriaResourceType: string;
 }
 
-export type WebSocketSubToken = MedplumBaseClaims & AdditionalWsBindingClaims;
+export type WebSocketSubToken = MedplumAccessTokenClaims & AdditionalWsBindingClaims;
```
ThatOneBro, line 47: This is just a more accurate type of the token we were already generating from the beginning 🙃

packages/server/src/fhir/repo.ts
```diff
@@ -958,6 +958,15 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         return;
       }
       await setActiveSubscription(project, criteriaResourceType, `Subscription/${resource.id}`, resource.criteria);
+      if (create) {
```
codyebberson, line 961: Makes sense for now. 

The `handleStorage` function is getting a bit oversized.  Mental note that we may want to refactor some of this out into separate helper methods in the future.

packages/server/src/subscriptions/websockets.ts
```diff
@@ -284,7 +284,12 @@ function unsubscribeWsFromAllSubscriptions(ws: WebSocket): void {
 
 export async function handleR4SubscriptionConnection(socket: WebSocket): Promise<void> {
   const redis = getCacheRedis();
+  const socketId = randomUUID();
   let onDisconnect: (() => Promise<void>) | undefined;
+  let userRef: string | undefined;
+  let socketProjectId: string | undefined;
+
+  globalLogger.info('[WS] FHIR subscription socket connected', { socketId });
```
codyebberson, line 292: I'm trying to think if we should use `globalLogger` or `getLogger()` here?

The main advantage of `getLogger()` is that it would attach `projectId`, `profile`, etc to the logs if it's available.

I can't remember if this method would be called in an "authenticated request context" or not.

Based on the lines below, it seems like it's not in an authenticated context, so we need to do that manually?

ThatOneBro: Correct, the WebSocket connection is established via an unauthenticated endpoint because of how WebSockets work in the browser, so a user's identity is not established until you actually bind to a token

I think we could still wrap everything in a context later on but I think that's a bit of a refactor

PR #8480, opened 2026-02-19
Updating the Foomedical chat experience

reshmakh:

PR #8556, opened 2026-02-28
Updating the legacy chat to use threaded inbox in Foomedical.  Also u…

reshmakh:
…pdate default model for spaces


reshmakh:
Fixed in https://github.com/medplum/medplum/pull/8480

PR #8563, opened 2026-03-01
feat(react): add allowPatientSelection prop to ThreadInbox for client facing apps

techdavidy:
- Adds allowPatientSelection prop to ThreadInbox and NewTopicDialog (defaults to false)
- When false, the patient field in the new message dialog is pre-filled from subject and disabled — appropriate for patient-facing apps like Foomedical where the logged-in user is always the patient
- When true, the patient field is an editable search input — appropriate for provider-facing apps where the provider selects which patient to message
- Foomedical's MessagesPage now passes the logged-in patient profile as subject
- Provider's MessagesPage sets allowPatientSelection={true} to retain the existing search behavior

Screenshot 2026-03-01 at 11 59 03 AM


techdavidy:
@claude  please review

PR #8570, opened 2026-03-02
fix(provider): rename page schedule to calendar

techdavidy:
For context, we would use resource names for pages when presenting corresponding resources in the same page.

### DIFF
--- a/examples/medplum-provider/src/App.tsx
+++ b/examples/medplum-provider/src/App.tsx
@@ -90,7 +90,7 @@ export function App(): JSX.Element | null {
                     label: 'Patients',
                     href: '/Patient?_count=20&_fields=name,email,gender&_sort=-_lastUpdated',
                   },
-                  { icon: <IconCalendarEvent />, label: 'Schedule', href: '/schedule' },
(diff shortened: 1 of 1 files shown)


examples/medplum-provider/src/App.tsx
```diff
@@ -203,7 +203,7 @@ export function App(): JSX.Element | null {
               <Route path="Task" element={<TasksPage />} />
               <Route path="Task/:taskId" element={<TasksPage />} />
               <Route path="/onboarding" element={<IntakeFormPage />} />
-              <Route path="/schedule" element={<SchedulePage />} />
+              <Route path="/calendar" element={<SchedulePage />} />
```
noahsilas, line 206: Should we add a redirect from `/schedule` to `/calendar`? Or is the intent here that eventually the `/schedule` route will be used for managing Schedule resources, and we want to break any links to it now instead of having them suddenly switch to a different page later?

codyebberson: I think the intent is that we can reserve `/schedule` and `/Schedule` (I think the router is case insensitive) for the `Schedule` FHIR resource, so probably not redirect?

codyebberson:
@claude please review

Issue #4850, opened 2024-07-11
getExtensionValue doesn't return arrays of value when multiple extension


tbigueres:
Hello, 

My understanding of FHIR Extension, is that if we want to have one extension that acts like an array, we can do something like this : 

`
{
      "extension": [
               {
                        "url":"https://acme.com/fhir/ext/example",
                        "valueString":"Value1"
               }, 
               {
                        "url":"https://acme.com/fhir/ext/example",
                        "valueString":"Value2"
               }, 
               
      ]
`

<img width="884" alt="image" src="https://github.com/medplum/medplum/assets/2139076/3d8bcc8f-4ebc-480a-a91c-7e3a20d9dca2">
https://cloud.google.com/healthcare-api/docs/concepts/fhir-extensions

In order to represent an array of valueString. 
If this is actually the case, I guess it would be awesome that getExtensionValue actually returns an array of value in that case. 
For the moment it looks like it returns only the first value. 

Thanks

PR #8580, opened 2026-03-03
Teach scheduling operations to use HealthcareService scheduling parameters

### LINKED ISSUE #8271: Support shared scheduling parameters in `HealthcareService` resources
In scheduling operations like [$find](https://github.com/medplum/medplum/issues/7686) and [$book](https://github.com/medplum/medplum/issues/7690), scheduling parameters are expected to be loaded from either specific `Schedule` resources or from shared `ActivityDefinition` resources.

When looking up scheduling parameters, priority is given to:
- SchedulingParameters extensions matching a specific `serviceType` provided in a request
- SchedulingParameters extensions on a `Schedule` over extensions on an `ActivityDefinition`

---

- [x] Blocker: adding an `ActivityDefinition.code` search parameter(#8057).

noahsilas:
When setting up scheduling for multiple providers, we want to be able to set the SchedulingParameters up in a single place and have it be shared across other resources, so that we don't have to keep many Schedule extensions in sync.

To that end, you may define these parameters on an HealthcareService whose `type` attribute will be used as the scheduling `serviceType`. `HealthcareService.availableTime`  can be used to set hours of availability.

This has a priority rule that we must follow:

- If a Schedule has any extensions exactly matching a requested service type, only those extensions are used. This lets an individual calendar override the shared rules.

- Next, any extensions from a HealthcareService that match a requested service type are used.

Resolves #8271

### DIFF
--- a/packages/server/src/fhir/operations/book.test.ts
+++ b/packages/server/src/fhir/operations/book.test.ts
@@ -1276,6 +1276,15 @@ describe('Appointment/$book', () => {
           ],
         });
       expect(response30.status).toEqual(400);
+      expect(response30.body).toHaveProperty('issue', [
+        {
+          severity: 'error',
+          code: 'invalid',
+          details: {
+            text: 'No matching scheduling parameters found',
+          },
+        },
+      ]);
     });
   });
 
--- a/packages/server/src/fhir/operations/book.ts
+++ b/packages/server/src/fhir/operations/book.ts
@@ -67,6 +67,18 @@ function assertAllMatch<T>(objects: T[], msg: string): T {
   return first;
 }
 
+function serviceTypeTokens(slots: Slot[]): string[] {
+  const tokenSet = new Set<string>();
+  for (const slot of slots) {
+    for (const concept of slot.serviceType ?? EMPTY) {
+      for (const coding of concept.coding ?? EMPTY) {
+        tokenSet.add(`${coding.system ?? ''}|${coding.code ?? ''}`);
+      }
+    }
+  }
+  return [...tokenSet.values()];
+}
+
 function chooseActiveParameters(
   proposedSlot: Slot,
   parameters: SchedulingParameters[],
@@ -147,14 +159,8 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
 
   // Collect all unique service type codes across all proposed slots, then fetch
   // matching HealthcareService resources in a single query.
-  const allServiceTypes = [
-    ...new Set(
-      proposedSlots
-        .flatMap((slot) => slot.serviceType ?? EMPTY)
-        .flatMap((concept) => concept.coding ?? EMPTY)
-        .map((coding) => `${coding.system ?? ''}|${coding.code ?? ''}`)
-    ),
-  ];
+  const allServiceTypes = serviceTypeTokens(proposedSlots);
+
   const healthcareServices: HealthcareService[] =
     allServiceTypes.length > 0
       ? await ctx.repo.searchResources<HealthcareService>({
@@ -169,7 +175,8 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
     async () => {
       await Promise.all(
         proposedSlots.map(async (proposedSlot, index) => {
-          const schedule = schedules.find((s) => `Schedule/${s.id}` === getReferenceString(proposedSlot.schedule));
+          const scheduleRefString = getReferenceString(proposedSlot.schedule);
+          const schedule = schedules.find((s) => `Schedule/${s.id}` === scheduleRefString);
           invariant(schedule, 'Slot.schedule not loaded');
 
           const actor = actors.find((a) => `${a.resourceType}/${a.id}` === schedule.actor[0].reference);
@@ -181,12 +188,9 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
             );
           }
 
-          const slotServiceTypes = (proposedSlot.serviceType ?? EMPTY)
-            .flatMap((concept) => concept.coding ?? EMPTY)
-            .map((coding) => `${coding.system ?? ''}|${coding.code ?? ''}`);
+          const slotServiceTypes = serviceTypeTokens([proposedSlot]);
 
-          const durationMinutes =
-            (new Date(proposedSlot.end).getTime() - new Date(proposedSlot.start).getTime()) / 60000;
+          const durationMinutes = (Date.parse(proposedSlot.end) - Date.parse(proposedSlot.start)) / 60000;
 
           const parameters = chooseSchedulingParameters(schedule, healthcareServices, slotServiceTypes)
             .map(([params]) => params)


codyebberson:
@claude - please review.  In particular, can you please comment on the merits of using `ActivityDefinition` for this use case?

packages/server/src/fhir/operations/book.ts
```diff
@@ -170,13 +146,31 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
+      : [];
+
   const bufferSlots: Slot[] = [];
 
   const createdResources = await ctx.repo.withTransaction(
     async () => {
       await Promise.all(
         proposedSlots.map(async (proposedSlot, index) => {
-          const schedule = schedules.find((s) => `Schedule/${s.id}` === proposedSlot.schedule.reference);
+          const schedule = schedules.find((s) => `Schedule/${s.id}` === getReferenceString(proposedSlot.schedule));
```
techdavidy: you should be able to do getReferenceString(s)

techdavidy:
based on feedback from Claude. do we care about ActivityDefinition.status ?

noahsilas:
> Technically this all looks reasonable besides a couple of the suggestions Claude made. On the modeling question, it does some like `HealthcareService` is more semantically aligned? Have we already aligned on ActivityDefinition?

Sorry I didn't put a note on here as I started this work, we did a bunch of discussion in #data-modeling and agreed that HealthcareService is a better fit. Lots more detail in the relevant commit now added: https://github.com/medplum/medplum/pull/8580/changes/f5c636e5e2f506937ca1131dc5576b8734d408e9

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -765,7 +749,9 @@ A bariatric surgery requiring surgeon, OR room, and anesthesiologist coordinatio
 <details>
 <summary>ActivityDefinition: Bariatric Surgery</summary>
 
-This ActivityDefinition defines a 120-minute surgical procedure requiring coordination of surgeon, anesthesiologist, and OR room with 45/30-minute buffers and multiple booking limits (2 per day, 8 per week).
+This ActivityDefinition defines a 120-minute surgical procedure requiring coordination of surgeon, anesthesiologist, and OR room.
```
rahul1:
```suggestion
This HealthcareService defines a 120-minute surgical procedure requiring coordination of surgeon, anesthesiologist, and OR room.
```

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -765,7 +749,9 @@ A bariatric surgery requiring surgeon, OR room, and anesthesiologist coordinatio
 <details>
 <summary>ActivityDefinition: Bariatric Surgery</summary>
```
rahul1:
```suggestion
<summary>HealthcareService: Bariatric Surgery</summary>
```

You probably need to modify the mermaid diagram as well

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
```diff
@@ -256,4 +256,281 @@ describe('parseSchedulingParametersExtensions', () => {
...
+      };
+
+      expect(parseSchedulingParametersExtensions(hs)).toMatchObject([
+        {
+          serviceType: [{ coding: [{ code: 'consult', system: 'http://example.com' }] }],
+          availability: [{ dayOfWeek: ['mon', 'tue'], timeOfDay: ['09:00:00'], duration: 480 }],
+          duration: 30,
+          bufferBefore: 0,
+          bufferAfter: 0,
+          alignmentInterval: 60,
```
rahul1, line 283: why does alignment interval default to 60? I would have expected it default to `duration`

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
```diff
@@ -256,4 +256,281 @@ describe('parseSchedulingParametersExtensions', () => {
...
+          { url: 'https://medplum.com/fhir/StructureDefinition/SchedulingParameters', extension: [durationExt] },
+        ],
+      };
+
+      expect(parseSchedulingParametersExtensions(hs)).toMatchObject([
+        { availability: [{ dayOfWeek: [], timeOfDay: ['00:00:00'], duration: 1440 }] },
+      ]);
+    });
+
+    test('availableTime with only startTime (no endTime) is filtered out', () => {
```
rahul1, line 319: if we had a profile for HealthcareService, we could add a constraints that enforce either `allDay` or `startTime` AND `endTime` are are set

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
```diff
@@ -256,4 +256,281 @@ describe('parseSchedulingParametersExtensions', () => {
...
+      // No availableTime on resource, no availability in extension — should not throw
+      const hs: HealthcareService = {
+        resourceType: 'HealthcareService',
+        extension: [
+          { url: 'https://medplum.com/fhir/StructureDefinition/SchedulingParameters', extension: [durationExt] },
+        ],
+      };
+
+      expect(() => parseSchedulingParametersExtensions(hs)).not.toThrow();
+      expect(parseSchedulingParametersExtensions(hs)).toMatchObject([{ availability: [], serviceType: [] }]);
```
rahul1: also check that duration is correct

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
```diff
@@ -256,4 +256,281 @@ describe('parseSchedulingParametersExtensions', () => {
...
+                  },
+                },
+              },
+            ],
+          },
+        ],
+      };
+
+      expect(() => parseSchedulingParametersExtensions(hs)).toThrow(
+        "Scheduling parameter attribute 'availability' is not allowed in this context"
```
rahul1: > "Scheduling parameter attribute 'availability' is not allowed in this context"

Instead of "in this context" can we be more clear ? 

> ...not allowed on HealthcareService

packages/docs/docs/scheduling/defining-availability.md
noahsilas: This is a slightly interesting case - the intent here is to show how an `ActivityDefinition` can define requirements for a service beyond what scheduling understands. In this case, the activity definition says that you need two practitioners (surgeon + anesthesiologist) and a location (OR room). This would inform the input to multi-schedule $find, saying that it should include three schedule references.

This might be too much to put into the docs now, maybe we should remove this example entirely?

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
noahsilas: I think this was a decision I made early on in the absence of a clear spec; defaulting to "starts on the hour" seemed reasonable, and it is well defined for all cases.

We can change this default to match the duration (though I'd want to do so in a separate PR). My concern would be that an alignment for durations that aren't clean divisors of 1 hour require a reference time - an alignment interval of "45 minutes" might suggest times of "8:00, 8:45, 9:30, ..." or might mean "9:00, 9:45, 10:30, ...". The natural reference time to use would be the start of the interval, which we can do, it just makes a bunch of the math trickier.

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
noahsilas: I started exploring what that could look like; I suspect that if we want to define that profile we want to use FSH, which requires introducing an extra build step. (example of what that could look like is here: https://github.com/medplum/medplum/pull/8111/changes#diff-7de550bd4030064bf12304c489c2b48bf1c91c80f8593cd053c0382c8d9da304R54)

The reason I think this is strongly desirable is that when I was exploring this for Schedule resources, the FSH definition for the extension + profile was around 80 lines (https://gist.github.com/noahsilas/06d6b2ae9191c2ba876b8a605e72de86#file-scheduling-parameters-extension-fsh) while the fully instantiated profile + extension was around 6k lines (the other two files in that gist).

In any case though, I think we probably need to assume that some HealthcareService resources that don't comply with our profile could exist, and handle them in some manner. Here I chose to ignore them, but we could also choose to throw loudly?

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
noahsilas: Good idea, thanks :+1:

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
```diff
@@ -106,23 +113,138 @@ function exactlyOne<T>(arr: T[], attribute: string): T {
...
       'duration'
     );
-    const rawAvailability = atLeastOne(
-      extension.extension.filter((ext) => ext.url === 'availability'),
-      'availability'
-    );
+
+    // `availability` is required in Schedule, not allowed in HealthcareService
+    // (where we read from HealthcareService.availableTimes instead)
+    const rawAvailability = extension.extension.filter((ext) => ext.url === 'availability');
```
rahul1, line 234: would using our utility function make sense here? https://www.medplum.com/docs/sdk/core.getextensionvalue

noahsilas: I considered it - the issue is that the helper only returns the first matching extension. So it could help for entries that have at most one value (things like `duration` and `bufferBefore`), but then we would lose the assertions that there are not multiple such entries.

One thought I had was creating a plural version of the utility: something like `getExtensionValues(resource, urls): Extension[]` (potentially resolving issues like #4850). Then we could defer to the broader mechanisms for extracting these values, even when we have multiple entries or constrain the cardinality. 

I chose not to take that route because this particular extension is relatively constrained and typed, so using these array filters keeps an element of type safety that `getExtensionValue` can't match yet (as it returns an `ExtensionValue` value, which means you still have to narrow the type more tightly after getting it to use it as a string or integer or whatever).

So I think that the alternative here could look roughly like:
```
function getExtensionValues(resource: any, ...urls: string[]): ExtensionValue[] { ... };

function isAvailability(value: ExtensionValue): value is Availability { ... }

const rawAvailability = getExtensionValues(extension, 'availability').filter(isAvailability)
```

In some ways this is better - it adds some runtime type guards that I'm sort of cheating on a bit by casting the entire extension at the beginning of this function. My intention when originally writing that was that this extension would be basically guaranteed by being part of a defined profile, but getting all that wired up has proven challenging... (questions like #8111 still persist).

All this is to say that I'm going to consider revisiting this in the future but I don't think I'm going to do it right now.

packages/docs/docs/scheduling/defining-availability.md
rahul1: Yeah, let's remove all the documentation around "requirements" until we actually formalize that a little further. Right now, let's keep it more scoped to what we're thinking about with availability and default.

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
rahul1: Yeah, I understand what you're saying. Feel free to punt on this for now, but maybe let's file a ticket to revisit this in the future. From a sales demo perspective, it's just that it starts to look a little funky for 30 minutes for all appointments to already have a 30-minute gap. It was also trying to get something as a smart, reasonable default for sales

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
rahul1: No, I think this is fine for no - no need to throw.

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
rahul1: Yeah, I think that's fair. In general, I'm just trying to make sure everyone is aware of our utility functions, as they're not well discoverable, but what you said makes sense if it's not a perfect fit

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
noahsilas: For follow up: https://github.com/medplum/medplum/issues/8734

noahsilas:
@claude This has been updated to use HealthcareService instead of ActivityDefinition. Please re-review the latest version.

packages/docs/docs/scheduling/defining-availability.md
noahsilas: Done: https://github.com/medplum/medplum/pull/8580/changes/d9a9e4f57c06c518f91bd75d820debb56ea03be8

packages/server/src/fhir/operations/utils/scheduling-parameters.test.ts
noahsilas: Done: https://github.com/medplum/medplum/pull/8580/changes/a02853ec3e0a47b4de8a90b9a0d8604b3931b112

packages/server/src/fhir/operations/find.ts
```diff
@@ -104,7 +82,18 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
     throw new OperationOutcomeError(badRequest('Search range cannot exceed 31 days'));
   }
 
-  const [schedule, slots] = await Promise.all([
+  const healthcareServiceSearch: Promise<HealthcareService[]> = ctx.repo.searchResources<HealthcareService>({
```
codyebberson, line 84: Is `service-type` a required property?  

Or can we skip this search if `serviceTypes` is empty?

noahsilas: In https://github.com/medplum/medplum/pull/8672 we made service-type a required input. Previously we had thought that it made sense to have a "wildcard" availability that could match, but we felt that introduced behavior that was hard to explain and complicated implementation, so we changed it from optional to required.

codyebberson: Got it.  K, reading more...

packages/server/src/fhir/operations/book.ts
```diff
@@ -164,13 +145,31 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
   assertAllOk(actors, 'Schedule.actor load failed', 'Parameters.parameter[%i].schedule.actor');
 
+  // Collect all unique service type codes across all proposed slots, then fetch
+  // matching HealthcareService resources in a single query.
+  const allServiceTypes = [
+    ...new Set(
+      proposedSlots
+        .flatMap((slot) => slot.serviceType ?? EMPTY)
+        .flatMap((concept) => concept.coding ?? EMPTY)
+        .map((coding) => `${coding.system ?? ''}|${coding.code ?? ''}`)
```
mattwiller: Would it be feasible to condense this down to a single `flatMap()`?

packages/server/src/fhir/operations/book.ts
```diff
@@ -164,13 +145,31 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
+      : [];
+
   const bufferSlots: Slot[] = [];
 
   const createdResources = await ctx.repo.withTransaction(
     async () => {
       await Promise.all(
         proposedSlots.map(async (proposedSlot, index) => {
-          const schedule = schedules.find((s) => `Schedule/${s.id}` === proposedSlot.schedule.reference);
+          const schedule = schedules.find((s) => `Schedule/${s.id}` === getReferenceString(proposedSlot.schedule));
```
mattwiller: Extracting the call to `getReferenceString()` outside this inner closure will save a bit of duplicated work

packages/server/src/fhir/operations/book.ts
```diff
@@ -182,8 +181,16 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
             );
           }
 
-          const extensions = parseSchedulingParametersExtensions(schedule);
-          const parameters = findMatchingSchedulingParameters(extensions, proposedSlot);
+          const slotServiceTypes = (proposedSlot.serviceType ?? EMPTY)
+            .flatMap((concept) => concept.coding ?? EMPTY)
+            .map((coding) => `${coding.system ?? ''}|${coding.code ?? ''}`);
```
mattwiller: This mapping sequence looks the same as the one above: could this bit of logic be extracted into a helper function?

packages/server/src/fhir/operations/book.ts
```diff
@@ -182,8 +181,16 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
+
+          const durationMinutes =
+            (new Date(proposedSlot.end).getTime() - new Date(proposedSlot.start).getTime()) / 60000;
```
mattwiller:
```suggestion
            (Date.parse(proposedSlot.end) - Date.parse(proposedSlot.start)) / 60000;
```

packages/server/src/fhir/operations/book.test.ts
```diff
@@ -1043,8 +1062,225 @@ describe('Appointment/$book', () => {
...
+                schedule: createReference(schedule),
+                serviceType: [serviceType],
+                start: '2026-01-15T14:00:00Z',
+                end: '2026-01-15T14:30:00Z',
+                status: 'free',
+              } satisfies Slot,
+            },
+          ],
+        });
+      expect(response30.status).toEqual(400);
```
mattwiller, line 1278: Consider adding a second assertion about the exact error response, to ensure this fails for the right reason

packages/server/src/fhir/operations/book.ts
noahsilas: Ah, nice, I should have thought of that. Thanks!

packages/server/src/fhir/operations/book.ts
noahsilas: Hm. The challenge here is that we want to access nested values in a form like `Slot.serviceType[*].coding[*]`.

We can shift from these chained flatMaps to a nested style. That might be better for having smaller temporary arrays. On the other hand, I bet that V8 can do work on this kind of tight iteration to make it real fast...

We also could switch to a more imperative style to avoid building up most of the intermediate arrays at all: roughly like this:
```ts
const serviceTokenSet = new Set<string>();
for (const slot of proposedSlots) {
  for (const concept of slot.serviceType) {
    for (const coding of concept.coding) {
      serviceTokenSet.add(`${coding.system ?? ''}|${coding.code ?? ''}`);
    }
  }
}

if (serviceTokenSet.size > 0) {
  await ctx.repo.searchResources({
    resourceType: 'HealthcareService',
    filters: [{
      code: 'service-type',
      operator: Operator.EQUALS,
      value: [...serviceTokenSet.values()].join(',')
    }]
}
```

Would something like that feel better to you? It scores higher on "cognitive complexity", but I think that's misleading as the nested loops here feel quite readable to me.

That's also all written assuming that what you care about here is the number of potentially large intermediate array allocations, if I'm trying to address the wrong problem let me know!

noahsilas: I took a stab at this in https://github.com/medplum/medplum/pull/8580/commits/914a29b477766bfd189e4894998c666e6ba5de4c - let me know if you would like to see what any other approach would look like!

PR #8581, opened 2026-03-03
Developed Healthie patient and document importer

Ljeytl:
We added two new things to the Healthie importer: providers and documents.

## Providers

We now pull the list of active clinicians from Healthie and create two FHIR resources for each one: a `Practitioner` (name, NPI, licenses, specialties) and a `PractitionerRole` (what they do, where). This happens before `patients` are imported, so `patients` can reference their assigned provider.

## Documents

We pull patient documents from Healthie, download the actual file content from Healthie's storage, and save it in Medplum as a `DocumentReference` + Binary. Smart optimization: we skip re-downloading files that haven't changed since the last import.


Ljeytl:
@claude  please review

PR #8582, opened 2026-03-03
Deploy worker service if specified

mattlong:


codyebberson:
@claude please review.  In particular, please validate the the bash expression for checking if the environment variables are set.

ThatOneBro (approved):

PR #8585, opened 2026-03-04
Provider: add "visit" set up after $book

noahsilas:
After using $book to create an Appointment and Slot resources, we end up in a situation where we want to configure the appointment in the same way as the older "visit" set up style. This entails selecting an "encounter class" and a PlanDefinition, and creating a variety of resources.

To this end, we split `createEncounter` into `createAppointment` and `createEncounter` - the older `<CreateVisit>` code path now calls both of these in sequence, while the newer `<AppointmentDetails>` component that has an Appointment from a `$book` operation can call `createEncounter` only.

This will support both the current medplum-provider $find/$book scheduling workflow, and a future workflow where a Patient-centered app will allow a logged in patient to create an appointment via $find/$hold.

The SchedulePage has been updated to reveal the newly created Appointment in the Details pane after creation, allowing the provider to immediately apply a template and create the related encounter when appropriate.

https://github.com/user-attachments/assets/6b3d36d9-ed16-4b05-ad56-60bf5f36c876

### DIFF
--- a/examples/medplum-provider/src/components/schedule/AppointmentDetails.test.tsx
+++ b/examples/medplum-provider/src/components/schedule/AppointmentDetails.test.tsx
@@ -386,6 +386,54 @@ describe('AppointmentDetails', () => {
       expect(createEncounter).not.toHaveBeenCalled();
     });
 
+    test('createEncounter failure shows an error message', async () => {
+      const user = userEvent.setup();
+
+      // Create a PlanDefinition so ResourceInput can find it (searched by `name`)
+      await medplum.createResource({
+        resourceType: 'PlanDefinition',
+        name: 'Test Plan',
+        title: 'Test Plan',
+        status: 'active',
+      });
+
+      const encounterError = new Error('Failed to create encounter');
+      vi.mocked(createEncounter).mockRejectedValue(encounterError);
+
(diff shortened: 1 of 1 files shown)


techdavidy:
@claude please review

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -63,12 +67,54 @@ function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
...
+      return;
+    }
+
+    const encounter = await createEncounter(
+      medplum,
+      encounterClass,
+      patientParticipant,
+      planDefinition,
+      props.appointment
+    );
```
techdavidy: show notification if error

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -63,12 +67,54 @@ function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
...
+
+    navigate(`/Patient/${patientParticipant.id}/Encounter/${encounter.id}`)?.catch(showErrorNotification);
```
techdavidy: we usually just console log for navigate

PR #8590, opened 2026-03-04
docs: WebSocket subscriptions reconnect behavior

rahul1:
## Summary
Add documentation on WebSocket subscription reconnect behavior and lifecycle hooks to catch up on missed events during network disconnections.

## Motivation
Users need clear guidance on how to handle WebSocket subscription reconnection scenarios and ensure they don't display stale data after temporary network interruptions.

## Test plan
- [ ] Review documentation rendering in the docs site
- [ ] Verify code examples are syntactically correct
- [ ] Check that the guidance is clear and actionable

### DIFF
--- a/packages/docs/docs/react/lifecycle-events.mdx
+++ b/packages/docs/docs/react/lifecycle-events.mdx
@@ -41,8 +41,9 @@ const [communications, setCommunications] = useState<Communication[]>([]);
 useSubscription(
   'Communication',
   (bundle: Bundle) => {
-    // Extract the new communication from the incoming bundle
-    // In a subscription bundle, the resource is typically in entry[1]
+    // Subscription bundles have a specific structure:
+    // - entry[0] contains SubscriptionStatus (metadata about the event)
+    // - entry[1] contains the actual resource that triggered the subscription
     const incoming = bundle.entry?.[1]?.resource as Communication | undefined;
     if (!incoming) {
       return;


packages/docs/docs/react/lifecycle-events.mdx
```diff
@@ -19,6 +19,39 @@ These are usually used for triggering rerenders within the context of React, but
...
+```tsx
+useSubscription(
+  'Communication',
+  (bundle: Bundle) => {
+    // Handle incoming messages
+  },
+  {
+    onSubscriptionConnect: useCallback(() => {
+      // Refresh state when subscription reconnects
+      medplum.searchResources('Communication', { _sort: '-_lastUpdated' });
```
ThatOneBro: We might want to show what to with the results from this search
See: 
https://github.com/medplum/medplum/blob/05268f0eb65676f52cc5d4b9f1e57cde93484b42/packages/react/src/chat/BaseChat/BaseChat.tsx#L74

We can also simplify and just show it getting set into a simple useState maybe

We also might want to show this with a `cache: 'no-cache'` so that the search gets the latest for all of those resources (eg. if we missed an update event)

https://github.com/medplum/medplum/blob/05268f0eb65676f52cc5d4b9f1e57cde93484b42/packages/react/src/chat/BaseChat/BaseChat.tsx#L179

I think that's an easy one to mess up

PR #8591, opened 2026-03-04
fix(provider): Use external id in Candid when Candid encounter id is missing

techdavidy:
- Background external ID lookup — when a claim exists but has no Candid encounter ID identifier, silently queries the Candid bot via externalId (the encounter's FHIR ID) to resolve the encounter and surface claim status without user action

- fetchCandidEncounter refactor — extracted into a reusable useCallback that falls back to externalId lookup when no candidEncounterId is present; also called after claim submission so status refreshes immediately

### DIFF
--- a/examples/medplum-provider/src/components/encounter/BillingTab.tsx
+++ b/examples/medplum-provider/src/components/encounter/BillingTab.tsx
@@ -146,7 +146,10 @@ export const BillingTab = (props: BillingTabProps): JSX.Element => {
     }
     const serviceLines = result?.fullEncounter?.serviceLines;
     if (serviceLines?.length) {
-      const totalCents = serviceLines.reduce((sum: number, line: CandidServiceLine) => sum + (line.chargeAmountCents ?? 0), 0);
+      const totalCents = serviceLines.reduce(
+        (sum: number, line: CandidServiceLine) => sum + (line.chargeAmountCents ?? 0),
+        0
+      );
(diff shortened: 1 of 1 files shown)


techdavidy:
@claude please review
1. Adding || backgroundChecking would cause an infinite loop.
2. requires to call the bot with different parameters.

techdavidy:
@claude please check again

techdavidy:
@claude please review

examples/medplum-provider/src/hooks/useEncounterChart.ts
```diff
@@ -78,45 +79,36 @@ export function useEncounterChart(
...
-  }, [medplum, encounterResource]);
-
-  // Fetch clinical impressions related to the encounter
-  const fetchClinicalImpressions = useCallback(async (): Promise<void> => {
-    if (!encounterResource) {
-      return;
+    const enc = encounterResource;
+
+    async function fetchTasks(): Promise<void> {
+      const taskResult = await medplum.searchResources('Task', `encounter=${getReferenceString(enc)}`, {
```
mattlong, line 90: Any concerns if there are more than 20 `Task` that match? Without a sort specified on the search, you could potentially be getting a different result set each time

techdavidy: good catch will extend in a new PR

PR #8592, opened 2026-03-04
Minor speedup to `getTypedPropertyValueWithSchema`

codyebberson:
I don't know if this is worth it, but putting up a PR for discussion.

In my testing, this yields a 10-12% speedup in schema validation.


codyebberson:
@claude please review.  Is the extra code complexity worth the minor speedup?

mattwiller (approved):

Issue #8573, opened 2026-03-02
Only add subscription to active subscriptions after a user binds to the Subscription token


ThatOneBro:
Currently we add a WebSocket subscription to the project's active list as soon as it's created. This leaves a gap where if a user never successfully binds to that Subscription via a token on their WebSocket, that this "active" subscription gets orphaned and will not be cleaned up until its cache entry either expires or if the server notices that the token is already expired.

If we only add it to the active list after a user binds to it, we ensure that the subscription is cleaned up whenever the users WebSocket is disconnected, and makes it more likely that the client will successfully clean up the subscription by unbinding.

ThatOneBro:
Fixed in #8596

PR #8596, opened 2026-03-05
fix(subscription): track sub token expiration, ref sub on bind

### LINKED ISSUE #8567: Cache token expiry with project subscription hash entry
In order to facilitate more effective and more frequent garbage collection of stale subscriptions that were not cleaned up, we can cache the token expiry alongside the subscription criteria

ThatOneBro:
Fixes #8567 and #8573

### DIFF
--- a/packages/server/src/fhir/operations/getwsbindingtoken.test.ts
+++ b/packages/server/src/fhir/operations/getwsbindingtoken.test.ts
@@ -218,7 +218,13 @@ describe('Get WebSocket binding token', () => {
       expect(res2.status).toBe(400);
       expect(res2.body).toMatchObject<OperationOutcome>({
         resourceType: 'OperationOutcome',
-        issue: [{ severity: 'error', code: 'invalid', details: { text: 'Invalid channel type for this operation: rest-hook' } }],
+        issue: [
+          {
+            severity: 'error',
+            code: 'invalid',
+            details: { text: 'Invalid channel type for this operation: rest-hook' },
+          },
+        ],
       });
     }));
 


packages/server/src/subscriptions/websockets.ts
```diff
@@ -337,7 +344,18 @@ export async function handleR4SubscriptionConnection(socket: WebSocket): Promise
       return;
     }
     const cacheEntry = JSON.parse(cacheEntryStr) as CacheEntry<Subscription>;
-    const criteriaResourceType = cacheEntry.resource.criteria.split('?')[0];
+
+    // We can cast here because these criteria are proven to be valid when calling $get-ws-binding-token
+    const criteriaResourceType = cacheEntry.resource.criteria.split('?')[0] as ResourceType;
+    const subRef = `Subscription/${verifiedToken.subscription_id}`;
+    // We know exp is always defined for these tokens
+    const expiration = verifiedToken.exp as number;
```
noahsilas: You could replace this type cast with an invariant assertion:
```
const expiration = verifiedToken.exp;
invariant(expiration); // works if `0` is not a valid expiration value
invariant(typeof expiration === 'number'); // excludes type `undefined`

File: packages/core/src/subscriptions/index.ts [hunk 2]
```diff
@@ -319,7 +319,8 @@ export class SubscriptionManager {
 
     // Get binding token
     const { parameter } = await this.medplum.get<Parameters>(
-      `fhir/R4/Subscription/${subscriptionId}/$get-ws-binding-token`
+      `fhir/R4/Subscription/${subscriptionId}/$get-ws-binding-token`,
+      { cache: 'reload' }
```
noahsilas: This means that this will fill the cache with the result of this GET but will not return a pre-existing cache entry, right? Is there a scenario where a cached token would no longer be valid that you're defending against here?

packages/server/src/pubsub.ts
```diff
@@ -8,28 +10,95 @@ export function publish(channel: string, message: string | Buffer): Promise<numb
 
 // --- Active WebSocket subscription hash helpers ---
 
-function getActiveSubsKey(projectId: string, resourceType: string): string {
-  return `medplum:subscriptions:r4:project:${projectId}:active:${resourceType}`;
```
noahsilas, line 12: Are there any considerations for deployment around cleaning up these keys that have been supplanted by v2 keys? Will these just naturally expire at some point?

packages/server/src/pubsub.ts
```diff
@@ -8,28 +10,95 @@ export function publish(channel: string, message: string | Buffer): Promise<numb
...
 ): Promise<number> {
-  return getPubSubRedis().hset(getActiveSubsKey(projectId, resourceType), subRef, criteria);
+  return getPubSubRedis().hset(getActiveSubsKey(projectId, resourceType), subRef, JSON.stringify(entry));
+}
+
+export function readActiveSubEntries(rawValues: string[]): ActiveSubscriptionEntry[] {
+  if (!rawValues.length) {
+    return [];
+  }
+  return JSON.parse('[' + rawValues.join(',') + ']') as ActiveSubscriptionEntry[];
```
noahsilas, line 67: Interesting strategy of turning all the raw values into a single JSON string to parse. Is the idea here that a single large JSON.parse call is going to be more performant than potentially many smaller JSON.parse steps? Should we be thinking about alternative implementations that could make parsing a known shape faster?

packages/server/src/pubsub.ts
```diff
@@ -50,3 +119,16 @@ export function removeUserActiveWebSocketSubscriptions(authorRef: string, refs:
...
+export async function cleanupActiveSubs(projectId: string, entryMap: ActiveSubscriptionMap): Promise<void> {
+  for (const [resourceType, refs] of getSubRefsByResourceType(entryMap)) {
+    await removeActiveSubscriptions(projectId, resourceType, refs);
+  }
+  try {
+    for (const [authorRef, refs] of getSubRefsByAuthorRef(entryMap)) {
+      await removeUserActiveWebSocketSubscriptions(authorRef, refs);
+    }
+  } catch (_err: unknown) {
+    globalLogger.error('Attempted to remove active sub entries but Redis was already closed');
```
noahsilas: This message seems like it could obscure other types of exception that might get caught/handled here. Is there a way to test if the error caught is the one you are anticipating here?

packages/server/src/pubsub.ts
ThatOneBro: Yes that is the idea, that merging them into one JSON object (albeit an array) incurs less penalty and also makes their allocations more contiguous. I haven't done perf tests on this strategy in a long time but I believe that this is faster especially for potentially thousands of small objects that will be created and removed very quickly.

Ideally wouldn't need to use JSON at all but went with JSON for flexibility in the payload without necessarily having to completely change the key again. Presumably we could make backwards-compatible changes to the payload of the hash entry without worrying about previous entries

packages/server/src/pubsub.ts
ThatOneBro: Good point, I might just make this more generic and log the whole thrown error with this

packages/server/src/pubsub.ts
ThatOneBro: Sadly these will not naturally expire. I think the entire key has to expire together before Redis v7.4? (https://redis.io/docs/latest/commands/hexpire/ was added then)

Which I don't think we can use yet 😢 

We probably want to clean them up though this might require some manual intervention. Don't think it's super pressing though

packages/core/src/subscriptions/index.ts
ThatOneBro: Sorry this is confusing. In our `MedplumClient` this doesn't exactly correlate with the browser caching behavior, and instead correlates to the `MedplumClient` cache itself. Any value like `no-cache` or `reload` will result in a cache purge. Though this really should maybe be `no-cache` to be clearer to the reader.

I'll make that change 👍 

See:  https://github.com/medplum/medplum/blob/main/packages/core/src/client.ts#L3460

codyebberson:
@claude please review

packages/server/src/fhir/repo.ts
```diff
@@ -945,61 +952,19 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
-        const userSubCount = await getUserActiveWebSocketSubscriptionCount(author);
-        getLogger().info('[WS] Subscription created', {
-          subscriptionId: resource.id,
-          criteria: resource.criteria,
-          user: author,
-          userSubscriptions: userSubCount,
-        });
-      }
+    if (resource.resourceType === 'Subscription' && resource.channel?.type === 'websocket' && create) {
+      getLogger().info('[WS] Subscription created', {
```
mattlong, line 956: Should this logging just be moved to wherever the WS is actually created instead of in `handleStorage`?

packages/server/src/fhir/repo.ts
```diff
@@ -847,7 +847,14 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
     // If this is a Subscription resource we are creating, we want to make sure the user is not over their per-user limit
     if (create && result.resourceType === 'Subscription' && result.channel?.type === 'websocket') {
-      await this.checkWebSocketSubscriptionLimit(result);
+      const projectId = result.meta?.project;
+      const author = result.meta?.author?.reference;
+      if (!(projectId && author)) {
+        throw new OperationOutcomeError(badRequest('No project or author connected to the specified Subscription.'));
+      }
+      const project = await this.getProjectById(projectId);
+      invariant(project);
```
mattlong, line 856: Isn't this exactly what `assert` from `node:assert` does?

packages/server/src/fhir/repo.ts
ThatOneBro: This is technically where it's created, it's just that it's not "ref-counted" until we actually bind it to the WebSocket

We want to have logging for both. Although I do agree that it feels like a weird edge case

packages/server/src/fhir/repo.ts
ThatOneBro: cc @noahsilas do we know the difference between `node:assert` and `invariant` is?

PR #8599, opened 2026-03-05
FooMedical Scheduling Demo

noahsilas:
The `<Scheduler>` component was being used as a demo, showing one potential UI for patient scheduling including collecting a Questionnaire response. I'm updating the FooMedical demo app to show how our new Scheduling operations can work.

- I'm removing the Questionnaire collection from the `<Scheduler>` component (as it added complexity to the component)
- I've split out a `fetchSlots` prop from the `schedule` prop, making a bunch of the type checking nicer along the way.
- There's now also `onSelectSlot` that the caller can use to book the slot (or decide to show something like a `<QuestionnaireForm>` if needed as part of the scheduling process).
- Added loading and error states to the component and demo app

This technically represents breaking changes to a component in `@medplum/react`, but the Scheduler was so non-functional that I think we can safely assume that no one was using it outside of the FooMedical demo application.

## Access Control
To use the scheduling operations from FooMedical, the patient needs an access policy with some extra access.
- Read access to 'Schedule'
- Create access to `Appointment`
- Read/Create access to `Slot`

I'm going to make some proposals in documentation later for updates to our suggested default access policy, but I think that we can make these fairly constrained. I'm holding off on making those adjustments for now while I am tightening the minimal options down.

## Screen Recording

This recording was made with network throttling set to "3G" speed to help make the various loading states clearer.

https://github.com/user-attachments/assets/289ab0d0-c08c-44d8-8266-67696fe69d6a

### DIFF
--- a/packages/react/src/Scheduler/Scheduler.tsx
+++ b/packages/react/src/Scheduler/Scheduler.tsx
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Button, Stack, Text } from '@mantine/core';
+import { Button, Loader, Stack, Text } from '@mantine/core';
 import { showNotification } from '@mantine/notifications';
 import type { WithId } from '@medplum/core';
 import { getReferenceString, isReference, isResource, normalizeErrorString } from '@medplum/core';
@@ -135,7 +135,11 @@ export function Scheduler(props: SchedulerProps): JSX.Element | null {
   }, [slots, date]);
 
   if (!slots) {
-    return null;
+    return (
+      <div className={classes.container} data-testid="scheduler">
+        <Loader />
+      </div>
+    );
   }
 
   return (


packages/react/src/Scheduler/Scheduler.tsx
```diff
@@ -82,28 +87,26 @@ export function Scheduler(props: SchedulerProps): JSX.Element | null {
...
+  // If a single Schedule or Reference<Schedule> is provided, set the actor from it
+  useEffect(() => {
+    if (props.schedule && !Array.isArray(props.schedule)) {
+      if (isResource(props.schedule)) {
+        setActor(onlyPractitioner(props.schedule));
+      } else {
+        medplum
+          .readReference<Schedule>(props.schedule)
+          .then((schedule) => setActor(onlyPractitioner(schedule)))
+          .catch(console.error);
```
techdavidy: maybe display the error

noahsilas: Is there a preferred method for that in the shared react package? I could display a Mantine [Alert](https://mantine.dev/core/alert/) inside this component, which doesn't make many assumptions about the containing application. It looks like lots of things in here already use `@mantine/notifications`, so maybe tying in to that system would be better?

noahsilas:
@techdavidy I ended up pushing all the FooMedical follow ups in on top of the cleanup you already did a first review pass of. A bunch more movement in here, but it should all be pretty legible when reviewed as individual commits. Let me know if you have any questions or concerns!

reshmakh:
Related - need to update: https://www.medplum.com/docs/access/access-policies#patient-access

PR #8547, opened 2026-02-27
CDK support for defining background worker only services

mattlong:
Note this is chained from https://github.com/medplum/medplum/pull/8546.

A minimal CDK configuration to pair with this could look like:

```json
"workerServices": [
  {
    "id": "BackgroundJobs",
    "desiredCount": 5
  }
]
```

PR #8603, opened 2026-03-06
Added dispatch worker

codyebberson:
This PR adds a new "Dispatch" BullMQ queue.

Before:
* At the end of every "write" operation we would call `addBackgroundJobs`
* `addBackgroundJobs` would do a fair amount of work to evaluate different jobs
  * Subscriptions: evaluate all subscriptions to see if any match the event, and if so add a job to the "Subscription" queue
  * Download: Look for external URL's that need to be downloaded, and if so add a job to the "Download" queue
  * Cron: Look for any cron jobs that need to be run, and if so add a job to the "Cron" queue
* Historically that was fine

What changed:
* Recently we have seen a significant uptick in use of Websocket subscriptions, which puts quite a bit of extra load on the "write" path
* We moved the BullMQ workers to a separate ECS service, to keep a lot of that work outside of the main HTTP listners (see #8546, #8547, #8582)
* But the "write" path still has to do all the work to evaluate the jobs, even though the actual work is done in a separate service

What this PR does:
* Adds a new "Dispatch" queue, which is responsible for evaluating the jobs and adding them to the appropriate queues
* The "write" path now only has to add a single job to the "Dispatch" queue, and then the "Dispatch" worker does all the work to evaluate the jobs and add them to the appropriate queues
* This should significantly reduce the load on the "write" path, and allow us to scale the "Dispatch" workers independently of the "write" listeners

Open questions:
* To ensure that the Dispatch worker looks at the right version of every changed resource, we're using `versionId` and `Repository.readVersion`.  That call does not use Redis, it goes direct to Postgres.  That is a very fast query, but I wanted to flag that it introduces an extra `SELECT` query on every dispatch execution.
* The bulk of the work here is updating 100's of tests that rely on a specific layout of BullMQ jobs and queue patterns. I'm a little torn about whether it makes sense to keep mocking all of that out, or if we should embrace these are "integration tests", and just let BullMQ run.  That would delete a lot of code, but it would also slow down the tests.

### DIFF
--- a/packages/server/src/workers/subscription.test.ts
+++ b/packages/server/src/workers/subscription.test.ts
@@ -2901,10 +2901,11 @@ describe('Subscription Worker', () => {
 
         // The valid subscription fires a notification — use it to confirm the worker ran
         const nextNotificationPromise = waitForNextSubNotification<Patient>();
-        await wsSubRepo.createResource<Patient>({
+        const patient = await wsSubRepo.createResource<Patient>({
           resourceType: 'Patient',
           name: [{ given: ['Alice'], family: 'Smith' }],
         });
+        await findAndExecDispatchJob(patient, 'create');
         await nextNotificationPromise;
 
         // The expired entry should have been cleaned from the hash
@@ -2967,10 +2968,11 @@ describe('Subscription Worker', () => {
 
         // Trigger the worker; valid sub fires, expired is cleaned up
         const nextNotificationPromise = waitForNextSubNotification<Patient>();
-        await wsSubRepo.createResource<Patient>({
+        const patient = await wsSubRepo.createResource<Patient>({
           resourceType: 'Patient',
           name: [{ given: ['Alice'], family: 'Smith' }],
         });
+        await findAndExecDispatchJob(patient, 'create');
         await nextNotificationPromise;
 
         // Expired entry should be removed from the user active set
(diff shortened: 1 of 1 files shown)


codyebberson:
@claude please review

packages/server/src/fhir/repo.ts
```diff
@@ -1431,7 +1431,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         });
       }
 
-      await addSubscriptionJobs(resource, resource, {
+      await addBackgroundJobs(resource, resource, {
```
mattlong, line 1405: It's not problematic that download and cron jobs will be processed for deletes?

codyebberson: Good question.  Let me confirm.  If nothing else, we can add a guard on `interaction`.

My goal here is to get all of that logic out of the main write path, and centralized in the dispatch job.

codyebberson:
Responding to Claude's questions:

> Question: What happens if these versions are already deleted/purged by the time the dispatch job runs? The PR description mentions this goes direct to Postgres (not Redis), which is good, but:

"Delete" is a soft delete, so you can still read the version by version ID.

"Purge" is a hard delete, and we do not run async jobs after purge.

> Do you have version retention policies that could purge old versions?

No automatic retention policies.

> Should there be explicit error handling for "version not found" scenarios?

This would be a highly unlikely scenario under odd circumstances.
1. User changes a resource, which adds it to the dispatch queue
2. User immediately purges the resource (within ~100 ms)
3. Dispatch queue tries to read the resource, but it is gone

In that case, the dispatch queue job would fail and retry 3 times and then permanently fail.

packages/server/src/fhir/repo.ts
codyebberson: I added a guard on `interaction !== 'delete'` in the dispatch worker.

PR #8606, opened 2026-03-06
Efax integration 

techdavidy:
based on https://github.com/medplum/medplum/pull/8601


techdavidy:
@claude please review

techdavidy:
@claude please review

examples/medplum-provider/src/components/fax/SendFaxModal.tsx
```diff
@@ -0,0 +1,327 @@
...
+        attachment = await medplum.createAttachment({
+          data: file,
+          contentType: file.type,
+          filename: file.name,
+        });
+      } else {
+        return;
+      }
+
+      const recipient = await medplum.createResource<Organization>({
```
finnbergquist: Can we upsert this Organization based on fax number to minimize duplicate Organizations

examples/medplum-provider/src/components/fax/FaxBoard.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  // Poll eFax once on mount, then refresh the list when done
+  useEffect(() => {
+    if (efaxPolledRef.current) {
+      return;
+    }
+    efaxPolledRef.current = true;
+
+    const pollEfax = async (): Promise<void> => {
+      const receiveEfaxUrl = medplum.fhirUrl('Communication', '$receive-efax');
```
finnbergquist, line 97: Should we consider polling for eFaxes outside of FaxBoard, similar to how we poll for DoseSpot notifications?

Say I am on the patient chart but get a new fax. I wouldn't be able to see it unless I click into the faxes page with no prior indication that there is a new inbound fax.

finnbergquist: Alternatively, the better version of this would be to get webhooks working in the eFax integration, and then use in-memory subscriptions to subscribe to inbound eFax Communication create requests.

So maybe it's not worth polling outside of this component. Just a thought tho. Either way, I think this is good enough to make this usable for people

examples/medplum-provider/src/components/fax/SendFaxModal.tsx
```diff
@@ -0,0 +1,364 @@
...
+  defaultPatient,
+  defaultAttachment,
+}: SendFaxModalProps): JSX.Element {
+  const medplum = useMedplum();
+  const profile = useMedplumProfile();
+
+  const [recipientOrg, setRecipientOrg] = useState<Organization | undefined>(undefined);
+  const [recipientName, setRecipientName] = useState('');
+  const [faxNumber, setFaxNumber] = useState('');
+  const [patient, setPatient] = useState<Reference<Patient> | undefined>(defaultPatient);
```
mattlong, line 33: The stale state issue that Claude flagged seems legit?

examples/medplum-provider/src/components/fax/SendFaxModal.tsx
```diff
@@ -0,0 +1,364 @@
...
+  const [subject, setSubject] = useState('');
+  const [coverNote, setCoverNote] = useState('');
+  const [file, setFile] = useState<File | null>(null);
```
mattlong, line 36: Same with the issue flagged for file getting stale.

examples/medplum-provider/src/components/fax/SendFaxModal.tsx
```diff
@@ -0,0 +1,364 @@
...
+        icon: <IconCircleOff />,
+        title: 'Error',
+        message: normalizeErrorString(error),
+      });
+    } finally {
+      setIsSubmitting(false);
+    }
+  };
+
+  const handleClose = (): void => {
```
mattlong, line 218: Maybe the stale state issues could be handled with something like?

```ts
useEffect(() => {
  handleClose();
}, [opened]);

**Comment by techdavidy** [comment]
@claude please check again after we addressed the feedback

**Comment by techdavidy** [comment]
@claude there is no way to 
```
const existing = await medplum.searchOne('Organization', {
  telecom: `fax|${digits}`,
});
```

mattlong (approved):

PR #8617, opened 2026-03-09
Reduce cache lookups in login flow

mattwiller:


packages/server/src/agent/websockets.ts
```diff
@@ -146,13 +145,13 @@ export async function handleAgentConnection(socket: WebSocket, request: Incoming
...
-    const authState = await getLoginForAccessToken(undefined, command.accessToken);
-    if (!authState) {
+    const repo = (await getLoginForAccessToken(undefined, command.accessToken))?.repo;
+    if (!repo) {
       sendError('Invalid access token');
       return;
     }
 
-    const repo = await getRepoForLogin(authState, true);
+    // const repo = await getRepoForLogin(authState, true);
```
ThatOneBro: Cleanup 👀

packages/server/src/agent/websockets.ts
```diff
@@ -200,13 +199,14 @@ export async function handleAgentConnection(socket: WebSocket, request: Incoming
...
-    if (!authState) {
+    const authResult = await getLoginForAccessToken(undefined, command.accessToken);
+    if (!authResult) {
       sendError('Invalid access token');
       return;
     }
+    const { authState, repo } = authResult;
 
-    const repo = await getRepoForLogin(authState, true);
+    // const repo = await getRepoForLogin(authState, true);
```
ThatOneBro: Cleanup 👀

packages/server/src/fhir/accesspolicy.ts
```diff
@@ -46,18 +47,20 @@ export async function getRepoForLogin(authState: AuthState, extendedMode?: boole
   const accessPolicy = await getAccessPolicyForLogin(authState);
 
   const globalSystemRepo = getGlobalSystemRepo();
-  const refs: Reference[] = [membership.project, realMembership.profile];
-  const resolved = await globalSystemRepo.readReferences<Resource>(refs);
-
-  if (resolved[0] instanceof Error) {
-    throw resolved[0];
+  let profile: WithId<ProfileResource | Bot | ClientApplication> | undefined = authState.profile;
```
ThatOneBro, line 50: Is this any reasonable case where this might be stale and we might need to make sure that these references still resolve?

Or at this point is it reasonable that these should always be fresh

mattwiller: For now, it's always the case that the "profile" resource is just the same "user" resource (a `ClientApplication`) that should always be fresh enough at this point.  In most, if not all, cases, it will have just been looked up to be added to the Repository context

ThatOneBro (approved):

PR #8621, opened 2026-03-09
Auto-disable Subscription after repeated failures

mattlong:
A failure is only recorded on the final retry for a given subscription job.


packages/server/src/workers/subscription-failure-tracker.ts
```diff
@@ -0,0 +1,77 @@
...
+  return undefined;
+}
+
+/**
+ * Clears the failure tracking for a subscription (called on success or after auto-disable).
+ * @param subscriptionId - The subscription ID.
+ */
+export async function clearSubscriptionFailures(subscriptionId: string): Promise<void> {
+  const redis = getCacheRedis();
+  await redis.del(getRedisKey(subscriptionId));
```
ThatOneBro: For sets especially it's better to use `unlink` since it does the actual reclamation of memory in a background thread and immediately frees the main thread

packages/server/src/workers/subscription.ts
```diff
@@ -829,6 +839,64 @@ async function catchJobError(subscription: Subscription, job: Job<SubscriptionJo
...
+
+async function autoDisableSubscription(
+  systemRepo: SystemRepository,
+  trigger: SubscriptionAutoDisableTrigger,
+  subscription: WithId<Subscription>,
+  failureCount: number
+): Promise<void> {
+  try {
+    // Re-read to get latest version and check it's still active
+    const current = await systemRepo.readResource<Subscription>('Subscription', subscription.id);
```
ThatOneBro: We could do this all in a patchResource as well, using a test to assert active = true, and then setting active to off and adding the error all in one operation

packages/server/src/config/types.ts
```diff
@@ -164,13 +164,28 @@ export interface MedplumServerConfig {
   /** TOTP authenticator window for MFA token validation (default: 1) */
   mfaAuthenticatorWindow?: number;
 
+  /**
+   * Optional configuration for automatically disabling subscriptions after repeated failures.
+   * Each trigger defines a threshold: if a subscription accumulates `maxConsecutiveFailures`
+   * final failures (after all retries exhausted) within `timeWindowSeconds`, it is set to status "off".
+   * Multiple triggers are evaluated independently — the subscription is disabled if ANY trigger fires.
+   */
+  subscriptionAutoDisable?: SubscriptionAutoDisableTrigger[];
```
rahul1, line 176: following along with our other pattenrs, can we make this overridable on a project level as well?

codyebberson:
@claude please review

PR #8622, opened 2026-03-10
feat(superadmin): add widget for per-user WS sub metrics

ThatOneBro:


ThatOneBro:
Closing this for now, we changed how much info we show for each project in the per-project widget and it's probably enough for most cases. Plus there are some perf implications when it comes to making this widget useful.

PR #8623, opened 2026-03-10
fix(sonar): boost memory for sonar analysis

ThatOneBro:
Fixes an issue with Sonar running out of memory during analysis


codyebberson:
@ThatOneBro  - Any conclusions here?  Is this worth it?

ThatOneBro:
@codyebberson Increasing memory even up to 12GB did not seem to help. Seems like it's something else causing the issue. Specifically something in `packages/server/sec/index.ts` seems to cause the analysis to hang whenever the cache is busted for that file

PR #8628, opened 2026-03-10
Updating sample access policy to support scenarios that are now enabl…

reshmakh:
…ed in patient app

### DIFF
--- a/packages/docs/docs/access/access-policies.md
+++ b/packages/docs/docs/access/access-policies.md
@@ -501,7 +501,20 @@ Binary resources cannot use compartment-based access controls. They require expl
     },
     {
       "resourceType": "Communication",
-      "criteria": "Communication?_compartment=%patient"
+      "criteria": "Communication?_compartment=%patient",
+      "interaction": ["create", "read", "search"]
+    },
+    {
+      "resourceType": "RequestGroup",
+      "criteria": "RequestGroup?_compartment=%patient"
+    },
+    {
+      "resourceType": "Task",
+      "criteria": "Task?focus=%patient"
+    },
+    {
+      "resourceType": "QuestionnaireResponse",
+      "criteria": "QuestionnaireResponse?_compartment=%patient"
     },
     {
       "resourceType": "Subscription",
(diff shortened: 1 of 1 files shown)


packages/docs/docs/access/access-policies.md
```diff
@@ -515,17 +528,52 @@ Binary resources cannot use compartment-based access controls. They require expl
...
     {
       "resourceType": "Slot",
-      "readonly": true
+      "interaction": ["create", "read", "search"],
+      "criteria": "Slot?_compartment=%patient"
     },
     {
-      "resourceType": "Binary"
+      "resourceType": "Slot",
+      "criteria": "Slot?_compartment:not=%patient",
```
finnbergquist: We should thoroughly test the different edge cases for these Slot conditions. Adding multiple AccessPolicy entries with different criteria <> interaction pairs can sometimes cause unexpected behavior. See [this example](https://github.com/medplum/medplum/pull/8521). We don't fully endorse it yet in our docs, but people do it. We are improving it but it remains relatively un-tested, and sometimes can behave differently based on the order of entries in AccessPolicy.

Some thoughts:

1. The Slot resource is not in the [Patient compartment definition](https://hl7.org/fhir/R4/compartmentdefinition-patient.html), so we are relying on AccessPolicy.compartment to tag the Slot with a Patient ref. 
This mechanism for assigning meta.accounts is very brittle, because it can get overwritten easily (on any subsequent update after initial create req). Or, if another User authors the Slot (Bot or Practitioner), then it won't have the Patient ref in the `Slot.meta.compartment` (not 100% sure if this will be an issue without further investigation)

2. I believe that, if the Patient calls $find, then it will use the patient's access. This _"should"_ work because the Patient can theoretically see their own Slots with write permission and other Patient's slots without write permission. We will need to test this as well though.

packages/docs/docs/access/access-policies.md
finnbergquist: We should consider if we still want to use `AccessPolicy.compartment` here.

packages/docs/docs/access/access-policies.md
```diff
@@ -515,17 +528,52 @@ Binary resources cannot use compartment-based access controls. They require expl
...
+      "resourceType": "DocumentReference",
+      "criteria": "DocumentReference?_compartment=%patient"
+    },
+    {
+      "resourceType": "Subscription",
+      "criteria": "Subscription?type=websocket&author=%patient"
+    },
+    {
+      "resourceType": "Appointment",
+      "interaction": ["create"],
```
finnbergquist: Patients will also need to be able to search and read their Appointments

packages/docs/docs/access/access-policies.md
finnbergquist: The reason I bring this up is that Slot Access will be crucial for $find to actually work as intended, so our recommendation for this is essentially a core part of the Scheduling data model

reshmakh: @noahsilas - this is really good feedback, and we need to have some way to test and verify this is working as expected.  TIL: Slots are _not_ in the patient compartment

noahsilas: Interesting - in testing locally I'm seeing Slots assigned into the Patient compartment when a Patient is set as meta.author, but it seems like that is unintentional behavior.

It makes sense that a Slot isn't intended to be part of the patient compartment though - in general they don't have any kind of Patient reference, and the one that exists in my testing in `meta.author` is exactly the one that we're worried about causing accidental leakage.

finnbergquist: @noahsilas @reshmakh removing `%patient` from the `AccessPolicy.compartment` will fix this. The Patient reference will never be included, by default, in the `Slot.meta.compartment`. Only when explicitly assigned via $set-accounts

packages/docs/docs/access/access-policies.md
finnbergquist: See thread [here](https://medplum.slack.com/archives/C02AMUU6SBH/p1773357985852589)

packages/docs/docs/access/access-policies.md
noahsilas: Ah, I understand what was happening now. Thanks for the explanation!

I've also put up a related PR adding a test case that shows the minimal access policy needed for $book to work on behalf of a Patient user. As we update that endpoint to touch other resources (like `HealthcareService`) that test case should start failing and help prompt us to add future entries in here too. https://github.com/medplum/medplum/pull/8700/changes#diff-dadd6a8931ecb168c051093cb8b65917a3e2a4c7c740b786d1ac1b3c590ccaf9R1236-R1262

finnbergquist (approved):

PR #8638, opened 2026-03-11
Add New Messaging Example Code for Documentation

everett-williams:
Messaging example code for documentation

### DIFF
--- a/packages/examples/src/communications/messaging-examples.ts
+++ b/packages/examples/src/communications/messaging-examples.ts
@@ -17,17 +17,17 @@ const profile = { resourceType: 'Practitioner' as const, id: 'example-user-id' }
 // Used in the Thread Lifecycle page to show how to find open threads
 await medplum.searchResources('Communication', {
   'part-of:missing': true,
-  'status:not': 'completed',
+  'status:not': 'completed,entered-in-error,stopped,unknown',
 });
 // end-block filterActiveThreadsTs
 
 /*
 // start-block filterActiveThreadsCli
-medplum get 'Communication?part-of:missing=true&status:not=completed'
+medplum get 'Communication?part-of:missing=true&status:not=completed,entered-in-error,stopped,unknown'
 // end-block filterActiveThreadsCli
 
 // start-block filterActiveThreadsCurl
-curl 'https://api.medplum.com/fhir/R4/Communication?part-of:missing=true&status:not=completed' \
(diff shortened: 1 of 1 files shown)


packages/examples/src/communications/messaging-examples.ts
```diff
@@ -0,0 +1,158 @@
...
+// end-block filterActiveThreadsCli
+
+// start-block filterActiveThreadsCurl
+curl 'https://api.medplum.com/fhir/R4/Communication?part-of:missing=true&status:not=completed' \
+  -H 'authorization: Bearer $ACCESS_TOKEN' \
+  -H 'content-type: application/fhir+json'
+// end-block filterActiveThreadsCurl
+
+// start-block unreadCountTs
+const unreadCount = await medplum.search('Communication', {
```
techdavidy: I think we would use something else besides status for read/unread.

techdavidy:
@claude please review

packages/examples/src/communications/messaging-examples.ts
everett-williams: Thanks. I'm going to update all of this to use tasks

everett-williams:
@claude Please re-review

packages/examples/src/communications/messaging-examples.ts
```diff
@@ -0,0 +1,201 @@
...
+  -H 'content-type: application/fhir+json'
+// end-block filterByPatientCurl
+*/
+
+// start-block filterMyThreadsTs
+// Filter to only the current user's active threads
+await medplum.searchResources('Communication', {
+  'part-of:missing': true,
+  recipient: getReferenceString(profile),
+  'status:not': 'completed',
```
techdavidy: I think we can either say active as: preparation | in-progress | not-done | on-hold | 

or active status is no stopped | completed | entered-in-error | unknown

everett-williams: Hmmm to clarify here - the status on the message thread header is what were using to define active threads. This would not be on an individual message communication resource. So for example, if a thread is still 'in progress', that communication resource (without a payload or part0f) would not have a 'completed' status. Does that make sense?

everett-williams: I don't think the other options would be something that would happen for a thread header communication resource, but please let me know if that is incorrect and we should plan for all of those cases being possible.

everett-williams: Updated to capture edge cases

PR #8644, opened 2026-03-11
fix: make count configurable in ValueSetAutocomplete

### LINKED ISSUE #8639: Configure Count in valueSet
In ValueSetAutocomplete.tsx, the count parameter passed to
  medplum.valueSetExpand() is hardcoded to 10. This should be configurable with
  a default of 10.

https://medplum.slack.com/archives/C0A2DPWBMV5/p1773248040778719

  Current Behavior

  // packages/react/src/ValueSetAutocomplete/ValueSetAutocomplete.tsx:68-76
  const valueSet = await medplum.valueSetExpand(
    {
      ...expandParams,
      url: binding,
      filter: input,
      count: 10,  // hardcoded
    },
    { signal }
  );

  expandParams is already accepted as a prop (Partial<ValueSetExpandParams>),
  but because count: 10 comes after the spread, any count value passed via
  expandParams is always overridden.

Ljeytl:
Move count: 10 before the expandParams spread so callers can override it via expandParams.count. Previously the hardcoded value always won.

Fixes #8639

### DIFF
--- a/packages/react/src/ValueSetAutocomplete/ValueSetAutocomplete.test.tsx
+++ b/packages/react/src/ValueSetAutocomplete/ValueSetAutocomplete.test.tsx
@@ -142,10 +142,7 @@ describe('AsyncAutocomplete', () => {
     const input = screen.getByPlaceholderText<HTMLInputElement>('Test');
     await enterSearchString(input, 'test');
 
-    expect(spy).toHaveBeenCalledWith(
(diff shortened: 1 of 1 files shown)


techdavidy:
@claude please review :D

techdavidy:
maybe let's write a test to catch it

PR #8651, opened 2026-03-12
Scheduling Docs Update

finnbergquist:
- Deleted outdated Scheduling info that was causing confusion (for people and llms)
- Centralized some basic data modeling elements
- Updated docs on the operations we currently support (book, and find)
- Next up: ActivityDefinition -> HealthcareServicce

### DIFF
--- a/packages/docs/docs/scheduling/defining-availability.md
+++ b/packages/docs/docs/scheduling/defining-availability.md
@@ -3,7 +3,7 @@ sidebar_label: Defining Availability (Alpha)
 sidebar_position: 10
 ---
 
-# Defining Availability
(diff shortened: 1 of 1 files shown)


packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -1,62 +1,21 @@
 ---
-sidebar_label: Defining Availability (Alpha)
+sidebar_label: Defining Availability
 ---
 
-# Defining Availability (Alpha)
+# Defining Availability
```
rahul1: Given some of the changes underway, I think this should still be alpha

PR #8653, opened 2026-03-12
Add tenanting prompt to docs

maddyli:


codyebberson (approved):
Minor preference for keeping this kind of content directly in the markdown rather than using the `MedplumCodeBlock`.

The purpose of `MedplumCodeBlock` is to make sure our example code is valid and compiles.  This is just markdown.

Your call though.

Thanks @maddyli ! 🙏 🚀

PR #8657, opened 2026-03-13
feat(react): Optional props to pass to formatting functions in display components

nate-watkins:
Our good friend Opus has assisted with porting over some of the React components to Vue in our codebase. I found adding these optional props that get passed directly to the respective `@medplum/core` formatting functions useful. This could be a maintenance burden if these function signatures are expected to change. Feel free to close of course if so. This follows the pattern established in `<HumanNameDisplay />` with `formatHumanName` options.


codyebberson (approved):

codyebberson:
Thanks @nate-watkins !

PR #8660, opened 2026-03-13
Time Zones and Timestamps docs

codyebberson:


packages/docs/docs/scheduling/timezones.md
```diff
@@ -0,0 +1,182 @@
...
+sidebar_position: 20
+---
+
+# Time Zones and Timestamps
+
+Handling time correctly in distributed systems can be surprisingly complex. Differences in time zones, daylight saving time transitions, and system clock configuration can introduce subtle bugs if timestamps are not handled consistently.
+
+Medplum follows established industry practices and the FHIR specification to ensure that timestamps remain unambiguous, interoperable, and predictable across systems.
+
+This document explains how Medplum handles time zones and timestamps.
```
mattwiller, line 12: I think this line could be removed without much loss of clarity

packages/docs/docs/scheduling/timezones.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Medplum will store this value exactly as written in the resource JSON.
+
+FHIR date and dateTime values use **ISO 8601-compatible formatting**, which allows either:
+
+- a UTC suffix (`Z`)
+- or an explicit time zone offset (`±HH:MM`)
+
+Examples:
```
mattwiller, line 37: It might also be worth adding some language and/or examples — here or in the section below — about "partial" dateTime values, e.g. `2026-02`, and how they're represented in the DB

packages/docs/docs/scheduling/timezones.md
```diff
@@ -0,0 +1,182 @@
...
+| `2026-03-09T09:00:00-05:00` | Local time with offset |
+
+FHIR also includes several different temporal data types:
+
+| Type       | Description                                            |
+| ---------- | ------------------------------------------------------ |
+| `date`     | A calendar date without a time                         |
+| `dateTime` | A date and optional time, possibly with offset         |
+| `instant`  | A precise moment in time (always includes a time zone) |
+| `time`     | A time of day without a date                           |
```
mattwiller, line 91:
```suggestion
| `time`     | A time of day without a date or time zone             |
```

packages/docs/docs/scheduling/timezones.md
```diff
@@ -0,0 +1,182 @@
...
+| `2026-03-09T14:00:00Z`      | UTC                    |
+| `2026-03-09T09:00:00-05:00` | Local time with offset |
+
+FHIR also includes several different temporal data types:
+
+| Type       | Description                                            |
+| ---------- | ------------------------------------------------------ |
+| `date`     | A calendar date without a time                         |
+| `dateTime` | A date and optional time, possibly with offset         |
+| `instant`  | A precise moment in time (always includes a time zone) |
```
mattwiller, line 90:
```suggestion
| `instant`  | A dateTime specified to at least the second precision, with a time zone |
```

packages/docs/docs/scheduling/timezones.md
```diff
@@ -0,0 +1,182 @@
...
+| --------------------------- | ---------------------- |
+| `2026-03-09T14:00:00Z`      | UTC                    |
+| `2026-03-09T09:00:00-05:00` | Local time with offset |
+
+FHIR also includes several different temporal data types:
+
+| Type       | Description                                            |
+| ---------- | ------------------------------------------------------ |
+| `date`     | A calendar date without a time                         |
+| `dateTime` | A date and optional time, possibly with offset         |
```
mattwiller, line 89: If the time is specified, the offset/timezone is actually required by the FHIR spec:
> If hours and minutes are specified, a time zone SHALL be populated
```suggestion
| `dateTime` | A date with optional time and time zone         |
```

PR #8666, opened 2026-03-13
changed so that project name is retrieved dynamically.

Ljeytl:


packages/server/src/auth/utils.ts
```diff
@@ -123,9 +123,25 @@ export async function sendLoginResult(res: Response, login: Login): Promise<void
...
+      uniqueProjectRefs.set(ref.reference, ref);
+    }
+  }
+  const freshProjectMap = new Map<string, Reference<Project>>();
+  await Promise.all(
+    [...uniqueProjectRefs.entries()].map(async ([refStr, ref]) => {
+      const project = await systemRepo.readReference<Project>(ref);
+      freshProjectMap.set(refStr, createReference(project));
+    })
+  );
```
mattwiller: This should ideally use `systemRepo.readReferences()` instead, so all the Projects can be retrieved in a single `MGET` call to Redis

packages/server/src/auth/utils.ts
```diff
@@ -123,9 +123,25 @@ export async function sendLoginResult(res: Response, login: Login): Promise<void
...
+
   const redactedMemberships = memberships.map((m) => ({
     id: m.id,
-    project: m.project,
+    project: freshProjectMap.get(m.project?.reference as string) ?? m.project,
```
mattwiller: Why build `freshProjectMap` if we just need one resolved Project here?

techdavidy:
shouldn't we update the display name to avoid fetching more resources ?

Ljeytl:
> shouldn't we update the display name to avoid fetching more resources ?

probably honestly. but requires us running a one time job on all project names to update right?

packages/server/src/auth/utils.ts
```diff
@@ -123,9 +123,19 @@ export async function sendLoginResult(res: Response, login: Login): Promise<void
   // Safe to rewrite attachments,
   // because we know that these are all resources that the user has access to
   const memberships = await getMembershipsForLogin(login);
+
+  const uniqueRefs = [...new Map(memberships.map((m) => [m.project.reference, m.project])).values()];
+  const projects = await systemRepo.readReferences<Project>(uniqueRefs);
+  const freshProjectMap = new Map<string, Reference<Project>>();
+  for (const project of projects) {
+    if (!(project instanceof Error)) {
+      freshProjectMap.set(`Project/${project.id}`, createReference(project));
```
techdavidy:
```suggestion
      freshProjectMap.set(getReferenceString(project), createReference(project));
```

Issue #8618, opened 2026-03-09
QuestionnaireForm UI Bug


finnbergquist:
## Background

In a Questionnaire, when you define a item of `type=choice `with `repeats: true,` the intended behavior is to allow the selection of multiple options for a single question.

In the QuestionnaireForm component, when you select answers, it starts to show other inputs below:

https://github.com/user-attachments/assets/80e92fc2-384e-4d94-a42f-8828608647e2

Here is example Questionnaire Response for Repro:
```
{
  "resourceType": "Questionnaire",
  "status": "active",
  "item": [
    {
      "linkId": "medications",
      "type": "choice",
      "text": "Medication",
      "repeats": true,
      "answerValueSet": "http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1010.4"
    }
  ]
}
```

_note: the confusing extra items that display in the QuestionnaireForm don't actually get written to the QuestionnaireResponse resource_

techdavidy:
@ianplunkett  @finnbergquist 
```
"repeats": true,
```
is not needed in this case. Due to value set, it will follow the path of dropdown. 

follow the example for multi select 

```
 {
            linkId: 'q5',
            text: 'Question 5 - Value Set Multiselect Dropdown',
            type: 'choice',
            answerValueSet: 'http://loinc.org/vs/LL4436-3',
            extension: [
              {
                url: 'http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl',
                valueCodeableConcept: {
                  coding: [
                    {
                      system: 'http://hl7.org/fhir/questionnaire-item-control',
                      code: 'multi-select',
                      display: 'Multi select',
                    },
                  ],
                  text: 'Multi select',
                },
              },
            ],
          },
```

PR #8674, opened 2026-03-14
fix(react): QuestionnaireForm multiple choice `repeats: true` bug

techdavidy:
solves https://github.com/medplum/medplum/issues/8618


codyebberson:
@claude please review

PR #8678, opened 2026-03-15
Fix nested slice sub-element validation scoping

### LINKED ISSUE #8677: Validator does not scope sub-element validation to matched slice context
## Summary

When a resource property (e.g., `Patient.identifier`) is sliced and a value matches a specific slice, the validator does not use the matched slice's sub-element definitions for nested validation. Instead, it uses the base element's schema, causing required sub-elements from **other** slices to be incorrectly enforced.

## Environment

Medplum v5.0.14 (server + SDK), FHIR R4

## StructureDefinition (simplified)

The NPHIES Patient profile defines identifier slices discriminated by `type`:

```json
{
  "path": "Patient.identifier",
  "slicing": {
    "discriminator": [{"type": "value", "path": "type"}],
    "rules": "closed"
  }
}
```

Two relevant slices:

**passport** (min=0) — requires a `country` extension (min=1):
```json
{
  "path": "Patient.identifier", "sliceName": "passport",
  "min": 0, "max": "1"
}
// passport's type discriminator:
{
  "path": "Patient.identifier.type",
  "patternCodeableConcept": {"coding": [{"system": "http://terminology.hl7.org/CodeSystem/v2-0203", "code": "PPN"}]}
}
// passport's extension — country required:
{
  "path": "Patient.identifier.extension", "min": 1,
  "slicing": {"discriminator": [{"type": "value", "path": "url"}]}
}
{
  "path": "Patient.identifier.extension", "sliceName": "country", "min": 1, "max": "1"
}
```

**nationalid** (min=0) — does NOT require `country` extension:
```json
{
  "path": "Patient.identifier", "sliceName": "nationalid",
  "min": 0, "max": "1"
}
// nationalid's type discriminator:
{
  "path": "Patient.identifier.type",
  "patternCodeableConcept": {"coding": [{"system": "http://terminology.hl7.org/CodeSystem/v2-0203", "code": "NI"}]}
}
// nationalid's extension — NOT required:
{
  "path": "Patient.identifier.extension", "min": 0
}
```

## Test cases

**Test 1: NI identifier without country — should PASS (nationalid doesn't require country)**
```json
{
  "identifier": [{
    "type": {"coding": [{"code": "NI", "system": "http://terminology.hl7.org/CodeSystem/v2-0203"}]},
    "value": "1099999990"
  }]
}
```
Result: **FAILS** — `Incorrect number of values provided for slice 'country': expected 1..1, but found 0`

**Test 2: PPN identifier without country — should FAIL (passport requires country)**
```json
{
  "identifier": [{
    "type": {"coding": [{"code": "PPN", "system": "http://terminology.hl7.org/CodeSystem/v2-0203"}]},
    "value": "A12345"
  }]
}
```
Result: **PASSES** — no validation error

**Test 3: Bogus identifier type — should fail with closed slicing error**
```json
{
  "identifier": [{
    "type": {"coding": [{"code": "BOGUS", "system": "http://example.com"}]},
    "value": "123"
  }]
}
```
Result: **FAILS** with `country` error, not a closed-slicing error

## Root cause

In `visitProperty` (`packages/core/src/typeschema/validation.ts`), after `checkSliceElement` matches a value to a slice, only the slice count is incremented. The recursive crawl of the matched value's children continues using the **base element's schema**, not the matched slice's sub-elements.

```typescript
const sliceName = checkSliceElement(value, element.slicing);
if (sliceName && sliceCounts) {
  sliceCounts[sliceName] += 1;
}
// Children are later crawled using base schema, not slice.elements
```

This means:
- The `country` extension requirement from the `passport` slice is visible in the base schema and applied to ALL identifiers
- The `nationalid` slice's sub-elements (which don't require `country`) are never used
- Test 2 passes because `hasDefaultExtensionSlice` skips extension slice setup entirely, so no extension validation happens

## Suggested fix direction

After `checkSliceElement` matches a slice, the matched `SliceDefinition.elements` should be used for validating that value's sub-properties instead of the base element's schema. The slice elements are already in memory (parsed from the StructureDefinition snapshot).

yingyangcode:
## Summary

When a value matches a slice (e.g., `Patient.identifier` matches the "nationalid" slice), the validator continues using the base element's schema for child property validation instead of the matched slice's sub-elements. This causes required sub-elements from **other** slices to be incorrectly enforced.

Fixes: #8677

## The problem

The NPHIES Patient profile defines identifier slices where only the "passport" slice requires a "country" extension (`min=1`), while "nationalid" does not (`min=0`). Before this fix, the country requirement was applied to all identifiers regardless of which slice they matched.

In `visitProperty`, after `checkSliceElement` matches a value to a slice, only the slice count is incremented. The recursive crawl of the matched value's children continues using the base element's schema, not the matched slice's sub-elements.

## The fix

Added `validateSliceSubElements()` which, after a slice match, validates the matched value's children against the slice-specific element definitions rather than the base schema. The slice elements are already in memory (parsed from the StructureDefinition snapshot), so this is an in-memory lookup swap.

## Test cases from #8677

**NI identifier without country** — should PASS (nationalid doesn't require country):
- Before: FAILS with `country: expected 1..1, found 0`
- After: PASSES

**PPN identifier without country** — should FAIL (passport requires country):
- Before: PASSES (no validation)
- After: Should correctly enforce country requirement

**Bogus identifier type** — should fail with closed slicing error:
- Before: FAILS with `country` error instead of closed-slicing error
- After: Should produce correct error

### DIFF
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -808,6 +808,128 @@ describe('FHIR resource validation', () => {
     });
   });
 
+  describe('Sliced sub-element scoping (#8677)', () => {
+    // Regression tests for https://github.com/medplum/medplum/issues/8677.
+    // When a value matches a slice, only the matched slice's sub-element constraints
+    // should be enforced — sub-element requirements declared on other slices must not
+    // be applied to values that did not match those slices.
+    const slicedIdentifierUrl = 'http://example.com/StructureDefinition/SlicedIdentifierTest';
+    const slicedIdentifierSd: StructureDefinition = {
+      resourceType: 'StructureDefinition',
+      url: slicedIdentifierUrl,
+      name: 'SlicedIdentifierTest',
+      status: 'active',
+      kind: 'logical',
+      abstract: false,
+      type: slicedIdentifierUrl,
+      snapshot: {
+        element: [
+          { path: 'SlicedIdentifierTest' },
+          {
+            path: 'SlicedIdentifierTest.identifier',
+            min: 0,
+            max: '*',
+            base: { path: 'SlicedIdentifierTest.identifier', min: 0, max: '*' },
+            slicing: {
+              discriminator: [{ type: 'value', path: 'system' }],
+              ordered: false,
+              rules: 'open',
+            },
+            type: [{ code: 'Identifier' }],
+          },
+          // "passport" slice — requires both system and value sub-elements
+          {
+            id: 'SlicedIdentifierTest.identifier:passport',
+            path: 'SlicedIdentifierTest.identifier',
+            sliceName: 'passport',
+            min: 0,
+            max: '*',
+            base: { path: 'SlicedIdentifierTest.identifier', min: 0, max: '*' },
+            type: [{ code: 'Identifier' }],
+          },
+          {
+            id: 'SlicedIdentifierTest.identifier:passport.system',
+            path: 'SlicedIdentifierTest.identifier.system',
+            min: 1,
+            max: '1',
+            base: { path: 'Identifier.system', min: 0, max: '1' },
+            type: [{ code: 'uri' }],
+            fixedUri: 'http://example.com/passport',
+          },
+          {
+            id: 'SlicedIdentifierTest.identifier:passport.value',
+            path: 'SlicedIdentifierTest.identifier.value',
+            min: 1,
+            max: '1',
+            base: { path: 'Identifier.value', min: 0, max: '1' },
+            type: [{ code: 'string' }],
+          },
+          // "nationalid" slice — only requires system; deliberately omits the value requirement
+          {
+            id: 'SlicedIdentifierTest.identifier:nationalid',
+            path: 'SlicedIdentifierTest.identifier',
+            sliceName: 'nationalid',
+            min: 0,
+            max: '*',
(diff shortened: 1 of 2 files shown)


mattwiller:
Thank you for the PR @yingyangcode!  A few requests before we'll be able to merge it:
- You'll need to [sign your commit](https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot) to satisfy the DCO check
- It looks like there are a few JSDoc lint errors to be fixed, probably just adding doc lines for function parameters
- Would you mind adding a few unit tests for the new behavior?  It sounds like you already had some test cases, but committing them to `packages/core/src/typeschema/validation.test.ts` will help ensure we don't regress on this change

yingyangcode:
Pushed an update addressing all three review items:

- **DCO**: added `Signed-off-by`.
- **JSDoc**: added `@param` lines for `validateSliceSubElements` (lint passes).
- **Tests**: added a `Sliced sub-element scoping (#8677)` describe block in `packages/core/src/typeschema/validation.test.ts` with four cases covering looser-slice / stricter-slice / mixed-identifier scenarios.

While writing the tests I tightened the implementation: missing-property results from `getNestedProperty` were counted as 1 (so `min=1` checks always passed) and array sub-elements weren't flattened before sub-slice classification. Now skips slice constraints on grandchildren paths (which need hierarchical context like `dataAbsentReason` to evaluate correctly) and counts direct sub-elements correctly. Full validation suite still passes.

PR #8682, opened 2026-03-16
Clear repo commit hooks on transaction rollback

kapetan:
Hi,

Running medplum we encountered some inconsistent state under heavy load when transactions fail because of serialization errors. E.g. subscription jobs running for non-existing resources and resources being returned successfully when fetched using ID but not existing in the database (wrongly cached in redis when transaction in rolled back).

The below is a log entry from the server where a subscription job runs for task `3255ae26-6c50-4392-97aa-11d82d04eaa6` which was never created in the database.

```js
{
  level: 'INFO',
  timestamp: '2026-03-12T15:04:35.319Z',
  msg: 'SubscriptionQueue worker: failed',
  jobId: '15646',
  jobTimestamp: 1773327875311,
  processedOn: 1773327875313,
  queuedDurationMs: 2,
  attemptsMade: 1,
  attemptsStarted: 1,
  subscription: 'Subscription/0b7f346d-cf81-413e-adae-fd8fc6faa36e',
  resource: 'Task/3255ae26-6c50-4392-97aa-11d82d04eaa6',
  versionId: '83a4733f-493b-4014-bf14-6eae067925b3',
  interaction: 'create',
  channelType: 'rest-hook',
  requestId: '90329045-4552-4f11-ac0b-8ebc393d290d',
  traceId: '4e5d18f6-e2de-408b-b63f-57a65821d96e',
  prev: 'active',
  error: 'Not found',
  stack: 'OperationOutcomeError: Not found\n' +
    '    at _Repository.readVersion (file:///usr/src/medplum/packages/server/dist/index.js:36646:15)\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)\n' +
    '    at async execSubscriptionJob (file:///usr/src/medplum/packages/server/dist/index.js:35316:31)\n' +
    '    at async /usr/src/medplum/node_modules/bullmq/dist/cjs/classes/worker.js:565:32'
}
```

The below task is returned when querying `GET /fhir/R4/Task/3255ae26-6c50-4392-97aa-11d82d04eaa6`.

```json
{
  "resourceType": "Task",
  "intent": "order",
  "status": "requested",
  "id": "3255ae26-6c50-4392-97aa-11d82d04eaa6",
  "meta": {
    "versionId": "83a4733f-493b-4014-bf14-6eae067925b3",
    "lastUpdated": "2026-03-12T15:04:35.196Z",
    "author": {
      "reference": "ClientApplication/83e0ed1f-1ba7-474c-9eab-00c91d5f8248"
    },
    "project": "97b7c2ce-1dfa-4edc-9b80-f21a8673e009",
    "compartment": [
      {
        "reference": "Project/97b7c2ce-1dfa-4edc-9b80-f21a8673e009"
      }
    ]
  }
}
```

The root cause is that the registered pre and post commit hooks on a repo instance are not cleared when a transaction fails and are executed if the transaction succeeds after a retry.

This pull-request clears the hooks if the transaction is rolled back. It also ensures that hooks work as expected with nested transactions by introducing a callback stack, otherwise there is a risk that hooks are executed when a nested transaction swallows the transaction error (as far as I can see there are currently no places in the code that do that).

The test case that creates a patient demonstrates how currently a patient that is never created in the database because of a rollback gets pushed to the job queue.


mattlong:
Thanks for flagging this, @kapetan! We'd like to go with an alternative implementation to avoid introducing new classes and having to merge arrays in the happy path:
* Like before, store pre and post callbacks in a single array each
* Track where each transaction level's callbacks begin in a stack, e.g. `{ pre: number, post: number}[]`
* When beginning a savepoint, push the current length of each array
* When releasing a savepoint, pop from the stack
* When rolling back a savepoint, pop from the stack and truncate/splice the arrays

It would be much appreciated if you'd like to implement that 🙏

kapetan:
@mattlong Here is the new implementation following your comments. I think I managed to implement it as you described.

kapetan:
@mattlong Hi, any updates?

mattwiller:
@kapetan It looks like the test case you added is currently failing in CI: would you mind taking a look to see how best to address it?

kapetan:
There was a change in main that broke the test. It should be fixed now.

kapetan:
Hi again, is it possible to get approval to run the tests.

mattlong (approved):
Thanks so much! This looks great!

PR #8684, opened 2026-03-16
docs(communications): Read Receipts & Message Status

everett-williams:
Adds documentation for read receipts and message status using Task resources.

- Read Receipts & Message Status page
- Sidebar and index updates
- Rename to Read Receipts Using Tasks, style updates
- Removes obsolete messaging-data-model and thread-lifecycle docs
- Alternate read-receipt pattern note

All commits are signed off.

### DIFF
--- a/packages/examples/src/communications/messaging-examples.ts
+++ b/packages/examples/src/communications/messaging-examples.ts
@@ -96,25 +96,31 @@ const threadReadStateUrl = 'https://medplum.com/fhir/StructureDefinition/thread-
 const header = await medplum.readResource('Communication', threadHeader.id);
 const currentUserRef = getReferenceString(profile);
 const readStateBlock = header.extension?.find((ext) => {
-  if (ext.url !== threadReadStateUrl || !ext.extension) {return false;}
+  if (ext.url !== threadReadStateUrl || !ext.extension) {
+    return false;
+  }
   const participantExt = ext.extension.find((e: { url?: string }) => e.url === 'participant');
   return (participantExt as { valueReference?: { reference?: string } })?.valueReference?.reference === currentUserRef;
 });
 if (readStateBlock?.extension) {
   const lastReadExt = readStateBlock.extension.find((e: { url?: string }) => e.url === 'lastRead');
   const lastReadAtExt = readStateBlock.extension.find((e: { url?: string }) => e.url === 'lastReadAt');
-  if (lastReadExt)
-    {(lastReadExt as { valueReference?: { reference: string } }).valueReference = {
+  if (lastReadExt) {
+    (lastReadExt as { valueReference?: { reference: string } }).valueReference = {
       reference: `Communication/${latestMessageId}`,
-    };}
-  else
-    {readStateBlock.extension.push({
+    };
+  } else {
+    readStateBlock.extension.push({
       url: 'lastRead',
       valueReference: { reference: `Communication/${latestMessageId}` },
-    });}
+    });
+  }
   const now = new Date().toISOString();
-  if (lastReadAtExt) {(lastReadAtExt as { valueDateTime?: string }).valueDateTime = now;}
-  else {readStateBlock.extension.push({ url: 'lastReadAt', valueDateTime: now });}
+  if (lastReadAtExt) {
+    (lastReadAtExt as { valueDateTime?: string }).valueDateTime = now;
+  } else {
+    readStateBlock.extension.push({ url: 'lastReadAt', valueDateTime: now });
+  }
   await medplum.updateResource(header);
 }
 // end-block updateThreadReadStateTs


everett-williams:
Updated to follow decision guide based on Finn's feedback.

packages/docs/docs/communications/read-receipts-and-message-status.mdx
```diff
@@ -0,0 +1,174 @@
...
+## Decision Guide
+
+1. Can a message have more than one recipient (i.e. more than two total participants in the thread)?
+
+- No → Use [Option A: Simple Communication-based model](#option-a-simple-model-11-threads-only). Searchable and simple; does not support group threads.
+- Yes → Go to question 2.
+
+2. Do you need to search for unread messages via the API?
+
+- No → Use [Option B: Extension on thread header](#option-b-extension-on-thread-header). Not searchable, but works for groups; you can still show a "user up to date" dot and sort by timestamps.
```
finnbergquist: I would leave a note that you can still display a dot to indicate that a User is not up to date in thread, and sort by timestamps. 

Even though this doesn't give true API search capabilities for unread messages, this may be good enough for a lot of people, and it is much simpler

packages/examples/src/communications/messaging-examples.ts
finnbergquist, line 23: Worth explaining why this filters out thread headers.

packages/docs/docs/communications/read-receipts-and-message-status.mdx
```diff
@@ -0,0 +1,174 @@
...
+tags: [communications, messaging, read receipts, task, lifecycle]
+keywords: [read receipt, message status, sent, received, read, unread, Task, extension]
+---
+
+import ExampleCode from '!!raw-loader!@site/../examples/src/communications/messaging-examples.ts';
+import MedplumCodeBlock from '@site/src/components/MedplumCodeBlock';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Read Receipts and Message Status
```
finnbergquist, line 13: Overall, are we assuming that people are using the threaded chat model? https://www.medplum.com/docs/communications/organizing-communications?

If that is not the case, we should note that _option 2 only works with the threaded model._

packages/docs/docs/communications/read-receipts-and-message-status.mdx
```diff
@@ -0,0 +1,174 @@
...
+  </TabItem>
+</Tabs>
+
+---
+
+## Option B: Extension on Thread Header
+
+Use this when you have group threads but do not need to search for unread messages via the API. Store per-participant read state in a custom [extension](https://www.hl7.org/fhir/extensibility.html) on the thread header [`Communication`](/docs/api/fhir/resources/communication) (the one with no `partOf`).
+
+Pros: Works for group threads; no extra resources beyond the thread header; simpler than the Task-based model.
```
finnbergquist, line 89: nice

packages/examples/src/communications/messaging-examples.ts
```diff
@@ -33,6 +36,113 @@ curl 'https://api.medplum.com/fhir/R4/Communication?part-of:missing=true&status:
...
+  status: 'in-progress' as const,
+  subject: { reference: 'Patient/homer-simpson', display: 'Homer Simpson' },
+  topic: { text: 'Lab results - April 10th' },
+  extension: [
+    {
+      url: 'https://medplum.com/fhir/StructureDefinition/thread-read-state',
+      extension: [
+        { url: 'participant', valueReference: { reference: 'Practitioner/doctor-alice-smith' } },
+        { url: 'lastRead', valueReference: { reference: 'Communication/latest-message-id' } },
+      ],
```
finnbergquist, line 84: we probably want to add a timestamp in here too

packages/examples/src/communications/messaging-examples.ts
```diff
@@ -33,6 +36,113 @@ curl 'https://api.medplum.com/fhir/R4/Communication?part-of:missing=true&status:
...
+// Satisfy TS6133 (unused variable); value only used for doc block extraction
+// eslint-disable-next-line no-void
+void threadWithReadState;
+
+// start-block patchThreadReadStateTs
+// Option B: Update a participant's last-read when they view the thread
+await medplum.patchResource('Communication', threadHeader.id, [
+  {
+    op: 'replace',
+    path: '/extension/0/extension/1/valueReference',
```
finnbergquist: Hmmmm, this path is patching the extension [] with an index of 0. We don't know for sure what other extensions people could have on their Communication.

We can't guarantee that the specific extension with the read status is at the 0 index. Instead we need to look for a system url match.

Medplum gives you getExtension() in @medplum/core to find by URL 

getExtension(resource, ...urls)

In this example, I would just use write back with updateResource(communication) (PUT)

packages/docs/docs/communications/read-receipts-and-message-status.mdx
```diff
@@ -0,0 +1,174 @@
...
+## Option C: Task-Based Model (Groups + Searchable)
+
+Use this when you have group threads and need to search for unread messages via the API. Each "message needs to be read by recipient" is represented by a [`Task`](/docs/api/fhir/resources/task) with a dedicated code (e.g. `read-receipt`). Tasks are created when messages are sent and completed when the recipient reads.
+
+Pros: Supports group threads; read receipts are searchable (e.g. unread count, unread per thread).  
+Con: Most complex data model; requires creating and updating Task resources (typically via a [Bot](/docs/bots/bot-basics) on message create and when the user marks as read).
+
+### Create a Read-Receipt Task When a Message is Sent
+
+Typically done in a Bot triggered by a [Subscription](/docs/subscriptions) on [`Communication`](/docs/api/fhir/resources/communication) create. Use `Task.code` to distinguish read-receipt Tasks from other Task types (e.g. response-routing).
```
finnbergquist: I don't know if you need to include this. 

I think it is equally valid to create a Communication and corresponding Tasks all from the client as part of a Transaction Bundle.

finnbergquist: If we want to dive into the whole _Communication Handler_ AKA _Communication Bus_, we should address that somewhere else in a doc. I think that universal Subscription for Communication-> Bot is more necessary when lots of business logic is involved (ex. Task reassignment based on provider role type)

packages/docs/docs/communications/read-receipts-and-message-status.mdx
everett-williams: We are assuming that people will be using the threaded model. I'll be reworking some of the documentation around the data model etc. to make it the approach we recommend. (e.g., "how to create your first thread") will be a page in the flow.

packages/examples/src/communications/messaging-examples.ts
finnbergquist: For the sake of simple examples, I would consider avoiding patches in favor of puts across the board on updates, especially when it is changing a deeply nested fields, but that is very much my opinion

packages/docs/docs/communications/read-receipts-and-message-status.mdx
everett-williams: OK removed this for now, will include a separate page on bot stuff

packages/examples/src/communications/messaging-examples.ts
everett-williams: Updated with the PUT approach

packages/examples/src/communications/messaging-examples.ts
everett-williams: Added

packages/examples/src/communications/messaging-examples.ts
everett-williams: Added explanation

packages/docs/docs/communications/read-receipts-and-message-status.mdx
everett-williams: Added clarification - old note said 'user up to date', changed to 'user not up to date'

packages/examples/src/communications/messaging-examples.ts
```diff
@@ -33,6 +37,138 @@ curl 'https://api.medplum.com/fhir/R4/Communication?part-of:missing=true&status:
...
+        { url: 'lastRead', valueReference: { reference: 'Communication/latest-message-id' } },
+        { url: 'lastReadAt', valueDateTime: '2024-03-15T14:30:00.000Z' },
+      ],
+    },
+  ],
+};
+// end-block threadHeaderWithReadExtensionTs
+// Satisfy TS6133 (unused variable); value only used for doc block extraction
+// eslint-disable-next-line no-void
+void threadWithReadState;
```
mattwiller, line 91: We usually just do e.g. `console.log(threadWithReadState)` to resolve these issue in examples, which would avoid the need for `void`

everett-williams: Ah got it - thanks for letting me know

PR #8686, opened 2026-03-17
Definition resources for marketplace

codyebberson:
Context: https://docs.google.com/document/d/1kb2xuGJ-loRmaPmOZitP-VA0Op3AlUPHtWEaD8MktDs/edit?tab=t.0#heading=h.ljwbe5uixkxa

### DIFF
--- a/packages/definitions/src/fhir/r4/profiles-medplum.json
+++ b/packages/definitions/src/fhir/r4/profiles-medplum.json
@@ -8237,7 +8237,13 @@
               "path" : "PackageRelease.version",
               "min" : 1,
               "max" : "1"
-            }
+            },
+            "constraint" : [{
+              "key": "package-release-version-semver",
+              "severity": "error",
+              "human": "Version must be in semantic versioning format (x.y.z)",
+              "expression": "matches('^(0|[1-9]\\\\d*)\\\\.(0|[1-9]\\\\d*)\\\\.(0|[1-9]\\\\d*)(-[0-9A-Za-z-]+(\\\\.[0-9A-Za-z-]+)*)?(\\\\+[0-9A-Za-z-]+(\\\\.[0-9A-Za-z-]+)*)?$')"
+            }]
           },
           {
             "id" : "PackageRelease.description",
@@ -8487,7 +8493,13 @@
               "path" : "PackageInstallation.version",
               "min" : 1,
               "max" : "1"
-            }
+            },
+            "constraint" : [{
+              "key": "package-installation-version-semver",
+              "severity": "error",
+              "human": "Version must be in semantic versioning format (x.y.z)",
+              "expression": "matches('^(0|[1-9]\\\\d*)\\\\.(0|[1-9]\\\\d*)\\\\.(0|[1-9]\\\\d*)(-[0-9A-Za-z-]+(\\\\.[0-9A-Za-z-]+)*)?(\\\\+[0-9A-Za-z-]+(\\\\.[0-9A-Za-z-]+)*)?$')"
+            }]
           },
           {
             "id" : "PackageInstallation.status",
--- a/packages/server/src/fhir/operations/packageinstall.test.ts
+++ b/packages/server/src/fhir/operations/packageinstall.test.ts
@@ -63,6 +63,24 @@ describe('PackageRelease $install', () => {
     jest.restoreAllMocks();
   });
 
+  test('Require semver version string', async () => {
+    const systemRepo = getGlobalSystemRepo();
+    await expect(async () =>
+      withTestContext(() =>
+        systemRepo.createResource<PackageRelease>({
+          resourceType: 'PackageRelease',
+          meta: { project: project.id },
+          package: { reference: 'Package/' + randomUUID() },
+          version: 'not-a-semver',
+          content: {
+            contentType: ContentType.FHIR_JSON,
+            url: `Binary/${randomUUID()}`,
+          },
+        })
+      )
+    ).rejects.toThrow(/Version must be in semantic versioning format/);
+  });
+
   test('Forbidden for non-admin user', async () => {
     const systemRepo = getGlobalSystemRepo();
     const packageRelease = await withTestContext(() =>


codyebberson:
@claude please review

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -7751,5 +7751,755 @@
...
+      "resourceType" : "StructureDefinition",
+      "id" : "Package",
+      "name" : "Package",
+      "url" : "https://medplum.com/fhir/StructureDefinition/Package",
+      "status" : "active",
+      "kind" : "resource",
+      "abstract" : false,
+      "type" : "Package",
+      "baseDefinition" : "http://hl7.org/fhir/StructureDefinition/DomainResource",
+      "description" : "Package account for automated actions.",
```
noahsilas: This looks like it might have some copy/paste leftover after cloning from the `Bot` definition?

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -7751,5 +7751,755 @@
...
+            }
+          },
+          {
+            "id" : "PackageRelease.content",
+            "path" : "PackageRelease.content",
+            "definition" : "The actual content of the package release.",
+            "min" : 1,
+            "max" : "1",
+            "base" : {
+              "path" : "Bot.content",
```
noahsilas:
```suggestion
              "path" : "PackageRelease.content",
```

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -7751,5 +7751,755 @@
...
+            },
+            "type" : [{
+              "code" : "Extension"
+            }]
+          },
+          {
+            "id" : "PackageInstallation.identifier",
+            "path" : "PackageInstallation.identifier",
+            "short" : "An identifier for this bot",
+            "definition" : "An identifier for this bot.",
```
noahsilas:
```suggestion
            "short" : "An identifier for this package installation",
            "definition" : "An identifier for this package installation.",
```

packages/fhirtypes/dist/Package.d.ts
```diff
@@ -0,0 +1,129 @@
...
+   * resource are required to check for modifier extensions.
+   *
+   * Modifier extensions SHALL NOT change the meaning of any elements on
+   * Resource or DomainResource (including cannot change the meaning of
+   * modifierExtension itself).
+   */
+  modifierExtension?: Extension[];
+
+  /**
+   * An identifier for this bot.
```
noahsilas:
```suggestion
   * An identifier for this Package.
```

packages/fhirtypes/dist/PackageInstallation.d.ts
```diff
@@ -0,0 +1,133 @@
...
+   * resource are required to check for modifier extensions.
+   *
+   * Modifier extensions SHALL NOT change the meaning of any elements on
+   * Resource or DomainResource (including cannot change the meaning of
+   * modifierExtension itself).
+   */
+  modifierExtension?: Extension[];
+
+  /**
+   * An identifier for this bot.
```
noahsilas:
```suggestion
   * An identifier for this package installation.
```

packages/fhirtypes/dist/PackageRelease.d.ts
```diff
@@ -0,0 +1,122 @@
...
+   * resource are required to check for modifier extensions.
+   *
+   * Modifier extensions SHALL NOT change the meaning of any elements on
+   * Resource or DomainResource (including cannot change the meaning of
+   * modifierExtension itself).
+   */
+  modifierExtension?: Extension[];
+
+  /**
+   * An identifier for this bot.
```
noahsilas:
```suggestion
   * An identifier for this package release.
```

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -7751,5 +7751,755 @@
...
+            },
+            "type" : [{
+              "code" : "Extension"
+            }]
+          },
+          {
+            "id" : "Package.identifier",
+            "path" : "Package.identifier",
+            "short" : "An identifier for this bot",
+            "definition" : "An identifier for this bot.",
```
noahsilas:
```suggestion
            "short" : "An identifier for this package",
            "definition" : "An identifier for this package.",
```

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -7751,5 +7751,755 @@
...
+            },
+            "type" : [{
+              "code" : "Extension"
+            }]
+          },
+          {
+            "id" : "PackageRelease.identifier",
+            "path" : "PackageRelease.identifier",
+            "short" : "An identifier for this bot",
+            "definition" : "An identifier for this bot.",
```
noahsilas:
```suggestion
            "short" : "An identifier for this package release",
            "definition" : "An identifier for this package release.",
```

packages/fhirtypes/dist/PackageInstallation.d.ts
```diff
@@ -0,0 +1,133 @@
...
+   */
+  meta?: Meta;
+
+  /**
+   * A reference to a set of rules that were followed when the resource was
+   * constructed, and which must be understood when processing the content.
+   * Often, this is a reference to an implementation guide that defines the
+   * special rules along with other profiles etc.
+   */
+  implicitRules?: string;
```
noahsilas, line 52: On the one hand, it's nice to have such an explicit listing of all the fields in a single type. On the other hand, it can feel like some of these fields are a bit obscuring of the more pertinent differentiators.

This makes me wonder if we should teach our type definitions about some of the shared types, such as `interface PackageInstallation extends DomainResource { ... }`.

Not something we should address in this PR, just a thought I had as I was reviewing this and feeling like there's a lot of boilerplate. Let me know if this seems desirable and I should take a stab at it!

codyebberson:
@claude - please double check your `installedBy` logic.  Note that Medplum only creates database columns for "search parameters", not all individual properties.  At present, `installedBy` is not a searchable property, and therefore it does not get its own column, it only lives in the JSON payload.

packages/server/src/fhir/operations/packageinstall.ts
```diff
@@ -0,0 +1,81 @@
...
+    const json = await readStreamToString(stream);
+    const bundle = JSON.parse(json);
+    getLogger().info('Installing package', {
+      profile: membership.profile,
+      package: packageRelease.package,
+      version: packageRelease.version,
+    });
+    const result = await processBatch(req, repo, router, bundle);
+    await systemRepo.updateResource({ ...installation, status: 'installed' });
+    validateBatchResponse(result);
```
noahsilas: Would it make sense to reorder these two lines so that the validation occurs before we mark the installation as "installed"?

I think this is likely to be low impact, as the error thrown in validation will update the installation to apply `status: "error"`, but it could result in surprising intermediate states, especially if something is subscribed to PackageInstallation updates.

packages/server/src/fhir/operations/packageinstall.test.ts
```diff
@@ -0,0 +1,236 @@
...
+          },
+          request: {
+            method: 'POST',
+            url: 'Patient',
+          },
+        },
+      ],
+    };
+
+    // Create Binary with malformed bundle
```
noahsilas: #nit: it took me a moment to realize that the "malformed" aspect of this bundle is supplied below in the MockBinaryStorage setup. It would be nice to amend this comment to be a little more explicit about that, or maybe colocate that setup with this comment.

noahsilas (commented):
Seems pretty reasonable overall!

I think that one big feature that isn't represented here yet that seems like a natural outgrowth of packages would be some notion of dependencies. I can imagine package authors wanting to rely on things like code systems or being able to invoke bots from other packages. However, as soon as you offer that you get into the real pain points of package managers around dependency management and resolution...

It seems like we can start by introducing a sort of "base" layer for packages that are fully self contained and solve a lot of immediate problems, and leave design space open for future contributors to handle the rich domain of interrelated packages in the future, so I think this makes a lot of sense as-is.

packages/definitions/src/fhir/r4/profiles-medplum.json
codyebberson: Eek, good catch

packages/fhirtypes/dist/PackageInstallation.d.ts
codyebberson: Yeah, I hear you.  This is perhaps another argument for making the leap and doing this with FSH.

packages/server/src/fhir/operations/packageinstall.test.ts
codyebberson: I updated the comment for clarity.

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -7781,5 +7781,755 @@
...
+              "targetProfile" : ["https://medplum.com/fhir/StructureDefinition/Package"]
+            }],
+            "base" : {
+              "path" : "PackageRelease.package",
+              "min" : 1,
+              "max" : "1"
+            }
+          },
+          {
+            "id" : "PackageRelease.version",
```
ThatOneBro, line 8228: Do we want a constraint on the string for version? Seems like we might want to push maintainers to use semver in order to make updating easier for consumers of packages

I know it's early, but I only mention it because it's pretty hard to retroactively enforce such a constraint 😅

codyebberson: Good suggestion...

I can imagine people wanting to do that `-` tag thing (i.e., `0.0.1-alpha`)

What if we changed it to the "id" type, which just enforces this regex: https://hl7.org/fhir/R4/datatypes.html#id

image


Hm, that wouldn't allow for underscores though 🤔

ThatOneBro: SemVer 2.0 has that extension for the metadata tags: https://semver.org/#:~:text=%5E(0%7C%5B1%2D9%5D%5Cd*)%5C.(0%7C%5B1%2D9%5D%5Cd*)%5C.(0%7C%5B1%2D9%5D%5Cd*)(%3F%3A%2D((%3F%3A0%7C%5B1%2D9%5D%5Cd*%7C%5Cd*%5Ba%2DzA%2DZ%2D%5D%5B0%2D9a%2DzA%2DZ%2D%5D*)(%3F%3A%5C.(%3F%3A0%7C%5B1%2D9%5D%5Cd*%7C%5Cd*%5Ba%2DzA%2DZ%2D%5D%5B0%2D9a%2DzA%2DZ%2D%5D*))*))%3F(%3F%3A%5C%2B(%5B0%2D9a%2DzA%2DZ%2D%5D%2B(%3F%3A%5C.%5B0%2D9a%2DzA%2DZ%2D%5D%2B)*))%3F%24

`^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`

Nasty looking regex but should be safe if we limit the length of the string

See the examples of things that work:
https://regex101.com/r/vkijKf/1/

codyebberson: Updated to require semver

PR #8690, opened 2026-03-17
fix(spaces): Bots include requesters info - fix component renderin

techdavidy:
- Update ai-fhir-request-tools to have context of the requester. 
Now it can answer questions like "who am I" or "show my info". Also when creating resources it can list request as owner or actor depending on the resource. 

- Fix conflict between Tooltip from recharts and mantine.


techdavidy:
@claude please review

PR #8696, opened 2026-03-17
docs(communications): add Searching and Querying Threads page

everett-williams:
Adds a new documentation page for searching and querying message threads.

## Changes
- **New page:** `packages/docs/docs/communications/searching-and-querying-threads.mdx` — search query cheat sheet, examples for querying all threads, messages in a thread, loading threads with messages, and filters (by patient, by current user); TypeScript / CLI / cURL tabs; Live updates section with `subscribeToCriteria` and `useSubscription` examples
- **Sidebar:** Added entry under Messaging & Communications
- **Index:** Added link on communications index

### DIFF
--- a/packages/examples/src/communications/messaging-examples.ts
+++ b/packages/examples/src/communications/messaging-examples.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // start-block imports
-import { MedplumClient, getReferenceString, SNOMED } from '@medplum/core';
+import { getReferenceString, MedplumClient, SNOMED } from '@medplum/core';
 import type { Communication } from '@medplum/fhirtypes';
 
 // end-block imports


packages/docs/docs/communications/searching-and-querying-threads.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+# Searching and Querying Threads
+
+## Search Query Cheat Sheet
+
+| Goal                      | Query                                                                  | Notes                         |
+| ------------------------- | ---------------------------------------------------------------------- | ----------------------------- |
+| All threads               | `Communication?part-of:missing=true`                                   | Returns only thread headers   |
+| Messages in a thread      | `Communication?part-of=Communication/{id}&_sort=sent`                  | Sorted chronologically        |
+| Threads with all messages | `Communication?part-of:missing=true&_revinclude=Communication:part-of` | Single request, full contents |
```
finnbergquist: Can you think of a scenario where you would want to get multiple threads and all message contents in those threads in the same request?

I think this would be over fetching. _revinclude is a pretty expensive operation. 

Lets get rid of this example

packages/docs/docs/communications/searching-and-querying-threads.mdx
```diff
@@ -0,0 +1,176 @@
...
+| Threads for a patient     | Add `&subject=Patient/{id}` to any query above                         | Filter by patient context     |
+| My threads                | Add `&recipient=Practitioner/{id}`                                     | Filter by assigned user       |
```
finnbergquist: This should be where I am the recipient **or** the sender

finnbergquist: Double check how provider app does this

packages/docs/docs/communications/searching-and-querying-threads.mdx
```diff
@@ -0,0 +1,176 @@
...
+Communication?part-of=Communication/{threadId}
+```
+
+When a notification arrives, append the new message to your local state. The notification `Bundle` contains two entries: a `SubscriptionStatus` at index 0, and the new `Communication` at index 1.
+
+```ts
+const emitter = medplum.subscribeToCriteria(`Communication?part-of=Communication/${threadId}`);
+
+emitter.addEventListener('message', (event) => {
+  const newMessage = event.payload.entry?.[1]?.resource;
```
finnbergquist: Can you make sure that you get the Communication here and not rely on it being the second entry in the array (ie. at an index of 1)

packages/docs/docs/communications/searching-and-querying-threads.mdx
```diff
@@ -0,0 +1,176 @@
...
+});
+```
+
+If using `@medplum/react`, the [useSubscription](/docs/react/use-subscription) hook handles connection management:
+
+```tsx
+useSubscription(
+  `Communication?part-of=Communication/${id}`,
+  (bundle: Bundle) => {
+    const newMessage = bundle.entry?.[1]?.resource as Communication;
```
finnbergquist: Can you make sure that you get the Communication here and not rely on it being the second entry in the array (ie. at an index of 1)

packages/docs/docs/communications/searching-and-querying-threads.mdx
```diff
@@ -0,0 +1,176 @@
...
+    if (newMessage) {
+      setMessages((prev) => [...prev, newMessage]);
```
finnbergquist, line 165: Maybe add an inline comment here to explain that this is where you can display any FE notifications that a new message was received in addition to displaying the new message

packages/docs/docs/communications/searching-and-querying-threads.mdx
finnbergquist: There isn't an elegant way to do that in a single search

packages/docs/docs/communications/searching-and-querying-threads.mdx
everett-williams: Removed the revinclude example

packages/docs/docs/communications/searching-and-querying-threads.mdx
everett-williams: Got it. Provider filters by recipient and assumes the thread creator is on the header as a recipient. I noted that there isn't a single search you can do, so otherwise you'd need to run two searches and merge client side

packages/docs/docs/communications/searching-and-querying-threads.mdx
everett-williams: Updated

packages/docs/docs/communications/searching-and-querying-threads.mdx
everett-williams: Updated

packages/docs/docs/communications/searching-and-querying-threads.mdx
everett-williams: Added an inline comment on this

packages/docs/docs/communications/searching-and-querying-threads.mdx
finnbergquist: Ahh, ok. 

> thread creator is on the header as a recipient


Last request is to make sure this is part of our recommendation elsewhere

finnbergquist (approved):
Nice! One last small request to make sure we recommend adding sender to Communication.recipient on thread header

mattlong (approved):

PR #8697, opened 2026-03-17
fix(provider): Handle CPT code on sample data

techdavidy:
Modify sample test data to be able to create charge items and do proper billing.


techdavidy:
@claude please review

Issue #7688, opened 2025-10-27
$hold Operation for Appointment Scheduling


rahul1:
## Overview

Implement the FHIR `Appointment/$hold` operation for temporarily reserving appointment slots, based on the [[IHE ITI.Scheduling specification](https://build.fhir.org/ig/IHE/ITI.Scheduling/OperationDefinition-appointment-hold.html)](https://build.fhir.org/ig/IHE/ITI.Scheduling/OperationDefinition-appointment-hold.html) with Medplum-specific enhancements.

## Endpoints

- **`[base]/Appointment/$hold`** - Hold appointment slot(s) temporarily

## MVP Scope

**Version 1 includes:**
- Accept Appointment resource as input
- Create `Slot` resources with `status=busy-tentative`
- Multi-schedule atomic holds
- Transaction-based conflict prevention

**Out of scope for Version 1** (future enhancements):
- Automatic hold expiration/cleanup
- Hold renewal/extension
- Hold notifications/webhooks

### Appointment Resource

**`appointment`** (Appointment, 1..1) - *Medplum Extension*
- Appointment resource from `$find` results
- Must include `start`, `end`, and `participant` with actor references
- Must have a service type with a reference to HealthcareService
- Must include `Slot` resources embedded as `contained` resources. These hold `Schedule` references that will be used to validate availability.
- Consider getting a suitable Appointment from [`/Appointment/$find`](https://www.medplum.com/docs/scheduling/appointment-find)

**Example:**
```json
{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "appointment",
      "resource": {
        "resourceType": "Appointment",
        "status": "proposed",
        "start": "2025-10-28T09:00:00-07:00",
        "end": "2025-10-28T09:30:00-07:00",
        "serviceType": [
          {
            "coding": [{"code": "office-visit"}],
            "extension": [{
               "url": "https://medplum.com/fhir/service-type-reference",
               "valueReference": { "reference": "HealthcareService/office-visit-id" }
            }]
          }
        ],
        "participant": [
          {
            "actor": {"reference": "Practitioner/dr-smith"},
            "required": "required",
            "status": "needs-action"
          }
        ],
        "contained": [
          {
            "resourceType": "Slot",
            "status": "busy",
            "schedule": { "reference": "Schedule/dr-smith-schedule" },
            "start": "2025-10-28T09:00:00-07:00",
            "end": "2025-10-28T09:30:00-07:00"
          }
        ]
      }
[...]

PR #8700, opened 2026-03-17
Add test case for Patient using $book operation

noahsilas:
This is likely to move into $hold when we implement it (see https://github.com/medplum/medplum/issues/7688); for now we have started using $book to demonstrate the flow in FooMedical (see https://github.com/medplum/medplum/pull/8599) and so we would like to add some test coverage supporting that use case.

Implementation notes:
- Adjusted `addTestUser` to accept an options object as its second argument. Previously this took an optional access policy, but switching to an options object lets us add features like `resourceType` in a clear way.

- `addTestUser` tries to log in as the newly created user, which requires a project ID for a user with a `Patient` profile. This means that `Practitioner` users are also now specifying that parameter, but my belief is that this was implicitly looked up from the user's projectMembership in that scenario and this represents no functional change.

### DIFF
--- a/packages/docs/docs/access/access-policies.md
+++ b/packages/docs/docs/access/access-policies.md
@@ -518,6 +518,10 @@ Binary resources cannot use compartment-based access controls. They require expl
       "criteria": "Subscription?type=websocket&author=%profile"
     },
     {
+      "resourceType": "HealthcareService",
+      "readonly": true
+    },
+    {
       "resourceType": "Organization",
       "readonly": true
     },
--- a/packages/server/src/admin/project.test.ts
+++ b/packages/server/src/admin/project.test.ts
@@ -157,7 +157,9 @@ describe('Project Admin routes', () => {
       })
     );
 
-    const bobRegistration = await addTestUser(aliceRegistration.project, { resourceType: 'AccessPolicy' });
+    const bobRegistration = await addTestUser(aliceRegistration.project, {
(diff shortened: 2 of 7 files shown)


noahsilas:
See also: https://github.com/medplum/medplum/pull/8628 which is updating the documentation to include some of the fields in the $book-minimal-access-policy to the documentation, at the point where we recommend a default patient access policy. (There are more changes than just this minimal set in that PR, it seems to be considering experiences across future Patient facing apps more broadly).

codyebberson:
@claude - please review.  It looks like this only touches tests, so should be safe.

The-Alchemist:
Hi, @noahsilas !  I fixed a merge conflict.  Could you make sure it's good?  https://github.com/medplum/medplum/pull/8700/changes/c1a0b21daa7a71b8983c80103a13d053469b7ed2

noahsilas:
> Hi, @noahsilas ! I fixed a merge conflict. Could you make sure it's good? [c1a0b21](https://github.com/medplum/medplum/commit/c1a0b21daa7a71b8983c80103a13d053469b7ed2)

The resolution looks good, but it looks like the test added is now failing. I'll dig into it now.

noahsilas:
@The-Alchemist I've rebased onto latest `main` and updated the test ( the shape of the extension had changed in #8755 so that was a subtle merge conflict that needed resolving).

The-Alchemist (approved):

PR #8708, opened 2026-03-18
Fixes #4365 Fixes #8673 - DomainConfiguration rules on external apps

### LINKED ISSUE #4365: Sign-in to graphiql fails if DomainConfiguration is present
Already in progress - just tracking

### LINKED ISSUE #8673: [bug] DomainConfiguration does not work when logging in from custom domain
## Repro
- DomainConfiguration set to `example.com`  to redirect to an Okta client
- User (admin@example.com) has multiple ProjetMemberships
- Start logging in to [provider.medplum.com](http://provider.medplum.com/)  with `admin@example.com`
- Log into Okta successfully
- we then redirect to [app.medplum.com](http://app.medplum.com/)  to select the specific project membership 

## Background
At [this point in the auth flow](https://github.com/medplum/medplum/blob/3fb7ee94e215ed941a5626fcd4ec57ccb3e5bc68/packages/server/src/auth/external.ts#L115-L150):
we DON’t have
- ClientId
- ProjectId
- Membership

Right now we are always redirected app.medplum.com to sign in

## Expected behavior
To select the project membership, you should be directed to the initiating sign in page

codyebberson:
This PR introduces a new optional field on `DomainConfiguration`:

```
allowedPostLoginRedirectUrls: string[]
```

#### Background

Some enterprise customers configure domain-level SSO policies (e.g. “all `@example.com` users must authenticate via Okta / Entra”), regardless of the `client_id` used.

This works well for Medplum-hosted apps, but creates challenges for custom applications:

* Domain-based IdP routing happens early in the login flow (before `client_id` / `redirect_uri` are known)
* After successful authentication, users need to be returned to the originating application
* Without safeguards, this introduces risk of open redirect vulnerabilities

#### What this change does

This PR allows domain configurations to explicitly define a set of **allowlisted post-login redirect URLs**.

During login:

1. If a domain-based IdP is triggered, the original sign-in URL is preserved in the auth state
2. After successful authentication, Medplum validates the URL against `allowedPostLoginRedirectUrls`
3. If valid, the user is redirected to that URL
4. Otherwise, the user falls back to the default Medplum app flow

#### Security considerations

* Redirects are only allowed to explicitly configured URLs
* Prevents open redirect vulnerabilities
* Does not change OAuth redirect URI validation (IdP callbacks still return to Medplum-owned endpoints)

#### Why this matters

* Enables domain-based SSO for developer-built apps
* Maintains compatibility with existing enterprise SSO configurations
* Provides a safe, explicit mechanism for post-login routing without requiring early `client_id` resolution

### DIFF
--- a/packages/server/src/auth/external.ts
+++ b/packages/server/src/auth/external.ts
@@ -145,7 +145,7 @@ export async function externalCallbackHandler(req: Request, res: Response): Prom
   }
 
   let signInPage: string;
-  if (validateReturnToUrl(body.returnTo, domainConfig)) {
+  if (isValidReturnToUrl(body.returnTo, domainConfig)) {
     // This is the case for external auth with a returnTo URL specified in the state.
     signInPage = body.returnTo;
   } else {
@@ -254,7 +254,7 @@ async function verifyExternalCode(
  * @param domainConfig - The domain configuration for the external auth request.
  * @returns True if the returnTo URL is valid, false otherwise.
  */
-function validateReturnToUrl(
+function isValidReturnToUrl(
   returnTo: string | undefined,
   domainConfig: DomainConfiguration | undefined
 ): returnTo is string {
@@ -264,13 +264,6 @@ function validateReturnToUrl(
 
   try {
     const returnToUrl = new URL(returnTo);
-
-    // Always allow .medplum.com hosts
-    if (returnToUrl.hostname.endsWith('.medplum.com')) {
-      return true;
-    }
-
-    // Check against allowed URLs - compare origin and ensure path starts with allowed path
     return !!domainConfig?.allowedPostLoginRedirectUrls?.some((allowedUrl) => {
       try {
         const allowed = new URL(allowedUrl);
--- a/packages/server/src/auth/external.test.ts
+++ b/packages/server/src/auth/external.test.ts
@@ -578,28 +578,6 @@ describe('External', () => {
     expect(redirect.searchParams.get('login')).toBeTruthy();
   });
 
-  test('returnTo URL to a .medplum.com host is always allowed', async () => {
-    const medplumReturnTo = 'https://app.medplum.com/signin';
-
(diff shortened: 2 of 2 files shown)


codyebberson:
@claude please review

packages/server/src/auth/external.ts
```diff
@@ -233,3 +247,44 @@ async function verifyExternalCode(
...
   }
 }
+
+/**
+ * Determines if the returnTo URL is valid based on the domain configuration.
+ * @param returnTo - The returnTo URL from the external auth state.
+ * @param domainConfig - The domain configuration for the external auth request.
+ * @returns True if the returnTo URL is valid, false otherwise.
+ */
+function validateReturnToUrl(
```
ThatOneBro: Small nit, most of our validation functions that use the `validate` verb are functions that throws a validation error.

Maybe we could call this `isValidReturnToUrl()`?

packages/server/src/auth/external.ts
```diff
@@ -233,3 +247,44 @@ async function verifyExternalCode(
...
+): returnTo is string {
+  if (!returnTo) {
+    return false;
+  }
+
+  try {
+    const returnToUrl = new URL(returnTo);
+
+    // Always allow .medplum.com hosts
+    if (returnToUrl.hostname.endsWith('.medplum.com')) {
```
mattlong: Are we sure we want to add a hardcoded value here?

codyebberson: Yeah, good point, this should probably be a config setting.

ThatOneBro (approved):

packages/server/src/auth/external.ts
codyebberson: Yep, done ✅

packages/server/src/auth/external.ts
codyebberson: Done ✅

PR #8709, opened 2026-03-18
Added SmartAppLaunch.fhirContext

codyebberson:


packages/server/src/oauth/token.ts
```diff
@@ -620,13 +620,15 @@ async function sendTokenResponse(res: Response, login: WithId<Login>, client?: C
...
   let encounter = undefined;
+  let fhirContext = undefined;
 
   if (login.launch) {
     const launch = await systemRepo.readReference(login.launch);
     // Prefer identifier.value if present (for SMART apps that need a specific external identifier),
     // otherwise fall back to the FHIR resource ID
     patient = launch.patient?.identifier?.value ?? resolveId(launch.patient);
     encounter = launch.encounter?.identifier?.value ?? resolveId(launch.encounter);
+    fhirContext = launch.fhirContext;
```
ThatOneBro, line 631: ~I think this is intended to be parsed as `launch.resourcetypetolower` for each resource:
https://build.fhir.org/ig/HL7/smart-app-launch/scopes-and-launch-context.html#:~:text=Any%20SMART%20EHR%20MAY%20extend%20this%20list%20to%20support%20additional%20context.%20When%20specifying%20resource%20types%2C%20convert%20the%20type%20names%20to%20all%20lowercase%20(e.g.%2C%20launch/diagnosticreport~

EDIT: I misunderstood, looks like this is a way to specify what to return for this specific SMART launch outside of the flow itself.

ThatOneBro (approved):

PR #8715, opened 2026-03-18
feat(app): Include payload content tab for Communication 

techdavidy:
In order to allow easy editing of payload content string of Communication resources, we include a new tab to update the string.

### DIFF
--- a/packages/app/src/resource/CommunicationPayloadPage.test.tsx
+++ b/packages/app/src/resource/CommunicationPayloadPage.test.tsx
@@ -103,6 +103,38 @@ describe('CommunicationPayloadPage', () => {
     expect(updated.payload?.[0].contentString).toBe('Updated');
   });
 
+  test('preserves attachment and reference payloads when saving string edits', async () => {
+    const medplum = new MockClient();
+    const communication = await medplum.createResource<Communication>({
+      resourceType: 'Communication',
+      status: 'completed',
+      payload: [
+        { contentString: 'Original text' },
+        { contentAttachment: { url: 'https://example.com/file.pdf', title: 'My PDF' } },
+        { contentReference: { reference: 'DocumentReference/123' } },
+      ],
+    });
+
+    await setup(`/Communication/${communication.id}/payload`, medplum);
+
+    const textarea = await screen.findByPlaceholderText('Enter content...');
+    await act(async () => {
+      fireEvent.change(textarea, { target: { value: 'Updated text' } });
+    });
+
+    await act(async () => {
+      fireEvent.click(screen.getByRole('button', { name: 'Save' }));
+    });
+
+    expect(await screen.findAllByText('Saved')).not.toHaveLength(0);
+
+    const updated = await medplum.readResource('Communication', communication.id);
+    expect(updated.payload).toHaveLength(3);
+    expect(updated.payload?.[0].contentString).toBe('Updated text');
+    expect(updated.payload?.[1].contentAttachment).toEqual({ url: 'https://example.com/file.pdf', title: 'My PDF' });
+    expect(updated.payload?.[2].contentReference).toEqual({ reference: 'DocumentReference/123' });
+  });
+
   test('adds a new text payload when no payload is available', async () => {
     const medplum = new MockClient();
     const communication = await medplum.createResource<Communication>({


packages/app/src/resource/CommunicationPayloadPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export function CommunicationPayloadPage(): JSX.Element | null {
+  const medplum = useMedplum();
+  const { resourceType, id } = useParams() as { resourceType: ResourceType; id: string };
+  const communication = useResource<Communication>({ reference: resourceType + '/' + id });
+  const [payload, setPayload] = useState<CommunicationPayload[]>([]);
+
+  useEffect(() => {
+    if (communication) {
+      setPayload((communication.payload ?? EMPTY).filter((item) => 'contentString' in item));
```
noahsilas: It seems like if this is used on a payload that includes any attachments or references, those would be invisible in the UI and then removed if the viewer saves the resource. That feels a little unsafe - how would you feel about keeping any entries without a `contentString` key in the state? Even if you simply skipped rendering them, we could preserve existing entries, although it might be best to render a placeholder even if it just says "An attachment" or something similar.

packages/react-hooks/src/useThreadInbox/useThreadInbox.ts
```diff
@@ -42,7 +42,7 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
 
   const fetchAllCommunications = useCallback(async (): Promise<void> => {
     const searchParams = new URLSearchParams(query);
-    searchParams.append('identifier:not', 'ai-message-topic');
+    searchParams.append('identifier:not', 'http://medplum.com/ai-message|');
```
noahsilas, line 45: It looks to me like AI messages get identifier `http://medplum.com/ai-message|ai-message-topic`, which I think will match both the old query (`"ai-message-topic"`) and the new one (`"http://medplum.com/ai-message|"`).

For my learning, is there a reason to prefer this version?

packages/app/src/resource/CommunicationPayloadPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+  if (!communication) {
+    return null;
+  }
+
+  return (
+    <Document>
+      <Stack>
+        {payload.map((item, index) => (
+          <Textarea
+            key={index}
```
noahsilas: I think this works now, but it is likely to become a danger if someone adds a "Remove payload" button that would drop an entry from the payload state.

I think it's safer to omit `key` entirely than to use an array index, although that may trigger a lint warning? We could decorate the state with some objects to provide stable identity within the UI, but that might be overkill up until we add the dangerous capability to this component.

I guess if it's not going to make the linter yell loudly I'd encourage dropping this line, but if it does we can rely on our future selves to fix this problem when needed.

packages/react-hooks/src/useThreadInbox/useThreadInbox.ts
techdavidy: Good question, we are introducing another communication.
```
http://medplum.com/ai-message|ai-message-systemprompt
```

techdavidy: So we filter any ai related not just topics (parent communication).

packages/app/src/resource/CommunicationPayloadPage.tsx
techdavidy: yeah this page would handle content string only. 
checking to see how it would look with place holders.

techdavidy: added foot note instead

packages/app/src/resource/CommunicationPayloadPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { resourceType, id } = useParams() as { resourceType: ResourceType; id: string };
+  const communication = useResource<Communication>({ reference: resourceType + '/' + id });
+  const keyCounter = useRef(0);
+  const [payload, setPayload] = useState<KeyedPayload[]>([]);
+
+  useEffect(() => {
+    if (communication) {
+      setPayload(
+        (communication.payload ?? EMPTY)
+          .filter((item) => 'contentString' in item)
```
codyebberson: I think the data loss bug is still present.  I think we need to remove this line, and do the filtering at display-time, not at load-time.

techdavidy:
@claude already tested no data loss

techdavidy:
@claude included tests

PR #8716, opened 2026-03-18
fix(fhircast): update topic key expiry upon access

ThatOneBro:
Previously we only set expiry on the first set, since SETNX only will set expiry when it's successful, we should refresh the expiry explicitly in every call so any get to the topic key will reset its TTL


codyebberson:
@claude please review

PR #8717, opened 2026-03-19
docs: reorganize sidebar for clarity and discoverability

everett-williams:
## Summary

Restructures the documentation sidebar to use clearer, more intuitive section names and groupings. No files are moved on disk and no URLs change — this is purely a navigational reorganization.

**Section renames:**
- "Model" → **"Clinical Workflows"** — self-explanatory for both technical and non-technical readers
- "Manage" → **"Security & Identity"** — precisely describes auth, users, and access control
- "Advanced" → **"Deploy & Operate"** — surfaces Self-Hosting (31 files) and Compliance prominently

**Items moved between sections:**
- FHIR Basics → Get Started (foundational prerequisite knowledge)
- Medplum App → Get Started (first tool used after registering)
- Migration → Get Started
- Analytics → Build (it's a guide about building analytics programs)
- Contributing → standalone top-level section (community, not "advanced")

**Other improvements:**
- Added subheaders to Integrate section (Connectivity, Clinical Partners, Business & Operations, Standards)
- Added missing sidebar entries for `scheduling/timezones` and `integration/candid-health`
- Demoted single-page Rate Limits from category to doc link
- Updated `home.mdx` prose reference from "Model in FHIR" to "Clinical Workflows"


packages/docs/docs/home.mdx
```diff
@@ -42,7 +42,7 @@ Medplum is a **headless EHR**. Using [Medplum products](./products) you can buil
 
 1. [Learn FHIR basics](./docs/fhir-basics) to understand the structure of FHIR resources, the healthcare standard on which Medplum is built.
 2. [Learn workflow patterns](./blog/fhir-workflow-patterns-to-simplify-your-life) to understand how resources are used to place requests, record events, and store information.
-3. Explore our docs on workflows under "Model in FHIR"! Popular workflows are included below. Step through specific use cases. Don't see a how-to guide you need? Reach out on our [Discord](https://discord.gg/medplum) or send us an email at [support@medplum.com](mailto:support@medplum.com).
+3. Explore our [Clinical Workflows](/docs/charting) guides! Popular workflows are included below. Step through specific use cases. Don't see a how-to guide you need? Reach out on our [Discord](https://discord.gg/medplum) or send us an email at [support@medplum.com](mailto:support@medplum.com).
```
rahul1: Why does this link to charting?  vs. some tio-level page?

packages/docs/docs/home.mdx
```diff
@@ -54,6 +54,7 @@ Medplum is a **headless EHR**. Using [Medplum products](./products) you can buil
 1. [CRUD using our FHIR datastore](./docs/fhir-datastore), and learn to [search your data](./docs/search).
 2. Explore our [React components](./docs/api/react), pre-built to support Medplum data.
 3. [Run the Medplum stack locally](./docs/self-hosting/running-full-medplum-stack-in-docker) to set up your developer environment.
+4. Plan your [self-hosted deployment](./docs/self-hosting) on AWS, GCP, Azure, or Docker.
```
rahul1, line 57: I think we purposefully left this off to gently guide readers to hosted. 

@maddyli ?

packages/docs/docs/home.mdx
everett-williams: There's no top level page for clinical workflows - charting is just the first subsection within clinical workflows. Open to alternatives if you have any suggestions

packages/docs/docs/home.mdx
everett-williams: My perspective here was that folks want to know about self hosting so burying it might not do us any favors. What we can do here is either:
1. Update our website so self hosting is more clearly teed up with tradeoffs and clarifications on valid reasons for it vs. not; and leave buried in the docs
2. Leave it prominent in the docs and have an upfront page that includes information on tradeoffs and clarifications before diving into the details (could still also do the website update in addition) - I think we link to a decision guide but might be worth making it a part of the initial page instead

packages/docs/docs/home.mdx
maddyli: Hm do we need a link here? The child pages are already linked in the subitems below

packages/docs/docs/home.mdx
maddyli: Hm that is interesting. I would prefer to leave it off the main menu of options: folks who understand open source will seek out the self hosting docs regardless, and I don't want newbies seeing it as a prominent option. I agree that we can probably update the header self-hosting doc to include the tradeoffs more prominently

packages/docs/docs/home.mdx
everett-williams: Fair point. I will remove this

packages/docs/docs/home.mdx
everett-williams: OK sounds good. I'll move it down (maybe before reference). And will update the doc itself as a separate PR later on

packages/docs/sidebars.ts
```diff
@@ -313,6 +370,13 @@ const sidebars: SidebarsConfig = {
       link: { type: 'doc', id: 'integration/hl7-interfacing/index' },
       items: [{ type: 'autogenerated', dirName: 'integration/hl7-interfacing' }],
     },
+    {
+      type: 'category',
+      label: 'FHIRcast',
```
maddyli: Any reason why FHIRcast in particular was moved up? It really only pertains to radiology workflows, most folks probably care about labs/meds/billing or Smart on FHIR more

everett-williams: Good point. I'll move down this 'connectivity' section and keep e-prescribe and billing / RCM as the top two integration sections

maddyli (approved):
YAY

PR #8718, opened 2026-03-19
feat(provider): allow prompts to loop in the translator tool to complete requests in the prompt

techdavidy:
Replaced the single-shot translator call with a loop (max 10 iterations). The translator bot  
  now receives tool results back and keeps calling FHIR until it has everything it needs, then
  returns content with no tool calls to exit.

### DIFF
--- a/examples/medplum-demo-bots/src/spaces-bots/fhir-translator-bot.test.ts
+++ b/examples/medplum-demo-bots/src/spaces-bots/fhir-translator-bot.test.ts
@@ -26,10 +26,7 @@ function mockSystemPromptCommunication(medplum: MockClient, payload1?: string): 
     resourceType: 'Communication',
     id: 'test-communication',
     status: 'completed',
-    payload: [
-      { contentString: SYSTEM_PROMPT },
-      ...(payload1 !== undefined ? [{ contentString: payload1 }] : []),
-    ],
+    payload: [{ contentString: SYSTEM_PROMPT }, ...(payload1 !== undefined ? [{ contentString: payload1 }] : [])],
   });
 }
 


examples/medplum-demo-bots/src/spaces-bots/fhir-translator-bot.ts
```diff
@@ -4,24 +4,33 @@
 import type { BotEvent, MedplumClient } from '@medplum/core';
 import type { Parameters, Reference } from '@medplum/fhirtypes';
 
-function buildProfileContext(profile: Reference): string {
-  const ref = `${profile.reference}`;
+const SYSTEM_PROMPT_IDENTIFIER = 'http://medplum.com/ai-spaces|ai-fhir-request-tools';
```
noahsilas, line 7: I think you can avoid Sonar throwing some false alarms on this if you choose a system URL that starts with "https"

examples/medplum-demo-bots/src/spaces-bots/fhir-translator-bot.ts
```diff
@@ -4,24 +4,33 @@
...
+  });
 
-When the user asks about themselves (e.g., "who am I", "show my profile", "show my info", "what are my details"),
-call fhir_request with GET **${ref}** to fetch their current data — do not answer from context alone.
+  const systemPrompt = communication?.payload?.[0]?.contentString;
+  if (!systemPrompt) {
+    throw new Error('ai-fhir-request-tools system prompt is not available');
+  }
+
+  const profileContextTemplate = communication?.payload?.[1]?.contentString;
```
noahsilas, line 24: It feels slightly opaque to have the ordering of the payload elements be meaningful in this way. What would you think about creating two Communication resources with different identifiers and a single payload each?

examples/medplum-demo-bots/src/spaces-bots/fhir-translator-bot.ts
```diff
@@ -4,24 +4,33 @@
...
+  if (!profileContextTemplate) {
+    throw new Error('ai-fhir-request-tools profile context template is not available');
+  }
 
-When the user asks about their own resources (e.g., "my tasks", "my appointments", "my observations"),
-query using the appropriate subject/owner/requester parameter pointing to **${ref}**.
-For example: GET Task?owner=${ref} or GET Appointment?actor=${ref}.
-`;
```
noahsilas, line 24: Interesting idea of moving these prompts into the database. Is the idea here that we want to try lots of variations here quickly, potentially scoped to specific projects, without requiring full releases each time we want to experiment?

Overall I think the approach makes a lot of sense. It makes changes to the prompt invisible in source control, but since each Communication resource storing the prompts is versioned I think that we should still have strong tracking on what changed when 👍

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
```diff
@@ -68,6 +68,7 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
   const scrollViewportRef = useRef<HTMLDivElement>(null);
   const isSendingRef = useRef(false);
   const loadVersionRef = useRef(0);
+  const componentStreamOpenedRef = useRef(false);
```
noahsilas, line 71: For my learning, why use a `Ref` here instead of State? I guess the idea here is that we use this within long-lived effects as a sort of "one time latch", and don't actually need any reactive behavior when this changes?

examples/medplum-provider/src/utils/spaceMessaging.ts
```diff
@@ -430,47 +441,18 @@ export async function processMessage(params: ProcessMessageParams): Promise<Proc
...
-      };
-    }
+  if (!loopCompleted) {
+    throw new Error(
+      `The AI reached the maximum number of steps (${MAX_AGENT_ITERATIONS}) without completing your request. Please try a more specific question.`
+    );
+  }
 
-    // Get summary response after tool execution (streaming if callback provided)
+  // Get summary response after tool execution (streaming if callback provided)
```
noahsilas, line 454: The diff here gets a bit awkward to read on github.com, but it looks like these two bot executions  (summary response generation and visualization component generation) got reordered. I sort of wish there was a way to help git recognize when a chunk of code should be considered as a unit more atomically...

I guess that could be an indicator that these might be useful to break into functions?

examples/medplum-demo-bots/src/spaces-bots/fhir-translator-bot.ts
techdavidy: I thought about it but since they are very related I kept both. 
We can review after some user testing.

examples/medplum-provider/src/utils/spaceMessaging.ts
techdavidy: in a following pr I'll be managing those 2 bots

PR #8723, opened 2026-03-20
fix(client/subscription): prevent React lifecycles from tracking unused WebSocket subs

### LINKED ISSUE #8662: Investigate `SubscriptionManager` keeping untracked subscriptions bound to WebSocket
We've done a lot of work around garbage collection on the server side but a lot of the cleanup assumes that a user's client will either:
1) Unbind an unused subscription
2) Close the WebSocket, which when detected server-side will clean up all WebSocket subscriptions associated with that WebSocket connection

It seems like some of the particular patterns that emerge from React component lifecycles may make some race conditions possible which may untrack subscriptions in the `SubscriptionManager` without actually sending an unbind message. This means that a subscription will not be cleaned up until a user closes their WebSocket connection, which may not happen frequently enough to free up unused subscriptions from the perspective of the `SubscriptionManager`

ThatOneBro:
Fixes #8662

### DIFF
--- a/packages/core/src/subscriptions/index.ts
+++ b/packages/core/src/subscriptions/index.ts
@@ -611,6 +611,10 @@ export class SubscriptionManager {
       this.maybeEmitDisconnect(entry);
       this.removeCriteriaEntry(entry);
     }
+    // Auto-close the WebSocket when the last entry has been finalized
+    if (entriesToRemove.length > 0 && this.criteriaEntries.size === 0) {
+      this.closeWebSocket();
+    }
   }
 
   private isEntryGettingRemoved(criteriaEntry: CriteriaEntry): boolean {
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

For context: websocket connect/disconnect lifecycle events have proven unreliable, and it feels like we might be fighting a losing battle of trying to rely on them. Is there a better way to do this, like maybe a "global list + GC" model?

packages/core/src/subscriptions/index.ts
```diff
@@ -518,9 +580,41 @@ export class SubscriptionManager {
       return;
     }
 
-    // If actually removing (refcount === 0)
-    this.maybeEmitDisconnect(criteriaEntry);
-    this.removeCriteriaEntry(criteriaEntry);
+    // Record when refCount hit 0 — gcUnrefEntries will finalize after the grace period.
+    // Re-subscribing before then rescues the entry (see addCriteria).
+    criteriaEntry.lastUnrefTime = Date.now();
+    criteriaEntry.generation++;
```
mattwiller: Instead of early-returning above, the code could be aligned with this comment like so:
```ts
if (criteriaEntry.refCount <= 0) {
  criteriaEntry.lastUnrefTime = Date.now();
  criteriaEntry.generation++;
}
```

ThatOneBro: Good callout

PR #8743, opened 2026-03-22
Rewrite Attachment.url in convertToTransactionBundle (#8664)

### LINKED ISSUE #8664: convertToTransaction bundle should replace Attachment.url
A common pattern when importing FHIR data from external sources is that attachment.url will point to a binary resource. This utility function should take that into account when rewriting references.

HeyangQin:
## Summary
- Extend `convertToTransactionBundle` to rewrite `Attachment.url` when it references a Binary resource in the bundle (e.g. `Binary/abc123`)
- Update `referenceReplacer` to handle `key === 'url'` in addition to `key === 'reference'`
- Update `findReferences` to detect `urn:uuid:` URLs for correct topological bundle ordering
- External URLs (e.g. `https://...`) are left untouched since they won't match any ID in the bundle

## Test plan
- [x] Added test: Attachment.url pointing to Binary in bundle is rewritten to `urn:uuid:`
- [x] Added test: External URLs are not rewritten
- [x] Added test: Binary is ordered before DocumentReference when referenced via Attachment.url
- [x] All 19 tests pass

Closes #8664

### DIFF
--- a/packages/core/src/bundle.test.ts
+++ b/packages/core/src/bundle.test.ts
@@ -234,6 +234,110 @@ describe('Bundle tests', () => {
       });
     });
 
+    test('Rewrites Attachment.url referencing a Binary in the bundle', () => {
+      const inputBundle: Bundle = {
+        resourceType: 'Bundle',
+        type: 'collection',
+        entry: [
+          {
+            fullUrl: 'https://example.com/DocumentReference/doc1',
+            resource: {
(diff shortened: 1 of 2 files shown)


codyebberson:
Hi @HeyangQin  - thank you for submitting!

Can you please sign your commit?  See here: https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot

PR #8747, opened 2026-03-22
User management cleanup

codyebberson:
Medplum's "Users" / "Clients" / "Bots" admin experience is long overdue for some cleanup.

In particular, the most common case of "I want to see a list of Clients" + "I want to see the client ID + client secret" has always required some extra hops and arcane knowledge of the data model.

This PR builds on @maddyli's "Member Table" cleanup in https://github.com/medplum/medplum/pull/8465

This PR introduces a proper "Membership Details" page for all membership details:

User:

image

ClientApplication:

image

### DIFF
--- a/packages/app/src/admin/MemberDetailsPage.tsx
+++ b/packages/app/src/admin/MemberDetailsPage.tsx
@@ -19,7 +19,7 @@ export function MemberDetailsPage(): JSX.Element {
       <Title>ProjectMembership Details</Title>
       <MedplumLink to={membership}>Go to ProjectMembership</MedplumLink>
       <ResourceTable value={membership} />
-      <Title mt="md">{profile.resourceType} Details</Title>
+      <Title mt="md">Profile Details</Title>
       <MedplumLink to={profile}>Go to {profile.resourceType}</MedplumLink>
       <ResourceTable value={profile} ignoreMissingValues />
     </>


packages/app/src/admin/MemberDetailsPage.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const profile = useResource(membership.profile);
+  if (!profile) {
+    return <Loading />;
+  }
+
+  return (
+    <>
+      <Title>ProjectMembership Details</Title>
+      <MedplumLink to={membership}>Go to ProjectMembership</MedplumLink>
+      <ResourceTable value={membership} ignoreMissingValues />
```
maddyli: Hm I think the ProjectMembership's accessPolicies and userConfiguration are not shown here because of the ignoreMissingValues being set to true. Same for project admin/not admin

Probably reasoable for keeping the UI clean, but it is somewhat unclear how you would add one of those based on this screen. It could be worth adding some functionality here to edit this properties 

Small nit, this is super nice!

codyebberson: Good point.  Maybe we remove `ignoreMissingValues` for the `ProjectMembership`?

packages/app/src/admin/MemberDetailsPage.tsx
```diff
@@ -0,0 +1,27 @@
...
+  if (!profile) {
+    return <Loading />;
+  }
+
+  return (
+    <>
+      <Title>ProjectMembership Details</Title>
+      <MedplumLink to={membership}>Go to ProjectMembership</MedplumLink>
+      <ResourceTable value={membership} />
+      <Title mt="md">{profile.resourceType} Details</Title>
```
mattlong: This still leaves room for confusion. For example, in the ClientApplication screen shot, it's probably not obvious if the "ClientAplication Details" section corresponds to the User or Profile of the ProjectMembership above since they have the same display string of "Cody Test Default Client"

Perhaps change the title here to "Profile Details" and add a section just above for "User Details" that shows `membership.user` to help tie together the page sections? But I dunno, maybe that's a mid-wit solution.

mattlong: Thinking more, a better solution might be to make `ResourceTable` be expandable on reference fields to show details for the referenced resourced. Then we wouldn't need the extra sections at all

codyebberson: The `User` object will often not be accessible by users.  We could "try and show it if available", but that's net new functionality, and I consider that out of scope.

I renamed the second section to "Profile Details".

I like the idea of "references become follow-able and expandable in the `ResourceTable`", but that's a significant project, and probably out of scope for this minor cleanup.

packages/app/src/admin/MemberDetailsPage.tsx
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import { Title } from '@mantine/core';
+import { Loading, MedplumLink, ResourceTable, useMedplum, useResource } from '@medplum/react';
+import type { JSX } from 'react';
+import { useParams } from 'react-router';
+
+export function MemberDetailsPage(): JSX.Element {
+  const medplum = useMedplum();
+  const { membershipId } = useParams() as { membershipId: string };
+  const membership = medplum.readResource('ProjectMembership', membershipId).read();
```
noahsilas, line 11: I notice that this line seems to be handling data loading via React Suspense, while the next line is using a `useResource` hook to handle it. It feels a little strange to be mixing those two styles in a single component.

Is the idea here that because `.read()` uses suspense, you can `useResource` in an unconditional way?

That's an interesting way to get around the limitations of the "rules of hooks" - I had been starting to build out a pattern of passing an `{ enabled }` option into dependent searches, but in some ways this style is nicer - assuming that you have a Suspense boundary in the right place.

Just commenting because this is something new for me to think about in the context of data loading in our applications 🤔

PR #8748, opened 2026-03-22
Fixes #8072 - flaky HL7 client tests

### LINKED ISSUE #8072: [Flaky test] Hl7Client
https://github.com/medplum/medplum/actions/runs/20239056049/job/58102142557?pr=8071

```
> @***/hl7@5.0.9 test
> jest --coverage

PASS src/connection.test.ts
PASS src/server.test.ts
FAIL src/client.test.ts (123.502 s)
  ● Hl7Client › Creates new connection whenever connection is closed from other side and a new message is sent

    thrown: "Exceeded timeout of 120000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      584 |   });
      585 |
    > 586 |   test('Creates new connection whenever connection is closed from other side and a new message is sent', async () => {
          |   ^
      587 |     const port = getRandomPort();
      588 |
      589 |     let serverSideConnection!: Hl7Connection;

      at test (src/client.test.ts:586:3)
      at Object.describe (src/client.test.ts:14:1)

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```

codyebberson:
Fix flaky HL7 client test + related production bugs

Fixes #8072

Background

The test "Creates new connection whenever connection is closed from other side and a new message is sent" (packages/hl7/src/client.test.ts:585) was
timing out in CI at roughly 1% frequency. The 120-second timeout made it hard to notice and impossible to reproduce locally. Root cause analysis revealed
 three separate bugs — two in production code, one in the test itself.

---
Bug 1 — Hl7Connection: pending messages hang forever when the connection is closed by the remote peer (connection.ts)

When a remote peer closes the socket, Hl7Connection's socket.on('close') handler only dispatched Hl7CloseEvent. Any sendAndWait promises that were in
flight at that moment were left in pendingMessages with no mechanism to ever resolve or reject them — they would hang until the process exited or the
test runner timed out.

Fix: Extracted the pending-message drain logic from close() into a private drainPendingMessages() method and called it from the socket close handler. The
 method is idempotent (no-op when the map is already empty), so the explicit close() call path is unaffected.

---
Bug 2 — client.test.ts: sleep(0) is not a reliable synchronization primitive for I/O events

The test called await sleep(0) after closing the server-side connection, expecting the client to have processed the TCP close by then. sleep(0) uses
setTimeout, which fires in Node's timers phase — before the poll phase where incoming I/O events (like the TCP FIN/RST from the server) are processed. On
 a loaded CI machine, the timer can fire before the client socket has received and processed the close, causing connect() to return the old dead
connection. Combined with Bug 1, this made sendAndWait hang forever.

Fix: Register a { once: true } listener on the client's 'close' event before closing the server-side connection, then await that promise. This guarantees
 the client has fully processed the close (and cleared this.connection) before the second sendAndWait is attempted.

---
Bug 3 — client.test.ts / server.ts: port allocation was racy

Tests used Math.random() to pick ports in the range 30000–40000, with a per-process usedPorts set to prevent intra-run collisions. This didn't protect
against conflicts with other processes or parallel CI workers.

The intermediate fix (getFreePort() — bind to 0, read port, release) still had a window between releasing the port and the test server binding to it.

Fix: Hl7Server.start() now accepts port 0 and returns Promise<number> (the OS-assigned port read from server.address() after binding). The OS never
releases the port between assignment and use, making this genuinely race-free. Tests that start an Hl7Server now use const port = await server.start(0).
Tests using raw net.createServer() use server.listen(0, ...) + server.address() inline. getFreePort() is kept only for the two tests that need an unbound
 port number (to test connecting to a port with nothing listening).

---
Additional improvements to Hl7Server.start()

- Returns the bound port (Promise<number> instead of Promise<void>) — useful for start(0) and as a general convenience
- Bounded retries on EADDRINUSE — previously retried indefinitely; now rejects after maxRetries (default 10) with a clear error message
- Exponential backoff with cap — retry delays are min(50 * 2^(n-1), 5000ms) instead of a flat 50ms, capped at 5s per attempt to bound worst-case wait
time
- Non-EADDRINUSE errors are now rejected — previously swallowed silently
- maxRetries option lives in a new Hl7ServerStartOptions interface (not Hl7ConnectionOptions, which is for per-connection settings)

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -21,6 +21,9 @@ mkdir -p coverage/combined
 SHOULD_RUN_SEED_TEST=$(date) time npx turbo run test:seed --filter=./packages/server -- --coverage
 cp "packages/server/coverage/coverage-final.json" "coverage/packages/coverage-server-seed.json"
 
+# Build
+npm run build
+
 # Test
 # Run them separately because code coverage is resource intensive
 


packages/hl7/src/server.ts
```diff
@@ -60,17 +73,30 @@ export class Hl7Server {
...
         if (e?.code === 'EADDRINUSE') {
-          await sleep(50);
-          server.close();
-          listenOnPort(port);
+          if (retries >= maxRetries) {
+            reject(new Error(`Failed to bind to port ${port} after ${maxRetries} retries`));
+            return;
+          }
+          retries++;
+          const exponentialBackoffTime = Math.min(50 * 2 ** (retries - 1), MAX_BACKOFF_MS);
```
ThatOneBro: This exponential backoff will affect how quickly we are able to detect and rebind to ports in the agent, I think we should not change the time between attempts because we want to rebind as fast as possible. Setting a default `maxRetries` makes sense but in cases where the old agent refuses to shut down (has ocassionally happened historically) this would mean we would stop attempting to take over ports after 10 retries.

We should also support an infinite retries configuration for the agent still, to preserve the behavior that will keep the agent attempting to bind to the port

scripts/test.sh
```diff
@@ -26,13 +26,17 @@ cp "packages/server/coverage/coverage-final.json" "coverage/packages/coverage-se
 
 for dir in `ls packages`; do
   if test -f "packages/$dir/package.json" && grep -q "\"test\":" "packages/$dir/package.json"; then
-    npx turbo run test --filter=./packages/$dir -- --coverage
+    pushd packages/$dir
+    npm t -- --coverage
```
ThatOneBro, line 33: I think this breaks how our CI job currently works, since we rely on the build being part of the dependency graph for Turbo in the `test` job 👀

codyebberson: We run `npm run build` in `build.yml`, no?

codyebberson: Oh, I see.  that's in a different "job", and therefore doesn't have the cache.  I see.

PR #8749, opened 2026-03-22
Disable redirect URL partial match

codyebberson:
The `getClientRedirectUri` function supported an `allowPartial` flag with known usability vs security tradeoffs. This PR tightens that behavior in two ways:

1. Partial matching is now opt-in at the `ClientApplication` level via a new (deprecated-on-arrival) `redirectUriAllowPartialMatch` field, rather than being controlled solely by the call site.
2. When partial matching is permitted, the new `isAllowedPartialRedirectUri` helper enforces that the origin must match exactly — only path and query string divergence is allowed.

A small number of known client applications have been identified as depending on partial match behavior and have been migrated to the explicit opt-in. All other clients default to exact matching.

The `allowPartial` code path and `redirectUriAllowPartialMatch` flag are deprecated and will be removed in a future release once remaining clients have been migrated.

### DIFF
--- a/packages/server/src/auth/external.test.ts
+++ b/packages/server/src/auth/external.test.ts
@@ -408,6 +408,94 @@ describe('External', () => {
     expect(tokenResponse.body.profile.display).toBe('External User');
   });
 
+  test('Block partial redirect URI match', async () => {
+    const subjectAuthClient = await withTestContext(async () => {
+      // Create a new client application with external subject auth
+      const client = await createClient(systemRepo, {
+        project,
+        name: 'Subject Auth Client',
+        redirectUri,
+      });
+
+      // Update client application with external auth
+      await systemRepo.updateResource<ClientApplication>({
+        ...client,
+        identityProvider: {
+          ...identityProvider,
+          useSubject: true,
+        },
+      });
+
+      return client;
+    });
+
+    const url = appendQueryParams('/auth/external', {
+      code: randomUUID(),
+      state: JSON.stringify({
+        redirectUri: redirectUri + '/extra',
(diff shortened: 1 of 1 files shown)


packages/docs/static/data/medplumDefinitions/clientapplication.json
```diff
@@ -448,6 +480,38 @@
...
+      "types": [
+        {
+          "datatype": "boolean"
+        }
+      ],
+      "path": "ClientApplication.redirectUriAllowPartialMatch",
+      "min": 0,
+      "max": "1",
+      "short": "",
+      "definition": "@deprecated This feature is deprecated. Redirect URIs must match exactly for security reasons.",
```
mattwiller: 👍

packages/server/src/oauth/clients.ts
```diff
@@ -65,14 +65,26 @@ export function getClientRedirectUri(
...
 }
 
+function isAllowedPartialRedirectUri(actualUri: string, requestedUri: string): boolean {
+  // This is a temporary workaround to allow partial matching of redirect URIs for legacy clients.
+  // It should be removed once all clients are migrated to use exact redirect URIs.
+  try {
+    const actualUrl = new URL(actualUri);
+    const requestedUrl = new URL(requestedUri);
+    return actualUrl.origin === requestedUrl.origin && requestedUrl.pathname.startsWith(actualUrl.pathname);
+  } catch (_err) {
```
mattwiller:
```suggestion
  } catch {
```

codyebberson:
Holding on this until after core business hours to avoid potential customer disruption.

PR #8750, opened 2026-03-23
docs(opentelemetry): fix environment variable references for OTLP configuration

jakub-bacic:
## Summary

- Corrects `OTLP_TRACES_ENDPOINT` to `OTLP_TRACE_ENDPOINT` in the OpenTelemetry self-hosting docs

The server reads `process.env.OTLP_TRACE_ENDPOINT` (singular) in [`instrumentation.ts:39`](https://github.com/medplum/medplum/blob/main/packages/server/src/otel/instrumentation.ts#L39), but the docs documented it as `OTLP_TRACES_ENDPOINT` (plural). This would cause tracing to silently not work for users following the docs.

`OTLP_METRICS_ENDPOINT` was already correct and is unchanged.

### DIFF
--- a/packages/docs/docs/self-hosting/opentelemetry.md
+++ b/packages/docs/docs/self-hosting/opentelemetry.md
@@ -55,7 +55,7 @@ OpenTelemetry cannot be configured through the normal `medplum.config.json` file
 Instead, you must add the following environment variables:
 
 ```bash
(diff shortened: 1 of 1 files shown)


mattwiller:
Great catch @jakub-bacic — thank you for the PR!

PR #8755, opened 2026-03-23
Update scheduling "availability" subextension to be shaped like R5/R6 Availability type

### LINKED ISSUE #8710: Adjust SchedulingParameters `availability` to be in line with future `Availability` type
In FHIR R5, a new `Availability` type is introduced (https://hl7.org/fhir/R5/metadatatypes.html#Availability).

- This type is natively used by the R5 `HealthcareService`, which we are planning on using as a container for shared scheduling definitions.
- The shape of this type is better suited to how humans express availability: startTime/endTime rather than startTime/duration ("9am-5pm" vs. "from 9am for 8 hours")
- This avoids some questions around fields in the `Timing` type that our extension was ignoring or erroring on. example: `Timing.repeat.durationUnit` needs to be constrained to consistent units and causes errors when units like "mo" (month) are used.
- This will give us a smoother upgrade path to R5/R6 in the future.

Question: what is the best way to represent the R5 `Availability` type in an R4 extension field?

Question: are we still early enough in alpha to make a fully breaking change here, or do we need to introduce a different extension key or otherwise version the `availability` field to support the old style as well?

DST Notes: This has the effect of locking end times to clock times; "12am+8hours" can be 7am, 8am, or 9am depending on how the clock behaves on a given day; "12am-8am" always ends at wall clock time 8am, but can be 7, 8, or 9 hours long. The challenge of inconsistent times ("01:30" happens twice some days, "02:30" never happens on others) already exists for the current implementation's start time, but can now also affect end times. The rules for this behavior should be tested and documented.

noahsilas:
The original shape of our "availability" extension used a FHIR [Timing datatype](https://hl7.org/fhir/R4B/datatypes.html#Timing) to store availability information.
- We used the `Timing.repeat` attribute to represent our availability recurrence rules. 
- Unsupported `Timing` fields were silently ignored
- Timing.repeat represents these with "startTime + duration", which has some challenges:
  - It doesn't match typical human usage; we generally discuss startTime + endTime (e.g. "9am – 5pm")
  - It can be hard to interpret when the unit used is not an hour: compare "9:30am – 5:00pm" with "starting at 9:30am and lasting for 450 minutes"

In https://github.com/medplum/medplum/pull/8580 we started supporting using `HealthcareService.availableTime` to set scheduling availability that can be shared across practitioners.

In FHIR R5+ there is a new [Availability datatype](https://hl7.org/fhir/R5/metadatatypes.html#Availability). R4's `HealthcareService.availableTime` becomes R5's `HealthcareService.availability` and takes on this type. To make our future transition to R6 easier, we would like to standardize on a shape in our extension now that matches that.

To that end, we are changing the shape of the "availability" sub extension. Instead of having `valueTiming`, it has nested extensions:

| Sub-extension         | Type          | Description                                    |
| --------------------- | ------------- | ---------------------------------------------- |
| `availableTime`       | (nested)      | One entry per availability window (repeatable) |
| ↳ `daysOfWeek`        | `Code`   | One entry per day (`mon`–`sun`); repeatable    |
| ↳ `allDay`            | `Boolean`| If `true`, window spans the full day           |
| ↳ `availableStartTime`| `Time`   | Opening time (ignored when `allDay` is `true`) |
| ↳ `availableEndTime`  | `Time`   | Closing time (ignored when `allDay` is `true`) |
| `notAvailableTime`    | (nested)      | Reserved for future use; not yet implemented        |

I had originally implemented this as a new "availability.r4" extension that we could support in parallel with our older "availability" shape, but our opinion is that we are still early enough in the Alpha phase of our Scheduling support that this breaking change is acceptable, as it sets us up for a better long-term data model. To that end, later commits in this PR remove the old "availability" subextension, and then rename "availability.r4" -> "availability" in its stead.

Resolves #8710


examples/foomedical/README.md
```diff
@@ -106,14 +106,19 @@ The "Get Care" page is configured to search for availability with service-type "
         },
         {
           "url": "availability",
-          "valueTiming": {
-            "repeat": {
-              "dayOfWeek": ["mon", "tue", "wed", "thu"],
-              "timeOfDay": ["09:00:00"],
-              "duration": 8,
-              "durationUnit": "h"
+          "extension": [
```
rahul1: Actually, for this, instead of inventing a full code block in this readme, can we just reference our medplum docs? That makes it easier to not have to change things in multiple places.

noahsilas:
@claude Please review

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
```diff
@@ -177,6 +193,31 @@ export function chooseSchedulingParameters(
...
+// Convert a single availability extension into SchedulingParametersAvailability entries.
+// notAvailableTime sub-extensions are ignored for now.
+function extractAvailabilityR4(ext: {
+  url: 'availability';
+  extension: (AvailabilityR4AvailableTime | AvailabilityR4NotAvailableTime)[];
+}): SchedulingParametersAvailability[] {
+  return ext.extension
+    .filter((sub): sub is AvailabilityR4AvailableTime => sub.url === 'availableTime')
+    .map((availTime) => {
+      const dayOfWeek = availTime.extension.filter((e) => e.url === 'daysOfWeek').map((e) => e.valueCode);
```
codyebberson, line 205: Super minor nit, feel free to ignore - these could be one pass through `extension` rather than 4.

Or, the `const start` and `const end` could be after the `if (allDay)` check.

noahsilas: > these could be one pass through extension rather than 4.

Is the idea to reduce the number of iterations over this extension array? Are you thinking of something like:
```ts
const dayOfWeek = [];
let start: string;
let end: string;

for (const extension of availTime.extension) {
  if (extension.url === 'daysOfWeek') { dayOfWeek.push(extension.valueCode); }
  if (extension.url === 'start') { start = extension.valueTime; }
  if (extension.url === 'end') { end = extension.valueTime; }
  if (extension.url === 'allDay') { start = end = '00:00:00'; }
}
```

That's kind of nice in that it would only do a single pass. I tend to like the functional style, but for something like this it does have some minor performance impact.

One utility I find myself thinking about writing an implementation of would be something like `_.groupBy`. I kind of want to do something more like:
```ts
const byUrl = groupBy(availTime.extension, ext => ext.url);
const daysOfWeek = byUrl['daysOfWeek'].map(x => x.valueCode);
...
```

Getting that to work in a type-safe way for the extension format is a bit tricky though...

I really like your suggestion of doing the pass for `allDay` first and using that as a flag to skip the start/end iterations. I'm also going to take a quick stab at the one-iteration version and see how it feels.

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -46,11 +46,11 @@ All scheduling constraints are managed through a single consolidated extension:
...
+| `serviceType`       | [CodeableConcept](/docs/api/fhir/datatypes/codeableconcept) | Schedule only                       | Required                                        | Applies configuration only to the specified service type, overriding defaults for that service                                                                                       | N/A - must be specified                                                                |
+| `availability`      | Nested extension                                            | Schedule only                       | Optional                                        | Bookings must fully fit within the recurring windows (day-of-week + start time + end time). See [below](#availability-extension) for details                                         | Time is implicitly available by default (unless blocked by Slots or other constraints) |
```
mattwiller: Could/should "Nested extension" link to the section below?

mattwiller (approved):

packages/docs/docs/scheduling/defining-availability.md
noahsilas: Good idea - I had added the link into the "Behavior when defined" section, but that's not as good a spot for it. Thanks!

PR #8760, opened 2026-03-23
feat(spaces): Improve AI agent loop handling and add resizable preview panel.

techdavidy:
- Max iterations — no more error throws: When the AI agent loop hits the step limit, instead of   
  throwing an error, the app now calls the summary bot on whatever tool results were gathered and   
  appends a note to the response suggesting the user try a more specific question or break the      
  request into smaller parts.                          
                                             
  - Resizable chat/preview split: Added a drag handle between the chat panel and the                
  resource/component preview panel. The preview can be resized between 50% and 75% of the available 
  container width (sidebar excluded). Defaults to 50%.

### DIFF
--- a/examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
+++ b/examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
@@ -83,8 +83,8 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
     document.body.style.userSelect = 'none';
 
     const onMouseMove = (moveEvent: MouseEvent): void => {
-      if (!isDraggingRef.current) { 
-        return; 
+      if (!isDraggingRef.current) {
+        return;
       }
       const delta = dragStartXRef.current - moveEvent.clientX;
       const containerWidth = chatContainerRef.current?.parentElement?.offsetWidth ?? window.innerWidth;


techdavidy:
@claude

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
```diff
@@ -65,10 +65,47 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
   const [streamingComponentCode, setStreamingComponentCode] = useState<string | undefined>();
   const [componentPanelOpen, setComponentPanelOpen] = useState(false);
   const [componentPreview, setComponentPreview] = useState<{ code: string; resources?: string[] } | undefined>();
+  const [panelWidth, setPanelWidth] = useState<number>(() => window.innerWidth * 0.5);
```
noahsilas, line 68: I think this might be a good use case for a `useLayoutEffect`:

```ts
const [panelWidth, setPanelWidth] = useState(0);
useLayoutEffect(() => {
  setPanelWidth(chatContainerRef.current?.parentElement?.offsetWidth ?? window.innerWidth)
}, []);
```

noahsilas (approved):
Instead of making the splitter a one-off adjustment inside this component, can we introduce a reusable component? I'm vaguely imagining something like `<ResizableSplit leftPane={<Main />} rightPane={<Preview />} />`, although I can also imagine an implementation like this:
```ts
<ResizableSplit>
  <ResizableSplit.Pane><Main /></ResizableSplit.Pane>
  <ResizableSplit.Pane min="50%" max="75%"><Preview /></ResizableSplit.Pane>
</ResizableSplit>
```

I think there are a couple compelling reasons for this:
- reusability for the next time we want a resizable component
- it would be easier to reason about the splitter in isolation from other complex components
- I think this probably should have resize detection through something like https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
- This could implement the ARIA window splitter pattern (https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/) which would be nice to do in a more contained place

If this would be a pain in this PR, I can give it a shot separately as a follow up 👍

PR #8783, opened 2026-03-25
Fix use of modifiers in _filter search

mattwiller:


packages/core/src/search/search.ts
```diff
@@ -396,7 +396,7 @@ function parsePrefix(input: string, defaultOperator: Operator): { operator: Oper
 }
 
 function parseModifier(modifier: string): Operator {
-  return MODIFIER_OPERATORS[modifier] ?? Operator.EQUALS;
+  return (MODIFIER_OPERATORS[modifier] ?? (modifier as Operator)) || Operator.EQUALS;
```
mattwiller: This is the primary component of the fix: using the (prefix) modifier directly if it's not a true modifier, to accommodate `_filter` expressions with `ne` instead of `:not`

ThatOneBro: Could we do something like:
`MODIFIER_OPERATORS[modifier] ?? PREFIX_OPERATORS[modifier] ?? Operator.EQUALS`

Not sure if that makes sense

EDIT: Wait, that does not work 😅

Issue #8337, opened 2026-01-31
Update $book to accept `Appointment` resources as input


noahsilas:
In #8300 we implemented the first pass and only handled accepting `Slot` resources as input parameters. The spec says we should also handle `Appointment` resources. We need to follow up and update the operation to work with both types.

PR #8793, opened 2026-03-26
Introduce Appointment/$find multi-schedule search

### LINKED ISSUE #8273: Multi-schedule `$find` operation on Appointment resource
In #7967 we added `$find` as an instance level operation on a single schedule (`GET /Schedule/:id/$find`). For the full spec, we want to introduce this as resource type level (`GET /Appointment/$find?schedule=...`), with the ability to search across multiple schedules simultaneously. Result appointments should be available only when _all_ schedule parameters would have suitable availability.

This is very similar to https://github.com/medplum/medplum/issues/8272, differing in the base resource type. Results will be `Appointment` resources rather than `Slot` resources.

Full spec in https://github.com/medplum/medplum/issues/7686.

noahsilas:
This type-level operation receives multiple schedule references as input, and finds times that a given appointment type can be created that are available for all the input schedules.

Before we leave the Alpha window we are thinking of removing `Schedule/:id/$find`:
- Simplifies the API that `$book` accepts lets us more strongly type it and keep the implementation simple (only takes a single appointment resource as input)
- Removes some cognitive load for implementers who would have had to choose between the two implementations

Implementation Notes:
- Handling of bufferBefore/bufferAfter has been adjusted; previously I was looking for availability wide enough to include the buffers, but that doesn't work well for multi-schedule situations where each schedule can have distinct buffer requirements.

  By applying the buffer requirements to each calendar directly (reducing the size of each availability window for that calendar by the buffer amounts), we ensure that all the times inside each window have the requisite availability. We can then intersect all those actual availability windows normally.

- The Slots that cause the appointment to be satisfied are returned as "contained" resources inside each appointment. This allows the single top level Appointment resource to include all the needed information for operations like $book, which otherwise would need to receive Schedule references in some other way. This may also keep $book simple, as it can validate these contained slot resources.

- The `$book` operation has not yet been updated to handle these Appointment results; this will be a fast-follow via https://github.com/medplum/medplum/issues/8337.

Resolves: https://github.com/medplum/medplum/issues/8273


codyebberson:
@claude please review

noahsilas:
@claude There have been major updates to rebuild this on top of changes in the `main` branch, particularly with respect to https://github.com/medplum/medplum/pull/8983. Please re-review.

packages/server/src/fhir/operations/find.ts
```diff
@@ -109,69 +148,243 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
...
-    throw new OperationOutcomeError(badRequest('SchedulingParameters not present on Schedule or HealthcareService'));
-  }
+  const serviceType = toCodeableReferenceLike(healthcareService);
+
+  const allAvailability = schedules.map((schedule, idx) => {
+    const schedulingParameters = parameterGroup.get(schedule);
+    invariant(schedulingParameters);
+    const actor = actors[idx];
+    const actorTimeZone = getTimeZone(actor);
+    const activeTimeZone = schedulingParameters.timezone ?? actorTimeZone;
```
mattlong:
```suggestion
    const activeTimeZone = schedulingParameters.timezone ?? getTimeZone(actor);
```

packages/server/src/fhir/operations/find.ts
```diff
@@ -109,69 +148,243 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
...
+    }));
 
-  const activeTimeZone = schedulingParameters.timezone ?? actorTimeZone;
+    // Optimization: restrict to windows long enough for the requested duration
+    // here before trying to do intersections with other schedules later. This
+    // also ensures that we don't return intervals having an `end` before the
+    // `start` after our previous buffer-trimming step.
+    availability = availability.filter((interval) => {
+      const durationMs = interval.end.getTime() - interval.start.getTime();
+      return durationMs > schedulingParameters.duration * 1000;
```
mattlong:
```suggestion
      return durationMs >= schedulingParameters.duration * 1000;
```

packages/server/src/fhir/operations/find.ts
```diff
@@ -109,69 +148,243 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
...
 
-  const schedulingParameters = chooseSchedulingParameters(schedule, healthcareService);
-  if (!schedulingParameters) {
-    throw new OperationOutcomeError(badRequest('SchedulingParameters not present on Schedule or HealthcareService'));
-  }
+  const serviceType = toCodeableReferenceLike(healthcareService);
+
+  const allAvailability = schedules.map((schedule, idx) => {
+    const schedulingParameters = parameterGroup.get(schedule);
+    invariant(schedulingParameters);
```
mattlong:
```suggestion
    assert(schedulingParameters);
```

Am I missing a difference between `assert` from `node:assert` and `invariant`?

The-Alchemist:
@claude : re-check

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
```diff
@@ -145,6 +145,51 @@ function exactlyZero(arr: unknown[], attribute: string, resourceType: string): v
   }
 }
 
+function allMatch<T>(values: T[]): boolean {
+  const first = values[0];
+  return values.every((value) => value === first);
```
mattlong, line 153: For lower level util functions, it would be nice to optimize and return early avoid the `.every` if `values.length <= 1`.

packages/server/src/fhir/operations/find.ts
```diff
@@ -109,69 +148,243 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
...
+ * Handles HTTP requests for the Schedule $find operation.
+ *
+ * Endpoints:
+ *   [fhir base]/Schedule/[id]/$find
+ *
+ * @param req - The FHIR request.
+ * @returns The FHIR response.
+ */
+export async function scheduleFindHandler(req: FhirRequest): Promise<FhirResponse> {
+  const params = parseInputParameters<ScheduleFindParameters>(scheduleFindOperation, req);
```
mattlong, line 331: Not sure if it'd be additive, but you might want to add an explicit validation step for the reference inputs; e.g. for `params.schedules` check if they all pass `isReference<T>(val, 'Schedule')` to ensure they are well formed reference strings with the expected resource type.

mattlong (changes requested):
Nice. Love the reuse of the core logic for two endpoints!

packages/server/src/fhir/operations/find.ts
noahsilas: Oops, nice catch. Thanks!

packages/server/src/fhir/operations/find.ts
noahsilas: It looks like there's not - I added this thinking that utility `node:assert` would not refine the type to exclude values like `undefined`, but it looks like `@types/node` has added annotations such that it will. I'll clean this up!

noahsilas:
re: extractCommonParameters throwing `Error`: I'm looking at some options to improve this as a separate follow up. The pain point here is that these utilities don't have enough access to the context to determine where the input came from, and so emitting an error with an expression that points to the correct place is hard.

I'm looking at an approach where we annotate objects with path information so that we can emit good errors even from this kind of context-free helper.

UPDATE: This proposal is now posted: https://github.com/medplum/medplum/pull/9088

packages/server/src/fhir/operations/find.ts
noahsilas: This has started getting used in a couple places around the server package, I'll follow up with a cleanup in a separate commit.

The other alternative that is worth considering would be promoting a utility like this into core - I have had times in places like ProviderUI where I want to refine a type that TS can't infer (such as accessing an optional attribute on an object that was fetched with a search query requiring that attribute's existence). Having a single way we address that across the entire monorepo (node + browser) might be nice.

mattlong: Ah, good point. I hadn't considered that `assert` is only available in node contexts. Adding it to core seems reasonable!

packages/server/src/fhir/operations/find.ts
```diff
@@ -109,69 +149,246 @@ export async function scheduleFindHandler(req: FhirRequest): Promise<FhirRespons
...
+    availability = availability.filter((interval) => {
+      const durationMs = interval.end.getTime() - interval.start.getTime();
+      return durationMs >= schedulingParameters.duration * 60 * 1000;
+    });
+
+    return availability;
   });
 
-  const serviceType = toCodeableReferenceLike(healthcareService);
+  const intersectingAvailability = allAvailability.reduce((acc, val) => overlappingIntervals(acc, val));
```
mattlong: Regarding the sonar issue, TIL the behavior of calling reduce without an initial value:

> If initialValue is not specified, accumulator is initialized to the first value in the array, and callbackFn starts executing with the second value in the array as currentValue. In this case, if the array is empty (so that there's no first value to return as accumulator), an error is thrown.

You're doing that intentionally, so it's fine to resolve the issue as a false positive in Sonar.

It'd probably be good to also add a comment in the code explaining the intentional lack of an initial value

mattlong (approved):

packages/server/src/fhir/operations/find.ts
noahsilas: After doing a little research, apparently a main reason for this rule is that calling reduce without an initial value and an empty array throws a TypeError. I think in this case we're assured to never have an empty array as input, but I hadn't considered that this style of reduce using the initial element is a less known feature (I think I've been on teams with too many folks steeped in a functional-programming background).

I think I can satisfy the linter and make this explicit instead of adding a comment:
```suggestion
  const intersectingAvailability = allAvailability.slice(1).reduce(
    (acc, val) => overlappingIntervals(acc, val),
    allAvailability[0]
  );
```

Somewhat interestingly, if the input array ever is empty, this now would return `undefined`, but we don't have `noUncheckedIndexedAccess` enabled in our TypeScript configuration so the type checking still works out here. Maybe I'll add an assertion afterwards to validate that assumption and make it easier to turn that flag on without revisiting this later.

PR #8799, opened 2026-03-27
feat(provider): Spaces check whether bots and ai are enabled

techdavidy:
- Reads project.features from medplum.getProject()
  - When either feature is missing, renders SpacesInbox blurred and non-interactive in the background
   with an overlay card explaining the feature is unavailable.

Screenshot 2026-03-27 at 8 51 45 AM


techdavidy:
@claude can you update this pr to main

PR #8801, opened 2026-03-27
fix(ccda): add null check for reaction manifestation before setting extension

maximevast:
## Problem

`processReaction` in `ccda-to-fhir.ts` crashes with:

```
TypeError: Cannot set properties of undefined (setting 'extension')
```

when converting CCDA documents that contain allergy reactions with unmappable observation codes.

## Root Cause

`mapCode()` can return `undefined` when a CCDA code element has no system, code value, display, translations, or original text. The manifestation array is initialized as:

```typescript
manifestation: [this.mapCode(reactionObs.value as CcdaCode)] as CodeableConcept[]
```

When `mapCode()` returns `undefined`, this creates `[undefined]`. The existing guard:

```typescript
if (reaction.manifestation && reaction.manifestation.length > 0 && reactionObs.text?.reference?.['@_value']) {
```

passes because the array has length 1, but `manifestation[0]` is `undefined`.

## Fix

Add `reaction.manifestation[0] &&` to the guard before accessing the element:

```typescript
if (reaction.manifestation && reaction.manifestation.length > 0 && reaction.manifestation[0] && reactionObs.text?.reference?.['@_value']) {
```

## Context

We encountered this in production processing real-world EHI (Electronic Health Information) exports. Multiple patient CCDAs had allergy sections with reaction observations whose codes couldn't be mapped, causing the entire CCDA-to-FHIR conversion to fail.

### DIFF
--- a/packages/ccda/src/ccda-to-fhir.ts
+++ b/packages/ccda/src/ccda-to-fhir.ts
@@ -752,7 +752,7 @@ class CcdaToFhirConverter {
     this.processSeverity(reactionObs, reaction);
 
     // Add reaction reference
-    if (reaction.manifestation && reaction.manifestation.length > 0 && reactionObs.text?.reference?.['@_value']) {
+    if (reaction.manifestation?.[0] && reactionObs.text?.reference?.['@_value']) {
       reaction.manifestation[0].extension = this.mapTextReference(reactionObs.text);
     }
 


packages/ccda/src/ccda-to-fhir.ts
```diff
@@ -752,7 +752,7 @@ class CcdaToFhirConverter {
     this.processSeverity(reactionObs, reaction);
 
     // Add reaction reference
-    if (reaction.manifestation && reaction.manifestation.length > 0 && reactionObs.text?.reference?.['@_value']) {
+    if (reaction.manifestation && reaction.manifestation.length > 0 && reaction.manifestation[0] && reactionObs.text?.reference?.['@_value']) {
```
mattwiller: I think these checks could be safely combined:
```suggestion
    if (reaction.manifestation?.[0] && reactionObs.text?.reference?.['@_value']) {
```

maximevast: true thank you!

maximevast:
> Looks good to me, thanks for the PR @maximevast! If you could [sign all the commits on the branch](https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot) to satisfy the DCO check, then I'll be able to merge it

My pleasure; thanks for the great work!

PR #8840, opened 2026-03-31
fix(provider): Display error messages available when submitting claims

techdavidy:
When error ocurres, display errorMessage if available.
Display tooltips when buttons are disabled.


examples/medplum-provider/src/components/encounter/BillingTab.tsx
```diff
@@ -334,18 +334,44 @@ export const BillingTab = (props: BillingTabProps): JSX.Element => {
       setClaim(updatedClaim);
       await fetchCandidEncounter();
     } catch (err) {
-      showErrorNotification(err);
+      let errorMessage: string | undefined;
+      try {
+        const parsed = JSON.parse((err as Error).message);
+        errorMessage = parsed?.errorMessage;
+      } catch {
+        // not a JSON error body
```
codyebberson: We should probably have a generic message here.

```suggestion
        errorMessage = 'Could not parse response';
```

PR #8841, opened 2026-03-31
feat(provider): Coverage eligibility - display request and responses

techdavidy:
Screenshot 2026-03-31 at 3 47 05 PM


examples/medplum-provider/src/components/insurance/EligibilityDetails.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <Table.Td>
+        {typeof value === 'string' ? <Text size="sm">{value}</Text> : value}
+      </Table.Td>
+    </Table.Tr>
+  );
+}
+
+type BenefitItem = NonNullable<NonNullable<NonNullable<CoverageEligibilityResponse['insurance']>[number]['item']>[number]['benefit']>[number];
+
+function formatBenefitValue(benefit: BenefitItem, prefix: 'allowed' | 'used'): string {
```
noahsilas: It looks like this type is exported from fhirtypes as `CoverageEligibilityResponseInsuranceItemBenefit`, if you want to remove a bit of this nested non-nullable application.


```suggestion
function formatBenefitValue(benefit: CoverageEligibilityResponseInsuranceItemBenefit, prefix: 'allowed' | 'used'): string {
```

examples/medplum-provider/src/components/insurance/EligibilityDetails.tsx
```diff
@@ -0,0 +1,249 @@
...
+  if (prefix === 'allowed') {
+    if (benefit.allowedUnsignedInt !== undefined) return String(benefit.allowedUnsignedInt);
```
noahsilas: Should this use something like `toLocaleString()` to get a formatted value?

examples/medplum-provider/src/components/insurance/EligibilityDetails.tsx
```diff
@@ -0,0 +1,249 @@
...
+  }
+  return '—';
+}
+
+function formatMoney(money: { value?: number; currency?: string }): string {
+  if (money.value !== undefined) {
+    return new Intl.NumberFormat('en-US', { style: 'currency', currency: money.currency ?? 'USD' }).format(money.value);
+  }
+  return '—';
+}
```
noahsilas: This looks very similar to core's [`formatMoney() helper`](https://github.com/medplum/medplum/blob/v5.1.6/packages/core/src/format.ts#L480-L490) - can we use that here?

examples/medplum-provider/src/components/insurance/EligibilityDetails.tsx
```diff
@@ -0,0 +1,249 @@
...
+      </Stack>
+    </ScrollArea>
+  );
+}
+
+function RequestSection({ request }: { request: CoverageEligibilityRequest }): JSX.Element {
+  const servicedDate = request.servicedDate
+    ? formatDate(request.servicedDate)
+    : request.servicedPeriod
+      ? `${formatDate(request.servicedPeriod.start)} – ${formatDate(request.servicedPeriod.end)}`
```
noahsilas: Could this use [`formatPeriod`](https://github.com/medplum/medplum/blob/v5.1.6/packages/core/src/format.ts#L300)?

examples/medplum-provider/src/components/insurance/EligibilityDetails.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <Stack gap="md">
+        <Title order={5}>Eligibility Response</Title>
+        <Text size="sm" c="dimmed">
+          No response received yet.
+        </Text>
+      </Stack>
+    );
+  }
+
+  const firstInsurance = response.insurance?.[0];
```
noahsilas: Would it make sense to map over the entries in this list instead of only showing the first entry?

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,372 @@
...
+  const [responseLoading, setResponseLoading] = useState(false);
+  const [checkingEligibility, setCheckingEligibility] = useState(false);
+
+  // Load eligibility bot
+  useEffect(() => {
+    medplum
+      .searchOne('Bot', { identifier: 'https://www.medplum.com/bots|eligibility' })
+      .then((bot) => setEligibilityBot(bot ?? null))
+      .catch(() => setEligibilityBot(null));
+  }, [medplum]);
```
noahsilas: This could get things like loading tracking "for free" by using the `useSearchOne` helper, and replace a `useState`+`useEffect` with a single hook call:


```suggestion
const [eligibilityBot, eligibilityBotLoading] = useSearchOne('Bot', { identifier: 'https://www.medplum.com/bots|eligibility' });
```

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,372 @@
...
+  // Load PractitionerRole for current profile
+  useEffect(() => {
+    if (!profile) {
+      return;
+    }
+    medplum
+      .searchOne('PractitionerRole', { practitioner: getReferenceString(profile) })
+      .then((role) => setPractitionerRole(role ?? null))
+      .catch(() => setPractitionerRole(null));
+  }, [medplum, profile]);
```
noahsilas: This one doesn't convert as nicely into `useSearchOne` today, but I have an [open PR that makes it possible](https://github.com/medplum/medplum/pull/8838):
```ts
const [practitionerRole, practitionerRoleLoading] = useSearchOne(
  'PractitionerRole', 
  profile ? { practitioner: getReferenceString(profile) } : undefined,
  { enabled: !!profile }
);
```

(Similar thoughts for below resources omitted)

noahsilas: Is there always single PractitionerRole per practitioner? I wonder if we should use `searchResources` instead of `searchOne`, and add some kind of alert if there are multiple?

techdavidy: 👀

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,361 @@
...
+      return;
+    }
+    const found = requests.find((r) => r.id === requestId);
+    if (found) {
+      setSelectedRequest(found);
+    } else if (requestId) {
+      medplum
+        .readResource('CoverageEligibilityRequest', requestId)
+        .then(setSelectedRequest)
+        .catch(showErrorNotification);
```
noahsilas, line 108: This feels a little suspicious. I think this probably usually emits an extra query; when this component first mounts, the `requests` state is set to an empty list, so no matching request will be found in it, and this readResource query will fire. It seems likely to get batched with the searchResources query above that will populate the requests list, and which will usually return this CoverageEligibilityRequest instance.

A couple scenarios I'm thinking about
- If the patient has > 100 requests, then the above query might not find this request, in which case this is a helpful guardrail
- If the URL gets mismatched somehow, this could put us into a UI state where the selectedRequest state belongs to a different patient than the entries in `requests` state.
- When the above query sets `requests` into state, this effect will re-run, but at that point the `readResource` query should resolve from cache, so that's probably fine.

After thinking about it some, I think that the guardrail holding against too many results in the other query is a good idea. Maybe we should just add a check the result that the patient matches before setting it into `selectedRequest`? Something like:

```ts
      medplum
        .readResource('CoverageEligibilityRequest', requestId)
        .then((request) => {
          if (resolveId(request.patient) !== patientId) {
            throw new Error("CoverageEligibilityRequest does not match Patient");
          }
          setSelectedRequest(request);
        })
        .catch(showErrorNotification);
```

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,361 @@
...
+    }
+  }, [medplum, requestId, requests]);
+
+  // Fetch response for selected request
+  useEffect(() => {
+    if (!selectedRequest?.id) {
```
noahsilas, line 114: When we have a requestId from the URL we could kick this query off earlier (potentially getting it batched with the other initial fetches on component mount). But that may require some extra handling to make sure that it stays in sync with the `request` state...

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,361 @@
...
+  const handleCheckEligibility = async (): Promise<void> => {
+    if (!eligibilityBot) {
+      showErrorNotification(new Error('To enable Insurance Eligibility please contact support.'));
+      return;
+    }
+    if (!practitionerRole) {
+      showErrorNotification(new Error('No PractitionerRole found for the assigned practitioner.'));
+      return;
+    }
+    if (!coverage || !patientId || !coverageId) {
```
noahsilas: Is there a scenario where `coverage` is set without `coverageId` being set too?

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,361 @@
...
+    }
+    setCheckingEligibility(true);
+    try {
+      const requestBody: CoverageEligibilityRequest = {
+        resourceType: 'CoverageEligibilityRequest',
+        status: 'active',
+        purpose: ['benefits'],
+        created: new Date().toISOString(),
+        patient: { reference: `Patient/${patientId}` },
+        insurer: coverage.payor?.[0] as { reference: string; display?: string },
```
noahsilas: This looks like it is hiding some weaknesses in the type system; I think this is casting from `Reference<Organization | Patient | RelatedPerson>` to `{ reference: string }` in order to be assignable to `Reference<Organization>`.

Maybe we can add a helper to do this in a type safe way? Something along these lines:
```ts
function isReferenceTo<T extends ResourceType>(
  resourceType: T,
  ref: Reference
): ref is Reference<ExtractResource<T>> {
  return ref.reference?.startsWith(`${resourceType}/`);
}

const payor = coverage.payor.find(payor => isReferenceToType('Organization', payor));
// payor has type Reference<Organization> | undefined
if (!payor) { throw new Error(...); }
// payor has type Reference<Organization>
```

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,361 @@
...
+      setCheckingEligibility(false);
+    }
+  };
+
+  const getRequestHref = (req: CoverageEligibilityRequest): string =>
+    `/Patient/${patientId}/Coverage/${coverageId}/CoverageEligibilityRequest/${req.id}`;
+
+  return (
+    <Box w="100%" h="100%">
+      <Flex h="100%">
```
noahsilas, line 188: Does wrapping this `<Flex>` with the Box to specify the full width do something tricky or clever here? Seems like we might be able to unnest this all be removing that outer wrapper.

examples/medplum-provider/src/App.tsx
```diff
@@ -191,6 +192,8 @@ export function App(): JSX.Element | null {
                 <Route path="ServiceRequest/:serviceRequestId" element={<LabsPage />} />
                 <Route path="MedicationRequest" element={<MedicationsPage />} />
                 <Route path=":resourceType" element={<PatientSearchPage />} />
+                <Route path="Coverage/:coverageId" element={<CoveragePage />} />
```
noahsilas, line 195: Is this linked to from somewhere? Or is the intent of this PR to get this page merged and accessible only via deep-link for now?

techdavidy: you can get to that screen from patient summary

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
techdavidy: Thinking about wrapping this into a component after some testing and adding pagination there.

techdavidy:
@claude all feedback handled please review again

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,350 @@
...
+    } catch (error) {
+      showErrorNotification(error);
+    } finally {
+      setRequestsLoading(false);
+    }
+  }, [medplum, patientId, coverageId]);
+
+  useEffect(() => {
+    fetchRequests().catch(showErrorNotification);
+  }, [fetchRequests]);
```
noahsilas, line 92: Hm. Looking at this, my first instinct was to try to rewrite it to `useSearchResources`, but the ergonomics of that are still not great. I think to capture all the work you are doing here it would look something like this:

```ts
const [allRequests, allRequestsLoading, allRequestsOutcome] = useSearchResources(
  'CoverageEligibilityRequest',
  patientId ? new URLSearchParams({
    patient: `Patient/${patientId}`,
    _count: '100',
    _sort: '-_lastUpdated',
  }) : undefined,
  { enabled: !!patientId }
);

useEffect(() => {
  if (allRequestsOutcome && !isOk(allRequestsOutcome)) {
    showErrorNotification(allRequestsOutcome);
  }
}, [allRequestsOutcome]);

const requests = useMemo(() => allRequests?.filter((req) => ...) ?? []), [allRequests]);
```

One thought I have is that the effect taking an outcome and showing an error is probably extractable into a reusable hook of its own, which makes this better, but still sort of awkward.

As I'm looking at the actual usage, I'm considering a potential API improvement: what if we offered a symbol that you could pass as the search query to disable it instead of using the `{ enabled }` flag? That might look something like this:
```ts
const [allRequests, allRequestsLoading, allRequestsOutcome] = useSearchResources(
  'CoverageEligibilityRequest',
  patientId ? new URLSearchParams({
    patient: `Patient/${patientId}`,
    _count: '100',
    _sort: '-_lastUpdated',
  }) : useSearchResources.disabled,
);
```

That avoids duplicating the condition into a guard and a flag, although it doesn't really help with the outcome follow up. Maybe in Provider as an application consideration we could write a wrapping hook:
```ts
import { useSearchResources as useSeachResourcesBase } from '@medplum/react';

export const useSearchResources: typeof useSearchResourcesBase = (resourceType, query, options) => {
  const [resources, loading, outcome] = useSearchResourcesBase(resourceType, query, options);
  useEffect(() => {
    if (outcome && !isOk(outcome)) {
      showErrorNotification(outcome);
    }
  }, [outcome]);

  return [resources, loading, outcome];
}
```

Anyways, writing all this down for discussion, but I don't want to hold up your PR for this, so I'll follow up with some more exploration about how we can keep making both the `@medplum/react` core package and the provider app easier to write good data loading features for.

examples/medplum-provider/src/App.tsx
noahsilas: Ah, I see! I had been looking for references to this URL, but this is coming from a more dynamic source.

https://github.com/medplum/medplum/blob/v5.1.6/examples/medplum-provider/src/pages/patient/PatientPage.tsx#L98 currently tries to navigate to here when a Coverage resource is clicked. I imagine that clicking these today is causing that `navigate` call to error and log an error in the JS console?

Thanks for the pointer!

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
```diff
@@ -0,0 +1,350 @@
...
+  const planName =
+    coverage.class?.find((c) => c.type?.coding?.[0]?.code === 'plan')?.name ??
+    coverage.type?.text ??
+    coverage.type?.coding?.[0]?.display;
+
+  const subscriberId = coverage.subscriberId ?? coverage.identifier?.[0]?.value;
+
+  const periodText = coverage.period
+    ? `${formatDate(coverage.period.start)} – ${formatDate(coverage.period.end)}`
+    : undefined;
```
noahsilas: Does core's `formatPeriod` helper work for this? I think if period is not set that returns `''`, which is still falsey and so will omit the entry in the Stack below.

examples/medplum-provider/src/pages/patient/CoveragePage.tsx
noahsilas: > Thinking about wrapping this into a component after some testing and adding pagination there.

Would this mean that the request resource would always come from that paginated list instead of sometimes being loaded by ID this way?

I think that the main reason I'm feeling slightly funky about this is that the search applies a criteria (patient ID matching) that this query doesn't, and that feels like something I've commonly seen eventually turn into weird data bugs. So if this always comes from that search query, or if we replicate the condition here, either of those options would feel real good to me!

techdavidy:
@noahsilas I think there is an scenario where you can be at page 3, but marked resource id is in page 1. 
You would render the details of the specified id and not selected resource in list.

noahsilas:
> @noahsilas I think there is an scenario where you can be at page 3, but marked resource id is in page 1. You would render the details of the specified id and not selected resource in list.

I see! I had assumed that these were always linked, but the intent is for the search query to power the list UI, and for the fetch query to power the detail view, and with pagination these can be representing two different pieces of state. Thanks for the clarification!

PR #8842, opened 2026-04-01
Add multi-select support to ResourceInput

### LINKED ISSUE #7246: ResourceInput Multiselect
Enable to select multiple or 1 resource in ResourceInput.

fingoltin:
Fixes #7246.

### DIFF
--- a/packages/react/src/ResourceInput/ResourceInput.tsx
+++ b/packages/react/src/ResourceInput/ResourceInput.tsx
@@ -87,9 +87,11 @@ export interface ResourceInputProps<T extends Resource = Resource> {
   readonly required?: boolean;
   readonly itemComponent?: (props: AsyncAutocompleteOption<T>) => JSX.Element | ReactNode;
   readonly onChange?: (value: T | undefined) => void;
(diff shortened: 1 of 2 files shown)


packages/react/src/ResourceInput/ResourceInput.tsx
```diff
@@ -107,6 +109,7 @@ export function ResourceInput<T extends Resource = Resource>(props: ResourceInpu
   const defaultValue = useResource(props.defaultValue, setOutcome);
```
techdavidy, line 109: Thanks for putting this together. 
Default value would require to support multiple values too.

fingoltin: Agreed, do you have any quick tips for supporting an array?

techdavidy: thanks for the input on this issue.
We moved ahead and create a new component for multi-select support.

https://github.com/medplum/medplum/pull/8949

fingoltin: Awesome thank you!!

PR #8847, opened 2026-04-01
fix: support _project search parameter in MockClient/MemoryRepository

### LINKED ISSUE #8831: MockClient: `_project` search parameter not supported in MemoryRepository / matchesSearchFilter
## Summary

`_project=<projectId>` is a Medplum-specific search filter widely used in production bots, but `MockClient`/`MemoryRepository` silently returns zero results whenever it is included in a search query.

## Root cause

`matchesSearchFilter` (in `packages/core/src/search/match.ts`) looks up the filter code in `globalSchema.types[resourceType].searchParams`. When the code is not found it returns `false`, which means the resource **does not match**:

```typescript
// packages/core/src/search/match.ts
function matchesSearchFilter(resource, searchRequest, filter) {
  const searchParam = globalSchema.types[searchRequest.resourceType]?.searchParams?.[filter.code];
  if (!searchParam) {
    return false;   // ← unknown param → every resource is excluded
  }
  ...
}
```

`_project` is not present in the built-in params added in `packages/core/src/types.ts` (which only includes `_lastUpdated`, `_compartment`, `_profile`, `_security`, `_tag`). It is also not defined in any of the `SEARCH_PARAMETER_BUNDLE_FILES` indexed in tests.

## Observed behaviour

A search like:

```typescript
medplum.searchResources('Encounter', {
  status: 'finished',
  _project: projectId,
})
```

returns **0 results** in tests even when matching encounters have been created with `meta.project: projectId`.

A search without `_project` correctly returns the encounters.

## Expected behaviour

`_project=<id>` should filter resources where `meta.project === <id>`, analogous to how the real Medplum server handles the parameter at the database layer.

## Suggested fix

Add `_project` to the built-in per-resource-type search params in `packages/core/src/types.ts`, e.g.:

```typescript
_project: {
  base: [resourceType],
  code: '_project',
  type: 'string',    // or 'reference'
  expression: resourceType + '.meta.project',
} as SearchParameter,
```

Alternatively, `matchesSearchFilter` could treat unknown parameters as "no-op" (return `true`) rather than "exclude" (return `false`), though a targeted fix for `_project` is preferable to preserve the strictness of other unknown-param behavior.

## Impact

Any bot or subscription test that scopes searches with `_project` will silently pass against a universe of all resources, making project-isolation assertions untestable with `MockClient`.

sshiao:
## Summary
Added `_project` search parameter support to `matchesSearchFilter` so that `MockClient`/`MemoryRepository` correctly filters resources by `meta.project`.

Fixes #8831

## What Changed
- Added early return in `matchesSearchFilter` for `_project` filter code before the `globalSchema.types` lookup (which fails since `_project` is not a registered standard search param)
- Added `matchesProjectFilter` function that checks `resource.meta.project` against the filter value
- Supports `EQUALS`, `NOT_EQUALS`, `MISSING`, `PRESENT` operators and comma-separated values (OR semantics), consistent with how other special params like `_compartment` are handled

## Test Results
- Added 6 test cases covering: exact match, non-match, missing meta.project, NOT_EQUALS, comma-separated OR, and MISSING operator
- All 55 tests in `match.test.ts` pass

Generated with [Claude Code](https://claude.ai/claude-code)


codyebberson:
Hi @sshiao  - thanks for submitting.  Apologies for the hiccup.  We had a change that broke PRs from forks due to restricted access to Github Secrets.  I just updated your branch, which should hopefully fix that.  Stay tuned.

packages/core/src/search/match.ts
```diff
@@ -46,6 +46,9 @@ const searchExprCache = new LRUCache<FhirPathAtom>(1000);
  * @returns True if the resource satisfies the search filter.
  */
 function matchesSearchFilter(resource: Resource, searchRequest: SearchRequest, filter: Filter): boolean {
+  if (filter.code === '_project') {
```
codyebberson: Hi @sshiao -

Rather than adding a special case in code here, I think we could fix this by adding the `_project` search parameter definition in `packages/core/src/types.ts`

See here: https://github.com/medplum/medplum/blob/main/packages/core/src/types.ts#L172-L230

I believe that we could add a new entry in that array for `_project`, such as:

```typescript
      _compartment: {
        base: [resourceType],
        code: '_project',
        type: 'reference',
        expression: resourceType + '.meta.project',
      } as SearchParameter,
```

codyebberson (changes requested):
Thanks!  See comment for proposed implementation strategy

sshiao:
thanks for the feedback!  definitely better.  making the change.

PR #8851, opened 2026-04-02
Clean up questionnaire choice type logic

codyebberson:
Push the logic into a single function for: "given this questionnaire item, what UI widget should we use to display it?"


techdavidy:
@claude please review

techdavidy (approved):

PR #8854, opened 2026-04-03
Document how to Execute Clinical Protocols

rahul1:


packages/docs/sidebars.ts
```diff
@@ -95,6 +95,7 @@ const sidebars: SidebarsConfig = {
       label: 'Care Coordination',
       link: { type: 'doc', id: 'careplans/index' },
       items: [
+        { type: 'doc', id: 'careplans/protocols' },
```
everett-williams: I think this doc page may make more sense to put under the 'Clinical Configuration' section given it's about general set up of protocols.

rahul1: That makes sense. I'll make that change

PR #8856, opened 2026-04-03
fix(hl7): wait before destroying HL7 conn socket

ThatOneBro:
In an attempt to deflake a lot of the HL7 related tests, I noticed we destroy the underlying `Socket` immediately after calling `.end()` on it in our `Hl7Connection#close`, which means we don't allow for proper graceful TCP closing in most circumstances via the `FIN` packet.

This is possibly causing a lot of our issues in tests and probably also showing up as timeouts on the other side of the TCP connection which is not ideal.

### DIFF
--- a/packages/hl7/src/connection.ts
+++ b/packages/hl7/src/connection.ts
@@ -30,6 +30,7 @@ export interface SendAndWaitOptions {
 
 export interface Hl7ConnectionOptions {
   messagesPerMin?: number;
+  gracefulCloseTimeoutMs?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/hl7/src/connection.ts
```diff
@@ -306,15 +330,21 @@ export class Hl7Connection extends Hl7Base {
...
-      this.socket.once('close', resolve);
+    await new Promise<void>((resolve) => {
+      const timer = setTimeout(() => {
+        this.socket.destroy();
+      }, this.gracefulCloseTimeoutMs);
+
+      this.socket.once('close', () => {
+        clearTimeout(timer);
+        resolve();
+      });
```
mattwiller, line 347: Can this be rewritten into a more imperative async/await style?  As a general sketch, would something like this work:
```ts
await sleep(this.gracefulCloseTimeoutMs);
this.socket.destroy();
```

ThatOneBro: The idea is that we don't want to wait unless we have to, it's supposed to be a race for either getting the close or hitting the timeout and then destroying the Socket forcefully

PR #8857, opened 2026-04-03
Provider: Schedule from shared scheduling parameters on HealthcareService resources

noahsilas:
In https://github.com/medplum/medplum/pull/8580 we added support to the Scheduling APIs to use HealthcareService resources as shared containers for scheduling parameters.

This means that provider app should start fetching relevant HealthcareService records and offering them as schedulable service types.

We start by pushing more of this feature's logic to be self-contained inside the `<FindPane>` component, and then teach that component to fetch HealthcareService records and list schedulable service types from them in the pane.

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/FindPane.tsx
+++ b/examples/medplum-provider/src/pages/schedule/FindPane.tsx
@@ -4,7 +4,7 @@ import { Button, Group, Stack, Title } from '@mantine/core';
 import type { WithId } from '@medplum/core';
 import { EMPTY, formatDateTime, isDefined } from '@medplum/core';
 import type { Appointment, Bundle, Schedule, Slot } from '@medplum/fhirtypes';
-import { CodeableConceptDisplay, useMedplum } from '@medplum/react';
+import { CodeableConceptDisplay, useMedplum, useSearchResources } from '@medplum/react';
 import { IconChevronRight, IconX } from '@tabler/icons-react';
 import type { JSX } from 'react';
 import { useCallback, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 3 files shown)


examples/medplum-provider/src/utils/scheduling.ts
```diff
@@ -22,3 +22,10 @@ export function serviceTypesFromSchedulingParameters(schedule: Schedule): Codeab
   const serviceTypes = extensions.map((ext) => getExtensionValue(ext, 'serviceType') as CodeableConcept | undefined);
   return serviceTypes.filter(isDefined);
 }
+
+export function hasSchedulingParameters(resource: Schedule | HealthcareService): boolean {
+  if (!resource.extension) {
+    return false;
+  }
+  return resource.extension.some((ext) => ext.url === SchedulingParametersURI);
+}
```
techdavidy, line 28:
```suggestion
 export function hasSchedulingParameters(resource: Schedule |           
  HealthcareService): boolean {
    return !!getExtension(resource, SchedulingParametersURI);
  }
```

examples/medplum-provider/src/pages/schedule/FindPane.tsx
```diff
@@ -13,26 +13,37 @@ import { BookAppointmentForm } from '../../components/schedule/BookAppointmentFo
...
+  serviceTypesFromSchedulingParameters,
+} from '../../utils/scheduling';
 
 const ONE_WEEK_MS = 7 * 24 * 60 * 60 * 1000;
 
 type FindPaneProps = {
   schedule: WithId<Schedule>;
   range: Range;
   onSuccess: (results: { appointments: Appointment[]; slots: Slot[] }) => void;
+  className?: string;
```
techdavidy, line 28: any reason not to have a FindPane.module.css ?

examples/medplum-provider/src/utils/scheduling.ts
noahsilas: Nice, thanks!

examples/medplum-provider/src/pages/schedule/FindPane.tsx
noahsilas: Yeah, the idea here was that `<SchedulePage>` wants to control the width of the `<FindPane>` (https://github.com/medplum/medplum/blob/main/examples/medplum-provider/src/pages/schedule/SchedulePage.module.css#L19-L21).

My intuition here is that the FindPane shouldn't be context-aware of its container, so instead this offers the container a prop to customize the styling if the FindPane renders.

codyebberson:
@claude please review

PR #8858, opened 2026-04-03
Remove inactive ProjectMemberships from logins

maddyli:


packages/server/src/auth/utils.ts
codyebberson, line 125: Should we even return `ProjectMemberships` where `active=false` ?

maddyli: The only reason I would keep it is if someone fatfingers and sets their own ProjectMembership to inactive, they can at least see what happened from the login screen. Otherwise, they just won't be able to see it at all 

But, I do agree. This is to keep some levels transparency/keep current behavior, but I agree that just not returning inactive membership is the cleanest way

packages/react/src/auth/ChooseProfileForm.tsx
```diff
@@ -46,14 +46,20 @@ export function ChooseProfileForm(props: ChooseProfileFormProps): JSX.Element {
...
-    .filter(filterMembership)
-    .slice(0, 10)
-    .map((item) => (
-      <Combobox.Option value={item.id as string} key={item.id} className={classes.option}>
-        <SelectOption {...item} />
-      </Combobox.Option>
-    ));
+  const filtered = props.memberships.filter(filterMembership);
+  const active = filtered.filter((m) => m.active !== false);
+  const inactive = filtered.filter((m) => m.active === false);
```
techdavidy: since active is optional, active flag as undefined would not show.
is that intended  ?

packages/react/src/auth/ChooseProfileForm.tsx
```diff
@@ -106,14 +112,16 @@ function getMembershipLabel(membership: ProjectMembership): string | undefined {
 
 function SelectOption(membership: ProjectMembership): JSX.Element {
   const label = getMembershipLabel(membership);
+  const isInactive = membership.active === false;
   return (
     <Group gap="xs" align="center">
-      <Box className={classes.iconBox}>
+      <Box className={isInactive ? `${classes.iconBox} ${classes.inactiveIcon}` : classes.iconBox}>
```
techdavidy: let's use cx

packages/react/src/auth/ChooseProfileForm.tsx
```diff
@@ -106,14 +112,16 @@ function getMembershipLabel(membership: ProjectMembership): string | undefined {
...
         <IconBriefcase size={16} stroke={2} />
       </Box>
       <div>
-        <Text size="sm" fw={500}>
+        <Text size="sm" fw={500} c={isInactive ? 'dimmed' : undefined}>
```
techdavidy: let's use cx here too

techdavidy:
Thanks for handling the feedback. 
nit: We can change the title of the PR since we changed the logic

PR #8861, opened 2026-04-03
mTLS auth docs

codyebberson:


codyebberson:
@claude please review

codyebberson:
@claude please change the "curl" blocks to "bash" blocks for consistency.

I don't want to go overboard with curl or Python complexity.  We should fix anything that is factually incorrect, but generating the perfect certificate is out of scope for this document.

PR #8852, opened 2026-04-02
Revert "Fix use of modifiers in _filter search"

codyebberson:
Reverts medplum/medplum#8783

PR #8863, opened 2026-04-03
Re-try fix allowing use of modifiers in _filter search

mattwiller:
The crux of the changes is that `parseParameter` now takes both `operator` and `modifier` as inputs so that `operator` can be used as a fallback/default return value.

Let's consider the search

```
Patient?_filter=_has:Observation:performer:_id ne 'some-id'
```

Part of the chain of execution is `buildFilterParameterComparison` --> `buildSearchFilterExpression` --> `parseChainedParameter` --> `parseParameter`

Now, `parseParameter` is called with operator `ne` and modifier as the empty string allowing the correct `ne` to be used.

Context:
Reversion of original PR: #8852 
Original PR: #8783


codyebberson:
@claude please review

packages/core/src/search/search.ts
```diff
@@ -357,14 +357,22 @@ export function parseParameter(searchParam: SearchParameter, modifier: string, v
...
       if (!isValidSearchValue(searchParam, value)) {
         throw new OperationOutcomeError(
           badRequest(`Invalid format for ${searchParam.type} search parameter: ${value}`)
         );
       }
-      return { code: searchParam.code, operator: parseModifier(modifier), value };
-
+      let operator = parseModifier(modifier);
+      if (!operator) {
+        if (modifier === Operator.NOT_EQUALS) {
```
codyebberson: It does seem like this case might not be possible?

If `modifier === Operator.NOT_EQUALS`, then `operator` would have been set on line 366.

packages/server/src/fhir/search.ts
```diff
@@ -1191,8 +1191,11 @@ function trySpecialSearchParameter(
         filter
       );
     }
-    case '_filter':
-      return buildFilterParameterExpression(repo, selectQuery, resourceType, table, parseFilterParameter(filter.value));
+    case '_filter': {
+      const filterExpr = parseFilterParameter(filter.value);
+      return buildFilterParameterExpression(repo, selectQuery, resourceType, table, filterExpr);
```
codyebberson, line 1187: Minor, but it seems like this didn't change?

packages/core/src/search/search.ts
mattwiller: This condition is necessary: if `modifier === Operator.NOT_EQUALS`, then `parseModifier()` would return and set `operator` to `undefined`

codyebberson: Hm 🤔 

If `NOT_EQUALS` is not a valid modifier and `parseModifier()` returns undefined....  what is the use case for that?  Is this just for `_filter`?  Should we just add `NOT_EQUALS` to `MODIFIER_OPERATORS`?

mattwiller: That's exactly what I had done in 549b803759fbadce8090dfa1dc2307242b283ce2, but it broke the test in `@medplum/app`.  Since `ne` is not actually a valid modifier, I think this explicit condition for `_filter` is a cleaner solution

mattlong (approved):

mattlong:
@claude take another look after the addition of the last five commits

codyebberson (approved):

PR #8864, opened 2026-04-04
fix(getwssubstats): update scanned keys for newer Redis Hash key names / payload

ThatOneBro:
After implementing #8596 which changed the payload stored in our WebSocket subscription Redis Hash, we failed to update the `$get-ws-sub-stats` and `$get-ws-sub-project-stats` operations to account for the new key scheme / payload, so the current dashboard shows stale and inaccurate data for long expired subscriptions and not up-to-date counts. This PR fixes that.

I also took the chance to pull some of the useful data in the new format and put it into the UI to make it a bit more informative for subscriptions / criteria:
Screenshot 2026-04-03 at 5 05 04 PM


codyebberson:
@claude please review

PR #8866, opened 2026-04-04
test(agent): use `getFreePort` in all tests

ThatOneBro:
Use `getFreePort` in more tests. Big focus on agent tests

PR #8873, opened 2026-04-04
fix(ci/test): add `turbo` back to tests

ThatOneBro:
It looks like we may have unintentionally removed `turbo` in #8748, which has caused our CI runs to become slower and more flaky. At first it seemed this was just a phase change, which caused a lot of known flaky tests to become even flakier. It did push us to fix those tests. See: #8894, #8866, #8859, #8856.

However, ultimately we do gain a lot of value from `turbo` caching, so this PR attempts to add it back while making sure we are actually testing everything we intend to test.

One thing I noticed is that we historically were not making sure that we test every job of the matrix (Node version, Postgres version, Redis version) as turbo does not automatically account for those types of platform-level things by default. [They recommend writing all platform-related things you intend to affect caching to a file.](https://turborepo.dev/docs/guides/handling-platforms)

This means that before if a job failed and a concurrent job in the matrix succeeded, a subsequent retry of another job could "reuse" the other job's cached run, not actually testing what we thought it was testing.

I have opted to instead introduce an env var which only affects the `@medplum/server` test caching, which is what we intend to test against the matrix anyways. The env var will prevent that "cached reuse" we previously could have seen and ensures we always test every cell in the matrix.

### DIFF
--- a/packages/server/src/fhir/operations/patientmatch.test.ts
+++ b/packages/server/src/fhir/operations/patientmatch.test.ts
@@ -1,0 +1,465 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { WithId } from '@medplum/core';
+import { ContentType } from '@medplum/core';
+import type { Bundle, BundleEntry, Patient } from '@medplum/fhirtypes';
+import express from 'express';
+import request from 'supertest';
+import { initApp, shutdownApp } from '../../app';
+import { loadTestConfig } from '../../config/loader';
+import { initTestAuth } from '../../test.setup';
+import { scoreCandidate } from './patientmatch';
+
+const app = express();
+let accessToken: string;
+
+describe('Patient $match Operation', () => {
+  beforeAll(async () => {
+    const config = await loadTestConfig();
+    await initApp(app, config);
+    accessToken = await initTestAuth();
+  });
+
+  afterAll(async () => {
+    await shutdownApp();
+  });
+
+  test('Returns 400 when resource parameter is missing', async () => {
+    const res = await request(app)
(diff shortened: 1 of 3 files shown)


packages/server/turbo.json
```diff
@@ -12,7 +12,7 @@
       "dependsOn": ["test:seed"],
       "outputs": ["coverage/**"],
       "inputs": ["src/**/*.tsx", "src/**/*.ts"],
-      "env": ["REDIS_PASSWORD_DISABLED_IN_TESTS", "POSTGRES_HOST", "POSTGRES_PORT"]
+      "env": ["REDIS_PASSWORD_DISABLED_IN_TESTS", "POSTGRES_HOST", "POSTGRES_PORT", "SERVER_TEST_CACHE_KEY"]
```
ThatOneBro, line 15: We need this so that we don't share test results between different jobs in the test matrix.

This is important because if one job fails (ie. due to a flaky test) and another one passes, if you re-run the failed one it could potentially use successful package runs from the cache from the other successful job.

.github/workflows/build.yml
```diff
@@ -239,6 +239,9 @@ jobs:
           POSTGRES_HOST: localhost
           POSTGRES_PORT: 5432
           REDIS_PASSWORD_DISABLED_IN_TESTS: 1
+          SERVER_TEST_CACHE_KEY: NODE${{ matrix.node-version }}-PG${{ matrix.pg-version }}-REDIS${{ matrix.redis-version }}
+          NO_COVERAGE: ${{ (matrix.node-version == 22 && matrix.pg-version == 14 && matrix.redis-version == 7) && '' || '1' }}
```
ThatOneBro: Coverage considerably bloats CPU and memory usage, we can get away with not calling jest with `--coverage` for runs where we are not going to use it

codyebberson: clever

PR #8874, opened 2026-04-05
Fix react eslint warnings

codyebberson:


codyebberson:
@claude - please review.  The goal here is for zero functional changes, just fixing React ESLint errors and warnings.

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -87,22 +87,20 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
...
   const autoSubmitRef = useRef<boolean>(autoSubmit);
-  autoSubmitRef.current = autoSubmit;
-
   const optionsRef = useRef(options);
-  optionsRef.current = options;
+
+  useLayoutEffect(() => {
+    searchRef.current = search;
+    timerRef.current = timer;
+    abortControllerRef.current = abortController;
```
noahsilas, line 98: This looks a lot like something that can be simplified with some reusable hooks; on first read it looks like using the `useDebouncedValue` hook could replace the timer work, and then that seems like it would allow a more direct `useEffect` that could contain an abort implementation in a single closure instead of spread out this way.

That's a bigger change than this more mechanical refactoring to clean up the warnings, I'll really dig in and look at it separately a little later.

packages/react/src/AttachmentButton/AttachmentButton.tsx
```diff
@@ -78,6 +78,7 @@ export function AttachmentButton(props: AttachmentButtonProps): JSX.Element {
         ref={fileInputRef}
         onChange={(e) => onFileChange(e)}
       />
+      {/* eslint-disable-next-line react-hooks/refs */}
       {props.children({ onClick, disabled: props.disabled })}
```
noahsilas, line 82: It's sort of interesting that this is getting flagged this way. I guess the linter sees that a function using the ref is getting passed into the children as a render prop, and isn't able to tell that it isn't called during the render step.

Sort of a pain, but this is an interesting intersection of two unusual patterns (refs and render props), and I don't see an obvious way to make them play nicely with the way that the linter wants this to work.

I _think_ this should generally be safe (I have a hard time thinking of a scenario where someone wants a file picker dialog to open on mount, so I don't think anyone will actually call `onClick` during the render phase).

When we're considering breaking changes in the next major version, we might think about if passing a function as a child here is the best way to do this. This might be better handled with something like a context or an imperative handle.

packages/react/src/PlanDefinitionBuilder/PlanDefinitionBuilder.tsx
noahsilas: This looks like it could avoid using a ref entirely:
```ts
  setValue((prevValue) => ({
    ...prevValue,
    [property]: newValue,
  } as PlanDefinition));
```

packages/react/src/PlanDefinitionBuilder/PlanDefinitionBuilder.tsx
```diff
@@ -118,7 +120,9 @@ interface ActionArrayBuilderProps {
 
 function ActionArrayBuilder(props: ActionArrayBuilderProps): JSX.Element {
   const actionsRef = useRef(props.actions);
-  actionsRef.current = props.actions;
+  useLayoutEffect(() => {
+    actionsRef.current = props.actions;
+  });
```
noahsilas: This looks like it's only used inside functions that are recreated each render, I think we can remove this ref and let those close over `props.actions` directly?

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -87,22 +87,20 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
...
   const abortControllerRef = useRef<AbortController>(abortController);
-  abortControllerRef.current = abortController;
-
   const autoSubmitRef = useRef<boolean>(autoSubmit);
-  autoSubmitRef.current = autoSubmit;
-
   const optionsRef = useRef(options);
-  optionsRef.current = options;
+
+  useLayoutEffect(() => {
```
mattlong, line 95: These `useRef` are ostensibly to avoid adding as many entries to deps array of `handleTimer` and `handleSearchChange`? Is there a particular reason its problematic for those functions to be redefined as often? It seems like at least a few of the refs could pretty easily be removed altogether, but it's unclear how they're load bearing or not.

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -87,22 +87,20 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
...
   const timerRef = useRef<number>(timer);
-  timerRef.current = timer;
-
   const abortControllerRef = useRef<AbortController>(abortController);
-  abortControllerRef.current = abortController;
-
   const autoSubmitRef = useRef<boolean>(autoSubmit);
-  autoSubmitRef.current = autoSubmit;
-
   const optionsRef = useRef(options);
```
mattlong: `optionsRef` isn't read anywhere and can be deleted.

packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -133,7 +133,9 @@ export function SearchControl(props: SearchControlProps): JSX.Element {
   });
 
   const stateRef = useRef(state);
-  stateRef.current = state;
```
mattlong, line 136: I'm pretty sure `stateRef` can also just be completely removed. Everywhere else it is being used, could be replaced with the update version of `setState`, e.g. `setState((prev) => ({ ...prev, searchResponse: response }));`

packages/react/src/PlanDefinitionBuilder/PlanDefinitionBuilder.tsx
```diff
@@ -54,7 +54,9 @@ export function PlanDefinitionBuilder(props: PlanDefinitionBuilderProps): JSX.El
   }
 
   const valueRef = useRef<PlanDefinition>(value);
-  valueRef.current = value;
+  useLayoutEffect(() => {
```
mattlong: As mentioned offline, seems like valueRef could similarly just go away by using the updater version of `setValue`

packages/react/src/PlanDefinitionBuilder/PlanDefinitionBuilder.tsx
```diff
@@ -118,7 +120,9 @@ interface ActionArrayBuilderProps {
 
 function ActionArrayBuilder(props: ActionArrayBuilderProps): JSX.Element {
   const actionsRef = useRef(props.actions);
-  actionsRef.current = props.actions;
+  useLayoutEffect(() => {
```
mattlong: `actionsRef` can go away and just be replaced by `props.action`?

codyebberson:
@mattlong  @noahsilas  - thanks for review.

- AsyncAutocomplete: Agreed this likely wants a more effect-driven/debounced structure, but that feels like a follow-up refactor rather than an ESLint-only change.
- AttachmentButton: Agreed, this looks like a linter limitation around render props + refs; I don’t see a safe cleaner option in this PR.
- PlanDefinitionBuilder: Good catch. These refs do look removable with updater-form state / direct closure over props. I can either fold that into this PR as a no-op simplification or leave it for follow-up.
- SearchControl: Possibly, but that one is a bit more involved because the ref is used from async completions and several handlers. I’d rather keep this  PR mechanical unless we want to refactor it more deliberately.

codyebberson:
@claude please re-review

PR #8889, opened 2026-04-06
Clean up Scheduling service-type param usage and docs

noahsilas:
During the original $find implementation, it seemed like the `service-type` parameter could take zero, one, or many values, so we set its cardinality to `0..*`.

- In commit b98f2b055 we made `service-type` a required parameter, changing the lower bound of the cardinality from 0 to 1. I neglected to update the documentation to reflect this, so we do that now.

- I had thought that `service-type` could take multiple values, but we're implementing this by taking a single string value and using commas as delimiters between tokens.

  This is in line with how FHIR search parameters generally work; a single parameter with comma separated values is treated as a disjunction over those values ("or" style query), while specifying the same parameter multiple times is a conjunction ("and" style query).

  That is, `?service-type=foo&service-type=bar` has a semantically different meaning than `?service-type=foo,bar`; it should match service types that include codes "foo" and "bar", instead of "foo" or "bar".

  We're not currently supporting the conjunctive form, so I'm reducing the upper bound of the cardinality from `*` to 1. I'm also updating the FindPane in the provider application so that it matches this updated understanding.

- While looking at the $find operation definition, I saw that I also misinterpreted the `type` attribute; this is used to flag when the operation is available at the type level. In this case, `type: true` implies that `/Schedule/$find` should be a valid endpoint, but it is not. Let's correct that to `false` while we are here.


codyebberson:
@claude please review

PR #8890, opened 2026-04-06
fix(ci): stop double caching node modules

ThatOneBro:
After v6 of the `setup-node` action, node modules for NPM are cached by default. That means we download the same cache twice from two different caches. 

See: https://github.com/actions/setup-node#:~:text=Caching%20is%20now,the%20cache%20input.

This PR stops that behavior.

I considered letting the `setup-node` do the caching and removing our explicit use of `actions/cache`, however, `setup-node` is very strict about restoring caches with only strict matching of `package-lock.json`, which diverges from our current caching behavior of falling back to prefix matches. There is an open issue for this behavior on the `setup-node` repo, but it looks like the maintainers are strictly against this and recommend disabling automatic caching and explicitly using the `actions/cache` action like we are already doing.

See: https://github.com/actions/setup-node/issues/328

### DIFF
--- a/.github/workflows/autofix-ci.yml
+++ b/.github/workflows/autofix-ci.yml
@@ -22,6 +22,21 @@ jobs:
       - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
         with:
           node-version: '24'
+          # After v6, the default for `package-manager-cache` is true. We however, have historically used caching with a fallback which is not supported by the setup-node action
+          # See: https://github.com/actions/setup-node/issues/1120
+          package-manager-cache: false
+
+      - name: Cache node modules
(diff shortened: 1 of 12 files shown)


mattwiller (approved):

codyebberson (approved):
👍 This makes sense.

Even with `npm ci`, caching `~/.npm` is valuable because it avoids re-downloading unchanged packages. The key detail is that we’re not reusing `node_modules`, we’re reusing the npm cache.

Ideally we’d rely on `actions/setup-node`’s built-in caching, but it currently behaves like an exact match cache. That means small lockfile changes lead to full cache misses and unnecessary network fetches.

This fallback strategy is a pragmatic workaround that improves cache hit rates. We should revisit once `setup-node` supports partial/fallback restore semantics.

PR #8899, opened 2026-04-07
test(server): fix flaky validate client cert test

ThatOneBro:
This test was flaky for what seems like no good reason, although it is possible that if the character that you were replacing was already 'X' that it would cause this `validateClientCert` not to throw. This PR hopefully should fix that flakiness

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -248,6 +248,9 @@ jobs:
           POSTGRES_HOST: localhost
           POSTGRES_PORT: 5432
           REDIS_PASSWORD_DISABLED_IN_TESTS: 1
+          SERVER_TEST_CACHE_KEY: NODE${{ matrix.node-version }}-PG${{ matrix.pg-version }}-REDIS${{ matrix.redis-version }}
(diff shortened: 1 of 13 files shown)


packages/server/src/oauth/cert.test.ts
```diff
@@ -95,9 +95,10 @@ describe('Certificate validation', () => {
       for (let i = 0; i < lines.length; i++) {
         // Find a line with substantial base64 data (not BEGIN/END markers, and long enough)
         if (lines[i].length > 20 && !lines[i].includes('BEGIN') && !lines[i].includes('END')) {
-          // Replace a character in the middle
+          // Replace a character in the middle with a different character
           const mid = Math.floor(lines[i].length / 2);
-          lines[i] = lines[i].substring(0, mid) + 'X' + lines[i].substring(mid + 1);
+          const replacement = lines[i][mid] === 'X' ? 'Y' : 'X';
```
mattwiller, line 100: Could replace with a character that isn't in base64 to make this easier, e.g. `=`

PR #8907, opened 2026-04-08
Fixed react test warnings

codyebberson:
This PR removes the noisy React test warnings in `@medplum/react`

The main functional fix is in `packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx`. `QuestionnaireForm` was invoking parent `onChange` handlers during render/initialization, which caused "Cannot update a component while rendering a different component" warnings in `NewTopicDialog`. The form now stages pending response changes and emits them from an effect after render.

The `Spotlight` warning was fixed in `packages/react/src/AppShell/Spotlight.tsx` by removing the unsupported `inputProps` nesting and passing the password-manager ignore attributes directly through Mantine’s supported `searchProps` input props.

The remaining warning cleanup was test-focused:

- `packages/react/src/chat/ThreadInbox/ThreadInbox.test.tsx` now awaits initial async render work and injects a mock navigate function into `MedplumProvider`, which eliminates the `act(...)` warnings and the `jsdom` navigation to another Document errors.
- `packages/react/src/chat/ThreadInbox/ParticipantFilter.tsx` now cancels pending debounced searches on cleanup.
- `packages/react/src/chat/ThreadInbox/ParticipantFilter.test.tsx` now uses fake timers with `userEvent` timer advancement and flushes pending timers in `act(...)`, which removes the debounce/transition warning noise.

### DIFF
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -10,7 +10,7 @@ import {
   useQuestionnaireForm,
 } from '@medplum/react-hooks';
 import type { JSX } from 'react';
-import { useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { Form } from '../Form/Form';
 import { SubmitButton } from '../Form/SubmitButton';
 import { SignatureInput } from '../SignatureInput/SignatureInput';
@@ -34,13 +34,13 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element | 
   const medplum = useMedplum();
   const [signatureRequiredSubmitted, setSignatureRequiredSubmitted] = useState(false);
   const propsRef = useRef(props);
+  const pendingChangeRef = useRef<QuestionnaireResponse | undefined>(undefined);
   useLayoutEffect(() => {
     propsRef.current = props;
   });
 
   const onFormChange = useCallback((response: QuestionnaireResponse) => {
-    setSignatureRequiredSubmitted(false);
-    propsRef.current.onChange?.(response);
+    pendingChangeRef.current = response;
   }, []);
 
   const formState = useQuestionnaireForm({
@@ -55,6 +55,17 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element | 
   const formStateRef = useRef(formState);
   useLayoutEffect(() => {
     formStateRef.current = formState;
+  });
+
+  useEffect(() => {
+    const pendingChange = pendingChangeRef.current;
+    if (!pendingChange) {
+      return;
+    }
+
+    pendingChangeRef.current = undefined;
+    setSignatureRequiredSubmitted(false);
+    propsRef.current.onChange?.(pendingChange);
   });
 
   const isSignatureRequired = useMemo(() => {
--- a/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
+++ b/packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
@@ -57,6 +57,23 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element | 
     formStateRef.current = formState;
   });
 
+  // Intentionally run after every commit.
+  //
+  // `useQuestionnaireForm` currently invokes its `onChange` callback while the form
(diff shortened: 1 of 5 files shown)


packages/react/src/chat/ThreadInbox/ThreadInbox.test.tsx
```diff
@@ -45,28 +47,39 @@ describe('ThreadInbox', () => {
...
+          <MemoryRouter>
+            <MedplumProvider medplum={medplum} navigate={mockNavigate}>
+              {children}
+            </MedplumProvider>
+          </MemoryRouter>
+        )
+      );
+
+      await Promise.resolve();
+      await Promise.resolve();
```
ThatOneBro: Any idea why these two `Promise.resolve`s are required? 👀

codyebberson: Those two lines were there to flush two microtask turns inside the act(...) block.

Why that helped in this test:

- ThreadInbox mounts and useThreadInbox immediately runs an effect.
- That effect calls an async function and then schedules more state updates off resolved promises:
    - medplum.search(...)
    - sometimes medplum.graphql(...)
    - then .catch(...) / .finally(...) with setLoading(false)
- In these tests, those mocks resolve immediately, so the work lands in the microtask queue rather than a later timer tick.

Why two await Promise.resolve() calls instead of one:

- One microtask turn was not always enough to get through the full chain.
- The first flush lets the initial async continuation run.
- The second flush lets the next .then(...) / .finally(...) state updates settle before leaving act(...).

So the short version is: it was a cheap way to keep the initial async mount work inside act, which stopped the not wrapped in act(...) warnings.

That said, it is a bit hacky. It is not expressing intent very clearly, and it depends on the current promise chain depth. A cleaner version would be one of these:

- explicitly wait for a stable UI condition in each test
- make setup() wait for a specific first-render state
- use a small helper like flushMicrotasks() with a comment explaining why

codyebberson:
@claude please review

packages/react/src/chat/ThreadInbox/ParticipantFilter.tsx
```diff
@@ -92,6 +92,7 @@ export function ParticipantFilter(props: ParticipantFilterProps): JSX.Element {
 
   useEffect(() => {
     debouncedSearch(searchQuery);
+    return debouncedSearch.cancel;
```
noahsilas, line 95: I think that with #8838 this can probably be rewritten in terms of `useSearch` so it can use the debouncer built in to that, but that's a more invasive change... and potentially going to be a pain due to that search tool being more type-opinionated... I might try to do that update separately from this PR.

packages/react/src/chat/ThreadInbox/ThreadInbox.test.tsx
noahsilas: I'm under the impression that this isn't needed when you have used the `await act(async () => { ... })` wrapping form. I think that in this scenario, React does some tricky tracking of events being scheduled that _should_ be able to avoid this pattern. (There are some scenarios where this doesn't work, I think particularly around macrotask scheduling like `setTimeout`, but it doesn't sound like what we're doing here should be hitting those constraints?)

Is it possible that these are leftover from an iteration where you wren't using the inner async function wrapper?

packages/react/src/QuestionnaireForm/QuestionnaireForm.tsx
```diff
@@ -57,6 +57,34 @@ export function QuestionnaireForm(props: QuestionnaireFormProps): JSX.Element |
...
+  // `pendingChangeRef`, and this effect flushes it after commit. The effect clears
+  // the ref before calling `setSignatureRequiredSubmitted(false)`, so the state
+  // update does not create an infinite loop: the rerender triggered by `setState`
+  // immediately exits because there is no longer a pending change to flush.
+  //
+  // A more complete fix would be to move `useQuestionnaireForm`'s `onChange`
+  // emission out of render entirely. Until then, this effect must run on every
+  // commit so it can detect newly staged ref-based changes that do not participate
+  // in React's dependency tracking.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
noahsilas, line 76: Tricky, thanks for the detailed notes!

noahsilas (approved):

packages/react/src/chat/ThreadInbox/ThreadInbox.test.tsx
codyebberson: Good catch guys.  Fixed.

PR #8915, opened 2026-04-08
Scheduling Provider UI Fixups

noahsilas:
- Some attempts to optimize React-Big-Calendar usage; this still is taking a long time to render on my laptop, but it helps some.
- Noticed that our "transient" ID system was being persisted. This is intended to be a UI only concern; started undecorating the virtual slots before passing them into $book.


techdavidy:
@claude please review

ThatOneBro (approved):

PR #8928, opened 2026-04-09
Added separate rate limit for login paths

codyebberson:


packages/server/src/ratelimit.ts
```diff
@@ -119,3 +147,13 @@ function getRateLimitForRequest(req: Request, config?: MedplumServerConfig): num
 
   return limit;
 }
+
+function getRateLimitCategory(req: Request): RateLimitCategoryConfig {
+  const url = req.originalUrl;
+  for (const category of categories) {
```
ThatOneBro: We could avoid an extra function call here for most requests if we, instead of iterating all categories, do a numeric for loop from 0 to categories.length - 1 👀 

Since we already have the fallback
```suggestion
  for (let i = 0; i < categories.length - 2; i++) {
    const category = categories[i];
```

ThatOneBro (approved):

codyebberson:
@copilot  - can you please make @ThatOneBro 's requested change?

codyebberson:
Deployed to staging.  Demonstrated it works, but the UI is kinda ugly:

image

ThatOneBro:
Eh, ugly UI is probably ok for something that most people should almost never see

PR #8931, opened 2026-04-09
optimize(core): cache `Hl7Message`: `toString()`; `getAllSegments()`

ThatOneBro:
This PR optimizes `Hl7Message#parse` and a few of the other methods by lazily parsing the raw segments only when needed to (when a segment is actually accessed) and caches the results of a segments `toString`, invalidating cache only on modification.


ThatOneBro:
Here are some benchmarks comparing "main" (latest before this PR) and "local" (this branch). The messages look like this:
```
  "MSH|^~\\&|SENDING_APP|SENDING_FAC|REC_APP|REC_FAC|20240218153044||ORU^R01|MSG00001|P|2.5.1",
    "PID|1||123456^^^MRN||DOE^JOHN^A||19800101|M|||123 MAIN ST^^ANYTOWN^NY^12345||555-555-1234",
    "PV1|1|I|ICU^101^A|E|||1234^SMITH^JAMES^A^^^MD|||MED||||7|||1234^SMITH^JAMES^A^^^MD|IP||||||||||||||||||||||||||20240218120000",
    "ORC|RE|ORD001|FIL001||CM|||||||1234^SMITH^JAMES^A^^^MD",
    "OBR|1|ORD001|FIL001|80053^COMPREHENSIVE METABOLIC PANEL^CPT|||20240218120000|||||||20240218130000||1234^SMITH^JAMES^A^^^MD||||||20240218150000|||F",
```
with a variable number of OBX segments which we use for testing `getAllSegments()` with varying number of segments.

`Hl7Message.parse`
Screenshot 2026-04-15 at 2 31 10 PM

`Hl7Message.parse` + `getAllSegments('OBX')`
Screenshot 2026-04-15 at 2 35 00 PM

`Hl7Message.parse` + `toString()` x 2
Screenshot 2026-04-15 at 2 35 12 PM

`Hl7Message.parse` + mutate (busts cache for toString) + `toString()` x 2
Screenshot 2026-04-15 at 2 35 20 PM

`Hl7Message.parse`  + `getAllSegments('OBX')` + `toString()` x2
Screenshot 2026-04-15 at 2 35 28 PM

codyebberson:
@claude please review

ThatOneBro:
@claude Please re-review based on your previous feedback and my latest commit which should address all your points

packages/core/src/hl7.ts
```diff
@@ -86,24 +85,67 @@ export class Hl7Context {
...
 export class Hl7Message {
   readonly context: Hl7Context;
-  readonly segments: Hl7Segment[];
+  /**
+   * Internal lazy-parsed segment storage. Entries are `string` until they
+   * are accessed (via {@link getSegment}, {@link getAllSegments}, {@link header},
+   * or the public {@link segments} getter), at which point they are parsed in
+   * place and replaced with their {@link Hl7Segment} form.
+   */
+  private readonly _segments: (Hl7Segment | string)[];
```
noahsilas, line 94: Would it make sense to split this into two sparse arrays whose types can be tighter? Very roughly, I'm imagining something like:

```ts
  private _unparsedSegments: string[];
  private _parsedSegments: Hl7Segment[];

  function getSegment(index: number): Hl7Segment | undefined {
    if (!_parsedSegments[index] && _unparsedSegments[index]) {
      _parsedSegments[index] = parseSegment(_unparsedSegments[index]);
      delete _unparsedSegments[index]; // free string now that we have a parsed version
    }
    return _parsedSegments[index];
  }
```

I think this could reduce some of the type casting used which would let the type checker provide a little more assurance.

packages/core/src/hl7.ts
```diff
@@ -86,24 +85,67 @@ export class Hl7Context {
...
+  get segments(): Hl7Segment[] {
+    if (!this.allSegmentsParsed) {
+      for (let i = 0; i < this._segments.length; i++) {
+        if (typeof this._segments[i] === 'string') {
+          this.parseSegment(i);
+        }
+      }
+      this.allSegmentsParsed = true;
+    }
+    return this._segments as Hl7Segment[];
```
noahsilas, line 140: I think this might be unsound - you've validated that the array currently holds only Hl7Segments, but I think this returns a reference to an array that can have strings pushed into it later, so a caller iterating over this result in an async manner might suddenly break.

ThatOneBro: While theoretically true the interface does not support strings being pushed in after initialization of the class, and type system would tell you this is wrong

packages/core/src/hl7.ts
ThatOneBro: Sparse arrays do typically trigger de-optimization for V8, since this is primarily a performance-focused PR I think we would want to check that it doesn't impact the performance significantly

packages/core/src/hl7.ts
noahsilas: Ah, I see, at init time you can receive strings but `setSegment()` called after that only takes an Hl7Segment. Nice, thanks for the clarification!

packages/core/src/hl7.ts
```diff
@@ -270,25 +354,65 @@ export class Hl7Message {
...
+
+  private findSegmentIndexByName(name: string): number {
+    return this.segmentsByName.get(name)?.[0] ?? -1;
+  }
+
+  private invalidateCache(): void {
+    this.segmentsByName = buildSegmentMap(this._segments);
+    this.cachedString = undefined;
+    this.bindSegments();
+    // NOTE: We don't clear allSegmentsMap here because beyond this point all new segments have to be `Hl7Segment` types, not strings
```
noahsilas:
```suggestion
    // NOTE: We don't clear allSegmentsParsed here because beyond this point all new segments have to be `Hl7Segment` types, not strings
```

packages/core/src/hl7.ts
```diff
@@ -86,24 +85,67 @@ export class Hl7Context {
...
+    return this._segments as Hl7Segment[];
   }
 
   /**
    * Returns the HL7 message header.
    * @returns The HL7 message header.
    */
   get header(): Hl7Segment {
-    return this.segments[0];
+    return this.parseSegment(0) as Hl7Segment;
```
noahsilas: I think this cast is revealing a potential problem - we don't require instantiation with any elements, so this hypothetically could return undefined if this was instantiated empty and expects the header segment to be appended later.

Maybe we should detect if this is undefined and throw an error instead of casting?

```suggestion
    const headerSegment = this.parseSegment(0);
    if (!headerSegment) {
      throw new Error("Can't get header before first segment added");
    }
    return headerSegment;
```

packages/core/src/hl7.ts
```diff
@@ -270,25 +354,65 @@ export class Hl7Message {
     }
 
     if (typeof index === 'number') {
-      if (index >= this.segments.length) {
+      if (index >= this._segments.length) {
         // Append as last segment
-        this.segments.push(segment);
+        this._segments.push(segment);
+        this.invalidateCache();
```
noahsilas, line 364: How many segments are we dealing with in one message? If the size gets large it might be worthwhile to special case these operations to avoid rebuilding things like `segmentsByName`. (This "add a last segment" case is relatively lightweight, just needing to append an index to one entry, the update cases are more troublesome).

packages/core/src/hl7.ts
```diff
@@ -395,10 +585,9 @@ export class Hl7Segment {
    * @returns The parsed HL7 segment.
    */
   static parse(text: string, context = new Hl7Context()): Hl7Segment {
-    return new Hl7Segment(
-      text.split(context.fieldSeparator).map((f) => Hl7Field.parse(f, context)),
-      context
-    );
+    const segment = new Hl7Segment(text.split(context.fieldSeparator), context);
+    segment.cachedString = text;
```
noahsilas, line 593: Nice trick to reuse the input text as the cached string!

packages/core/src/hl7.test.ts
```diff
@@ -679,3 +679,830 @@ describe('HL7 Setter Functions', () => {
...
+    const msg = Hl7Message.parse(text);
+
+    // getAllSegments triggers per-segment parsing but does not mutate, so the
+    // message-level cached string must still be the original input.
+    msg.getAllSegments('OBX');
+    expect(msg.toString()).toBe(text);
+    expect(msg.toString()).toBe(text);
+  });
+
+  test('mutating non-OBX fields produces the exact expected toString output', () => {
```
noahsilas, line 1064: Just for my education, is there something special about OBX fields (other than this field being repeated in the `buildSampleMessage` helper)?

packages/core/src/hl7.test.ts
```diff
@@ -679,3 +679,830 @@ describe('HL7 Setter Functions', () => {
...
+  test('Parsed message returns original string without rebuilding', () => {
+    const text =
+      'MSH|^~\\&|Main_HIS|XYZ_HOSPITAL|iFW|ABC_Lab|20160915003015||ACK|9B38584D|P|2.6.1|\r' +
+      'MSA|AA|9B38584D|Everything was okay dokay!|';
+    const msg = Hl7Message.parse(text);
+
+    // First call returns original text
+    expect(msg.toString()).toBe(text);
+    // Second call returns same result
+    expect(msg.toString()).toBe(text);
```
noahsilas, line 788: My initial thought was that this was doing an identity check and so showed that we weren't rebuilding the string from components, but on second thought I think that matching strings share the same identity, so maybe this doesn't truly test that? But I don't really have a great way to test the underlying behavior... we could split the reconstruction into a helper like `_toString()` and add a spy on it, but that feels a little like having the tests drive the implementation. No great answer jumps to my mind here, I think this is probably fine as-is.

packages/core/src/hl7.test.ts
```diff
@@ -679,3 +679,830 @@ describe('HL7 Setter Functions', () => {
...
+    }
+    const subsequentNs = Number(process.hrtime.bigint() - subsequentStart) / repeats;
+
+    // Sanity: the cached value matches the rebuilt one and the expected mutation.
+    expect(msg.toString()).toBe(firstResult);
+    expect(msg.toString()).toBe(expected);
+
+    // Ratio assertion. Subsequent calls should be returning a cached string,
+    // which is dramatically faster than re-joining hundreds of segments.
+    const ratio = firstNs / Math.max(subsequentNs, 1);
```
noahsilas, line 1459: Double checking my understanding: `Math.max` here is preventing a division by zero in the case where we took less than 1 nanosecond?

noahsilas (approved):
Thanks for great test coverage!

packages/core/src/hl7.test.ts
ThatOneBro: Nothing really special, they are just a commonly larger field that are often repeated

packages/core/src/hl7.ts
ThatOneBro: This is a good point. In reality almost all `Hl7Message`s are constructed via the static `Hl7Message.parse` but  that doesn't mean we should over-index on that.

We should probably move the check for MSH (the header segment that is required for all HL7v2 messages) into the constructor of `Hl7Message`

packages/core/src/hl7.ts
ThatOneBro: I think on average an HL7 message is on the order of 10 segments? We were seeing some odd messages that were 10,000 segments, but seemed almost 100% to be an error. We just didn't want very large messages to clog up the pipes, so much of this stuff is probably diminishing returns with normal messages

I agree though, that "append to the end" case for both segments and fields seems like it might be a fun follow-up to go through and prevent a lot of unnecessary duplicate work

packages/core/src/hl7.test.ts
ThatOneBro: Yeah, you're right. The framing of this test is probably incorrect. All this proves is that whatever `toString` is doing is consistent if no modifications are made

But going to leave it alone for now...

PR #8932, opened 2026-04-09
fix(health-gorilla): strip non-Coding fields from diagnoses to prevent lab order submission failures

deam65:
## Problem

When a clinician submitted a lab order with ICD-10 diagnoses, the order failed silently: the "Error creating lab order" toast appeared, no order was sent to Health Gorilla, and no `AuditEvent` was recorded for the `send-to-health-gorilla` bot.

Investigation revealed orphaned per-test `ServiceRequest` resources in Medplum with `basedOn` references pointing to a root `ServiceRequest` that did not exist. CloudWatch logs surfaced the root cause:

  `"Invalid additional property \"designation\" (ServiceRequest.reasonCode[1].coding[0].designation)"`

## Root Cause

`ValueSetAutocomplete` returns `ValueSetExpansionContains` objects from the terminology server. These include fields that are not valid on the FHIR `Coding` type (`designation`, `abstract`, `inactive`, `contains`). In both `OrderLabsPage`, the `ValueSetAutocomplete` `onChange` handler was casting these objects directly as `Coding` with no sanitization:

  `items.map((item) => ({ coding: [item] }))`

When the selected ICD-10 code had a `designation` array in its terminology expansion, the root lab order `ServiceRequest` failed FHIR schema validation (422) on write. Because the project did not have the `transaction-bundles` feature flag enabled, the batch processed entries independently: per-test `ServiceRequest`s committed as orphans, the root `ServiceRequest` was never created, and the bot was never triggered.

## Fixes

**`examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx`**
- Updated the diagnoses `onChange` handler to ensure only valid `Coding` fields are passed to `setDiagnoses`.

### DIFF
--- a/packages/health-gorilla-core/src/lab-order.test.ts
+++ b/packages/health-gorilla-core/src/lab-order.test.ts
@@ -314,6 +314,58 @@ describe('createLabOrderBundle', () => {
     ]);
   });
 
+  test<TestContext>('strips non-Coding fields from diagnoses (e.g. designation from ValueSetExpansionContains)', async (ctx) => {
+    const { medplum, patient, requester, performingLab } = ctx;
+
+    const selectedTests: TestCoding[] = [{ code: TEST_CODES[0] } as TestCoding];
+    const testMetadata: Record<string, LabOrderTestMetadata> = { [TEST_CODES[0]]: {} };
+
+    // Simulate what ValueSetAutocomplete returns: a ValueSetExpansionContains with stray fields
+    // that are not valid on a FHIR Coding (designation, abstract, inactive, contains).
+    const diagnosisWithStrayFields = {
+      system: 'http://hl7.org/fhir/sid/icd-10-cm',
+      code: 'J06.9',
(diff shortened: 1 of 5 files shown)


packages/health-gorilla-core/src/lab-order.ts
rahul1, line 1: Would not change these - since diagnoses SHOULD already be the right types, we don't need to sanitize here

packages/health-gorilla-core/src/lab-order.test.ts
rahul1, line 1: I think this is overkill to defend against this. 

really, we should have caught this issue at the level in the UX components

deam65:
Branch only includes changes at the Provider app level so this can be reflected within hosted Medplum Provider as soon as it's merged 

https://github.com/medplum/medplum/pull/8933 contains the proper package exports and usage for a later release

rahul1 (approved):

PR #8941, opened 2026-04-11
feature(data warehouse): export from Postgres to Parquet / Iceberg (S3 Tables)

The-Alchemist:
## Summary

Introduces `packages/server/data-warehouse`, meant to run as a background worker. It uses an in-process **DuckDB** instance to:
1.  attach to **Medplum Postgres**, 
2. read rows from the `"AuditEvent"`, etc. tables
3. write results either to a local sink( **Parquet** files on disk), or to **AWS S3 Tables** as a managed **Iceberg** table.

This is intentionally narrow scope and only serves as an MVP:

### Out of Scope
* `packages/cdk` changes for `s3tables` will be a separate PR
* pre-deployment migrations to create tables in AWS S3 Tables will be a separate PR as well

## How it works

1. **CLI** (`src/index.ts`): Loads env (dotenv), validates Postgres connection (URL or host/name/user/password), requires `--start-window` / `--end-window`, and requires either `--s3-bucket` (Parquet) or `--aws-s3-table-arn` (Iceberg on S3 Tables).
1. **Local sink mode** (no S3 Table ARN): `COPY (SELECT * FROM pg_db."AuditEvent" WHERE lastUpdated …)` to `s3://<bucket>/audit_events/window_<start>_<end>.parquet` (sanitized timestamps in the filename).
1. **S3 Tables / Iceberg mode** (with ARN): **ATTACH** Iceberg with `ENDPOINT_TYPE s3_tables`, ensure `default` schema and `audit_events` table (created empty from Postgres schema with `LIMIT 0`), **DELETE** rows in the same time window from the Iceberg table, then **INSERT** the window’s rows from Postgres—so re-runs for the same window replace that slice.

Tests use **Testcontainers** Postgres and a local path instead of real S3 where needed (`export.test.ts`, `export.unit.test.ts`).

---

## Assumptions & limitations

- **One Postgres table → one lake table / file set:** The MVP maps `"AuditEvent"` to `audit_events` (Iceberg) or `audit_events/` Parquet prefixes only; generalizing to many FHIR tables is not in this change.
- **Sharding / scale:** Production Medplum may use **sharded Postgres** or very large tables; this tool assumes a single attachable database URL and does not coordinate shards, partition pruning across nodes, or parallel writers. **Sharding and multi-tenant isolation would need explicit design** (per-shard jobs, catalog layout, conflict-free windows) but are **not required** to validate this MVP.
- **Correctness of idempotency** on the Iceberg path depends on **`lastUpdated`** alignment with the window and on no concurrent writers fighting the same rows; overlapping jobs are not addressed here.

---

## Mermaid sequence diagram

```mermaid
sequenceDiagram
  participant CLI as Background Worker
  participant Duck as DuckDB (in-process)
  participant PG as PostgreSQL (Medplum)
  participant Dest as Iceberg

  CLI->>Duck: export --start-window --end-window (+ DB + S3 or ARN)
  Duck->>PG: ATTACH postgres as pg_db
  alt Parquet path (s3-bucket)
    Duck->>PG: SELECT * FROM "AuditEvent" WHERE lastUpdated in window
    Duck->>Dest: COPY ... TO s3://.../audit_events/window_*.parquet
  else S3 Tables path (aws-s3-table-arn)
    Duck->>Dest: ATTACH Iceberg (s3_tables)
    Duck->>Dest: CREATE TABLE IF NOT EXISTS audit_events (schema from PG, LIMIT 0)
    Duck->>Dest: DELETE WHERE lastUpdated in window
    Duck->>PG: SELECT * FROM "AuditEvent" WHERE lastUpdated in window
    Duck->>Dest: INSERT into audit_events
  end
```


packages/data-warehouse/src/aws.ts
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Creates a managed Iceberg table with partition spec, {@link buildWarehouseIcebergWriteOrder},
+   * and S3 Tables compaction strategy **sort** (aligned with the sort order).
+   *
+   * **Existing tables:** Adding or changing `writeOrder` / compaction is not an in-place update in
+   * this helper; evolve or replace the table in AWS if you need new physical layout.
+   *
+   * @param options - Table bucket ARN, namespace, and Iceberg table name (SQL-sanitized identifiers).
+   */
+  async createIcebergTable(options: EnsureIcebergTableOptions): Promise<void> {
```
codyebberson: I think all of this maintenance of creating and deleting tables should be done in CDK, not from application code.

packages/data-warehouse/src/config.ts
```diff
@@ -0,0 +1,198 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Default third arguments for commander `.option(…, …, default)` (from the process environment at module load).
+ */
+export const dataWarehouseCliEnvDefaults = {
+  dbHost: process.env.MEDPLUM_DATABASE_HOST,
```
codyebberson: Strong preference to replace this with existing `server` config code.

packages/data-warehouse/src/delete-tables.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Deletes managed Iceberg tables in S3 Tables for each warehouse source (by derived Iceberg name).
+ * Missing tables are reported in {@link DeleteWarehouseTablesSummary.missing} and do not fail the run.
+ *
+ * @param options - S3 table bucket ARN, region, namespace, and warehouse sources to delete.
+ * @returns Lists of Iceberg table names that were deleted vs not found.
+ */
+export async function deleteWarehouseIcebergTables(
```
codyebberson: Is this essentially a utility for debugging, troubleshooting, and cleanup?  If so, we typically put these types of things in the `medplum-ee` repo.

packages/data-warehouse/src/download.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  const response = JSON.parse(stdout) as GetTableMetadataLocationResponse;
+  if (!response.metadataLocation) {
+    throw new Error('Failed to resolve metadata location from AWS S3 Tables');
+  }
+
+  return response.metadataLocation;
+}
+
+async function downloadFileWithS3Client(
```
codyebberson: I assume this is another debugging/troubleshooting script?

packages/data-warehouse/src/export.ts
```diff
@@ -0,0 +1,379 @@
...
+  readonly icebergTable: string;
+  /** Keys row-threshold overrides and sync log lines (same as `icebergTable`). */
+  readonly tableKey: string;
+}
+
+export interface ExportOptions {
+  databaseUrl: string;
+  s3Bucket?: string;
+  s3Region: string;
+  athenaOutputLocation?: string;
```
codyebberson: Can you please help me understand why we need athena code here?  I assumed S3 tables handled everything?

codyebberson (commented):
Exciting stuff @The-Alchemist  - looking forward to getting this merged.

As we discussed offline, the big thing is that I believe most of the `data-warehouse` code should move to existing packages

- The infra setup code should probably move to `packages/cdk`
- Any code generators should move to `packages/generator`
- Troubleshooting and debugging scripts should move to the `medplum-ee` repo

I know that you're ramping up on a lot of Medplum context very quickly, so please let me know if you'd like to run through that on a call

packages/data-warehouse/src/download.ts
The-Alchemist: it's the ability to download the Parquet files from S3 tables (Iceberg).  Useful for debugging, yes, but also self-hosted environments I believe.  But moving to `medplum-ee` makes sense as well.

packages/data-warehouse/src/export.ts
The-Alchemist: great find, some dead code used for debugging.  removed in latest commit

packages/data-warehouse/src/config.ts
The-Alchemist: removing the `data-warehouse` submodule and moving code to `packages/server`, so it'll use the existing server config code

packages/server/src/data-warehouse/postgres-testcontainer.util.ts
```diff
@@ -0,0 +1,98 @@
...
+
+import { PostgreSqlContainer } from '@testcontainers/postgresql';
+import { execSync } from 'node:child_process';
+import { mkdtempSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+type StartedContainer = Awaited<ReturnType<PostgreSqlContainer['start']>>;
+
+export async function startPostgresTestContainer(image = 'postgres:17'): Promise<{
```
codyebberson: I don't think I understand why we need test containers here?  Can you describe the use case?

packages/server/src/data-warehouse/warehouse-sql.ts
```diff
@@ -0,0 +1,128 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/** Default Iceberg catalog schema when {@link asSqlIdentifier} is applied to namespace. */
+export const DEFAULT_NAMESPACE = 'default';
+
+export function escapeSqlLiteral(value: string): string {
+  return value.replaceAll("'", "''");
```
codyebberson: Is there a way we can avoid manual escaping?  I.e., can DuckDB or other client library handle this?

Alternatively, we could use the existing `SqlBuilder` class in `sql.ts`

packages/server/package.json
```diff
@@ -125,7 +128,8 @@
     "openapi3-ts": "4.5.0",
     "set-cookie-parser": "3.1.0",
     "supertest": "7.2.2",
-    "superwstest": "3.0.0"
+    "superwstest": "3.0.0",
+    "testcontainers": "11.14.0"
```
codyebberson: Let's explore how we can avoid this

codyebberson (commented):
Nice, this is looking much closer than the previous rev, we're getting much closer.

I don't think I understand the `testcontainers` use case, so that's one we should discuss.

The SQL generator is interesting.  I suspect we could use our existing `SqlBuilder` class and utilities, which is probably worth investigating.

From code organization perspective, there are a couple of conventions that I'd prefer we try to uphold

1. Keep cloud-specific code in the cloud-specific directories.  For AWS, that's `packages/server/src/cloud/aws`.  The common pattern that we have used is to have is to have (1) an `interface` and a "local" implementation in the feature directory, and then the cloud-specific implementation of the interface in the cloud-specific directoroy.

2. Minor nit, but i have a preference for using "warehouse" over "data-warehouse" for folder names and file names.  I admit that's subjective and personal preference though.

Let me know if you'd like to discuss those

packages/server/package.json
The-Alchemist: `testcontainers` is just a library to spin up/down Docker containers for tests, in this case, Postgres.  It's low effort to remove.  The tests can use the "global" Postgres database instead.

packages/server/src/data-warehouse/postgres-testcontainer.util.ts
The-Alchemist: testcontainers removed

packages/server/package.json
The-Alchemist: testcontainers removed

packages/server/src/data-warehouse/warehouse-sql.ts
The-Alchemist: new commit uses `SqlBuilder` where feasible without changing `SqlBuilder` too much

packages/server/src/cloud/aws/data-warehouse-client.ts
```diff
@@ -0,0 +1,55 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { CreateNamespaceCommand, GetTableCommand, S3TablesClient } from '@aws-sdk/client-s3tables';
+
+export interface DataWarehouseAwsClientOptions {
+  region: string;
+}
+
+export class DataWarehouseAwsClient {
```
codyebberson: Does this need to be a class?  Or could this just be 2 simple `function`s?

packages/server/src/config/loader.ts
```diff
@@ -157,6 +160,19 @@ export async function loadTestConfig(): Promise<MedplumServerConfig> {
   config.defaultSuperAdminClientId = randomUUID();
   config.defaultSuperAdminClientSecret = randomUUID();
   config.mtlsCertHeader = 'x-mtls-cert';
+  /*
+   * Sample `medplum.config.json` defaults `dataWarehouse.enabled` to false for self-hosters.
+   * The Jest harness still opts in so `initWorkers` → `refreshDataWarehouseSyncScheduler`
+   * exercises the enabled BullMQ scheduler path (cron + ARN come from the file when present).
+   */
+  config.dataWarehouse = {
```
codyebberson: Does this mean that data warehouse sync will run on all tests?  That might be overkill.

packages/server/src/config/loader.ts
```diff
@@ -218,6 +234,10 @@ function loadEnvConfig(): MedplumServerConfig {
       key = key.substring('WORKERS_'.length);
       currConfig = config.workers ??= {};
       section = 'workers';
+    } else if (key.startsWith('DATA_WAREHOUSE_')) {
```
codyebberson, line 225: 👍

packages/server/src/config/types.ts
```diff
@@ -307,6 +314,32 @@ export interface MedplumWorkersConfig {
...
+  enabled?: boolean;
+  /**
+   * BullMQ cron pattern used to schedule sync runs.
+   */
+  cron?: string;
+  /**
+   * Optional statement timeout applied when deriving the sync Postgres URL from server config.
+   * Uses PostgreSQL duration syntax, e.g. "15min" or "900s".
+   */
+  databaseStatementTimeout?: string;
```
codyebberson: Does this need to be a separate setting from the existing database statement timeout?

packages/server/src/config/utils.ts
```diff
@@ -30,6 +30,11 @@ export function addDefaults(config: MedplumServerConfig): ServerConfig {
   config.botLambdaLayerName ||= 'medplum-bot-layer';
   config.bcryptHashSalt ||= 10;
   config.bullmq = { concurrency: 20, removeOnComplete: { count: 1 }, removeOnFail: { count: 1 }, ...config.bullmq };
+  config.dataWarehouse = {
```
codyebberson: Minor: Thoughts on `dataWarehouse === undefined` being equivalent to `dataWarehouse.enabled === false`?

packages/server/src/config/validate-config.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  return {
+    baseUrl: 'http://localhost:8103',
+    database: { host: 'h', dbname: 'd', username: 'u', password: 'p' },
+    redis: { host: 'localhost', port: 6379 },
+    ...overrides,
+  } as MedplumServerConfig;
+}
+
+describe('validateDataWarehouseConfig', () => {
+  it('no-ops when dataWarehouse is absent', () => {
```
codyebberson: Minor Medplum convention:  across all tests across all projects, we use `test()` rather than `it()`

packages/server/src/data-warehouse/destination.ts
```diff
@@ -0,0 +1,73 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
```
codyebberson, line 1: I have mild concerns about having `sink.ts` and `sync.ts` in the same directory.  I accept that those are two valid and distinct concepts.  Just seems ripe for confusion.

Would `sink -> destination` be acceptable?  Or is that too contrived?

packages/server/src/data-warehouse/sync.ts
```diff
@@ -0,0 +1,149 @@
...
+  action: SyncAction;
+}
+
+export interface SyncResult {
+  resources: SyncResourceResult[];
+}
+
+export type SyncAction = 'skip-empty' | 'insert';
+
+export function getSyncAction(count: number): SyncAction {
```
codyebberson: I don't think this needs to be a separate function with separate tests.  It's only called in one place.

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,165 @@
...
+  const syncConfig = config.dataWarehouse;
+  if (!syncConfig?.enabled) {
+    throw new Error('dataWarehouse.enabled must be true to run scheduled sync');
+  }
+
+  assertEnabledDataWarehouseComplete(syncConfig);
+  const { namespace } = syncConfig;
+
+  const [databaseConfig, proxyEndpoint] = config.readonlyDatabase
+    ? [config.readonlyDatabase, config.readonlyDatabaseProxyEndpoint]
```
codyebberson: I'd be comfortable to not support `databaseProxyEndpoint` and `readonlyDatabaseProxyEndpoint` here.  That has been deprecated for a while, and will likely be removed in Medplum v6 later this year.  We do not use them ourselves.

packages/server/src/data-warehouse/warehouse-sql.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+export function buildCountFromHistoryTableQuery(sourceHistoryTable: string, whereClause: string): string {
+  const catalogQualifiedTableName = buildCatalogQualifiedTableIdentifier(POSTGRES_CATALOG, sourceHistoryTable);
+  const contentColumn = `"${catalogQualifiedTableName}"."content"`;
+  const query = new SelectQuery(catalogQualifiedTableName)
+    .raw('COUNT(*) AS count')
+    .where('content', '!=', null)
+    // can't use .where() here because it would be treated as a bind value and generate $1, etc.
+    .whereExpr(new Constant(`${contentColumn} != ''`))
```
codyebberson: This is worth calling out, because there's some history here.  Today, "deleted resources" are represented as empty values.  We know that needs to change to some kind of tombstone values.  Mental note that we will need to come back to this in the future.

packages/server/src/data-warehouse/warehouse-sql.ts
```diff
@@ -0,0 +1,279 @@
...
+    .whereExpr(new Constant(`(${whereClause})`));
```
codyebberson: This seems dangerous, because it could easily be misused for raw SQL.  This method is only called in one place, and it looks like it's just passing in `lastUpdated`.  All dynamic parameters absolutely must go through SQL params.  That's a pretty strict compliance rule.

packages/server/src/fhir/sql.ts
```diff
@@ -522,6 +522,11 @@ export class SqlBuilder {
     return this;
   }
 
+  appendLiteral(str: string): this {
```
codyebberson: This is dangerous.  I see that it is called in three places, and those call sites are safe enough, but I don't feel comfortable with the casual name of this method.  

`appendRawLiteral` or something might be more appropriate to signal the danger?

packages/server/src/workers/data-warehouse-sync.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+jest.mock('../data-warehouse/config', () => {
+  const actual: typeof DataWarehouseConfigModule = jest.requireActual('../data-warehouse/config');
+  return {
+    ...actual,
+    getWarehouseSyncPostgresTableNames: jest.fn(() => ['Patient_history', 'Observation_history']),
+    resolveWarehouseSourcesFromPostgresTableNames: jest.fn((tableNames: string[]) =>
+      tableNames.map((tableName) => ({
+        postgresTable: tableName,
+        icebergTable: tableName.toLowerCase(),
```
codyebberson: 👍

packages/server/src/workers/data-warehouse-sync.test.ts
```diff
@@ -0,0 +1,239 @@
...
+});
+jest.mock('../data-warehouse/sync', () => ({
+  syncData: jest.fn(async () => ({ resources: [{ action: 'insert' }, { action: 'skip-empty' }] })),
+}));
+jest.mock('bullmq');
+
+const mockedSyncData = jest.mocked(syncData);
+const mockedWorker = jest.mocked(Worker);
+
+function buildConfig(overrides?: Partial<MedplumServerConfig>): MedplumServerConfig {
```
codyebberson: Most of this is identical to `initTestConfig()`, so maybe could leverage that?

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,165 @@
...
+        type: 'exponential',
+        delay: 1000,
+      },
+    },
+  });
+
+  let worker: Worker<DataWarehouseSyncJobData> | undefined;
+  if (options?.workerEnabled !== false) {
+    const workerBullmq = getWorkerBullmqConfig(config, 'data-warehouse-sync') ?? {};
+    const perWorkerConcurrency = config.workers?.bullmq?.['data-warehouse-sync']?.concurrency;
```
codyebberson: Is that the same as

```suggestion
    const perWorkerConcurrency = workerBullmq.concurrency;
```

?

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,165 @@
...
+    worker = new Worker<DataWarehouseSyncJobData>(
+      DataWarehouseSyncQueueName,
+      async (job) => processDataWarehouseSyncJob(config, job),
+      {
+        ...defaultOptions,
+        ...workerBullmq,
+        concurrency: perWorkerConcurrency ?? 1,
```
codyebberson: This might be unnecessary, because it's just part of `workerBullmq`?

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,165 @@
...
+      await queue.removeJobScheduler(DataWarehouseSyncSchedulerId);
+    } catch (err) {
+      globalLogger.warn('Failed removing disabled data warehouse sync scheduler', { err });
+    }
+    return;
+  }
+
+  assertEnabledDataWarehouseComplete(syncConfig);
+
+  await queue.upsertJobScheduler(
```
codyebberson, line 130: For my own understanding.... That means that every server startup will call `queue.upsertJobScheduler()`?  I think that's probably ok (?), but perhaps worth calling out.

packages/server/src/database.ts
```diff
@@ -53,14 +53,19 @@ export async function initDatabase(serverConfig: MedplumServerConfig): Promise<v
   }
 }
 
-async function initPool(config: MedplumDatabaseConfig, proxyEndpoint: string | undefined): Promise<Pool> {
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
codyebberson: Please use the type for the return type.  I think `PoolConfig` is an existing type?

codyebberson (commented):
Getting closer 👍 🙏

codyebberson:
See here for fixing DCO: https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot

codyebberson:
@claude please review

packages/server/src/cloud/aws/data-warehouse-client.ts
The-Alchemist: nope, doesn't need to be a class. i'll convert it

packages/server/src/config/loader.ts
The-Alchemist: > Does this mean that data warehouse sync will run on all tests? That might be overkill.

what do you mean by "run on all tests"?  data warehouse tests are (or should be, if they aren't) independent.

i think Claude Code flagged this too.  let me confirm

packages/server/src/config/utils.ts
The-Alchemist: > Minor: Thoughts on dataWarehouse === undefined being equivalent to dataWarehouse.enabled === false?

IMHO, they are semantically identical.  i.e., if `config.dataWarehouse` is missing, it's the same as `config.dataWarehouse.enabled` is false.  confirm that's your expected behavior?

packages/server/src/workers/data-warehouse-sync.ts
The-Alchemist: yeah, i believe so.

although, now that i think about it, shouldn't concurrency always be 1?  i.e., we shouldn't allow two simultaneous data warehouse sync's.

does that sound right?

packages/server/src/workers/data-warehouse-sync.ts
The-Alchemist: yeah, i wasn't sure about this one either.  AI says:

> Yes—if a job is truly cron/repeat-based, upsertJobScheduler is generally the right pattern in BullMQ.
> 
> Why:
> 
> It is idempotent: you can call it repeatedly and it updates/replaces the scheduler by ID instead of creating duplicates.
> It handles config changes cleanly (new cron expression, changed payload/options).
> It matches your current design in both data-warehouse-sync and cron, where you also use removeJobScheduler when disabling.
> When not to use it:
> 
> One-off async work triggered by events/requests (queue.add(...)) should stay as-is.
> Fire-and-forget jobs that are not recurring do not need scheduler APIs.
> Practical rule:
> 
> Recurring by identity (e.g., “the DW sync schedule”, “bot X’s schedule”) → upsertJobScheduler(schedulerId, ...).
> Per-event work item (e.g., “resource updated”, “run this migration once”) → queue.add(...).
> So your instinct is correct: for other cron-like workers, prefer upsert + stable scheduler IDs.

Logic is sound to me, but WDYT?

packages/server/src/workers/data-warehouse-sync.ts
The-Alchemist: good catch

packages/server/src/workers/data-warehouse-sync.ts
codyebberson: Probably fine?  Not sure how else we would do it, other than manual action?

packages/server/src/workers/data-warehouse-sync.test.ts
The-Alchemist: done

packages/server/src/config/types.ts
The-Alchemist: good catch; removed duplicate config

packages/server/src/config/loader.ts
The-Alchemist: fixed

The-Alchemist:
@claude : review

packages/server/src/data-warehouse/sync.ts
The-Alchemist: done

The-Alchemist:
@claude review

packages/server/src/data-warehouse/warehouse-sql.ts
```diff
@@ -0,0 +1,268 @@
...
+import { Column, Condition, Constant, Disjunction, InsertQuery, SelectQuery, SqlBuilder, Subquery } from '../fhir/sql';
+
+const DEFAULT_COMPRESSION_TYPE = 'zstd';
+const DEFAULT_FILE_FORMAT = 'PARQUET';
+
+/** Default Iceberg catalog schema for warehouse objects. */
+export const DEFAULT_NAMESPACE = 'default';
+
+/** DuckDB catalog alias for attached S3 Tables Iceberg; not a Postgres schema/catalog name. */
+export const S3_TABLES_CATALOG = 's3_tables_db';
```
codyebberson: Minor naming nit: Is this S3-specific?  

I might be missing something, but it appears that this is used in the generic portion, and not necessarily tied to anything about S3.  Would `ICEBERG_CATALOG` be more accurate?

packages/server/medplum.config.json
```diff
@@ -34,5 +34,8 @@
     "removeOnFail": { "count": 1 },
     "removeOnComplete": { "count": 1 }
   },
-  "shutdownTimeoutMilliseconds": 30000
+  "shutdownTimeoutMilliseconds": 30000,
+  "dataWarehouse": {
```
codyebberson: Is this effectively the default value?  Can we remove?

codyebberson (commented):
Nice, good job @The-Alchemist  - this is looking good.

I think the next steps here are to deploy to staging and beat the crap out of it.  I want to make sure that `server` is robust against all kinds of borked configurations:

1. Disabled warehouse should be 100% no-op
2. Enabled warehouse with missing S3 Tables should emit errors/warnings, but should not otherwise impact server performance
3. Same for partial S3 Tables, such as S3 Tables with missing column definitions
4. Missing S3 Table permissions

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  const defaultOptions: QueueBaseOptions = {
+    connection: getBullmqRedisConnectionOptions(config),
+  };
+
+  const queue = new Queue<DataWarehouseSyncJobData>(DataWarehouseSyncQueueName, {
+    ...defaultOptions,
+    defaultJobOptions: {
+      attempts: 3,
```
codyebberson: Do we actually want to have a "retry" here?  Or should it just fail, and the next tick will automatically pick it up?

packages/server/src/fhir/sql.ts
```diff
@@ -35,38 +35,48 @@ export const ColumnType = {
 } as const;
 export type ColumnType = (typeof ColumnType)[keyof typeof ColumnType];
 
-export type OperatorFunc = (sql: SqlBuilder, column: Column, parameter: any, paramType?: string) => void;
+export type ConditionOperand = Column | Subquery;
+
+export type OperatorFunc = (sql: SqlBuilder, operand: ConditionOperand, parameter: any, paramType?: string) => void;
+
+function appendConditionOperand(sql: SqlBuilder, operand: ConditionOperand): void {
```
codyebberson: Is there a proper term for the common parent type of `Column` and `Subquery`? 

I think it might just be `Expression`?

Could we just use `sql.appendExpression()` in all of these cases rather than introducing new `appendConditionOperand()`?

packages/server/src/data-warehouse/destination.ts
The-Alchemist: renamed to destination

packages/server/medplum.config.json
The-Alchemist: good catch, i restored `medplum.config.json` to the original, because DW is disabled by default.

packages/server/src/data-warehouse/warehouse-sql.ts
The-Alchemist: done.

packages/server/src/workers/data-warehouse-sync.ts
The-Alchemist: done

packages/server/src/fhir/sql.ts
The-Alchemist: figured out a way to avoid introducing a new type (smaller blast radius).

see https://github.com/medplum/medplum/pull/8941/changes#diff-2a795e2ef605fd3a9e697b771267dd2d9e4e88735e2d84440d0c980aedb51edc

The-Alchemist:
@codyebberson 
> Nice, good job @The-Alchemist - this is looking good.

👍 

> I think the next steps here are to deploy to staging and beat the crap out of it. I want to make sure that `server` is robust against all kinds of borked configurations:

I had a config validation functions and tests (see https://github.com/medplum/medplum/pull/8941/changes#diff-bd4b3ab6f4ad9520c05211fe88ce3404528f1ba7b45fcc8d510266896d96c832), but this is only what I can think of, I'm sure there other cases, and I can add them to the tests.

> 
> 1. Disabled warehouse should be 100% no-op

https://github.com/medplum/medplum/pull/8941/changes#diff-bd4b3ab6f4ad9520c05211fe88ce3404528f1ba7b45fcc8d510266896d96c832R18

> 2. Enabled warehouse with missing S3 Tables should emit errors/warnings, but should not otherwise impact server performance
> 3. Same for partial S3 Tables, such as S3 Tables with missing column definitions
> 4. Missing S3 Table permissions

I need some direction on what are fatal and what are non-fatal errors.  I'm not very familiar with the background worker API and how errors are handled.

But I do have configuration validation centralized at `validate-config.ts`.  https://github.com/medplum/medplum/pull/8941/changes#diff-7f0d13fc1253f436f1d7b7921f67778a6c41ffb6d78fe435dc31d94fcbd1cd65

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,181 @@
...
+  const workerBullmq = getWorkerBullmqConfig(config, 'data-warehouse-sync') ?? {};
+  const worker = new Worker<DataWarehouseSyncJobData>(
+    DataWarehouseSyncQueueName,
+    async (job) => processDataWarehouseSyncJob(config, job),
+    {
+      ...defaultOptions,
+      ...workerBullmq,
+      lockDuration: workerBullmq.lockDuration ?? DATA_WAREHOUSE_SYNC_LOCK_DURATION_MS,
+      // Data warehouse sync is intentionally serialized.
+      concurrency: 1,
```
mattlong, line 87: This only results in one job being processed at a time per worker, not globally. BullMQ does have the notion of [global concurrency](https://docs.bullmq.io/guide/queues/global-concurrency), but we don't use it anywhere else. We were considering it at one point for the post-deploy migration queue where it is also important that only one job be running at a time, but wound up deciding against it to not be as dependent on BullMQ-specific behavior should we ever want to switch background job frameworks.

Another options would be a Postgres advisory lock; see `acquireAdvisoryLock` in `src/database.ts`.

packages/server/src/workers/data-warehouse-sync.ts
```diff
@@ -0,0 +1,181 @@
...
+export const DataWarehouseSyncSchedulerId = 'data-warehouse-sync';
+
+/**
+ * Default BullMQ lock duration for long-running warehouse sync jobs.
+ *
+ * This is useful because we want a high-frequency sync, but if it takes a long time,
+ * BullMQ will assume the job is dead.  We alleviate this by job.updateProgress and this
+ * increased default lock duration.
+ */
+export const DATA_WAREHOUSE_SYNC_LOCK_DURATION_MS = 5 * 60 * 1000;
```
mattlong, line 39: This is a pretty big change from the default 30 seconds. Is this worker CPU intensive enough that it wouldn't be able to renew its lock for five minutes at a time? [More info on job locking](https://docs.bullmq.io/bull/important-notes)

packages/server/src/data-warehouse/sync.ts
```diff
@@ -0,0 +1,138 @@
...
+    const resultTableName = options.destination.getDestinationName(spec);
+    await options.destination.ensureTargetExists(spec, namespace);
+
+    const countReader = await runParameterizedWarehouseSqlReadAll(
+      connection,
+      buildCountFromHistoryTableQuery(postgresTable, sourcePredicate)
+    );
+    const count = Number((countReader.getRowObjectsJson() as { count: number }[])[0]?.count ?? 0);
+
+    if (count > 0) {
```
mattlong, line 97: Instead of checking the count first and then writing the rows, couldn't it optimistically try writing the rows and no-op if there are no rows to write?

packages/server/src/database.ts
```diff
@@ -53,19 +53,23 @@ export async function initDatabase(serverConfig: MedplumServerConfig): Promise<v
   }
 }
 
-async function initPool(config: MedplumDatabaseConfig, proxyEndpoint: string | undefined): Promise<Pool> {
+export function initPoolConfig(
```
mattlong: Does this need to be exported?

packages/server/src/data-warehouse/sync.int.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  beforeAll(async () => {
+    const config = await loadTestConfig();
+    const db = config.database;
+    host = db.host ?? '';
+    port = db.port ?? 5432;
+    database = db.dbname ?? '';
+    username = db.username ?? '';
+    password = db.password ?? '';
+
+    const client = new pg.Client({
```
mattlong: This should use `loadTestConfig()` and `getDatabasePool(DatabaseMode.WRITER)` instead of directly creating a pg.Client. See `src/fhir/operations/db-column-statistics.test.ts` for a similar use case

packages/server/src/workers/data-warehouse-sync.ts
The-Alchemist: sounds like the pg-specific `acquireAdvisoryLock` is the way to go.  will implement

packages/server/src/workers/data-warehouse-sync.ts
The-Alchemist: it's not CPU intensive at all, it's all DB IO.  `SELECT`ing a large table can take more than 30 seconds, and this seemed a quick way to resolve it.  Do you see a different way?

packages/server/src/data-warehouse/sync.ts
The-Alchemist: i actually removed this block entirely, as `COUNT` is unnecessary and slow and only used for logging anyway.  Now it writes and logs the count afterwards.

BTW, Trying it optimistically isn't an option because there is no optimistic write a la `INSERT INTO... ON CONFLICT`/`MERGE` for Iceberg tables.

packages/server/src/data-warehouse/sync.int.test.ts
The-Alchemist: done

packages/server/src/database.ts
The-Alchemist: good catch, it was an experiment.  fixed

codyebberson (approved):

packages/server/src/workers/data-warehouse-sync.ts
mattlong: @The-Alchemist  I could very well be misunderstanding how job locking works, but my understanding is that the job could still phone home to renew its lock on the job while it's waiting for DB IO.

PR #8942, opened 2026-04-11
React hooks for DoseSpot prescriber self-enrollment bot

oleg-mp:
Enables prescriber self-enrollment. Works together with the bot `dosespot-self-enroll-prescriber-bot`


oleg-mp:
@claude review

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
```diff
@@ -0,0 +1,82 @@
...
+    try {
+      const roles = await medplum.searchResources('PractitionerRole', {
+        practitioner: `Practitioner/${practitionerId}`,
+        active: 'true',
+        _count: '10',
+      });
+
+      setAuthorizedNotEnrolled(hasDoseSpotPractitionerRole(roles));
+    } catch (err) {
+      console.warn('Failed to check DoseSpot PractitionerRole:', err);
```
techdavidy: show error here

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -1,20 +1,106 @@
...
+        }
+      } catch (err: unknown) {
+        if (!cancelled) {
+          onErrorRef.current?.(err);
+        }
+      }
+    };
+
+    run().catch(() => {
+      // Handled via onErrorRef when !cancelled
```
techdavidy: should be show message or handle error ?

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
codyebberson: Example:

```ts
import { notifications, showNotification } from '@mantine/notifications';

// ...

        showNotification({
          title: 'Missing Diagnosis',
          message: 'Please add at least one diagnosis before submitting a claim',
          color: 'red',
        });
```

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const practitionerId = profile?.resourceType === 'Practitioner' ? profile.id : undefined;
+
+  const checkPractitionerRole = useCallback(async () => {
+    if (enrolled || !practitionerId || checkedRef.current) {
+      setLoading(false);
+      return;
+    }
+
+    checkedRef.current = true;
```
noahsilas: Is the idea of this that to limit such that the effect will run at-most once? Is there some scenario where something like the practitionerId changes that we need to protect against?

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
```diff
@@ -0,0 +1,82 @@
...
+    try {
+      const roles = await medplum.searchResources('PractitionerRole', {
+        practitioner: `Practitioner/${practitionerId}`,
+        active: 'true',
+        _count: '10',
+      });
```
noahsilas: Can the `useSearchResources()` hook work for this? It can handle things like the loading state for you, and then things like `authorized` become a little clearer as derived state:

```ts
  const [roles, rolesLoading, rolesOutcome] = useSearchResources(
    'PractitionerRole',
    { practitioner: `Practitioner/${practitionerId}`, ... },
    { enabled: enrolled && !!practitionerId }
  );
  const authorized = useMemo(() => roles && hasDoseSpotPractitionerRole(roles), [roles]);
  const authorizedNotEnrolled = authorized && !enrolled;
```

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
```diff
@@ -0,0 +1,82 @@
...
+  useEffect(() => {
+    checkPractitionerRole().catch(console.error);
+  }, [checkPractitionerRole]);
+
+  const hasAccess = enrolled || authorizedNotEnrolled;
+
+  return {
+    enrolled,
+    authorizedNotEnrolled,
+    hasAccess,
```
noahsilas, line 58: It feels a little strange to be exposing `enrolled`, `hasAccess`, and `authorizedNotEnrolled`, which seems like it is conceptually derived from the other two (`authorizedNotEnrolled = hasAccess && !enrolled` is how it would feel most natural to me, but here you are actually deriving `hasAccess` from the others which is kind of interesting!)

Can we remove `authorizedNotEnrolled` to simplify this type?

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
```diff
@@ -0,0 +1,82 @@
...
+    needsSelfEnroll: authorizedNotEnrolled && !enrolled,
```
noahsilas: Similar thought here, although I think that `authorizedNotEnrolled` actually can only be set when `enrolled` is false. Can we drop this attribute from the hook's return value?

I think this one actually has a stronger case to keep it than `authorizedNotEnabled` though, as this name is more proscriptive in how the value should be interpreted.

examples/medplum-provider/src/pages/patient/PatientPage.utils.ts
```diff
@@ -37,11 +37,19 @@ export function getPatientPageTabOrThrow(tabId: string): PatientPageTabInfo {
...
+ *   When provided, controls DoseSpot tab visibility directly (supports
+ *   self-enrollment via PractitionerRole in addition to existing identifiers).
+ *   When omitted, falls back to checking the membership for a DoseSpot identifier.
  * @returns Filtered array of patient page tabs.
  */
-export function getPatientPageTabs(membership: ProjectMembership | undefined): PatientPageTabInfo[] {
-  const hasDoseSpot = hasDoseSpotIdentifier(membership);
+export function getPatientPageTabs(
+  membership: ProjectMembership | undefined,
+  hasDoseSpotAccess?: boolean
```
noahsilas: How would you feel about making this an options argument instead of a boolean? That leaves it more open to adding other flags or options in the future.

```ts
export function getPatientPageTabs(
  membership: ProjectMembership | undefined,
  options?: { hasDoseSpotAccess?: boolean }
): PatientPageTabInfo[] {
  const hasDoseSpot = options?.hasDoseSpotAccess ?? hasDoseSpotIdentifier(membership);
```

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -1,20 +1,106 @@
...
+  const onIframeSuccessRef = useRef(onIframeSuccess);
+  const onSelfEnrollSuccessRef = useRef(onSelfEnrollSuccess);
+  const onErrorRef = useRef(onError);
+
+  useEffect(() => {
+    onPatientSyncSuccessRef.current = onPatientSyncSuccess;
+    onIframeSuccessRef.current = onIframeSuccess;
+    onSelfEnrollSuccessRef.current = onSelfEnrollSuccess;
+    onErrorRef.current = onError;
+  }, [onPatientSyncSuccess, onIframeSuccess, onSelfEnrollSuccess, onError]);
```
noahsilas, line 47: This feels a bit suspicious. Is the intention to allow changes to the options argument without retriggering the effect below? If that's the case, this might be a good use case for [`useEffectEvent()`](https://react.dev/reference/react/useEffectEvent).

Overall this seems like it might not be necessary at all, but it also might be good to keep as a protection to prevent someone from inadvertently passing an unstable function reference and re-calling the bot on every render...

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -1,20 +1,106 @@
...
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const run = async (): Promise<void> => {
+      try {
+        if (selfEnroll && !hasDoseSpotIdentifier(medplum)) {
+          const enrollResult = (await medplum.executeBot(
```
noahsilas: Would it make sense to add some kind of generic callback that the UI can use to reflect what's going on? I'm imagining something sort of like this:
```ts
        if (selfEnroll && !hasDoseSpotIdentifier(medplum)) {
          statusCallback.current.?("Enrolling Practitioner with DoseSpot");
          const enrollResult = await medplum.executeBot(...)
          ...
        }

        if (patientId) {
          statusCallback.current.?("Syncing Patient information to DoseSpot"
          await medplum.executeBot(DOSESPOT_PATIENT_SYNC_BOT, { patientId });
          ...
        }
```

Not a blocker, this could be an addon later if we think it's helpful. But this is a long series of interactions that seem likely to leave folks wanting faster feedback.

packages/dosespot-react/src/useDoseSpotIFrame.ts
noahsilas: Should the run function get unwrapped from the try/catch and move the error handling inside this .catch handler?

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -1,20 +1,106 @@
...
+    onSelfEnrollSuccessRef.current = onSelfEnrollSuccess;
+    onErrorRef.current = onError;
+  }, [onPatientSyncSuccess, onIframeSuccess, onSelfEnrollSuccess, onError]);
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const run = async (): Promise<void> => {
+      try {
+        if (selfEnroll && !hasDoseSpotIdentifier(medplum)) {
```
noahsilas: When I first read this line I was a big surprised. It doesn't make a ton of sense to me to ask if the MedplumClient instance has DoseSpot, but I see that under the covers that's checking `medplum.getProjectMembership()`.

I'd gently advocate for taking a ProjectMembership as input instead of the medplum client. Not a blocker, but I would have understood something like this much more intuitively:
```ts
        if (selfEnroll && !hasDoseSpotIdentifier(medplum.getProjectMembership())) {
```

packages/dosespot-react/src/useDoseSpotSelfEnrollment.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const { result, loading, error } = useDoseSpotSelfEnrollment(options);
+  return (
+    <div>
+      <div>loading: {String(loading)}</div>
+      <div>status: {result?.status ?? 'none'}</div>
+      <div>error: {error ? formatEnrollmentTestError(error) : 'none'}</div>
+      <div>nextSteps: {result?.nextSteps?.join('; ') ?? 'none'}</div>
+    </div>
+  );
+}
```
noahsilas: You may be able to remove a bit of this boilerplate by using the [`renderHook`](https://testing-library.com/docs/react-testing-library/api/#renderhook) helper.

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
oleg-mp: This is a guard for the double invocation when running locally in strict mode. A way to bail early if/when another call is in progress.

examples/medplum-provider/src/hooks/useDoseSpotAccess.ts
oleg-mp: Done — rewrote the hook to use `useSearchResources` with `{ enabled }`, which handles loading state, deduplication, and the strict-mode guard internally. Removed `authorizedNotEnrolled` from the interface, keeping only `enrolled`, `hasAccess`, `needsSelfEnroll`, and `loading`. Also added `showNotification` on error per the earlier feedback.

packages/dosespot-react/src/useDoseSpotIFrame.ts
```diff
@@ -1,20 +1,103 @@
...
+      }
+
+      if (patientId) {
+        await medplum.executeBot(DOSESPOT_PATIENT_SYNC_BOT, { patientId });
+        if (cancelled) {
+          return;
+        }
+        onPatientSyncSuccessRef.current?.();
+      }
+      const result = await medplum.executeBot(DOSESPOT_IFRAME_BOT, { patientId });
```
noahsilas, line 71: Would it make sense to run these bots in parallel instead of in series? They don't seem to depend on each other's results. That might reduce how long callers are waiting on results.

This looks like it does match the current `useEPrescribingIFrame` behavior, so maybe best to leave as-is for now and try to add optimizations like this in a separate PR?

packages/dosespot-react/src/useDoseSpotSelfEnrollment.ts
```diff
@@ -0,0 +1,90 @@
...
+    } catch (err: unknown) {
+      setError(err);
+      onErrorRef.current?.(err);
+    } finally {
+      setLoading(false);
+    }
+  }, [medplum, enabled]);
+
+  useEffect(() => {
+    enroll().catch(console.error);
```
noahsilas, line 86: The error handling in here feels pretty intense; an error is logged to the console, passed into the `onError` callback prop, and returned as an element of the hook return value. I wonder if it would make sense to reduce our API surface by omitting some of those?

noahsilas (approved):

PR #8949, opened 2026-04-13
feat(react): Include ResourcesInput component to search/select multiple resources 

techdavidy:
solves https://github.com/medplum/medplum/issues/7246

Screenshot 2026-04-13 at 10 17 04 AM


techdavidy:
@claude please review

packages/react/src/ResourceInput/ResourcesInput.tsx
```diff
@@ -0,0 +1,246 @@
...
+    let cancelled = false;
+    const items = initialDefaultValue.current;
+
+    if (items && items.length > 0) {
+      Promise.allSettled(
+        items.map((item): Promise<T> => {
+          if (isResource(item)) {
+            return Promise.resolve(item as T);
+          }
+          if (isReference(item) && isPopulated(item)) {
```
noahsilas: I think that if `isReference(item)` returns true, then `isPopulated(item)` should also always be true, so I think we can drop one part of that condition:

```suggestion
          if (isReference(item)) {
```

packages/react/src/ResourceInput/ResourcesInput.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  useEffect(() => {
+    let cancelled = false;
+    const items = initialDefaultValue.current;
+
+    if (items && items.length > 0) {
+      Promise.allSettled(
+        items.map((item): Promise<T> => {
+          if (isResource(item)) {
+            return Promise.resolve(item as T);
```
noahsilas: This type cast looks like it can be dropped; the type here gets refined b `isResource` into `T & Resource`, which is assignable to `T`.
```suggestion
            return Promise.resolve(item);
```

packages/react/src/ResourceInput/ResourcesInput.tsx
```diff
@@ -0,0 +1,246 @@
...
+          }
+          if (isReference(item) && isPopulated(item)) {
+            return medplum.readReference(item as Reference<T>) as Promise<T>;
```
noahsilas: You can drop one of these type casts:
```suggestion
            return medplum.readReference(item as Reference<T>);
```

I'm a little surprised that the inner cast is needed, but it looks like the type is being refined to `Reference<T> & Reference<Resource>` rather than simply `Reference<T>`, and `readReference` seems to then decide that its return type can be any resource. Arg.

You could choose to annotate the type to readReference instead of doing a cast, but I think that's functionally pretty similar.
```suggestion
            return medplum.readReference<T>(item);
```

packages/react/src/ResourceInput/ResourcesInput.tsx
```diff
@@ -0,0 +1,246 @@
...
+          }
+          return Promise.reject(new Error('Not a resolvable item'));
+        })
+      )
+        .then((settled) => {
+          if (!cancelled) {
+            setDefaultResources(
+              (settled as PromiseSettledResult<T>[])
+                .filter((r): r is PromiseFulfilledResult<T> => r.status === 'fulfilled')
```
noahsilas: I don't think these casts are doing anything meaningful - it's changing the inner type of `settled` from `Awaited<T>` to `T`, but T shouldn't be a thenable. The filter then can also be cleaned up, as TS can infer the correct type automatically:

```suggestion
              settled.filter((r) => r.status === 'fulfilled')
```

packages/react/src/ResourceInput/ResourceInput.tsx
```diff
@@ -92,37 +23,14 @@ export interface ResourceInputProps<T extends Resource = Resource> {
...
-function toOption<T extends Resource>(resource: T): AsyncAutocompleteOption<T> {
-  return {
-    value: getReferenceString(resource) ?? '',
-    label: getDisplayString(resource),
-    resource,
-  };
-}
-
+/**
+ * @deprecated Use ResourcesInput instead, which supports multiple default and selected values.
```
noahsilas: I'd like to advocate to remove this deprecation warning - I think there is a common case where you want to offer a single resource selection, and having a component that does the wrapping/unwrapping of array inputs for you seems quite handy!

packages/react/src/ResourceInput/MultiResourceInput.tsx
```diff
@@ -0,0 +1,246 @@
...
+  // Capture initial defaultValue at mount — treated as uncontrolled (initial-only).
+  // The Mantine MultiSelect component does not reliably handle changes to defaultValue after mount.
+  const initialDefaultValue = useRef(props.defaultValue);
+
+  // undefined = still resolving references, [] = ready with no defaults, T[] = ready with resolved resources
+  const [defaultResources, setDefaultResources] = useState<T[] | undefined>(
+    !props.defaultValue || props.defaultValue.length === 0 ? [] : undefined
+  );
+
+  useEffect(() => {
```
noahsilas, line 168: Does it make sense to extract a reusable hook like `useReadReferences<T>(references: Reference<T>[])`? This seems like something we will want to do commonly, but maybe it's better to hold off on such complication until we see more usage cases.

noahsilas (approved):
Looks good to me! I think I would recommend removing the deprecation on `ResourceInput` before merging, I think this is probably a highly used component we should continue supporting.

The diff is a little hard to read due to the copying into a new filename, so I focused on the most important diff via `git diff origin/main:packages/react/src/ResourceInput/ResourceInput.tsx origin/david/resourcesinput-component:packages/react/src/ResourceInput/ResourcesInput.tsx`.

That looks good to me, with some opportunities to do a little type cleanup. (One side question: did Claude generate the first pass here? I have noticed that when I'm using Claude it seems to get fairly aggressive with type casting that I end up cleaning up later).

packages/react/src/ResourceInput/MultiResourceInput.tsx
techdavidy: I think we can hold for now

ThatOneBro:
I kind of prefer the name `MultiResourceInput` since the current name only differs by one letter. But that's just my two cents, otherwise LGTM

techdavidy:
@ThatOneBro updated to MultiResourceInput

ThatOneBro (approved):

PR #8964, opened 2026-04-14
fix(otel): use `ExponentialHistogram` view

ThatOneBro:
Previously our histograms have been using the default explicit buckets for histograms, which are arbitrarily set (0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000), leading to truncated graphs which don't make as much sense whenever data points suddenly spike but stay within the arbitrarily set lower value buckets. This PR fixes that by using the automatically calculated buckets via an exponential histogram view, which is now the default for most other OTel SDKs in other languages. 

The exponential histogram uses up to 160 buckets by default, selected to support a high-resolution distribution covering 1ms to 100s with less than 5% relative error — making it strictly better than simply tweaking the explicit bucket boundaries.


mattwiller (approved):
Looks good to me, though we should deploy to Staging and verify both histogram and other (e.g. counter) metrics appear correctly

PR #8966, opened 2026-04-14
fix(publish): add new `scriptsure-react` pkg

ThatOneBro:
Adds new `scriptsure-react` package to the `publish.sh` script


codyebberson:
When publishing a new package, we always need to create the package manually first and enable trusted publishing.

I published empty version 0.0.1 here: https://www.npmjs.com/package/@medplum/scriptsure-react

And then enabled trusted publishing:

image


image

codyebberson (approved):

PR #8968, opened 2026-04-14
Collapse duplicate problems in ProblemList

codyebberson:


codyebberson:
@copilot  - please fix the feedback

codyebberson:
@claude please review

techdavidy (approved):

packages/react/src/PatientSummary/ProblemList.tsx
```diff
@@ -55,34 +94,64 @@ export function ProblemList(props: ProblemListProps): JSX.Element {
...
-                        status={problem.clinicalStatus?.coding?.[0]?.code}
-                      />
-                    )}
-                    <Text size="xs" fw={500} c="dimmed">
-                      {formatDate(problem.onsetDateTime)}
-                    </Text>
-                  </Group>
+            {groupedProblems.map(([key, group], groupIndex) => {
+              const isExpanded = expandedGroups.has(key);
+              const displayProblems = isExpanded ? group : [group[0]];
```
noahsilas, line 99: I wonder if it would make sense to do a deeper inspection on the problems to choose how to render the group (particularly if there are instances with different status badges).

That said, I think that these results are sorted by `-_lastUpdated`, so this should show the status of the latest touched instance of a problem in this group, which seems like a great incremental step.

packages/react/src/PatientSummary/ProblemList.tsx
```diff
@@ -28,6 +39,34 @@ export function ProblemList(props: ProblemListProps): JSX.Element {
   );
   const [editCondition, setEditCondition] = useState<Condition>();
   const [opened, { open, close }] = useDisclosure(false);
+  const [expandedGroups, setExpandedGroups] = useState<Set<string>>(() => new Set());
+
+  const groupedProblems = useMemo(() => {
+    const groups = new Map<string, Condition[]>();
```
noahsilas, line 45: I think that our current TS target would allow you to use `Object.groupBy()` if you want to lean on some newer browser builtins 😁

noahsilas (approved):

PR #8978, opened 2026-04-15
feat(server): support externalAuthProviders via environment variable

ianplunkett:
## Summary

- Adds `externalAuthProviders` to the `objectKeys` set in the env config loader, enabling it to be set via the `MEDPLUM_EXTERNAL_AUTH_PROVIDERS` environment variable
- Previously the field was silently stored as a raw string instead of a parsed JSON array, making env-based configuration non-functional
- Adds a test covering the env var loading path

Usage would look like 

`MEDPLUM_EXTERNAL_AUTH_PROVIDERS='[{"issuer":"<ISSUER>","userInfoUrl":"<USER_INFO_URL>"}]'`

### DIFF
--- a/packages/server/src/config/loader.test.ts
+++ b/packages/server/src/config/loader.test.ts
@@ -107,6 +107,19 @@ describe('Config', () => {
     expect(config.botCustomFunctionsEnabled).toBe(true);
   });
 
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

codyebberson (approved):

PR #8979, opened 2026-04-16
Small document touch up

reshmakh:


packages/docs/docs/access/access-policies.md
```diff
@@ -560,6 +560,36 @@ The [patient access policy](#patient-access) above can be combined with [policy
...
+2. **Read access to each subscribed resource type** — Users must also have `read` access for every resource type they intend to subscribe to. For example, if your app calls `useSubscription('Communication?...')`, the policy must grant at least `read` on `Communication`.
+
+```json
+{
+  "resourceType": "AccessPolicy",
+  "name": "WebSocket Subscription Policy",
+  "resource": [
+    {
+      "resourceType": "Subscription",
+      "criteria": "Subscription?type=websocket&author=%user"
```
ThatOneBro: I made a mistake here, this should be
```suggestion
      "criteria": "Subscription?type=websocket&author=%profile"
```

packages/docs/docs/access/access-policies.md
```diff
@@ -560,6 +560,36 @@ The [patient access policy](#patient-access) above can be combined with [policy
...
 ```
 
+### WebSocket Subscriptions (`useSubscription` Hook)
+
+To use the [`useSubscription`](/docs/react/use-subscription) hook, users need two things in their access policy:
+
+1. **Permission to create `Subscription` resources** — The hook creates an in-memory `Subscription` on behalf of the user. Use a criteria scoped to websocket subscriptions owned by the current user so that users can only manage their own subscriptions:
+
+   ```
+   Subscription?type=websocket&author=%user
```
ThatOneBro:
```suggestion
   Subscription?type=websocket&author=%profile
```

finnbergquist (approved):
Agree with Derrick's comments, and then good to go.  We could also consider adding this to default patient access policy?

ThatOneBro (approved):

PR #8982, opened 2026-04-17
Failed login cleanup

codyebberson:
Two changes:

1. Fixed loading button state on `<SignInForm>`
2. Added explicit log message for failed login via `/auth/login`

### DIFF
--- a/packages/react/src/auth/AuthenticationForm.tsx
+++ b/packages/react/src/auth/AuthenticationForm.tsx
@@ -164,7 +164,7 @@ export function PasswordForm(props: PasswordFormProps): JSX.Element {
   const issues = getIssuesForExpression(outcome, undefined);
 
   const handleSubmit = useCallback(
-    (formData: Record<string, string>) => {
+    (formData: Record<string, string>) =>
       medplum
         .startLogin({
           ...baseLoginRequest,
@@ -172,8 +172,7 @@ export function PasswordForm(props: PasswordFormProps): JSX.Element {
           remember: formData.remember === 'on',
         })
         .then(handleAuthResponse)
-        .catch((err: unknown) => setOutcome(normalizeOperationOutcome(err)));
-    },
+        .catch((err: unknown) => setOutcome(normalizeOperationOutcome(err))),
     [medplum, baseLoginRequest, handleAuthResponse]
   );
 
--- a/packages/server/src/auth/login.ts
+++ b/packages/server/src/auth/login.ts
@@ -1,11 +1,14 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
+import { sleep } from '@medplum/core';
 import type { ResourceType } from '@medplum/fhirtypes';
 import type { Request, Response } from 'express';
 import { body } from 'express-validator';
 import { randomUUID } from 'node:crypto';
+import { getConfig } from '../config/loader';
 import { getLogger } from '../logger';
 import { tryLogin } from '../oauth/utils';
+import { peekRateLimitState } from '../ratelimit';
 import { makeValidationMiddleware } from '../util/validator';
 import { getProjectIdByClientId, sendLoginResult } from './utils';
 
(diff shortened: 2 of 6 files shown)


packages/react/src/auth/AuthenticationForm.tsx
```diff
@@ -164,16 +164,15 @@ export function PasswordForm(props: PasswordFormProps): JSX.Element {
   const issues = getIssuesForExpression(outcome, undefined);
 
   const handleSubmit = useCallback(
-    (formData: Record<string, string>) => {
+    (formData: Record<string, string>) =>
```
codyebberson, line 167: This is for our `<Form>` and `<SubmitButton>` react components.

If the `onSubmit` prop to `<Form>` returns a `Promise`, then it will automatically use the `<Button>` "loading" state in `SubmitButton`.

See Mantine docs: https://mantine.dev/core/button/#loading-state

Before: this callback returned `void`, and therefore the `<SubmitButton>` never used loading state.

After: this callback returns a `Promise`, and `<SubmitButton>` uses loading state.

This becomes very obvious when you're getting throttled for multiple seconds.

packages/server/src/auth/login.ts
```diff
@@ -27,26 +30,39 @@ export async function loginHandler(req: Request, res: Response): Promise<void> {
...
+      remoteAddress: req.ip,
+      userAgent: req.get('User-Agent'),
+      allowNoMembership: req.body.projectId === 'new',
+      origin: req.get('Origin'),
+    });
+    getLogger().info('Login success', { email: req.body.email, projectId });
+    await sendLoginResult(res, login);
+  } catch (err) {
+    const config = getConfig();
+    const rateLimitState = await peekRateLimitState(req, config);
```
codyebberson: This is perhaps an abuse of the rate limiter, but I think it's the most straightforward way to get this out.

packages/server/src/auth/login.ts
```diff
@@ -27,26 +30,39 @@ export async function loginHandler(req: Request, res: Response): Promise<void> {
...
+    const attempts = rateLimitState?.consumedPoints ?? 0;
+    const startAttempt = config.failedLoginThrottleStartAttempt;
+    const baseDelayMs = config.failedLoginThrottleBaseDelayMs;
+    const maxDelayMs = config.failedLoginThrottleMaxDelayMs;
+    const exponent = Math.max(attempts - startAttempt, 0);
+    const rawDelayMs = baseDelayMs * 2 ** exponent;
+    const delayMs = Math.min(maxDelayMs, rawDelayMs);
+    getLogger().warn('Login failed', { email: req.body.email, projectId, attempts, delayMs });
+    await sleep(delayMs);
```
mattwiller: This isn't obvious to me: why would we want to hold the connection open for the sleep duration, before returning the error?  Also, how would this block the user from issuing more login requests?

packages/server/src/ratelimit.ts
```diff
@@ -158,3 +158,15 @@ function getRateLimitCategory(req: Request): RateLimitCategoryConfig {
...
+
+/**
+ * Returns the current rate limit state for the given request without consuming any points.
+ * @param req - The request to peek the rate limit state for.
+ * @param config - Optional server config.
+ * @returns The current rate limit state, or null if no rate limit applies.
+ */
+export function peekRateLimitState(req: Request, config?: MedplumServerConfig): Promise<RateLimiterRes | null> {
+  const rateLimiter = getRateLimiter(req, config);
+  const key = getRateLimitKey(req);
```
mattwiller: Rather than piggybacking on the general auth rate limiter, we probably want to use a dedicated "failed login" limiter here to ensure correct blocking behavior

packages/server/src/auth/login.ts
codyebberson: > why would we want to hold the connection open for the sleep duration

This approach was based on guidance from our HITRUST audit consultant and is a pattern used in some systems to slow down credential enumeration by increasing the cost per attempt.

I agree with the concern around parallel requests and resource usage. Holding connections open isn’t effective against concurrent attacks and can create unnecessary load.

I’m updating the implementation to remove the server-side delay and instead rely on authentication-specific rate limiting / temporary blocking (HTTP 429), which achieves the same goal more robustly.

codyebberson:
@claude please review

The original PR had more substance, but I reverted most of it.  Now it's just two minor cleanup changes.

noahsilas (approved):

PR #8983, opened 2026-04-17
Scheduling: Use Reference<HealthcareService> in place of CodeableConcept[]

noahsilas:
There were some challenges relating to Scheduling when everything was being implemented in terms of linking ServiceType attributes that were CodeableConcept arrays:

- Comparing two CodeableConcept entries is not a strongly defined operation; we had implemented such that two concepts are considered "matching" if they have even a single overlapping Coding. This could result in surprising matches (if a single code was present in multiple places) or mismatches (if a Coding didn't match exactly, such as if one had a `system` attribute and one didn't).

- Following a Reference<HealthcareService> lets us read from the primary key index instead of doing a search on the `service-type` search parameter in more places.

- If passing a reference to a HealthcareService that the caller can't read (either due to permissions or because it doesn't exist), we can emit a more useful error message.

- Previously, changing `HealthcareService.type` to a new code would break relationships from any Schedules pointing at it. By encoding the links as References, they are now stable (although it can be confusing if they store denormalized data, as in `Schedule.serviceType[].coding`, that gets out of sync).

Implementation Notes:
=====================

- When linking a Schedule to a HealthcareService, in Schedule.serviceType we expect the CodeableConcept entries to carry an extra extension holding an explicit reference to a HealthcareService. This is similar in concept to an R5 CodeableReference (which this field will be changing to in the future).

  In the meantime, this style lets us perform cleaner checks if a Schedule is intended to be linked to a specific HealthcareService by reading these references.

- The SchedulingParameters extension on Schedule now has a subextension `url: "service", valueReference: Reference<HealthcareService>`. This makes the intention of overriding a specific HealthcareService clear.

  Aside: Down the line we think there will be configurations where a Schedule wants to override availability on many/all HealthcareService rows. At the time that we implement that we will likely make the contraint on this field weaker. For now, we consider it as a requirement and fail without this explicit linkage.

- Changed the parameter type on $find - `service-type` used to take a comma-separated list of code tokens. It now takes `service-type-reference`, which must be exactly one `Reference<HealthcareService>`. This makes it abundantly clear what the intention is: "schedule an appointment of this type", where the caller has made the type selection already by inspecting the list of available HealthcareServices.

- Added a Schedule Settings Page to Provider app to ease configuration.

  When the viewing the Provider app SchedulePage for a ScheduleResource with a Medplum SchedulingParameters extension, we add a gear icon in the top right that links to this page. It allows easily setting Schedule.serviceType entries by toggling switches for HealthcareService resources.

UI Demo:
=======

https://github.com/user-attachments/assets/1ba269b1-2d3d-46ea-8f6f-d257655051a3

### DIFF
--- a/packages/server/src/fhir/operations/book.ts
+++ b/packages/server/src/fhir/operations/book.ts
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
+import type { WithId } from '@medplum/core';
 import {
   badRequest,
   conflict,
@@ -24,9 +25,9 @@ import type {
 import { getAuthenticatedContext } from '../../context';
 import { addMinutes, areIntervalsOverlapping } from '../../util/date';
 import { invariant } from '../../util/invariant';
+import { extractReferencesFromCodeableReferenceLike } from '../../util/servicetype';
 import { buildOutputParameters, parseInputParameters } from './utils/parameters';
 import { applyExistingSlots, getTimeZone, resolveAvailability } from './utils/scheduling';
-import type { SchedulingParameters } from './utils/scheduling-parameters';
 import { chooseSchedulingParameters } from './utils/scheduling-parameters';
 
 const bookOperation = {
@@ -77,33 +78,6 @@ function serviceTypeTokens(slots: Slot[]): string[] {
     }
   }
   return [...tokenSet.values()];
-}
-
-function chooseActiveParameters(
-  proposedSlot: Slot,
-  parameters: SchedulingParameters[],
-  actorTimeZone: string,
-  existingSlots: Slot[]
-): SchedulingParameters | undefined {
-  const startDate = new Date(proposedSlot.start);
-  const endDate = new Date(proposedSlot.end);
-  return parameters.find((params) => {
-    const timeZone = params.timezone ?? actorTimeZone;
-    const range = {
-      start: addMinutes(startDate, -1 * params.bufferBefore),
-      end: addMinutes(endDate, params.bufferAfter),
-    };
-    const availability = resolveAvailability(params, range, timeZone);
-    const result = applyExistingSlots({
-      availability,
-      slots: existingSlots,
-      range,
-      serviceType: proposedSlot.serviceType ?? EMPTY,
-    });
-    return result.some(
-      (interval) => interval.start.getTime() <= range.start.getTime() && interval.end.getTime() >= range.end.getTime()
-    );
-  });
 }
 
 /**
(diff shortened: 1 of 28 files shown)


packages/server/src/fhir/operations/book.ts
```diff
@@ -157,17 +131,43 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
+      throw new OperationOutcomeError(badRequest('HealthcareService not found'));
+    }
+  } else {
+    // Collect all unique service type codes across all proposed slots, then fetch
+    // matching HealthcareService resources in a single query.
+    //
+    // Q: Do we support this style or require the CodeableReference style above?
+    const allServiceTypes = serviceTypeTokens(proposedSlots);
+
+    const healthcareServices: WithId<HealthcareService>[] =
```
codyebberson, line 158: To confirm my understanding...  The intent here is to support two methods of passing in `HealthcareService`

1. By referencing `HealthcareService` in `Slot.serviceType` extension
2. Searching for `HealthcareService` with `HealthcareService?service-type=${slot.serviceType}` 

?

packages/server/src/fhir/operations/book.ts
```diff
@@ -157,17 +131,43 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
+            filters: [{ code: 'service-type', operator: Operator.EQUALS, value: allServiceTypes.join(',') }],
+          })
+        : [];
+
+    if (healthcareServices.length === 0) {
+      throw new OperationOutcomeError(badRequest('No matching HealthcareService found'));
+    }
+
+    if (healthcareServices.length > 1) {
+      throw new OperationOutcomeError(badRequest('Multiple matching HealthcareServices found'));
```
codyebberson, line 171: In theory, this check would go away in the future?

packages/server/src/fhir/operations/book.ts
```diff
@@ -187,24 +187,21 @@ export async function appointmentBookHandler(req: FhirRequest): Promise<FhirResp
...
-            .filter((params) => params.duration === durationMinutes);
-
-          if (parameters.length === 0) {
+          if (parameters?.duration !== durationMinutes) {
             throw new OperationOutcomeError(badRequest('No matching scheduling parameters found'));
           }
 
-          const bufferBeforeMax = Math.max(...parameters.map((p) => p.bufferBefore));
-          const bufferAfterMax = Math.max(...parameters.map((p) => p.bufferAfter));
+          const timeZone = parameters.timezone ?? actorTimeZone;
```
codyebberson, line 201: Super minor nit, but my OCD would love it if we could standardize on `timeZone` vs `timezone`

packages/server/src/fhir/operations/find.ts
```diff
@@ -37,7 +32,7 @@ const findOperation = {
   parameter: [
     { use: 'in', name: 'start', type: 'dateTime', min: 1, max: '1' },
     { use: 'in', name: 'end', type: 'dateTime', min: 1, max: '1' },
-    { use: 'in', name: 'service-type', type: 'string', min: 1, max: '1' },
+    { use: 'in', name: 'service-type-reference', type: 'string', min: 1, max: '1', searchType: 'reference' },
```
codyebberson, line 35: 🤔   This makes sense to me, because we're now requiring `HealthcareService`.

Making a mental note that this deviates from the Argonaut spec: https://www.fhir.org/guides/argonaut/scheduling/OperationDefinition-appointment-find.html

Which, I guess `service-type` already deviated from `visit-type`.

In theory, in the future, we could support those Argonaut inputs, and map them to `HealthcareService` resources, yeah?  For similar functionality, but with an Argonaut-compatible API?

packages/server/src/util/servicetype.ts
```diff
@@ -0,0 +1,91 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { WithId } from '@medplum/core';
+import { createReference, EMPTY, getExtensionValue, getReferenceString, isDefined } from '@medplum/core';
+import type { CodeableConcept, HealthcareService, Reference } from '@medplum/fhirtypes';
+
+/**
+ * In R5/R6, `serviceType` attributes are changing from `CodeableConcept[]` to
```
codyebberson, line 8: These are good utils.  Making a mental note, we may want to move these to `core` for general availability, after the dust settles.

codyebberson (approved):
Nice, thanks @noahsilas  - this all looks good to me.  A few minor questions, mostly clarifications.

codyebberson:
@claude please review

packages/server/src/fhir/operations/book.ts
noahsilas: Yeah, you got it. This is one I felt slightly hung up on, so for now I left the older style here too.

- The first scenario is that we have an explicit reference in the custom extension. That should be the case for any results coming direct from $find, for instance.
- The second scenario is falling back to the older behavior trying to match serviceType on codes. This could let folks use $book without the custom CodeableConcept extensions. But since you need the SchedulingParameters extension in at least one place anyways, maybe I shouldn't get hung up on this and should stay more consistent and drop this branch?

I'd love to hear your thoughts on this particular API edge case!

packages/server/src/fhir/operations/book.ts
noahsilas: I think that we want to tie a given appointment to a single HealthcareService (which is where parameters like slot duration is likely to come from). On this `else` branch we didn't have an explicit reference, so we're doing a search on service-type codes, which makes it easy to accidentally get multiple matching services and cause confusion about what should happen.

We can probably try to get more clever about choosing a matching healthcare service if there are multiple matches, but my hope is that mostly we would be on the above path where we got an explicit reference.

packages/server/src/fhir/operations/book.ts
noahsilas: +1. I think I started with `timezone` and then saw that a lot of the newer JS world seems to be using `timeZone` and didn't go back and clean up. I'll follow up on this!

packages/server/src/fhir/operations/find.ts
noahsilas: Yeah, we could start by using the `visit-type` codes to search `HealthcareService.service-type`, and resolve that to a single healthcareService to pass into our `$find` implementation.

Somewhat interestingly, on that page the parameter is documented with name `visit-type` but the Documentation column of that same row also says "`service-type=[system]|[code]`". 🤔

PR #8986, opened 2026-04-17
docs(self-hosting): add docs for enabling rollbacks

ThatOneBro:
Although rollbacks are not recommended as Medplum is a continuously deployed application, it is an often requested feature.

These docs are for showing how you can enable rollbacks and safely rollback to the previous minor version, assuming you haven't applied the post-deploy migrations in the previous version

### DIFF
--- a/packages/docs/docs/self-hosting/enabling-rollbacks.md
+++ b/packages/docs/docs/self-hosting/enabling-rollbacks.md
@@ -32,10 +32,10 @@ Suppose your cluster is running `v4.2.6` and you want to upgrade to `v4.3.x`:
 ```
 v4.2.6 (latest patch of previous minor) ─┐
                                          │  ← safe rollback window
-v4.3.0 deployed, post-deploy deferred ───┘
+v4.3.15 deployed, post-deploy deferred ───┘
 ```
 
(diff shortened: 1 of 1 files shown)


The-Alchemist (approved):

packages/docs/docs/self-hosting/enabling-rollbacks.md
```diff
@@ -0,0 +1,142 @@
...
+sidebar_position: 110
+tags:
+  - self-hosting
+  - upgrades
+  - rollbacks
+---
+
+# Enabling Rollbacks for Medplum Server
+
+Rolling back a Medplum server deployment is normally not supported, because post-deploy migrations can alter data in ways that are not reversible by simply deploying an older image. However, starting with the `4.x` series, Medplum's minor version cadence gives you a narrow, well-defined window where a rollback is safe — provided you take control of **when** post-deploy migrations run.
```
codyebberson: Technically correct, but I think we can start with a more positive take.


```suggestion
As of the `4.x` series, Medplum supports a safe and well-defined rollback window for server deployments, provided you control when post-deploy migrations run.

:::note
Rolling back a Medplum server deployment has historically been unsafe, because post-deploy migrations can alter data in ways that are not reversible by simply deploying an older image. The `4.x` release process introduces a clearer separation between reversible and irreversible changes, enabling controlled rollbacks under specific conditions.
:::
```

PR #8987, opened 2026-04-17
Transaction dead connection management

### LINKED ISSUE #8984: Connection pool exhaustion: rollbackTransaction doesn't release pool slot when connection is killed by idle_in_transaction_session_timeout
## Description

Running Medplum server v5.1.8 on AWS ECS with Aurora PostgreSQL 16.13, we observe monotonic connection pool exhaustion over ~16 hours until all 100 pool slots are occupied and the server becomes unresponsive (users cannot sign in).

## Environment

- **Medplum server**: v5.1.8 (also reproduced on v5.1.6)
- **Database**: Aurora PostgreSQL 16.13 (db.t3.medium, `max_connections` = 450)
- **Deployment**: ECS Fargate, 1-2 tasks
- **Pool max**: 100 (default)

## Symptoms

1. After a fresh ECS task deployment, `DatabaseConnections` in CloudWatch starts at ~5
2. Connections climb steadily at ~5-6/hour
3. After ~16 hours, connections reach 100 and plateau
4. All new requests hang indefinitely — pool has no available slots
5. RDS PostgreSQL logs show repeated:
   ```
   FATAL: terminating connection due to idle-in-transaction timeout
   LOG: unexpected EOF on client connection with an open transaction
   ```
6. Only fix is ECS task restart, then the cycle repeats

## CloudWatch evidence (48-hour window)

```
Apr 15 12:28  avg=7    (restart)
Apr 15 17:28  avg=36   (climbing ~5/hr)
Apr 16 04:28  avg=86
Apr 16 11:28  avg=99   (pegged at 100)
Apr 17 11:28  avg=99   (still pegged, 30+ hours)
```

## Root cause analysis

The `idle_in_transaction_session_timeout=30000` added in PR #8630 (v5.1.4) correctly kills idle-in-transaction connections on the PostgreSQL side. However, when this happens, the `pg.Pool` slot on the Node.js side is never reclaimed.

In `packages/server/src/fhir/repo.ts`, `rollbackTransaction()` (line ~2625 on main):

```typescript
private async rollbackTransaction(error: Error): Promise<void> {
    this.assertInTransaction();
      await conn.query('ROLLBACK');          // ← throws if connection already killed
      this.transactionDepth--;               // ← never reached
      this.truncateCommitCallbacks();        // ← never reached  
      this.releaseConnection(error);         // ← never reached → pool slot is zombie
    } else {
      // ...
    }
}
```

When PostgreSQL kills the connection via `idle_in_transaction_session_timeout`, the subsequent `ROLLBACK` query throws because the socket is dead. Since there is no `try/finally`, `transactionDepth` is never decremented and `releaseConnection()` is never called.

### The safety net (`Symbol.dispose`) is unreliable

The Express error handler calls `closeRequestContext()` → `repo[Symbol.dispose]()` which checks `transactionDepth > 0` and calls `releaseConnection(error)`. However, when `rollbackTransaction` throws without decrementing `transactionDepth`, and `withTransaction` retries the loop, `beginTransaction` increments `transactionDepth` again (now = 2). The state is corrupted — subsequent cleanup may not release the slot correctly.

### What PR #8682 addresses (and doesn't)

PR #8682 (v5.1.8) added `truncateCommitCallbacks()` which correctly clears stale pre/post-commit callbacks on rollback. This reduces the **frequency** of transactions getting stuck (fewer spurious post-rollback callbacks holding connections). But it does **not** add a `try/finally` around the `ROLLBACK` query, so the pool slot leak still occurs when the connection is already dead.

## Suggested fix

Wrap the ROLLBACK query in try/finally to ensure cleanup always runs:

```typescript
private async rollbackTransaction(error: Error): Promise<void> {
    this.assertInTransaction();
      try {
        await conn.query('ROLLBACK');
      } catch {
        // Connection may already be dead (killed by idle_in_transaction_session_timeout)
      } finally {
        this.transactionDepth--;
        this.truncateCommitCallbacks();
        this.releaseConnection(error);  // always release, passing error removes from pool
      }
    } else {
      try {
        await conn.query('ROLLBACK TO SAVEPOINT sp' + this.transactionDepth);
      } catch {
        // Connection may already be dead
      } finally {
        this.transactionDepth--;
        this.truncateCommitCallbacks();
      }
    }
}
```

Additionally, consider exposing `connectionTimeoutMillis` in `MedplumDatabaseConfig` so that pool exhaustion causes requests to fail fast (with an error) instead of hanging forever.

## Related

- Issue #8620 — Database lock condition under load testing (showed 198 idle-in-transaction connections)
- PR #8630 — Added `idle_in_transaction_session_timeout=30000` (v5.1.4)
- PR #8767 — Re-use repo DB connection for system access (v5.1.5)
- PR #8682 — Clear repo commit hooks on transaction rollback (v5.1.8)
- node-postgres issue #2852 — pg-pool slot accounting desync when server kills connection
- node-postgres PR #3597 — `evictOnOpenTransaction` option (still open/unmerged)

mattlong:
Fix connection leak when transaction rollback fails on dead backend

  When a transaction's client is killed mid-flight (e.g., by
  idle_in_transaction_session_timeout), rollbackTransaction's ROLLBACK
  query also rejects. The subsequent depth-- / releaseConnection() lines
  never run, leaving the client checked out forever. Over time the pool
  saturates; on shutdown, pool.end() hangs waiting for the leaked client.

  - Wrap ROLLBACK / ROLLBACK TO SAVEPOINT in try/catch. On failure,
    abortTransaction() resets depth, clears the callback stack, and
    releases the client with the original error so pg-pool discards it.
  - releaseConnection() guards against double-release and clears
    this.conn even on non-disposable repos when an error is passed, so a
    dead client is never reused.
  - rollbackTransaction() no-ops at depth 0, preventing "Not in
    transaction" from masking the root cause when cleanup already ran.
  - commitTransaction() intentionally unchanged: RELEASE SAVEPOINT
    failures (aborted-transaction state) are still recoverable by the
    outer scope's rollback, which can run ROLLBACK TO SAVEPOINT against
    an aborted transaction.
  - Add medplum.db.totalConnections gauge to the otel heartbeat so slow
    leaks are observable before shutdown (idleCount / waitingCount alone
    don't show stuck-checked-out clients).
  - Regression test asserts state fully cleans up when ROLLBACK rejects.

Fixes #8984

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2623,26 +2623,17 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     const conn = await this.getConnection(DatabaseMode.WRITER);
     if (this.transactionDepth === 1) {
       await this.processPreCommit();
-      try {
-        await conn.query('COMMIT');
-      } catch (err) {
-        // COMMIT failed (e.g. connection was terminated by idle_in_transaction_session_timeout).
-        // Postgres has already rolled back, so skip ROLLBACK attempts.
-        this.abortTransaction(err as Error);
-        throw err;
-      }
+      await conn.query('COMMIT');
       this.transactionDepth--;
       this.releaseConnection();
       this.clearCallbackStack();
       await this.processPostCommit();
     } else {
-      try {
-        await conn.query('RELEASE SAVEPOINT sp' + this.transactionDepth);
-      } catch (err) {
-        // Savepoint release failed; outer transaction is toast. Abort the whole chain.
-        this.abortTransaction(err as Error);
-        throw err;
-      }
+      // If RELEASE SAVEPOINT fails (e.g. transaction in aborted state), let the error propagate.
+      // withTransaction's catch will invoke rollbackTransaction, which can run ROLLBACK TO SAVEPOINT
+      // even against an aborted transaction to recover — aborting here would discard work the outer
+      // scope can still commit. rollbackTransaction's own catch handles the truly-dead-connection case.
+      await conn.query('RELEASE SAVEPOINT sp' + this.transactionDepth);
       this.transactionDepth--;
       this.popCallbackFrame();
     }
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2516,8 +2516,21 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    * @param err - Optional error to remove the connection from the pool.
    */
   private releaseConnection(err?: boolean | Error): void {
-    if (this.conn && this.disposable) {
-      this.conn.release(err);
+    if (!this.conn) {
+      return;
+    }
+    if (this.disposable) {
+      try {
+        this.conn.release(err);
+      } catch (releaseErr) {
+        // pg-pool throws if release() is called twice (e.g. socket already errored out).
+        // We've done our part; just log and move on.
+        getLogger().warn('Error releasing database client', { err: normalizeErrorString(releaseErr) });
+      }
+      this.conn = undefined;
+    } else if (err) {
+      // Shared connection is known to be dead. Drop our reference so we don't reuse it.
+      // The owner of the connection is responsible for the actual release.
       this.conn = undefined;
     }
   }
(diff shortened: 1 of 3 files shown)


packages/server/src/otel/otel.ts
```diff
@@ -117,6 +117,10 @@ export function initOtelHeartbeat(): void {
     const writerPool = getDatabasePool(DatabaseMode.WRITER);
     const readerPool = getDatabasePool(DatabaseMode.READER);
 
+    setGauge('medplum.db.totalConnections', writerPool.totalCount, {
```
mattwiller, line 120: 👍

packages/server/src/fhir/repo.ts
```diff
@@ -2610,33 +2623,76 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+      if (isOuter) {
+        await conn.query('ROLLBACK');
+      } else {
+        await conn.query('ROLLBACK TO SAVEPOINT sp' + this.transactionDepth);
+      }
+    } catch (rollbackErr) {
+      // ROLLBACK itself failed — connection is almost certainly dead (e.g. killed by
+      // idle_in_transaction_session_timeout). Pass the original triggering error to
+      // abortTransaction so the client is released with the right root cause.
+      getLogger().warn('Error rolling back transaction', {
```
mattwiller, line 2660: 👍

packages/server/src/fhir/repo.ts
```diff
@@ -2616,27 +2629,61 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       this.clearCallbackStack();
       await this.processPostCommit();
     } else {
+      // If RELEASE SAVEPOINT fails (e.g. transaction in aborted state), let the error propagate.
+      // withTransaction's catch will invoke rollbackTransaction, which can run ROLLBACK TO SAVEPOINT
+      // even against an aborted transaction to recover — aborting here would discard work the outer
+      // scope can still commit. rollbackTransaction's own catch handles the truly-dead-connection case.
```
mattwiller, line 2635: 👍

mattwiller (approved):

PR #8988, opened 2026-04-18
DICOM

codyebberson:
Adds first-pass DICOMweb support to Medplum so DICOM data can be received by the Medplum Agent, forwarded to the server over HTTPS using STOW-RS, stored as Medplum resources, and retrieved by DICOMweb clients such as OHIF Viewer.

The main data model additions are three Medplum resource types:

- `DicomStudy`
- `DicomSeries`
- `DicomInstance`

Each uploaded DICOM instance is stored as a raw `Binary`, indexed into the study/series/instance resource hierarchy, and processed asynchronously to extract pixel data into retrievable `Binary` resources.

## What Changed

- Adds generated definitions, schemas, docs data, and TypeScript types for `DicomStudy`, `DicomSeries`, and `DicomInstance`.
- Adds database migration `v109` to create tables, history tables, reference tables, and search indexes for the new DICOM resources.
- Moves the DICOMweb router to `/dicomweb/` and adds CORS support for that prefix.
- Implements DICOMweb STOW-RS upload handling for `POST /dicomweb/studies` and `POST /dicomweb/studies/:studyUid`.
- Implements initial QIDO-RS support for:
  - `GET /dicomweb/studies`
  - `GET /dicomweb/studies/:studyUid/series`
- Implements initial WADO-RS support for:
  - `GET /dicomweb/studies/:studyUid/series/:seriesUid/metadata`
  - `GET /dicomweb/studies/:studyUid/series/:seriesUid/instances/:instanceUid/frames/:frame`
- Adds a DICOM background worker that processes `DicomInstance.raw`, extracts pixel data, stores it as `Binary`, and patches `DicomInstance.pixelData`.
- Updates the dispatch worker to enqueue DICOM processing when a `DicomInstance` raw binary changes.
- Updates the Medplum Agent DICOM channel to convert inbound DIMSE C-STORE payloads into DICOMweb STOW-RS requests instead of creating FHIR `Binary` resources directly.
- Adds `medplum dicomweb stow <file>` for manual STOW-RS upload testing from the CLI.
- Adds server-side DICOM utility helpers for DICOM JSON cleanup, metadata conversion, date/time conversion, person-name conversion, and multipart body writing.
- Adds `dcmjs` and `dicomweb-multipart-parser` dependencies where needed.

## Supported Scope

This PR is focused on the end-to-end path needed for the hackathon demo:

1. Receive DICOM data through the Medplum Agent.
2. Forward the DICOM payload to Medplum Server using DICOMweb STOW-RS.
[...]

PR #8989, opened 2026-04-19
Add MedplumClient support for ReadableStream and Readable

codyebberson:
Extracted from https://github.com/medplum/medplum/pull/8988

`MedplumClient` currently supports a few input types for HTTP request content body:

1. `string`
2. `Blob`
3. `File`
4. `Uint8Array`
5. Or a JavaScript object, which is automatically converted to a JSON string

This PR adds two more types:

1. `ReadableStream` - browser stream (detected as `.getReader`)
2. `Readable` - Node.js stream (detected via `.pipe` and `.on`)

If a stream is detected, `MedplumClient` will automatically add `duplex: 'half'` (see https://developer.mozilla.org/en-US/docs/Web/API/Request/duplex)

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -450,6 +450,11 @@ export interface MedplumRequestOptions extends RequestInit {
    * Only applies when the client is configured with auto-batching enabled.
    */
   disableAutoBatch?: boolean;
+
+  /**
+   * See: https://developer.mozilla.org/en-US/docs/Web/API/Request/duplex
+   */
+  duplex?: 'half';
 }
 
 export interface PushToAgentOptions extends MedplumRequestOptions {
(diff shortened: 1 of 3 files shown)


packages/core/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "composite": true,
     "module": "esnext",
-    "types": ["vitest/globals", "vite/client"],
+    "types": ["node", "vitest/globals", "vite/client"],
```
ThatOneBro, line 6: Do we want to add `node` to `core`? 👀 I think that might add global types that the browser won't have

packages/core/src/client.ts
```diff
@@ -3934,6 +3939,13 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
       (typeof Uint8Array !== 'undefined' && (data instanceof Uint8Array || data?.constructor.name === 'Uint8Array'))
     ) {
       options.body = data;
+    } else if (
+      typeof data === 'object' &&
+      data !== null &&
+      (typeof data.getReader === 'function' || (typeof data.pipe === 'function' && typeof data.on === 'function'))
+    ) {
+      options.body = data;
+      options.duplex = 'half'; // Required for ReadableStream in Node 18+
```
ThatOneBro:
```suggestion
      options.duplex = 'half'; // Required for ReadableStream in Node 18+ and in most browsers
```

ThatOneBro (approved):
Nice

packages/core/tsconfig.json
codyebberson: This was to make TypeScript 6 happy.  We're already using Node.js types in a bunch of tests, which were full of red squigglies.  I don't know if there's a good way to say "allow `node` types in tests, but not in final build"?

ThatOneBro: Yeah.... I'm guessing the "right" way to do this is probably via some kind of includes, excludes config... sounds complicated 🥲

PR #8991, opened 2026-04-20
Dependency upgrades 2026-04-20

medplumbot:
Dependency upgrades


ThatOneBro:
I excluded `@mantine/*` major updates for now since it looks like with some backwards-incompatible changes we might have to introspect on the version of Mantine in order to conditionally use the correct props for a component if we wanted to support both 8 and 9, which seems gross. Might be worth looking at it out of band.

codyebberson (approved):

PR #8998, opened 2026-04-21
Added ai-realtime websocket for speech-to-text

codyebberson:


packages/server/src/ws/ai-realtime.ts
```diff
@@ -0,0 +1,200 @@
...
+      sendError('OpenAI API key not configured in project secrets');
+      closeClient(1011);
+      return;
+    }
+
+    upstreamSocket = openAIRealtimeWebSocketFactory(apiKey, 'transcription');
+    bindUpstreamSocket(upstreamSocket);
+  }
+
+  function bindUpstreamSocket(upstream: WebSocket): void {
```
ThatOneBro: We should probably be using ReconnectingWebSocket here

If we can't due to the ping-pong we baked in, we can add a constructor param to disable that

Otherwise joining two potentially flaky sockets seems 🙃

codyebberson: Good call, makes sense.

packages/server/src/ws/ai-realtime.ts
```diff
@@ -0,0 +1,200 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import { badRequest, normalizeErrorString } from '@medplum/core';
+import type { IncomingMessage } from 'node:http';
+import type { WebSocket as IncomingWebSocket, RawData } from 'ws';
+import WebSocket from 'ws';
+import { globalLogger } from '../logger';
+import { getLoginForAccessToken } from '../oauth/utils';
+
```
ThatOneBro, line 13: Can we get the same metrics we have for all the WebSocket handlers here?

techdavidy (approved):

PR #8999, opened 2026-04-21
React and Provider App components for Meds Ordering

oleg-mp:
Supports teeing up medications for order with DAW Scriptsure


packages/core/src/medication-order-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+  readonly refill?: number;
+  readonly drugOrder?: number;
+  readonly sigLine3?: string;
+  readonly useSubstitution?: boolean;
+}
+
+/**
+ * Vendor-neutral input for an e-prescribing order-medication bot (matches ScriptSure bot shape).
+ */
+export interface MedicationOrderRequest {
```
codyebberson, line 44: Let's discuss offline.  Can these be FHIR `MedicaitonRequest` resources?

packages/cli/Project_data_55555_aaaaaa_bbbbb_ccc_ddddd_eeeeee_ndjson.ndjson
```diff
@@ -0,0 +1 @@
+download data
```
codyebberson: Is this file used?

packages/react-hooks/src/useMedicationOrder/useMedicationOrder.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Generic hook for e-prescribing drug search and order-medication bots.
+ *
+ * @param searchBotIdentifier - Bot identifier for drug search (returns Medication[]).
+ * @param orderBotIdentifier - Bot identifier for creating a pending order (returns MedicationOrderResponse from medplum-core).
+ * @returns Callbacks to search medications and submit an order bot request.
+ */
+export function useEPrescribingOrder(
+  searchBotIdentifier: Identifier,
```
codyebberson: Let's discuss w/ FDE team.  I think they're trying to make a push to using Bots via custom FHIR operations rather than Bot identifiers.

packages/react-hooks/src/useEPrescribingOrder/useEPrescribingOrder.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Generic hook for e-prescribing drug search and order-medication bots.
+ *
+ * @param searchBotIdentifier - Bot identifier for drug search (returns Medication[]).
+ * @param orderBotIdentifier - Bot identifier for creating a pending order (returns MedicationOrderResponse from medplum-core).
+ * @returns Callbacks to search medications and submit an order bot request.
+ */
+export function useEPrescribingOrder(
```
codyebberson: Minor nit, but let's discuss hook naming.  Naming things is hard.  Something about the "E" feels off to me.

codyebberson (commented):
Let's discuss offline.  The main question I have is whether we can use FHIR `MedicationRequest` rather than inventing a new type.  I know the FHIR `MedicationRequest` is famously unpopular, so I'm open to this discussion, but I want to make sure we've done a proper investigation.

examples/medplum-provider/src/components/meds/quantity-qualifiers.ts
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * Rules are sorted by `name` length descending so that more specific entries
+ * win over shorter prefixes (e.g. "Applicatorful" beats "Applicator").
+ *
+ * @param catalog - Live response from `GET /v3/prescription/quantityqualifier`
+ *   (and/or the static fallback rows).
+ * @returns A {@link QualifierMatcher} closed over the catalog.
+ */
+export function buildQualifierMatcher(catalog: readonly { potencyUnit: string; name: string }[]): QualifierMatcher {
+  const ordered = [...catalog]
```
noahsilas: I think that you can drop this cloning of `catalog` into a new array, as `.filter()` should create a shallow copy itself.

packages/cli/Project_data_55555_aaaaaa_bbbbb_ccc_ddddd_eeeeee_ndjson.ndjson
oleg-mp: no, this is an artifact of local testing. I keep deleting it but it comes back. I'll remove

packages/core/src/medication-order-utils.ts
oleg-mp: The bot supports MedicationRequest as an input as well. Wanted to have an alternative option for when someone decides NOT to use it on purpose, so yeah can discuss. The main pattern now is this: we create a draft MedicationRequest that gets fed into the bot, which transfers it to DAW as a new order, the result is a confirmation with ID and widget URL that let's prescriber complete the order.

packages/react-hooks/src/useMedicationOrder/useMedicationOrder.ts
rahul1: +1 push for operations. Less clunky, instance-level operations steamline call patterns, and in the future makes it easier to shadow built-in operations

packages/scriptsure-react/src/useScriptSureOrderMedication.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * React hook for ScriptSure drug search and pending order creation (iframe launch URL).
+ *
+ * Thin wrapper around `useEPrescribingOrder` from `@medplum/react-hooks` with ScriptSure bot identifiers.
+ *
+ * @returns The same API as `useEPrescribingOrder` (search + order callbacks).
+ */
+export function useScriptSureOrderMedication(): UseScriptSureOrderMedicationReturn {
+  return useEPrescribingOrder(SCRIPTSURE_DRUG_SEARCH_BOT, SCRIPTSURE_ORDER_MEDICATION_BOT);
```
rahul1: Again, if we use shared operation names, we may not even need to store these bot ids

examples/medplum-provider/src/pages/patient/MedicationsPage.tsx
```diff
@@ -1,71 +1,146 @@
...
+
+  const fetchOrders = useCallback(async (): Promise<void> => {
+    if (!patientReference) {
+      showErrorNotification('Patient not found');
+      return;
+    }
+    try {
+      const searchParams = new URLSearchParams({
+        subject: patientReference,
+        _count: '100',
```
techdavidy: we can add pagination to the page

examples/medplum-provider/src/pages/patient/MedicationsPage.tsx
```diff
@@ -75,85 +150,225 @@ export function MedicationsPage(): JSX.Element {
...
-  const fields = search.fields ?? ['_id', '_lastUpdated'];
-  const offset = search.offset ?? 0;
-  const count = search.count ?? DEFAULT_SEARCH_COUNT;
-  return {
-    ...search,
-    fields,
-    offset,
-    count,
-  };
+function filterMedicationRequestsByTab(requests: MedicationRequest[], tab: MedTab): MedicationRequest[] {
```
techdavidy: rather than doing it locally, we can send a new request to fetch with the corresponding filter

`status=draft`

examples/medplum-provider/src/pages/patient/MedicationsPage.tsx
```diff
@@ -75,85 +150,225 @@ export function MedicationsPage(): JSX.Element {
...
+              <Flex h={64} align="center" justify="space-between" p="md" wrap="wrap" gap="xs">
+                <Group gap="xs">
+                  <Tabs
+                    value={activeTab}
+                    onChange={(v) => setActiveTab((v as MedTab) || 'active')}
+                    variant="unstyled"
+                    className="pill-tabs"
+                  >
+                    <Tabs.List>
+                      <Tabs.Tab value="active">Active</Tabs.Tab>
```
techdavidy, line 325: We follow a pattern where the filter navigates to a url with the filter applied. 
Similar to app.medplum.com
`https://app.medplum.com/MedicationRequest?_count=20&_fields=id,_lastUpdated&_sort=-_lastUpdated&status=draft`

it is easier to share urls that way.

examples/medplum-provider/src/components/meds/MedicationRequestDetails.tsx
```diff
@@ -0,0 +1,400 @@
...
+}
+
+function DetailRow(props: { label: string; children: ReactNode }): JSX.Element {
+  const { label, children } = props;
+  return (
+    <Group align="flex-start" gap="lg" wrap="nowrap">
+      <Text fw={500} size="sm" style={{ width: '150px', flexShrink: 0 }} c="dimmed">
+        {label}
+      </Text>
+      <Stack gap={4} style={{ flex: 1 }}>
```
techdavidy: nit: flex={1}

examples/medplum-provider/src/components/meds/MedicationRequestDetails.tsx
```diff
@@ -0,0 +1,400 @@
...
+      </Text>
+    </Group>
+  );
+}
+
+function DetailRow(props: { label: string; children: ReactNode }): JSX.Element {
+  const { label, children } = props;
+  return (
+    <Group align="flex-start" gap="lg" wrap="nowrap">
+      <Text fw={500} size="sm" style={{ width: '150px', flexShrink: 0 }} c="dimmed">
```
techdavidy: w={150}

techdavidy:
@claude please review

oleg-mp:
@claude review latest changes and fixes in the context of addressing PR comments

examples/medplum-provider/src/components/meds/MedicationRequestDetails.tsx
```diff
@@ -0,0 +1,400 @@
...
+
+          <Stack gap="sm">
+            {medicationRequest.medicationCodeableConcept?.coding &&
+              medicationRequest.medicationCodeableConcept.coding.length > 0 && (
+                <DetailRow label="Medication codes">
+                  {medicationRequest.medicationCodeableConcept.coding.map((c, i) => (
+                    <Text key={i} size="sm">
+                      {c.display ? `${c.display} · ` : ''}
+                      {c.system} | {c.code}
+                    </Text>
```
noahsilas, line 216: I wonder if this can use the existing `<CodableConceptDisplay>` component? It looks like you might need to teach that to accept an `includeCode` prop (similar to `<CodingDisplay includeCode>`) to achieve your goal here though.

examples/medplum-provider/src/components/meds/MedListItem.module.css
```diff
@@ -0,0 +1,30 @@
...
+  white-space: normal;
+  word-break: break-word;
+}
+
+.title {
+  white-space: nowrap !important;
+  overflow: hidden !important;
+  text-overflow: ellipsis !important;
+  width: 100% !important;
+  max-width: 100% !important;
```
noahsilas: All the `!important` here is a bit of a code smell, is this overriding some other rules somewhere?

examples/medplum-provider/src/components/meds/MedListItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export function MedListItem(props: MedListItemProps): JSX.Element {
+  const { item, selectedItem, activeTab, onItemSelect, medicationOrderExtensions } = props;
+  const isSelected = selectedItem?.id === item.id;
+  const requester = useResource(item.requester) as Practitioner | undefined;
+  const pendingStatus = getPendingMedicationOrderStatus(item, medicationOrderExtensions);
+
+  return (
+    <MedplumLink to={onItemSelect(item)} underline="never">
```
noahsilas: Calling `onItemSelect` during rendering is surprising to me, maybe the name of that callback could be something like `getItemUrl` to make that behavior more obvious?

examples/medplum-provider/src/components/meds/MedListItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+    </MedplumLink>
+  );
+}
+
+function getMedicationDisplay(mr: MedicationRequest): string {
+  const mc = mr.medicationCodeableConcept;
+  if (mc?.text) {
+    return mc.text;
+  }
+  return formatCodeableConcept(mc) || 'Medication order';
```
noahsilas: I think that `formatCodeableConcept()` will use the text attribute itself when present, so I think you can simplify this:


```suggestion
  const mc = mr.medicationCodeableConcept;
  return formatCodeableConcept(mc) || 'Medication order';
```

examples/medplum-provider/src/components/meds/PrescriptionIFrameModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+        intervalId = undefined;
+      }
+    };
+
+    const tick = async (): Promise<void> => {
+      try {
+        if (skipPollRef.current) {
+          return;
+        }
+        const mr = await medplum.readResource('MedicationRequest', medicationRequestIdToWatch);
```
noahsilas: I think this might want to skip the cache when reading:

```suggestion
        const mr = await medplum.readResource('MedicationRequest', medicationRequestIdToWatch, { cache: 'reload');
```

packages/core/src/medication-order-utils.ts
```diff
@@ -0,0 +1,565 @@
...
+ * @param value - Unknown bot JSON payload.
+ * @returns True when the value is a non-empty or empty array of Medication resources.
+ */
+export function isMedicationArray(value: unknown): value is Medication[] {
+  if (!Array.isArray(value)) {
+    return false;
+  }
+  if (value.length > 0) {
+    const first = value[0];
+    return typeof first === 'object' && first !== null && first.resourceType === 'Medication';
```
noahsilas: Testing only the first object here seems like it could hide type errors; if there is a tuple type like `[Medication, MedicationRequest]`or something, this would incorrectly validate it.

How about something like:
```ts
return value.every(element => isResource(element, 'Medication'))
```

If we're doing that though, maybe this should be promoted to a more generic utility? `isResourceArray<Medication>(value, 'Medication')` would mirror how `isResource` works.

packages/core/src/medication-order-utils.ts
```diff
@@ -0,0 +1,565 @@
...
+ * @param medicationRequest - MR to read.
+ * @param ext - Vendor extension URL configuration.
+ * @returns The launch URL, if present.
+ */
+export function getMedicationOrderIframeUrl(
+  medicationRequest: MedicationRequest,
+  ext: MedicationOrderExtensions
+): string | undefined {
+  const e = medicationRequest.extension?.find((x) => x.url === ext.iframeUrlExtension);
+  return e?.valueUrl;
```
noahsilas: We have a utility in core that encapsulates this pattern:

```suggestion
  return getExtensionValue(medicationRequest, ext.iframeUrlExtension);
```

packages/react-hooks/src/useMedicationOrder/useMedicationOrder.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const searchMedications = useCallback(
+    async (params: MedicationSearchParams): Promise<Medication[]> => {
+      const url = medplum.fhirUrl('Medication', '$drug-search');
+      const body = medicationSearchParamsToParameters(params);
+      const response = await medplum.post(url, body);
+      if (!isResource<Bundle<Medication>>(response, 'Bundle')) {
+        throw new Error(INVALID_MEDICATION_SEARCH_RESPONSE);
+      }
+      return (response.entry ?? []).map((e) => e.resource).filter((r): r is Medication => Boolean(r));
```
noahsilas: It would be nice to validate the entry types:
```suggestion
      return (response.entry ?? []).map((e) => e.resource).filter((r): isResource<Medication>(r, 'Medication'));
```

packages/scriptsure-react/src/loadScriptSureQuantityQualifiers.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { MedplumClient } from '@medplum/core';
+import { isResource, medicationSearchParamsToParameters } from '@medplum/core';
+import type { Parameters } from '@medplum/fhirtypes';
+
+/**
+ * Reads `{ code, label }` rows from a `Parameters` resource produced by the
+ * `$drug-quantity-qualifiers` custom operation. Tolerates either flat parts
+ * (`code` + `label` as sibling primitives — current bot output) or grouped
+ * `quantityQualifier` part-bearing entries (defensive against a future shape
+ * change).
```
noahsilas: Is this comment correct? If so, I think I might not be understanding something about the expected input with "`code` + `label` as sibling primitives". Maybe a test case for each of these inputs would make this nice and obvious?

.gitignore
```diff
@@ -57,15 +57,16 @@ tsconfig.build.tsbuildinfo
...
 # Agent artifacts
+packages/agent/agent.properties
+packages/agent/src/.*-audit.json
 packages/agent/*agent-*-linux*
 packages/agent/sea-prep.blob
 packages/agent/src/upgrader-logs-*.txt
-packages/agent/agent.properties
-packages/agent/src/.*-audit.json
-packages/agent/src/medplum-agent-main-*.log
-packages/agent/src/medplum-agent-channels-*.log
```
noahsilas, line 66: Dropping these agent entries looks like it may have been a mistake in a merge resolution (Derrick recently added these in d8e58f1bca)

examples/medplum-provider/src/pages/integrations/ScriptSurePage.tsx
```diff
@@ -19,15 +21,25 @@ export function ScriptSurePage(): JSX.Element {
...
 
+  // The bot returns a UI-agnostic URL; the app appends `darkmode=on|off` once
+  // per iframe load (theme toggles mid-session intentionally do not reload
+  // the iframe so any in-progress state is preserved).
+  const colorScheme = useComputedColorScheme('light');
+  const [renderedUrl, setRenderedUrl] = useState<string | undefined>(undefined);
+  useEffect(() => {
+    setRenderedUrl(applyDarkmode(iframeUrl, colorScheme));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [iframeUrl]);
```
noahsilas: I think this is better expressed with `useMemo` than `useState`+`useEffect`:

```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps
  const renderedUrl = useMemo(() => applyDarkmode(iframeUrl, colorScheme), [iframeUrl]);
```

examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
```diff
@@ -258,6 +266,19 @@ export function OrderLabsPage(props: OrderLabsPageProps): JSX.Element {
   );
 }
 
+/**
+ * Local copy — not all published `@medplum/react` bundles export this helper in `.d.ts`.
```
noahsilas, line 270: I'm not entirely sure I understand this comment - is there some build process that isn't including type annotations for https://github.com/medplum/medplum/blob/main/packages/react/src/CodingInput/CodingInput.tsx#L49 ?

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+
+export interface OrderMedicationPageProps {
+  /** When set (e.g. from the patient chart), the form shows and uses this patient immediately. */
+  patient?: Patient;
+  onOrderComplete?: (result: { launchUrl: string; medicationRequestId?: string }) => void;
+}
+
+function getRoutedMedIdFromMedication(m: Medication): number | undefined {
+  const v =
+    m.identifier?.find((i) => i.system === SCRIPTSURE_ROUTED_MED_ID_SYSTEM)?.value ??
```
noahsilas: This (and similar places in this file) could use `@medplum/core`'s `getIdentifier` helper:

```suggestion
    getIdentifier(m, SCRIPTSURE_ROUTED_MED_ID_SYSTEM) ??
```

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+    m.code?.coding?.find((c) => c.system === SCRIPTSURE_ROUTED_MED_ID_SYSTEM)?.code;
```
noahsilas: This (and similar places in this file) could use `@medplum/core`'s `getCodeBySystem`:

```suggestion
    (m.code && getCodeBySystem(m.code, SCRIPTSURE_ROUTED_MED_ID_SYSTEM));
```

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+    const key = match.toLowerCase();
+    const direct = SIG_NUMBER_WORDS[key];
+    if (direct !== undefined) {
+      return String(direct);
+    }
+    const collapsed = key.replaceAll(/[ -]/g, '');
+    const collapsedHit = SIG_NUMBER_WORDS[collapsed];
+    return collapsedHit === undefined ? match : String(collapsedHit);
+  });
+}
```
noahsilas, line 216: This all feels fairly tricky, this might be a good candidate to add some extra test coverage for.

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+const SIG_NUMBER_WORDS_ALTERNATION = Object.keys(SIG_NUMBER_WORDS)
+  .map((w) => w.replaceAll(/[ -]/g, '[ -]'))
+  .join('|');
+const SIG_NUMBER_WORDS_RE = new RegExp(String.raw`\b(?:${SIG_NUMBER_WORDS_ALTERNATION})\b`, 'gi');
+
+/**
+ * Replaces spelled-out English numbers with digits in `text` so the interval
+ * regexes can capture them. Idempotent on text that already uses digits.
+ *
+ * @param text - Lowercased sig line.
```
noahsilas: It looks like `SIG_NUMBER_WORDS_RE` has the ignore case flag set, and then any match found is coerced to lower case, so I think that this doesn't need its input to already be lowercased.

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+    sigLine3: opts.sigLine3,
+    useSubstitution: opts.useSubstitution,
+  };
+}
+
+function medicationToCodeableConcept(m: Medication): MedicationRequest['medicationCodeableConcept'] {
+  return {
+    coding: m.code?.coding,
+    text: m.code?.text,
+  };
```
noahsilas, line 461: Is there a reason not to return `m.code` directly here? Hypothetically that could copy over `extension` attributes in the codeable concept, which might or might not be desirable, I'm not well enough versed in the context here to know. It would also make entries without a code translate more directly as `undefined` instead of `{ coding: undefined, text: undefined }`.

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+  useEffect(() => {
+    if (patientProp) {
+      return undefined;
+    }
+    if (!patientId) {
+      return undefined;
+    }
+    medplum.readResource('Patient', patientId).then(setPatient).catch(showErrorNotification);
+    return undefined;
+  }, [patientProp, patientId, medplum]);
```
noahsilas, line 713: It would be nice if this could reuse the `useResource` hook to get the patient, but looking at it there might be some slightly unnatural contortions to match this calling pattern. Might be something worth thinking about how we can make that hook's signature work for this use case in the future.

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+
+  useEffect(() => {
+    let cancelled = false;
+    const resolveRequester = async (): Promise<void> => {
+      const profile = medplum.getProfile() as Practitioner | PractitionerRole | undefined;
```
noahsilas, line 718: This looks like it is defined such that `PractitionerRole` isn't a valid value: https://github.com/medplum/medplum/blob/main/packages/core/src/utils.ts#L58

Is there a scenario where you are getting PractitionerRole based profiles returned? Should we update that type in core to include it?

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+
+      onOrderComplete?.({ launchUrl: res.launchUrl, medicationRequestId: res.medicationRequestId });
+    } catch (e) {
+      // The bot call (or downstream FHIR write) failed after the draft MR was created.
+      // Clean up the orphan draft so the MedicationsPage doesn't grow stale "draft" rows
+      // for every aborted attempt. We swallow delete failures so the original error
+      // is the one the user sees.
+      if (createdMr?.id) {
+        try {
+          await medplum.deleteResource('MedicationRequest', createdMr.id);
```
noahsilas, line 923: This feels slightly worrisome - it's possible that `orderMedication` creates a request that makes it all the way to the prescriber, but doesn't have its response make it all the way back here for some reason. 

I wonder if this should do something like `updateResource({ ...createdMr, status: 'unknown', statusReason: { coding: { code: 'medplum-provider-received-error' } } })` instead?

I suppose that the deletion here is really a "soft delete" and so this resource could still be looked up if needed?

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+      })
+      .catch((e) => {
+        if (!cancelled) {
+          showErrorNotification(e);
+        }
+      });
+    return (): void => {
+      cancelled = true;
+    };
+  }, [medplum, patient]);
```
noahsilas, line 1356: I wonder if this can use the `useSearch` family of hooks? That does a bunch of this boilerplate, and also provides things like loading state tracking:

```ts
  const [patientCoverages, patientCoveragesLoading] = useSearchResources(
    'Coverage',
    patient ? {
      patient: `Patient/${patient.id},
      status: 'active',
      _count: '50'
    } : undefined,
    { enabled: Boolean(patient) }
)
```

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+      .then((rows) => {
+        if (cancelled) {
+          return;
+        }
+        const filtered: { org: WithId<Organization>; isPrimary: boolean }[] = [];
+        for (const row of rows) {
+          if (row) {
+            filtered.push(row);
+          }
+        }
```
noahsilas: You can use the `isDefined` helper from `@medplum/core` to make this nice and tidy:
```suggestion
        const filtered = rows.filter(isDefined);
```

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -0,0 +1,1631 @@
...
+    },
+    [medplum, patientId]
+  );
+
+  return (
+    <>
+      <Divider label="Optional context" labelPosition="center" />
+      <AsyncAutocomplete<Condition>
+        label="Condition (diagnosis)"
+        placeholder={patient ? 'Search this patient\u2019s conditions' : 'Select patient first'}
```
noahsilas, line 1465: Went down a bit of a rabbit hole looking into this. #TIL that `\u2019` (RIGHT SINGLE QUOTATION MARK) is the preferred code point for a punctuation apostrophe.

packages/react-hooks/src/useMedicationOrderSet/useMedicationOrderSet.ts
```diff
@@ -0,0 +1,171 @@
...
+      const next = await callOperation(req);
+      setUrl(next);
+      return next;
+    } catch (err: unknown) {
+      setError(err);
+      return undefined;
+    } finally {
+      setLoading(false);
+    }
+  }, [callOperation]);
```
noahsilas, line 155: This isn't bound by the same `cancel` semantics as the effect below, so it could end up overriding data in weird ways. How would you feel about adjusting this slightly so we could reuse the same effect? I'm imagining something sort of like this:

```ts
const [refreshId, setRefreshId] = useState(0);
const refresh = useCallback(() => setRefreshId((prev) => prev + 1));
useEffect(() => { ... }, [..., refreshId]);
```

examples/medplum-provider/src/components/meds/quantity-qualifiers.ts
```diff
@@ -0,0 +1,272 @@
...
+    if (synonym === null) {
+      continue;
+    }
+    if (synonym) {
+      for (const pattern of synonym.patterns) {
+        compiled.push({ code, pattern, priority: synonym.priority, nameLength: name.length });
+      }
+      continue;
+    }
+    const escaped = name.replaceAll(REGEX_ESCAPE, String.raw`\$&`);
```
noahsilas: Based on Provider setting its TSConfig target to `"esnext"`, I think you can use `RegExp.escape(row.name.trim())`, which covers some tricky edge cases (https://github.com/tc39/proposal-regex-escaping)

noahsilas (commented):
Flagged some small opportunities for cleanup, but I think things are looking pretty good. There's a lot going on so I'll take one more pass through in a few minutes after I've had a little caffeine before adding a +1. 😅

packages/core/src/medication-order-utils.ts
```diff
@@ -0,0 +1,565 @@
...
+}
+
+/**
+ * Vendor-neutral output from an order-medication bot.
+ */
+export interface MedicationOrderResponse {
+  readonly orderId: number;
+  /**
+   * Vendor-side patient id (numeric in ScriptSure today; other vendors may use a different
+   * shape). Kept as `number` for backwards compatibility with the ScriptSure bot output.
```
noahsilas, line 78: If we think other vendors will use a different shape, maybe we should coerce to `string` now? That gives us a much wider support range in the future.

Alternatively, maybe `unknown`?

examples/medplum-provider/src/components/meds/MedListItem.module.css
oleg-mp: good catch. fixed

examples/medplum-provider/src/components/meds/MedListItem.tsx
oleg-mp: Renamed onItemSelect to getItemUrl with a note that it's called during render so callers should keep the callback pure (and memoize it).

packages/core/src/medication-order-utils.ts
oleg-mp: possibly... it's unlikely going to change anytime soon anyway. DoseSpot doesn't support this functionality at all.

packages/react-hooks/src/useMedicationOrderSet/useMedicationOrderSet.ts
oleg-mp: First attempt was your suggested nonce -> effect shape with a resolver queue (so refresh() could still return Promise<string | undefined>) - that deadlocked under act: await act(async () => await result.current.refresh()) blocks act waiting for a promise that needs act to flush the nonce-driven re-render. Final shape uses a shared runIdRef instead: both refresh() and the effect bump runIdRef.current at fetch start, capture myRunId, and only commit state (and only return the URL) when runIdRef.current === myRunId at resolution. Same "latest run wins" guarantee, no deadlock. Added a test covering the in-flight-refresh-vs-input-change race.

packages/scriptsure-react/src/loadScriptSureQuantityQualifiers.ts
oleg-mp: Rewrote the JSDoc to describe only the implemented canonical shape and link to the OD registration. Added two tests covering empty parameter and mixed-entry inputs

examples/medplum-provider/src/components/meds/quantity-qualifiers.ts
```diff
@@ -123,6 +123,15 @@ const QUALIFIER_SYNONYMS_BY_NAME: Readonly<Record<string, SynonymGroup | null>>
   'international unit': null,
 };
 
+/** Escape a string for use as a literal inside a RegExp pattern. */
+function escapeRegexLiteral(text: string): string {
+  // RegExp.escape (ES2025) lands in Node 23+; CI still runs Node 22.
```
noahsilas, line 132: Arg, good catch. For some reason I thought this one was more widely available than it is. Thanks for making what's happening here explicit!

examples/medplum-provider/src/components/meds/MedicationRequestDetails.tsx
oleg-mp: yeah, agreed. Might be a bit cleaner in it's own PR

.gitignore
oleg-mp: oh wait these are not dropped - these a the same, just generalized and lines rearranged. No-Op

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
oleg-mp: Yep - medplum.getProfile() for our PractitionerRole-based logins returns a PractitionerRole, but ProfileResource (packages/core/src/utils.ts:58) is Patient | Practitioner | RelatedPerson which forces the cast. Updating the union touches every isProfileResource / getDisplayString call site in the repo though, so I'd rather ship that as a separate focused @medplum/core PR. Documenting it here for the record

noahsilas (approved):

PR #9000, opened 2026-04-21
feat(server): inline DocumentReference attachments in Patient $everything

ianplunkett:
## Summary

- Adds `_inlineAttachments=true` query parameter to `GET /fhir/R4/Patient/:id/$everything`. When set, DocumentReference attachments pointing to Binary storage are fetched and embedded as base64 `attachment.data`, with `attachment.url` removed.
- Adds `_maxAttachmentSize` parameter (bytes, default 10 MB) to cap which attachments are inlined — attachments over the limit silently keep their `url`. The stream is read incrementally and abandoned early when the limit is exceeded.
- Exports `normalizeBinaryUrl` from `rewrite.ts` for reuse by the new inlining logic.
- Adds "Inline attachments" checkbox and max-size input (MB) to the Patient Export form under the FHIR Everything format option.


techdavidy:
@claude please review

PR #9004, opened 2026-04-22
Backport R5 `HealthcareService.offeredIn` field to Medplum R4

### LINKED ISSUE #8902: Support HealthcareService.offeredIn
### Issue

In FHIR R4, there is no way to create hierarchy of HealthcareServices, similar to what you can do with Organization's `partOf` field.

Proposing adding `HealthcareService.offeredIn`, which is introduced in FHIR R5, and it's corresponding search param, [offered-in](https://build.fhir.org/healthcareservice-search.html#HealthcareService-offered-in)

noahsilas:
In FHIR R4, there is no way to create hierarchy of HealthcareServices, similar to what you can do with Organization's partOf field.

In FHIR R5, a new field `HealthcareService.offeredIn` is introduced, along with a search parameter `offered-in`.

Backporting this field allows folks to start taking advantage of this structure today.

## Implementation Notes
- The generator package is a bit out of sync; this PR includes an initial commit that runs the generator, a later commit that re-runs the generator after making changes to the definitions, and a final commit that reverts the change from the original generator run. This way the end result only includes the downstream effects from our changes

- Resolves #8902


packages/definitions/src/fhir/r4/profiles-resources.json
```diff
@@ -184013,6 +184013,34 @@
...
+            "key" : "ele-1",
+            "severity" : "error",
+            "human" : "All FHIR elements must have a @value or children",
+            "expression" : "hasValue() or (children().count() > id.count())",
+            "source" : "http://hl7.org/fhir/StructureDefinition/Element"
+          }],
+          "mustSupport" : false,
+          "isModifier" : false,
+          "isSummary" : false
+        },
```
noahsilas, line 184043: This block was pulled from https://hl7.org/fhir/R5/healthcareservice.profile.json.html. Should we customize it to mention that this is non-standard in R4 somewhere?

ThatOneBro (commented):
I think we probably need to bump `Repository.VERSION` for this one

noahsilas:
> I think we probably need to bump `Repository.VERSION` for this one

Discussed a bit in person - my belief is that because this is adding a new optional column that can have no pre-existing entries, this is safe to roll forward without a bump. The new search-param column doesn't need to be backfilled from data, so no reindex is necessary.

I think it is more technically correct to bump the version, but could lead to extra overhead down the line, so I opted to skip it. Let me know if there are edge cases I didn't think of though!

finnbergquist (approved):
Looks good to me

packages/definitions/src/fhir/r4/profiles-resources.json
codyebberson: I'm ok to skip

codyebberson (approved):

packages/definitions/src/fhir/r4/profiles-resources.json
noahsilas: I added a note to `comment` in a69dd19eb4e0ba20db5a7cb04bc097e24a02b7dd

PR #9006, opened 2026-04-22
Fix agent channel reload when endpoint protocol changes

### LINKED ISSUE #6729: Changing agent channel protocol doesn't trigger channel type to change on reload
Changing the protocol of the channel endpoint (channels with same host and port but different protocol) are not correctly detected by the diff. This should result in the previous channel closing and a new channel of the correct type opening instead of a no-op.

degoke:
## What this fixes

This fixes a bug where changing a channel endpoint from one protocol to another on the same channel name was getting treated like a no-op reload.

So if a channel changed from something like `mllp://...` to `tcp://...`, we were reusing the existing channel instance instead of shutting down the old one and starting the right channel type.

## What changed

- if the existing channel type and the new endpoint type are the same, we still use the normal `reloadConfig()` path
- If the type changed, we now stop the old channel, remove it, create the right channel type, and start that instead
- pulled channel creation into a small helper so we don’t repeat the same switch in two places
- added coverage to the existing reload-config test to make sure a protocol change actually replaces the channel instance

## Testing

- ran the agent reload-config test
- verified same-type reloads still work
- verified protocol changes swap the channel instance instead of being ignored
- verified that the old channel gets closed after the replacement

Closes #6729

### DIFF
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -611,7 +611,7 @@ export class App {
       this.channels.delete(definition.name);
 
       let replacementChannel: Channel;
-      
+
       try {
         replacementChannel = this.createChannel(nextType, definition, endpoint);
       } catch (err) {
@@ -624,7 +624,7 @@ export class App {
       return;
     }
 
-    let channel: Channel
+    let channel: Channel;
 
     try {
       const channelType = getChannelType(endpoint);
--- a/packages/agent/src/app.ts
+++ b/packages/agent/src/app.ts
@@ -609,19 +609,6 @@ export class App {
 
       await existingChannel.stop();
       this.channels.delete(definition.name);
-
-      let replacementChannel: Channel;
-
-      try {
-        replacementChannel = this.createChannel(nextType, definition, endpoint);
-      } catch (err) {
-        this.log.error(normalizeErrorString(err));
-        return;
-      }
-
-      await replacementChannel.start();
-      this.channels.set(definition.name, replacementChannel);
-      return;
     }
 
     let channel: Channel;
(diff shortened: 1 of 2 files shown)


packages/agent/src/app.ts
```diff
@@ -596,28 +596,39 @@ export class App {
...
+
+      if (previousType === nextType) {
+        await existingChannel.reloadConfig(definition, endpoint);
+        return;
+      }
+
+      await existingChannel.stop();
+      this.channels.delete(definition.name);
+
+      let replacementChannel: Channel;
```
ThatOneBro: Instead of handling the replacement channel as a separate case, could we remove the rest of this block and just let it fall into the normal `createChannel` case?

ThatOneBro (commented):
Looks great! One minor suggestion about removing a logically duplicate block for the replacement channel, but otherwise looks good for merge after that.

Thank you for your contribution! 🙏

packages/agent/src/app.ts
degoke: Yes we can, that way the block is not duplicated. I'll make an update

degoke:
> Looks great! One minor suggestion about removing a logically duplicate block for the replacement channel, but otherwise looks good for merge after that.
> 
> Thank you for your contribution! 🙏  

Thank you for taking the time to review this! 🙏🏿 Medplum is an excellent project, and I plan to make more contributions as I get up to speed with the codebase.

ThatOneBro:
Thank you @degoke !

We appreciate you taking interest in the project and truly value all our community contributors.

Your latest change looks great!

It looks like your last commit was not signed, do you mind doing:
`git commit --amend -s`
And then force-pushing your branch?

ThatOneBro (approved):
Thanks @degoke, great work 🙏 🚀

PR #9031, opened 2026-04-24
Enforce Project.exportedResourceType on cached reads

rahul1:
## Summary

`Repository.canPerformInteraction` only verified project-compartment membership, so a read that hit the Redis cache could return a resource belonging to a linked project even when its `resourceType` was not in that project's `exportedResourceType`. The SQL path in `addProjectFilters` already applies this filter, so the two paths disagreed and the cache effectively bypassed the export policy.

### Repro

With a target project linking to a source project whose `exportedResourceType: [\"Bot\", \"OperationDefinition\"]`, a `ClientApplication` in the source project could be read from the target project if it was cached (e.g. warmed by a superuser read or by any flow using the owning project's repo).

### Fix

Mirror the carve-outs from `addProjectFilters` inside `canPerformInteraction`:
- primary (first) project is always allowed
- current project is always allowed
- `Project` resource type is always allowed
- linked project with empty/missing `exportedResourceType` is allowed
- otherwise, the resource type must be in the linked project's `exportedResourceType`

## Test plan

- [x] New regression test `Project.exportedResourceType enforced on cached reads` warms the cache via the linked project's repo, then asserts the exported type (`Organization`) reads successfully and the non-exported type (`Patient`) returns `notFound`.
- [x] New sanity test `Project.exportedResourceType allows resources in primary project` confirms `exportedResourceType` on the *primary* project does not block the owner's own reads.
- [x] Verified the new cache-path test fails on pre-fix code and passes with the fix.
- [x] `npm run test:seed` in `packages/server`
- [x] `repo.test.ts` full suite passes (169/169)
- [x] `accesspolicy` suite passes (57/57)

## Notes / follow-ups

- Auth flows (`getClientApplication`, `clientInfoHandler`, `tryLogin`, OAuth endpoints) still use `getGlobalSystemRepo()` to resolve `ClientApplication` and do not consult `exportedResourceType`. That appears intentional for sign-in UX but is worth a separate discussion; not addressed here.

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2259,7 +2259,19 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       }
       // Non-Superusers can only access resources in their Project, with read-only access to linked Projects
       if (readInteractions.includes(interaction)) {
-        if (!this.context.projects?.some((p) => p.id === resource.meta?.project)) {
+        const resourceProject = this.context.projects?.find((p) => p.id === resource.meta?.project);
+        if (!resourceProject) {
+          return undefined;
+        }
+        // Enforce `exportedResourceType` on linked projects; mirrors addProjectFilters.
+        // The primary (owning) and "current" projects always expose all resource types.
+        if (
+          resource.resourceType !== 'Project' &&
+          resourceProject.id !== this.context.projects?.[0]?.id &&
+          resourceProject.id !== this.context.currentProject?.id &&
+          resourceProject.exportedResourceType?.length &&
+          !resourceProject.exportedResourceType.includes(resource.resourceType as ResourceType)
+        ) {
           return undefined;
         }
       } else if (resource.meta?.project !== this.context.projects?.[0]?.id) {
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -2259,7 +2259,19 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+        if (!resourceProject) {
+          return undefined;
+        }
+        // Enforce `exportedResourceType` on linked projects; mirrors addProjectFilters.
+        // The primary (owning) and "current" projects always expose all resource types.
+        if (
+          resource.resourceType !== 'Project' &&
+          resourceProject.id !== this.context.projects?.[0]?.id &&
+          resourceProject.id !== this.context.currentProject?.id &&
+          resourceProject.exportedResourceType?.length &&
```
ThatOneBro, line 2272: Can you add some comments on each line about what each of these lines are doing, eg. why we check projects[0], why if `exportedResourceType` is empty that we skip this, etc.

codyebberson (approved):
Context: https://medplum.slack.com/archives/C051SKJKZ50/p1776988534986979

PR #9035, opened 2026-04-24
Allow passing Reference parameters via GET querystring

noahsilas:
For operations like $find it is nice to be able to support both GET requests (encoding the parameters as query-string entries) and as POST requests (encoding the parameters in the body as a `Parameters` object).

To support our use case of a `Reference` parameter, we need to teach our decoder that such a querystring parameter should be wrapped in a `{reference: %s }` value.

This makes our documentation for this endpoint (which shows POST style access) accurate, without breaking our GET support.


noahsilas:
We're updating the docs to use GET requests instead. We can revisit this if/when we do introduce multi-method endpoints.

PR #9042, opened 2026-04-26
Tightens access controls for linked projects

codyebberson:
Linked Projects were originally used in a limited way (primarily for sharing reference data like `CodeSystem` and `ValueSet`). As usage has expanded, we’re refining how administrative resources are scoped and exposed across project boundaries.

Project admin resources are now scoped strictly to their originating project and are no longer visible via linked projects. This applies to:

* `ProjectMembership`
* `User`
* `UserSecurityRequest`

Notes:
* Linked Projects still provide read-only access to non-admin resources as before
* This aligns behavior with the intended project isolation model
* Projects can optionally configure exported resource types to explicitly control what is shared. We expect to make this required in the future as Linked Projects see broader use

--------

Notes to reviewers:

* The diff is big because I changed the behavior of `createTestProject()`, which is used all over the place
* Before, it made some assumptions and called `new Repository()` directly
* After, it uses `getRepoForLogin()` which makes it more representative of real world usage

### DIFF
--- a/packages/server/src/fhir/accesspolicy.ts
+++ b/packages/server/src/fhir/accesspolicy.ts
@@ -281,33 +281,22 @@ function applyProjectAdminAccessPolicy(
     // then grant limited access to the project admin resource types
     accessPolicy.resource = accessPolicy.resource?.filter((r) => !projectAdminResourceTypes.includes(r.resourceType));
 
-    // Project admins can read all projects that they have access to
-    accessPolicy.resource.push({
-      resourceType: 'Project',
-      readonlyFields: ['features', 'link', 'systemSetting'],
-      hiddenFields: ['superAdmin', 'systemSecret', 'strictMode'],
-      interaction: ['read', 'vread', 'history', 'search'],
-    });
-
-    // But they can only edit their own project
-    accessPolicy.resource.push({
-      resourceType: 'Project',
-      criteria: `Project?_id=${resolveId(membership.project)}`,
-      readonlyFields: ['features', 'link', 'systemSetting'],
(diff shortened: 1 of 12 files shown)


packages/server/src/fhir/accesspolicy.ts
```diff
@@ -300,14 +310,17 @@ function applyProjectAdminAccessPolicy(
     accessPolicy.resource.push(
       {
         resourceType: 'ProjectMembership',
+        criteria: `ProjectMembership?_project=${resolveId(membership.project)}`,
```
codyebberson, line 302: Changes in this file are the main part

packages/server/src/fhir/repo.ts
```diff
@@ -1665,10 +1666,17 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    */
   private getPermittedProjectIds(resourceType: string): string[] | undefined {
     if (!this.context.projects?.length) {
+      // The repository is system-level, so all projects are permitted.
       return undefined;
     }
 
     const projectIds = [this.context.projects[0].id]; // Always include the first project
+
+    if (resourceType !== 'Project' && projectAdminResourceTypes.includes(resourceType as ResourceType)) {
```
codyebberson, line 1675: Technically, this isn't required, and the changes in `accesspolicy.ts` cover it.  I added this as a sanity test, just to confirm.

codyebberson:
@claude please review

reshmakh (approved):

codyebberson:
@claude - feedback fixed.  please review again

PR #9045, opened 2026-04-27
fix: matchesTokenValue |value syntax should match resources with any system (#1241)

### LINKED ISSUE #1241: Mock client doesn't handle `system|value` search syntax

devimano2011:
Fixes #1241

## Summary

When a token search uses `|value` syntax (pipe with empty system prefix), `matchesTokenValue` in `packages/core/src/search/match.ts` was incorrectly requiring that the matching resource have **no system property**.

This caused `searchOne('ServiceRequest', 'identifier=customer/UUID|${id}')` to return `undefined` even when a matching resource existed, while `searchOne('ServiceRequest', 'identifier=${id}')` (no system) worked correctly.

## Root Cause

The `else if (!system)` branch on line 141 had:
```ts
// BEFORE (buggy)
return !resourceValue.system && resourceValue.value?.toLowerCase() === value;
```

Per the [FHIR spec](https://hl7.org/fhir/search.html#token), `|[code]` means match any resource where the code matches **regardless of system**. The `!resourceValue.system` guard was rejecting all resources that had a system set.

## Fix

```ts
// AFTER (fixed)
return resourceValue.value?.toLowerCase() === value;
```

Removed the `!resourceValue.system &&` guard. Now `|value` matches resources with any system (or no system), as long as the value matches.

## Affected file

`packages/core/src/search/match.ts` — 2 lines changed

### DIFF
--- a/packages/core/src/search/match.test.ts
+++ b/packages/core/src/search/match.test.ts
@@ -1130,4 +1130,44 @@ describe('Search matching', () => {
     };
     expect(matchesSearchRequest(resource, search1)).toBe(false);
   });
+
+  test('token: |value syntax matches resource with any system (issue #1241)', () => {
+    // Resource with a system set on the identifier
+    const resource: ServiceRequest = {
+      resourceType: 'ServiceRequest',
+      identifier: [{ system: 'http://customer.example/ids', value: 'abc123' }],
+    };
+
+    // |value (pipe with empty system) MUST match regardless of whether resource has a system
+    // Before fix: !resourceValue.system was checked, so this returned false when system was set
+    expect(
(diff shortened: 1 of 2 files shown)


mattwiller:
@devimano2011 My understanding of the [FHIR spec](https://hl7.org/fhir/R4/search.html#token) is that the `|value` syntax means to match where "the Coding/Identifier has no system property".  I believe the existing behavior is correct

PR #9051, opened 2026-04-27
Range search 

mattwiller:


packages/server/src/fhir/range-column.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+    case 'date':
+    case 'dateTime':
+      return parseDateTimeToRange(typed.value);
+    case 'instant': {
+      const d = new Date(typed.value);
+      return { lInc: true, left: d, right: d, rInc: true };
+    }
+    default:
+      throw new Error('Cannot build datetime range from ' + typed.type);
```
mattlong, line 118: We have reports of `Cannot build datetime range from Timing` having been thrown while handling a GQL query. We should track that down.

mattlong (changes requested):

mattwiller:
@mattlong I added support for `Timing`, in addition to picking up a few other fixes as well (support for `Quantity.comparator`, and missing/empty values in complex types)

mattlong (approved):

PR #9052, opened 2026-04-27
Extract distinct functionality from `repo.ts`

mattlong:
`repo.ts` and its test suite are monolithic files. This extracts several aspects of them into separate source files to make them easier to work with for humans and robots.

This is strictly code re-organization; no intended change in functionality.

I'm deferring resolving the merge conflicts until range search (#9051) lands.

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -1,13 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type {
-  BackgroundJobInteraction,
-  Filter,
-  SearchRequest,
-  TypedValueWithPath,
-  ValidatorOptions,
-  WithId,
-} from '@medplum/core';
+import type { BackgroundJobInteraction, Filter, SearchRequest, WithId } from '@medplum/core';
 import {
   AccessPolicyInteraction,
   accessPolicySupportsInteraction,
@@ -22,7 +15,6 @@ import {
   extractAccountReferences,
   forbidden,
   formatSearchQuery,
-  getReferenceString,
   getStatus,
   gone,
   isGone,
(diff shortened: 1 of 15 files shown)


packages/server/src/fhir/routes.ts
```diff
@@ -385,7 +386,7 @@ function initInternalFhirRouter(): FhirRouter {
   // Validate create resource
   router.add('POST', '/:resourceType/$validate', async (req: FhirRequest) => {
     const ctx = getAuthenticatedContext();
-    await ctx.repo.validateResourceStrictly(req.body);
+    await validateRepositoryResource(ctx.repo, req.body, true);
```
noahsilas: #nit: In places like Github's UI where there isn't great tooling to see the annotations on the parameters, the bare `true` at the end is less obvious. It looks like that just switches to an inner function `validateRepositoryResourceStrictly`, which is also exported. Can that be called directly here?

packages/server/src/fhir/repo.ts
```diff
@@ -1826,7 +1525,9 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    * @param resource - The resource.
    */
   private async writeResource(client: PoolClient, resource: Resource): Promise<void> {
-    await new InsertQuery(resource.resourceType, [this.buildResourceRow(resource)]).mergeOnConflict().execute(client);
+    await new InsertQuery(resource.resourceType, [buildResourceRow(resource, Repository.VERSION)])
```
noahsilas, line 1553: It feels a little strange that `buildResourceRow` takes `Repository.VERSION`. I think that what's happening here is that the things that the version is encoding are more subtly baked in (through things like search parameter definitions), and so don't need to be explicitly passed in.

That does make me wonder if we should hoist this constant into somewhere like `repository/version.ts` so it can be similarly shared in a more implicit manner instead of being explicitly passed this way.

That said, it looks like this is pretty minimal in where it's happening, so I think keeping it as-is is just fine 👍

noahsilas (approved):
I did a scan through and everything seems legit here. I didn't do an exhaustive comparison, as the code is mostly just moving, so if there are areas that you would like special attention paid flag it and I'll re-review with closer attention paid.

mattlong:
@claude do your thing

PR #9053, opened 2026-04-27
fix: close OAuth callback server after extra browser requests post-login (#6144)

### LINKED ISSUE #6144: CLI login not resolving
When running `medplum login`, I have observed two different odd outcomes. The authentication still works, but it is confusing.

1. The process never resolves. But when I kill the process and then run `medplum whoami`, the login was successful.
2. I get this error, 
``` 
[Error]: Unauthorized
    at It.handleUnauthenticated (/opt/homebrew/lib/node_modules/@medplum/cli/node_modules/@medplum/core/dist/cjs/index.cjs:43:12710)
    at It.request (/opt/homebrew/lib/node_modules/@medplum/cli/node_modules/@medplum/core/dist/cjs/index.cjs:43:9170)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  outcome: {
    resourceType: 'OperationOutcome',
    id: 'unauthorized',
    issue: [ [Object] ]
  },
  cause: undefined
```
But again, when I run  run `medplum whoami`, the login appears to be successful.

devimano2011:
Fixes #6144

## Problem
When `medplum login` completes the OAuth code exchange, the `finally` block correctly calls `server.close()` and `process.exit(0)`. However, browsers typically fire additional HTTP requests after the redirect (e.g. `/favicon.ico`), which hit the `else` branch ("Not found") and were silently ignored — keeping the Node.js process alive indefinitely.

This is why users see the process hang even though login was successful.

## Fix
Added an `authComplete` flag (initialized to `false`) that is set to `true` inside the `finally` block after successful auth. The `else` branch now checks this flag and calls `server.close()` + `process.exit(0)` if auth has already completed, allowing the process to cleanly terminate.

## Changes
- `packages/cli/src/auth.ts` — 3 lines added

### DIFF
--- a/packages/cli/src/auth.ts
+++ b/packages/cli/src/auth.ts
@@ -64,6 +64,7 @@ async function startLogin(medplum: MedplumClient, profile: Profile): Promise<voi
 }
 
 async function startWebServer(medplum: MedplumClient): Promise<void> {
+  let authComplete = false;
   const server = createServer(async (req, res) => {
     const url = new URL(req.url as string, 'http://localhost:9615');
     const code = url.searchParams.get('code');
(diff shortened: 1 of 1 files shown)


mattwiller:
@devimano2011 I don't think this code works correctly: it sets a variable in one `if` branch, but checks the same variable in the opposite `else` case where it could never be set.  It also doesn't seem to fully address the linked issue, which references an `Unauthorized` error that simply isn't present here

PR #9054, opened 2026-04-27
fix: add missing Appointment date search parameter (#5614)

### LINKED ISSUE #5614: Appointment-date filter not working as expected
From the community:

> Hi team I have a concert about filtering against the Appointment  resource, I tried to use the following filter.
> 
> ```json
> {
>     "date": "ge2024-11-25"
> }
> ```
> 
> but this is not working so I use this instead
> 
> ```json
> {
>     "_filter": "date ge 2024-11-25"
> }
> ```
> 
> Is that expected for that Resource? since the date  filter is working for other resources.

devimano2011:
Fixes #5614

The `date` search parameter for `Appointment` was missing from `search-parameters-medplum.json`.

This caused `Appointment?date=ge2024-11-25` to return no results, while the equivalent `_filter=date ge 2024-11-25` worked correctly — because `_filter` uses a different code path that does not rely on the SearchParameter definitions.

**Fix:** Added `Appointment-date` SearchParameter with `expression: "Appointment.start"` per the [FHIR R4 spec](https://hl7.org/fhir/R4/appointment.html#search).

**File changed:** `packages/definitions/src/fhir/r4/search-parameters-medplum.json` (+18 lines)

### DIFF
--- a/packages/definitions/src/fhir/r4/search-parameters-medplum.json
+++ b/packages/definitions/src/fhir/r4/search-parameters-medplum.json
@@ -781,6 +781,24 @@
       }
     },
     {
+      "fullUrl": "https://medplum.com/fhir/SearchParameter/Appointment-date",
+      "resource": {
+        "resourceType": "SearchParameter",
+        "id": "Appointment-date",
+        "url": "https://medplum.com/fhir/SearchParameter/Appointment-date",
+        "version": "4.0.1",
+        "name": "date",
+        "status": "draft",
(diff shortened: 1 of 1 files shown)


mattwiller:
@devimano2011 `Appointment?date` is a standard FHIR search parameter, not one added by Medplum.  I don't believe this is the correct fix for the linked 2024 issue, which I think may actually be addressed by some recent work on date search

PR #9056, opened 2026-04-27
test(subscription): add thorough tests around user auth revocation

ThatOneBro:
Adds some tests that sanity check that users cannot receive messages after their `ProjectMembership` is set to `active: false` or if their `Login` is revoked.

### DIFF
--- a/packages/server/src/ws/subscriptions.test.ts
+++ b/packages/server/src/ws/subscriptions.test.ts
@@ -1408,6 +1408,7 @@ describe('WebSocket Subscription', () => {
         project: { features: ['websocket-subscriptions'] },
         withAccessToken: true,
         withRepo: true,
+        withClient: true,
       });
 
       const subscription = await testRepo.createResource<Subscription>({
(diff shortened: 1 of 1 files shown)


mattwiller (approved):

packages/server/src/ws/subscriptions.test.ts
```diff
@@ -1394,6 +1396,221 @@ describe('WebSocket Subscription', () => {
...
+          await Promise.race([sleep(150), messageGuard]);
+        })
+        .close()
+        .expectClosed()
+        .exec(async () => {
+          await sleep(0);
+        });
+    }));
+
+  test('No notification delivered after Login is revoked', () =>
```
mattlong, line 1510: Could you consolidate these into a test.each to avoid the logic diverging over time? Maybe something like:
```ts
test.each([
  ['inactive project memberhsip', (systemRepo, pm, login) => systemRepo.updateResource({...pm, active: false })],
  ['revoked login', (systemRepo, pm, login) => systemRepo.updateResource({...login, revoked: true })],
])('No notification delivered after %s')((_reason, resourceUpdater) =>
```

mattlong (changes requested):

PR #9060, opened 2026-04-28
docs: document authCredentialsMethod: 'header' for OAuth servers rejecting body auth (#5897)

### LINKED ISSUE #5897: Authentication issue with OAuth 2.0  servers that do not support request-body token authentication

devimano2011:
Fixes #5897

## Problem

Some OAuth 2.0 servers reject client credentials in the request body and require `Authorization: Basic` header authentication per RFC 6749 §2.3.1. Users hit "Failed to fetch tokens" with no guidance on how to resolve it.

## What was already in place

`MedplumClient` already supports this via `authCredentialsMethod: 'header'` in `MedplumClientOptions`. The existing `fetchTokens()` implementation correctly strips `client_id`/`client_secret` from the body and sets the header when this option is set. The option simply wasn't documented.

## Changes (docs-only, zero logic changes)

- Improved `authCredentialsMethod` JSDoc with RFC 6749 §2.3.1 reference and clear bullet-point description of `'body'` vs `'header'`
- - Added `@remarks` to `startClientLogin()` explaining when and how to use `authCredentialsMethod: 'header'` with a code example

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -1,4 +1,4 @@
-§// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import type {
   AccessPolicy,
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+§// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import type {
   AccessPolicy,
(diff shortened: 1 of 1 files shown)


mattwiller:
This issue is already resolved, and in my opinion JSDoc isn't really the right place to put this type of documentation.  This kind of example should probably live somewhere in the actual documentation pages instead.

PR #9061, opened 2026-04-28
fix: show extension fields in BackboneElementDisplay (#6999)

### LINKED ISSUE #6999: Extension fields unpopulated in table view and timeline
When viewing an Observation resource with an extension present in the JSON the file, the corresponding column on the Observation table does not include a the value of the extension. Furthermore, the extension is also not shown in the Observation Timeline when it is first added to the Observation. 

Image
Image
Image

devimano2011:
Fixes #6999

## Problem

Extension fields were not shown in the Observation table view or timeline, even when present in the resource JSON.

## Root cause

In `BackboneElementDisplay.tsx`, a local `IGNORED_PROPERTIES` constant filtered out extension keys and suppressed them from display:

```ts
const EXTENSION_KEYS = ['extension', 'modifierExtension'];
const IGNORED_PROPERTIES = DEFAULT_IGNORED_PROPERTIES.filter((prop) => !EXTENSION_KEYS.includes(prop));
// ...
} else if (IGNORED_PROPERTIES.includes(key)) {
  return null;
}
```

## Fix

As noted by @codyebberson in the issue, remove:
- The `IGNORED_PROPERTIES` derived constant
- - The `else if (IGNORED_PROPERTIES.includes(key))` guard
- - The now-unused `DEFAULT_IGNORED_PROPERTIES` import
Extension fields now render normally in table view and timeline.

### DIFF
--- a/packages/react/src/BackboneElementDisplay/BackboneElementDisplay.tsx
+++ b/packages/react/src/BackboneElementDisplay/BackboneElementDisplay.tsx
@@ -5,7 +5,7 @@ import { buildElementsContext, getPathDisplayName, isEmpty, tryGetDataType } fro
 import type { AccessPolicyResource } from '@medplum/fhirtypes';
 import type { JSX } from 'react';
 import { useContext, useMemo } from 'react';
-import { DEFAULT_IGNORED_NON_NESTED_PROPERTIES, DEFAULT_IGNORED_PROPERTIES } from '../constants';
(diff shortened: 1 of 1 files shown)


mattwiller:
@devimano2011 As noted in the [linked issue](#6999):

> We may want to explore if there's a way to systematically display extension data without filling the screen with extension URLs everywhere.

I don't think this can be considered a complete fix: we would need to add design consideration for how to handle the display of resources with many different extensions

PR #9062, opened 2026-04-28
`$resend` with `versionId`

dillonstreator:
# Add `versionId` support to `$resend`

## Why

Subscription handlers sometimes diff `previousVersion` against the current resource to decide whether to fire (lightweight idempotency). When a delivery fails and the resource is later updated, today's `$resend` always replays against the **current** version — handlers then evaluate against drifted state and can incorrectly no-op or fire.

Replay needs the `(version, previousVersion)` pair from the original failure point.

## What

- New optional `versionId` parameter on `POST /[Type]/[id]/$resend`. When set, the server loads that historical version and derives `previousVersion` from history.
- New `Repository.readPreviousVersion(resource)` helper — single-row SQL lookup, replaces ad-hoc paginated history scans.
- When `versionId` is supplied without `interaction`, the server auto-derives it: first version → `create`, otherwise → `update`.

### DIFF
--- a/packages/docs/docs/api/fhir/operations/resend.md
+++ b/packages/docs/docs/api/fhir/operations/resend.md
@@ -20,7 +20,7 @@ When a subscription notification fails or you need to reprocess a resource throu
 - **Debugging Subscriptions**: Test subscription logic during development by manually triggering notifications
 - **Data Synchronization**: Force re-sync of specific resources to external systems after fixing integration issues
 - **Selective Replay**: Re-trigger a specific subscription for a resource instead of all subscriptions
-- **Historical Replay**: Replay a notification against a specific historical `versionId` so subscription handlers that diff `previous` against `current` see the same state as the original event, even if the resource has since been updated
+- **Historical Replay**: Replay a notification against a specific historical `versionId` so subscription handlers that diff `%previous` against `%current` see the same state as the original event, even if the resource has since been updated
 - **Integration Testing**: Verify that subscriptions fire correctly without creating new test data
 
 :::note[Admin Required]
@@ -36,7 +36,7 @@ The operation takes an optional `option` parameter, which is an object containin
 | `verbose`      | Indicates if verbose logging should be enabled.                                                                                                                                                                                                                                                                                      | `boolean`                                | `false`       |
 | `interaction`  | [`Subscriptions`](/docs/api/fhir/resources/subscription) can be configured to trigger only when a resource is created or deleted as opposed to any update. This option allows you to specify which interaction type will be sent.                                                                                                    | `update` &#124; `create` &#124; `delete` | `update`      |
 | `subscription` | A specific [`Subscription`](/docs/api/fhir/resources/subscription) to trigger, formatted as `Subscription/<id>`. If left undefined, all [`Subscriptions`](/docs/api/fhir/resources/subscription) will be triggered.                                                                                                                  | `string`                                 | `undefined`   |
-| `versionId`    | Resend subscriptions for a specific historical version of the resource instead of the current version. When set, the `previous` passed to subscription evaluation is the version immediately prior to this one in history (if any). Useful for replaying a failed notification when the resource has since been updated.     | `string`                                 | `undefined`   |
+| `versionId`    | Resend subscriptions for a specific historical version of the resource instead of the current version. When set, the `%previous` passed to subscription evaluation is the version immediately prior to this one in history (if any). Useful for replaying a failed notification when the resource has since been updated.     | `string`                                 | `undefined`   |
 
 ## Invoke the `$resend` operation
 
--- a/packages/docs/docs/api/fhir/operations/resend.md
+++ b/packages/docs/docs/api/fhir/operations/resend.md
@@ -20,6 +20,7 @@ When a subscription notification fails or you need to reprocess a resource throu
 - **Debugging Subscriptions**: Test subscription logic during development by manually triggering notifications
 - **Data Synchronization**: Force re-sync of specific resources to external systems after fixing integration issues
 - **Selective Replay**: Re-trigger a specific subscription for a resource instead of all subscriptions
+- **Historical Replay**: Replay a notification against a specific historical `versionId` so subscription handlers that diff `previousVersion` against `resource` see the same state as the original event, even if the resource has since been updated
 - **Integration Testing**: Verify that subscriptions fire correctly without creating new test data
 
 :::note[Admin Required]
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -780,6 +787,78 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+  async readPreviousVersion<T extends Resource>(resource: WithId<T>): Promise<T | undefined> {
+    const lastUpdated = resource.meta?.lastUpdated;
+    if (!lastUpdated) {
+      throw new OperationOutcomeError(preconditionFailed);
+    }
+    if (!isUUID(resource.id)) {
+      throw new OperationOutcomeError(notFound);
+    }
+    // Fetch up to 2 candidates so we can detect ties at the same `lastUpdated`.
+    // We use `<=` (rather than `<`) and filter out the supplied versionId
```
mattwiller, line 934: What would happen if there are three or more resources with the same `lastUpdated` timestamp?

packages/server/src/fhir/repo.ts
```diff
@@ -780,6 +787,78 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    // Compare timestamps via numeric ms epoch. `new Date(x).getTime()` accepts
+    // both Date instances (which the pg driver currently returns for
+    // TIMESTAMPTZ columns) and ISO date strings (defensive against future
+    // driver/config changes), and yields a primitive number that is safe to
+    // compare with `===`. An invalid value would produce `NaN`, which would
+    // simply not match — we'd skip the warn but still return the first row,
+    // so this can't break the correctness of the prior-version lookup.
+    if (
+      rows.length > 1 &&
+      new Date(rows[0].lastUpdated).getTime() === new Date(rows[1].lastUpdated).getTime()
```
mattwiller: It might be a little cleaner to move the timestamp equality check into a helper function with the above comment

packages/server/src/fhir/repo.ts
```diff
@@ -780,6 +787,78 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    ) {
+      // Two or more candidate prior versions share the same `lastUpdated`
+      // timestamp, so the chosen prior is nondeterministic. This is rare
+      // (high-throughput writes within the same millisecond, or callers
+      // setting protected meta directly) but worth surfacing.
+      // TODO: emit a metric for this collision rate so we can decide whether
+      // a monotonic ordering column on `_History` is justified.
```
mattwiller: This can be removed if you're not intending to add such a metric

packages/server/src/fhir/routes.test.ts
```diff
@@ -676,25 +676,160 @@ describe('FHIR Routes', () => {
...
+      .send({});
+    expect(res.status).toBe(412);
+
+    // interaction='create' avoids the prior-version requirement
+    const resCreate = await request(app)
+      .post(`/fhir/R4/${getReferenceString(profile)}/$resend`)
+      .set('Authorization', 'Bearer ' + accessToken)
+      .send({ interaction: 'create' });
+    expect(resCreate.status).toBe(200);
+  });
```
mattwiller: It looks like this test may duplicate parts of the one below: can it be removed?

packages/server/src/fhir/repo.ts
dillonstreator: the ambiguity still gets flagged but the scope of the ambiguity is not fully captured
could optionally run a count query to see the scope and add a metric
an analytics dashboard would likely be a more appropriate way to flag these issues proactively

packages/server/src/fhir/repo.ts
dillonstreator: thoughts on centralizing to a helper when there are multiple needs for this type of check?

packages/server/src/fhir/repo.ts
dillonstreator: gotcha can remove for now
was mainly using this as a way to share the idea with y'all

packages/server/src/fhir/routes.test.ts
dillonstreator: yup — duplicate. Removed test, folded unique case into the versionId test

packages/server/src/fhir/repo.ts
dillonstreator: Longer-term, aligning on some disambiguation standard seems reasonable.
FWIW I opened a [thread in chat.fhir.org](https://chat.fhir.org/#narrow/channel/179166-implementers/topic/.25previous.20and.20_history.20under.20lastUpdated.20ties/with/591466882) and https://github.com/medplum/medplum/issues/9112

packages/server/src/fhir/repo.ts
```diff
@@ -780,6 +787,73 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    // compare with `===`. An invalid value would produce `NaN`, which would
+    // simply not match — we'd skip the warn but still return the first row,
+    // so this can't break the correctness of the prior-version lookup.
+    if (rows.length > 1 && new Date(rows[0].lastUpdated).getTime() === new Date(rows[1].lastUpdated).getTime()) {
+      // Two or more candidate prior versions share the same `lastUpdated`
+      // timestamp, so the chosen prior is nondeterministic. This is rare
+      // (high-throughput writes within the same millisecond, or callers
+      // setting protected meta directly) but worth surfacing.
+      getLogger().warn('readPreviousVersion: ambiguous prior version (lastUpdated tie)', {
+        resource: { reference: getReferenceString(resource) },
```
mattwiller: This way is a bit simpler, and produces potentially richer data for logging:
```suggestion
        resource: createReference(resource),
```

packages/server/src/fhir/repo.ts
```diff
@@ -780,6 +787,73 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    // TIMESTAMPTZ columns) and ISO date strings (defensive against future
+    // driver/config changes), and yields a primitive number that is safe to
+    // compare with `===`. An invalid value would produce `NaN`, which would
+    // simply not match — we'd skip the warn but still return the first row,
+    // so this can't break the correctness of the prior-version lookup.
+    if (rows.length > 1 && new Date(rows[0].lastUpdated).getTime() === new Date(rows[1].lastUpdated).getTime()) {
+      // Two or more candidate prior versions share the same `lastUpdated`
+      // timestamp, so the chosen prior is nondeterministic. This is rare
+      // (high-throughput writes within the same millisecond, or callers
+      // setting protected meta directly) but worth surfacing.
```
dillonstreator: There's another case of ambiguity that is not flagged with this check.
If the incoming reference resource and `rows[0]` share the same `lastUpdated`, there is ambiguity between which is `current` and `previous` but that does not get flagged.

dillonstreator: addressed by https://github.com/medplum/medplum/pull/9062/changes/7545c521e8a0a20a2f0d95ad867f4f8d7467b158

packages/docs/docs/api/fhir/operations/resend.md
```diff
@@ -20,6 +20,7 @@ When a subscription notification fails or you need to reprocess a resource throu
 - **Debugging Subscriptions**: Test subscription logic during development by manually triggering notifications
 - **Data Synchronization**: Force re-sync of specific resources to external systems after fixing integration issues
 - **Selective Replay**: Re-trigger a specific subscription for a resource instead of all subscriptions
+- **Historical Replay**: Replay a notification against a specific historical `versionId` so subscription handlers that diff `previousVersion` against `resource` see the same state as the original event, even if the resource has since been updated
```
dillonstreator: Swap `previousResource` and `resource` with `previous` and `current` for more semantically meaningful names

packages/docs/docs/api/fhir/operations/resend.md
```diff
@@ -72,6 +74,23 @@ curl 'https://api.medplum.com/fhir/R4/<resourceType>/<resourceId>/$resend' \
   </TabItem>
 </Tabs>
 
+### Replay a specific historical version
+
+To replay a subscription using a specific past version of the resource (rather than the current version), pass `versionId`. The server will load that exact version, set `previous` to the version immediately preceding it in history, and evaluate subscriptions against that pair.
```
mattwiller: I wasn't immediately sure what `previous` was referring to here, would it be possible to include a link to https://www.medplum.com/docs/subscriptions/subscription-extensions#expression-based-criteria somewhere to clarify?

packages/server/src/fhir/repo.ts
```diff
@@ -780,6 +787,91 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+      ];
+      getLogger().warn('readPreviousVersion: ambiguous prior version (lastUpdated tie)', {
+        resource: { reference: getReferenceString(resource) },
+        versionId: resource.meta?.versionId,
+        lastUpdated,
+        tiedWithIncoming,
+        tiedAmongPriors,
+        tiedVersionIds,
+      });
+    }
```
mattwiller: Could all of this be extracted into a helper function?

dillonstreator: ya seems reasonable since it's quite noisy and breaks the lexical flow of the get previous method - updated here https://github.com/medplum/medplum/pull/9062/changes/6e4a833dc18ad1a827f9168222a9310d8ab08c1c

packages/docs/docs/api/fhir/operations/resend.md
dillonstreator: good call - updated here https://github.com/medplum/medplum/pull/9062/changes/6e4a833dc18ad1a827f9168222a9310d8ab08c1c and also added renamed to `%previous` for consistency with the variable naming

packages/server/src/fhir/repo.ts
dillonstreator: good catch - updated here https://github.com/medplum/medplum/pull/9062/changes/6e4a833dc18ad1a827f9168222a9310d8ab08c1c

mattwiller (approved):
Thanks @dillonstreator! ✅

packages/server/src/fhir/repo.ts
```diff
@@ -765,8 +787,111 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+    // Rare (high-throughput writes within the same millisecond, or callers
+    // setting protected meta directly) but worth surfacing.
+    // Include all versionIds sharing the tied timestamp so the log reflects
+    // the actual ambiguity set, regardless of which case triggered it.
+    const tiedVersionIds = [
+      ...(tiedWithIncoming && resource.meta?.versionId ? [resource.meta.versionId] : []),
+      ...rows.filter((r) => new Date(r.lastUpdated).getTime() === firstMs).map((r) => r.versionId),
+    ];
+    getLogger().warn('readPreviousVersion: ambiguous prior version (lastUpdated tie)', {
+      resource: createReference(resource),
```
mattwiller, line 997: It looks like you might need to import `createReference` above

dillonstreator: `createReference` is imported in this file. This is a bit confusing. Any ideas?

PR #9064, opened 2026-04-28
Add wider retry spacing and jitter to subscription queue

mattwiller:
If a subscription — generally a webhook — fails, the current retry policy exhausts the default of 3 retries in under 8 seconds.  This leaves very little time for the service receiving the webhook to recover, and makes it very likely that all retries will fail as well.  This causes extra load on the database, since each failure usually generates a corresponding AuditEvent.  Spacing the retries out more should help by giving them a better chance to actually succeed, and spreads out any resulting database load


packages/server/src/workers/subscription.ts
```diff
@@ -81,9 +81,9 @@ const REQUEST_TIMEOUT = 120_000; // 120 seconds, 2 mins
 
 /**
  * The upper limit on the number of times a job can be attempted.
- * Using exponential backoff, 19 attempts is about 73 hours (2^18 seconds).
+ * Using exponential backoff, 15 attempts is ~91 hours (20*2^14 seconds).
```
noahsilas: If I'm understanding the jitter parameter correctly, setting it to `0.1` below will also make this take up to 10% longer (with a uniform distribution in that potential delay space, so an expected duration that is 5% longer). On this time frame that does add up to a couple hours, might be worth explicitly noting in the comment.

noahsilas (approved):

packages/server/src/workers/subscription.ts
mattwiller: Per the [BullMQ docs](https://docs.bullmq.io/guide/retrying-failing-jobs#built-in-backoff-strategies), it actually looks like the `jitter` is only applied to __lower__ the delay, so it could be up to 10% shorter overall and would be expected to be ~5% shorter than the listed duration.  Definitely worth noting here!

packages/server/src/workers/subscription.ts
```diff
@@ -81,9 +81,10 @@ const REQUEST_TIMEOUT = 120_000; // 120 seconds, 2 mins
 
 /**
  * The upper limit on the number of times a job can be attempted.
- * Using exponential backoff, 15 attempts is ~91 hours (20*2^14 seconds).
+ * Using exponential backoff capped at 8 hours delay, 19 attempts takes ~72-80 hours (10 attempts to reach ~5 hours delay + 9 attempts spaced 8 hours apart).
+ * Jitter is applied, and may cause the actual delay to be up to 10% shorter than the full calculated delay.
```
noahsilas: I think we lost the `jitter` parameter in this commit - do we need to implement it ourselves as part of the custom backoff strategy?

packages/server/src/workers/subscription.ts
```diff
@@ -883,25 +895,25 @@ async function catchJobError(
...
+  if (job.attemptsMade >= maxJobAttempts) {
+    // All retries exhausted - record as a final failure for auto-disable tracking
+    globalLogger.debug(`Max attempts made for job ${job.id}, subscription: ${subscription.id}`);
+    const triggers = await getSubscriptionAutoDisableTriggers(systemRepo, subscription.meta?.project);
+    const result = await recordSubscriptionFailure(subscription.id, triggers);
+    if (result) {
+      await autoDisableSubscription(systemRepo, result.trigger, subscription, result.failureCount);
+    }
+    // Throw a special error type to force BullMQ to move the job to failed state, instead of retrying again
+    throw new UnrecoverableError(err instanceof Error ? err.message : 'Subscription exhausted available retries');
```
noahsilas, line 908: Discussed in person, recording note here for posterity: previously jobs in this max-attempts failed condition would have entered a success state, now they will be moved into an error state. This will affect how those jobs show in metrics, and if they are retained in Redis (in accordance with settings like `bullmq.removeOnComplete` and `bullmq.removeOnFail`).

This all feels like an increase in correctness 👍

packages/server/src/workers/subscription.ts
```diff
@@ -148,12 +155,18 @@ export const initSubscriptionWorker: WorkerInitializer = (config, options?: Work
...
-      attempts: MAX_JOB_ATTEMPTS, // 1 second * 2^18 = 73 hours
-      backoff: {
-        type: 'exponential',
-        delay: 1000,
+    settings: {
+      backoffStrategy: (attemptsMade: number, type?: string, _err?: Error, _job?: MinimalJob) => {
+        if (type !== 'cappedExponential') {
+          throw new Error('Invalid backoff strategy for subscription queue');
+        }
+        const jitterFactor = 0.9 + 0.2 * Math.random(); // 90–110% of the calculated delay is applied
```
noahsilas, line 163: Nice method of getting a value in the right range! 😁

noahsilas (approved):

Issue #8412, opened 2026-02-11
MedplumClient fails on Node.js 25 due to localStorage behavior change


ianplunkett:
## Description

MedplumClient fails to initialize on Node.js 25 with `TypeError: this.storage.getItem is not a function`.

Node.js 25 changed `localStorage` behavior ([nodejs/node#60303](https://github.com/nodejs/node/issues/60303)). Instead of being `undefined`, it now returns an empty object `{}` without the standard Storage API methods (`getItem`, `setItem`, etc.).

MedplumClient detects that `localStorage` exists and attempts to use it for session storage, but fails because the object lacks the required methods.

## Reproduction

```javascript
import { MedplumClient } from '@medplum/core';

const client = new MedplumClient({
  baseUrl: 'https://api.medplum.com',
});
```

### Node 24 (works)

```
--- localStorage behavior ---
typeof localStorage: undefined
localStorage access threw: ReferenceError - localStorage is not defined

--- MedplumClient initialization ---
MedplumClient created successfully
```

### Node 25 (fails)

```
--- localStorage behavior ---
typeof localStorage: object
localStorage: {}
localStorage.getItem: undefined

--- MedplumClient initialization ---
TypeError: this.storage.getItem is not a function
    at st.getString (.../node_modules/@medplum/core/dist/esm/index.mjs:4:27258)
    at st.getObject (.../node_modules/@medplum/core/dist/esm/index.mjs:4:27413)
    at Yt.getActiveLogin (.../node_modules/@medplum/core/dist/esm/index.mjs:44:6142)
    at Yt.attemptResumeActiveLogin (.../node_modules/@medplum/core/dist/esm/index.mjs:4:47734)
    at new Yt (.../node_modules/@medplum/core/dist/esm/index.mjs:4:46992)
```

## Environment

- @medplum/core: 5.0.14 (latest)
- Node.js: v25.6.1
- OS: macOS

## Suggested Fix

The storage detection logic should verify that `localStorage` has the required methods before using it:

```typescript
// Instead of just checking if localStorage exists:
if (typeof localStorage !== 'undefined') { ... }

// Check if it's a valid Storage implementation:
if (typeof localStorage !== 'undefined' && typeof localStorage.getItem === 'function') { ... }
```

## Workaround

Pass a custom storage implementation to MedplumClient:

```javascript
import { MedplumClient, MemoryStorage } from '@medplum/core';

const client = new MedplumClient({
  baseUrl: 'https://api.medplum.com',
[...]

PR #9067, opened 2026-04-28
chore(node): tighten Node engine ranges and require Node LTS version 22 or 24

The-Alchemist:


scripts/build-agent-sea-linux.sh
```diff
@@ -16,8 +16,8 @@ npm run build
 # If not the expected default version, then warn
 node_version=$(node -v)
 major_version=$(echo "$node_version" | cut -d '.' -f 1)
-if [ "$major_version" != "v20" ]; then
-  echo "WARNING: Expected to be on v20.x.x but on $node_version"
+if [ "$major_version" != "v22" ]; then
+  echo "WARNING: Expected to be on v22.x.x but on $node_version"
```
ThatOneBro: I think all these agent ones should be v24 now

package.json
```diff
@@ -81,6 +81,6 @@
   },
   "packageManager": "npm@10.9.8",
   "engines": {
-    "node": "^22.18.0 || >=24.2.0"
+    "node": ">=22.0.0 <23.0.0 || >=24.0.0 <25.0.0"
```
ThatOneBro: We explicitly can only support v22.18.0 and over, as well as v24.2.0 and over. There was some added features to Node that we need after our move to ESM:
https://github.com/medplum/medplum/pull/7358#:~:text=we%20are%20changing%20the%20minimum%20required%20version%20to%20%22engines%22%3A%20%7B%20%22node%22%3A%20%22%5E22.18.0%20%7C%7C%20%3E%3D24.2.0%22%20%7D%20for%20access%20to%20import.meta.main%2C%20as%20described%20below.

ThatOneBro (changes requested):
See comment about Node engine version changes

package-lock.json
```diff
@@ -53,7 +53,7 @@
         "vitest-websocket-mock": "0.5.0"
       },
       "engines": {
-        "node": "^22.18.0 || >=24.2.0"
+        "node": ">=22.0.0 <23.0.0 || >=24.0.0 <25.0.0"
```
ThatOneBro: I think we might need to fix this by running `npm i`

ThatOneBro (approved):
Looks good after we get agreement in the lockfile

Thanks for cleaning up the repo 🙏

codyebberson (dismissed):
Can we please remove Node 23?  We have had lots of bug reports due to inconsistent behavior on non-LTS versions.  We explicitly do not support them in our version policy: https://www.medplum.com/docs/compliance/versions#nodejs

ThatOneBro:
@codyebberson Good point, we should probably do:
`>=22.18.0 <23.0.0 || >=24.2.0 <25.0.0`?

codyebberson:
Good question... I'm trying to remember where I got that string from.  It was for one of our dependencies....

codyebberson:
Oh yeah, I remember now, it was for `import.meta.main`: https://nodejs.org/api/esm.html#importmetamain

Added in v24.2.0, v22.18.0

I don't think we need to actively exclude 23 and 25?  it doesn't seem like any other packages do that.

The-Alchemist:
@codyebberson :
> I don't think we need to actively exclude 23 and 25? it doesn't seem like any other packages do that.

Yeah, most packages don't seem to exclude non-LTS versions.  But, on the flip side, we don't include non-LTS versions in our CI test matrix and exclude LTS releases from https://www.medplum.com/docs/compliance/versions#nodejs

(brainstorm) So, from a new-user onboarding, we shouldn't exclude non-LTS versions.  But we don't wanna "support" that kind of behavior.

(brainstorm) How about a warning on startup via `import { deprecate } from 'node:util'` or a `DeprecationWarning` to discourage non-LTS versions?  That might be going too far for comfort, or just out of scope for this PR.

Please advise, @codyebberson and @ThatOneBro .

Sounds like we're leaning on `>=22.18.0 <23.0.0 || >=24.2.0 <25.0.0` to exclude v23 and v25?

ThatOneBro:
> I don't think we need to actively exclude 23 and 25? it doesn't seem like any other packages do that.

I remember @ianplunkett did note certain versions of Node 25 were not working due to a backwards-incompat change that they reverted with `localStorage`, maybe we could exclude versions before the revert?

ianplunkett:
> > I don't think we need to actively exclude 23 and 25? it doesn't seem like any other packages do that.
> 
> I remember @ianplunkett did note certain versions of Node 25 were not working due to a backwards-incompat change that they reverted with `localStorage`, maybe we could exclude versions before the revert?

I opened an issue on it - https://github.com/medplum/medplum/issues/8412. Recently, I think @AndyStoneman also ran into an issue trying to run Medplum server with Node 25.

The-Alchemist:
@codyebberson : based on @ianplunkett @ThatOneBro 's comments, seems like it makes sense to exclude v23 and v25, i.e., `>=22.18.0 <23.0.0 || >=24.2.0 <25.0.0`.  Any other issues you see?

ThatOneBro:
Yeah that sounds good to me

PR #9071, opened 2026-04-28
fix(server): intersect valueSet and concept in  (#9057)

### LINKED ISSUE #9057: ValueSet/$expand ignores compose.include.system and concept when include.valueSet is set

compose: {
    include: [
      {
        concept: [
        ],
      }
    ]
  }

iAbhi001:
## Description

This PR fixes a bug in the `ValueSet/$expand` operation where `system` and `concept` filters were ignored if a `valueSet` was present within the same `include` element.

According to the [[FHIR ValueSet Compose definition](https://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.include)](https://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.include), these elements are cumulative (logical AND). If a single `include` contains both a `valueSet` reference and specific `concepts`, the expansion should only return the intersection—the specific concepts that also belong to the referenced value set.

### Changes
- Modified `computeExpansion` in `expand.ts` to remove the `continue` shortcut when `valueSet` is present.
- Implemented a local `includeExpansion` array to allow for intersection logic between `valueSet` codes and specific `concept` lists.
- Added logic to filter by `system` even when a `valueSet` is provided.
- Added a new regression test in `expand.test.ts` to verify the intersection of a parent ValueSet and specific concept codes.

## Related Issues
Closes #9057

## Checklist
- [x] I have read the [[Contributing Guide](https://github.com/medplum/medplum/blob/main/CONTRIBUTING.md)](https://github.com/medplum/medplum/blob/main/CONTRIBUTING.md).
- [x] My code follows the code style of this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed. (Optional: Check this once CI starts)


mattwiller:
Thanks for the PR @iAbhi001 — it looks like you may have forgotten to commit the test case(s) referenced in the PR description.  Would you mind adding those so we can review?  Also, it looks like you'll need to [sign your commits](https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot) to satisfy the DCO check before we can merge it

iAbhi001:
@mattwiller
Please help me by reviewing the PR. thanks!

mattwiller:
@iAbhi001 It looks like [your test is failing to run](https://github.com/medplum/medplum/actions/runs/25119522016/job/73631456713?pr=9071#step:8:203) — please ensure the test case succeeds on your local machine and then push an updated commit

iAbhi001:
@mattwiller Please help me. Sorry!:(

mattwiller:
Hey @iAbhi001 — thanks again for the PR, but it looks like this implementation is still pretty incomplete: we'll definitely need to handle the intersection of `ValueSet.compose.include.valueSet` and `ValueSet.compose.include.filter`.  I suspect that might end up being rather complicated, so it might be simpler for us to tackle this one

mattwiller:
@iAbhi001 We're going to need to handle this as part of a broader effort to make `valueSet` compose with the other filters, not just `concept`.  After some initial investigation, that work looks significantly too tricky and full of performance edge cases for us  to handle it by extending this branch: we'll need to spin up a dedicated workstream to make sure the various issues are resolved carefully.  I'll close this PR for now, since the later work will supersede it

PR #8744, opened 2026-03-22
feat(agent): add Agent/$stats operation for remote stats retrieval

Islam0953:
## Summary
Closes #7221

Add a new `$stats` FHIR operation for the Agent resource that allows fetching runtime statistics remotely via WebSocket.

### Changes
- **`packages/core/src/agent.ts`** — Added `AgentStatsRequest` and `AgentStatsResponse` message interfaces
- **`packages/server/src/fhir/operations/agentstats.ts`** — New operation handler (follows `$fetch-logs` pattern)
- **`packages/server/src/fhir/routes.ts`** — Registered `GET /Agent/$stats` and `GET /Agent/:id/$stats` routes
- **`packages/agent/src/app.ts`** — Handle `agent:stats:request`, collect and return stats (queue depths, HL7 client counts, channel/client stats, heartbeat status)

### Endpoints
- `GET [fhir base]/Agent/$stats` — all agents
- `GET [fhir base]/Agent/[id]/$stats` — specific agent

### Stats returned
- WebSocket/HL7 queue depths
- HL7 client count
- Live status & outstanding heartbeats
- Per-channel and per-client statistics

PR #9078, opened 2026-04-29
feat(agent): add `Agent/$stats` operation

### LINKED ISSUE #7221: Create `Agent/$stats` operation
As a follow-up to #7009, we could add an operation to collect the agent stats remotely which would enable things like an agent "cluster" dashboard with the health of each agent

ThatOneBro:
Fixes #7221

Replaces #8744


packages/app/src/resource/ToolsPage.test.tsx
```diff
@@ -167,7 +167,7 @@ describe('ToolsPage', () => {
       fireEvent.click(screen.getByLabelText('Ping'));
     });
 
-    await expect(screen.findByText('statistics', { exact: false })).resolves.toBeInTheDocument();
+    await expect(screen.findByText('ping statistics', { exact: false })).resolves.toBeInTheDocument();
```
ThatOneBro, line 170: Previously "statistics" only appeared in the ping results, however now it's always in the body since the stats section also contains it

codyebberson:
@claude please review

codyebberson:
> `channel.stats?.getStats() can be undefined if channel.stats is not initialized, `

It's not obvious to me if this is true.  If so, may want to fix?  Otherwise everything else looks stylistic.

packages/app/src/resource/ToolsPage.tsx
```diff
@@ -92,22 +192,24 @@ export function ToolsPage(): JSX.Element | null {
...
 
   useEffect(() => {
-    if (loadingStatus || reloadingConfig || upgrading || pinging) {
+    if (loadingStatus || reloadingConfig || upgrading || pinging || fetchingStats) {
       setWorking(true);
       return;
     }
     setWorking(false);
-  }, [loadingStatus, reloadingConfig, upgrading, pinging]);
+  }, [loadingStatus, reloadingConfig, upgrading, pinging, fetchingStats]);
```
noahsilas: #nit: This state+effect looks like it can be replaced with an inline variable (any of these other pieces of state changing will trigger a re-render anyways):

```
  const working = loadingStatus || reloadingConfig || upgrading || pinging || fetchingStats;
```

packages/app/src/resource/ToolsPage.tsx
```diff
@@ -84,6 +85,105 @@ function UpgradeConfirmContent(props: UpgradeConfirmContentProps): JSX.Element {
...
+] as const satisfies readonly (keyof AgentStats)[];
+
+function formatStatValue(value: Record<string, unknown> | boolean | number | string): string {
+  if (value === null || value === undefined) {
+    return '';
+  }
+  if (typeof value === 'object') {
+    return JSON.stringify(value);
+  }
+  return value.toString();
```
noahsilas, line 108: #nit: For potentially large numbers it would be nice to use `toLocaleString()`

packages/server/src/fhir/operations/agentstats.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  let port: number;
+  let accessToken: string;
+
+  beforeAll(async () => {
+    const config = await loadTestConfig();
+    server = await initApp(app, config);
+    accessToken = await initTestAuth({ membership: { admin: true } });
+
+    await new Promise<void>((resolve) => {
+      server.listen(0, 'localhost', 8514, () => {
```
noahsilas, line 47: 8514 is sort of a surprising backlog size value - is there something special happening here?

packages/server/src/fhir/operations/agentstats.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param req - The FHIR request.
+ * @returns The FHIR response.
+ */
+export async function agentStatsHandler(req: FhirRequest): Promise<FhirResponse> {
+  return handleBulkAgentOperation(req, async (agent) => fetchStats(agent));
+}
+
+async function fetchStats(agent: WithId<Agent>): Promise<FhirResponse> {
+  return sendAndHandleAgentRequest<AgentStatsResponse>(agent, { type: 'agent:stats:request' }, 'agent:stats:response', {
```
noahsilas, line 37: For running agents that haven't upgraded to support this stats request yet, it seems like the agent will log an error locally and not process the message, and then this will timeout after 5s. Is that right?

Do we know the version of code that the agent is running? Maybe we could fail early if we know that the agent doesn't support this to avoid having these operations run long?

ThatOneBro: Good question, today we already have this for a few message types on older agents. See: https://github.com/medplum/medplum/issues/9118

packages/server/src/fhir/operations/agentstats.test.ts
ThatOneBro: Not really sure where this initially got introduced, this is mainly copy-pasta boilerplate from tests across the entire codebase... 🤔

noahsilas (approved):

PR #9080, opened 2026-04-29
Document login rate limit category

ianplunkett:
## Summary

- The rate limiter has three categories (login, auth, default) but the docs only described two
- Adds the login category row (5 req/IP/min) covering `/auth/login`, `/auth/newuser`, and `/auth/newproject`
- Clarifies that the auth category covers _other_ `/auth/*` and `/oauth2/*` endpoints, and that `/auth/me` falls under default

## Test plan

- [x] Verify table renders correctly in the docs site


maddyli (approved):
nice!

PR #9081, opened 2026-04-29
Revoke active sessions on password reset

### LINKED ISSUE #9038: Revoke all other sessions on password change

mattwiller:
Resolves #9038

### DIFF
--- a/packages/server/src/auth/setpassword.ts
+++ b/packages/server/src/auth/setpassword.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { allOk, badRequest } from '@medplum/core';
-import type { User, UserSecurityRequest } from '@medplum/fhirtypes';
+import { allOk, badRequest, EMPTY, getReferenceString, Operator } from '@medplum/core';
+import type { Login, User, UserSecurityRequest } from '@medplum/fhirtypes';
 import type { Request, Response } from 'express';
 import { body } from 'express-validator';
 import { pwnedPassword } from 'hibp';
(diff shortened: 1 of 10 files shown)


packages/server/src/auth/setpassword.ts
```diff
@@ -47,6 +47,16 @@ export async function setPasswordHandler(req: Request, res: Response): Promise<v
...
   await setPassword({ ...user, emailVerified: true }, req.body.password);
   await systemRepo.updateResource<typeof securityRequest>({ ...securityRequest, used: true });
+
+  const activeSessions = await systemRepo.search<Login>({
+    resourceType: 'Login',
+    filters: [{ code: 'user', operator: Operator.EQUALS, value: getReferenceString(user) }],
+  });
+  for (const entry of activeSessions.entry ?? EMPTY) {
+    const login = entry.resource as Login;
+    await systemRepo.updateResource({ ...login, revoked: true });
```
ThatOneBro, line 64: Do we want to make this a patch just to make it a little safer?

ThatOneBro (commented):
I think there are a few more places where we would expect this login revocation to happen. Namely: `/auth/resetpassword`, and `/auth/changepassword`

packages/server/src/auth/setpassword.ts
mattwiller: I'm not too worried about clobbering anything on the Login, since we're revoking it anyways

mattwiller:
@ThatOneBro Good call on `/changepassword`: I combined the logic for both code paths to ensure they stay in sync.  I think `/resetpassword` shouldn't revoke the logins yet, since anyone can request the password reset email and that could be a DoS attack vector

ThatOneBro:
@mattwiller That's a fair point. Maybe we should make the revoke happen at the point that you actually do set a new password though?

Scenario: Someone compromises your account, changes your password. You reset your password and login but the other user is still logged in until you manually revoke those sessions somehow

EDIT: Sorry, this was a defect in my understanding of the flow. LGTM

ThatOneBro (approved):

PR #9082, opened 2026-04-29
Consistent Repository connection state tracking and sharing

### LINKED ISSUE #9029: Derived `Repository` instances can share a transactional client without sharing transaction state
# Repository instances can share a transactional client without sharing transaction state

## Summary

`Repository.getSystemRepo()` and `Repository.clone()` can reuse the same `PoolClient` as the parent repository, but transaction state is still stored per `Repository` instance.

That means two `Repository` objects can participate in the same underlying Postgres transaction while disagreeing about whether they are "in a transaction".

## Affected code

- `src/fhir/repo.ts`
  - `constructor(...)`
  - `clone(...)`
  - `getSystemRepo(...)`
  - `withTransaction(...)`
  - `beginTransaction(...)`
  - `ensureInTransaction(...)`
  - `preCommit(...)`
  - `postCommit(...)`
  - cache helpers such as `getCacheEntry(...)`, `getCacheEntries(...)`, `setCacheEntry(...)`, `deleteCacheEntry(...)`, `deleteCacheEntries(...)`

## Problem

`Repository` currently mixes two different concepts:

1. Repository-instance state
   - auth context
   - mode
   - ownership of the JS object
2. Transaction state
   - whether the shared DB client is currently inside `BEGIN ... COMMIT/ROLLBACK`
   - savepoint depth
   - pre/post-commit callback queues
   - transaction-scoped side effects and bookkeeping

Those should not be the same thing.

Today, transaction state is effectively tracked per repository instance via fields like:

- `transactionDepth`
- `preCommitCallbacks`
- `postCommitCallbacks`
- `callbackStack`

But derived repos can reuse the same `PoolClient`:

- `clone(...)` passes `conn ?? this.conn`
- `getSystemRepo(...)` passes `this.conn`

So a derived repo can share the active transactional client without sharing the transaction state that governs how that client must be used.

## Why this is wrong

If repo A has already issued `BEGIN` on a `PoolClient`, then repo B using that same client is also inside that transaction, regardless of what `repoB.transactionDepth` says.

The current model allows this situation:

```ts
await repo.withTransaction(async () => {
  const systemRepo = repo.getSystemRepo();

  // Uses the same PoolClient / same SQL transaction.
  await systemRepo.readResource('Project', projectId);

  // But systemRepo.transactionDepth is still 0.
});
```

At that point, `systemRepo` is participating in the live SQL transaction, but its local state says "not in transaction".

### 1. Cache behavior becomes incorrect

Historically, cache guards used `this.transactionDepth`.

That means a derived repo inside a parent transaction could still:

- read from Redis mid-transaction
- write to Redis before the outer transaction commits
- delete cache entries before commit

This can leak uncommitted state across the caching layer or read stale state during an in-flight transaction.

### 2. `postCommit` / `preCommit` semantics are wrong across derived repos

`preCommit(...)` and `postCommit(...)` currently key off the local repository instance:

- if `this.transactionDepth > 0`, queue the callback
- otherwise invoke it immediately

So a derived repo on the same live transactional client can run a supposed "post-commit" callback immediately, even though the outer transaction has not committed yet.

That can affect:

- background jobs
- binary handling side effects
- audit/event emission behavior
- any future side effect that assumes commit has already happened

### 3. `ensureInTransaction(...)` can make the wrong decision

`ensureInTransaction(...)` currently checks `this.transactionDepth`.

If a derived repo is already on a live transactional client but has `transactionDepth === 0`, it will call `withTransaction(...)` again instead of reusing the current transaction context.

### 4. `withTransaction(...)` can issue `BEGIN` on an already-active client

`beginTransaction(...)` does:

```ts
this.transactionDepth++;
  await conn.query('BEGIN ISOLATION LEVEL ' + isolationLevel);
} else {
  await conn.query('SAVEPOINT sp' + this.transactionDepth);
}
```

For a derived repo that shares an already-active `PoolClient`, `transactionDepth` still starts at `0`, so the first call attempts another `BEGIN` rather than creating a savepoint or joining the existing transaction context.

## Reproduction sketch

Minimal example:

```ts
await repo.withTransaction(async () => {
  const systemRepo = repo.getSystemRepo();

  // Same SQL transaction, different Repository instance.
  await systemRepo.postCommit(async () => {
    // Expected: run after outer commit
    // Actual with per-instance state: may run immediately
  });

  await systemRepo.readResource('Project', projectId);
});
```

Another problematic shape:

```ts
await repo.withTransaction(async () => {
  const systemRepo = repo.getSystemRepo();

  await systemRepo.ensureInTransaction(async (client) => {
    // Expected: reuse current transaction
    // Actual with per-instance state: may attempt nested BEGIN semantics incorrectly
  });
});
```

## Expected behavior

Any `Repository` instances that share the same live transactional `PoolClient` should also share the same transaction context.

That shared context should define:

- whether a transaction is active
- nesting/savepoint depth
- pre-commit callbacks
- post-commit callbacks
- transaction-scoped bookkeeping

Derived repos should not have to guess transaction state from:

- their own local `transactionDepth`
- a shared telemetry object
- whether `this.conn` happens to be set

## Actual behavior

Transaction participation is currently determined by the underlying shared client, while transaction control flow is determined by per-instance fields.

That mismatch means:

- one repo can be in the transaction from Postgres's perspective
- while the same repo instance believes it is not in a transaction

## Proposed direction

Introduce an explicit shared `TransactionContext` (name flexible) and keep it separate from request/repo state.

### Suggested shape

`TransactionContext` should own:

- the shared `PoolClient`
- transaction nesting depth / savepoint depth
- pre-commit callback queue(s)
- post-commit callback queue(s)
- transaction-scoped access aggregation

`Repository` should own:

- auth/access context
- mode
- non-transaction request behavior
- a reference to the current `TransactionContext`, if any

### Suggested rules

1. `repo.withTransaction(...)` creates or joins a shared `TransactionContext`.
2. `repo.getSystemRepo()` and `repo.clone()` should carry forward the active `TransactionContext`, not just the raw `PoolClient`.
3. Cache deferral and callback behavior should consult the shared transaction context, not a per-instance counter.
4. Request-level or logging-level aggregation should be separate from transaction lifecycle.

## Acceptance criteria

- A repo derived inside an active transaction behaves as "in transaction" for cache reads/writes and callback deferral.
- `preCommit(...)` and `postCommit(...)` registered from any participating repo run exactly once in the correct phase of the outer transaction.
- `ensureInTransaction(...)` and nested `withTransaction(...)` on derived repos reuse the active shared context instead of issuing a fresh top-level `BEGIN`.
- Transaction retries, rollback handling, and savepoint behavior remain correct across `repo`, `repo.getSystemRepo()`, and `repo.clone()`.

mattlong:
Fixes #9029


The-Alchemist (approved):

packages/server/src/fhir/repo.ts
```diff
@@ -379,13 +384,18 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
    * @returns a SystemRepository for the same shard as this repository.
    */
   getSystemRepo(): SystemRepository {
-    return createSystemRepository(this.shardId, this.conn, {
+    const contextDefaults = {
       skipBackgroundJobs: this.context.skipBackgroundJobs,
-    });
+    };
+    if (this.connection.hasConnection()) {
```
mattlong, line 390: I'm not sure if this branch makes sense; always sharing `this.connection` when elevating to a system repo is probably preferable but would be a change in behavior for a widespread and sensitive code path.

mattlong: To clarify, as the code is now with the `if` matches the current/existing behavior of `Repository`

codyebberson:
@claude  - can you please re-review this PR?

We received a community report that it caused a regression (full report pasted below).  Do you see anything in this PR that could cause that behavior?

--------

## Summary

Medplum server v5.1.11 crash-loops due to a race condition in the shared `RepositoryConnection` introduced by PR #9082 ("Consistent Repository connection state tracking and sharing"). The server crashes within ~3 minutes of startup, regardless of external load.

## Environment

- **Medplum version**: `medplum/medplum-server:5.1.11`
- **Infrastructure**: AWS ECS Fargate
- **Database**: Aurora PostgreSQL 16.13 (RDS)
- **Previous stable version**: `5.1.10` (no crashes)

## Reproduction

1. Deploy `medplum/medplum-server:5.1.11` with default configuration
2. Allow WebSocket subscription clients to connect
3. Server crashes within ~3 minutes with `Uncaught exception: terminating connection due to idle-in-transaction timeout`

No external load is required — the crash occurs with only health check traffic and WebSocket subscription bindings.

## Evidence

### 1. PostgreSQL server logs (RDS) — at the exact crash timestamp

```
2026-05-19 10:39:35 UTC — WARNING: there is already a transaction in progress (port 48730)
2026-05-19 10:39:35 UTC — WARNING: there is no transaction in progress (port 48754)
2026-05-19 10:39:35 UTC — WARNING: there is already a transaction in progress (port 48738)
2026-05-19 10:39:35 UTC — WARNING: there is no transaction in progress (port 48754)
... (9+ connections affected simultaneously)
```

30 seconds later:
```
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48736)
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48732)
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48730)
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48740)
```

### 2. Node.js runtime warning (same timestamp)

```
(node:1) DeprecationWarning: Calling client.query() when the client is already executing a query is deprecated and will be removed in pg@9.0.
```

### 3. Medplum application error

```
ERROR: Database error: "Not in transaction"
  at _RepositoryConnection.assertInTransaction (dist/index.js:32202)
  at _RepositoryConnection.commitTransaction (dist/index.js:32139)
  at _RepositoryConnection.withTransaction (dist/index.js:32082)
  at _Repository.writeToDatabase (dist/index.js:39453)
  at _Repository.handleStorage (dist/index.js:39428)
  at _Repository.updateResourceImpl (dist/index.js:39372)
```

### 4. Final crash

```
ERROR: Uncaught exception thrown: "terminating connection due to idle-in-transaction timeout"
  at parseErrorMessage (pg-protocol/dist/parser.js:305)
```

## Root Cause Analysis

PR #9082 moved `transactionDepth` from per-`Repository`-instance into a shared `RepositoryConnection` object. When multiple concurrent async operations access the same `RepositoryConnection` (e.g., via `repo.getSystemRepo()` or internal server processes), the shared `transactionDepth` counter becomes corrupted:

1. **Operation A** calls `beginTransaction()` → sets `transactionDepth = 1`, issues `BEGIN`
2. **Operation B** (same `RepositoryConnection`) calls `beginTransaction()` → issues `BEGIN` on the **same PoolClient** that already has a transaction → PostgreSQL warns: "there is already a transaction in progress"
3. One operation's error path calls `abortTransaction()` → resets `transactionDepth = 0`
4. The other operation calls `commitTransaction()` → `assertInTransaction()` throws "Not in transaction" (because depth is now 0)
5. The orphaned `BEGIN` (from the duplicate) leaves connections idle in transaction state
6. After 30s, PostgreSQL kills these connections → uncaught exception → process crash

### Why v5.1.10 didn't crash

In v5.1.10, `transactionDepth` was a **per-Repository-instance** field. Even when `getSystemRepo()` shared the same `PoolClient`, each instance maintained its own counter, so they couldn't corrupt each other's transaction tracking.

### Supporting context

- PR #4776 ("each repo can only do one query at a time") explored guarding against this exact class of bug in July 2024 but was never merged.
- The `pool.on('error')` handler from PR #8987 properly handles dead connections, but cannot prevent the crash because the `transactionDepth` corruption causes an uncaught `OperationOutcomeError` to propagate out of the Express error handler before the idle timeout fires.

## Timeline (from a single server instance)

| Time (UTC) | Event |
|---|---|
| 10:36:35 | Server starts (v5.1.11) |
| 10:36:52 | WebSocket subscription connections established |
| 10:39:35 | pg deprecation warning + "Not in transaction" errors cascade |
| 10:39:35 | PostgreSQL: "already a transaction in progress" / "no transaction in progress" |
| 10:40:05 | PostgreSQL FATAL: kills 4 idle-in-transaction connections |
| 10:40:05 | Server crash: Uncaught exception |

## Expected behavior

Concurrent operations that share a `RepositoryConnection` should not corrupt each other's transaction state. Either:
- Serialize access to the shared connection (mutex/queue), or
- Prevent concurrent `beginTransaction()` on a connection that already has an active transaction (fail fast with a clear error instead of issuing a duplicate `BEGIN`)

## Workaround

Downgrade to `medplum/medplum-server:5.1.10` which contains the dead connection pool fix (PR #8987) but does not include the `RepositoryConnection` sharing change (PR #9082).

PR #9088, opened 2026-04-30
Scheduling error path improvements

noahsilas:
As I've been working on these Scheduling APIs, there have been some ongoing challenges related to giving good feedback. A main issue here is that the different API shapes mean that things like SchedulingParameters can come from all kinds of sources.

The main idea in this commit is that we can annotate our in-memory objects with a special symbol key. The insight here is that Symbol keys are not enumerable in regular iteration, and so these annotations will not be serialized to JSON or the database.

One place that this showed immediate potential is in how we access extension values; when dealing with our complex extension, core's `getExtension` helper is of limited utility:
- It only finds the first extension with a given URL, but we may have many
- It doesn't include path information about where the returned extension came from (making creating helpful error messages hard)

I've added a `getExtensions()` helper (note the plural) that can extract extensions with a specific URL, and annotates the results with the real index in the extensions array. We can then use that in our scheduling parameters helpers.

One final difficulty that arose from that: I had made a simplifying assumption originally that the data in these extensions would be validated by having a FHIR Profile applied that would restrict the values. In the original implementation then I had applied a cast to the extension values, like `... as SchedulingParametersExtension`. I was then relying on an explicit discriminated union type of that shape to refine the types of the sub-extensions.

As we haven't shipped that profile definition, these types can be violated, which can cause confusing errors.

The new `getExtensions()` helper is typed to return `Extension[]`, without trying to narrow a specialized extension type. This means that before accessing the attributes, we add some explicit assertions that the extensions have the right shape. Those assertions can be quite tidy because the extensions have been annotated with path information and so can generate helpful error messages!


noahsilas:
@claude Please review

noahsilas:
@claude review the latest updates

packages/server/src/util/withpath.ts
```diff
@@ -0,0 +1,77 @@
...
+
+// Gets the path from an annotated object
+export function getPath(obj: PathedObject): string;
+export function getPath(obj: object): undefined;
+export function getPath(obj: { [PATH_SYMBOL]?: string }): string | undefined {
+  if (PATH_SYMBOL in obj) {
+    return obj[PATH_SYMBOL];
+  }
+  return undefined;
+}
```
noahsilas: This is sort of the core idea of this PR - that we can annotate objects with a special symbol to track how they should be represented when an error message needs to be emitted referencing them.

This seemed cleaner than trying to do some kind of context tracking, as each object can carry the information with it, and the type system can be aware of which objects have been decorated in this way.

The rest of the changes are trying to take this idea and apply it to the complex parameters that can be involved in the Scheduling APIs and updating helper functions to use this information.

packages/server/src/util/extension.ts
```diff
@@ -0,0 +1,77 @@
...
+    throw new OperationOutcomeError(badRequest('Extension valueCode has wrong type', getPath(extension)));
+  }
+}
+
+export function assertExtensionDuration(
+  extension: WithPath<Extension>
+): asserts extension is WithPath<Extension> & { valueDuration: Duration } {
+  if (!('valueDuration' in extension)) {
+    throw new OperationOutcomeError(badRequest('Extension valueDuration missing', getPath(extension)));
+  }
```
mattlong, line 112: This one doesn't check if `extension.valueDuration` is actually the correct type. If intentional, could you add a comment saying as much?

packages/server/src/util/withpath.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Utilities for tracking paths through manipulations
+ *
+ * Path information is annotated onto objects with a unique symbol
+ * key, allowing us to read/write path information into the JS objects
+ * in a way that is not revealed during normal iteration.
+ */
+const PATH_SYMBOL = Symbol('MedplumPath');
+
+type PathedObject = { [PATH_SYMBOL]: string };
```
mattlong: Should this be an interface? My understanding is that interfaces are generally more performant for the type checker.

packages/server/src/util/withpath.ts
```diff
@@ -0,0 +1,77 @@
...
+): WithPath<T>[] {
+  return flatMapFilter(objects, (obj, idx) => {
+    if (predicate(obj)) {
+      return withPath(obj, `${pathPrefix}[${idx}]`);
+    }
+    return undefined;
+  });
+}
+
+// Copies paths from a source array to objects in the destinations array
```
mattlong: Could you turn this comment into a docstring and expand on the description a bit? Without seeing it used somewhere, it's not immediately obvious when I'd want to use it.

Docstrings on all the other functions `filterWithPaths`, `withPaths`, etc. would be appreciated 🙏

packages/server/src/util/withpath.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    expect(pathed.a).toBe(1);
+  });
+
+  test('preserves all existing properties', () => {
+    const obj = { x: 'hello', y: 42, z: true };
+    const pathed = withPath(obj, 'bar');
+    expect(pathed).toMatchObject({ x: 'hello', y: 42, z: true });
+  });
+
+  test('path is not visible in normal enumeration', () => {
```
mattlong, line 27: Could you add a test checking that `stringify` from `@medplum/core` on a full resource with nested extensions excludes paths?

packages/server/src/fhir/operations/book.test.ts
```diff
@@ -428,6 +430,10 @@ describe('Appointment/$book', () => {
         details: {
           text: 'Mismatched service types',
         },
+        expression: [
+          'Parameters.slot[0].serviceType[0].extension[0].valueReference.reference',
```
mattlong, line 434: 🔥

mattlong (changes requested):
Cool concept! As a followup, we should evaluate if this could be combined/merged with/supercede the logic behind `TypedValueWithPath` in `@medplum/core`.

My request for docstrings in `withpath.ts` extends (hah) to all the other exported util/helper functions 🙏

packages/server/src/util/withpath.ts
noahsilas: Good question. I think I mostly default to `type` over `interface` because there are some things that `type` can express that `interface` can't, but I don't know how these impact performance. To stay consistent with other places in the codebase I'm happy to update this to be an interface 👍

packages/server/src/util/extension.ts
noahsilas: Oops. `valueDuration` is more complex to type check, so I skipped it when on my first pass but forgot to leave a TODO for myself to revisit... Thanks for the spot!

noahsilas: I had thought I would need to more fully validate the types of the `Duration` object, but my current plan is to validate only that `typeof valueDuration === 'object'`. I think that this is a good middle ground between taking anything and doing the deep enforcement of object properties.

noahsilas:
> Cool concept! As a followup, we should evaluate if this could be combined/merged with/supercede the logic behind `TypedValueWithPath` in `@medplum/core`.

Definitely worth exploring! My thought was to let this bake a little in `server` before pulling it into `core`, but I'm definitely planning on hoisting it up there later and taking advantage of this style in some more places. 😁 

I've addressed the rest of your feedback in commit df4e2b8.

mattlong (approved):

PR #9092, opened 2026-04-30
Provider: Make `<AlphaBanner>` dark-mode friendly

noahsilas:
Tweaking the colors a bit here makes this much nicer when someone is using the dark theme.

## Before
Screenshot 2026-04-30 at 11 44 03 AM

## After
Screenshot 2026-04-30 at 11 43 39 AM

### DIFF
--- a/examples/medplum-provider/src/components/AlphaBanner.tsx
+++ b/examples/medplum-provider/src/components/AlphaBanner.tsx
@@ -2,20 +2,20 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { BoxProps } from '@mantine/core';
 import { Box, Group, Pill } from '@mantine/core';
+import cx from 'clsx';
 import type { JSX } from 'react';
+import classes from './AlphaBanner.module.css';
 
 interface AlphaBannerProps extends BoxProps {
   children: React.ReactNode;
 }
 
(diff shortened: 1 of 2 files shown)


examples/medplum-provider/src/components/AlphaBanner.tsx
```diff
@@ -2,20 +2,20 @@
...
 interface AlphaBannerProps extends BoxProps {
   children: React.ReactNode;
 }
 
 export function AlphaBanner(props: AlphaBannerProps): JSX.Element {
-  const { children, ...boxProps } = props;
+  const { children, className, ...boxProps } = props;
   return (
-    <Box bg="violet.0" p="sm" {...boxProps}>
+    <Box p="sm" {...boxProps} className={cx(classes.alphaBanner, className)}>
```
techdavidy, line 16: we wouldn't need className here since it would be applied to <AlphaBanner> ?

codyebberson (approved):

examples/medplum-provider/src/components/AlphaBanner.tsx
noahsilas: The idea here is that if the caller is applying a `className` prop, we add our own to whatever they supplied instead of overwriting it if it had been spread in via `{...boxProps}` before our attribute is applied.

PR #9093, opened 2026-04-30
Deploy graphiql in staging.yml

mattlong:


codyebberson (approved):

packages/graphiql/.env.defaults
```diff
@@ -1,2 +1,3 @@
 MEDPLUM_BASE_URL=http://localhost:8103/
 MEDPLUM_INTROSPECTION_URL=https://graphiql.medplum.com/schema/schema-latest.json
+STAGING_MEDPLUM_INTROSPECTION_URL=https://graphiql.staging.medplum.dev/schema/schema-latest.json
```
mattlong: hmm, I don't think this is actually going to do anything

PR #9098, opened 2026-05-01
fix: preserve agent log files during uninstall

devimano2011:
Delete specific agent files during uninstallation while preserving log files. Ensure the installation directory is removed only if empty.


ThatOneBro:
Fixed in #9693

PR #9107, opened 2026-05-01
fix(core): "graph[vertex] is not iterable" when importing FHIR bundles without ID

The-Alchemist:
occurs when importing first bundle from the Synthea SyntheticMass v2 dataset (https://synthea.mitre.org/downloads)


packages/core/src/bundle.test.ts
```diff
@@ -375,6 +378,20 @@ describe('Bundle tests', () => {
       expect(attachmentUrl).toBe(binaryEntry?.fullUrl);
       expect(attachmentUrl).toMatch(/^urn:uuid:/);
     });
+
+    test('Synthea collection bundle (Abbott509) matches BatchPage convert path', () => {
+      const syntheaPath = join(dirname(fileURLToPath(import.meta.url)), '__fixtures__', 'Abbott509_Aaron203_44.json');
+      let bundle = JSON.parse(readFileSync(syntheaPath, 'utf8')) as Bundle;
+      if (bundle.type !== 'batch' && bundle.type !== 'transaction') {
+        bundle = convertToTransactionBundle(bundle);
+      }
```
mattwiller:
```suggestion
      expect(bundle.type).toStrictEqual('collection');
      bundle = convertToTransactionBundle(bundle);
```

packages/core/src/bundle.test.ts
```diff
@@ -375,6 +378,20 @@ describe('Bundle tests', () => {
       expect(attachmentUrl).toBe(binaryEntry?.fullUrl);
       expect(attachmentUrl).toMatch(/^urn:uuid:/);
     });
+
+    test('Synthea collection bundle (Abbott509) matches BatchPage convert path', () => {
+      const syntheaPath = join(dirname(fileURLToPath(import.meta.url)), '__fixtures__', 'Abbott509_Aaron203_44.json');
```
mattwiller: Would it be possible to construct a smaller input for this test case?

packages/core/src/bundle.ts
```diff
@@ -175,7 +175,7 @@ function topologicalSortWithCycles(graph: AdjacencyList): { sorted: string[]; cy
 
     // Visit all neighbors
     let hasCycle = false;
-    for (const neighbor of graph[vertex]) {
+    for (const neighbor of graph[vertex] ?? []) {
```
codyebberson: Minor: you can use the `EMPTY` constant.

mattwiller: We have a static empty array in `@medplum/core` for exactly this scenario, to prevent unnecessary allocations:
```suggestion
    for (const neighbor of graph[vertex] ?? EMPTY) {
```

codyebberson (approved):

packages/core/src/bundle.ts
```diff
@@ -232,8 +232,8 @@ function buildAdjacencyList(bundle: Bundle): AdjacencyList {
     if (entry.resource) {
```
mattlong: Not to make it too many cooks in the kitchen, but it seems like just making this change would:
1. obviate the `fullUrl as string` cast below
2. obviate the defensive change above 
```suggestion
    if (fullUrl && entry.resource) {
```

mattlong (approved):

packages/core/src/bundle.test.ts
The-Alchemist: got it down to ~300 lines

packages/core/src/bundle.ts
The-Alchemist: done

PR #9113, opened 2026-05-03
fix(core): serialize cross-tab token refresh with Web Locks API

dillonstreator:
## Problem

Medplum refresh tokens are single-use with rotation. The server validates the incoming `refresh_secret` against the stored `Login.refreshSecret`, generates a new secret, and persists it (`packages/server/src/oauth/token.ts:336-343`). Any subsequent refresh with the previous token fails with `400 invalid_request "Invalid token"` (verified by `packages/server/src/oauth/token.test.ts:1276-1282`).

`MedplumClient.refresh()` only deduplicates concurrent refreshes within a single tab via the in-memory `this.refreshPromise`. It does not coordinate across tabs. When two tabs detect an expired access token at the same time, both `POST /oauth2/token` with the same refresh token. One wins, one fails — and the failure cascades:

1. Tab A and Tab B both hold refresh token `RT0` (secret `S0`).
2. Tab A refreshes → server rotates `S0 → S1` → Tab A receives `RT1`.
3. Tab B refreshes with `RT0` → server: `S0 ≠ S1` → `400 invalid_request`.
4. Tab B's `fetchTokens` hits the `!response.ok` branch → calls `clearActiveLogin()` and `onUnauthenticated?.()` (`client.ts:4456-4458`).
5. `clearActiveLogin` writes `activeLogin = undefined` to `localStorage`.
6. Tab A's `storage` listener (`client.ts:4520-4541`) sees `e.key === null` (or a cleared `activeLogin`) → triggers `locationUtils.reload()` and/or `this.clear()`.
7. **All tabs are logged out**, even the one that successfully refreshed.

This affects every user with multiple Medplum tabs open across an access-token-expiry boundary. Multi-tab auth refresh race events are more prevalent for web clients with websocket subscriptions bound to refetching and/or polling configured.

## Fix

Wrap the refresh network call in a [Web Lock](https://developer.mozilla.org/docs/Web/API/Web_Locks_API) scoped to this client's storage namespace. Tabs on the same origin serialize on the lock; the loser waits, re-reads `activeLogin` from storage when it acquires the lock, and skips the network call if a peer tab has already produced a fresh access token.

- Lock name: `medplum-refresh:<storage-key-for-activeLogin>` — namespaced via `storage.makeKey('activeLogin')` so multiple Medplum apps on the same origin (different `storagePrefix`) don't block each other.
- The in-memory `refreshPromise` dedup is preserved — it remains the cheap intra-tab path; the lock only kicks in when the tab actually issues the network request.
- `fetchTokens` already calls `setActiveLogin`, which writes the new tokens to storage before the lock releases. The next tab in the queue sees the fresh tokens immediately on lock acquisition.
- Browser auto-releases the lock if the tab is closed mid-refresh.

## Backward compatibility

- No new public API or option. The `MedplumClientOptions` surface is unchanged.
- Behavior is unchanged in environments without `navigator.locks` (Node, older Safari, Web Workers without Locks): a runtime check falls back to the existing in-memory dedup path.
- All existing call sites of `refresh()` / `refreshIfExpired()` continue to return a `Promise<void>` with identical resolve/reject semantics.

## Notes / follow-ups

- The existing `storage` event listener (`client.ts:4518-4544`) already syncs in-memory `accessToken` / `refreshToken` for non-refresh paths when another tab writes `activeLogin`. With this PR, that listener and the lock are complementary: the lock prevents the bad request in the first place; the listener keeps non-refresh code paths in sync.
- Could add an optional `lockTimeoutMs` later if we want to guard against a stuck lock holder, but Web Locks already release on tab close so this is unlikely to be needed.


packages/core/src/client.ts
```diff
@@ -4067,21 +4075,54 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
+   */
+  private async runRefreshWithLock(): Promise<ProfileResource | void> {
+    const run = (): Promise<ProfileResource | void> => {
+      // Re-read latest tokens from storage before hitting the network.
+      // A peer tab may have completed a refresh while we were queued on the lock.
+      const latest = this.getActiveLogin();
+      if (latest && latest.accessToken && latest.accessToken !== this.accessToken) {
+        this.setAccessToken(latest.accessToken, latest.refreshToken);
+      }
+      if (this.isAuthenticated(0)) {
```
ThatOneBro: I think this check here needs to respect the grace period, otherwise a call to `refreshIfExpired` with a grace period of 5 mins will call refresh which will then call this function and it could still be with grace period 0, `isAuthenticated` returns true and you fail to refresh.

ThatOneBro (commented):
Cool. I've never used Web Locks before, this is great. One comment about grace period but overall think this looks good

ThatOneBro:
@dillonstreator Thanks for making the fixes. 🙏

Going to deploy this branch to our staging environment in a bit to sanity check everything but looks good to me so far 👍

dillonstreator:
Thank you for helping to expedite @ThatOneBro
I'm fairly confident this is the source of latent mid workflow session logouts reported by frustrated users that was difficult to reproduce and debug, especially given the sporadic occurrences

ThatOneBro:
Update on the review:

I was still able to repro the multi-tab signout case on this branch by changing the access token lifetime to 1 minute. It seems this problem might be a little harder to solve then it seems...

I am seeing requests on other tabs during the refresh triggering sign out, which I guess means that if you don't get the lock you need a promise that resolves when the localStorage changes. Doing some further investigation

dillonstreator:
> I am seeing requests on other tabs during the refresh triggering sign out, which I guess means that if you don't get the lock you need a promise that resolves when the localStorage changes. Doing some further investigation

I wonder if part of this is due to the contrived test. Is the gracePeriod still 5min or did you lower that as well to a duration less than token lifetime? Is the token effectively always considered to be in a state of needing refresh? Lock acquisition should be blocking with the default configurations for `mode` and `ifAvailable` so it's odd that you'd see requests on other tabs since they should all be waiting for lock acquisition. The local storage active login is already set/updated inside the lock before releasing.

ThatOneBro:
Good point, I'll check out what happens if grace period is lower than lifetime. I guess that could affect further refreshes

ThatOneBro:
Okay just to confirm, even with a grace period of 25 seconds I'm still seeing the same behavior on a local copy of Medplum App

What have you been using to test this behavior?

dillonstreator:
I haven't setup/run a live repro and I'm not setup to test medplum client / server code / config changes on the fly locally. This PR came after tracing production logs and reviewing `client.ts`. Do you know which path is triggering `clearActiveLogin`?

dillonstreator:
@ThatOneBro I was able to reproduce the fix working as expected after working through a couple gotchas in locally testing:

## 1. Infra

```sh
docker compose up -d  # postgres + redis
```

## 2. Build workspace deps (first time only)

```sh
npx turbo run build --filter=@medplum/server
```

`@medplum/core` and other workspace packages need dist output before the server's `tsx watch` can resolve them. After this, the app's `vite dev` reads `@medplum/core` directly from `../core/src` (alias in `packages/app/vite.config.ts`), so the PR change is picked up live without rebuilding core.

## 3. Two source tweaks for fast repro

### a) Short access-token lifetime

`packages/server/src/oauth/keys.ts`

```diff
-const DEFAULT_ACCESS_LIFETIME = '1h';
+const DEFAULT_ACCESS_LIFETIME = '1m';
```

### b) Tight client-side grace period

`packages/app/src/index.tsx` — inside `new MedplumClient({...})`:

```diff
     storagePrefix: '@medplum:',
     cacheTime: 60000,
     autoBatchTime: 100,
+    refreshGracePeriod: 5000,
```

Default grace is 5 minutes; with a 1-minute token that means refresh would fire on every request. 5s grace = refresh kicks in only when the token has <5s left, which lets you observe the actual race.

## 4. Gotcha 1 — `remember` must be true

`AuthenticationForm` only sends `remember: true` if the **"Remember me" checkbox is ticked**. Default is unchecked. Without it, the server's `includeRefreshToken()` returns false and **no refresh token is issued at all** — token expires → 401 → `refresh()` early-returns at:

```ts
if (!this.refreshToken && !(this.clientId && this.clientSecret)) {
  return undefined;
}
```

…and you get logged out instead of refreshing. The Web Lock path is never entered.

Alternative: sign in with scope `offline` or `offline_access`.

## 5. Gotcha 2 — super admin bypass

Even with `remember: true`, the default `admin@example.com` user belongs to the Super Admin project, and `setLoginMembership` strips `refreshSecret` for super admins:

`packages/server/src/oauth/utils.ts` (~line 492):

```diff
-  if (project.superAdmin) {
-    // Disable refresh tokens for super admins
-    updatedLogin.refreshSecret = undefined;
-  }
+  // TEMP: comment out for multi-tab refresh testing
+  // if (project.superAdmin) {
+  //   updatedLogin.refreshSecret = undefined;
+  // }
```

Cleaner alternative is to register a non-super-admin project + user via `/register` and test there — but commenting out the gate is faster.

## 6. Run

```sh
cd packages/server && npm run dev   # 8103
cd packages/app    && npm run dev   # 3000 (or next free port)
```

## 7. Sign in & verify refresh token issued

1. Open the app, sign in as `admin@example.com` / `medplum_admin` **with "Remember me" checked**.
2. DevTools console:

```js
JSON.parse(localStorage['@medplum:activeLogin']).refreshToken
```

Must be a JWT string. If `undefined`, one of the gotchas above is in play.

## 8. Multi-tab repro

1. Open a second tab to the same origin (auth carries via `localStorage`).
2. DevTools → Sources → `client.ts`, add **logpoints** (right-click gutter → Add logpoint) at:
   - `clearActiveLogin` impl (~line 1218): `console.trace('clearActiveLogin')`
   - Refresh-failure path (`fetchTokens`, ~line 4499): `console.trace('refresh 400')`
   - Storage listener (~line 4523): `console.trace('storage listener clear')`
   - `runRefreshWithLock` after `locks.request` (~line 4127): `console.log('lock acquired', lockName)`
3. Widen the race window — paste in **both tabs**:

   ```js
   const orig = window.fetch;
   window.fetch = (...a) =>
     a[0]?.toString?.().includes('/oauth2/token')
       ? new Promise(r => setTimeout(() => r(orig(...a)), 2000))
       : orig(...a);
   ```

4. Wait ~55s after login (or until access token has <5s left).
5. Trigger an authed request in **each tab within ~1s of each other** (refresh page, click a resource, etc.).

### Expected with PR fix

- One tab acquires `medplum-refresh:@medplum:activeLogin` Web Lock, hits `/oauth2/token`, writes new tokens to storage.
- Other tab's logpoint shows lock acquired *after* first releases, re-reads storage, skips network call.
- `clearActiveLogin` logpoints stay silent. Both tabs remain logged in.

### Without the fix (`git stash` the PR change)

- Second tab POSTs `/oauth2/token` with the already-rotated refresh secret → 400.
- `clearActiveLogin` fires at line 4499 → wipes `localStorage`.
- First tab's storage listener fires → reload / clear.
- Both tabs end up at `/signin`.

## 9. Watch lock state live

```js
await navigator.locks.query()
```

`held` lists active holder, `pending` lists queued tabs.

## 10. Cleanup

```sh
git checkout packages/server/src/oauth/keys.ts \
             packages/server/src/oauth/utils.ts \
             packages/app/src/index.tsx
docker compose down
```

ThatOneBro:
Thanks for the detailed write-up @dillonstreator! It looks like the key detail I consistently kept missing was the `Remember` step, which I had forgotten was added in a relatively recent upgrade to the Sign In box 😅 

Looks like it's working as expected after that, sorry about the confusion.

LGTM 👍

ThatOneBro (approved):

mattlong:
@claude please review

PR #9121, opened 2026-05-05
Tenant Selector Options

finnbergquist:
Documenting different options for building a tenant selector.


The-Alchemist (approved):

everett-williams:
I feel like this could benefit from a diagram or two. Also maybe a quick cleanup of formatting, I have a docs formatting guide I can send so you can pass this through AI and get it to look and feel a bit more like our other documentation

everett-williams (approved):

PR #9122, opened 2026-05-05
chore(provider): use LinkTabs in provider patient page

techdavidy:
Use LinkTabs in patient screen for consistent UX across app and provider. 
 

Screenshot 2026-05-05 at 10 48 01 AM


examples/medplum-provider/src/pages/patient/PatientPage.tsx
```diff
@@ -121,7 +82,17 @@ export function PatientPage(): JSX.Element {
         </div>
 
         <div className={classes.content}>
-          <PatientTabsNavigation tabs={tabs} currentTab={currentTab} onTabChange={onTabChange} />
+          <Paper w="100%" radius={0} style={{ borderBottom: '1px solid var(--app-shell-border-color)' }}>
+            <ScrollArea>
+              <LinkTabs
+                baseUrl={patientPathPrefix(patientId)}
+                tabs={resolvedTabs.map((t) => ({ label: t.label, value: t.url || t.id }))}
```
noahsilas: If you move this map function inside of `useMemo`, then this prop will have a stable reference (which may help React rendering performance).

packages/react/src/LinkTabs/LinkTabs.tsx
```diff
@@ -27,8 +27,15 @@ export function LinkTabs(props: LinkTabsProps): JSX.Element {
   const navigate = useMedplumNavigate();
 
   const [currentTab, setCurrentTab] = useState(() => {
-    const tab = locationUtils.getPathname().split('/').pop();
-    return tab && tabs.some((t) => t.value === tab) ? tab : tabs[0].value;
+    const segment = locationUtils.getPathname().split('/').pop();
+    if (segment) {
+      const segmentLower = segment.toLowerCase();
+      const matched = tabs.find((t) => t.value.split(/[?#]/)[0].toLowerCase() === segmentLower);
```
noahsilas, line 33: Just re-stating my understanding: this is removing any querystring (?...) or fragment identifier (#...) parts of a URL from the tab value, which leaves the path component only for us to compare with the current URL pathname component.

packages/react/src/LinkTabs/LinkTabs.tsx
```diff
@@ -38,7 +45,7 @@ export function LinkTabs(props: LinkTabsProps): JSX.Element {
   }
 
   return (
-    <Tabs value={currentTab.toLowerCase()} onChange={onTabChange} {...rest}>
+    <Tabs value={currentTab} onChange={onTabChange} {...rest}>
```
noahsilas, line 48: I think removing the `toLowerCase()` here makes sense. I see that we did introduce some new case manipulation above though - would it make sense to remove the `toLowerCase()` above too?

noahsilas (approved):
Nice 👍

PR #9130, opened 2026-05-05
feat(provider): spaces render markdown responses

techdavidy:
Screenshot 2026-05-05 at 1 41 55 PM

### DIFF
--- a/examples/medplum-provider/src/components/spaces/Markdown.tsx
+++ b/examples/medplum-provider/src/components/spaces/Markdown.tsx
@@ -1,0 +1,94 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import { Anchor, Code, List, Table, Text, Title } from '@mantine/core';
+import type { JSX } from 'react';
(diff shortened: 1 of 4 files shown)


techdavidy:
@claude please review

codyebberson (approved):

examples/medplum-provider/src/components/spaces/Markdown.tsx
```diff
@@ -0,0 +1,94 @@
...
+      {children}
+    </Text>
+  ),
+  em: ({ children }) => (
+    <Text component="span" fs="italic">
+      {children}
+    </Text>
+  ),
+  code: ({ className, children }) => {
+    const isBlock = (className ?? '').includes('language-');
```
noahsilas, line 35: We may also want to capture blocks that don't have a language tag. I think you can detect that with something like this:
```ts
const hasNewline = typeof children === 'string' && children.includes('\n');
const hasLanguage = (className ?? '').includes('language-');
const isBlock = hasNewline || hasLanguage;
```

noahsilas (commented):
I sort of wish we had some kind of code-splitting situation in here so that this dependency could be loaded only when someone interacts with it. Something to think about for later. Looks good to me!

PR #9132, opened 2026-05-05
issue-8934-fix Add projectId and subscriptionId to failed auto-disabl…

### LINKED ISSUE #8934: Add projectId and subscriptionId to the Failed to auto-disable subscription log line
Similar to the information included when a subscription is successfully auto-disabled. While we're there, replace `Failed to auto-disable subscription` with `errorMessage` for consistency.

csrinivas2130:
Fixes #8934.

Updates the subscription auto-disable failure log to use the generated auto-disable error message and include structured identifiers for the affected project and subscription.

Changes:
- Moves `errorMessage` outside the `try` block so it can be reused in the failure log.
- Adds `projectId`, `subscriptionId`, and `failureCount` to the failure log context.
- Adds regression coverage for the failure path.

This does not change auto-disable behavior; it only improves observability when auto-disable itself fails.


packages/server/src/fhir/repo.ts
```diff
@@ -247,6 +247,14 @@ function addSyntheticR4ProjectIfMissing(context: RepositoryContext): void {
...
 }
 
+function addSyntheticR4ProjectIfMissing(context: RepositoryContext): void {
+  if (context.projects && !context.projects.some((project) => project.id === syntheticR4Project.id)) {
+    // Repositories with a project scope can also see the synthetic R4 project,
+    // but repeated construction from the same context must not append duplicates.
+    context.projects.push(syntheticR4Project);
+  }
+}
+
```
mattwiller, line 257: This doesn't look like it's being called from anywhere; could it be removed?

packages/server/src/workers/subscription.test.ts
```diff
@@ -3678,10 +3678,62 @@ describe('Subscription Worker', () => {
...
+
+        const patient = await repo.createResource<Patient>({
+          resourceType: 'Patient',
+          name: [{ given: ['Test'], family: 'AutoDisableLogFailure' }],
+        });
+
+        await expect(findAndExecSubscriptionJob(patient, 'create')).rejects.toThrow(UnrecoverableError);
+
+        expect(warnSpy).toHaveBeenCalledWith(
+          'Automatically disabled after 1 consecutive failed events in the last 600 seconds',
```
mattwiller, line 3723: This doesn't look like it actually tests the code path you changed; would you mind updating the test to hit the `Failed to auto-disable subscription` log?

PR #9137, opened 2026-05-06
auth/login: differentiate "no membership" from "user not found"

tr00x:
## Summary

When `tryLogin` succeeds at password verification but the user has zero
active `ProjectMembership` rows matching the request (`memberships.length === 0`
in `packages/server/src/oauth/utils.ts`), the response is currently:

```
400 OperationOutcome { issue: [{ code: 'invalid', details: { text: 'User not found' } }] }
```

That's the same shape and message as the case where the email doesn't
exist at all, and it's misleading. The user *does* exist — the operator
just needs to know the membership got dropped (e.g. by a backup restore,
an accidental admin delete, or unfinished invite flow).

This PR keeps the same HTTP status (400) and `OperationOutcome` shape,
just swaps the message body so operators can tell the two failure modes
apart from the response alone.

## What changed

- `packages/server/src/oauth/utils.ts` — single `badRequest()` call now
  returns a more actionable string and a short comment explaining why the
  message differs from the genuinely-not-found path on line 176.
- `packages/server/src/auth/login.test.ts` — both existing tests that
  asserted the old `"User not found"` message ("No membership" and
  "Inactive membership") updated to the new one.

The genuinely-not-found path higher up (line 176) is intentionally left
as `'User not found'` — anti-enumeration still matters there because the
caller has not yet authenticated.

## Why we hit this

Building a self-hosted referral portal on top of Medplum, a
`ProjectMembership` row got orphaned during a partial backup restore. The
misleading `"User not found"` cost ~30 min of looking at the wrong layer.
The fix is one string + a comment — but the diagnostic value is real.

## Test plan

- [ ] Updated unit tests cover the two cases that asserted the old
      message; full upstream suite verified by CI in this PR.
- [ ] `grep -rn "User not found" packages/` confirms no other test
      asserts the literal old message in the no-membership path.
- [ ] Maintainer review on the new wording.

## Open questions

- Whether the no-membership case should be `forbidden` (403) rather than
  `invalid` (400) — semantically it's "you authenticated, but have no
  scope here." Kept at 400 in this PR to avoid changing existing client
  expectations; happy to switch if maintainers prefer.

---

Two more small auth fixes are sketched out as PR-ready proposals (admin-mediated `auth/setpassword`, first-class magic-link) in [`docs/upstream/`](https://github.com/tr00x/vendo/tree/main/docs/upstream) of the downstream repo where this fix surfaced. Will send them as separate PRs once this one lands or gets feedback.


packages/server/src/oauth/utils.ts
```diff
@@ -211,7 +211,13 @@ export async function tryLogin(request: LoginRequest): Promise<WithId<Login>> {
   }
 
   if (memberships.length === 0 && !request.allowNoMembership) {
-    throw new OperationOutcomeError(badRequest('User not found'));
+    // The user authenticated successfully, but has no active ProjectMembership
+    // matching the request. Surface this distinctly from "user does not exist"
+    // so operators can tell the two failure modes apart — same status (400)
+    // and same OperationOutcome shape, just a more actionable message.
+    throw new OperationOutcomeError(
+      badRequest('User has no active project memberships. Contact a project admin to be re-invited.')
```
mattlong: Let's shorten this to just `User has no active project memberships`

mattlong (changes requested):
Thanks, @tr00x! Much more actionable and less misleading error message 😅 

To get the DCO check passing, you'll need to sign off on each commit and rebase as described [here](https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot).

packages/server/src/oauth/utils.ts
tr00x: Done — shortened to `User has no active project memberships` in both `oauth/utils.ts` and the two `login.test.ts` assertions. Pushed as a force-update on the same branch (DCO sign-off preserved).

PR #9138, opened 2026-05-06
Backport R5 type guard syntax in search parameter expressions

mattwiller:
The `as`/`.as()` syntax used in FHIR R4 and R4B is incorrect: since it should throw an error when encountering a different type, it was replaced in R5 with the `.ofType()` function.  Backporting this allows search parameter type guards to work correctly with Medplum's FHIRPath implementation


mattlong (approved):
LGTM. Curious what your process was for updating the `expression` entries in `search-parameters.json`

mattwiller:
I hand-searched for " as " and replaced usages that weren't in other description fields, and did a find-and-replace for `.as(` to `.ofType(`

PR #9143, opened 2026-05-06
fix(provider): Pre fill patient in lab ordering

techdavidy:
Screenshot 2026-05-06 at 4 05 27 PM


examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
```diff
@@ -94,6 +96,12 @@ export function OrderLabsPage(props: OrderLabsPageProps): JSX.Element {
     }
   }, [tests, setTests]);
 
+  useEffect(() => {
+    if (defaultPatientResource) {
```
mattlong: Does this not need to check if `patient` was already some other way? Seems like this would override a patient chosen with the `ResourceInput` below

mattlong (approved):

PR #9146, opened 2026-05-07
Provider: Reveal more slots in SchedulePage

noahsilas:
Previously, we were not showing "busy" slots, presumably with the assumption that they mapped one-to-one to Appointments that would be displayed. This sometimes causes confusion when that relationship doesn't hold, such as when someone has deleted an Appointment without also deleting the related Slot resources.

We now explicitly hide Slots that are referenced by an Appointment that we are displaying _and_ exactly match the start/end times.

We also explicitly hide slots with status "entered-in-error", as those should not affect availability.

### DIFF
--- a/examples/medplum-provider/src/components/Calendar.tsx
+++ b/examples/medplum-provider/src/components/Calendar.tsx
@@ -12,7 +12,6 @@ import type { Event, SlotInfo, ToolbarProps, View } from 'react-big-calendar';
 import { Calendar as ReactBigCalendar, dayjsLocalizer } from 'react-big-calendar';
 import 'react-big-calendar/lib/css/react-big-calendar.css';
 import type { Range } from '../types/scheduling';
-import { SchedulingTransientIdentifier } from '../utils/scheduling';
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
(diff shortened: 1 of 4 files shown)


codyebberson:
@claude please review

noahsilas:
> Dead code inside forEach

Done ✅ 

> String equality for date matching (Calendar.tsx:208)

I don't think there's a meaningful difference between this string comparison and a date comparison, this just was easier/cheaper 🤷 

> Slot merging interaction

I sort of think we should remove this merging behavior and reflect the real Slot resources on the calendar. I'm not entirely sure what the original intention here was though, so I haven't addressed it yet. This weird interaction does seem like a good reason to do so though...

examples/medplum-provider/src/components/Calendar.tsx
```diff
@@ -183,18 +183,35 @@ export function Calendar(props: {
...
-    () => slotsToEvents(props.slots.filter((slot) => !SchedulingTransientIdentifier.get(slot))),
-    [props.slots]
-  );
+  const events = useMemo(() => appointmentsToEvents(props.appointments), [props.appointments]);
+
+  const backgroundEvents = useMemo(() => {
+    const appointmentIndex = props.appointments.reduce<Record<string, Appointment>>((acc, appointment) => {
+      (appointment.slot ?? []).forEach((slotRef) => {
+        const key = getReferenceString(slotRef);
+        if (key) {
```
codyebberson, line 192: When is `key` not truthy?

examples/medplum-provider/src/components/Calendar.tsx
```diff
@@ -183,18 +183,35 @@ export function Calendar(props: {
...
-
-  const backgroundEvents = useMemo(
-    () => slotsToEvents(props.slots.filter((slot) => !SchedulingTransientIdentifier.get(slot))),
-    [props.slots]
-  );
+  const events = useMemo(() => appointmentsToEvents(props.appointments), [props.appointments]);
+
+  const backgroundEvents = useMemo(() => {
+    const appointmentIndex = props.appointments.reduce<Record<string, Appointment>>((acc, appointment) => {
+      (appointment.slot ?? []).forEach((slotRef) => {
```
codyebberson: We have a `EMPTY` constant in `@medplum/core` to avoid creating empty arrays.

codyebberson (approved):
Nice 👍

examples/medplum-provider/src/components/Calendar.tsx
noahsilas: In practice I think it always should be, but the type of `appointment.slot` is `Reference<Slot>[]`, which is allowed under typescript to not have a `reference` attribute, in which case `getReferenceString()` returns undefined.

So this is mostly making Typescript happy, but also does cover a technically valid case.

PR #9152, opened 2026-05-07
Add admin API endpoint for rate limit utilization status

maddyli:
## Summary
- Adds `GET /admin/projects/:projectId/rate-limits` endpoint that reads current FHIR quota consumption directly from Redis
- Accepts optional `?membershipId=<id>` query params (repeatable) to check specific project memberships
- Returns per-membership and per-project quota utilization (consumed points, remaining points, time to reset) without consuming any rate limit points
- Protected by existing `verifyProjectAdmin` middleware — only project admins and super admins can access

## Test plan
- [ ] Verify endpoint returns rate limit status for all project members when no `membershipId` is specified
- [ ] Verify filtering by one or more `membershipId` query params
- [ ] Verify 403 response for non-admin users
- [ ] Verify 403 response when requesting a membership from a different project
- [ ] Verify `null` fhirQuota for members with no recent activity (no Redis key)
- [ ] Verify consumed points appear after FHIR activity


packages/server/src/admin/project.ts
```diff
@@ -54,6 +55,7 @@ projectAdminRouter.post(
   }
 );
 
+projectAdminRouter.get('/:projectId/rate-limits', getRateLimitStatusHandler);
```
mattwiller: I think we should probably make this a Project-level operation endpoint, e.g. `/fhir/R4/Project/<id>/$rate-limits`

packages/server/src/admin/ratelimit.ts
```diff
@@ -0,0 +1,110 @@
...
+    points: userLimit,
+    duration: 60,
+  });
+
+  const projectLimiter = new RateLimiterRedis({
+    keyPrefix: 'medplum:rl:fhir:project:',
+    storeClient: redis,
+    points: projectLimit,
+    duration: 60,
+  });
```
mattwiller: This logic should probably live alongside/be shared with the rate limiter itself, to ensure that the configuration stays in sync

packages/server/src/admin/ratelimit.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export async function getRateLimitStatusHandler(req: Request, res: Response): Promise<void> {
+  const ctx = getAuthenticatedContext();
+  const project = ctx.project;
+
+  const membershipIds = req.query.membershipId;
+  let idsArray: string[] | undefined;
+  if (membershipIds !== undefined) {
+    idsArray = Array.isArray(membershipIds)
```
ThatOneBro: I think we have an `arrayify` util we could use here

packages/server/src/admin/ratelimit.ts
```diff
@@ -0,0 +1,108 @@
...
+  const projectLimiter = new RateLimiterRedis({
+    keyPrefix: 'medplum:rl:fhir:project:',
+    storeClient: redis,
+    points: projectLimit,
+    duration: 60,
+  });
+
+  const results: MembershipRateLimitStatus[] = await Promise.all(
+    memberships.map(async (membership) => {
+      const result = await membershipLimiter.get(membership.id as string);
```
ThatOneBro: Up to 1000 single GETs in a tight loop to Redis seems a little much, especially if we give this to users to call however often they want

ThatOneBro: We should check if there is a multi-get API exposed by the rate limiter

ThatOneBro (commented):
Overall makes sense

We should probably cost this operation, since it could be potentially expensive itself to search for all project memberships and then up to 2000 singular calls to Redis

Preferably we could condense that down to one big call, but we should at least assign some cost to the operation either way

packages/server/src/admin/project.ts
maddyli: Addressed, added a Project-level operation endpoint that still requires project admin credentials

packages/server/src/fhir/operations/project-rate-limits.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export async function projectRateLimitsHandler(req: FhirRequest): Promise<FhirResponse> {
+  const ctx = getAuthenticatedContext();
+  if (!ctx.repo.isProjectAdmin() && !ctx.repo.isSuperAdmin()) {
+    return [forbidden];
+  }
+
+  const project = ctx.project;
+  const idsArray = arrayify(req.query.membershipId);
```
mattwiller: The operation parameters should ideally be defined in an OperationDefinition in this file, so you can use the built-in parameter parsing utilities like parseInputParameters() and buildOutputParameters() to ensure consistency.  This would allow removing e.g. the `quotaParts()` helper below

packages/server/src/fhir/operations/project-rate-limits.ts
```diff
@@ -0,0 +1,125 @@
...
+            ...(m.profileReference ? [{ name: 'profileReference', valueString: m.profileReference }] : []),
+            ...quotaParts(m.quota, userLimit),
+          ],
+        })),
+      ],
+    } as any,
+  ];
+}
+
+function parseQuotaResult(consumed: string | null, pttl: number, limit: number): QuotaStatus | undefined {
```
mattwiller: [nit] "parse" doesn't really fit here; I'd recommend something like `buildQuotaStatus()` instead to more clearly show the intent of the function

mattwiller:
@maddyli The implementation in general looks solid to me — would just be good to bring this in line with our other operations endpoints using an `OperationDefinition` and helper functions to make the operation input and output format more declarative

mattwiller (approved):

PR #9154, opened 2026-05-07
feat(server,app): add `User/$rescope` operation

### LINKED ISSUE #8963: User re-scope operation
Users commonly request that server-scoped users be rescoped to project-scoped users, and occasionally vice-versa. 

This issue is to create an operation that re-scopes server users to be project-scoped, and perhaps to do the opposite. This is likely going to be a super admin operation, due to the permissions structures that will need to change due to this operation.

ThatOneBro:
Fixes #8963

Rescoping as project admin from membership page:
Screenshot 2026-05-07 at 12 57 20 PM
Screenshot 2026-05-07 at 12 57 40 PM

Rescoping as super admin from super config page:

Super admins can move both from global scope -> project scope and project scope -> global scope.
Selecting a project and either global or project target scope narrows which users are searchable in the User input to users with project memberships in the selected project.
Screenshot 2026-05-08 at 9 19 42 AM
Screenshot 2026-05-08 at 9 19 53 AM


packages/server/src/fhir/operations/rescope.ts
```diff
@@ -0,0 +1,215 @@
...
+  mutate: (user: WithId<User>, systemRepo: ReturnType<typeof getGlobalSystemRepo>) => Promise<WithId<User>>
+): Promise<WithId<User>> {
+  await ctx.repo.readResource('User', userId);
+  const systemRepo = getGlobalSystemRepo();
+  return systemRepo.withTransaction(
+    async () => {
+      const user = await systemRepo.readResource<User>('User', userId);
+      return mutate(user, systemRepo);
+    },
+    { serializable: true }
```
mattwiller: Might be worth documenting why this requires `SERIALIZABLE` (the searches in the mutation functions below?) in a comment

packages/server/src/fhir/operations/rescope.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    // Verify the caller has write access to this User: super admin, or project admin
+    // of the project the User currently belongs to.
+    if (!callerIsSuperAdmin && user.project.reference !== getReferenceString(ctx.project)) {
+      throw new OperationOutcomeError(forbidden);
+    }
+
+    const { project: _project, ...userWithoutProject } = user;
+    const { project: _metaProject, ...metaWithoutProject } = user.meta ?? {};
```
mattwiller: Since the User object is local to this function, it might be faster/simpler to just `delete user.project` and `delete user.meta?.project` in this case

packages/server/src/fhir/operations/rescope.ts
```diff
@@ -0,0 +1,215 @@
...
+  return withRescopeTxn(ctx, userId, async (user, systemRepo) => {
+    if (!user.project) {
+      throw new OperationOutcomeError(badRequest('User is already global-scoped'));
+    }
+
+    // Verify the caller has write access to this User: super admin, or project admin
+    // of the project the User currently belongs to.
+    if (!callerIsSuperAdmin && user.project.reference !== getReferenceString(ctx.project)) {
+      throw new OperationOutcomeError(forbidden);
+    }
```
mattwiller: This kind of duplicates the checks in `userRescopeOperation()`, and there are no corresponding checks in `rescopeUserToProject()`: could these permission checks be centralized somehow?

packages/server/src/fhir/operations/rescope.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  if (!['project', 'global'].includes(params.scope)) {
+    return [badRequest('Invalid scope: must be "project" or "global"')];
+  }
+
+  // Promoting a User to project scope (or moving between projects) requires super admin.
+  // Demoting to global scope allows a project admin as long as they have write access to the User.
+  const rescopeToProjectAllowed = ctx.project.superAdmin;
+  const rescopeToGlobalAllowed = ctx.project.superAdmin || ctx.membership.admin;
```
mattwiller: Should this incorporate the Project ID check from below?

packages/server/src/fhir/operations/rescope.ts
```diff
@@ -0,0 +1,215 @@
...
+ *   re-read User and the system repo; returns the updated User.
+ * @returns The updated User as returned by the mutate callback.
+ */
+async function withRescopeTxn(
+  ctx: AuthenticatedRequestContext,
+  userId: string,
+  mutate: (user: WithId<User>, systemRepo: ReturnType<typeof getGlobalSystemRepo>) => Promise<WithId<User>>
+): Promise<WithId<User>> {
+  await ctx.repo.readResource('User', userId);
+  const systemRepo = getGlobalSystemRepo();
```
mattwiller: Why not use `ctx.systemRepo` here?

ThatOneBro: Good question, I haven't been updated on what the difference is of any of those things 👀 

We should probably have a summary of that somewhere. I just saw some other operations calling `getGlobalSystemRepo` but even then was unsure

packages/server/src/fhir/operations/rescope.ts
ThatOneBro: Yeah funny enough I was just looking at that

packages/server/src/fhir/operations/rescope.ts
ThatOneBro: Yeah I think I'm going to probably remove this helper anyways, I think it kind of obscures things. Originally I thought the duplication was getting flagged in the split between the two flows for project and global but looks like it was just the operation definition JSON 🙃

packages/server/src/fhir/operations/rescope.ts
mattlong: That summary is in repo.ts

ThatOneBro: Oh I remember now, I think because Users are going to be globally scoped if I remember correctly? @mattlong

mattwiller (approved):

PR #9156, opened 2026-05-07
Fixes #9039 - Server option to require verified email address

### LINKED ISSUE #9039: Add server config option to require email verification for new accounts

codyebberson:
Adds an optional server configuration setting (`requireVerifiedEmailForProjectCreation`) that requires users to verify their email address before creating a new project via the self-registration flow.

**This setting does NOT affect:**
- New users created through the Invite flow
- Logins for existing users

### How it works

**Registration flow with verification enabled:**

1. User submits the registration form → `POST /auth/newuser` with `projectId: 'new'`
2. Server creates the user + login, sends a verification email, and returns `{ login, emailVerificationRequired: true }` (instead of proceeding to project creation)
3. UI shows an instructional message: *"Please check your email for a verification link…"*
4. User clicks the link in the email → `/verifyemail/<id>/<secret>`
5. User clicks "Verify email" → `POST /auth/verifyemail` marks `user.emailVerified = true` and returns a redirect URI
6. UI follows the redirect back to `/register?login=<loginId>`
7. User enters a project name and completes registration via `POST /auth/newproject`

### Configuration

Set in `medplum.config.json`:

```json
{
  "requireVerifiedEmailForProjectCreation": true
}
```

Default is `false` (opt-in, no change in behavior for existing deployments).

### Technical notes

- `UserSecurityRequest` (type `verify-email`) is reused from the existing `$update-email` operation — no new FHIR resource type needed
- PKCE state (`pkceState`, `codeVerifier`) was migrated from `sessionStorage` to `MedplumClient.storage` (localStorage by default). This allows the PKCE flow to survive the cross-tab redirect introduced by email verification. **This is a behavioral change for all PKCE flows** — the verifier now persists across sessions rather than being cleared on tab close
- Verification links do not expire (consistent with existing password-reset behavior)
- The `/auth/verifyemail` handler returns an `OperationOutcome` with HTTP 200 (rather than a true 302 redirect) so the React SPA can extract the URI and navigate without a full-page reload
- `oauth/utils.ts`: `setLoginMembership` now correctly sets `login.project` alongside `login.membership` (bug fix bundled in this PR)

Fixes #9039

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -3981,8 +3981,6 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
     const codeVerifier = getRandomString().slice(0, 128);
     this.storage.setString('codeVerifier', codeVerifier);
 
-    console.log('CODY startPkce', { pkceState, codeVerifier });
-
     try {
       const arrayHash = await encryptSHA256(codeVerifier);
       const codeChallenge = arrayBufferToBase64(arrayHash)
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -517,7 +517,6 @@ export interface NewUserRequest {
   readonly remember?: boolean;
   readonly projectId?: string;
   readonly clientId?: string;
-  readonly redirectUri?: string;
 }
 
 export interface NewProjectRequest {
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -543,6 +543,7 @@ export interface GoogleLoginRequest extends BaseLoginRequest {
 
 export interface LoginAuthenticationResponse {
   readonly login: string;
+  readonly emailVerificationRequired?: boolean;
   readonly mfaEnrollRequired?: boolean;
   readonly mfaRequired?: boolean;
   readonly enrollQrCode?: string;
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -512,11 +512,12 @@ export interface NewUserRequest {
   readonly lastName: string;
   readonly email: string;
   readonly password: string;
-  readonly recaptchaToken: string;
+  readonly recaptchaToken?: string;
   readonly recaptchaSiteKey?: string;
   readonly remember?: boolean;
   readonly projectId?: string;
   readonly clientId?: string;
+  readonly redirectUri?: string;
 }
 
 export interface NewProjectRequest {
@@ -1204,9 +1205,6 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    */
   clear(): void {
     this.storage.clear();
-    if (isBrowserEnvironment()) {
-      sessionStorage.clear();
-    }
     this.clearActiveLogin();
   }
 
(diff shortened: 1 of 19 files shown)


codyebberson:
@claude please review

packages/server/src/auth/newuser.ts
```diff
@@ -115,6 +123,31 @@ export async function createUser(request: Omit<NewUserRequest, 'recaptchaToken'>
...
   });
+
   globalLogger.info('User created', { id: result.id, email });
   return result;
 }
+
+async function sendVerificationEmail(user: WithId<User>, login: WithId<Login>): Promise<void> {
+  const redirectUri = concatUrls(getConfig().appBaseUrl, `register?login=${login.id}`);
+  const { id, secret } = await verifyEmail(user, redirectUri);
+  const url = concatUrls(getConfig().appBaseUrl, `verifyemail/${id}/${secret}`);
```
mattwiller, line 134: Could this live inside `verifyEmail()`?  Might be nice to colocate it with the email verification code

mattwiller (approved):

packages/core/src/client.ts
```diff
@@ -3977,10 +3975,10 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
    */
   async startPkce(): Promise<{ codeChallengeMethod: CodeChallengeMethod; codeChallenge: string }> {
     const pkceState = getRandomString();
-    sessionStorage.setItem('pkceState', pkceState);
+    this.storage.setString('pkceState', pkceState);
```
mattlong, line 3978: We've gotten a little bit of heat for using localStorage instead of cookies for storing sensitive info; this seems like we're going further down that road?

packages/server/src/auth/newuser.ts
codyebberson: Yeah, good question, I wondered about that.

There were two significant differences between this one and the other call site:
1. The redirect URL is different
2. The email message was different

So it started to feel premature.

I felt like it might be cleaner separate for now, thinking of "Rule of Three": https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

packages/server/src/auth/register.ts
```diff
@@ -38,6 +38,7 @@ export interface RegisterResponse {
 
 /**
  * Registers a new user and/or new project.
+ * This method is now only used in tests! Do not use this method in production code!
```
mattlong, line 41: We should move it to `test.setup.ts` and rename it to `newTestRegistration` or something

packages/server/src/auth/verifyemail.ts
```diff
@@ -42,18 +42,23 @@ export async function verifyEmailHandler(req: Request, res: Response): Promise<v
     await systemRepo.updateResource<UserSecurityRequest>({ ...securityRequest, used: true });
   });
 
-  sendOutcome(res, allOk);
+  if (securityRequest.redirectUri) {
+    // Send a "redirect", but don't actually follow it, because the client may want to handle the redirect themselves (e.g. in a React app).
+    sendOutcome(res, { ...redirect(new URL(securityRequest.redirectUri)), id: 'ok' });
```
mattlong: Claude's recommendation seemed reasonable here.

mattlong (commented):
>New users created through the Invite flow

Having divergent behavior for the invite flow seems like a smell. What's the reasoning for excluding it?

packages/core/src/client.ts
codyebberson: This particular data is explicitly client-only, so a cookie would not be the right call in this case.

packages/server/src/auth/register.ts
codyebberson: Yeah, agreed, I didn't want to overload this PR.  I'll follow up with that separately.

codyebberson:
> > New users created through the Invite flow
> 
> Having divergent behavior for the invite flow seems like a smell. What's the reasoning for excluding it?

That would be a net-new flow for 100% of Medplum users.  That's a pretty big change when you think through the full details.

PR #9160, opened 2026-05-08
refactor(server/agent): use 1 cb channel per host

### LINKED ISSUE #9089: [Agent] Optimize `pushToAgent` Redis PubSub callback system
Currently we create a temporary Redis PubSub subscriber for every call to `pushToAgent` to wait for the response from the agent to send back via the HTTP response. This has worked well and not seemed to been a bottleneck in the past. However, we have used a different pattern for WebSocket subscriptions which seems to be more efficient: creating one subscriber per host and at the host level routing the responses to their caller.

This could reduce I/O overhead on Redis during high throughput scenarios and is probably just something worth collecting stats on in regards to performance, both in terms of end-user latency at scale and also Redis performance / connection usage.

ThatOneBro:
Fixes #9089

Part 1 of 2. `Agent/$push` (and the other `waitForResponse` operations) opened a dedicated Redis subscriber connection per in-flight request, subscribed to a channel named after the callback id. This replaces that with a single subscriber per server process, listening on one channel keyed off the hostname (`agent:cb:${hostname}`), with responses demultiplexed by callback id. Connection count becomes O(1) per process instead of O(N) per in-flight push.

A callback id that yields no channel at all is malformed: the response is logged and dropped rather than published somewhere arbitrary.

### Rolling deploy

The initiator of a request and the process that owns the agent's WebSocket are usually different pods, so mid-deploy the two can be on different releases. Only one direction actually breaks — a new initiator subscribed to `agent:cb:${hostname}` whose response is published by an old process, which writes to the callback id verbatim — but both sides are doubled up here so the schemes fully overlap:

- **Write:** `publishAgentCallback` publishes to the derived per-hostname channel *and* to the callback id itself. (For an old-style id, `Agent/<id>-<uuid>`, the two are the same channel and it publishes once.)
- **Read:** the initiator adds the callback id to the shared subscriber for the lifetime of the request and drops it when the request settles. That's an extra channel on the existing connection, not an extra connection, so the O(1) connection count still holds.

#10582 removes both once this has been released and fully rolled out.


The-Alchemist:
@claude review

The-Alchemist:
@ThatOneBro : is claude's comment about "Wire-format change is not backward-compatible during a rolling upgrade" relevant?  WDYT?

ThatOneBro:
Yeah it's definitely relevant-- a known issue. I think this will have to be a multi-PR progression with 2 minor version bumps which is why I haven't pushed on it much for now. I think coming up in this next wave of agent PR's I'll start pushing on this one for the "write old protocol and read from both" PR that will be split off from this one. Just so we can coordinate other things that might require a minor bump

packages/server/src/fhir/operations/utils/agentutils.ts
```diff
@@ -174,35 +179,19 @@ export async function publishAgentRequest<T extends AgentResponseMessage = Agent
...
+    // Tag the message with a hostname-keyed callback id so the response is routed
+    // back to this server process via the shared per-hostname callback channel,
+    // avoiding a new Redis subscriber connection per in-flight request.
+    message.callback = buildAgentCallbackId(randomUUID());
+
+    await ensureCallbackSubscriber();
+    // Rolling-deploy compatibility: a server still on the previous release publishes the
+    // response to the callback id verbatim, so listen on that channel too until every
+    // process is upgraded.
+    await subscribeLegacyCallbackChannel(message.callback);
```
mattwiller: When can this be removed?

packages/server/src/fhir/operations/utils/agentcallback.ts
```diff
@@ -0,0 +1,239 @@
...
+ * only needed until every process has been upgraded; see {@link subscribeLegacyCallbackChannel}.
+ *
+ * A callback id that yields no channel is malformed: the response is logged and dropped
+ * rather than published somewhere arbitrary.
+ *
+ * @param callback - The callback id echoed back by the agent.
+ * @param message - The serialized agent response.
+ */
+export async function publishAgentCallback(callback: string, message: string): Promise<void> {
+  const channel = getCallbackChannelFromId(callback);
```
mattwiller: It might be good to add a check here that the channel string is the expected shape here, as defense in depth against malformed callback IDs getting passed in

mattwiller (approved):

PR #9163, opened 2026-05-08
docs: Restructure our charting documentation and add additional content based on decision guide

everett-williams:
## Summary

Restructures Medplum charting documentation into a slimmer section with three primary guides (**Designing charting**, **Visit templates**, **Chart data model**), consolidates patient/encounter FHIR patterns into `chart-data-model.mdx`, and moves external file guidance to `fhir-datastore/external-documents.mdx`.

## Changes

- Replace several standalone charting pages with the consolidated model; fold SOAP workflow into visit templates.
- Broaden `structured-data-capture.md` intro (not charting-only); link to visit templates.
- Add client redirects in `docusaurus.config.ts` for legacy `/docs/charting/*` URLs.
- Update cross-links (blogs, ONC, tasks, binary-data, etc.); fix MDX heading patterns and anchor targets for a clean build.
- Relocate external-documents example to `packages/examples/src/fhir-datastore/`.


packages/docs/docs/charting/allergy-intolerances.md
```diff
@@ -1,109 +0,0 @@
-# Representing Allergies
```
codyebberson, line 1: Is there a way for us to preserve some of these deleted files without too much duplicate content?

Deleted files -> broken links -> SEO penalties

packages/docs/sidebars.ts
```diff
@@ -54,7 +54,11 @@ const sidebars: SidebarsConfig = {
       type: 'category',
       label: 'Charting',
       link: { type: 'doc', id: 'charting/index' },
-      items: [{ type: 'autogenerated', dirName: 'charting' }],
```
codyebberson, line 57: Will this still preserve a `/docs/charting` URL?  Even if it's an autogenerated index page, it's still valuable to have some page that lands there for SEO purposes.

everett-williams: Yes, there will still be an index.md landing page

packages/docs/docs/charting/allergy-intolerances.md
everett-williams: We have implemented client redirects via Docusaurus, so the old links will still resolve to the right place. I believe this shouldn't impact SEO, but let me know if that's incorrect

maddyli (approved):

mattwiller (approved):

PR #9167, opened 2026-05-08
feat(server/bots): prune old bot lambda versions

### LINKED ISSUE #9166: Prune old Lambda version for Bots on deploy
Currently we never prune old versions of deployed Lambdas for Lambda Bots, which has led to a lot of accumulation of old bot versions that have not been used but still contribute to our `Code Storage` quota for AWS Lambda. We should make sure that we clean up old versions (anything older than the newest 2 versions, or something similar) on deploy to keep our usage below the quota.

ThatOneBro:
Fixes #9166

### DIFF
--- a/packages/cdk/src/backend.ts
+++ b/packages/cdk/src/backend.ts
@@ -383,6 +383,7 @@ export class BackEnd extends Construct {
             'lambda:UpdateFunctionCode',
             'lambda:UpdateFunctionConfiguration',
             'lambda:InvokeFunction',
+            'lambda:ListVersionsByFunction',
           ],
           resources: [`arn:aws:lambda:${region}:${accountNumber}:function:medplum-bot-lambda-*`],
         }),
--- a/packages/cdk/src/backend.ts
+++ b/packages/cdk/src/backend.ts
@@ -384,6 +384,7 @@ export class BackEnd extends Construct {
             'lambda:UpdateFunctionConfiguration',
             'lambda:InvokeFunction',
             'lambda:ListVersionsByFunction',
+            'lambda:DeleteFunction',
           ],
           resources: [`arn:aws:lambda:${region}:${accountNumber}:function:medplum-bot-lambda-*`],
         }),
--- a/packages/server/src/cloud/aws/deploy.ts
+++ b/packages/server/src/cloud/aws/deploy.ts
@@ -15,7 +15,7 @@ import {
   UpdateFunctionCodeCommand,
   UpdateFunctionConfigurationCommand,
 } from '@aws-sdk/client-lambda';
-import { sleep } from '@medplum/core';
+import { normalizeErrorString, sleep } from '@medplum/core';
 import type { Bot } from '@medplum/fhirtypes';
 import { ConfiguredRetryStrategy } from '@smithy/util-retry';
 import JSZip from 'jszip';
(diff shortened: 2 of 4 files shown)


packages/cdk/src/backend.ts
```diff
@@ -383,6 +383,8 @@ export class BackEnd extends Construct {
             'lambda:UpdateFunctionCode',
             'lambda:UpdateFunctionConfiguration',
             'lambda:InvokeFunction',
+            'lambda:ListVersionsByFunction',
```
ThatOneBro, line 386: Deleting bots required the addition of these permissions for the `TaskExecutionRole`

packages/server/src/cloud/aws/deploy.ts
```diff
@@ -26,6 +28,7 @@ export const LAMBDA_HANDLER = 'index.handler';
 export const LAMBDA_MEMORY = 1024;
 export const DEFAULT_LAMBDA_TIMEOUT = 10;
 export const MAX_LAMBDA_TIMEOUT = 900; // 60 * 15 (15 mins)
+export const LAMBDA_VERSIONS_TO_KEEP = 2;
```
ThatOneBro, line 32: Keeping 2 versions just to be safe

packages/server/src/cloud/aws/deploy.ts
```diff
@@ -172,11 +175,48 @@ export async function deployLambdaInternal(
...
+  do {
+    const response = await client.send(new ListVersionsByFunctionCommand({ FunctionName: name, Marker: marker }));
+    for (const v of response.Versions ?? []) {
+      const parsed = Number(v.Version);
+      if (Number.isInteger(parsed)) {
+        versions.push(parsed);
+      }
+    }
+    marker = response.NextMarker;
+  } while (marker);
```
ThatOneBro, line 212: We page through all versions to exhaustively prune old versions. Since things will be continually pruned on each deploy, this should really only happen once, or in our case since we have already done this via a script, 0 times.

ThatOneBro (commented):
Tested on staging with the deployed CDK changes and looks like it's working well

packages/server/src/cloud/aws/deploy.ts
```diff
@@ -172,11 +175,48 @@ export async function deployLambdaInternal(
 
   if (await lambdaExists(client, name)) {
     await updateLambda(bot, client, name, zipFile);
+    await cleanupOldLambdaVersions(client, name);
```
codyebberson: 1. Do we want to do this synchronously on every deploy?  Bot deploy is already pretty slow, and this could add ~seconds.  My original assumption was that this would be an async task.
2. (Apologies for pedantic nit) One word "cleanup" is a noun.  "Clean up" verb form is [two words](https://www.merriam-webster.com/dictionary/cleanup).   For what it's worth, I'd probably prefer "delete" anyway.

ThatOneBro: Yeah originally I didn't await here but after testing, deleting even up to 10 versions was almost instantaneous on staging. I guess better to just avoid this potential hang though, I'll change it back

packages/server/src/cloud/aws/deploy.ts
```diff
@@ -172,11 +176,56 @@ export async function deployLambdaInternal(
 
   if (await lambdaExists(client, name)) {
     await updateLambda(bot, client, name, zipFile);
+    const { project } = getAuthenticatedContext();
+    // Don't block on delete since this could take a while
+    deleteOldLambdaVersions(client, name).catch((err) => {
+      globalLogger.error('Error occurred while deleting old Lambdas', {
```
codyebberson, line 182: 👍

codyebberson (approved):

codyebberson:
Thanks @ThatOneBro 🙏

PR #9170, opened 2026-05-11
fix(email): improve error messages when email feature is unavailable

### LINKED ISSUE #1246: Better error message with email feature is not enabled
At present the [sendEmail](https://www.medplum.com/docs/sdk/classes/MedplumClient#sendemail) function is not available for free accounts (to prevent spammers).  The error message that those who attempt to use it encounter is a bit cryptic and could be improved.

galenzo17:
## Summary
- Separate the combined `forbidden` check in email routes into two distinct conditions with descriptive messages
- Return `"Email feature is not enabled for this project"` when the project lacks the email feature
- Return `"Only project administrators can send emails"` (403) when the user is not a project admin
- Throw `"Email is not configured on this server"` instead of silently succeeding when no email provider is configured

## Test plan
- [ ] Verify non-admin users receive 403 with descriptive message
- [ ] Verify projects without email feature receive 400 with descriptive message
- [ ] Verify unconfigured email provider throws an error instead of silently returning 200

Closes #1246


codyebberson:
Thanks @galenzo17 - Conceptually, looks good to me.  Let us know if you'd like any help with updating the tests.

galenzo17:
@codyebberson Thanks for the feedback! I pushed a fix — sendEmail now logs a
  warning instead of throwing when email isn't configured, and moved the          
  configuration check to the route handler. This way the API endpoint still
  returns a clear error, but internal callers like inviteUser don't break.        
                                                                          
  I ran the tests locally and everything is passing (screenshot attached). Would
  you mind taking a look at the updated PR and approving the CI run when you get a
   chance? I'd really appreciate it.
                                                                                  
  Thanks again for the help! 
  
image
image

galenzo17:
Tests added — ready for another look.

PR #9174, opened 2026-05-11
feat(provider): realtime ai speech to text 

techdavidy:
Uses $ai-realtime operation and hook from https://github.com/codyebberson/medplum-ai-realtime/blob/main/src/hooks/useWhisper.ts

Screenshot 2026-05-11 at 1 06 05 PM


examples/medplum-provider/src/hooks/useWhisper.ts
```diff
@@ -0,0 +1,272 @@
...
+export type UseWhisperResult = {
+  status: WhisperStatus;
+  error: unknown;
+  transcripts: TranscriptItem[];
+  start: () => Promise<void>;
+  stop: () => void;
+  isListening: boolean;
+};
+
+export function useWhisper({
```
codyebberson, line 39: Should we put this in `@medplum/react-hooks`?

codyebberson (approved):
Overall, lgtm.  I do wonder if we should move the `useWhisper` into `@medplum/react-hooks`, but we can do that separately.

examples/medplum-provider/src/hooks/useWhisper.ts
techdavidy: yes I was thinking about that

PR #9176, opened 2026-05-11
Fix implementations of FHIRPath functions

mattwiller:
Correcting the imeplementation of "as" (both operator and function versions), as well as a small fix for the concatenation operator


packages/core/src/fhirpath/atoms.ts
```diff
@@ -155,15 +155,11 @@ export class AsAtom extends InfixOperatorAtom {
...
+    return functions.as(context, this.left.eval(context, input), this.right);
   }
 }
 
-export abstract class BooleanInfixOperatorAtom extends InfixOperatorAtom {
-  abstract eval(context: AtomContext, input: TypedValue[]): TypedValue[];
-}
-
-export class ArithemticOperatorAtom extends BooleanInfixOperatorAtom {
+export class ArithemticOperatorAtom extends InfixOperatorAtom {
```
codyebberson: `Arithemtic` 😭 

We should probably fix this typo 😭

packages/core/src/fhirpath/atoms.test.ts
```diff
@@ -46,8 +48,8 @@ describe('Atoms', () => {
   });
 
   test('ConcatAtom', () => {
-    expect(evalFhirPath('{} & {}', [])).toStrictEqual([]);
-    expect(evalFhirPath('x & y', [])).toStrictEqual([]);
+    expect(evalFhirPath('{} & {}', [])).toStrictEqual(['']);
```
codyebberson, line 51: I had my pitchfork ready for this, but apparently `&` is concatenate in this case?  I don't like it, but yeah, this looks correct 😠

codyebberson (approved):
Nice to see all of those `test.skip`'s get fixed.  Thanks 🙏

packages/core/src/fhirpath/atoms.test.ts
mattwiller: Yup, `&` (concatenate) is a weird one 😕

PR #9148, opened 2026-05-07
Add $hold Scheduling operation

noahsilas:
This is similar to $book in that it takes a result from `$find` and persists it. There are some minor differences between the two in how the input is manipulated before saving.

I'm taking advantage of the fact that we didn't implement this operation with respect to the early `Schedule/:id/$find` slot-based operation; since that original implementation we have decided that we should reduce the surface area of the API and only offer the more powerful `Appointment/$find` variant (as it can be used for both single-schedule and multi-schedule booking).

This meant that I had a clean slate to start from. I've tried to implement as much as I can in reusable helper functions so that we can easily extend this to `$book` next.

PR #9177, opened 2026-05-11
FooMedical: Use Appointment/$find + Appointment/$hold

noahsilas:
Previously this was implemented with `Schedule/:id/$find` -> `Appointment/$book` as an example of using the planned single-calendar workflow.

We always planned on moving this to `$hold` over `$book`, as that creates a tentative appointment awaiting approval from the practitioner being scheduled. We have added that endpoint in #9148, so we should start using it.

Implementation Notes:
---------------------
- The $hold implementation accepts Appointment resources coming from `Appointment/$find`. This required re-wiring some of the underlying react components that were focused on Slot usage.

- Added the ability for FooMedical to use a `.env` file to make localdev a little nicer.


codyebberson:
@claude please review

codyebberson (approved):
See Claude comments.  The mutation one might be worth fixing.:

> Bug: Shallow copy mutates original appointment
> examples/foomedical/src/pages/GetCarePage.tsx:56-60

Otherwise lgtm 👍

PR #9184, opened 2026-05-12
chore(react-hooks): Move whisper hook to react-hooks package 

techdavidy:


techdavidy:
@claude please review

examples/medplum-provider/src/pages/spaces/ChatInput.tsx
```diff
@@ -70,32 +70,46 @@ export function ChatInput({
     },
```
noahsilas, line 70: Before we publish this in `@medplum/react-hooks`, how would you feel about pushing a bunch of this work inside of the hook? It seems like the common use case would be to get the complete transcript rather than just the latest segment, so having the hook build up the full string instead of consumers needing the ref+logic seems like a win to me.

packages/react-hooks/src/useWhisper/useWhisper.ts
noahsilas: I think that if some component calls `start` multiple times things here can get sort of funky - we can construct multiple audio streams and web sockets and lose some of the references to them.

I also wonder if we should split this into a couple smaller functions. I wonder if we will want to initialize the microphone permissions separately from setting up the websocket.

Should we keep a single websocket connection alive for the life of this hook? Can it be reused for multiple audio clips if someone sends a few messages in a row?

examples/medplum-provider/src/pages/spaces/ChatInput.tsx
techdavidy: Not 100% sold on the idea to pass full transcript. The hook would be responsible to pass the chunks received from the web socket.

packages/react-hooks/src/useWhisper/useWhisper.ts
techdavidy: broke down the function.

examples/medplum-provider/src/pages/spaces/ChatInput.tsx
noahsilas: On inspection, the useWhisper hook is already collecting all the transcripts in an array, so there's another possible implementation that this consumer could use that is slightly less stateful:
```ts
const { start, stop, status, transcripts } = useWhisper({ model: 'gpt-4o-transcribe' });
const transcript = useMemo(() => transcripts.map(t => t.trim()).filter(Boolean).join(' '));
useEffect(() => onInputChange(transcript), [transcript, onInputChange]);
```

That said, I see your point that it's nice to have the shared hook be more minimal, and it's always possible for us (or any consumer) to add a wrapper to handle merging intermediate states later.

packages/react-hooks/src/useWhisper/useWhisper.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+    });
+
+    const processor = { onaudioprocess: null, connect: jest.fn(), disconnect: jest.fn() };
+    const audioContext = {
+      destination: {},
+      createMediaStreamSource: jest.fn().mockReturnValue({ connect: jest.fn() }),
+      createScriptProcessor: jest.fn().mockReturnValue(processor),
+      close: jest.fn().mockResolvedValue(undefined),
+    };
+    (globalThis as any).AudioContext = jest.fn().mockImplementation(() => audioContext);
```
noahsilas:
```suggestion
    globalThis.AudioContext = jest.fn().mockImplementation(() => audioContext);
```

noahsilas (commented):
I haven't dug in deeply to audio API stuff yet, but Sonar seems to be flagging a bunch of the APIs we are using as deprecated - is that something we should be concerned by?

packages/react-hooks/src/useWhisper/useWhisper.ts
techdavidy: I would say if recordings are infrequent it adds complexity without much win.
Probably we can review after people test it.

techdavidy:
Addressing the deprecated api in a separate PR. Want to move forward with this PR since we need to update $ai-realtime which is depending on this PR.

packages/react-hooks/src/useWhisper/useWhisper.ts
```diff
@@ -187,58 +187,56 @@ export function useWhisper({
...
-      setStatus('connecting');
-
-      const url = buildWebSocketUrl(medplum.getBaseUrl());
-      console.debug('[useWhisper] connecting to', url);
-      const websocket = new ReconnectingWebSocket(url);
-
-      websocketRef.current = websocket;
-
-      websocket.onopen = () => setStatus('connected');
+  const acquireMicrophone = useCallback(async (): Promise<MediaStream> => {
```
mattlong, line 190: Should this and `openWebSocket` be [`useEffectEvent`](https://react.dev/reference/react/useEffectEvent)? I have very little experience with them, but based on how they get called, I think it would be appropriate.

mattlong (approved):

packages/react-hooks/src/useWhisper/useWhisper.ts
techdavidy: good q, will experiment with the following pr

PR #9186, opened 2026-05-12
refactor(server): combine `systemLogger` + `globalLogger`

ThatOneBro:
For historical reasons we have both a `globalLogger` and a `systemLogger`, which seem to serve mostly the same purpose. The main difference is that `globalLogger` writes to stdout via `console.log` (which is not very performant, see: ) and `systemLogger` defaults to `LogLevel.ERROR`, which is a bit of a footgun.

The main reason why `systemLogger` defaulting to `LogLevel.ERROR` is especially pernicious is that calling `getLogger()` outside of a `RequestContext` (eg. in background jobs) will fallback to `systemLogger`, making any `getLogger().info` calls to be silently ignored in background jobs.

This PR aims to combine `globalLogger` and `systemLogger` to get the best of both worlds

Separately we have not made sure to respect backpressure when calling `process.stdout.write`, so this PR also makes sure that we await the drain promise before continuing to write to the I/O buffer


packages/server/src/fhir/operations/dbschemadiff.test.ts
```diff
@@ -15,8 +16,11 @@ describe('$db-schema-diff', () => {
     const config = await loadTestConfig();
     await initApp(app, config);
 
-    // The migration script can log a lot to the console sometimes
-    console.log = jest.fn();
+    // The migration script can log a lot sometimes
+    jest.spyOn(globalLogger, 'info').mockImplementation(() => undefined);
+    jest.spyOn(globalLogger, 'warn').mockImplementation(() => undefined);
+    jest.spyOn(globalLogger, 'error').mockImplementation(() => undefined);
+    jest.spyOn(globalLogger, 'debug').mockImplementation(() => undefined);
```
mattwiller:
```suggestion
    jest.spyOn(globalLogger, 'log').mockImplementation(() => undefined);
```

packages/server/src/workers/subscription.test.ts
```diff
@@ -1684,8 +1684,7 @@ describe('Subscription Worker', () => {
   test('Subscription -- Unexpected throw inside of satisfiesAccessPolicy (regression in #3978, see #4003)', () =>
     withTestContext(async () => {
       globalLogger.level = LogLevel.WARN;
-      const originalConsoleLog = console.log;
-      console.log = jest.fn();
+      const warnSpy = jest.spyOn(globalLogger, 'warn').mockImplementation(() => undefined);
```
mattwiller: I would also recommend just mocking out the `log` method in all these tests, unless we really only care about one level and not the others — that way the test could still pass if we decide to change the log level

mattwiller (approved):

packages/server/medplum.config.json
```diff
@@ -34,5 +34,6 @@
     "removeOnFail": { "count": 1 },
     "removeOnComplete": { "count": 1 }
   },
+  "logRequests": true,
```
mattlong: Undo this change?

mattlong (changes requested):

mattlong (approved):

mattlong (approved):

packages/server/src/logger.ts
```diff
@@ -1,23 +1,43 @@
...
+  writeLineToStdout,
   undefined,
   process.env.NODE_ENV === 'test' ? LogLevel.ERROR : LogLevel.INFO
 );
 
-function write(msg: string): void {
-  process.stdout.write(msg + '\n');
-}
-
-export const systemLogger = new Logger(write, undefined, LogLevel.ERROR);
```
codyebberson, line 16: 👋 👍

codyebberson (approved):

PR #9187, opened 2026-05-12
Lambda Cleaner uses lower concurrency and has improved logging

mattlong:


packages/server/src/cloud/aws/lambda.ts
```diff
@@ -54,7 +51,7 @@ export async function deleteOldLambdaVersions(
   stats?: DeleteOldLambdaVersionStats
 ): Promise<void> {
   const keepLatest = options.keepLatest ?? 1;
-  const deleteConcurrency = options.deleteConcurrency ?? 2;
+  const deleteConcurrency = options.deleteConcurrency ?? 1;
```
mattwiller: Should/could this reference `DEFAULT_DELETE_CONCURRENCY`?

mattwiller (approved):

PR #7437, opened 2025-09-21
Move Sonar and Coveralls back to Build action

codyebberson:
This remediates the "Dangerous Workflow" warnings flagged by the OpenSSF Scorecard:

https://scorecard.dev/viewer/?uri=github.com/medplum/medplum

image

The nature of this comes down to how and when Github Actions has access to Github Secrets.  When a PR is opened on a branch in the `medplum/medplum` repository, it has access to secrets.  When a PR is opened on a branch from a fork, it does not have access to secrets.  Sonar needs the `SONAR_TOKEN` secret.

We have tried to fix this before using a variety of configurations (see https://github.com/medplum/medplum/pull/7218 https://github.com/medplum/medplum/pull/7227 https://github.com/medplum/medplum/pull/7232).  We have been fighting with this for [almost 4 years](https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/30).  But we have come to conclusion there just isn't a safe way to support Sonar with PRs from forks due to permission issues.

Personally, I think this is quite sad.  If anything, PRs from forks benefit from Sonar the most.  Hopefully through some combination of Github Actions permissions improvements or Sonar action improvements eventually we can bring this back.

PR #9189, opened 2026-05-12
Attempt to revive coveralls PR comments

mattlong:
The comments stopped being added after https://github.com/medplum/medplum/pull/7437. Trying to figure out what's going on.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -259,7 +259,6 @@ jobs:
         uses: coverallsapp/github-action@0a51d2e0b5417d06e4ecceb534aec87defc53926 # master
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          debug: true
       - name: SonarCloud Scan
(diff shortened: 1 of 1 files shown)


ThatOneBro:
@mattlong Could it be because the token passed to coveralls action no longer has write perms?

.github/workflows/build.yml
```diff
@@ -254,9 +256,10 @@ jobs:
       - name: Coveralls GitHub Action
         if: ${{ matrix.node-version == 22 && matrix.pg-version == 14 && matrix.redis-version == 7 }}
         continue-on-error: true
-        uses: coverallsapp/github-action@09b709cf6a16e30b0808ba050c7a6e8a5ef13f8d # master
+        uses: coverallsapp/github-action@0a51d2e0b5417d06e4ecceb534aec87defc53926 # master
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          debug: true
```
mattwiller: Do we want to keep this on?

mattlong: good catch!

mattwiller (approved):

PR #9192, opened 2026-05-13
Add active consumer index to $rate-limits operation

maddyli:


codyebberson:
@claude  - please review

@ThatOneBro  - I'd appreciate your review on the Redis commands here.  This is a surgical fix, and it all makes sense on the surface.  Using raw Redis commands in handlers is a little spooky, because we're tightly coupling the feature to Redis layout, which will be tricky to manage over time.  I think it's probably ok.  Maybe something for us to noodle on for the future.

packages/server/src/fhir/operations/project-rate-limits.ts
```diff
@@ -76,16 +83,20 @@ export async function projectRateLimitsHandler(req: FhirRequest): Promise<FhirRe
     }
     memberships = reads;
   } else {
-    memberships = await ctx.repo.searchResources<ProjectMembership>({
-      resourceType: 'ProjectMembership',
-      filters: [{ code: 'project', operator: Operator.EQUALS, value: getReferenceString(project) }],
-      count: 1000,
-    });
+    const activeKey = getActiveRateLimitKey(project.id);
+    const activeMembershipIds = await redis.zrevrange(activeKey, 0, 999);
```
ThatOneBro, line 84: Do we think `msBeforeReset` is important for the dashboard? If not, we could save querying all those extra keys and just use the points in the "active" set by adding "WITHSCORES" here and just grabbing all the project memberships and building the output with those two things

ThatOneBro (commented):
Pipelining all these commands makes sense, and doesn't block the request which is good.

Overall think this works as intended but one question about an optimization we could do if `msBeforeReset` is not important

packages/server/src/fhir/operations/project-rate-limits.ts
maddyli: It is not included in the dashboard as it currently stands, but I would want the returned values from this endpoint to include the `msBeforeReset`, since I suspect users would want to use this endpoint to create their own dashboards and want might to include this information. As such, I would prefer to keep this in

packages/server/src/fhir/operations/project-rate-limits.ts
```diff
@@ -76,16 +83,20 @@ export async function projectRateLimitsHandler(req: FhirRequest): Promise<FhirRe
...
+    if (activeMembershipIds.length > 0) {
+      const reads = await Promise.allSettled(
+        activeMembershipIds.map((id) => ctx.repo.readResource<ProjectMembership>('ProjectMembership', id))
```
ThatOneBro: If we can map the ids to references first and then get memberships by doing a call to `ctx.repo.readReferences` that will mitigate a lot of the overhead

ThatOneBro (approved):
LGTM, thanks for bearing with my feedback 🙏

PR #9193, opened 2026-05-13
docs(dosespot): document self-service prescriber enrollment bot

ianplunkett:
## Summary

- Adds documentation for `dosespot-self-enroll-prescriber-bot` to the [Prescriber Enrollment guide](https://www.medplum.com/docs/integration/dosespot/enroll-user)
- Adds a comparison table at the top so readers can choose between the admin-driven and self-service approaches
- Covers: prerequisites (`PractitionerRole` setup + role type codes), auto-advancing EPCS flow (sequence diagram), Practitioner resource requirements, and input/output reference

### DIFF
--- a/packages/docs/docs/integration/dosespot/enroll-user.md
+++ b/packages/docs/docs/integration/dosespot/enroll-user.md
@@ -7,7 +7,16 @@ import TabItem from '@theme/TabItem';
 
 # Prescriber Enrollment
 
-This guide explains how to enroll a new prescriber in DoseSpot via the Medplum **Enroll Prescriber Bot**, including the full workflow for enabling Electronic Prescriptions for Controlled Substances (EPCS).
+This guide explains how to enroll a new prescriber in DoseSpot. Medplum provides two enrollment bots:
+
+| Approach | Bot | Best for |
+|---|---|---|
+| **Admin-driven** | `dosespot-enroll-prescriber-bot` | IT admin enrolls each provider manually; full control over each step |
+| **Self-service** | `dosespot-self-enroll-prescriber-bot` | Provider enrolls themselves the first time they open the DoseSpot iFrame; no back-and-forth with admin |
+
+Both bots support full EPCS enrollment (IDP + TFA). The self-service bot auto-advances through each registration stage on each call, while the admin bot requires explicit `initIdp`/`initTfa` flags.
+
+## Admin-Driven Enrollment (dosespot-enroll-prescriber-bot)
 
 :::note[Prerequisites]
 The user executing the bot must:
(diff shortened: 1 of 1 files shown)


packages/docs/docs/integration/dosespot/enroll-user.md
```diff
@@ -363,3 +372,121 @@ After each run, the bot updates the Practitioner resource with:
...
+
+## Self-Service Enrollment (dosespot-self-enroll-prescriber-bot)
+
+Self-service enrollment lets a provider enroll themselves in DoseSpot without any admin involvement. The first time a provider opens the DoseSpot iFrame, the bot runs automatically, creates the clinician record, and auto-advances through each registration stage on every subsequent call. Providers complete remaining steps (legal agreement, IDP, TFA) directly in the DoseSpot iFrame.
+
+This eliminates the multi-turn admin/provider coordination required by the admin-driven bot.
+
+### Prerequisites
+
+- The `DOSESPOT_USER_ID` project secret must be configured with an admin-level DoseSpot clinician ID (typically a Proxy/Clinic Admin user)
```
finnbergquist: Maybe note that Medplum support needs to set this up for you.

packages/docs/docs/integration/dosespot/enroll-user.md
```diff
@@ -363,3 +372,121 @@ After each run, the bot updates the Practitioner resource with:
...
+| Code | Description |
+|------|-------------|
+| `PrescribingClinician` | Can prescribe medications |
+| `ReportingClinician` | Can view/report but not prescribe |
+| `EpcsCoordinator` | Coordinates EPCS enrollment |
+| `ClinicianAdmin` | Clinic administrator |
+| `PrescribingAgentClinician` | Prescribes on behalf of another |
+| `ProxyClinician` | Proxy access |
+
+### Step 2: Enable Self-Enrollment in the Provider App
```
finnbergquist, line 423: You wouldn't really be able to change code in provider app, right?

Maybe we say something like

If you are using provider app -> you are good to go, and you re-launches of the iframe will move you through the process. I think that selfEnrollment is already set to true in useDosespotIframe() in provider app.

If you are using useDoseSpotIFrame() in your own app -> then you need to set selfEnroll to true and make sure you are on at least version 5.x.x

finnbergquist (approved):

packages/docs/docs/integration/dosespot/enroll-user.md
ianplunkett: Good call — added a note that Medplum support needs to configure this for hosted deployments.

packages/docs/docs/integration/dosespot/enroll-user.md
ianplunkett: Updated — now distinguishes between Provider App users (already works out of the box) and custom app users (need `selfEnroll: true` on `useDoseSpotIFrame` and version 5.x.x+).

PR #9194, opened 2026-05-13
Disable GIN Fastupdate on Appointment, Slot in tests

noahsilas:
While working on scheduling tests that are exercising endpoints using serializable transactions, I'm seeing CI failing irregularly with "could not serialize access due to read/write dependencies among transactions". I'm surprised because the tests shouldn't be conflicting with each other (they are touching distinct rows), but it seems like Postgres can sometimes end up locking more than I expected.

@MattLong suggested that FastUpdate could be the culprit - the default GIN index uses this feature, which can push all the updates to a table into a single pending list, which could cause this kind of problem. In this commit we turn off the fastUpdate feature for these tables in the test environment.

Includes some other scheduling test updates:
- Fixes a $book implementation issue where references to objects can leak across transaction retries.
- Improves $book tests to check `response.body.issue` before `response.status` to provide more helpful errors on test failure

### DIFF
--- a/packages/server/src/seed.test.ts
+++ b/packages/server/src/seed.test.ts
@@ -4,6 +4,8 @@ import type { Project } from '@medplum/fhirtypes';
 import { initAppServices, shutdownApp } from './app';
 import { loadTestConfig } from './config/loader';
 import { DatabaseMode, getDatabasePool } from './database';
+import type { OutputAction } from './fhir/operations/db-configure-indexes';
+import { configureGinIndexes, vacuumTable } from './fhir/operations/db-configure-indexes';
 import type { SystemRepository } from './fhir/repo';
 import { getGlobalSystemRepo } from './fhir/repo';
 import { SelectQuery } from './fhir/sql';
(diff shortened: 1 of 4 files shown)


packages/server/src/seed.test.ts
```diff
@@ -63,8 +65,17 @@ describe('Seed', () => {
...
     await withTestContext(async () => {
+      const repo = getGlobalSystemRepo();
       // Run post-deploy migrations synchronously
-      await synchronouslyRunAllPendingPostDeployMigrations(getGlobalSystemRepo());
+      await synchronouslyRunAllPendingPostDeployMigrations(repo);
+
+      const actions: OutputAction[] = [];
+      const tables = ['Appointment', 'Appointment_References', 'Slot', 'Slot_References'];
+      const client = repo.getDatabaseClient(DatabaseMode.WRITER);
+      await configureGinIndexes(client, actions, tables, { fastUpdate: false });
```
mattwiller, line 82: Would you mind adding a comment about why this is necessary / what problem we're solving here, for future us to reference?

mattwiller (approved):

PR #9205, opened 2026-05-13
Avoid duplicated code from op defs

mattlong:
OperationDefinition are prone to triggering duplicate code detection so introduce a helper to further cut down on the boilerplate.

### DIFF
--- a/packages/server/src/fhir/operations/ai.ts
+++ b/packages/server/src/fhir/operations/ai.ts
@@ -10,54 +10,57 @@ import { sendFhirResponse } from '../response';
 import { makeOperationDefinition } from './definitions';
 import { parseInputParameters } from './utils/parameters';
 
-const operation = makeOperationDefinition({ scope: 'system' }, {
-  id: 'ai',
-  url: 'https://medplum.com/fhir/OperationDefinition/ai',
-  name: 'ai',
-  code: 'ai',
-  parameter: [
-    {
-      name: 'messages',
-      use: 'in',
-      min: 1,
-      max: '1',
-      type: 'string',
-      documentation: 'JSON string containing the conversation messages array',
-    },
-    {
-      name: 'model',
-      use: 'in',
-      min: 1,
-      max: '1',
-      type: 'string',
-      documentation: 'OpenAI model to use (e.g., gpt-4, gpt-3.5-turbo)',
-    },
-    {
-      name: 'tools',
-      use: 'in',
-      min: 0,
-      max: '1',
-      type: 'string',
-      documentation: 'JSON string containing the tools array (optional)',
-    },
-    {
-      name: 'content',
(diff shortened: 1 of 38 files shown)


packages/server/src/fhir/operations/ai.ts
```diff
@@ -2,25 +2,19 @@
...
-const operation: OperationDefinition = {
-  resourceType: 'OperationDefinition',
+const operation = makeOperationDefinition({ scope: 'system' }, {
   id: 'ai',
   url: 'https://medplum.com/fhir/OperationDefinition/ai',
   name: 'ai',
-  status: 'active',
-  kind: 'operation',
   code: 'ai',
-  resource: ['Parameters'],
```
noahsilas, line 20: This looks like it was dropped, but I suspect intentionally so as `Parameters` doesn't seem like a true resource type?

packages/server/src/fhir/operations/ai.ts
```diff
@@ -2,25 +2,19 @@
...
-  system: false,
```
noahsilas, line 21: Looks like this changed, but I think that the new value (`scope: 'system'`) is correct based on the URL that this is invoked through.

packages/server/src/fhir/operations/projectinit.ts
```diff
@@ -21,49 +14,26 @@ import { getLogger } from '../../logger';
...
+const projectInitOperation = makeOperationDefinition(
+  { scope: 'type', resource: 'Project' },
+  {
+    name: 'project-init',
+    code: 'init',
+    parameter: [
+      param('in', 'name', 'string', 1, '1'),
+      param('in', 'owner', 'Reference', 0, '1'),
+      param('in', 'ownerEmail', 'string', 0, '1'),
+      param('out', 'return', 'Project', 1, '1'),
```
noahsilas, line 33: Double checking my understanding: this is compressing the code by using positional parameters instead of named keys, which makes this less likely to be flagged as duplicate code by the scanner, right?

noahsilas (approved):
Looks good to me! I think there is a minor conflict in that I just merged the $hold operation which should get updated at some point. That's an extremely safe conflict though, and I can follow up to make it conform to this style later.

packages/server/src/fhir/operations/projectinit.ts
mattlong: Yep, pretty much. See the last commit on #9154 was to get around [this Sonar blocker](https://github.com/medplum/medplum/pull/9154/checks?check_run_id=75693161311). It looks like failing report is no longer available on Sonar...

packages/server/src/fhir/operations/ai.ts
mattlong: Yes, that was my read.

PR #9209, opened 2026-05-14
fix(server): index fragment CodeSystems in CodingTable

JeffryL:
Support CodeSystem.content = fragment during terminology indexing, since fragment is a valid FHIR CodeSystem content mode and commonly used for partial terminology distributions such as SNOMED subsets.

## Summary

- `CodingTable.batchIndexResources` skipped CodeSystems with `content: 'fragment'`, so their concepts were never written to the `Coding` lookup table
- This caused `$expand` to return empty expansions for fragment CodeSystems (e.g. SNOMED CT fragments uploaded by projects), while `complete` and `example` CodeSystems worked correctly
- Fix: add `'fragment'` to the content check alongside `'complete'` and `'example'` in `coding.ts`

## Changes

- `packages/server/src/fhir/lookups/coding.ts` — one-line fix in `batchIndexResources`
- `packages/server/src/fhir/operations/expand.test.ts` — new test: _Expands ValueSet with explicit concepts from fragment CodeSystem_

## Test plan

- [x] New test `'Expands ValueSet with explicit concepts from fragment CodeSystem'` in `expand.test.ts` covers the fix: creates a `content: 'fragment'` CodeSystem, a ValueSet referencing it, and verifies `$expand` returns the expected concepts
- [x] Existing expand tests remain green
- [x] UPDATED `coding.test.ts`: renamed and updated assertion to reflect fragment CodeSystems are now indexed


JeffryL:
Closing — needs local testing first before submitting upstream.

JeffryL:
Local testing confirmed: all 46 tests in `expand.test.ts` pass with this fix applied, including the new `'Expands ValueSet with explicit concepts from fragment CodeSystem'` test case. Tested against PostgreSQL 16 + Redis 7 running in Docker.

JeffryL:
Added DCO.

JeffryL:
Fixed the failing CI test in coding.test.ts. The existing test 'Omits codings from incomplete CodeSystem resource' was asserting the old behavior (fragment CodeSystems not indexed → 0 rows expected), which is exactly what this PR fixes. Updated the test name and assertion to reflect the new correct behavior.

Local testing passed because only expand.test.ts was run to verify the fix — the pre-existing test in coding.test.ts was not run separately. CI caught it by running the full test suite. My bad!

mattwiller (approved):

PR #9210, opened 2026-05-14
Add `Appointment/:id/$confirm` operation

noahsilas:
This operation is invoked for a "pending" appointment created by `$hold`. It updates the appointment to the "booked" status, and updates all associated Slot resources to match.

### DIFF
--- a/packages/server/src/fhir/operations/confirm.test.ts
+++ b/packages/server/src/fhir/operations/confirm.test.ts
@@ -120,18 +120,6 @@ describe('Appointment/:id/$confirm', () => {
     expect(appointments[0]).toMatchObject({ id: appointment.id, status: 'booked' });
   });
 
-  test('Sets appointment status to booked', async () => {
-    const slot = await makeSlot('busy-tentative');
-    const appointment = await makeAppointment('pending', [slot]);
-
-    await request
-      .post(`/fhir/R4/Appointment/${appointment.id}/$confirm`)
-      .set('Authorization', `Bearer ${project.accessToken}`);
-
-    const updated = await systemRepo.readResource<Appointment>('Appointment', appointment.id);
-    expect(updated.status).toEqual('booked');
-  });
-
   test('Updates busy-tentative slots to busy', async () => {
     const slot = await makeSlot('busy-tentative');
     const appointment = await makeAppointment('pending', [slot]);
@@ -193,7 +181,7 @@ describe('Appointment/:id/$confirm', () => {
   });
 
   test.each(['booked', 'cancelled', 'fulfilled', 'noshow', 'entered-in-error'] as Appointment['status'][])(
-    'Returns 409 for non-confirmable status: %s',
+    'Returns 400 for non-confirmable status: %s',
     async (status) => {
       const appointment = await makeAppointment(status);
 
@@ -210,7 +198,7 @@ describe('Appointment/:id/$confirm', () => {
           },
         ],
       });
-      expect(response.status).toEqual(409);
+      expect(response.status).toEqual(400);
     }
   );
 
--- a/packages/server/src/fhir/operations/confirm.test.ts
+++ b/packages/server/src/fhir/operations/confirm.test.ts
@@ -1,0 +1,246 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { WithId } from '@medplum/core';
+import { createReference, isDefined, isResource } from '@medplum/core';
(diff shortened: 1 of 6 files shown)


packages/server/src/fhir/operations/confirm.test.ts
```diff
@@ -0,0 +1,246 @@
...
+
+describe('Appointment/:id/$confirm', () => {
+  let project: TestProjectResult<{ withAccessToken: true }>;
+  let practitioner: WithId<Practitioner>;
+  let schedule: WithId<Schedule>;
+
+  beforeAll(async () => {
+    const config = await loadTestConfig();
+    // try to be more resilient to concurrent tests touching the same tables
+    config.transactionAttempts = 5;
```
mattwiller, line 26: 👍

packages/server/src/fhir/operations/confirm.test.ts
```diff
@@ -0,0 +1,246 @@
...
+  test('Sets appointment status to booked', async () => {
+    const slot = await makeSlot('busy-tentative');
+    const appointment = await makeAppointment('pending', [slot]);
+
+    await request
+      .post(`/fhir/R4/Appointment/${appointment.id}/$confirm`)
+      .set('Authorization', `Bearer ${project.accessToken}`);
+
+    const updated = await systemRepo.readResource<Appointment>('Appointment', appointment.id);
+    expect(updated.status).toEqual('booked');
```
mattwiller: Could this assertion be combined into the previous test to avoid duplicating the test setup work?

packages/server/src/fhir/operations/confirm.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @returns The FHIR response.
+ */
+export async function appointmentConfirmHandler(req: FhirRequest): Promise<FhirResponse> {
+  const ctx = getAuthenticatedContext();
+  const appointmentId = req.params.id;
+  const updatedResources = await ctx.repo.withTransaction(
+    async () => {
+      const appointment = await ctx.repo.readResource<Appointment>('Appointment', appointmentId);
+      if (appointment.status !== 'pending' && appointment.status !== 'proposed') {
+        throw new OperationOutcomeError(conflict(`Appointment cannot be confirmed in '${appointment.status}' status`));
```
mattwiller: I think `409 Conflict` may be technically correct, but clashes a bit with the other ways we typically use that status code: I would recommend a simple `400 Bad Request` here instead

packages/server/src/fhir/operations/confirm.ts
```diff
@@ -0,0 +1,71 @@
...
+      const updatedAppointment = await ctx.repo.updateResource<Appointment>({ ...appointment, status: 'booked' });
+
+      return [updatedAppointment, ...updatedSlots];
+    },
+    { serializable: true }
+  );
+  const bundle = {
+    resourceType: 'Bundle',
+    type: 'transaction-response',
+    entry: updatedResources.map((resource) => ({ resource })),
```
mattwiller, line 67: We should also consider populating `Bundle.entry.response` and `Bundle.entry.fullUrl` to match the batch implementation and provide additional context

packages/server/src/fhir/operations/confirm.ts
```diff
@@ -0,0 +1,71 @@
...
+      assertAllLoaded(slots, 'Loading slots failed');
+
+      // Mark `busy-tentative` slots as `busy`
+      const updatedSlots = await Promise.all(
+        slots.map(async (slot) => {
+          if (slot.status === 'busy-tentative') {
+            return ctx.repo.updateResource<Slot>({ ...slot, status: 'busy' });
+          }
+          return slot;
+        })
```
mattwiller: Removing the double-nested return statements inside this closure might aid readability some:
```suggestion
        slots.map(async (slot) => slot.status === 'busy-tentative' ? ctx.repo.updateResource<Slot>({ ...slot, status: 'busy' }) : slot)
```

packages/server/src/fhir/operations/confirm.test.ts
noahsilas: The above test actually already does assert this, removing this test now. Good catch, thank you!

packages/server/src/fhir/operations/confirm.ts
noahsilas: Good idea, I should probably do this in places like $hold too (https://github.com/medplum/medplum/blob/v5.1.12/packages/server/src/fhir/operations/hold.ts#L83). I'd like to split this into a follow up and revisit across all the scheduling APIs at once, if that seems ok?

mattwiller (approved):

PR #9157, opened 2026-05-08
Add $cancel Scheduling operation

noahsilas:
Adds the `/Appointment/$cancel` FHIR operation, which cancels a booked or pending appointment by setting its status to cancelled and deleting all referenced Slot resources in a serializable transaction.

Resolves: #7691

PR #9213, opened 2026-05-14
Provider: Add "Cancel Visit" button

noahsilas:
This button invokes the endpoint added in https://github.com/medplum/medplum/pull/9157 to cancel the appointment and delete any related slots.

Not handled yet: For appointments linked to an Encounter, clicking the appointment immediately navigates you to the Encounter detail page, so there is nowhere to show this scheduling operation. In the future, we will add an intermediate step showing the appointment details drawer with a link to the encounter details. This will offer a place to add scheduling UI such as changing the time/date of the appointment. 

In the short term, this will be immediately useful for workflows such as patient-booking - a Patient creates an appointment from an app such as FooMedical, and no Encounter is created. A team member should then either Confirm (https://github.com/medplum/medplum/pull/9210) or Cancel the appointment. 

- Added a specialized `<SideDrawer />` component that includes the awkward styling needed to add a button at the bottom of the drawer.
- Adjusted a bunch of provider types to be marked as `WithId` 

Preview
======

https://github.com/user-attachments/assets/8f4b1034-aaf2-4e78-9593-3030c6cff5c4


examples/medplum-provider/src/components/SideDrawer.tsx
```diff
@@ -0,0 +1,48 @@
...
+//   <Stack justify="space-between" h="100%">
+//     <div>
+//       Top of drawer content goes here.
+//     </div>
+//     <div>
+//       Bottom of drawer content goes here.
+//     </div>
+//   </Stack>
+// </SideDrawer>
+export function SideDrawer(props: SideDrawerProps): JSX.Element {
```
codyebberson: Why did we need this over using Mantine's `<Drawer>` directly?

examples/medplum-provider/src/pages/schedule/AppointmentCancelButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+import type { JSX } from 'react';
+import { useCallback, useState } from 'react';
+import { showErrorNotification } from '../../utils/notifications';
+
+type AppointmentCancelButtonProps = {
+  readonly appointment: WithId<Appointment>;
+  onCancel?: (appointment: WithId<Appointment>) => void;
+};
+
+export function AppointmentCancelButton(props: AppointmentCancelButtonProps): JSX.Element {
```
codyebberson: Did this need a separate component?

codyebberson (commented):
Main feature is good, but this feels like a lot of code for the one event handler.

`medplum-provider` is getting a little bit heavy, so I'm admittedly more sensitive to code size at the moment.

examples/medplum-provider/src/pages/schedule/AppointmentCancelButton.tsx
noahsilas: Not necessarily, this is only used in the one place so it could be inlined. My thought was that the SchedulePage that uses it was starting to get pretty large with lots of state, and so pushing some of that down into smaller components would help keep things more local and legible.

examples/medplum-provider/src/components/SideDrawer.tsx
noahsilas: The main reason is that this adds some tricky CSS to allow pushing things like buttons to the bottom of the drawer (see examples/medplum-provider/src/components/SideDrawer.module.css).

So far we're only doing that in the one place, but it seemed likely to me that we would want to do so again in the future, so I pulled this into a reusable position.

I might make a PR against Mantine to see if they would update their drawer styles to natively make the body content fill the component this way, it would be nice not to try to reach in to the mantine drawer styles in this way 🤔

noahsilas:
@codyebberson Here's a more minimal version; it pushes the handler into the `AppointmentDetails` component (trying still to push as much as I can down from `<SchedulePage>`, which is a container for a lot of state/behavior, and re-rendering it is expensive due to the calendar). I removed the Drawer wrapper that pushes the controls into a footer position, and instead just inserted a divider with some extra whitespace. That ends up looking like this:

Screenshot 2026-05-22 at 1 51 23 PM

What do you think?

codyebberson:
@claude - please re-review

codyebberson (approved):
lgtm 👍  - your call on Claude's remaining minor issues

PR #9228, opened 2026-05-15
Fixes #9183 - offline_access when refresh_token in grant_types

### LINKED ISSUE #9183: MCP: AI clients (e.g. Claude) don't request offline_access scope, causing frequent session drops
## Problem

AI clients connecting to the Medplum MCP server via OAuth authenticate using the `authorization_code` flow but do not request the `offline_access` scope. As a result, Medplum never issues a refresh token, and the session silently dies when the access token expires (~1 hour).

This is a known gap in how MCP clients implement OAuth, as they initiate the auth flow without `offline_access`, so there is no mechanism for the server to issue or the client to use a refresh token.

## Impact

Users are forced to manually reconnect the MCP integration every ~1 hour, which is disruptive for any ongoing session.

## Proposed Solutions

**Force `offline_access` on the server side**–If the Medplum MCP OAuth flow can inject `offline_access` into the scope regardless of what the client requests, a refresh token would always be issued. The client would then be able to silently refresh without user intervention.

## Workaround (for now)

Use [Client Credentials](https://www.medplum.com/docs/auth/client-credentials) with the [Medplum CLI](https://www.medplum.com/docs/cli) for a persistent connection that doesn't rely on a browser session.

## References

- Related OAuth spec on MCP client refresh token support: https://www.redcaller.com/docs/references/mcp-client-oauth-refresh-token-support

codyebberson:
The immediate motivation here is to enable refresh tokens via `grant_types` so that Claude MCP integration can get refresh tokens.

1. Added `grantType` and `responseType` to `ClientApplication`
2. Include refresh tokens when `ClientApplication.grantType` includes `refresh_token`
3. Drive-by fix: marked `'offline'` scope as deprecated, and changed them all to `'offline_access'`, which is more standard
4. Drive-by fix: added logging for deprecated `'offline'` scope and deprecated `remember` flag
5. Drive-by fix: fixed some client test utilities to support all `ClientApplication` fields

See https://github.com/medplum/medplum/issues/9183 for more details

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -503,7 +503,7 @@ export interface BaseLoginRequest {
 export interface EmailPasswordLoginRequest extends BaseLoginRequest {
   readonly email: string;
   readonly password: string;
-  /** @deprecated Use scope of "offline" or "offline_access" instead. */
+  /** @deprecated Use "offline_access" scope instead. */
   readonly remember?: boolean;
 }
 
--- a/packages/definitions/src/fhir/r4/fhir.schema.json
+++ b/packages/definitions/src/fhir/r4/fhir.schema.json
@@ -61936,6 +61936,20 @@
           },
           "type": "array"
         },
+        "grantType": {
+          "description": "Optional OAuth grant type for the client application. This specifies the allowed grant types for the client application.",
+          "items": {
+            "$ref": "#/definitions/string"
+          },
+          "type": "array"
+        },
+        "responseType": {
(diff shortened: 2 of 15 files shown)


codyebberson:
@claude please review

ThatOneBro (approved):

ThatOneBro (commented):
Agree with Claude on casting return type of `includeRefreshToken` to a boolean. Otherwise, LGTM

We can follow up with grant type / response type enforcement later unless you want to take another pass. I can create an issue for that if we want to add that to backlog

ThatOneBro (approved):

PR #9230, opened 2026-05-16
Fixed the RECAPTCHA_SITE_KEY always being reverted to the default

### LINKED ISSUE #8615: Setting RECAPTCHA_SITE_KEY to an empty string does not disable reCAPTCHA.
medplum-app docker-entrypoint.sh treats empty RECAPTCHA_SITE_KEY as unset and falls back to the test key. When self-hosting medplum-app with the official container image, setting RECAPTCHA_SITE_KEY to an empty string does not disable reCAPTCHA.

According to #1693
- if RECAPTCHA_SITE_KEY is explicitly set to an empty string, it should remain empty
- this should allow self-hosted deployments to explicitly disable reCAPTCHA on the app side

TheDarkula:
Closes [8615](https://github.com/medplum/medplum/issues/8615).

### DIFF
--- a/packages/app/docker-entrypoint.sh
+++ b/packages/app/docker-entrypoint.sh
@@ -7,7 +7,7 @@ set -e
 : ${MEDPLUM_BASE_URL:="http://localhost:8103/"}
 : ${MEDPLUM_CLIENT_ID:=""}
 : ${GOOGLE_CLIENT_ID:=""}
-: ${RECAPTCHA_SITE_KEY:="6LfHdsYdAAAAAC0uLnnRrDrhcXnziiUwKd8VtLNq"}
+: ${RECAPTCHA_SITE_KEY="6LfHdsYdAAAAAC0uLnnRrDrhcXnziiUwKd8VtLNq"}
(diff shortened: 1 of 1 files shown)


codyebberson (approved):
This change is intentional and correct.

In POSIX shell parameter expansion:

* `${VAR:=default}` uses the default if the variable is unset *or empty*
* `${VAR=default}` uses the default only if the variable is unset

For most config values, treating empty string as “missing” is reasonable. However, for `RECAPTCHA_SITE_KEY`, an empty string is meaningful and represents “disable reCAPTCHA”.

Previously:

```sh
: ${RECAPTCHA_SITE_KEY:="...test key..."}
```

would replace an explicitly empty value with the Google test key, making it impossible for self-hosted deployments to intentionally clear/unset the site key.

Changing to:

```sh
: ${RECAPTCHA_SITE_KEY="...test key..."}
```

preserves explicitly empty values while still providing the default test key when the variable is entirely unset.

In other words:

* unset → use test key
* empty string → disable reCAPTCHA
* custom value → use custom value

Issue #8161, opened 2026-01-06
Put calls are removing the organization reference in the compartment as well as accounts


kvpendergast-unity:
Hello, 

We're using access policies to assign organizations to the meta.accounts and meta.compartment field. When the resource is created with a properly scoped access policy, the organization is added to both of these. 

However, when a resource is updated the organization reference is removed from the accounts and compartment. Our understanding is that the expected behavior is that the organization is retained in updates and cannot be changed/replaced/removed. 

Are we misunderstanding this functionality? Or is it possible there is a bug here? For reference, we believe we've traced the issue here

https://github.com/medplum/medplum/blob/8bffe8d686b93753ac1a9c72c6227533f049f50c/packages/server/src/fhir/repo.ts#L2016

the getAccounts method doesn't appear to build accounts on update for resources that aren't part of a patient compartment. 

Thanks for looking.

kvpendergast-unity:
After spending more time on this with the team, it seems we assumed that because compartment could not be set on updates that it also could not be removed on updates. 

Does this align with the intended purpose of this feature? For now, we've resolved by explicitly passing in the compartment/accounts on the PUT call in the transaction bundle where it gets executed.

nate-watkins:
Here's a repro script to show the behavior. For what it's worth, we've discussed internally modifying our fork to change the behavior of `getAccounts` to keep existing accounts with the presence of a request header (potentially `x-medplum-retain-accounts`) and passing `canWriteAccount` check if the updated resource does not specify account/accounts at all.

The current behavior affects us most in ETL flows where we repeatedly make upsert requests `on-behalf-of` a 'scoped' ClientApplication, which has a `ProjectMembership.accessPolicy` where the access policy has an `AccessPolicy.compartment`, typically referencing an Organization for an MSO-like model.

ETA: I don't think this is a bug. Seems to be working as intended. I do think the current behavior can be slightly unintuitive for workflows that leverage `upsertResource`/`conditionalUpdate`. The 'magic' accounts/compartment propagation via AccessPolicy.compartment is _extremely_ useful, at least for our use case. I suspect the current guidance would be to follow-up with a subsequent `$set-accounts` operation, although having the option to intentionally avoid losing the accounts and avoid the additional request would be nice to have.

``` typescript
/**
 * Creates an org, access policy, and client app, then performs two identical upserts on the same Observation.
 * First upsert should have proper accounts references, second upsert loses them.
 */
[...]

PR #9255, opened 2026-05-20
Build compartments from user-writeable meta.accounts

mattwiller:
In order to support more complex MSO use cases and access control, the behavior of the `meta.accounts` field and related derived fields will change:
- `AccessPolicy.compartment` values will still be added to `meta.accounts` on resource creation only — but not when updating existing resources
- `meta.accounts` will no longer be re-derived on resource updates to pull in accounts from linked Patients: only values explicitly written to the field will be used
  - There is still an edge case for Project Admins, where modifying any part of a resource's `meta` field without explicitly carrying forward the previous value of `meta.accounts` will clear `meta.accounts`, due to the ambiguity inherent in FHIR `update`
- `meta.compartment` will continue to be re-derived on every write to combine the references in `meta.accounts` with eligible Patient references from the resource, as well as any accounts present on those linked Patients
- The requirement that a user must be a Project Admin to write to `meta.accounts` is reinforced, ensuring that regular users cannot set the `meta.accounts` field when creating or updating resources

The net effect of this change is that references written to `meta.accounts` — especially those populated by `AccessPolicy.compartment` — will be more resilient.  The final `meta.compartment` values computed by the server should not be changing at all.

In combination with this change, we should likely also change the `$set-accounts` operation to only rebuild `meta.compartment` instead of writing to `meta.accounts` for all compartment resources.  That work can be sequenced after this initial change is merged


nate-watkins:
Nice- think this resolves #8161

The-Alchemist:
@mattwiller looks like the merge conflicts are pretty straightforward.

LGTM otherwise

The-Alchemist:
@claude explain what compartments are, and review this PR

PR #9256, opened 2026-05-20
fix(server): derive bare-id reference prefix from SearchParameter.target (closes #9254)

### LINKED ISSUE #9254: Bare-UUID reference search is FHIR R4 non-compliant — only `patient`/`subject` resolve; `encounter`, `practitioner`, etc. silently return zero results
### Summary

Medplum's reference search hard-codes the bare-UUID → typed-reference fixup to two column names (`patient`, `subject`). Every other single-target reference search parameter (`encounter`, `practitioner`, `location`, `requester`, `performer`, `recorder`, `owner`, `author`, `asserter`, …) treats a bare UUID as a literal string and silently returns an empty bundle, even though the FHIR R4 spec defines bare `[id]` as the standard form for these params.

`packages/server/src/fhir/search.ts → buildReferenceSearchFilter`:

```ts
values = values.map(
  (v) => !v.includes("/") && (impl.columnName === "subject" || impl.columnName === "patient") ? `Patient/${v}` : v
);
```

### Why this is a FHIR R4 spec-compliance bug (not an Epic-specific quirk)

[FHIR R4 §3.1.1.4.12 — reference](https://www.hl7.org/fhir/R4/search.html#reference) defines the interpretation of a reference search parameter as:

> - `[parameter]=[id]` — the logical [id] of a resource using a local reference
> - `[parameter]=[type]/[id]` — **for when the reference can point to different types of resources** (e.g. `Observation.subject`)
> - `[parameter]=[url]` — an absolute URL

The spec explicitly limits the `[type]/[id]` form to **multi-target** reference params. For **single-target** params (the vast majority of reference search params — `Encounter.patient`, `DeviceRequest.encounter`, `PractitionerRole.practitioner`, `Encounter.location`, etc.), bare `[id]` is the documented, standard form. The same section even calls this out for the `patient` param:

> When using the patient search parameter, there is no need to specify `:Patient` as a modifier, or `Patient/` in the search value, as this must always be true.

That logic applies identically to every other single-target reference search param. Medplum hard-codes `patient`/`subject` and silently breaks the spec-compliant form for everything else.

### Repro (against any Medplum server with seeded data)

```http
# Works — `patient` is hard-coded
GET /Encounter?patient=07d3b8a9-6c5e-5a8f-8a3b-bb8b8a935e40

# All return zero with bare UUIDs, even though they have matching data:
GET /DeviceRequest?encounter=f8f7d662-c0ed-51d1-b479-1d76f38bc941
GET /ServiceRequest?encounter=f8f7d662-c0ed-51d1-b479-1d76f38bc941
GET /Condition?encounter=f8f7d662-c0ed-51d1-b479-1d76f38bc941
GET /PractitionerRole?practitioner=52b2d25f-13f3-542a-9f45-9c01e3f5a8c4
GET /Encounter?location=4394d513-a467-581b-8282-50fddb91f100

# All work if you prepend the type prefix:
GET /DeviceRequest?encounter=Encounter/f8f7d662-c0ed-51d1-b479-1d76f38bc941
```

### Interop impact

Every other R4 server I've checked accepts bare UUIDs for single-target reference params per the spec:

| Server | Bare-UUID single-target ref search |
|---|---|
| HAPI FHIR (JPA reference implementation) | ✅ Accepts |
| Microsoft Azure Health Data Services (HAPI-derived) | ✅ Accepts |
| IBM FHIR Server | ✅ Accepts |
| Smile CDR (HAPI-based) | ✅ Accepts |
| Aidbox | ✅ Accepts |
| Epic | ✅ Accepts (e.g. [`DeviceRequest.Search`](https://fhir.epic.com/Specifications?api=10243)) |
| **Medplum** | ❌ Returns empty bundle |

A spec-compliant FHIR client that works against any other server silently misbehaves against Medplum: the query returns HTTP 200 OK with `Bundle.total = 0` instead of an OperationOutcome, so callers (a) don't know anything went wrong and (b) don't know to retry with a `Type/id` prefix.

### Suggested fix

Two options, in order of preference:

1. **Derive the prefix from the search parameter's `target` list when the param has exactly one target type.** This is the spec-correct, future-proof fix — handles every resource type without further maintenance and avoids growing a hard-coded list.
2. **Extend the hard-coded list** to cover every documented single-target reference column (mirrors the current shape but is brittle and likely to drift).

### Local workaround we're using

We patch the bundled `packages/server/dist/index.js` at container start to extend the column list. Snippet:

```js
values = values.map((v) => {
  if (v.includes("/")) return v;
  const p = { patient: "Patient", subject: "Patient", encounter: "Encounter", practitioner: "Practitioner", location: "Location" }[impl.columnName];
  return p ? p + "/" + v : v;
});
```

This is obviously not a real solution — it's a one-off bandaid for our local-dev stack. Happy to send a PR for option 1 (derive-from-targets) if that's the preferred direction.

---

AbhiMsft:
## What

Fixes #9254.

Reference search parameters with a single target resource type now accept bare ids on the wire — the spec-compliant form per [FHIR R4 §3.1.1.4.12](https://www.hl7.org/fhir/R4/search.html#reference). Previously, the bare-id → typed-reference fixup in `buildReferenceSearchFilter` was hard-coded to the `subject` and `patient` columns, so spec-compliant queries like `DeviceRequest?encounter=<uuid>` silently returned an empty bundle.

## How

1. `ColumnSearchParameterImplementation` gains an optional `singleTargetType: ResourceType` field, populated at impl-build time when:
   - `searchParam.type === 'reference'`
   - `impl.type !== SearchParameterType.CANONICAL` (canonical refs intentionally excluded)
   - `searchParam.target?.length === 1`

   This mirrors the existing `searchParam.target?.length === 1` pattern in `parseChainLink` (search.ts:1872) for chained search.

2. `buildReferenceSearchFilter` now consults `impl.singleTargetType`:
   - If the value contains `/`, leave it alone (already a typed reference, URN, etc.).
   - If `singleTargetType` is set and the value is a UUID (`isUUID(v)`), prepend `Type/`. The `isUUID` guard avoids rewriting unrelated slashless values (URNs, identifiers, non-Medplum ids).
   - Otherwise, preserve the existing `subject`/`patient` → `Patient/` fallback verbatim for multi-target back-compat (e.g. `Observation.subject` targets [Patient, Group, Device, Location]).

## Why this design

- **Spec compliance.** R4 §3.1.1.4.12 defines bare `[id]` as the standard form and reserves `[type]/[id]` for multi-target params. The spec even calls this out for `patient`: > _"When using the patient search parameter, there is no need to specify `:Patient` as a modifier, or `Patient/` in the search value, as this must always be true."_ The same logic applies to every other single-target reference param.
- **No behaviour change for existing callers.** The legacy `subject`/`patient` fallback is preserved bit-for-bit. Multi-target params still default to `Patient/` for bare values; nothing that worked before stops working.
- **Pre-computed.** The target is resolved once at impl-build time, not per-query.
- **Canonical refs excluded.** Canonical reference params store URLs, not `Type/id` literals, so prefixing would corrupt them.
- **UUID gate on the new path.** Medplum stores its own resource ids as UUIDs; gating on `isUUID` keeps the new behaviour scoped to the canonical bug shape (bare UUIDs) and prevents URN/identifier collateral.

## Out of scope

- `buildReferenceEqualsCondition` (called only from `searchByReferenceImpl`) is intentionally unchanged. Its existing callers pass typed `Type/id` strings via `getReferenceString`, and the per-refValue mutation pattern at search.ts:194 would need a separate refactor to safely participate in normalization. Happy to do that in a follow-up if you'd like.

## Tests

New `Reference search bare-id prefix derivation` describe in `search.test.ts` covers:

- `DeviceRequest?encounter=<uuid>` — single-target, the canonical bug repro
- `DeviceRequest?encounter=Encounter/<uuid>` — typed-prefix control (still works)
- `PractitionerRole?practitioner=<uuid>` — single-target, non-Patient
- `Encounter?location=<uuid>` — single-target on an **array** reference column
- `Observation?subject=<bare-uuid>` — back-compat, multi-target subject still defaults to Patient
- `DeviceRequest?encounter=not-a-uuid` — non-uuid bare value is not rewritten (returns empty rather than corrupting the filter)

## Checks

- `npx eslint` on the changed files: clean
- `npx prettier --check` on the changed files: clean
- `@medplum/server` builds successfully (full `tsc` clean for the package)

### DIFF
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -26,6 +26,7 @@ import type {
   Communication,
   Composition,
   Condition,
+  DeviceRequest,
   DiagnosticReport,
   Encounter,
   EvidenceVariable,
@@ -38,6 +39,7 @@ import type {
   Patient,
   PlanDefinition,
   Practitioner,
+  PractitionerRole,
   Project,
   Provenance,
   Questionnaire,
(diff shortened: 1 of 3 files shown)


codyebberson (approved):
Nice, clean, thanks @AbhiMsft

AbhiMsft:
@ThatOneBro - Could you please help help review this PR?

codyebberson:
@AbhiMsft  - sorry about the delay.  We'll get this merged today.

codyebberson:
Thanks @AbhiMsft !

codyebberson:
Reverted for now to restore compatibility and unblock the customer workflow.

After digging deeper, I think the underlying behavior change from this PR is technically correct per the FHIR spec, specifically around reference search parameter handling for `Task.subject` / `Task.for`.

However, it also exposed that we had longstanding implicit shorthand behavior where:

```text
subject={id}
```

was being accepted as equivalent to:

```text
subject=Patient/{id}
```

even though `Task.for` is `Reference(Any)` and therefore ambiguous without a resource type qualifier.

The revert is intended as a short-term stability measure while we decide how we want to handle backward compatibility here.

Potential next steps:

* Restore the stricter/spec-correct behavior and require fully-qualified references
* Preserve legacy shorthand matching for backwards compatibility
* Potentially scope the compatibility behavior only to cases where the target search parameter resolves to a single resource type in practice

I think we should treat this as a compatibility/design discussion rather than a correctness bug in the PR itself.

PR #9259, opened 2026-05-20
feat(server): delete Lambda when a Bot is deleted

### LINKED ISSUE #9168: Remove old Lambdas upon Bot deletion
Whenever a Bot resource is deleted, we should check if the given bot has been deployed, and if it has, we should remove the associated Lambda in order to prevent orphaned Bots from taking up too much Code Storage quota in AWS Lambda

ThatOneBro:
Fixes #9168

### DIFF
--- a/packages/server/src/cloud/aws/deploy.ts
+++ b/packages/server/src/cloud/aws/deploy.ts
@@ -3,6 +3,7 @@
 import type { GetFunctionConfigurationCommandOutput, LambdaClient } from '@aws-sdk/client-lambda';
 import {
   CreateFunctionCommand,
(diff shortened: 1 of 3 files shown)


ThatOneBro:
Tested on staging, looks like it's working as expected 👍

mattwiller (approved):
Would it be possible to make a single delete call, without checking if the lambda already exists?

ThatOneBro:
> Would it be possible to make a single delete call, without checking if the lambda already exists?

I think that is possible, we can just check if `ResourceNotFoundException` is thrown 👍

techdavidy (approved):

PR #9260, opened 2026-05-20
Serialize transaction state transitions

### LINKED ISSUE #9244: v5.1.11: RepositoryConnection sharing race condition causes crash-loop (PR #9082 regression)
## Summary

Medplum server v5.1.11 crash-loops due to a race condition in the shared `RepositoryConnection` introduced by PR #9082 ("Consistent Repository connection state tracking and sharing"). The server crashes within ~3 minutes of startup, regardless of external load.

## Environment

- **Medplum version**: `medplum/medplum-server:5.1.11`
- **Infrastructure**: AWS ECS Fargate
- **Database**: Aurora PostgreSQL 16.13 (RDS)
- **Previous stable version**: `5.1.10` (no crashes)

## Reproduction

1. Deploy `medplum/medplum-server:5.1.11` with default configuration
2. Allow WebSocket subscription clients to connect
3. Server crashes within ~3 minutes with `Uncaught exception: terminating connection due to idle-in-transaction timeout`

No external load is required — the crash occurs with only health check traffic and WebSocket subscription bindings.

### 1. PostgreSQL server logs (RDS) — at the exact crash timestamp

```
2026-05-19 10:39:35 UTC — WARNING: there is already a transaction in progress (port 48730)
2026-05-19 10:39:35 UTC — WARNING: there is no transaction in progress (port 48754)
2026-05-19 10:39:35 UTC — WARNING: there is already a transaction in progress (port 48738)
2026-05-19 10:39:35 UTC — WARNING: there is no transaction in progress (port 48754)
... (9+ connections affected simultaneously)
```

30 seconds later:
```
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48736)
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48732)
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48730)
2026-05-19 10:40:05 UTC — FATAL: terminating connection due to idle-in-transaction timeout (port 48740)
```

### 2. Node.js runtime warning (same timestamp)

```
(node:1) DeprecationWarning: Calling client.query() when the client is already executing a query is deprecated and will be removed in pg@9.0.
```

### 3. Medplum application error

```
ERROR: Database error: "Not in transaction"
  at _RepositoryConnection.assertInTransaction (dist/index.js:32202)
  at _RepositoryConnection.commitTransaction (dist/index.js:32139)
  at _RepositoryConnection.withTransaction (dist/index.js:32082)
  at _Repository.writeToDatabase (dist/index.js:39453)
  at _Repository.handleStorage (dist/index.js:39428)
  at _Repository.updateResourceImpl (dist/index.js:39372)
```

### 4. Final crash

```
ERROR: Uncaught exception thrown: "terminating connection due to idle-in-transaction timeout"
  at parseErrorMessage (pg-protocol/dist/parser.js:305)
```

## Root Cause Analysis

PR #9082 moved `transactionDepth` from per-`Repository`-instance into a shared `RepositoryConnection` object. When multiple concurrent async operations access the same `RepositoryConnection` (e.g., via `repo.getSystemRepo()` or internal server processes), the shared `transactionDepth` counter becomes corrupted:

1. **Operation A** calls `beginTransaction()` → sets `transactionDepth = 1`, issues `BEGIN`
2. **Operation B** (same `RepositoryConnection`) calls `beginTransaction()` → issues `BEGIN` on the **same PoolClient** that already has a transaction → PostgreSQL warns: "there is already a transaction in progress"
3. One operation's error path calls `abortTransaction()` → resets `transactionDepth = 0`
4. The other operation calls `commitTransaction()` → `assertInTransaction()` throws "Not in transaction" (because depth is now 0)
5. The orphaned `BEGIN` (from the duplicate) leaves connections idle in transaction state
6. After 30s, PostgreSQL kills these connections → uncaught exception → process crash

### Why v5.1.10 didn't crash

In v5.1.10, `transactionDepth` was a **per-Repository-instance** field. Even when `getSystemRepo()` shared the same `PoolClient`, each instance maintained its own counter, so they couldn't corrupt each other's transaction tracking.

### Supporting context

- PR #4776 ("each repo can only do one query at a time") explored guarding against this exact class of bug in July 2024 but was never merged.
- The `pool.on('error')` handler from PR #8987 properly handles dead connections, but cannot prevent the crash because the `transactionDepth` corruption causes an uncaught `OperationOutcomeError` to propagate out of the Express error handler before the idle timeout fires.

## Timeline (from a single server instance)

| Time (UTC) | Event |
|---|---|
| 10:36:35 | Server starts (v5.1.11) |
| 10:36:52 | WebSocket subscription connections established |
| 10:39:35 | pg deprecation warning + "Not in transaction" errors cascade |
| 10:39:35 | PostgreSQL: "already a transaction in progress" / "no transaction in progress" |
| 10:40:05 | PostgreSQL FATAL: kills 4 idle-in-transaction connections |
| 10:40:05 | Server crash: Uncaught exception |

## Expected behavior

Concurrent operations that share a `RepositoryConnection` should not corrupt each other's transaction state. Either:
- Serialize access to the shared connection (mutex/queue), or
- Prevent concurrent `beginTransaction()` on a connection that already has an active transaction (fail fast with a clear error instead of issuing a duplicate `BEGIN`)

## Workaround

Downgrade to `medplum/medplum-server:5.1.10` which contains the dead connection pool fix (PR #8987) but does not include the `RepositoryConnection` sharing change (PR #9082).

mattlong:
This is the first of two steps to improve transaction management. Serializes the critical moments of transaction management within RepositoryConnection such that competing `withTransaction` cannot clobber eachother.

Fixes #9244

### DIFF
--- a/packages/server/src/fhir/repo.test.ts
+++ b/packages/server/src/fhir/repo.test.ts
@@ -1703,6 +1703,238 @@ describe('FHIR Repo', () => {
     ]);
   });
 
+  test('withTransactionStateLock serializes concurrent transaction begins', async () => {
+    const beginIssued = Promise.withResolvers<undefined>();
+    const allowBegin = Promise.withResolvers<undefined>();
+    const finishFirstTransaction = Promise.withResolvers<undefined>();
+    const secondCallbackStarted = Promise.withResolvers<undefined>();
+    const queries: string[] = [];
+    let beginCount = 0;
+    const query = jest.fn(async (sql: string) => {
+      queries.push(sql);
+      if (sql === 'BEGIN ISOLATION LEVEL REPEATABLE READ' && ++beginCount === 1) {
+        // Pause the first BEGIN before beginTransaction can publish transactionDepth = 1.
+        beginIssued.resolve(undefined);
+        await allowBegin.promise;
+      }
+      return { rows: [] };
+    });
+    const client = {
+      query,
+      release: jest.fn(),
+    } as unknown as PoolClient;
+    const repo = getShardSystemRepo(
+      'test-shard',
+      RepositoryConnection.borrowClient(client, { mode: DatabaseMode.WRITER })
+    );
+
+    const tx1 = repo.withTransaction(async () => {
+      await finishFirstTransaction.promise;
+    });
+    await beginIssued.promise;
+
+    // Start a second transaction while the first transaction is suspended in BEGIN. Without the state
+    // lock, this second call can observe transactionDepth = 0 and incorrectly issue another BEGIN.
+    const tx2 = repo.withTransaction(async () => {
+      secondCallbackStarted.resolve(undefined);
+    });
+    // Let the second transaction run any queued promise continuations. If it is not blocked by the
+    // lock, it will append its own SQL before this assertion.
+    await allowPendingMicrotasks();
+
+    // The second begin must wait until the first BEGIN has completed and published transactionDepth.
+    expect(queries).toStrictEqual(['BEGIN ISOLATION LEVEL REPEATABLE READ']);
+
+    allowBegin.resolve(undefined);
+    await secondCallbackStarted.promise;
+    finishFirstTransaction.resolve(undefined);
+    await Promise.all([tx1, tx2]);
+
+    expect(queries).toStrictEqual([
+      'BEGIN ISOLATION LEVEL REPEATABLE READ',
+      'SAVEPOINT sp2',
+      'RELEASE SAVEPOINT sp2',
+      'COMMIT',
+    ]);
+  });
+
+  test('withTransactionStateLock serializes concurrent transaction commits', async () => {
+    const firstStarted = Promise.withResolvers<undefined>();
+    const secondStarted = Promise.withResolvers<undefined>();
+    const finishTransactions = Promise.withResolvers<undefined>();
+    const releaseSavepointIssued = Promise.withResolvers<undefined>();
+    const allowReleaseSavepoint = Promise.withResolvers<undefined>();
+    const queries: string[] = [];
+    let releaseSavepointCount = 0;
+    const query = jest.fn(async (sql: string) => {
+      queries.push(sql);
+      if (sql === 'RELEASE SAVEPOINT sp2' && ++releaseSavepointCount === 1) {
+        // Hold the inner commit in the database call before transactionDepth is decremented.
+        releaseSavepointIssued.resolve(undefined);
+        await allowReleaseSavepoint.promise;
+      }
+      return { rows: [] };
+    });
+    const client = {
+      query,
+      release: jest.fn(),
+    } as unknown as PoolClient;
+    const repo = getShardSystemRepo(
+      'test-shard',
+      RepositoryConnection.borrowClient(client, { mode: DatabaseMode.WRITER })
+    );
+
+    const tx1 = repo.withTransaction(async () => {
+      firstStarted.resolve(undefined);
+      await secondStarted.promise;
+      await finishTransactions.promise;
+    });
+    await firstStarted.promise;
+
+    const tx2 = repo.withTransaction(async () => {
+      secondStarted.resolve(undefined);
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/repository/repository-connection.ts
```diff
@@ -193,15 +193,16 @@ export class RepositoryConnection implements Disposable {
   ): Promise<TResult> {
     this.assertNotClosed();
     const isolationLevel = options?.serializable ? 'SERIALIZABLE' : 'REPEATABLE READ';
-    this.assertCompatibleTransactionIsolationLevel(isolationLevel);
```
mattlong, line 196: This was moved within a serialized block; not deleted outright.

mattlong:
@claude review

packages/server/src/fhir/repository/repository-connection.ts
```diff
@@ -261,100 +264,147 @@ export class RepositoryConnection implements Disposable {
...
+   * - release connection state
+   * @param callback - The callback to execute with the transaction state lock.
+   * @returns Passthrough result of the callback.
+   */
+  private async withTransactionStateLock<T>(callback: () => Promise<T>): Promise<T> {
+    // capture the current tail of the queue
+    const previous = this.transactionStateLock;
+
+    // create a new unresolved promise and make it the new queue tail immediately
+    const { promise, resolve } = Promise.withResolvers<undefined>();
```
noahsilas, line 287: #TIL about Promise.withResolvers - neat!

packages/server/src/fhir/repo.test.ts
```diff
@@ -1703,6 +1703,238 @@ describe('FHIR Repo', () => {
...
+    let beginCount = 0;
+    const query = jest.fn(async (sql: string) => {
+      queries.push(sql);
+      if (sql === 'BEGIN ISOLATION LEVEL REPEATABLE READ' && ++beginCount === 1) {
+        // Pause the first BEGIN before beginTransaction can publish transactionDepth = 1.
+        beginIssued.resolve(undefined);
+        await allowBegin.promise;
+      }
+      return { rows: [] };
+    });
```
noahsilas, line 1721: Nice, this is quite a clever way to demonstrate the existing problem. Getting all the synchronization right to do that is quite tricky!

packages/server/src/fhir/repo.test.ts
```diff
@@ -1703,6 +1703,238 @@ describe('FHIR Repo', () => {
...
+    await secondCallbackStarted.promise;
+    finishFirstTransaction.resolve(undefined);
+    await Promise.all([tx1, tx2]);
+
+    expect(queries).toStrictEqual([
+      'BEGIN ISOLATION LEVEL REPEATABLE READ',
+      'SAVEPOINT sp2',
+      'RELEASE SAVEPOINT sp2',
+      'COMMIT',
+    ]);
```
noahsilas, line 1758: @mattlong and I Talked about this in person - this test is demonstrating the current state of the world and one way it is possible to get into an undesirable state.

My understanding: it's incorrect for two independent transactions to end up creating a savepoint inside a single transaction; the intention is for this to happen only when `withTransaction` helpers are nested.

So this test is showing that currently an un-awaited `withTransaction`, or possibly something like systemRepo (which might be invoked from multiple requests/jobs concurrently), can end up in this bad state.

@mattlong is going to follow up to correct this behavior.

noahsilas (approved):

packages/server/src/fhir/repository/repository-connection.ts
ThatOneBro: Same

packages/server/src/fhir/repo.test.ts
mattlong: #9285 changes the behavior of this call pattern

PR #9261, opened 2026-05-20
fix(provider): Filter selfpay coverage from visit eligibility check

techdavidy:


examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
```diff
@@ -62,9 +63,9 @@ export function EncounterCoverageEligibilityModal(props: EncounterCoverageEligib
     medplum
       .searchResources('Coverage', `patient=Patient/${patient.id}&status=active`)
       .then((results) => {
-        const sorted = results.sort((a, b) => (a.order ?? 99) - (b.order ?? 99));
```
mattlong, line 65: No sorting needed?

mattlong (approved):

examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
techdavidy: no, we would show it as it was created in the db

PR #9263, opened 2026-05-20
Automatically provision default patient access policy in new projects

maddyli:


packages/server/src/fhir/operations/projectinit.ts
```diff
@@ -155,3 +162,39 @@ export async function createProject(
...
+async function createDefaultPatientAccessPolicy(
+  systemRepo: SystemRepository,
+  project: WithId<Project>
+): Promise<WithId<AccessPolicy>> {
+  return systemRepo.createResource<AccessPolicy>({
+    resourceType: 'AccessPolicy',
+    meta: { project: project.id },
+    name: 'Default Patient Access Policy',
+    compartment: { reference: '%patient' },
+    resource: [
```
ThatOneBro, line 174: We might prefer to use `_compartment=%patient` for a lot of these just for the sake of standardization. It could be that some of these won't be in the patient compartment though so it's probably not a silver bullet

packages/server/src/fhir/operations/projectinit.ts
```diff
@@ -141,6 +142,12 @@ export async function createProject(
     description: 'Default client for ' + project.name,
   });
 
+  const accessPolicy = await createDefaultPatientAccessPolicy(systemRepo, project);
+  project = await systemRepo.updateResource<Project>({
```
mattlong: We should get in the habit of using patch over update when possible

packages/server/src/auth/newpatient.test.ts
```diff
@@ -88,74 +88,31 @@ describe('New patient', () => {
       login: res4.body.login,
       projectId: projectId,
     });
-    expect(res5.status).toBe(400);
-
-    // As Christina, create a default access policy for new patients
```
mattlong, line 93: We should still test that patient registration fails if someone were to delete the default patient access policy

mattlong (changes requested):

mattlong (approved):
Do we have a plan for what to do when we want to make some change to the default access policy and wish it were retroactively applied to existing projects? I suspect the answer is that there's probably not much that could reasonably be done; nor would we want to change permissions without warning since a project could be relying on the old behavior...

maddyli:
Yup, I think we just wouldn't touch those existing policies, especially since its provisioned within the project and we expect some users will edit it. This change is also strictly forward-looking for a similar reason; if folks have already implemented patient default access policies, they should exist as is.

PR #9265, opened 2026-05-20
Add lab account number to provider lab order form

AndyStoneman:
Simple change to include the lab account number as a value you can pass in when ordering labs via Medplum Provider.

### DIFF
--- a/examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
+++ b/examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
@@ -75,6 +75,7 @@ export function OrderLabsPage(props: OrderLabsPageProps): JSX.Element {
     updateBillingInformation,
     setSpecimenCollectedDateTime,
     setOrderNotes,
+    setPerformingLabAccountNumber,
     createOrderBundle,
   } = labOrderReturn;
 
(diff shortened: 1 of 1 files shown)


examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
```diff
@@ -177,6 +178,10 @@ export function OrderLabsPage(props: OrderLabsPageProps): JSX.Element {
               performingLab={performingLab}
               error={createError?.validation?.performingLab}
             />
+            <TextInput
+              label="Lab account number"
+              onChange={(e) => setPerformingLabAccountNumber(e.currentTarget.value || undefined)}
```
techdavidy, line 183: is this something we can auto populate  ?
like we do LabCorp when ordering labs from encounter ?

finnbergquist: Yea, this makes sense. I would propose an Organization search over the Account Number identifier system:

{
      "type": {
        "coding": [
          { "code": "AN" }
        ]
      },
      "value": "04843980"
    }

as outlined https://developer.healthgorilla.com/docs/billing-and-accounts#requester-and-account-numbers

finnbergquist (changes requested):

PR #9280, opened 2026-05-21
feat(provider): Use operations instead of bots for claim submit

techdavidy:
- Replaced bot-driven submissions with FHIR operations. submitClaim now invokes
  POST /Claim/{id}/$candid-submit-claim and submitToStedi invokes POST 
  /Claim/{id}/$stedi-submit-claim, each posting an empty Parameters body. Stedi
  submission still updates the claim's insurance via updateResource first.

  - Unified post-submit state via ClaimResponse. Removed the old Candid executeBot +
   readResource dance and the Stedi executeBot path. The operation returns a
  ClaimResponse; if present it is set directly on state, otherwise we fall back to
  fetchClaimResponse.


techdavidy:
@claude please review again

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
```diff
@@ -95,14 +95,10 @@ interface ClaimReviewPanelProps {
...
   insuranceCoverages: WithId<Coverage>[];
   selectedCoverage: WithId<Coverage> | undefined;
   selfPayCoverage: WithId<Coverage> | undefined;
   initialBillingType: BillingType;
-  showCandidButton?: boolean;
-  showStediButton?: boolean;
-  stediSubmitting?: boolean;
   onClose: () => void;
   onSubmitClaim: (coverages: Reference<Coverage>[]) => void;
   onSubmitToStedi?: (insurance: Reference<Coverage>[]) => void;
```
noahsilas, line 104: How would you feel about renaming `onSubmitClaim` to `onSubmitToCandid`? That would make these two entries more parallel, and it would be much more obvious to me as a newcomer that I should expect one or the other to be invoked. (As written now, my first thought was that `onSubmitClaim` would be invoked for either integration.)

examples/medplum-provider/src/components/encounter/BillingTab.tsx
```diff
@@ -542,73 +395,49 @@ export const BillingTab = (props: BillingTabProps): JSX.Element => {
...
+          <SubmitClaimModal
+            opened={confirmModalOpen}
+            coverages={coverages}
+            selectedCoverage={coverage}
+            patient={patient}
+            conditions={conditions}
+            practitioner={practitioner}
+            onClose={() => setConfirmModalOpen(false)}
+            onSubmitClaim={submitToCandid}
+            onSubmitToStedi={submitToStedi}
```
noahsilas, line 425: This prop is sort of serving double-duty - it's also now used to decide if the "submit to stedi" button is shown. Should we be doing something to check if Stedi is enabled here? I'm imagining something like:

```ts
            onSubmitToStedi={stediEnabled ? submitToStedi : undefined}
```

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
```diff
@@ -272,44 +251,37 @@ const ClaimReviewPanel = (props: ClaimReviewPanelProps): JSX.Element => {
...
-            size="md"
-            rightSection={<IconArrowUpRight size={16} />}
-            loading={submitting}
-            disabled={!canSubmit || !!stediSubmitting}
-            onClick={handleConfirm}
-          >
-            Submit to Candid
-          </Button>
-        )}
+        <Button
```
noahsilas, line 264: Should this be gated on the presence of the `onSubmitClaim` prop? It seems like before this "Submit to Candid" button could be hidden but now it is always visible, is that intentional?

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
```diff
@@ -159,33 +151,20 @@ const ClaimReviewPanel = (props: ClaimReviewPanelProps): JSX.Element => {
...
   };
-  const handleConfirm = (): void => {
+
+  const resolveCoverages = async (): Promise<Reference<Coverage>[]> => {
     if (billingType === 'self-pay') {
-      onSubmitClaim(selfPayCoverage ? [createReference(selfPayCoverage)] : []);
-    } else {
-      onSubmitClaim(insuranceCoverages.filter((c) => selectedIds.has(c.id)).map(createReference));
+      const coverage = ensureSelfPayCoverage ? await ensureSelfPayCoverage() : selfPayCoverage;
+      return coverage ? [createReference(coverage)] : [];
```
noahsilas: It looks like this component is only used from the BillingTab, which will always supply the `ensureSelfPayCoverage` prop. We can probably simplify this a bit in that case by marking that prop as required:

```suggestion
      const coverage = await ensureSelfPayCoverage();
      return [createReference(coverage)];
```

examples/medplum-provider/src/components/encounter/BillingTab.tsx
```diff
@@ -325,104 +235,79 @@ export const BillingTab = (props: BillingTabProps): JSX.Element => {
...
           notifications.show({
             color: 'red',
             icon: <IconCircleOff />,
             title: 'Error',
-            message: errorMessage,
+            message: parsed?.errorMessage,
           });
         } catch {
-          showErrorNotification(err);
+          showErrorNotification(normalizeErrorString(err));
```
noahsilas: showErrorNotification already passes its argument into `normalizeErrorString()` internally, so you don't need to do it here.

```suggestion
          showErrorNotification(err);
```

examples/medplum-provider/src/components/encounter/ClaimSubmittedPanel.tsx
```diff
@@ -2,21 +2,33 @@
...
-  candidEncounterId?: string;
+  claimResponse: ClaimResponse | Reference<ClaimResponse>;
   exportMenu: ReactNode;
 }
 
-export const ClaimSubmittedPanel = (props: ClaimSubmittedPanelProps): JSX.Element => {
-  const { status, claimAmount, createdAt, candidEncounterId, exportMenu } = props;
+export const ClaimSubmittedPanel = (props: ClaimSubmittedPanelProps): JSX.Element | null => {
+  const { claimResponse, exportMenu } = props;
+  const claimResponseResource = useResource(claimResponse);
```
noahsilas, line 20: Something for us to think about for when we can make future breaking changes w/ Medplum v6: I think this hook should return a tuple `[resource, loading, outcome]`, similar to `useSearch*`, so that components like this can easily show a `<Loader />` or similar when appropriate.

noahsilas (commented):
partial review, will pick this back up and finish first thing in the morning.

examples/medplum-provider/src/components/encounter/BillingTab.tsx
techdavidy: for now we assumed both are enabled.
We creating a single path and the server handles it.

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
techdavidy: Soon we are calling $submit. 
We will update here.

examples/medplum-provider/src/components/encounter/ClaimSubmittedPanel.tsx
techdavidy: I am onboard

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
noahsilas: Ah, so this is an intermediate step and the idea is to merge these two operations into a single operation later? I see, thanks for the clarification.

examples/medplum-provider/src/components/encounter/BillingTab.tsx
```diff
@@ -129,89 +101,27 @@ export const BillingTab = (props: BillingTabProps): JSX.Element => {
     fetchCoverage().catch((err) => showErrorNotification(err));
   }, [medplum, patient]);
 
-  useEffect(() => {
-    medplum
-      .searchOne('Bot', { identifier: 'https://medplum.com/integrations/candid-health|send-to-candid' })
-      .then((bot) => setBillingBot(bot ?? null))
-      .catch(() => setBillingBot(null));
-  }, [medplum]);
```
noahsilas, line 137: Double checking my understanding of the strategy in play here: originally these effects would get bots with specific integration identifiers. If those bots existed, we could show UI like a "Send claim to Stedi" button, and that button would call `medplum.executeBot()` with that bot ID.

In the new version, that bot is registered as a [custom FHIR operation](https://www.medplum.com/docs/bots/custom-fhir-operations). This client code always shows the "Send claim to Stedi" and "Send claim to Candid" buttons, and clicking one tries to invoke an operation like `Claim/:id/$candid-submit-claim`. If that operation is not registered, that will return a 404 error.

So this cuts down on some round trips and some load time, at the expense of requiring some extra setup (new `OperationDefinition` resources required). In the intermediate state we also show both "submit" buttons, but the plan is to unify those into a single button that will call another new `$submit` operation which will itself invoke the appropriate integration behind the scenes.

Did I understand the plan here correctly? Is there something special happening to register the custom operations in all the right projects?

examples/medplum-provider/src/components/encounter/ClaimSubmittedPanel.tsx
mattlong: https://github.com/medplum/medplum/issues/9350

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
mattlong, line 172: Should `canSubmit` consider whether there is an active submit request still in flight?

Also, if the claim was already submitted, it seems like you might want to prevent accidentally submitting it again.

mattlong (approved):
Generally LGTM. This will be further consolidated after https://github.com/medplum/medplum/pull/9295 lands, correct?

examples/medplum-provider/src/components/encounter/SubmitClaimModal.tsx
techdavidy: There is a variable that handles submitting that disables the button and proceed in the workflow.

PR #9281, opened 2026-05-21
Provider: Navbar visual tweaks + HeaderDropdown Refactor

kevinwadeshaw:
This PR updates elements of our global Navbar to create visual consistency with other elements across the app—more specifically the Auth project selection screen and the other dropdown/context menus we use for filters, etc.

Changes include:
- updated Link border radius to match those of the new List component
- cursor change for opening/closing Navbar
- refactor of HeaderDropdown
  - includes same visual representation of Projects we use for Auth
  - creates visual consistency across Project representation, with selected state for current Project
  - adds Appearance as menu item with updated light/dark/system toggle

---

Screenshot 2026-05-21 at 3 39 33 PM

Screenshot 2026-05-21 at 4 07 30 PM


packages/react/src/AppShell/HeaderDropdown.module.css
```diff
@@ -0,0 +1,75 @@
+.divider {
+  border-color: light-dark(var(--mantine-color-gray-2), var(--mantine-color-dark-4));
+}
+
+.accountSection {
+  padding: var(--mantine-spacing-xs) var(--mantine-spacing-sm);
```
noahsilas, line 6: For my learning: what factors in to your decisions to place styles in css modules like this vs. as style props like `px="sm" py="xs"`?

packages/react/src/auth/membership.utils.ts
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { ProjectMembership } from '@medplum/fhirtypes';
+
+export function getMembershipLabel(membership: ProjectMembership): string | undefined {
+  return membership.identifier?.find((i) => i.system === 'https://medplum.com/identifier/label')?.value;
```
noahsilas: This looks like it could use core's `getIdentifier` helper:

```suggestion
  return getIdentifier(membership, 'htps://medplum.com/identifier/label');
```

packages/react/src/AppShell/Navbar.module.css
```diff
@@ -201,13 +202,17 @@
...
   display: flex;
   align-items: center;
   justify-content: center;
   padding: 0;
   transition: background-color 100ms ease;
-  cursor: ew-resize;
+  cursor: e-resize;
+
+  &[data-opened] {
+    cursor: w-resize;
```
noahsilas, line 214: This is a very nice little detail 😁

packages/react/src/auth/ProjectLoginOption.tsx
```diff
@@ -0,0 +1,47 @@
...
+export function ProjectLoginOption(props: ProjectLoginOptionProps): JSX.Element {
+  const { projectDisplay, profileDisplay, label, selected } = props;
+  const projectTitle = label ? `${projectDisplay} - ${label}` : projectDisplay;
+
+  return (
+    <Group gap="xs" align="center" wrap="nowrap">
+      <Box className={classes.iconBox}>
+        <IconBriefcase size={16} stroke={2} />
+      </Box>
+      <div style={{ flex: 1, minWidth: 0 }}>
```
noahsilas, line 25: `minWidth: 0` wasn't obvious to me - I originally thought "`flex: 1` includes `flex-shrink: 1`, so won't this naturally have an appropriate minimum width?".

After poking at it for a minute though I realized: this is supporting the `truncate` flag in its children - without this, the flexbox default is `min-width: auto`, which expands to contain its children. In that case the children wouldn't get truncated with ellipsis the way we want.

I don't have any suggestions for ways to make this clearer, so this comment is just me writing down what I learned here for the benefit of anyone following along. 👍

noahsilas (approved):
Looks great!

packages/react/src/auth/membership.utils.ts
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { ProjectMembership } from '@medplum/fhirtypes';
+
+export function getMembershipLabel(membership: ProjectMembership): string | undefined {
```
techdavidy: I think we can keep the function in the original file

packages/react/src/AppShell/Navbar.tsx
```diff
@@ -23,6 +23,7 @@ import { MedplumLink } from '../MedplumLink/MedplumLink';
 import { ResourceAvatar } from '../ResourceAvatar/ResourceAvatar';
 import { ResourceTypeInput } from '../ResourceTypeInput/ResourceTypeInput';
 import { HeaderDropdown } from './HeaderDropdown';
+import headerDropdownClasses from './HeaderDropdown.module.css';
```
techdavidy, line 26: Description only mentions changes in the nav bar. Are the changes during log in too?

packages/react/src/AppShell/HeaderDropdown.module.css
kevinwadeshaw: Good question! This could honestly go either way, but I included it in the separate css module to group with the other HeaderDropdown styling rules. The idea being that If we ever wanted to tweak this further, we could make any edits just from this css file.

packages/react/src/auth/ProjectLoginOption.tsx
kevinwadeshaw: This was a Claude suggestion—definitely non-obvious but seems appropriate for getting the truncating right!

packages/react/src/auth/membership.utils.ts
kevinwadeshaw: Ok! I'll revert and use `getIdentifier` per @noahsilas' comment https://github.com/medplum/medplum/pull/9281#discussion_r3289964628

packages/react/src/AppShell/Navbar.tsx
kevinwadeshaw: Sorry, yes, I copied the updated Project icon/text color changes we made in the HeaderDropdown back to the Project login auth screen to keep these consistent. These changes are for [ProjectLoginOption.module.css](https://github.com/medplum/medplum/pull/9281/changes/BASE..cd649f0a001e114ddd1216156257bd8da19e2987#diff-87e05d8fc62a0ca9348fba4f0468d5772d5c1d1a1381fa96edfc596fbc47a2f7). Let me know if I should break this out into a separate PR or update the description!

kevinwadeshaw:
@techdavidy this is ready for another review btw—main changes are here:
https://github.com/medplum/medplum/pull/9281/changes/e3259bed44b055965156e29a0b70ff071800e5c5

kevinwadeshaw:
For anyone following along, this last commit (https://github.com/medplum/medplum/pull/9281/commits/54e6fbce39a4e6eab8147927492aaffa2c6201b1) adds support for the version toggle on the core Medplum app (which still has our legacy header as a layout option). Here's what that looks like:

**v2:**
Screenshot 2026-05-29 at 9 26 01 AM
Screenshot 2026-05-29 at 9 27 09 AM

**v1:**
Screenshot 2026-05-29 at 9 26 18 AM

techdavidy:
@claude please review

techdavidy:
@claude please review. And verify there are no regression when switching active logins from the dropdown menu

techdavidy (approved):

PR #9283, opened 2026-05-22
Added order set bundle and import button to provider

AndyStoneman:


examples/medplum-provider/src/data/order-set-example-bundle.json
```diff
@@ -0,0 +1,301 @@
...
+                {
+                  "system": "https://medplum.com/CodeSystem/order-set-scope",
+                  "code": "shared",
+                  "display": "Shared (practice-wide)"
+                }
+              ]
+            }
+          }
+        ],
+        "description": "Sample **shared** order set for type 2 diabetes in older adults.\n\n- Metformin 500 mg BID with meals (renal gate).\n- Empagliflozin 10 mg each morning (lower SGLT2 dose for age ≥75).\n- Sitagliptin 25 mg daily (DPP-4; low hypoglycemia risk).\n\n**Sync:** POST this bundle to Medplum, then invoke `scriptsure-orderset-sync-bot` with `{ \"planDefinitionId\": \"<id>\" }`. Do not pre-stamp `https://spa.scriptsure.com/orderset-id` — the bot adds it after the first successful sync.",
```
oleg-mp: A small caveat here - this is a DAW-specific set of instructions, so might want to remove for the sake of keeping it vendor-neutral, while we can have vendor specific instructions elsewhere

examples/medplum-provider/src/data/order-set-example-bundle.json
```diff
@@ -0,0 +1,301 @@
+{
+  "resourceType": "Bundle",
+  "type": "collection",
+  "meta": {
```
oleg-mp: This part might need to go, as we're adding it as a vendor-neutral bundle

examples/medplum-provider/src/data/order-set-example-bundle.json
```diff
@@ -0,0 +1,301 @@
...
+    "tag": [
+      {
+        "system": "https://medplum.com/scriptsure/demo-export",
+        "code": "orderset-sync-demo",
+        "display": "Shared PlanDefinition + ActivityDefinitions for scriptsure-orderset-sync-bot (no ScriptSure orderset-id yet)"
+      }
+    ]
+  },
+  "identifier": {
+    "system": "https://medplum.com/scriptsure/demo-export",
```
oleg-mp: we may wanna figure out a more generic one that doesn't mention scriptsure

oleg-mp (approved):
🚢

PR #9284, opened 2026-05-22
Use batched searches for preferred pharmacies and orgs for med prescribing

oleg-mp:


examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -1337,67 +1340,66 @@ export function OptionalContextFields(props: Readonly<OptionalContextFieldsProps
...
     }
-    let cancelled = false;
-    medplum
-      .searchResources('Coverage', `patient=Patient/${patient.id}&status=active&_count=50`)
-      .then((rows) => {
-        if (!cancelled) {
-          setPatientCoverages(rows.filter((c): c is WithId<Coverage> => Boolean(c.id)));
-        }
+    const orgById = new Map(
+      (organizations ?? []).filter((org): org is WithId<Organization> => Boolean(org.id)).map((org) => [org.id, org])
```
noahsilas, line 1378: `organizations` looks like it is typed as `ResourceArray<WithId<Organization>> | undefined`, so I think you can skip the entire `filter` part of this - if there are any entries, they should be guaranteed to have an `id` attribute.

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -1337,67 +1340,66 @@ export function OptionalContextFields(props: Readonly<OptionalContextFieldsProps
...
+        .filter(isDefined)
+        .join(','),
+    [preferredPharmacyRefs]
+  );
+
+  const [organizations, , organizationsOutcome] = useSearchResources(
+    'Organization',
+    { _id: preferredPharmacyOrgIds, _count: String(Math.max(preferredPharmacyRefs.length, 1)) },
+    { enabled: preferredPharmacyOrgIds.length > 0 }
+  );
```
noahsilas, line 1371: This feels like a pattern that we should see if we end up using in more places and consider if it would fit into its own hook. Loading from multiple references seems likely to be a common pattern.

One minor thing here: this is going to require `Organization.search` permissions instead of the previous implementation's `Organization.read` permission. Probably always fine, but something to be aware of.

There's also some slight difference in caching behavior; previously each reference would be individually passed into `medplum.readReference`, and so could be cached individually. This search will cache the entire collection together. Also probably fine for this use case (seems likely either all or none would already be cached).

noahsilas (approved):

PR #9285, opened 2026-05-23
Transaction scoped repos

mattlong:
### Update

Consider starting with the docstring of `RepositoryConnection.withTransaction` which explains the permissions/usability for repository transactions.

### Original description

Prevents unexpected transaction interleaving. Consider the following code. Note the absence of `await`. Previously, this surprisingly would result in nested transactions instead of two serial transactions.

```typescript
repo.withTransaction(() => { ... });
repo.withTransaction(() => { ... });
```

### Before
```sql
BEGIN ISOLATION LEVEL REPEATABLE READ;
SAVEPOINT sp2;
RELEASE SAVEPOINT sp2;
COMMIT;
```

### After
The second `repo.withTransaction()` throws.

### What's different

To achieve this, the parent transaction cannot be used within the `withTransaction` callback. Instead, a new Repository instance is provided as a parameter to the `withTransaction` callback:

```typescript
repo.withTransaction(async (txRepo) => {
  await txRepo.createResource(...);
  txRepo.preCommit(() => { ... });
  await txRepo.withTransaction(async (nestedRepo) => {
    await nestedRepo.updateResource(...);
    nestedRepo.postCommit(() => { ... });
  });
});
```

**Attempting to use a parent repo while it has an active transaction scope will throw**. An eslint rule was added to facilitate this change in behavior.

### Alternative approach

Instead of explicit transaction-scoped repositories, an alternative solution was to synchronize the beginning and ending of each layer of transaction/savepoints with a stack of transaction locks facilitated by AsyncLocalStorage. it would avoid the API change of needing to provide transaction-scoped repositories, but at the expense of a more implicit magic in the Repository class that becomes even harder to reason about. Keeping the API explicit seemed like the better choice. If there is strong dislike for the current proposal, we can explore the alternative approach more in-depth.

This PR is chained on to #9260.


packages/server/src/fhir/repo.ts
```diff
@@ -296,13 +305,15 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
     if (!this.context.author?.reference) {
       throw new Error('Invalid author reference');
     }
+    this.guardRepositoryMethods();
```
mattlong: Alternatively, we could consider only explicitly adding the guard to `Repository.getDatabaseClient` and all the caching methods. The weakness of that approach is that it's opt-in instead of opt-out like this one is which leaves a bit of risk for future additions to the Repository API.

packages/server/src/fhir/repo.ts
```diff
@@ -2227,6 +2254,54 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
+        const original = descriptor.value as (...args: unknown[]) => unknown;
+        const isAsyncFunction = original.constructor.name === 'AsyncFunction';
+        Object.defineProperty(this, propertyKey, {
+          configurable: true,
+          value: function (this: Repository, ...args: unknown[]): unknown {
+            try {
+              this.assertNotClosed();
+              this.assertAvailable();
+            } catch (err) {
+              if (isAsyncFunction) {
```
ThatOneBro: Since this will be static for the given method, can we factor this out into two static paths?
```
value: isAsyncFunction ? asyncVersion : syncVersion
```

packages/server/src/test-matchers.ts
```diff
@@ -0,0 +1,71 @@
...
+      toEqualUnordered(expected: readonly unknown[]): any;
+    }
+    interface InverseAsymmetricMatchers {
+      toEqualUnordered(expected: readonly unknown[]): any;
+    }
+  }
+}
+
+expect.extend({
+  toEqualUnordered(received: unknown, expected: readonly unknown[]) {
```
ThatOneBro, line 31: Nice 👍

packages/server/src/fhir/operations/codesystemimport.ts
```diff
@@ -103,7 +104,8 @@ export async function codeSystemImportHandler(req: FhirRequest): Promise<FhirRes
   }
 
   try {
-    await repo.withTransaction(async (db) => {
+    await repo.withTransaction(async (txRepo) => {
+      const db = txRepo.getDatabaseClient(DatabaseMode.WRITER);
```
ThatOneBro, line 108: I think that `db` here being either `Pool` or `PoolClient` instead of only `PoolClient` does change assumptions about whether importCodeSystem is in a transaction or not, right?

EDIT: I don't think it will ever be that a `txRepo` will return `Pool` from `getDatabaseClient` but the types say that it can. Can we narrow the type somehow within a `withTransaction` block so we can avoid adding `Pool` to typing across a lot of the functions that we added it to in this PR?

ThatOneBro (commented):
Looks great for the most part, just one point about type changes that are a little misleading in regards to what kind of connection we get while in a transaction

packages/server/src/fhir/operations/codesystemimport.ts
mattlong: Attempting to enforce proper usage by narrowing its input to `PoolClient` can be thwarted by `getDatabasePool(...).connect()`. That's a little contrived but from the type system, it demonstrates the issue.

Option 1:
If `importCodeSystem` should always run within a transaction, it should ensure that itself. The most straightforward way to achieve that today would be for it to take a `Repository` instead of a `Pool/PoolClient` as its input and wrap its contents in `repo.ensureInTransaction`. That'd require changing the `LookupTable.batchIndexResources` signature and other lookup table implementations. This feels directionally the way we should go by relying on `Repository` as the clearinghouse for database context/routing logic. We should push the Repository dependency as low as possible and only grab a DB connection when necessary. In fact, #9343 takes this even further with the move towards `Repository.executeSql` and effectively deprecates `Repository.getDatabaseClient` altogether. Perhaps it should be formally deprecated in that PR...

Option 2:
Add a proper `TransactionClient` type/interface that extends PoolClient and use it instead of `PoolClient` in functions where we want that guarantee. We could either add back that client as a parameter in the withTransaction/ensureInTransaction callbacks and/or add a `Repository.getTransactionClient` that either returns a `TransactionClient` if in a transaction or throws.

To me this doesn't feel quite right since it detracts from functional composability. At the end of the day, there's no reason `importCodeSystem`, `processProperties`, `resolveProperty`, and other lower level functions involved couldn't technically operate outside of a transaction; they just need a way to issue queries against a DB and that's all they should care about in their signature.

codyebberson: Yeah, personally, not a fan of Option 2.

> can be thwarted by getDatabasePool(...).connect()

True, but any calls to `getDatabasePool()` should be viewed with immediate high scrutiny.

I think I see what @ThatOneBro  is saying.

`Repository` used to be generic with the connection type, right?  It looks like that has been removed.  I don't have context on how/when/why that happened.  If it was still generic, then `txRepo` could be type `Repository<PoolClient>`, `db` would be type `PoolClient`, and everyone would be happy?  Not to trivialize it, that might be more complicated than it's worth.

mattlong: The `TClient` generic wasn't particularly useful since the return type of `Repository.getDatabaseClient` is a hardcoded `Pool | PoolClient`.

The `TClient` generic goes away in this PR since the only place it was being used was in the withTransaction/ensureInTransaction callbacks which got replaced with the transaction-scoped repository. We can add back `client: PoolClient` as a second parameter in the transaction callbacks, but that really feels like a step in the wrong direction since there would be two ways to get at the DB connection: directly with `client: PoolClient` and indirectly with `txRepo.getDatabaseClient(): Pool | PoolClient`.

mattlong: Updated proposal after iterating on this a lot.

Move away from our existing pattern of using `Pool | PoolClient` as a type. In Medplum's codebase, it's exclusively shorthand for something that can be used to perform `client.query(sql, ...)`. Let's just codify that in a new type:

```typescript
type PgQueryable = Pick<Pool, 'query'> & Pick<PoolClient, 'query'>;
```

`PgQueryable` replaces  all usages of `Pool | PoolClient` throughout the codebase and becomes the default return type of `Repository.getDatabaseClient()` (currently returns `Pool | PoolClient`) by bringing back the `TClient` generic to `Repository`. Transaction scoped repositories refine `TClient` to `PoolClient` reflecting their bound connection:

```typescript
class Repository<TClient extends PgClientLike = PgClientLike> extends FhirRepository implements Disposable {
  getDatabaseClient(mode: DatabaseMode): TClient;

  // `& this` to preserve the `SystemRepository` subclass distinction from `Repository`
  private createTransactionScopedRepo(): TransactionRepository & this
}

type TransactionRepository = Repository<PoolClient>;
```

With this:
* `PgQueryable` becomes the default type used in functions that are not particularly concerned with if its being called in a transaction or not; i.e. functions that previously used `Pool | PoolClient`.
* `PoolClient` can be used to effectively enforce a function to only be used within a transaction. I maintain that this is detrimental to the composability of our code. We have `Repository.ensureInTransaction` for exactly the purpose of guaranteeing a transaction at the proper abstraction level; all child function shouldn't care what sort of DB client they're operating on.

codyebberson:
@claude - can you please take an extra close look at:

* `packages/server/src/fhir/repo.ts`
* `packages/server/src/fhir/repository/repository-connection.ts`

In particular, can we take a fine toothed comb looking for corner cases of race conditions.  In the past, we have occasionally had very tricky bugs because certain state values were not set atomically because of an `await` in the middle of a state transition block.

mattlong:
@claude Take another look. The critical issue has been addressed.

mattlong:
@claude one more time. chained precommits and stale transaction-scoped repos have been addressed. Not planning on addressing the `withStatementTimeout` issue at this time since it is so rarely used.

mattlong:
@claude I addressed the withStatementTimeout issue. Everything should be fixed except the contrieved statle transaction repo edge case.

packages/server/src/fhir/repo.ts
```diff
@@ -1094,8 +1161,11 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       }
     }
 
-    await this.batchWriteLookupTables(conn, resources, false);
-    await this.batchWriteResources(conn, resources);
+    await this.ensureInTransaction(async (txRepo) => {
```
ThatOneBro, line 1167: 👍

packages/server/src/fhir/repo.ts
```diff
@@ -1187,25 +1262,26 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       }
 
       if (!this.isCacheOnly(resource)) {
-        await this.ensureInTransaction(async (conn) => {
+        await this.ensureInTransaction(async (txRepo) => {
           const columns = buildDeletedResourceRow(resourceType, id, resource.meta?.project);
 
-          await new InsertQuery(resourceType, [columns]).mergeOnConflict().execute(conn);
+          const client = txRepo.getDatabaseClient(DatabaseMode.WRITER);
```
ThatOneBro, line 1266: 👍

mattlong:
@claude take another look. the implementation has changed quite a bit since you last reviewed. For context, see the docstring from `RepositoryConnection.withTransaction`

packages/server/src/fhir/repository/repository-connection.ts
```diff
@@ -171,29 +259,139 @@ export class RepositoryConnection implements Disposable {
...
   ): Promise<TResult> {
-    this.assertNotClosed();
-    this.assertOwnsClient();
-    if (this.transactionDepth > 0) {
-      throw new Error('Cannot set statement timeout during an active transaction');
-    }
+    const client = await this.withConnectionStateLock(async () => {
+      this.assertNotClosed();
+      this.assertOwnsClient();
+      if (this.transactionDepth > 0) {
```
ThatOneBro: Can `this.transactionDepth` go away now that we're tracking depth with scope?

ThatOneBro (approved):
We did a synchronous review offline, LGTM.

A few small questions but nothing blocking at this point

PR #9286, opened 2026-05-23
Fixes #9269 - GC IP userinfo support

### LINKED ISSUE #9269: Medplum Token Exchange Support for GC IP
## Background

Medplum's token exchange flow (RFC 8693) validates an external IdP token by calling the configured \`userInfoUrl\` using a hard-coded OIDC contract:

```
GET <userInfoUrl>
Authorization: Bearer <subject_token>
```

This works for standard OIDC providers (Auth0, Okta, Google OAuth). It does **not** work for **Google Cloud Identity Platform (GCIP) / Firebase email-password authentication**, which issues its own ID tokens that can only be verified via:

```
POST https://identitytoolkit.googleapis.com/v1/accounts:lookup?key=<API_KEY>
Content-Type: application/json

{ "idToken": "<firebase-idToken>" }
```

GCIP is a common choice for patient-facing applications built on Medplum, where a standard OIDC provider (Auth0, Okta) handles practitioners/admins and GCIP handles patients. Currently there is no way to configure a \`ClientApplication\` to use GCIP for token exchange without deploying an external adapter service.

## Root cause

`getExternalUserInfo` in `packages/server/src/oauth/utils.ts` hard-codes:
- HTTP method: always `GET`
- Token placement: always `Authorization: Bearer <token>`

Two additional mismatches with GCIP:
- GCIP requires an API key as a query parameter (`?key=<API_KEY>`)
- GCIP returns `{ "users": [{ "email": "...", "localId": "..." }] }` rather than a flat claims object

## Success criteria

Token exchange for GC IP will need to be able to support GC IP unique contract, instead of standard userinfo url:
```
POST https://identitytoolkit.googleapis.com/v1/accounts:lookup?key=<API_KEY>
Content-Type: application/json

{ "idToken": "<firebase-idToken>" }
```

codyebberson:


codyebberson:
@claude please review

ThatOneBro (approved):

PR #9288, opened 2026-05-23
fix(cdk): forward loadBalancerLoggingPrefix to ALB access logs

### LINKED ISSUE #9287: @medplum/cdk: `loadBalancerLoggingPrefix` is silently ignored; ALB access logs are written under a prefix equal to the bucket name
## Summary

In `@medplum/cdk`, `BackEnd.createLoadBalancer` is supposed to forward the
configured ALB access-log key prefix to `loadBalancer.logAccessLogs(bucket, prefix)`,
but both call sites pass `config.loadBalancerLoggingBucket` (the bucket **name**)
as the `prefix` argument. As a result, ALB access-log objects are written under
an S3 key prefix equal to the bucket name, and the value users configure in
`loadBalancerLoggingPrefix` is never read.

## Root cause

`packages/cdk/src/backend.ts` declares `createLoadBalancer` with a
`loadBalancerLoggingPrefix: string | undefined` parameter and matching JSDoc
(L966, L980), and uses it correctly inside the method:

```ts
// packages/cdk/src/backend.ts
private createLoadBalancer(
  stackName: string,
  namePrefix: string,
  port: number,
  internetFacing: boolean,
  certificateArn: string,
  loadBalancerLoggingPrefix: string | undefined,   // ← expects the prefix
  ...
)

if (this.loggingBucket) {
  loadBalancer.logAccessLogs(this.loggingBucket, loadBalancerLoggingPrefix);
}
```

But both call sites pass `config.loadBalancerLoggingBucket` as that argument:

```ts
// L552 — main ALB
this.loadBalancer = this.createLoadBalancer(
  config.stackName,
  '',
  config.apiPort,
  config.apiInternetFacing !== false,
  config.apiSslCertArn,
  config.loadBalancerLoggingBucket,   // ← should be config.loadBalancerLoggingPrefix
  ...
);

// L568 — optional mTLS ALB
this.mtlsLoadBalancer = this.createLoadBalancer(
  ...
  config.mtlsSslCertArn,
  config.loadBalancerLoggingBucket,   // ← same bug
  ...
);
```

The `loadBalancerLoggingPrefix` property is already declared on
`MedplumInfraConfig` (`packages/core/src/config.ts:268`) and on `BaseConfig`
(`:95`), so operators can — and do — set it. It is simply never read.

## Observed behaviour

With a config like:

```jsonc
{
  "loadBalancerLoggingBucket": "medplum-logs-us-east-1",
  "loadBalancerLoggingPrefix": "elb"
}
```

`npx cdk synth <stack> -c config=<config>.json | grep -A1 access_logs.s3.prefix`
yields:

```
- Key: access_logs.s3.prefix
  Value: medplum-logs-us-east-1
```

Logs land under `s3://medplum-logs-us-east-1/medplum-logs-us-east-1/AWSLogs/...`
instead of the intended `s3://medplum-logs-us-east-1/elb/AWSLogs/...`.

## Expected behaviour

The value of `config.loadBalancerLoggingPrefix` should be forwarded to
`logAccessLogs` and end up in `access_logs.s3.prefix`. Log objects should land
under `<prefix>/AWSLogs/...` as documented by
[`ApplicationLoadBalancer.logAccessLogs`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_elasticloadbalancingv2.ApplicationLoadBalancer.html).

## Workaround

None that produces the right template. Setting `loadBalancerLoggingBucket` to
the *desired prefix* would make `access_logs.s3.prefix` correct but breaks the
bucket lookup at `backend.ts:538` (`s3.Bucket.fromBucketName(...)`). Operators
have to live with the wrong prefix until the wiring is fixed.

## Impact

- Access logs from every Medplum deployment write to `<bucket-name>/AWSLogs/...`
  rather than an isolated `<configured-prefix>/AWSLogs/...`, causing collisions
  in shared logging buckets.
- Operators who configure `loadBalancerLoggingPrefix` find that downstream
  consumers keyed on that prefix (S3 lifecycle rules, IAM scopes, Athena
  partitions, log-aggregation pipelines) silently fail to match real objects.
- Both the main ALB (L552) and the optional mTLS ALB (L568) are affected.

## Suggested fix

In `packages/cdk/src/backend.ts`, change both `this.createLoadBalancer(...)`
call sites to pass `config.loadBalancerLoggingPrefix` (not
`config.loadBalancerLoggingBucket`) as the `loadBalancerLoggingPrefix`
argument. The method signature, JSDoc, and config type are already correct;
only the two arguments need to change.

A regression test should assert against the synthesized template, e.g.:

```ts
template.hasResourceProperties('AWS::ElasticLoadBalancingV2::LoadBalancer', {
  LoadBalancerAttributes: Match.arrayWith([
    { Key: 'access_logs.s3.prefix', Value: 'elb' },
  ]),
});
```

The existing `packages/cdk/src/index.test.ts` fixture already sets
`loadBalancerLoggingPrefix: 'elb'` but the test only verifies that `synth()`
succeeds, so the bug is not caught today.

jfiorato:
`BackEnd.createLoadBalancer` already declares a `loadBalancerLoggingPrefix: string | undefined` parameter and forwards it to `loadBalancer.logAccessLogs(this.loggingBucket, prefix)`, but both call sites in `BackEnd` were passing `config.loadBalancerLoggingBucket` (the bucket name) as that argument. The synthesized `AWS::ElasticLoadBalancingV2::LoadBalancer` therefore carried `access_logs.s3.prefix` equal to the bucket name, and the value users configure in `loadBalancerLoggingPrefix` was silently ignored — even though the property is already defined on `BaseConfig` / `MedplumInfraConfig` and the existing test fixture sets it to `'elb'`.

Both the main ALB and the optional mTLS ALB are affected; both call sites are updated to pass `config.loadBalancerLoggingPrefix`. The method signature, JSDoc, and config types were already correct, so this is purely a two-line wiring fix.

Tests:
- New regression test in `packages/cdk/src/index.test.ts` (`'ALB access log prefix uses loadBalancerLoggingPrefix'`) instantiates `MedplumStack` directly, runs `Template.fromStack(...)`, and asserts that exactly two `AWS::ElasticLoadBalancingV2::LoadBalancer` resources are synthesized, that both carry `LoadBalancerAttributes` containing `{ Key: 'access_logs.s3.prefix', Value: 'elb' }`, and that zero ALBs carry the bucket name as the prefix.
- The existing `index.test.ts` suite only asserts `synth()` doesn't throw, which is why the bug was live undetected; the new test runs against the synthesized template so the regression is caught at the assertion layer.
- `npm test -w @medplum/cdk`: 67/67 pass. Verified the new test fails when either call site is reverted to `config.loadBalancerLoggingBucket`.

Closes #9287.


codyebberson (approved):

codyebberson:
Thanks @jfiorato !

PR #9289, opened 2026-05-24
Fixes #9276 - unified external auth

### LINKED ISSUE #9276: External IDP Token Exchange Project Membership Selection
## Problem
For implementers self-hosting medplum, **token exchange cannot satisfy server scoped use cases where the users authenticating with external IDP via token exchange need to be able to access multiple projects**. This is because the current token exchange feature stores the external ID on the project membership, meaning there can only be one project membership affiliated with the external user in Auth0 or Okta for example, Additionally, the current token exchange feature requires a client application within a specific project that token exchange will be used within. This makes things tricky because that ClientApplication will now need to be duplicated across all projects. 

## Desired Behavior

1. Similar to MedPlum's Google Auth, the desired behavior here is to allow for an intermediary project membership selection step as part of token exchange.
2. The external IDP information can be stored on some super admin ClientApplication **or** server config, similar to our direct external auth feature.

codyebberson:
Fixes #9276

This is an incremental step toward unifying Medplum external authentication flows behind the existing `IdentityProvider` type and `/auth/external` callback.

This PR does **not** attempt to replace Google auth or the existing bearer-token external auth flow yet. It lays groundwork for a shared external auth model while preserving existing behavior.

## What Changed

- Added new `IdentityProvider` fields:
  - `identitySource`
  - `identityMappingMode`
- Added CodeSystem / ValueSet definitions for the new enum values.
- Updated `/auth/external` to resolve external identity using the new two-axis model:
  - `identitySource`: which claim/value to read from the external credentials
  - `identityMappingMode`: which Medplum identity concept to map that value to
- Preserved legacy behavior:
  - Existing email-based external auth maps to `User.email`
  - Existing `useSubject` auth maps to `ProjectMembership.externalId`
  - `useSubject` remains supported as legacy shorthand
- Extended server config external auth provider shape:
  - `MedplumExternalAuthConfig.identityProvider?: IdentityProvider`
  - `MedplumExternalAuthConfig.userInfoUrl` is retained as deprecated legacy config
- Added support for `/auth/external` to select a server-configured identity provider using `state.issuer`.

## Compatibility

Backwards compatibility is intended to be preserved.

Existing configurations without `identitySource` or `identityMappingMode` continue to use legacy interpretation:

- `useSubject: true` implies:
  - `identitySource = "subject"`
  - `identityMappingMode = "project-membership-external-id"`
- no `useSubject` implies:
  - `identitySource = "email"`
  - `identityMappingMode = "user-email"`

Legacy `externalAuthProviders[].userInfoUrl` remains supported for existing bearer-token external auth.

## Larger Plan

The long-term goal is to unify Medplum's external auth implementations:

- Google auth
- DomainConfiguration external auth
- ClientApplication external auth
- server-level `externalAuthProviders`
- bearer-token external auth

The intended end state is a shared flow:

1. Resolve an `IdentityProvider`
2. Validate external credentials
3. Normalize external claims
4. Resolve Medplum identity according to configured mapping mode
5. Create a Medplum `Login`
6. Let existing membership/profile selection complete the login

## Still in progress

- Google auth is not yet refactored into the shared `/auth/external` internals.
- `project-membership-profile` / `fhirUser` mapping is not yet implemented for `/auth/external`.
- Server-scoped subject mapping is not implemented yet.

For server-scoped subject mapping, we should avoid using `User.externalId` because it is deprecated. A likely future direction is mapping `{issuer, sub}` to `User.identifier`, but that needs a separate design pass.

- ID token cryptographic validation in generic external auth remains unchanged.
- The existing server-level bearer-token auth flow is only lightly adapted to the new config shape; it is not fully unified yet.

### DIFF
--- a/packages/server/src/auth/external.ts
+++ b/packages/server/src/auth/external.ts
@@ -5,6 +5,7 @@ import {
   concatUrls,
   ContentType,
   encodeBase64,
+  isString,
   OAuthGrantType,
   OAuthTokenAuthMethod,
   OperationOutcomeError,
@@ -213,7 +214,10 @@ async function resolveExternalLoginIdentity(
   const identityMappingMode = getIdentityMappingMode(idp);
 
   if (identitySource === 'email' && identityMappingMode === 'user-email') {
-    const email = (userInfo.email as string | undefined)?.toLowerCase();
+    let email: string | undefined;
+    if (isString(userInfo.email)) {
+      email = userInfo.email.toLowerCase();
+    }
     if (!email) {
       throw new OperationOutcomeError(badRequest('External token does not contain email address'));
     }
@@ -221,7 +225,10 @@ async function resolveExternalLoginIdentity(
   }
 
   if (identitySource === 'subject' && identityMappingMode === 'project-membership-external-id') {
-    const externalId = userInfo.sub as string | undefined;
+    let externalId: string | undefined;
+    if (isString(userInfo.sub)) {
+      externalId = userInfo.sub;
+    }
     if (!externalId) {
       throw new OperationOutcomeError(badRequest('External token does not contain subject'));
     }
@@ -232,10 +239,12 @@ async function resolveExternalLoginIdentity(
 }
 
 function getIdentitySource(idp: IdentityProvider): IdentitySource {
+  // Fallback to preserve legacy behavior
   return idp.identitySource ?? (idp.useSubject ? 'subject' : 'email');
 }
 
 function getIdentityMappingMode(idp: IdentityProvider): IdentityMappingMode {
+  // Fallback to preserve legacy behavior
   return idp.identityMappingMode ?? (idp.useSubject ? 'project-membership-external-id' : 'user-email');
 }
 
--- a/packages/server/src/auth/external.ts
+++ b/packages/server/src/auth/external.ts
@@ -30,6 +30,7 @@ import { getDomainConfiguration } from './method';
 
 export interface ExternalAuthState {
   domain?: string;
+  issuer?: string;
   projectId?: string;
   clientId?: string;
   scope?: string;
@@ -39,6 +40,14 @@ export interface ExternalAuthState {
   codeChallengeMethod?: CodeChallengeMethod;
   redirectUri?: string;
   returnTo?: string;
+}
+
+type IdentitySource = NonNullable<IdentityProvider['identitySource']>;
+type IdentityMappingMode = NonNullable<IdentityProvider['identityMappingMode']>;
+
+interface ExternalLoginIdentity {
+  email?: string;
+  externalId?: string;
 }
 
 export async function externalCallbackHandler(req: Request, res: Response): Promise<void> {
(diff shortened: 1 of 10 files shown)


codyebberson:
### How to use with Okta

You will need:
1. Your Okta domain
2. Okta client ID
3. Okta client secret

#### Step 1:  Update your Medplum server config

```json
  "externalAuthProviders": [
    {
      "issuer": "https://${oktaDomain}/oauth2/default",
      "identityProvider": {
        "authorizeUrl": "https://${oktaDomain}/oauth2/v1/authorize",
        "tokenUrl": "https://${oktaDomain}/oauth2/v1/token",
        "userInfoUrl": "https://${oktaDomain}/oauth2/v1/userinfo",
        "clientId": "your client ID",
        "clientSecret": "your client secret",
        "identitySource": "email",
        "identityMappingMode": "user-email"
      }
    }
  ],
```

#### Step 2: Add client side login code

```ts
  async function redirectToOkta(): Promise<void> {
    const oktaUrl = new URL(`https://${oktaDomain}/oauth2/v1/authorize`);
    oktaUrl.searchParams.set('client_id', oktaClientId);
    oktaUrl.searchParams.set('response_type', 'code');
    oktaUrl.searchParams.set('scope', 'openid profile email');
    oktaUrl.searchParams.set('redirect_uri', 'http://localhost:8103/auth/external');
    oktaUrl.searchParams.set(
      'state',
      JSON.stringify({
        ...(await medplum.ensureCodeChallenge(baseLoginRequest)),
        issuer: `https://${oktaDomain}/oauth2/default`,
        returnTo: locationUtils.getLocation(),
      })
    );
    locationUtils.assign(oktaUrl.toString());
  }
```

codyebberson:
@claude please review

packages/server/src/auth/external.ts
```diff
@@ -194,9 +198,47 @@ async function getIdentityProvider(
...
+      throw new OperationOutcomeError(badRequest('External token does not contain subject'));
+    }
+    return { externalId };
+  }
+
+  throw new OperationOutcomeError(badRequest('Unsupported identity provider configuration'));
+}
+
+function getIdentitySource(idp: IdentityProvider): IdentitySource {
+  return idp.identitySource ?? (idp.useSubject ? 'subject' : 'email');
```
mattlong, line 243:
```suggestion
    // fallbacks to preserve legacy behavior
  return idp.identitySource ?? (idp.useSubject ? 'subject' : 'email');
```

packages/server/src/auth/external.ts
```diff
@@ -194,9 +198,47 @@ async function getIdentityProvider(
...
+}
+
+function getIdentityMappingMode(idp: IdentityProvider): IdentityMappingMode {
+  return idp.identityMappingMode ?? (idp.useSubject ? 'project-membership-external-id' : 'user-email');
```
mattlong, line 248:
```suggestion
  // fallbacks to preserve legacy behavior
  return idp.identityMappingMode ?? (idp.useSubject ? 'project-membership-external-id' : 'user-email');
```

packages/server/src/auth/external.ts
```diff
@@ -194,9 +198,47 @@ async function getIdentityProvider(
...
 
+async function resolveExternalLoginIdentity(
+  idp: IdentityProvider,
+  userInfo: Record<string, unknown>
+): Promise<ExternalLoginIdentity> {
+  const identitySource = getIdentitySource(idp);
+  const identityMappingMode = getIdentityMappingMode(idp);
+
+  if (identitySource === 'email' && identityMappingMode === 'user-email') {
+    const email = (userInfo.email as string | undefined)?.toLowerCase();
```
mattlong: I suppose the `toLowerCase()` effectively validates that email is a string, but avoiding casts would be nice.
```suggestion
    let email: string | undefined;
    if (typeof userInfo.email === 'string') {
      email = userInfo.email.toLowerCase();
    }
```

packages/server/src/auth/external.ts
```diff
@@ -194,9 +198,47 @@ async function getIdentityProvider(
...
+    if (!email) {
+      throw new OperationOutcomeError(badRequest('External token does not contain email address'));
+    }
+    return { email };
+  }
+
+  if (identitySource === 'subject' && identityMappingMode === 'project-membership-external-id') {
+    const externalId = userInfo.sub as string | undefined;
```
mattlong: Similar type validation as above

mattlong (approved):

PR #9294, opened 2026-05-26
Various oauth2 fixes for spec compliance

codyebberson:
Preparing to apply for OAuth2 / Open ID Connect certification.

It has been a while since I ran the conformance tests.  The tool has changed a little, and our docs were out of date.  

The only functional change is that we need to demonstrate we can return all user info fields, so this adds support for: 

- `middle_name`
- `nickname`
- `preferred_username`
- `website`
- `zoneinfo`

After that, we're mostly green on the conformance suite:

image

### DIFF
--- a/packages/server/src/oauth/README.md
+++ b/packages/server/src/oauth/README.md
@@ -8,23 +8,77 @@ OpenID Implementer's Guide: https://openid.net/specs/openid-connect-basic-1_0.ht
 
 For both local and production testing, setup your Practitioner profile with all of the OpenID user info properties:
 
-| User info property    | Practitioner property | Example value |
-| --------------------- | --------------------- | ------------- |
-| name                  | name[0]               | Alice         |
-| given_name            | name[0].given[0]      |               |
-| middle_name           | name[0].given[0]      |               |
-| family_name           | name[0].family        |               |
-| gender                | name[0].gender        |               |
-| picture               | photo[0].url          |               |
-| preferred_username    |                       |               |
-| nickname              |                       |               |
-| website               |                       |               |
-| zoneinfo              |                       |               |
-| email                 | telecom               |               |
-| email_verified        |                       |               |
-| phone_number          | telecom               |               |
-| phone_number_verified |                       |               |
-| address               | address               |               |
+| User info property    | Practitioner property       | Example value |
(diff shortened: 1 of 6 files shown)


packages/server/src/oauth/README.md
```diff
@@ -195,7 +223,7 @@ Be sure to logout between each test by visiting <http://host.docker.internal:810
 
 In "EHR Launch Sequence":
 
-Make sure "Scopes" includes "fhirUser launch launch/patient offline_access openid profile user/*._ patient/_.\_"
+Make sure "Scopes" includes "fhirUser launch launch/patient offline*access openid profile user/\*.* patient/\_.\_"
```
ThatOneBro: Is this supposed to be `offline`?

packages/server/src/oauth/README.md
```diff
@@ -33,90 +87,59 @@ Follow these instructions to setup your local dev environment with the Conforman
...
 - The Medplum API server cookies require HTTPS
-  - That means we cannot simply use `vite` and `tsx`
-  - Instead, we will use [Caddy](https://caddyserver.com/) for an easy localhost HTTPS proxy
+- That means that the Medplum API server cannot simply use "localhost"
+- Instead, we will use ngrok to create a public URL that tunnels to our local dev environment, and configure the Medplum API server to use that URL as its base URL
+
+The Medplum OpenID configuration requires HTTPS. We recommend using ngrok for this purpose. We recommend using multiple tunnels: one for the API server and one for the app server. This allows the Conformance Suite to access the API server directly, without going through the app server.
 
-The Medplum OpenID configuration requires HTTPS. We recommend using .
+For the purposes of this example, we will use the following ngrok URLs:
```
ThatOneBro, line 96: 👍

ThatOneBro (commented):
Nice, makes sense for the most part. One question about updated docs

packages/server/src/oauth/README.md
codyebberson: Eek, markdown got mangled.

We have always supported both `offline` and `offline_access`, but apparently `offline_access` is the conventional option.

ThatOneBro: Oh I had it backwards

ThatOneBro (approved):

PR #9295, opened 2026-05-26
feat(server): include claim submit operation

techdavidy:
Submits claim to the corresponding custom operation.

Resolution order:
1. 'processor' parameter in the request body ('stedi' or 'candid').
2. Project secrets: STEDI_CLAIM_API_KEY (→ stedi), then CANDID_SECRET_ID (→ candid).

Endpoints:
POST /Claim/$submit                (Claim passed via the 'resource' input parameter)
POST /Claim/:id/$submit            (Claim read from the URL)


finnbergquist:
@techdavidy , what do we do in the event where both project secrets are set?

techdavidy:
@finnbergquist there is a parameter to go with an given processor.

codyebberson (changes requested):
@techdavidy  - nice, this is a good feature.

👍  that both the Candid and Stedi integrations are conforming to the same `Claim/$submit` operation definition. That's exactly the right direction architecturally, since it allows implementations to be swapped without changing client behavior.

One thought: rather than having the server know about specific vendors, I wonder if we can make this one level more generic. For example, we could have a single project setting such as `CLAIM_SUBMIT_BOT_ID`. If configured, the operation would invoke that bot. If not configured, we could return a clear "Claim submit not configured" error.

That would keep the core server implementation vendor-neutral while still supporting Candid, Stedi, and any future claim processors that implement the same operation contract.

What do you think?

techdavidy:
@claude please review

packages/server/src/fhir/operations/claimsubmit.ts
```diff
@@ -0,0 +1,88 @@
...
+  const { project, repo } = getAuthenticatedContext();
+
+  const customOperationCode = project.setting?.find((s) => s.name === CLAIM_SUBMIT_OPERATION_SETTING)?.valueString;
+  if (!customOperationCode) {
+    return [badRequest('Claim submit is not configured: set the CLAIM_SUBMIT_OPERATION project setting.')];
+  }
+
+  const claimId = (req.params as { id?: string } | undefined)?.id;
+  let claim: Claim;
+  if (claimId) {
```
mattlong: It should probably be an error if both `params.resource` from the POST body and `claimId` from the URL are specified.

packages/server/src/fhir/operations/claimsubmit.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  const subRequest: FhirRequest = {
+    ...req,
+    url: `/Claim/$${customOperationCode}`,
+    body: claim,
+  };
+  const result = await tryCustomOperation(subRequest, repo);
+  if (!result) {
+    return [badRequest('Claim submit operation is not available.')];
```
mattlong: The amount of guidance in the error handling in this handler should probably be consistent. The other `badRequest` provide clues on how to fix it while this one does not. Either approach is defensible, let's just make it consistent.

mattlong (changes requested):

mattlong (approved):

codyebberson:
@claude - one last final review, everything look good?

packages/server/src/fhir/operations/claimsubmit.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Dispatches the Claim passed via the 'resource' input parameter to the configured claim processor.
+ *
+ * Endpoint:
+ *   [fhir base]/Claim/$submit
+ *
+ * @param req - The FHIR request.
+ * @returns The FHIR response from the underlying custom operation.
+ */
+export async function claimSubmitPostHandler(req: FhirRequest): Promise<FhirResponse> {
+  // try {
```
codyebberson: Minor cleanup

```suggestion
```

packages/server/src/fhir/operations/claimsubmit.ts
```diff
@@ -0,0 +1,128 @@
...
+export async function claimSubmitGetHandler(req: FhirRequest): Promise<FhirResponse> {
+  const { repo } = getAuthenticatedContext();
+  const claimId = req.params.id;
+
+  if (!claimId) {
+    return [badRequest('Claim ID is required')];
+  }
+
+  const claim = await repo.readResource<Claim>('Claim', claimId);
+  const result = await handleClaimSubmit(req, claim);
```
codyebberson: Minor cleanup

```suggestion
  return handleClaimSubmit(req, claim);
```

codyebberson (approved):
lgtm.  couple small cleanup items if you would like

PR #9302, opened 2026-05-26
Provider: Add Appointment confirm button

noahsilas:
After using `$hold` (such as from the FooMedical example), an appointment is in the "pending" state.

- Represent "pending" status visually with an outlined style
  Screenshot 2026-05-27 at 9 46 53 AM

- Add a button to invoke `$confirm` for pending appointments
  Screenshot 2026-05-27 at 9 47 00 AM

### DIFF
--- a/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
+++ b/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
@@ -3,11 +3,11 @@
 import { Button, Divider, Group, Stack, Text, Tooltip } from '@mantine/core';
 import { showNotification } from '@mantine/notifications';
 import type { WithId } from '@medplum/core';
-import { createReference, formatHumanName, formatPeriod, isReference } from '@medplum/core';
-import type { Appointment, Coding, Patient, PlanDefinition, Practitioner } from '@medplum/fhirtypes';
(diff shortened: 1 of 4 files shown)


examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -158,18 +159,48 @@ export function AppointmentDetails(props: {
...
     } finally {
       setCancelLoading(false);
     }
-  }, [medplum, appointment, cancellable, onUpdate]);
+  }, [medplum, appointment, cancellable, onAppointmentUpdate]);
+
+  const confirmable = appointment.status === 'pending';
+  const [confirmLoading, setConfirmLoading] = useState(false);
+  const handleConfirm = useCallback(async () => {
+    if (!confirmable) {
```
techdavidy, line 168: should be surface a warning to user ?

techdavidy (approved):

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
noahsilas: This shouldn't be possible a possible outcome (the button should be hidden when this is not confirmable) - this is a bit of a legacy of my first pass before I wired that up. Oops. Let me remove this instead of trying to figure out error copy.

PR #9303, opened 2026-05-27
feat(react): AI real time questionnaire form

techdavidy:


kevinwadeshaw:
Looking good! Just a few questions/notes here:

- Is there a reason we need the banner content area to be toggle-able (via arrow icon button)? Thought it was cleaner just exposing while dictation is running, since the content is primarily supporting that state—e.g. seeing the Transcript input isn't helpful when dictation is off.
- For the Listening state, I think it's helpful to keep the "Listening…" label constant as the system can listen and process at the same time. 
  - This is why I had moved the "Processing…" state to the button—so we can see these happening concurrently
- The "Processing…" label as button state is also helpful as it restricts the user from stopping dictation while it's processing and potentially seeing a mix of conflicting states like the following, where it looks like it's listening and processing while dictation is actually off:
Screenshot 2026-05-27 at 11 02 28 AM

---

Let me know what you think about these issues and if you want to discuss live!

The-Alchemist:
Strange build error:

```
viewport.scrollTo is not a function
 ❯ ../../packages/react/src/utils/diff.ts:115:11
```

packages/react/src/QuestionnaireForm/AIRealTimeQuestionnaireForm.tsx
```diff
@@ -0,0 +1,402 @@
...
+    checkBotAvailability().catch(console.error);
+    return () => {
+      cancelled = true;
+    };
+  }, [medplum]);
+
+  const isProjectVoiceEnabled = medplum.getProject()?.features?.includes('ai-realtime') ?? false;
+  const isVoiceEnabled = isProjectVoiceEnabled && botAvailability === 'available';
+  const idleLabel: ReactNode =
+    botAvailability === 'unavailable' ? (
```
mattlong: If `isProjectVoiceEnabled` is false, this could still show `DEFAULT_IDLE_LABEL` which seems incorrect?

packages/react/src/QuestionnaireForm/AIRealTimeQuestionnaireForm.tsx
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  let statusState: 'finishing' | 'recording' | 'idle' = 'idle';
+  if (isFinishing) {
+    statusState = 'finishing';
+  } else if (activeStatusLabel) {
+    statusState = 'recording';
+  }
+
+  let statusIcon: JSX.Element;
```
mattlong: Seems like this if/else if block could be merged with the `statusState` one

mattlong (approved):

PR #9307, opened 2026-05-27
Added order set sync bot trigger after importing example order set

AndyStoneman:
Simple PR to trigger the order set sync bot after importing our example order set in the "Getting Started" tab. This makes the new order set automatically sync into DAW upon import.

### DIFF
--- a/examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
+++ b/examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
@@ -18,6 +18,7 @@ import { showNotification } from '@mantine/notifications';
 import { convertToTransactionBundle } from '@medplum/core';
 import type { Bundle, BundleEntry } from '@medplum/fhirtypes';
 import { MedplumLink, useMedplum } from '@medplum/react';
+import { SCRIPTSURE_ORDERSET_SYNC_BOT } from '@medplum/scriptsure-react';
 import {
   IconApps,
   IconArrowUpRight,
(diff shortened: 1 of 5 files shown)


examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
```diff
@@ -129,6 +130,13 @@ export function GetStartedPage(): JSX.Element {
       const resourceCount =
         result.entry?.filter((entry: BundleEntry) => entry.response?.status?.startsWith('2')).length || 0;
 
+      const pdLocation = result.entry?.find((e) => e.response?.location?.startsWith('PlanDefinition/'))?.response
```
oleg-mp, line 134: My main concern is that it assumes existence of DAW integration for users, also this leaks specific implementation of e-prescribing (DAW) to the top level vs keeping it vendor agnostic and resolve downstream based on existence of specific implementation.

examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
```diff
@@ -129,6 +130,13 @@ export function GetStartedPage(): JSX.Element {
...
+        ?.location;
```
oleg-mp, line 135: We also need to make sure the PlanDefinition is actually an order set and filter out the ones that are not. Safe in this particular case of curated bundle, but in general good hygiene to have a guard and only allow the ones where `"http://terminology.hl7.org/CodeSystem/plan-definition-type", code: "order-set"`

oleg-mp:
@claude please review

oleg-mp (approved):
Looks good. The issue that Claude found (syncOrderSet error can mask a successful import) looks like a false positive, so not a blocker

PR #9308, opened 2026-05-27
feat(server,app): add email MFA method

### LINKED ISSUE #9070: Add email and/or SMS support for MFA
**Issue:** Medplum currently only supports TOTP for MFA.

**Impact:** This creates a significant accessibility barrier for certain patient populations (e.g. elderly patients, patients without smartphones, or those with limited digital literacy).

**Proposed fix**: Add support for email and/or SMS-based OTP as an alternative MFA method.

ThatOneBro:
Fixes #9070

## Summary

Medplum's MFA only supported TOTP authenticator apps. As #9070 notes, that's an accessibility barrier for patients without smartphones or with limited digital literacy. This PR generalizes MFA into a **multi-method** system and adds **email-based MFA** — a single-use 6-digit code emailed at sign-in — as the first alternative.

It lets a project:
- offer **email MFA, authenticator (TOTP) MFA, or both**, via an `allowedMfaMethods` project setting;
- let a user enroll in **more than one factor** and satisfy a challenge with **any** of them;
- **add or remove individual factors** (or disable MFA entirely) from a redesigned settings page.

Fully backward compatible: existing TOTP-enrolled users and projects that set nothing behave exactly as before (no migration needed).

## Data model

| Field | Type | Purpose |
|---|---|---|
| `User.mfaMethod` | `('totp' \| 'email')[]` | Factors the user is enrolled in. Bound to a new `user-mfa-method` CodeSystem/ValueSet. |
| `Login.emailMfa` | `{ codeHash: string; expiresAt: instant }` | Per-login holder for the emailed code: a **bcrypt hash** plus expiry. Cleared on verification. |

New constant `EMAIL_MFA_CODE_EXPIRATION_MS` = 20 minutes. A user with `mfaEnrolled` but no `mfaMethod` is treated as `['totp']`, so legacy accounts are unaffected.

## How it works — server

- `getAllowedMfaMethods(project)` reads the comma-delimited `allowedMfaMethods` setting (defaults to `['totp']`); `getEnrolledMfaMethods(user)` returns the user's factors with the legacy fallback.
- **Issuing a code** (`sendMfaEmailCode`): a 6-digit `randomInt` code is bcrypt-hashed onto the `Login` with an expiry and emailed in plaintext; it's single-use and cleared once verified.
- **Verifying** (`verifyMfaToken`): checks the emailed code first (expiry + `bcrypt.compare`), and on success also sets `User.emailVerified = true` and clears `emailMfa`; otherwise falls back to TOTP.
- **Login response** (`sendLoginResult`): when email is the user's only factor, the code is emailed immediately so the UI can go straight to code entry. Responses now carry `mfaMethods` / `email` (and `allowedMfaMethods` on enroll-required).

**Endpoints** (`auth/mfa.ts`):
- `GET /status` → returns `enrolledMethods`, `allowedMethods`, `email`; always ensures a TOTP secret exists so the QR can be shown even when adding TOTP to an email-enrolled account.
- `POST /login-enroll` / `POST /enroll` → take a `method`; email needs no token, TOTP still verifies the authenticator code. Per-method dedupe guard and project-allowed check.
- `POST /send-email` (login-scoped) and `POST /send-email-challenge` (authenticated) → (re)send a code during sign-in / to authorize a settings change.
- `POST /disable` → optional `method` removes just that factor and keeps the rest; omitting it disables MFA entirely. Both require proof of a connected factor via `verifyConnectedFactor` (TOTP code *or* emailed code), regenerate the TOTP secret when TOTP is removed, and consume any outstanding code.

## How it works — UI

- **`@medplum/react`:** new `MfaEnrollForm` (method chooser / email-only / TOTP-QR) and `MfaVerificationForm` (authenticator-first with "Get a code by email instead" / "Resend code"). `MfaForm` gains a `footer` slot and `ReactNode` descriptions. `SignInForm` is rewired to handle auto-sent codes, email mode, and email enrollment; `LoginAuthenticationResponse` gains `allowedMfaMethods` / `mfaMethods` / `email`.
- **`@medplum/app`:** `MfaPage` becomes a full management surface — list enabled methods, add email/authenticator, remove a single factor, or disable MFA, each behind a modal. `SecurityPage` always links to it ("Manage MFA" / "Enroll").

## Flows enabled

1. Project picks its policy (email-only, TOTP-only, or both).
2. Enroll — choose a method, or go straight in for single-method projects.
3. Sign in with email MFA — code emailed automatically, user types it in.
4. Sign in with both factors — authenticator by default, email fallback.
5. Add a second factor to an enrolled account.
6. Remove one factor while keeping others, or disable MFA — each gated on re-proving a connected factor.

## Security

Emailed codes are bcrypt-hashed at rest, single-use, 6 digits, 20-minute expiry; every settings change requires re-proving a connected factor; verifying an emailed code also verifies the user's email.

## Tests

`auth/mfa.test.ts` (+~1400 lines incl. legacy regression + edge cases), additional oauth/auth coverage, and `MfaPage` / `SignInForm` tests migrated to Vitest.


packages/server/src/auth/utils.ts
```diff
@@ -25,6 +28,73 @@ import { TODO_SHARD_ID } from '../fhir/sharding';
...
+ * @param login - The login to attach the hashed code to.
+ * @param user - The user to email.
+ */
+export async function sendMfaEmailCode(login: WithId<Login>, user: User): Promise<void> {
+  const systemRepo = getGlobalSystemRepo();
+  const code = randomInt(0, 1_000_000).toString().padStart(6, '0');
+  const codeHash = await bcryptHashPassword(code);
+  const expiresAt = new Date(Date.now() + EMAIL_MFA_CODE_EXPIRATION_MS).toISOString();
+  await systemRepo.updateResource<Login>({ ...login, emailMfa: { codeHash, expiresAt } });
+  const expirationMinutes = Math.round(EMAIL_MFA_CODE_EXPIRATION_MS / 60_000);
```
mattwiller: Might be better to underestimate, just in case:
```suggestion
  const expirationMinutes = Math.floor(EMAIL_MFA_CODE_EXPIRATION_MS / 60_000);
```

packages/server/src/auth/utils.ts
```diff
@@ -25,6 +28,73 @@ import { TODO_SHARD_ID } from '../fhir/sharding';
...
+  await sendEmail(systemRepo, {
+    to: user.email,
+    subject: `Your Medplum verification code: ${code}`,
+    text: [
+      'Here is your Medplum verification code:',
```
mattwiller: This copy should probably include a) a reference to the fact that this is a login code, and b) basic instructions for what to do with the code to complete the log in

packages/server/src/auth/utils.ts
```diff
@@ -94,12 +181,26 @@ export async function sendLoginResult(res: Response, login: Login): Promise<void
...
     return;
   }
 
   if (user.mfaEnrolled && login.authMethod === 'password' && !login.mfaVerified) {
-    res.json({ login: login.id, mfaRequired: true });
+    const mfaMethods = getEnrolledMfaMethods(user);
+    // If email is the user's only MFA method, send the verification code
+    // immediately so they can enter it without any further interaction. A set
+    // emailMfa means a code has already been issued for this login, so
+    // don't re-send (e.g. when the login status endpoint is queried again).
```
mattwiller, line 220: 👍

packages/server/src/auth/mfa.ts
```diff
@@ -13,18 +13,41 @@ import { invalidRequest, sendOutcome } from '../fhir/outcomes';
...
+/**
+ * Returns the union of the user's existing MFA methods plus the new method.
+ * @param user - The user.
+ * @param method - The method to add.
+ * @returns The updated list of MFA methods.
+ */
+function addMfaMethod(user: User, method: MfaMethod): MfaMethod[] {
+  const methods = new Set<MfaMethod>(getEnrolledMfaMethods(user));
+  methods.add(method);
+  return Array.from(methods);
```
mattwiller: This double allocation isn't necessary, could just scan the array instead:

```suggestion
  const methods = getEnrolledMfaMethods(user);
  if (!methods.includes(method)) {
    methods.push(method);
  }
  return methods;
```

packages/server/src/auth/mfa.ts
```diff
@@ -149,20 +253,58 @@ mfaRouter.post(
...
+      methodToRemove = requested;
+      if (!enrolledMethods.includes(methodToRemove)) {
+        sendOutcome(res, badRequest('User not enrolled in MFA method'));
+        return;
+      }
     }
 
+    // If the user is enrolled in TOTP, require a valid authenticator code to
+    // make any change. Email-only users are already authenticated for this
+    // request, so no second factor is available to verify.
```
mattwiller: This feels distinctly less secure than TOTP: shouldn't users be forced to prove they have access to the email to make a change?

packages/server/src/auth/utils.ts
ThatOneBro: Done!

packages/server/src/auth/mfa.ts
ThatOneBro: Yeah made it so that you have to authenticate with either/or when disabling one or all factors

packages/fhirtypes/dist/Login.d.ts
```diff
@@ -242,3 +248,21 @@ export interface Login {
    */
   pictureUrl?: string;
 }
+
+/**
+ * Single-use 6-digit code for email-based MFA. Set when a code is
+ * issued for this login and cleared on verification.
+ */
+export interface LoginEmailMfa {
```
codyebberson, line 256: Can we avoid the nested type, and use the existing `Login.expiresAt`?

ThatOneBro: This is the expiration of the verification code, not the login itself

The two expiries can co-exist when a Login is valid but a user requests a login challenge while already logged in, in order to remove their authenticator if they lost access, for example

This enables self-service of re-configuring a user's authenticator if a project has both email and TOTP enabled

Scenario:
1. User logs in. `Login.expiresAt` set.
2. User requests verification code to remove authenticator. That code is only valid for X minutes (20 for now). We set both `Login.emailMfa.codeHash` and `Login.emailMfa.expiresAt` together. 

They always occur in a pair, so I thought it made sense to keep them as a single type that is either there or not.

The-Alchemist:
Cursor said:

Most systems **block removal of the last MFA method** when MFA is mandatory. The goal is to keep the account compliant with policy and avoid locking the user out.

## Typical behavior

**1. Hard block (most common)**  
The remove/delete action is disabled or returns an error such as:

- “You must keep at least one MFA method while MFA is required.”
- “Add another method before removing this one.”

The user stays at exactly one enrolled method; they cannot go to zero.

**2. Replace-before-remove (also very common)**  
Instead of “remove then add,” the flow is **replace**:

- Add the new method first (SMS, authenticator app, security key, etc.).
- Only after the new method is verified can the old one be removed.
- If there’s only one method, the UI often says **“Change”** or **“Replace”** rather than **“Remove.”**

**3. Atomic swap**  
Some products treat it as one operation: verify with the current method → enroll the new one → drop the old one in the same session. The account never has zero MFA enrolled.

## What usually does *not* happen

When MFA is **forced** (org policy, regulated industry, admin setting), systems rarely:

- Allow zero MFA and rely on “you’ll be prompted on next login” (that’s more common during **initial** enrollment, not after MFA is locked on).
- Let the user opt out without an admin or policy change.
- Remove MFA silently without a clear message.

## How “lost my only MFA method” is handled separately

That’s a different path, not “remove my last method”:

- Backup/recovery codes  
- Identity verification (email, support, ID check)  
- Admin reset of MFA  
- Break-glass / helpdesk procedures  

So: **you can’t self-serve down to zero MFA**, but **you can recover** if you lose access.

## Policy layers

| Who enforces MFA | User can remove last method? |
|------------------|------------------------------|
| User optional MFA | Often yes — account may fall back to password-only |
| User chose “require MFA on my account” | Usually no — same as org-forced |
| Org/admin mandatory MFA | No — only swap methods |
| Compliance (healthcare, finance, etc.) | No — often audit-logged; admin reset only |

## UX details most products share

- **Disable** the remove control on the last method, or show an explanatory modal.
- **Re-authenticate** (password + current MFA) before any MFA change.
- **Distinguish** “remove optional second factor” vs “remove required factor.”
- **Log** MFA enrollment/removal for security/compliance.

**Bottom line:** when MFA is mandatory, most sign-up/auth systems enforce a **minimum of one enrolled method**. Users can **change** how they MFA, but not **delete** down to none without an admin/policy change or a dedicated account-recovery flow.

The-Alchemist:
Also, after discussion with @ThatOneBro , it should prompt for email verification after enrollment in MFA.

The-Alchemist:
With @ThatOneBro , we discussed a UI issue:

* user is authentication through Google
* user enrolled in MFA (email or TOTP)
* UI says "MFA Enrolled: true"
* _but_ the system will not prompt for MFA because it's a Google account

Perhaps it should "MFA Enrolled: invalid" or something like that?  @codyebberson

codyebberson (approved):

PR #9311, opened 2026-05-27
Document Alpha/Beta feature expectations

noahsilas:
- include a link to the page from Provider's `AlphaBanner` component.
- update Scheduling operations with TSDoc `@experimental` annotations
- update "Alpha" admonitions around docs with link to alpha/beta definitions

### DIFF
--- a/packages/docs/docs/compliance/alpha-beta.md
+++ b/packages/docs/docs/compliance/alpha-beta.md
@@ -1,0 +1,87 @@
+# Alpha & Beta Features
+
+Some Medplum features are released before they reach general availability (GA). This page explains what the **Alpha** and **Beta** labels mean, what guarantees apply at each stage, and what you are opting in to when you build on a pre-GA feature.
+
+## Stability Stages
+
+| Stage           | API stability                                    | Breaking changes                                                           | Production use         |
+|-----------------|--------------------------------------------------|----------------------------------------------------------------------------|------------------------|
+| **Alpha**       | Experimental — subject to change at any time     | May occur in any release, including patch releases, without advance notice | Not recommended        |
+| **Beta**        | Mostly stable — significant changes are unlikely | Announced in release notes with ≥ one minor version of lead time           | Permitted with caution |
+| **GA (Stable)** | Stable — full semver guarantees apply            | Only in major releases, with ≥ 30 days written notice                      | Fully supported        |
+
+## Alpha
+
+A feature tagged **Alpha** is an early preview. It is functional enough to be useful for exploration or prototyping, but it is not yet complete.
+
+**What this means in practice:**
+
+- The API shape, request/response schema, and behavior can change between any two releases — including weekly patch releases — without a migration guide.
+- The feature may be disabled, renamed, or removed entirely if the design direction changes.
+- There is no SLA on availability or correctness for alpha features.
+- Alpha features are not covered by Medplum's deprecation policy (which only applies to GA features).
+
+**How alpha is marked:**
+
+- Documentation pages carry an `Alpha` admonition at the top.
+- Status columns in overview tables show **Alpha**.
+- TypeScript APIs may be annotated `@experimental` in TSDoc.
+
+**Who should use alpha features:**
+
+Alpha features are intended for developers who want to give early feedback or who are willing to track changes closely. Do not base production workflows or compliance-sensitive integrations on alpha features.
+
+This is the stage when feedback can be most rapidly incorporated, so we encourage Medplum users to test features at this stage and let us know when we can make their usage easier/better.
+
+## Beta
+
+A feature tagged **Beta** has stabilized enough that the core contract is unlikely to change, but it has not yet completed the full validation required for GA.
+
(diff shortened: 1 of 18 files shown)


packages/docs/docs/compliance/alpha-beta.md
```diff
@@ -0,0 +1,87 @@
+# Alpha & Beta Features
+
+Some Medplum features are released before they reach general availability (GA). This page explains what the **Alpha** and **Beta** labels mean, what guarantees apply at each stage, and what you are opting in to when you build on a pre-GA feature.
+
+## Stability Stages
+
+| Stage           | API stability                                    | Breaking changes                                                           | Production use         |
+|-----------------|--------------------------------------------------|----------------------------------------------------------------------------|------------------------|
+| **Alpha**       | Experimental — subject to change at any time     | May occur in any release, including patch releases, without advance notice | Not recommended        |
+| **Beta**        | Mostly stable — significant changes are unlikely | Announced in release notes with ≥ one minor version of lead time           | Permitted with caution |
```
ThatOneBro: I think this is sensible policy for Beta, but would be interested what customer-facing folks think

It's both easy to gate client versions on the end-user side and since our minor versions are relatively infrequent (on the order of months) I think this is conservative enough

packages/docs/docs/compliance/alpha-beta.md
```diff
@@ -0,0 +1,87 @@
...
+A feature can also be **retired** from alpha or beta without reaching GA if the approach is abandoned. In that case, Medplum will post a notice in the release notes and, where practical, document migration options.
+
+## Quick reference
+
+| Signal                              | Where you see it                  | Stage                               |
+|-------------------------------------|-----------------------------------|-------------------------------------|
+| `:::info[Alpha]` admonition in docs | Doc page header                   | Alpha                               |
+| `(Alpha)` in sidebar label          | Docs sidebar                      | Alpha                               |
+| **Alpha** in a status table         | Docs overview table               | Alpha                               |
+| `@experimental` TSDoc tag           | TypeScript source / API reference | Alpha                               |
```
ThatOneBro, line 56: Yeah we should definitely be using this TSDoc tag 👍

packages/docs/docs/compliance/alpha-beta.md
```diff
@@ -0,0 +1,87 @@
...
+
+This is the stage when feedback can be most rapidly incorporated, so we encourage Medplum users to test features at this stage and let us know when we can make their usage easier/better.
+
+## Beta
+
+A feature tagged **Beta** has stabilized enough that the core contract is unlikely to change, but it has not yet completed the full validation required for GA.
+
+**What this means in practice:**
+
+- Breaking changes are possible but will be called out in the release notes for the minor version that introduces them.
```
mattlong: Previously, minor version bumps were only used when wanting to rely on previous post-deploy migrations having been run. This adds a second reason which I think is generally fine. One detriment is that our self-hosting guidance of deploying the latest patch version of each minor version might result in some not strictly necessary deploys. Not a big deal, just something to keep in mind.

codyebberson (changes requested):
Let's discuss in person. Recent experiences with public-facing alpha features have been suboptimal.  We should discuss with @reshmakh regarding how customers and sales prospects perceive the alpha label, and whether we want to use it all externally.

codyebberson:
Thanks for discussing today @noahsilas 

Overall I like the idea of having a public Alpha/Beta policy page. Since Medplum develops in the open, I think it's helpful to set expectations around what users can safely build on and what guarantees apply at each stage.

A few concerns/suggestions:

1. The current document reads more like a formal policy than a customer-facing expectations guide. Some of the language creates fairly specific commitments that we may not always be able to meet as a small team.

For example:

* "announced ≥ one minor version of lead time"
* "Medplum will provide a migration path"
* "≥ 30 days written notice"

I'd consider softening these to things like:

* "whenever practical"
* "reasonable effort"
* "advance notice through release notes and documentation"

The goal is to communicate intent without accidentally creating obligations that become difficult to maintain.

2. I would consider shortening the page significantly. Most users are really trying to answer two questions:

* Can I use this in production?
* How likely is it to break?

The table at the top already does a good job of communicating that. The rest could potentially be condensed into a much shorter explanation.

3. I think we should explicitly frame public alpha development as a feature, not a weakness. Something along the lines of:

"Medplum develops in the open. Alpha features are available publicly so that developers can experiment early and provide feedback before APIs are finalized."

That better reflects how we operate as an open source company.

4. The GA section currently ties a number of compliance, certification, and SLA concepts to feature maturity. I'd double-check that we're comfortable making those statements globally. Some of them may be true in practice but are stronger claims than I would normally put on a docs page.

Net/net: I like the direction and think we should have a page like this. My main feedback is to make it a bit shorter and less contractual, while keeping the core message:

* Alpha = experimental, may change anytime
* Beta = production-capable but still evolving
* GA = covered by normal versioning and support expectations

codyebberson (approved):

PR #9314, opened 2026-05-28
Remove deprecated /auth/exchange endpoint (closes #9117)

### LINKED ISSUE #9117: Remove deprecated `/auth/exchange` endpoint
The `/auth/exchange` endpoint was added in March 2023: https://github.com/medplum/medplum/pull/1691

It was deprecated in April 2023: https://github.com/medplum/medplum/pull/1847

It was replaced by the proper OAuth 2.0 "Token Exchange" flow.

Since that time, the implementation of `/auth/exchange` has essentially been a shim over the OAuth standard, preserved only for backwards compatibility.

In the past 6 months, we have had zero successful requests on that endpoint.  There have been many 4xx requests, which all appear to be security researchers and pentesters (hackerone, etc).

At this point, I think it's safe to remove.

HRSHDG8:
## Summary

Removes the deprecated `/auth/exchange` endpoint, as requested in #9117.

- The endpoint was added in #1691 (March 2023) and deprecated in #1847 (April 2023) when proper OAuth 2.0 Token Exchange support was added.
- Per #9117, it has had zero successful requests in the past 6 months; the only traffic is 4xx requests from security researchers.
- Migration: callers should use `POST /oauth2/token` with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`, or the SDK helper `MedplumClient.exchangeExternalAccessToken()` (which already targets the standard endpoint).

## Changes

- Delete `packages/server/src/auth/exchange.ts` and its test file.
- Remove route registration from `packages/server/src/auth/routes.ts`.
- Unexport `exchangeExternalAuthToken` in `packages/server/src/oauth/token.ts` — after the shim is gone it is only called internally by `handleTokenExchange` (the OAuth 2.0 token exchange handler).
- Drop the "Legacy `/auth/exchange` endpoint" section from `packages/docs/docs/auth/token-exchange.md`.

The SDK method, the example code under `packages/examples/`, and the OAuth 2.0 token exchange documentation are all unchanged — they already use the standard endpoint.

This is a breaking change and is labeled `v6` on the issue, so should land in the v6 release.

## Test plan

- [x] `npx turbo run test --filter=@medplum/server` passes (jest suite ran fresh, no cache).
- [x] Grep for `/auth/exchange`, `exchangeHandler`, `exchangeValidator` across the repo returns 0 matches.
- [x] `exchangeExternalAuthToken` is referenced only inside `packages/server/src/oauth/token.ts` (definition + the standards-compliant caller).
- [ ] CI build/lint/test on the PR.


The-Alchemist:
hi, @HRSHDG8 , have you fix the merge conflicts?  thanks!

PR #9316, opened 2026-05-28
testing(hl7): migrate @medplum/hl7 from Jest to Vitest

The-Alchemist:
Migrates `@medplum/hl7` from Jest to Vitest.

## Changes
- Add `vite.config.ts` (node environment, serial execution for socket tests)
- Switch `package.json` test script to `vitest run`
- Convert tests from `jest.*` to `vi.*`
- **`client.ts`**: use `node:net` import for ESM-compatible spying


.envrc
```diff
@@ -0,0 +1,34 @@
+PATH_add /opt/homebrew/opt/node@24/bin
```
codyebberson: I think we want to remove this file?

codyebberson (approved):

PR #9325, opened 2026-05-28
fix(server): derive bare-id reference prefix from SearchParameter.target

### LINKED ISSUE #9254: Bare-UUID reference search is FHIR R4 non-compliant — only `patient`/`subject` resolve; `encounter`, `practitioner`, etc. silently return zero results

codyebberson:
Restores #9256

Reverts #9324

This PR restores the original reference search handling implementation while also preserving backwards compatibility for legacy shorthand queries.

Two additional fixes:

1. The original PR included code for backwards compatibility, however it assumed that `singleTargetType` would be false-y, but it could sometimes be populated with `'Resource'`.
2. The original PR did the `ElementDefinition` cross-walk on every write operation, which is a lot of duplicate work per write.  I moved it into `getSearchParameterDetails()`, which is cached and therefore only executed once.

Credit to @AbhiMsft  for the underlying implementation and standards alignment.

Original description:

* fix(server): derive bare-id reference prefix from SearchParameter.target

Closes #9254

Reference search parameters with a single target resource type now accept bare ids (the spec-compliant form per FHIR R4 ┬º3.1.1.4.12). Previously, the bare-id ΓåÆ typed-reference fixup in `buildReferenceSearchFilter` was hard-coded to the `subject` and `patient` columns, so queries like `DeviceRequest?encounter=<uuid>` silently returned an empty bundle.

The new behaviour:

  - For literal (non-canonical) reference search parameters whose underlying `SearchParameter.target` has exactly one entry, the column impl now carries a pre-computed `singleTargetType`. `buildReferenceSearchFilter` uses it to prefix bare UUID values with `Type/` before they hit the WHERE clause.
  - The existing `subject`/`patient` ΓåÆ `Patient/` fallback is preserved verbatim for back-compat on multi-target refs (e.g. `Observation.subject` targets [Patient, Group, Device, Location]).
  - Canonical reference params and non-UUID slashless values are left untouched to avoid corrupting URNs, canonical URLs, and other non-id values.

This mirrors the `searchParam.target?.length === 1` pattern already used by chained search (`parseChainLink`) in the same file.

Tests cover `DeviceRequest.encounter`, `PractitionerRole.practitioner`, the array reference column `Encounter.location`, the typed-prefix control, back-compat for `Observation.subject`, and the non-uuid safety case.

* Fix single-target reference derivation for shared search parameters

Shared search parameters like 'encounter' have multiple global targets (Encounter, EpisodeOfCare), but individual resource types like DeviceRequest.encounter only target Encounter. The previous check (searchParam.target.length === 1) missed these cases, causing bare UUID searches to return zero results.

Now derives the single target type from the resource's element definition targetProfile when the global SearchParameter has multiple targets but the element narrows to exactly one.

### DIFF
--- a/packages/server/src/fhir/searchparameter.ts
+++ b/packages/server/src/fhir/searchparameter.ts
@@ -172,61 +172,15 @@ function buildSearchParameterImplementation(
   const writeable = impl as Writeable<ColumnSearchParameterImplementation>;
   writeable.searchStrategy = 'column';
   writeable.columnName = convertCodeToColumnName(code);
-  if (searchParam.type === 'reference' && impl.type !== SearchParameterType.CANONICAL) {
-    const singleTarget = getSingleReferenceTarget(searchParam, impl);
-    if (singleTarget) {
-      writeable.singleTargetType = singleTarget;
-    }
+  if (
+    searchParam.type === 'reference' &&
+    impl.type !== SearchParameterType.CANONICAL &&
+    impl.referenceTargetTypes?.length === 1
+  ) {
+    writeable.singleTargetType = impl.referenceTargetTypes[0];
   }
 
   return impl;
-}
-
-const HL7_FHIR_STRUCTURE_PREFIX = 'http://hl7.org/fhir/StructureDefinition/';
-
-/**
- * Derives the single reference target type for a search parameter on a
- * specific resource type.  Shared search parameters (e.g. `encounter`) may
- * list multiple targets globally, but the element definition on a concrete
- * resource (e.g. `DeviceRequest.encounter`) often narrows it to exactly one.
- *
- * Returns the target resource type name if exactly one HL7 target profile is
- * found, otherwise `undefined`.
- *
- * @param searchParam - The search parameter definition.
- * @param impl - The search parameter implementation details.
- * @returns The single target resource type, or undefined if there are zero or multiple targets.
- */
-function getSingleReferenceTarget(
-  searchParam: SearchParameter,
-  impl: SearchParameterDetails
-): ResourceType | undefined {
-  // Fast path: the SearchParameter itself already has a single target.
-  if (searchParam.target?.length === 1) {
-    return searchParam.target[0];
-  }
-
-  // Derive from element definitions (resource-specific target profiles).
-  const elementDefs = impl.elementDefinitions;
-  if (elementDefs) {
-    const targetTypes = new Set<string>();
-    for (const ed of elementDefs) {
-      for (const t of ed.type) {
-        if (t.code === 'Reference' && t.targetProfile) {
-          for (const profile of t.targetProfile) {
-            if (profile.startsWith(HL7_FHIR_STRUCTURE_PREFIX)) {
-              targetTypes.add(profile.slice(HL7_FHIR_STRUCTURE_PREFIX.length));
-            }
-          }
-        }
-      }
-    }
-    if (targetTypes.size === 1) {
-      return targetTypes.values().next().value as ResourceType;
-    }
-  }
-
-  return undefined;
 }
 
 /**
--- a/packages/server/src/fhir/searchparameter.ts
+++ b/packages/server/src/fhir/searchparameter.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import type { SearchParameterDetails } from '@medplum/core';
-import { capitalize, getSearchParameterDetails } from '@medplum/core';
+import { capitalize, getSearchParameterDetails, SearchParameterType } from '@medplum/core';
 import type { ResourceType, SearchParameter } from '@medplum/fhirtypes';
 import { AddressTable } from './lookups/address';
 import { CodingTable } from './lookups/coding';
@@ -23,6 +23,19 @@ export const SearchStrategies = {
 export interface ColumnSearchParameterImplementation extends SearchParameterDetails {
   readonly searchStrategy: typeof SearchStrategies.COLUMN;
   readonly columnName: string;
+  /**
+   * For literal (non-canonical) reference search parameters that resolve to
+   * exactly one target resource type on this resource, the name of that
+   * target type (e.g. `Encounter` for `DeviceRequest.encounter`).
+   *
+   * Derived from the resource's element definition target profiles when the
+   * global SearchParameter has multiple targets but the element narrows to one.
+   * Pre-computed at impl build time so that {@link buildReferenceSearchFilter}
+   * can prepend a `Type/` prefix to spec-compliant bare-id search values
+   * without re-reading the SearchParameter for every query. Undefined for
+   * multi-target references, canonical references, and non-reference params.
+   */
+  readonly singleTargetType?: ResourceType;
 }
 
 export interface LookupTableSearchParameterImplementation extends SearchParameterDetails {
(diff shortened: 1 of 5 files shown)


packages/server/src/fhir/searchparameter.ts
```diff
@@ -159,10 +172,63 @@ function buildSearchParameterImplementation(
   const writeable = impl as Writeable<ColumnSearchParameterImplementation>;
   writeable.searchStrategy = 'column';
   writeable.columnName = convertCodeToColumnName(code);
+  if (searchParam.type === 'reference' && impl.type !== SearchParameterType.CANONICAL) {
+    const singleTarget = getSingleReferenceTarget(searchParam, impl);
+    if (singleTarget) {
+      writeable.singleTargetType = singleTarget;
+    }
```
mattwiller: This can be simplified:
```suggestion
    writeable.singleTargetType = getSingleReferenceTarget(searchParam, impl);
```

packages/server/src/fhir/searchparameter.ts
```diff
@@ -159,10 +172,63 @@ function buildSearchParameterImplementation(
...
+          }
+        }
+      }
+    }
+    if (targetTypes.size === 1) {
+      return targetTypes.values().next().value as ResourceType;
+    }
+  }
+
+  return undefined;
```
mattwiller: This can be simplified somewhat to reduce cognitive complexity for the function, with the added bonus of avoiding allocating the `Set`:
```suggestion
  let targetType: ResourceType | undefined;
  for (const ed of elementDefs ?? EMPTY) {
    const refType = ed.type.find((t) => t.code === 'Reference');
    for (const profile of refType?.targetProfile ?? EMPTY) {
      if (profile.startsWith(HL7_FHIR_STRUCTURE_PREFIX) && !targetType) {
        // Found (currently) single type: store for now
        targetType = profile.slice(HL7_FHIR_STRUCTURE_PREFIX.length) as ResourceType;
      } else if (targetType) {
        // Multiple target types exist
        return undefined;
      }
    }
  }
  return targetType;
```

mattwiller (approved):

packages/server/src/fhir/searchparameter.ts
```diff
@@ -159,10 +172,63 @@ function buildSearchParameterImplementation(
...
+  // Derive from element definitions (resource-specific target profiles).
+  const elementDefs = impl.elementDefinitions;
+  if (elementDefs) {
+    const targetTypes = new Set<string>();
+    for (const ed of elementDefs) {
+      for (const t of ed.type) {
+        if (t.code === 'Reference' && t.targetProfile) {
+          for (const profile of t.targetProfile) {
+            if (profile.startsWith(HL7_FHIR_STRUCTURE_PREFIX)) {
+              targetTypes.add(profile.slice(HL7_FHIR_STRUCTURE_PREFIX.length));
```
mattlong: Micro-optimization: could `return undefined` if targetTypes.size === 1 here if we want to shoot the moon on Cognitive Complexity

packages/server/src/fhir/searchparameter.ts
```diff
@@ -159,10 +172,63 @@ function buildSearchParameterImplementation(
...
+ * @param impl - The search parameter implementation details.
+ * @returns The single target resource type, or undefined if there are zero or multiple targets.
+ */
+function getSingleReferenceTarget(
+  searchParam: SearchParameter,
+  impl: SearchParameterDetails
+): ResourceType | undefined {
+  // Fast path: the SearchParameter itself already has a single target.
+  if (searchParam.target?.length === 1) {
+    return searchParam.target[0];
```
mattlong: Is it possible for this to be `Resource`? Is that problematic?

packages/server/src/fhir/searchparameter.ts
mattlong: Actually it'd be even a little more complicated than that since its a Set

packages/server/src/fhir/searchparameter.ts
codyebberson: Yes, fixed

codyebberson:
@claude please review

PR #9326, opened 2026-05-28
feat(data warehouse): add startDate parameter

The-Alchemist:
## Requirement

The data warehouse needs a "low water mark", aka `startDate`, config parameter to determine the _earliest_ date when to start syncing from.  e.g., do not try to sync data earlier than this date.

This allows us to deploy a data warehouse without having it automatically backfill all data, which is unnecessary and will be a separate process. 


packages/server/src/config/utils.ts
```diff
@@ -204,6 +208,28 @@ export function isObjectConfig(key: string): boolean {
...
+ * @param raw - A `Date` instance or date string.
+ * @returns The parsed `Date`, or `undefined` when the string is empty.
+ */
+export function normalizeDate(raw: Date | string): Date | undefined {
+  if (raw instanceof Date) {
+    return raw;
+  }
+  const trimmed = raw.trim();
+  if (!trimmed) {
+    return undefined;
```
mattwiller: Would this cause the config to be silently ignored?  It might be preferable to throw instead, so the user can correct their configuration before deployment/startup

packages/server/src/config/utils.ts
```diff
@@ -204,6 +208,28 @@ export function isObjectConfig(key: string): boolean {
...
+export function isDateConfig(key: string): boolean {
+  return dateKeys.has(key);
+}
+
+/**
+ * Coerces a `Date` or ISO 8601 string to a `Date`.
+ * @param raw - A `Date` instance or date string.
+ * @returns The parsed `Date`, or `undefined` when the string is empty.
+ */
+export function normalizeDate(raw: Date | string): Date | undefined {
```
codyebberson: What are the conditions that we might receive a `Date` object as a config input?

packages/server/src/config/types.ts
```diff
@@ -338,6 +338,11 @@ export interface MedplumDataWarehouseConfig {
   localBasePath?: string;
   /** Optional Iceberg namespace used by sync. */
   namespace?: string;
+  /**
+   * Earliest resource `lastUpdated` timestamp to include in sync.
+   * History rows with `lastUpdated` before this value are excluded.
+   */
+  startDate?: Date;
```
codyebberson: I know this might sound weird, but I have a strong preference for avoiding the `Date` type, and instead using ISO-8601 strings.  That's what we do pretty much everywhere else.

This type needs to be serializable.

packages/server/src/config/utils.ts
The-Alchemist: done

codyebberson: Silently ignoring empty string is fine, and probably correct, because many people will use environment variables for config settings, and we should gracefully support that as a no-op.

Malformed strings should probably be caught and error thrown.

packages/server/src/config/validate-config.ts
```diff
@@ -39,6 +39,10 @@ export function getDataWarehouseConfigErrors(config: MedplumServerConfig): strin
     errors.push('dataWarehouse.destination must be "s3tables" or "local"');
   }
 
+  if (dw.startDate !== undefined && !isDate(dw.startDate)) {
```
codyebberson: 👍

packages/server/src/config/validate-config.ts
```diff
@@ -67,3 +71,7 @@ export function warnInvalidDataWarehouseConfig(config: MedplumServerConfig, logg
     errors,
   });
 }
+
+function isDate(value: Date | undefined): boolean {
```
codyebberson: I'm pretty sure we have a version of this utility already floating around.  Probably more than one.

If we want to be precise, we typically follow the convention that "date" means `YYYY-MM-DD` and "dateTime" or "instant" means `YYYY-MM-DDTHH:MM:SS[TZ]`

So, if we do need to add a new validation path, it should probably be called `isDateTime` or `isInstant`

However, I'd reiterate that I think this setting should be a string, and not a `Date` type, because we basically don't use `Date` type for any persistence, only as a utility for generating timestamps.

codyebberson (changes requested):
Conceptually this looks good.

The main thing, which unfortunately propagates through in quite a few places, is that we prefer not to use the `Date` type in our types.

Let me know if you'd like to discuss more offline

packages/server/src/config/validate-config.ts
The-Alchemist: changed it to a `string`, and validate using `validator` package's `isISO8601()`.  WDYT?

codyebberson:
@claude - my feedback is resolved. please review

codyebberson (approved):
Nice.  See remaining comment about silently ignoring empty string.  Otherwise lgtm

PR #9329, opened 2026-05-28
Updating Candid docs

AndyStoneman:
First pass at updated candid docs. Two big things to verify are:
1. Accuracy of "coming soon" labels in `index.md` file–are they actually coming soon? 
2. Accuracy of FHIR bundle I use as an example.


finnbergquist (commented):
This looks great!

Two pieces of feedback:
- Can you document using the custom fhir operation instead of direct bot invocation
- Can you test uploading our documented bundle, making sure the references all resolve, and then calling the claim submission operation on it

mattlong (approved):

PR #9334, opened 2026-05-29
Require HTTPS for rest-hook Subscription URLs by default

codyebberson:
- Require `https://` endpoints for `Subscription` rest-hook delivery by default
- Add `allowInsecureRestHookUrl` server config override for self-hosted deployments that still need `http://` webhooks
- Add subscription worker tests for default HTTP rejection and explicit insecure URL opt-in

### DIFF
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -40,6 +40,7 @@ import fetch from 'node-fetch';
 import { createHmac } from 'node:crypto';
 import type { Operation } from 'rfc6902';
 import { executeBot } from '../bots/execute';
+import { getConfig } from '../config/loader';
 import type { SubscriptionAutoDisableTrigger } from '../config/types';
 import { WEBSOCKET_SUB_PUBLISH_CHANNEL } from '../constants';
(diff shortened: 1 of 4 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -787,6 +789,25 @@ async function sendRestHook(
...
+
+  if (parsedUrl.protocol === 'https:') {
+    return;
+  }
+
+  if (parsedUrl.protocol === 'http:' && getConfig().allowInsecureRestHookUrl) {
+    return;
+  }
+
+  throw new Error('Invalid rest-hook URL: HTTPS is required unless allowInsecureRestHookUrl is enabled');
```
mattwiller, line 808:
```suggestion
  if (parsedUrl.protocol === 'http:' && getConfig().allowInsecureRestHookUrl) {
    return;
  }

  if (parsedUrl.protocol !== 'https:') {
    throw new Error('Invalid rest-hook URL: HTTPS is required unless allowInsecureRestHookUrl is enabled');
  }
```

mattwiller (approved):

PR #9337, opened 2026-05-29
feat(server): record authenticating ClientApplication on AuditEvent.agent[]

### LINKED ISSUE #9332: AuditEvent.agent[] omits the authenticating ClientApplication
Disclaimer since I'm new to interacting with the repo but the below was written by Claude Code. I've been trying to find a nice workaround for this for the last several hours and hitting a wall. We really want to be able to see which actions were performed by a ClientApplication on behalf of a user in a structured, queryable way. This appears to be supported natively on FHIR and seems like it might be a straightforward enhancement which I'm happy to open a PR for

## Summary

The per-interaction `AuditEvent` Medplum auto-generates records a **single** `agent` — the resource author/profile — and **never records the authenticating `ClientApplication`** as a participant. This holds across auth modes: a normal human-via-app session omits the app; and under `X-Medplum-On-Behalf-Of` the **acting client disappears entirely**, leaving the event indistinguishable from the user acting directly. For reads (no resource `meta` to inspect) there's no other record of who acted.

This looks like a general under-population of `agent[]` relative to the FHIR spec (`agent` is `1..*`, meant to capture both the human and the application). On-behalf-of is the motivating, most acute case. We'd like the authenticating client recorded as an additional `agent[]` participant, and we're happy to send a PR.

## Motivation / use case

A server-side application performs FHIR operations on behalf of a specific user via `X-Medplum-On-Behalf-Of`, so the **user's exact `AccessPolicy` applies** (we don't want the client to exceed the user's scopes — on-behalf-of is ideal). For compliance/forensics we need to answer *"which actions, including reads, did the delegated client perform on behalf of this user vs. the user directly?"* — which the native audit trail can't express today.

The **SMART on FHIR `authorization_code` consent flow** is the canonical instance of this: a user explicitly consents to a third-party `ClientApplication`, which then reads/writes on their behalf (including later via `offline_access`), yet the consented app is recorded **nowhere** on the per-interaction `AuditEvent` (`agent[]`, `meta.author`, and `meta.onBehalfOf` all show only the user). Granted SMART scopes are enforced (effective = user `AccessPolicy` ∩ scopes) but likewise aren't recorded on the AuditEvent — so "which consented app, under what scope, performed this action" is unanswerable from the audit log.

## There's already a precedent in the codebase

Medplum's **bot-execution** AuditEvent already does exactly the dual attribution we're asking for. `createBotAuditEvent` (`packages/server/src/util/auditevent.ts`) emits two agents:

```jsonc
// type.code = "execute"
"agent": [
  { "who": <triggering user>, "requestor": true  },
  { "who": <bot runAs profile>, "requestor": false }
],
"source": { "observer": <Bot> }
```

So multi-agent AuditEvents are already an accepted pattern here. The gap is that the **generic per-interaction path** — `createAuditEvent`, called from `logEvent` in `packages/server/src/fhir/repo.ts` — that *both* bot FHIR ops *and* on-behalf-of requests flow through still emits a single agent. We're asking to bring that path to parity.

There's also a write/audit asymmetry worth noting: writes already record `meta.author` + `meta.onBehalfOf` (`repo.ts`), but `createAuditEvent` ignores both — the audit record drops the second identity the write path already has.

## Current behavior (v5.1.13; `main` is identical)

- `createAuditEvent` (`packages/server/src/util/auditevent.ts`) sets `agent: [{ who, requestor: true, network }]` — one agent — and `source.observer = the server`. No slot for the client.
- `logEvent` passes `who = this.context.author`. `context.author` comes from `getRepoForLogin` (`packages/server/src/fhir/accesspolicy.ts`) as the profile of `membership = onBehalfOfMembership ?? realMembership`.
  - **Normal user session:** `author` = the user's profile; the authenticating `ClientApplication` (`login.client`, populated in `oauth/utils.ts`) is **dropped** from the AuditEvent.
  - **On-behalf-of:** `author` = the **on-behalf-of user**; the acting (admin) client that sent the header is recorded **nowhere** on the event. Most acute case.
  - **`client_credentials`:** covered only incidentally, because there the client *is* the profile.
- Applies to reads/searches too (when `saveAuditEvents` is enabled) — the case with no resource-`meta` fallback.

### Example

```http
GET /fhir/R4/Patient/<id>
Authorization: Bearer <token for ClientApplication/agent-client>
X-Medplum-On-Behalf-Of: ProjectMembership/<user-membership>
```

Resulting `AuditEvent` (abridged) — only the user appears; no sign the delegated client read this:

```json
{
  "resourceType": "AuditEvent",
  "agent": [{ "who": { "reference": "Practitioner/<user>" }, "requestor": true }],
  "source": { "observer": { "identifier": { "value": "https://<server>/" } } }
}
```

## Proposed change

On the per-interaction path, record the authenticating `ClientApplication` as an additional `agent[]` participant (when present), alongside the acting profile — the same shape `createBotAuditEvent` already produces:

```json
{
  "resourceType": "AuditEvent",
  "agent": [
    {
      "who": { "reference": "Practitioner/<user>" },
      "requestor": true,
      "type": { "coding": [{ "system": "http://dicom.nema.org/resources/ontology/DCM", "code": "humanuser" }] }
    },
    {
      "who": { "reference": "ClientApplication/agent-client" },
      "requestor": false,
      "type": { "coding": [{ "system": "http://dicom.nema.org/resources/ontology/DCM", "code": "application" }] }
    }
  ]
}
```

- Keep the existing profile agent as `agent[0]` (`requestor: true`) and **append** the client as a non-requestor agent, so consumers assuming `agent[0]` is the actor don't break.
- Only add the client agent when it's available (`login.client` is absent on the pure `client_credentials`/system path) — don't synthesize one.
- Feasible: `authState.login.client` is in context; it currently just isn't threaded into `createAuditEvent`/`logEvent`.
- *(Secondary, optional)* the granted SMART scope could likewise be recorded on the AuditEvent (it lives on the `Login` today), so the trail captures not just *who* acted but *under what authorization*.

## Questions

1. Would you accept recording the authenticating `ClientApplication` as an additional `AuditEvent.agent[]` participant on the per-interaction path — generalizing what `createBotAuditEvent` already does — upstream?
2. Preferred shape — append a non-requestor `agent[]` entry (with `type` role codings) — and should it be gated behind `X-Medplum: extended` / config to preserve the current single-agent default?
3. Is the client intentionally omitted from per-interaction AuditEvents today for a reason we're missing?

If the direction works, **we're glad to open a PR** — it looks localized to `createAuditEvent`/`logEvent` plus threading `login.client` (and the real-vs-on-behalf-of split) through from context. The same change naturally covers the SMART consent flow and bot FHIR interactions (all share this path).

## Environment

`medplum/medplum-server:5.1.13` (verified identical on `main`), accessed via `X-Medplum-On-Behalf-Of` from a project-admin ClientApplication.

---

Refs: [On-Behalf-Of](https://www.medplum.com/docs/auth/methods/on-behalf-of) · [AuditEvent](https://www.medplum.com/docs/api/fhir/resources/auditevent) · [Bot run-as-user](https://www.medplum.com/docs/bots/bot-run-as-user)

WonderPanda:
Closes #9332

## Summary

The per-interaction `AuditEvent` Medplum auto-generates records a **single** `agent` — the resource author/profile — and never records the authenticating `ClientApplication`. Under `X-Medplum-On-Behalf-Of` the acting client disappears entirely, leaving the event indistinguishable from the user acting directly; the same gap applies to SMART-on-FHIR apps acting on behalf of a user. For reads there is no resource `meta` fallback, so there's no record of which client acted.

This PR threads the authenticating `ClientApplication` (`login.client`) through the repository context and records it as an additional **non-requestor** `agent[]` participant on the per-interaction path — generalizing the dual-agent attribution `createBotAuditEvent` already produces.

## Changes

- `RepositoryContext` gains an optional `client?: Reference<ClientApplication>`.
- `getRepoForLogin` populates it from `authState.login.client`.
- `createAuditEvent` accepts an optional `client` and **appends** it as a non-requestor agent with the DICOM `Application` role coding (`110150`).
- `logEvent` passes `this.context.client` through.

## Behavior

```jsonc
"agent": [
  { "who": { "reference": "Practitioner/<user>" }, "requestor": true },
  {
    "who": { "reference": "ClientApplication/<client>" },
    "requestor": false,
    "type": { "coding": [{ "system": "http://dicom.nema.org/resources/ontology/DCM", "code": "110150", "display": "Application" }] }
  }
]
```

- **Backward compatible:** `agent[0]` is unchanged (still the actor, `requestor: true`), so consumers assuming `agent[0]` is the actor are unaffected.
- The client agent is **appended only when it differs from the actor** — so `client_credentials` (where the client *is* the author) doesn't get a duplicate, and the pure system path (no `login.client`) is unaffected.
- Covers reads/searches (when `saveAuditEvents` is enabled), on-behalf-of, SMART-on-FHIR consent flows, and bot FHIR interactions — all share this path.

## Tests

- Unit tests in `auditevent.test.ts` covering: client appended as second agent; no duplicate when client is the actor; omitted when absent.
- (Drive-by) made the existing `Logs Bot output` test self-sufficient — it relied on leaked global config from an earlier test.

## Open question for maintainers

Issue #9332 raised whether this should be gated behind config / `X-Medplum: extended` to preserve the single-agent default. I've implemented it ungated since the change is additive and `agent[0]` is preserved, but happy to gate it if you'd prefer. The optional secondary suggestion in the issue (recording the granted SMART scope on the AuditEvent) is **not** included here to keep this PR focused.

### DIFF
--- a/packages/server/src/fhir/accesspolicy.ts
+++ b/packages/server/src/fhir/accesspolicy.ts
@@ -99,6 +99,7 @@ export async function getRepoForLogin(authState: AuthState, extendedMode?: boole
     checkReferencesOnWrite: project.checkReferencesOnWrite,
     validateTerminology: project.features?.includes('validate-terminology'),
     onBehalfOf: authState.onBehalfOf ? createReference(authState.onBehalfOf) : undefined,
+    client: login.client,
   });
 }
 
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -53,6 +53,7 @@ import type {
   Binary,
   Bundle,
   BundleEntry,
+  ClientApplication,
   Meta,
   OperationOutcome,
   Project,
@@ -154,6 +155,17 @@ export interface RepositoryContext {
    * This value will be included in every resource as meta.onBehalfOf.
    */
   onBehalfOf?: Reference;
+
+  /**
+   * The authenticating ClientApplication for the current login, when present.
+   * This is the application that obtained the access token, which may differ
+   * from the acting `author` (e.g. when using `X-Medplum-On-Behalf-Of`, or for
+   * a SMART on FHIR app acting on behalf of a user). It is recorded as an
+   * additional non-requestor `agent[]` participant on per-interaction AuditEvents
+   * so the audit trail captures which client performed an action.
+   * Absent on the pure `client_credentials` / system paths.
+   */
+  client?: Reference<ClientApplication>;
 
   remoteAddress?: string;
 
@@ -2072,6 +2084,7 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
         resource,
         searchQuery: query,
         durationMs: options?.durationMs,
+        client: this.context.client,
       }
     );
     logAuditEvent(auditEvent);
--- a/packages/server/src/util/auditevent.test.ts
+++ b/packages/server/src/util/auditevent.test.ts
@@ -8,11 +8,13 @@ import type { BotExecutionRequest } from '../bots/types';
 import { loadTestConfig } from '../config/loader';
 import { globalLogger } from '../logger';
 import {
+  ApplicationAgentType,
   AuditEventOutcome,
   createAuditEvent,
   createBotAuditEvent,
   CreateInteraction,
   logAuditEvent,
+  ReadInteraction,
   RestfulOperationType,
 } from './auditevent';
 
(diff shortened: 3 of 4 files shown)


codyebberson (approved):

codyebberson:
Thanks @WonderPanda !

PR #9339, opened 2026-05-29
fix(agent): fix duplicated messages during upgrade

### LINKED ISSUE #9338: [Agent] Pushed messages can be duplicated in rapid succession when upgrading takes long
Messages being pushed to a device can be duplicated by the upgrading agent for a brief window during upgrade if the handoff from old agent to new agent takes too long

ThatOneBro:
Fixes #9338


packages/agent/src/app.test.ts
```diff
@@ -1779,6 +1779,76 @@ describe('App', () => {
     expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('Transmit response missing callback'));
   }, 5000);
 
+  test('Agent transmit response while status is off returns disabled error', async () => {
```
ThatOneBro, line 1783: Noticed this branch was not covered so opportunistically added a test for it

packages/agent/src/app.ts
```diff
@@ -183,20 +194,23 @@ export class App {
...
+        // handling data-plane messages.
+        this.isPrimary = true;
       } catch (_err) {
         this.log.info('Unable to create agent PID file, trying again...');
         attempt++;
         if (attempt === maxAttempts) {
           throw new Error('Too many unsuccessful attempts to create agent PID file');
         }
-        await sleep(500);
+        await sleep(50);
```
ThatOneBro, line 248: This is to shorten the window before we "unlock" and allow the agent to respond to incoming messages

packages/agent/src/app.test.ts
```diff
@@ -1779,6 +1779,76 @@ describe('App', () => {
...
+    const timeout = setTimeout(() => {
+      shouldThrow = true;
+    }, 2500);
+    while (!state.agentError) {
+      if (shouldThrow) {
+        throw new Error('Timeout');
+      }
+      await sleep(100);
+    }
+    clearTimeout(timeout);
```
mattwiller, line 1842: This seems like something we'd want to re-use across tests; should it be extracted into a helper?

mattwiller (approved):

packages/agent/src/app.test.ts
ThatOneBro: Yeah tbh this is copy-pasta across many generations of tests. Might be worth it at this point

ThatOneBro: I'll put that in a follow-up

PR #9340, opened 2026-05-29
fix(server): don't refresh if membership inactive

ThatOneBro:
Prevents users with an inactive ProjectMembership but a non-revoked / expired refresh token from spamming with 401s. Previously we would give an access token anyways, which the client would attempt to use and immediately receive a 401, triggering another refresh, in a tight loop

### DIFF
--- a/packages/server/src/oauth/token.ts
+++ b/packages/server/src/oauth/token.ts
@@ -302,7 +302,7 @@ async function handleRefreshToken(req: Request, res: Response): Promise<void> {
 
   if (login.membership) {
     const membership = await systemRepo.readReference<ProjectMembership>(
-      login.membership as Reference<ProjectMembership>
+      login.membership
     );
     if (membership.active === false) {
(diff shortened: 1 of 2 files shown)


The-Alchemist (approved):

packages/server/src/oauth/token.ts
```diff
@@ -300,6 +300,14 @@ async function handleRefreshToken(req: Request, res: Response): Promise<void> {
     return;
   }
 
+  if (login.membership) {
+    const membership = await systemRepo.readReference<ProjectMembership>(login.membership);
+    if (membership.active === false) {
+      sendTokenError(res, 'invalid_grant', 'Profile not active');
```
mattwiller:
```suggestion
      sendTokenError(res, 'access_denied', 'Profile not active');
```

mattwiller (approved):
I think another error code would be more correct, but good either way

PR #9341, opened 2026-05-29
fix(core,react-hooks): don't refresh subs if unauthenticated

ThatOneBro:
It's currently possible that if a developer does not guard against unauthed access to component which mount `useSubscription` that unauthenticated requests to create subscriptions could be automatically attempted and cause noisy errors. This PR should clean that up by making the `useSubscription` hook no-op when unauthenticated.

Separately but related, we also now check that a user is authenticated before attempting to refresh subscriptions that have tokens which will expire soon (or maybe already expired while a tab was backgrounded).

### DIFF
--- a/packages/react-hooks/src/useSubscription/useSubscription.ts
+++ b/packages/react-hooks/src/useSubscription/useSubscription.ts
@@ -92,7 +92,10 @@ export function useSubscription(
     }
 
     let shouldSubscribe = false;
-    if (prevCriteriaRef.current !== effectiveCriteria || !deepEquals(prevMemoizedSubPropsRef.current, memoizedSubProps)) {
+    if (
+      prevCriteriaRef.current !== effectiveCriteria ||
+      !deepEquals(prevMemoizedSubPropsRef.current, memoizedSubProps)
+    ) {
       shouldSubscribe = true;
     }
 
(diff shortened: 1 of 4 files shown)


packages/react-hooks/src/useSubscription/useSubscription.ts
```diff
@@ -44,6 +44,12 @@ export function useSubscription(
   options?: UseSubscriptionOptions
 ): void {
   const medplum = useMedplum();
+  const profile = useMedplumProfile();
+  // When the user is not authenticated, the subscription becomes a no-op. Subscribing would
+  // create a `Subscription` resource and request a WebSocket binding token, both of which 401
+  // without a profile. Treating criteria as `undefined` skips all subscription work; when the
+  // user authenticates, `profile` changes and the effect re-runs to subscribe.
+  const effectiveCriteria = profile ? criteria : undefined;
```
mattwiller, line 52: To avoid having two variables and needing to update all the code below to use the right one, and to align with the above comment, I'd suggest just reassigning the input param directly here:
```suggestion
  if (!profile) {
    criteria = undefined;
  }
```

mattwiller (approved):

packages/react-hooks/src/useSubscription/useSubscription.ts
ThatOneBro: It seems our `react-hooks/immutability` rule doesn't like parameter reassignment. It's only a warning but eh I still hate seeing those lines.

I think I do prefer what you presented but our rules do not. Might be worth re-evaluating the rule?

mattwiller: No worries, I'm fine with it either way

PR #9343, opened 2026-05-29
Repository resource type access tracking

mattlong:
**Note**: This PR is chained on #9285 since they both touch the same parts of Repository and related code.

Add resource-type access tracking to the FHIR repository

This branch instruments the server's FHIR repository layer to record which resource types are touched by every database and cache operation, with the goal of identifying transactions that mix "special" platform resource types (Project, ProjectMembership, User) with ordinary resource types. This is groundwork for a future effort to split the repository/datastore (logged under the [RepoSplit] tag). Down the road, having this resource type access metadata will enable enforcement of sharding boundaries by throwing when violations are detected rather than just logging.

### What changed

New access-tracking layer (packages/server/src/fhir/repository/access-tracker.ts)
- Adds RepositoryAccessTracker, which partitions accessed resource types into "special" (Project, ProjectMembership, User) vs. "other" and maintains a stack of per-transaction frames (counts of SQL/cache reads & writes, the resource types touched, and the calling source).
- Frames follow the transaction lifecycle: pushed on BEGIN/savepoint, merged on nested commit/rollback, popped on outer commit/rollback, and cleared on connection reset.
- Emits structured getLogger().info('[RepoSplit]...') logs at the statement level (when a single op mixes special + other types) and at the transaction level (when a committed/rolled-back transaction touched both classes).
- Exposes getResourceTypesFromReferences() helper.

Repository connection (repository-connection.ts)
- getDatabaseClient(mode) --> getDatabaseClient(options), where options carry { operation, mode, resourceTypes, source }, and each call records access.
- withTransaction now requires TransactionSqlOptions (resource types + source) and no longer passes the raw PoolClient to callbacks; withStatementTimeout similarly drops the client param.
- withStatementTimeout is relaxed to allow use inside an active transaction (using set_config(..., is_local=true)), instead of throwing; it still guards against mutating a borrowed connection's config. assertOwnsClient takes an optional
custom message.

fhir-router (batch.ts, repo.ts, core/search.ts)
- Introduces TransactionOptions { resourceTypes, serializable? }; FhirRepository.withTransaction now takes it as a required argument.
- Batch/transaction preprocessing collects the set of resource types from parsed routes (documented limitations: system-level interactions, GraphQL, and custom operations are under-reported).
- Conditional create/update/delete/patch pass the search's resource types via the new getSearchResourceTypes() helper in @medplum/core.

Repo & search wiring (repo.ts, search.ts)
- Every getDatabaseClient, withTransaction, and withStatementTimeout call site now supplies resourceTypes and a descriptive source string (e.g. repo.writeToDatabase, search.getSearchEntries).
- Search query building threads a trackedResourceTypes set through select/filter/chained-search construction so chained and _type-multi searches accumulate all involved resource types.

Callers updated to satisfy the new signatures: admin (invite/project/super), auth (verifyemail), value-set/code-system/concept-map operations, scheduling (book/cancel), expunge, rescope, rotatesecret, update-user-email, refresh-reference-display, explain, db-configure-column-statistics, reindex, subscription worker, seed, and migration utils.

Tests & helpers
- Adds mockStdoutWrite() test helper to suppress logger output.
- Updates repo.test.ts, transaction.test.ts, repo-guard.test.ts, search.test.ts, reference.test.ts, reindex.test.ts, and batch.test.ts for the new APIs, plus new coverage for batch resource-type collection and access logging.

Notes

- The tracking is observational (logging only) and does not change query behavior or transaction semantics.
- Resource-type coverage for GraphQL and custom operations is intentionally incomplete and noted in code comments.

### DIFF
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -36,8 +36,10 @@ export type ReadHistoryOptions = {
   limit?: number;
 };
 
+type ResourceTypeInput = ResourceType | readonly ResourceType[] | ReadonlySet<ResourceType>;
+
 export interface TransactionOptions {
-  readonly resourceTypes: Iterable<ResourceType>;
+  readonly resourceTypes: ResourceTypeInput;
   readonly serializable?: boolean;
 }
 
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -12,6 +12,7 @@ import {
   deepClone,
   evalFhirPath,
   generateId,
+  getSearchResourceTypes,
   globalSchema,
   matchesSearchRequest,
   multipleMatches,
@@ -20,7 +21,7 @@ import {
   preconditionFailed,
   stringify,
 } from '@medplum/core';
-import type { Bundle, OperationOutcome, Parameters, Reference, Resource } from '@medplum/fhirtypes';
+import type { Bundle, OperationOutcome, Parameters, Reference, Resource, ResourceType } from '@medplum/fhirtypes';
 
 export type CreateResourceOptions = {
   assignedId?: boolean;
@@ -34,6 +35,11 @@ export type ReadHistoryOptions = {
   offset?: number;
   limit?: number;
 };
+
+export interface TransactionOptions {
+  readonly resourceTypes: Iterable<ResourceType>;
+  readonly serializable?: boolean;
+}
 
 export const RepositoryMode = {
   READER: 'reader',
@@ -195,10 +201,12 @@ export abstract class FhirRepository {
    * Runs a callback function within a transaction.
    *
    * @param callback - The callback function to be run within a transaction.
+   * @param options - The transaction options.
+   * @returns The result of the callback function.
    */
   abstract withTransaction<TResult>(
     callback: (txRepo: this) => Promise<TResult>,
-    options?: { serializable?: boolean }
+    options: TransactionOptions
   ): Promise<TResult>;
 
   /**
@@ -280,7 +288,10 @@ export abstract class FhirRepository {
         const createdResource = await txRepo.createResource(resource, options);
         return { resource: createdResource, outcome: created };
       },
-      { serializable: true } // Requires strong transactional guarantees to ensure unique resource creation
+      {
+        resourceTypes: getSearchResourceTypes(search),
+        serializable: true, // serializable to ensure unique resource creation
+      }
     );
   }
 
@@ -340,7 +351,7 @@ export abstract class FhirRepository {
         const updated = await txRepo.updateResource({ ...resource, id: existing.id }, options);
         return { resource: updated, outcome: allOk };
       },
-      { serializable: true }
+      { serializable: true, resourceTypes: getSearchResourceTypes(search) }
     );
   }
 
@@ -374,7 +385,7 @@ export abstract class FhirRepository {
         const resource = matches[0];
         await txRepo.deleteResource(resource.resourceType, resource.id);
       },
-      { serializable: true }
+      { serializable: true, resourceTypes: getSearchResourceTypes(search) }
     );
   }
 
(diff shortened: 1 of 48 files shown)


mattlong:
@claude review

codyebberson (approved):

codyebberson:
@claude - please re-review since my previous approval

packages/fhir-router/src/repo.ts
```diff
@@ -36,6 +36,13 @@ export type ReadHistoryOptions = {
   limit?: number;
 };
 
+type ResourceTypeInput = ResourceType | readonly ResourceType[] | ReadonlySet<ResourceType>;
```
codyebberson, line 39: Can we avoid this and just pick one for an internal type?

packages/fhir-router/src/repo.ts
```diff
@@ -664,3 +676,7 @@ const sortComparator = <T extends Resource>(a: T, b: T, sortRule: SortRule): num
   const bStr = JSON.stringify(evalFhirPath(expression, b));
   return aStr.localeCompare(bStr) * (sortRule.descending ? -1 : 1);
 };
+
+function getSearchResourceTypes(searchRequest: SearchRequest): ResourceTypeInput {
+  return searchRequest.types ?? [searchRequest.resourceType];
```
codyebberson: I think this essentially only returns `ResourceType[]`, yeah?

packages/server/src/fhir/lookups/reference.test.ts
```diff
@@ -34,6 +34,12 @@ describe('ReferenceTable', () => {
     return a.code.localeCompare(b.code);
   }
 
+  function getReferenceTestClient(
+    resourceTypes: ResourceType | ResourceType[]
+  ): ReturnType<typeof systemRepo.getDatabaseClient> {
```
codyebberson: We can use proper type here, no?  `Pool | PoolClient`?

packages/server/src/fhir/repository/access-tracker.ts
```diff
@@ -0,0 +1,317 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
```
codyebberson, line 1: Initial reaction is that this file feels like a lot. 

Is it your expectation that we only need this while we're navigating the transition to sharding and this goes away?  Or are we keeping this forever?

packages/server/src/fhir/repo.ts
```diff
@@ -416,6 +418,61 @@ export class Repository extends FhirRepository implements Disposable {
     }
   }
 
+  async sqlRead<R extends QueryResultRow = any>(
```
codyebberson, line 423: K, I feel like we need to back up and discuss strategy.  We have made it a very long way without needing this much `executeSql` code, and I'm getting a bit spooked that we're adding so much hardware for these.  Let's discuss offline

codyebberson (changes requested):
I see where we're going overall, but I'm afraid that the tail is wagging the dog here, and we ended up introducing a disproportionate amount of machinery.  Let's discuss in next eng sync.

codyebberson (approved):

packages/server/src/fhir/repository/access-tracker.ts
mattlong: Discussed offline.
tl;dr The access tracking is intended to be transitory. Eventually it will morph into the driver of Repository smart routing and enforcement in a sharded environment.

packages/server/src/fhir/repo.ts
mattlong: Discussed offline.

The concern is that these methods encourage one-off DB access that doesn't flow through the established interface of a Repository, i.e. createResource, updateResource, etc. `sqlRead` and `sqlWrite` are window dressing on top of the new `executeSql` which is itself meant to completely replace `getDatabaseClient`. The intent is definitely not to encourage wider use of Repository's DB access methods; instead it is meant to route all DB access through a Repository itself alongside resource type declarations to enable DB routing rather than a raw DB PoolClient.

PR #7677, opened 2025-10-24
feat(agent): add durable queue for HL7 messages

ThatOneBro:
Fixes #7578

PR #9346, opened 2026-05-29
feat(agent): add durable queue for HL7 messages

### LINKED ISSUE #7578: Make agent HL7 queue durable
Some overlap with #7006, the goal is to make the HL7 message queue for channels durable to service restarts, network outages, etc. by using a local SQLite database

Some interesting considerations include:
- What to do in regards to database migrations between versions
- How to handle downgrades which may have different database schemas

ThatOneBro:
Fixes #7578. Replaces #7677 (closed). This is a significant simplification of that original, which is why it lands after months of agent hardening that made the zero-downtime upgrade less flaky and the artifact-download path more observable.

## Summary

The Agent buffers inbound HL7 v2 traffic exclusively in the JS heap (`App.webSocketQueue`, per-connection sequence counters, pending `sendAndWait` callbacks). If the agent crashes, is killed, or is upgraded, in-flight inbound messages are lost. This is worst in **enhanced mode**, where `Hl7Connection` sends the sender a **commit ACK (CA) on receipt** — so the source system believes the message is durably persisted when it never left RAM. That is a silent violation of the durable-commit contract.

This PR replaces the in-memory inbound path with a **durable, SQLite-backed, per-channel FIFO**. Every inbound message is persisted **on the same TCP turn it is received, before any ACK is returned**, and a per-channel worker drains the table to the Medplum server in order. Queued work survives process restart; work interrupted mid-flight is quarantined for review rather than silently dropped or blindly re-sent.

**Scope:** inbound only (channel → server). The outbound path (server → channel) stays in-memory and is unchanged. Processing is **serial per channel** in v1 (one message in flight per channel; different channels run in parallel).

## The durable contract, precisely

`CA` is sent **only after the durable write commits**. To make this honest, the agent keeps `enhancedMode` *off* the `Hl7Connection` (suppressing the library's synchronous auto-ACK) and sends the commit ACK itself after the row is written:

- **Stored OK →** `sendCommitAck` (CA, or AA in aaMode). The row's `received_at` is the commit timestamp.
- **Duplicate control ID (reject mode, or content-mismatch in idempotent mode) →** terminal NACK (CR / AR) + a `nacked` audit row.
- **Storage error (disk full / DB locked / busy) →** **retryable** NACK (CE / AE) so the peer safely retransmits; only a best-effort `nacked` audit row is written, never a committed one.

## Architecture

```
source device (HL7/MLLP/TCP)
        │
        ▼
AgentHl7Channel (per-port server)
        │  1. INSERT inbound row (synchronous)        ──┐
        │  2. sendCommitAck on commit (CA/AA)           │  packages/agent/src/hl7.ts
        │  3. sendCommitNack on error  (CE/AE | CR/AR)  ──┘
        ▼
inbound_hl7_messages  ── SQLite (node:sqlite, WAL)  ── packages/agent/src/queue/{schema,durable-queue}.ts
        │  (worker notified)
        ▼
ChannelQueueWorker (1 per channel, serial)          ──  packages/agent/src/queue/worker.ts
        │  agent:transmit:request  (over WS)
        ▼
Medplum server → Bot  ──→  agent:transmit:response
        │
        ▼
back to worker → channel.sendToRemote(response)  (app-level ACK to source)
```

**New modules (all under `packages/agent/src/queue/`):**

| Module | Responsibility |
| --- | --- |
| `durable-queue.ts` | Owns the `node:sqlite` handle. Typed CRUD: `enqueue` (sole dedup + seq-no authority), `claimNext`, `markSent`, `requeue`, `markProcessed/Rejected/Failed`, `setAckOutcome`, `recordServerResponse`, `recoverOnStartup`, `checkpointWalIfDirty`, `close`. **Also owns the dispatch-lease loop** — `startDispatchLease`/`stopDispatchLease`/`isLeader` plus the lease primitives and the `assertNotDemoted()` dispatch gate (formerly a separate `DispatchLeaseManager`; see *Retention & lease*). |
| `schema.ts` | Versioned DDL + migration runner (incl. the single-row `_lease` table). |
| `worker.ts` | `ChannelQueueWorker` — one per channel; serial dequeue loop; correlates WS responses back to the row; self-terminates on `QueueLeaseError` when a peer takes the lease. |
| `retention.ts` | `RetentionSweeper` — time + size purge with a protection floor. |
| `queries.ts` / `types.ts` | Centralized hot-path SQL; `MessageState` / `AckOutcome` / `QueueErrorCode` / `InboundRow`. |

`packages/agent/src/hl7.ts` holds the deferred-commit-ACK state machine and idempotent-replay logic; `packages/agent/src/app.ts` wires queue lifecycle into start/stop, the dispatch-lease handoff, and the heartbeat.

### Storage: `node:sqlite` + WAL

- **`node:sqlite`** (stable in Node 24; behind `--experimental-sqlite` in 22.18+) — comparable to better-sqlite3 but **no native dependency to cross-compile per platform for the agent's SEA build**.
- **WAL** with `synchronous = NORMAL` (durable across process crash; writers don't block readers), plus tuned `cache_size` / `mmap_size` / `busy_timeout`.
- DB file: `<logDir>/medplum-agent-queue.sqlite` (+ `-wal` / `-shm` sidecars). Overridable via `Agent.setting[name=queueDbPath].valueString`. **Ops note:** the new file *and its `-wal`/`-shm` sidecars* must be on a persistent path / volume mount (installer + Dockerfile).

### Two-axis disposition model

A row's fate lives on **two independent axes**, because the two network legs fail independently:

- **`state`** — the **Bot leg**: `queued → claimed → inflight → processed | rejected | failed` (plus `nacked` for intake rejects). The split between `claimed` (worker has claimed the row but its `agent:transmit:request` is **not yet on the wire** — still buffered in the in-memory WS queue) and `inflight` (request **written to the socket**, `sent_at` stamped) is what makes crash recovery safe — see below.
- **`ack_outcome`** — the **source leg**: `pending | delivered | undelivered | not_owed`.

This split is what lets the system represent "the Bot accepted the message (`processed`) but the source connection had already closed (`undelivered`)" as a **non-failure that is never re-dispatched** — see Testing for why this mattered against Mirth.

`QueueErrorCode` further classifies `failed` rows as **transient** (`server-error`/5xx, `server-rate-limited`/429 — provably produced no result) vs **ambiguous** (`response-timeout`, `interrupted`, `worker-stopped`, `dispatch-failed` — the server *may* have processed it). This classification is recorded now and is the gate for the auto-retry fast-follow (below).

### Crash recovery

`recoverOnStartup()` runs once the process acquires the dispatch lease (at startup, or on a later takeover — it is idempotent), before any channel starts claiming:

- `queued` rows **resume** automatically.
- `claimed` rows — the request provably never hit the socket (`sent_at` is `NULL`) — are returned to `queued` (attempt un-counted) for **zero-duplicate-risk** retry.
- `inflight` rows are promoted to `failed` (`interrupted`), with `ack_outcome` left `pending` (genuinely unknown) — quarantined for review/replay rather than blindly retried, because the request reached the wire and the Bot may have processed it.

The server response body + status code are persisted on the row, so an interrupted ACK can be completed on recovery.

### Two recovery paths for in-flight work

- **Path 1 — automatic requeue on WS disconnect.** If a row is `claimed` (its `agent:transmit:request` still sitting **unsent** in the in-memory WS queue) when the WS drops, the server provably never saw it → the request is pulled and the row returns to `queued` (`DurableQueue.requeue`, which also un-counts the attempt) for zero-duplicate-risk retry on reconnect. If the request already went out (`inflight`), delivery is ambiguous and the row is left to the response timeout. The on-disk `claimed`/`inflight` split mirrors this unsent-vs-sent distinction durably.
- **Path 2 — manual operator replay (v1).** `failed` rows can be reset to `queued` via documented SQL (`UPDATE … SET state='queued', attempt_count=0 WHERE id=? AND state='failed'`). Never `rejected`, never `processed`.

### Retention & lease

- **`RetentionSweeper`**: time-based + size-driven purge, oldest-first, with a protection floor that keeps recent `failed` rows and `processed`+`undelivered` rows (the actionable operator signal) beyond the normal window. Runs an immediate sweep at startup, then on an interval. Runs **regardless of dispatch leadership** (a follower still accepts intake writes, so it has its own WAL/disk to manage).
- **Dispatch lease** (folded into `DurableQueue` — `startDispatchLease`/`stopDispatchLease`/`isLeader` over a single-row `_lease` table; formerly a standalone `DispatchLeaseManager`): exactly one process may **dispatch** at a time, so the brief two-agent overlap during a zero-downtime upgrade can't double-process. **The lease gates dispatch only** — a follower still persists every inbound message (intake), checkpoints its WAL, and runs retention; only *claiming + driving rows* and `recoverOnStartup` are leader-restricted. Acquire on start, heartbeat to extend, release on `stop()` **before** closing the DB (so a waiting peer takes over immediately instead of waiting out the TTL), or take over after a dead holder's TTL lapses. **There is no `onLostLeadership` callback**: leadership loss surfaces as a `QueueLeaseError` thrown by `assertNotDemoted()` on every dispatch-class mutation (the load-bearing, data-layer check), and the demoted worker tears *itself* down — leaving its in-flight row unsettled (`claimed`/`inflight`) for the new leader's `recoverOnStartup` to reconcile. A heartbeat-driven watchdog cancels a wedged in-flight dispatch so a demoted worker doesn't wait out the full response timeout.

## Notable fixes & decisions made along the way

The branch is mostly one feature commit followed by a long hardening tail. The decisions worth calling out:

**ACK protocol correctness**
- **All deferred-ACK logic lives in the agent, not `@medplum/hl7`** (`f42b325c` → reverted/reimplemented in `59b0af7b`). After trying to thread the deferred ACK through `Hl7Connection`, the cleaner end state was to revert `@medplum/hl7` entirely and keep the whole ACK state machine in the agent, where it can be reasoned about in one place and the shared library stays untouched.
- **Retryable vs terminal NACK codes** (`6812709e`): storage failures now send **CE/AE (retryable)** rather than **CR/AR (terminal)**. Since a storage failure writes only an audit row (never a committed one), the peer's retransmit is safely accepted as fresh — telling it to give up would lose a message we never stored.
- **Don't discard late ACKs** (`cd5cf134`): a server response arriving after the response-timeout used to fall through and replay a *stale* ACK. Now the server is authoritative — a late 2xx/4xx settles the already-timed-out row to `processed`/`rejected`, and duplicate responses for settled rows are ignored. (A late response that arrives at an agent which has *also* been demoted is dropped rather than written — `f3edc71a`.)
- **Suppress the app-level ACK in aaMode** (`ac1df23c`): in aaMode the source already received its AA at intake (the deferred commit ACK), so the Bot's later app-level AA is a redundant *second* AA that races sources which close on ACK. Mark `processed`+`delivered` without re-sending.
- **Record ACK stats on every path** (`66469354`, `98640abc`): the deferred / suppressed / replayed ACK paths must call `recordAckReceived`, or the control ID lingers in the pending-RTT map until the 5-minute GC warns "never got a response."

**Dedup & sequence-number integrity**
- **Single dedup authority + atomic seq-no** (`164abf6a`, `39be3b68`): all duplicate detection moved into `DurableQueue.enqueue()` (the only writer), and MSH.13 assignment happens via an `assignSeqNo` callback **inside the same transaction as the insert**. A duplicate never consumes a sequence number, and a committed row's number is guaranteed persisted. This removed a pre-check race between the connection layer and the queue.
- **Idempotent retransmit replay** (`hl7.ts`): a byte-for-byte resend replays the stored server-response ACK (or the commit ACK if not yet dispatched) rather than re-dispatching to the Bot; a *different* body reusing a committed control ID is rejected (AR/CR) with a `nacked` audit row.

**Dispatch leadership & worker lifecycle**
- **Split `processing` into `claimed` and `inflight`** (`3613c10d`, with `7cdd751c` rolling `inflight` → `claimed` back if the socket write fails): a single `processing` state couldn't distinguish "claimed but the request never hit the wire" (provably unseen by the server — safe to requeue) from "request sent, response unknown" (ambiguous — must quarantine). Splitting it records the recovery decision on the row itself (`sent_at` NULL vs stamped) instead of a fragile in-memory check.
- **Fold the dispatch lease into `DurableQueue`** (`6f4b16f8`): the leader-election loop was a separate `DispatchLeaseManager`. Folding it in makes the queue the single point of interaction for the `App` and lets the loop drive the same lease primitives it already owned — no second object to keep in sync. The lease gates **dispatch only** (claim + `recoverOnStartup`), never intake/maintenance/diagnostics, which is what makes the upgrade overlap safe.
- **No `onLostLeadership` callback — enforce demotion at the data layer** (`d9aa4398`): rather than pushing leadership loss down a callback chain, every dispatch-class mutation (`claimNext`, `recordServerResponse`, `markProcessed/Rejected/Failed`, `requeue`) first calls `assertNotDemoted()`, which throws `QueueLeaseError` once a peer owns the `_lease` row. A demoted process is *physically unable* to write dispatch state regardless of what its in-memory leader flag believes; the worker catches the error, stops itself, and leaves its in-flight row for the new leader to reconcile.
- **Harden the leader-callback path** (`6033804e`, `1055784d`): `onBecameQueueLeader` and `forEachChannelWorker` catch per-channel errors so one channel failing to start (or a throwing WS handler) can't abort bringing up the rest.

**Durability & lifecycle**
- **Two-axis disposition** (`99342610`): split Bot-leg `state` from source-leg `ack_outcome` (see Architecture) so independent leg failures stop being conflated.
- **Requeue provably-unsent transmits on disconnect** (`862857df`): the Path-1 logic above — distinguishes "provably unsent" (`claimed`, safe to requeue) from "ambiguous" (`inflight`, leave to timeout).
- **WAL durability on teardown and when idle** (`28743a49`, `1514409f`, `59f80813`): explicit `PRAGMA wal_checkpoint(TRUNCATE)` on `close()` (a peer connection during upgrade overlap defeats SQLite's implicit last-connection checkpoint); a bounded drain (`raceWithTimeout` + `allSettled`) so one hung channel can't block queue teardown; WAL checkpointing pushed *into* `DurableQueue` (`checkpointWalIfDirty()` on each heartbeat, leadership-independent) and an immediate retention sweep at startup so idle / frequently-restarted agents don't grow the WAL or skip retention.
- **Error classification** (`e24e9f99`): the `QueueErrorCode` transient/ambiguous taxonomy that underpins the future auto-retry.
- **Schema hardening** (`6b330cb4`, `bf9cec79`): centralized hot-path SQL into `queries.ts` with `EXPLAIN QUERY PLAN` regression guards.

### Integration tests — `packages/agent/src/queue/integration.test.ts` (~20 scenarios)

End-to-end: a real `App` + a real HL7 client + a mock Medplum WS server + a real SQLite queue. **Purpose:** prove the durable-CA contract holds across the failure modes a heap-only queue cannot survive. **Scope** includes:

- Happy path (CA after enqueue → `processed`+`delivered`); aaMode commit-AA balances pending RTT.
- Server **5xx → `failed`** (transient, ack not owed) and **4xx → `rejected`** (permanent, never re-dispatched).
- Idempotent duplicates: replay the prior server-response ACK (post-dispatch) or the commit ACK (pre-response); content mismatch on a seen control ID → AR.
- `assignSeqNo`: retransmit dedupes on the *original* despite a new MSH.13; duplicates and storage errors **don't consume a sequence number**; audit rows record the right number.
- **Restart recovery**: rows left `inflight` across a process boundary become `failed` (`interrupted`); rows left `claimed` (unsent) requeue and re-dispatch.
- **Leadership handoff**: a graceful `stop()` releases the lease before closing the DB and a waiting peer takes over; an ungraceful lease yank from a still-running leader (in-flight row wedged) stops it driving the queue *without* settling the row, and the new owner's `recoverOnStartup` reconciles the abandoned row.
- **`spam-while-stopping`**: messages hammer the channel as `App.stop()` tears down — every CA the source received has a durable row after shutdown (verified against WAL state).
- Heartbeat ticks flush the WAL; `stop()` flushes the WAL even when a channel fails to stop.

### Unit tests

- `durable-queue.test.ts` — enqueue/dedup, FIFO `claimNext`, `markSent`, `requeue`, terminal transitions, `claimed`/`inflight` recovery (requeue claimed, fail inflight), peek-vs-commit seq-no semantics, lease primitives, WAL checkpoint, migration idempotency.
- `worker.test.ts` — FIFO dispatch, 5xx/4xx/429 handling, response timeout, **late-response settlement**, offline queueing + drain on reconnect, **unsent-requeue (`claimed`) vs sent-ambiguous (`inflight`)** on disconnect, and **lease theft** — a peer steals the lease while the worker is idle (next `claimNext` throws → worker self-stops) or wedged in-flight (heartbeat watchdog cancels the dispatch before the response timeout, row left unsettled).
- `durable-queue-lease.test.ts` / `dispatch-lease.test.ts` — dispatch-lease leader election: acquire/expire/takeover/heartbeat/release, idempotent `startDispatchLease`/`stopDispatchLease`, and demotion enforced at the data layer via `QueueLeaseError` (there is no lost-leadership callback).
- `retention.test.ts` — time + size purge, the undelivered/recent-error protection floor, immediate startup sweep.
- `hl7-connection.test.ts` — CA/AA/CE/CR/AE/AR codes and the legacy-vs-durable enhanced-mode wiring (auto-ACK on vs off).

### Regression tests (added alongside their fixes)

| Guards against | Test |
| --- | --- |
| aaMode dangling pending RTT | `aaMode happy path: the commit AA balances the pending RTT entry …` |
| channel-stop failure skipping queue teardown | `stop() closes the durable queue and flushes the WAL even when a channel fails to stop` |
| seq-no consumed on storage error / mismatch | `assignSeqNo: a storage error / content-mismatch …` (3 tests) |
| ACK-undeliverable + retransmit replay | `ack-delivery failure (source connection closed) → processed+undelivered … retransmit replays and flips to delivered` |
| dropping non-UTF-8 messages at intake | `non-UTF-8 channel encoding (iso-8859-1): message is accepted, CA sent, and forwarded text preserved` |
| late server response at a demoted agent | `late ACK but this agent has been demoted already` |

### Query-plan / index regression guards — `query-plan.test.ts`

For every hot-path query, `EXPLAIN QUERY PLAN` asserts it uses the intended index and **never full-scans `inbound_hl7_messages`** — so a future query reshape that drops an index fails CI immediately.

### Manual testing against a live Mirth deployment

- **Full bidirectional round trip:** Mirth sending channel → Agent (durable inbound queue) → Medplum → Bot → creates `MessageHeader` → Subscription fires → second Bot → `pushToAgent` → Agent (in-memory outbound) → Mirth receiving channel. Confirms the new inbound queue carries real traffic end-to-end without breaking the existing outbound/round-trip flow.
- **Connectivity loss under sustained load** (~4 messages/sec/channel): dropped connectivity mid-stream; queued work resumed and provably-unsent (`claimed`) transmits requeued on reconnect with no loss.
- **Connection thrashing (no keep-alive — a new TCP connection per inbound message):** surfaced a real bug in the **ACK-delivery / replay** path. Because the per-message source connection had already closed by the time the Bot responded, `sendToRemote` couldn't deliver the app-level ACK — the row correctly went `processed`+`undelivered`, and Mirth (never having seen the ACK) retransmitted. The fix makes that retransmit **replay the stored server-response ACK and flip the leg to `delivered`, with no Bot re-dispatch** (`hl7.ts` `replayServerAck` + `setAckOutcome`, on the two-axis model from `99342610`). Now covered by the `ack-delivery failure …` regression test above.
- **Agent upgrade during high traffic:** zero-downtime upgrade (dispatch-lease handoff + WAL flush on teardown) under load with **no lost messages**.

## Follow-ups (fast-follow)

- **Auto-retry of transient failures.** `failed` is not uniform. Only **transient** codes (5xx, 429) are safe to auto-redispatch — the request provably produced no result. **Ambiguous** codes (`response-timeout`, `interrupted`, `worker-stopped`, `dispatch-failed`) mean the server *may* have processed the message, so auto-redispatch risks duplicate upstream processing and stays operator-gated. The `QueueErrorCode` classification in this PR exists precisely to gate that policy; the redispatch loop itself is the fast-follow.
- **Per-channel concurrency** (`maxConcurrentPerQueue`). v1 keeps one message in flight per channel; a configurable concurrency limit lets a channel keep N messages in flight to the Bot at once for higher throughput, while preserving the durability guarantees here.


ThatOneBro:
@claude There is a lot of context here, please review this PR, taking in the context of the architecture in `DURABLE_QUEUE_ARCHITECTURE.md` and also the comments throughout the diff

ThatOneBro:
@claude I addressed all your issues except #2 and #5. #2 does not look like a real issue since it's static and #5 is already documented in the comments as being correct but wasteful, it's just simpler for now. Can you take another pass at a review?

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
```diff
@@ -0,0 +1,801 @@
...
+Initial migration (v1) is the DDL above. Subsequent migrations are append-only — never edited in place.
+
+---
+
+## 4. Message Lifecycle
+
+A row's disposition lives on **two independent axes**, because the two network
+legs fail independently:
+
+- **Bot leg** — `state` — the message's journey to the Medplum server and back,
```
mattlong: "Bot" and "server" are used interchangeably throughout a lot of the documentation.

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
```diff
@@ -0,0 +1,801 @@
...
+     (purged)
+```
+
+**Bot-leg states** (`state`), exhaustively:
+
+| State | Set when | Set by | Terminal? |
+|---|---|---|---|
+| `queued` | row inserted, before WS send | `DurableQueue.enqueue` | no |
+| `claimed` | worker claimed the row off the queue, but the `agent:transmit:request` is not yet on the wire (still buffered in the in-memory WS queue) | `DurableQueue.claimNext` | no |
+| `inflight` | the request was written to the socket (`sent_at` stamped); awaiting the server response | `DurableQueue.markSent` (from `App.sendToWebSocket`) | no |
```
mattlong: It seems like there must be a gap between `claimed` and `inflight`  where its unknown if the write to the socket succeeded.

packages/agent/src/app.ts
```diff
@@ -857,29 +1141,150 @@ export class App {
...
+  }
+
+  /**
+   * Invokes `fn` for every channel that currently has a durable-queue worker running.
+   * @param fn - Callback applied to each running {@link ChannelQueueWorker}.
+   */
+  private forEachChannelWorker(fn: (worker: ChannelQueueWorker) => void): void {
+    for (const channel of this.channels.values()) {
+      if (channel instanceof AgentHl7Channel && channel.worker) {
+        fn(channel.worker);
```
mattlong: What should happen if `fn` throws?

packages/agent/src/app.ts
```diff
@@ -544,10 +601,192 @@ export class App {
...
+    if (recovered.requeued > 0 || recovered.failed > 0) {
+      this.log.info(
+        `Acquired queue lease — recovered interrupted rows: ${recovered.requeued} requeued (unsent), ` +
+          `${recovered.failed} failed (ambiguous, in-flight).`
+      );
+    }
+    // Tell every HL7 channel to start its worker now that we're leader.
+    for (const channel of this.channels.values()) {
+      if (channel instanceof AgentHl7Channel) {
+        channel.onBecameQueueLeader();
```
mattlong: What happens on errors?

packages/agent/src/queue/lease-manager.ts
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Begins the acquire-and-heartbeat loop. `onBecameLeader` fires every time
+   * we transition from follower to leader (typically once, but can repeat if
+   * we lose the lease mid-run and reclaim it). `onLostLeadership` fires when a
+   * heartbeat discovers a peer stole the lease, so the caller can drain workers
+   * before the peer starts acting on the shared data.
+   * @param onBecameLeader - Callback invoked when this process takes the lease.
+   * @param onLostLeadership - Callback invoked when a heartbeat finds the lease lost.
+   */
+  start(onBecameLeader: () => void, onLostLeadership?: () => void): void {
```
mattlong: What should happen if `start()` is called when already the leader? It seems like everything would be fine, might be worth making that explicit in the docs

packages/agent/src/queue/retention.ts
```diff
@@ -0,0 +1,206 @@
...
+      const db = this.queue.getDb();
+      const cutoffProcessed = now - this.retentionMs;
+      const cutoffErrored = now - this.erroredRetentionMs;
+
+      // Phase 1: time-based purge of fully-done processed rows. A processed row
+      // whose ACK never reached the source (ack_outcome = 'undelivered') is NOT
+      // fully done — it must survive long enough for the source to retransmit
+      // (which replays the stored ACK) and is an operator signal — so it's
+      // excluded here and protected by the errored floor in phase 3.
+      const phase1 = db.prepare(RETENTION_PHASE1_DELETE).run(cutoffProcessed);
```
mattlong: It does seem like its worth cleaning up these repeated prepares knowing how hostile agent environments can be and how rarely some of them are managed.

packages/agent/src/app.ts
```diff
@@ -544,10 +601,192 @@ export class App {
...
+  }
+
+  /**
+   * Called by the {@link QueueLeaseManager} when a heartbeat discovers a peer
+   * stole the lease. Stops every channel worker so the demoted process stops
+   * claiming and dispatching rows from the now peer-owned queue. A later
+   * re-acquisition fires {@link onBecameQueueLeader} again, which restarts the
+   * workers.
+   */
+  private onLostQueueLeadership(): void {
```
mattlong: Should retentionSweeper be stopped here?

packages/agent/src/app.ts
```diff
@@ -544,10 +601,192 @@ export class App {
...
+      }
+    }
+
+    // Flush the WAL on every heartbeat tick (same idiom as ChannelStatsTracker /
+    // Hl7ClientPool GC). SQLite only attempts checkpoints piggybacked on commits,
+    // so once traffic stops nothing else would drain the WAL until the hourly
+    // sweep or close(). checkpointWalIfDirty() is a no-op on an idle queue.
+    if (!this.queueCheckpointListener) {
+      const queue = this.durableQueue;
+      this.queueCheckpointListener = () => queue.checkpointWalIfDirty();
```
mattlong: This lets _this_ agent invoke `queue.checkpointWalIfDirty()` even if it doesn't own the lease.

This is representative of other similar questions of durableQueue ownership and intent. There's an implicit dance and careful coordination needed between `this.leaseManager` and `this.durableQueue` where some operations are gated off being the leader but others are allowed to continue.

It seems like the model would be easier to reason about and more explicit if if `App` didn't access `durableQueue` directly, but requests it through the lease manager: `this.leaseManager.getDurableQueue()` which could return `undefined` or maybe throw an error if its not the current leader. If there's a need to overrule the leadership bit and get the durable queue regardless, so be it, just make it explicit in the method/getter name or with an optional parameter.

mattlong (commented):
I haven't looked at everything yet, but wanted to leave these comments so far for you to consider.

packages/agent/src/app.ts
ThatOneBro: Yeah this is related to the ideas you pointed out with `checkpointWalIfDirty`. Neither of those actually are unsafe to do while not the leader as they are idempotent and re-entrant, so in both cases it's actually okay to do them when not the leader.


I found the better logical split between what the `QueueLeaseManager` actually gates vs not to be "dispatch" and "everything else". Basically only pulling messages to be dispatched / dispatching messages need to be gated by the lease manager. Everything is fair game for concurrent stuff. As such I just renamed it to `DispatchLeaseManager`

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: Yes I think it's okay though, mainly we just care if we do know it succeeded to go to the network buffer or not, before that we can treat any exception the same as `claimed` and handle error cases similarly

packages/agent/src/queue/types.ts
```diff
@@ -136,6 +136,34 @@ export class QueueError extends Error {
...
+  readonly localHolder: string | undefined;
+  /** The holder that actually owns the lease now (the peer that took over). */
+  readonly currentHolder: string | undefined;
+
+  constructor(localHolder: string | undefined, currentHolder: string | undefined) {
+    super(
+      `Queue lease held by a peer; this process is no longer leader ` +
+        `(local holder=${localHolder ?? 'unset'}, current holder=${currentHolder ?? 'none'})`
+    );
+    this.name = 'QueueLeaseError';
```
mattlong, line 161: Curious why setting name is necessary. Does super() set it to something else? Like `Error`?

packages/agent/src/hl7.ts
```diff
@@ -268,13 +731,60 @@ export class AgentHl7ChannelConnection {
...
+  private sendCommitAck(message: Hl7Message): void {
+    const enhancedMode = this.channel.getEnhancedMode();
+    if (!enhancedMode) {
+      return;
+    }
+    const ackCode: AckCode = enhancedMode === 'standard' ? 'CA' : 'AA';
+    const response = message.buildAck({ ackCode });
+    this.hl7Connection.send(response);
+    // Reuse the existing 'enhancedAckSent' listener for logging (handleEnhancedAckSent).
+    this.hl7Connection.dispatchEvent(new Hl7EnhancedAckSentEvent(this.hl7Connection, response));
```
mattlong, line 755: Is there a reason `this.hl7Connection` doesn't dispatch its own events within `send()`? This feels fragile

packages/agent/src/queue/retention.ts
```diff
@@ -0,0 +1,215 @@
...
+        deletedProcessed,
+        deletedErrored,
+        dbSizeBytesAfter: this.queue.getDbSizeBytes(),
+      };
+      this.lastResult = result;
+      this.lastSweepAt = now;
+
+      if (deletedProcessed > 0 || deletedErrored > 0) {
+        this.log.info(
+          `Retention sweep: deleted ${deletedProcessed} processed, ${deletedErrored} errored. DB ${result.dbSizeBytesAfter} bytes.`
```
mattlong: Worth logging the the duration of the sweep? Or perhaps in a different `this.log.warn` if the duration is too long

packages/agent/src/queue/types.ts
ThatOneBro: Yeah generally if you want the Error to have a custom name you have to set it in the constructor

packages/agent/src/hl7.ts
ThatOneBro: Generally I don't know if we want any commit-level ACK sent over the wire by the user to trigger this event being emitted, since historically this event was just to correlate logs of a system-generated "enhanced mode ACK" being sent. 

The `send` method is very simple and does nothing but encode the HL7 message into a buffer and push it onto the socket, and it would be a divergence from that simplicity of the method to inspect messages and emit events based on the contents of the message.

Before the durable queue was added, the machinery for sending this commit ACK and emitting the event was handled by the connection when configured in `enhancedAck` mode, but now that we require the agent itself to drive the sending of the enhanced mode ACK after actually ensuring it has been committed to the database, we can no longer let the `Hl7Connection` control this. That's why we pushed this machinery into these methods for the HL7 channel.

Maybe we could push this into `Hl7Connection` with methods like `Hl7Connection.sendCommitAck` ?

packages/agent/src/app.ts
ThatOneBro: We talked about this offline; I've made the split more clear about what is allowed to be done without the lock and what needs the lock (namely taking things out of the queue for processing), and we've refactored the `DurableQueue` class to reflect that with an `assertNotDemoted` check before each action that requires the lease, similar to how we handle things in `Repository` in `@medplum/server`

packages/agent/src/queue/retention.ts
ThatOneBro: 👍 Done!

packages/agent/src/app.ts
ThatOneBro: Ended up removing `onLostQueueLeadership` since it wasn't strictly necessary and muddied the waters around what required the queue lease and what didn't 👍

packages/agent/src/queue/retention.ts
ThatOneBro: Good call, added

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: I standardized on `Bot/server`, hopefully that's not confusing

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: Actually you're right about this, I wrapped the `markSent` and `ws.send` in a transaction so that if it does get set to `inflight` atomically and actually gets set before sending now but only committed after send.

packages/agent/src/queue/lease-manager.ts
ThatOneBro: This has been moved to `DurableQueue#startDispatchLease`, but I did update the TSDoc to note that it is idempotent

packages/agent/src/app.ts
ThatOneBro: Ultimately errors thrown from here will come up in the `DurableQueue#tryAcquireDispatchLease` and get caught:
https://github.com/medplum/medplum/blob/derrick-durable-queue-take-2/packages/agent/src/queue/durable-queue.ts#L958

But we can still do better and try to make sure as many channels get this method called on them properly and rethrow all the errors at the end as an "aggregate error" (without using Node's nasty `AggregateError` class 😬 )

ThatOneBro: Updated 🙏

packages/agent/src/app.ts
ThatOneBro: Now we will catch and rethrow at the end,

This has two paths in practice, the one in `this.webSocket.on('close')` and the one in handling `agent:connect:response`. 

They both have error handling, so I think we should be good now. I did refactor the WebSocket error handler to actually include the error now since previously it assumed that the only error that could be thrown was one from the WebSocket connection itself, but in Node that is not the case like it is in the browser.

ThatOneBro:
@claude Can you give a final review of this PR in it's current state?

mattlong (approved):
:shipit:

PR #9349, opened 2026-05-29
SMART Health Cards and Links

codyebberson:
Adds initial Medplum server support for SMART Health Cards and SMART Health Links as FHIR operations.

https://build.fhir.org/ig/HL7/smart-health-cards-and-links/

New operations:

- `POST /fhir/R4/Patient/:id/$generate-smart-health-card`
- `POST /fhir/R4/$verify-smart-health-card`
- `POST /fhir/R4/Patient/:id/$generate-smart-health-link`
- `POST /fhir/R4/$resolve-smart-health-link`

Also adds a public generated-SHL manifest endpoint:

- `POST /fhir/R4/.well-known/smart-health-links/:id/manifest.json`

# Changes

- Added SMART Health Card generation from a patient compartment bundle.
- Added SHC output artifacts:
  - compact JWS credential
  - `shc:/` numeric URI
  - `.smart-health-card` JSON file content
  - optional QR code data URL
- Added SHC verification for:
  - compact JWS credentials
  - `shc:/` numeric values
  - `.smart-health-card` file JSON
- Reused existing OAuth signing key infrastructure for SHC signing.
  - `getSigningKey(alg?)` now optionally returns the first active key matching an algorithm.
  - SHC generation uses the active `ES256` key from existing `JsonWebKey` resources.
  - This removes the original prototype's ephemeral in-memory signing key.
- Added external SHC verification support.
  - Local Medplum JWKS is checked first.
  - External issuer JWKS is fetched from `https://issuer/.well-known/jwks.json`.
  - Verification response distinguishes:
    - `signatureValid`
    - `issuerTrusted`
    - `verified`
- Added initial SMART Health Link generation and resolution.
  - Generates `shlink:/` payloads.
  - Encrypts bundled FHIR JSON as JWE `dir` + `A256GCM`.
  - Supports optional passcode protection with `bcrypt`.
  - Serves manifests for Medplum-generated links.
  - Resolves Medplum-generated links and returns decrypted FHIR resources.
- Added focused tests for happy paths, failure paths, external SHC verification, key selection, and passcode behavior.

# Notes / Limitations

This is still an initial server-side prototype, not a complete production rollout.

- SHL persistence is currently in-memory.
- SHL encrypted files are embedded in the manifest rather than stored durably in Binary/storage.
- SHL external resolution is not implemented yet.
- SHC external issuer trust is intentionally conservative:
  - External signatures can validate cryptographically.
  - External issuers are not marked trusted.
  - `verified` is only true for Medplum-trusted issuer/key combinations.
- External JWKS fetching has first-pass safety controls:
  - HTTPS only
  - no redirects
  - timeout
  - basic localhost/private IP literal rejection
  - DNS private-range resolution checks should be added before treating this as hardened remote fetch infrastructure.

### DIFF
--- a/packages/server/src/fhir/operations/smarthealthcards.ts
+++ b/packages/server/src/fhir/operations/smarthealthcards.ts
@@ -150,6 +150,7 @@ export async function verifySmartHealthCardHandler(req: FhirRequest): Promise<Fh
   let issuerTrusted: boolean | undefined = undefined;
   let header: ProtectedHeaderParameters | undefined = undefined;
   let valid = true;
+  let error: string | undefined = undefined;
 
   try {
     ({ issuerTrusted } = await verifySmartHealthCardCredential(credential));
@@ -157,6 +158,7 @@ export async function verifySmartHealthCardHandler(req: FhirRequest): Promise<Fh
   } catch (err) {
     getLogger().warn('SMART Health Card validation failed', { err });
     valid = false;
+    error = normalizeErrorString(err);
   }
 
   return [
@@ -168,6 +170,7 @@ export async function verifySmartHealthCardHandler(req: FhirRequest): Promise<Fh
       issuer: unverifiedPayload.iss,
       keyId: header?.kid,
       fhirBundle: JSON.stringify(unverifiedPayload.vc.credentialSubject.fhirBundle),
+      error,
     }),
   ];
 }
--- a/packages/server/src/fhir/operations/smarthealthcards.ts
+++ b/packages/server/src/fhir/operations/smarthealthcards.ts
@@ -3,13 +3,14 @@
 import { allOk, badRequest, normalizeErrorString, OAuthSigningAlgorithm } from '@medplum/core';
 import type { FhirRequest, FhirResponse } from '@medplum/fhir-router';
 import type { Bundle, Patient, Resource } from '@medplum/fhirtypes';
-import type { JWK, KeyLike } from 'jose';
+import type { JWK, KeyLike, ProtectedHeaderParameters } from 'jose';
 import { CompactSign, compactVerify, decodeProtectedHeader, importJWK } from 'jose';
 import { isIP } from 'node:net';
 import { deflateRawSync, inflateRawSync } from 'node:zlib';
 import QRCode from 'qrcode';
 import { getConfig } from '../../config/loader';
 import { getAuthenticatedContext } from '../../context';
+import { getLogger } from '../../logger';
 import { getJwks, getSigningKey } from '../../oauth/keys';
 import { makeOperationDefinition, makeParameters } from './definitions';
 import { getPatientEverything } from './patienteverything';
@@ -80,7 +81,6 @@ export const verifySmartHealthCardOperation = makeOperationDefinition(
       { use: 'in', name: 'shcUri', type: 'string', min: 0, max: '1' },
       { use: 'in', name: 'file', type: 'string', min: 0, max: '1' },
       { use: 'out', name: 'valid', type: 'boolean', min: 1, max: '1' },
-      { use: 'out', name: 'signatureValid', type: 'boolean', min: 0, max: '1' },
       { use: 'out', name: 'issuerTrusted', type: 'boolean', min: 0, max: '1' },
       { use: 'out', name: 'verified', type: 'boolean', min: 0, max: '1' },
       { use: 'out', name: 'issuer', type: 'uri', min: 0, max: '1' },
@@ -91,6 +91,15 @@ export const verifySmartHealthCardOperation = makeOperationDefinition(
   }
 );
 
+/**
+ * Handles the Patient/$generate-smart-health-card operation.
+ *
+ * Builds a Patient/$everything bundle for the requested patient, signs it as a SMART Health Card credential, and
+ * returns the credential in JWS, SHC URI, and SMART Health Card file formats.
+ *
+ * @param req - FHIR request containing the patient ID and optional parameters for filtering, expiration, and QR code generation.
+ * @returns FHIR response containing the generated SMART Health Card credential in multiple formats and related metadata.
+ */
 export async function generateSmartHealthCardHandler(req: FhirRequest): Promise<FhirResponse> {
   try {
     const ctx = getAuthenticatedContext();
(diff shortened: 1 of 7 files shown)


codyebberson:
@claude please review

codyebberson:
@claude 

>  1. CodeQL: bcrypt rounds too low [smarthealthlinks.ts:18]

OWASP now recommends "10 or more".  See: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

> 2. IPv6 ULA fd prefix missing [smarthealthcards.ts:448]

Fixed

> 3. FHIR bundle returned from unverified decode on failure

This is by-design.  The vast majority of sample QR codes and sample SHC links, even from the spec itself, have invalid signatures, so this was a pragmatic choice to allow users to test sample data.

> 4. Signing key fetched twice in generateSmartHealthCardHandler [smarthealthcards.ts:129]

Fixed

> 5. recipient is accepted but silently ignored [smarthealthlinks.ts:29, 86–90]

This will be fixed in future PR.  There are a number of open future tasks with SHL.

> 6. No rate limiting on the manifest endpoint [routes.ts:177]

There is a global rate limiter in the middleware that protects this route.

> 7. affectsState: false on $generate-smart-health-card may need re-checking [smarthealthcards.ts:58]

Noted.  Accepted risk.

> 8. params.shlink as string cast without runtime guard [smarthealthlinks.ts:147]

Fixed

packages/server/src/fhir/operations/smarthealthcards.ts
```diff
@@ -0,0 +1,461 @@
...
+  return jwks.keys ?? [];
+}
+
+/**
+ * Checks whether a hostname should be rejected for outbound SMART Health Card JWKS lookup.
+ *
+ * @param hostname - Hostname from an external issuer URL.
+ * @returns True when the hostname is localhost, link-local, loopback, private, or otherwise unsafe.
+ */
+function isUnsafeHostname(hostname: string): boolean {
```
codyebberson, line 451: In future PR, I'd like to move this to `packages/server/src/util` and use it everywhere we call external URLs:
1. Rest-hook subscriptions
2. Auto downloader
3. Here

packages/server/src/fhir/operations/smarthealthcards.ts
```diff
@@ -0,0 +1,461 @@
...
+    });
+    const issuer = getConfig().issuer;
+    const { credential, keyId } = await createSmartHealthCardCredential(bundle, issuer, params.exp);
+    const shcUri = encodeSmartHealthCardUri(credential);
+    const file = JSON.stringify(writeSmartHealthCardFile(credential));
+    const qrCodeDataUrl = params.includeQrCode ? await QRCode.toDataURL(shcUri) : undefined;
+
+    return [
+      allOk,
+      makeParameters({
```
mattwiller: This should be able to use the existing `buildOutputParameters()` helper to ensure consistency with the above OperationDefinition

packages/server/src/fhir/operations/smarthealthcards.ts
```diff
@@ -0,0 +1,461 @@
...
+        shcUri,
+        file,
+        qrCodeDataUrl,
+        issuer,
+        keyId,
+      }),
+    ];
+  } catch (err) {
+    return [badRequest(normalizeErrorString(err))];
+  }
```
mattwiller: Is this necessary?  Why would all errors get wrapped/normalized to 400 in this operation?

packages/server/src/fhir/operations/smarthealthcards.ts
```diff
@@ -0,0 +1,461 @@
...
+ * @param req - FHIR request containing the SMART Health Card credential in one of the supported input formats.
+ * @returns FHIR response containing the verification results, including validity, issuer trust, and decoded payload details.
+ */
+export async function verifySmartHealthCardHandler(req: FhirRequest): Promise<FhirResponse> {
+  const params = parseInputParameters<VerifySmartHealthCardParams>(verifySmartHealthCardOperation, req);
+  const credential = readSmartHealthCardCredential(params);
+  const unverifiedPayload = decodeSmartHealthCardPayload(credential);
+  let issuerTrusted: boolean | undefined = undefined;
+  let header: ProtectedHeaderParameters | undefined = undefined;
+  let valid = true;
```
mattwiller: I have a minor aesthetic preference for starting with `valid = false` and changing it at the end of the `try` block below

packages/server/src/fhir/operations/smarthealthcards.ts
```diff
@@ -0,0 +1,461 @@
...
+    exp,
+    vc: {
+      type: SHC_VC_TYPE,
+      credentialSubject: {
+        fhirVersion: FHIR_VERSION,
+        fhirBundle: bundle,
+      },
+    },
+  };
+  const compressedPayload = deflateRawSync(Buffer.from(JSON.stringify(payload)));
```
mattwiller: I suspect it's fine because we're dealing with relatively small payloads here, but is there any DoS concern around the sync versions of these functions?

packages/server/src/fhir/operations/smarthealthcards.ts
```diff
@@ -0,0 +1,461 @@
...
+      : normalizedHostname;
+  const ipVersion = isIP(ipHostname);
+  if (ipVersion === 0) {
+    return ['localhost', 'localhost.localdomain'].includes(normalizedHostname);
+  }
+  if (ipVersion === 4) {
+    return /^(10\.|127\.|169\.254\.|172\.(1[6-9]|2\d|3[0-1])\.|192\.168\.)/.test(ipHostname);
+  }
+  return (
+    ipHostname === '::1' || ipHostname.startsWith('fe80:') || ipHostname.startsWith('fc') || ipHostname.startsWith('fd')
```
mattwiller, line 465: Where are these cases sourced from?  Are we sure they're relatively exhaustive?

packages/server/src/fhir/operations/smarthealthlinks.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+  return undefined;
+}
+
+function encodeSmartHealthLink(payload: SmartHealthLinkPayload): string {
+  return `shlink:/${base64url.encode(Buffer.from(JSON.stringify(payload)))}`;
+}
+
+function parseSmartHealthLink(input: string): SmartHealthLinkPayload {
+  const raw = input.includes('#shlink:/') ? input.substring(input.indexOf('#shlink:/') + 1) : input;
```
mattwiller: You can avoid scanning the `input` string twice by first calling `input.indexOf('#shlink:/')` and checking the return value against `-1`

packages/server/src/fhir/operations/smarthealthcards.ts
codyebberson: These were hand-rolled rather than sourced from a single canonical list, and you're right that they're not exhaustive — string matching on IP literals misses cases like IPv4-mapped IPv6 (::ffff:127.0.0.1), expanded ::1, 0.0.0.0/8, and CGNAT, and it only triggers on raw IP literals at all (a hostname resolving to a private IP isn't caught).

The HTTPS requirement mitigates the most obvious cases for now. I'd like to address SSRF hardening properly in a fast-follow PR — robust IP classification (e.g. ipaddr.js, already in our tree) plus DNS resolution and IP pinning — rather than partially here.

packages/server/src/fhir/operations/smarthealthcards.ts
codyebberson: Fixed

packages/server/src/fhir/operations/smarthealthcards.ts
codyebberson: I removed it.  The main reason was that some code paths can result in "Not found" outcomes, which would result in a generic HTTP 404 "Not found".  It looks like we allow this in a bunch of our other FHIR operations, so I'll do the simple thing for now.

In the future, I feel like we could probably return more helpful error messages in those cases.  "HTTP 400, requested client ID not found" seems better than "HTTP 404 Not Found"

packages/server/src/fhir/operations/smarthealthcards.ts
codyebberson: Fixed

packages/server/src/fhir/operations/smarthealthcards.ts
codyebberson: I switched this to the async zlib API via `promisify(deflateRaw)`, which should avoid blocking the event loop. For these QR-sized payloads the practical risk was already low, but this makes the implementation more robust without much downside.

packages/server/src/fhir/operations/smarthealthlinks.ts
codyebberson: Fixed

mattwiller (approved):

PR #9354, opened 2026-05-29
Track idle in transaction time

mattlong:
Adds an approximation of postgres idle in transaction time as measured from the application.

We may want to change what triggers metrics/log generation: currently it is the max idle in transaction to match the semantics of `idle_in_transaction_session_timeout`. Thinking about it more, maybe we want to add or change the trigger to total transaction duration? Or maybe even have multiple triggers...


mattlong:
@claude review. Pay close attention to the overriding of clienty.query

mattlong:
@claude review again

The-Alchemist (approved):

codyebberson (approved):

PR #9355, opened 2026-05-29
feat(react): realtime voice - keep socket active

techdavidy:
Refactor useWhisper to keep the WebSocket connection across recordings. 
There is noticeable latency while re-connecting on every start adds latency. 
Remove deprecated calls from useWhisper.


techdavidy:
@claude please review

codyebberson (approved):

PR #9356, opened 2026-05-30
Support `resource:number` in `convertToTransactionBundle`

codyebberson:
Adds support for SMART Health Card-style bundle-local references such as:

```json
{
  "fullUrl": "resource:0"
}
```

and

```json
{
  "patient": {
    "reference": "resource:0"
  }
}
```

when converting arbitrary FHIR bundles to transaction bundles using `convertToTransactionBundle()`.

### Background

FHIR allows references within a bundle to resolve against matching `Bundle.entry.fullUrl` values. While `urn:uuid:` is the most common representation, SMART Health Cards use compact `resource:N` URIs (for example `resource:0`, `resource:1`, etc.) to reduce QR code payload size.

Example:

```json
{
  "fullUrl": "resource:0",
  "resource": {
    "resourceType": "Patient"
  }
}
```

```json
{
  "patient": {
    "reference": "resource:0"
  }
}
```

These are valid bundle-local references and are commonly encountered when importing SMART Health Card payloads.

### Problem

`convertToTransactionBundle()` relies on reference analysis to:

* determine resource dependencies
* perform topological sorting
* rewrite references for transaction processing
* generate conditional create logic

Bundles containing `resource:N` references were not previously recognized as having valid intra-bundle relationships, resulting in incomplete dependency resolution.

### Solution

Extend bundle reference resolution logic to recognize `resource:N` URIs as bundle-local references whenever a matching `Bundle.entry.fullUrl` exists.

This allows SMART Health Card bundles and other bundles using compact `fullUrl` identifiers to participate in the same dependency analysis and transaction conversion workflow as bundles using `urn:uuid:` identifiers.

### Impact

* SMART Health Card bundles can now be passed directly to `convertToTransactionBundle()`
* Topological sorting correctly preserves resource dependencies
* Conditional create and reference rewriting continue to work as expected
* Improves compatibility with valid FHIR bundle-local reference patterns encountered in the wild

### DIFF
--- a/packages/core/src/__fixtures__/example-00-a-fhirBundle.json
+++ b/packages/core/src/__fixtures__/example-00-a-fhirBundle.json
@@ -1,0 +1,103 @@
+{
+  "resourceType": "Bundle",
+  "type": "collection",
+  "entry": [
+    {
+      "fullUrl": "resource:0",
+      "resource": {
+        "resourceType": "Patient",
+        "name": [
+          {
+            "family": "Anyperson",
+            "given": [
+              "John",
+              "B."
+            ]
+          }
+        ],
+        "birthDate": "1951-01-20"
+      }
+    },
+    {
+      "fullUrl": "resource:1",
+      "resource": {
+        "resourceType": "Immunization",
+        "status": "completed",
+        "vaccineCode": {
+          "coding": [
+            {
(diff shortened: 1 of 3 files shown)


ThatOneBro (approved):
Interesting 👀 

LGTM

PR #9366, opened 2026-06-01
fix(provider): Create/update claim during export or submit

techdavidy:
Currently, In provider while updating an encounter. It would by try to update/create a claim with the same updates.
(charge items, conditions, practitioner, etc). This would create a lot of extra network calls and code to support it,

Simplify to create/update claim while exporting or submitting.


techdavidy:
@claude please review

mattlong (approved):

PR #9378, opened 2026-06-02
feat(data warehouse): adding config to set which tables to sync, and minor changes

The-Alchemist:
## Requirement

Current data warehouse parameters allow one to configure:
*  `startDate`: earliest sync date
* `enabled`: `true` or `false`

After discussion with @codyebberson , we have realized a need to specify which tables--or more accurately, resource types--to sync, as a stepping stone to enabling all tables.


packages/server/src/data-warehouse/config.ts
```diff
@@ -83,15 +83,31 @@ export interface WarehouseSourceTable {
...
  * matching migrations (`resourceType + '_History'`).
  * Used by the scheduled data warehouse sync worker.
  *
+ * @param resourceTypes - Optional FHIR resource types to include. When omitted or empty, all types are included.
  * @returns The list of Postgres table names.
  */
-export function getWarehouseSyncPostgresTableNames(): string[] {
-  return getResourceTypes().map((resourceType) => `${resourceType}_History`);
+export function getWarehouseSyncPostgresTableNames(resourceTypes?: string[]): string[] {
+  const allTableNames = getResourceTypes().map(toHistoryPostgresTableName);
```
mattlong: Only suggesting an optmization since I believe this gets executed on every worker run.
```suggestion
  if (!resourceTypes?.length) {
    return getResourceTypes().map(toHistoryPostgresTableName);
  }
  const results = [];
  const selectedTypes = new Set(resourceTypes);
  for (const resourceType of getResourceTypes()) {
    if (selectedTypes.has(resourceType)) {
      results.push(toHistoryPostgresTableName(resourceType));
    }
  }
  return result;
```

packages/server/src/config/types.ts
```diff
@@ -348,6 +348,11 @@ export interface MedplumDataWarehouseConfig {
    * History rows with `lastUpdated` before this value are excluded.
    */
   startDate?: string;
+  /**
+   * FHIR resource types to sync (e.g. `Patient`, `Observation`).
+   * When omitted, all indexed resource history tables are synced.
+   */
+  resourceTypes?: string[];
```
mattlong, line 355: What happens if someone were to have a set of resource types listed here and then months down the road decides they need to add more resources? Does the backfill happen naturally as the include list expands?

The-Alchemist: great point.   the `startDate` parameter will dictate what data will be synced, making care of "small" backfills.

a "large" backfill, such as for all of 2025,  however, will need to be handled out of band.  that's the current plan, at least

packages/server/src/data-warehouse/config.ts
The-Alchemist: done

codyebberson:
@claude please review

codyebberson (approved):

PR #9386, opened 2026-06-02
feat: log AuditEvents for Bot channel subscriptions

### LINKED ISSUE #9385: Bot subscriptions do not generate subscription delivery AuditEvents
## Bug

When a `Subscription` uses a Bot as its channel endpoint (`channel.endpoint = "Bot/<id>"`), the `subscription-audit-event-destination` extension is silently ignored and no subscription delivery `AuditEvent` is ever created — regardless of whether the extension specifies `resource`, `log`, or both.

## Root Cause

In `packages/server/src/workers/subscription.ts`, the `execBot` function calls `executeBot(...)` but never calls `createSubscriptionAuditEvent()`:

```ts
async function execBot(...): Promise<void> {
  // ...
  await executeBot({ subscription, bot, runAs, ... });
  // ← no createSubscriptionAuditEvent() call here
}
```

By contrast, the HTTP rest-hook path calls `createSubscriptionAuditEvent()` in two places — after a successful/failed fetch and in the exception handler — respecting the `subscription-audit-event-destination` extension.

## Steps to Reproduce

1. Create a `Subscription` with `channel.endpoint = "Bot/<id>"` and the extension:
   ```json
   {
     "url": "https://medplum.com/fhir/StructureDefinition/subscription-audit-event-destination",
     "valueCode": "log"
   }
   ```
2. Trigger the subscription by creating/updating a matching resource.
3. Observe: no `AuditEvent` with `type.code = "transmit"` appears in logs or DB.

## Expected Behavior

Bot subscriptions should generate subscription delivery `AuditEvent` resources (and/or log them) in the same way HTTP rest-hook subscriptions do, honoring the `subscription-audit-event-destination` extension.

## Suggested Fix

Wrap the `executeBot` call in `execBot` with try/catch and call `createSubscriptionAuditEvent()` on both success and failure, mirroring the HTTP rest-hook path. The import for `createSubscriptionAuditEvent` and `AuditEventOutcome` already exists at the top of the file.

ianplunkett:
## Problem

Bot-channel subscriptions never called `createSubscriptionAuditEvent()`, so subscription delivery was invisible in the audit trail. The existing `createBotAuditEvent()` (called inside `executeBot`) creates an `execute`-type AuditEvent that includes the subscription as an entity, but it only wrote to the database — it never emitted to logs.

Fixes #9385

## Solution

Rather than adding a second `transmit`-type AuditEvent per bot execution (which would double logging volume), this PR reconciles the existing bot execution AuditEvent to be the single source of truth:

1. **Always emit to logs** — `createBotAuditEvent` now always calls `logAuditEvent()` regardless of `bot.auditEventDestination`. Log emission can no longer be disabled.
2. **DB write remains optional** — controlled by `bot.auditEventDestination` (defaults to `resource` when unset, same as before).
3. **No separate `transmit` event for bots** — the `execute` AuditEvent already captures the subscription as an entity, so a second event would be redundant.
4. **`execBot` simplified** — the try/catch wrapper and duplicate `createSubscriptionAuditEvent` calls are removed. Rest-hook subscriptions are unaffected.

The `requestTime`, `traceId`, and `headers` fields are now passed into `executeBot` from the subscription job, which ensures the bot's own `execute` AuditEvent has an accurate period start time and trace context.

## Test plan

- `Bot subscription execution AuditEvent is always emitted to logs` — verifies the `execute` AuditEvent appears via `globalLogger.write` with `logAuditEvents: true`, and no `transmit` AuditEvent is written to the DB.
- Existing bot subscription tests verify the DB AuditEvent count is unchanged (1 `execute` event per delivery).
- Existing `Subscription AuditEvent destination with logging` tests verify rest-hook behavior is unaffected.


techdavidy:
@claude please review

codyebberson (changes requested):
I'd prefer to avoid 2 `AuditEvent` resources per subscription, because we this is a significant amount of logging data. 

Scattered thoughts:

1. At present, the "log destination" extension allows you to independently toggle the log `AuditEvent` and the database `AuditEvent`
2. I think the consensus is that we always want to log `AuditEvent`, and it should not be allowed to disable that, because it is effectively part of the official audit trail
3. The "log destination" extension should be respected for both webhooks and bot execution
4. Then there's still the question of whether there should be one or two `AuditEvent` resources
5. In conversation, the soft understanding was that they're effectively the same shape, except one is `execute` and one is `transmit`
6. TODO: Confirm that the `AuditEvent` resources are otherwise identical
7. TODO: Make a call on whether that justifies a copy

mattwiller:
@claude Review the updated PR, which is significantly reduced in scope

ThatOneBro (approved):

packages/server/src/util/auditevent.ts
```diff
@@ -353,23 +353,18 @@ export async function createBotAuditEvent(
...
-      }
-      case 'log':
-        logAuditEvent({
-          ...auditEvent,
-          outcomeDesc: tail(outcomeDesc, config.maxBotLogLengthForLogs ?? defaultBotOutputLength),
-        });
-        break;
-    }
+  // Always emit to logs
+  logAuditEvent({
```
codyebberson, line 357: Nice

codyebberson (approved):
ship it 🚀

PR #9388, opened 2026-06-02
Constrain scheduling operations with `Schedule.planningHorizon`

### LINKED ISSUE #8432: $find operation should respect Schedule.planningHorizon boundaries
The $find operation does not enforce Schedule.planningHorizon.start or Schedule.planningHorizon.end when computing available slots. It only uses the caller-provided start/end parameters and the SchedulingParameters.availability extension.

https://discord.com/channels/905144809105260605/1471895487224353012

noahsilas:
- In `$find` we can use this to reduce our search space for available slots early on, potentially saving some effort. We avoid returning appointment results that are outside of the planning horizon for any requested calendar.
- In `$book` and `$hold`, we validate that the proposed appointment is not outside the planning horizon.
- We add some minor touchups to Provider UI to handle errors and empty results in $find.
- If the schedule doesn't define `planningHorizon.start` or `planningHorizon.end`, the relevant checks are skipped. (As though the planning horizon extends in that direction indefinitely).

Resolves #8432

### DIFF
--- a/examples/medplum-provider/src/pages/schedule/FindPane.tsx
+++ b/examples/medplum-provider/src/pages/schedule/FindPane.tsx
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Button, Group, Stack, Title } from '@mantine/core';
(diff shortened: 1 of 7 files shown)


noahsilas:
@claude Please review

codyebberson (approved):

PR #9389, opened 2026-06-02
Scheduling: default availability to "always"

noahsilas:
In the original spec and our documentation, we listed the "availability" sub-extension as an optional constraint. When not present, we should treat it as though there is no additional availability constraint (aka 24/7 availability).

The first implementation treated this sub-extension as required. In this PR we loosen that constraint to match the spec and docs.


codyebberson:
@claude please review

codyebberson (approved):

PR #9392, opened 2026-06-03
docs(agent): add docs page for `Agent/$upgrade`

ThatOneBro:
Adds docs for the `Agent/$upgrade` operation. Apparently we skipped this operation but had docs for using it via CLI

### DIFF
--- a/packages/docs/docs/agent/upgrade.md
+++ b/packages/docs/docs/agent/upgrade.md
@@ -38,11 +38,15 @@ For example:
 medplum get 'Agent/$upgrade?status=active'
 ```
 
+Here, `status=active` is an `Agent` search parameter used to select which agents to upgrade. See [Using Agent search parameters in bulk operations](./using-search-parameters.md) for more ways to select operation targets.
+
 ## Parameters
 
 - `version` (optional; default: latest published version): The version to upgrade the agent(s) to. When omitted, the agent upgrades to the latest available version.
 - `timeout` (optional; default: `45000`, max: `56000`): How long, in milliseconds, the server waits for the agent to acknowledge the upgrade request before returning an error. Values above the maximum are clamped to `56000`.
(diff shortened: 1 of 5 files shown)


packages/docs/docs/agent/upgrade.md
```diff
@@ -0,0 +1,276 @@
...
+
+```bash
+medplum get 'Agent/$upgrade?status=active'
+```
+
+## Parameters
+
+- `version` (optional; default: latest published version): The version to upgrade the agent(s) to. When omitted, the agent upgrades to the latest available version.
+- `timeout` (optional; default: `45000`, max: `56000`): How long, in milliseconds, the server waits for the agent to acknowledge the upgrade request before returning an error. Values above the maximum are clamped to `56000`.
+- `force` (optional; default: `false`): When `true`, instructs the agent to upgrade even if it is already running the requested version.
```
noahsilas, line 47: It would be nice to include something here about the usage of Agent search parameters. On first inspection, the example above with `?status=active` doesn't match any of these listed parameters and so could be confusing, although hopefully it becomes clear when the reader gets to the multi-agent section below.

packages/docs/docs/agent/upgrade.md
```diff
@@ -0,0 +1,276 @@
...
+
+Upgrade a specific agent by identifier:
+
+```bash
+medplum get 'Agent/$upgrade?identifier=agent-007'
+```
+
+### Multiple Agents by Name
+
+Upgrade all agents with a specific name prefix:
```
noahsilas: It seems like this endpoint does have some pagination considerations, where this will request will upgrade up to 10 agents. I think it would be a good idea to be explicit about this, so that someone with a larger agent fleet doesn't issue this single command and get surprised that only a subset upgrade.

noahsilas (approved):

packages/docs/docs/agent/upgrade.md
ThatOneBro: Fair, worth calling out

Issue #9237, opened 2026-05-18
Project Rename Operation


finnbergquist:
### Issue
When you rename a project, ProjectMembership.project.display text fields do not get updates. Those stale text fields are what show up in the login project selector, causing confusion.

### Proposed Solution
**Update** To avoid the potential large fanout of udpating ProjectMembership resources, proceed with the targeted fix of reading `Project` resources for the project selector.

Create a new operation that can rename a project and propagate change to ProjectMembership.project.display fields for users in that project.

finnbergquist:
Resolved in #9402

PR #9402, opened 2026-06-03
fix(server-app): Read project resource to display name

techdavidy:
Before 
=========

Screenshot 2026-06-03 at 11 43 34 AM
Screenshot 2026-06-03 at 11 46 06 AM

After
========
Screenshot 2026-06-03 at 11 43 55 AM
Screenshot 2026-06-03 at 1 00 11 PM

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -3139,6 +3139,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
           this.profilePromise = undefined;
           const profileChanged = this.sessionDetails?.profile?.id !== result.profile.id;
           this.sessionDetails = result;
+          this.syncStoredLoginProject();
           if (profileChanged) {
             this.dispatchEvent({ type: 'change' });
           }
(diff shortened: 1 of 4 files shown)


packages/core/src/client.ts
```diff
@@ -3152,6 +3153,33 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
+      return;
+    }
+    const updated: LoginState = {
+      ...activeLogin,
+      project: { ...activeLogin.project, display: projectName },
+    };
+    this.storage.setObject('activeLogin', updated);
+    this.storage.setObject(
+      'logins',
+      this.getLogins().map((login) => (login.profile?.reference === updated.profile?.reference ? updated : login))
```
mattlong: Is it safe to drop `login` which is the updated `activeLogin` into the `logins` object? Or should the map function similarly update the matching `login`?

mattlong (approved):

techdavidy:
@claude review

packages/core/src/client.ts
ThatOneBro: I think we could use `this.addLogin(updated);` here, if I'm reading this correctly

mattlong:
Fixes #9237

ThatOneBro (approved):
Nice LGTM

PR #9405, opened 2026-06-03
feat(data warehouse): include and exclude filters for tables to sync

The-Alchemist:
## Requirement

We want to be able to include/exclude tables  from the data warehouse sync process.

This change adds two new config values, `included` and `excluded`.

## Configuration

Along the way, added support for env var config for arrays, so one can set:

`MEDPLUM_DATA_WAREHOUSE_RESOURCE_TYPES_INCLUDED=Patient,Observation`

... instead of having to embed JSON, like:

`MEDPLUM_DATA_WAREHOUSE_RESOURCE_TYPES_INCLUDED='["Patient","Observation"]'`

This aligns with the long-term goal of transitioning to env vars for config as well.

### DIFF
--- a/packages/server/src/config/types.ts
+++ b/packages/server/src/config/types.ts
@@ -326,13 +326,6 @@ export interface MedplumWorkersConfig {
 
 export type MedplumDataWarehouseDestinationType = 's3tables' | 'local';
 
-export interface MedplumDataWarehouseResourceTypesConfig {
-  /** FHIR resource types to include (e.g. `Patient`, `Observation`). When omitted, all types are candidates. */
-  included?: string[];
-  /** FHIR resource types to exclude from sync. Applied after `included`, if set. */
-  excluded?: string[];
-}
-
 export interface MedplumDataWarehouseConfig {
   /**
    * Enables/disables the scheduled sync worker. Defaults to false.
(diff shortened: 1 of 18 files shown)


codyebberson:
@claude please review

packages/server/src/config/types.ts
```diff
@@ -349,10 +356,10 @@ export interface MedplumDataWarehouseConfig {
    */
   startDate?: string;
   /**
-   * FHIR resource types to sync (e.g. `Patient`, `Observation`).
-   * When omitted, all indexed resource history tables are synced.
+   * FHIR resource types to include and/or exclude from sync.
+   * When omitted, all resource history tables are synced.
    */
-  resourceTypes?: string[];
+  resourceTypes?: MedplumDataWarehouseResourceTypesConfig;
```
codyebberson: Strong preference for keeping these config types flat.

```ts
includeResourceTypes?: string[];
excludeResourceTypes?: string[];
```

Self-hosting customers very frequently get tripped up on server config settings, so simple = better.

The-Alchemist: done

codyebberson:
@claude - please re-review.  are all previous items resolved?

codyebberson (approved):

PR #9410, opened 2026-06-03
feat(server): add measure-score search parameter for MeasureReport

### LINKED ISSUE #9401: Add `measure-score` search parameter for `MeasureReport`
## Summary

Add a Medplum custom search parameter that indexes `MeasureReport.group.measureScore` as a `quantity`-typed parameter (`code: measure-score`), so `MeasureReport` resources can be filtered and sorted by their computed score with the standard numeric comparators (`eq`/`gt`/`lt`/`ge`/`le`).

## Motivation

`MeasureReport` is the semantically correct FHIR resource for a derived, per-subject metric or score (`group.measureScore`). However, FHIR R4 does **not** define a search parameter for `measureScore` — the core set is limited to `measure`, `patient`/`subject`, `period`, `status`, `reporter`, `evaluated-resource`, etc. As a result, there's no way to query by the value of the score itself.

This forces a common, generically-useful pattern — "find/list/count subjects whose computed score crosses a threshold" — onto less appropriate resources. Today, to get a natively-queryable numeric value you have to either:

- materialize the value onto an `Observation` (`value-quantity` is indexed), which pollutes the clinical Observation record with operational/analytics data and is semantically questionable; or
- add a custom search parameter + database column via a self-hosted migration, which isn't available on hosted Medplum.

Indexing `measureScore` lets `MeasureReport` be used for exactly what it's designed for, and benefits any workflow involving quality measures, risk scores, or other per-patient computed metrics.

## Concrete use cases

- List all subjects whose measure report scored ≥ a threshold: `GET /MeasureReport?measure=<measure>&measure-score=ge0.8`
- Count subjects above/below a threshold via `_summary=count`.
- Sort cohorts by score for review/triage.

## Proposed change

Add a single entry to the Medplum custom search-parameter bundle (`packages/definitions/src/fhir/r4/search-parameters-medplum.json`):

```json
{
  "resourceType": "SearchParameter",
  "code": "measure-score",
  "base": ["MeasureReport"],
  "type": "quantity",
  "expression": "MeasureReport.group.measureScore"
}
```

…then regenerate the schema migration via the standard generator.

## Why this should be low-risk / low-complexity

- **Strong precedent for this category of addition.** Medplum already defines custom search parameters on standard FHIR resources, including numeric ones — e.g. `Task.priority-order` (`type: number`) and `Task.due-date` (`type: date`). `measure-score` is the same pattern.
- **No new indexing code.** A `quantity` parameter already routes through the existing `range-column` strategy, and because `MeasureReport.group` is repeating, the existing **NUMMULTIRANGE** multi-value path handles multiple group scores. The schema migration (range column + sort column + GIST index) and the post-deploy reindex are produced by the existing migration generator.

## Semantics note (for discussion)

A `MeasureReport` may contain multiple `group` entries, so `MeasureReport.group.measureScore` can yield multiple values. With multi-range indexing, `measure-score=ge0.8` would match a report if **any** group's score qualifies. This seems like the desired behavior, but flagging it explicitly in case maintainers prefer to scope the expression (e.g. to a single primary group) or document the multi-group matching semantics.

## Willing to contribute

Happy to open the PR (bundle entry + generated migration + test coverage for comparator queries) if maintainers are supportive of the approach. Wanted to confirm the direction — particularly the multi-group semantics above — before doing so.

WonderPanda:
## Summary

FHIR R4 defines no search parameter for `MeasureReport.group.measureScore`, so there is currently no way to query measure reports by their computed score. This adds a Medplum custom `measure-score` quantity search parameter, enabling comparator queries and sorting on the score:

```
GET /MeasureReport?measure=<measure>&measure-score=ge0.8
GET /MeasureReport?measure=<measure>&measure-score=gt0.5&_summary=count
GET /MeasureReport?measure=<measure>&_sort=-measure-score
```

This makes `MeasureReport` usable for the common "find/list/count subjects whose computed score crosses a threshold" pattern, without forcing that data onto `Observation`.

Closes #9401.

## Changes

- Add the `measure-score` SearchParameter (type `quantity`, expression `MeasureReport.group.measureScore`) to the medplum definitions bundle.
- Generated schema migration `v109` (range columns `__measureScore` / `__measureScoreSort` + legacy column) and post-deploy data migration `v39` (gin + gist indexes), via `npm run migrate`.
- Add `search.test.ts` coverage for comparator filtering (`gt`/`lt`) and sorting.

## Semantics note

A `MeasureReport` may contain multiple `group` entries, so `MeasureReport.group.measureScore` yields multiple values. With multi-range indexing, `measure-score=ge0.8` matches a report if **any** group's score qualifies. Happy to scope the expression to a single primary group instead if maintainers prefer — flagging the multi-group behavior explicitly.

## Heads-up: a generator bug I hit

When generating the data migration, the migration generator emitted the range-gist index's sort column **unquoted**:

```sql
... USING gist ("__measureScore", __measureScoreSort)   -- folds to lowercase, "column does not exist"
```

Existing committed migrations (e.g. `data/v35.ts`) quote it (`"__measureScoreSort"`), which is what actually works. I used the quoted form in `v39`. This bug is latent — it only surfaces when adding the first brand-new range-column search parameter since the v105/v35 bulk range migration (which this PR is). Might be worth fixing the generator's `buildIndexSql` rendering of the `sorted` expression column separately.

## Testing

- `seed.test.ts` passes — confirms `v109` + `v39` apply via the real migration runner (incl. post-deploy version finalization).
- `search.test.ts` (481), `searchparameter.test.ts`, `range-column.test.ts`, `migrate.test.ts` all pass.
- The new test guards the legacy (non-`range-search`) variant with `if (!features) return;`, matching existing array-range-column tests, since array quantity params aren't supported on the legacy column path.

### DIFF
--- a/packages/definitions/src/fhir/r4/search-parameters-medplum.json
+++ b/packages/definitions/src/fhir/r4/search-parameters-medplum.json
@@ -1958,6 +1958,23 @@
         "type": "reference",
         "expression": "HealthcareService.offeredIn"
       }
+    },
+    {
+      "fullUrl": "https://medplum.com/fhir/SearchParameter/MeasureReport-measure-score",
+      "resource": {
+        "resourceType": "SearchParameter",
+        "id": "MeasureReport-measure-score",
+        "url": "https://medplum.com/fhir/SearchParameter/MeasureReport-measure-score",
+        "version": "4.0.1",
+        "name": "measure-score",
+        "status": "draft",
+        "publisher": "Medplum",
+        "description": "The computed score for a group within a measure report",
+        "code": "measure-score",
+        "base": ["MeasureReport"],
+        "type": "quantity",
+        "expression": "MeasureReport.group.measureScore"
+      }
     }
   ]
 }
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -1476,6 +1476,61 @@ describe.each<Project['features']>([undefined, ['range-search']])('project-scope
       expect(bundle3.entry?.[0]?.resource?.id).toStrictEqual(observation3.id);
     }));
 
+  test('Filter MeasureReport by measure-score', () =>
+    withTestContext(async () => {
+      if (!features) {
+        return; // measure-score is an array range column, only supported with the range-search feature
+      }
+      const measure = 'Measure/' + randomUUID();
+
+      const report1 = await repo.createResource<MeasureReport>({
+        resourceType: 'MeasureReport',
+        status: 'complete',
+        type: 'individual',
+        measure,
+        period: { start: '2024-01-01', end: '2024-12-31' },
+        group: [{ measureScore: { value: 0.25 } }],
+      });
+
+      const report2 = await repo.createResource<MeasureReport>({
+        resourceType: 'MeasureReport',
+        status: 'complete',
+        type: 'individual',
+        measure,
+        period: { start: '2024-01-01', end: '2024-12-31' },
+        group: [{ measureScore: { value: 0.9 } }],
+      });
+
+      const greaterThan = await repo.search<MeasureReport>({
+        resourceType: 'MeasureReport',
+        filters: [
+          { code: 'measure', operator: Operator.EQUALS, value: measure },
+          { code: 'measure-score', operator: Operator.GREATER_THAN, value: '0.5' },
+        ],
+      });
+      expect(greaterThan.entry?.length).toStrictEqual(1);
+      expect(greaterThan.entry?.[0]?.resource?.id).toStrictEqual(report2.id);
+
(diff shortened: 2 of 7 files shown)


maddyli:
Hi @WonderPanda ! Thanks so much for opening this! I added a comment to the attached issue, let me know if that seems to fulfill your requirements. Open to a having a discussion there!

WonderPanda:
> Hi @WonderPanda ! Thanks so much for opening this! I added a comment to the attached issue, let me know if that seems to fulfill your requirements. Open to a having a discussion there!

Thank you! I've replied on the issue there. I'd be happy to update the PR to resolve conflicts

maddyli:
Closing in favor of discussion on issue

PR #9425, opened 2026-06-05
testing(packages/react): porting from jest to vitest

The-Alchemist:


packages/react/src/chat/ThreadChat/ThreadChat.test.tsx
```diff
@@ -17,14 +18,14 @@ type SubscriptionControllerEvents = {
   subscription: { type: 'subscription'; criteria: string; bundle: Bundle };
 };
 
-jest.mock('@medplum/react-hooks', () => {
+vi.mock('@medplum/react-hooks', async (importOriginal) => {
   // eslint-disable-next-line @typescript-eslint/no-require-imports
   const { TypedEventTarget } = require('@medplum/core');
   const _subscriptionController = new TypedEventTarget() as TypedEventTarget<SubscriptionControllerEvents>;
-  const original = jest.requireActual('@medplum/react-hooks');
+  const original = await importOriginal<typeof MedplumReactHooks>();
```
noahsilas: You can simplify the types here a little by inlining the react-hooks import:

```ts
vi.mock(import('@medplum/react-hooks'), async (importOriginal) => {
  const original = await importOriginal();
  ...
});
```

Separately, I think there's a lot of room to tidy up some of the strangeness related to this controller stub, but that probably should happen separately from this larger jest/vitest movement.

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.test.ts
```diff
@@ -34,9 +34,10 @@ describe('assignValuesIntoSlices', () => {
       for (const url of profilesToLoad) {
         const sd = USCoreStructureDefinitions.find((sd) => sd.url === url);
         if (!sd) {
-          fail(`could not find structure definition for ${url}`);
+          expect.fail(`could not find structure definition for ${url}`);
+        } else {
```
noahsilas: It looks like this `else` nesting after `expect.fail()` gets introduced in a few places in this PR. Semantically I think this is the same as omitting the nesting, but it makes me wonder if there is a difference between how jest's `fail` and vitest's `expect.fail` are typed? Is this solving a problem in the type layer, or just a stylistic preference?

packages/react/src/AppShell/AppShell.test.tsx
```diff
@@ -98,45 +98,25 @@ describe('AppShell v1', () => {
...
-      fireEvent.keyDown(input, { key: 'ArrowDown', code: 'ArrowDown' });
-    });
-
-    // Press "Enter"
-    await act(async () => {
-      fireEvent.keyDown(input, { key: 'Enter', code: 'Enter' });
-    });
-
-    expect(navigateMock).toHaveBeenCalledWith('/test-code');
+    await selectAutocompleteOption(input, 'Test');
```
noahsilas, line 105: Nice extraction, I was surprised browsing through at just how often this is useful!

packages/react/src/SignatureInput/SignatureInput.test.tsx
```diff
@@ -4,19 +4,10 @@ import { MockClient } from '@medplum/mock';
 import { MedplumProvider } from '@medplum/react-hooks';
 import type { ReactNode } from 'react';
 import SignaturePad from 'signature_pad';
+import type { Mock } from 'vitest';
 import { act, render, screen } from '../test-utils/render';
 import { SignatureInput } from './SignatureInput';
 
-jest.mock('signature_pad', () => {
```
noahsilas, line 10: For any other readers: this mock setup gets moved into packages/react/src/test-mocks/signature_pad.ts.

packages/react/src/test.setup.ts
```diff
@@ -3,8 +3,40 @@
...
+  var IS_REACT_ACT_ENVIRONMENT: boolean | undefined;
+  interface Window {
+    IS_REACT_ACT_ENVIRONMENT?: boolean;
+  }
+}
+
+// Vitest's jsdom environment exposes `self` separately from `globalThis`. React reads
+// IS_REACT_ACT_ENVIRONMENT from globalThis while Testing Library toggles it on self, so
+// async RTL act() calls do not suppress spurious "not wrapped in act(...)" warnings.
+// See: https://github.com/vitest-dev/vitest/issues/1146
```
noahsilas: Is this still needed today? It looks like this linked issue should have been closed by https://github.com/vitest-dev/vitest/pull/1256, which I hope would have removed the need for this trickiness.

The-Alchemist:
@cursoragent review

packages/react/src/chat/ThreadChat/ThreadChat.test.tsx
The-Alchemist: done!  yeah, a separate PR makes sense for the controller stub strangeness

packages/react/src/ResourceArrayInput/ResourceArrayInput.utils.test.ts
The-Alchemist: done!  I think it was Cursor's stylistic preference

packages/react/src/test.setup.ts
The-Alchemist: yup, you're right.  done

packages/react/src/SignatureInput/SignatureInput.test.tsx
The-Alchemist: not sure what you mean?  Vitest uses a resolve.alias to `test-mocks/signature_pad.ts` so these changes are semantically equivalent.  but if you suggest a different approach, i'll definitely take a look

noahsilas: Sorry I wasn't more clear - I was initially surprised that tests were passing with this removal. It took me a moment to realize that the mock was still being applied, just from a different place. So this is purely here as a pointer for any future readers of this diff to help them if they encounter the same question.

No action needed here. 👍

noahsilas (approved):

PR #9433, opened 2026-06-05
logging(data warehouse): decreased logging and duplicates

The-Alchemist:


packages/server/src/data-warehouse/sync.ts
```diff
@@ -82,7 +84,10 @@ async function runWarehouseTableSync(
   for (const [index, spec] of options.warehouseSources.entries()) {
     const { icebergTable, postgresTable } = spec;
     const tablesCompleted = index + 1;
+    const watermarkStartTime = new Date();
```
mattwiller: You could use `Date.now()` directly here instead of needing to call `.getTime()` on each date below

mattwiller (approved):

PR #9434, opened 2026-06-06
Data Warehouse DuckDB Lifecycle Cleanup

codyebberson:
This PR makes a small, targeted cleanup change to the DuckDB lifecycle used by data warehouse sync jobs and removes an unnecessary source-table sort.

The motivation is production ECS telemetry showing increasing `Storage write (Bytes/Second)` over time on tasks running the S3 Tables data warehouse sync. The rate resets after deploys, which suggests process-lifetime DuckDB/native/temp-resource buildup rather than normal cumulative storage growth.

We do not know yet whether this fully explains the production behavior, but the missing `DuckDBInstance.closeSync()` is an obvious lifecycle gap and is worth fixing. Removing the unconditional sort also reduces a query shape that can cause DuckDB to spill intermediate data to local storage.

## Changes

- Close the `DuckDBInstance` after each sync, not just the DuckDB connection.
- Remove the unconditional `ORDER BY lastUpdated` from history-table export queries.
- Use async `fs.promises.rm()` for deleting the per-sync DuckDB temp directory.
- Close DuckDB instances in related integration tests.

## Why Remove `ORDER BY`

The previous query sorted every source history table by `lastUpdated` before writing to the destination. For Iceberg/S3 Tables, row order is not a correctness contract, and consumers should order explicitly when needed.

The sort can force DuckDB to materialize and spill intermediate state for large tables, which is a plausible contributor to elevated local write throughput. Removing it avoids a global sort during each table sync.

Tradeoff: local Parquet and Iceberg writes no longer have deterministic row order. That should not affect warehouse correctness.

### DIFF
--- a/packages/server/src/data-warehouse/config.int.test.ts
+++ b/packages/server/src/data-warehouse/config.int.test.ts
@@ -79,6 +79,7 @@ describe('config (integration)', () => {
       expect(row.t).toBe('3s');
     } finally {
       connection.closeSync();
+      instance.closeSync();
     }
   }, 30_000);
 });
--- a/packages/server/src/data-warehouse/sync.int.test.ts
+++ b/packages/server/src/data-warehouse/sync.int.test.ts
@@ -129,6 +129,7 @@ describe('syncData local destination (integration)', () => {
       expect(row.project_id).toBe('project-from-json');
     } finally {
       c.closeSync();
+      instance.closeSync();
     }
   }, 30_000); // longer timeout because of database operations
 });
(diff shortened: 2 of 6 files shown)


codyebberson:
Deploying to staging: https://github.com/medplum/medplum/actions/runs/27048403069

codyebberson:
@claude please review

reshmakh (approved):

codyebberson:
@claude please re-review

PR #9435, opened 2026-06-06
feat(data warehouse): restore ORDER BY

The-Alchemist:
ORDER BY in the query makes sure the data is ordered in the Parquet files,
which means the metadata in the Parquet files will contain the proper metadata
for each row group, such as min/max.

This metadata is then used by Iceberg and query engines, and suboptimal metadata
will result can result in slow/inefficient query plans.

The risk of spilling to disk is slow, as the data is small and ECS Fargate ephemeral disk
tasks have a default of 20GB, which is more than enough for each resource type.

### DIFF
--- a/packages/server/src/data-warehouse/warehouse-sql.test.ts
+++ b/packages/server/src/data-warehouse/warehouse-sql.test.ts
@@ -19,7 +19,7 @@ describe('warehouse SQL query builders', () => {
     sql.appendExpression(query);
 
     expect(sql.toString()).toBe(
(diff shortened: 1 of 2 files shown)


codyebberson (approved):
Broad strokes, I'm ok with restoring this.

> The risk of spilling to disk is slow, as the data is small and ECS Fargate ephemeral disk
> tasks have a default of 20GB, which is more than enough for each resource type.

Historically, Medplum hosts have used zero disk, including zero ephemeral disk, so this is an architectural change, even if it feels minor and insignificant.  I am still interested in exploring whether we can do everything in-memory or stream the Parquet files directly to S3 to avoid using disk all together.

See follow-ups here: https://github.com/medplum/medplum/issues/9436

PR #9446, opened 2026-06-08
Hold quota delay until after transaction in async mode

mattwiller:


packages/server/src/fhir/fhirquota.ts
```diff
@@ -19,6 +19,11 @@ export interface FhirQuotaConfig {
   projectLimit: number;
 }
 
+export const READ_QUOTA_COST = 1;
```
mattlong: To tidy up the imports elsewhere, could this be something like 
```typescript
export const FhirQuotaCost = {
  READ: 1,
  HISTORY: 10,
  SEARCH: 20,
  WRITE: 100
} as const;
```

packages/server/src/fhir/repo.ts
```diff
@@ -359,8 +359,22 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
 
-  private rateLimiter(): FhirRateLimiter | undefined {
-    return this.isSuperAdmin() ? undefined : tryGetRequestContext()?.fhirRateLimiter;
+  async recordFhirQuota(points: number): Promise<void> {
+    const ctx = tryGetRequestContext();
+    const limiter = this.isSuperAdmin() ? undefined : ctx?.fhirRateLimiter;
+    if (ctx instanceof AuthenticatedRequestContext && ctx.isAsync) {
+      // Do not enforce rate limits in async context; instead, slow down the consumer
+      // in proportion to the weight of the operation being performed
+      const delay = points * (getConfig().asyncDelayScaling ?? 1);
```
mattlong: Could you add asyncDelayScaling to `DefaultConfigKeys` since a default is set in `addDefaults`? That should avoid needing the `?? 1` fallback here.

mattlong (approved):
Couple small change request, but otherwise LGTM!

PR #9447, opened 2026-06-08
fix(core): recreate WebSocket subscriptions on reconnect

d-vincent:
## Summary

WebSocket subscriptions stop delivering after the socket reconnects, until a full page reload. The server drops the `Subscription` resource bound to a connection when that connection closes, but on reconnect the SDK reuses the **same** `Subscription.id` — so `$get-ws-binding-token` fails with the now-deleted id on every reconnect, forever. This clears `subscriptionId` during `refreshAllSubscriptions` so the entry is recreated on the new connection.

## Problem

On a socket reconnect we observe, for each previously-bound criteria:

```
GET /fhir/R4/Subscription/{id}/$get-ws-binding-token  →  400 Bad Request
OperationOutcome diagnostics: "Error reading subscription: Not found"
```

The criteria's entry reverts to `idle`/`active`, `error` is emitted, and the next rebind requests the **same** `{id}` again — looping. The WebSocket itself stays `OPEN`; only the bind is dead. Recovery requires a page reload (which creates fresh `Subscription` resources).

Triggered by any socket drop — laptop sleep, deploy, network blip, or the SDK's own auto-reconnect.

## Root cause

`refreshAllSubscriptions()` resets the entry's token/state on reconnect but deliberately preserved `subscriptionId`:

```ts
// We preserve subscriptionId so rebindCriteriaEntry can reuse
// the existing Subscription resource instead of creating a new one.
```

Since the server drops the `Subscription` when the prior connection closes, the preserved id is no longer bindable, and `rebindCriteriaEntry()` (which only creates when `subscriptionId` is unset) retries the dead id indefinitely.

## Change

Two commits:

1. **`fix(core): recreate WebSocket subscriptions on reconnect`** — clear `subscriptionId` in `refreshAllSubscriptions()` alongside `token`/`tokenExpiry`. On the next `rebindCriteriaEntry()` the entry has no `subscriptionId`, so it creates a fresh `Subscription` and binds it on the new connection. One line, plus a comment update. _(Per maintainer feedback — thanks! — this replaces an earlier catch-block recovery approach with this more surgical fix.)_

2. **`fix(core): tolerate WebSocket error events without an error property`** — `event.error.toString()` in the `SubscriptionManager` error handler throws a `TypeError` when the `error` event has no `.error` (the normal shape of a browser WebSocket error event; also how `mock-socket` fires errors), which surfaced as unhandled errors during the reconnect tests. Null-guarded with `event.error?.toString()`.

## Test

_"should recreate the Subscription rather than reuse a dropped subscriptionId after reconnect"_ — after a socket drop + reconnect, asserts `createResource` is called a second time and a fresh `Subscription` id binds. **Fails on `main`** (`createResource` called once — reuses the dead id); passes with change 1. Verified deterministic across repeated runs; `prettier` / `eslint` / `tsc` clean for the changed files.

## Validation

Also validated against the production Medplum backend (`api.medplum.com`): orphaned subscriptions are recreated on reconnect and delivery resumes without a page reload.

---


ThatOneBro:
Thanks @d-vincent, this is a great find!

Specifically the cause of the bug is that we clean up the WebSocket subscriptions on WebSocket disconnect, so we cannot reuse the subscriptionId on reconnect. However, we do not clear the `subscriptionId` in `refreshAllSubscription`, which is the source of this bug.

I think the more surgical fix would be add:
`criteriaEntry.subscriptionId = undefined`
here:
https://github.com/medplum/medplum/blob/main/packages/core/src/subscriptions/index.ts#L468

And then if we could change my previous erroneous comment to reflect that we are intentionally NOT keeping `subscriptionId` since the Subscription will have been cleaned up on the server side, and we need to create a new one, that would be perfect.

Also would love a regression test 🙏 

Thanks again and also great work! 😄

packages/core/src/subscriptions/index.ts
```diff
@@ -223,7 +223,7 @@ export class SubscriptionManager {
               details: {
                 text: 'WebSocket connection closed due to an error',
               },
-              diagnostics: event.error.toString(),
+              diagnostics: event.error?.toString(),
```
ThatOneBro: I don't believe this should be necessary, the type of this event is `ErrorEvent` which guarantees an `error` property. Can we figure out why there was an event that didn't have an `error` property? It could be an event other than an error event that was emitted?

d-vincent: @ThatOneBro  I think I have a reasonable path for this behavior:
 
 SubscriptionManager's handler is registered on the ReconnectingWebSocket, and the RWS forwards the underlying socket's native error event:

  - reconnecting-websocket.ts:592 — this._ws.addEventListener('error', this._handleError)
  
  - _handleError(event: ErrorEvent) (:546) is typed ErrorEvent, but that's an unchecked assertion — it's wired to the native socket's error event.
  
  - A native WebSocket error event is a plain Event, not an ErrorEvent — per the WHATWG spec the WebSocket interface fires a simple Event on error, so there's no .error (or
  .message). This is true in real browsers, not just mock-socket.
  
  - _handleError then dispatches it via cloneEvent

  function cloneEvent(e) { return new e.constructor(e.type, e); }

  - For a native event that's new Event('error', …) → a bare Event, still no .error. So the SubscriptionManager receives an event typed ErrorEvent whose .error is undefined, and
  event.error.toString() throws.
  
  Instead of my simple `?.toString()` fix, we could do something like this in _handleError in `reconnecting-websockets.ts`: 
  
private readonly _handleError = (event: ErrorEvent): void => {
    const error = event.error instanceof Error ? event.error : new Error(event.message || 'WebSocket error');
    this._debug('error event', error.message);
    this._disconnect(undefined, error.message === 'TIMEOUT' ? 'timeout' : undefined);
    if (this.onerror) this.onerror(event);
    this.dispatchEvent(new Events.ErrorEvent(error, this));   // not cloneEvent — guarantees `.error`
    this._connect();
  };

ThatOneBro: Hmm yeah long term we should probably fix the typings in `ReconnectingWebSocket`, but that might be some work to get it right.

In the meantime can we try to get `event.error?.toString() ?? event.message`

ThatOneBro (approved):
🙏 Thank you for this @d-vincent

PR #9453, opened 2026-06-09
Server: Scheduling Parameters can be inherited from HealthcareService

### LINKED ISSUE #8383: Scheduling Parameters: Refactor appointment property fields
## Background
Right now, our scheduling parameters extension has two types of fields:
1. Things that define the availability of the actor
2. Things that define the properties of the appointment itself

For the second category, we should be able to define them at the actor level, but also be able to inherit those properties from the ActivityDefinition level. 

## Proposal
* Move all fields that are properties of the appointment into a sub-extension. 
* For our inheritance logic at the additional capability that we merge the activity definition level of these properties with the actor level definition of these properties, with the actor level taking precedence.

noahsilas:
When a Practitioner defines their availability, they may not intend to deviate from other practitioner's scheduling setup, particularly with` respect to fields like `duration` and `alignmentInterval` (as these must match across Schedule resources to be jointly used in $find/$book). The intent here is to make this common case easy.

To that end, when computing the SchedulingParameters to use, we now merge several "layers" of configuration:
- Entries from the `Schedule` scheduling-parameters extension matching the requested service type
- Entries from the `HealthcareService` scheduling-parameters extension
- The system default values
- The timezone information found on `Schedule.actor[0].extension("http://hl7.org/fhir/StructureDefinition/timezone").value`

In the future we may add additional layers (such as being able to specify a "wildcard" availability within your schedule to apply to multiple service-types).

We track these layers in a new data structure, the "Layered Dictionary". This allows fast lookups, but also provides a resolution mechanism for us to identify the source of a given configuration value. This lets us emit helpful error messages now, and may be used as a tool to help explain the mechanics of the system later. It has some nice mechanics for TypeScript to understand the fields inside the dictionary.

Resolves #8383 - This is slightly less precise than what was requested there. In this PR I bias towards flexibility, and still allow individual `Schedule` resources to override fields like `duration` and `alignmentInterval`, although I've documented that this is "not recommended". I chose this after experimenting a bit and feeling like the more complex nesting structure added significantly more pain to simple scenarios / getting started. I also felt like it was conceptually nice to say "all fields default to inheriting from HealthcareService". We may still want to adjust this before exiting the `beta` window for this feature.

## Potential Breaking Change

Common cases we have seen so far either specify SchedulingParameters only on one layer (HealthcareService or Practitioner), or fully specify SchedulingParameters on the Practitioner. These scenarios will not see any changes as a result of this PR.

In the scenario where someone has defined scheduling parameters on `HealthcareService`, and defined scheduling parameters on a `Practitioner`'s `Schedule`, and omitted a value on the schedule level, that would have been getting the default value and now would start inheriting the value from `HealthcareService`.

**Example**:
A HealthcareService defines `{ duration: 30, alignmentInterval: 30 }`:
```
{
  "resourceType": "HealthcareService",
  "id": "hs-1",
  "extension": [
    {
      "url": "https://medplum.com/fhir/StructureDefinition/SchedulingParameters",
      "extension": [
        { "url": "duration", "valueDuration": { "value": 30, "unit": "min" } },
        { "url": "alignmentInterval", "valueDuration": { "value": 30, "unit": "min" } }
      ]
    }
  ]
}
```

A Practitioner's Schedule defines `{ duration: 30, bufferBefore: 10 }` for this service:
```
{
  "resourceType": "Schedule",
  "id": "s-2",
  "extension": [
    {
      "url": "https://medplum.com/fhir/StructureDefinition/SchedulingParameters",
      "extension": [
        { "url": "service", "valueReference": { "reference": "HealthcareService/hs-1" } },
        { "url": "duration", "valueDuration": { "value": 30, "unit": "min" } },
        { "url": "bufferBefore", "valueDuration": { "value": 10, "unit": "min" } }
      ]
    }
  ]
}
```

Before this change, the practitioner's schedule would have used the default alignment interval of `60` ("on-the-hour" scheduling). After this change, it will inherit the alignment interval of `30` from the HealthcareService.

### DIFF
--- a/packages/docs/docs/scheduling/defining-availability.md
+++ b/packages/docs/docs/scheduling/defining-availability.md
@@ -915,7 +915,7 @@ This Schedule will use the default availability for the "Follow-Up" service (Mon
       "extension": [
         "url": "https://medplum.com/fhir/service-type-reference",
         "valueReference": {
-          "reference": "HealthcareService/new-patient-visit,
+          "reference": "HealthcareService/new-patient-visit",
           "display": "New Patient Visit"
         }
       ]
@@ -931,7 +931,7 @@ This Schedule will use the default availability for the "Follow-Up" service (Mon
       "extension": [
         "url": "https://medplum.com/fhir/service-type-reference",
         "valueReference": {
-          "reference": "HealthcareService/follow-up,
+          "reference": "HealthcareService/follow-up",
           "display": "Follow-up Visit"
         }
       ]
--- a/packages/docs/docs/scheduling/defining-availability.md
+++ b/packages/docs/docs/scheduling/defining-availability.md
@@ -5,7 +5,11 @@ sidebar_position: 10
 
 # Defining Availability (Alpha)
 
-This guide covers how to configure availability using the `SchedulingParameters` extension — at both the actor level (per Schedule) and the service type level (via HealthcareService). It covers scheduling constraints, override behavior, timezone handling, and multi-resource scheduling patterns.
+This guide covers how to configure availability using the `SchedulingParameters` extension — at both the actor level (per Schedule) and the service type level (via HealthcareService). It covers scheduling constraints, field-level inheritance, override behavior, timezone handling, and multi-resource scheduling patterns.
+
+Parameters may be defined on a HealthcareService and shared amongst all Schedules that book that type
+of appointment. Each Schedule may also override these parameters to define behaviors specific to that
+schedule as needed.
 
 The diagram below shows how availability can be defined at both
 - The [actor level](/docs/scheduling/defining-availability#actor-level-availability) (via Schedule)
@@ -15,8 +19,8 @@ The diagram below shows how availability can be defined at both
 %%{init: {'theme':'base', 'themeVariables': {'fontSize':'14px'}, 'flowchart': {'useMaxWidth': false, 'htmlLabels': true}}}%%
 graph TD
     C1[Practitioner<br/>*Dr. Smith*] --> B1
-    A[HealthcareService<br/>*Initial Visit Defaults*<br/><br/><b>Service-level<br/>defined availability</b><br/>ex. <i>1hr slot duration</i>] -.-> B1[Schedule<br/>*Dr. Smith's Schedule*<br/><br/><b>Actor-level<br/>defined availability</b><br/>ex. <i>Mon–Fri, 9am–5pm</i>]
-    A -.-> B2[Schedule<br/>*Dr. Johnson's Schedule*<br/><br/><b>Actor-level<br/>defined availability</b><br/>ex. <i>Mon–Fri, 9am–5pm</i>]
+    A[HealthcareService<br/>*Initial Visit Defaults*<br/><br/><b>Service-level defaults</b><br/>ex. <i>1hr slot duration, 10min buffer</i><br/><i>Mon–Fri 9am–5pm availability</i>] -.->|inherited by| B1[Schedule<br/>*Dr. Smith's Schedule*<br/><br/><b>Actor-level overrides</b><br/>ex. <i>availability: Mon–Wed only</i>]
+    A -.->|inherited by| B2[Schedule<br/>*Dr. Johnson's Schedule*<br/><br/><b>no overrides</b><br/><i>uses all service defaults</i>]
     C2[Practitioner<br/>*Dr. Johnson*] --> B2
 
     B1 --> D1[Slot<br/>*status: busy*]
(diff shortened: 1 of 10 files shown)


noahsilas:
@claude please review

noahsilas:
> **Error path precision regression in assertAllMatch**

This was something I went back-and-forth on for a bit while working on this PR that comes from a limitation of the `withPath` helper - we can't decorate primitive values like `number`, and so the two obvious options here are to create a layer per sub-extension, or to reduce the specificity.

It's nice to have the very exact pointer to the precise sub-extension where a field came from, but I think that this level of precision is not necessary - most sub-extensions are only allowed once, and some others like `availability` can be present multiple times and are merged (which, if tracked, introduces a different style of layer combination: deep-merge instead of pure overlay).

Instead, sticking with tracking only at the granularity of the SchedulingParameters top-level extension, we reduce the number of layers that are tracked, which should help keep the system more legible for future work.

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -881,6 +927,13 @@ This Schedule will use the default availability for the "Follow-Up" service (Mon
           "system": "http://example.org/appointment-types",
           "code": "follow-up"
         }
+      ],
+      "extension": [
+        "url": "https://medplum.com/fhir/service-type-reference",
+        "valueReference": {
+          "reference": "HealthcareService/follow-up,
```
mattwiller: Missing the trailing `"` here:
```suggestion
          "reference": "HealthcareService/follow-up",
```

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -872,6 +911,13 @@ This Schedule will use the default availability for the "Follow-Up" service (Mon
           "system": "http://example.org/appointment-types",
           "code": "new-patient-visit"
         }
+      ],
+      "extension": [
+        "url": "https://medplum.com/fhir/service-type-reference",
+        "valueReference": {
+          "reference": "HealthcareService/new-patient-visit,
```
mattwiller: Missing the trailing quote here:
```suggestion
          "reference": "HealthcareService/new-patient-visit",
```

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -328,17 +333,31 @@ export async function getSchedulingParametersGroup(
...
-        throw new OperationOutcomeError(badRequest('No timezone specified', getPath(actor)));
+
+      let parameters = getScheduleSchedulingParameters(schedule, healthcareService, serviceParams);
+
+      const timezone = getTimeZone(actor);
+      if (timezone) {
+        // Tricky: `timezone` is defined to prefer scheduling-parameter
+        // definitions coming from HealthcareService or Schedule extensions
+        // over per-actor configuration. We put the actor-based layer at the
+        // bottom of the stack with `prepend` to give it lowest priority.
```
mattwiller, line 349: 👍

packages/server/src/util/layereddict.ts
```diff
@@ -0,0 +1,108 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
```
mattwiller, line 1: [nit] May want to align the name of these file with the others, e.g. `layered-dict.ts`

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
```diff
@@ -381,90 +315,165 @@ function extractAvailability(
...
+): LayeredDict<ServiceSchedulingParameters> {
+  const defaultsLayer = withPath(
+    {
+      service: createReference(healthcareService),
+      ...SERVICE_DEFAULTS,
+    },
+    getPath(healthcareService)
+  );
+
+  let result: LayeredDict<ServiceSchedulingParameters> = LayeredDict.empty().addLayer(defaultsLayer);
```
mattwiller: Would it make sense to add a way to construct a new `LayeredDict` directly from `defaultsLayer`, without creating and throwing away the original `LayeredDict.empty()`?

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
```diff
@@ -381,90 +315,165 @@ function extractAvailability(
...
-    const bufferBefore = atMostOne(getExtensions(extension, 'bufferBefore'), 'bufferBefore');
-    const bufferAfter = atMostOne(getExtensions(extension, 'bufferAfter'), 'bufferAfter');
-    const alignmentOffset = atMostOne(getExtensions(extension, 'alignmentOffset'), 'alignmentOffset');
-    const rawAlignmentInterval = atMostOne(getExtensions(extension, 'alignmentInterval'), 'alignmentInterval');
+  const extensions = getExtensions(healthcareService, SchedulingParametersURI);
+  if (extensions.length === 0) {
+    // Proposal: make this an error before scheduling GA launch. Consider
+    // making `duration` a required field at the same time. This would simplify
+    // the type logic (`duration` could always be guaranteed, removing the
+    // difference between ServiceSchedulingParameters and SchedulingParameters.
```
mattwiller, line 360: Do we need to make a decision on this?

mattwiller (approved):
I like the new `LayeredDict` abstraction, just a few small nits but otherwise makes sense to me!

packages/docs/docs/scheduling/defining-availability.md
noahsilas: Oops, thank you!

packages/server/src/util/layereddict.ts
noahsilas: That was my initial thought too, but the other entries in [this folder](https://github.com/medplum/medplum/tree/v5.1.17/packages/server/src/util) are already in this format, following the example of `auditevent.ts`. Do you think these should all be converted to kebab-case? Happy to follow up on that!

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
noahsilas: We need to make this choice at some point. It's a little awkward if we do go down this path, as it introduces a potentially significant breaking change (scheduling that worked before w/ `parameters only defined on the `Schedule` layer would stop working). My thought was to follow up on this separately, probably with some advice on how to detect the problem and prepare for it.

For now, this PR preserves the current behavior, and deals with a bit of extra complexity in the types to do so.

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
noahsilas: Yeah, that's probably sensible. What do you think about `LayeredDict.from(baseLayer)`?

mattwiller: I like that 👍

packages/server/src/util/layereddict.ts
mattwiller: I feel like it would be generally nice to standardize, but agree that maybe it would be better to do a more thorough pass at a later time

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
noahsilas: Done in [64ba3c1](https://github.com/medplum/medplum/pull/9453/commits/64ba3c14ff90578f81aea9be9c8bc74961b0cfa1)

PR #9459, opened 2026-06-09
[Docs] - Shared Projects

finnbergquist:


packages/docs/docs/access/projects.md
```diff
@@ -21,19 +21,64 @@ Additionally, [`Projects`](/docs/api/fhir/medplum/project) each have their own u
...
 
+### Setting Up a Shared Project
+
+Configuring project linking is a two-step process that requires super-admin access:
+
+**Step 1: Create the shared project**
+
+Create a new [`Project`](/docs/api/fhir/medplum/project) that will hold the shared resources. This project will not be used directly by end users — it exists solely as a container for shared content.
+
+It is **highly recommended** to set `exportedResourceType` on the shared project to restrict which resource types are visible to linked projects. If `exportedResourceType` is omitted, all resource types in the shared project are accessible.
```
deam65: For consistency, maybe we should recommend that `exportedResourceType` is always specified. Line 112 to `EOF` says to always specify this value

deam65 (approved):

packages/docs/docs/access/projects.md
finnbergquist: Good call

PR #9460, opened 2026-06-09
Rebuild Operation Definitions

### LINKED ISSUE #9424: Include `OperationDefinition` resources in the "Rebuild R4" flow
We currently have:
1. Rebuild Structure Definitions
2. Rebuild Search Parameters
3. Rebuild Value Sets

Which sync the definition resources in `@medplum/definitions` into the "R4" project.

Customer feature request to include `OperationDefinition` resources.

That makes me wonder if we should just globally import all of the definition resources.

Those are currently 3 different super admin endpoints.

I also wonder if it should simply be one operation, and "Resource Types" is an optional input parameter where the caller can specify which resource types to either include or exclude.

mattwiller:
Resolves #9424


packages/server/src/fhir/operations/rebuild-base-definitions.ts
```diff
@@ -0,0 +1,177 @@
...
+
+type InputParams = {
+  resourceType?: ResourceType[];
+};
+
+export async function rebuildBaseDefinitionsOperation(req: FhirRequest): Promise<FhirResponse> {
+  const { repo } = getAuthenticatedContext();
+  if (!repo.isSuperAdmin()) {
+    return [forbidden];
+  }
```
mattlong:
```suggestion
  const { repo } = requireSuperAdmin();
```

packages/server/src/fhir/operations/rebuild-base-definitions.ts
```diff
@@ -0,0 +1,177 @@
...
+      if (entry.resource?.resourceType === 'StructureDefinition') {
+        await processStructureDefinitionEntry(repo, entry as BundleEntry<StructureDefinition>);
+      } else if (entry.resource?.resourceType === 'OperationDefinition') {
+        await processOperationDefinitionEntry(repo, entry as BundleEntry<OperationDefinition>);
+      }
+    });
+  }
+
+  // Terminology Resources: Code Systems and Value Sets
+  if (types?.includes('CodeSystem') !== false || types.includes('ValueSet')) {
```
mattlong: The if statements throughout are hard to grok at first glance since they have different structures given the possibility of `types` being undefined. It might be worth breaking out the `undefined` cases somehow or maybe just add a comment explaining the intent.

packages/server/src/fhir/operations/rebuild-base-definitions.ts
```diff
@@ -0,0 +1,177 @@
...
+  return systemRepo.withTransaction(async () => {
+    const bundle = await systemRepo.search({
+      resourceType: resource.resourceType,
+      filters: [
+        { code: 'url', operator: Operator.EQUALS, value: resource.url as string },
+        { code: '_project', operator: Operator.EQUALS, value: projectId },
+      ],
+    });
+    for (const entry of bundle.entry ?? EMPTY) {
+      const existing = entry.resource as WithId<CodeSystem | ValueSet>;
```
mattlong:
```suggestion
      const existing = entry.resource as WithId<T>;
```

packages/server/src/fhir/operations/rebuild-base-definitions.ts
```diff
@@ -0,0 +1,177 @@
...
+  await replaceExisting(repo, param, r4ProjectId);
+}
+
+async function processStructureDefinitionEntry(
+  repo: Repository,
+  entry: BundleEntry<StructureDefinition>
+): Promise<void> {
+  const sd = entry.resource;
+  if (!sd?.name) {
+    return;
```
mattlong, line 95: Worth logging that something is being skipped?

packages/server/src/fhir/operations/rebuild-base-definitions.ts
```diff
@@ -0,0 +1,177 @@
...
+async function processTerminologyDefinitionEntry(
+  repo: Repository,
+  entry: BundleEntry<CodeSystem | ValueSet>
+): Promise<void> {
+  const resource = entry.resource as CodeSystem | ValueSet;
+  await replaceExisting(repo, resource, r4ProjectId);
+}
+
+async function replaceExisting<T extends Resource & { url?: string }>(
+  systemRepo: Repository,
```
mattlong:
```suggestion
  systemRepo: SystemRepository,
```

packages/server/src/fhir/operations/rebuild-base-definitions.ts
```diff
@@ -0,0 +1,177 @@
...
+function cleanSeedResource<T extends Resource>(resource: T): T {
+  const clean = {
+    ...resource,
+    meta: {
+      ...resource.meta,
+      project: r4ProjectId,
+      lastUpdated: undefined,
+      versionId: undefined,
+    },
+    text: undefined,
```
mattlong, line 162: Curious why we can't add differential here like:
```suggestion
    text: undefined,
    differential: undefined,
```

packages/server/src/seeds/searchparameters.ts
```diff
@@ -1,37 +1,12 @@
...
 import type { Repository } from '../fhir/repo';
-import { globalLogger } from '../logger';
 
 /**
  * Creates all SearchParameter resources.
  * @param systemRepo - The system repository to use
  */
 export async function rebuildR4SearchParameters(systemRepo: Repository): Promise<void> {
-  const client = systemRepo.getDatabaseClient(DatabaseMode.WRITER);
-  await client.query('DELETE FROM "SearchParameter" WHERE "projectId" = $1', [r4ProjectId]);
```
mattlong, line 16: The project-wide delete is lost for `SearchParameter` and `StructureDefinition` (but it wasn't there for ValueSet?) That increases chance for the R4 project to diverge from what's in the definition files if rogue resources were added and changes the semantics of rebuilding away from a full wipe and rebuild. Is that the direction we want to go?

mattlong (changes requested):
Biggest question is on the loss of the `DELETE`.

packages/server/src/fhir/operations/rebuild-base-definitions.ts
mattwiller: It wasn't being logged before, so I'm inclined to keep it that way

packages/server/src/seeds/searchparameters.ts
mattwiller: I discussed previously with @codyebberson and was persuaded that it's preferable to ensure that there's no moment when the definitions aren't available in the database, since that could have adverse downstream effects.  We could wrap the entire `DELETE FROM ...` and resource creation loop, but that might end up being a relatively large chunk of work, so I thought the individual transactions would be easier to deal with.  Additionally, it made testing much easier, since the entire table doesn't get wiped out by the tests for this code

ThatOneBro: Can we do conditional PUT with the version ID?

EDIT: I guess we would have to actually assign a version ID deterministically somehow 😅 

I guess we could also read the resource and do a deep equals before attempting to write?

EDIT2: Another less FHIR-y approach would be to store a hash of the resource in the resource itself, as an identifier or similar. That would avoid the read and make the check fast

mattwiller: A conditional PUT would fail if the server had gotten into a state where two resources with the same `url` exist, which is potentially a situation we would expect the rebuild operation to fix.  Is the primary concern here the performance, something else, or some combination?

codyebberson:
@claude please review

mattlong (approved):

PR #9466, opened 2026-06-09
Provider: fix back button after fax page loading

noahsilas:
Noticed today during Provider-sync meeting.

See also: ea52bb251


codyebberson (approved):
Nice catch @noahsilas

techdavidy (approved):

PR #9467, opened 2026-06-09
Propagate AWS Trace IDs

mattwiller:


packages/server/src/otel/instrumentation.ts
```diff
@@ -62,6 +64,10 @@ export function initOpenTelemetry(): void {
     metricReader = new PeriodicExportingMetricReader({ exporter });
   }
 
+  const propagator = new CompositePropagator({
+    propagators: [new AWSXRayPropagator(), new W3CTraceContextPropagator()],
```
mattlong, line 68: [The docs](https://github.com/open-telemetry/opentelemetry-js/tree/b711a81d5262904245d70f1857b6f3bc811b22cd/experimental/packages/opentelemetry-sdk-node#textmappropagator) mention that the default includes a fallback to a "baggage propagator", do we need to include that?

mattwiller: We're not currently using baggage anywhere, so it should be safe to omit until we are

The-Alchemist (approved):

PR #9468, opened 2026-06-09
Add logo wordmark

codyebberson:
https://github.com/user-attachments/assets/93e1f44a-b266-492c-89e5-6497fa158596


ThatOneBro (dismissed):

kevinwadeshaw:
@codyebberson Love this idea, but in practice is doesn't look great with logos like ours where the symbol and wordmark are optically (not metrically) centered. So you end up with the symbol uncentered/unaligned in the collapsed state:

Logo Alignment

---

If we want to represent the full logo in the navbar, I recommend we pair this with the collapse/expand action button when expanded and then, in the collapsed state, only show this icon button in place of the symbol. Most other products that use a logo in a sidebar like this use a similar pattern (see Claude on web for example).

Navbar with Logo

The-Alchemist (approved):

codyebberson:
@kevin - yeah, agreed, there are some rough edges when you play with it in practice.  

This particular PR cheated a little bit so that the word "Medplum" was outside.  At present, if we make it centered and the correct size, then you see part of the "M", which looks dumb.

To do this right, we will need to do a bit deeper integration into `<Navbar>`.  Right now, the caller passes in a `logo`, and `<Navbar>` uses it "as is'.  To get the ideal behavior, we'll need to sort that out.

PR #9471, opened 2026-06-10
fix(cli): throw error when bot source/dist file is not found

### LINKED ISSUE #3692: Bot Deployment Silently Fails in CLI
When running the `medplum bot deploy` command with the cli, we have found that it fails silently when an error is encountered. 

For example: we had a bot's id mismatched in one of our environments which encountered a "NOT FOUND" error (logged out by the CLI), but it still returned a successful status code, and subsequently passed in our CI/CD.

I recommend making it return a non-zero status code.

MadhukarJukanti:
## Summary

  - `saveBot` and `deployBot` silently returned early when the configured source/dist file did not exist on disk, incrementing success counters and exiting with code 0
  - This caused CI/CD pipelines to see a successful deployment that never actually ran
  - Replace the silent early-return with a thrown `Error` so failures are reported to stderr and the process exits with code 1

  ## Root Cause

  `readFileContents` returns an empty string when a file does not exist. Both `saveBot` and `deployBot` checked `if (!code) { return; }` — silently skipping deployment without adding to the error list, so
  `botWrapper` still counted them as successes.

  ## Changes

  - `packages/cli/src/utils.ts`: `saveBot` and `deployBot` now throw an `Error` instead of silently returning when the code file is missing
  - `packages/cli/src/bot.test.ts`: Added tests for "Save bot source file not found" and "Deploy bot dist file not found" — both expect exit code 1

  ## Test plan
  - [ ] `Save bot source file not found` — `process.exit(1)` and stderr message
  - [ ] `Deploy bot dist file not found` — `process.exit(1)` and stderr message
  - [ ] All existing bot tests pass

  Fixes #3692


MadhukarJukanti:
@ThatOneBro  I've rebased this branch into latest main and resolved the conflicts. This PR is ready for review when you get a chance.

PR #9476, opened 2026-06-10
Wire up FHIRPath Patch in server

### LINKED ISSUE #1263: Implement Fhirpatch
Implement the patch semantics for Fhirpath . Will open up a host of applications for reading/writing/aliasing properties of resources

mattwiller:
Resolves #1263


packages/server/src/fhir/repo.ts
```diff
@@ -1246,7 +1251,15 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
...
-        patchObject(resource, patch);
+        if (Array.isArray(patch)) {
+          patchObject(resource, patch);
+        } else if (patch.parameter) {
+          const params = parseParametersFromDefinitions(
+            patchOperationDefinition.parameter as OperationDefinitionParameter[],
+            patch.parameter
+          );
+          fhirpathPatchTypedValue(toTypedValue(resource), params.operation as FhirPathPatch[]);
+        }
```
mattlong, line 1278: Should there be an `else` block for defense in depth against against bad inputs?

packages/server/src/fhir/routes.test.ts
```diff
@@ -481,6 +481,36 @@ describe('FHIR Routes', () => {
     expect(res.status).toBe(200);
   });
 
+  test('FHIRPath Patch resource success', async () => {
+    const res = await request(app)
```
mattlong, line 486: Might be worth adding an expect before the request to confirm what you're operating on:

```typescript
expect(updatedPatient.name?.[0].given).toStrictEqual(['Alice']);
```

packages/server/src/fhir/repo.ts
mattwiller: I suppose we could just return `resource` here?  It would have otherwise fallen through to `updateResourceImpl()` as a no-op update, which would effectively do the same thing

codyebberson:
@claude please review

codyebberson:
@claude please re-review

codyebberson:
@mattwiller  - thoughts on updating the `MemoryRepository` function declarations?

codyebberson (approved):

codyebberson (approved):

PR #9479, opened 2026-06-11
fix(agent,ws): reconnect agent in various error states

### LINKED ISSUE #9478: [Agent] Make reconnect more robust against poor connectivity scenarios
There are a few scenarios which the agent does not fully account for during the reconnect lifecycle for the agent WebSocket connection, including transient DNS timeouts during auth or reconnect, WebSocket disconnecting during initial connection handshake, and a few other long-tail scenarios. We should make sure to fix those and add coverage for as many of those scenarios as we have currently identified.

ThatOneBro:
Fixes #9478

### DIFF
--- a/packages/agent/src/app.test.ts
+++ b/packages/agent/src/app.test.ts
@@ -342,6 +342,133 @@ describe('App', () => {
     mockServer.stop();
   });
 
+  test('Attempt to reconnect when WebSocket is open but agent never becomes live', async () => {
+    const state = {
+      connectRequestCount: 0,
+      respondToConnect: false,
+    };
+
+    const mockServer = new Server('wss://example.com/ws/agent');
+    mockServer.on('connection', (socket) => {
+      socket.on('message', (data) => {
+        const command = JSON.parse((data as Buffer).toString('utf8'));
+        if (command.type === 'agent:connect:request') {
+          state.connectRequestCount += 1;
+          // Simulate the server failing to process the connect request (e.g. a transient error
+          // while validating the token or reading the Agent resource) -- the socket stays open
+          // but no `agent:connect:response` is ever sent
+          if (state.respondToConnect) {
+            socket.send(Buffer.from(JSON.stringify({ type: 'agent:connect:response' })));
+          }
+        }
+      });
+    });
+
+    const agent = await medplum.createResource<Agent>({
+      resourceType: 'Agent',
+      name: 'Test Agent',
+      status: 'active',
+    });
+
+    const app = new App(medplum, agent.id, LogLevel.INFO);
+    app.heartbeatPeriod = 100;
+    await app.start();
+
+    // Wait for the first connect request, which goes unanswered
+    while (state.connectRequestCount < 1) {
+      await sleep(50);
+    }
+
+    // The agent should notice it never became live and force a reconnect,
+    // which re-sends the connect request -- this time we answer it
+    state.respondToConnect = true;
+    while (state.connectRequestCount < 2) {
+      await sleep(50);
+    }
+
+    expect(state.connectRequestCount).toBeGreaterThanOrEqual(2);
+
+    await app.stop();
+    mockServer.stop();
+  });
+
+  test('WebSocket queue worker recovers after a failed send', async () => {
+    const state = {
+      mySocket: undefined as Client | undefined,
+      live: false,
+      transmitRequestCount: 0,
+    };
+
+    const mockServer = new Server('wss://example.com/ws/agent');
+    mockServer.on('connection', (socket) => {
+      state.mySocket = socket;
+      socket.on('message', (data) => {
+        const command = JSON.parse((data as Buffer).toString('utf8'));
+        if (command.type === 'agent:connect:request') {
+          socket.send(Buffer.from(JSON.stringify({ type: 'agent:connect:response' })));
+        }
+        if (command.type === 'agent:transmit:request') {
+          state.transmitRequestCount += 1;
+        }
+      });
+    });
+
+    const agent = await medplum.createResource<Agent>({
+      resourceType: 'Agent',
+      name: 'Test Agent',
+      status: 'active',
+    });
+
+    const app = new App(medplum, agent.id, LogLevel.INFO);
+    // Long heartbeat period so the heartbeat retry doesn't interfere with this test --
+    // recovery should be driven by the next enqueued message alone
+    app.heartbeatPeriod = 30_000;
+    await app.start();
+
+    // Wait for the WebSocket to connect and the agent to become live
+    while (!state.mySocket) {
+      await sleep(50);
+    }
+    while (!app.getStats().live) {
+      await sleep(50);
+    }
+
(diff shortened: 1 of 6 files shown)


packages/agent/src/app.ts
```diff
@@ -857,10 +897,15 @@ export class App {
...
-        this.webSocketWorker = undefined;
+      .catch((err) => {
+        this.log.error(`WebSocket worker error: ${normalizeErrorString(err)}`);
       })
-      .catch((err) => console.log('WebSocket worker error', err));
+      .finally(() => {
+        // Always clear the worker, even on error -- otherwise the queue could never drain again.
+        // The failed message was put back on the queue; the next enqueue (or the heartbeat)
+        // restarts the worker.
+        this.webSocketWorker = undefined;
```
noahsilas, line 907: Moving this into a `finally` seems like a strong improvement. One way this seems like it could still get into trouble is if sending a message hangs and never resolves for some reason. Should we add some kind of timeout around `trySendToWebSocket()` to catch that kind of scenario? Or do we know that the underlying requests already have timeouts baked in?

Thinking about it, this might actually be more of a feature request for `core.MedplumClient` to decorate more requests it makes with `AbortSignal.timeout` so that every implementer using it doesn't have to think about this 🤔

packages/core/src/websockets/reconnecting-websocket.ts
```diff
@@ -541,9 +548,15 @@ export class ReconnectingWebSocket<WS extends IWebSocket = WebSocket>
     if (!this._ws) {
       return;
     }
+    const ws = this._ws;
     this._removeListeners();
+    // The abandoned socket can still emit 'error' after close() -- e.g. the `ws` library aborts a
+    // still-CONNECTING handshake and emits 'error' on the next tick, after this method returns.
+    // On an EventEmitter-backed implementation, an 'error' event with no listeners is thrown as an
+    // uncaught exception and kills the process, so keep a no-op listener attached.
```
noahsilas, line 556: #TIL about this special case for the `error` event. Oof, good catch.

packages/agent/src/app.test.ts
```diff
@@ -342,6 +342,259 @@ describe('App', () => {
...
+
+    // Wait for the first connect request, which goes unanswered
+    while (state.connectRequestCount < 1) {
+      await sleep(50);
+    }
+
+    // The agent should notice it never became live and force a reconnect,
+    // which re-sends the connect request -- this time we answer it
+    state.respondToConnect = true;
+    while (state.connectRequestCount < 2) {
```
noahsilas: If something goes wrong here, this while loop won't exit and the test will eventually timeout, which is a less helpful/debuggable error message than hitting the assertion below. It would be nice if we tracked the MS delay so far and exited this loop before the test timeout.

Similar thoughts on some testing loops below.

Not a merge blocker IMO.

packages/agent/src/app.ts
```diff
@@ -266,16 +273,38 @@ export class App {
...
+    if (!this.live) {
+      // The WebSocket can be open without the agent ever becoming live: the
+      // `agent:connect:request` may have failed to send, or the server may have failed to
+      // process it (without closing the socket). Neither side sends traffic on such a
+      // connection, so nothing would ever break us out of this state -- force a reconnect
+      // after enough heartbeat periods. This also acts as a backstop while the underlying
+      // ReconnectingWebSocket is retrying on its own after a close.
+      this.notLiveHeartbeats += 1;
+      if (this.notLiveHeartbeats > MAX_NOT_LIVE_HEARTBEATS) {
+        this.notLiveHeartbeats = 0;
```
noahsilas, line 285: When reconnect is called, does the heartbeat stop?

Without full context, this looks like it hits a boundary condition and then immediately resets to below that boundary, which seems like it could cause infinite loops. Should we fail in some other way if we hit this path multiple times?

One way this could look might be something like this:
```ts
      this.notLiveHeartbeats += 1;
      if (this.notLiveHeartbeats % NOT_LIVE_HEARTBEAT_CYCLE === 0) {
        if (this.notLiveHeartbeats > NOT_LIVE_HEARTBEAT_MAX) {
          throw new Error(...);
        }
        this.log.warn(...);
        this.webSocket.reconnect();
      }
```

I might also be misunderstanding the nature of what goes on behind the scenes in the reconnect handler and this might not be an issue at all.

noahsilas (approved):
These seem like good improvements to me! A couple thoughts/questions in comments, but nothing that seemed like it should be a blocker 👍

packages/core/src/websockets/reconnecting-websocket.ts
techdavidy: I know, learning so much here.

packages/agent/src/app.test.ts
mattlong: We should generally be able to avoid sleeps altogether with the `Promise.withResolvers()` approach, right?

```typescript
const ignoredConnectRequest = Promise.withResolvers<undefined>();
const respondedConnectRequest = Promise.withResolvers<undefined>();

mockServer.on('connection', (socket) => {
  socket.on('message', (data) => {
    const command = JSON.parse((data as Buffer).toString('utf8'));
    if (command.type === 'agent:connect:request') {
      state.connectRequestCount += 1;
      // Simulate the server failing to process the connect request (e.g. a transient error
      // while validating the token or reading the Agent resource) -- the socket stays open
      // but no `agent:connect:response` is ever sent
      if (state.respondToConnect) {
        socket.send(Buffer.from(JSON.stringify({ type: 'agent:connect:response' })));
        respondedConnectRequest.resolve(undefined)
      } else {
        ignoredConnectRequest.resolve(undefined);
      }
    }
  });
});

...

await ignoredConnectRequest.promise;
state.respondToConnect = true;
await respondedConnectRequest.promise;

expect(state.connectRequestCount).toBeGreaterThanOrEqual(2);
```

packages/agent/src/app.ts
ThatOneBro: The tricky part is a timeout is hard to implement for the function as a whole, as the only actually async here is the `refreshIfExpired` call to get a new token if a token is expired, and we're not actually waiting for a response to come back here. It's technically possible for the token refresh that to hang I think? Assuming the server never actually responds. Though I think in practice we have a 60 second load balancer timeout that would be triggered. 

But yeah wonder if we would want to bake in a timeout to that in `MedplumClient` itself

packages/agent/src/app.test.ts
ThatOneBro: I think the issue is that it would still timeout if we didn't get one of the connect requests, I think @noahsilas 's point was to timeout within a reasonable amount of time before the whole test times out, so we hit the assertion which clues us into the fact we didn't get all the connection requests we expected

packages/agent/src/app.ts
ThatOneBro: Think it's a reasonable callout, in practice though this will be on the order of ~20 seconds without getting a response back and so attempting to reconnect after 20 seconds in an "infinite" loop is par for the course for the agent, since uptime is a cornerstone of the agent 💯

PR #9480, opened 2026-06-11
testing(app): migrate from jest to vitest

The-Alchemist:
99% search-replace

Feedback welcome!


techdavidy:
@claude please review

packages/app/src/index.test.tsx
```diff
@@ -1,28 +1,82 @@
...
-    });
-
-    expect(await screen.findByTestId('search-control')).toBeInTheDocument();
+    const root = document.createElement('div');
+    root.id = 'root';
+    document.body.appendChild(root);
+    await initApp();
+
+    // createBrowserRouter + jsdom does not reliably complete HomePage URL redirects, so verify
+    // initApp mounted an authenticated App shell instead of search-control specifically.
```
noahsilas, line 73: This is surprising, I wonder what's up with this? (Not a blocker, just curious why it would have worked in `jest` but sometimes doesn't with your more detailed mocks in this `vitest` suite)

packages/app/jest.config.json
```diff
@@ -1,19 +0,0 @@
...
-  "testEnvironmentOptions": {
-    "globalsCleanup": "on"
-  },
-  "testTimeout": 120000,
-  "transform": {
-    "^.+\\.(js|jsx|ts|tsx)$": "babel-jest"
-  },
-  "moduleFileExtensions": ["ts", "tsx", "js", "jsx", "json", "node"],
-  "moduleNameMapper": {
-    "\\.css$": "identity-obj-proxy"
```
noahsilas, line 12: I think with this removal we might be able to also remove this package from the root package.json

noahsilas (approved):

techdavidy (approved):

packages/app/jest.config.json
The-Alchemist: i think you're right!  i'll submit a separate PR removing `identity-obj-proxy`.

packages/app/src/index.test.tsx
The-Alchemist: I can look into this, but it might be because vitest is stricter in some areas than jest.  I asked Cursor to generate comments for areas that might be confusing, FWIW.

PR #9481, opened 2026-06-11
Patient everything inline attachments

codyebberson:
Continuing @ianplunkett's work from here: https://github.com/medplum/medplum/pull/9000

* Moving `Binary.securityContext` check into `Repository` rather than checking in many places
* Added new `inlineAttachmentsMaxTotalBytes` server config setting that applies to entire response, not per attachment


packages/server/src/fhir/operations/patienteverything.ts
```diff
@@ -196,6 +221,103 @@ function collectReferences(resource: any, foundReferences = new Set<string>()):
...
+      }
+      try {
+        let binary: Binary;
+        if (versionId) {
+          binary = await repo.readVersion<Binary>('Binary', id, versionId);
+        } else {
+          binary = await repo.readResource<Binary>('Binary', id);
+        }
+        const stream = await getBinaryStorage().readBinary(binary);
+        const buffer = await readStreamToBufferWithLimit(stream, remainingBytes);
```
noahsilas, line 261: Feels a bit of a shame to potentially read in lots of bytes only to drop them if we go over the limit. My first intuition was that we should try to record the file size somewhere on `Binary` when it is written, so that we can short circuit here with less IO. There's not really an obvious place it can go on a FHIR Binary resource though.

It looks like these `Binary` resources are coming via `DocumentReference.attachment`, which does have a `size` attribute, that could be another option? Still would require other updates to write to that field, and we probably would still want this code-path as a backup for any resources that don't have the `size` attribute at all.

Final thought here is that this might be something we should add some monitoring on: if we're hitting this limit often, it might be worthwhile to do the extra IO for an S3 `HEAD` request to get the filesize up front. My intuition is that this is uncommon and that would be more expensive than the optimistic path of trying to read each file.

All that is to say I think what you have already written is probably the best initial approach 👍

packages/server/src/fhir/operations/patienteverything.ts
```diff
@@ -196,6 +221,103 @@ function collectReferences(resource: any, foundReferences = new Set<string>()):
...
+        if (buffer === null) {
+          skippedCount++;
+          getLogger().debug('Skipping attachment inline: exceeds remaining inline attachment budget', {
+            binaryId: id,
+            remainingBytes,
+            maxBytes,
+          });
+          continue;
```
noahsilas, line 270: Would it make sense to write something into the result that indicates that this was intentionally not inlined, so that consumers know it was not a bug?

packages/server/src/fhir/operations/patienteverything.ts
```diff
@@ -196,6 +221,103 @@ function collectReferences(resource: any, foundReferences = new Set<string>()):
...
+        getLogger().debug('Error inlining DocumentReference attachment', { error: err });
+      }
+    }
+  }
+  getLogger().debug('Finished inlining DocumentReference attachments', { inlinedCount, skippedCount, totalBytes });
+}
+
+/**
+ * Reads a stream into a Buffer, aborting and returning null if the total byte count exceeds maxBytes.
+ * The URL is left untouched for attachments that exceed the limit.
```
noahsilas: I think this line of the comment about "the url" is not applicable to this function (probably should move to the caller?)

packages/server/src/fhir/repo.ts
```diff
@@ -537,11 +523,21 @@ export class Repository extends FhirRepository<PoolClient> implements Disposable
       await this.setCacheEntry(resource);
     }
 
+    return this.authorizeBinarySecurityContext(resource);
+  }
+
+  private async authorizeBinarySecurityContext<T extends Resource>(resource: WithId<T>): Promise<WithId<T>> {
```
noahsilas: This doesn't look like it depends on the `id` field at all, I think you could drop this annotation and remove an `as WithId<Resource>` cast below.

noahsilas (approved):

packages/server/src/fhir/operations/patienteverything.ts
codyebberson: Noted for future improvement

packages/server/src/fhir/operations/patienteverything.ts
codyebberson: Noted for future improvement

codyebberson:
@claude please review

PR #9483, opened 2026-06-11
feat(ai): support LiteLLM via configurable base URL and dynamic model list

maddyli:
## Summary
- Add an optional `OPENAI_BASE_URL` project secret to the `$ai` operation so it can target any OpenAI-compatible endpoint (e.g. a LiteLLM proxy). Defaults to `https://api.openai.com/v1` when unset, so existing projects are unchanged.
- Make the Spaces model dropdown configurable via an `aiModels` `Project.setting` (JSON array of `{value, label}`), with a built-in default fallback in `spaceModels.ts`.
- Add a generic **Settings** admin page (mirroring the Secrets page) plus a server `POST /admin/projects/:id/settings` endpoint to manage `Project.setting` from the main admin app.
- Document the LiteLLM routing requirement: each `value` must match a `model_name` alias in the proxy's `config.yaml` (no validation in UI or `$ai`).

## Details
- `$ai` is wire-compatible with LiteLLM (same `/chat/completions` body, response shape, and `Bearer` auth), so only the base URL is configurable — no request/response translation. Trailing slashes on the secret are normalized.
- Realtime transcription WebSocket is intentionally out of scope.

## Test plan
- [ ] `packages/server`: `npm test -- src/fhir/operations/ai.test.ts` (new base-URL default/custom/trailing-slash cases) — *blocked locally: no Postgres/Redis in this env*
- [ ] `packages/server`: `npm test -- src/admin/project.test.ts` (new save-settings + 403 cases) — *blocked locally: no DB*
- [x] `examples/medplum-provider`: `spaceModels.test.ts` (9 cases) pass
- [x] `packages/app`: `SettingsPage.test.tsx` + ProjectPage/Secrets/Sites suites pass
- [ ] Manual: set `OPENAI_BASE_URL` + `aiModels` on a project, confirm `$ai` routes to a LiteLLM proxy and the Spaces dropdown reflects the configured models


packages/server/src/fhir/operations/ai.ts
```diff
@@ -134,6 +135,10 @@ export async function aiOperation(
     return [badRequest('OpenAI API key not configured in project secrets')];
   }
 
+  const baseUrl = (
+    ctx.project.secret?.find((s) => s.name === 'OPENAI_BASE_URL')?.valueString ?? 'https://api.openai.com/v1'
```
techdavidy: I would say do LLM_BASE_URL or similar

maddyli: For sure. My main consideration is that if we have non-OpenAI formatted calls, such as directly to Anthropic's models or other, should we also include a new base URL for those ones? Or should this be an array of configurations?

techdavidy: I think we can store aws bedrock configs in secrets and use that to find out whether we should use that or not.

techdavidy: e.g Use bedrock else use  LLM_BASE_URL else error
```
 const region = getSecret('AWS_BEDROCK_REGION') ?? getConfig().awsRegion;
  const accessKeyId = getSecret('AWS_BEDROCK_ACCESS_KEY_ID');
  const secretAccessKey = getSecret('AWS_BEDROCK_SECRET_ACCESS_KEY');
  if (accessKeyId && secretAccessKey) {
    return new BedrockRuntimeClient({
      region,
      credentials: { accessKeyId, secretAccessKey, sessionToken: getSecret('AWS_BEDROCK_SESSION_TOKEN') },
    });
  }
```

packages/app/src/admin/SettingsPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <form
+      noValidate
+      autoComplete="off"
+      onSubmit={(e: FormEvent) => {
+        e.preventDefault();
+        medplum
+          .post(`admin/projects/${projectId}/settings`, settings)
+          .then(() => medplum.get(`admin/projects/${projectId}`, { cache: 'reload' }))
+          .then(() => showNotification({ color: 'green', message: 'Saved' }))
+          .catch(console.log);
```
techdavidy: show error message

packages/app/src/admin/SettingsPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const projectId = getProjectId(medplum);
+  const projectDetails = medplum.get(`admin/projects/${projectId}`).read();
+  const [schemaLoaded, setSchemaLoaded] = useState(false);
+  const [settings, setSettings] = useState<ProjectSetting[] | undefined>();
+
+  useEffect(() => {
+    medplum
+      .requestSchema('Project')
+      .then(() => setSchemaLoaded(true))
+      .catch(console.log);
```
techdavidy: display erro message

packages/server/src/fhir/operations/ai.ts
maddyli: I see, but when it comes to the `/v1/chat/completions` that the OpenAI calls are formatted with, wouldn't that be different depending on the LLM_BASE_URL used? 

Happy to rename the config regardless, I think we just need to have another configuration eventually for the exact path after the base URL

techdavidy:
@claude please review

techdavidy:
yeah let's solve return [badRequest('Failed to call AI API: ' + (error as Error).message)];

techdavidy (approved):

PR #9484, opened 2026-06-11
feat(react): ResourceBoard - shell that encapsulates inbox logic

techdavidy:
⏺ ResourceBoard is a generic master-detail "inbox" shell for FHIR resources — a left sidebar list plus
  a right-hand detail pane, with the data-fetching built in (backed by the useResourceBoard hook in
  @medplum/react-hooks).

  Layout it renders:
  - Sidebar: an optional header row (pill tabs that navigate via URI + a right-aligned headerActions
  slot), a scrollable resource list, and built-in pagination.
  - Detail: a flex area for the selected resource, with an empty-state fallback.

  What it owns vs. what you supply:
  - Owns — the search/fetch lifecycle, loading + selected state, pagination, and the per-item chrome
  (each row is wrapped in a <div> that paints hover/selected).
  - You supply — search (resourceType + filters + count/offset), selectedId, renderItem(item, ctx) for
  row content, and renderDetail(selected, ctx) for the detail pane. Slots for
  skeleton/emptyList/emptyDetail.

⏺ ListWithDetailPane is a generic, purely presentational master-detail shell. It owns layout only — no data fetching, no routing. You hand it the already-loaded data + render functions, and it draws a two-pane UI and emits callbacks.
  
 ⏺ First implementation of  ResourceBoard is in FaxBoard/Page.


techdavidy:
@claude please review

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,157 @@
...
+  // Adjust state during render (SearchControl pattern) so a new search or fetcher
+  // shows the skeleton without a synchronous setState inside an effect.
+  if (!deepEquals(search, memoizedSearch)) {
+    setMemoizedSearch(search);
+    setLoading(true);
+  }
+  if (loadItems !== prevLoadItems) {
+    setPrevLoadItems(() => loadItems);
+    setLoading(true);
+  }
```
noahsilas: It looks like this is the only place `prevLoadItems` is used, and the result here is to re-set the `loading` bit each time it changes? It seems like this is functionally the same as this effect:

```ts
useEffect(() => { setLoading(true); }, [loadItems]);
```

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,157 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { MedplumClient, SearchRequest } from '@medplum/core';
+import { deepEquals, formatSearchQuery } from '@medplum/core';
+import type { Resource } from '@medplum/fhirtypes';
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
+import { useMedplum } from '../MedplumProvider/MedplumProvider.context';
+
+export interface ResourceBoardLoadResult {
+  readonly items: Resource[];
```
noahsilas: A bunch of the types in here feel like they might be better expressed with generics:

```suggestion
export interface ResourceBoardLoadResult<T extends Resource> {
  readonly items: WithId<T>[];
```

That could reduce the usage of casting in resource board users to translate `Resource` into the more specific types.

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
```diff
@@ -107,19 +110,120 @@ export function ThreadInbox(props: ThreadInboxProps): JSX.Element {
...
-    total,
-    handleThreadStatusChange,
-    addThreadMessage,
-    refreshThreadMessages,
-  } = useThreadInbox({
-    query,
-    threadId,
-  });
+  // Last message per thread id, rebuilt by every load and read by renderItem.
+  const lastMessagesRef = useRef<Map<string, Communication>>(new Map());
```
noahsilas: This feels pretty awkward. I think it works, because the `loadItems` function won't resolve until this is updated, meaning it should be populated before the `renderItem` function depending on it is ever run. But it still seems like the kind of workaround that suggests that we are contorting our data flow in a bizarre way.

 I wonder if we can attach this "last-message" information to the results in some other way? 

One thought is that maybe the return type of `loadItems` could be something like `Promise<{ resource: Communication, lastMessage: Communication}>` or something similar? This would break the conceptualization of "resourceboard" a bit, as the result isn't exactly a "Resource" subtype. Maybe this suggests that there are two components we want here: one that is something like `<ListWithDetailPane>` that can handle an arbitrary input list, and then `<ResourceBoard>` which can do the more advanced search handling and wrap that inner component. 

Another strategy option that comes to mind: does this nested query need to be batched through this graphQL query at this layer? If we let `loadItems` be responsible just for loading the Communication list, and deferred loading the `lastMessage` component until the ThreadListItem was trying to use it, then maybe it would all get auto-batched by the Medplum client anyways? That could provide for a future world where items that are below the scroll line don't trigger their load right away, but defer until they come into view.

packages/react/src/ListWithDetailPane/ListWithDetailPane.tsx
```diff
@@ -0,0 +1,195 @@
...
+  readonly selected: boolean;
+  readonly index: number;
+  readonly items: WithId<T>[];
+}
+
+export interface ListWithDetailPaneDetailContext {
+  readonly refresh: () => Promise<void>;
+}
+
+export interface ListWithDetailPaneProps<T extends Resource = Resource> {
```
noahsilas: One idea to consider: should we make this generic less constrained? I think the only requirement we really have is that elements have the `id` attribute, so I think we could write:

```suggestion
export interface ListWithDetailPaneProps<T extends { id: string } = WithId<Resource>> {
```

Hypothetically, that could be used for components like `ThreadInbox` that require more complex data. 
```tsx
type ThreadRecord = { id: string, thread: WithId<Communication>, latestMessage?: WithId<Communication> };

const { threadMessages } = useThreadInbox(...);
const items = threadMessages.map(([thread, latestMessage]) => ({
  id: thread.id,
  thread,
  latestMessage
}));
return (
  <ListWithDetailPaneProps<ThreadRecord> items={items} ... />
);
```

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,164 @@
...
+        : await defaultLoadItems<T>(memoizedSearch, medplum);
+      if (requestId !== loadRequestIdRef.current) {
+        return;
+      }
+      setItems(result.items);
+      setTotal(result.total);
+      optionsRef.current.onLoad?.(result.items, result.total);
+      if (optionsRef.current.selectedId === undefined && result.items.length > 0) {
+        optionsRef.current.onSelectFirst?.(result.items[0]);
+      }
```
noahsilas, line 111: Since we will have to support this API indefinitely, I'm inclined to bias towards a smaller surface area - how would you feel about removing `onSelectFirst`, and letting the caller handle this as part of an `onLoad` callback?

You are closer to the usage than I am, so I'll also happily follow your lead on this design decision if you think this is the correct API as-is.

techdavidy:
@claude please review

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,164 @@
...
+  readonly refresh: () => Promise<void>;
+}
+
+async function defaultLoadItems<T extends Resource>(
+  search: SearchRequest,
+  medplum: MedplumClient
+): Promise<ResourceBoardLoadResult<T>> {
+  const bundle = await medplum.search(
+    search.resourceType,
+    formatSearchQuery({ ...search, total: search.total ?? 'accurate', fields: undefined }),
```
noahsilas: What's the story of `fields: undefined` in here? Is there some search parameter that we are excluding this wa?

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,164 @@
...
+    } finally {
+      if (requestId === loadRequestIdRef.current) {
+        setLoading(false);
+      }
+    }
+  }, [memoizedSearch, loadItems, medplum]);
+
+  // Effects
+  useLayoutEffect(() => {
+    optionsRef.current = options;
```
noahsilas, line 133: This seems like it could be `useEffect` instead of `useLayoutEffect` (which is preferable as a default, because layout effects block the render from painting to the screen). It might even not need an effect, and just be something we can put directly in the render function?

Is there a reason to do this in a layout effect? If so, we should probably leave a comment about it.

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,164 @@
...
+        if (requestId === selectionRequestIdRef.current) {
+          setSelectedState({ id: selectedId, value: undefined });
+          (optionsRef.current.onError ?? console.error)(error);
+        }
+      }
+    };
+    resolveSelection().catch(console.error);
+  }, [selectedId, items, medplum, resourceType]);
+
+  return { items, total, loading, selected, memoizedSearch, refresh: executeLoad };
```
noahsilas: The `memoizedSearch` name feels a little like it is leaking an implementation detail into the public API, but maybe it is important for the caller to understand that they will receive a stable value here?

My intuition is that we should name this something less proscriptive like `search` to leave ourselves room to maneuver in the future.

examples/medplum-provider/src/components/fax/FaxBoard.tsx
```diff
@@ -103,127 +87,60 @@ export function FaxBoard({ faxId, activeTab, inboxUri, sentUri, query, getFaxUri
...
+            fax={fax}
+            selectedFax={selected ? fax : undefined}
+            activeTab={activeTab}
+            getFaxUri={getFaxUri}
+            hideDivider={selected || (index < items.length - 1 && items[index + 1]?.id === faxId)}
+          />
         )}
-      </Flex>
+        emptyList={<EmptyFaxState activeTab={activeTab} />}
+        renderDetail={(fax) => <FaxDetailPanel fax={fax} onFaxChange={refreshList} />}
```
noahsilas, line 133: Strategy alternative thought: would it make sense to rely on react-router for this? The detail page is typically at a nested URL (in this case, `/Fax/Communication/:faxId`). You can imagine the routes looking something like:
```tsx
<Route path="Fax/Communication" element={FaxBoard}>
  <Route index element={FaxEmptyState} />
  <Route path=":id" element={FaxDetailPanel} />
</Route>
```

The downside of this kind of approach is that synchronizing the cache between the list view and the detail view is harder (although we could do it with something like a React Context to pass a handle to the currently loaded page of items into the detail view, or by putting the fetched items into the Medplum client cache). The upside is that things like "selection state" become more obviously controlled by the URL.

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
```diff
@@ -0,0 +1,164 @@
...
+    const resolve = async (): Promise<WithId<T> | undefined> => {
+      const custom = optionsRef.current.resolveSelected;
+      if (custom) {
+        return custom(selectedId, items, medplum);
+      }
+      const found = items.find((item) => item.id === selectedId);
+      if (found !== undefined) {
+        return found;
+      }
+      return (await medplum.readResource(resourceType, selectedId)) as WithId<T>;
```
noahsilas, line 158: One potential concern: this can find items by ID that don't match the search criteria. If a resource changes state and no longer should appear under on tab, this might be confusing.

I wonder if we can detect the difference between the situation where an element matches the search but is not on the current page vs. when the element does not match the search?

Maybe we can replace this `readResource` with a search query augmented with the ID? Something like:
```ts
  const resource = await medplum.searchOne(resourceType, { ...memoizedSearch, _id: selectedId });
```

examples/medplum-provider/src/components/fax/FaxBoard.tsx
techdavidy: we'll keep handling all url params as discussed offline

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
techdavidy: the feedback has been just to auto select one

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
techdavidy: had it as leftover

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
techdavidy: yeah there is a case where your filters display a set of items but the id of the display element is not par of the group.

examples/medplum-provider/src/components/fax/FaxBoard.tsx
noahsilas: Some additional notes from our discussion:

- The `ResourceBoard` component that this wraps should stay neutral to the choice of framework libraries like `react-router`, so it's not really appropriate to use something like `<Outlet>` inside it.
- We anticipate scenarios where the `selected` state is not stored in the URL; in cases like this `FaxBoard` can perform navigation, but it's not a requirement to use the ResourceBoard inner element.

packages/react-hooks/src/useResourceBoard/useResourceBoard.ts
noahsilas: I think there's probably room to improve here, but I think it can be an additive change later (probably something like showing an `<Alert>` if the resource doesn't match the search). Not a blocker on this PR.

noahsilas (approved):

PR #9486, opened 2026-06-11
testing(server): migrate from jest to vitest

The-Alchemist:
### mock hoisting
In Jest, `jest.mock()` is hoisted before everything, including setupFiles.

In Vitest, `vi.mock()` is only hoisted within each test file. test.setup.ts runs first as a setupFile, and its mocks apply globally — but per-test-file mocks do not run before setup finishes.

So the order is:

1. test.setup.ts loads and registers its mocks
2. Test file loads and registers its mocks
3. Test file imports run

If production code is imported during step 1 (or early in step 3), it can bind to the wrong thing.

### circular imports

```
vi.mock('pg') factory runs
  → import('./migrations/migration-versions')
    → migrations/data/v31.ts
      → import { DatabaseMode } from '../../database'
        → database.ts starts loading
          → import { Pool } from 'pg'   // while pg mock factory is still running!
```

### fetch mock helpers 
Vitest types `vi.spyOn(globalThis, 'fetch')` against the real fetch signature (`Promise<Response>`), whereas Jest’s mock types were loose enough that tests could return partial objects like `{ status, json: () => ... }` without TypeScript errors.  Helpers were created in `test.setup.ts` to assist.


packages/server/vite.config.ts
```diff
@@ -0,0 +1,83 @@
...
+      '@google-cloud/secret-manager': resolve(serverDir, 'src/__mocks__/@google-cloud/secret-manager.ts'),
+      '@google-cloud/storage': resolve(serverDir, 'src/__mocks__/@google-cloud/storage.ts'),
+    },
+  },
+  test: {
+    globals: true,
+    environment: 'node',
+    server: {
+      deps: {
+        inline: ['pg'],
```
ThatOneBro, line 52: Okay interesting, so this is for mocking I'm guessing?

Gemini said:
>Why use server.deps.inline?
> * Fixes ESM/CJS compatibility issues: Node.js sometimes struggles to process packages that output unbundled or poorly formatted ES modules. Inlining them lets Vite's pipeline process them safely.
> * Resolves native Node errors: It fixes common errors like `ERR_UNSUPPORTED_DIR_IMPORT` or issues where Vitest cannot parse syntax inside a node_modules package.
> * **Enables mocking/aliasing**: By default, Vitest cannot easily apply vi.mock() or resolve.alias to externalized packages. Inlining them pulls them into the workspace pipeline so mocking works flawlessly.

packages/server/src/admin/super.test.ts
```diff
@@ -5,30 +5,48 @@ import type { Bot, Login, Practitioner, Project, ProjectMembership, User } from
...
-import type { ReindexJobData } from '../workers/reindex';
-import { getReindexQueue } from '../workers/reindex';
+import type * as CronModule from '../workers/cron';
+import type * as LambdaCleanerModule from '../workers/lambda-cleaner';
+import type * as ReindexModule from '../workers/reindex';
+
+type RepositoryType = RepoModule.Repository;
+
+// Keep in sync with Repository.VERSION in fhir/repo.ts
+const REPOSITORY_VERSION = 15;
```
ThatOneBro: It would be preferable to avoid this somehow, otherwise any time we bump the `Repository.VERSION` we're going to have to remember to bump this

codyebberson:
Nice job @The-Alchemist 

@claude - please review.

Please confirm that this does not affect any runtime behavior, only tests and mocks.  Based on file names alone, I believe this only touches tests and mocks files.

codyebberson (approved):
Let's goooooo 🚀

packages/server/src/admin/super.test.ts
```diff
@@ -5,30 +5,48 @@ import type { Bot, Login, Practitioner, Project, ProjectMembership, User } from
...
+type CronJobData = CronModule.CronJobData;
+type LambdaCleanerJobData = LambdaCleanerModule.LambdaCleanerJobData;
+type ReindexJobData = ReindexModule.ReindexJobData;
+
+let getCronQueue: typeof CronModule.getCronQueue;
+let getLambdaCleanerQueue: typeof LambdaCleanerModule.getLambdaCleanerQueue;
+let getReindexQueue: typeof ReindexModule.getReindexQueue;
+
+// Keep in sync with LAMBDA_NAME_REGEX_PATTERN in cloud/aws/deploy.ts
+const LAMBDA_NAME_REGEX_PATTERN = '^medplum-bot-lambda-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$';
```
ThatOneBro: Same as above, any way to avoid this ?

ThatOneBro:
Looks good overall, excited to get this merged 🎉

packages/server/src/admin/super.test.ts
ThatOneBro: Although to be fair, this one probably won't change much

packages/server/src/admin/super.test.ts
The-Alchemist: this was a crazy change.  fixed.  see https://github.com/medplum/medplum/pull/9486/changes#diff-e2fe37702ef3f020d047310a6ed562212367199f36257cce94aca12de56afb9f

packages/server/src/admin/super.test.ts
The-Alchemist: also removed as part of https://github.com/medplum/medplum/pull/9486/changes#diff-e2fe37702ef3f020d047310a6ed562212367199f36257cce94aca12de56afb9f

packages/server/vite.config.ts
The-Alchemist: looks like the tests pass without the inlining!  it seems newer versions of vitest handle mocking CJS modules like `pg` better.

Issue #8734, opened 2026-03-20
Choose and set default scheduling `alignmentInterval`


noahsilas:
In the initial implementation we set the default `alignmentInterval` to "60", meaning "appointments start on the hour".

This satisfies some common use cases, but is awkward for others. For example, 30 minute appointments leave large gaps between schedule-able slots.

One option to explore: setting the alignment interval to match the slot duration (maybe inclusive of bufferBefore/bufferAfter times?). This could make "tip-to-tail" scheduling work nicely.

Potential complication: what happens when the duration isn't a nice divisor of an hour? The original `alignmentInterval` implementation assumed that valid values are ones like "30", "20", "15", which unambiguously divide every hour. If you have a value like "45", this could represent times like "8:00, 8:45, 9:30" or times like "9:00, 9:45, 10:30".

What happens when the duration is longer than an hour?

It seems like this sort of requires that alignment carries some notion of "reference time" to operate with respect to, which is probably the starting time of the availability block? We may need to carry that information through the scheduling flow to make sure we can respect it.

As one final side note, I think that today `$find` works fairly well for this, although it probably generates unexpected slots in some conditions where the availability start is modified by slots already on the calendar. In `$book` we punted on checking alignment to allow folks to set this in interesting ways with the intent of cleaning up and resolving later (see https://github.com/medplum/medplum/pull/8331).

rahul1:
Necessarily need to reinvent the wheel here. These are informed by the [Healthie API](https://docs.gethealthie.com/guides/scheduling/appointments/) and the [Cal.com API](https://cal.com/docs/api-reference/v2/orgs-teams-event-types/create-an-event-type#body-slot-interval). 

We can just copy what they do, which I think is what you proposed, where the interval start time is the reference time.

codyebberson:
My recommendation would be:

* Default `alignmentInterval` to the appointment duration.
* Use the availability block start time as the alignment reference point.
* Generate slots as `availabilityStart + N * alignmentInterval`.

For example, if availability starts at 9:00 AM and the duration/alignment interval is 45 minutes, the generated slots would be:

* 9:00
* 9:45
* 10:30
* 11:15
* ...

This works for arbitrary durations (45, 50, 73 minutes, etc.) without requiring a separate reference-time concept or imposing restrictions such as "must evenly divide an hour" or "must evenly divide a day".

This feels like a simple, explainable rule that we can document and support long-term.

noahsilas:
[...]

PR #9488, opened 2026-06-12
Support nondivisible alignment times in `$find`

noahsilas:
When using `alignmentInterval` that doesn't evenly divide the hour, you can end up with surprising results. For example, when using an alignment like "90 minutes", the grid alignment depended on the availability window that was being searched. This meant that searching for appointments on a single day could return slots aligned to two distinct grids as your search `start` time changed.

To resolve this, we need to choose a standard grid to use for this alignment, and ensure that we are matching it when generating appointment times in $find. In this PR we choose to align a daily grid to midnight (potentially offset by `alignmentOffset` scheduling parameter).

One tricky aspect to this is that the alignment grid should shift with local timezone rules. For example, imagine you are using a 90 minute alignment interval and your slots start at 9am PDT (-7). After the DST transition, if the alignment grid does not also shift, your slots would start beginning at 10am PST (-8).

Most businesses don't change their hours on these transitions, and so when they configure their alignment, the intention is usually for it to be consistent: "My day starts at 9am PT, and my first slot should align with that." To that end, we introduce a new SchedulingParameter attribute, `alignmentTimezone`. This lets you choose the zone that your alignment grid is anchored with respect to.

## Backwards compatibility note:
In cases where your slots are divisible by one hour, this does not make a difference[^1] (10,20,15,30,60 minute slots all evenly fill the entire daily grid). This was the only officially supported scenario in our previous alpha implementation; folks using other alignment values may see their scheduling grid suddenly jump when this is released.

I'm setting the default alignmentTimezone value to `Etc/UTC` - this is an easy-to-explain default that works well for some places (Arizona, Hawaii, Japan, anywhere that doesn't observe DST). It is likely that most users will want to override this to match their local timezone (e.g., `America/New_York`, `America/Los_Angeles`, `Europe/Paris`).

## Future Related Changes:
- This is building towards a solution for #8734, in which we would change the `alignmentInterval` default value. The most requested default value is one that matches the scheduling duration, allowing tip-to-tail booking by default. In that world, we can almost guarantee that alignment grids will diverge from evenly divisible values ("90 minutes" being a common example). For that reason, some kind of change like this is a prerequisite.

- With this updated slot generation in place, we should revisit enforcing alignment in $book. The original implementation assumed that the input value was evenly divisible, which broke on some outputs from $find in these non-divisible alignment values. We removed that restriction in #8331, which means that currently $book can be used to create appointments that violate scheduling constraints. We should update $book/$hold to test alignment with these new rules.

[^1]: Okay, it makes a difference in some rare edge cases, such as when you are on Lord Howe Island on a day in which a Daylight Savings Time transition shifts the clock by thirty minutes. I suspect that scheduling discontinuities in such environments are unsurprising.

### DIFF
--- a/packages/server/src/fhir/operations/utils/scheduling-parameters.ts
+++ b/packages/server/src/fhir/operations/utils/scheduling-parameters.ts
@@ -130,6 +130,10 @@ const SERVICE_DEFAULTS = Object.freeze({
   alignmentTimezone: 'Etc/UTC',
 });
 
+// This is a `Temporal.Instant` singleton that we instantiate once for
+// performance.
+const epochInstant = Temporal.Instant.fromEpochMilliseconds(0);
+
 function isReferenceTo<T extends Resource>(reference: Reference<T> | undefined, resource: WithId<T>): boolean {
   if (!reference?.reference) {
     return false;
@@ -175,7 +179,7 @@ function assertValidTimezone(ext: WithPath<Extension>): void {
   // "US/Pacific" is an alias for "America/Los_Angeles"), and is
   // case-insensitive (we accept "america/los_angeles" as valid).
   try {
-    Temporal.Instant.fromEpochMilliseconds(0).toZonedDateTimeISO(ext.valueCode);
+    epochInstant.toZonedDateTimeISO(ext.valueCode);
   } catch {
     throw new OperationOutcomeError(badRequest(`Invalid timezone '${ext.valueCode}'`, getPath(ext)));
   }
--- a/packages/server/src/fhir/operations/utils/scheduling-parameters.ts
+++ b/packages/server/src/fhir/operations/utils/scheduling-parameters.ts
@@ -149,6 +149,11 @@ function durationToMinutes(extension: WithPath<Extension>): number {
   if (value === undefined) {
     throw new OperationOutcomeError(badRequest('Got duration without value', getPath(extension)));
   }
+
+  if (value < 0) {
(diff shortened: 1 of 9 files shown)


noahsilas:
@claude please review

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
```diff
@@ -159,6 +168,19 @@ function durationToMinutes(extension: WithPath<Extension>): number {
...
 }
 
+function assertValidTimezone(ext: WithPath<Extension>): void {
+  assertExtensionCode(ext);
+  // Check that we can build a Temporal.ZonedDateTime with the given timezone.
+  // Note that this accepts non-canonical timezone identifiers (example:
+  // "US/Pacific" is an alias for "America/Los_Angeles"), and is
+  // case-insensitive (we accept "america/los_angeles" as valid).
+  try {
+    Temporal.Instant.fromEpochMilliseconds(0).toZonedDateTimeISO(ext.valueCode);
```
mattwiller: Technically, I think you could avoid re-creating `Temporal.Instant.fromEpochMilliseconds(0)` each time and could re-use a constant defined outside the function body to make this slightly more performant

mattwiller (approved):

packages/server/src/fhir/operations/utils/scheduling-parameters.ts
noahsilas: Nice catch, thanks!

PR #9490, opened 2026-06-12
feat(server): Stage 2 $install hook and uncapped install Bundles

oleg-mp:
## Summary

Three changes to package installation. The first two are confined to `packageinstall.ts` and its tests; the third is an access-control change in `core/access.ts` and `repo.ts`.

**Stage 2 install hook.** `$install` gains a second, imperative stage: it validates optional `Parameters` against the release's bundled `Questionnaire`, invokes the package's setup bot, and records the result idempotently so a re-run recovers rather than duplicating. Stage 1 failures restart from the beginning (the transaction did not commit); Stage 2 failures skip the committed Stage 1 bundle and retry setup.

The setup bot is resolved from the package's **shared impl project**, not from the calling project. This came out of review — every other bot in a package is published once and versioned, and there was no reason the hook should be the exception. Two things made the old shape actively wrong rather than merely inconsistent:

* It ran with `runAsUser: false`, so it executed as its own `ProjectMembership`, which is not a project admin and is never granted admin. Writing `Project.secret` from the install answers — the hook's main job — requires admin. Published as an impl bot with `runAsUser: true` it runs as the admin who invoked `$install`, which is the privilege the work needs and never more than the installer already had.
* Resolution cannot go through the caller. `RepositoryContext.projects` is snapshotted from `Project.link` when the request's `Repository` is built, so the link this handler appends a few lines earlier is invisible to `ctx.repo` until a later request. A system search scoped to the impl project the release names is deterministic on a first install. The escalation is bounded to the single bot the release names, and the caller has already proven access to that release — the same rationale as reading the install Bundle `Binary`.

`runAsUser: true` is **enforced** on the resolved bot rather than assumed, since the bot is published from another repo (medplum-ee#954) and nothing here would otherwise guarantee it. Without the check, `getBotProjectMembership` falls through to the bot's own membership in the impl project, and the hook would write the installing customer's settings into the shared impl project — visible to every other install of that package.

That also deletes the `deployBot` fallback, which existed only because a Bundle cannot carry `executableCode`. It had no test coverage, and it wrote as the caller, so it could never have targeted the impl project anyway.

**Reconciliation is version-aware.** `existing` is found by package reference, not by version, so the record for an installed v1 is what a caller installing v2 hits. Neither `installed` branch was safe across versions: identical settings hash the same and returned `allOk` having applied nothing, and differing settings took `skipStage1` while `upsertInstalling` relabelled the record to v2 — leaving an installation claiming a version whose Bundle never ran. The version is now compared directly. (Folding it into the config hash, the obvious alternative, only converts the no-op into a skipped Stage 1; the new Bundle still never applies.)

| Existing state | Same version | Different version |
| --- | --- | --- |
| `installed` | no-op if settings unchanged, else re-run Stage 2 | refused as a conflict naming both versions |
| `error` | resume past Stage 1 only if the failure was in the setup bot | full install, Stage 1 included |
| `installing` (recent) | 409 in-progress | 409 in-progress |
| `installing` (stale) / `requested` | full install | full install |

A failed install stays recoverable by a newer release deliberately: `PackageInstallation` is `readonly` for project admins, so refusing every version mismatch would wedge a project with no way to clear the record.

Moving between versions otherwise belongs to `$upgrade` (Ticket 4b), which also has to run declared migrations and rewrite what Stage 1 already wrote.

**Install Bundles are no longer entry-capped.** Install Bundles are written as conditional upserts so an install can be re-run, which forced the whole transaction to be serializable and capped it at `maxSerializableTransactionEntries`. Rather than raise the cap — which just defers the problem and conflates isolation with blast-radius control — `$install` now pre-resolves the conditional queries itself before applying the Bundle. The transaction becomes an ordinary one, and the cap stays where it is.

One consequence worth reviewing: an entry that matches nothing is applied as a plain `POST`, so a package cannot rely on a resource keeping an id it chose. That is already true for the installing project admin, who lacks the create-by-id privilege.

**Catalog visibility for project admins.** `Package` and `PackageRelease` are `projectAdminResourceTypes`, and `getPermittedProjectIds` scoped all admin resource types to the caller's own project. A published catalog lives in a separate publisher project, so a customer-project admin could not see it — `GET /Package` returned an empty searchset even with admin rights. This introduces `catalogResourceTypes` and exempts those two types from current-project-only scoping so they flow through the existing `Project.link` + `exportedResourceType` logic. `PackageInstallation` stays scoped to the owning project, since installations are per-customer state.

Note this makes `Package`/`PackageRelease` behave like ordinary link-exported resources, so any pre-existing link that does not restrict `exportedResourceType` now exposes them too — consistent with the link model, and intended.

## Reviewing

**Please review the access-control change on its own merits.** It was originally split into #10121 to get an independent review, but that PR was opened against this branch rather than `main` and merged here, so it never received one. Rather than re-split and churn the history again, it is folded back into this PR's scope — but it is the riskiest change in the diff and the one most worth a careful look. The relevant surface is:

* `packages/core/src/access.ts` — the new `catalogResourceTypes` list
* `packages/server/src/fhir/repo.ts` — the `getPermittedProjectIds` exemption
* `packages/server/src/fhir/repo.test.ts` — cross-link read with and without export, plus `PackageInstallation` scoping

Note that `$install` still uses the caller-scoped `PackageRelease` read as its authorization gate; only the install Bundle `Binary` is read via `systemRepo`, because `Binary` is never cross-project exportable.

One implicit contract worth knowing while reading `parseSettings`: install `Parameters` are keyed by `parameter.name` and validated against `Questionnaire.item.linkId`, so the two have to match. Documented in place.

## Companion change

Publishing hooks as impl bots is a contract change, so the manifest side moves with it in medplum-ee#954: `hooks.postInstall` now names a declared impl bot instead of a source path, and the install Bundle no longer carries a copy of the hook. The two need to land together.

While verifying this, one previously-logged open risk turned out to be already solved in core and is now documented rather than tracked: `getBotSecrets` layers the bot's project secrets under the run-as project's, so an impl bot invoked with `runAsUser: true` sees shared vendor defaults overridden by the customer's own values.

## Scope note

This PR no longer carries the `updateResourceImpl` null-meta guard. That was split into #10119 and **closed without merging** — review established that no supported write path produces a resource stored without `meta`, so the guard would have silently repaired corrupt data and hidden the integrity problem instead of surfacing it. Nothing here depends on it.

## Test plan

* `packageinstall.test.ts` — 17/17 passing
* `repo.test.ts` — 98/98 passing
* The three tests added for the review findings were each checked to **fail** with the corresponding source change reverted, so they are regression guards rather than tests that happen to pass: a different version over an installed package is refused and applies nothing, a failed install is still recoverable by a newer release with Stage 1 re-run, and a setup bot published without `runAsUser` is refused before it executes
* Two earlier tests cover the setup-bot relocation: a release naming a setup bot without an impl project is rejected before the bot runs, and a same-named bot in the calling project cannot stand in for the published one
* The uncapped-Bundle test first asserts that the same 20-entry Bundle submitted directly is rejected with "too many entries", so it fails if the fix regresses
* That test enables the `transaction-bundles` project feature, without which the Bundle degrades to batch semantics and the cap never applies
* Verified locally against Postgres/Redis: `packages/server` `npm run build` exits 0, full-package `tsc --noEmit` is clean, and `prettier --check` / `eslint` pass on every changed file

---

## Update: making a bot reachable across a project link safe to execute

Three further commits, added after building the linked-exposure model in medplum-ee#955 and running it end to end against a local server. Each is independent and reviewable on its own.

**`fix(server): make a bot reachable across a project link safe to execute`.** A `runAsUser: true` bot resolves to the caller's membership wherever its row lives, and `Bot` is not a project-admin resource type, so a caller-scoped identifier search already crosses `Project.link`. Marketplace packages now rely on that, and two things about it were wrong:

* The `bots` feature was checked only against the project that *owns* the bot, so a project with the feature disabled could execute a linked project's bots — the feature was evaluated against the publisher's project rather than the caller's. `executeBot` now calls `isBotEnabledForProject` on the run-as project (the installing admin's project for the setup bot), not the impl project that owns the bot row.
* `getBotForRequest` used `searchOne` with no sort rules. That search spans every linked project exporting `Bot`, so an identifier matching in two of them resolved to whichever row the database returned first — and executed it. It now fetches two and fails with a diagnosable error, mirroring the `Conditional PUT matched multiple resources` guard already in `packageinstall.ts`.

**`feat(server): let a custom operation stream its response`.** `tryCustomOperation` passed no response stream and wrapped the return value in `buildOutputParameters`, so a bot behind a custom operation could never stream even with `streamingEnabled` set and `Accept: text/event-stream` requested. This one is separable — nothing else here depends on it — if you would rather it went in on its own.

**`fix(server): make an install auditable and its setup-bot outcome meaningful`.** `$install` wrote no `AuditEvent` at all, while publish has been audited from the start. That gap matters because the operation mutates `PackageInstallation` in place and `clearErrorState` clears the error extensions on success — so an install that failed twice and succeeded on the third attempt reads afterwards exactly like one that installed cleanly. Every attempt now records who installed which release, the phase reached, whether the install bundle was applied or resumed, what it commissioned, and the config hash. A failed audit write is logged rather than failing an install that already happened.

The same commit fixes two setup-bot problems that only appeared once a real package was installed:

* **An informational outcome failed the install.** Reporting through an `OperationOutcome` is the setup-bot contract — it is how one-shot credentials come back — but `normalizeBotExecutionResult` marks any bot returning a non-OK outcome as failed, and an outcome is only OK if it says so. A hook that merely described what it seeded therefore failed the install *after* doing the work. In practice this meant the reference Plugin could not install successfully on any server. Only `error` and `fatal` issues are now failures.
* **The reason was discarded.** A hook reporting that way logs nothing, so `logResult || 'Setup bot execution failed'` threw away the only description of the problem. The real cause in the first local run was a missing project feature, invisible until the outcome was read by hand. The outcome's error text is now the message.

### Test plan for this update

* `packageinstall.test.ts` 25/25; `execute` / `custom` / `deploy` / `botinit` / `bots` suites green — 94 tests across the touched files
* The three audit tests were confirmed to **fail** with the audit write disabled (`expected [] to have a length of 1 / 1 / 2`), so they are regression guards
* Setup-bot outcome semantics are covered in all three directions: informational-only succeeds, warning-only succeeds, and an `error` issue fails *with its text as the message*
* A streaming bot in a linked project, executed by identifier from a customer project, returns `text/event-stream` — the combination the marketplace actually needs, and one neither the linked tests nor the existing streaming tests covered alone
* `tsc --noEmit` and `eslint` clean on every changed file
* Verified live against a localhost server on this branch: publish, install as a project admin, then execute across the link. The customer project received **zero** bots, `Project.link` gained the impl project, and both a streaming and a non-streaming linked bot ran — the streaming one returning real SSE tokens

### Note on the rebase

These commits sit on top of `aa55e5082 [autofix.ci] apply automated fixes`, which had corrected the same `jsdoc/no-multi-asterisks` error I had fixed locally. Autofix left a stray trailing asterisk (`requires*`); the rebase resolves in favour of removing both.

---

## Update: `package-secret`

One more commit, closing the last gap the local end-to-end left open.

**The extension was documented in three places and implemented in none.** Core had never heard of the URL — it lives in the manifest tooling, and `$install` reads the bundled `Questionnaire` only to check that required items have values. So a package that flagged an answer got nothing, and the failure stayed silent until a bot tried to read the secret. Every install so far had the value set by hand afterwards.

Core owns the extension now, for the same reason it owns `packageRelease-setup-bot` and `packageRelease-impl-project`: the install operation is what acts on it, and a convention only the publisher knows is one the server cannot honor.

Details worth a reviewer's attention:

* **Upserted by name**, so rotating a value on a re-install replaces it instead of leaving two entries under the same name with the stale one first.
* **Runs before the setup bot**, so a hook can verify the state it will run against, and **outside the resume skip**, so a reconfigure takes effect on a re-invoke that skips the committed bundle.
* **Uses `systemRepo`**, even though a project admin may write `Project.secret` themselves — `secret` is not among the policy's `readonlyFields`. The reason is the neighbouring `hiddenFields` on the same entry: `superAdmin`, `systemSecret`, and `strictMode` are hidden from a project admin, so a read-modify-write through the caller's repo risks writing the project back without them. Authorization remains the operation's existing `superAdmin || membership.admin` gate; this is only about not truncating the resource. `linkImplProject` already does it this way for the same class of reason.

### Two things deliberately not done

**Names are not namespaced per package.** `Project.secret` is one flat namespace shared with the platform — `$ai` reads `OPENAI_API_KEY` from it — so prefixing would break the very secret the package exists to supply. Collisions are real, silent, and last-write-wins. That is a review concern, and it is documented as one rather than mangled here.

**Non-secret settings stay unpersisted.** `PackageInstallation` has no `settings` element, so the docs promising one were describing a field that does not exist; settings were already being parsed, validated, hashed, handed to the setup bot, and discarded. Adding the element is a resource-type change and nothing needs it until `$upgrade` does. The docs are corrected in medplum-ee#955. One upside of the gap: no secret was ever persisted anywhere it should not have been.

### Test plan

* `packageinstall.test.ts` 28/28; 97 tests across the touched suites
* Three new tests: flagged answers are written and unflagged ones excluded — including a secret nested inside a group, which proves the walk recurses; a rotated value replaces rather than duplicating; and the audit record carries `secretsWritten` as a count with neither name nor value in it
* Verified live on a local install: `secretsWritten=1` in the audit record, and the reference Plugin's streaming bot works end to end with **no manual step** — 15/15 harness checks, up from 14/15


packages/server/src/fhir/operations/packageinstall.ts
```diff
@@ -31,44 +115,351 @@ export async function packageInstallHandler(
...
+  installation: WithId<PackageInstallation>,
+  settings: InstallSettings
+): Promise<Resource | undefined> {
+  const setupBotIdentifier = getExtensionValue(packageRelease, PackageReleaseSetupBotUrl);
+  if (!setupBotIdentifier || typeof setupBotIdentifier !== 'string') {
+    return undefined;
   }
+
+  // The setupBot ships inside the Stage 1 Bundle, so it lives in the calling project.
+  const userBot = await ctx.repo.searchOne<Bot>({
```
finnbergquist: Why wouldn't this bot live in the shared project?

finnbergquist:
To make sure I understand this correctly:

- We have a new shared project that everyone gets with PackageRelease resources (representing marketplace of packages that can be installed)
- You can call `POST /PackageRelease/:id/$install` which adds resources to customer project **and**  links resources from the shared project


One thing I am a little confused on is how the versioning of shared projects will work. Will there be a different shared project per version?

packages/server/src/fhir/operations/packageinstall.ts
oleg-mp: It technically can. Good catch. In fact it'll simplify things and make it more consistent. Gonna update the PR

techdavidy:
@claude review

techdavidy:
@oleg-mp 
would it be ok to have any user be able to view packages but only admin to install them ?

oleg-mp:
> @oleg-mp would it be ok to have any user be able to view packages but only admin to install them ?

you mean to view in the catalog? good question. I'd say probably maybe yes, but depends on the business rules of specific practice. I don't see any obvious security issues with it, more like a who is allowed to see what. i.e. should a doctor be able to see this far down the technical stack

techdavidy:
mind taking a look at these issue from claude `Version upgrades are silently treated as no-ops` ? @oleg-mp

techdavidy:
@claude any chance to be verbose about the name of the stages rather than stage 1 and stage2 ?

oleg-mp:
> mind taking a look at these issue from claude `Version upgrades are silently treated as no-ops` ? @oleg-mp

already submitted the fix

oleg-mp:
@claude re-review this refactored version

packages/server/src/bots/utils.ts
```diff
@@ -146,12 +146,33 @@ export function normalizeBotExecutionResult(result: BotExecutionResult): BotExec
...
+ * projects must have the feature. Checking only the bot's project would let a
+ * project that has bots disabled execute a linked project's bots, since the
+ * feature would be evaluated against the publisher's project rather than the
+ * caller's.
  * @param bot - The bot resource.
+ * @param runAs - The membership the bot will execute as, when known.
  * @returns True if the bot is enabled.
  */
-export async function isBotEnabled(bot: Bot): Promise<boolean> {
+export async function isBotEnabled(bot: Bot, runAs?: ProjectMembership): Promise<boolean> {
```
codyebberson, line 160: Let's discuss offline.

codyebberson (commented):
@claude - please review

I believe the meat of this PR is in `packageinstall.ts` - it seems like there is a significant amount of duplicate code with the other existing bot execute flows.  Can you please confirm?

codyebberson:
Question for @oleg-mp and maybe @claude  - why does the `$install` operation need to invoke `executeBot` at all?  The original design was that `$install` would simply execute a `Bundle` batch operation, which could itself include calls to `Bot/$deploy` and `Bot/$install`.  Perhaps a little more work for the package author, but it would presumably keep the `$install` operation flat and simple?

oleg-mp:
@claude review the refactored version of this PR

Issue #7006, opened 2025-07-09
Ability to replay Agent messages


maddyli:
Feature to be able to replay messages at an agent. 

Would be an Agent operation, where a user could specify a timeframe and all messages that were received within that timeframe are replayed back the Agent.

PR #9491, opened 2026-06-12
feat(agent): auto-retry of messages already committed but failed

### LINKED ISSUE #9492: [Agent] Configurable auto-retry of committed messages
Built on the back of #7578, we should add configuration options to automatically retry messages that failed due to enumerated classes of errors, for example, if a message was send but never received an ACK, or if a message failed to even send due to a connection outage.

ThatOneBro:
Fixes #9492 

Since #7006 requires a server change, having a configurable auto-replay of different classes of failed messages as a first step addresses concerns with guaranteed delivery without the need for a server upgrade.

Related to #7006
Dependent on #9346


packages/agent/src/queue/schema.ts
```diff
@@ -46,17 +46,27 @@ export const MIGRATIONS: readonly Migration[] = [
...
         attempt_count         INTEGER NOT NULL DEFAULT 0,
         callback_id           TEXT    NOT NULL,
         server_response_body  BLOB,
         server_status_code    INTEGER,
         ack_outcome           TEXT    NOT NULL DEFAULT 'pending',
         last_error            TEXT,
+        -- error_code is the machine-readable failure classification (QueueErrorCode);
+        -- the retry policy gates on it, never on the free-form last_error string.
         error_code            TEXT,
+        -- next_attempt_at is the earliest time (ms) a retry-scheduled 'queued' row
```
ThatOneBro: Since we've cut v5.1.22, we can no longer merge this into one schema, we need a second migration 👀

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
```diff
@@ -286,29 +290,56 @@ since the durable write that triggers CA happens synchronously at intake and wou
...
+
+> **The default is guaranteed-delivery mode** (the right column below). This is a deliberate, historical choice: the Medplum server's agent WebSocket transmit handler (`packages/server/src/ws/agent.ts`) collapses **every** Bot-execution failure to HTTP `400` and does not disambiguate a permanent rejection from a transient/ephemeral one. With no reliable transient-vs-permanent signal from the server, the only safe default is to assume any error might be ephemeral and keep retrying — otherwise a genuinely retryable failure would be silently dropped, defeating the durable queue. The cost is possible duplicate delivery. Operators who need exactly-once should either opt out per channel (`guaranteedDelivery=false`, the left column) or **dedupe in their Bot** — e.g. record processed message control IDs (MSH-10) on a FHIR resource such as `MessageHeader` and skip a control ID that has already been handled. Once the server differentiates status codes (5xx/429 for transient, 4xx for permanent — the agent already classifies these, see `classifyStatusCode`), normal mode becomes a safe default again.
```
mattlong: > Operators who need exactly-once should either opt out per channel (`guaranteedDelivery=false`, the left column) or **dedupe in their Bot**

Normal mode isn't more aligned with exactly-once, is it? Perhaps this should be reworded to "Operators who cannot tolerate duplicate deliveries should either..."

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
```diff
@@ -421,20 +452,34 @@ We document this behavior change clearly in the channel URL `enhanced=aa` docs.
...
+| `queueSweepIntervalSecs`            | `valueInteger` | `3600`                                                          | How often the retention sweeper runs.                                                                                                                                                                                                                            |
+| `channelAutoRetry`                  | `valueBoolean` | `true`                                                          | Agent-wide default for auto-retry of retryable dispatch failures (§4.1). On by default; set `false` to opt out.                                                                                                                                                  |
+| `channelGuaranteedDelivery`         | `valueBoolean` | `true`                                                          | Agent-wide default for guaranteed-delivery mode (§4.1). On by default while the server returns 400 for every failure (see §4.1); set `false` to opt into normal mode and dedupe in your Bot. Requires auto-retry.                                                |
```
mattlong: Instead of `channelAutoRetry` and `channelGuaranteedDelivery`, could there be a single `channelRetryMode`  that can be `none | normal | guaranteed`?

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
```diff
@@ -421,20 +452,34 @@ We document this behavior change clearly in the channel URL `enhanced=aa` docs.
...
+| `channelAutoRetryBaseDelayMs`       | `valueInteger` | `1000`                                                          | Delay before the first retry.                                                                                                                                                                                                                                    |
```
mattlong: Maybe I'm missing something, but it seems like the agent-wide settings should not be prefixed with `channel` and the per-channel settings in the next table should have the prefix?

packages/agent/src/hl7.ts
```diff
@@ -915,6 +947,143 @@ export function parseDuplicateBehavior(rawValue: string | undefined, logger: ILo
...
+    return undefined;
+  }
+  const normalized = rawValue.toLowerCase();
+  if (normalized === 'true') {
+    return true;
+  }
+  if (normalized === 'false') {
+    return false;
+  }
+  logger.warn(`Invalid ${name} value '${rawValue}'; expected 'true' or 'false'. Ignoring.`);
```
mattlong: It seems like any unrecognized or invalid config params should throw and shut down the process rather than be ignored.

packages/agent/src/queue/schema.ts
```diff
@@ -124,6 +127,25 @@ export const MIGRATIONS: readonly Migration[] = [
       ) STRICT;
     `,
   },
+  {
+    // Auto-retry support for the Bot leg. Adds two columns to the existing
+    // inbound_hl7_messages table; SQLite ALTER TABLE ... ADD COLUMN is a
+    // metadata-only operation, so this is cheap even on a populated DB.
```
mattlong: Even specifying a default value is metadata only?

packages/agent/src/queue/worker.ts
```diff
@@ -480,6 +692,25 @@ function classifyStatusCode(statusCode: number): QueueErrorCode {
...
 
+/**
+ * Extracts the MSA-1 acknowledgment code from an HL7 ACK body, if there is one.
+ * Used in guaranteed-delivery mode, where the upstream ACK code — not the HTTP
+ * status — decides whether the message is settled. Anything unparseable returns
+ * undefined and the caller falls back to status-code handling.
+ * @param body - The server response body (expected to be an HL7 ACK message).
+ * @returns The uppercased MSA-1 value (e.g. 'AA', 'AE', 'AR', 'CA', 'CE', 'CR'), or undefined.
+ */
+function parseAckCode(body: string | undefined): string | undefined {
```
mattlong: Would it be reasonable to have this return `AA | AE | AR ...` instead of just string? That way you could use switch statements to help ensure exhaustive covering of values elsewhere.

packages/agent/src/queue/worker.ts
```diff
@@ -365,25 +489,52 @@ export class ChannelQueueWorker {
...
+      this.handleFailure(row, classifyStatusCode(statusCode), `Server returned ${statusCode}: ${response.body ?? ''}`);
       return;
     }
 
-    // The Bot accepted the message (2xx): the row is `processed`. Delivering the
-    // ACK back to the source is a SEPARATE leg recorded in ack_outcome — a closed
-    // source connection yields `processed` + `undelivered`, never a Bot-leg
-    // failure, and is therefore never re-dispatched. The source recovers it by
-    // retransmitting, which replays the stored ACK (handleDuplicate, hl7.ts).
+    // The Bot accepted the message (2xx, or guaranteed-mode AA/CA): the row is
```
mattlong, line 700: It seems like some sort of validation or check that response is in the expected state after the previous early exits since there is a decently sized matrix of possibilities between `statusCode` and `ackCode` which is a lot to keep in your head. It'd be especially nice if we could lean on the types via a  `satisfies` or something here.

mattlong (changes requested):

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: The thought was to leave the client side open to auto-retry in the future.

Channels are just one direction of retries

packages/agent/src/hl7.ts
ThatOneBro: The convention so far is to ignore unrecognized parameters, since the most common flow for updating agent config is edit resource -> reload config while the agent is already running. Crashing the program because of an unrecognized new configuration rather than just continuing with the old value does not usually make sense, because it puts the agent in a crash loop until the Agent resource is repaired, which may take several minutes to detect, causing avoidable downtime.

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: That seems sensible

packages/agent/src/queue/schema.ts
ThatOneBro: Good question, I think this was written before that default was added... going to take a look. Regardless I think this is still a safe migration since the migration will happen before the agent starts responding to messages... I should actually validate that though (on startup, we apply migration before attempting to commit incoming messages)

packages/agent/src/hl7.ts
ThatOneBro: After thinking about this more, I think you are correct that we shouldn't try to apply a partial config update... We should validate the config as a separate step probably, and exit before applying a partially valid config... and only apply updates atomically

ThatOneBro: I think we should do that in a follow-up PR though as the current model already works in a similar way:
https://github.com/medplum/medplum/issues/9721

packages/agent/src/queue/worker.ts
ThatOneBro: 💯, done

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: Good point, I've fixed those references 👍

packages/agent/DURABLE_QUEUE_ARCHITECTURE.md
ThatOneBro: Done 🙏

packages/agent/src/queue/schema.ts
ThatOneBro: It looks like it is metadata only unless there is a CHECK constraint or if there is a generated COLUMN with a not null constraint. I think it will just read the default from the metadata in the cases of rows that have not been populated

ThatOneBro: See: https://www.sqlite.org/lang_altertable.html#altertableaddcolumn

packages/agent/src/queue/worker.ts
ThatOneBro: I've added some better type hints via making `InboundRow` a discriminated union and adding asserts in a few places to assert the state of a row before continuing...

Also split out `handleProcessed` here which makes it more readable I think

ThatOneBro:
Multi-angle review from Claude, done locally:

---

## Review findings: auto-retry of committed-but-failed messages

**Overview:** The core retry state machine is solid and well documented. Every issue found lives in the *interactions* between the new retry loop and pre-existing paths that assumed a row settles exactly once (duplicate replay, late responses, config reload, crash recovery). Process: 8 finder angles → 37 raw candidates → 17 distinct mechanisms adversarially verified → 12 confirmed, 4 plausible, 3 refuted. Top 8, most severe first.

### 1. Default guaranteed mode swallows bot-authored NACKs — CONFIRMED, major
[`packages/agent/src/queue/worker.ts:552`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/queue/worker.ts#L552)

A bot that deliberately returns an AR/AE ACK runs successfully, so the server replies HTTP **200** with that ACK as body (`ws/agent.ts`: `statusCode: result.success ? 200 : 400`). On main this marked the row processed and relayed the ACK to the source. Now, with zero config changes: AR/CR → `markRejected` + `ack_outcome='not_owed'` — **nothing is sent to the source**, whose MLLP client hangs until timeout (it only recovers the AR if it retransmits under idempotent dedup; a standard-enhanced source that already got its CA never retransmits). AE/CE → `UpstreamError`, retried indefinitely — the bot re-executes every ≤60s forever while the channel head is blocked. §4.1 and the tests cover the HTTP-400+ACK case but never the HTTP-200 bot-returns-NACK pattern, and no test asserts the `sendAck` suppression.

### 2. Agent-wide retry-setting changes never reach running channels — CONFIRMED, major
[`packages/agent/src/hl7.ts:257`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/hl7.ts#L257)

`resolveRetryPolicy` + `worker.setRetryPolicy` only run inside `configureHl7ServerAndConnections`, which `reloadConfig` skips when the endpoint address is unchanged. Main's gate was correct because everything it configured came from the URL; this PR newly injects `getChannelRetrySettings()` into that gated function. An operator who sets `channelRetryMode='none'` mid-incident gets a successful reload while every running worker keeps the old guaranteed policy (and keeps stamping `guaranteedDelivery=true` onto new rows) until process restart. The PR's own comment ("push the new policy at it if it's already running") shows propagation was intended.

### 3. Stale failed-attempt ACK replayed to a retransmitting source — CONFIRMED, major
[`packages/agent/src/hl7.ts:672`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/hl7.ts#L672)

`SCHEDULE_RETRY` clears `sent_at`/`processing_started_at` but **not** `server_response_body`/`server_status_code`, creating the first long-lived non-terminal state that carries a response body — and `handleDuplicate`'s replay gate checks only that the body is non-empty, never the row state. Routine trigger: attempt 1 gets an AE, row requeues for retry, the source (still owed its only ACK) retransmits during the backoff and receives the stale AE as final verdict. It may resubmit under a new control ID (bypassing MSH-10 dedup → true upstream duplicate), then the successful retry sends a contradictory AA. Fix candidates: clear the response columns in `SCHEDULE_RETRY`, or don't replay for non-settled rows.

### 4. Late response from attempt N settles attempt N+1 — CONFIRMED, major
[`packages/agent/src/queue/worker.ts:356`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/queue/worker.ts#L356)

`callback_id` is minted once per row and never rotated across retries, and `onServerResponse` matches the pending dispatch purely by callback. Attempt 1 times out at 60s; guaranteed mode redispatches ~1s later; attempt 1's late response then resolves attempt 2's pending promise — a late AR/CR can terminally reject a row whose in-flight attempt succeeded, and a late 500 schedules attempt 3 even if attempt 2 succeeded (cascading, since attempt 3 shares the id too). Unreachable on main, where no row was ever re-dispatched while a response could still arrive. A per-attempt dispatch token (or rotating the callback on claim) would close it.

### 5. Poison-message wedge is the shipped default — PLAUSIBLE (documented trade-off to adjudicate), major
[`packages/agent/src/queue/worker.ts:694`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/queue/worker.ts#L694)

Bot throws deterministically on one message → server collapses it to 400 with a non-HL7 body → `ServerRejected` is not in `GUARANTEED_TERMINAL_CODES` → retried forever, and the head-of-line predicate stalls the entire channel. Meanwhile intake keeps CA-ACKing new messages, and all three retention DELETEs target only terminal states, so the `queued` backlog is exempt from the size cap and the DB grows without bound. §4.1 documents this as "the contract," but on main a 400 was terminal and the channel kept flowing — a zero-config upgrade flips that. Worth adjudicating the default, or adding a guardrail (backlog-depth alarm, terminal-after-N-days, or intake backpressure).

### 6. `RECOVER_INFLIGHT`'s COALESCE preserves a stale retryable code — CONFIRMED, minor
[`packages/agent/src/queue/queries.ts:249`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/queue/queries.ts#L249)

`CLAIM_NEXT` clears only `next_attempt_at`, so a row that retried after a 503 carries `error_code='server-error'` into its next attempt; if the agent crashes mid-flight, recovery COALESCEs and keeps that code instead of writing `interrupted`. The row is ambiguous (may have been processed) yet labeled with the one code documented as "provably never accepted, safe to re-dispatch" — exactly the triage signal normal mode exists for. On main the COALESCE was vacuously safe (no live row could carry a code); `RECOVER_INFLIGHT_GUARANTEED` overwrites unconditionally, showing the intended semantics.

### 7. Late-response disambiguation is effectively dead in the default mode — CONFIRMED, minor
[`packages/agent/src/queue/worker.ts:392`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/queue/worker.ts#L392)

The late-settle path only fires for rows in `failed`, but under guaranteed mode a timed-out row is immediately back in `queued`, where late responses are discarded ("outcome is owned elsewhere" — untrue for a backoff-parked row with no other owner). A late authoritative AA is thrown away and the retry re-executes the bot; the doc comment ("avoids a redundant re-dispatch") is now stale for the default mode. Settling `queued` rows with a pending `next_attempt_at` is tractable since everything runs on one event loop.

### 8. `assertRowState(QUEUED)` after enqueue can false-NACK a committed message — PLAUSIBLE, minor
[`packages/agent/src/queue/durable-queue.ts:564`](https://github.com/medplum/medplum/blob/519fd3e68270d59bf19da199773b32d96bdcc2df/packages/agent/src/queue/durable-queue.ts#L564)

The re-read runs *after* the insert commits, so during zero-downtime-upgrade overlap the peer leader's `claimNext` can flip the row to `claimed` in the window before `getById` — the assert throws, and `handleMessageDurable`'s catch sends the source a storage-error NACK for a message that is durably committed and actively dispatching. Main returned the row regardless of state. Narrow window, but the fix is just tolerating non-`nacked` states here.

### Also noted (below the severity cap)

- `worker.ts:103` — `RetryPolicy`'s two derived booleans re-admit the "guaranteed but not retrying" hybrid the docs call unrepresentable — and **worker.test.ts already constructs it at 7 sites** via `{ ...DEFAULT_RETRY_POLICY, enabled: false }`; those tests exercise the guaranteed ACK-parsing branch and pass only because their bodies aren't parseable HL7. Carrying `mode: RetryMode` instead is a drop-in at all five use sites.
- `queries.ts:270` — crash recovery follows the intake-time `guaranteed_delivery` snapshot, not current policy, so a retryMode flip doesn't govern the (at most one per channel) inflight row recovered after a crash.
- `app.ts:803` — `parseChannelRetryModeSetting` duplicates `parseRetryModeParam`; and agent settings are silently clamped where URL params warn-and-ignore (`hl7.ts:1029`).

### Verified-and-refuted (for the record)

Triple-HL7-parse efficiency concern (core's parse is lazy; only +1 cheap parse vs main); "ACK (unknown)" logging regression (all `Hl7EnhancedAckSentEvent` producers build ACKs from typed `AckCode`s); AA-body-over-HTTP-400 trust (server 400 bodies are never parseable HL7; ACK-over-status precedence is documented and tested).

**Common thread for findings 1–7:** `applyServerResponse`, `handleDuplicate`, `onServerResponse`, and crash recovery all predate retries and assume a row settles exactly once. A per-attempt identity (rotating callback/dispatch token) plus clearing response columns on `SCHEDULE_RETRY` would resolve findings 3, 4, and 7 at the root.

packages/agent/src/queue/schema.ts
```diff
@@ -124,6 +127,32 @@ export const MIGRATIONS: readonly Migration[] = [
...
+    // even on a populated DB *for these two columns*: it only rewrites the
+    // schema text and leaves existing rows untouched. That's NOT universally
+    // true -- adding a column with a CHECK constraint, or a generated column
+    // with NOT NULL, forces a full-table read/rewrite proportional to row count.
+    // Neither column here does that (next_attempt_at is nullable; the
+    // guaranteed_delivery NOT NULL sits on a plain, non-generated column with a
+    // DEFAULT), so both stay cheap. See
+    // https://www.sqlite.org/lang_altertable.html#altertableaddcolumn.
+    version: 2,
+    sql: `
```
mattlong, line 142: Curious what your process for testing these migrations are. I suspect it's somewhat manual which is totally fine, just wondering if it's documented incase another developer needs to develop on it

mattlong (approved):

packages/agent/src/queue/schema.ts
ThatOneBro: There is actually a test for migrations here, although it's kind of hidden in the middle of `durable-queue.test.ts`

https://github.com/medplum/medplum/pull/9491/changes#diff-775bb35ff0cd6a983f1c4a56394033864ffea13c91e54263308a330ad2b4c972R755

PR #9494, opened 2026-06-12
fix(ws): tear down WebSocket subscription Redis listener on shutdown

The-Alchemist:
## Summary
- Add `closeWebSocketSubscriptionHandler()` to disconnect and clear the module-level Redis pub/sub subscriber used for WebSocket subscription fan-out.
- Call it from `closeWebSockets()` during graceful shutdown, alongside existing FHIRcast and echo heartbeat cleanup.
- Recreate the subscriber on bind when the prior connection has reached ioredis `'end'` status, not only when the reference is undefined.

## Motivation
After shutdown, `redisSubscriber` could remain truthy while pointing at a dead connection. That blocked `setupSubscriptionHandler()` from running on the next bind and left stale pub/sub state across server restarts.

## Test plan
- [ ] Bind a WebSocket subscription client, shut down the server, restart, and verify notifications still arrive on rebind
- [ ] `npm test -w @medplum/server -- src/ws/subscriptions.test.ts`

### DIFF
--- a/packages/server/src/ws/subscriptions.ts
+++ b/packages/server/src/ws/subscriptions.ts
@@ -75,8 +75,14 @@ let subscriptionMessagesSent = 0;
 let subscriptionMessagesReceived = 0;
 
 async function setupSubscriptionHandler(): Promise<void> {
-  redisSubscriber = getPubSubRedisSubscriber();
-  redisSubscriber.on('message', async (channel: string, events: string) => {
+  const subscriber = getPubSubRedisSubscriber();
+  redisSubscriber = subscriber;
+  subscriber.on('end', () => {
+    if (redisSubscriber === subscriber) {
+      redisSubscriber = undefined;
+    }
+  });
+  subscriber.on('message', async (channel: string, events: string) => {
     globalLogger.debug('[WS] redis subscription events', { channel, events });
     const subEventPayload = JSON.parse(events) as V1SubEventPayload | V2SubEventPayload;
     let resource: WithId<Resource>;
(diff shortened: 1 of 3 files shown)


ThatOneBro (approved):
Nice, good fix 👍

packages/server/src/ws/subscriptions.ts
```diff
@@ -75,8 +75,14 @@ let subscriptionMessagesSent = 0;
 let subscriptionMessagesReceived = 0;
 
 async function setupSubscriptionHandler(): Promise<void> {
-  redisSubscriber = getPubSubRedisSubscriber();
-  redisSubscriber.on('message', async (channel: string, events: string) => {
+  const subscriber = getPubSubRedisSubscriber();
+  redisSubscriber = subscriber;
+  subscriber.on('end', () => {
```
ThatOneBro, line 80: I guess one thing that would be good to add for a sanity check is a test checking that calling `.end` on a subscriber triggers `redisSubscriber` to be `undefined` and binding a new WebSocket subscription socket actually retriggers `setupSubscriptionHandler` after that

The-Alchemist: done.   see https://github.com/medplum/medplum/pull/9494/changes#diff-af8dcb88f7c517f956883bc957531c4dea9b58d29d47ff341a450cb67e506727R2034

ThatOneBro (approved):

ThatOneBro (approved):
Refactor LGTM, thanks @The-Alchemist

PR #9498, opened 2026-06-12
Ensure that Coding properties are attached to primary row

### LINKED ISSUE #9489: `CodeSystem/$import` attaches properties to designation synonym rows, silently dropping codes from filter-based `$expand`/`$validate-code`
## Summary

When a code in a CodeSystem has designations (stored internally as synonym rows in the `Coding` table), a subsequent `CodeSystem/$import` call that imports **properties** for that code can attach the property rows to a **synonym row** instead of the primary coding row. The result is silent and order-dependent:

- The code disappears from any `ValueSet/$expand` and `ValueSet/$validate-code` whose compose uses a property filter (`=` / `in`), because the property `EXISTS` join matches the synonym row while the primary row (the one returned by expansion queries with the default `language IS NULL` filter) has no property row.
- `CodeSystem/$lookup` stops returning the property, because lookup emits rows with `synonymOf` set as designations only and never reads their properties.

There is no error or warning — membership and semantics just quietly degrade for the affected codes.

Observed on self-hosted Medplum `5.1.11`; the relevant code is unchanged on `main` as of 2026-06-11. Related but distinct: #8243 (closed by #8287, which fixed designations without language codes — a different problem).

## Root cause

`packages/server/src/fhir/operations/utils/terminology.ts` — `selectCoding()` selects by `(system, code)` and returns **both** primary and synonym rows (it selects the `synonymOf` column but never filters on it):

```ts
export function selectCoding(systemId: string, ...code: string[]): SelectQuery {
  return new SelectQuery('Coding')
    .column('id')
    .column('code')
    .column('display')
    .column('synonymOf')
    .column('language')
    .where('system', '=', systemId)
    .where('code', 'IN', code);
}
```

`packages/server/src/fhir/operations/codesystemimport.ts` — `processProperties()` then resolves the row to attach each property to with a first-match `find()` over that result set:

```ts
const sourceCodingId = codingIds.find((r) => r.code === imported.code)?.id;
...
const targetCodingId = codingIds.find((r) => r.code === imported.value)?.id;
```

When primary and synonym rows both exist for a code, which row `find()` returns depends on result-set order, which is unspecified (no `ORDER BY`). In practice, re-importing a concept (the upsert path uses `ON CONFLICT ... DO UPDATE`) rewrites the primary tuple and tends to move it after the synonym row in heap order, making the synonym the first match. The same first-match pattern affects:

- the **target** of `code`-typed properties (`targetCodingId` can resolve to a synonym row, corrupting hierarchy traversal), and
- the **source** of designation imports (a new synonym can chain `synonymOf` onto another synonym row instead of the primary).

## Reproduction

1. Create a CodeSystem with a declared property:

```http
POST /fhir/R4/CodeSystem
```
```json
{
  "resourceType": "CodeSystem",
  "url": "http://example.org/fhir/CodeSystem/repro",
  "status": "active",
  "content": "not-present",
  "property": [{ "code": "category", "type": "string" }]
}
```

2. Import a concept **with a designation**:

```http
POST /fhir/R4/CodeSystem/$import
```
```json
{
  "resourceType": "Parameters",
  "parameter": [
    { "name": "system", "valueUri": "http://example.org/fhir/CodeSystem/repro" },
    { "name": "concept", "valueCoding": { "code": "A", "display": "Alpha" } },
    { "name": "designation", "part": [
      { "name": "code", "valueCode": "A" },
      { "name": "language", "valueCode": "fr" },
      { "name": "value", "valueString": "Alpha (fr)" }
    ]}
  ]
}
```

3. Re-import the concept (simulates a routine idempotent re-run; the upsert rewrites the primary row):

```json
{
  "resourceType": "Parameters",
  "parameter": [
    { "name": "system", "valueUri": "http://example.org/fhir/CodeSystem/repro" },
    { "name": "concept", "valueCoding": { "code": "A", "display": "Alpha" } }
  ]
}
```

4. Import a property for the code in a separate call:

```json
{
  "resourceType": "Parameters",
  "parameter": [
    { "name": "system", "valueUri": "http://example.org/fhir/CodeSystem/repro" },
    { "name": "property", "part": [
      { "name": "code", "valueCode": "A" },
      { "name": "property", "valueCode": "category" },
      { "name": "value", "valueString": "demo" }
    ]}
  ]
}
```

5. Create a ValueSet filtering on the property and expand/validate it:

```json
{
  "resourceType": "ValueSet",
  "url": "http://example.org/fhir/ValueSet/repro",
  "status": "active",
  "compose": { "include": [{
    "system": "http://example.org/fhir/CodeSystem/repro",
    "filter": [{ "property": "category", "op": "=", "value": "demo" }]
  }]}
}
```

## Suggested fix

Restrict property/designation source and target resolution to primary rows. Either:

- add a primary-only variant of `selectCoding` (e.g. `.where('synonymOf', '=', null)`) and use it in `importCodeSystem`'s property and designation resolution paths (note `$lookup` intentionally wants synonym rows, so the existing selector must remain available), or
- keep the query as-is and filter at the call sites: `codingIds.find((r) => r.code === imported.code && isEmpty(r.synonymOf))` (and the same for `targetCodingId` and the designation `sourceCodingId`).

A data repair for already-affected rows (re-pointing `Coding_Property.coding` from synonym rows to their `synonymOf` target) may also be worth considering, since affected deployments have silently wrong expansions today.

mattwiller:
Resolves #9489

### DIFF
--- a/packages/server/src/fhir/operations/codesystemimport.test.ts
+++ b/packages/server/src/fhir/operations/codesystemimport.test.ts
@@ -1,7 +1,8 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import { ContentType, EMPTY } from '@medplum/core';
-import type { CodeSystem } from '@medplum/fhirtypes';
+import type { CodeSystem, ValueSet } from '@medplum/fhirtypes';
+import { randomUUID } from 'crypto';
 import express from 'express';
 import request from 'supertest';
 import { initApp, shutdownApp } from '../../app';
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

packages/server/src/fhir/operations/codesystemimport.test.ts
```diff
@@ -324,6 +325,95 @@ describe('CodeSystem $import', () => {
...
+    const valueSet = resValueSet.body as ValueSet;
+
+    const resExpand = await request(app)
+      .get(`/fhir/R4/ValueSet/$expand?url=${valueSet.url}`)
+      .set('Authorization', 'Bearer ' + accessToken)
+      .set('Content-Type', ContentType.FHIR_JSON)
+      .send();
+    expect(resExpand.status).toStrictEqual(200);
+    const expanded = resExpand.body as ValueSet;
+    console.log(JSON.stringify(expanded.expansion));
```
codyebberson:
```suggestion
```

codyebberson (approved):

PR #9499, opened 2026-06-12
Adding Twilio hook + new provider messaging UI

AndyStoneman:
Twilio SMS integration into medplum-provider chat component. This adds a `useTwilioSms` hook and wires it into `ThreadChat/BaseChat` so providers can switch between sending an in-app message and sending an SMS to the patient from the same chat UI. Includes a channel selector dropdown, a handful of scroll/layout fixes that came up along the way (particularly when refreshing the messaging page), and a testing for the hook and UI changes.

Open to criticism or other approaches for the scroll logic in BaseChat.tsx. It felt overly complex when implementing but I couldn't get the annoying scroll bugs fixed any other way I tried.


techdavidy (commented):
I think we can use the logic to use the hook in ThreadInbox. Expose a message sent in ThreadChat and offload the sms work to ThreadInbox

PR #9493, opened 2026-06-12
WIP: fix(server): disable ioredis ready check on pub/sub subscribers, do not merge yet

The-Alchemist:
## Summary
- Disable ioredis `enableReadyCheck` when duplicating Redis connections for pub/sub subscribers.
- Prevents reconnect failures where ioredis sends `INFO` on a connection already in Redis subscriber mode (where only subscribe/unsubscribe-style commands are allowed).
- Long-lived subscriber clients (WebSocket notifications, FHIRcast, agent messaging) now rely on ioredis `autoResubscribe` for reconnect recovery instead of the default ready check.

## Testing
* No changes.
* This change avoids this flaky error:

```
 FAIL  src/workers/subscription.test.ts > Subscription Worker > WebSocket Subscriptions
ReplyError: ERR Can't execute 'info': only (P|S)SUBSCRIBE / (P|S)UNSUBSCRIBE / PING / QUIT / RESET are allowed in this context
```

## Related upstream issue

https://github.com/redis/ioredis/issues/2099

PR #9501, opened 2026-06-13
WIP: fix(server): await Redis subscriber ready before SUBSCRIBE

The-Alchemist:
## Summary

Fixes a flaky ioredis race where `SUBSCRIBE` can complete before the connection's INFO ready check, leaving the client in subscriber mode and causing:

```
Error: Connection in subscriber mode, only subscriber commands may be used
```

This surfaced in CI on `subscription.test.ts` > **WebSocket Subscriptions** `beforeAll`, especially when `post-deploy-migration.test.ts` runs immediately before it (both call `initAppServices` / `shutdownApp` in lifecycle hooks). Reproduced locally at ~50% failure rate without this fix.

- Add `awaitPubSubRedisSubscriberReady()` in `redis.ts` (waits for `'ready'`, with a post-listener status re-check to close the event race)
- Call it before the first `SUBSCRIBE` in WS subscription handler, agent, echo, FHIRcast, MCP SSE, and agent push utils
- Echo uses a deferred async promise so socket message handlers can still await subscription safely
- Tests: await ready in subscription worker WebSocket suite `beforeAll`; mocked subscribers use `status: 'ready'`

## Test plan

- [x] `eslint` on changed files
- [x] `jest src/workers/subscription.test.ts -t "WebSocket Subscriptions"` (WebSocket suite runs; one unrelated fetch-mock test may fail on main)
- [x] Vitest flake repro on jest-to-vitest branch: `post-deploy-migration` + WebSocket Subscriptions — 10/10 pass with this fix
- [ ] Full CI


ThatOneBro:
It seems like this might be an alternative approach to solving the same thing #9493 is solving?

FWIW, I looked into what happens when you send subscribe before awaiting ready and it seems like this would only be an issue in two cases:
1. offline queue disabled. I think that doesn't apply in our case
2. autoReconnect is on, and `enableReadyCheck` is on (bingo)

ThatOneBro:
The real risk I can think of is that this could introduce cases where we don't immediately subscribe to a channel and miss some messages. I'm generally in favor of https://github.com/medplum/medplum/pull/9493 but in the interim it might be worth disabling `autoReconnect` altogether for pub subscriber clients and do it manually by listening for events? 

So that would be:
* Turn off ready check AND autoReconnect 
* Attempt to manually reconnect whenever we get offline or whatever the event is for that case. Maybe we could even rip the upstream fix from that PR for now and vendor it

I think waiting for the ready check is not relevant for our subscriber clients since they are listen only anyways

PR #9504, opened 2026-06-13
Kubernetes docs fixes

codyebberson:


charts/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: |
   Medplum is an open source healthcare development platform.
   Provides FHIR server, auth, workflows, and APIs for building healthcare applications.
 type: application
-version: 5.1.17
+version: '5.1.17'
```
codyebberson, line 7: This is more cosmetic, just aligning that version strings should be in quotes.

charts/values.yaml
```diff
@@ -58,7 +58,7 @@ deployment:
   replicaCount: 1
   image:
     repository: medplum/medplum-server
-    tag: latest
+    tag: '5.1.17'
```
codyebberson, line 61: This is a more significant change.

Using the `latest` tag as a deployment target in Infrastructure-as-Code tools such as CDK, Terraform, or Helm is generally no longer considered a best practice. Earlier container deployment workflows sometimes relied on mutable tags like latest combined with mechanisms such as "force new deployment" to trigger rollouts. Over time, the industry has largely shifted toward explicit, immutable version references (version numbers, image digests, or release tags) because they provide better reproducibility, auditability, rollback behavior, and change tracking.

codyebberson:
@claude please review

codyebberson:
@claude 

> Pre-existing nit (not introduced by this PR): Line 92 uses third-person: "they can now run the helm install command" while the rest of the doc uses second-person ("you"). Worth fixing in a follow-up for consistency.

Fixed

> The hardcoded 5.1.17 in the upgrade examples is expected in docs (it's illustrative), but it will become stale over time. Consider whether a placeholder like <VERSION> would be better, or if a docs convention for this exists elsewhere.

That value will now be replaced on every versioned release.  See the changes in `scripts/prepare-release.sh`

reshmakh (approved):

PR #9506, opened 2026-06-14
Bulk import should recognize and retry on rate limits

codyebberson:
Updated 2026-08-10:

--------

Fixes #9506.

Medplum CLI bulk imports send resources in FHIR transaction bundles. When a project has real transaction bundles enabled, exceeding a rate limit returns an HTTP 429 response. When transactions are not enabled, the server processes the transaction bundle using batch semantics and returns an HTTP 200 response with 429 errors embedded in individual bundle entries.

The SDK's standard retry handling only sees the outer HTTP response, so it cannot recognize the embedded 429 responses. Its normal retry limit is also unsuitable for a bulk import that may legitimately run for hours.

This change adds retry handling specifically to the CLI bulk import flow:

- Disable the SDK's bounded retries for each bulk import request and handle retries in an outer loop.
- Retry outer HTTP 429 responses until the request succeeds.
- For transaction bundles processed with batch semantics, retry only the entries that returned 429.
- Do not resend entries that already succeeded, which avoids creating duplicate resources.
- Preserve the existing behavior for all non-429 entry failures.

## Rate limit reset metadata

Rate-limited `OperationOutcome` resources now include the reset delay as an extension:

```json
{
  "url": "https://medplum.com/fhir/StructureDefinition/rate-limit-reset",
  "valueUnsignedInt": 60
}
```

The value is the number of seconds until the rate limit resets, matching the `t` value in the `RateLimit` response header. The extension is emitted by both the HTTP request limiter and the FHIR interaction quota limiter.

The existing JSON metadata in `OperationOutcome.issue.diagnostics` remains unchanged. The client-side helper also reads `diagnostics.msBeforeNext` as a fallback for compatibility with servers that do not yet emit the extension.

If neither form of reset metadata is available, bulk import uses the most recent `RateLimit` header known to `MedplumClient`, with a 500 ms minimum fallback.

--------

Original PR:

--------

See: https://hl7.org/fhir/R4/http.html#transaction-response

> For a failed transaction, the server returns a single [OperationOutcome](https://hl7.org/fhir/R4/operationoutcome.html) instead of a Bundle.

The main motivation here is to make it so that transactions that fail due to rate limits return HTTP 429 so that `MedplumClient` can detect, sleep, and retry.

The Medplum CLI bulk import uses transaction bundles, which quickly fail on rate limits.

This is not backwards compatible.  We should discuss with the team before rolling out to check if there are any customer concerns.  If there are concerns, one potential compromise is to use the failed `OperationOutcome` for HTTP status, but continue to return the `Bundle`.  I'm more inclined to rip off the band-aid and return the proper response though.

### DIFF
--- a/packages/core/src/outcomes.ts
+++ b/packages/core/src/outcomes.ts
@@ -3,6 +3,7 @@
 import type { OperationOutcome, OperationOutcomeIssue } from '@medplum/fhirtypes';
 import { arrayify } from './array';
 import type { Constraint } from './typeschema/types';
+import { EMPTY } from './utils';
 
 const OK_ID = 'ok';
 const CREATED_ID = 'created';
@@ -225,8 +226,8 @@ export const tooManyRequests: OperationOutcome = {
  * @param msBeforeNext - Milliseconds until the rate limit resets.
  */
 export function setRateLimitReset(outcome: OperationOutcome, msBeforeNext: number): void {
-  outcome.issue[0].extension = [
-    ...(outcome.issue[0].extension ?? []).filter((e) => e.url !== RATE_LIMIT_RESET_EXTENSION_URL),
+  outcome.extension = [
+    ...(outcome.extension ?? EMPTY).filter((e) => e.url !== RATE_LIMIT_RESET_EXTENSION_URL),
     {
       url: RATE_LIMIT_RESET_EXTENSION_URL,
       valueUnsignedInt: Math.ceil(msBeforeNext / 1000),
@@ -240,9 +241,7 @@ export function setRateLimitReset(outcome: OperationOutcome, msBeforeNext: numbe
  * @returns Milliseconds until reset, or undefined if not present.
  */
 export function getRateLimitReset(outcome: OperationOutcome): number | undefined {
-  const seconds = outcome.issue
-    .flatMap((issue) => issue.extension ?? [])
-    .find((e) => e.url === RATE_LIMIT_RESET_EXTENSION_URL)?.valueUnsignedInt;
+  const seconds = outcome.extension?.find((e) => e.url === RATE_LIMIT_RESET_EXTENSION_URL)?.valueUnsignedInt;
   if (seconds !== undefined) {
     return seconds * 1000;
   }
--- a/packages/core/src/outcomes.ts
+++ b/packages/core/src/outcomes.ts
@@ -18,6 +18,7 @@ const CONTENT_TOO_LARGE_ID = 'content-too-large';
 const UNSUPPORTED_MEDIA_TYPE_ID = 'unsupported-media-type';
 const MULTIPLE_MATCHES_ID = 'multiple-matches';
 const TOO_MANY_REQUESTS_ID = 'too-many-requests';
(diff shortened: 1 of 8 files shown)


codyebberson:
@claude please review

packages/fhir-router/src/fhirrouter.ts
```diff
@@ -80,6 +81,16 @@ async function batch(req: FhirRequest, repo: FhirRepository, router: FhirRouter)
...
 
   const result = await processBatch(req, repo, router, bundle);
+
+  if (bundle.type === 'transaction') {
+    const outcome = result.entry?.[0]?.response?.outcome;
+    if (outcome && !isOk(outcome)) {
+      // For a failed transaction, the server returns a single OperationOutcome instead of a Bundle.
+      // See: https://hl7.org/fhir/R4/http.html#transaction-response
+      return [outcome];
+    }
```
mattwiller:
```suggestion
    ```suggestion
    const errorOutcome = result.entry?.find((e) => e.response?.outcome && !isOk(e.response.outcome));
    if (errorOutcome) {
      // For a failed transaction, the server returns a single OperationOutcome instead of a Bundle.
      // See: https://hl7.org/fhir/R4/http.html#transaction-response
      return [errorOutcome];
    }
```

mattwiller (commented):
Any error in the Bundle should probably result in an overall error (see suggestion).  Also,I believe this is only an issue when the "transaction" is actually running in batch mode; rolling out full transaction support would cause individual errors to abort the transaction as soon as they occur and return a bare `OperationOutcome` as expected

mattlong (commented):
If we're processing a `bundle.type === 'transaction'` in batch mode since transactions aren't enabled it effectively becomes a `batch` and shouldn't get transaction semantics for error handling. If we did, you could be obscuring possibly successful entry results by returning an `OperationOutcome` instead.

As mentioned above, we already return a sole `OperationOutcome` for true transactions.

That being said, I could be convinced that if every entry result is an (the same?) error, returning an `OperationOutcome` would be reasonable.

codyebberson:
Thank you @mattlong and @mattwiller - discussed offline, decided to move this into `MedplumClient.bulkImport` to handle more surgically.

packages/core/src/outcomes.ts
```diff
@@ -218,6 +219,47 @@ export const tooManyRequests: OperationOutcome = {
...
+}
+
+/**
+ * Returns the number of milliseconds until a rate limit resets.
+ * @param outcome - The rate limit outcome.
+ * @returns Milliseconds until reset, or undefined if not present.
+ */
+export function getRateLimitReset(outcome: OperationOutcome): number | undefined {
+  const seconds = outcome.issue
+    .flatMap((issue) => issue.extension ?? [])
```
mattwiller: Why `flatMap` here — isn't the rate limit extension guaranteed to be at the top level?

codyebberson: Good point. This is response-wide rate-limit metadata, and we control where it is written, so there’s no need to search every issue. I moved the extension to OperationOutcome.extension and simplified this to a direct lookup.

mattwiller (approved):

PR #6389, opened 2025-04-17
Using .env file to specify environment variables

finnbergquist:
Starting with one example app before replicating this pattern to all of the other example apps

Issue #5736, opened 2025-01-05
Standardize config settings across all example apps


codyebberson:
Almost all of our example apps have a handful of the same config settings:
1. `MEDPLUM_BASE_URL` - for which API server to use
2. `GOOGLE_CLIENT_ID` - for Google auth
3. `RECAPTCHA_SITE_KEY` - for reCAPTCHA

The example apps are not consistent in how these settings are applied.  A few patterns:
1. Hardcoded strings
2. A `config.ts` file
3. A `.env` file

I was playing with the idea of an NPM initializer here: https://github.com/medplum/medplum/pull/5734

It would be nice and convenient if all of the example apps followed the same convention for these config settings, so that the NPM initializer could use the same flow.  I think it would probably be a better experience for users too, because it would be more consistent and predictable.

My personal preference would probably be to move everything to `.env` files, but I wanted to solicit feedback from the team on that first.  The exact mechanism is less important than making everything more consistent.

PR #9507, opened 2026-06-14
Simplified example app config settings

codyebberson:
Potentially controversial follow-up to https://github.com/medplum/medplum/issues/5736 and https://github.com/medplum/medplum/pull/6389

This PR standardizes configuration loading across the Vite-based example apps in `examples/`.

Changes include:

- Added or normalized `.env.defaults` for every browser example app.
- Added the Vite startup copy behavior so each app creates `.env` from `.env.defaults` when missing.
- Standardized Vite public env prefixes for Medplum auth-related settings.
- Removed the old `src/config.ts` app config wrappers.
- Replaced hardcoded `SignInForm` auth settings with `import.meta.env` reads.
- Updated README references that pointed users at deleted `src/config.ts` files.

## Motivation

The example apps previously used a mix of config patterns:

- Some read from `.env`.
- Some used `src/config.ts`.
- Some hardcoded Medplum or Google auth values directly in the app.

That made the examples harder to compare, harder to copy from, and easier to misconfigure. The goal of this change is to keep behavior the same while making configuration predictable across examples.

## New Convention

Each browser example app now has a `.env.defaults` file with the shared baseline values:

```dotenv
MEDPLUM_BASE_URL=https://api.medplum.com/ # Or http://localhost:8103/ for localhost
MEDPLUM_CLIENT_ID= # Optional
GOOGLE_CLIENT_ID=921088377005-3j1sa10vr6hj86jgmdfh2l53v3mp7lfi.apps.googleusercontent.com # Optional
RECAPTCHA_SITE_KEY=6LfFd_8gAAAAAOCVrZQ_aF2CN5b7s91NEYIu5GxL # Optional
```

Apps with additional app-specific settings keep those settings in the same `.env.defaults` file after the shared block. For example, Photon keeps its `PHOTON_*` values and FooMedical keeps `MEDPLUM_PROJECT_ID`.

Each browser example app's `vite.config.ts` now creates `.env` from `.env.defaults` when `.env` is missing:

```ts
if (!existsSync(path.join(__dirname, '.env'))) {
  copyFileSync(path.join(__dirname, '.env.defaults'), path.join(__dirname, '.env'));
}
```

Each app also exposes the shared env prefixes through Vite:

```ts
envPrefix: ['MEDPLUM_', 'GOOGLE_', 'RECAPTCHA_'],
```

Apps with legitimate app-specific public env variables may include additional prefixes. For example, the Photon integration also exposes `PHOTON_`.

## Why Remove `src/config.ts`

The old `src/config.ts` files were mostly thin wrappers around `import.meta.env`. Keeping them created a second configuration pattern without adding much value, and it made the examples look more different than they really were. Reading env values directly at the usage site keeps the examples simpler and makes Vite the single source of truth for browser-exposed configuration.

## Localhost Notes

Older `.env.defaults` files included repeated commented-out localhost blocks. Those were removed in favor of the inline `MEDPLUM_BASE_URL` hint:

```dotenv
MEDPLUM_BASE_URL=https://api.medplum.com/ # Or http://localhost:8103/ for localhost
```

This keeps the defaults files small and uniform while preserving the most important local development guidance.


codyebberson:
@claude please review

codyebberson:
@claude please re-review

ThatOneBro (approved):

PR #9523, opened 2026-06-16
Remove Slot-based alpha scheduling implementations

noahsilas:
In the original alpha scheduling specification, we thought we would offer both a `Slot` based implementation and an `Appointment` based implementation. After building out the `Slot` version first, we decided that it would be better for the platform to be more opinionated and offer a smaller set of scheduling primitives.

We settled on using `Appointment` as the basis for our scheduling APIs. The deciding factor here is that it was clear how this could more cleanly satisfy complex workflows involving multiple Schedule resources.

Breaking Changes
===============
- `GET /Schedule/:id/$find` has been removed. Users of this API should migrate to `GET /Appointment/$find` instead. See https://www.medplum.com/docs/scheduling/appointment-find for updated usage details.
- `POST /Appointment/$book` no longer accepts `slot` or `patient-reference` parameters.  Users should pass an `appointment` parameter. The `patient-reference` should be embedded in `appointment.participants`. The `slot` should be embedded in `appointment.contained`. See https://www.medplum.com/docs/scheduling/appointment-book for updated usage details.


techdavidy:
@claude review

packages/server/src/fhir/operations/book.ts
```diff
@@ -89,218 +38,14 @@ async function bookFromProposedAppointmentHandler(proposedAppointment: Appointme
...
-        slot: createdSlots.map((slot) => ({ reference: getReferenceString(slot) })),
-        participant,
-        start,
-        end,
-      });
-      return [appointment, ...createdSlots, ...createdBufferSlots];
-    },
-    { serializable: true }
+    withPath(params.appointment, 'Parameters.appointment'),
+    (appointment, _slots) => {
```
techdavidy: we can remove slot from the customizer ?

techdavidy (approved):

packages/server/src/fhir/operations/book.ts
noahsilas: We can. I thought it was sort of nice to be able to see what other input this function can see without jumping to the definition, but you're right that this is functionally nothing.

PR #9524, opened 2026-06-16
Enhance patient invitation logic in admin module

### LINKED ISSUE #8843: Apply Project.defaultPatientAccessPolicy to Admin patient invites
**Issue:** Project.defaultPatientAccessPolicy is not applied when inviting patient Users through the admin console. It is currently applied via [open patient registration](https://www.medplum.com/docs/user-management/open-patient-registration). Invites with no AccessPolicy selected create a ProjectMembership without AccessPolicy, so patients can get overly broad access until the membership is fixed manually.

**Suggested fix:** If resourceType is Patient (and AccessPolicy is not provided on the invite), the server should set ProjectMembership.accessPolicy from Project.defaultPatientAccessPolicy

Admin -> Invite New User:
Image

deam65:
- Updated invite functionality to apply default patient access policy when no explicit policy is provided.
- Added tests to verify behavior for inviting patients with and without explicit access policies.
- Ensured that inviting practitioners does not apply the default patient access policy.
- `upsertProjectMembership()` flow in `invite.ts` owns the default policy assignment. 
- Removed `defaultPatientAccessPolicy` assignment in SCIM user provisioning, letting `inviteUser() > upsertProjectMembership()` flow handle it. 

This improves the handling of access policies during user invitations, ensuring consistency and expected behavior in various scenarios.

Signed-off-by: Darren Eam <darreneam65@gmail.com>

### DIFF
--- a/packages/server/src/admin/invite.test.ts
+++ b/packages/server/src/admin/invite.test.ts
@@ -2,7 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 import { SendEmailCommand, SESv2Client } from '@aws-sdk/client-sesv2';
 import { allOk, ContentType, createReference, getReferenceString, normalizeErrorString } from '@medplum/core';
-import type { AccessPolicy, BundleEntry, Practitioner, ProjectMembership, User } from '@medplum/fhirtypes';
+import type {
+  AccessPolicy,
+  BundleEntry,
+  Patient,
+  Practitioner,
+  ProjectMembership,
+  RelatedPerson,
+  User,
+} from '@medplum/fhirtypes';
 import type { AwsClientStub } from 'aws-sdk-client-mock';
 import { mockClient } from 'aws-sdk-client-mock';
 import 'aws-sdk-client-mock-jest';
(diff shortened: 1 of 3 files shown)


deam65:
@claude please review

deam65:
@claude please review once more. Feedback has been addressed

deam65:
@claude please do one final review

deam65:
> The main open question is whether the upsert-overwrite behavior (existing custom policy → silently replaced by default) is intentional.

@claude I think this shouldn't be an issue because there isn't a scenario where we would invite someone that already exists as a Patient. Do you foresee any problems with this? Does it make sense to fail the invite flow on discovery of a duplicate?

deam65:
@claude actually, it seems like inviting a patient that already exists as a Patient resource can be problematic. Another client just went through this when inviting a clinician who had multiple Practitioner resources, pointing to the same email in one project. They got a misleading error which has been fixed in https://github.com/medplum/medplum/pull/9536 

so for this, we may want to overwrite the existing Patient AP - since that is the most recent upsert action for Patients that already exist. Does this make sense, or do you need more clarification?

codyebberson (approved):
Nice 👍

PR #9530, opened 2026-06-16
feat(terraform/gcp): namespace resources for multi-deployment + fixes

ianplunkett:
## Summary

Makes the GCP Terraform config (`terraform/gcp/`) reusable so multiple Medplum
deployments can live in the **same** GCP project without resource-name collisions,
plus a few correctness fixes found while deploying it end-to-end.

## Changes

- **Namespacing** — new `var.namespace` and `locals.tf` derive every named resource
  (VPC, subnets, GKE, Cloud NAT, global addresses, Cloud SQL, Redis, buckets, CDN LB,
  Cloud Armor policies, service account, firewall rule) from a single prefix. Default
  `namespace = "medplum"` preserves the previous names.
- **`var.gke_zones`** — decouples the autopilot cluster zones from the hardcoded
  `us-west1-a/b` pair.
- **DB credentials** — new `var.db_password` (generates a strong random password via
  `random_password` when unset) plus a sensitive `db_password` output, replacing the
  hardcoded `medplum`/`medplum` user/password.
- **Bucket-key bugfix** — `cdn.tf` referenced `module.buckets["medplum-storage-p"]` /
  `["medplum-app-p"]`, which never matched the `gcs_buckets` keys (`plan` would fail).
  Buckets are now keyed by logical name (`storage`/`app`) and named `<namespace>-<key>`.
- **GCS presigned URLs** — adds `roles/iam.serviceAccountTokenCreator` on the runtime
  GSA *to itself*, required for V4 SignBlob under Workload Identity (the previously
  "manual" GCP signing step).
- **Provider bump** — `google` 5.45.2, which fixes the spurious
  `node_pool_auto_config.insecure_kubelet_readonly_port_enabled` "Must specify a field
  to update" error on re-apply of the Autopilot cluster.

`terraform.tfvars` remains a placeholder template (updated for the corrected bucket
keys and new `namespace`/`gke_zones` variables); no environment-specific values are
committed.

## Testing

Deployed the full stack from this config into a live GCP project and verified
end-to-end: `terraform apply` is clean/idempotent, the GKE backend pod reads its
config from Secret Manager via Workload Identity (`/healthcheck` reports
`postgres:true, redis:true`), and the frontend serves over HTTPS through the CDN.


codyebberson:
@claude please review.  Please specifically call out if any of these are breaking changes.

codyebberson:
@ianplunkett  - thoughts on these?  I think most GCP customers currently use our Terraform as reference material rather than as an actual dependency, so it might be ok.  We should probably document these changes though for anyone who intends to upgrade.

ianplunkett:
@codyebberson I agree that most folks are just using this as a reference implementation. I think everyone will be forking this as at least some of the files notably `terraform/gcp/terraform.tfvars` needs customized values.  Do you think we need more documentation than the PR notes here?

ianplunkett:
Closing this for now

PR #9533, opened 2026-06-17
WIP: fix(ws): send subscription bind handshake before heartbeat can fire

The-Alchemist:
## Summary

- Fix a race in WebSocket subscription bind where a periodic heartbeat could be delivered before the bind handshake
- After `subscribeWsToSubscription`, the server registered the heartbeat handler and awaited Redis before sending the handshake; with a short heartbeat interval (25ms in tests, or under load in prod), clients could receive `SubscriptionStatus.type: heartbeat` as the first message instead of `handshake`
- Send the handshake immediately after subscribing the socket, before `ensureHeartbeatHandler()` and any `await`

## Test plan

- [ ] `cd packages/server && npx vitest run src/ws/subscriptions.test.ts -t "Recovers when the subscription handler fails to initialize"` (stress loop if possible)
- [ ] `cd packages/server && npx vitest run src/ws/subscriptions.test.ts`
- [ ] Manual bind: confirm first JSON message after `bind-with-token` is always handshake

### DIFF
--- a/packages/server/src/ws/subscriptions.ts
+++ b/packages/server/src/ws/subscriptions.ts
@@ -418,6 +418,10 @@ export async function handleR4SubscriptionConnection(socket: WebSocket, request:
       membershipId: verifiedToken.membership_id,
     });
     subscribeWsToSubscription(socket, verifiedToken.subscription_id, rawToken, criteriaResourceType);
+    // Send handshake before ensureHeartbeatHandler or any await so a heartbeat tick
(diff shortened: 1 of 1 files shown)


ThatOneBro:
I think this is a duplicate of https://github.com/medplum/medplum/pull/9531 ?

The-Alchemist:
@ThatOneBro : 

> I think this is a duplicate of #9531 ?


you're totally right.  closing.  it was causing some flaky tests for me

PR #9534, opened 2026-06-17
fix(ccda): normalize address part text nodes in mapAddresses

jdjkelly:
## Problem

In the C-CDA XML ITS, address parts (`ADXP`) and name parts (`ENXP`) may legally carry attributes — most commonly `partType` (e.g. `<city partType="CTY">Anytown</city>`). When an element has both text content **and** an attribute, `fast-xml-parser` represents it as an object rather than a string:

```json
{ "#text": "Anytown", "@_partType": "CTY" }
```

`mapCcdaNameToFhirHumanName` already normalizes every name part through `nodeToString`, so names are unaffected. But `mapAddresses` assigned the parsed values directly:

```ts
city: addr.city,   // -> { "#text": "Anytown", "@_partType": "CTY" } when an attribute is present
```

So FHIR `Address` fields can come out as objects instead of strings whenever the source emits the attributed form. Downstream this breaks FHIR validation (`$validate` reports `Invalid additional property "#text"` / `"@_partType"`) and any consumer that treats `Address.city` as a string (e.g. `city.trim()` throws `is not a function`).

This form is legal, just verbose — generators built on generic HL7 V3 datatype serializers commonly stamp `partType` on every part — so the converter should handle it on addresses the same way it already does on names.

## Fix

- Run every address part (`line`, `city`, `state`, `postalCode`, `country`) through `nodeToString`, mirroring `mapCcdaNameToFhirHumanName`.
- Widen the `CcdaAddr` string fields to `string | CcdaText` to reflect what the parser actually produces.

## Test

Adds `address.test.ts` covering both the `partType`-attributed form and the plain-string form, asserting the FHIR `Address` parts are clean strings with no `#text` / `@_partType` artifacts.

`npx vitest run` in `packages/ccda`: **266 passed, 1 skipped** (no regressions).

### DIFF
--- a/packages/ccda/src/address.test.ts
+++ b/packages/ccda/src/address.test.ts
@@ -1,0 +1,71 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { Patient } from '@medplum/fhirtypes';
+import { convertCcdaToFhir } from './ccda-to-fhir';
+import { convertXmlToCcda } from './xml';
+
+/**
+ * Address parts that carry an XML attribute (e.g. the HL7 `partType` on ADXP
+ * elements) are parsed by fast-xml-parser into `{ '#text': value, '@_partType': ... }`
+ * objects rather than bare strings. `mapAddresses` must normalize these via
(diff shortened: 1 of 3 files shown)


mattwiller (approved):
Looks good to me, thanks @jdjkelly!

codyebberson (approved):

PR #9541, opened 2026-06-17
Handle additional values for boolean Coding properties

mattwiller:


packages/server/src/fhir/operations/codesystemlookup.ts
```diff
@@ -123,16 +123,20 @@ export async function lookupCoding(
   return output;
 }
 
+const TRUE_VALUES = ['true', 't', '1'];
```
The-Alchemist, line 126: i like sonarqube's feedback here (use a `Set`), but it's very minor and borderline pedantic

The-Alchemist (approved):

packages/server/src/fhir/operations/codesystemlookup.ts
mattwiller: For small collections like this, I believe it's probably faster to do equality testing on the elements all nearby in memory than to do the hashing required for set

PR #9544, opened 2026-06-18
test: add _project search parameter tests for MemoryRepository (#8831)

### LINKED ISSUE #8831: MockClient: `_project` search parameter not supported in MemoryRepository / matchesSearchFilter

rtmalikian:
## Summary

Adds comprehensive test coverage for the `_project` search parameter in `MemoryRepository` and `matchesSearchRequest`, closing #8831.

The `_project` search parameter was registered in `globalSchema` by #9042, but the issue remained open because there were **zero unit tests** verifying that `_project` filtering works correctly in the `MemoryRepository` / `matchesSearchRequest` code path used by `MockClient`.

### Root Cause (from issue)

`matchesSearchFilter` looks up the filter code in `globalSchema.types[resourceType].searchParams`. Before #9042, `_project` was not present, so the lookup returned `false` and **every resource was excluded**. After #9042 added `_project` to `getOrInitTypeSchema()`, the lookup succeeds and the token filter matches correctly — but no tests existed to verify this.

### Tests Added

**`packages/core/src/search/match.test.ts`** (6 new tests):
- `_project` EQUALS filter matches correct project, rejects others and resources without project
- `_project` NOT_EQUALS operator
- `_project` MISSING true/false
- `_project` PRESENT true/false
- `_project` combined with `_tag` filter (AND logic)
- `_project` on non-Patient resource types (Observation)

**`packages/fhir-router/src/repo.test.ts`** (2 new tests):
- `MemoryRepository.search` with `_project` filter — verifies project isolation
- `MemoryRepository.search` with `_project` + `name` filter combined

### Verification

```
✓ src/search/match.test.ts (60 tests) 332ms
✓ src/repo.test.ts (15 tests) 341ms
```

All tests pass, including the 8 new tests.

Closes #8831

---

📧 rtmalikian@gmail.com

---

**Disclosure:** This code was developed with assistance from **mimo-v2.5-pro** (Xiaomi) via **Hermes Agent** (Nous Research). All changes were reviewed, tested against the actual codebase, and verified for correctness.

### DIFF
--- a/packages/core/src/search/match.test.ts
+++ b/packages/core/src/search/match.test.ts
@@ -1139,4 +1139,153 @@ describe('Search matching', () => {
     };
     expect(matchesSearchRequest(resource, search1)).toBe(false);
   });
+
+  describe('_project', () => {
+    test('Filter by _project matches correct project', () => {
+      const resource1: Patient = {
+        resourceType: 'Patient',
+        id: 'patient-1',
+        meta: { project: 'project-1' },
+      };
+
+      const resource2: Patient = {
+        resourceType: 'Patient',
+        id: 'patient-2',
+        meta: { project: 'project-2' },
+      };
+
+      const resource3: Patient = {
+        resourceType: 'Patient',
+        id: 'patient-3',
+        // no project
+      };
+
+      const search: SearchRequest = {
+        resourceType: 'Patient',
+        filters: [{ code: '_project', operator: Operator.EQUALS, value: 'project-1' }],
+      };
+
+      expect(matchesSearchRequest(resource1, search)).toBe(true);
+      expect(matchesSearchRequest(resource2, search)).toBe(false);
+      expect(matchesSearchRequest(resource3, search)).toBe(false);
+    });
+
+    test('_project with NOT_EQUALS operator', () => {
+      const resource1: Patient = {
+        resourceType: 'Patient',
+        id: 'patient-1',
+        meta: { project: 'project-1' },
+      };
+
+      const resource2: Patient = {
+        resourceType: 'Patient',
+        id: 'patient-2',
+        meta: { project: 'project-2' },
+      };
+
+      const search: SearchRequest = {
+        resourceType: 'Patient',
+        filters: [{ code: '_project', operator: Operator.NOT_EQUALS, value: 'project-1' }],
+      };
+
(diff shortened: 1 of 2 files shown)


The-Alchemist:
@cursoragent review

PR #9553, opened 2026-06-18
[Server] Enforce scheduling alignment in $book

noahsilas:
In #9488 (commit 6bd3b68ae) we established some rules for how scheduling slots should align to a grid when they are not evenly divisible by the hour. This should resolve the tension that we saw when we originally implemented the alignment testing: before that commit, `$find` could emit slots that would not pass validation, so we had disabled this enforcement (see #8331, commit f970721a9 for details).

With stronger understanding of the grid and more edge cases handled, it should now be the case that any results generated by `$find` (even with awkward alignment values) should be accepted by this validation.


noahsilas:
@claude Please review.

codyebberson (approved):
Nice 👍

PR #9565, opened 2026-06-19
fix: include _id search param in field selector display (Fixes #6043)

### LINKED ISSUE #6043: "_id" search param not mapped to "ID" display string
Image

rtmalikian:
Fixes #6043

## Problem

The `_id` search parameter was not included in the field selector dropdown because the `getFieldsList` function in `SearchFieldEditor` deduplicated fields by display name. Since both the `id` property and `_id` search parameter resolve to the display name "ID", the `_id` search parameter was filtered out.

When `_id` appeared in the `fields` array (e.g., from URL parsing like `Patient?_fields=_id,name`), Mantine's `MultiSelect` couldn't find a matching entry in the data list and displayed the raw `_id` string instead of the user-friendly "ID" label.

## Solution

Removed the display-name-based deduplication from `getFieldsList`. The function now deduplicates by field code only (using the `keys` set), which correctly treats `id` and `_id` as distinct fields while still preventing true duplicates.

## Verification

All existing tests pass:

```
✓ src/SearchFieldEditor/SearchFieldEditor.test.tsx (3 tests) 241ms
✓ src/SearchControl/SearchControlField.test.ts (1 test) 2ms
✓ src/SearchControl/SearchUtils.test.tsx (21 tests) 5ms
```

ESLint passes with no new warnings (only a pre-existing `react-hooks/set-state-in-effect` warning unrelated to this change).

---

📧 rtmalikian@gmail.com

---

**Disclosure:** This code was developed with assistance from **mimo-2.5-pro** (Xiaomi) via **Hermes Agent** (Nous Research). All changes were reviewed, tested against the actual codebase, and verified for correctness.

### DIFF
--- a/packages/react/src/SearchFieldEditor/SearchFieldEditor.tsx
+++ b/packages/react/src/SearchFieldEditor/SearchFieldEditor.tsx
@@ -132,23 +132,19 @@ function getFieldsList(
 ): string[] {
   const result = [] as string[];
   const keys = new Set<string>();
-  const names = new Set<string>();
 
   // Add properties first
   for (const key of Object.keys(typeSchema.elements)) {
     result.push(key);
     keys.add(key.toLowerCase());
-    names.add(buildFieldNameString(key));
   }
 
-  // Add search parameters if unique
+  // Add search parameters if unique by code
   if (searchParams) {
     for (const code of Object.keys(searchParams)) {
-      const name = buildFieldNameString(code);
-      if (!keys.has(code) && !names.has(name)) {
+      if (!keys.has(code)) {
         result.push(code);
         keys.add(code);
-        names.add(name);
       }
     }
   }


codyebberson (commented):
Thanks @rtmalikian  - We need to take a deeper look at this one.  Some of that behavior is intentional (combining "elements" and "search params" into a single "field")

rtmalikian:
Thanks for the feedback @codyebberson! I understand — combining "elements" and "search params" into a single "field" entry is intentional design.

The core issue from #6043 is that when `_id` appears in the URL `fields` param (e.g., `Patient?_fields=_id,name`), the Mantine MultiSelect shows the raw `_id` string instead of the user-friendly "ID" label, because the deduplication removes `_id` from the data list before the component can match it.

Could you suggest the preferred approach? Some options:

1. **Keep deduplication but fix the display resolution** — Instead of removing `_id` from the list, keep it but mark it as an alias that maps to the combined "ID" entry. The MultiSelect would then display "ID" regardless of whether `id` or `_id` was in the URL fields.

2. **Separate the display list from the value list** — Keep deduplication for the dropdown display, but add a value-to-display mapping so URL-parsed `_id` resolves to the "ID" label.

3. **Other approach** you prefer?

Happy to adjust the PR to match the intended architecture.

rtmalikian:
Friendly ping @codyebberson — checking in on the architectural direction for the _id field display fix. Happy to adjust the PR to match the intended approach when you have a moment to share your preference. Thanks!

PR #9566, opened 2026-06-19
fix: resolve $mantine-breakpoint-xs in Header.module.css to prevent CSS leakage (Fixes #9257)

### LINKED ISSUE #9257: @medplum/react: shipped CSS contains unresolved Mantine placeholder ($mantine-breakpoint-xs), forces postcss-preset-mantine on consumers
## Summary
The CSS published in `@medplum/react@5.1.12` (`dist/esm/index.css`) contains an unresolved Mantine SCSS-style placeholder, `$mantine-breakpoint-xs`. The substitution is currently deferred to consumer build pipelines, which forces every consuming app to add `postcss-preset-mantine` to its PostCSS chain and to switch CSS minifiers away from Vite's default `lightningcss`. The placeholder should be resolved at the package's own build time so consumers receive valid CSS.

## Affected version
`@medplum/react@5.1.12` (latest on npm as of filing).

## Source location
`packages/react/src/AppShell/Header.module.css:30`

```css
.userName {
  ...
  @media (max-width: $mantine-breakpoint-xs) {
    display: none;
  }
}
```

The unresolved token survives into the shipped artifact at `dist/esm/index.css` inside the `.Header_userName` selector:

```
.Header_userName{...@media (max-width: $mantine-breakpoint-xs){display:none}}
```

## Reproduction
1. `npm i @medplum/react@5.1.12 @mantine/core @mantine/hooks`
2. Create a Vite app and import the bundled CSS from `@medplum/react`.
3. `vite build` with the default `cssMinify: 'lightningcss'`. The build fails because `lightningcss` rejects `$mantine-breakpoint-xs` as an invalid media query value.

## Current consumer workaround
To get the package to build, consumers must:
- Add `postcss-preset-mantine` to their PostCSS chain so the placeholder is rewritten before minification.
- Set `build.cssMinify = 'esbuild'` in `vite.config.ts` (lightningcss validates media queries even with the preset wired, due to plugin-ordering interactions at minify time).

## Expected behavior
The Mantine breakpoint substitution should happen during `@medplum/react`'s own build, so `dist/esm/index.css` ships with the placeholder already resolved to the appropriate `em`/`px` value, or hoisted to a `--mantine-breakpoint-xs` CSS custom property that consumers can override. Consumers of a prebuilt CSS bundle should not need a SCSS-aware PostCSS plugin to use it.

## Environment
- `@medplum/react`: 5.1.12
- Mantine: 8.x (per `@medplum/react` peer deps)
- Build tool: Vite (default config); also reproduces with any tool whose CSS minifier validates media queries (lightningcss, csso strict mode).

## Impact
Low severity but high friction: every consumer either ships an invalid CSS bundle, adds the workaround above, or rewrites the affected CSS Module locally.

rtmalikian:
Fixes #9257

## Problem

`@medplum/react` ships CSS in `dist/esm/index.css` that contains an unresolved Mantine SCSS variable `$mantine-breakpoint-xs`. The esbuild build pipeline for the react package does not run PostCSS plugins (`postcss-preset-mantine`, `postcss-simple-vars`), so the raw variable leaks into the distributed CSS.

This causes build failures for consumers using Vite with `lightningcss` (the default CSS minifier), which validates media queries and rejects `$mantine-breakpoint-xs` as an invalid value.

## Solution

Replace the raw `$mantine-breakpoint-xs` variable in `Header.module.css` with its resolved value `36em`. This is the same value defined in `postcss.config.mjs` and matches the Mantine default breakpoint.

The `postcss.config.mjs` in the react package defines:
- `$mantine-breakpoint-xs`: `36em`

The `@mixin smaller-than` syntax used in `packages/app/src/HomePage.module.css` is the correct PostCSS pattern, but it only works when PostCSS is in the pipeline. For distributed CSS, the resolved value must be used directly.

## Verification

1. Confirmed no other CSS files in `packages/react/src` contain unresolved `$mantine-` variables
2. The value `36em` matches the Mantine default and the project's `postcss.config.mjs`
3. The `@media (max-width: 36em)` query is valid CSS that works without any PostCSS processing

## Changelog

| Date | Change | Author |
|------|--------|--------|
| 2026-06-19 | Replace $mantine-breakpoint-xs with 36em in Header.module.css | rtmalikian |

### Files Changed
- `packages/react/src/AppShell/Header.module.css` — Replace unresolved SCSS variable with resolved value

### Verification
- Confirmed no remaining `$mantine-` variables in `packages/react/src/**/*.css`
- Value `36em` matches project's `postcss.config.mjs` definition

---

📧 rtmalikian@gmail.com

---

**Disclosure:** This code was developed with assistance from **mimo-v2.5-pro** (Xiaomi) via **Hermes Agent** (Nous Research). All changes were reviewed, tested against the actual codebase, and verified for correctness.

### DIFF
--- a/packages/react/src/AppShell/Header.module.css
+++ b/packages/react/src/AppShell/Header.module.css
@@ -27,7 +27,10 @@
   line-height: 1px;
   margin-right: 3px;
 
-  @media (max-width: $mantine-breakpoint-xs) {
+  /* Use resolved value (36em) instead of $mantine-breakpoint-xs to avoid
+     leaking unresolved SCSS variables into distributed CSS.
+     See: https://github.com/medplum/medplum/issues/9257 */
+  @media (max-width: 36em) {
     display: none;
   }
 }


codyebberson:
Thanks for digging into this and for the detailed analysis. I agree you've identified a real issue here: consumers can encounter build failures when Mantine-specific CSS syntax leaks through into the distributed package.

That said, I don't think hardcoding `36em` is the right fix for us. `@medplum/react` is intentionally coupled to Mantine, and we increasingly rely on Mantine's theming and styling system. Replacing Mantine variables with literal values would duplicate theme configuration and make it easier for things to drift over time.

I think the underlying issue is that these styles are not being processed through the Mantine PostCSS pipeline before distribution. We'd prefer to address that root cause (or document the required Mantine PostCSS setup for consumers) rather than hardcoding breakpoint values in the source.

Relevant Mantine docs:

* [Mantine PostCSS preset](https://mantine.dev/styles/postcss-preset/)
* [Mantine CSS modules](https://mantine.dev/styles/css-modules/)

For now I'm going to close this PR, but the underlying issue is worth discussing and we'd definitely welcome a proposal that preserves Mantine's theming capabilities while fixing the build/distribution problem.

PR #9567, opened 2026-06-19
Fixes #1563 - Replace node-fetch with built-in

### LINKED ISSUE #1563: Replace node-fetch with Node.js 18 built-in fetch

codyebberson:
Fixes #1563

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48168,14 +48168,8 @@
         "medplum": "dist/cjs/index.cjs"
       },
       "devDependencies": {
-<<<<<<< HEAD
         "@medplum/fhirtypes": "5.1.21",
         "@medplum/mock": "5.1.21",
-        "@types/node-fetch": "2.6.13",
-=======
-        "@medplum/fhirtypes": "5.1.18",
-        "@medplum/mock": "5.1.18",
->>>>>>> ccf573c80 (Removed node-fetch from cli and examples)
         "@types/semver": "7.7.1",
         "aws-sdk-client-mock": "4.1.0"
       },
@@ -48830,16 +48824,8 @@
         "ioredis": "5.8.2",
         "jose": "5.10.0",
         "jszip": "3.10.1",
-<<<<<<< HEAD
-<<<<<<< HEAD
         "node-fetch": "2.7.0",
         "nodemailer": "9.0.1",
-=======
-=======
-        "node-fetch": "2.7.0",
->>>>>>> fc7395f66 (Restore node-fetch for vmcontext bots)
-        "nodemailer": "8.0.11",
->>>>>>> b8ca2e6f5 (Replace node-fetch with built-in)
         "otplib": "12.0.1",
         "pdfmake": "0.2.21",
         "pg": "8.22.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -48161,7 +48161,6 @@
         "dotenv": "17.4.2",
         "fast-glob": "3.3.3",
         "iconv-lite": "0.7.2",
-        "node-fetch": "2.7.0",
         "semver": "7.8.4",
         "tar": "7.5.16"
       },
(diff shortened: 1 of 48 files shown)


packages/server/src/bots/vmcontext.ts
```diff
@@ -75,7 +74,7 @@ export async function runInVmContext(request: BotExecutionContext): Promise<BotE
 
   const sandbox = {
     console: botConsole,
-    fetch,
+    fetch: globalThis.fetch,
```
ThatOneBro: This is an API change for VM context users, especially around timeouts

ThatOneBro (commented):
Nice, finally! 

In line with the idea of replacing all instances in server, but one note about API change that we should be aware of. Maybe it's fine since it's contained within `vmcontext` but still worth pointing out

packages/server/src/bots/vmcontext.ts
codyebberson: Hm.  Good point.  Maybe we should save that one as a breaking change for v6.

Self-hosted customers on Azure and GCP definitely use that.

I'll file a separate ticket to audit all of our enterprise bots for any usage of the `node-fetch` timeout code.

Good catch, thanks @ThatOneBro

techdavidy: yeah when working on SSE I remember moving away from node-fetch.
But reverted due to timeouts.

codyebberson:
1. Reverted the changes to vmcontext, holding that back until v6
2. Removed `node-fetch` from all enterprise bots: https://github.com/medplum/medplum-ee/pull/834

packages/server/src/bots/vmcontext.ts
codyebberson: @techdavidy  - do you remember where that was at?  Was it in `server` or a bot?  I think between this PR and [the ee PR](https://github.com/medplum/medplum-ee/pull/834) I removed all references to `node-fetch`, but want to make sure I'm not missing something in another repo.

codyebberson:
@claude - please review

codyebberson:
@claude - To explain why `node-fetch` is still in `packages/bot-layer/package.json` and `packages/server/package.json`, we need to discuss how VMContext bots are used.

Both Medplum's enterprise bots and customer's self-managed bots can run custom code for arbitrary business logic or proprietary integrations.  When running on an AWS cluster using AWS Lambda for bots, we need to continue to include `node-fetch` as a dependency that is included in the AWS Lambda Layer, so those bots can continue to `import fetch from 'node-fetch';`.  When running on a production cluster that uses VMContext or Fission for bots, a similar principle applies.

Medplum typically publishes one major release per year, usually in October.  We can mark the `node-fetch` dependency as deprecated, but I don't know if we'll actually remove it just yet, because we need to give enterprise customers adequate time to update code, update dependencies, test, verify, etc.  I think the more likely guess is that we will need to continue to include `node-fetch` in `bot-layer` for at least one more full cycle.

packages/server/src/bots/vmcontext.ts
techdavidy: yeah it is the line that Derrick mentioned. I see it is reverted back too in this. Should be good to go

techdavidy (approved):

PR #9568, opened 2026-06-19
fix(react): Use ListWithDetailPane in ThreadInbox

techdavidy:
Refactor ThreadInbox in react packages to use ListWithDetailPane.


packages/react/src/chat/ThreadInbox/ChatList.tsx
noahsilas, line 1: Removing this component is a breaking change, I think we should mark it as deprecated and leave it in place until the next major version. 😪

packages/react/src/chat/ThreadInbox/ThreadDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+                {thread.status !== 'completed' && (
+                  <Menu.Dropdown>
+                    <Menu.Item onClick={() => onStatusChange('completed')}>Completed</Menu.Item>
+                  </Menu.Dropdown>
+                )}
+              </Menu>
+            </Flex>
+            <Divider />
+            <Flex direction="column" style={{ flex: 1 }} h="100%">
```
noahsilas: This Flex container looks like it only has one child, so I suspect that we can probably replace this with a `Box` or even a `div`. 

EDIT: Just realized that this is code that mostly was moved rather than authored, if you want to leave this as-is to keep that relationship clear, that also seems fine to me.

techdavidy:
@claude please review

packages/react/src/chat/ThreadInbox/ThreadDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+                onViewInDocuments={onViewInDocuments}
+              />
+            </Flex>
+          </Stack>
+        </Paper>
+      </Flex>
+
+      {/* Right sidebar - Patient summary */}
+      {thread.subject && showPatientSummary && (
+        <Flex direction="column" w={300} h="100%">
```
noahsilas: `Flex` with a single child here too

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
```diff
@@ -173,165 +153,78 @@ export function ThreadInbox(props: ThreadInboxProps): JSX.Element {
...
-                      boundaries={1}
-                    />
-                  </Center>
-                </Box>
-              )}
-            </Paper>
-          </Flex>
+  // The list renders the parent thread (topic) of each tuple; the last message is
+  // looked up by thread id when rendering each row.
+  const items = useMemo(() => threadMessages.map(([topic]) => topic as WithId<Communication>), [threadMessages]);
```
noahsilas: I wondered if we can update the type of `useThreadInbox` to annotate its result as `WithId<Communication>` instead of applying this cast. It looks like usually messages it returns are the results of fetching data that are guaranteed to have an ID, but that messages can also be pushed in via `addThreadMessage` which doesn't have the same guarantee.

I think the result of that is that this cast is unsafe, which is troublesome because it means that `items` doesn't actually satisfy the constraint for `ListWithDetailPane`...

Maybe we filter out ID-less items? Or maybe we assign them some kind of transient identifier so that we can display them and link to them?

Or maybe we can make `addThreadMessage` only accept `WithId<Communication>` - that is technically a breaking change, but only in the type layer. Is that something we feel we can adjust in a minor version release?

noahsilas (commented):
Overall I really like how this separates the display markup from the component logic in ThreadInbox! One easy-to-fix issue is that I think we need to retain `ChatList` as an exported component since this is in our public library.

The slightly harder question is what to do with threads that don't have an ID. I _think_ that in this component's usage that will never happen - `NewTopicDialog` only emits a message after it has been persisted and has an ID. I can imagine future components that optimistically update before receiving the response from the server, but in that case they probably need some kind of transient local identifier to ensure that updates are applied correctly, so maybe that entails a much larger update to support anyways...

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
techdavidy: this is an edge case that we just present the thread but if no message is sent. We don't display the thread anymore

packages/react/src/chat/ThreadInbox/ChatList.tsx
techdavidy: It should be fine since it is not exported outside the package.

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
mattlong: I'm very wary of inaccurate casts. It'd be preferable to make some change to avoid it.

My proposal: It looks like `ListWithDetailPane` is actually already setup internally to handle `items` without an id, but its prop types require an ID. Can we loosen the generics in the signatures of `ListWithDetailPaneItemContext`, `ListWithDetailpanePropsBase`, `ListWithDetailPaneProps`, `ListWithDetailPane` to all allow missing ids? e.g.

```typescript
export function ListWithDetailPane<T extends { id?: string } = Resource>
```

With those changes, the cast can be removed along with a few corresponding type/generic changes in `ThreadDetail.tsx` and `ThreadInbox.tsx`.

mattlong (changes requested):

mattlong (approved):

PR #9571, opened 2026-06-19
Pin dockerfile dependencies

codyebberson:


codyebberson:
Sonar disagreed with this configuration.  I want to avoid these tools fighting with each other.  Closing this PR, and trying a new Renovate config: https://github.com/medplum/medplum/pull/9572

PR #9574, opened 2026-06-19
feat(provider): Include patient claims

techdavidy:
First pass of patient Claim in provider.
Claim details show corresponding claim response.

Screenshot 2026-06-19 at 1 35 21 PM


codyebberson:
@claude please review

techdavidy:
@claude review

kevinwadeshaw:
(Sorry, updated the wrong PR… hope the merge is ok!)

examples/medplum-provider/src/components/claims/ClaimBoard.tsx
```diff
@@ -0,0 +1,125 @@
...
+          ...loadSearch,
+          total: loadSearch.total ?? 'accurate',
+          include: [{ resourceType: 'Claim', searchParam: 'patient' }],
+        }),
+        { cache: 'no-cache' }
+      );
+      const claims = (bundle.entry ?? [])
+        .filter((entry) => entry.search?.mode !== 'include')
+        .map((entry) => entry.resource)
+        .filter((resource): resource is WithId<Claim> => resource?.resourceType === 'Claim');
```
noahsilas, line 66: The type here looks like it should already be `WithId<Claim> | undefined`, so all we need to do is rule out the "undefined" case. There's a helper in core called `isDefined` that should be able to do exactly that:
```suggestion
        .filter(isDefined);
```

examples/medplum-provider/src/components/claims/ClaimBoard.tsx
```diff
@@ -0,0 +1,125 @@
...
+        ...new Set(
+          claims.map((claim) => resolveId(claim.insurance?.[0]?.coverage)).filter((id): id is string => Boolean(id))
+        ),
+      ];
+      if (coverageIds.length > 0) {
+        await client.search(
+          'Coverage',
+          `_id=${coverageIds.join(',')}&_include=Coverage:payor&_count=${coverageIds.length}`,
+          { cache: 'no-cache' }
+        );
```
noahsilas, line 81: I wonder a bit about how much pre-warming the cache with this search helps. It seems like if we skip this step, then we would expect each row to end up emitting a GET query for a single Coverage resource, but the MedplumClient should hopefully be able to auto-batch that and cache each of those results individually.

While you are right that this search result will pre-warm the cache for the individually found resources, it also _won't_ itself be in cache unless the exact set of coverages it is looking for were already batched. 

For example, imagine looking at page 1 that has links to Coverage A,B, and C, and then clicking into the next page that only has links to Coverage B and C. It can't reuse the search cache, but it could have reused the individual cached resources.

This does seem to also be a pattern that comes up with some frequency, I'm starting to think we need a function like `readReferences()` that can take care of this kind of trickiness 🤔

examples/medplum-provider/src/components/claims/ClaimBoard.tsx
```diff
@@ -0,0 +1,125 @@
...
+    <Box w="100%" h="100%">
+      <ResourceBoard<Claim>
+        search={search}
+        loadItems={loadItems}
+        selectedId={claimId}
+        renderItem={(claim, { index, items }) => (
+          <ClaimListItem
+            claim={claim}
+            getClaimUri={getClaimUri}
+            hideDivider={index < items.length - 1 && items[index + 1]?.id === claimId}
```
noahsilas, line 99: I wonder if it's possible to make this use the `selected` attribute instead. It looks like the intent here is to put a divider between items, except not before the selected item, right? So what about something along these lines:
```tsx
        renderItem={(claim, { selected }) => (
          <ClaimListItem claim={claim} getClaimUri={getClaimUri} selected={selected} />
        )}
```

```css
.claimListItem + .claimListItem:not(.selected)::before {
  content: '';
  background: light-dark(var(--mantine-color-gray-1), var(--mantine-color-dark-4));
  width: calc(100% - 1rem);
  height: 1px;
  margin-left: 0.5rem;
  padding: 0 0.5rem;
}
```

examples/medplum-provider/src/components/claims/ClaimDetailPanel.tsx
```diff
@@ -0,0 +1,138 @@
...
+  // on mount and the initial state already reflects the loading/empty baseline.
+  useEffect(() => {
+    let active = true;
+
+    medplum
+      .searchResources(
+        'ClaimResponse',
+        { request: `Claim/${claim.id}`, _sort: '-_lastUpdated', _count: '1' },
+        { cache: 'no-cache' }
+      )
```
noahsilas, line 31: Could this reuse `useSearchOne()`?

examples/medplum-provider/src/components/claims/ClaimListItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface ClaimListItemProps {
+  readonly claim: Claim;
+  readonly getClaimUri: (claim: Claim) => string;
+  readonly hideDivider?: boolean;
+}
+
+export function ClaimListItem({ claim, getClaimUri, hideDivider }: ClaimListItemProps): JSX.Element {
+  // Prefer the reference display (set by createReference); only fetch when it's missing.
+  const patient = useResource(claim.patient?.display ? undefined : claim.patient);
+  const patientName = claim.patient?.display ?? (patient ? getDisplayString(patient) : undefined);
```
noahsilas, line 21: This might be something that would be a good reusable component - this seems like the behavior we want basically any time we are rendering a Reference.

PR #9579, opened 2026-06-20
Provider: Add patient Documents tab

### LINKED ISSUE #9431: Provider: Update Documents view in Patient Profile
- A unified list of uploaded files and faxes
  - Sent/received faxes that have been assigned to the Patient are included and tagged with a Fax badge
  - This dis-includes any document related to lab results (as the Labs section contains these) and also dis-includes addendums added to signed/locked Visit notes.
  - This list is also sortable by newest/oldest
- Ability to upload a new document directly from your local file system
- A Document preview for all formats with a section for metadata
- Ability to easily rename uploaded documents by clicking the title in the preview header
- Ability to download documents or fax them directly from the document preview

kevinwadeshaw:
## Summary

Adds a dedicated patient **Documents** tab to the provider example app — a master/detail view over `DocumentReference` and message/fax `Communication` resources, replacing the generic search-table view for this tab.

### Features
- **List + detail layout** — scrollable document list with a preview panel (PDF/iframe, image, and video previews; metadata for everything else).
- **Upload** — upload a file directly as a `DocumentReference`.
- **Source filtering** — filter by Fax / Lab / Upload / Message.
- **Inline edit & soft-delete** — edit title/type/category/author; delete marks the resource `entered-in-error` (kept in the project, hidden from the list).
- **Keyboard navigation** — Arrow up/down moves selection in lockstep with focus.
- **Noise filtering** — Health Gorilla lab artifacts (by category, identifier, or addendum type) are hidden from the list.

### Changes
- New components under `examples/medplum-provider/src/components/documents/`.
- New `DocumentsPage` (`pages/patient/DocumentsPage.tsx`) + styles, wired into two routes in `App.tsx` (`DocumentReference`, `DocumentReference/:documentId`).
- Documents tab search param switched from `subject=` to `patient=` in `PatientPage.utils.ts` — `patient` resolves against the same `DocumentReference.subject` field but is constrained to Patient references, matching the page's own fetch and the other patient-scoped tabs.

## Test plan
- `npx tsc --noEmit` — clean
- `npx eslint` (page, components, tests) — clean
- `npx vitest run src/pages/patient/DocumentsPage.test.tsx` — **12 passing** (empty state, loading, list render, upload, soft-delete, source filter, keyboard nav, redirect-to-first, HG artifact filtering)

### DIFF
--- a/examples/medplum-provider/src/App.tsx
+++ b/examples/medplum-provider/src/App.tsx
@@ -37,6 +37,7 @@ import { ScriptSurePage } from './pages/integrations/ScriptSurePage';
 import { MessagesPage } from './pages/messages/MessagesPage';
 import { CommunicationTab } from './pages/patient/CommunicationTab';
 import { CoveragePage } from './pages/patient/CoveragePage';
+import { DocumentsPage } from './pages/patient/DocumentsPage';
 import { DoseSpotTab } from './pages/patient/DoseSpotTab';
 import { EditTab } from './pages/patient/EditTab';
 import { ExportTab } from './pages/patient/ExportTab';
@@ -214,6 +215,8 @@ export function App(): JSX.Element | null {
                 <Route path="ServiceRequest/:serviceRequestId" element={<LabsPage />} />
                 <Route path="MedicationRequest" element={<MedicationsPage />} />
                 <Route path="MedicationRequest/:medicationRequestId" element={<MedicationsPage />} />
+                <Route path="DocumentReference" element={<DocumentsPage />} />
+                <Route path="DocumentReference/:documentId" element={<DocumentsPage />} />
                 <Route path=":resourceType" element={<PatientSearchPage />} />
                 <Route path="Coverage" element={<CoveragePage />} />
                 <Route path="Coverage/:coverageId" element={<CoveragePage />} />
--- a/examples/medplum-provider/src/components/documents/DocumentDetailPanel.module.css
+++ b/examples/medplum-provider/src/components/documents/DocumentDetailPanel.module.css
@@ -1,0 +1,18 @@
+/*
+ * One-shot fade for the "Added by …" author clause. Applied only the first time a document's
+ * original author resolves this visit; on cached revisits the clause renders without this class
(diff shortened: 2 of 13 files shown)


codyebberson (commented):
1. Thanks @kevinwadeshaw  - this is a good feature
2. @techdavidy  - let's work together to figure out how we can generalize this.  We need to find a way to be able to add pages/experiences like this without big 2k line PRs.  I'm nervous that we're accumulating way too much specialty code, and we're not generalizing enough.  That's going to make this very difficult to maintain over time.   There are a lot of good ideas and good features in here that should be moved into general purpose utilities

Some things already have existing React components that we should use:
1. `<iframe>` should be replaced with `<AttachmentDisplay>`
2. `<AttachmentPreview>` should be replaced with `<AttachmentDisplay>`
3. `handleOpenInBrowser` should be moved into `<AttachmentDisplay>`
4. `window.open()` should be replaced with `locationUtils.open()`
5. Could `DocumentFilterMenu` be generalized into a reusable menu in `<ResourceBoard>`?
6. All of `DocumentListItem.module.css‎` should be generalized
7. All of `DocumentsPage.module.css‎` should be generalized
8. `useIsTruncated` should be possible with 100% CSS and no JS (I think?)

In some of those cases, we may need to push some of the functionality into existing React components (i.e., add functionality to `<AttachmentDisplay>` to support this), but I think that's worth the effort.

I don't want to be obstructionist.  I promise that I'm trying to promote long term maintainability here.

codyebberson:
@claude - can you please review, with a particular eye to opportunities for code reuse and general purpose utilities?  At this point, I think we have hit the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) and can safely say that we need to refactor.  

@techdavidy  took a big step recently by introducing the `<ResourceBoard>` component, which I think we should lean into as a framework for creating this style of experience.

techdavidy:
@claude I think we can just search for DocumentReference given that Communication would with attachment would create a DocumentReference.

PR #9580, opened 2026-06-20
Fixed SMART app launch launchUri permissions

codyebberson:
This PR tightens access control for the `ClientApplication/$smart-launch` operation by reading the target `ClientApplication` through the authenticated user's project-scoped repository before using its `launchUri`.

Previously, this operation loaded the target `ClientApplication` through the system repository. That meant an authenticated request with a known `ClientApplication` ID could reach launch handling before the usual project-scoped read checks were applied. The returned data was limited to redirect behavior based on the configured `launchUri`, but the operation should still follow the same access-control pattern as related operations.

## Changes

- Read the target `ClientApplication` via `ctx.repo.readResource(...)` in `$smart-launch`.
- Preserve existing launch behavior for clients the requester is allowed to read.
- Add a regression test confirming a requester from one project cannot launch a `ClientApplication` from another project.

### DIFF
--- a/packages/server/src/fhir/operations/launch.test.ts
+++ b/packages/server/src/fhir/operations/launch.test.ts
@@ -73,6 +73,19 @@ describe('ClientApplication/:id/$smart-launch', () => {
     expect(launch.id).toEqual(launchId);
   });
 
+  test('Does not launch ClientApplication from another project', async () => {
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

reshmakh (approved):

PR #9582, opened 2026-06-20
Replace helm-s3 plugin with native Helm + AWS CLI

codyebberson:
Our Helm chart publishing workflow currently depends on the `helm-s3` plugin:

```bash
helm plugin install https://github.com/hypnoglow/helm-s3.git
helm s3 push ...
```

After upgrading to Helm 4, plugin installation now fails with:

```text
Error: plugin source does not support verification. Use --verify=false to skip verification
```

While we could disable verification, this introduces additional plugin-specific maintenance and keeps a third-party dependency in our release pipeline.

### Solution

Replace `helm-s3` with native Helm and AWS CLI commands.

The workflow now:

1. Packages the chart using `helm package`
2. Downloads the existing `index.yaml` from S3
3. Regenerates the repository index using `helm repo index --merge`
4. Uploads the chart archive to S3
5. Uploads the updated `index.yaml` to S3

Consumers are unaffected because they already use the public HTTPS repository:

```bash
helm repo add medplum https://charts.medplum.com
```

Only the publishing implementation changes.

### Benefits

* Removes dependency on the `helm-s3` plugin
* Avoids Helm 4 plugin verification issues
* Uses only native Helm functionality (`helm package`, `helm repo index`)
* Simplifies the release pipeline
* Reduces long-term maintenance burden and supply-chain surface area

### Notes

The repository format remains unchanged (`index.yaml` + chart archives stored in S3). Existing chart installation and upgrade workflows continue to work without modification.

### DIFF
--- a/.github/workflows/build-helm-charts.yml
+++ b/.github/workflows/build-helm-charts.yml
@@ -1,8 +1,8 @@
 name: Build Helm Charts
 
 concurrency:
-  group: ci-build-helm-${{ github.sha }}
-  cancel-in-progress: true
+  group: ci-build-helm
+  cancel-in-progress: false
 
 on:
   push:
@@ -19,6 +19,8 @@ jobs:
     timeout-minutes: 45
     permissions:
       id-token: write
+      contents: read
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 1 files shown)


codyebberson:
@claude please review

codyebberson:
@claude please re-review

codyebberson:
@claude please re-review

reshmakh (approved):

PR #9587, opened 2026-06-21
Fixes #9400 - Support External Token Exchange Across Project Memberships

### LINKED ISSUE #9400: Support Token Exchange for ProjectMemberships across different Projects
## Problem

`medplum.exchangeExternalAccessToken(token, clientId, membershipId)` should let a user exchange a token into any of their memberships, including ones in **different projects**. Today it only works for memberships in the **same project as the `ClientApplication`**, because the project and identity provider are read from the `ClientApplication` instead of from the membership.

See `exchangeExternalAuthToken` in `packages/server/src/oauth/token.ts`:

```404:415:packages/server/src/oauth/token.ts
  const projectId = await getProjectIdByClientId(clientId, undefined);
  const client = await systemRepo.readResource<ClientApplication>('ClientApplication', clientId);
  const idp = client.identityProvider;
  ...
  userInfo = await getExternalUserInfo(idp.userInfoUrl, subjectToken, idp);
```

## Expected

A user with memberships across multiple projects can pass any of their membershipIds and be authenticated into that membership's project.

## Fix

- Read identityProvider.userInfoUrl from the server config (universal external auth) instead of the ClientApplication. _Note: there could be multiple different identityProvider objects in the server config_
- When membershipId is provided, derive the project from the membership rather than the ClientApplication.

codyebberson:
This PR adds support for server-level `externalAuthProviders` in external access token exchange while preserving the existing `ClientApplication.identityProvider` behavior.

The token exchange flow now:

- First looks for an `externalAuthProviders` entry with a matching top-level `clientId`.
- Falls back to the existing `ClientApplication.identityProvider` path when no server-level provider matches.
- Derives the login project from `membership_id` only when using the server-level provider path.
- Preserves the legacy client-scoped project behavior when using `ClientApplication.identityProvider`.
- Keeps `ClientApplication` scoped to its own project and still uses it as the OAuth client for issued tokens.

The deprecated `/auth/exchange` endpoint uses the same resolution behavior via its existing `clientId` request field.


codyebberson:
@claude - please review

Please ignore Sonar's warning about lack of code coverage.  I'm more interested in correctness for now.

On this:

```
export interface MedplumExternalAuthConfig {
  // ...
  /** Optional client ID used to select this external auth provider during token exchange. */
  readonly clientId?: string;
```

Is `clientId` necessary at this level?  or could we use the existing `clientId` inside `identityProvider`?  Can those be the same, or would we need to support different values?

codyebberson:
@claude please re-review

The-Alchemist:
@cursoragent fix the merge conflict at packages/server/src/oauth/token.test.ts when merging in main

packages/server/src/oauth/token.ts
```diff
@@ -451,6 +455,110 @@ export async function exchangeExternalAuthToken(
...
+  res: Response,
+  idp: IdentityProvider,
+  subjectToken: string
+): Promise<Record<string, unknown> | undefined> {
+  try {
+    return await getExternalUserInfo(idp.userInfoUrl, subjectToken, idp);
+  } catch (err: any) {
+    const outcome = normalizeOperationOutcome(err);
+    sendTokenError(res, 'invalid_request', normalizeErrorString(err), getStatus(outcome));
+    return undefined;
```
mattlong, line 517: `return null` to match the behavior of `resolveServerExternalAuthProjectId` where null means a response was already sent?

packages/server/src/oauth/token.ts
```diff
@@ -393,53 +400,50 @@ export async function exchangeExternalAuthToken(
...
   }
 
-  let email: string | undefined = undefined;
-  let externalId: string | undefined = undefined;
-  if (idp.useSubject) {
-    externalId = userInfo.sub as string;
-  } else {
-    email = userInfo.email as string;
+  const userInfo = await tryGetExternalUserInfo(res, idp, subjectToken);
+  if (!userInfo) {
```
mattlong, line 450: Paired with the other comment, change this to `userInfo === null`

packages/server/src/oauth/token.ts
```diff
@@ -451,6 +455,110 @@ export async function exchangeExternalAuthToken(
...
+    return undefined;
+  }
+}
+
+function getExternalAuthLoginIdentity(
+  idp: IdentityProvider,
+  userInfo: Record<string, unknown>
+): { email?: string; externalId?: string } {
+  if (idp.useSubject) {
+    return { externalId: userInfo.sub as string };
```
mattlong: Are the casts in the function necessary?

mattlong (approved):

packages/server/src/oauth/token.ts
codyebberson: Good catch, this was the result of an awkward refactor to avoid Sonar complexity warnings.  I'm changing it back to inline to avoid the hoop jumping.

packages/server/src/oauth/token.ts
codyebberson: Good catch

```diff
-  userInfo: Record<string, unknown>
+  userInfo: JWTPayload
```

PR #9588, opened 2026-06-21
fix(server): persist client-credentials auth Logins to Postgres

jdjkelly:
## Why

medplum stores `client` (and `execute`) `Login` resources cache-only — Redis, never Postgres — via `isCacheOnly()` (upstream #1506). Since the cache Redis uses ElastiCache's default eviction (`volatile-lru`) and Login entries have a TTL, a still-valid Login can be evicted under memory pressure. The read path then returns `undefined`, so a valid, unexpired token gets a **401** that clients retry until expiry.

We saw exactly this happen under relatively extreme conditions with very high key eviction. Obviously reducing memory pressure is also sensible, but the number of keys being inserted is high enough to make the economics of expanding size unattractive.

## What

`client` Logins now persist to Postgres (like `password`/`google`), so an evicted cache entry is harmless — the read falls back to the DB row. Auth path unchanged. `execute` (highest-volume) Logins stay cache-only.

**Tradeoff:** reverses #1506 for `client`, so the `Login` table grows. But limited to only `client` auth method.

## Alternative

**`feat/auth-redis`** instead keeps Logins cache-only but isolates them onto a dedicated `noeviction` Redis: https://github.com/flexpa/medplum/pull/2/commits/a8283bf4ee40bf5c05cd75687f7fb512c53d9715

I can open the alternative as a PR if it's preferable

### DIFF
--- a/packages/server/src/auth/login-persistence.test.ts
+++ b/packages/server/src/auth/login-persistence.test.ts
@@ -1,0 +1,61 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { WithId } from '@medplum/core';
+import { createReference } from '@medplum/core';
+import type { ClientApplication, Login } from '@medplum/fhirtypes';
+import { randomUUID } from 'node:crypto';
+import { initAppServices, shutdownApp } from '../app';
+import { loadTestConfig } from '../config/loader';
+import { DatabaseMode, getDatabasePool } from '../database';
+import { getGlobalSystemRepo } from '../fhir/repo';
+import { deleteResourceCacheEntry } from '../fhir/repository/resource-cache';
+import { withTestContext } from '../test.setup';
+
+const systemRepo = getGlobalSystemRepo();
+
+async function countLoginRows(id: string): Promise<number> {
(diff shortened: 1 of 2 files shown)


jdjkelly:
A cleaner solution might be to just conditionally skip the Postgres writes as currently written but dependent on a configuration. Can also go that direction.

The other solution I considered was making MedplumClient resilient to this (it's currently not) but forceably re-minting an access token when an otherwise valid one 401s due to not being in the cache. It actually loops and causes an OOM right now in this scenario so might be worth considering as a supplement: https://github.com/flexpa/medplum/pull/3

jdjkelly:
Any thoughts @mattlong?

The-Alchemist:
Thanks for your patience, @jdjkelly .

I can't speak for @mattlong ,  but I believe we're trying to avoid any additional load on the writer instance of the database.  https://github.com/flexpa/medplum/pull/3 seems like a more surgical fix.  Could you make a PR out of that for us to review?

jdjkelly:
> Thanks for your patience, @jdjkelly .
> 
> I can't speak for @mattlong , but I believe we're trying to avoid any additional load on the writer instance of the database. [flexpa#3](https://github.com/flexpa/medplum/pull/3) seems like a more surgical fix. Could you make a PR out of that for us to review?

@The-Alchemist opened: https://github.com/medplum/medplum/pull/9680

PR #9589, opened 2026-06-21
fix(server):Fix flaky migrate.test.ts by isolating DW test schema

### LINKED ISSUE #9586: Flaky test in migration generator + temp data warehouse tables
I get this failed test about 25% of the time when I run all server tests on localhost.

I think it's because we're dynamically creating the `DwWorkerSyncIntTest` table, which then messes up the migration generator logic if they run concurrently.

I'm not sure what the best fix is.  I suppose we could add an exclusion check for `IntTest`?

```
 FAIL  src/migrations/migrate.test.ts (5.808 s)
  ● Generator › generateMigrationActions › returns PhasalMigration with preDeploy and postDeploy arrays

    error: relation "DwWorkerSyncIntTest_history" does not exist

      193 |   tableName: string
      194 | ): Promise<(CheckConstraintDefinition & { valid: boolean })[]> {
    > 195 |   const rs = await db.query<{
          |              ^
      196 |     table_name: string;
      197 |     conname: string;
      198 |     contype: string;

      at ../../node_modules/pg-pool/index.js:45:11
      at getCheckConstraints (src/migrations/migrate.ts:195:14)
      at getTableDefinition (src/migrations/migrate.ts:178:18)
      at buildStartDefinition (src/migrations/migrate.ts:162:17)
      at generateMigrationActions (src/migrations/migrate.ts:103:27)
      at Object.<anonymous> (src/migrations/migrate.test.ts:67:22)
```

The-Alchemist:
Run data warehouse sync integration tests in a dedicated Postgres schema instead of public so parallel migrate introspection in migrate.test.ts no longer races with table teardown. Drop the schema with CASCADE after the test for cleanup.

Fixes #9586

### DIFF
--- a/packages/server/src/workers/data-warehouse-sync.int.test.ts
+++ b/packages/server/src/workers/data-warehouse-sync.int.test.ts
@@ -5,7 +5,7 @@
  * Integration tests for `data-warehouse-sync.ts` through `processDataWarehouseSyncJob`.
  *
  * Exercises the scheduled job entry point end-to-end: server config → `getDataWarehouseSyncOptions`
- * → real `syncData` → local Parquet on disk, against `medplum_test` with a dedicated history table.
+ * → real `syncData` → local Parquet on disk, against `medplum_test` with a dedicated test schema.
  * Mocks only `getWarehouseSyncPostgresTableNames` to limit scope. Does not use BullMQ or Redis.
  *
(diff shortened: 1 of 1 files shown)


codyebberson (approved):
Thanks @The-Alchemist  🙏

PR #9593, opened 2026-06-22
Restore CORS preflight request caching

codyebberson:
Adds a 5 minute `Access-Control-Max-Age` for allowed CORS preflight requests.

This should reduce repeated browser preflight traffic for API calls without restoring the old HTTP response caching behavior that caused operational risk.

## Details

- Sets CORS `maxAge` to `600` seconds for allowed origins and paths.
- Re-enables the existing preflight max-age test.
- Updates the test to exercise the real Express app and an allowed FHIR preflight request.
- Verifies that preflight responses keep `Cache-Control: no-store, no-cache, must-revalidate`.

## Risk

Low. This uses the existing `cors` middleware's `maxAge` option, which emits `Access-Control-Max-Age` for browser preflight caching. It does not set `Cache-Control: public` or otherwise enable intermediary/shared HTTP caching of the preflight response.

### DIFF
--- a/packages/server/src/app.test.ts
+++ b/packages/server/src/app.test.ts
@@ -304,12 +304,18 @@ describe('App', () => {
     expect(await shutdownApp()).toBeUndefined();
   });
 
-  test.skip('Preflight max age', async () => {
+  test('Preflight max age', async () => {
     const app = express();
-    const res = await request(app).options('/');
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

techdavidy (approved):

PR #9595, opened 2026-06-22
fix(server): render CMS-1500 Box 24E diagnosis pointers

ajwwong:
## Problem

`getClaimPDFDocDefinition` hard-codes CMS-1500 Box 24E (diagnosis pointer) to an empty string:

```ts
// 24E. Diagnosis pointer
createText('', 335, y),
```

So every generated CMS-1500 PDF omits the per-line link between each service line and the Box 21 diagnoses that justify it — even when the Claim's `item.diagnosisSequence` is populated.

## Fix

Render the pointers from `item.diagnosisSequence` as reference letters (1 → A, 2 → B, …), entered with **no separators** per the NUCC CMS-1500 instructions (e.g. a line justified by diagnoses A and B shows as `AB`).

The mapping is extracted into a small pure `formatDiagnosisPointers()` helper, exported and unit-tested alongside the existing formatters (`formatHumanName`, `getSimplePhone`).

## Testing

Added unit tests for `formatDiagnosisPointers` covering single/multiple pointers, the A–L range, empty/undefined input, and out-of-range values. `getClaimPDFDocDefinition` itself requires an authenticated repo, so the full rendering path remains covered by the existing `claimexport` integration tests.

### DIFF
--- a/packages/server/src/fhir/operations/utils/cms1500pdf.test.ts
+++ b/packages/server/src/fhir/operations/utils/cms1500pdf.test.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import type { HumanName } from '@medplum/fhirtypes';
-import { formatHumanName, getSimplePhone } from './cms1500pdf';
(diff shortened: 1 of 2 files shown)


codyebberson (approved):

codyebberson:
Thanks @ajwwong !

ajwwong:
Thanks for the quick review, @codyebberson! 🙏

PR #9600, opened 2026-06-22
feat(react): add support for mute in useWhisper

techdavidy:
## Summary

Adds mute support to `useWhisper`. The hook now exposes `muted` state and a `setMuted(value)` toggle.

Muting works by toggling `enabled` on the captured mic track rather than tearing down the session. A disabled track emits silence, so the server-side VAD simply hears nothing — the warm WebSocket + OpenAI session and the audio worklet keep running, so unmuting is instant.


packages/react-hooks/src/useWhisper/useWhisper.ts
```diff
@@ -68,6 +70,18 @@ export function useWhisper({
   const startingCaptureRef = useRef(false);
   const sessionReadyRef = useRef(false);
   const capturingRef = useRef(false);
+  const [muted, setMutedState] = useState(false);
+  const mutedRef = useRef(false);
```
noahsilas, line 74: It looks like nothing is ever reading from this ref, can we drop it?

noahsilas (approved):

techdavidy:
@claude review

PR #9603, opened 2026-06-22
[Provider] Use FullCalendar@v7.0.0

noahsilas:
# Implementation Notes
- Dark mode support is baked in to the FullCalendar default themes
- Rendering the calendar, or the calendar's parent component, feels much snappier under FullCalendar than React-Big-Calendar.
- React-big-calendar has some confusing rendering behavior when used with the DayJS package. (issue filed here: https://github.com/bigcalendar/react-big-calendar/issues/2807). The FullCalendar rendering has some nice touches:
  - The day always starts at 12:00
  - The day defaults to being scrolled to 6:00am, hiding less commonly scheduled hours and bringing more frequently viewed hours into the default frame.
  - As a result of showing fewer hours in the default view, each slot is larger and easier to label. 
- FullCalendar v7 requires the temporal-polyfill package. This matches how we built the scheduling APIs in medplum server, so there is nice conceptual matching. It does not require `dayjs`, `luxon`, or `moment`, _all_ of which react-big-calendar lists as dependencies. The ultimate impact here is a bit muted, as it appears that `luxon` and `moment` are transitive dependencies in other places in the monorepo, and so aren't nicely pruned out of package-lock.json yet.
  - I do want to look at bundle splitting at some point, this feels like lots of code that could be loaded in a deferred way (only when someone hits the scheduling page). Not touching that today though.
- The `useCalendarController()` hook is a nice way to get an imperative handle on the calendar. It makes inlining the header toolbar quite convenient. (There's an opportunity to use the built-in header toolbar, but it didn't match the old theme as closely. If we go down this path we may investigate a more full custom theme).
- It's not always obvious how best to pull in TypeScript types from FullCalendar when writing functions like event handlers. The same was true of react-big-calendar though 🤷 
- We remove some janky height computations that I think were trying to get the old calendar to fit under a header banner that no longer exists?
- For future impovement FullCalendar supports a rich [`businessHours`](https://v7.fullcalendar.io/businessHours) feature that we can use to visualize schedule availability in the future. This is not supported out of the box for react-big-calendar (https://github.com/bigcalendar/react-big-calendar/issues/783), and mantine schedule components support only one block of business hours per day (https://mantine.dev/schedule/week-view/#business-hours-per-day).

## Notable changes to dependencies affecting bundle size:

Package | Change | Size
-- | -- | --
dayjs | removed | -23.78K
lodash | removed | -100.19K
react-big-calendar | removed | -188.86K
@popperjs/core | removed | -48.33K
temporal-polyfill | added | +46.57K
@fullcalendar/react | added | +434.73K
@full-ui/headless-calendar | added | +26.92K

## Screenshot: FullCalendar
Screenshot 2026-06-22 at 3 16 46 PM

## Screenshot: FullCalendar (Dark Mode)
Screenshot 2026-06-22 at 3 16 52 PM

## Screenshot: Previous React-Big-Query
Screenshot 2026-06-22 at 3 23 12 PM


techdavidy:
@claude review

noahsilas:
@claude Nice catch on the color scheme issue, this in fact did not work with "auto". Corrected (along with the other issues) in [bbef5e3](https://github.com/medplum/medplum/pull/9603/commits/bbef5e399383f5ca141546a96deaaae9d80b665e). Please re-check the latest version.

techdavidy (approved):

PR #9606, opened 2026-06-23
Add Patient to user invite endpoint to support RelatedPerson

maddyli:
Currently, if using the invite endpoint to invite a RelatedPerson, it will fail unless the RelatedPerson profile already exists in Medplum. 

This is because RelatedPerson resource requires a Patient (`RelatedPerson.patient`), and when the invite endpoint provisions the RelatedPerson, it does not have a Patient to tie to the RelatedPerson. As such, it must rely on matching the email to an existing RelatedPerson profile, or else it will error. 

This PR adds a `patient` field to the invite endpoint, and fails the invite for RelatedPersons (as it does currently) if not patient is provided. Furthermore, it also adds patient selection to the Invite User form in the UI for RelatedPersons. 

This is also in support of adding a RelatedPerson default access policy, where the parameter of the access policy depends upon the attached patient.


techdavidy:
@claude review

techdavidy:
is this assuming the related person is already a patient ?

maddyli:
@techdavidy it partially assumes that the RelatedPerson resource might already exist, and match to that RelatedPerson if so. RelatedPerson as a resource requires a Patient, so this aims to add that param to the invite endpoint if the RelatedPerson does not already exist as a resource and thus needs a Patient. 

> 3. UI: server "Patient is required" error won't surface under the field

This suggestion wasn't addressed because I still want the UI to be able to invite existing RelatedPerson users where the RelatedPerson already exists, and thus the Patient isn't required. Adding a patient if the RelatedPerson already exists will not override the patient, but want to be fully open regardless.

packages/server/src/admin/invite.ts
```diff
@@ -215,21 +227,55 @@ async function upsertProfileResource(
...
+        referencedPatient = await systemRepo.readReference<Patient>(patient);
+      } catch (err) {
+        // Convert notFound into a descriptive badRequest, since a bad patient
+        // reference in the request is a client error (consistent with access policies).
+        if (err instanceof OperationOutcomeError && isNotFound(err.outcome)) {
+          throw new OperationOutcomeError(badRequest(`Patient ${getReferenceString(patient)} does not exist`));
+        }
+        throw err;
+      }
+      if (referencedPatient.meta?.project !== project.id) {
```
ThatOneBro, line 246: It's worth noting this potentially leaks some information about a patient being in a project or not, but not too worried about it since you have to have the reference to the Patient to start with.

I can see the value in disambiguating for users with different projects for prod/staging/dev for example, who may try to invite a related person to the wrong project using a Patient reference who they know exists but is not found in this project.

packages/server/src/admin/invite.ts
```diff
@@ -215,21 +227,55 @@ async function upsertProfileResource(
...
+          operator: Operator.EQUALS,
+          value: email,
+        },
+      ];
+
+      // Inviting by email may match an existing profile (e.g. a previously
+      // created RelatedPerson), in which case no patient is required. Only
+      // enforce the patient requirement when a new RelatedPerson would be created.
+      if (resourceType === 'RelatedPerson' && !patient) {
+        const existing = await systemRepo.searchOne<RelatedPerson>({ resourceType, filters });
```
ThatOneBro: It's worth noting this is (sadly) non-deterministic if there are more than one `RelatedPerson` that matches the filter. So we would essentially be choosing at random if there were duplicates with the same email.

I think we should probably search for 2 and throw if results.length > 1, see [here](https://github.com/medplum/medplum/blob/ffa2fe66187a7bc6c999523108c990137b542e3d/packages/server/src/fhir/operations/rescope.ts#L155-L167) for an example. 

There is an outstanding design question about whether we should always guarantee there is only one result with `searchOne` and automatically throw if not. But that's a separate issue unrelated to this PR

ThatOneBro (commented):
Looks good for the most part, just one major question around an ambiguous case. If you wouldn't mind adding a test case for that multiple `RelatedPerson` with same email case, I think this will good to ship 👍

ThatOneBro:
@claude Can you also give this PR a review? Aside from the noted issue of the duplicate `RelatedPerson` with the same email case, can you give a verdict on how the PR is looking otherwise?

ThatOneBro (approved):
Nice, LGTM

PR #9540, opened 2026-06-17
Scheduling API: Beta release

noahsilas:
A handful of documentation updates along side tagging the scheduling API with "beta" annotations.

Adds a blog post announcing the "Beta" window.

Added some extra test coverage of a few utility functions.

Pulled in some additional documentation enhancements from #9607 - thanks @reshmakh !

PR #9607, opened 2026-06-23
Scheduling update

reshmakh:
Updating documentation based on testing.

### DIFF
--- a/packages/docs/docs/scheduling/defining-availability.md
+++ b/packages/docs/docs/scheduling/defining-availability.md
@@ -3,6 +3,9 @@ sidebar_label: Defining Availability (Alpha)
 sidebar_position: 10
 ---
 
+import ExampleCode from '!!raw-loader!@site/../examples/src/scheduling/defining-availability.ts';
+import MedplumCodeBlock from '@site/src/components/MedplumCodeBlock';
+
 # Defining Availability
 
 :::info[Alpha]
@@ -16,6 +19,8 @@ This guide covers how to configure availability using the `SchedulingParameters`
 Parameters may be defined on a HealthcareService and shared amongst all Schedules that book that type
 of appointment. Each Schedule may also override these parameters to define behaviors specific to that
 schedule as needed.
+
+A few constraints trip people up most often — see [Common Pitfalls](#common-pitfalls) for a quick reference before you start.
 
 The diagram below shows how availability can be defined at both
 - The [actor level](/docs/scheduling/defining-availability#actor-level-availability) (via Schedule)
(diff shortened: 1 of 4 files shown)


codyebberson:
@claude please review

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -304,69 +143,28 @@ The [`Schedule`](/docs/api/fhir/resources/schedule) resource is the foundation f
...
+Instead, use `comment` — a free-text field that's a good place for a human-readable description or label (for example, `"Dr. Smith's Office Visit availability"`).
+
+:::
+
+#### Choosing the actor: `Practitioner` vs `PractitionerRole`
+
+`Schedule.actor` may reference a [`Practitioner`](/docs/api/fhir/resources/practitioner), [`PractitionerRole`](/docs/api/fhir/resources/practitionerrole), [`Location`](/docs/api/fhir/resources/location), or [`Device`](/docs/api/fhir/resources/device). When the actor is a person:
+
+- Use **`Practitioner`** when availability is for the individual regardless of role or location.
+- Use **`PractitionerRole`** when availability is specific to a role, organization, or location binding (for example, when licensure varies by state — see [state-by-state licensure](/docs/scheduling/state-by-state-licensure)).
```
noahsilas, line 159: This might be something we should figure out some additional guidance for - if a single Practitioner has multiple PractitionerRole resources, each with a separate Schedule, then the underlying Practitioner can end up double-booked.

I think this is especially likely to be an issue for tele-health practice; someone might have overlapping availability via two PractitionerRole schedules representing their ability to interact with clients in multiple jurisdictions.

In a world like that, should $find check _all_ the practitioner's schedules for availability? Should $book create availability blocks on all of them? This might be application dependent, but if it is needed it should probably be inside the transaction of $book, so maybe this suggests we should have an additional parameter to include.

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -304,69 +143,28 @@ The [`Schedule`](/docs/api/fhir/resources/schedule) resource is the foundation f
...
+
+In all cases the referenced actor (Practitioner, PractitionerRole, Location, or Device) must carry the FHIR timezone extension (`http://hl7.org/fhir/StructureDefinition/timezone`). Pick one convention per example and use it consistently for both the `actor` reference and any related text.
```
noahsilas, line 161: We have relaxed this a little bit - the timezone must be defined somewhere, but it may now also be specified in places like the HealthcareService or Schedule in the SchedulingParameters extension.

We still recommend that this gets set on a per-actor basis, as this common HL7 extension is most canonical and can be useful in other contexts, but it's not strictly required there anymore.

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -1370,3 +572,31 @@ Let `$find` calculate available windows dynamically.
...
+{
+  "request": {
+    "method": "POST",
+    "url": "Organization",
+    "ifNoneExist": "identifier=http://example.org/organizations|bayview-surgery-center"
+  }
+}
+```
+
+See [Example 4](#example-4-canonical-seed-bundle--one-service-multiple-practitioners-and-rooms) for a complete, validated seed bundle.
```
noahsilas, line 592: This feels like it should probably be more general guidance in some kind of "getting started with Medplum" docs.

I think that general transactions are also still behind a feature flag (https://www.medplum.com/docs/fhir-datastore/fhir-batch-requests#batches-vs-transactions), so I'm not sure how much we want to nudge folks in this direction. Using `ifNoneExist` without a transaction is probably a pretty good middle ground: safety from duplicates when re-running, no requirement that the transaction feature be enabled.

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -1370,3 +572,31 @@ Let `$find` calculate available windows dynamically.
...
+
+## Common Pitfalls
+
+A few constraints trip people up most often when configuring availability:
+
+- **`Schedule` has no `name` element** in FHIR R4. Use `comment` for a human-readable label — sending `name` fails validation with `Invalid additional property "name"`. See [The Schedule Resource](#the-schedule-resource).
+- **Seed configuration with a `transaction` bundle, not `batch`.** A `batch` commits good entries and silently drops bad ones with no rollback, which can leave a partially-built configuration and orphaned `Slot`s. See [Seeding configuration](#4-seed-configuration-with-a-transaction-bundle-not-batch).
```
noahsilas, line 599: I don't think it's correct to say that the batch "silently drops" failed updates, I think that it will return an `OperationOutcome` for each one. It is the caller's responsibility to check within those for error results, which is a bit of extra work though.

noahsilas (commented):
I have some more thinking to do about the `PractitionerRole` support long term. I think that your description here is probably correct, but I worry about encouraging folks to use it before we've got a good story set out. That might be something that we extend the API to better support during our beta window?

I do think we should probably remove most of the comments in here about the transaction support, and consider if we can add them to a shared "setup" docs page that helps folks onboarding generally. Scheduling is only one area where this kind of multi-resource setup is useful, and having it in a single place we can link to would be great. I think there's also a fair bit of nuance in how that works, and getting it fully documented will probably take a little more work.

Thanks for transplanting all the example code into a `.ts` file - I've definitely accidentally introduced errors when it was in the `.md` and not checked!

noahsilas:
Several of the things Claude identified are already being fixed in https://github.com/medplum/medplum/pull/9540/changes#diff-d9f5f94d5f7a3f966c8f3d5038cd7afc2b2d5654c1c7cd93363dfbff15b22249 👍

noahsilas:
@reshmakh I've pulled most of this into the PR that touches some of these docs and marks the scheduling API as a beta. I've removed most of the content about transactions for now - I think that topic requires a bit more nuance and detail, so I linked to the full docs page on the topic. Let me know if you think we should try to bring more of that content back into the scheduling page.

I'm also thinking that we might want to split some "getting started" content out into a separate page - before the GA release I'll do another pass on this and see how we can make that nice and clear.

reshmakh:
Closing this in favor of @noahsilas recommendation!

PR #9611, opened 2026-06-23
fix(mock): don't reuse ID 123 for two Patients

The-Alchemist:
## Summary

- ID `123` was re-used, was a silent bug hiding inconsistency between the mock client and real client
- Assign Dr. Alice Smith `Practitioner/124` instead of reusing `123` (Homer Simpson remains `Patient/123`, Test Organization remains `Organization/123`).
- Create constants `MOCK_HOMER_PATIENT_ID`, `MOCK_ALICE_PRACTITIONER_ID`, `MOCK_TEST_ORGANIZATION_ID` for easier referencing

## Why?

Lookup tables such as `HumanName` index by `resourceId` only. The collision between `Patient/123` and `Practitioner/123` caused the practitioner's name row to overwrite the patient's during mock seeding, breaking SQL-backed name search.

### DIFF
--- a/packages/mock/src/constants.ts
+++ b/packages/mock/src/constants.ts
@@ -8,4 +8,4 @@ export const MOCK_HOMER_PATIENT_ID = '123';
 export const MOCK_ALICE_PRACTITIONER_ID = '124';
 
 /** Well-known MockClient Organization id for Test Organization. */
-export const MOCK_TEST_ORGANIZATION_ID = '123';
+export const MOCK_TEST_ORGANIZATION_ID = '125';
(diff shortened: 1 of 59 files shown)


packages/mock/src/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/** Well-known MockClient Patient id for Homer Simpson. */
+export const MOCK_HOMER_PATIENT_ID = '123';
+
+/** Well-known MockClient Practitioner id for Dr. Alice Smith. */
+export const MOCK_ALICE_PRACTITIONER_ID = '124';
+
+/** Well-known MockClient Organization id for Test Organization. */
+export const MOCK_TEST_ORGANIZATION_ID = '123';
```
noahsilas: This one seems less likely to result in surprising conflicts, but while we are doing this work to make these more distinct would it make sense to also change this ID to `125` or something similar?

noahsilas (approved):

packages/mock/src/constants.ts
The-Alchemist: good idea!  done

PR #9615, opened 2026-06-24
fix(core): flaky base64 tests

The-Alchemist:
## Summary

- Mock `getBuffer` instead of `isNodeEnvironment` in the base64 Error test
- Aligns the test with what `encodeBase64` / `decodeBase64` actually check in `base64.ts`
- Prevents flaky failures when mock return values leak from the Node.js test case
- Flaky tests occurred in high concurrency scenarios


The-Alchemist:
@claude review

ThatOneBro (approved):

PR #9619, opened 2026-06-24
[Provider] Always reveal appointment details drawer from calendar

noahsilas:
Previously, clicking on an appointment that had an encounter would navigate you directly to the encounter page. This meant that there was not a good way to interact with the appointment resource, such as if you want to cancel the appointment.

Now, selecting an appointment always starts by revealing the appointment in the AppointmentDetails drawer. That pane has a button linking to the encounter detail page when an Encounter resource already exists.

We also do some work on the styling of the AppointmentDetails drawer and the metadata revealed inside.

You may double-click on an appointment that has an associated encounter to jump directly to it.

Screenshot 2026-06-24 at 11 04 47 AM

### DIFF
--- a/examples/medplum-provider/src/components/Calendar.tsx
+++ b/examples/medplum-provider/src/components/Calendar.tsx
@@ -89,7 +89,9 @@ export function Calendar(props: {
   );
 
   // Add slight delay to click handler to permit double-clicks to register
-  const handleSelectEvent = useDebouncedCallback(handleSelectEventRaw, 100);
+  const handleSelectEventDebounced = useDebouncedCallback(handleSelectEventRaw, 100);
+
+  const handleSelectEvent = onDoubleClickAppointment ? handleSelectEventDebounced : handleSelectEventRaw;
 
   // FullCalendar creates new elements on each render rather than recycling them,
   // so dblclick listeners are cleaned up automatically when the old element is GC'd
--- a/examples/medplum-provider/src/components/Calendar.tsx
+++ b/examples/medplum-provider/src/components/Calendar.tsx
@@ -211,8 +211,10 @@ export function Calendar(props: {
           },
         }}
         eventDidMount={(info) => {
-          eventDataRef.current.set(info.el, info.event.extendedProps as ExtendedEvent);
-          info.el.addEventListener('dblclick', handleDblClick);
+          if (onDoubleClickAppointment) {
+            eventDataRef.current.set(info.el, info.event.extendedProps as ExtendedEvent);
+            info.el.addEventListener('dblclick', handleDblClick);
+          }
         }}
       />
     </div>
--- a/examples/medplum-provider/src/components/Calendar.tsx
+++ b/examples/medplum-provider/src/components/Calendar.tsx
@@ -10,12 +10,13 @@ import '@fullcalendar/react/themes/classic/palette.css';
 import '@fullcalendar/react/themes/classic/theme.css';
 import timeGridPlugin from '@fullcalendar/react/timegrid';
 import { Button, Group, SegmentedControl, Title, useComputedColorScheme } from '@mantine/core';
+import { useDebouncedCallback } from '@mantine/hooks';
 import { EMPTY, getReferenceString } from '@medplum/core';
 import type { Appointment, Slot } from '@medplum/fhirtypes';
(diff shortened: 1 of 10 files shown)


examples/medplum-provider/src/components/Calendar.tsx
```diff
@@ -86,6 +88,27 @@ export function Calendar(props: {
     [onSelectAppointment, onSelectSlot]
   );
 
+  // Add slight delay to click handler to permit double-clicks to register
+  const handleSelectEvent = useDebouncedCallback(handleSelectEventRaw, 100);
+
+  // FullCalendar creates new elements on each render rather than recycling them,
+  // so dblclick listeners are cleaned up automatically when the old element is GC'd
```
mattlong, line 97: It looks like `eventDataRef.current` grows indefinitely right now. Besides the set growing over time, wouldn't that also result in the old elements not being GC-able since something's referencing them?

examples/medplum-provider/src/components/Calendar.tsx
```diff
@@ -86,6 +88,27 @@ export function Calendar(props: {
     [onSelectAppointment, onSelectSlot]
   );
 
+  // Add slight delay to click handler to permit double-clicks to register
+  const handleSelectEvent = useDebouncedCallback(handleSelectEventRaw, 100);
```
mattlong: We don't use `dblclick` anywhere else and it isn't a common pattern on the web , so I'm a bit hesitant of introducing it here. But I'm out of the loop on provider and scheduling in general; is double clicking common in calendars? Asking since the 100ms delay will be perceptible to some folks and cause them to click again or just generally perceive the UI as being slow/laggy.

Can the debouncing be selectively applied only when a double click listener is defined?

examples/medplum-provider/src/components/Calendar.tsx
noahsilas: The advantage of a WeakMap is that it does not prevent GC of the objects it references, and when an Event unmounts the reference in the weak map becomes a candidate for garbage collection itself.

examples/medplum-provider/src/components/Calendar.tsx
noahsilas: It's pretty uncommon, and is the result of us trying to serve multiple scenarios with a single interface:
- one common case is that a practitioner is looking at their own schedule and wants to jump into the encounter an Appointment is linked to. This is how the calendar UI currently behaves.
- a use case we are introducing here is for scheduling, where you want to be able to perform operations like canceling or rescheduling a visit, and so we need some kind of detail view of the `Appointment` separate from the `Encounter` view.

There are a couple ways to address this in the UI. We could differentiate the views - maybe there should be something like a practitioner dashboard that shows upcoming appointments in a list format with a variety of navigation options, or maybe the Calendar UI becomes more modal? I'm already considering adding a mode distinguishing "scheduling" from "availability" (drag+drop to create `Slot` resources that mark time as free/busy), so maybe there's a mode that is something like "practice" that doesn't offer any of the scheduling options but just lets you dive in to the encounters?

(cc @kevinwadeshaw maybe we can spend a couple minutes together this week thinking about some of the product design aspects here?)

We definitely can only use the debounce handler when we have a double-click handler registered, that's a great idea. I'll apply that now 👍

examples/medplum-provider/src/components/Calendar.tsx
mattlong: TIL; missed that part. Neat!

mattlong (approved):

techdavidy:
@claude please review

examples/medplum-provider/src/utils/encounter.ts
```diff
@@ -280,3 +288,14 @@ export async function updateEncounterStatus(
 
   return medplum.updateResource(updatedEncounter);
 }
+
+export function encounterUrl(encounter: WithId<Encounter>): string {
+  // If the encounter subject is a Patient, deep link to the encounter
+  // inside that patient's context
+  if (isReference(encounter.subject, 'Patient')) {
+    return `/${encounter.subject.reference}/Encounter/${encounter.id}`;
```
techdavidy: Nit:  return `/${encounter.subject.reference}/${getReferenceString(enounter)}`;

techdavidy (approved):

PR #9621, opened 2026-06-24
Provider: Add Message Settings to Messages

kevinwadeshaw:
## Overview

This PR adds **Message Settings** to the provider Messages experience, allowing providers to edit a thread's Topic and Practitioner participants after it's created. Also included here is a small refresh of the **New Message** dialog and some general quality-of-life improvements for the section.

## Problem

The thread inbox let providers start a message and change its status, but once a thread existed there was no way to:
- Edit its Topic, or add/remove Practitioner participants.
- Recover gracefully when a thread had no Practitioner recipient (e.g. legacy/patient-only threads).

The New Message dialog also relied on a `QuestionnaireForm` to pick practitioners, which is now better served by our new-ish `MultiResourceInput`. The section also required a manual click to open any thread, unlike the Tasks and Faxes sections which auto-open the first list item.

## Goals

- Let providers edit a thread's **Topic** and **Practitioner participants** after creation
- Make thread creation simpler and stricter: a Patient is required; at least one Practitioner is required—the Practitioner creating the thread is auto-selected/populated as the main participant
- Match the Tasks/Faxes experience by **auto-selecting the first thread**
- Keep the inbox list stable and correct during edits (doesn't drop drafts, doesn't reorder on metadata changes).

## Non-goals

- Changing the Patient on an existing thread (the Subject is fixed after creation)
- Editing non-Practitioner participants (RelatedPerson, CareTeam, etc.) — these are preserved untouched

### 1. Message Settings dialog (`EditTopicDialog`)
- New info-icon button in the thread header opens a **Message Settings** modal
- Edit the thread **Topic** and **Practitioner** recipients; the patient is shown as read-only
- All reference fields are resolved in parallel before render so they appear together rather than popping in one at a time; a failed patient read degrades independently instead of blanking the form.
- Sender fallback: threads with no Practitioner recipient (e.g. legacy/patient-only) surface the `sender` Practitioner so the field isn't empty; saving migrates it into `recipient`.
- On save, **non-Practitioner recipients are preserved**—only the Practitioner set is replaced.

### 2. New Message dialog refresh (`NewTopicDialog`)
- Replaced the `QuestionnaireForm` Practitioner picker experience with `MultiResourceInput`, defaulting the signed-in Practitioner
- `Next` is disabled until a Patient and at least one Practitioner are selected.
- Applied a shared modal style (`MESSAGE_MODAL_STYLES`) matching the Send Fax modal—bold title, header divider, and consistent `lg` spacing around the action button.

### 3. Inbox behavior
- The first thread is auto-selected on load, mirroring the Tasks/Faxes boards. Implemented via a new optional `replace` flag on the Medplum navigate function so the implicit selection replaces history (no back-button bounce).
- **In-place thread updates** (`useThreadInbox.updateThreadParent`): editing a thread updates the list entry in place instead of refetching, so a **draft thread (no reply yet) stays in the list** after editing its settings.
- Inbox is **ordered by latest message time**, and metadata-only changes (status/topic edits) do not reorder the list

## Testing

- `EditTopicDialog.test.tsx` (new): renders after resolution, omits patient field for practitioner-only threads, saves edited topic while preserving the patient recipient, disables save with no practitioners, and the sender-practitioner fallback.
- `NewTopicDialog.test.tsx`: updated for the new label and required-patient gating.
- Full suites green: ThreadInbox 77, useThreadInbox 12; ESLint clean (one pre-existing unrelated warning).

## Commits
This PR is organized into three reviewable commits:
1. `feat(react): rework New Message dialog with MultiResourceInput`
2. `feat(react): add Message Settings dialog for threads`
3. `feat(react): auto-select first thread in inbox`

## Reference Images:

Screenshot 2026-06-24 at 1 46 54 PM

Screenshot 2026-06-24 at 1 47 19 PM


packages/react/src/chat/ThreadInbox/messageModalStyles.ts
```diff
@@ -0,0 +1,21 @@
...
+ * matching the Send Fax modal: a bold (700) title, a header with an inset bottom border
+ * acting as the divider under the header, and zero body padding so the body can lay out
+ * its own `lg` spacing around content and the bottom button.
+ */
+export const MESSAGE_MODAL_STYLES: ModalProps['styles'] = {
+  body: { padding: 0 },
+  title: { fontWeight: 700 },
+  header: {
+    padding: 'var(--mantine-spacing-md) var(--mantine-spacing-lg)',
+    backgroundImage: `linear-gradient(var(--mantine-color-gray-2), var(--mantine-color-gray-2))`,
```
noahsilas: This looks like it would match `backgroundColor: 'var(--mantine-color-gray-2)'` - is there something tricky going on with the usage of a one color gradient hre?

packages/react/src/chat/ThreadInbox/NewTopicDialog.tsx
```diff
@@ -104,77 +94,56 @@ export const NewTopicDialog = (props: NewTopicDialogProps): JSX.Element => {
...
+          <Stack gap={0}>
+            <Text fw={500}>Topic (optional)</Text>
+            <Text c="dimmed">Enter a topic for the message</Text>
+
+            <TextInput placeholder="Enter your topic" value={topic} onChange={(e) => setTopic(e.target.value)} />
+          </Stack>
+
+          <Box pt="xs">
+            <Divider />
+          </Box>
```
noahsilas: I think you can apply this style prop directly to the Divider:
```suggestion
          <Divider pt="xs" />
```

packages/react-hooks/src/useThreadInbox/useThreadInbox.ts
```diff
@@ -176,13 +186,25 @@ export function useThreadInbox({ query, threadId }: UseThreadInboxOptions): UseT
     doAdd().catch((err: Error) => setError(err));
   };
 
+  // Update an already-listed thread's parent in place (e.g. after editing its topic/participants),
+  // preserving its last-message slot. A full refetch would drop a draft thread that has no reply
+  // yet, since fetchAllCommunications only keeps threads with a message; this keeps the draft
+  // visible in the list until the user navigates away.
+  const updateThreadParent = useCallback((updated: Communication): void => {
```
noahsilas: I was a bit thrown off here - I think in the rest of this hook we have sort of been conceptualizing that a "thread" _is_ the parent container of related "message" entries. So talking about a "thread's parent" here made me imagine that there was another layer above that.

My first instinct here would be to name this something like `updateThread`, and avoid the "parent" nomenclature. What do you think?

packages/react/src/chat/ThreadInbox/EditTopicDialog.tsx
```diff
@@ -0,0 +1,185 @@
...
+ * @param onSaved - Callback fired with the updated Communication resource after a successful save.
+ */
+export interface EditTopicDialogProps {
+  thread: Communication;
+  opened: boolean;
+  onClose: () => void;
+  onSaved?: (communication: Communication) => void;
+}
+
+export const EditTopicDialog = (props: EditTopicDialogProps): JSX.Element => {
```
noahsilas: "Topic" here seems like a synonym for "Thread" - how would you feel about calling this `EditThreadDialog` to stick with the theme?

packages/react/src/chat/ThreadInbox/EditTopicDialog.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const medplum = useMedplum();
+
+  const patientRef = thread.subject as Reference<Patient> | undefined;
+
+  // Existing practitioner recipients, identified by the "Practitioner/" reference prefix
+  // (the patient subject and any other recipient types are excluded).
+  const initialPractitioners = useMemo(() => {
+    const fromRecipients = (thread.recipient ?? []).filter((r): r is Reference<Practitioner> =>
+      Boolean(r.reference?.startsWith('Practitioner/'))
+    );
```
noahsilas: We have a handy utility that we can reuse here:
```suggestion
    const fromRecipients = (thread.recipient ?? []).filter((r) => isReference<Practitioner>(r, 'Practitioner'));
```

This does make me wonder if we should create specialized versions of this per-resource to make this easier. What if we added something to core so you could do something like this?
```ts
    const fromRecipients = (thread.recipient ?? []).filter(isReference.Practitioner);
```

packages/react/src/chat/ThreadInbox/EditTopicDialog.tsx
```diff
@@ -0,0 +1,185 @@
...
+    if (fromRecipients.length > 0) {
+      return fromRecipients;
+    }
+    // Fallback for threads with no practitioner recipient (e.g. legacy/patient-only threads):
+    // surface the sender when it is a Practitioner, so the field isn't empty. Saving then
+    // migrates this practitioner into the recipient list.
+    const sender = thread.sender;
+    if (sender?.reference?.startsWith('Practitioner/')) {
+      return [sender as Reference<Practitioner>];
+    }
```
noahsilas: This can also use the same utility and avoid a cast:
```suggestion
    if (isReference<Practitioner>(sender, 'Practitioner')) {
      return [sender];
    }

File: packages/react/src/chat/ThreadInbox/EditTopicDialog.tsx [hunk 7]
```diff
@@ -0,0 +1,185 @@
...
+    // CareTeam, etc.) and replace only the Practitioner entries with the edited set,
+    // so editing the topic/practitioners never silently drops other participants.
+    const preservedRecipients = (thread.recipient ?? []).filter((r) => !r.reference?.startsWith('Practitioner/'));
+    const updated: Communication = {
+      ...thread,
+      recipient: [
+        ...preservedRecipients,
+        ...practitioners.map((practitioner) => ({
+          reference: practitioner.reference,
+        })),
```
noahsilas: `practitioners` looks like it already of type `Reference<Practitioner>[]`, can we use it directly here?
```suggestion
        ...practitioners,
```

packages/react/src/chat/ThreadInbox/EditThreadDialog.tsx
```diff
@@ -0,0 +1,185 @@
...
+              <Stack gap={0}>
+                <Text fw={500}>Patient</Text>
+
+                <ResourceInput resourceType="Patient" name="patient" defaultValue={resolved.patient} disabled={true} />
+              </Stack>
+            )}
+
+            <Stack gap={0}>
+              <Text fw={500}>Practitioner</Text>
+              <Text c="dimmed">Select one or more practitioners</Text>
```
noahsilas: As a follow up, we should consider adding a `description` prop to `ResourceInput`/`MultiResourceInput` so that you don't need to match the input styling yourself this way 🤔

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
```diff
@@ -158,6 +158,16 @@ export function ThreadInbox(props: ThreadInboxProps): JSX.Element {
...
   }, [error]);
 
+  // Auto-select the first thread when none is selected (matching the Tasks/Faxes boards).
+  // Uses replace so the implicit selection doesn't add a back-button entry that would
+  // bounce straight back into this effect.
+  useEffect(() => {
+    if (loading || threadId !== undefined || threadMessages.length === 0) {
+      return;
+    }
+    navigate(getThreadUri(threadMessages[0][0]), { replace: true });
```
noahsilas: Thanks for updating so that the `replace` option is available here - broken "back" buttons always bum me out, so this is a very nice touch to me!

noahsilas (commented):
I have a couple thoughts about some of the names in here that I think are worth considering now as we introduce this (as the long term support requirements on things in these packages are real), and then a couple minor suggestions inline. I don't see any blockers in the implementation details. Let me know what you think about the naming question and I can approve! :grin:

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
techdavidy: we would need to update ThreadInbox to use ListWithDetailPane.
So we remove the navigate from the component itself.

kevinwadeshaw: Sounds good! Updating to remove from this PR—can implement with `ListWithDetailPane` in a separate PR

packages/react/src/chat/ThreadInbox/messageModalStyles.ts
kevinwadeshaw: Good catch… it's unfortunately a trick! The backgroundImage is actually rendering the divider line below the modal header—copied from the SendFaxModal header in Faxes. Think we can leave this in place for now but should refactor both to use a proper divider line instead of backgroundImage. I can create a separate PR to address this as a fast-follow.

packages/react-hooks/src/useThreadInbox/useThreadInbox.ts
kevinwadeshaw: ✅

packages/react/src/chat/ThreadInbox/EditTopicDialog.tsx
kevinwadeshaw: ✅

packages/react/src/chat/ThreadInbox/messageModalStyles.ts
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import type { ModalProps } from '@mantine/core';
+
+/**
+ * Shared Modal styling for the thread inbox dialogs (New Message / Message Settings),
+ * matching the Send Fax modal: a bold (700) title, a header with an inset bottom border
+ * acting as the divider under the header, and zero body padding so the body can lay out
+ * its own `lg` spacing around content and the bottom button.
+ */
+export const MESSAGE_MODAL_STYLES: ModalProps['styles'] = {
```
techdavidy: this could be a module.css ?

packages/react/src/chat/ThreadInbox/EditThreadDialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+      {resolved ? (
+        <Stack gap={0}>
+          <Stack gap="lg" p="lg">
+            {/* Only show the patient when the thread has a subject. The patient cannot be changed
+                or added after creation, so practitioner-only threads omit this field entirely. */}
+            {resolved.patient && (
+              <Stack gap={0}>
+                <Text fw={500}>Patient</Text>
+
+                <ResourceInput resourceType="Patient" name="patient" defaultValue={resolved.patient} disabled={true} />
```
techdavidy: shouldn't passing the reference be enough?

packages/react/src/chat/ThreadInbox/EditThreadDialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+      recipient: [...preservedRecipients, ...practitioners],
+      topic: { text: topic },
+    };
+
+    try {
+      const saved = await medplum.updateResource(updated);
+      onSaved?.(saved);
+      onClose();
+    } catch (error) {
+      showNotification({
```
techdavidy: showErrorNotification

packages/react/src/chat/ThreadInbox/EditThreadDialog.tsx
kevinwadeshaw: The main reason for this method is to have the fields all render together—when passing references, the fields render independently and sometimes appear to "pop" on load (especially if patient/practitioner is not cached). 

I updated this to pass the reference and you can see this slight delay + pop now. Not sure if it's worth saving the LOC but will leave it to you to decide! Or maybe there's another method that can solve this?

techdavidy:
@claude review

techdavidy:
@claude in the hook when using const getLastMessageTime = (message: Communication): number => 
would this brake the pagination since it is sorting locally ?

techdavidy:
@claude review

techdavidy:
@claude review

techdavidy:
@claude review

techdavidy (approved):

PR #9625, opened 2026-06-24
Update terminology table indexes in preparation for UMLS 2026

mattwiller:
In preparation for changing the "empty" value of the `"Coding_Property".target` field from `NULL` to `-1`, this updates the partial index for reverse relationship lookups to avoid indexing rows with the new "empty" value for relationship target


codyebberson:
@claude please review

codyebberson (approved):

packages/server/src/migrations/data/data-version-manifest.json
```diff
@@ -144,5 +144,8 @@
   },
   "v38": {
     "serverVersion": "5.1.13"
+  },
+  "v39": {
+    "serverVersion": "5.1.22"
```
ThatOneBro: I think this needs to be updated

ThatOneBro (changes requested):

packages/server/src/migrations/data/data-version-manifest.json
```diff
@@ -144,5 +144,8 @@
   },
   "v38": {
     "serverVersion": "5.1.13"
+  },
+  "v39": {
```
ThatOneBro, line 148: Think we have another data manifest conflict 🙃

packages/server/src/migrations/data/data-version-manifest.json
ThatOneBro, line 149: Should we bump this to `5.1.24` too? Since we just cut that release under this PR 🫠

ThatOneBro (approved):

packages/server/src/migrations/data/data-version-manifest.json
mattwiller: Whoops, part of my change got left behind

PR #9627, opened 2026-06-24
Update `Patient/$match` to use CMS matching criteria

codyebberson:
Implements [CMS_Patient_Matching_Proposal_v3.2.2](https://docs.google.com/document/d/1f5nJsS09Sr-MzXG0f9j-eDvEwcayaiT-/edit?usp=sharing&ouid=110925489055255831269&rtpof=true&sd=true)


packages/docs/docs/api/fhir/operations/patient-match.md
```diff
@@ -4,13 +4,20 @@ sidebar_position: 35
...
+The `$match` operation implements [Master Patient Index (MPI)](https://en.wikipedia.org/wiki/Master_patient_index) patient matching. It accepts a (possibly partial) Patient resource, searches your project for candidates, and returns a `Bundle` of matches — each annotated with a score and match grade.
+
+Medplum's matching is based on the **CMS Patient Matching framework**, an evidence-based model that defines a fixed set of identity-attribute combinations strong enough to identify a patient. Rather than an ad-hoc weighted heuristic, matching is grounded in these approved combinations, with consistent normalization and conservative, constrained fuzzy matching.
+
+:::note CMS guidelines are in draft
+The CMS Patient Matching framework is a draft proposal and is **subject to change as the guidelines evolve**. The criteria table and behavior described here reflect the current draft and may be updated in future releases.
+:::
```
noahsilas: Docusaurus 3.x admonitions are a little stricter in formatting 😪 

```suggestion
:::note[CMS guidelines are in draft]

The CMS Patient Matching framework is a draft proposal and is **subject to change as the guidelines evolve**. The criteria table and behavior described here reflect the current draft and may be updated in future releases.

:::
```

codyebberson:
@claude please review

codyebberson:
@claude please re-review

packages/server/src/fhir/operations/utils/cms-patient-match.ts
```diff
@@ -0,0 +1,387 @@
...
+    exactCount,
+    fuzzyCount,
+    noneCount,
+    criteriaId: criteria?.id,
+    matchType: criteria?.matchType,
+    suffixConflict,
+  };
+}
+
+export function extractCmsMatchFields(patient: Patient): CmsPatientMatchFields {
```
noahsilas, line 153: Do you have a sense for how much this code is going to be used in a common path? If this is an infrequent operation, then it's probably fine, but if this is going to end up in a tight loop comparing many patients, then we may want to consider making this lazier.

For example, this is digging through identifiers to extract `legalId` for every input patient, but the result of that field isn't used until criteria 09, so that might be wasted cycles for common cases.

Naively, I'm imagining something along the lines of:
```ts
export function extractCmsMatchFields(patient: Patient): CmsPatientMatchFields {
  const cache: Map<string, string[]> = new Map();
  return {
    get firstName() {
      if (cache.has('firstName')) { return cache.get('firstName'); }
      const result = extractStrings(patient, 'Patient.name.given');
      cache.set('firstName', result);
      return result;
    },
  }
  ...
}
```


Edit: Nevermind, I see now that all of these fields are used to compute values like `exactCount` which are used for scoring multiple candidates, so this may be required work and adding the complexity for lazy eval + caching would itself be wasteful.

Edit 2: Ah, but the scoring isn't used if one of the criteria matches, in which case we consider it a "certain" match, so maybe we can sometimes skip a bunch of work? On the other hand, an exact match is maybe the less common case, and so it might be better to optimize for the case of an inexact match. Hmmm, maybe we can add some metrics to this later to guide us in a round of performance optimization instead of guessing the way I am now.

packages/server/src/fhir/operations/utils/cms-patient-match.ts
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Counts single-character insertions, deletions, substitutions, and adjacent
+ * transpositions. For the only threshold this framework uses (distance <= 1) the
+ * restricted form is equivalent to true Damerau-Levenshtein (§V.E.2).
+ *
+ * @param a - First string (already normalized).
+ * @param b - Second string (already normalized).
+ * @returns The edit distance between `a` and `b`.
+ */
+export function damerauLevenshtein(a: string, b: string): number {
```
noahsilas, line 352: This also seems like a good optimization target. Since we are only testing if the distance is <= 1, we could write a restricted version of this that does less work. For example, if the length of the input strings differs by more than 1, then we could stop immediately. 

I think there is a version of this algorithm that can short circuit earlier in other cases as well. For example, if `a` has a prefix of "ab" and `b` has a prefix of "12" then I think we can guarantee that the distance is > 1 without looking at any more characters, but the implementation there is way less obvious.

The-Alchemist:
Cursor review against the spec:

## Overall alignment

The PR implements the **core matching engine** the proposal describes: all **26 Table 2 combinations**, constrained fuzzy rules, generational-suffix blocking, uniqueness gating for disclosure, and basic audit logging. The docs explicitly target spec **3.2.2**.

What is implemented well:

| Proposal area | PR coverage |
|---|---|
| Table 2 combinations 01–26 | All implemented in `cms-patient-match.ts` |
| Fuzzy only on `*` fields, max 1 fuzzy field | `starred()` + per-field `allowFuzzy` |
| Damerau-Levenshtein distance ≤ 1, min 5 chars | Implemented |
| No Soundex/phonetics | Not used |
| Case/whitespace/punctuation/diacritic folding | `foldString()` |
| Full DOB only (`YYYY-MM-DD`), no imputation | `normalizeFullDate()` |
| SSN/ITIN last 4 only | `normalizeLast4()` |
| Generational suffix conflict negates match | `hasGenerationalSuffixConflict()` |
| Match all known values (names, addresses, telecom, identifiers) | FHIRPath over all entries |
| Unique candidate required for release | `onlyCertainMatches: true` + empty bundle on ambiguity/truncation |
| Namespace-bound identifiers (rule 26) | `system\|value` matching, CMS systems excluded |

---

## Gaps (by proposal section)

### III.D — IAL2 Claims Tokens (major gap)

Not implemented at all.

The proposal requires patient-facing apps to include a CSP-issued IAL2 token, with validation (JWKS signature, `iss`/`aud`/`exp`/`jti` anti-replay), and using token demographics as query input when valid.

The PR accepts demographics directly on the `Patient` resource with no token validation path.

### V — Normalization (partial gaps)

**Implemented:** string folding, email trim/lowercase, US phone digit normalization, partial DOB rejection, suffix expansion (small hardcoded map).

**Missing or incomplete:**

1. **Nickname reference table (§V.B.1)** — Explicitly out of scope per docs: no `Bob` ↔ `Robert` equivalence. Proposal requires a versioned nickname table applied before fuzzy matching.

2. **Placeholder suppression (§V.D)** — Not implemented. No tables/regexes for unknown/test/newborn names, placeholder DOBs, etc. Docs acknowledge this.

3. **Project US@ address normalization (§V.C.1)** — Street lines use `foldString()` only (strip punctuation/spaces). No US@ standardization (unit handling, USPS formatting, etc.). This weakens rule **01** (First + Last + DOB + Street Line).

4. **E.164 phone normalization (§V.C.3)** — Phones are digit-stripped (US `+1` handled), not normalized to true E.164 (`+16175550100`). International numbers are digit-only without country-code semantics.

5. **Invalid DOB ranges (§V.D.6)** — No rejection of DOBs before (current year − 120) or after (today + 2 days).

6. **MBI / Legal ID “exact” handling (Table 2 notes)** — MBI and legal IDs are run through `foldString()` (hyphens/punctuation removed). Proposal says these must match exactly (no fuzzy); deterministic normalization may be OK, but folding alters CMS MBI formatting.

7. **Legal ID scope (§III.C, rules 09/10/22/24/25)** — `legalId` is hardcoded to `https://bluebutton.cms.gov/resources/identifiers/beneficiary-id`. The proposal means DL/passport/legal ID **with issuing-authority namespace**, not only CMS beneficiary ID. Driver’s license, passport, state ID systems are not modeled.

### V.B — Name handling (minor gaps)

- **All name components separated (§V.B.2)** — Only `given` and `family` are used; prefix, middle, title are ignored (middle dismissed in proposal anyway).
- **Suffix in given name** — Suffixes only read from `HumanName.suffix`, not parsed from `"Robert Jr"` in `given`.
- **Versioned suffix table (§V.B.6)** — Small inline map, not versioned/maintained reference data.

### Core matching behavior — candidate discovery (operational gap)

The comparison logic is solid, but **candidate gathering** is narrow:

```154:184:packages/server/src/fhir/operations/patientmatch.ts
    // Strategy 1: search by identifier (strongest signal)
    ...
    // Strategy 2: search by telecom (phone/email) when present.
    ...
    // Strategy 3: search by name + birthdate.
```

Missing search strategies for:

- Address / street line (rule **01**)
- SSN/ITIN last 4
- MBI-only or legal-ID-only anchors

More importantly: searches use **raw** name values, while matching uses **normalized/fuzzy** names. A query `Robret` vs stored `Robert` may never enter the candidate set, so a valid fuzzy Table 2 match can be missed — a **false negative**, which the proposal treats as high harm.

Disclosure mode (`onlyCertainMatches: true`) depends entirely on this gather step; truncation at 100 results also suppresses release (`gather-truncated`).

### VI — Validation requirements (not in PR)

- No ONC patient-matching test dataset validation (§VI.A)
- No precision/recall/FPR reporting
- No framework-level P(collision) empirical validation (§VI.B)

These are process/compliance obligations; reasonable to defer, but they are gaps vs the full proposal.

### VII — Audit & monitoring (partial)

Disclosure mode logs an `AuditEvent` with outcome, criteria IDs, match types, matched IDs, and uniqueness — good start.

**Missing vs §VII:**

| Required | PR |
|---|---|
| Log **every** Table 2 query evaluation | Only logs when `onlyCertainMatches: true` |
| Attribute-level comparison outcomes | Not logged |
| Software/rules version for reproducibility | Not logged |
| Ongoing monitoring process (ambiguous rates, fuzzy violations, etc.) | Not implemented (org process) |

Discovery mode (`onlyCertainMatches: false`) performs no CMS audit logging.

### VIII — Governance & versioning (N/A for code)

Governance, public comment, semantic versioning of the framework itself — outside this PR’s scope.

### Other proposal principles not encoded

- **P(collision) threshold / risk framework** — Used to justify Table 2; not computed at runtime (expected).
- **Full SSN/ITIN transmission prohibition (§III.E)** — Last 4 is extracted if a full value is stored, but there is no guard rejecting full SSN/ITIN in incoming queries.
- **Network multi-responder behavior (§II.4)** — Single-system `$match`; no federated/network aggregation.
- **Disagreement in non-qualifying fields (§III.C note)** — Not explicitly handled; non-qualifying mismatches don’t block a Table 2 match (aligned with proposal intent, but no “flag for review” path).
- **DOB ± 1 day** — Listed as public comment topic only; not in v3.2.2 requirements.

---

## Summary scorecard

| Category | Status |
|---|---|
| Table 2 matching rules | **Strong** — all 26 rules, fuzzy constraints, suffix block |
| Uniqueness / disclosure gate | **Strong** — unique-only release, ambiguity suppression |
| Basic normalization | **Good** — strings, phone, email, DOB |
| Advanced normalization | **Weak** — no nicknames, placeholders, US@ addresses, true E.164 |
| Legal ID modeling | **Gap** — beneficiary-id only, not general legal ID namespaces |
| IAL2 token handling | **Missing** |
| Candidate search / false negatives | **Gap** — narrow gather + exact-name search vs fuzzy match |
| Audit / monitoring | **Partial** — disclosure-only, incomplete fields |
| Validation / governance | **Not in PR** (expected for a server feature) |

---

## Highest-priority gaps to close

If the goal is closer CMS compliance for **record disclosure**, I'd prioritize:

1. **IAL2 token validation** for patient-facing flows (§III.D).
2. **Candidate search** aligned with normalization/fuzzy matching (and anchors for address, MBI, SSN last 4).
3. **Placeholder suppression** and **nickname tables** (§V.B, §V.D) — explicitly called out in the proposal and acknowledged as missing in docs.
4. **Legal ID** support beyond beneficiary-id, with namespace-bound matching (§III.C).
5. **Project US@** address normalization for rule 01.
6. **Richer audit logging** for all disclosure evaluations, including field-level outcomes (§VII).

The PR is a solid **responder-side Table 2 matcher** for FHIR `$match`, but it is not yet a full CMS framework implementation — especially around IAL2, advanced normalization, legal ID breadth, candidate discovery, and operational validation/audit requirements.

I can go deeper on any section (e.g. a rule-by-rule test matrix, or a concrete plan for the candidate-search gap) if that would help.

packages/server/src/fhir/operations/patientmatch.ts
```diff
@@ -124,169 +153,159 @@ async function gatherCandidates(repo: Repository, input: Patient): Promise<WithI
...
+/**
+ * Logs an audit record for a CMS disclosure decision (§VII): query initiator, the Table 2
+ * combinations evaluated, matched record id(s), and the final determination. Ambiguous and
+ * truncated outcomes are flagged as minor failures so they surface in monitoring.
+ * @param ctx - The authenticated request context.
+ * @param matches - The candidates that satisfied a combination.
+ * @param truncated - Whether the candidate gather was capped (uniqueness unprovable).
+ */
+function logCmsAuditEvent(ctx: AuthenticatedRequestContext, matches: ScoredPatient[], truncated: boolean): void {
+  const released = !truncated && matches.length === 1;
```
noahsilas, line 258: Is "released" a term of art in this context?

codyebberson:
> IAL2 token validation for patient-facing flows (§III.D).

This is out of scope for this PR (requires contracts with IAL2 providers such as ID.me, Clear, etc)

> Placeholder suppression and nickname tables (§V.B, §V.D) — explicitly called out in the proposal and acknowledged as missing in docs.

CMS has not yet published these tables, so leaving that as out of scope for now.

> Legal ID support beyond beneficiary-id, with namespace-bound matching (§III.C).

Out of scope for now.  We can add more Legal ID systems in the future as needed.

> Project US@ address normalization for rule 01.

This is out of scope.  Requires third party address normalization service contracts.

> Richer audit logging for all disclosure evaluations, including field-level outcomes (§VII).

Hm, open to consider that in the future, although this will get incredibly noisy for patient match operations, so I'm inclined to pursue a more minimalist interpretation for now.

packages/server/src/fhir/operations/utils/cms-patient-match.ts
```diff
@@ -0,0 +1,387 @@
...
+
+export function normalizeLast4(value: string): string {
+  return foldString(value).slice(-4);
+}
+
+export function normalizePhone(value: string): string {
+  const digits = value.replace(/\D/g, '');
+  if (digits.length === 11 && digits.startsWith('1')) {
+    return digits.slice(1);
+  }
```
noahsilas, line 203: In the long term, it might be better to normalize to a format like E.164 instead of to a 10-digit telephone number - that would better support international usage. But normalizing phone numbers is it's own whole can of worms (and probably requires hinting from other fields to know the country code of the TN), and it seems like we could refine this later if it becomes important.

packages/server/src/fhir/operations/utils/cms-patient-match.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    expect(extractStrings({ resourceType: 'Patient', multipleBirthBoolean: true }, 'Patient.multipleBirth')).toEqual(
+      new Set()
+    );
+  });
+
+  test('matches fields exactly, fuzzily, or not at all', () => {
+    expect(matchField(new Set(['robert', 'alice']), new Set(['alice', 'robret']))).toBe('exact');
+    expect(matchField(new Set(['robert']), new Set(['robret']))).toBe('fuzzy');
+    expect(matchField(new Set(['smith']), new Set(['smyth']))).toBe('fuzzy');
+    expect(matchField(new Set(['ann']), new Set(['ana']))).toBe('none');
```
noahsilas, line 88: This might be worth breaking into its own test case for explanatory power:

```ts
test('does not allow fuzzy matching on short strings', () => {
    expect(matchField(new Set(['ann']), new Set(['ana']))).toBe('none');
});

noahsilas (approved):

packages/server/src/fhir/operations/patientmatch.ts
codyebberson: Yes, this is related to the CMS logging requirements.

Zooming out, the reason why this spec is so detailed and has so much associated research is that the end goal is to be able to say "if the user's identity is verified by an IAL2 (such as ID.me) and they match with one of these criteria, then EHRs must actually release the patient records".  That's a pretty radical change when you think about it.  That's why the term "release" is being used here, and it's why there's so much ceremony.

PR #9628, opened 2026-06-24
Token Exchange Example Scripts

finnbergquist:
Setup instructions and testing code for how to use changes in https://github.com/medplum/medplum/pull/9587

Specifically, this example should show self hosted customers how to:
- setup server config for external auth providers with token exchange
- make it easy to first get external token with external IDPs auth
- how to call 
```
const profile = await medplum.exchangeExternalAccessToken(
    externalAccessToken,
    clientId, //which server config auth provider to use
    membershipId //the membership to auth with
  );
  ```

### CODE REVIEW

**Comment by codyebberson** [comment]
@claude please review

File: examples/medplum-token-exchange-demo/.env.defaults [hunk 1]
```diff
@@ -0,0 +1,40 @@
+# Medplum server base URL.
```
codyebberson, line 1: Can we rename this to `.env.defaults`?  Trying to make all examples follow the same convention

codyebberson:
@claude please re-review

codyebberson (approved):
Probably need to fix the `EXTERNAL_` thing, but otherwise lgtm 👍

PR #9630, opened 2026-06-25
Fast fail batches with invalid `fullUrl` uuids

noahsilas:
It's tempting to try to use named entries here, but if you use something other than a valid UUID and try to reference back to it from inside other documents, we can end up persisting invalid pointers.

Instead, let's spot these early and fail the entire batch or transaction bundle.

As one example, in commit https://github.com/medplum/medplum/pull/9540/changes/238e13ea2394fdf8233f2c747b9e7b3c07f6e2c6 I rewrite some docs to use real UUIDs, because when I tried to run the sample bundle it created bad data in my project.

### DIFF
--- a/packages/fhir-router/src/batch.ts
+++ b/packages/fhir-router/src/batch.ts
@@ -175,6 +175,16 @@ class BatchProcessor {
         );
         continue;
       }
+
+      // An invalid fullUrl is a structural error that makes the bundle's reference
+      // graph undefined — fail the entire bundle so we don't persist invalid pointers
+      const fullUrl = entry.fullUrl;
+      if (fullUrl && /^urn(?::|%3A)uuid/i.test(fullUrl) && !localBundleReference.test(fullUrl)) {
+        throw new OperationOutcomeError(
+          badRequest('Invalid fullUrl: must be a valid UUID URN', `Bundle.entry[${i}].fullUrl`)
+        );
+      }
+
       const outcome = await this.preprocessEntry(entry, i, seenIdentities);
       if (outcome) {
         if (!this.isTransaction()) {
--- a/packages/fhir-router/src/batch.ts
+++ b/packages/fhir-router/src/batch.ts
@@ -29,6 +29,7 @@ const maxUpdates = 50;
 const maxSerializableTransactionEntries = 8;
 
 const localBundleReference = /urn(:|%3A)uuid(:|%3A)[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
+const exactLocalBundleReference = new RegExp(`^${localBundleReference.source}$`, localBundleReference.flags);
 
 type BundleEntryIdentity = { placeholder: string; reference: string };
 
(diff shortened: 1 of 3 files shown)


packages/fhir-router/src/batch.ts
```diff
@@ -175,6 +175,16 @@ class BatchProcessor {
         );
         continue;
       }
+
+      // An invalid fullUrl is a structural error that makes the bundle's reference
+      // graph undefined — fail the entire bundle so we don't persist invalid pointers
+      const fullUrl = entry.fullUrl;
+      if (fullUrl && /^urn(?::|%3A)uuid/i.test(fullUrl) && !localBundleReference.test(fullUrl)) {
```
The-Alchemist: could we perhaps extract this to a `isUUIDURN()` function or something like it and put it in one of the `utils.ts` files?  it's a gnarly regex, and it'll be very useful in other places too I bet, aside from just batch requests.

The-Alchemist (changes requested):

packages/fhir-router/src/batch.ts
```diff
@@ -175,6 +175,16 @@ class BatchProcessor {
...
+        throw new OperationOutcomeError(
+          badRequest('Invalid fullUrl: must be a valid UUID URN', `Bundle.entry[${i}].fullUrl`)
+        );
```
mattwiller, line 189: This would throw an error for folks who are using bundles with malformed `fullUrl` strings, even if they aren't trying to use the intra-Bundle references at all and it wouldn't have caused any issue.  Should this check be moved into e.g. `rewriteIdsInString()` to more accurately target only cases where the malformed `fullUrl` would be used?

noahsilas: My thought here was that it's difficult to tell whether an invalid URL is actually something that would be used inside the rewriting. Today we do so by testing for the URN:UUID regex, but an invalid URL wouldn't hit that. We could test all the strings against the list of invalid fullUrl entries, but that could get expensive for large batches. And, of course, if we could detect the string effectively, we could decide that we should do the replacement instead of failing (even if it is, strictly speaking, out of alignment with the spec).

What do you think? I can look at some more alternatives and see if I can come up with anything that feels better, but so far this felt like the best compromise. Maybe this is something we should do more outreach about though? Or maybe we should punt on this until the next major release when we can make it part of the upgrade guide's "breaking changes" list?

mattwiller: Fair enough, I was hoping that by moving the check into somewhere like `rewriteIdsInString()` the check would only be run on strings actively being replaced.  If that's not viable, then I think it's probably fine to ship this as-is.  Failing fast seems like a better behavior than silently not rewriting something, and the resolution if someone hits the new error is straightforward: just ensure the UUID is lowercase and/or use something like a v8 (application-specific) UUID format where most of the hex digits can be set to whatever:
```
xxxxxxxx-xxxx-8xxx-8xxx-xxxxxxxxxxxx
```

noahsilas: Yeah, unfortunately as I understand the implementation, we are iterating through the bundle and calling `rewriteIdsInString()` on every string inside of the JSON. I think the idea there is that these references can be embedded arbitrarily deeply, including as awkward sub-strings, and all of those have to be found and replaced.

That does offer some explanation for why the spec wants to be fairly precise in requiring UUIDs here - the regex at least can be compiled once and match all the potential strings, while processing arbitrary URLs would require more costly search steps. 

(All that said, running this regex on every string in a big bundle feels wasteful, but so far my thoughts on optimization are pretty under-developed and I don't feel especially confident any of them would be large improvements... might be worth spending a little time on if we see lots of large batches coming in the future though?)

packages/fhir-router/src/batch.ts
noahsilas: What do you think of the approach in [8d56ab4](https://github.com/medplum/medplum/pull/9630/commits/8d56ab496a4615731c7ac3100c3cee7634201efe)?

PR #9631, opened 2026-06-25
fix(react): display un-profiled extensions in resource views (#6999)

### LINKED ISSUE #6999: Extension fields unpopulated in table view and timeline

lolkat247:
## Summary

Fixes #6999. Extensions added to a resource were invisible in the **search/table column**, the **Timeline**, and the **resource detail view**. The generic display stack intentionally hid un-sliced extensions to avoid noise from Synthea junk data — but that also hid legitimate user extensions.

## Approach (addresses the feedback on the prior attempt, #9061)

PR #9061 was closed because simply un-hiding extensions doesn't handle resources with many/junk extensions. This PR addresses that directly, using patterns already in the codebase:

- **Render un-sliced extensions the same way profile-*sliced* extensions already render** — as inline labeled rows (`DescriptionListEntry`), with the label derived from the extension URL. No new UI concept.
- **Noise control via a `DEFAULT_IGNORED_EXTENSION_URLS` ignore-list**, mirroring the existing `DEFAULT_IGNORED_PROPERTIES` mechanism. Seeded conservatively with the common Synthea synthetic-data extensions (DALY/QALY); legitimate US-Core extensions are not suppressed. The list is easy to extend.

## Changes

- `BackboneElementDisplay`: stop hiding un-sliced extension arrays.
- `ResourceArrayDisplay`: render each un-sliced extension as a labeled row, filtered by the ignore-list. This single change fixes the **detail view** and the **table column** (both route through here).
- `ResourceDiffTable`: resolve the `extension` element definition directly from the schema so added extensions render fully in the **Timeline diff** (previously the lookup failed and dropped all but the first extension).
- New `getExtensionDisplayName(url)` helper and `DEFAULT_IGNORED_EXTENSION_URLS` constant.

## Testing

- New tests for all three surfaces (detail view, Timeline diff, table column), the ignore-list, and the URL→label helper.
- Full `@medplum/react` suite passes (1204 tests); `tsc`, lint, and api-extractor build are clean.

## Notes for reviewers

- The ignore-list is currently an internal constant (consistent with `DEFAULT_IGNORED_PROPERTIES`). Happy to make it configurable via prop/context if preferred.
- Per-URL/FHIRPath column targeting in the table is intentionally out of scope.

### DIFF
--- a/packages/react/src/BackboneElementDisplay/BackboneElementDisplay.test.tsx
+++ b/packages/react/src/BackboneElementDisplay/BackboneElementDisplay.test.tsx
@@ -101,4 +101,40 @@ describe('BackboneElementDisplay', () => {
     });
     expect(screen.getByText('Foo not implemented')).toBeInTheDocument();
   });
+
+  test('Displays un-sliced extensions as labeled rows', async () => {
+    await setup({
+      path: 'Patient',
+      value: {
+        type: 'Patient',
+        value: {
+          resourceType: 'Patient',
+          extension: [{ url: 'http://example.com/eye-color', valueString: 'blue' }],
+        },
+      },
+    });
+    expect(await screen.findByText('Eye Color')).toBeInTheDocument();
+    expect(screen.getByText('blue')).toBeInTheDocument();
+  });
+
+  test('Hides ignored extension URLs', async () => {
+    await setup({
+      path: 'Patient',
+      value: {
+        type: 'Patient',
+        value: {
+          resourceType: 'Patient',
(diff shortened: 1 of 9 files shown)


lolkat247:
Force push was to add DCO.

PR #9632, opened 2026-06-25
feat(server): improve history tombstones with meta.deleted

The-Alchemist:
## Summary

Improves the `_History` tombstone for deleted resources. Without this change, delete history rows stored an empty string, which made it difficult for the data warehouse to identify and sync deletions and caused `readVersion` on a delete version to fail when parsing empty content.

- **Minimal history tombstones** — Delete history rows now store a metadata-only stub (`resourceType`, `id`, `meta`) with `meta.deleted: true` instead of an empty string. The main resource row still uses empty `content` and `deleted: true`.
- **Tombstone metadata** — Each delete history tombstone includes `meta.versionId`, `meta.lastUpdated`, `meta.author`, `meta.project` (when present on the deleted resource), and `meta.deleted: true`.
- **Consistent versionId** — A single `versionId` is generated per delete and used for both the `_History` row and `meta.versionId` in the tombstone.
- **Preserved 410 semantics** — `readHistory` and `readVersion` still return HTTP 410 for delete versions (no resource body in history bundles). `readVersion` on a tombstone now correctly returns 410 instead of throwing on `JSON.parse('')`.
- **R5-style history entries** — Delete entries in history bundles use `request.method: DELETE` on the instance URL (`Patient/123`) rather than `GET` on the version URL.
- **Restore helpers** — Added `restore.ts` utilities for detecting deleted resources in history and building restored resources by clearing `meta.deleted`.

Delete auditing continues via `AuditEvent` (`logEvent`); no searchable `Provenance` resource is created per delete.

## Documentation updates

The [Deleting Data](https://github.com/medplum/medplum/blob/main/packages/docs/docs/fhir-datastore/deleting-data.md) doc was updated for two reasons:

1. **Internal vs public behavior** — The new tombstone is stored in `_History` for auditing and data warehouse sync, but it is **not** exposed as the `resource` body when clients read history. The doc now states this explicitly so readers do not assume `GET .../_history` returns the tombstone JSON for delete versions (FHIR requires HTTP 410 with an `OperationOutcome` instead).

2. **Restore example** — The restore script was updated to match current metadata: restoration clears `meta.deleted` rather than filtering a `DELETED` security tag from `meta.tag`. Debug logging was removed and the example types were tightened.

## What the FHIR spec requires

From [FHIR R4 HTTP](https://hl7.org/fhir/R4/http.html):

1. **Delete is logical, not physical** — history is retained; delete adds a special history entry with no content, marked as deleted.
2. **Non-version reads return 410 Gone** — `GET Patient/123` after delete → 410, not 404.
3. **Unknown vs deleted** — servers that track deletes distinguish 410 (deleted) from 404 (never existed).
4. **Version-specific reads** — `GET Patient/123/_history/{vid}` on the delete version → 410; prior versions remain readable.
5. **History includes deleted versions** — `GET Patient/123/_history` still works and includes the delete entry.
6. **Undelete** — allowed via `PUT`/`update`; may include `ETag` on 410 responses for concurrency.

R5 formalizes history bundle shape further: delete entries use `request.method: DELETE`, no `resource` body, and `response.status: 410` ([HistoryBundle profile](https://hl7.org/fhir/R5/history-bundle.html)).

## Common patterns across implementations

| Server | Soft delete | History tombstone | vread of delete version | Hard / physical delete | Audit of delete |
|--------|-------------|-------------------|-------------------------|------------------------|-----------------|
| **HAPI FHIR** | Yes (`RES_DELETED_AT`) | New version, metadata-only (`DELETED_AT`, no body) | 410 | `$expunge`, `$delete-expunge` (Smile CDR) | Server logging |
| **Azure FHIR** | Default (`hardDelete=false`) | Included in `_history` | 410 | `?hardDelete=true` (resource + all history) | Platform audit logs |
| **Oracle HDR FHIR** | Yes | New version marked deleted; `Location` header on 410 | 410 | `$expunge`, `DELETE?_expunge=true` | Server audit logs |
| **Google Healthcare API** | Yes (unless versioning disabled) | All versions in `_history`, including deleted | 410 | `Resource.purge` (history only) | Cloud audit logs |
| **Medplum** (this branch) | Yes | Minimal tombstone in `*_History` (`resourceType`, `id`, `meta.deleted`); history entry uses `request.method: DELETE` on instance URL | 410 | `expungeResource` / `expungeResources` | `AuditEvent` via `logEvent` |

Medplum's main differentiator vs HAPI/Azure is the **enriched tombstone** stored in `_History` (minimal resource JSON with `meta.deleted`) rather than purely empty history content, while still keeping the public FHIR API behavior unchanged.

## Out of Scope

Fixing the _old_ data in the `_History` tables is out of scope.  That's a separate PR at https://github.com/medplum/medplum/pull/9728


The-Alchemist:
@claude review, and make sure it matches the FHIR spec

The-Alchemist:
@claude re-review

packages/server/src/fhir/repo.ts
```diff
@@ -726,8 +731,9 @@ export class Repository extends FhirRepository implements Disposable {
       const entries: BundleEntry<T>[] = [];
 
       for (const row of rows) {
-        const resource = row.content ? this.removeHiddenFields(JSON.parse(row.content as string)) : undefined;
-        const outcome: OperationOutcome = row.content
+        const isDeleted = isDeleteTombstone(row.content as string);
```
codyebberson: `isDeleteTombstone` calls `JSON.parse()`, so we're double parsing on this line and the next line.  Might as well parse once and use the parsed value.

packages/server/src/fhir/repo.ts
```diff
@@ -745,8 +751,8 @@ export class Repository extends FhirRepository implements Disposable {
         entries.push({
           fullUrl: getFullUrl(resourceType, row.id),
           request: {
-            method: 'GET',
-            url: `${resourceType}/${row.id}/_history/${row.versionId}`,
+            method: isDeleted ? 'DELETE' : 'GET',
```
codyebberson, line 752: I think the old `'GET'` was probably wrong.

> Element Id `Bundle.entry.request.method`
> Definition	In a transaction or batch, this is the HTTP action to be executed for this entry. In a history bundle, this indicates the HTTP action that occurred.

So:
1. First entry should be `POST` for the create
2. Updates should be `PUT`
3. Deletes should be `DELETE`

In a perfect world, we would use `PATCH` if it was a patch, but i don't think we persist patch vs update in the database, so we can't represent that properly today.  Mental note for future improvements.

packages/server/src/fhir/repo.ts
```diff
@@ -804,6 +810,10 @@ export class Repository extends FhirRepository implements Disposable {
         throw new OperationOutcomeError(notFound);
       }
 
+      if (isDeleteTombstone(rows[0].content as string)) {
```
codyebberson: Same thing here, `isDeleteTombstone` calls `JSON.parse()` and we also parse on line 808/818.

For context, I know that calling `JSON.parse()` might seem like a dumb thing to worry about, but it is one of our most CPU intensive operations after input validation.

packages/server/src/fhir/repo.ts
```diff
@@ -914,6 +924,7 @@ export class Repository extends FhirRepository implements Disposable {
       lastUpdated: this.getLastUpdated(existing, validatedResource),
       author: this.getAuthor(),
       onBehalfOf: this.context.onBehalfOf,
+      deleted: undefined,
```
codyebberson, line 927: 👍

codyebberson (commented):
Nice job, all of the changes to `profiles-types.json` and `Meta.d.ts` all look perfect 👍 

I left some comments about double calling `JSON.parse()`.  We went on a campaign to tighten that up a while back, so I'd appreciate if we can hold a high bar.

The thing about `Bundle.entry.request.method` is not high priority, but if we're making changes to that line, might as well clean it up now.  That's P2 though.

The-Alchemist:
@claude re-review

packages/server/src/fhir/repo.ts
The-Alchemist: let's make it a separate PR.  i almost have it ready, but i don't wanna hold this one up.

The-Alchemist:
Followup PR for handling history entry HTTP method (`GET` vs `POST` vs `DELETE`) at https://github.com/medplum/medplum/pull/9649/

packages/server/src/fhir/repository/row-builder.ts
```diff
@@ -34,22 +34,69 @@ import { buildTokenColumns } from '../token-column';
...
+    return { tombstone: true };
+  }
+  const resource = JSON.parse(content) as Resource;
+  if (resource.meta?.deleted === true) {
+    return { tombstone: true };
+  }
+  return { tombstone: false, resource };
+}
+
+export function isDeleteTombstone(content: string | null | undefined): boolean {
```
codyebberson: This function is no longer used and can be deleted.

packages/server/src/fhir/repository/row-builder.ts
```diff
@@ -34,22 +34,69 @@ import { buildTokenColumns } from '../token-column';
...
 
+export interface DeleteHistoryContentOptions {
+  versionId: string;
+  lastUpdated: Date;
+  author: Reference;
+}
+
+export type ParsedHistoryContent = { tombstone: true } | { tombstone: false; resource: Resource };
+
+export function parseHistoryContent(content: string | null | undefined): ParsedHistoryContent {
```
codyebberson: 1. We try to avoid `null` and only use `undefined` - can we remove that here?
2. We're trying to reduce the number of one-off "helper" types, because it creates a lot of noise and slows down build time.  Claude and Codex love them, so we have to be extra vigilant to stop the onslaught.

In this case, I think this function can just return `Resource`, no?

```ts
return content ? JSON.parse(content) as Resource : { meta: { deleted: true } };
```

?

packages/server/src/fhir/repo.ts
```diff
@@ -726,8 +731,10 @@ export class Repository extends FhirRepository implements Disposable {
       const entries: BundleEntry<T>[] = [];
 
       for (const row of rows) {
-        const resource = row.content ? this.removeHiddenFields(JSON.parse(row.content as string)) : undefined;
-        const outcome: OperationOutcome = row.content
+        const parsed = parseHistoryContent(row.content as string);
```
codyebberson, line 731: We could probably just inline that whole `parseHistoryContent()` function right here


```suggestion
        const parsed = row.content ? JSON.parse(row.content as string) as Resource : { meta: { deleted: true }};
```

packages/server/src/fhir/repo.ts
```diff
@@ -726,8 +731,10 @@ export class Repository extends FhirRepository implements Disposable {
...
+        const isDeleted = parsed.tombstone;
```
codyebberson:
```suggestion
        const isDeleted = parsed.meta?.deleted;
```

codyebberson:
@claude - please look over my feedback.  Please make this PR as surgical as possible.

The-Alchemist:
@claude please look over @codyebberson 's feedback

The-Alchemist:
@claude try again

The-Alchemist:
@claude try again

The-Alchemist:
@claude review

The-Alchemist:
Deployed to staging resulting `content` JSON becomes:

```json
{
   "resourceType":"Patient",
   "id":"d81f2c86-2e63-46f0-9449-72f22db11a2c",
   "meta":{
      "versionId":"5d6f7335-1d54-4349-9a25-40f33d9f7009",
      "lastUpdated":"2026-06-29T15:00:05.734Z",
      "author":{
         "reference":"Practitioner/f5e630be-d617-4109-9bc4-f9fd2a85b4bc",
         "display":"Karl ProjectAdmin"
      },
      "deleted":true,
      "project":"24d16fc3-9aaa-42c5-aa7f-716c55151540"
   }
}
```

packages/server/src/fhir/repo.ts
```diff
@@ -2056,6 +2084,7 @@ export class Repository extends FhirRepository implements Disposable {
       meta.account = undefined;
       meta.accounts = undefined;
       meta.compartment = undefined;
+      meta.deleted = undefined;
```
codyebberson, line 2089: Good catch 👍

packages/server/src/fhir/repository/row-builder.ts
```diff
@@ -34,22 +34,49 @@ import { buildTokenColumns } from '../token-column';
...
+    id: resource.id,
+    meta,
+  });
+}
+
 export function buildDeletedResourceRow(
-  resourceType: ResourceType,
-  id: string,
-  projectId: string | undefined
+  resource: Resource,
```
codyebberson: I think we can guarantee the resource always has an `id` at this point, correct?  and then we can avoid the cast on line 73?

```suggestion
  resource: WithId<Resource>,
```

packages/server/src/fhir/repository/row-builder.ts
```diff
@@ -34,22 +34,49 @@ import { buildTokenColumns } from '../token-column';
...
+  const resourceType = resource.resourceType;
+  const id = resource.id as string;
   const columns = {
     id,
     lastUpdated,
     deleted: true,
-    projectId: projectId ?? systemResourceProjectId,
-    content,
+    projectId: resource.meta?.project ?? systemResourceProjectId,
+    content: '',
```
codyebberson: Sorry, I've forgotten where we landed on this.  We decided to punt the tombstone JSON until future commit?  What would happen if we added it now?

codyebberson: Oh oh oh.  I see.  This is the main resource table.  So, we include the deleted JSON in the `_History` table, and we use empty string in the resource table. 🤔  Ok, I think that makes sense.

packages/server/src/fhir/repo.test.ts
```diff
@@ -52,6 +55,30 @@ import * as tokenColumnModule from './token-column';
...
+}
+
+function buildRestoredResource<T extends Resource>(latestVersion: WithId<T>): WithId<T> {
+  if (!latestVersion.meta?.deleted) {
+    return latestVersion;
+  }
+  const { deleted: _, ...meta } = latestVersion.meta;
+  return { ...latestVersion, meta } as WithId<T>;
+}
+
```
codyebberson: Quick note on code conventions:
1. We try to avoid excessive helpers in test files, because it makes the tests harder to follow.  It looks like almost all of these utility functions are only used once, so probably better to just inline them in the tests directly.
2. If we do include utility functions, we put them at the bottom of test files.  That improves readability when developers come into these files.  JavaScript hoisting works fine in these cases.

codyebberson (commented):
Good job @The-Alchemist - I feel pretty good about the overall structure and core implementation here.  I think everything else is minor nits.

packages/server/src/fhir/repo.test.ts
The-Alchemist: done

packages/server/src/fhir/repository/row-builder.ts
The-Alchemist: done

packages/server/src/fhir/repository/row-builder.ts
The-Alchemist: correct, this is the main resource table.

i think one could make an argument for putting metadata in the main resource table (non-history) as well.  For symmetry, if anything else.

but i didn't want to introduce anything that would change non-`_History` tables in this PR.

codyebberson:
@claude please re-review

packages/definitions/src/fhir/r4/profiles-types.json
```diff
@@ -22029,6 +22029,22 @@
           "type" : [{
             "code" : "Reference"
           }]
+        },
+        {
+          "id" : "Meta.deleted",
```
ThatOneBro, line 22034: Should we make this an extension?

Rather than yet another field to track in meta between Medplum's meta and standard 🙃

ThatOneBro: cc @mattlong @codyebberson @mattwiller

codyebberson: > Should we make this an extension?

It's a fair question, probably out of scope for this PR.  If we go that route, we should do all of the "extended" meta fields at the same time.

This one is particularly harmless, because it's only deleted resources, which are completely undefined by the spec.

codyebberson (approved):

PR #9634, opened 2026-06-25
security(docker): upgrade Docker base images to Node 24.18

The-Alchemist:
## Summary
- Upgrade the production server Dockerfile from Docker Hardened Images `node:24.16` to `node:24.18`
- Updates both the build stage (`24.18-dev`) and runtime stage (`24.18`)

## Test plan
- no regressions


mattwiller:
This Node.js version specifically includes the [fix](https://github.com/nodejs/node/pull/64004) for the `ERR_STREAM_PREMATURE_CLOSE` [issue](https://github.com/nodejs/node/issues/63989) we had observed previously

mattwiller (approved):

PR #9635, opened 2026-06-25
Update invite user docs

AndyStoneman:


maddyli (approved):

packages/docs/docs/app/invite/admin-users-page.png
alexanderxlin, line 1: Maybe leave out our emails so we don't get people reaching out directly via those.

alexanderxlin (commented):
Looks good, left 1 comment

alexanderxlin (approved):

PR #9636, opened 2026-06-25
feat(provider): Add documents tab in patient

techdavidy:
Screenshot 2026-06-29 at 3 35 34 PM


packages/react/src/ListWithDetailPane/ListWithDetailPane.tsx
```diff
@@ -41,6 +42,8 @@ export interface ListWithDetailPaneProps<T extends { id: string } = WithId<Resou
   readonly listWidth?: number;
 
   // Header
+  /** Plain title shown at the left of the header when no `tabs` are provided. */
```
noahsilas: If we wanted to be most descriptive, we could make this props type a union so that it is valid to pass either `headerText` or `tabs` but not both. I think that might look something like this:

```tsx
export type ListWithDetailPaneProps<T extends { id: string } = WithId<Resource>> =
  | ListWithDetailPanePropsBase
  | ListWithDetailPanePropsBase & { tabs: ListWithDetailPaneTab[] }
  | ListWithDetailPanePropsBase & { headerText: ReactNode }
```

examples/medplum-provider/src/pages/patient/DocumentsPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const handleDeleted = useCallback((): void => {
+    setRefreshKey((k) => k + 1);
+    navigate(`/Patient/${patientId}/DocumentReference${location.search}`)?.catch(console.error);
+  }, [navigate, patientId, location.search]);
+
+  const patientRef: Reference<Patient> = { reference: `Patient/${patientId}` };
+
+  const headerActions = (
+    <Tooltip label="Upload document" position="bottom" openDelay={500}>
+      <ActionIcon radius="xl" variant="filled" color="blue" size={32} onClick={() => setUploadOpened(true)}>
```
noahsilas: My understanding is that the `<Tooltip>` contents are not visible to screen readers, so for best accessibility we should also add `aria-label="Upload document"` to the `<ActionIcon>` component.

examples/medplum-provider/src/pages/patient/DocumentsPage.tsx
```diff
@@ -0,0 +1,162 @@
...
+        onSelectFirst={(doc) => navigate(docUri(doc.id), { replace: true })?.catch(console.error)}
+        onChange={(s) =>
+          navigate(`/Patient/${patientId}/DocumentReference${formatSearchQuery(s)}`)?.catch(console.error)
+        }
+        onError={showErrorNotification}
+      />
+
+      <UploadDocumentModal
+        opened={uploadOpened}
+        onClose={() => setUploadOpened(false)}
```
noahsilas: #nit: If you create this state with [`useDisclosure`](https://mantine.dev/hooks/use-disclosure/), then you get a nice stable callback you can use here:
```tsx
const [uploadOpened, uploadOpenedHandlers] = useDisclosure();
<UploadDocumentModal opened={uploadOpened} onClose={uploadOpenedHandlers.close} ... />
``

File: examples/medplum-provider/src/pages/patient/UploadDocumentModal.tsx [hunk 4]
```diff
@@ -0,0 +1,142 @@
...
+            onDragLeave={() => setIsDragging(false)}
+            onDrop={(e) => {
+              e.preventDefault();
+              setIsDragging(false);
+              const dropped = e.dataTransfer.files?.[0];
+              if (dropped) {
+                setFile(dropped);
+              }
+            }}
+            className={cx(classes.dropzone, isDragging && classes.dropzoneDragging)}
```
noahsilas, line 106: This looks like it has a lot of overlap wth `SendFaxModal`, maybe there is some unification into a common"uploadModal" or extracting a shared "FileInput" component that can happen here as a follow up later.

techdavidy:
@claude please review

examples/medplum-provider/src/pages/patient/DocumentListItem.module.css
```diff
@@ -0,0 +1,31 @@
...
+
+/*
+ * Arrow-key navigation moves selection in lockstep with focus, so the selected background already
+ * marks the active option for keyboard users. Suppress the default focus ring to avoid a redundant
+ * outline stacked on top of that highlight. (Mouse clicks never showed a ring: it's :focus-visible.)
+ */
+.item:focus,
+.item:focus-visible {
+  outline: none;
+}
```
noahsilas: It seems like there's some work relating to arrow key navigation in this PR, but to me it seems a bit inscrutable - hiding this focus ring makes it hard to discern when an element in the list has focus, and the tab ordering seems surprising when I"m trying to move through. I haven't been able to actually get the arrow key navigation within the list to work in Chrome or Safari (although it _might_ be invisibly working and I just can't tell without this focus indicator?)

I wonder if `ListWithDetailPane` can encapsulate some of this work to make navigating through the list appropriately easy instead of delegating it to each instance...

examples/medplum-provider/src/pages/patient/DocumentListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const metaLine = documentType ? `${metaPrefix}: ${documentType}` : metaPrefix;
+
+  // MedplumLinkProps only types Mantine's style props, so the listbox-option DOM
+  // attributes are passed via a spread (they reach the underlying <a> through ...rest).
+  const optionProps: AnchorHTMLAttributes<HTMLAnchorElement> = {
+    id,
+    role: 'option',
+    tabIndex: isSelected ? 0 : -1,
+    'aria-selected': isSelected,
+  };
```
noahsilas: This feels like something `<MedplumLink>` should be able to understand. It looks like something about the way we are extending Mantine's `AnchorProps` is losing the fact that the underlying component is an `<a>` tag. I bet we can enhance that to also be aware of the DOM props that an anchor can take to reduce this kind of tricky workaround. (The Mantine `<Anchor>` component _does_ seem to accept these props, but the typing in there is extremely complex due to being a polymorphic component...)

examples/medplum-provider/src/pages/patient/DocumentListItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <MedplumLink
+      {...optionProps}
+      to={getItemUri(item)}
+      underline="never"
+      className={cx(classes.item, isSelected && classes.selected)}
+    >
+      <Stack gap={0} miw={0}>
+        <Tooltip label={name} multiline maw={320} withinPortal openDelay={300}>
+          <Text fw={700} truncate="end" miw={0}>
+            {name === referenceString ? 'Untitled Document' : name}
```
noahsilas, line 40: I wonder if it would make sense to teach `getDisplayString()` to take an extra fallback argument. It seems like there is a common scenario where we want to avoid showing reference strings, so maybe something like `getDisplayString(item, { fallback: 'Untitled Document' })` might be useful generally.

noahsilas (approved):
I have some potential cleanup/followup thoughts in comments. I think that the arrow-navigation / tabindex stuff is potentially quite confusing and would be good to drop from this PR for now (in a quick test, dropping the focus style overrides and `tabIndex` lets me at least tab through the items in a natural feeling way, even if the arrow-key navigation doesn't work the way it sounds like it is meant to).

PR #9638, opened 2026-06-25
chore(fax): create DocumentReference resources for attachment in fax communications

techdavidy:


packages/docs/docs/integration/efax/index.md
finnbergquist: I feel like our example code snippet should use our provider app endorsement of contentReference here.

But maybe leave an inline comment that we support both contentReference & contentAttachment

examples/medplum-provider/src/components/fax/SendFaxModal.tsx
```diff
@@ -173,7 +193,11 @@ export function SendFaxModal({
         subject: patient,
         sender: createReference(profile),
         recipient: [createReference(recipient)],
-        payload: [{ contentAttachment: attachment }],
+        payload: [
+          documentReference
```
finnbergquist, line 197: Won't documentReference always be truthy here assuming the create request succeeded? And therefore, the contentReference -> DocumentReference will always be used.

Is there a reason to include fallback logic to attachment in provider app?

finnbergquist (approved):

examples/medplum-provider/src/components/fax/SendFaxModal.tsx
finnbergquist: Might be better UX to just surface an error message if DocumentRef failed

PR #9642, opened 2026-06-25
Remove more jest dependencies

codyebberson:
Reduces `node_modules` from 1.69 GB to 1.48 GB (12% drop).

Before removing all of Jest and Babel:

image

After:

image

`node_modules` continues to be huge though 😢 

image


techdavidy (approved):

packages/dosespot-core/package.json
```diff
@@ -45,9 +45,7 @@
     "@medplum/core": "5.1.22",
     "@medplum/definitions": "5.1.22",
     "@medplum/fhirtypes": "5.1.22",
-    "@medplum/mock": "5.1.22",
-    "@types/node": "22.19.21",
-    "vitest": "4.1.9"
```
noahsilas, line 50: double-checking my understanding: removing these entries means that when these packages are used, they will get looked up from the available packages in the root `package.json`. So this change is effectively a no-op, but means that this package can't drift from the shared version in future updates, right?

noahsilas (approved):
🎉 🎉 🎉

packages/dosespot-core/package.json
codyebberson: Yep!

codyebberson:
@copilot resolve the merge conflicts in this pull request

Issue #1255, opened 2022-12-10
Access policy defaults - should we require access policies?


reshmakh:
Currently, [Access Policies](https://www.medplum.com/docs/auth/access-control) need to be created and applied to each [ProjectMembership](https://www.medplum.com/docs/api/fhir/medplum/projectmembership).  

Consider setting up default access policies for each project as part of initialization. 
Consider requiring project memberships to have an AccessPolicy

reshmakh:
Proposal: we should enable 3 default access policies for all new projects:

1. Patient - use [this](https://www.medplum.com/docs/auth/access-control#patient-access) as default access policy
2. Admin - Read/Write all project resources 
3. Practitioner - Read/write all clinical resources, but no Medplum specific resources

Todo: create policies

rahul1:
Proposed default access matrix: 
https://docs.google.com/spreadsheets/d/1xdMuS0mDPHOokJu9XZoIPpd1r_ATd1EahvI0K5P39n4/edit?usp=sharing

rahul1:
I'd put together an initial proposal recommendations for what the policies should be: 
https://docs.google.com/spreadsheets/d/1xdMuS0mDPHOokJu9XZoIPpd1r_ATd1EahvI0K5P39n4/edit?gid=0#gid=0

reshmakh:
Is there a reason we aren't using Linked Projects for this?

codyebberson:
> Is there a reason we aren't using Linked Projects for this?

That's a good question.  We could.

If we think about "AWS Managed Policies" as a reference point, then "Medplum Managed Access Policies" could be a thing that live in a linked project.  That would reduce duplication, and give us an easy pathway for making changes.

The downside is that it increases complexity, especially for self hosting customers.  We don't have a great way to setup default projects and automatically link them.  The "R4" project is a bit of a special case with a bunch of extra logic to support it.

The other consideration is whether we expect people to actually use them exactly as-is versus using them as a starting point.

maddyli:
Patient: https://github.com/medplum/medplum/pull/9263
RelatedPerson: https://github.com/medplum/medplum/pull/9643 and https://github.com/medplum/medplum/pull/9606
Practitioner and admin Practitioner: https://github.com/medplum/medplum/pull/9724

PR #9643, opened 2026-06-26
Add RelatedPerson Default Access Policy

maddyli:
- **New `Project.defaultAccessPolicies` array field** — replaces the scalar `defaultPatientAccessPolicy` pattern with a general-purpose array of `{ profileType, accessPolicy }` entries. Each entry maps a member role (`'Patient' | 'Practitioner' | 'RelatedPerson' | 'Admin'`) to an `AccessPolicy` reference. `defaultPatientAccessPolicy` is marked `@deprecated`.

- **Default RelatedPerson access policy** — on project creation, a "Default RelatedPerson Access Policy" is now provisioned alongside the existing patient policy. It uses `%patient` as a parameter placeholder, granting a RelatedPerson access to the patient's compartment resources. Both policies are registered in `defaultAccessPolicies`.

- **RelatedPerson invite applies the policy automatically** — when inviting a RelatedPerson with no explicit access policy, the server looks up the `RelatedPerson` entry in `defaultAccessPolicies` and applies it as a parameterized access rule, substituting the patient reference from the resolved `RelatedPerson.patient` field (not the request body). This correctly handles both new RelatedPerson creation and matches against existing records.

- **Patient invite respects the new array** — the patient invite path now checks `defaultAccessPolicies` for a `Patient` entry first, falling back to the legacy `defaultPatientAccessPolicy` field for projects that haven't been migrated.


packages/fhirtypes/dist/Project.d.ts
```diff
@@ -191,6 +192,27 @@ export interface Project {
...
+
+/**
+ * Default access policy to apply to project members of a given role.
+ */
+export interface ProjectDefaultAccessPolicy {
+
+  /**
+   * The member role this policy applies to.
+   */
+  profileType: 'Patient' | 'Practitioner' | 'RelatedPerson' | 'Admin';
```
codyebberson, line 210: I assume this was hand edited?  And `Admin` is not a resource type.

If we want the code generator do do this properly, then we will need to configure the `StructureDefinition` to refer to a `CodeSystem`/`ValueSet`.  That might be too much though.  There's a simpler way, which is to use the pre-existing `ValueSet` for resource types.  I can dig that up

packages/fhirtypes/dist/Project.d.ts
```diff
@@ -144,6 +144,7 @@ export interface Project {
 
   /**
    * The default access policy for patients using open registration.
+   * @deprecated Use defaultAccessPolicies instead.
```
codyebberson: This looks hand edited.  We should update the string in `packages/definitions/src/fhir/r4/profiles-medplum.json`, otherwise this will be overwritten with the next run of the generator.

packages/server/src/admin/invite.ts
```diff
@@ -396,13 +396,33 @@ async function upsertProjectMembership(
...
   // Patients only. RelatedPerson and Practitioner invites are unchanged.
   // Also applies on upsert when no policy is provided in the request.
+  // Prefers defaultAccessPolicies over the legacy defaultPatientAccessPolicy field.
+  if (request.resourceType === 'Patient' && !partialMembership.accessPolicy && !partialMembership.access?.length) {
+    const defaultPolicy = project.defaultAccessPolicies?.find((p) => p.profileType === 'Patient');
+    if (defaultPolicy) {
+      partialMembership.accessPolicy = defaultPolicy.accessPolicy;
+    } else if (project.defaultPatientAccessPolicy) {
+      // Fallback to legacy field for backwards compatibility
+      partialMembership.accessPolicy = project.defaultPatientAccessPolicy;
```
codyebberson, line 406: 👍

packages/server/src/admin/invite.ts
```diff
@@ -396,13 +396,33 @@ async function upsertProjectMembership(
...
+    !partialMembership.access?.length
   ) {
-    partialMembership.accessPolicy = project.defaultPatientAccessPolicy;
+    const defaultPolicy = project.defaultAccessPolicies?.find((p) => p.profileType === 'RelatedPerson');
+    const patientRef = (profile as RelatedPerson).patient;
+    if (defaultPolicy && patientRef) {
+      partialMembership.access = [
+        {
+          policy: defaultPolicy.accessPolicy,
+          parameter: [{ name: 'patient', valueReference: patientRef }],
```
codyebberson, line 422: I see.  This is the default behavior if nothing is submitted 🤔  

I see the pragmatism.  I don't love baking in this custom logic at this level though 🤔

packages/server/src/fhir/operations/projectinit.ts
```diff
@@ -191,7 +205,7 @@ async function createDefaultPatientAccessPolicy(
       { resourceType: 'Observation', criteria: 'Observation?_compartment=%patient' },
       { resourceType: 'Procedure', criteria: 'Procedure?_compartment=%patient' },
       { resourceType: 'QuestionnaireResponse', criteria: 'QuestionnaireResponse?_compartment=%patient' },
-      { resourceType: 'RelatedPerson', criteria: 'RelatedPerson?_compartment=%patient' },
+      { resourceType: 'RelatedPerson', criteria: 'RelatedPerson?patient=%patient' },
```
codyebberson: Hm.  I'm trying to think if there are any cases where this would cause a regression?  Nothing obvious, only pretty contrived cases, right?  

Does this support "Model 3" in our "Family Relationships" guide? packages/definitions/src/fhir/r4/profiles-medplum.json

packages/fhirtypes/dist/Project.d.ts
maddyli: Yup, it was hand edited. I don't think it would make sense for the ValueSet to be for all resource types, since really only a few can have memberships. 

My mental model is that we only want to apply default access policies to the users (RelatedPerson, Practitioner, and Patient profiles), and maybe add default ones for ClientApplications and Bots. The `Admin` option is meant to represent Practitioners that are also admins, since they will have a different set of permissions than normal practitioners, and I want the auto-provisioned first user of the Project to have a very open access policy (read/write all). 

This would mean being fairly opinionated on the initial set of default access policies, but then I plan on adding documentation about how to add more nuanced access policies, and the logic for which policies are applied there will probably exist in business logic outside of here

packages/server/src/admin/invite.ts
maddyli: Fair, though if the invite call passes in an access policy in the request, that access policy would trump this default. I think if folks are doing something more complex (like your comment below at family relationship models where patients are also relatedpersons) then either the default access policy should be edited, or the user should implement some custom business logic when inviting users. My 2c on that so far

packages/server/src/fhir/operations/projectinit.ts
maddyli: I think these lines are actually functionality equivalent. RelatedPerson is in the patient compartment if the RelatedPerson.patient is that patient. I'll change it back just to keep it straightforward

packages/fhirtypes/dist/Project.d.ts
codyebberson: You can use `"https://medplum.com/fhir/ValueSet/resource-types"`

For example:

```json
          {
            "id": "Project.defaultAccessPolicies.profileType",
            "path": "Project.defaultAccessPolicies.profileType",
            "definition": "The member role this policy applies to.",
            "min": 1,
            "max": "1",
            "type": [{"code": "code"}],
            "base": {
              "path": "Project.defaultAccessPolicies.profileType",
              "min": 1,
              "max": "1"
            },
            "binding" : {
              "strength" : "required",
              "description" : "One of the resource types defined as part of this version of FHIR.",
              "valueSet" : "https://medplum.com/fhir/ValueSet/resource-types|4.0.1"
            }
          },
```

codyebberson: >  I don't think it would make sense for the ValueSet to be for all resource types, since really only a few can have memberships.

Ok.  Let me know if you'd like assistance with this.  We will need a new `CodeSystem` and `ValueSet` in `packages/definitions/src/fhir/r4/valuesets-medplum.json` for Medplum profile types.

packages/server/src/admin/invite.ts
codyebberson: K.  I understand.  I think the "real" fix here is https://github.com/medplum/medplum/issues/1255

Until then I'm ok with this 👍

packages/fhirtypes/dist/Project.d.ts
maddyli: Yup, added those to `valuesets-medplum.json`

codyebberson:
@claude please review

codyebberson:
@claude please re-review

codyebberson (approved):
See note about remaining edit in `Project.d.ts`.  Ok to punt

PR #9660, opened 2026-06-27
Support more SMART Health Link formats

codyebberson:


codyebberson:
@claude please reivew

codyebberson:
@claude Thanks for the careful review. I addressed the two correctness/security issues and left the direct SMART Health Card payload behavior as an intentional current scope limit.

## Flag Check

Agreed that `flag` is a set of characters and direct links can combine `U` with other flags such as `L` or `P`.

Updated the external direct SHL check from exact equality with `U` to requiring that the flag contains `U`:

```ts
if (!payload.flag?.includes('U')) {
  throw new Error('Only direct SMART Health Links are supported');
}
```

I also updated the external direct-link success test to use `flag: 'LU'`, and added coverage that a missing flag is rejected before attempting to fetch the payload.

## Outbound Fetch / SSRF Hardening

Agreed on the SSRF and information disclosure risk. The external SHL URL is now passed through the existing `validateOutboundUrl()` utility before `fetch`, which matches the pattern used elsewhere in the server codebase. This enforces HTTPS by default and rejects unsafe literal hostnames/IPs such as localhost, loopback, link-local, private, reserved, and metadata-service addresses.

I also made the fetch behavior stricter:

- Added `redirect: 'error'`
- Added a 5 second `AbortSignal.timeout(...)`
- Removed the response body from non-OK error messages
- Added a 10 MiB declared `Content-Length` limit before reading and decrypting the JWE payload

One follow-up note: this PR intentionally avoids adding a one-off streaming response limiter inside the SMART Health Links operation. If we want to harden all outbound HTTP requests against servers that omit or lie about `Content-Length`, that should be a shared outbound fetch/response-reading utility and then applied consistently across the server.

The tests now cover:

- `redirect: 'error'` and timeout signal are passed to `fetch`
- HTTP error response bodies are not echoed back to the caller
- Oversized declared `Content-Length` payload responses are rejected
- Unsafe metadata IP URLs are rejected before `fetch`

## Direct SMART Health Card Payloads

Agreed that real-world direct `U` links may contain SMART Health Cards, including `application/smart-health-card` JWS payloads.

I did not expand support for that in this PR. The current `$resolve-smart-health-link` response shape returns decrypted FHIR resources, and supporting embedded SMART Health Cards would require a separate design decision about whether to verify the SHC, extract the embedded Bundle, return credential/trust metadata, or preserve the JWS. For now, external direct SHL resolution remains limited to encrypted FHIR JSON payloads.

Verification:

```bash
npm test --workspace=@medplum/server -- smarthealth.test.ts
```

reshmakh (approved):

PR #9662, opened 2026-06-28
Unified outbound URL validation

codyebberson:
Adds a small shared server utility for validating outbound URLs before `fetch()` calls that use user-controlled or configured URLs.

The new `validateOutboundUrl()` helper validates that a URL:

- parses as an absolute URL
- uses HTTPS by default
- does not use unsafe literal hostnames such as localhost, loopback, private, link-local, reserved, or multicast IP addresses

The helper supports a single `allowHttp` option for the existing insecure rest-hook compatibility setting.

## Changes

- Added `packages/server/src/util/url.ts`
  - `validateOutboundUrl()`
  - `isUnsafeHostname()`
  - `isUnsafeIpAddress()`
- Added focused unit tests in `packages/server/src/util/url.test.ts`
- Added `ipaddr.js` as a direct server dependency for IP parsing and range classification
  - `ipaddr.js` was already in `package-lock.json` as a transitive dependency
- Replaced local URL validation at outbound fetch call sites:
  - Subscription rest-hook webhooks
  - Auto-download worker
  - External auth token exchange
  - External auth userinfo / token exchange userinfo
  - SMART Health Card issuer JWKS lookup
- Removed the local SMART Health Card `isUnsafeHostname()` implementation

## Notes

The auto-download worker still preserves its existing distinction between internal Medplum URLs and external URLs. Internal Binary/storage URLs are excluded before applying outbound URL validation to candidate external URLs.

This PR intentionally does not add DNS resolution or reachability checks. Those are useful follow-up protections, but they introduce network behavior, latency, and deployment tradeoffs that should be considered separately.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46345,6 +46345,7 @@
         "hibp": "14.1.3",
         "http-graceful-shutdown": "3.1.16",
         "ioredis": "5.8.2",
+        "ipaddr.js": "2.4.0",
         "jose": "5.10.0",
         "jszip": "3.10.1",
         "node-fetch": "2.7.0",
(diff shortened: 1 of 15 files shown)


reshmakh (approved):

codyebberson:
@claude please review

codyebberson:
@claude please re-review

codyebberson:
@claude please re-review

PR #9664, opened 2026-06-28
Use vendored patch lib

codyebberson:
In https://github.com/medplum/medplum/pull/9335 we vendored the `rfc6902` library to enable custom [FHIRPatch](https://build.fhir.org/fhirpatch.html) semantics.

This PR replaces all references to `rfc6902` with `@medplum/core` and removes the dependency.

The patch methods were not previously exported and therefore not previously run through `api-documenter`, so this also required some cosmetic changes to JSDoc comments and renaming `Options` to `PatchOptions` to avoid naming collision.


codyebberson:
@claude - please review

reshmakh (approved):

PR #9665, opened 2026-06-28
Add SMART Health Link import feature to Provider app

codyebberson:


codyebberson:
@claude please review

codyebberson:
@claude please re-review

codyebberson:
@claude 

> 🟡 Still open — a failed $match surfaces an error over a usable state

In this case, we know that `Patient/$match` will be available and will not fail.  It will either succeed with a match or succeed with empty results.

> Minor / nits (unchanged, non-blocking)
> SmartHealthLinkImportPage.tsx:175 — setImportResult(result) immediately followed by navigate(...) still unmounts before the success Alert at line 367 can render, so that Alert is effectively dead code. If the redirect is the intended UX, consider dropping the Alert block (or vice-versa).

This application uses the Mantine notification system, which has persistent notifications across page transitions, so the `navigate()` call will not clear the notification toast.

> getResourceSummary (utils 98–105) and getResourceDate (213–222) still duplicate the same effectiveDateTime ?? issued ?? … fallback chain — could share one helper.

Good feedback, saving this for future work.

> New edge case from the refactor: if a referenced resource is deselected, its reference (e.g. DiagnosticReport.result → Observation/xyz) stays a literal ResourceType/id and isn't a urn:uuid: in the transaction, so the server will treat it as an existing-resource reference. Same behavior as before the refactor, just noting it — a deselect that breaks a kept resource's reference could fail the transaction. Probably acceptable for an example app.

Good feedback, saving this for future work.

codyebberson:
@claude - can you please do one final check after the last cleanup commits?

codyebberson:
@claude Thanks for the follow-up review. I addressed the resource identity issue and checked the `shlink:/` normalization behavior.

## Bundle Entry Keys

Agreed that the previous refactor made `resource.id` a firmer runtime assumption than intended. The import cart now keeps the original `BundleEntry` objects instead of flattening immediately to `Resource[]`, so selection state can use the bundle entry identity.

I added a small shared helper:

```ts
export function getSmartHealthLinkBundleEntryKey(entry: BundleEntry): string | undefined {
  return entry.fullUrl ?? (entry.resource ? getReferenceString(entry.resource) : undefined);
}
```

The page now uses that helper consistently for initial selection, React row keys, checkbox state, select-all, and import filtering. `buildSmartHealthLinkImportBundle()` also uses the same key helper, so the UI and transaction builder agree on what was selected.

I added a focused regression test for an id-less resource selected by `entry.fullUrl`.

## Wrapped SMART Health Link URLs

I confirmed that the server already handles viewer URLs that wrap the SHL in a fragment, such as:

```text
https://viewer.example/launch#shlink:/...
```

The provider page sends the input to `$resolve-smart-health-link`, and the server resolve path calls `parseSmartHealthLink()`. That parser already extracts `#shlink:/` fragments before validating the payload:

```ts
const fragmentIndex = input.indexOf('#shlink:/');
const raw = fragmentIndex === -1 ? input : input.substring(fragmentIndex + 1);
```

There is also existing core test coverage for this behavior, including generic viewer-fragment parsing and HL7 example viewer URLs. So I left the client-side `.trim()` behavior as-is.

reshmakh (approved):

PR #9408, opened 2026-06-03
Redesign /products page: three-tier framework + workflows section

everett-williams:
## Summary

Redesigns `/products` around a three-tier story — **Apps → Capabilities → Foundations**, all on one FHIR data model — with an interactive architecture diagram for the foundations.

- **Apps**: your custom app, plus reference apps (Medplum Provider, Medplum App).
- **Capabilities**: 8 clinical capabilities, each shipping with pre-built UIs.
- **Foundations**: 10 building blocks, explorable in an interactive diagram, over a Compliance & Certification band.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9670, opened 2026-06-29
docs(products): redesign /products page

kevinwadeshaw:
This builds off of Everett's initial commit here (#9408), our latest discussion, and the copy doc [here](https://docs.google.com/document/d/1hrjpxYsXdZw9Qao0OcPfT-LUcNVBNk1L0zcTLzNX0q8/edit?usp=sharing).

---

### Major changes and challenges with this version include:

- Replacing hero copy and image
    - Current image is a conceptual placeholder but genenerally should move away from a code block being the first thing our audience sees
- Updated copy throughout the site, per the most recent copy doc, with some additional tweaks
- Removing 3 tier “data model” section
    - (These products aren’t really the data model and are redundant with the rest of the content on the page)
- Removed custom application from Apps section:
    - This isn’t a real product that we offer, but think we should lead with this as the CTA at the bottom of the page (not implemented yet)
- Reformatted Apps into a single, tabbed component
    - This sets us up to expand tabs as other Apps roll out
    - Should we have different icons to represent each app? 
    (This is a brand hierarchy question and tied to naming)
- Reformatted Capabilities to be cards instead of vertical tabs
    - Most of these are table stakes so the images aren’t going to be that interesting—definitely not worth the effort to click through each or to view a larger than thumbnail
    - The size of these cards now also starts to add a visual hierarchy to the page: Apps are big products, Capabilities are smaller building blocks for those, etc.
- For the Foundations section, I’ve tried simplifying the diagram layout and interactions here but this still isn’t resolved
    - The main issue is that the Apps and Capabilities sections are now straightforward  representations but the diagram for Foundations makes the products here more difficult to parse—i.e. it’s difficult to tell what exactly a Foundation is here.. Why are we seeing the Apps again? Are Integrations Foundations? Etc. The numbers help, but it’s still not as clear as it could be
    - I can propose some alternate ways of handling this but wanted to put it out there for feedback first


reshmakh:
This is exciting and I'm on board with the overall direction. Agree we should handle the new naming as a separate step so we don't block this.

A few things on this version:

**Capitalization.** We're using capital-case for Apps, Capabilities, and Foundations as if they're formal, named concepts. I'd push back on that. It introduces more terms for the reader to learn and makes the page feel heavier than it needs to be. Related: the hero line "build on the strongest foundations in healthcare" reuses capital-F Foundations, which reads as a reference to the section and is confusing. I'd lowercase these unless we've genuinely decided they're proper product tiers (and that's the naming conversation we're deferring).

**Where do patient experiences go?** It's not clear where patient experiences live in this structure. My guess is Capabilities, but wherever it lands, it should be represented somewhere on the page.

**Foundations diagram.** On the diagram you flagged as unresolved: I have a slight preference for the earlier version Everett had (Brian liked it too). The current treatment reads as a list, whereas Everett's gave a sense of how the pieces fit together, even if it was a bit of a cartoon. That relational quality is what makes a diagram worth having over a list. Was the current approach a deliberate move away from that, or were we looking too dev-y on it? Want to make sure we don't lose the "how it fits together" view.

**Small copy fix.** Should be "Compliance and certifications" (certifications plural).

Generally looks good.

everett-williams:
I generally agree with Reshma's feedback on capitalization and the foundations diagram (IMO the current version is harder to parse / understand). I also still wonder a bit on **whether it's the right call to remove custom applications from the apps section.** I recognize that's not necessarily a 'product' of ours, but I do want to make it clear that we are not an out of the box solution, and our main value prop is building your own app so IMO it's important to recognize in this section. Curious what others think.

I'm also excited to hear what you are planning for the hero image!

kevinwadeshaw:
@reshmakh @everett-williams Thanks for the thoughts here! A few notes and responses below:

**Category Capitalization**
I’ll update to treat these product categories as lowercase—can revisit later as part of the naming discussion, if we want to.

**Representing Patient Experience**
It should be mostly implicit throughout the page that most of the capabilities and foundations can be used to support patient-facing experiences—e.g. Messaging/Scheduling/etc. can also be patient-facing. Though, this could be more explicit and I’ll try making some changes to copy or visuals to better communicate this.

**Foundations & Diagram**
The current approach was based on trying to clarify the exact foundational pieces in the diagram, since it’s also representing concepts other than the foundations. I think you actually said it though: this diagram is meant to show “how it all fits together” which is different from the goal of communicating “what are the Medplum foundations” in this section. I wonder if we should break these up again (🙃), so this section is just for showing/listing the foundations—and not as part of a diagram. And then maybe we have a separate section below that’s more about “how these products fit together” with the diagram as the main visual (referencing the apps, capabilities, and foundations above).

**Custom Applications**
I’ll try adding a callout for this in the Apps section, but separate from our own Medplum apps. I’ll also try a version as a larger section toward the bottom of the page—similar to a “how it fits together” section, I think it’s helpful to see the different products and then see a “build your own custom apps” section that can reference the apps, capabilities, and foundations above. Potentially even showing ‘recipes’ of how to combine these products to create custom apps for different solutions.

—

I’ll keep going on this but @codyebberson please let us know if you have any thoughts on these and/or other feedback. The latest version we're referencing can be found here: https://medplum-www-git-kevin-products-page-medplum.vercel.app/products

kevinwadeshaw:
@reshmakh @everett-williams @codyebberson  

Here's the latest on the Products page:
https://medplum-7z6kr7o3e-medplum.vercel.app/products

And here's what changed from the last version:

**Global changes**
- Updated copy throughout to make apps, capabilities, and foundations all lowercase
- Updated copy in key moments to reference patient experience/engagement

**Hero section**
- Made the top navigation bar match the homepage: starts clear and only gains shadow on scroll
- Updated image to feature Provider app with floating tags that circle around to indicate active capabilities + foundations
- Note: if this looks alright, I’ll touch up these tags to be less visually heavy

**Foundations section**
- Updated the foundations section to strip out diagram and instead be accordion-based slideshow
- Representational images of each foundations including visual of code
- Updated copy of header description and copy of individual foundations

**(New) "How it all works together" section**
- Brought back the original architecture diagram and restyled to match the other page elements
- New headline and copy

**Transformed bottom CTA to focus on custom apps**
- New headline, description, and button copy
- Rotating images of (speculative) custom apps (2 currently but will add more for breadth)

Please let me know what thoughts or feedback you have on these latest changes—thanks!

everett-williams:
Thanks @kevinwadeshaw for all these updates! I like the direction you've taken. One final point thought below.

For the foundations section - I like the new visuals, but do you think some of them should be larger / take up a bit more of the blue box?

I think the new diagram format works but think readability could potentially be improved by lightening fill colors, increasing text and icon size, and potentially resizing the custom apps vs. medplum apps to be more even.

kevinwadeshaw:
@reshmakh @codyebberson @everett-williams @maddyli — 

I just added one last small commit here to cover some outstanding items:

- Tweaks to fully support dark mode, using patterns we established on the home page
- Updates to support mobile viewports: stacking buttons, horizontal scrolling for diagram, text size adjustments, etc.
- Per Everett's diagram feedback: increased text sizes for headers and labels, larger icon tiles, and more prominent Your Apps section
- Updated hero image, animation, and tags styling
- Updated existing speculative app images in Custom Apps section and added one more image of a patient-facing app

@everett-williams I didn't update the foundations images since these are only illustrative mock-ups and we don't want people to read too much into them. If we have ideas of how to make these more 'real' we can update as a fast-follow.

Please let me know if you have any final thoughts or questions on this—here's the link to review:
https://medplum-k5wk8p4m3-medplum.vercel.app/products

Otherwise, I think it's pretty much ready to publish!

maddyli (approved):
so exciting!

everett-williams (approved):
Look good to me - and sounds good regarding the visuals for foundations!

codyebberson:
@claude - please review.  The content looks good.  I'm primarily interested in implementation as a Docusaurus page and Docusaurus components.  Any redundancy that can be removed?  Are we following Docusaurus styles best practices?  any existing styles or components that we could reuse to help slim this down?

everett-williams:
@claude please re-review. Note that I skipped updating #4 and #6 given there's limited value to pursuing these and they would risk the visuals / animations. Net impact of updates was roughly 380 source lines saved with no visual changes.

PR #9672, opened 2026-06-29
perf(server): cache parsed patient-compartment FHIRPath expressions

jdjkelly:
## Summary

`getPatients()` runs on every resource write (to compute `meta.compartment`). For resources in the patient compartment it evaluates a FHIRPath expression for each compartment search parameter, and it parsed each expression from scratch on every call:

```ts
const values = evalFhirPath(searchParam.expression as string, resource); // re-tokenizes + re-parses every call
```

The set of compartment expressions is fixed, so the same handful were re-parsed for every such resource. This PR caches the parsed `FhirPathAtom` per expression string.

The benefit therefore scales with the share of patient-compartment writes. Nearly all of a clinical chart, but none of, say, a terminology / `StructureDefinition` import.

## Why

Profiling a bulk transaction insert (a 500-resource patient chart) showed FHIRPath `tokenize` + `parse` accounting for ~5% of active CPU, traced entirely to this call site.

## Impact

[We built a benchmark harness to performance profile batch inserts](https://github.com/flexpa/medplum/pull/7). It runs a 500 resource batch built from Synthea. 

This PR was benchmarked with three runs of 25 timed iterations each against `main` and against this branch (this commit cherry-picked onto the same harness), run back-to-back to control for machine state.

Averages:

| Metric | `main` | `main` + this PR | Δ |
|---|---|---|---|
| **CPU / transaction** | 474.3 ms | 417.3 ms | **−12.0%** |
| Wall / transaction | 1149.3 ms | 1067.1 ms | −7.2% |
| Throughput | 435 res/s | 469 res/s | +7.8% |

<details><summary>Per-run medians</summary>

| Run | main CPU (ms) | main wall (ms) | +PR CPU (ms) | +PR wall (ms) |
|----:|----:|----:|----:|----:|
| 1 | 465.4 | 1135.3 | 407.8 | 1039.8 |
| 2 | 480.7 | 1178.0 | 410.4 | 1054.5 |
| 3 | 476.7 | 1134.7 | 433.7 | 1107.1 |

</details>

## Correctness

- `parseFhirPath` is deterministic; the cached AST is identical to re-parsing
- The cache is keyed by expression string and bounded by the number of distinct compartment search parameters (a few dozen).
- `@medplum/core` unit tests, server typecheck, and lint all pass.

### DIFF
--- a/packages/server/src/fhir/patient.ts
+++ b/packages/server/src/fhir/patient.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { WithId } from '@medplum/core';
-import { EMPTY, evalFhirPath, getReferenceString, getSearchParameter } from '@medplum/core';
+import type { FhirPathAtom, WithId } from '@medplum/core';
+import { EMPTY, evalFhirPath, getReferenceString, getSearchParameter, parseFhirPath } from '@medplum/core';
 import { readJson } from '@medplum/definitions';
 import type {
   CompartmentDefinition,
@@ -44,6 +44,25 @@ export function getPatientCompartmentParams(resourceType: string): string[] | un
   return undefined;
 }
 
+const compartmentExpressionCache = new Map<string, FhirPathAtom>();
+
+/**
+ * Returns the parsed FHIRPath atom for a patient-compartment search parameter expression, caching
+ * the result. getPatients() runs on every resource write, so this avoids re-tokenizing and
+ * re-parsing the same fixed set of compartment expressions for every resource. The cache is bounded
+ * by the number of distinct compartment search parameters.
+ * @param expression - The search parameter's FHIRPath expression.
+ * @returns The parsed FHIRPath atom.
+ */
+function getParsedCompartmentExpression(expression: string): FhirPathAtom {
+  let parsed = compartmentExpressionCache.get(expression);
+  if (!parsed) {
+    parsed = parseFhirPath(expression);
+    compartmentExpressionCache.set(expression, parsed);
+  }
+  return parsed;
+}
+
 /**
  * Returns the patient compartment ID for a resource.
  * If the resource is in a patient compartment (i.e., an Observation about the patient),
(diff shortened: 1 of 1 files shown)


packages/server/src/fhir/patient.ts
```diff
@@ -44,6 +44,25 @@ export function getPatientCompartmentParams(resourceType: string): string[] | un
   return undefined;
 }
 
+const compartmentExpressionCache = new Map<string, FhirPathAtom>();
```
mattwiller, line 47: Normally we would use an `LRUCache` for this, but since this is a bounded set of compartment search parameters (< 500 elements) a `Map` is perfectly fine 👍

mattwiller (approved):

mattwiller:
Thanks for the PR @jdjkelly!  It looks like you'll need to [sign your commit](https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot) to satisfy the DCO check, but otherwise looks good to me ✅

jdjkelly:
> Thanks for the PR @jdjkelly! It looks like you'll need to [sign your commit](https://www.medplum.com/docs/contributing/dco#2-fixing-existing-commits-if-you-forgot) to satisfy the DCO check, but otherwise looks good to me ✅

Done @mattwiller

codyebberson:
@jdjkelly  - thanks for digging into this. I think the optimization is absolutely valid; we definitely don't want to be reparsing the same FHIRPath expressions on every write.

After chatting with the team, though, I think we'd prefer to solve this one level higher. `SearchParameterDetails` already stores a cached `parsedExpression`, so rather than introducing another cache in this code path, we'd like to steer callers toward using that. That gives us a single canonical parsed representation instead of maintaining multiple caches for the same expressions.

Longer term, I still think there's merit to a small default cache in the generic FHIRPath evaluator as well, since that would help all callers, not just SearchParameters. But for this specific case, using `SearchParameterDetails.parsedExpression` feels like the cleaner architecture.

If we can get that change in today, I'd lean toward doing that instead of merging this as-is. Otherwise, we may merge this first and follow up with the refactor. Either way, thanks for tracking this down—it was a good catch.

jdjkelly:
@codyebberson makes sense - I am hunting for other hot calls and FHIRPath is in general a big one. I'll watch the commit history here and try not to flood PRs for anything that's shaped too similarly.

codyebberson:
> @codyebberson makes sense - I am hunting for other hot calls and FHIRPath is in general a big one. I'll watch the commit history here and try not to flood PRs for anything that's shaped too similarly.

Yes, one of the things we're debating is whether we should enable the global AST parsing cache in `packages/core/src/fhirpath/parse.ts`.  

1. If user passes in a cache, use that
2. If user does not pass in a cache, use a globally configured LRU cache of ~1000 items

That's probably the most convenient solution, but then it has magical global state, which could be undesirable.

jdjkelly:
Solved elsewhere

PR #9676, opened 2026-06-29
Add MedicationStatement resource to PatientSummary component

codyebberson:
Adds display-layer support for `MedicationStatement` resources in the React `PatientSummary` medications section.

Historically, PatientSummary only displayed `MedicationRequest` resources. That aligned with US Core guidance, which recommends `MedicationRequest` and does not include `MedicationStatement`. We have generally discouraged use of `MedicationStatement` for operational medication workflows.

However, we have recently seen multiple customer examples where imported medication history arrives as `MedicationStatement`. This change lets PatientSummary surface those imported records alongside `MedicationRequest` resources so clinicians can see the data that is already present.

## Scope

- Fetches both `MedicationRequest` and `MedicationStatement` resources for the built-in medications section.
- Displays both resource types in the existing Medications list.
- Sorts the combined list by `meta.lastUpdated` descending using a shared date utility.
- Keeps add/edit behavior limited to `MedicationRequest`.
- Routes `MedicationStatement` row clicks through `onClickResource` instead of opening the `MedicationRequest` editor.

## Guidance

`MedicationRequest` remains the recommended resource for operational medication flows, including ePrescribe. `MedicationStatement` support here is intentionally limited to displaying imported data; it does not add authoring or workflow support for `MedicationStatement`.

### DIFF
--- a/packages/react/src/PatientSummary/Medications.test.tsx
+++ b/packages/react/src/PatientSummary/Medications.test.tsx
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import { createReference } from '@medplum/core';
-import type { MedicationRequest } from '@medplum/fhirtypes';
+import type { MedicationRequest, MedicationStatement } from '@medplum/fhirtypes';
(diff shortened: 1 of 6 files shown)


techdavidy:
@claude please review

techdavidy (approved):

PR #9678, opened 2026-06-29
feat(server): Project-level `mfaRequired` setting

### LINKED ISSUE #8853: Add Project-level MFA requirement
**Issue**: MFA enforcement is driven primarily by the User.mfaRequired flag (e.g. set at invite time). Teams that want MFA on for everyone in a project must remember to set it per user or patch User resources. 

**Suggested Fix:** Add a new Project.mfaRequired boolean such that if Project.mfaRequired = true:

1. All new User accounts would have it enabled automatically
2. All new login attempts would require it be enabled

ThatOneBro:
Fixes #8853

### DIFF
--- a/packages/server/src/auth/me.test.ts
+++ b/packages/server/src/auth/me.test.ts
@@ -106,6 +106,7 @@ describe('Me', () => {
     expect(res6.body).toBeDefined();
     expect(res6.body.config).toMatchObject(config);
     expect(res6.body.security).toBeDefined();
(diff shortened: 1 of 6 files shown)


The-Alchemist (approved):

ThatOneBro:
Confirmed on Staging that a server-scoped user that does not have `mfaRequired` set on their user is prompted to enroll into MFA when invited to a project with `mfaRequired` set. Also will not allow a user to login to the requiring project if they unenroll while logged into a separate project that does not require MFA.

PR #9685, opened 2026-06-29
fix(data warehouse): cap DuckDB Postgres connections during warehouse sync

The-Alchemist:
## Summary
- DuckDB by default creates multiple connections and splits queries between them
- That's great for performance, but the data warehouse sync, we'd prefer a slow path which doesn't stress the database instance much
- Settings apply:
  - one connection
  - one thread

### DIFF
--- a/packages/server/src/data-warehouse/destination.ts
+++ b/packages/server/src/data-warehouse/destination.ts
@@ -9,7 +9,6 @@ import type { DuckdbConnection } from './warehouse-sql';
 import {
   buildCopySelectToParquetQuery,
   buildDuckdbPostgresAttachQuery,
-  buildDuckdbPostgresConnectionSettingsQueries,
   buildProjectedSelectFromHistoryTable,
   runParameterizedWarehouseSql,
 } from './warehouse-sql';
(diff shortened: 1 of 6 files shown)


codyebberson:
@claude  please review

packages/server/src/data-warehouse/destination.ts
```diff
@@ -43,7 +44,12 @@ export class LocalParquetWarehouseDestination implements DataWarehouseDestinatio
   }
 
   getSetupQueries(connectionString: string): string[] {
-    return ['INSTALL postgres', 'LOAD postgres', buildDuckdbPostgresAttachQuery(connectionString)];
+    return [
+      'INSTALL postgres',
+      'LOAD postgres',
+      ...buildDuckdbPostgresConnectionSettingsQueries(),
```
codyebberson: might be easier to just include the string literals right in here?  

```
      'SET threads = 1',
      'SET pg_use_ctid_scan = false',
      'SET pg_connection_limit = 1',
      'SET pg_pool_max_connections = 1',
      "SET pg_pool_acquire_mode = 'try'",
```

codyebberson (approved):
I'd recommend inlining the commands, keeping it flatter and simpler, but your call.

packages/server/src/data-warehouse/destination.ts
The-Alchemist: done

The-Alchemist:
@claude re-review

The-Alchemist:
@claude that change, "The default (local) destination no longer caps connections" was intentional.  Connections should only be capped for the S3 Tables path / sink

PR #9686, opened 2026-06-29
[Provider] Hide slots related to cancelled appointments

noahsilas:
If you have created an appointment with attributes like "bufferBefore", then when you use `$cancel` the Slot resource created to enforce that is also soft-deleted.

When you do this from the Provider UI, we want to clear such Slots from the calendar. We don't get the Slot resources back in the update response (as they have been deleted), so we inspect `Appointment.slot` to find references to the related Slot resources we can hide.


codyebberson:
@claude please review

codyebberson (approved):

PR #9689, opened 2026-06-29
Expand error logging metadata

### LINKED ISSUE #8934: Add projectId and subscriptionId to the Failed to auto-disable subscription log line

mattlong:
Fixes #8934

### DIFF
--- a/packages/server/src/workers/subscription.ts
+++ b/packages/server/src/workers/subscription.ts
@@ -978,7 +978,10 @@ async function autoDisableSubscription(
   } catch (err) {
     globalLogger.warn('Failed to auto-disable subscription', {
       subscription: subscription.id,
+      project: subscription?.meta?.project,
+      failureCount,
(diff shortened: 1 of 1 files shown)


packages/server/src/workers/subscription.ts
```diff
@@ -978,7 +978,10 @@ async function autoDisableSubscription(
   } catch (err) {
     globalLogger.warn('Failed to auto-disable subscription', {
       subscription: subscription.id,
+      project: subscription?.meta?.project,
```
mattwiller: This should probably align with the standard logging field:
```suggestion
      projectId: subscription?.meta?.project,
```

mattwiller (approved):

PR #9693, opened 2026-06-30
fix(agent): preserve .log files during uninstall (#3672)

### LINKED ISSUE #3672: Agent uninstall removes all files but should leave created log files in-place
Based on the https://github.com/medplum/medplum/blob/main/packages/agent/installer.nsi#L309 - The files within the `Medplum Agent` directory are removed but the directory is left behind. The Agent log files that are created, should be excluded from the deletion process since these could have information necessary for retention.

devnoct:
fix #3672
The uninstaller used `RMDir /r "$INSTDIR"`, which removed the agent's
`.log` files too. These can hold information worth retaining.

This iterates over `$INSTDIR` and deletes everything except `*.log`
files (using `/REBOOTOK` for files in use, matching the existing style),
then removes the directory only if it ends up empty.

Closes #3672

### DIFF
--- a/packages/agent/installer.nsi
+++ b/packages/agent/installer.nsi
@@ -369,6 +369,63 @@ Function InstallApp
 
 FunctionEnd
 
+# Determine whether a file should be preserved during uninstall (gh #3672):
+# a bare "*.log", or a numbered rotation "*.log.1" through "*.log.999".
+# All checks work off a fixed 8-char tail (wide enough for "*.log.999") so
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Tested a built version of this commit on a real agent installation. Looks like it works 👍 

Screenshot 2026-07-02 at 3 40 22 PM

ThatOneBro (approved):
Thanks @devnoct, I did tweak this a little to catch all log files but looks good now 🙏

PR #9695, opened 2026-06-30
Allow `$export` with read-only scopes

codyebberson:
`$export` operations (`Group/{id}/$export`, `Patient/$export`, project-level `$export`) are
fundamentally **read** operations, but they create internal bookkeeping resources to track job
state and stage output:

- an `AsyncJob` to track the export, and
- one `Binary` per resource type to hold the NDJSON output.

Until now, `BulkExporter` created both of these using the **caller's** repository, so they were
subject to the caller's access policy. A caller with read-only access (for example a
`client_credentials` client granted `system/*.read`) would get an `HTTP 403 Forbidden` when
starting an export, because creating the `AsyncJob` is a write.

This PR makes `BulkExporter` create those bookkeeping resources with the **system repo**, scoped
to the caller's project. A read-only caller can now run `$export` end to end.

## Why this is safe

This does **not** widen access to exported data. The actual FHIR data is still read through the
caller's own repository (`getPatientEverything` / `exportResourceType`), so the caller's access
policy fully governs what ends up in the export.

Only the server-side artifacts of servicing the request move to the system repo:

- They are explicitly scoped to the caller's project (`meta.project`), so the caller can still
  poll the `AsyncJob` status and download the `Binary` files through their own (project-scoped)
  repo.
- They contain only server-controlled fields plus data that was already access-checked when it
  was read.

This also aligns the three creation/update points: `BulkExporter.close()` already updated the
`AsyncJob` via the system repo, so `start()` and `getWriter()` were the inconsistent outliers.

## Changes

- `packages/server/src/fhir/operations/utils/bulkexporter.ts`
  - `start()`: create the `AsyncJob` via `this.repo.getSystemRepo()`, scoped to the caller's
    project.
  - `getWriter()`: create the output `Binary` the same way.
- `packages/server/src/fhir/operations/groupexport.test.ts`
  - Regression test: an export run with a read-only access policy
    (`{ resourceType: '*', readonly: true }`) now succeeds — `start()` no longer throws
    `Forbidden`, the `AsyncJob` lands in the caller's project, and writing output (creating a
    `Binary`) also succeeds.

## Principle of least surprise

Requiring write capability to perform a read-only `$export` is a sharp edge: the `AsyncJob` is an
implementation detail of how Medplum tracks async jobs, not something the caller explicitly asked
to create. Creating an `AsyncJob` directly via `POST /AsyncJob` should still require write access;
doing it implicitly as the machinery of a read operation should not.

## Notes / follow-ups

- This removes a real latent fragility regardless of any specific incident: any read-leaning
  export client was one access-policy/scope tightening away from a 403 at `start()`.
- If we want to be fully explicit about the project rather than deriving it from
  `repo.currentProject()`, a follow-up could thread `project` through the `BulkExporter`
  constructor. The current approach keeps the change minimal and is a no-op for the existing
  system-repo unit tests.

### DIFF
--- a/packages/server/src/fhir/operations/utils/bulkexporter.ts
+++ b/packages/server/src/fhir/operations/utils/bulkexporter.ts
@@ -50,11 +50,16 @@ export class BulkExporter {
   }
 
   async start(url: string): Promise<WithId<AsyncJob>> {
-    this.resource = await this.repo.createResource<AsyncJob>({
+    // The AsyncJob is internal bookkeeping for the export, not user-authored data.
+    // $export is a read operation, so creating it must not require the caller to have
+    // write access -- a read-only scope (e.g. system/*.read) is sufficient. Create it
+    // with the system repo, scoped to the caller's project so they can still poll it.
+    this.resource = await this.repo.getSystemRepo().createResource<AsyncJob>({
       resourceType: 'AsyncJob',
       status: 'active',
       request: url,
       requestTime: new Date().toISOString(),
+      meta: { project: this.repo.currentProject()?.id },
     });
     return this.resource;
   }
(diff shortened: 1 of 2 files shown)


ThatOneBro (approved):

packages/server/src/fhir/operations/utils/bulkexporter.ts
```diff
@@ -50,21 +50,30 @@ export class BulkExporter {
...
+    // The AsyncJob is internal bookkeeping for the export, not user-authored data.
+    // $export is a read operation, so creating it must not require the caller to have
+    // write access -- a read-only scope (e.g. system/*.read) is sufficient. Create it
+    // with the system repo, scoped to the caller's project so they can still poll it.
+    this.resource = await this.repo.getSystemRepo().createResource<AsyncJob>({
       resourceType: 'AsyncJob',
       status: 'active',
       request: url,
       requestTime: new Date().toISOString(),
+      meta: { project: this.repo.currentProject()?.id },
```
ThatOneBro: This is required, otherwise these will become global resources rather than project-scoped resources 👍

codyebberson:
@claude please review

PR #9696, opened 2026-06-30
feat(provider): Include DiagnosticReports and ServiceRequests in labs

techdavidy:
Patient/Labs now provide two options. 
Completed - DiagnosticReport?status=final
Open -  ServiceRequest?status=active%2Cdraft%2Con-hold

Completed tab allows us to browse final results from DiagnosticReports whether they have ServiceRequest or not.
Open tab allows us to browse active, draft and on-hold ServiceRequest resources.


techdavidy:
@claude please review

techdavidy:
@claude review again

examples/medplum-provider/src/utils/documentReference.ts
```diff
@@ -31,6 +31,43 @@ export async function fetchLabOrderRequisitionDocuments(
...
+ * Resolves DiagnosticReport.presentedForm attachments into displayable attachments.
+ * Some integrations (e.g. HGDX) set Attachment.url to a DocumentReference reference
+ * such as "DocumentReference/123" instead of a binary URL. For those entries, fetch
+ * the DocumentReference and use its content attachment instead, preferring the
+ * content whose contentType matches the presentedForm entry.
+ * @param medplum - The Medplum client
+ * @param presentedForm - The DiagnosticReport.presentedForm attachments
+ * @returns The displayable attachments
+ */
+export async function resolvePresentedFormAttachments(
```
noahsilas, line 44: What would you think about moving this into a new component? I'm imagining something roughly like this:

```tsx
{diagnosticReport.presentedForm?.map((attachment, idx) => (
  <PresentedForm attachment={attachment} key={`${attachment.url}_${idx}`} />
))}
```

The queries to resolve any attachments should get auto-batched by the medplum client, but each one could manage things like loading state. Down the line we could even teach the component to hold off on doing the fetch until it was near the viewport to potentially save some queries on reports with many attachments.

examples/medplum-provider/src/components/labs/LabReportContent.tsx
```diff
@@ -0,0 +1,66 @@
...
+  // Resolve presentedForm entries that point at a DocumentReference instead of binary content
+  useEffect(() => {
+    resolvePresentedFormAttachments(medplum, report.presentedForm)
+      .then(setPresentedFormAttachments)
+      .catch(console.error);
+
+    // Clear attachments when the report changes so the previous report's documents do not linger
+    return () => {
+      setPresentedFormAttachments([]);
+    };
```
noahsilas: Keeping this state in sync feels tricky; this seems like it should work, but I wonder if we should add something like `key={report.id}` to callers of this component.

examples/medplum-provider/src/components/labs/LabResultDetails.tsx
```diff
@@ -1,171 +1,62 @@
...
+                      return result.code.coding.map((coding) => coding.display).join(', ');
+                    }
+
+                    // If there's a text field and only one code, use the text field
+                    if (result.code?.text) {
+                      return result.code.text;
+                    }
+
+                    // Otherwise, show the first code or fallback
+                    return result.code?.coding?.[0]?.display || 'Lab Result';
```
noahsilas: This block is tricky; I'd probably bias towards using `<CodeableConceptDisplay value={result.code} />`, but admittedly this is doing some extra handling that is maybe better? Might be worth splitting out a follow-up to improve that shared component so we can use it in more places later.

examples/medplum-provider/src/components/labs/LabResultDetails.tsx
```diff
@@ -1,171 +1,62 @@
...
+      <Paper h="100%">
+        <Stack gap="0">
+          <Stack gap="md" p="md">
+            <Stack gap="md">
+              <Stack gap="0">
+                <Text size="xl" fw={800}>
+                  {(() => {
+                    // If there are multiple codes (2 or more), show them separated by commas
+                    if (result.code?.coding && result.code.coding.length >= 2) {
+                      return result.code.coding.map((coding) => coding.display).join(', ');
```
noahsilas: Should this reuse core's `formatCoding()` helper so it will fall back to the `code` attribute if `display` is not set?
```suggestion
                      return result.code.coding.map((coding) => formatCoding(coding)).join(', ');
```

examples/medplum-provider/src/components/labs/LabResultDetails.tsx
```diff
@@ -1,171 +1,62 @@
...
+                    }
+
+                    // Otherwise, show the first code or fallback
+                    return result.code?.coding?.[0]?.display || 'Lab Result';
+                  })()}
+                </Text>
+                <Text size="sm" c="gray.7">
+                  {result.effectiveDateTime
+                    ? `Issued ${formatDate(result.issued)} • Collected ${formatDate(result.effectiveDateTime)}`
+                    : `Issued ${formatDate(result.issued)}`}
```
noahsilas: #nit: extracting the prefix here would be a nice touch:

```suggestion
                  Issued {formatDate(result.issued)}
                  {result.effectiveDateTime && (` • Collected ${formatDate(result.effectiveDateTime)}`}
```

This also feels like it could be some kind of `<li>` markup that can auto-insert the bullet in between adjacent items:

```tsx
<ul className={classes.bulletBetween}>
  <li>Issued {formatDate(result.issued)}</li>
  {result.effectiveDateTime && (<li>Collected {formatDate(result.effectiveDateTime)}</li>)}
</ul>
```

```css
.bulletBetween > li {
  display: inline;
}
.bulletBetween > li + li::before {
  content: "•";
  display: inline-block;
  margin: 0 0.4rem;
}
```

examples/medplum-provider/src/components/labs/LabResultDetails.tsx
```diff
@@ -1,171 +1,62 @@
...
-                {diagnosticReport?.category?.map((category, index) => (
-                  <Text key={index} size="sm">
-                    {category.text || category.coding?.[0]?.display}
-                  </Text>
-                ))}
+    <ScrollArea h="100%">
+      <Paper h="100%">
+        <Stack gap="0">
+          <Stack gap="md" p="md">
+            <Stack gap="md">
```
noahsilas: This Stack looks like it only has one child, which I think means it's not functionally doing anything and can probably be dropped?

examples/medplum-provider/src/components/labs/LabDetailPane.tsx
```diff
@@ -0,0 +1,56 @@
...
+export function LabDetailPane(props: LabDetailPaneProps): JSX.Element {
+  const { item } = props;
+
+  // A ServiceRequest is its own order. A DiagnosticReport may be based on one,
+  // in which case we prefer the order view over the raw result.
+  const basedOnRef =
+    item.resourceType === 'DiagnosticReport'
+      ? (item.basedOn?.find((ref) => ref.reference?.startsWith('ServiceRequest/')) as
+          | Reference<ServiceRequest>
+          | undefined)
```
noahsilas: We have a nice helper for this in core that you can use to remove the cast:
```suggestion
      ? item.basedOn?.find((ref) => isReference<ServiceRequest>(ref, 'ServiceRequest'))
```

examples/medplum-provider/src/pages/patient/LabsPage.tsx
```diff
@@ -1,223 +1,219 @@
...
 
-  const [activeTab, setActiveTab] = useState<LabTab>('completed');
-  const [openOrders, setOpenOrders] = useState<ServiceRequest[]>([]);
-  const [completedOrders, setCompletedOrders] = useState<ServiceRequest[]>([]);
+  const patient = usePatient();
+  const patientReference = useMemo(() => (patient ? getReferenceString(patient) : undefined), [patient]);
+
+  // Tabs are URL-driven: the DiagnosticReport route is "Completed", the
+  // ServiceRequest route is "Open".
+  const activeTab: LabTab = location.pathname.includes('/DiagnosticReport') ? 'completed' : 'open';
```
noahsilas: This is sort of an interesting paradigm - you are mounting this same component under two routes, and testing the location against each one. This feels like something that should probably be in the routing layer; I wonder if we could set those up like `<Route path="DiagnosticReport" element={<LabsPage tab="completed" />}>` so that this becomes less URL-aware?

noahsilas (approved):
Some small questions/thoughts, overall seems legit 👍

examples/medplum-provider/src/components/labs/LabReportContent.tsx
techdavidy: will add the <PresentedForm attachment={attachment} > in the following PR

PR #9704, opened 2026-06-30
Multi-medication cart checkout for ScriptSure e-prescribing

oleg-mp:
## Summary

Adds vendor-neutral React hooks and Provider App UI for the multi-medication cart workflow: clinicians build a per-patient cart of draft `MedicationRequest`s during a visit, then check out once into a single ScriptSure MedCart approval widget. Pairs with medplum-ee PR https://github.com/medplum/medplum-ee/pull/858 (`$checkout-medications`, `$remove-cart-medication`, `$clear-cart` bots).

- **`@medplum/core`**: `MedicationCheckoutRequest`/`Response` and cart remove/clear `Parameters` encode/decode helpers; pending-order-status utilities for cart-staged drafts.
- **`@medplum/react-hooks`**: `useMedicationCheckout` (batch checkout → approval URL) and `useMedicationCart` (remove one / clear all).
- **`@medplum/scriptsure-react`**: thin `useScriptSureCheckout` / `useScriptSureCart` re-exports.
- **`medplum-provider`**: Draft tab as a pure cart (Checkout / Clear cart / per-row trash, no selection checkboxes); order modal with **Add to cart** primary, **Prescribe this medication** secondary, and a cart-count indicator; `PrescriptionIFrameModal` `cart-checkout` mode polling multiple MR ids until webhook reconciliation.

## Test plan

- [x] `npx vitest run` — `MedicationsPage.test.tsx`, `OrderMedicationPage.test.tsx`, `useMedicationCheckout.test.tsx`
- [x] `npx tsc --noEmit` in `examples/medplum-provider`
- [ ] Staging smoke: add 2+ meds to cart → Checkout → approve in MedCart widget → drafts reconcile to active
- [ ] Staging smoke: per-row trash and Clear cart remove drafts from Draft tab
- [ ] Staging smoke: order modal shows cart indicator when cart non-empty; **Prescribe this medication** sends only the current med

## Notes

- Excludes generated `package-lock.json` and `packages/fhirtypes/dist/ProjectMembership.d.ts` from this commit.


oleg-mp:
@claude please review

oleg-mp:
Addressed Claude review + SonarQube new-code coverage in c1cf06ab1.

**Correctness (#1, #2)**
- **Whole-cart checkout:** `handleCheckout` now paginates through all draft `MedicationRequest`s (`fetchAllDraftIds`, 100/page) before calling `$checkout-medications`, so `Checkout (N)` and `Clear cart (N)` both use `total` and off-page drafts are no longer silently dropped.
- **All-lines-failed:** when every item comes back `status: 'failed'`, we surface the failure notification and **do not** open the MedCart widget or start polling.

**Test coverage (#3) / SonarQube (47% → target ≥80%)**
- `packages/core`: cart-management Parameters encode/decode round-trips + type guard tests.
- `packages/react-hooks`: new `useMedicationCart.test.tsx` (remove + clear + invalid response).
- `medplum-provider`: MedicationsPage tests for per-row trash, Clear cart, and all-failed checkout (no modal).

**Minor (#4, #5)**
- Cross-repo notes on `SCRIPTSURE_PENDING_ORDER_STATUS_IN_CART` / `SCRIPTSURE_MESSAGE_ID_SYSTEM` (consumed by `medplum-ee` bots).
- Dropped redundant guard in `submitSingle` (validation lives in `buildSingleDraftBody`).

**Left as-is (#6):** `cartCount` vs `total` can briefly diverge — both advisory; acceptable for now.

Verified locally: 76 tests across core + hooks + provider pages; `tsc --noEmit` clean.

packages/react-hooks/src/useMedicationCart/useMedicationCart.ts
```diff
@@ -0,0 +1,76 @@
...
+ * projects swap vendors by deploying a different bot under the same code.
+ *
+ * Mirrors `useMedicationCheckout`: requests are encoded as `Parameters` bodies
+ * and the `Parameters` responses are decoded by
+ * `parametersToMedicationCartManageResponse`. Per-line outcomes arrive in
+ * `response.items`.
+ *
+ * @returns Callbacks to remove one item or clear the whole cart.
+ */
+export function useMedicationCart(): UseMedicationCartReturn {
```
noahsilas, line 73: From an API design perspective, what do you think about merging `useMedicationCart` and `useMedicationCheckout`? It seems relatively light-weight to return some extra functions that the caller may not need from a hook, and it seems like it would be ergonomic for consumers if they could use all the management tools from a single place.

Would it make sense to include an `addToCart(medicationRequest: MedicationRequest)` function as well? If I'm reading right, it seems like that is a pretty thin wrapper around creating the resource, but maybe encapsulating some of the loading state.

If these were all controlled from the same hook, one potential improvement sticks out to me: we could inspect the loading state of the `addToCart` function to ensure that there isn't a pending request at the time that `checkout()` is invoked.

packages/scriptsure-react/src/useScriptSureCart.ts
```diff
@@ -0,0 +1,30 @@
...
+ * clear, the patient's ScriptSure MedCart (`DELETE /v3/medcart/{patientId}/{rxId}`
+ * / per-item clear). Reverts the affected draft `MedicationRequest`s to plain
+ * (non-in-cart) drafts server-side.
+ *
+ * A near-empty re-export of `useMedicationCart` from `@medplum/react-hooks`: the
+ * vendor binding lives entirely on the server, where each Medplum project's
+ * `OperationDefinition`s for `$remove-cart-medication` / `$clear-cart` point at
+ * the ScriptSure cart-manage bot via the `operationDefinition-implementation`
+ * extension. See the deploy script in
+ * `medplum-ee/packages/scriptsure/src/scripts/deploy.ts`.
```
noahsilas: This reference is not public, but this package is open-source; is there another open reference we can link to instead (or in addition to, if we want to leave this here for internal usage)?

examples/medplum-provider/src/components/meds/PrescriptionIFrameModal.tsx
```diff
@@ -25,7 +25,13 @@ export interface PrescriptionIFrameModalProps {
    * Skips auto-close if the MR was already synced when the modal opened (e.g. reopening iframe).
    */
   medicationRequestIdToWatch?: string;
```
noahsilas: Would it make sense to remove this parameter, as it is strictly less powerful than the array format?

examples/medplum-provider/src/components/meds/PrescriptionIFrameModal.tsx
```diff
@@ -145,43 +166,65 @@ export function PrescriptionIFrameModal(props: PrescriptionIFrameModalProps): JS
...
+    // omitted so the caller can tell a complete sample from a partial one.
+    const sampleSyncedById = async (): Promise<Map<string, boolean>> => {
+      const syncedById = new Map<string, boolean>();
+      await Promise.all(
+        ids.map(async (id) => {
+          try {
+            const mr = await medplum.readResource('MedicationRequest', id, { cache: 'reload' });
+            syncedById.set(id, medicationRequestLooksSynced(mr));
+          } catch {
+            // Leave id absent on transient failure (see attempts handling below).
```
noahsilas: This is a petty implicit error channel, what would you think about gathering these caught errors and returning them too? `return [syncedById, errors]` might be one way. Another option would be to make the map something like `Map<string, boolean | Error>` so you can store the results inline.

examples/medplum-provider/src/components/meds/PrescriptionIFrameModal.tsx
```diff
@@ -66,8 +81,13 @@ export function PrescriptionIFrameModal(props: PrescriptionIFrameModalProps): JS
...
   const [pollTimedOut, setPollTimedOut] = useState(false);
-  /** null = not sampled yet; true = first sample was still draft/pending; false = first sample already synced (never auto-close). */
-  const pollStartedUnsyncedRef = useRef<boolean | null>(null);
+  /**
+   * null = not sampled yet; otherwise the set of watched MR ids that were still
+   * draft/pending on the first sample. Auto-close fires once every id in this set
+   * has synced. An empty set means everything was already synced on open (never
+   * auto-close).
+   */
+  const pollStartedUnsyncedRef = useRef<Set<string> | null>(null);
```
noahsilas: There's a lot of nuance in this sync-polling; would it make sense to extract a hook that can encapsulate a bunch of this in a more testable unit?

```ts
  // construct stable array from input
  const resourcesToSync = useMemo(() => ([
    ...(medicationRequestIdsToWatch ?? []),
    ...(medicationRequestIdToWatch ? [medicationRequestIdToWatch] : [])),
  ]), [watchKey]);

  // start polling loop
  const { synced, unsynced, errors } = useSyncPolling(resourcesToSync, {
    test: medicationRequestLooksSynced,
    pollInterval: POLL_MS,
  });

  // invoke effect when all resources have been sync'd
  useEffect(() => {
    if (unsynced.length === 0) { onFhirSynced(); }
  }, [unsynced, onFhirSynced])
```

examples/medplum-provider/src/hooks/useMedicationRequestSyncPolling.ts
```diff
@@ -0,0 +1,182 @@
...
+        window.clearInterval(intervalId);
+        intervalId = undefined;
+      }
+    };
+
+    const sampleSyncedById = async (): Promise<[Map<string, boolean>, Map<string, Error>]> => {
+      const syncedById = new Map<string, boolean>();
+      const sampleErrors = new Map<string, Error>();
+      await Promise.all(
+        watchedIds.map(async (id) => {
```
noahsilas, line 107: Potential performance enhancement: if we are trying to sync multiple resources, we could pass only the unsync'd IDs into this function. That would avoid repetitive reads on already sync'd elements if just one entry is slow or error-prone. It does add some state tracking complexity, so it seems fine to leave this one as-is for now, as I think we anticipate the number of items being sync'd to be generally quite low(1-5ish?).

examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
```diff
@@ -86,6 +86,17 @@ export interface OrderMedicationPageProps {
    * Approve Queue flow. Called with the created draft on success.
    */
   onAddedToCart?: (medicationRequest: WithId<MedicationRequest>) => void;
+  /**
+   * Optional cart-hook `addToCart` from the parent (shared with Draft-tab
+   * checkout so pending adds can block checkout). When omitted, Add to cart
+   * falls back to `medplum.createResource`.
+   */
+  persistCartDraft?: (medicationRequest: MedicationRequest) => Promise<MedicationRequest>;
```
noahsilas, line 94: Related thought: @techdavidy and I started thinking about exposing some more events through MedplumClient - https://github.com/medplum/medplum/pull/9822 shows an example of a debug UI that listens to all requests going through the client, and https://github.com/medplum/medplum/pull/9821 shows some more targeted "resource modified" events and a consuming hook.

Maybe this could be implemented in a similar way, where the parent component could register its interest in "MedicationRequest" changes, and tie in without needing this explicit pass-through.

This is not really feedback for your PR here, more it is an invitation to join this conversation about how we think these complex trees of components should interact. Maybe we can hop on a call later, as I'd love to hear what you think about these ideas.

noahsilas (approved):

PR #9705, opened 2026-06-30
perf: serve Patient $everything from the reader when searchOnReader is enabled

jdjkelly:
## Why

`Patient/$everything` is routed as a FHIR operation, not a search, so it never went through `setSearchRepositoryMode` and always ran in the default writer mode, even for projects that opted into the [`searchOnReader`](https://github.com/medplum/medplum/blob/main/packages/docs/docs/self-hosting/project-settings.md) `systemSetting`.

But effectively this operation _is_ a search: it fans out into patient-compartment searches (plus an access-check read) to resolve everything related to the patient. That left this read-only work hitting the writer pool instead of the read replica.

## What

This wires `$everything` into the same opt-in: when `searchOnReader` is enabled, flip the repository to reader mode before the reads, mirroring how plain FHIR searches honor the setting. `req.config.searchOnReader` is already populated for every FHIR request in `routes.ts`, so no plumbing changes were needed.

```ts
if (req.config?.searchOnReader) {
  ctx.repo.setMode(RepositoryMode.READER);
}
```

Just passing `DatabaseMode.READER` on the individual queries is not enough on its own: the connection defaults to `RepositoryMode.WRITER`, and the resolver in `repository-connection.ts` forces the writer pool until the logical mode is flipped:

```ts
return getDatabasePool(this.mode === RepositoryMode.WRITER ? DatabaseMode.WRITER : mode);
```

Hence the explicit `setMode(RepositoryMode.READER)`.

## Testing

Added two tests in `routes.test.ts`: one asserting `$everything` reads from the writer pool by default, and one asserting it reads from the reader pool when `searchOnReader` is enabled.

- `tsc --noEmit` clean
- Both `$everything` tests pass


codyebberson:
Makes sense 👍

jdjkelly:
@codyebberson updated to latest master, should be good now

PR #9706, opened 2026-06-30
test(agent): fix flaky dicom test

ThatOneBro:
Fixes a new flaky test, likely coming from the move to `vitest` with different semantics regarding module resolution


The-Alchemist:
great find!

The-Alchemist (approved):

PR #9711, opened 2026-07-01
fix(core/fhirpath): `as` should return empty collection on wrong type

ThatOneBro:
According to the FHIRPath spec there are 3 exception cases, two of which should throw errors and one of which should return the empty collection.

See: https://hl7.org/fhirpath/N1/#:~:text=If%20the%20identifier%20cannot%20be%20resolved%20to%20a%20valid%20type%20identifier%2C%20the%20evaluator%20will%20throw%20an%20error.%20If%20there%20is%20more%20than%20one%20item%20in%20the%20input%20collection%2C%20the%20evaluator%20will%20throw%20an%20error.%20Otherwise%2C%20this%20operator%20returns%20the%20empty%20collection.

See also: https://build.fhir.org/ig/HL7/FHIRPath/en/#unary-operators--and--:~:text=Observation.value%20as%20Quantity%20//%20returns%20value%20as%20a%20Quantity%20if%20it%20is%20of%20type%20Quantity%2C%20and%20an%20empty%20result%20otherwise

If the left operand is of the specified type or a subclass, return the value of the left operand, otherwise:

Case 1: Cannot resolve identifier to valid type identifier -- Throw
Case 2: More than one item in collection -- Throw
Finally: Return empty collection <-- This is the case we don't handle properly today

This PR fixes that

### DIFF
--- a/packages/core/src/fhirpath/fhirpath.test.ts
+++ b/packages/core/src/fhirpath/fhirpath.test.ts
@@ -646,11 +646,14 @@ describe('FHIRPath Test Suite', () => {
     });
 
     test('testPolymorphismAsB', () => {
-      expect(() => evalFhirPath('(Observation.value as Period).unit', observation)).toThrow();
+      // The official suite flags this as a semantic error (Period has no `unit`).
+      // Medplum evaluates dynamically, so `value as Period` on a Quantity yields the
+      // empty collection (rather than throwing), and `.unit` on empty is empty.
+      expect(evalFhirPath('(Observation.value as Period).unit', observation)).toStrictEqual([]);
     });
 
     test('testPolymorphismAsBFunction', () => {
-      expect(() => evalFhirPath('Observation.value.as(Period).start', observation)).toThrow();
+      expect(evalFhirPath('Observation.value.as(Period).start', observation)).toStrictEqual([]);
     });
   });
 
--- a/packages/core/src/fhirpath/fhirpath.test.ts
+++ b/packages/core/src/fhirpath/fhirpath.test.ts
@@ -1620,6 +1620,24 @@ describe('FHIRPath Test Suite', () => {
     });
   });
 
+  describe('testOfType', () => {
+    test('unqualified type name', () => {
+      expect(evalFhirPath('Patient.ofType(Patient).type().name', patient)).toStrictEqual(['Patient']);
+    });
+
+    test('FHIR-qualified type name', () => {
+      expect(evalFhirPath('Patient.ofType(FHIR.Patient).type().name', patient)).toStrictEqual(['Patient']);
+    });
+
+    test('FHIR-qualified type name with backticks', () => {
+      expect(evalFhirPath('Patient.ofType(FHIR.`Patient`).type().name', patient)).toStrictEqual(['Patient']);
+    });
+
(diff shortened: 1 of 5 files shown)


The-Alchemist:
@claude review

packages/core/src/fhirpath/fhirpath.test.ts
```diff
@@ -3562,7 +3583,7 @@ describe('FHIRPath Test Suite', () => {
     });
   });
 
-  describe.skip('testType', () => {
+  describe('testType', () => {
```
ThatOneBro, line 3586: Was able to get most of this block passing after fixing `ofType` and `is` to use the new `getTypeName` util

packages/core/src/fhirpath/functions.ts
```diff
@@ -30,6 +30,25 @@ export type FhirPathFunction = (context: AtomContext, input: TypedValue[], ...ar
...
+ * @returns The resolved type name, or the empty string if the atom is not a type identifier.
+ */
+export function getTypeName(typeAtom: Atom): string {
+  if (typeAtom instanceof SymbolAtom) {
+    return typeAtom.name;
+  }
+  if (typeAtom instanceof DotAtom && typeAtom.left instanceof SymbolAtom && typeAtom.right instanceof SymbolAtom) {
+    return typeAtom.left.name + '.' + typeAtom.right.name;
+  }
+  return '';
```
ThatOneBro, line 49: There is still an open question of whether this should just throw here, as we did in `as`.

It seems almost all of the callers expect an invalid type identifier to throw...

See: 

`is`: https://hl7.org/fhirpath/N1/#:~:text=returns%20false.-,If%20the%20identifier%20cannot%20be%20resolved%20to%20a%20valid%20type%20identifier%2C%20the%20evaluator%20will%20throw%20an%20error.,-If%20the%20input

`as`: https://hl7.org/fhirpath/N1/#:~:text=a%20subclass%20thereof.-,If%20the%20identifier%20cannot%20be%20resolved%20to%20a%20valid%20type%20identifier%2C%20the%20evaluator%20will%20throw%20an%20error.,-If%20there%20is

`ofType`: https://hl7.org/fhirpath/N1/#:~:text=result%20is%20empty.-,The%20type%20argument%20is%20an%20identifier%20that%20must%20resolve%20to%20the%20name%20of%20a%20type%20in%20a%20model.,-For%20implementations%20with (although not as explicit here)

packages/core/src/fhirpath/fhirpath.test.ts
```diff
@@ -646,11 +646,14 @@ describe('FHIRPath Test Suite', () => {
     });
 
     test('testPolymorphismAsB', () => {
-      expect(() => evalFhirPath('(Observation.value as Period).unit', observation)).toThrow();
+      // The official suite flags this as a semantic error (Period has no `unit`).
+      // Medplum evaluates dynamically, so `value as Period` on a Quantity yields the
+      // empty collection (rather than throwing), and `.unit` on empty is empty.
+      expect(evalFhirPath('(Observation.value as Period).unit', observation)).toStrictEqual([]);
```
mattwiller, line 652: 👍

mattwiller (approved):

PR #9712, opened 2026-07-01
[Provider] Reveal all appointment participants in detail pane

noahsilas:
We were only showing the first Patient, with the understanding that the current schedule's Practitioner was implied. In the world with Appointments that have multiple associated participants, this breaks down a bit.

We now explicitly show all appointment participants, which may include multiple patients, practitioners, locations, devices, and more.

Relatedly, we adjust the `<CreateEncounterForm>` to block if there are multiple Patient or Practitioner resources associated with the appointment. Previously we were just taking the first one we found, which could lead to misconfigured Encounters; in the long run we should enhance this to let the viewer select the appropriate Patient/Practitioner record to work with. For now we will just warn when this ambiguity is hit.

Screenshot 2026-06-30 at 5 39 21 PM

### DIFF
--- a/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
+++ b/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
@@ -13,7 +13,7 @@ import {
   isReference,
   isResource,
 } from '@medplum/core';
-import type { Appointment, Bundle, CodeableConcept, Patient, Practitioner, Slot } from '@medplum/fhirtypes';
+import type { Appointment, Bundle, CodeableConcept, Patient, Slot } from '@medplum/fhirtypes';
 import {
   CodeableConceptDisplay,
   Form,
@@ -117,7 +117,6 @@ export function AppointmentDetails(props: {
   // one of each.
   const participants = props.appointment.participant.map((p) => p.actor);
   const patientRef = participants.find((r) => isReference<Patient>(r, 'Patient'));
-  const practitionerRef = participants.find((r) => isReference<Practitioner>(r, 'Practitioner'));
 
   const patient = useResource(patientRef);
 
@@ -221,7 +220,7 @@ export function AppointmentDetails(props: {
         <OperationOutcomeAlert outcome={encounterOutcome} title="Loading Encounter failed" />
       )}
       {patientRef && !encounter && !encounterLoading && encounterOutcome && isOk(encounterOutcome) && (
-        <CreateEncounterForm appointment={appointment} patientRef={patientRef} practitionerRef={practitionerRef} />
+        <CreateEncounterForm appointment={appointment} />
       )}
 
       <Divider />
--- a/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
+++ b/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
@@ -21,6 +21,7 @@ import {
   OperationOutcomeAlert,
   ResourceAvatar,
   ResourceInput,
+  ResourceName,
   useMedplum,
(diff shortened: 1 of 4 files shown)


codyebberson:
@claude please review

mattlong:
@claude review again

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -109,17 +110,13 @@ export function AppointmentDetails(props: {
...
-  const patientRef = participants.find((r) => isReference<Patient>(r, 'Patient'));
-  const practitionerRef = participants.find((r) => isReference<Practitioner>(r, 'Practitioner'));
-
-  const patient = useResource(patientRef);
+  const [encounter, encounterLoading, encounterOutcome] = useSearchOne(
+    'Encounter',
+    {
+      appointment: getReferenceString(appointment),
+    },
+    { debounceMs: 0 }
```
mattlong: Disabling debouncing since the input `appointment` isn't driven by user input but by a prop? Might be worth a comment

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -172,22 +169,40 @@ export function AppointmentDetails(props: {
     }
   }, [medplum, appointment, confirmable, onAppointmentUpdate, onSlotUpdate]);
 
+  const sortedParticipants = appointment.participant
+    .flatMap((participant) => {
```
mattlong: Why does this need to be a `flatMap`? It seems like `map` and removing the array wrappers in the return values would be sufficient? Or just switching to a for loop and pushing onto an array.

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -172,22 +169,40 @@ export function AppointmentDetails(props: {
...
+      const actor = participant.actor;
+      if (!isReference(actor)) {
+        return [];
+      }
+      const [resourceType] = parseReference(actor);
+      return [{ actor, resourceType }];
+    })
+    .sort((a, b) => (a.resourceType === 'Patient' ? 0 : 1) - (b.resourceType === 'Patient' ? 0 : 1));
+
+  const hasPatient = appointment.participant.some((p) => isReference(p.actor, 'Patient'));
```
mattlong: Searching through `sortedParticipants` would be more consistent since it drives the return value.

mattlong (approved):
LGTM. Few minor comments

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
noahsilas: I forget why I originally wrote it this way; I suspect that it was intended to handle the case where `actor.reference` is not defined (currently returns `[]` which flattens to nothing). It would be a little clearer to be explicit in the intention to remove unresolvable references:
```ts
const sortedParticipants = appointment.participant
  .map((participant) => {
    if (!isReference(participant.actor)) {
      return undefined;
    }
    const [resourceType] = parseReference(participant.actor);
    return { actor: participant.actor, resourceType }
  }).filter(isDefined)
  .sort(...)
```

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
noahsilas: We can actually do even better; because `sortedParticipants` would have hoisted any "Patient" actors to the front of the list, we actually only need to check the first entry and don't need to scan the array at all!

PR #9714, opened 2026-07-01
feat(agent): add cursor pagination to `$fetch-logs`

### LINKED ISSUE #8149: [Agent] Make logs that have been rotated fetchable
Currently only the current log file being written is fetchable by the `$fetch-logs` operation. This is because the `winston-daily-log-rotate` package does not support querying across multiple log files natively, so in order to accomplish this, we could either:

1. Manually read the files ourselves, and perhaps stitch them together before adding them to a separate Winston instance specifically for the fetch logs operation
2. Fork the package and add this natively into the daily rotate driver

Option 1 seems more sustainable if not upstreaming the change, but also been considering trying to upstream option 2 if there is a want for this from the package maintainers.

ThatOneBro:
Adds `before` parameter and cursor-based pagination to `$fetch-logs`, which expands the queryability of the agent logs.

Also fixes #8149

### DIFF
--- a/packages/agent/src/logger.ts
+++ b/packages/agent/src/logger.ts
@@ -423,7 +423,7 @@ export class WinstonWrapperLogger implements ILogger {
 
     let nextBefore: string | undefined;
     if (hasMore && logs.length > 0) {
-      const oldest = logs[logs.length - 1].timestamp;
+      const oldest = logs.at(-1).timestamp;
       // Number of returned entries sharing the oldest timestamp in this page.
       const sameTimestampInPage = logs.filter((l) => l.timestamp === oldest).length;
       // If this page continued within the same timestamp block as the incoming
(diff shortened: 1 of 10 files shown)


packages/agent/src/logger.ts
```diff
@@ -331,27 +350,89 @@ export class WinstonWrapperLogger implements ILogger {
...
         }
       );
     });
+
+    const hasMore = queryLogs.length > limit;
+    const logs = queryLogs.slice(0, limit);
+
+    let nextBefore: string | undefined;
+    if (hasMore && logs.length > 0) {
+      const oldest = logs[logs.length - 1].timestamp;
```
The-Alchemist:
```suggestion
      const oldest = logs.at(-1).timestamp;
```

The-Alchemist: nitpick: i like sonarqube's suggestion here

The-Alchemist (changes requested):

The-Alchemist (approved):

PR #9715, opened 2026-07-01
feat(server): add server-scoped subscriptions

ThatOneBro:
This implements server-scoped subscriptions behind a server configuration setting, in order to enable large self-hosted deployments to use the project-per-tenant model while still creating a central data processing flow for all tenants


packages/server/src/util/auditevent.ts
```diff
@@ -409,7 +409,11 @@ export async function createSubscriptionAuditEvent(
   const auditEvent: AuditEvent = {
     resourceType: 'AuditEvent',
     meta: {
-      project: auditedEvent.meta?.project,
+      // Scope the audit event to the triggering resource's project rather than the
+      // subscription's. For project-scoped subscriptions these are identical, but for
+      // server-scoped subscriptions (which live in the system project) this ensures the
+      // audit event still lands in the resource's own project.
+      project: resource.meta?.project ?? auditedEvent.meta?.project,
```
ThatOneBro: I don't think this is actually needed 🤔 

Think it could just be:

ThatOneBro:
```suggestion
      project: resource.meta?.project,
```

codyebberson: Yeah.... i don't understand this line

```ts
const auditedEvent = subscription ?? resource;
```

We should probably figure out why that is there.

codyebberson: @mattwiller  ?

codyebberson:
@claude please review

packages/server/src/util/auditevent.ts
codyebberson: @ThatOneBro  - I took a look, and all 3 call sites for `createSubscriptionAuditEvent()` pass in a `subscription`, so we should be able to make both `outcomeDesc` and `subscription` into required arguments.

codyebberson: And then we could remove this line, because `auditedEvent === subscription`

codyebberson:
@claude please re-review

packages/server/src/workers/subscription.ts
```diff
@@ -240,6 +240,24 @@ async function satisfiesAccessPolicy(
...
+  // enforcement, but it is only reachable for a Subscription with no `meta.project`, which is safe
+  // on two independent grounds:
+  //   1. Creation is restricted. A normal user's write always has `meta.project` defaulted to their
+  //      own project context (see Repository.getProjectId in fhir/repo.ts), so only a super admin
+  //      (operating with no project context) can create a project-less Subscription at all.
+  //   2. Evaluation is gated. Project-less subscriptions are only ever loaded and evaluated when the
+  //      `serverScopedSubscriptions` server config flag is enabled (see getSubscriptions below);
+  //      when it is off they are never matched against any resource.
+  // A normal project-scoped Subscription always has `meta.project` set and is fully checked below.
+  if (!subscription.meta?.project) {
```
ThatOneBro, line 257: Worth pointing out this change. This avoids the warning below that the author of a `Subscription` does not have a membership in the project.

Since these are server-scoped subscriptions, the author has to be a super admin anyways, and the server-scoped subscription feature flag needs to be on in order for us to get here without a `meta.project` on the subscription, so it's safe to skip the access policy check only in that case

codyebberson:
@claude please re-review

packages/server/src/config/types.ts
```diff
@@ -128,6 +128,15 @@ export interface MedplumServerConfig {
...
   preCommitSubscriptionsEnabled?: boolean;
 
+  /**
+   * Flag to enable server-scoped rest-hook subscriptions (default: false).
+   * When enabled, the subscription worker evaluates not only the subscriptions within a
+   * resource's own project, but also subscriptions that are not scoped to any project
+   * (i.e. stored in the system project). This allows a single set of subscriptions to
+   * apply across every project on the server.
+   */
+  serverScopedSubscriptions?: boolean;
```
mattlong:
```suggestion
  serverScopedSubscriptionsEnabled?: boolean;
```

packages/server/src/workers/subscription.test.ts
```diff
@@ -613,6 +613,120 @@ describe('Subscription Worker', () => {
       }
     }));
 
+  test('Server-scoped subscription fires across projects when enabled', () =>
```
mattlong, line 610: For both of the new tests, can we add a normal project-scoped subscription and ensure it is triggered in both cases?

mattlong (changes requested):

mattlong (approved):

PR #9718, opened 2026-07-01
Add typeahead code matching to `ValueSet$expand`

maddyli:


mattwiller:
This new query condition isn't well-indexed, and is likely to make `$expand` unacceptably slow due to the large possible result sets.  Have you tested this with any known ValueSets on Staging?

PR #9723, opened 2026-07-01
UI/UX fix for ScriptSure Pharmacy Search

oleg-mp:
Now allows filtering by categories of pharmacies and correctly states format in the form hints

### DIFF
--- a/packages/react/src/PatientSummary/PharmacyDialog.tsx
+++ b/packages/react/src/PatientSummary/PharmacyDialog.tsx
@@ -6,7 +6,7 @@ import type { AddFavoriteParams, AddPharmacyResponse, PharmacySearchParams } fro
 import { formatAddress, normalizeErrorString } from '@medplum/core';
 import type { Organization, Patient } from '@medplum/fhirtypes';
 import type { JSX, ReactNode } from 'react';
-import { useCallback, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 import styles from './PharmacyDialog.module.css';
 
 /**
@@ -42,7 +42,11 @@ export interface PharmacyDialogProps {
   readonly onAddToFavorites: (params: AddFavoriteParams) => Promise<AddPharmacyResponse>;
   /** Optional fields rendered inside the search form above the submit button. */
   readonly renderBeforeSearchButton?: ReactNode;
-  /** Extra search parameters merged into the bot request (e.g. vendor-specific filters). */
+  /**
+   * Extra search parameters merged into the bot request (e.g. vendor-specific filters).
+   * `handleSearch` depends on this callback, so callers should memoize it (e.g. `useCallback`)
+   * to avoid re-creating the search handler on every render.
+   */
   readonly getExtraSearchParams?: () => Record<string, unknown>;
   /** Optional per-field placeholder overrides for the search form. */
   readonly searchPlaceholders?: PharmacySearchFieldPlaceholders;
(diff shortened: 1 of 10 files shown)


oleg-mp:
@claude review plz

packages/react/src/PatientSummary/PharmacyDialog.tsx
```diff
@@ -5,16 +5,51 @@ import { showNotification } from '@mantine/notifications';
...
+
 export interface PharmacyDialogProps {
   readonly patient: Patient;
   readonly onSubmit: (pharmacy: Organization) => void;
   readonly onClose: () => void;
-  readonly onSearch: (params: PharmacySearchParams) => Promise<Organization[]>;
+  readonly onSearch: (params: PharmacySearchParams & Record<string, unknown>) => Promise<Organization[]>;
   readonly onAddToFavorites: (params: AddFavoriteParams) => Promise<AddPharmacyResponse>;
+  /** Optional fields rendered inside the search form above the submit button. */
+  readonly renderBeforeSearchButton?: ReactNode;
```
noahsilas, line 44: Would it make sense to make this a generic `children` prop?

packages/react/src/PatientSummary/PharmacyDialog.tsx
```diff
@@ -221,7 +274,18 @@ export function PharmacyDialog(props: PharmacyDialogProps): JSX.Element {
         Object.entries(searchParams).filter(([_, v]) => typeof v === 'string' && v.trim() !== '')
       ) as PharmacySearchParams;
 
-      if (Object.keys(cleanParams).length === 0) {
+      const extraParams = getExtraSearchParams?.() ?? {};
+      const mergedParams = { ...cleanParams, ...extraParams };
```
noahsilas, line 278: Should extra params go through the "cleaning" process above to remove values like `undefined` and `null` and `''`?

EDIT: Looks like it can't quite be the same, because we also allow arrays and objects in "extra" params. 🤔

packages/react/src/PatientSummary/PharmacyDialog.tsx
```diff
@@ -221,7 +274,18 @@ export function PharmacyDialog(props: PharmacyDialogProps): JSX.Element {
...
+
+      const hasTextCriteria = Object.keys(cleanParams).length > 0;
+      const hasExtraCriteria = Object.entries(extraParams).some(([, value]) => {
+        if (Array.isArray(value)) {
+          return value.length > 0;
+        }
```
noahsilas, line 284: Should we add a similar case for empty objects like `{}`?

packages/react/src/PatientSummary/PharmacyDialog.tsx
```diff
@@ -5,16 +5,51 @@ import { showNotification } from '@mantine/notifications';
...
+  address: 'Street address (min 3 chars)',
+  ncpdpID: 'National Council for Prescription Drug Programs ID',
+};
+
 export interface PharmacyDialogProps {
   readonly patient: Patient;
   readonly onSubmit: (pharmacy: Organization) => void;
   readonly onClose: () => void;
-  readonly onSearch: (params: PharmacySearchParams) => Promise<Organization[]>;
+  readonly onSearch: (params: PharmacySearchParams & Record<string, unknown>) => Promise<Organization[]>;
```
noahsilas, line 41: I think this widens out and prevents the type checker from spotting some common problems like a typo in a search parameter name.

What do you think about making this one generic with respect to the extra search parameters as well?
```ts
export interface PharmacyDialogProps<T extends PharmacySearchParams = PharmacySearchParams> {
  readonly onSearch(params: T) => Promise<Organization[]>;
  ...
}
```

noahsilas (approved):

PR #9727, opened 2026-07-01
[Provider] Improve SchedulePage navigation + Schedule creation

noahsilas:
- When changing the viewed Practitioner, go into a loading state immediately instead of leaving the previous calendar on-screen. This reduces confusion on slow connections while the change is happening.

- When no Schedule is found for the selected practitioner, show an empty state that allows the viewer to create a Schedule resource for them. Previously, the last-viewed schedule would stay on screen, leading to confusing UI state.

- When first clicking in to the Scheduling screen from the sidebar, if we don't find an existing schedule for the viewer we show the empty state with a create button instead of automatically creating a new Schedule resource.

  - This is particularly nice if the viewer is a member of an operations team that should not have their own Schedule resource, but still would like to view the schedules of other Practitioners.

- Added improved error handling, with a special descriptive "not found" message.

Screenshot 2026-07-01 at 2 28 34 PM

Screenshot 2026-07-01 at 9 57 01 PM

## Implementation Notes
The `<SchedulePage>` component was doing a lot of heavy lifting and sort of conflating two concerns:
- Choosing a `Schedule` resource by actor, and potentially creating one for the viewer
- Showing the slots and appointments of the selected `Schedule` resource

To facilitate this set of work improving the first concern, I started by splitting out the inner `<ScheduleDetails>` component that handles the data loading of related resources and rendering the calendar. That meant that the outer `<SchedulePage>` component was now nicely light-weight and it is easier to track the statefulness of these changes.

For that reason, this PR is best reviewed commit-by-commit: the first extracts the inner component, and the second makes the functional changes to the outer layer.


noahsilas:
@claude please review

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -38,290 +24,203 @@ export function SchedulePage(): JSX.Element | null {
...
+  );
+  const [loading, setLoading] = useState(true);
+  const [readOutcome, setReadOutcome] = useState<OperationOutcome | undefined>();
+  // Tracks the last ID whose fetch has settled (success or failure), so that
+  // isLoadingById clears even when readResource rejects and schedule stays stale.
+  const [resolvedId, setResolvedId] = useState<string | undefined>(undefined);
+
+  // True immediately when the URL id changes (derived, no extra render) and stays
+  // true until the fetch for that id settles or the schedule already matches.
+  const isLoadingById = Boolean(id) && schedule?.id !== id && resolvedId !== id;
```
noahsilas, line 42: I'm trying to avoid introducing new "synchronous setState in effect" warnings, so this is a bit of a departure.

Historically we would have used slightly more intuitive code like this (simplified example):
```ts
  useEffect(() => {
    setLoadingById(true);
    medplum.readResource('Schedule', id).then(...).finally(() => setLoadingById(false));
  }, [id]);
```

That synchronous setState causes an extra render cycle when the component first mounts.

In this component, we can derive the loading state by comparing the `id` URL parameter with the loaded `schedule` state. 

There's a complication though; if the load fails for any reason, then we want to go into an error state, and so we needed a way to detect that. For that reason we also are tracking the `resolvedId` state, which we update whenever an active schedule fetch settles (regardless of success or failure state).

examples/medplum-provider/src/pages/schedule/SchedulePage.tsx
```diff
@@ -38,290 +24,203 @@ export function SchedulePage(): JSX.Element | null {
...
+
+  // ReferenceInput does not take a `value` prop, so we use this stateful `key` to trigger
+  // it to remount and take new values from `initialValue` on demand. :confounded:
+  const [stateKey, setStateKey] = useState(0);
```
noahsilas, line 46: This feels frustratingly tricky, I'd like to follow up separately on making `<ReferenceInput value={myRefState} onChange={setMyRefState} />` into a valid usage. This does seem to involve some tricky prop drilling through a few layers of components that all need to get the treatment, so for this PR I'm using this painful manual management. 😪

codyebberson (approved):

PR #9732, opened 2026-07-02
fix(core): use statusOptions in pollStatus so status polls do not resend the request body

### LINKED ISSUE #9731: `MedplumClient.pollStatus` ignores its own `statusOptions` — async jobs with a request body fail with "Request with GET/HEAD method cannot have body" (breaks Super Admin reindex form)
## Bug

`MedplumClient`'s private `pollStatus()` constructs a corrected options object for the status GET, then never uses it:

https://github.com/medplum/medplum/blob/acb726dd16355e8aca822de33b69907c8a6c45fe/packages/core/src/client.ts#L3819-L3836

```ts
private async pollStatus<T>(statusUrl: string, options: MedplumRequestOptions, state: RequestState): Promise<T> {
  const statusOptions: MedplumRequestOptions = { ...options, method: 'GET', body: undefined, redirect: 'follow' };
  if (state.pollCount === undefined) {
    // First request - try request immediately
    if (options.headers && typeof options.headers === 'object' && 'Prefer' in options.headers) {
      statusOptions.headers = { ...options.headers };
      delete statusOptions.headers.Prefer;
    }
    state.statusUrl = statusUrl;
    state.pollCount = 1;
  } else {
    // Subsequent requests - wait and retry
    const retryDelay = options.pollStatusPeriod ?? 1000;
    await sleep(retryDelay, { signal: options.signal });
    state.pollCount++;
  }
  return this.request(statusUrl, { ...options, method: 'GET' }, state); // <-- statusOptions is never used
}
```

Because the final call spreads the original `options` instead of `statusOptions`:

1. **Any async request that carries a body always fails client-side.** If the initiating request had a `body` (e.g. `startAsyncRequest(url, { method: 'POST', body, pollStatusOnAccepted: true })`), the body is carried into the status polling GET, and fetch throws `TypeError: Request with GET/HEAD method cannot have body` (Node's undici and browser fetch both reject it). The async job is created server-side and keeps running; only the client polling breaks, so the caller always sees a failure.
2. The `Prefer` header cleanup above is dead code — `Prefer: respond-async` is re-sent on every poll.
3. `redirect: 'follow'` is never applied.

## User-visible symptom

The Super Admin **Reindex Resources** form is broken out of the box: `packages/app/src/admin/SuperAdminPage.tsx` calls `startAsyncJob(medplum, 'Reindexing Resources', 'admin/super/reindex', formData)`, which POSTs a JSON body with `pollStatusOnAccepted: true`. The first poll throws, and the notification reports "Request with GET/HEAD method cannot have body" even though the reindex job actually started and completes server-side.

The neighboring Super Admin async jobs that send **no** body (Rebuild StructureDefinitions / SearchParameters / ValueSets, Reload Cron) work fine, which makes this look like a reindex-specific server failure when it's actually a client polling bug.

## To reproduce

1. `docker compose -f docker-compose.full-stack.yml up`
2. Sign in to the app as a super admin → Super Admin → **Reindex Resources** → enter `Patient` → Reindex.
3. The notification immediately turns red with "Request with GET/HEAD method cannot have body" (a `TypeError` from fetch in the browser console). Meanwhile the `AsyncJob` resource exists and progresses server-side.

Programmatic minimal repro against any endpoint that returns `202` + `Content-Location`:

```ts
await medplum.startAsyncRequest('admin/super/reindex', {
  method: 'POST',
  body: JSON.stringify({ resourceType: 'Patient' }),
  pollStatusOnAccepted: true,
});
// rejects: TypeError: Request with GET/HEAD method cannot have body
```

## Expected behavior

The status poll should be a bare GET of the status URL — exactly what the unused `statusOptions` already builds.

## Fix

One line — pass `statusOptions` to `this.request()`:

```diff
-    return this.request(statusUrl, { ...options, method: 'GET' }, state);
+    return this.request(statusUrl, statusOptions, state);
```

This also revives the intended `Prefer` header removal and `redirect: 'follow'` behavior. Happy to open a PR.

## Environment

- Present on `main` @ acb726d (2026-07-02) and in released `@medplum/core` (observed on 5.1.18).
- Runtime-independent: undici (Node 18+) and browser fetch both throw on a GET with a body.

Searched existing issues/PRs for "cannot have body", "GET/HEAD", and reindex polling failures — no prior report found.

atanderson19:
Fixes #9731

## What

`MedplumClient.pollStatus()` builds a `statusOptions` object (`{ ...options, method: 'GET', body: undefined, redirect: 'follow' }`, plus `Prefer` header removal on the first poll) but then passes `{ ...options, method: 'GET' }` to `this.request()`, so `statusOptions` is dead code. The original request body is carried into the status polling GET, and fetch throws `TypeError: Request with GET/HEAD method cannot have body` (both undici and browsers reject it).

User-visible symptom: the Super Admin **Reindex Resources** form always reports failure (the reindex `AsyncJob` actually starts and completes server-side; only client polling breaks). Any body-carrying request with `pollStatusOnAccepted: true` is affected. See #9731 for a full repro on a stock `docker-compose.full-stack.yml` stack.

## Fix

One line: pass `statusOptions` to `this.request()`. This also revives the intended `Prefer` header removal and `redirect: 'follow'` behavior.

## Testing

Added a regression test in the `Prefer async` describe block asserting that the initial POST carries the body and `Prefer: respond-async`, and that every status poll is a GET with no body and no `Prefer` header.

- New test fails on `main` with `AssertionError: expected '{"resourceType":"Patient"}' to be undefined`, passes with the fix.
- `vitest run src/client.test.ts`: 199 passed (199).
- `tsc --noEmit` and `eslint` clean on the changed files.

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -4126,6 +4126,47 @@ describe('Client', () => {
       expect(fetch).toHaveBeenCalledTimes(4);
       expect((response as any).resourceType).toStrictEqual('Patient');
     });
+
+    test('Status polls do not resend the request body', async () => {
+      const fetch = vi.fn();
+
+      // First time, return 202 Accepted with Content-Location
+      fetch.mockImplementationOnce(async () =>
+        mockFetchResponse(202, {}, { 'content-location': 'https://example.com/content-location/1' })
+      );
+
+      // Second time, return 202 Accepted with Content-Location
+      fetch.mockImplementationOnce(async () =>
+        mockFetchResponse(202, {}, { 'content-location': 'https://example.com/content-location/1' })
+      );
+
+      // Third time, return 200 with JSON
+      fetch.mockImplementationOnce(async () => mockFetchResponse(200, { resourceType: 'AsyncJob' }));
+
+      const client = new MedplumClient({ fetch });
+      await client.startAsyncRequest('/test', {
+        method: 'POST',
+        body: '{"resourceType":"Patient"}',
+        pollStatusOnAccepted: true,
+        pollStatusPeriod: 1,
+      });
+      expect(fetch).toHaveBeenCalledTimes(3);
+
+      // The initial request carries the body and the Prefer header
+      const initialOptions = fetch.mock.calls[0][1];
+      expect(initialOptions.method).toStrictEqual('POST');
+      expect(initialOptions.body).toStrictEqual('{"resourceType":"Patient"}');
+      expect(initialOptions.headers['Prefer']).toStrictEqual('respond-async');
+
+      // Status polls must be GETs without a body (fetch throws
+      // "Request with GET/HEAD method cannot have body" otherwise)
+      // and without the Prefer header
+      for (const [, pollOptions] of fetch.mock.calls.slice(1)) {
+        expect(pollOptions.method).toStrictEqual('GET');
+        expect(pollOptions.body).toBeUndefined();
+        expect(pollOptions.headers['Prefer']).toBeUndefined();
+      }
+    });
   });
 
   describe('Token refresh', () => {
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -3832,7 +3832,7 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
       await sleep(retryDelay, { signal: options.signal });
       state.pollCount++;
     }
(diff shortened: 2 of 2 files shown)


ThatOneBro (approved):
Yeah looks like the usage of `statusOptions` might have been lost in a refactor of a round of old PRs getting merged in. Thanks for finding this 🙏

PR #9734, opened 2026-07-02
Use cloned Repository for out-of-band save

### LINKED ISSUE #9646: Fire-and-forget AuditEvent write is not transaction-scoped → hard `Repository is in an active transaction` 400 on multi-write operations (`$book`, `$deploy`) when `saveAuditEvents` is on, since #9285 (5.1.19+)
### Describe the bug

`Repository.logEvent()` writes the post-interaction AuditEvent fire-and-forget — `updateResourceImpl(auditEvent).catch(...)` is dispatched on the **request repo (outer `ctx.repo`) without `await`** and is **not transaction-scoped**. When the surrounding interaction is a multi-write operation that holds a `withTransaction` on the same pooled connection, that audit write reuses the connection mid-transaction.

Since **5.1.19 (PR #9285, "Transaction scoped repos")** the guard is strict, so this now throws a **hard 400**:

```
Repository is in an active transaction; use the transaction-scoped repository passed to the callback
```

Before 5.1.19 the same race was silent (`"Not in transaction"` / `"Cannot start SERIALIZABLE transaction..."`, occasional dropped writes). Still present in **5.1.22** (latest).

This is **not** the operation handlers using the wrong repo — they were correctly migrated to the scoped repo. It is specifically the audit write that was missed.

### Reproducer 1 — `Appointment/$book` (clinical, runtime, deterministic)

`createProposedAppointment` in `packages/server/src/fhir/operations/utils/scheduling.ts` correctly uses the scoped `txRepo`:

```ts
const createdResources = await repo.withTransaction(async (txRepo) => {
  await validateAllAvailability(txRepo, slots, healthcareService, ...);
  const createdSlots = await Promise.all(slots.map((slot) => txRepo.createResource<Slot>(slot)));
  const createdAppointment = await txRepo.createResource<Appointment>({ ... });
  return [createdAppointment, ...createdSlots];
}, { serializable: true });
```

The operation's own writes are fine. But the **AuditEvent write fired by the first `txRepo.createResource(Slot)` runs on the outer repo while the serializable transaction is still open** → guard trips → `$book` returns 400.

Steps:
1. Self-hosted, single process (HTTP + workers together), `medplum/medplum-server:5.1.22`, `MEDPLUM_SAVE_AUDIT_EVENTS=true`.
2. `POST /fhir/R4/Appointment/$book` with a `Parameters` body (`slot` + `patient-reference`) against a Schedule with scheduling parameters.
3. → 400 `Repository is in an active transaction; use the transaction-scoped repository passed to the callback`.
4. Set `MEDPLUM_SAVE_AUDIT_EVENTS` empty (audit off) → `$book` succeeds every time. Confirms the audit write is the racing party.

### Reproducer 2 — `Bot/$deploy`

Same class. `deployBot` (`packages/server/src/fhir/operations/deploy.ts`) does `createResource(Binary)` then `updateResource(Bot)` back-to-back; the Binary's audit write races the Bot update. Fails with the same 400 when `saveAuditEvents` is on; succeeds with audit off.

### Expected behavior

Enabling `MEDPLUM_SAVE_AUDIT_EVENTS` (required for audit-trail compliance, e.g. NEN 7513) must not break `$book`, `$deploy`, or any multi-write operation. The post-interaction AuditEvent write should be transaction-scoped (use the same `txRepo` / be sequenced inside the operation's transaction), or run on its own connection — never reuse a connection that is mid-transaction.

Impact: with audit on (mandatory in production), appointment booking is broken. There is no config workaround at runtime because audit cannot be disabled in production.

### Environment

- medplum-server **5.1.22** (Docker), client **5.1.22**
- Self-hosted, single-container (HTTP API + BullMQ workers in one process), Postgres + Redis
- `MEDPLUM_DEFAULT_BOT_RUNTIME_VERSION=vmcontext`, `MEDPLUM_SAVE_AUDIT_EVENTS=true`

### Likely related

#9285 (Transaction scoped repos, 5.1.19), #9260 (Serialize transaction state transitions, 5.1.14), #9526 (Refactor FHIR repository author handling, 5.1.19). The guard is correct; the gap is that `logEvent`'s fire-and-forget write wasn't migrated to be transaction-scoped.

---

mattlong:
Fixes #9646
Replaces #9655

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -2211,14 +2211,13 @@ export class Repository extends FhirRepository implements Disposable {
 
     if (getConfig().saveAuditEvents && isResource(resource) && resource?.resourceType !== 'AuditEvent') {
       auditEvent.id = this.generateId();
-      // Clone the repository to obtain a separate RepositoryConnection is used for two reasons
+      // Clone the repository to obtain a separate RepositoryConnection for two reasons:
       // 1. the unawaited save must outlive the current repo's connection scope, which is marked 'ended'
-      // and closed/unusable as soon as post-commit callbacks return.
+      // and closed/unusable as soon as post-commit callbacks returns (before the unawaited save completes).
       // 2. the unawaited save begins a transaction (in handleStorage) which would clash with any subsequent
-      // transactions in the current Repository and cause one of them to fail.
+      // mainline transactions started on the current Repository and cause one of them to fail.
       // To reduce AuditEvent overhead, we could consider further decoupling AuditEvent saves from request processing
-      // by pushing them onto an in-process queue (or the existing BullMQ infrastructure) drained by one long-lived writer.
-      // on an interval.
+      // by pushing them onto an in-process queue (or BullMQ) and drain/write them to the DB on an interval.
       const saveRepo = this.clone();
       saveRepo
         .updateResourceImpl(auditEvent, true)
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -332,10 +332,11 @@ export class Repository extends FhirRepository implements Disposable {
    * Convenience method to create a new repository with the same context but a new connection.
    * Since a cloned Repository doesn't share a connection, cloning is safe at any point of the
    * Repository's life-cycle for out-of-band work, even after the Repository has been disposed.
+   * @param contextOverrides - Optional overrides for the RepositoryContext.
    * @returns A new repository with the same context but a new connection.
    */
-  clone(): Repository {
-    return new Repository(this.context);
+  clone(contextOverrides?: Partial<RepositoryContext>): Repository {
+    return new Repository({ ...this.context, ...contextOverrides });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ThatOneBro:
@claude Can you make sure that cloning at any point in the lifecycle of the `Repository` is sound?

ThatOneBro:
Here is what Fable 5 found about this:

**Verdict: the change is sound.** The clone-per-save approach fixes two real bugs in the old fire-and-forget audit save, and the details hold up under scrutiny.

### Bugs this fixes

1. **Scope-ended race** — the unawaited `this.updateResourceImpl(auditEvent, true)` ran on the request repo's connection, and lost a race two ways:
   - `closeRequestContext()` disposes the request repo when the response finishes, so an in-flight save could hit `assertUsable()` → "Already closed".
   - For transactional writes, success audit events fire in post-commit callbacks, and the transaction scope flips to `ended` as soon as those callbacks return — before the unawaited save completes — so the save could hit "The transaction has ended".

   Either way the AuditEvent was silently dropped (error log only).

2. **Connection-state clash** — the save begins its own transaction in `handleStorage` (`ensureInTransaction` → `BEGIN`). On the shared connection that pushes a scope that makes subsequent mainline operations on the same repo throw "Repository is in an active transaction". Read/search audit events fire mid-request (not deferred to post-commit), so this was reachable.

### Details verified

- Removing `assertUsable()` from `clone()` is safe: `clone()` only reads `this.context`, and `[Symbol.dispose]` never mutates or frees the context — it only sets `closed` and releases the connection. Cloning a disposed repo is genuinely fine, as the new doc comment claims.
- No other caller depended on the removed guard. It only ever blocked ancestor-scoped repos mid-transaction (`txRepo.clone()` always passed it), and the other production call sites (`explain.ts`, `asyncjobexecutor.ts`) clone at root scope.
- The clone lazily acquires its own owned pooled connection (the constructor doesn't connect), so the `.finally()` dispose properly releases it, and no connection is leaked if the save fails before acquiring one.
- The dispose in `.finally()` can't throw on any path traced (which would otherwise become an unhandled rejection): `Repository[Symbol.dispose]` is idempotent, and even the rollback-failed/dead-connection path resets scopes to root and releases defensively before the `finally` runs, so `releaseConnection` no-ops; `conn.release()` itself is wrapped in try/catch.
- `updateResourceImpl` is `async`, so synchronous throws land in the `.catch` rather than escaping `logEvent`.
- No audit recursion: the save calls `updateResourceImpl` directly (no `logEvent` wrapper on that path) and `resourceType !== 'AuditEvent'` guards entry.
- Access-policy propagation works through the shared context: the new transaction.test.ts test confirms account/compartment metadata lands on the saved AuditEvent.
- `SystemRepository` never reaches this path (`isSystem` early return in `logEvent`), so `clone()` returning base `Repository` rather than the concrete class is harmless today. Minor: `createTransactionScopedRepo` deliberately preserves the concrete class via `this.constructor` — worth mirroring in `clone()` if it ever gets broader use.

### Operational caveats (both already acknowledged in the in-code comment)

- Each audited operation now briefly holds a **second pool connection**, so peak connection demand roughly doubles with `saveAuditEvents` on. The pool sets no `connectionTimeoutMillis`, so under saturation audit saves queue indefinitely — no deadlock (requests never await them), but sustained overload piles up pending saves in memory with no backpressure.
- In-flight saves are still lost on process crash (pre-existing fire-and-forget semantics). The suggested queue/BullMQ follow-up would address both.

Both changed test files pass (transaction.test.ts: 54, method-guards.test.ts: 36) and eslint is clean on `repo.ts`.

ThatOneBro (approved):

mattlong:
> Each audited operation now briefly holds a second pool connection, so peak connection demand roughly doubles with saveAuditEvents on. The pool sets no connectionTimeoutMillis, so under saturation audit saves queue indefinitely — no deadlock (requests never await them), but sustained overload piles up pending saves in memory with no backpressure.

This isn't accurate; the pool connection from the original transaction is released after `COMMIT` is issued.

PR #9735, opened 2026-07-02
fix(server): reject Prefer: respond-async in $set-accounts without propagate

maddyli:
## Summary

Fixes #8926.

When calling `$set-accounts` with `Prefer: respond-async` but without `propagate: true`, the server silently ran the operation synchronously instead of honoring the async preference. The caller had no indication their preference was ignored — the response looked like a normal synchronous 200.

This happened because the async branch was gated on `respond-async && params.propagate`; without `propagate`, execution short-circuited to the synchronous path.

The async worker only exists for the potentially long-running compartment propagation (the worker hardcodes `propagate: true`), so a non-propagate call has nothing to run in the background. Rather than silently falling back, this change returns a **400 Bad Request** so callers know the combination is unsupported (Option 1 from the issue).

## Changes

- `packages/server/src/fhir/operations/set-accounts.ts`: return `badRequest` when `respond-async` is requested without `propagate`.
- `packages/server/src/fhir/operations/set-accounts.test.ts`: add a test asserting the 400 response.

## Testing

- Typechecks cleanly.
- ⚠️ The vitest suite could not be run in my environment (server tests require Postgres + Redis, which weren't available). Please run `npm test -- src/fhir/operations/set-accounts.test.ts` in `packages/server` against a database to confirm.

### DIFF
--- a/packages/server/src/fhir/operations/set-accounts.test.ts
+++ b/packages/server/src/fhir/operations/set-accounts.test.ts
@@ -452,6 +452,25 @@ describe('Patient Set Accounts Operation', () => {
     );
   });
 
+  test('Rejects async response without propagate', async () => {
+    const res = await request(app)
+      .post(`/fhir/R4/Patient/${patient.id}/$set-accounts`)
+      .set('Authorization', 'Bearer ' + accessToken)
+      .set('Content-Type', ContentType.FHIR_JSON)
+      .set('Prefer', 'respond-async')
+      .send({
+        resourceType: 'Parameters',
+        parameter: [
+          {
+            name: 'accounts',
+            valueReference: createReference(organization1),
+          },
+        ],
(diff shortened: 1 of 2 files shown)


mattwiller:
I don't totally understand the motivation for this change: why would returning an error be the correct result here?

> The caller had no indication their preference was ignored — the response looked like a normal synchronous 200.

The immediate `200 OK` response is in fact the indication: it tells the user that the request was processed and obtained a successful result without needing to process asynchronously.  If the request had instead been processed asynchronously, it would have returned a `202 Accepted`.

In this case, since there is no heavy work to be done, I don't believe it makes sense to spawn an async job.  The user's preference for an async response is overridden by the reality that it's significantly faster to do just one update synchronously — so we opt to do the sensible thing

PR #9736, opened 2026-07-02
fix(data warehouse): reset DuckDB connection per table and limit postgres reader load

The-Alchemist:
## Summary
    
* Close and reopen the DuckDB database between table syncs to avoid Postgres `SSL connection has been closed unexpectedly`
* Configure DuckDB with single-threaded postgres reads and disabled CTID scans to be gentler on reader instances.
* Improve warehouse sync logging for better diagnostics
    

## Details

DuckDB's postgres' queries get killed with `SSL connection has been closed unexpectedly`, which means the _server_ kills the connection.  Current hypothesis is that after many queries, some connections get idle timed out and killed next time DuckDB tries to do a Postgres query.

One way around this, tested here:
* just use one connection
* create a separate connections for every resource type / table
   * (creating a connection once every few seconds is not ideal, but low-impact)


The-Alchemist:
@claude review

codyebberson (approved):

PR #9740, opened 2026-07-02
Prepare `ConceptMap/$import` and `ConceptMap/$translate` for SNOMED-to-ICD mapping

mattwiller:


codyebberson:
@claude please review

packages/server/src/fhir/operations/conceptmapimport.ts
```diff
@@ -319,14 +318,18 @@ async function writeMappingRows(
   attributes: (Omit<AttributeRow, 'mapping'>[] | undefined)[]
 ): Promise<void> {
   if (mappings.length) {
-    const insertMappings = new InsertQuery('ConceptMapping', mappings)
-      .mergeOnConflict(['conceptMap', 'sourceSystem', 'sourceCode', 'targetSystem', 'targetCode'])
```
mattlong, line 341: Why don't we merge on conflict anymore? Could you please add comments outlining the strategy and steps of the importer?

packages/server/src/fhir/operations/conceptmaptranslate.ts
```diff
@@ -167,13 +181,69 @@ async function handleUnmappedCodes(
...
+      }
+      currentMappingId = id;
+    }
+
+    if (kind && match) {
+      const propertyName = kind as 'property' | 'dependsOn' | 'product';
+      if (propertyName === 'property') {
+        const property: ConceptMapTranslateMatchProperty = { uri, value: { type, value: JSON.parse(value) } };
+        match.property = append(match.property, property);
+      } else {
```
mattlong: To avoid the `kind as ...` cast above
```suggestion
      } else if (propertyName === 'dependsOn' || propertyName === 'product') {
```

packages/server/src/migrations/data/v41.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const migration: CustomPostDeployMigration = {
+  type: 'custom',
+  prepareJobData: (asyncJob) => prepareCustomMigrationJobData(asyncJob),
+  run: async (repo, job, jobData) => runCustomMigration(repo, job, jobData, callback),
+};
+
+// prettier-ignore
+async function callback(client: PoolClient, results: MigrationActionResult[]): Promise<void> {
+  await fns.query(client, results, `DROP INDEX CONCURRENTLY IF EXISTS "ConceptMapping_map_source_target_idx"`);
```
mattlong: Confirming that you made this migration manually (i.e. not generated based on changes in `migrate.ts`) since the index was never tracked in `migrate.ts`? Only in `src/migrations/schema/v98.ts`?

mattlong (approved):

mattlong (changes requested):
Oops, meant to hit Request changes on my review; mostly for the request for comments outlining the import process

packages/server/src/migrations/data/v41.ts
mattwiller: I regenerated the migration with the drop index flag after rebasing; it's now a schema migration

packages/server/src/fhir/operations/conceptmapimport.ts
mattwiller: Added some comments, but in general we removed the unique index on `ConceptMapping` because the SNOMED-to-ICD map breaks that assumption.  That map contains complex logic that can map the same source/target pair more than once, with different `dependsOn` and `property` attributes

mattlong (approved):

PR #9741, opened 2026-07-02
Prefer complete over example CodeSystem and skip retired terminology resources

maddyli:
## Summary

Addresses a terminology selection issue raised in Slack: a `ValueSet` composed of a CodeSystem URL that has **two** CodeSystems in the same Project — one `content: example` and one `content: complete` — was sometimes selecting the `example` one. In the US Core v5 project this caused `http://hl7.org/fhir/us/core/ValueSet/us-core-provider-role` to expand to only the 4 example `http://nucc.org/provider-taxonomy` codes instead of the complete set.

`findTerminologyResource()` previously selected among same-URL resources using only `version` and `date`, breaking ties by Project linking order — with no preference between `example` and `complete`, and no consideration of resource `status`.

## Changes

- **Prefer `complete` over `example`**: added a tie-breaker sort on `CodeSystem.content` (`content-mode`). `complete` sorts ahead of `example` (and the other content modes) lexically, so a complete CodeSystem is preferred when `version` and `date` are equal. This only applies to `CodeSystem` (the field doesn't exist on `ValueSet`/`ConceptMap`).
- **Check status / skip retired**: added a `status ne retired` filter so retired (i.e. deactivated) resources are excluded from selection entirely. This unblocks the workflow of deactivating a duplicate CodeSystem/ValueSet by setting its `status` to `retired` — it will no longer be selected regardless of version. Applies to all terminology resource types.

## Testing

Added tests in `codesystemlookup.test.ts`:
- Two CodeSystems with the same URL + version (`example` created first, `complete` second) — a `$lookup` for a code that only exists in the `complete` one succeeds, and a code only in the `example` one is not found, confirming `complete` was selected.
- A retired CodeSystem is excluded from selection (`$lookup` returns 404).

### DIFF
--- a/packages/server/src/fhir/operations/utils/terminology.ts
+++ b/packages/server/src/fhir/operations/utils/terminology.ts
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { WithId } from '@medplum/core';
+import type { Filter, WithId } from '@medplum/core';
 import { badRequest, createReference, OperationOutcomeError, Operator, resolveId } from '@medplum/core';
 import type {
   CodeSystem,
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/operations/utils/terminology.ts
```diff
@@ -41,19 +41,31 @@ export async function findTerminologyResource<T extends TerminologyResource>(
...
+  if (resourceType === 'CodeSystem') {
+    // Prefer more complete content, regardless of version or date: 'complete' sorts ahead of 'example'
+    // (and the other content modes) lexically, so a complete CodeSystem is always preferred over an
+    // example one -- even when the example is newer or has a higher version.
+    sortRules.push({ code: 'content-mode', descending: false });
+  }
+  // Select highest version (by lexical sort -- no version is assumed to be "current")
+  sortRules.push({ code: 'version', descending: true });
+  // Break ties by selecting more recently-updated resource (lexically -- no date is assumed to be current)
+  sortRules.push({ code: 'date', descending: true });
```
mattwiller: This is starting to get relatively complex: I wonder if it would be worth moving this sorting into code so we can have more fine-grained control and handle e.g. semver versions correctly eventually.

mattwiller (approved):

PR #9743, opened 2026-07-03
perf(ccda): 1.4x faster C-CDA to FHIR conversion

jdjkelly:
## Why

Profiling showed most of the C-CDA to FHIR conversion cost was overhead, not parsing or mapping: `fast-xml-parser` built a jPath string for every tag and attribute just to answer "is this element an array?", `sanitizeNullFlavors` deep-cloned the entire document on every conversion, and parser/builder instances were rebuilt per call.

## What

No new dependencies; all changes are inside `packages/ccda/src`.

- `convertXmlToCcda` uses `jPath: false` with `ARRAY_PATHS` indexed by last tag in a `Map`, matching parent-tag suffixes against the parser's live `MatcherView` instead of building a jPath string and scanning all 87 paths with `endsWith` per element
- `XMLParser`/`XMLBuilder` instances hoisted to module singletons; the builders also set `jPath: false`
- `sanitizeNullFlavors` is copy-on-write instead of cloning the whole tree
- 9 new tests lock in the new semantics

## Benchmarks

Interleaved before/after runs, Node 22, means of 10-30 samples.

ONC 170.315(b)(1) fixtures (per 6 documents):

| Operation | Before | After | Speedup |
|---|---|---|---|
| `convertXmlToCcda` | 10.9 ms | 7.8 ms | 1.40x |
| `convertCcdaToFhir` | 1.10 ms | 0.84 ms | 1.31x |
| Full pipeline | 12.0 ms | 9.0 ms | 1.33x |

Synthea sample C-CDAs (per document):

| Document | Parse before / after | Convert before / after |
|---|---|---|
| 0.3 MB (238 resources) | 12.2 / 9.0 ms (1.36x) | 1.8 / 1.2 ms (1.50x) |
| 0.8 MB (806 resources) | 29.8 / 21.2 ms (1.41x) | 4.3 / 2.5 ms (1.72x) |
| 10.4 MB (4,110 resources) | 266 / 191 ms (1.39x) | 27.2 / 18.0 ms (1.51x) |
| 20.1 MB (19,050 resources) | 539 / 391 ms (1.38x) | 93.8 / 70.9 ms (1.32x) |

Full pipeline on the 21 MB document: 633 ms to 462 ms (1.37x).

Parse output is byte-identical to the previous implementation on all `testdata/` fixtures and on Synthea sample documents up to 21 MB.


The-Alchemist:
Thank you, @jdjkelly .  Is there a way to get some of these perf benefits without adding `path-expression-matcher`?  We are very hesitant to add a new dependency unless absolutely necessary.

jdjkelly:
@The-Alchemist done

Note that this dependency is already in the tree, the idea with pinning it was just to make detecting drift easier, but the PR does not require a package.json edit.

PR #9744, opened 2026-07-03
Fix DiagnosticReport conclusion field matching

### LINKED ISSUE #4761: DiagnosticReport.conclusion and DiagnosticReport.conclusionCode both believe they are for the same SearchParameter
I am using a simple table view of DiagnosticReport just showing two fields: conclusion and conclusionCode.

Screenshot 2024-06-29 at 10 02 24 PM

When I print out the `fields` object at the end of `getFieldDefinitions` within `SearchControlField.ts`, it shows the following:

```
[
  {
    "name": "conclusion",
    "elementDefinition": {
      "description": "Concise and clinically contextualized summary conclusion (interpretation/impression) of the diagnostic report.",
      "path": "DiagnosticReport.conclusion",
      "min": 0,
      "max": 1,
      "isArray": false,
      "constraints": [],
      "type": [
        {
          "code": "string",
          "targetProfile": null,
          "profile": null
        }
      ],
      "binding": null
    },
    "searchParams": [
      {
        "base": [
          "DiagnosticReport"
        ],
        "code": "conclusion",
        "type": "token",
        "expression": "DiagnosticReport.conclusionCode",
        "target": null
      }
    ]
  },
  {
    "name": "conclusionCode",
    "elementDefinition": {
      "description": "One or more codes that represent the summary conclusion (interpretation/impression) of the diagnostic report.",
      "path": "DiagnosticReport.conclusionCode",
      "min": 0,
      "max": null,
      "isArray": true,
      "constraints": [],
      "type": [
        {
          "code": "CodeableConcept",
          "targetProfile": null,
          "profile": null
        }
      ],
      "binding": {
        "strength": "example",
        "valueSet": "http://hl7.org/fhir/ValueSet/clinical-findings"
      }
    },
    "searchParams": [
      {
        "base": [
          "DiagnosticReport"
        ],
        "code": "conclusion",
        "type": "token",
        "expression": "DiagnosticReport.conclusionCode",
        "target": null
      }
    ]
  }
]
```

which seems to suggest that both fields want to refer to the same search parameter. This leads to the weird behavior that I noticed while fixing [this bug](https://github.com/medplum/medplum/issues/4748) that sorting by DiagnosticReport.conclusion actually ends up sorting by DiagnosticReport.conclusionCode.

Is this intended or a bug? I noticed lower down in the same method, when we determine whether a field has a SearchParameter instead [using pathRegex](https://github.com/medplum/medplum/blob/2bd203ece21c8ffbf86d420617ee859bb7e42a3f/packages/react/src/SearchControl/SearchControlField.ts#L120) that it does that right thing and concludes that there is no SearchParameter for DiagnosticReport.conclusion.

driavysinus:
Fixes #4761.

This fixes the SearchControl field matching for `DiagnosticReport.conclusion` and `DiagnosticReport.conclusionCode`.

FHIR R4 defines a `conclusion` SearchParameter whose expression points to `DiagnosticReport.conclusionCode`. Because `SearchControlField` previously treated an exact element name and exact SearchParameter code as a match without checking the expression, the free-text `DiagnosticReport.conclusion` element could be associated with the token SearchParameter for `DiagnosticReport.conclusionCode`.

This PR verifies that same-code exact SearchParameters actually match the element path before associating them with an exact element. It preserves existing alias behavior for SearchParameters whose code differs from the selected field name.

Validation:
- `npm --workspace packages/react test -- SearchControlField.test.ts`
- `npm --workspace packages/react test -- SearchControlField.test.ts SearchPopupMenu.test.tsx`
- `npx prettier --check packages/react/src/SearchControl/SearchControlField.ts packages/react/src/SearchControl/SearchControlField.test.ts`
- `git diff --check`

Note: the full local `npm --workspace packages/react test` suite was also run, but unrelated locale-sensitive assertions failed on this Windows environment (currency, date, and number formatting).

### DIFF
--- a/packages/react/src/SearchControl/SearchControlField.test.ts
+++ b/packages/react/src/SearchControl/SearchControlField.test.ts
@@ -19,4 +19,23 @@ describe('getFieldDefinitions', () => {
     expect(identifierField?.searchParams?.length).toBe(1);
     expect(identifierField?.searchParams?.find((sp) => sp.code === 'identifier')).toBeDefined();
   });
+
+  test('DiagnosticReport conclusion does not match conclusionCode search parameter', () => {
+    const fieldDefs = getFieldDefinitions({
+      resourceType: 'DiagnosticReport',
+      fields: ['conclusion', 'conclusionCode'],
+    });
+
+    expect(fieldDefs.length).toBe(2);
+    const conclusionField = fieldDefs.find((field) => field.name === 'conclusion');
+    const conclusionCodeField = fieldDefs.find((field) => field.name === 'conclusionCode');
+
+    expect(conclusionField?.elementDefinition?.path).toBe('DiagnosticReport.conclusion');
+    expect(conclusionField?.searchParams).toBeUndefined();
+
+    expect(conclusionCodeField?.elementDefinition?.path).toBe('DiagnosticReport.conclusionCode');
+    expect(conclusionCodeField?.searchParams?.length).toBe(1);
+    expect(conclusionCodeField?.searchParams?.[0]?.code).toBe('conclusion');
+    expect(conclusionCodeField?.searchParams?.[0]?.expression).toBe('DiagnosticReport.conclusionCode');
+  });
 });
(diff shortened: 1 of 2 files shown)


driavysinus:
Update: DCO, labeler, Socket, Snyk, and Chromatic/UI checks are green on the latest commit. The remaining Vercel failures are team-authorization gates for fork previews rather than build failures.

@mattlong @mattwiller, when convenient, could you review this fix and authorize the remaining preview deployments?

PR #9745, opened 2026-07-04
Reentrant async batch

### LINKED ISSUE #9124: Store async bundles in object storage
Instead of including them in the bullmq job

### LINKED ISSUE #9681: Re-entrant async batch processing
Medplum supports asynchronously processing `batch` type Bundles in the background via a BullMQ job. Instead of applying rate limits to async batches, async batches incur sleeps, sometimes on the order of 100s of milliseconds between batch entries. For bundles with many entries, this can lead to the total processing time of an async batch to be more than a few minutes. The job currently isn't re-entrant making it vulnerable to being interrupted by a new server deploy. Furthermore, the partial results of the interrupted batch are not made available.

These jobs need to be re-entrant so that processing a large batch survives server deploys or unexpected crashes of bullmq workers. This will require persisting the in-flight state of `BatchProcessor` out of the process to somewhere more durable, e.g. Redis or object storage (S3), including but not limited to:

* `BundlePreprocessInfo`
* `BatchProcessor.resolvedIdenties`
* `BatchProcessor.resultEntries`
* `BatchProcessor.entryIndex`

An alternative, reentrant API of `BatchProcessor` is needed for the BullMQ job to interact with. It should persist initial state stemming from preprocessing the batch, and then update its progress after each entry is processed.

The main processing loop is exposed rather than being encapsulated within `run()`. e.g. a `BatchProcessor.processNextEntry()` method. This way, server shutdown is detected, the BullMQ job can simply stop calling processNextEntry and delay/requeue the job.

Once the job is resumed, `BatchProcessor` needs a way to hydrate state from the durable store, e.g. a static `BatchProcessor.fromState(asyncJob....)` where `asyncJob` is assumed to be the input for computing durable storage keys.

There will be additional API changes to `BatchProcessor` to facilitate re-entrant processing, but this proposal covers the crux of durable storage.

### What about storing intermediate state in the BullMQ job itself?
This could be a viable approach, but its not yet clear if that approach is scalablesince the total size of intermediate state is non-trivial; especially for very large batches.

#9124 should likely be rolled into this issue.

mattlong:
Action items:

- [x] Add more verbose logging particularly for timing to `BatchCheckpointStore`
- [ ] Look in to having the duration persisting intermediate state detract from FHIR quota sleep duration between entries
- [x] Documentation pass

Fixes #9681 
Fixes #9124

### DIFF
--- a/packages/fhir-router/src/batch.ts
+++ b/packages/fhir-router/src/batch.ts
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { Event, WithId } from '@medplum/core';
 import {
-  append,
   badRequest,
   getReferenceString,
   getStatus,
@@ -39,6 +38,45 @@ type BundlePreprocessInfo = {
 };
 
 /**
+ * A JSON-serializable form of `BundlePreprocessInfo`, where `resourceTypes` is an
+ * array rather than a `Set` so it can be persisted to durable storage.
+ */
+export type SerializableBundleInfo = {
+  ordering: number[];
+  requiresStrongTransaction: boolean;
+  updates: number;
+};
+
+/**
+ * The durable, JSON-serializable state produced by preprocessing a batch bundle. This is
+ * everything required to (re)hydrate a {@link BatchProcessor} and resume processing after a
+ * server restart or worker crash, EXCEPT the progress marker (the index into `bundleInfo.ordering`
+ * of the next entry to process) and the per-entry results produced so far, both of which the
+ * caller (e.g. the async batch worker) is responsible for persisting separately.
+ *
+ * @see BatchProcessor.preprocess
+ * @see BatchProcessor.fromState
+ */
+export interface BatchInitialState {
+  /**
+   * The preprocessed bundle. During preprocessing the bundle is mutated in place: `create`
+   * entries are assigned resource IDs and conditional `update`/`delete` URLs are rewritten to
+   * absolute references. This mutated form MUST be persisted and reloaded verbatim — re-running
+   * preprocessing on the original bundle would assign fresh random IDs and break references to
+   * resources that were already created.
+   */
+  bundle: Bundle;
+  bundleInfo: SerializableBundleInfo;
+  resolvedIdentities: Record<string, string>;
+  /**
+   * Result entries produced during preprocessing itself (i.e. error responses for malformed
+   * entries in a non-transaction bundle), keyed by their index in the original bundle. These
+   * entries are never re-processed, so they are captured here rather than in per-entry progress.
+   */
+  preprocessResults: Record<number, BundleEntry>;
+}
+
+/**
  * Processes a FHIR batch request.
  *
  * See: https://www.hl7.org/fhir/http.html#transaction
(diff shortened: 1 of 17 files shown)


codyebberson:
@claude please review

packages/fhir-router/src/batch.ts
```diff
@@ -82,38 +130,177 @@ class BatchProcessor {
...
+    return this.position;
+  }
+
+  /**
+   * Returns the result entries produced since the last call (keyed by their index in the
+   * original bundle) and clears the pending buffer. The caller persists these to durable
+   * storage as a checkpoint before advancing the durable progress marker.
+   * @returns Newly-produced result entries, keyed by original bundle index.
+   */
+  takePendingResults(): Record<number, BundleEntry> {
```
mattwiller: This could return `undefined` when there are no pending results, which would make some of the call sites cleaner since they wouldn't have to check `Object.keys(results).length > 0` (which allocates an unnecessary array as well)

packages/fhir-router/src/batch.ts
```diff
@@ -364,69 +551,60 @@ class BatchProcessor {
...
-        if (this.isTransaction()) {
-          throw err;
-        }
+    const n = this.position;
+    const entryIndex = bundleInfo.ordering[n];
+    const entry = entries[entryIndex];
+    const rewritten = this.rewriteIdsInObject(entry);
+    try {
+      this.setResult(entryIndex, await this.processBatchEntry(rewritten));
+      this.position = n + 1;
```
mattwiller, line 605: Why store `n` above, instead of just incrementing `position` here?
```suggestion
      this.position++;
```

packages/server/src/cloud/aws/storage.ts
```diff
@@ -97,6 +103,16 @@ export class S3Storage extends BaseBinaryStorage {
     return output.Body as Readable;
   }
 
+  async deleteFile(key: string): Promise<void> {
+    // S3 DeleteObject is idempotent: deleting a missing key returns success.
```
mattwiller, line 107: 👍

packages/server/src/fhir/batch.test.ts
```diff
@@ -1261,6 +1286,143 @@ describe('Batch and Transaction processing', () => {
...
+    let checks = 0;
+    const isClosingSpy = vi.spyOn(queueRegistry, 'isClosing').mockImplementation(() => checks++ >= 2);
+
+    // The delayed job re-throws DelayedError so BullMQ can re-queue it.
+    await expect(execBatchJob(job)).rejects.toBeInstanceOf(DelayedError);
+    isClosingSpy.mockRestore();
+
+    // Progress was checkpointed into the job data so a future worker can resume.
+    expect(job.data.position).toBeGreaterThan(0);
+    expect(job.data.position).toBeLessThan(4);
```
mattwiller: It would probably be simpler and clearer to just assert that this has the expected value, e.g.
```suggestion
    expect(job.data.position).toStrictEqual(3);
```

packages/fhir-router/src/batch.ts
```diff
@@ -364,69 +551,76 @@ class BatchProcessor {
...
    */
-  private async processBatch(
-    bundleInfo: BundlePreprocessInfo,
-    resultEntries: (BundleEntry | OperationOutcome)[]
-  ): Promise<Bundle> {
-    const bundleType = this.bundle.type;
+  async processNextEntry(): Promise<void> {
+    if (this.isTransaction()) {
+      throw new Error('Re-entrant batch processing does not support transaction semantics');
+    }
```
mattwiller, line 573: Does this need to be re-checked for each entry, or could it be moved up?

mattlong:
@claude take another look. I've addressed all your feedback but item 5

mattwiller (approved):
Generally looks good to me, just had some minor questions/nits

packages/fhir-router/src/batch.ts
mattlong: `processNextEntry` is the outermost layer of `BatchProcessor`. I was considering making `BatchProcessor.isTransaction` public instead of private to avoid the repeated calls, but that seemed fragile since the caller needs to know to check `inTransaction()` before using `processNextEntry` whereas keeping it within the processor makes it self-enforcing

PR #9747, opened 2026-07-04
feat: implement :in/:not-in search modifier for token params

### LINKED ISSUE #5664: Correct implementation of `:in` search modifier
The [current implementation](https://github.com/medplum/medplum/blob/26ca7725254a3fe4ab4b73e6e064cec2650becb4/packages/server/src/fhir/lookups/token.ts#L565) of the `:in` modifier for `token` search parameters is overly broad: it matches resources that contain any code from a system referenced by the value set, independent of which codes are included in the ValueSet.  For example, the search string `Observation?code:in=http://hl7.org/fhir/us/core/ValueSet/us-core-smoking-status-observation-codes` should only match resources with `code=http://loinc.org|72166-2` (the only code included in the ValueSet), but will instead produce SQL like this:

```sql
SELECT "Condition"."id"
FROM "Condition"
LEFT JOIN "Condition_Token" ON (
  "Condition_Token"."resourceId"="Condition"."id"
  AND "Condition_Token"."code"='code'
  AND "Condition_Token"."system"=ANY(
    (
      SELECT "ValueSet"."reference"
      FROM "ValueSet"
      WHERE "ValueSet"."url"='http://hl7.org/fhir/us/core/ValueSet/us-core-smoking-status-observation-codes'
      LIMIT 1
    )::TEXT[]
  )
)
```

This is effectively a query to match any LOINC code, since that is the system in `ValueSet:reference`:
```sql
SELECT "Condition"."id"
FROM "Condition"
LEFT JOIN "Condition_Token" ON (
  "Condition_Token"."resourceId"="Condition"."id"
  AND "Condition_Token"."code"='code'
  AND "Condition_Token"."system"=ANY(ARRAY['http://loinc.org']::TEXT[])
)
```

The implementation of `:in` should be updated to use the same query fragments as the `ValueSet/$expand` and/or `ValueSet/$validate-code` operations to ensure it's actually matching against the exact codes included in the target ValueSet.

sahilnyk:
Fixes #5664

  **What this does**

  The `Condition.code:in=http://example.com/ValueSet/test` thing now works. It finds the ValueSet, expands it to get the actual codes, and then searches for those exact codes. The same happens with `:not-in`, it
  searches for everything except those codes. Before, both would give an "Invalid modifier" error.

  This fixes the original bug where `:in` matched any code from the ValueSet's system instead of only the codes actually in the ValueSet.

  **Why I did it this way**

  I reused the existing `expandValueSet` from the `$expand` operation instead of writing separate expansion logic, so the code stays consistent with how ValueSets are expanded everywhere else. Once expanded, each
  `system|code` pair is turned into a normal token filter (`EQUALS` for `:in`, `NOT_EQUALS` for `:not-in`) so the existing search path does the matching.

  One known limit: expansion is capped at 1000 codes, so very large ValueSets aren't fully covered yet. Can be handled as a follow-up.

  **Here is what I did to test it**

  - [x] I ran `npx vitest run src/fhir/tokens.test.ts`. All 101 tests passed
  - [x] I ran `npx tsc --noEmit`. There were no type errors
  - [x] I made sure Postgres was running and the schema was set up before I started testing


sahilnyk:
Hi, CI needs approval to run on this PR. Could someone take a look? Thanks!

mattwiller:
Thanks for the PR @sahilnyk — I'd like to discuss the implementation a bit before merging.  In particular, I was curious about this part:

> During a normal search, there is no logged-in user context, so that code would crash. Instead I wrote a simpler expander that reads codes from the database directly using the search connection no login needed.

Normal search requests are usually authenticated, so I'm unsure what necessitated the parallel `$expand` implementation.  Preferably, we would re-use the existing expansion logic similarly to the [draft implementation of the `:below` modifier](#9527).  Would you mind attempting to align with that approach?

sahilnyk:
@mattwiller oops, thanks for the feedback I'm working on it

sahilnyk:
@mattwiller, I replaced custom inlineExpandValueSet with the shared expandValueSet from expand.ts, following the :below approach. All 101 token tests have passed. Let me know if anything else needs to be changed!

sahilnyk:
I have updated the PR description to reflect the current approach, please refer to it for what's changed.

PR #9748, opened 2026-07-04
Server config setting for enabled search parameters

codyebberson:
Adds new server configuration options for disabling selected background services and controlling which standard FHIR search parameters are indexed at startup.

The main change is `enabledSearchParameters`, which allows deployments to restrict the standard FHIR search parameters loaded into the server schema. This can significantly reduce search/indexing overhead for deployments that only need a smaller subset of standard search parameters. Medplum-specific search parameters are still always enabled.

Also adds feature flags for disabling dispatch and subscription processing:

- `dispatchEnabled`
- `subscriptionsEnabled`

Both default to `true`, preserving existing behavior unless explicitly disabled.

## Details

- Adds `enabledSearchParameters?: string[]` to `MedplumServerConfig`.
- Updates structure definition loading to accept server config and only index enabled standard search parameters.
- Keeps Medplum search parameters always enabled.
- Adds `dispatchEnabled` config support to skip dispatch job execution when disabled.
- Adds `subscriptionsEnabled` config support to skip subscription job creation when disabled.
- Adds defaults for the new service flags.
- Adds test coverage for disabled dispatch and subscription branches.

Recommended config for https://github.com/HealthSamurai/fhir-server-performance-benchmark 

```json
  "rateLimitsEnabled": false,
  "dispatchEnabled": false,
  "subscriptionsEnabled": false,
  "autoDownloadEnabled": false,
  "enabledSearchParameters": [
    "conformance-date",
    "conformance-version",
    "conformance-url",
    "Patient-name",
    "individual-address",
    "individual-birthdate",
    "Organization-name",
    "Observation-category",
    "clinical-code",
    "clinical-date",
    "Observation-value-quantity",
    "Encounter-status",
    "Encounter-class",
    "MedicationRequest-subject",
    "medications-encounter",
    "MedicationRequest-requester"
  ]
```


codyebberson:
More config settings

```
docker pull medplum/medplum-server-staging:c991f761bdcdd95fd5cacaff7845d11562d3813f
```

https://hub.docker.com/repository/docker/medplum/medplum-server-staging/tags/c991f761bdcdd95fd5cacaff7845d11562d3813f/sha256:e6febc8d1bef5918706d02afb2c8b7d443e1fb0f5e867836079099e8df9047fb

packages/server/src/config/types.ts
```diff
@@ -122,6 +122,12 @@ export interface MedplumServerConfig {
   /** Optional threshold in milliseconds for logging and recording high idle time within transactions */
   idleInTransactionLogThresholdMs?: number;
 
+  /** Flag to enable/disable the background worker dispatch service. (default 'true' for enabled) */
```
mattlong: This may not be the right spot, but do we need to layout which features are gated behind dispatch?

packages/server/src/config/utils.ts
```diff
@@ -37,6 +37,8 @@ export function addDefaults(config: MedplumServerConfig): ServerConfig {
   config.defaultProjectFeatures ??= [];
   config.defaultProjectSystemSetting ??= [];
   config.emailProvider ||= config.smtp ? 'smtp' : 'awsses';
+  config.dispatchEnabled ??= true;
```
mattlong, line 40: Need to be added to `DefaultConfigKeys`. I suspect that list is pretty out of date :(

packages/server/src/config/types.ts
```diff
@@ -215,6 +221,9 @@ export interface MedplumServerConfig {
 
   /** Optional flag to allow external auth providers to use insecure HTTP or local URLs. */
   allowInsecureExternalAuthUrl?: boolean;
+
+  /** Optional list of enabled search parameters by SearchParameter.id.  Default is all search parameters enabled. */
```
mattlong: Worth mentioning that medplum resource type search params are always enabled regardless of this setting?

packages/server/src/workers/dispatch.ts
```diff
@@ -113,6 +114,10 @@ async function addDispatchJobData(job: DispatchJobData): Promise<void> {
  * @param job - The dispatch job details.
  */
 export async function execDispatchJob(job: Job<DispatchJobData>): Promise<void> {
+  if (!getConfig().dispatchEnabled) {
```
mattlong, line 117: Should the config also disable even queueing the events?

packages/server/src/workers/subscription.ts
```diff
@@ -313,6 +313,10 @@ export async function addSubscriptionJobs(
   context: BackgroundJobContext,
   options?: ResendSubscriptionsOptions
 ): Promise<void> {
+  if (!getConfig().subscriptionsEnabled) {
```
mattlong, line 316: Same question here, seems like the config should dictate of jobs get queued or not.

mattlong (changes requested):

packages/server/src/workers/dispatch.ts
codyebberson: mm, good call

packages/server/src/workers/subscription.ts
codyebberson: This is pre-queue (`addSubscriptionJobs`) and handles multiple entry points (normal dispatch flow and also `$resend` operation).

mattlong (approved):

PR #9749, opened 2026-07-05
Implement MCP search and fetch tools

### LINKED ISSUE #9616: Implement the MCP 'search' and 'fetch' tools (currently dummy stubs)
In `packages/server/src/mcp/server.ts`, `search` (L31) and `fetch` (L36) currently return a hardcoded dummy document. From the comment they look like placeholders so ChatGPT will connect, rather than working tools yet.

The only functional tool is `fhir-request` (L51), which returns `JSON.stringify(response)`. For an LLM client this is the core problem: a real patient can have thousands of resources, and a single search can return a large bundle, so dumping the full raw response into the model's context is token-heavy and blows past the context window fast. There is no read-only or token-bounded option, and the L48 comment already flags the raw passthrough as suboptimal.

Implementing `search` and `fetch` as real, read-only tools would give clients a bounded path:
- `search`: typed params for a resource type, returning a small projected result set with bounded counts instead of the full bundle
- `fetch`: read a single resource by reference

I would back both with `ctx.repo` so the user's AccessPolicy still applies, and keep them strictly read-only. That keeps payloads small enough to actually fit in context, and read-only also seems like a sensible first step toward the PHI support the docs note is not there yet.

Happy to open the PR if this direction fits.

driavysinus:
## Summary

Fixes #9616.

This replaces the placeholder MCP `search` and `fetch` tools with read-only FHIR-backed implementations:

- `search` parses a FHIR search query, defaults `_count` to the core default, clamps requested counts to 100, and returns MCP search results with FHIR references, display titles, and resource URLs.
- `fetch` reads a single FHIR resource by reference and returns the resource JSON plus basic metadata.
- Both tools declare read-only/idempotent MCP annotations and run through the authenticated repository context, so existing access policy checks are applied.

## Tests

- `npm --workspace @medplum/server exec eslint -- src/mcp/server.ts src/mcp/routes.test.ts`
- `npm --workspace @medplum/server run build`
- `npm --workspace @medplum/server test -- src/mcp/routes.test.ts` *(blocked locally: app setup cannot connect to local Postgres/Redis, failing during seed before the MCP test body runs)*

### DIFF
--- a/packages/docs/docs/ai/mcp.md
+++ b/packages/docs/docs/ai/mcp.md
@@ -26,6 +26,46 @@ _Note: LLMs can sometimes cache sessions. If you experience issues, try disconne
 
 The Medplum MCP integration exposes several powerful tools for interacting with FHIR data. Here are the tools available in your MCP server:
 
+#### `search`
+
+- **Title:** FHIR Search
+- **Description:** Performs a read-only FHIR search through the authenticated repository. Pass a standard FHIR search string such as `Patient?name=Smith&_count=10`. Results contain only `id`, `title`, and `url`; `_count` defaults to 20 and is capped at 100. Use `_offset` for pagination and `_total=accurate` when the total number of matches is needed. `_include` and `_revinclude` are ignored; retrieve related resources explicitly with `fetch`.
+- **Annotations:** Read-only and idempotent.
+- **Schema:**
+
+```json
+{
+  "type": "object",
+  "properties": {
+    "query": {
+      "type": "string",
+      "description": "FHIR search string, for example 'Patient?name=Smith&_count=10'."
+    }
+  },
+  "required": ["query"]
+}
+```
+
+#### `fetch`
+
+- **Title:** FHIR Read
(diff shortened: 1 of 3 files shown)


driavysinus:
DCO is fixed on the latest commit. The remaining failures are Vercel preview deployments requiring Medplum team
  authorization for this fork PR.
  Could a team member please authorize the Vercel deployments / approve the checks?

driavysinus:
Update: I rebased this branch onto the current `main` and pushed ceaadcab9. `search` now enforces the default/cap at the repository boundary, strips `_include`/`_revinclude`, returns match entries only, and includes `total` when requested. `fetch` validates literal `ResourceType/id` references, rejects `Binary`, bounds titles and large resource payloads, and both tools return structured FHIR `OperationOutcome` errors. The MCP docs and regression coverage were updated as well.

Validation:
- ESLint and Prettier pass
- `@medplum/core` and `@medplum/server` builds pass
- The targeted MCP integration suite still requires local Postgres/Redis; in this environment it stops during app seeding with `ECONNREFUSED` before the test body runs

Because #9783 overlaps this work, @ThatOneBro @mattlong please let me know which PR you’d prefer to continue. I’m happy to coordinate either way.

PR #9752, opened 2026-07-06
fix(ValueSet): Intersect valueSet expansion with system/filter (#9057)

### LINKED ISSUE #9057: ValueSet/$expand ignores compose.include.system and concept when include.valueSet is set

sahilnyk:
Fixes #9057

## What this change does

When a `compose.include` references a valueSet AND also constrains it with a `system`, `concept`, or `filter`, the expansion now returns only the codes common to both, instead of ignoring the system/concept/filter as the old code did.

The outer `system` is applied before the referenced valueSet expands, and the outer `filter` is layered into the expansion query so it runs before the `MAX_EXPANSION_SIZE` truncation. Remaining `concept`/`filter` cases are intersected after expansion.

## Known limitation

A nested valueSet defined by an explicit `concept` list plus an outer filter matching more than 1000 codes can still truncate at `MAX_EXPANSION_SIZE` before the intersection. Same 1000 code ceiling that applies across expansion today, left as a follow up.

## Testing

- [x] `vitest`: 3 new tests pass (multi system truncation, filter layered before truncation, system plus filter intersection).
- [x] `prettier --check`, `eslint`, `tsc --noEmit`: clean on both files.


packages/server/src/fhir/operations/expand.ts
```diff
@@ -181,26 +181,52 @@ async function computeExpansion(
...
+            if (c) {
+              systemExpansion.push({ system: codeSystem.url, code: c.code, display: c.display });
+            }
+          }
+        } else {
+          await includeInExpansion(include, systemExpansion, codeSystem, params);
+        }
+
+        const validKeys = new Set(systemExpansion.map((c) => `${c.system}|${c.code}`));
+        filtered = vsExpansion.filter((c) => validKeys.has(`${c.system}|${c.code}`));
```
mattwiller: Manually intersecting two expansions could cause issues, for example in the following scenario:

```jsonc
{
  "resourceType": "ValueSet",
  "status": "draft",
  "url": "http://example.com/ValueSet/nested",
  "compose": {
    "include": [
      // Contains all of two large code systems, each with thousands of codes
      { "system": "http://example.com/CodeSystem/A" },
      { "system": "http://example.com/CodeSystem/B" }
    ]
  }
}

{
  "resourceType": "ValueSet",
  "status": "draft",
  "url": "http://example.com/ValueSet/compound",
  "compose": {
    "include": [
      {
        "valueSet": "http://example.com/ValueSet/nested",
        "system": "http://example.com/CodeSystem/B"
      }
    ]
  }
}
```

I believe that in this case, the expansion of the nested ValueSet would return the first 1,000 concepts from `CodeSystem/A` and would incorrectly result in an empty intersection with `CodeSystem/B`.  This issue stems from using the (truncated) expansions as intermediate steps: instead, the code will likely need to construct the correct SQL query to layer the `system` and `filter` rules from the outer ValueSet onto the expansion query of the included ValueSet

mattwiller (changes requested):

packages/server/src/fhir/operations/expand.ts
sahilnyk: umm right. I will fix it by applying the outer system and filter rules directly into the nested ValueSet's expansion query, so the filtering happens at the SQL level before the count limit.

sahilnyk:
@mattwiller Reworked this so the outer system gets pushed into the nested ValueSet's compose before expansion runs,
  instead of filtering after the count limit. In your example that drops the CodeSystem/A include up
  front and only expands CodeSystem/B, so no empty intersection.

  Also handled pre-expanded nested ValueSets (the ones with expansion.contains and no compose) since they
  skip the compose filter. Added tests for both paths, all 52 in expand.test.ts pass.

  One caveat: a single huge {system: X} include (1000+ codes) with an outer filter could still truncate
  before the filter applies. Fixing that fully means layering the filter into the expansion SQL, can do
  it here or as a follow-up.

sahilnyk:
@mattwiller reworked per your review (filter now layered into the expansion query before truncation) with tests added, could you re-review when you get a chance? Happy to change anything if needed.

mattwiller:
@sahilnyk After some initial investigation, the work to correctly compose `valueSet` and the various other include filters looks significantly too tricky and full of performance edge cases for us  to handle it by extending this branch: we'll need to spin up a dedicated workstream to make sure the various issues are resolved carefully.  I'll close this PR for now, since the later work will supersede it

PR #9756, opened 2026-07-06
feat(app): add version picker to Agent Tools page

### LINKED ISSUE #9757: [Agent] Add UI for upgrading to a specific version to Medplum App
Currently you can only upgrade to the latest version from Medplum App. The `$upgrade` operation already supports upgrading to any available version, but the UI is not there for that functionality. We should add a version picker to Medplum App to allow users to upgrade to the version that they want.

Caveat: We should warn users about downgrading their agent version, since that could introduce regressions.

ThatOneBro:
Adds a version picker, allowing users to select an exact version to upgrade to, rather than just latest as previously allowed. When attempting to downgrade versions (go to a lower version), a prompt will warn the user about potential regressions in features and bugs before allowing the user to downgrade.

Fixes #9757


packages/core/src/version-utils.ts
```diff
@@ -39,6 +39,46 @@ export function assertReleaseManifest(candidate: unknown): asserts candidate is
...
+  }
+  const res = await fetch(url.toString());
+  if (res.status !== 200) {
+    let message: string | undefined;
+    try {
+      message = ((await res.json()) as { message: string }).message;
+    } catch (err) {
+      console.error(`Failed to parse message from body: ${normalizeErrorString(err)}`);
+    }
+    throw new Error(`Received status code ${res.status} while ${errorContext}. Message: ${message}`);
```
mattlong: Shouldn't error messages typically be static with details in the second `cause` object?

packages/core/src/version-utils.ts
```diff
@@ -97,6 +122,24 @@ export function isValidMedplumSemver(version: string): boolean {
   return /^\d+\.\d+\.\d+(-[0-9a-z]{7})?$/.test(version);
 }
 
+/**
+ * Compares two Medplum semver version strings, ignoring any trailing commit-hash suffix.
+ * @param a - The first version to compare.
+ * @param b - The second version to compare.
+ * @returns A negative number if `a` is older than `b`, a positive number if `a` is newer than `b`, or `0` if they resolve to the same release.
+ */
+export function compareVersions(a: string, b: string): number {
```
mattlong: This feels a bit fragile since its only meant to support standard `MAJOR.MINOR.PATCH-HASH` formats. If we were to ever publish any other sort of version string it may not behave as expected. You're already filtering inputs that make it to this function, so it'd be nice to avoid misuse of it by locking down the inputs with a template literal type matching your regex. I We don't use template literals much, but it feels warranted here:

```typescript
type MedplumSemver = `${number}.${number}.${number}` | `${number}.${number}.${number}-${string}`;

export function isValidMedplumSemver(version: string): version is MedplumSemver {
  return /^\d+\.\d+\.\d+(-[0-9a-z]{7})?$/.test(version);
}
```

mattlong (changes requested):

packages/core/src/version-utils.ts
ThatOneBro: I think we could move the message into another error and put it in the `cause` prop, yeah

mattlong (approved):

PR #9759, opened 2026-07-06
fix: prevent logger serialization errors from crashing the server process

amcgivern:
If an unhandled promise rejection carries an object with circular references — e.g. an AbortError from an aborted HTTPS request referencing a live IncomingMessage → TLSSocket → ClientRequest graph — the unhandledRejection handler logs it, JSON.stringify throws on the cycle, and the throw escapes the handler as an uncaughtException, which exits the process. A log line can kill the server.

## What

- `Logger.log` tries plain `JSON.stringify` first; on failure it retries with a replacer that emits `"[Circular]"` for cycles, then falls back to a minimal message if serialization still fails — so logging never throws.
- Tests: a circular AbortError through the core logger and through the server's installed `unhandledRejection` listener.

### DIFF
--- a/packages/core/src/logger.ts
+++ b/packages/core/src/logger.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { isError } from './outcomes';
+import { isError, normalizeErrorString } from './outcomes';
 
 /*
  * Once upon a time, we used Winston, and that was fine.
@@ -134,7 +134,7 @@ export class Logger implements ILogger {
     }
 
     this.write(
-      JSON.stringify({
+      stringifyLogMessage({
         level: LogLevelNames[level],
         timestamp: new Date().toISOString(),
         msg: this.prefix ? `${this.prefix}${msg}` : msg,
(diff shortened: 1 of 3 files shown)


packages/core/src/logger.ts
```diff
@@ -145,6 +145,51 @@ export class Logger implements ILogger {
...
+ * @param logMessage - The log message to stringify.
+ * @returns The JSON string representation of the log message.
+ */
+function stringifyLogMessage(logMessage: Record<string, any>): string {
+  try {
+    return JSON.stringify(logMessage);
+  } catch {
+    try {
+      return JSON.stringify(logMessage, createCircularReplacer());
+    } catch (err) {
```
mattwiller, line 164: Is there a reason not to always use `JSON.stringify(logMessage, createCircularReplacer())` here?

mattwiller (commented):
Thanks for the PR @amcgivern — this looks good to me.  I just had one question about the implementation

packages/core/src/logger.ts
codyebberson: Must resist temptation to micro benchmark `createCircularReplacer()`

ThatOneBro: De-opting `JSON.stringify` is very dangerous to performance. Definitely preferable to try .. catch and only then try here

PR #9761, opened 2026-07-06
Fix transaction bundle attachment references

### LINKED ISSUE #8664: convertToTransaction bundle should replace Attachment.url

driavysinus:
## What changed

- Updated `convertToTransactionBundle()` reference rewriting so `Attachment.url` values that point to bundled resources are remapped to the generated `urn:uuid` fullUrl.
- Kept URL rewriting scoped to Attachment-like objects, so non-attachment URL fields such as `Extension.url` are preserved.

## Why

Fixes #8664.

## Validation

- `npm test --workspace=@medplum/core -- bundle.test.ts`


driavysinus:
Update: DCO, labeler, Socket, Snyk, and Chromatic/UI checks are green on the latest commit. The remaining Vercel statuses point to Medplum team authorization for this fork rather than build failures.

@mattlong, when you have a chance, could you review this fix and authorize the preview deployments?

PR #9762, opened 2026-07-06
[codex] Fail CLI bot deploy on error outcome

### LINKED ISSUE #3692: Bot Deployment Silently Fails in CLI

driavysinus:
## What changed

- Updated `deployBot()` to treat non-OK `OperationOutcome` responses from `Bot/$deploy` as deployment failures.
- Added a regression test that returns a `notFound` outcome from `$deploy` and verifies `medplum bot deploy` exits non-zero.

## Why

Fixes #3692. Previously the CLI printed the deploy result text but did not fail when the server returned an error `OperationOutcome`, which could let CI pass after a failed bot deployment.

## Validation

- `npm test --workspace=@medplum/cli -- bot.test.ts`

### DIFF
--- a/packages/cli/src/bot.test.ts
+++ b/packages/cli/src/bot.test.ts
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { allOk } from '@medplum/core';
+import { allOk, notFound } from '@medplum/core';
 import type { Bot } from '@medplum/fhirtypes';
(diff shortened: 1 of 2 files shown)


ThatOneBro (approved):

codyebberson:
Thanks @driavysinus !

PR #9763, opened 2026-07-06
[codex] Preserve agent logs on uninstall

### LINKED ISSUE #3672: Agent uninstall removes all files but should leave created log files in-place

driavysinus:
## What changed

- Replaced recursive deletion of the Windows Agent install directory during uninstall with deletion of installer-managed files only.
- Left `$INSTDIR` in place when runtime logs or other generated files remain.
- Still removes `$INSTDIR` when it becomes empty after deleting known installed files.

## Why

Fixes #3672. The previous uninstaller used `RMDir /r /REBOOTOK "$INSTDIR"`, which removed generated Agent log files that may be needed for retention or troubleshooting.

## Validation

- `git diff --check`

`makensis` is not installed in my local environment, so I could not compile the NSIS installer script locally.

### DIFF
--- a/packages/agent/installer.nsi
+++ b/packages/agent/installer.nsi
@@ -381,8 +381,17 @@ Section Uninstall
     # Uninstall the Start menu shortcuts
     RMDir /r /REBOOTOK "$SMPROGRAMS\${APP_NAME}"
 
-    # Delete the files
-    RMDir /r /REBOOTOK "$INSTDIR"
+    # Delete installer-managed files, but preserve runtime logs in $INSTDIR for retention.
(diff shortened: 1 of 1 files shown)


ThatOneBro:
Hello @driavysinus, thanks for the PR but we have closed the mentioned issue in #9693. Closing this one out

PR #9764, opened 2026-07-06
Filter included ValueSet expansions by concepts

### LINKED ISSUE #9057: ValueSet/$expand ignores compose.include.system and concept when include.valueSet is set

driavysinus:
## Summary

- Fix `ValueSet/$expand` when a single `compose.include` imports another ValueSet and also specifies `system` / `concept` constraints.
- Intersect nested ValueSet expansions with the include's `system` and `concept` values before adding them to the parent expansion.
- Add an API regression test and a small unit test for the filtering helper.

Fixes #9057

## Testing

- `npm test --workspace=@medplum/server -- src/fhir/operations/expand.unit.test.ts`
- `npx eslint src/fhir/operations/expand.ts src/fhir/operations/expand.test.ts src/fhir/operations/expand.unit.test.ts`

Unable to run `npm test --workspace=@medplum/server -- src/fhir/operations/expand.test.ts` locally because Postgres/Redis are not running in this environment (`ECONNREFUSED` during `initAppServices`).

### DIFF
--- a/packages/server/src/fhir/operations/expand.ts
+++ b/packages/server/src/fhir/operations/expand.ts
@@ -190,11 +190,11 @@ async function computeExpansion(
           includedValueSet,
           {
             ...params,
-            count: maxCount - expansion.length,
+            count: include.system || include.concept ? MAX_EXPANSION_SIZE : maxCount - expansion.length,
           },
           terminologyResources
         );
-        expansion.push(...nestedExpansion);
+        expansion.push(...filterExpansionByInclude(nestedExpansion, include).slice(0, maxCount - expansion.length));
 
         if (expansion.length >= maxCount) {
           // Skip further expansion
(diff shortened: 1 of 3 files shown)


packages/server/src/fhir/operations/expand.ts
```diff
@@ -236,6 +236,21 @@ async function computeExpansion(
   return expansion;
 }
 
+export function filterExpansionByInclude(
```
mattwiller: Post-filtering the results of the initial expansion will not produce correct results for large ValueSets over multiple code systems: the addition of filters onto the nested ValueSet expansion must happen at initial query time

mattwiller (changes requested):

driavysinus:
Thanks for the review. I pushed a follow-up commit that carries the include `system` / `concept` constraints into nested ValueSet expansion instead of only post-filtering the initially expanded result. I also fixed the DCO signoff on the latest commit.

Could you please take another look when you have a chance?

mattwiller:
@driavysinus We're going to need to handle this as part of a broader effort to make `valueSet` compose with the other filters, not just `concept`.  After some initial investigation, that work looks significantly too tricky and full of performance edge cases for us  to handle it by extending this branch: we'll need to spin up a dedicated workstream to make sure the various issues are resolved carefully.  I'll close this PR for now, since the later work will supersede it

PR #9773, opened 2026-07-07
feat(provider): Include lab/uploads filters for Documents

techdavidy:
Allow to filter documents between labs or any other upload. 

Screenshot 2026-07-10 at 12 49 46 PM


techdavidy:
@claude review

examples/medplum-provider/src/pages/patient/DocumentsPage.tsx
```diff
@@ -23,42 +26,57 @@ export function DocumentsPage(): JSX.Element {
...
-  // Patient-scoped DocumentReference search. The page offset/count AND the sort are read from the
-  // URL — sort is URL-driven, so deep links and the pagination round-trip stay authoritative. When
-  // the URL carries no `_sort`, we fall back to most-recent-first (and the effect below pins that
-  // default into the URL).
-  const search = useMemo<SearchRequest>(() => {
+  // The URL is the source of truth for the FULL search sent to the Medplum client — filters,
+  // sort, count, offset, and total. The subject filter is the one exception: the patient is
+  // already in the path, so it stays out of the query string. Missing pieces are defaulted
+  // here, and the effect below pins the normalized query back into the URL so the address bar
+  // always shows exactly what is fetched.
```
noahsilas, line 33: It seems like we have a lot of variants of code like this that parses querystring params + some defaults from the component into a SearchRequest, and then syncs updates back into the querystring. I wonder if we can extract a reusable hook that encapsulates a lot of that work?

Not a blocker for this PR

examples/medplum-provider/src/pages/patient/DocumentsPage.tsx
```diff
@@ -23,42 +26,57 @@ export function DocumentsPage(): JSX.Element {
...
   }, [location.search, patientId]);
 
-  // Sort is URL-driven; when the URL carries no `_sort`, pin the default into the URL (history
-  // replace) so the address bar always reflects the active sort. The resulting search is deep-equal
-  // to the pre-redirect one, so ResourceBoard's memoized search does not refetch.
+  // Serializes a search for the URL: the subject filter is dropped because the patient id is
+  // carried by the path.
+  const toQuery = useCallback(
+    (s: SearchRequest): string =>
+      formatSearchQuery({ ...s, filters: (s.filters ?? []).filter((f) => f.code !== 'subject') }),
```
noahsilas, line 62: Above in `extraFilters`, we are excluding `status` as well, saying it is coming from the route. Should we drop it here too?

examples/medplum-provider/src/pages/patient/DocumentSourceFilterMenu.utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  readonly source: 'lab' | 'upload';
+  readonly label: string;
+  readonly headerText: string;
+  readonly filters: Filter[];
+}
+
+// Each source is identified by marks on the document itself: Health Gorilla stamps its
+// identifier on every lab document it writes (reports, specimen labels, requisitions), while
+// uploads have no identifier and no related resource. (Its debug-log documents carry a
+// ServiceRequest in context.related instead, which is why Upload also requires related:missing.)
```
noahsilas: I'm having a tough time understanding this. I think particularly "Its debug-log documents" leaves me unclear on what "It" is and what a "debug-log document" is.

"Upload" documents being ones that have no identifiers or related context documents is interesting; this feels a bit like something that will eventually be confusing. Is there an opportunity to attach a specific identifier to "uploads" when they are being added?

It sort of feels like we should also have an "other" category for any documents that are not in either of these categories - right now a document reference with any non-HG identifier would be invisible in this UI, right?

examples/medplum-provider/src/pages/patient/DocumentsPage.tsx
techdavidy: in this case we would want to drop the entered in error status

examples/medplum-provider/src/pages/patient/DocumentSourceFilterMenu.utils.ts
techdavidy: We have DocumentReferences that works as debug payloads from integrations so they get filtered. 
We can create a new filter named 'other' for any other documents.

examples/medplum-provider/src/pages/patient/DocumentSourceFilterMenu.utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    headerText: 'Lab Documents',
+    filters: [{ code: 'identifier', operator: Operator.EQUALS, value: `${HEALTH_GORILLA_SYSTEM}|` }],
+  },
+  {
+    source: 'other',
+    label: 'Other Documents',
+    headerText: 'Other Documents',
+    filters: [
+      { code: 'identifier', operator: Operator.MISSING, value: 'true' },
+      { code: 'related', operator: Operator.MISSING, value: 'true' },
```
noahsilas, line 36: I think getting explicit about the intention of this filter will be helpful for future readers:
```suggestion
      // The Health Gorilla integration can create DocumentReference entries
      // containing debugging information; we don't want to display those here.
      // Such documents have a reference to a `ServiceRequest` in `context.related`.
      { code: 'related', operator: Operator.MISSING, value: 'true' },
```

I wonder if we can add some more explicit way to identify these debug documents, as this does seem likely to eventually conflict with some real document carrying a "related" reference. That tracking is maybe better done as a separate issue though, maybe we can follow up on that separately.

examples/medplum-provider/src/pages/patient/DocumentSourceFilterMenu.utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    label: 'Lab',
+    headerText: 'Lab Documents',
+    filters: [{ code: 'identifier', operator: Operator.EQUALS, value: `${HEALTH_GORILLA_SYSTEM}|` }],
+  },
+  {
+    source: 'other',
+    label: 'Other Documents',
+    headerText: 'Other Documents',
+    filters: [
+      { code: 'identifier', operator: Operator.MISSING, value: 'true' },
```
noahsilas: This excludes all documents that have _any_ identifier, right? Should this be using a more targeted query like `operator: Operator.NOT_EQUALS, value: `${HEALTH_GORILLA_SYSTEM}|`?

examples/medplum-provider/src/pages/patient/DocumentSourceFilterMenu.utils.ts
noahsilas: This comment should probably remove the "uploads" reference now that the second tab is renamed to "other".

noahsilas (approved):

PR #9779, opened 2026-07-07
Fixes for pagination and job recovery in $set-accounts

mattwiller:
- Fixed search pagination to handle new `cursor` output from `Patient/$everything`
- Added verbose queue logging for async `$set-accounts`
- Check for `AsyncJob.status` on each `$set-accounts` loop iteration to fail fast on job cancel

### DIFF
--- a/packages/server/src/workers/set-accounts.ts
+++ b/packages/server/src/workers/set-accounts.ts
@@ -13,7 +13,7 @@ import { getShardSystemRepo } from '../fhir/repo';
 import { PLACEHOLDER_SHARD_ID } from '../fhir/sharding';
 import type { AuthState } from '../oauth/middleware';
 import type { WorkerInitializer, WorkerInitializerOptions } from './utils';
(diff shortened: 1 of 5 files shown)


packages/server/src/workers/utils.ts
```diff
@@ -308,3 +308,5 @@ export function defaultQueueOptions(config: MedplumServerConfig): QueueOptions {
     },
   };
 }
+
+export class CancelledError extends Error {}
```
ThatOneBro: Should we set the name in constructor?

packages/server/src/workers/set-accounts.ts
```diff
@@ -53,6 +53,9 @@ export const initSetAccountsWorker: WorkerInitializer = (config, options?: Worke
         ...workerBullmq,
       }
     );
+    addVerboseQueueLogging<SetAccountsJobData>(queue, worker, (job) => ({
```
ThatOneBro, line 56: Adding the logging infra 👍

ThatOneBro (approved):

PR #9780, opened 2026-07-08
fix(provider): Use ListWithDetailPane in tasks

techdavidy:
Provide a consisten UI/UX in TaskBoard by using ListWithDetailPane.


techdavidy:
@claude review

examples/medplum-provider/src/components/fax/FaxBoard.tsx
```diff
@@ -120,14 +120,8 @@ export function FaxBoard({
...
-        renderItem={(fax, { selected, index, items }) => (
-          <FaxListItem
-            fax={fax}
-            selectedFax={selected ? fax : undefined}
-            activeTab={activeTab}
-            getFaxUri={getFaxUri}
-            hideDivider={selected || (index < items.length - 1 && items[index + 1]?.id === faxId)}
-          />
+        renderItem={(fax, { selected }) => (
+          <FaxListItem fax={fax} selectedFax={selected ? fax : undefined} activeTab={activeTab} getFaxUri={getFaxUri} />
```
noahsilas, line 124: Unrelated: while we're in here, maybe we should update the `selectedFax` prop to be `isSelected` or similar 🤔

examples/medplum-provider/src/components/tasks/TaskBoard.tsx
```diff
@@ -117,10 +101,10 @@ export function TaskBoard({
         return;
       }
 
-      let results: Task[] = [];
+      let results: WithId<Task>[] = [];
 
       if (bundle.entry) {
-        results = bundle.entry.map((entry) => entry.resource as Task).filter((r): r is Task => r !== undefined);
+        results = bundle.entry.map((entry) => entry.resource).filter((r): r is WithId<Task> => r !== undefined);
```
noahsilas: My first thought here was that we could use a helper from core:
```suggestion
        results = bundle.entry.map((entry) => entry.resource).filter((r) => isResourceWithId<Task>(r, 'Task'));
```

We can do a little better though: the type of `entry.resource` is being inferred as `WithId<Task> | undefined`, so we just need to exclude the `undefined` case. For that, there's a simpler helper in core:
```suggestion
        results = bundle.entry.map((entry) => entry.resource).filter(isDefined);
```

Maybe even better though would be to use the `searchResources` method of MedplumClient so that the return type starts out as exactly what we want in here:

```ts
    try {
      const tasks = await medplum.searchResources('Task', query, { cache: 'no-cache' });
      const total = tasks.bundle.total;
      let results = [...tasks];
     ...
```

noahsilas (approved):
Some minor cleanup opportunities, nothing blocking. Nice tidy up! 🎉

PR #9782, opened 2026-07-08
feat(react): Display address in DiagnosticReportDisplay

techdavidy:
Screenshot 2026-07-08 at 7 03 12 PM
Screenshot 2026-07-08 at 7 03 03 PM

### DIFF
--- a/packages/react/src/DiagnosticReportDisplay/DiagnosticReportDisplay.tsx
+++ b/packages/react/src/DiagnosticReportDisplay/DiagnosticReportDisplay.tsx
@@ -15,6 +15,7 @@ import { useMedplum, useResource } from '@medplum/react-hooks';
 import cx from 'clsx';
 import type { JSX } from 'react';
 import { useEffect, useState } from 'react';
+import { AddressDisplay } from '../AddressDisplay/AddressDisplay';
 import { CodeableConceptDisplay } from '../CodeableConceptDisplay/CodeableConceptDisplay';
 import { MedplumLink } from '../MedplumLink/MedplumLink';
 import { NoteDisplay } from '../NoteDisplay/NoteDisplay';
@@ -96,7 +97,7 @@ interface DiagnosticReportHeaderProps {
 
 function DiagnosticReportHeader({ value, hideSubject = false }: DiagnosticReportHeaderProps): JSX.Element {
   return (
-    <Group mt="md" gap={30}>
+    <Group mt="md" gap={30} align="flex-start">
       {value.subject && !hideSubject && (
         <div>
           <Text size="xs" tt="uppercase" c="dimmed">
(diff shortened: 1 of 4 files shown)


packages/react/src/DiagnosticReportDisplay/DiagnosticReportDisplay.tsx
```diff
@@ -141,6 +142,25 @@ function DiagnosticReportHeader({ value, hideSubject = false }: DiagnosticReport
...
+  const performer = useResource(value);
+  const address = performer?.resourceType === 'Organization' ? performer.address?.[0] : undefined;
+  return (
+    <>
+      <ResourceBadge value={value} link={true} />
+      {address && (
+        <Text size="sm" c="dimmed" mt={4}>
+          <AddressDisplay value={address} />
+        </Text>
+      )}
```
noahsilas, line 172: This feels a little awkwardly sized/spaced in the screenshot, I wonder if this should go down to `size="xs"`? Should it be aligned under the `ResourceName` instead of under the avatar?

noahsilas (approved):

packages/react/src/DiagnosticReportDisplay/DiagnosticReportDisplay.tsx
```diff
@@ -95,52 +96,84 @@ interface DiagnosticReportHeaderProps {
...
+        </Stack>
       )}
-      {value.status && (
-        <div>
-          <Text size="xs" tt="uppercase" c="dimmed">
-            Status
-          </Text>
-          <StatusBadge status={value.status} />
-        </div>
+      {hasLeftColumn && hasRightColumn && <Divider orientation="vertical" />}
```
noahsilas, line 118: I wonder if we can create a component similar to `<Group>` that adds these dividers automatically between entries. It would be kind of nice to be able to write this like:
```tsx
<Group gap="xl" dividers>
  <LeftColumn />
  {hasRightColumn && (<RightColumn />)}
</Group>
```

I think this should be doable, although there are some tricky edge cases around empty containers to think about...

Looks like Mantine is open to [adding a feature like this](https://github.com/orgs/mantinedev/discussions/2367), maybe worth exploring this upstream instead of creating our own wrapping component?

packages/react/src/DiagnosticReportDisplay/DiagnosticReportDisplay.tsx
```diff
@@ -95,52 +96,84 @@ interface DiagnosticReportHeaderProps {
...
-      ))}
-      {value.issued && (
-        <div>
-          <Text size="xs" tt="uppercase" c="dimmed">
-            Issued
-          </Text>
-          <Text>{formatDateTime(value.issued)}</Text>
-        </div>
+    <Group mt="md" gap="xl" align="stretch" wrap="nowrap">
+      {hasLeftColumn && (
```
noahsilas, line 104: Would it be helpful to leave a placeholder element when there is no left column so that the "right column" components are always visually in the same place?

Probably not a serious issue as I suspect that `subject` will nearly always be present.

noahsilas (approved):

PR #9785, opened 2026-07-08
fix(server): honor target type in _include and _revinclude

### LINKED ISSUE #9765: _include and _revInclude targetType are not supported
The FHIR spec specifies that targetType is a valid parameter to search by, for example `_include: ["AuditEvent:entity:Task"]`, where targetType is "Task".

The targetType ends up getting dropped and all entities are being returned in the bundle. The expectation is that only entities of resourceType === targetType are returned. This impacts both `_include` and `_revInclude` I believe. 

**Script to reproduce**

```
/**
 * Repro: Medplum ignores the _include target-type suffix.
 *
 * FHIR: _include=AuditEvent:entity:Task should only include Task resources
 * referenced from AuditEvent.entity.what — not other entity references
 * (e.g. another AuditEvent).
 *
 * Bug: getSearchIncludeEntries evaluates AuditEvent.entity.what for every
 * reference and never filters by include.targetType.
 *
 * Usage:
 *   MEDPLUM_BASE_URL=https://api.medplum.com \
 *   MEDPLUM_CLIENT_ID=... \
 *   MEDPLUM_CLIENT_SECRET=... \
 *   npx tsx repro-include-target-type-bug.ts
 *
 * Or with an access token:
 *   MEDPLUM_BASE_URL=... MEDPLUM_ACCESS_TOKEN=... npx tsx repro-include-target-type-bug.ts
 */

import { MedplumClient, getReferenceString } from '@medplum/core';
import type { AuditEvent, Bundle, Task } from '@medplum/fhirtypes';

const baseUrl = process.env.MEDPLUM_BASE_URL ?? 'http://localhost:8103';
const accessToken = process.env.MEDPLUM_ACCESS_TOKEN;
const clientId = process.env.MEDPLUM_CLIENT_ID;
const clientSecret = process.env.MEDPLUM_CLIENT_SECRET;

const medplum = new MedplumClient({ baseUrl, ...(accessToken ? { accessToken } : {}) });

function includedResources(bundle: Bundle): string[] {
  return (bundle.entry ?? [])
    .filter((entry) => entry.search?.mode === 'include')
    .map((entry) => getReferenceString(entry.resource!)!)
    .sort();
}

async function main(): Promise<void> {
  if (!accessToken) {
    if (!clientId || !clientSecret) {
      throw new Error('Set MEDPLUM_ACCESS_TOKEN or MEDPLUM_CLIENT_ID + MEDPLUM_CLIENT_SECRET');
    }
    await medplum.startClientLogin(clientId, clientSecret);
  }

  const recorded = new Date().toISOString();

  console.log('Creating Task...');
  const task = await medplum.createResource<Task>({
    resourceType: 'Task',
    status: 'completed',
    intent: 'order',
    description: 'include-target-type repro',
  });

  console.log('Creating referenced AuditEvent...');
  const referencedAuditEvent = await medplum.createResource<AuditEvent>({
    resourceType: 'AuditEvent',
    recorded,
    type: { system: 'http://terminology.hl7.org/CodeSystem/audit-event-type', code: 'rest' },
    action: 'R',
    agent: [{ requestor: true }],
    source: { observer: { display: 'repro' } },
  });

  console.log('Creating AuditEvent with entity -> Task + AuditEvent...');
  const auditEvent = await medplum.createResource<AuditEvent>({
    resourceType: 'AuditEvent',
    recorded,
    type: { system: 'http://terminology.hl7.org/CodeSystem/audit-event-type', code: 'rest' },
    action: 'E',
    agent: [{ requestor: true }],
    source: { observer: { display: 'repro' } },
    entity: [
      { what: { reference: `Task/${task.id}` } },
      { what: { reference: `AuditEvent/${referencedAuditEvent.id}` } },
    ],
  });

  console.log('\n--- Search WITHOUT target type ---');
  console.log(`GET AuditEvent?_id=${auditEvent.id}&_include=AuditEvent:entity\n`);

  const unrestricted = await medplum.search('AuditEvent', {
    _id: auditEvent.id!,
    _include: 'AuditEvent:entity',
  });

  const unrestrictedIncluded = includedResources(unrestricted);
  console.log('Included resources:', unrestrictedIncluded);
  console.log('Expected: both Task and referenced AuditEvent');

  console.log('\n--- Search WITH :Task target type ---');
  console.log(`GET AuditEvent?_id=${auditEvent.id}&_include=AuditEvent:entity:Task\n`);

  const restricted = await medplum.search('AuditEvent', {
    _id: auditEvent.id!,
    _include: 'AuditEvent:entity:Task',
  });

  const restrictedIncluded = includedResources(restricted);
  console.log('Included resources:', restrictedIncluded);
  console.log('Expected: Task only');
  console.log(`Actual:   ${restrictedIncluded.join(', ')}`);

  const bugPresent =
    restrictedIncluded.includes(`AuditEvent/${referencedAuditEvent.id}`) &&
    restrictedIncluded.includes(`Task/${task.id}`);

  console.log('\n--- Result ---');
  if (bugPresent) {
    console.log('BUG REPRODUCED: _include=AuditEvent:entity:Task included a non-Task resource.');
  } else if (restrictedIncluded.length === 1 && restrictedIncluded[0] === `Task/${task.id}`) {
    console.log('OK: only Task was included.');
  } else {
    console.log('Unexpected result — inspect bundle entries above.');
  }

  console.log('\nCreated resource IDs:');
  console.log(`  Task/${task.id}`);
  console.log(`  AuditEvent/${referencedAuditEvent.id} (referenced entity)`);
  console.log(`  AuditEvent/${auditEvent.id} (search match)`);
}

main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

A possible fix I have locally

```
async function getSearchIncludeEntries(
  repo: Repository,
  include: IncludeTarget,
  resources: Resource[]
): Promise<BundleEntry[]> {
  // Previously this wasn't using targetType
  const { resourceType, searchParam: code, targetType } = include;
  const searchParam = getSearchParameter(resourceType, code);
  if (!searchParam) {
    throw new OperationOutcomeError(badRequest(`Invalid include parameter: ${resourceType}:${code}`));
  }
  
  const fhirPathResult = evalFhirPathTyped(searchParam.expression as string, resources.map(toTypedValue));
  const references: Reference[] = [];
  const canonicalReferences: string[] = [];
  for (const result of fhirPathResult) {
    if (result.type === PropertyType.Reference) {
      const reference = result.value as Reference;
      // if targetType exists, then check if the reference is of the right type before adding it to included references.
      if (!targetType || isReference(reference, targetType as ResourceType)) {
        references.push(reference);
      }
    } else if (canonicalReferenceTypes.includes(result.type)) {
      canonicalReferences.push(result.value);
    }
  }

  const includedResources = (await repo.readReferences(references)).filter((v) => isResource(v)) as WithId<Resource>[];
  const canonicalTargetTypes = targetType
    ? searchParam.target?.filter((candidateType) => candidateType === targetType)
    : searchParam.target;
  if (canonicalTargetTypes && canonicalReferences.length > 0) {
    const canonicalSearches = canonicalTargetTypes.map((resourceType) => {
      const searchRequest = {
        resourceType: resourceType,
        filters: [
          {
            code: 'url',
            operator: Operator.EQUALS,
            value: canonicalReferences.join(','),
          },
        ],
        count: DEFAULT_MAX_SEARCH_COUNT,
        offset: 0,
      };
      const query = getSelectQueryForSearch(repo, searchRequest);
      return getSearchEntries(repo, searchRequest, query);
    });

    const searchResults = await Promise.all(canonicalSearches);
    for (const result of searchResults) {
      for (const entry of result.entry) {
        includedResources.push(entry.resource as WithId<Resource>);
      }
    }
  }

  return includedResources.map((resource) => ({
    fullUrl: getFullUrl(resource.resourceType, resource.id),
    search: { mode: 'include' },
    resource,
  }));
}
```

Let me know if there's a reason targetType is not implemented and if you're open to a PR to fix this.

oldrobotdev:
Fixes #9765.

`_include` and `_revinclude` take an optional target type, like `_include=AuditEvent:entity:Task`, to limit the include to references of that type. The parser reads the target type already, but the server ignored it: `getSearchIncludeEntries` returned every resource referenced by `AuditEvent.entity.what`, so the `:Task` suffix still pulled in the referenced AuditEvents.

For `_include`, the references are now filtered to the target type before they're read, and the canonical branch only searches the matching target. For `_revinclude`, the base resources are filtered to the target type before they're followed, returning early when none match. Neither path changes when no target type is present.

I added a test for each: an AuditEvent whose `entity.what` references both a Task and another AuditEvent, checking that `_include=AuditEvent:entity:Task` returns only the Task, plus the equivalent revinclude case. Both fail on `main` and pass with the change, and the rest of the search suite stays green.

### DIFF
--- a/packages/server/src/fhir/search.ts
+++ b/packages/server/src/fhir/search.ts
@@ -540,7 +540,7 @@ async function getSearchIncludeEntries(
   include: IncludeTarget,
   resources: Resource[]
 ): Promise<BundleEntry[]> {
-  const { resourceType, searchParam: code } = include;
+  const { resourceType, searchParam: code, targetType } = include;
   const searchParam = getSearchParameter(resourceType, code);
   if (!searchParam) {
     throw new OperationOutcomeError(badRequest(`Invalid include parameter: ${resourceType}:${code}`));
@@ -557,9 +557,18 @@ async function getSearchIncludeEntries(
     }
   }
 
-  const includedResources = (await repo.readReferences(references)).filter((v) => isResource(v)) as WithId<Resource>[];
-  if (searchParam.target && canonicalReferences.length > 0) {
-    const canonicalSearches = searchParam.target.map((resourceType) => {
+  // `_include=ResourceType:code:targetType` restricts the include to references of
+  // `targetType`; without the suffix every referenced resource type is returned.
+  const targetReferences = targetType
+    ? references.filter((reference) => reference.reference?.startsWith(targetType + '/'))
+    : references;
+
+  const includedResources = (await repo.readReferences(targetReferences)).filter((v) =>
+    isResource(v)
+  ) as WithId<Resource>[];
+  const canonicalTargets = targetType ? searchParam.target?.filter((t) => t === targetType) : searchParam.target;
+  if (canonicalTargets?.length && canonicalReferences.length > 0) {
+    const canonicalSearches = canonicalTargets.map((resourceType) => {
       const searchRequest = {
         resourceType: resourceType,
         filters: [
@@ -605,16 +614,37 @@ async function getSearchRevIncludeEntries(
   revInclude: IncludeTarget,
   resources: Resource[]
 ): Promise<BundleEntry[]> {
-  const { resourceType, searchParam: code } = revInclude;
+  const { resourceType, searchParam: code, targetType } = revInclude;
   const searchParam = getSearchParameter(resourceType, code);
   if (!searchParam) {
     throw new OperationOutcomeError(badRequest(`Invalid include parameter: ${resourceType}:${code}`));
   }
 
-  const references =
-    getSearchParameterImplementation(resourceType, searchParam).type === SearchParameterType.CANONICAL
-      ? flatMapFilter(resources, (r) => getCanonicalUrl(r))
-      : resources.map(getReferenceString);
+  // `_revinclude=ResourceType:code:targetType` restricts the reverse include to
+  // base resources of `targetType`. Build the references in a single pass,
+  // filtering to the target type as we go.
+  const isCanonical =
+    getSearchParameterImplementation(resourceType, searchParam).type === SearchParameterType.CANONICAL;
+  const references: string[] = [];
+  for (const resource of resources) {
+    if (targetType && resource.resourceType !== targetType) {
+      continue;
+    }
+    if (isCanonical) {
+      const canonicalUrl = getCanonicalUrl(resource);
+      if (canonicalUrl) {
+        references.push(canonicalUrl);
+      }
+    } else {
+      const reference = getReferenceString(resource);
+      if (reference) {
+        references.push(reference);
+      }
+    }
+  }
+  if (references.length === 0) {
+    return [];
+  }
   const searchRequest = {
     resourceType: resourceType as ResourceType,
     filters: [{ code, operator: Operator.EQUALS, value: references.join(',') }],
--- a/packages/server/src/fhir/search.test.ts
+++ b/packages/server/src/fhir/search.test.ts
@@ -2771,6 +2771,86 @@ describe.each<Project['features']>([undefined, ['range-search']])('project-scope
       ).toStrictEqual(expected);
     }));
 
+  test('_include with target type', () =>
+    withTestContext(async () => {
+      // gh-9765: AuditEvent.entity.what can reference many types, so
+      // _include=AuditEvent:entity:Task must include only the Task, not every
+      // referenced resource type.
+      const task = await repo.createResource<Task>({
+        resourceType: 'Task',
+        status: 'completed',
+        intent: 'order',
+      });
+      const referencedAuditEvent = await repo.createResource<AuditEvent>({
+        resourceType: 'AuditEvent',
+        recorded: '2026-01-01T00:00:00.000Z',
+        type: { system: 'http://terminology.hl7.org/CodeSystem/audit-event-type', code: 'rest' },
+        agent: [{ requestor: true }],
+        source: { observer: { display: 'test' } },
+      });
+      const auditEvent = await repo.createResource<AuditEvent>({
+        resourceType: 'AuditEvent',
+        recorded: '2026-01-01T00:00:00.000Z',
+        type: { system: 'http://terminology.hl7.org/CodeSystem/audit-event-type', code: 'rest' },
+        agent: [{ requestor: true }],
+        source: { observer: { display: 'test' } },
+        entity: [
+          { what: { reference: getReferenceString(task) } },
+          { what: { reference: getReferenceString(referencedAuditEvent) } },
+        ],
+      });
+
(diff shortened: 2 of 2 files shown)


packages/server/src/fhir/search.ts
```diff
@@ -605,16 +614,23 @@ async function getSearchRevIncludeEntries(
...
+  if (targetResources.length === 0) {
+    return [];
+  }
+
   const references =
     getSearchParameterImplementation(resourceType, searchParam).type === SearchParameterType.CANONICAL
-      ? flatMapFilter(resources, (r) => getCanonicalUrl(r))
-      : resources.map(getReferenceString);
+      ? flatMapFilter(targetResources, (r) => getCanonicalUrl(r))
+      : targetResources.map(getReferenceString);
```
mattwiller: Could this construct `references` in a single pass over `resources`?  Something like this should be relatively clear and efficient:

```ts
const references: string[] = [];
for (const target of resources) {
  if (targetType && target.resourceType !== targetType) {
    continue;
  }

  if (getSearchParameterImplementation(resourceType, searchParam).type === SearchParameterType.CANONICAL) {
    const canonicalUrl = getCanonicalUrl(target);
    if (canonicalUrl) {
      references.push(canonicalUrl);
    }
  } else {
    references.push(getReferenceString(target));
  }
}
```

oldrobotdev: Good call, done. Switched it to a single pass that filters to the target type and builds the references inline, and pulled the canonical check out of the loop since it doesn't change per resource. Also took the Copilot suggestion above and gated the canonical include branch on the target count so it doesn't kick off an empty search. Signed off the commit for DCO as well.

packages/server/src/fhir/search.ts
```diff
@@ -605,16 +614,34 @@ async function getSearchRevIncludeEntries(
...
+    if (targetType && resource.resourceType !== targetType) {
+      continue;
+    }
+    if (isCanonical) {
+      const canonicalUrl = getCanonicalUrl(resource);
+      if (canonicalUrl) {
+        references.push(canonicalUrl);
+      }
+    } else {
+      references.push(getReferenceString(resource));
```
mattwiller: This line looks like it's causing a build error — mind taking a look?  Please ensure that the code builds locally and passes the tests for this file before re-submitting

mattwiller (changes requested):

packages/server/src/fhir/search.ts
oldrobotdev: Sorry about that, fixed. getReferenceString returns string-or-undefined for a plain resource, so pushing the result straight into the references array tripped the build. I guarded it the same way as the canonical branch above. Ran the server build and the search tests locally this time, both green.

mattwiller (approved):

codyebberson:
Thanks @oldrobotdev !

PR #9790, opened 2026-07-08
React: Extract useSchedulingResources hook

noahsilas:
In preparing for building UI that allows viewers to interact with multiple schedules at once, here we extract a hook that fetches the appropriate resources (related `Slot` and `Appointment` resources in the search range), and provides wrappers around the Scheduling APIs that can keep this state in sync with mutations.

This is intended to be pulled into `@medplum/react-hooks` once it feels it has stabilized here in the Provider app.

### Alternative Strategies Considered
- Make `MedplumClient` a more full featured event source - if we could receive events from the MedplumClient when mutations/operations happen to key resource types, we could avoid needing to expose our own API wrapper. This feels like a much larger lift and something to explore more broadly later.
- Using webhook subscriptions to receive changes - this might be something we layer in to this hook later (it would be nice if changes to your calendar from other sources would update in real time). This requires the "websocket-subscriptions" feature enabled on the project, so doesn't feel suitable as the basic building block (and could still result in some awkward windows between when a mutation completes and when the subscription event arrives).

### DIFF
--- a/examples/medplum-provider/src/components/OutcomeAlert.tsx
+++ b/examples/medplum-provider/src/components/OutcomeAlert.tsx
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { AlertProps } from '@mantine/core';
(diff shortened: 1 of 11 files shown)


noahsilas:
Closing in favor of #9821 and #9931.

PR #9794, opened 2026-07-08
Fix low-contrast text in the Choose a Project combobox

### LINKED ISSUE #9791: Low contrast on projects combobox
Image

When using arrow keys to focus on the project options in the Choose a Project screen, the contrast on the user text is very low.

Filupnot:
Fixes #9791 

## Screenshots (TL;DR)

|  | Before | After |
|--------|--------|--------|
| Not Hovered | image | image |
| Hovered | image | image | 

## Root cause (selected option)

When an option gets keyboard-highlighted, Mantine's `Combobox.module.css` applies a blue background and white text color to the option. Text normally inherits that white color, but since we hard-set `c="dimmed"` on the subtitle `<Text>`, its color wouldn't change when the option became highlighted, leaving dimmed gray text on a blue background.

## Fix implemented

Keep the subtitle text `--mantine-color-dimmed` by default, but switch it to `--mantine-color-white` when the option is selected.

## Root cause (hover on a selected option)

Mantine deliberately makes its selected-option styling low-specificity, so that consumer styles can override it if needed. Our custom hover background was more specific, so hovering the already-selected option replaced Mantine's blue background with our gray hover background. But the white text color stuck around, leaving faint white text on a light gray background.

## Fix implemented

Exclude the selected option from our custom hover styling, matching how Mantine's own hover rule already excludes it, so hovering a selected option keeps its normal blue background and white text unchanged.


kevinwadeshaw:
@Filupnot thanks for catching this! 

One thought on the fix here: for most other keyboard-focus-highlights, we are using a blue border instead of a blue fill. With an outline, we can maintain the same background color change we have on hover and don't have to change the text colors (any more than they already change on hover). Could this same treatment work for this Project selection screen?

Here's how it looks for the same Project element in the navbar dropdown:

navbar-dropdown

Filupnot:
> @Filupnot thanks for catching this!
> 
> One thought on the fix here: for most other keyboard-focus-highlights, we are using a blue border instead of a blue fill. With an outline, we can maintain the same background color change we have on hover and don't have to change the text colors (any more than they already change on hover). Could this same treatment work for this Project selection screen?
> 
> Here's how it looks for the same Project element in the navbar dropdown:
> 
> navbar-dropdown

I like that - updated!

One thing to call out though, the focus ring that you're seeing is the browser's focus ring on a Menu component, with no special handling. The project selector on the other hand uses a _Combobox_ component, which actually keeps focus on the search bar, and has its own handling for showing the selected projectbox. That being said, from a visual perspective, the CSS changes to skew from Mantine's stock styling is pretty minimal. 

Let me know what you think of this. 

https://github.com/user-attachments/assets/5fa5b133-1642-4db2-ab0a-79162702232b

kevinwadeshaw:
@Filupnot Hey, thanks for this update and sorry for the slow reply—looks perfect to me!

mattlong (approved):

PR #9796, opened 2026-07-08
[React] Skip rendering "OK" in OperationOutcomeAlert

noahsilas:
The most common OperationOutcome results are informational, telling that an operation completed successfully. In general, we don't want to render "Alert" components for such a result. In cases where showing those results is desirable, you may add the new `displayOkOutcomes` prop.

One common example: The `useSearch*()` family of hooks in `@medplum/react-hooks` return an OperationOutcome that usually is a "success", but might have an issue indicating an access policy violation or an invalid search parameter. This provides an ergonomic way to consume that result in your application with less ceremony.

**Before**:
```
import { isOk } from '@medplum/core';
import { useSearch } from '@medplum/react-hooks';
import { OperationOutcomeAlert } from '@medplum/react';

function MyComponent() {
  const [patients, loading, outcome] = useSearch('Patient');
  if (loading) { return <Loader />; }
  if (outcome && !isOk(outcome)) {
    return <OperationOutcomeAlert outcome={outcome} />
  }
  return (
    <>
      {patients?.map((patient) => <ListItem patient={patient} />)}
    </>
  );
}
```

**After**:
```
import { useSearch } from '@medplum/react-hooks';
import { OperationOutcomeAlert } from '@medplum/react';

function MyComponent() {
  const [patients, loading, outcome] = useSearch('Patient');
  if (loading) { return <Loader />; }
  return (
    <>
      <OperationOutcomeAlert outcome={outcome} />
      {patients?.map((patient) => <ListItem patient={patient} />)}
    </>
  );
}
```

### DIFF
--- a/packages/react/src/OperationOutcomeAlert/OperationOutcomeAlert.tsx
+++ b/packages/react/src/OperationOutcomeAlert/OperationOutcomeAlert.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { AlertProps } from '@mantine/core';
 import { Alert } from '@mantine/core';
-import { operationOutcomeIssueToString } from '@medplum/core';
+import { isOk, operationOutcomeIssueToString } from '@medplum/core';
 import type { OperationOutcome, OperationOutcomeIssue } from '@medplum/fhirtypes';
 import { IconAlertCircle } from '@tabler/icons-react';
 import type { JSX } from 'react';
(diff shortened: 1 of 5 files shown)


packages/react/src/OperationOutcomeAlert/OperationOutcomeAlert.tsx
```diff
@@ -2,22 +2,29 @@
...
+  const { outcome, issues: issuesProp, displayOkOutcomes, ...spacingProps } = props;
+
   const issues = outcome?.issue || issuesProp;
   if (!issues || issues.length === 0) {
     return null;
   }
+
+  if (outcome && isOk(outcome) && !displayOkOutcomes) {
+    return null;
+  }
```
noahsilas, line 26: This is technically a breaking change, if anyone was relying on this component to display "OK" outcomes they would need to now also add the `displayOkOutcomes` flag, but the nature of an "OK" outcome feels like it is naturally in tension with the usage of an `<Alert>` component. What do folks think? Can we ship this in a minor release?

techdavidy: if this case the developer using this can avoid rendering it if outcome is ok ?

noahsilas:
@claude please review

packages/react/src/OperationOutcomeAlert/OperationOutcomeAlert.tsx
noahsilas: Yeah, I think that is the pretty typical pattern: folks have been guarding rendering this with a check like `if (!isOK(outcome)) { return <OperationOutcomeAlert outcome={outcome} />; }`.

In that situation, everything still works exactly as it did before (well, the outcome will be tested against `isOk()` an extra time but that should be safe).

But my goal here is to make it easier to use without that extra step, so that the common case behaves more similarly to how the caller would expect (not throwing up red "All OK" alerts).

mattlong (approved):

Issue #6686, opened 2025-05-28
Dashlane Extension Phantom Click Issue


codyebberson:
## Problem Summary

A customer reported that users are automatically redirected from the `/signin` page to the `/register` page without clicking anything. Investigation revealed:

- Issue only occurs for users with the Dashlane Chrome extension installed
- The Dashlane extension's auto-login feature programmatically clicks the "Register" link while scanning for forms to autofill
- This creates phantom clicks with suspicious properties:
  - `isTrusted: false`
  - All coordinates at `(0,0)` 
  - `pointerId: -1`
- Customer can work around by disabling Dashlane's "auto-login" feature, but this affects their workflow

## Root Cause

Dashlane's form detection logic programmatically interacts with page elements while scanning for login/register forms, inadvertently triggering click handlers on navigation elements.

### Option 1: Filter Untrusted Events ⭐ **RECOMMENDED**
```javascript
const handleRegisterClick = (event) => {
  if (!event.isTrusted) {
    console.log('Blocked untrusted click event');
    return;
  }
  // Normal register logic here
};
```

### Option 2: Coordinate-Based Detection
```javascript
const handleRegisterClick = (event) => {
  if (event.clientX === 0 && event.clientY === 0 && event.pageX === 0 && event.pageY === 0) {
    console.log('Blocked phantom click (zero coordinates)');
    return;
  }
  // Normal register logic here
};
```

### Option 3: HTML Attributes to Discourage Automation
```html
<a href="/register" 
   role="button" 
   data-no-autofill="true"
   data-form-type="navigation">
   Register
</a>
```

### Option 4: Convert to Button Element
```html
<button onClick={() => navigate('/register')} type="button">
  Register
</button>
```

## Recommendation: Option 1 (Untrusted Event Filtering)

**Why this is the best approach:**

- **Standard Web API**: `event.isTrusted` is specifically designed to distinguish genuine user interactions from programmatic events
- **Reliable**: More robust than coordinate checking, which could have false positives
- **Future-proof**: Will work against other extensions or automation tools, not just Dashlane
- **Well-supported**: Excellent browser compatibility
- **Semantically correct**: Uses the web platform as intended

## Implementation Strategy

**Don't apply everywhere** - Only add `!event.isTrusted` checks to sensitive click handlers:

- ✅ **Navigation triggers** (like Register link)
- ✅ **Destructive actions** (delete, logout)
- ✅ **Financial/sensitive operations**
[...]

PR #9798, opened 2026-07-08
Add login link to registration screen

### LINKED ISSUE #9788: Can't navigate back to Login page from Register page
When you're in the registration page, there is no button that navigates you back to the login page. You need to navigate backwards, or remove `/register` from the URL path. 

https://github.com/user-attachments/assets/2cae3a18-d2b6-4b0d-9d2e-cf4cfb437ddc

Filupnot:
Fixes #9788 

## Problem 
There was not a good way to get back to the login page from the registration page. The user would have to navigate backwards in the browser, or update the URL path to go back. 

## Changes for users
Now, there will be a link at the bottom of the registration page, that will let the user go back to the login screen. This matches the same pattern that exists in the login page _("Don't have an account? Register")_ and in the reset password form _("Back to Sign In")_. 

## Code changes
- Add a new optional prop for `onSignIn?: () => void` to the new user form and the registration form.
- Wire `onSignIn={() => navigate('/signin')}` to the register page
- Update relevant tests appropriately
- Tell password managers to ignore the link (following pattern done in #6706)

### DIFF
--- a/packages/react/src/auth/NewUserForm.tsx
+++ b/packages/react/src/auth/NewUserForm.tsx
@@ -143,7 +143,7 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
             size="sm"
             mt="lg"
             c="dimmed"
-            style={{ textAlign: 'center' }}
+            ta="center"
             data-dashlane-ignore="true"
             data-lp-ignore="true"
             data-no-autofill="true"
--- a/packages/react/src/auth/NewUserForm.tsx
+++ b/packages/react/src/auth/NewUserForm.tsx
@@ -145,7 +145,7 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
             {' and '}
             <Anchor href="https://www.medplum.com/terms">Terms&nbsp;of&nbsp;Service</Anchor>.
           </Text>
-          <Text c="dimmed" size="xs" pt="xs" ta="center">
+          <Text c="dimmed" size="xs" pb="xs" ta="center">
             This site is protected by reCAPTCHA and the Google{' '}
             <Anchor href="https://policies.google.com/privacy">Privacy&nbsp;Policy</Anchor>
             {' and '}
@@ -155,7 +155,7 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
       </Form>
       {props.onSignIn && (
         <>
-          <Divider mt="lg" mb="xs" />
+          <Divider mt="lg" pb="xs" />
           <Text size="sm" mt="lg" c="dimmed" ta="center">
             Already have an account? <Anchor onClick={props.onSignIn}>Sign In</Anchor>
           </Text>
--- a/packages/react/src/auth/NewUserForm.tsx
+++ b/packages/react/src/auth/NewUserForm.tsx
@@ -161,7 +161,7 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
           {' and '}
           <Anchor href="https://www.medplum.com/terms">Terms&nbsp;of&nbsp;Service</Anchor>.
         </Text>
-        <Text c="dimmed" size="xs" pb="xs" ta="center">
+        <Text c="dimmed" size="xs" ta="center">
           This site is protected by reCAPTCHA and the Google{' '}
           <Anchor href="https://policies.google.com/privacy">Privacy&nbsp;Policy</Anchor>
           {' and '}
--- a/packages/react/src/auth/NewUserForm.tsx
+++ b/packages/react/src/auth/NewUserForm.tsx
@@ -21,6 +21,7 @@ export interface NewUserFormProps {
   readonly googleClientId?: string;
   readonly recaptchaSiteKey?: string;
   readonly children?: ReactNode;
+  readonly onSignIn?: () => void;
   readonly handleAuthResponse: (response: LoginAuthenticationResponse) => void;
 }
 
@@ -137,6 +138,20 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
           pb="xs"
         />
         <SubmitButton fullWidth>Register Account</SubmitButton>
+        {props.onSignIn && (
+          <Text
+            size="sm"
+            mt="lg"
+            c="dimmed"
+            style={{ textAlign: 'center' }}
+            data-dashlane-ignore="true"
+            data-lp-ignore="true"
+            data-no-autofill="true"
+            data-form-type="navigation"
+          >
+            Already have an account? <Anchor onClick={props.onSignIn}>Sign In</Anchor>
+          </Text>
+        )}
         <Text c="dimmed" size="xs" pt="lg" ta="center">
           By clicking "Register Account" you agree to the Medplum{' '}
           <Anchor href="https://www.medplum.com/privacy">Privacy&nbsp;Policy</Anchor>
--- a/packages/react/src/auth/NewUserForm.tsx
+++ b/packages/react/src/auth/NewUserForm.tsx
@@ -145,7 +145,7 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
             {' and '}
             <Anchor href="https://www.medplum.com/terms">Terms&nbsp;of&nbsp;Service</Anchor>.
           </Text>
-          <Text c="dimmed" size="xs" ta="center">
+          <Text c="dimmed" size="xs" pt="xs" ta="center">
             This site is protected by reCAPTCHA and the Google{' '}
             <Anchor href="https://policies.google.com/privacy">Privacy&nbsp;Policy</Anchor>
             {' and '}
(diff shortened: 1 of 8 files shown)


packages/react/src/auth/NewUserForm.tsx
```diff
@@ -137,6 +138,20 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
...
+            size="sm"
+            mt="lg"
+            c="dimmed"
+            style={{ textAlign: 'center' }}
+            data-dashlane-ignore="true"
+            data-lp-ignore="true"
+            data-no-autofill="true"
+            data-form-type="navigation"
+          >
+            Already have an account? <Anchor onClick={props.onSignIn}>Sign In</Anchor>
```
noahsilas: The `onSignIn` function works, but I think it is a little less clear what the intention of that prop name is. You can imagine a similarly named prop being invoked when the registration is complete and the viewer becomes "signed in" as a new User.

What do you think about instead passing a prop like `signInUrl`? Then this could use `<Anchor href={props.signInUrl}>`, or `<MedplumLink to={props.signInUrl}>` to defer to the navigation function (which is already configured in apps like `app` and `provider` to use react router).

That would also make the callers a little tidier: `<RegisterForm signInUrl="/signin">` vs. `<RegisterForm onSignIn={() => navigate("/signin").catch(console.error)}>`.

packages/react/src/auth/NewUserForm.tsx
```diff
@@ -137,6 +138,20 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
           pb="xs"
         />
         <SubmitButton fullWidth>Register Account</SubmitButton>
+        {props.onSignIn && (
+          <Text
+            size="sm"
+            mt="lg"
+            c="dimmed"
+            style={{ textAlign: 'center' }}
```
noahsilas: Mantine has a builtin for this one:

```suggestion
            ta="center"
```

packages/react/src/auth/NewUserForm.tsx
```diff
@@ -137,6 +138,20 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
...
+            data-dashlane-ignore="true"
+            data-lp-ignore="true"
+            data-no-autofill="true"
+            data-form-type="navigation"
```
noahsilas, line 150: These attributes are a little surprising to me - this Text component doesn't seem like it has anything that would obviously be affected by these. I'm guessing that some of this is about making this work nicely with password managers like lastpass/dashlane? Maybe we should move all this footer content out of the `<Form>` component to make those browser extensions less likely to interact with it?

codyebberson:
@claude please review

packages/react/src/auth/NewUserForm.tsx
noahsilas: On inspection, it looks like you probably went with this to match the `SignInForm`, which accepts an `onRegister` prop that looks like this. Probably a good idea to stay consistent between these components for now, and maybe we can evaluate a breaking change to both components to simplify this later with the "*Url" style props.

EDIT: Claude apparently said the same thing as I was typing this 😂

packages/react/src/auth/RegisterForm.test.tsx
```diff
@@ -212,6 +212,32 @@ describe('RegisterForm', () => {
...
+    await setup({
+      type: 'project',
+      recaptchaSiteKey,
+      onSuccess: vi.fn(),
+      onSignIn,
+    });
+
+    await act(async () => {
+      fireEvent.click(screen.getByText('Sign In'));
+    });
```
noahsilas: This `fireEvent` helper from `@testing-library/react` already handles wrapping the behavior in `act()`, so you can omit it here:

```suggestion
    fireEvent.click(screen.getByText('Sign In'));
```

noahsilas (approved):

packages/react/src/auth/NewUserForm.tsx
Filupnot: Yeah, that was the intention (I left a short note about that in the original PR description). Looks like moving it outside of the form should prevent that issue from coming up though, so I went and implemented your recommended solution. 

Since the login link is at the bottom now, I figured adding a divider would differentiate it better from the disclaimer stuff. Let me know if you think the divider should be removed :) 

image

techdavidy:
@kevinwadeshaw mind taking a look on your side ?

packages/react/src/auth/NewUserForm.tsx
```diff
@@ -38,118 +39,128 @@ export function NewUserForm(props: NewUserFormProps): JSX.Element {
...
-                  );
-                } catch (err) {
-                  setOutcome(normalizeOperationOutcome(err));
-                }
-              }}
-            />
-          </Box>
-          <Divider label="or" labelPosition="center" my="lg" />
+          <Divider mt="lg" />
+          <Text size="sm" mt="lg" c="dimmed" ta="center">
```
techdavidy: The NewUserForm.tsx diff looks much bigger than it is — the only real change is the new onSignIn prop and the  "Sign In" text block, but placing that block outside the <Form> required wrapping the whole component in a fragment, which re-indented all >100 lines.

techdavidy (changes requested):
Mind taking a look at the NewUserForm.tsx diff looking much bigger than it is.
It also makes it inconsistent with the other auth forms.

packages/react/src/auth/NewUserForm.tsx
noahsilas: @techdavidy In the github UI you can see the much tidier diff by enabling "Hide whitespace changes" from the gear in the top right of the diff view, or adding `?w=1` to your URL. In the `git` CLI you can use `git diff --ignore-space-change` (aka `git diff -b`) to see a similar diff without the indent causing a big delta. 😁 

One final thought for the future (not necessary to adjust now): in this diff you have added explicit margin-top to two adjacent components, which will evenly space them all. Another option would have been to wrap the components in a [`<Stack gap="lg">`](https://mantine.dev/core/stack/) to handle putting even space between all the elements of the stack. This is nice because it's more intention revealing, and makes it harder for the vertical spacing to drift if someone tweaks only one of these `mt` props.

techdavidy: Yeah, I can see a tidy diff but still a big diff. 
There is a similar PR that achieves the same with less lines of code.

techdavidy:
Moved it out of the merge queue to get feedback from Kevin.
This component is seen by a lot of users so asked for his input on this one.

packages/react/src/auth/NewUserForm.tsx
techdavidy: https://github.com/medplum/medplum/pull/9801/changes#diff-ee23a413a29cc0c014d1a472ef94fa21afa3336a3e08362a0ec62416f899dcd3R141-R145

Filupnot: From what I can tell, this PR is essentially what I did in my first iteration, except I included the `data-*` attributes on the `<Text>` element to prevent weirdness from coming up with password managers (see #6686). 

This PR is more lines of code because: 
- We indented the form to put the login outside the `<Form>` (https://github.com/medplum/medplum/pull/9798#discussion_r3547826534)
- We're also updating the examples

packages/react/src/auth/NewUserForm.tsx
Filupnot: After thinking more about this... I think we should not change this. It's better to be consistent with how we are in other places, both in the code and from the user's perspective. Let me know if you disagree with this, happy to talk further.

techdavidy:
@claude review

packages/react/src/auth/NewUserForm.tsx
Filupnot: Actually, thinking more about this, I do agree that it would be better to keep this consistent with `AuthenticationForm` by keeping this link inside the form. I do think we should keep the `data` attributes though.

Filupnot: @noahsilas - Good point on Stack gap for the future. I kept the explicit `mt` here since the spacing in this stack matches how AuthenticationForm spaces its register link.

techdavidy (approved):

techdavidy:
Thanks @Filupnot for taking care of this. 
Thanks @kevinwadeshaw for the input.

techdavidy:
Screenshot 2026-07-09 at 10 04 15 AM (1)

@Filupnot  let's keep the input from Kevin but inside the form please.

Filupnot:
> Screenshot 2026-07-09 at 10 04 15 AM (1)
> 
> @Filupnot let's keep the input from Kevin but inside the form please.

You might've been looking at an earlier commit- the most recent commit should've moved it back inside the form

image

techdavidy:
@kevinwadeshaw thoughts ?

PR #9799, opened 2026-07-08
feat(react): DiagnosticReportDisplay add more examples in storybook 

techdavidy:
Screenshot 2026-07-08 at 4 28 01 PM
Screenshot 2026-07-08 at 4 27 53 PM


packages/react/src/DiagnosticReportDisplay/DiagnosticReportDisplay.stories.tsx
```diff
@@ -101,6 +102,106 @@ export const HideNotes = (): JSX.Element => {
...
+ * and performers, and one remaining "final" result.
+ * @returns The lab panel with corrections story.
+ */
+export const LabPanelWithCorrections = (): JSX.Element => {
+  const medplum = useMedplum();
+  const [loaded, setLoaded] = useState(false);
+
+  useEffect(() => {
+    (async (): Promise<boolean> => {
+      await medplum.createResource(LabPanelSpecimen);
```
noahsilas: This resource sample has an explicit `id` set, and so this component ends up rejecting if it renders twice (including cases such as someone navigating around different stories and coming back to this one a second time).

It seems like you should be able to use ``medplum.createResourceIfNoneExist(LabPanelSpecimen, `_id=${LabPanelSpecimen.id}`)``, but in my testing that doesn't seem to work. I have a suspicion that what I'm seeing is related to React Strict Mode running the effect twice and hitting a race condition in the mock implementation, but I haven't dug too deep yet. Maybe that won't be a problem in the deployed storybook?

As a side note, maybe we can improve the PR preview generation to see when `packaegs/react` was touched and also do a storybook preview release 🤔

techdavidy: good point seems the deployed storybook has the issue.
creating a new issue to check what is happening.

noahsilas (approved):

PR #9801, opened 2026-07-09
Add sign in link to the register page

AbhinavMir:
Candidate for #9788.

demo


mattwiller:
Already fixed by #9798

PR #9802, opened 2026-07-09
feat(warehouse): async high water mark calculations

The-Alchemist:
- Instead of creating and tearing down a DuckDB instance for **every** history table, open **one** instance for the whole job and destroy it at the end (so Iceberg catalog metadata does not carry stale watermarks across ticks).
- Resolve every table’s high watermark **before** attaching Postgres:
  - fan out Iceberg watermark reads across up to 8 DuckDB connections for _massive_ performance boost since some large tables like `patient_history` can take as much as 30 seconds for `MAX(last_updated)`
  - close watermark connections, then open one write connection, attach Postgres once, and run per-table inserts
- Deferring Postgres attach until after watermark work avoids idle-connection resets during Iceberg reads.
- On Iceberg commit conflicts (`CommitFailedException` / `Conflict_409`, typically from AWS compaction), skip that table and continue; the next tick will pick up the data rather than blocking the job on an unbounded compaction wait.

### DIFF
--- a/packages/server/src/cloud/aws/data-warehouse-destination.ts
+++ b/packages/server/src/cloud/aws/data-warehouse-destination.ts
@@ -42,6 +42,35 @@ export class S3TablesWarehouseDestination implements DataWarehouseDestination {
     });
   }
 
+  /*
+   * DuckDB session settings for managed Iceberg sync connections.
+   *
+   * These are connection-scoped (`SET` without `GLOBAL`) and must run on every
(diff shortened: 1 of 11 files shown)


The-Alchemist:
@claude review

The-Alchemist:
@claude review

The-Alchemist:
@claude review

codyebberson (approved):

PR #9803, opened 2026-07-09
Make "View Password" toggle keyboard accessible

### LINKED ISSUE #9786: View Password icon (eyeball) isn't keyboard navigable
In the registration workflow, you're unable to use a keyboard to navigate to the eyeball icon that allows you to view your password. It's not part of the tab order and can't be accessed via arrow keys (or other keyboard methods). 

Claude thinks that this can be fixed with a Mantine component toggle. 

https://github.com/user-attachments/assets/54243dcf-86da-4099-b6f2-3b01339ead0e

Filupnot:
Fixes #9786

## Previously

The stock Mantine Password component by default hides the View Password button from the tab order, making it not keyboard accessible and violating [WCAG 2.1.1.](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html).\

## Changes for users 

Users can now tab to this button and interact with it via spacebar. 

https://github.com/user-attachments/assets/3303252e-0267-4339-b0c0-45f70edc5749

## Changes to the code 

Mantine's documented pattern for making this keyboard accessible is to add the following: 
```
<MantinePasswordInput
      {...rest}
      visibilityToggleButtonProps={{
        'aria-label': 'Toggle password visibility',
        tabIndex: 0,
        ...visibilityToggleButtonProps,
      }}
    />
```

Adding this to each of our `<PasswordInput>` fields would add a good amount of bloat, so instead, we'll create a new Medplum owned `<PasswordInput>` component that extends the Mantine one with these accessibility props.


Filupnot:
@claude review this

The-Alchemist (approved):

techdavidy:
👍🏻 Given this is a new component I would say let's add claude's suggestions.

techdavidy:
@claude would this be a breaking change in any package ?

PR #9804, opened 2026-07-09
test(agent,hl7): fix flaky `getFreePort` port collision

ThatOneBro:
`getFreePort` closed its probe server before returning, so two sequential calls could be handed the same ephemeral port by the OS. Tests that allocate multiple ports (e.g. the 'test' and 'prod' channels in "Setting a channel.endpoint.status to 'off'") then declared the same port and app.start() threw a config conflict.

Remember every issued port and hold a probe server open when it lands on an already-issued port, so the OS can't reissue it in the same round

### DIFF
--- a/packages/agent/src/test-utils.ts
+++ b/packages/agent/src/test-utils.ts
@@ -110,24 +110,47 @@ export function createTestHl7ClientPool(
   return { pool, messageTracker, heartbeatEmitter };
 }
 
+// Every port ever handed out by `getFreePort`, so we never issue the same one twice
+// within a single test process — see the comment on `getFreePort` below.
+const issuedPorts = new Set<number>();
+
 // Used only for tests that need a free port number with *nothing* listening on it.
 // For tests that start an Hl7Server, prefer `server.start(0)` which returns the OS-assigned
 // port and never has a release-then-rebind window.
+//
+// Because we close the probing server before returning, the port is freed immediately and the
+// OS may hand the *same* ephemeral port to a subsequent call. Callers that allocate two ports
+// (e.g. one per channel) would then collide. To avoid this, remember every port we issue and
+// keep any probing server that lands on an already-issued port open until we find a fresh one,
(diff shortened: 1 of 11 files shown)


packages/agent/src/test-utils.ts
```diff
@@ -110,24 +110,47 @@ export function createTestHl7ClientPool(
...
-      });
+  const heldServers: ReturnType<typeof createServer>[] = [];
+  const closeServer = async (server: ReturnType<typeof createServer>): Promise<void> =>
+    new Promise((resolve, reject) => {
+      server.close((err) => (err ? reject(err) : resolve()));
     });
-    server.on('error', reject);
-  });
+  try {
+    for (let attempt = 0; attempt < 20; attempt++) {
```
mattwiller: Should the number of attempts be configurable somehow?

packages/agent/src/test-utils.ts
```diff
@@ -110,24 +110,47 @@ export function createTestHl7ClientPool(
...
+      const server = createServer();
+      const port = await new Promise<number>((resolve, reject) => {
+        server.on('error', reject);
+        server.listen(0, () => {
+          resolve((server.address() as { port: number }).port);
```
mattwiller: I'd recommend casting to the actual `AddressInfo` type here, which includes `port`

packages/hl7/src/test-utils.ts
```diff
@@ -80,22 +80,45 @@ export class MockServer extends EventEmitter {
...
+        continue;
+      }
+      issuedPorts.add(port);
+      await closeServer(server);
+      return port;
+    }
+    throw new Error('Unable to find a free port after 20 attempts');
+  } finally {
+    await Promise.all(heldServers.map((server) => closeServer(server).catch(() => undefined)));
+  }
```
mattwiller: This logic seems mostly duplicated; can it be pulled into a helper function?

mattwiller (approved):

PR #9806, opened 2026-07-09
Adding yang su and meena ramachandran to plumcon agenda

timray721:


mattwiller (approved):

maddyli (approved):
Would you mind moving them above Community Demos, below Stuart? Also, could you convert their images to webp? Otherwise looks great!

PR #9809, opened 2026-07-09
Make MemoryRepository conditionalCreate more transactional 

noahsilas:
We don't have a real DB here, but we can lean on the single-threaded nature of the JS runtime to get transaction-like behavior. We start by creating private "*Sync" versions of some methods that do not introduce async boundaries. We then add a `conditionalCreate` that wraps those synchronous methods to do the find + create without the opportunity for an async call to interleave between them.

I spotted this while investigating why some Storybook stories would error during re-rendering; in that context we are occasionally have code like `medplum.createResource({ resourceType: 'Patient', id: 'homer-simpson' })`, which throws when running a second time. There are a few avenues of remediation, but a natural one was to move towards ``medplum.createResourceIfNoneExist(homer, `_id=${homer.id}`)``. There are still some tricky aspects to work out in the stories (this doesn't work if the mock client doesn't have the search params for the requested type loaded), so in this PR I'm just introducing this small prerequisite step.

### DIFF
--- a/packages/fhir-router/src/repo.test.ts
+++ b/packages/fhir-router/src/repo.test.ts
@@ -291,5 +291,33 @@ describe('MemoryRepository', () => {
       expect(resultAsc[getReferenceString(patients[0])].map((o) => o.valueString)).toStrictEqual(['0', '1', '2']);
       expect(resultAsc[getReferenceString(patients[1])].map((o) => o.valueString)).toStrictEqual(['0', '1']);
     });
+
+    test('concurrent conditionalCreate', async () => {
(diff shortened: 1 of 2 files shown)


noahsilas:
Side note: even without navigating, when viewing the storybook the effects defined in components can be run twice. I thought this was due to [React Strict Mode](https://react.dev/reference/react/StrictMode), which runs effects twice in when development, but the same thing seems to happen even in prod builds? This might be due to https://github.com/storybookjs/storybook/issues/33584.

ThatOneBro (approved):

PR #9811, opened 2026-07-09
Reset batch processor state on transaction retry

mattwiller:
A transaction Bundle that requires `SERIALIZABLE` isolation can fail with a serialization conflict error (`40001`) at `COMMIT` time: `RepositoryConnection.withTransaction` treats this as retryable and re-runs the transaction callback.
 
However, `BatchProcessor` keeps its progress (`this.position`) and per-entry results (`this.resultEntries`) as instance state that is **not** reset between attempts. When the first attempt processed every entry and then failed at `COMMIT`, the retry re-enters `processEntriesAndBuild()` with `position` already at the end, so `hasMoreEntries()` is false.  The retry therefore executes zero statements, commits an empty transaction successfully, and returns the stale success responses from the (rolled-back) first attempt.
 
The client receives a `200 OK` response with a `transaction-response` bundle reporting results as successful, but none of those writes were committed to the database.
 
This issue is more likely on a small Postgres database, where `SERIALIZABLE` isolation uses relation/page-level predicate locks and serialization conflicts are common

### DIFF
--- a/packages/fhir-router/src/batch.ts
+++ b/packages/fhir-router/src/batch.ts
@@ -177,9 +177,26 @@ export class BatchProcessor {
       return this.processEntriesAndBuild();
     }
 
-    return this.repo.withTransaction((txRepo) => this.withRepo(txRepo, () => this.processEntriesAndBuild()), {
-      serializable: bundleInfo.requiresStrongTransaction,
-    });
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude please review

codyebberson (approved):

PR #9813, opened 2026-07-10
Co-locate all mutable state

mattlong:
Followup to #9811

### DIFF
--- a/packages/fhir-router/src/batch.ts
+++ b/packages/fhir-router/src/batch.ts
@@ -86,7 +86,7 @@ export async function processBatch(
   return processor.run();
 }
 
-type ProcessingState = {
+interface ProcessingState {
   /** Index into `bundleInfo.ordering` of the next entry to process. */
   position: number;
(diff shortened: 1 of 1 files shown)


packages/fhir-router/src/batch.ts
```diff
@@ -86,6 +86,17 @@ export async function processBatch(
   return processor.run();
 }
 
+type ProcessingState = {
```
codyebberson: Minor, but the TypeScript team recommends using `interface` over `type` unless you have to: https://www.typescriptlang.org/play/?#example/types-vs-interfaces

```ts
// That said, we recommend you use interfaces over type
// aliases. Specifically, because you will get better error
// messages. If you hover over the following errors, you can
// see how TypeScript can provide terser and more focused
// messages when working with interfaces like Chicken.
```

Interfaces allegedly have better compiler performance too 🤷‍♂️

codyebberson (approved):

PR #9819, opened 2026-07-10
Display inline errors on missing AutoComplete option set

### LINKED ISSUE #9687: Medplum Provider shows many errors if UMLS is missing
Notifications on ValueSet not found:

Image

I feel like this could potentially be valid once, and then dismissed.  Right now, this comes back on every autocomplete.

Filupnot:
Fixes #9687

## Previous behavior

If an autocomplete field is bound to an unavailable option set, then we would throw error toasts to the user like this: 

https://github.com/user-attachments/assets/87cf0775-1a7d-4bfd-8176-f0610f1527df

## New Behavior
We'll get rid of toasts in favor of inline styling. This will: 
1. Make it easier to associate fields to their respective errors 
2. Reduce user spam

### ValueSet bound fields
On form load, we'll perform validation to identify all the fields that are missing their option lists _upfront_ (instead of only once a user interacts with the field). We'll also have an info button to show the user which specific ValueSet is missing for the given field. 

#### Fields accepting custom values

image
image

#### Fields _not_ accepting custom values 

We will **disable** fields in this situation, since the user would not be able to type anything into them anyways. 

### Non-ValueSet bound fields  

Integration-backed fields (e.g. Health Gorilla autocomplete bot) and FHIR search fields (patient picker) will also have their toasts replaced with inline errors. However we'll only validate them when the user interacts with them (not on form load). 

#### Why not validate on form load?
- ValueSet validation is cheaper and more reliable than the others; if we get a 400/404 error, then we know that the value set doesn't exist, so the result is cacheable per URL per session. On the other hand, integration fields are based on arbitrary search functions (e.g. Health Gorilla APIs), so validation on form load would mean running integration bots / calling external APIs on every page render, and we can't classify failures as permanent because these services can change. 
- Surfacing missing integrations upfront is tracked separately in #9824 .


packages/react/src/ValueSetAutocomplete/ValueSetAutocomplete.tsx
```diff
@@ -58,22 +59,37 @@ function createValue(input: string): ValueSetExpansionContains {
...
-          count: 10,
-          ...expandParams,
-          url: binding,
-          filter: input,
-        },
-        { signal }
-      );
+      let valueSet: ValueSet;
+      try {
+        valueSet = await medplum.valueSetExpand(
```
techdavidy: nit: maybe we can keep as const and handle the response inside this try or in a separate function.

Filupnot: Updated

Filupnot:
@kevinwadeshaw - Added you as reviewer to this PR since it changes UI around showing field errors to users. Happy to discuss. 

@claude review this please.

packages/react/src/ValueSetAutocomplete/valueSetAvailability.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+  return entry.snapshot;
+}
+
+/**
+ * Subscribes to the availability of a ValueSet, probing it on first use.
+ * @param binding - The ValueSet URL, or undefined for unbound inputs.
+ * @returns The current availability of the ValueSet.
+ */
+export function useValueSetAvailability(binding: string | undefined): ValueSetAvailability {
```
techdavidy: We should move this useValueSetAvailability to react-hooks and move the rest to core ?

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -131,16 +134,33 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
...
+    combobox,
+    loadOptions,
+    onChange,
+    toOption,
+    minInputLength,
+    setTimer,
+    setAbortController,
+    setOptions,
+    setAutoSubmit,
+    setLoadError,
```
noahsilas: Thanks for fixing these warnings while you were in here. It's sort of frustrating that the React Compiler sets different rules for what dependencies should be included in this list!

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -131,16 +134,33 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
         }
       })
       .catch((err) => {
-        if (!(newAbortController.signal.aborted || err.message.includes('aborted'))) {
-          showNotification({ color: 'red', message: normalizeErrorString(err) });
+        const message = normalizeErrorString(err);
+        if (!(newAbortController.signal.aborted || message.includes('aborted'))) {
+          setLoadError(message);
```
noahsilas: This component is fairly tricky, and used under the hood in a bunch of places. As a more minimal change, what would you feel about moving this error tracking into the `<ValueSetAutocomplete>` component? In there we have a relatively straightforward `loadValues()` function that could catch errors when making the request and pass it to this component as an `error` prop.

It seems like that could lose some of the special handling you are doing in here around retry/Enter key handling though, so maybe this is actually inherent complexity to the problem? Although in the scenario we're considering, it seems like this would never have resolved to an option list and so maybe these scenarios don't happen?

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
Filupnot: Of course! :)

noahsilas (commented):
Strategy question: I see a lot of work being done around caching in packages/core/src/valueSetAvailability.ts, but this is ultimately delegating to `MedplumClient.valueSetExpand()`, which uses `MedplumClient.get()`, which will maintain a cache internal to the client. Can we rely on that client cache instead of adding these extra wrapping layers?

packages/core/src/valueSetAvailability.ts
```diff
@@ -0,0 +1,172 @@
...
+  | { readonly status: 'available' }
+  | { readonly status: 'unavailable'; readonly message: string };
+
+const CHECKING: ValueSetAvailability = { status: 'checking' };
+const AVAILABLE: ValueSetAvailability = { status: 'available' };
+
+// How long an 'unavailable' verdict is trusted before a new consumer triggers a re-probe,
+// so fields recover after a value set is imported/linked without hammering known-missing
+// URLs on every page navigation
+const UNAVAILABLE_RETRY_INTERVAL_MS = 60_000;
```
techdavidy: it would be fine trigger a check on every refresh.

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
techdavidy: yeah let's keep this component as is :D

packages/react/src/QuestionnaireForm/QuestionnaireFormItem.tsx
```diff
@@ -420,23 +421,34 @@ function getValueSetOptions(
     .then((valueSet: ValueSet) => valueSet.expansion?.contains ?? []);
 }
 
-function useValueSetOptions(valueSetUrl: string | undefined): [ValueSetExpansionContains[], boolean] {
+function useValueSetOptions(
+  valueSetUrl: string | undefined
+): [ValueSetExpansionContains[], boolean, boolean | undefined] {
```
techdavidy: let's create a type for the return otherwise boolean boolean gets confusing.

packages/react/src/QuestionnaireForm/QuestionnaireFormItem.tsx
```diff
@@ -42,7 +43,7 @@ import { DateTimeInput } from '../DateTimeInput/DateTimeInput';
 import { QuantityInput } from '../QuantityInput/QuantityInput';
 import { ReferenceInput } from '../ReferenceInput/ReferenceInput';
 import { ResourcePropertyDisplay } from '../ResourcePropertyDisplay/ResourcePropertyDisplay';
-import { ValueSetAutocomplete } from '../ValueSetAutocomplete/ValueSetAutocomplete';
+import { UnavailableNote, ValueSetAutocomplete } from '../ValueSetAutocomplete/ValueSetAutocomplete';
```
techdavidy: UnavailableNote from ValueSetAutocomplete seems weird. Maybe it can have an isolated import

packages/react-hooks/src/useValueSetAvailability/useValueSetAvailability.ts
```diff
@@ -0,0 +1,74 @@
...
+ * A filter-free probe means a 400/404 unambiguously describes the value set itself (unlike a
+ * user-typed search, whose 400 can be filter-specific), so the verdict is safe to act on. Repeated
+ * probes of the same URL are deduplicated by the `MedplumClient` request cache, which caches
+ * rejections too, so many fields bound to the same missing value set cost one request. Recovery
+ * after a value set is imported happens on the next mount (i.e. a page refresh) — there is no live
+ * subscription.
+ * @param binding - The ValueSet URL, or undefined for unbound inputs (always available).
+ * @returns undefined while the probe is in flight, true if available, false if unavailable.
+ */
+export function useValueSetAvailability(binding: string | undefined): boolean | undefined {
```
noahsilas: The `<IntakeFormPage>` component looks like it is trying to test multiple value sets at once, would it make sense to adjust this hook to take an array and return a list of available or unavailable value sets?

Maybe that unnecessarily complicates that common case of wanting to test a single value set though. Maybe we leave this API as-is for the single case, but internally it could delegate to an array-aware version?

```ts
export function useValueSetAvailability(binding: string | undefined): boolean | undefined {
  const result = useValueSetAvailabilities([binding]); // not sure about this name yet 🤔 
  if (result.loading) return undefined;
  return result.available.includes(binding);
}
```

noahsilas: Should the parameter here be named `url` instead of `binding`? I think that the "binding" terminology comes from specific fields within other resources that are "bound" to a specific value set, but that the identifier used to refer to the value set is its URL.

packages/react/src/ValueSetAutocomplete/ValueSetAutocomplete.tsx
```diff
@@ -100,6 +145,25 @@ export function ValueSetAutocomplete(props: ValueSetAutocompleteProps): JSX.Elem
...
+  if (first && second) {
+    return (
+      <>
+        {first}
+        <br />
+        {second}
+      </>
+    );
+  }
+  return first || second;
```
noahsilas, line 169: I think our typical pattern here would be to combine these elements by wrapping them in a `<Stack>`, which provides some extra ability to control spacing between the elements.

The main case where this makes a difference is when the text on the `first` node wraps; in that world we would like to have some whitespace separating the two nodes to help clarify that they are distinct.

In the scenario where one of the input nodes is missing, I think that it should be omitted and the Stack will not insert extra spacing.

So, what do you think about an implementation like:
```ts
function combineNodes(first: ReactNode, second: ReactNode): ReactNode {
  return <Stack gap="xs">{first}{second}</Stack>;
}
```

noahsilas (commented):
I think this is getting real close now!

My main strategy feedback is that the `<IntakeFormPage>` seems to suggests we want a more powerful building block that can support that component's use case as well as the `<ValueSetAutocomplete>` style usage that works on a single valueset. I wonder if we can build the hook in such a way that it supports both, or expose a thin wrapper to make the single-value-set scenario easy while delegating to the more powerful hook.

That said, if going down that path causes the complexity here to balloon, we should probably go with this as-is. Let me know what you think!

noahsilas (approved):

PR #9820, opened 2026-07-10
fix(server): CMS-1500 PDF Box 25 and Box 24B read the wrong sources

ajwwong:
The `Claim/$export` CMS-1500 PDF prints two boxes from the wrong sources:

**Box 25 — Federal Tax ID.** The generator reads the TAX-typed identifier from the *insurer* (`coverage.payor[0]`):

```ts
const taxIdentifier = insurer.identifier?.find((id) => id.type?.coding?.find((code) => code.code === 'TAX'));
```

Per the NUCC 1500 Instruction Manual, Box 25 is the **billing provider's** federal tax ID (EIN/SSN). Printing the payer's tax id (or, more commonly, nothing — payer Organizations rarely carry a TAX identifier) makes the exported form incorrect for filing. This PR reads the same TAX-typed identifier from `claim.provider` instead.

**Box 24B — Place of Service.** The generator prints `item.locationAddress.state` — a state abbreviation — into the two-digit POS box. FHIR's `Claim.item.location[x]` supports `locationCodeableConcept`, which is where a place-of-service *code* lives. This PR adds a `getPlaceOfService` helper that prefers a `locationCodeableConcept` coding (the CMS place-of-service code system `https://www.cms.gov/Medicare/Coding/place-of-service-codes` first, then any coding) and keeps `locationAddress.state` as a legacy fallback, so no existing behavior is lost for claims that only carry an address.

Unit tests added for `getPlaceOfService` (CMS-system preference, any-coding fallback, legacy address fallback, undefined).

Follow-up candidates (not in this PR): Box 32 service facility is currently hardcoded empty; Box 27 accept-assignment and Box 33b taxonomy are not rendered.

### DIFF
--- a/packages/server/src/fhir/operations/utils/cms1500pdf.test.ts
+++ b/packages/server/src/fhir/operations/utils/cms1500pdf.test.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { HumanName } from '@medplum/fhirtypes';
-import { formatDiagnosisPointers, formatHumanName, getSimplePhone } from './cms1500pdf';
+import type { ClaimItem, HumanName } from '@medplum/fhirtypes';
+import { formatDiagnosisPointers, formatHumanName, getPlaceOfService, getSimplePhone } from './cms1500pdf';
 
 describe('CMS 1500 PDF Utils', () => {
   test('formats full name with middle name', () => {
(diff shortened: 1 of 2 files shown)


techdavidy:
@claude review

techdavidy:
@claude validate the system is correct.

techdavidy:
@claude since 24 b box is POS code, adding any other system as fallback or state would be wrong.

techdavidy:
Thanks @ajwwong for this.
Mind applying the last suggestion from claude.

ajwwong:
Applied in 97da4bad2 — `getPlaceOfService` is now strict: Box 24B only ever comes from a `locationCodeableConcept` coding in the CMS Place of Service system. Both fallbacks are gone (a coding from another system or a `locationAddress.state` now leaves the box blank rather than printing an unfilable value), tests updated to assert the strict behavior plus a position-independence case. Also reattached the `formatDiagnosisPointers` JSDoc that my earlier insertion had orphaned, per the first review note.

techdavidy:
@ajwwong our commits require to be signed. Please follow https://github.com/medplum/medplum/pull/9820/checks?check_run_id=87811539034

ajwwong:
@techdavidy Hi David!  No worries. On it. Thanks for all your help! :-)

techdavidy (approved):

PR #9821, opened 2026-07-10
MedplumClient emits resourceModified events

noahsilas:
Concept: Instead of pushing state updates through patterns like React context managers or [prop-drilling various callbacks through large sub-trees](https://github.com/medplum/medplum/pull/9790), allow components to subscribe to mutations being routed through the same MedplumClient instance they are using.

This has the benefit of allowing a component to be reactive to updates that come from anything else under the same `<MedplumProvider>` root. As one example, if we have a calendar widget showing Appointments and Slots for a practitioner, and they edit one of those resources in a modal, the calendar can get notified of the mutation without explicit wiring. This PR shows the SchedulingDetails component taking advantage of the new hook.

Strategy Note: I considered an implementation of these hooks closer to the HTTP request level, which could be used to power other interfaces like a "FHIR XRay" UI showing what the application is doing behind the scenes. I've decided that the current goal is better achieved by staying more resource-focused, but we can layer in more event types in a similar pattern later. That idea is explored more in https://github.com/medplum/medplum/pull/9822. 

We don't handle some complex query types:
- Batch/Transaction queries: We don't have enough context in the client to know the intention of each component request; earlier versions of this PR tried to parse the HTTP request to infer it, but for now we omit that complexity and require users to emit the relevant notices themselves. (This matches current behavior where someone using `medplum.batchExecute(...)` would have to perform explicit medplum client cache invalidations. This suggests to me that in some future world we will want a more resource-focused batch API that can compose the batch operation from higher level operations we can interpret more readily.
- GraphQL mutations: For similar reasons, GraphQL mutations require manual event dispatch of related events.
- FHIR Operations (example from this PR: scheduling operations like `$book`, `$cancel`): The meaning of these are implementation dependent, and  out client doesn't know enough about what they mean to handle them effectively. They can also be overridden on a per-project basis with custom implementations, so even if the client offers some default behavior for a few common operations we may need to offer hooks to customize them as well.

Follow ups: I plan on extracting the scheduling queries into a reusable hook that can layer in these events for immediate local updates, and webhook subscriptions (when enabled) for getting real-time updates from other clients mutating the same resources. In the interest of keeping this PR more minimal, that is omitted for now. The Provider-app updates are here to demonstrate the usage of the events and hooks.

### DIFF
--- a/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
+++ b/examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
@@ -52,14 +52,13 @@ function ServiceTypeDisplay(props: { value: CodeableConcept }): JSX.Element {
 
 type UpdateAppointmentFormProps = {
   appointment: WithId<Appointment>;
-  onUpdate: (appointment: WithId<Appointment>) => void;
 };
 
 function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
   const medplum = useMedplum();
   const [patient, setPatient] = useState<Patient | undefined>(undefined);
 
-  const { appointment, onUpdate } = props;
+  const { appointment } = props;
   const handleSubmit = useCallback(async () => {
     if (!patient) {
       return;
@@ -75,15 +74,12 @@ function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
       ],
     } satisfies Appointment;
 
-    let result: WithId<Appointment>;
     try {
-      result = await medplum.updateResource(updated);
+      await medplum.updateResource(updated);
     } catch (error) {
       showErrorNotification(error);
-      return;
     }
-    onUpdate?.(result);
-  }, [medplum, patient, appointment, onUpdate]);
+  }, [medplum, patient, appointment]);
 
   return (
     <Form onSubmit={handleSubmit}>
@@ -104,11 +100,8 @@ function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
   );
 }
 
-export function AppointmentDetails(props: {
-  appointment: WithId<Appointment>;
-  onAppointmentUpdate: (appointment: WithId<Appointment>) => void;
-}): JSX.Element {
-  const { appointment, onAppointmentUpdate } = props;
+export function AppointmentDetails(props: { appointment: WithId<Appointment> }): JSX.Element {
+  const { appointment } = props;
   const medplum = useMedplum();
 
   const [encounter, encounterLoading, encounterOutcome] = useSearchOne(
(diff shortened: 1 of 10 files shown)


noahsilas:
This is an alternative to https://github.com/medplum/medplum/pull/9790, which doesn't touch `MedplumClient` but used a bunch of prop-drilling to provide state management within a subtree. Posted early for discussion.

techdavidy:
@claude review

noahsilas:
Exploring the error handling scenarios separately in https://github.com/medplum/medplum/pull/9872 - the best case scenario there is we make all event listeners safer, but this does create a change in error behavior for existing clients & events. If that's too intrusive to make outside of a big version bump I can do something more targeted here.

examples/medplum-provider/src/components/schedule/BookAppointmentForm.tsx
```diff
@@ -135,6 +132,19 @@ export function BookAppointmentForm(props: BookAppointmentFormProps): JSX.Elemen
           (obj: WithId<Slot> | WithId<Appointment>): obj is WithId<Appointment> => obj.resourceType === 'Appointment'
         );
 
+        // $book is a custom operation, so the client cannot classify its modifications
+        // itself; announce them to invalidate caches and notify interested components
+        // (e.g. the schedule calendar).
+        medplum.notifyResourceModified({
+          resourceType: 'Appointment',
+          operation: 'create',
+          id: appointment?.id,
```
techdavidy: we can move this inside the if (appointment), to avoid the appointment?.id

techdavidy: or even better check for appoint if (!appointment)  and return if not found

packages/core/src/client.ts
```diff
@@ -2736,6 +2814,14 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
     this.cacheResource(result, options);
     this.invalidateUrl(this.fhirUrl(resource.resourceType, resource.id, '_history'));
     this.invalidateSearches(resource.resourceType);
+    if (!notModified) {
```
techdavidy: nit: 

const isModified = result !== undefined;
if (isModified)  to avoid the double negation

examples/medplum-provider/src/components/schedule/BookAppointmentForm.tsx
noahsilas: This is all mostly here to satisfy typescript - $book always returns an Appointment if it succeeded. It does feel awkward in here now though, I think I'll try to make that implicit belief an explicit assertion and remove all the rest of these guards.

techdavidy: yeah thanks just to avoid the awkwardness

examples/medplum-provider/src/components/schedule/AppointmentDetails.tsx
```diff
@@ -72,15 +72,12 @@ function UpdateAppointmentForm(props: UpdateAppointmentFormProps): JSX.Element {
       ],
     } satisfies Appointment;
 
-    let result: WithId<Appointment>;
```
techdavidy, line 77: love this

techdavidy (approved):

PR #9823, opened 2026-07-10
feat(react): add `additionalColumns` prop to `SearchControl`

ThatOneBro:
## Summary

Adds an optional `additionalColumns` prop to `SearchControl` for rendering read-only, **computed** columns after the search-result columns.

Unlike columns derived from `search` fields, an additional column:
- is not backed by a search parameter,
- has no sort/filter menu,
- renders arbitrary content per row via `renderCell(resource)`.

This is useful for values that must be computed or fetched separately from the searched resource (e.g. a related resource's status).

```ts
export interface SearchControlAdditionalColumn {
  readonly name: string;
  readonly renderCell: (resource: Resource) => ReactNode;
}
```

## Motivation

Split out of the MFA admin UI work (`derrick-add-mfa-pswd-reset-ui`), where the Users admin table needs to show computed per-member enrollment columns. That branch currently works around the absence of this prop with a companion table; once this lands it can consume `additionalColumns` directly.

## Test plan

- Added `SearchControl` test "Renders additional columns" verifying the header and a computed cell render.
- Full `SearchControl.test.tsx` suite passes (39 tests).


packages/react/src/SearchControl/SearchControl.tsx
```diff
@@ -418,6 +436,13 @@ export function SearchControl(props: SearchControlProps): JSX.Element {
                 </Menu>
               </Table.Th>
             ))}
+            {props.additionalColumns?.map((col) => (
+              <Table.Th key={col.name}>
+                <Text fw={500} p={2}>
```
noahsilas, line 441: #nit I think you could move these style props onto the `Table.Th` component to remove a node from the tree. I sort of like that idea for `p` to make the container more responsible for spacing and sort of don't like it for `fw` which feels more like a property of the text. Probably ignore this #nit and move on 😂

noahsilas (approved):
Nice cleanup of some complex trickery in `app` by making the `<SearchControl>` component more powerful!

PR #9825, opened 2026-07-10
SSRF-Safe Outbound Fetch

codyebberson:
- Add a shared undici `safeAgent` that resolves outbound hostnames, blocks unsafe resolved IPs, and connects to the resolved address directly.
- Install the safe agent as the global dispatcher during server bootstrap unless `allowUnsafeOutbound` is enabled.
- Replace per-call outbound URL validation on fetch paths with dispatcher-level enforcement, while keeping lightweight static checks where they prevent known-bad queue jobs or preserve SMART Health operation error messages.
- Replace legacy `allowInsecureExternalAuthUrl` and `allowInsecureRestHookUrl` config flags with the deployment-level `allowUnsafeOutbound` escape hatch.

## Details

The safe dispatcher closes two SSRF gaps in hosted deployments:

- DNS rebinding: the dispatcher performs DNS lookup in the connect hook and validates the resolved address before opening the socket.
- Redirects: undici applies the dispatcher to each connection, including redirect hops.

For background workers, static pre-queue filtering remains in place for download and rest-hook subscription URLs. This avoids enqueueing jobs that are known to fail and prevents needless retry consumption. The dispatcher remains the authoritative connection-time protection for DNS and redirect behavior.

## Configuration

New server config:

- `allowUnsafeOutbound?: boolean`
- Env var: `MEDPLUM_ALLOW_UNSAFE_OUTBOUND=true`

This is intended only for self-hosted or on-premises deployments that need outbound access to trusted local/private services. It should not be enabled in hosted or cloud-managed environments.

Removed server config:

- `allowInsecureExternalAuthUrl`
- `allowInsecureRestHookUrl`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45366,6 +45366,7 @@
         "rate-limiter-flexible": "11.2.0",
         "semver": "7.8.5",
         "temporal-polyfill": "0.3.2",
+        "undici": "7.28.0",
         "uuid": "14.0.1",
         "validator": "13.15.35",
         "ws": "8.21.0",
@@ -45412,6 +45413,15 @@
       },
       "engines": {
         "node": ">=10"
+      }
+    },
+    "packages/server/node_modules/undici": {
+      "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-7.28.0.tgz",
+      "integrity": "sha512-cRZYrTDwWznlnRiPjggAGxZXanty6M8RV1ff8Wm4LWXBp7/IG8v5DnOm74DtUBp9OONpK75YlPnIjQqX0dBDtA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.18.1"
       }
     }
   }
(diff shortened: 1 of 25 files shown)


codyebberson:
@claude please review

ThatOneBro (approved):
This seems sound to me. Worth noting that this removes two existing config options but they were only a few weeks old so think it's worth the simplification

LGTM 👍

codyebberson:
This was tested and validated in our AWS staging environment, but we identified two legitimate cases of insecure HTTP:

1. AWS SDK connecting to IMDS for auth
2. Sending metrics and traces to OTel collector sidecar

Those both work today, but only because they don't use Undici or Node.js built-in fetch.  We probably shouldn't count on that to be true forever.

I'm going to pull back on `setGlobalDispatcher(safeAgent)` and make this opt-in at the `fetch` call sites.

codyebberson:
@claude please re-review

codyebberson:
@claude please re-review

codyebberson:
Deploying to staging: https://github.com/medplum/medplum/actions/runs/29163540857

codyebberson:
Server deploy failed, see: https://github.com/medplum/medplum/issues/9837

PR #9826, opened 2026-07-10
feat(server): add `bullmq` globalConcurrency setting

ThatOneBro:
## Summary

Adds a `globalConcurrency` option to `MedplumBullmqConfig` that sets a **cluster-wide** concurrency limit on a BullMQ queue. Unlike the existing per-worker `concurrency`, this limit is enforced globally across all workers of a queue via Redis.

Because global concurrency is a queue-level property persisted in Redis, it can only be changed imperatively via `queue.setGlobalConcurrency(n)` / `queue.removeGlobalConcurrency()` — not through constructor options.

## Changes

- **`config/types.ts`** — new `globalConcurrency?: number` on `MedplumBullmqConfig`.
- **`config/utils.ts`** — registered `bullmq.globalConcurrency` as an integer config key so env-var config parses it as a number.
- **`workers/utils.ts`** — new `applyGlobalConcurrency(queue, config)` helper:
  - configured → `setGlobalConcurrency(n)`
  - omitted → `removeGlobalConcurrency()` (clears a previously-set limit; the value lives in Redis so it must be explicitly removed)
  - fire-and-forget with `.catch()` logging, so it doesn't block the synchronous worker init flow.
- **`workers/index.ts`** — called in `initWorkers` for each registered queue, using the merged global + per-worker config from `getWorkerBullmqConfig`.
- **`workers/utils.test.ts`** — tests for set / remove / undefined-config / rejection-logging.

## Config example

```json
{ "workers": { "bullmq": { "batch": { "globalConcurrency": 5 } } } }
```

## Note

This applies on every instance at startup, so all instances in a fleet should carry the same config — matching how the rest of the `bullmq` config is expected to be fleet-consistent.

## Test plan

- [x] `npx vitest run src/workers/utils.test.ts` — 17 passed
- [x] `npx tsc --noEmit` — clean
- [x] eslint + prettier — clean


packages/server/src/workers/utils.ts
```diff
@@ -259,6 +259,30 @@ export function addVerboseQueueLogging<TDataType>(
...
 
+/**
+ * Applies the configured global concurrency limit to a queue.
+ *
+ * Global concurrency is a queue-level property persisted in Redis, so it must be set imperatively
+ * rather than through queue/worker constructor options. When `globalConcurrency` is configured, the
+ * limit is set; when it is omitted, any previously-set limit is removed so that clearing the config
+ * takes effect on the next startup.
+ *
+ * The underlying Redis calls are fire-and-forget: failures are logged but do not block worker init.
```
mattlong: Not blocking worker init on errors feels risky here since it means the intended config isn't reflected. e.g. if some process is designed around having a global concurrency of one and that isn't actually enforced, it could have unintended consequences.

I know I tend to be on the paranoid side about such things, but the chance that these calls failing is so low, we might as well just hard fail on it. I could be convince otherwise if we were confident our logging/alerting would flag it to us quickly.

ThatOneBro: That's fair, I can change this to just `await promise` instead and let it throw if it throws

That actually does seem more "correct" to me

ThatOneBro: Okay updated

mattwiller (approved):

mattlong (approved):

ThatOneBro:
@claude can you do a final review on this making sure that there are no regressions for existing configs that predate the `globalConcurrency` feature?

PR #9829, opened 2026-07-11
Do not redownload already local async batch result chunks

mattlong:
Avoid needlessly fetching result chunks already in memory. e.g., in the happy path of an uninterrupted batch, no results need be fetched from durable storage.


ThatOneBro (approved):
Nice 👍

PR #9842, opened 2026-07-13
fix(app): clear session on Register click to prevent redirect loop

### LINKED ISSUE #9787: Register button from change project screen gets stuck
**From a user's perspective:** When you're logged in and attempt to change projects, it presents you with another login screen. From there, when you hit "Register", nothing happens, and you stay in the login page. 

**What's actually happening** (from my understanding): The first login screen is used to have the user log into the account with the desired project. If you hit "Register", then the _new_ register screen asks the user what account they want to use to create a _new_ project. These are different workflows, but visually, they look pretty much the same, and I can see an end user getting confused about this pretty easily. 

https://github.com/user-attachments/assets/6aefadfa-5d81-4732-a904-fb027e6d5da4

sahilnyk:
## Summary

Fixes #9787

When a logged-in user hits "Register" (either from the plain sign-in page or the `project=new` re-auth screen), `RegisterPage` bounces them back since they already have a profile, leaving them stuck with a dead link.

## Fix

Dropped the earlier `signOut()` approach entirely, no session changes on click anymore. Register now only renders when `!profile`, so it's simply hidden whenever the user is logged in, regardless of screen or query param. No bounce, no dead click, no logout side effect.

https://github.com/user-attachments/assets/3603c2c7-6e41-4baf-bde5-294feae7b7d4

## Checklist

- [x] Manual testing passed (local dev)
- [x] TypeScript compiles clean
- [x] ESLint passes
- [x] Build succeeds
- [x] Unit tests updated and passing
- [x] No new dependencies or config changes


techdavidy:
@claude please review

techdavidy:
@claude should there be a signup button at that point when we are just re authenticating ?

sahilnyk:
All the review feedback has been implemented

codyebberson (changes requested):
This is a pretty significant change in behavior.  This would log out the user as soon as they hit "Register", correct?  We need to discuss internally before we can do anything of this magnitude

```ts
        isRegisterEnabled() && searchParams.get('project') !== 'new'
          ? async () => {
              // Only clear a live session; RegisterPage bounces logged-in users
              // back to /signin?project=new otherwise. Fall back to clear() if
              // the server logout call fails.
              if (medplum.getProfile()) {
                await medplum.signOut().catch(() => medplum.clear());
              }
              navigate('/register')?.catch(console.error);
            }
          : undefined
      }
```

sahilnyk:
> This would log out the user as soon as they hit "Register", correct? 


@codyebberson Yes, it would log the user out. Since that's the significant change in behavior you flagged, I dropped it, no signOut() call left anywhere. Register just doesn't show up when there's an active session, so nothing of that magnitude is happening on click anymore. Video attached shows the current behavior.

Please review if that's what you want.

https://github.com/user-attachments/assets/4618df04-3925-42c3-a14a-59574a93d89c

Following up on the "discuss internally" part, want to lay out the actual tradeoff so it's easier to talk through.

RegisterPage sends any logged-in user back (if (medplum.getProfile()) navigate('/signin?project=new')) no matter how they got there. So the loop was never really about project=new specifically, that's just the path that showed it. There are two ways to fix it:

1. Keep the feature, but log out first (the version you flagged): clicking Register while logged in signs the user out, then sends them to /register, so they can still make a second account without extra steps. Downside: the logout happens without warning.
2. What's in the PR now: just hide Register whenever someone's logged in. No surprise logout, but now there's no way to register a second account without logging out manually first.

So the real question is: does anyone actually use "register a second account while still logged in"? If yes, option 1 (with a warning before logging out) is probably worth it. If not, option 2 is simpler and it's already done.

Let me know which one you'd rather go with, happy to switch back if needed.

PR #9800, opened 2026-07-09
WIP, do not merge: testing upgrade to TypeScript 7

The-Alchemist:
## Summary

Experimental upgrade to **TypeScript 7.0.2** (native Go compiler) for faster `tsc` builds, using the side-by-side alias pattern recommended in the [TypeScript 7.0 announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/).

Medplum was already on **TypeScript 6.0.3**, which is the recommended prerequisite. No source or `tsconfig` changes were required — the root config is already aligned with TS 7 defaults (`strict`, `module: esnext`, `moduleResolution: bundler`, `erasableSyntaxOnly`, no deprecated options).

### 1. Side-by-side compiler setup (root `package.json`)

TypeScript 7 does not ship a stable programmatic API until 7.1. A direct `typescript@7`-only upgrade would break **ESLint** — `@medplum/eslint-config` uses `projectService: true`, which requires `require('typescript/lib/tsserverlibrary')`, and the TS 7 package does not provide that.

**API Extractor is not affected** — `@microsoft/api-extractor@7.58.9` bundles its own TypeScript compiler and is invoked via CLI, independent of the project `typescript` package.

The fix is npm aliases:

| Package | Role |
|---------|------|
| `@typescript/native` → `typescript@^7.0.2` | Fast `tsc` for all build scripts (`npx tsc` → 7.0.2) |
| `typescript` → `@typescript/typescript6@^6.0.2` | TS 6 API for `typescript-eslint` only |

Also updated `overrides.typescript` to `npm:@typescript/typescript6@^6.0.2` so all workspaces resolve the same shim.

### 2. Workspace dependency cleanup (23 `package.json` files)

Removed redundant explicit `"typescript": "6.0.3"` devDependencies from:

- `packages/react`, `packages/react-hooks`, `packages/docs`
- All 21 example apps that pinned it

These packages now rely on root hoisting + the npm override for a single consistent version.

### 3. ESLint — no patch required

The side-by-side setup initially looked like it would need a consumer-side `tsserverlibrary` shim ([microsoft/TypeScript#63422](https://github.com/microsoft/TypeScript/issues/63422)). That issue is **closed** — the fix shipped in `@typescript/typescript6@6.0.1+` (shim added in [TypeScript-Website#3559](https://github.com/microsoft/TypeScript-Website/pull/3559)). Pinning to `^6.0.2` is sufficient; `npm run lint` passes without changes to `@medplum/eslint-config`.

[...]

PR #9844, opened 2026-07-13
Dependency upgrades 2026-07-13

medplumbot:
Dependency upgrades


package.json
```diff
@@ -29,33 +29,33 @@
...
     "@vitejs/plugin-react": "6.0.3",
     "cross-env": "10.1.0",
     "esbuild": "0.28.1",
     "formidable": "3.5.4",
     "jsdom": "29.1.1",
-    "npm-check-updates": "22.2.7",
+    "npm-check-updates": "22.2.9",
     "nyc": "18.0.0",
-    "prettier": "3.8.4",
+    "prettier": "3.9.5",
```
noahsilas, line 47: This upgrade causes some autofix churn (see ccb6fb58473cff84f4e5b868d4f92ea50356acae).

package-lock.json
```diff
@@ -8641,6 +8486,9 @@
       "cpu": [
         "arm64"
       ],
+      "libc": [
+        "glibc"
+      ],
```
noahsilas, line 8491: I'm not sure why running `scripts/reinstall.sh --update` adds this churn to the lockfile (run with npm v11.16.0 on MacOS, which seems like it should emit the same lockfile as the CI build 🤔 )

noahsilas (approved):
We're holding back two major version upgrades:
- Typescript: Already in progress elsewhere, see #9800 
- GraphQL: Some dependencies are still in cooldown, this should apply cleanly next week when this runs again.

Of note: a `prettier` upgrade has introduce some formatting changes, the most notable one being that TS unions now stay on one line when they fit.

PR #9845, opened 2026-07-13
Support project-scoped external JWT bearer authentication

codyebberson:
This PR allows Medplum to accept externally issued JWTs directly as bearer tokens when the target project is identified by a project-scoped API URL.

For a request such as:

```text
/projects/{projectId}/...
/api/projects/{projectId}/...
```

Medplum can resolve external token verification settings from a `ClientApplication` in that project, verify the JWT against the application's configured JWKS endpoint, and authenticate the request as that client application's project membership.

Existing server-wide external bearer configuration and the existing `fhirUser` / `sub` identity mapping behavior remain supported.

## Motivation

Some external systems issue valid JWT bearer tokens that identify the issuer but do not include a Medplum project, membership, or profile identifier. The ONC Inferno CRD validator is one example.

Before project-scoped API URLs were available, the same external issuer and token shape could not be routed safely to different Medplum projects. A hardcoded membership or a server-wide issuer-to-project mapping would work for only one project and would not provide tenant isolation.

Project-scoped API URLs now provide the missing routing signal. This PR uses that signal to locate the external authentication configuration and client identity within the target project.

## Behavior

External bearer authentication now resolves its configuration in this order:

1. Look for a matching issuer in the existing server-wide `externalAuthProviders` configuration.
2. If no server-wide provider matches and the request uses a project-scoped URL, search that project's `ClientApplication` resources for a unique application whose `identityProvider.issuer` matches the JWT issuer.
3. If neither source produces a provider, reject the token.

When configuration is resolved from a project-local `ClientApplication`:

- the client search runs against the project's physical shard;
- the search is also constrained by `_project` for logical project isolation;
- exactly one client must match the issuer;
- the JWT is verified using the client's `identityProvider.jwksUrl`;
- the request is authenticated as that client's existing `ProjectMembership`;
- the external-auth cache key includes the URL project ID.

Zero or multiple matching clients fail closed.

## IdentityProvider changes

`IdentityProvider` now supports verification-only configurations:

- Adds `issuer`.
- Adds `jwksUrl`.
- Makes browser OAuth fields optional:
  - `authorizeUrl`
  - `tokenUrl`
  - `userInfoUrl`
  - `clientId`
  - `clientSecret`

This allows a client application to describe an external JWT issuer without pretending to be a complete OAuth authorization-code provider.

Code paths that perform browser login or token exchange now validate the optional fields they require and return a configuration error when they are missing.

## Token verification

External bearer tokens are verified as follows:

- If `jwksUrl` is configured, Medplum verifies the JWT signature and issuer using the remote JWKS.
- Otherwise, Medplum preserves the existing userinfo-based verification behavior.
- Outbound userinfo and JWKS requests use `safeFetch`.

## Compatibility

- Existing server-wide `externalAuthProviders` continue to work.
- Existing userinfo-based external bearer authentication continues to work.
- Existing legacy bearer identity mapping remains unchanged:
  - `fhirUser` or `ext.fhirUser` maps through the referenced profile.
  - `sub` maps to `ProjectMembership.externalId`.
- Existing non-project-scoped API routes remain unchanged.
- A project-local client fallback is only available when the request URL contains a project ID.
- Browser OAuth login and token exchange continue to require the endpoint and client fields used by those flows.

### Server-wide issuer-to-project configuration

We considered adding project IDs to the server's external provider configuration. That would keep tenant configuration in deployment config, would not scale well for hosted multi-tenant environments, and would duplicate identity already represented by project-local client applications.

### Mapping arbitrary token claims to users or memberships

We explored configurable email, subject, and FHIR-user mapping modes. They did not solve the motivating case because the validator token has no claim that identifies the target Medplum project or membership. The final implementation keeps the existing `fhirUser` and `sub` compatibility paths and uses the project URL plus `ClientApplication` for identity-less tokens.

### Explicit verification-method and audience configuration

We also explored a separate verification-method field and configurable audience validation. Those were broader external-auth features than this change requires. The final implementation selects JWKS verification when `jwksUrl` is present and otherwise preserves userinfo verification.

## Security properties

- JWT signatures are verified against the configured JWKS.
- The JWT issuer must match the configured identity-provider issuer.
- The URL project is included in the external-auth cache key.
- Project-local client lookup uses both the correct physical shard and an explicit `_project` constraint.
- Ambiguous issuer matches within a project are rejected.
- The authenticated identity is an existing client membership; the URL does not create a membership or grant access by itself.
- Existing authenticated project-scope validation still rejects requests where the resolved membership does not belong to the URL project.

### DIFF
--- a/packages/core/src/base-schema.json
+++ b/packages/core/src/base-schema.json
@@ -4238,8 +4238,21 @@
   },
   "IdentityProvider": {
     "elements": {
+      "issuer": {
+        "type": [
+          {
+            "code": "string"
+          }
+        ]
+      },
+      "jwksUrl": {
+        "type": [
+          {
+            "code": "string"
+          }
+        ]
+      },
       "authorizeUrl": {
-        "min": 1,
         "type": [
           {
             "code": "string"
@@ -4247,7 +4260,6 @@
         ]
       },
       "tokenUrl": {
-        "min": 1,
         "type": [
           {
             "code": "string"
@@ -4262,7 +4274,6 @@
         ]
       },
       "userInfoUrl": {
-        "min": 1,
         "type": [
           {
             "code": "string"
@@ -4284,7 +4295,6 @@
         ]
       },
       "clientId": {
-        "min": 1,
         "type": [
           {
             "code": "string"
@@ -4292,7 +4302,6 @@
         ]
       },
       "clientSecret": {
-        "min": 1,
         "type": [
           {
             "code": "string"
--- a/packages/definitions/src/fhir/r4/fhir.schema.json
+++ b/packages/definitions/src/fhir/r4/fhir.schema.json
@@ -62924,6 +62924,14 @@
     "IdentityProvider": {
       "description": "External Identity Provider (IdP) configuration details.",
       "properties": {
+        "issuer": {
+          "description": "Issuer URL for the external Identity Provider.",
+          "$ref": "#/definitions/string"
+        },
+        "jwksUrl": {
+          "description": "Remote URL for the external Identity Provider JWKS endpoint.",
+          "$ref": "#/definitions/string"
+        },
         "authorizeUrl": {
           "description": "Remote URL for the external Identity Provider authorize endpoint.",
           "$ref": "#/definitions/string"
@@ -62988,14 +62996,7 @@
           ]
         }
       },
-      "additionalProperties": false,
-      "required": [
-        "authorizeUrl",
-        "tokenUrl",
-        "userInfoUrl",
-        "clientId",
-        "clientSecret"
-      ]
+      "additionalProperties": false
     },
     "ProjectMembership_Access": {
       "description": "Extended access configuration using parameterized access policies.",
--- a/packages/definitions/src/fhir/r4/fhir.schema.json
+++ b/packages/definitions/src/fhir/r4/fhir.schema.json
@@ -62928,6 +62928,10 @@
           "description": "Issuer URL for the external Identity Provider.",
           "$ref": "#/definitions/string"
         },
+        "audience": {
+          "description": "Expected audience for tokens issued by the external Identity Provider.",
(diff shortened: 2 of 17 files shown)


codyebberson:
@claude - please review.  Please pay particular attention to any changes to existing functionality.

For context, there are 3 pathways to external auth:
1. OAuth authorization code flow (normal 3-legged OAuth) which results in a callback to `/auth/external` endpoint
2. OAuth token exchange (`urn:ietf:params:oauth:grant-type:token-exchange`) with the `/oauth2/token` endpoint
3. And the newly introduced direct bearer token

The newly introduced direct bearer token should only affect Medplum clusters with the `externalAuthProviders` server config setting.

We need to ensure that authorization code flow and token exchange flow are functionally unaffected by this PR.

The-Alchemist:
project-scoped API endpoints PR at https://github.com/medplum/medplum/pull/10173

codyebberson:
@claude - please review

Please pay extra attention to:
* any possible breaking changes to existing behavior
* any possible security considerations

codyebberson:
@claude - please re-review

> 🔴 Performance / DoS — unbounded client scan on the hot path, pre-verification

Fixed.  please re-review

> 🟠 Security — no audience (aud) validation on JWKS-verified tokens

Fixed.  please re-review

> 🟠 Correctness — ambiguity check can be bypassed by the 1000 cap

Fixed.  please re-review

> JWKS set not reused (utils.ts:977):

Noted.  The alternative of caching JWKS values has its own set of risks.

> Cache TTL vs token exp (utils.ts:1262):

Noted.

> method.ts:50 — isExternalAuth now returns undefined unless idp.authorizeUrl && idp.clientId are both set.

Those were previously required fields, and are now optional.  Any existing client must have those values set.

> Redundant lookup: when a server-wide provider matches

I believe this is fixed by the performance fix

ThatOneBro (approved):

PR #9846, opened 2026-07-13
Adding public twilio docs

AndyStoneman:


packages/docs/docs/integration/twilio-sms/setup.md
```diff
@@ -0,0 +1,71 @@
...
+}
+```
+
+`TWILIO_FROM_NUMBER` is optional — omit it if callers will always supply the sending number via the `https://medplum.com/twilio-from-number` extension on the `Communication` resource. If provided, the install operation automatically registers the inbound webhook with Twilio for that number.
+
+Re-running the install operation is safe — secrets already set in the project will not be overwritten.
+
+## Project Secrets
+
+All secrets are written automatically by `$twilio-sms-install` and read at runtime by the integration bots:
```
mattlong, line 32: Should the docs explain that the integration is powered by bots and other resources in a linked project (assuming that's the case)?

packages/docs/docs/integration/twilio-sms/index.md
```diff
@@ -0,0 +1,36 @@
...
+sidebar_position: 0
+tags: [integration, communications]
+---
+
+# Twilio SMS Integration
+
+Medplum provides a first-party integration with Twilio to send and receive SMS messages directly from your healthcare application. Messages are stored as FHIR `Communication` resources, enabling seamless integration with your clinical workflows.
+
+:::caution[Medplum Team Setup Required]
+This integration requires setup by the Medplum team. [Contact us](mailto:info@medplum.com?subject=Twilio%20SMS%20Integration) to enable Twilio SMS for your project.
```
mattlong: It'd be nice if this could more explicitly say that this is only available on hosted Medplum to avoid confusion for self-hosters and folks poking around the repo.

mattlong (changes requested):

packages/docs/docs/integration/twilio-sms/index.md
```diff
@@ -0,0 +1,36 @@
+---
+sidebar_position: 0
+tags: [integration, communications]
+---
+
+# Twilio SMS Integration
+
+Medplum provides a first-party integration with Twilio to send and receive SMS messages directly from your healthcare application. Messages are stored as FHIR `Communication` resources, enabling seamless integration with your clinical workflows.
```
finnbergquist: I think we should probably remove "first party" if customer needs to bring their own account

packages/docs/docs/integration/twilio-sms/setup.md
```diff
@@ -0,0 +1,71 @@
...
+```json
+{
+  "resourceType": "Parameters",
+  "parameter": [
+    { "name": "TWILIO_FROM_NUMBER", "valueString": "+15005550006" }
+  ]
+}
+```
+
+The phone number is normalized to E.164 before lookup — see [Phone Number Formats](/docs/integration/twilio-sms/sending-sms#phone-number-formats) for accepted input formats. This operation can be called multiple times for customers with multiple Twilio numbers.
```
finnbergquist: I think it is a little awkward to reference _customers_ in docs. Maybe rephrase to
 
This operation can be called multiple times to setup different Twilio numbers

packages/docs/docs/integration/twilio-sms/setup.md
```diff
@@ -0,0 +1,71 @@
+---
+sidebar_position: 1
+---
+
+# Setup
+
+## Step 1 — Install the integration
+
+A project admin invokes the `$twilio-sms-install` operation once per customer project, passing the Twilio credentials. This stores the required secrets and creates the webhook `ClientApplication`.
```
finnbergquist: -> operation once in your project

finnbergquist (approved):

mattlong (approved):

PR #9847, opened 2026-07-13
Run generator

mattwiller:


packages/core/src/base-schema.json
```diff
@@ -2928,6 +2928,13 @@
             "code": "Reference"
           }
         ]
+      },
+      "deleted": {
+        "type": [
+          {
+            "code": "boolean"
```
codyebberson, line 2935: I think this will fail tests because there's a test that exercises all Medplum "extended" properties.

packages/definitions/src/fhir/r4/fhir.schema.json
```diff
@@ -787,6 +788,9 @@
     },
     {
       "$ref": "#/definitions/LoginEmailMfa"
+    },
+    {
+      "$ref": "#/definitions/ProjectDefaultAccessPolicies"
```
codyebberson, line 793: This ship sailed here, but this is a reminder to use singular property names, never plural 😢

codyebberson (approved):

PR #9849, opened 2026-07-13
Limit codesystem imports

mattwiller:


mattlong (commented):
Any concerns about backward-incompatibility?

mattwiller:
@mattlong Since this limit applies at the operation endpoint level (and not the import functionality itself), explicit `CodeSystem/$import` calls are the only surface area to be concerned with.  For those, our internal imports should all be using a batch size of 500 currently — giving us 2x headroom.  I think this limit is a good one in general: we want to strongly encourage spreading out work on the server.  If any users were previously sending batches over the allowed size limit, my hope is that the fix is simple enough: just split the oversize import into multiple calls

The-Alchemist (approved):

mattlong (approved):

codyebberson:
@claude please review

codyebberson:
@claude - please re-review after rebasing

PR #9850, opened 2026-07-13
Surface missing shared-project dependencies at sign-in

### LINKED ISSUE #9824: Medplum Provider should surface missing shared-project dependencies upfront
## Context 

Medplum Provider assumes several Medplum "shared projects" are linked into the user's project (e.g. UMLS value sets, US Core profiles, or integration bots). Hosted projects get these out-of-the-box via project linking, but self-hosted projects or projects running from source have none of them. 

Today, nothing checks this assumption, which can lead to some weird behavior like #9687 . 

We don't surface these errors very well; from an end user's perspective, a single missing project can result in multiple fields showing errors across the app when they're missing associated ValueSets / suggestions. A few examples: 

- ValueSet bound fields (like diagnoses bound to ICD-10 codes) fail `$expand`
- The "Performing lab" field in Order Labs errors with "Not found" on every search because the Health Gorilla autocomplete bot doesn't exist 
- Profile-driven forms (e.g. the US Core Patient edit form) show "Not found" when the profile StructureDefinition is missing 

Each looks like a distinct bug, but they all have the same root cause, and the user is not told what that is & how to fix it. 

## Proposal 

We should detect the missing dependencies upfront, and present them in a message to the user. 

1. Define the provider app's expected dependencies (known ValueSet URLs, profile URLs, bot IDs) in one place.
2. Probe them once per session (cheap requests: `$expand` with `count: 1`, profile/bot reads) after sign-in.
3. Surface one consolidated message, like _"The following projects are missing: UMLS terminology, Health Gorilla integration..."_, with links to relevant setup docs, or a message to contact their system administrator.

Individual fields should keep their local error messages from #9819 , but this fix would complement that and give the user more guidance on what to do to fix it. 

## Related 
- #9819 fixes the error toast spam issue called out in #9687 
- The Intake form already has a page-specific version of this fix, as a top-level message that tells the user that some fields are missing ValueSets. This issue generalizes this fix to the whole app.

Filupnot:
Fixes #9824 

## Problem Summary

Medplum Provider implicitly depends on several Medplum "shared projects" being linked into the user's project (UMLS terminology, US Core profiles, Health Gorilla integration bots). Hosted projects get these via project linking, but self-hosted / from-source projects have none of them, and nothing checks the assumption, so a single missing project surfaces as many unrelated-looking "Not found" errors across the app.

## Changes

This PR detects the missing dependencies once after sign-in and surfaces them in a single consolidated, dismissible banner with links to setup docs.

image

The app's dependencies will be declared in a new file/manifest, `config/appDependencies.ts` , grouped by project (UMLS terminology, US Core profile, Health Gorilla), each with a docs link and a way to "probe" the source to check if it exists or not. 

A new hook, `hooks/useMissingDependencies.ts`, does a shallow existence-check ("probe") on each group after the user signs in: 
- ValueSets are checked via `@medplum/react`'s shared `checkValueSetAvailability` (introduced in #9819 ) 
- Profiles and Bots are checked via `searchOne`

A group is marked missing if _at least one_ of its probes definitively resolves as missing (400/404/empty search) and _none_ resolve as present. A single present probe proves the shared project is linked, so it overrides everything else in that group- even if another probe in the same group came back missing or inconclusive.

## Relationship to #9819

This PR re-uses `checkValueSetAvailability` introduced in #9819 , so that the banner probe and the inline field-level checks share the same client cache (we only check for presence of a ValueSet once, and a missing one recovers automatically once linked). We're keeping the inline checks introduced by that PR, but this one just also adds a top-level "global" check for the root cause of the issue at sign-in. 

#9819 should be merged first. After that, this PR should then target a merge base of `main`. 

## Note on maintenance 

This manifest is grouped by shared _project_, not by individual resource; each project needs only one or two canary probes, since one missing ValueSet/profile/bot from a project implies the whole project is unlinked. Because of that, most new fields don't require a manifest update: they typically reuse an already-tracked project. Drift is only a risk when a field depends on a genuinely new shared project.

A fully dynamic list isn't practical here: some bindings (e.g. a customer's own `Questionnaire.answerValueSet`) are data-driven and can't be enumerated from source at all, not just impractically so. A static manifest is cheap to check upfront, at the cost of needing a human to add an entry when a new shared project is introduced.


Filupnot:
@claude review this

examples/medplum-provider/src/config/appDependencies.ts
```diff
@@ -0,0 +1,89 @@
...
+      { kind: 'valueSet', url: 'http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1010.4' },
+    ],
+  },
+  {
+    id: 'us-core-profiles',
+    name: 'US Core profiles',
+    docsUrl: 'https://www.medplum.com/docs/fhir-datastore/profiles',
+    probes: [{ kind: 'profile', url: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient' }],
+  },
+  {
```
techdavidy: Probably we don't need to surface HG as it is an integration.

examples/medplum-provider/src/config/appDependencies.ts
```diff
@@ -0,0 +1,89 @@
...
+export const DEPENDENCY_GROUPS: readonly DependencyGroup[] = [
+  {
+    id: 'umls-terminology',
+    name: 'UMLS terminology',
+    docsUrl: 'https://www.medplum.com/docs/terminology',
+    probes: [
+      // ICD-10-CM (diagnoses) and a representative UMLS/NLM value set. Both are backed by the
+      // UMLS terminology shared project, so probing one or two representatives is sufficient.
+      { kind: 'valueSet', url: ICD10_CM_BILLABLE_VALUESET },
+      { kind: 'valueSet', url: 'http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1010.4' },
```
noahsilas: Is there a scenario where one of these probes is not sufficient to check that the UMLS data has been linked in?

It seems like things would simplify a bit if we could unwrap this one layer and always provide a single probe, but I might not be considering some scenarios you are covering.

examples/medplum-provider/src/components/MissingDependenciesBanner.tsx
```diff
@@ -0,0 +1,80 @@
...
+  } catch {
+    return false;
+  }
+}
+
+function writeDismissed(key: string): void {
+  try {
+    sessionStorage.setItem(key, 'true');
+  } catch {
+    // Ignore storage failures; dismissal falls back to component state.
```
techdavidy: console.error

Filupnot:
@claude re-review after the recent changes.

examples/medplum-provider/src/App.tsx
```diff
@@ -178,6 +179,7 @@ export function App(): JSX.Element | null {
       resourceTypeSearchDisabled={true}
       spotlightPatientsOnly={true}
     >
+      {profile && <MissingDependenciesBanner />}
```
noahsilas, line 182: What would you think about putting this in the "GetStartedPage" instead of in the app root? Right now that probably resolves the common case of someone "getting started" with a new project in localdev or a self-hosted environment, but could would avoid running these queries on every page load when someone is navigating into an established project.

examples/medplum-provider/src/components/MissingDependenciesBanner.tsx
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * See https://github.com/medplum/medplum/issues/9824.
+ * @returns The banner, or null when nothing is missing / it has been dismissed.
+ */
+export function MissingDependenciesBanner(): JSX.Element | null {
+  const medplum = useMedplum();
+  const sessionKey = medplum.getProject()?.id ?? medplum.getProfile()?.id ?? 'default';
+  const { missingGroups, loading } = useMissingDependencies();
+  const dismissedKey = DISMISSED_KEY_PREFIX + sessionKey;
+  const [dismissed, setDismissed] = useState(() => readDismissed(dismissedKey));
```
noahsilas: Could we adjust this so that if someone has already dismissed the banner we can skip the queries? As an example, what about something like this:

```suggestion
  const dismissedKey = DISMISSED_KEY_PREFIX + sessionKey;
  const [dismissed, setDismissed] = useState(() => readDismissed(dismissedKey));
  const { missingGroups, loading } = useMissingDependencies({ enabled: !dismissed });
```

examples/medplum-provider/src/hooks/useMissingDependencies.ts
```diff
@@ -0,0 +1,105 @@
...
+      // detectMissingGroups doesn't reject in practice (runProbe catches every probe error and
+      // returns a verdict); this no-op handler only satisfies the no-floating-promises rule.
+      .catch(() => undefined);
+    return () => {
+      active = false;
+    };
+  }, [medplum, projectId]);
+
+  const resolved = result?.projectId === projectId ? result : undefined;
+  return { missingGroups: resolved?.missing ?? [], loading: !resolved };
```
noahsilas, line 115: I wonder if we can cache successful results so that we don't need to run these queries on every page load - the intuition here being that after these resources are available at all, that result probably doesn't change (it's relatively rare for a project to _lose_ access to a linked project).

One idea might be writing successful results into local storage (possibly including a version ID or some kind of hash of the probes used so that if that set changes this could re-run automatically to pick up added dependencies?)

examples/medplum-provider/src/hooks/useMissingDependencies.ts
```diff
@@ -0,0 +1,105 @@
...
+          return 'present';
+        }
+        return availability.status === 'unavailable' ? 'missing' : 'unknown';
+      }
+      case 'profile': {
+        const found = await medplum.searchOne('StructureDefinition', { url: probe.url });
+        return found ? 'present' : 'missing';
+      }
+      case 'bot': {
+        const found = await medplum.searchOne('Bot', { identifier: `${probe.system}|${probe.value}` });
```
noahsilas: Would it make sense to make this more general and let probes include arbitrary resource queries?

```tsx
      case 'resource': {
        const found = await medplum.searchOne(probe.resourceType, probe.query);
```

On the one hand it's nice to be explicit in what these queries are as you have been, but on the other hand pushing the full query into the probe definition would be more readable in that one place.

I don't have a strong opinion on the answer here, if you like the current version I'm happy with it as-is too 👍

techdavidy:
Thanks for taking care of this. 

 At its core the feature is: run 4 existence checks after sign-in, show one dismissible Alert. 
 To do that it adds ~275 lines of source across 3 new files.
 
 I think we can accomplish the same with less LOC. (easier to understand)

examples/medplum-provider/src/hooks/useMissingDependencies.ts
```diff
@@ -0,0 +1,105 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { MedplumClient } from '@medplum/core';
+import { checkValueSetAvailability, useMedplum } from '@medplum/react';
+import { useEffect, useState } from 'react';
+import type { DependencyGroup, DependencyProbe } from '../config/appDependencies';
+import { DEPENDENCY_GROUPS } from '../config/appDependencies';
+
+type ProbeResult = 'present' | 'missing' | 'unknown';
```
techdavidy, line 10: isAvailable?:  boolean
It could be true, false, undefined.

examples/medplum-provider/src/App.tsx
techdavidy: @noahsilas maybe this is another great use case for the listeners in the MedplumClient.

examples/medplum-provider/src/hooks/useMissingDependencies.ts
Filupnot: I think I'd prefer to keep this one as-is... this potentially invites confusing behavior with `if (isAvailable)` / `!isAvailable`, where `undefined` and `false` have different meanings that might not be super clear with a boolean. That being said I don't have too strong opinions either way, so let me know if you feel strongly about this.

examples/medplum-provider/src/config/appDependencies.ts
Filupnot: Honestly I think you're right. This use case was more relevant for the `bot` cases, but we removed that from another comment (don't think health gorilla should necessarily be listed as a "dependency" for all Provider instances). Changed to just contain one probe per entry.

examples/medplum-provider/src/App.tsx
Filupnot: I pushed some commits that make the banner more intelligent about when it actually performs the probes; we don't re-probe if the user dismissed the banner, or if we already get back a `present` response. So I don't think performance is a concern here. 

Part of me feels like if we're confident that a required valueset/profile is missing, then we should make that warning as obvious as possible, even for "established" projects. If users have been using a project for a while that's been missing these dependencies, we'd probably want to make it obvious to them, and they'd miss it because they wouldn't be launching the app in the "getting started" page. 

Let me know if you disagree / if I'm misunderstanding the use case here

examples/medplum-provider/src/components/MissingDependenciesBanner.tsx
Filupnot: Agreed, although I implemented this in a slightly different way. Let me know what you think.

examples/medplum-provider/src/hooks/useMissingDependencies.ts
Filupnot: Agreed, implemented an update for this

examples/medplum-provider/src/hooks/useMissingDependencies.ts
Filupnot: I don't feel strongly either, so I'll opt to keep it as-is 😄

Filupnot:
> Thanks for taking care of this.
> 
> At its core the feature is: run 4 existence checks after sign-in, show one dismissible Alert. To do that it adds ~275 lines of source across 3 new files.
> 
> I think we can accomplish the same with less LOC. (easier to understand)

It is a fair bit of code. Some justification: 
- We went with a declarative probe manifest instead of just inlining the URLs/logic at each dependent field, which should be more robust & extensible 
- We added a bit of caching to make sure we're not querying when we don't need to
- This includes some React plumbing to make sure we're handling switching switching projects correctly 
- A good chunk of the updated code is tests

If you think that it's worth attempting to trim down the LLOC on this PR then I can look more into that.

examples/medplum-provider/src/config/appDependencies.ts
Filupnot: Removed

examples/medplum-provider/src/components/MissingDependenciesBanner.tsx
Filupnot: Added

techdavidy:
@claude please review

Filupnot:
Closing this PR in favor of #9883  instead.

PR #9851, opened 2026-07-13
Validate intake form response before creating the patient

### LINKED ISSUE #8827: `New Patient` intake form: patient still gets created on validation error
In the `New Patient` form, clicking submit without filling in all required fields can return an error, however the patient still gets created (even though validation error was thrown) 

Steps to repro: 

1. Navigate to `New Patient` form in Provider app
2. Fill in all but one required field
3. Click submit - observe that an error appears for an invalid empty value
    - example: `Invalid empty non-primitive value (QuestionnaireResponse.item[8].item[2].answer[0])`
4. Navigate to Patient list view in Provider app, observe that the patient was created

AbhinavMir:
Closes #8827

 Check everything that can fail before writing anything. Two checks were added at the top of onboardPatient:
1. Ask the server to validate the questionnaire response (same check that would have failed mid-way now runs first).
2. Since the coverage code reads the fields unconditionally, it crashed on partial input. Fix makes sure the insurance section has all answers it needs.

demo

### DIFF
--- a/examples/medplum-provider/src/utils/intake-form.test.ts
+++ b/examples/medplum-provider/src/utils/intake-form.test.ts
@@ -1,8 +1,9 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { allOk, badRequest, OperationOutcomeError } from '@medplum/core';
+import { badRequest, OperationOutcomeError } from '@medplum/core';
 import type { Questionnaire, QuestionnaireResponse, QuestionnaireResponseItemAnswer } from '@medplum/fhirtypes';
 import { MockClient } from '@medplum/mock';
+import type { Mock } from 'vitest';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 
 const mockAnswers = {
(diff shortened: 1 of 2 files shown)


maddyli:
Hi @AbhinavMir ! Thanks so much for this, would you mind also linking the related issue?

AbhinavMir:
Hi @maddyli - added. I am still experimenting, but I think it shud be good soon

AbhinavMir:
@maddyli PR shud be ready now! Sorry I forgot to tag earlier!

Filupnot:
@claude review this please

Filupnot:
@AbhinavMir - Thanks for writing this PR! Can you take a look at Claude's review above, specifically the 1st point? Also to pass our CI/CD DCO make sure you sign the commits (you can do this by appending `-s` to your commit line).

AbhinavMir:
Looking!

AbhinavMir:
@Filupnot I'll be honest, I used Fable and and tried to get this to a more complete state, but it seems like it depends on `transaction-bundles` in this case, and I don't understand the scope of that feature.

"Atomic rollback requires the project's `transaction-bundles` feature; without it the server applies batch semantics" - does this sound about right?

AbhinavMir:
@Filupnot updated! Let me know!

Filupnot:
Thanks for the update, and sorry for the delayed response!

This recent PR may have introduced a bug: The recorder approach re-routes every `add*` helper call through a fake MedplumClient, but most of those helpers (`addAllergy`, `addCondition`, `addCoverage`, `addImmunization`, `addFamilyMemberHistory`, `addPharmacy`, `upsertObservation`) call `medplum.upsertResource(resource, query)`, which is a conditional PUT that updates an existing matching resource instead of creating a duplicate. The recorder only forwards `resource`, not `query`, and `convertToTransactionBundle` always emits a plain POST. So all of those resource types lose their dedup behavior: a retried or resubmitted intake form would now create duplicate allergies, conditions, coverage records, etc., instead of updating in place. That's the "transaction-bundles feature" scope you flagged not understanding; it's a real corner of the FHIR API (conditional create/update inside a bundle via `request.ifNoneExist`), and it's exactly the kind of thing that's hard to get right without someone who's used it before checking the diff.

Given that, I think the cleanest path is to split this back into just the first commit, the validation-before-create fix. I think that's the entire fix for the linked issue, it's already well-tested, and it doesn't need any of the transaction-bundle machinery. If we want atomic submission as a follow-up, it deserves its own PR and issue, since it's a materially bigger change than [#8827](https://github.com/medplum/medplum/issues/8827) asked for, and honestly, given how much FHIR-transaction subtlety it involves, it's probably not a great fit for an open community contribution anyway.

Thanks again for sticking with this one!

PR #9855, opened 2026-07-14
Self-hosting Considerations Documentation

maddyli:


packages/docs/docs/self-hosting/considerations.md
```diff
@@ -2,45 +2,52 @@
...
 
-**Personal Education**
-You want to gain experience with self-hosting infrastructure, and your Medplum instance will only be used for personal projects without production requirements.
+**Staffing on-call coverage.** Production healthcare systems are used on nights and weekends, and your team will need 24/7 on-call coverage for outages. Beyond the upfront work of building incident response protocols and standing up observability and alerting, this is a continuous operational load: staffing on-call rotations, maintaining all supporting infrastructure, and responding to incidents as they occur. Because your team did not build Medplum, diagnosing the root cause of an incident often takes longer than it would for the engineers who wrote and operate the system every day, which counts against your recovery during an active outage.
 
-## Operational Considerations for Self-Hosting
+## Why Medplum Operates Medplum Efficiently
```
The-Alchemist, line 19: LGTM, but is the duplicate "Medplum" here intentional?  "Why Medplum Operates _Medplum_ Efficiently?"

packages/docs/docs/self-hosting/considerations.md
```diff
@@ -2,45 +2,52 @@
...
 
-**Personal Education**
-You want to gain experience with self-hosting infrastructure, and your Medplum instance will only be used for personal projects without production requirements.
+**Staffing on-call coverage.** Production healthcare systems are used on nights and weekends, and your team will need 24/7 on-call coverage for outages. Beyond the upfront work of building incident response protocols and standing up observability and alerting, this is a continuous operational load: staffing on-call rotations, maintaining all supporting infrastructure, and responding to incidents as they occur. Because your team did not build Medplum, diagnosing the root cause of an incident often takes longer than it would for the engineers who wrote and operate the system every day, which counts against your recovery during an active outage.
```
everett-williams: 'We've seen self-hosting customers struggle to diagnose the root cause of an incident quickly or independently, given their teams don't have the same level of institutional knowledge as the engineers who built and maintain the system every day - which counts against recovery time during an active outage'

Softened a little - thoughts?

packages/docs/docs/self-hosting/considerations.md
everett-williams: I believe that is intentional

everett-williams (approved):
Left a couple minor comments but think it looks good overall 👍

packages/docs/docs/self-hosting/considerations.md
maddyli: added

PR #9856, opened 2026-07-14
ci(server): smoke test 'npm run dev'

The-Alchemist:
Runs `npm run dev` and wait for `/healthcheck` in the e2e CI job  to catch tsx/dev boot failures that `npm start` misses.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -590,6 +590,8 @@ jobs:
       - name: Build Medplum
         run: npm run build:fast
       - name: Smoke test npm run dev
+        env:
+          REDIS_PASSWORD_DISABLED_IN_TESTS: 1
         run: |
           pushd packages/server
           setsid npm run dev > server-dev.log 2>&1 &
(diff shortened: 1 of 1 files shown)


The-Alchemist:
@claude review

The-Alchemist:
@claude review

.github/workflows/build.yml
```diff
@@ -589,6 +589,56 @@ jobs:
...
+          pushd packages/server
+          setsid npm run dev > server-dev.log 2>&1 &
+          SERVER_PID=$!
+          popd
+
+          MAX_RETRIES=24
+          RETRY_COUNT=0
+
+          until curl -s -f http://localhost:8103/healthcheck > /dev/null; do
+            if ! kill -0 "$SERVER_PID" 2>/dev/null; then
```
noahsilas, line 605: TIL about `kill -0`, which does not send a signal (there is no signal with id 0), but does perform error checking, and so effectively tests if the process with the given PID is running and the user has permission to send a signal to it. Neat!

.github/workflows/build.yml
```diff
@@ -589,6 +589,56 @@ jobs:
...
+          until ! curl -s -f http://localhost:8103/healthcheck > /dev/null; do
+            if [ $SHUTDOWN_RETRY_COUNT -ge $SHUTDOWN_MAX_RETRIES ]; then
+              echo "Max retries reached. Dev server did not release port 8103."
+              exit 1
+            fi
+
+            SHUTDOWN_RETRY_COUNT=$((SHUTDOWN_RETRY_COUNT + 1))
+            echo "Waiting for port 8103 to free up... (Attempt $SHUTDOWN_RETRY_COUNT/$SHUTDOWN_MAX_RETRIES)"
+            sleep 1
+          done
```
mattwiller: Under what circumstances could this fail?  Shouldn't the `wait` call above ensure that the server process has terminated?

.github/workflows/build.yml
```diff
@@ -589,6 +589,56 @@ jobs:
...
+
+          echo "Port 8103 is free."
+        env:
+          REDIS_PASSWORD_DISABLED_IN_TESTS: 1
```
noahsilas, line 631: #style #nit - because the previous key has such a long body, this `env` key gets a little lost. IMO it would be more readable above the `run` block. Pure style thought, not a blocker.

noahsilas (approved):
Awesome, glad we can detect and prevent this issue from recurring!

.github/workflows/build.yml
The-Alchemist: i learned that too!

.github/workflows/build.yml
The-Alchemist: agree.  done

.github/workflows/build.yml
The-Alchemist: Great question.  I actually removed it and then my LLM warned me so I put it back.

It's purely defensive, maybe too defensive.

This:
```SERVER_PID=$!```

...technically grabs the PID of  the _shell_ that runs`npm` not `node`, and `wait()` only kills the parent.  So `kill` and `wait` are not 100% sufficient to make sure that the port is open and that `npm start` at the next step will succeed. 

I agree with you that it's overkill.  I replaced it with a `kill -9` after the `wait`.

Issue #8075, opened 2025-12-15
executeBatch does not invalidate search cache for affected resource types


benripka:
## Summary

When using `executeBatch()` to create/update/delete resources, the client's internal request cache is not invalidated for the affected resource types. This causes subsequent `search()` calls to return stale data until the cache TTL expires (60s in browser).

## Expected Behavior

`executeBatch()` should invalidate the search cache for all resource types affected by the transaction, consistent with the high-level methods:

```typescript
// These all call invalidateSearches() ✅
await medplum.createResource(patient);
await medplum.updateResource(patient);
await medplum.deleteResource('Patient', id);

// This does not ❌
await medplum.executeBatch(bundle);
```

## Current Behavior

```typescript
const bundle = {
  resourceType: 'Bundle',
  type: 'transaction',
  entry: [{
    resource: { resourceType: 'Appointment', ... },
    request: { method: 'POST', url: 'Appointment' }
  }]
};

await medplum.executeBatch(bundle);

// Returns stale data - the new appointment is missing
const appointments = await medplum.search('Appointment', { date: '2024-01-15' });
```

## Root Cause

In `client.ts`, `executeBatch` is implemented as a simple POST without cache invalidation:

```typescript
executeBatch(bundle: Bundle, options?: MedplumRequestOptions): Promise<Bundle> {
  return this.post(this.fhirBaseUrl, bundle, undefined, options);
  // No invalidateSearches() call
}
```

## Suggested Fix

After the POST succeeds, iterate through the response entries and invalidate searches for each resource type:

```typescript
async executeBatch(bundle: Bundle, options?: MedplumRequestOptions): Promise<Bundle> {
  const result = await this.post(this.fhirBaseUrl, bundle, undefined, options);
  
  // Invalidate cache for all affected resource types
  if (result.entry) {
    const resourceTypes = new Set<ResourceType>();
    for (const entry of result.entry) {
      if (entry.resource?.resourceType) {
        resourceTypes.add(entry.resource.resourceType as ResourceType);
      }
    }
    for (const resourceType of resourceTypes) {
      this.invalidateSearches(resourceType);
    }
  }
  
  return result;
}
```

## Workaround

Manually invalidate after `executeBatch`:

```typescript
const result = await medplum.executeBatch(bundle);
for (const entry of result.entry || []) {
  if (entry.resource) {
    medplum.invalidateSearches(entry.resource.resourceType as ResourceType);
  }
}
```

## Environment

[...]

PR #9859, opened 2026-07-14
fix(core): invalidate search cache after executeBatch resolves

sahilnyk:
## Problem

`executeBatch()` posts the bundle but never invalidates the search cache, unlike `createResource`, `updateResource`, and `deleteResource`. Resources created/updated/deleted via batch stay invisible to `search()` until the cache TTL expires (60s default).

## Fix

After the POST resolves, walk the *request* bundle entries and invalidate the cache only for mutation methods (POST/PUT/PATCH/DELETE)  reads never touch the cache. Resource type is derived from the entry's request URL, supporting both relative (`Patient/123`) and absolute (same-server) URLs.

Follows the approach [suggested in the issue](https://github.com/medplum/medplum/issues/8075) by @benripka.

## Changes

- `packages/core/src/client.ts`: `executeBatch` now async, invalidates cache per mutation entry's resource type
- `packages/core/src/client.test.ts`: tests for mutation-only invalidation, PUT/PATCH/DELETE mutations, and absolute URL handling

## Testing

- [x] Unit tests added: mutation-only invalidation, GET reads excluded, PUT/PATCH/DELETE (incl. no-body DELETE), absolute URL entries
- [x] `client.test.ts` 202/202 pass
- [x] eslint clean


packages/core/src/client.test.ts
```diff
@@ -3118,6 +3118,35 @@ describe('Client', () => {
...
+            },
+          },
+        ],
+      });
+      const client = new MedplumClient({ fetch });
+      const invalidateSpy = vi.spyOn(client, 'invalidateSearches');
+      await client.executeBatch(bundle);
+      expect(invalidateSpy).toHaveBeenCalledWith('Patient');
+      expect(invalidateSpy).toHaveBeenCalledWith('Appointment');
+      invalidateSpy.mockRestore();
```
mattwiller: If one of the `expect()` statements above throws, this cleanup step won't get run

packages/core/src/client.ts
```diff
@@ -2897,8 +2897,20 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
...
-  executeBatch(bundle: Bundle, options?: MedplumRequestOptions): Promise<Bundle> {
-    return this.post(this.fhirBaseUrl, bundle, undefined, options);
+  async executeBatch(bundle: Bundle, options?: MedplumRequestOptions): Promise<Bundle> {
+    const result = await this.post(this.fhirBaseUrl, bundle, undefined, options);
+    if (result.entry) {
+      const resourceTypes = new Set<ResourceType>();
+      for (const entry of result.entry) {
+        if (entry.resource?.resourceType) {
+          resourceTypes.add(entry.resource.resourceType as ResourceType);
+        }
```
mattwiller: This would unconditionally invalidate searches for entire resource types even if the underlying operation was just a read by ID, which would degrade the usefulness of the cache especially when autobatching is turned on.  This should only invalidate the search cache when the corresponding operation is some kind of mutation (i.e. POST/PUT/PATCH/DELETE)

mattwiller (changes requested):

sahilnyk:
@mattwiller Fixed both. Only invalidates now if the entry's method is POST/PUT/PATCH/DELETE, reads leave the cache alone.

While fixing that I noticed DELETE entries have no resource in the response, so the old check missed them entirely, invalidation now reads the resource type off the request URL instead so DELETE gets covered too.

Also dropped mockRestore since the spy's on a local client per test, nothing to clean up.

Whenever you get a chance, please take another look.

PR #9866, opened 2026-07-14
fix: resolve API Extractor ae-unresolved-link warnings

The-Alchemist:
## Summary
- Qualify cross-package TSDoc `@link` targets (e.g. `@medplum/react-hooks#useMedicationIFrame`) so API Extractor can resolve them from DoseSpot/ScriptSure wrappers
- Qualify interface member links in `UseMedicationCartReturn` (fixes cascade warnings in `@medplum/react`)
- Replace `@link` with backticks for private non-exported symbols (`MAX_AUTH_ATTEMPTS`, `RequestState`) in `MedplumClient`

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -4011,8 +4011,8 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
   /**
    * Handles an unauthenticated (HTTP 401) response from the server.
    *
-   * Bounded and terminal: at most {@link MAX_AUTH_ATTEMPTS} attempts per request (1 initial
-   * + 1 recovery, tracked via {@link RequestState.authAttempt}). The recovery re-mints via a
+   * Bounded and terminal: at most `MAX_AUTH_ATTEMPTS` attempts per request (1 initial
+   * + 1 recovery, tracked via `RequestState.authAttempt`). The recovery re-mints via a
    * forced {@link MedplumClient.refresh} (bypassing the {@link MedplumClient.isAuthenticated}
    * short-circuit on the rejected token), single-flight so concurrent 401s share one re-mint.
    * A second 401 is terminal: clear auth, `onUnauthenticated`, reject — never recurse.
--- a/packages/dosespot-react/src/useDoseSpotIFrame.ts
+++ b/packages/dosespot-react/src/useDoseSpotIFrame.ts
@@ -19,11 +19,12 @@ export interface DoseSpotIFrameOptions extends MedicationIFrameOptions {
   readonly onSelfEnrollSuccess?: (result: DoseSpotSelfEnrollmentResult) => void;
 }
 
+/* eslint-disable jsdoc/escape-inline-tags -- TSDoc cross-package {@link @package#symbol} references */
 /**
  * React hook that syncs a patient to DoseSpot and returns the iframe URL.
  *
  * Runs optional self-enrollment, then the patient-sync bot (when `patientId`
- * is set), then the iframe bot — aligned with {@link useMedicationIFrame}
+ * is set), then the iframe bot — aligned with {@link @medplum/react-hooks#useMedicationIFrame}
  * behavior plus DoseSpot-specific enrollment.
  *
  * @param options - Configuration and callback options.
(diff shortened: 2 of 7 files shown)


packages/core/src/client.ts
```diff
@@ -4011,8 +4011,8 @@ export class MedplumClient extends TypedEventTarget<MedplumClientEventMap> {
   /**
    * Handles an unauthenticated (HTTP 401) response from the server.
    *
-   * Bounded and terminal: at most {@link MAX_AUTH_ATTEMPTS} attempts per request (1 initial
-   * + 1 recovery, tracked via {@link RequestState.authAttempt}). The recovery re-mints via a
+   * Bounded and terminal: at most `MAX_AUTH_ATTEMPTS` attempts per request (1 initial
+   * + 1 recovery, tracked via `RequestState.authAttempt`). The recovery re-mints via a
```
mattwiller, line 4015: Why can't these ones use `@link`?

The-Alchemist: Turns out API Extractor only allows _exported_ symbols to use `@link`.  We can export `MAX_AUTH_ATTEMPTS` and/or `RequestState.authAttempt`, but they seem like internal implementations details.  wdyt?

The-Alchemist:
@cursoragent rebase and fix merge conflicts

The-Alchemist:
@cursoragent run 'npm run lint:fix' and fix these errors:


```
/home/runner/work/***/***/packages/dosespot-react/src/useDoseSpotIFrame.ts
Error:   26:1  error  Unexpected inline JSDoc tag. Did you mean to use {@***}, \@***, or ?  jsdoc/escape-inline-tags

/home/runner/work/***/***/packages/dosespot-react/src/useDoseSpotPharmacySearch.ts
Error:   13:1  error  Unexpected inline JSDoc tag. Did you mean to use {@***}, \@***, or ?  jsdoc/escape-inline-tags
```

The-Alchemist:
@claude review

The-Alchemist:
@cursoragent 

 The \@ escaped cross-package links — please verify these actually resolve

Files: useDoseSpotIFrame.ts, useDoseSpotPharmacySearch.ts, useScriptSureIFrame.ts, useScriptSureOrderSet.ts, useScriptSurePharmacySearch.ts

{@link \@medplum/react-hooks#useMedicationIFrame}
The backslash was added to satisfy the ESLint jsdoc/escape-inline-tags rule (from eslint-plugin-jsdoc). My concern: eslint-plugin-jsdoc speaks JSDoc, but API Extractor speaks TSDoc, and the two disagree about this syntax.

The TSDoc-correct form of a cross-package reference is {@link @medplum/react-hooks#useMedicationIFrame} — the leading @ is the scoped-package sigil and is expected by TSDoc's declaration-reference grammar.
A literal backslash (\@medplum/...) is not a valid first character for a TSDoc module source. TSDoc's declaration-reference parser doesn't do CommonMark-style backslash unescaping inside {@link} code destinations, so \@medplum/react-hooks is likely parsed as a malformed reference rather than the package @medplum/react-hooks.
Why green CI doesn't settle this: the build runs api-extractor run --local (see each package's package.json), and ae-unresolved-link is a warning, not an error. In --local mode API Extractor treats warnings as non-fatal (succeeded = errorCount === 0), and the base api-extractor.json doesn't promote it to an error. So the "Build ✅" and the earlier "no ae-unresolved-link warnings" claim are consistent with the link silently not resolving — the whole reason warnings like this were fixable-but-not-blocking in the first place.

Please confirm manually (I couldn't — no node_modules in this sandbox and network installs are blocked):

cd packages/dosespot-react && npm run build 2>&1 | grep -iE 'unresolved|tsdoc|\\bwarn'
Then inspect the generated dist/types.d.ts (or the docs output) to confirm the @link renders as a real link to useMedicationIFrame, not as literal text \@medplum/....

If it turns out the \@ doesn't resolve, the cleaner fix is to keep the TSDoc-correct reference and silence the JSDoc linter locally, e.g.:

/**
 * ...
 * Thin wrapper around the generic {@link @medplum/react-hooks#usePharmacySearch} hook,
 * ...
 */
// eslint-disable-next-line jsdoc/escape-inline-tags
or disable/scope jsdoc/escape-inline-tags for these packages in the shared config, since it's a JSDoc rule being applied to TSDoc content.

can you test and verify?

mattwiller (approved):

PR #9867, opened 2026-07-14
fix(core): use standards-compliant keys() in storage.ts

The-Alchemist:
In `ClientStorage.clear()`, enumerate prefixed `localStorage` keys with the Web Storage API `key(i)` / `length` loop instead of `Object.keys()`, which is not part of the Storage spec and is unreliable across implementations.

Found accidentally because it fails on Node v26.


codyebberson:
@claude please review

codyebberson (approved):

PR #9869, opened 2026-07-14
Don't require a care template when creating an encounter

### LINKED ISSUE #9816: Unclear "Please fill out required fields" error on new patient encounter
In New Encounter we seem to require a template, even though it's not marked as such. This results in a confusing workflow for users who don't have any templates in their system (there is no guidance that one is missing, so the workflow can't be completed).

https://github.com/user-attachments/assets/0db22ea0-f7af-49b8-b792-f9af7aa8e2b1

Filupnot:
Fixes #9816

## Summary
No longer require a Care Template to be filled out when creating an encounter. Encounters can be created from three places: 
- `EncounterModal.tsx` - "new encounter" modal in patient screen
- `CreateVisit.tsx` - schedule "create visit" flow 
- `CreateEncounterForm.tsx` - post-booking "set up encounter" form
  - (Note: This form is tricky to open because it requires an appointment that's not linked to an encounter. All Provider workflows create an Appointment and Encounter together.) 

We'll also update the starting page to remove the blurb about a care template being "required".

### DIFF
--- a/examples/medplum-provider/src/components/schedule/CreateEncounterForm.tsx
+++ b/examples/medplum-provider/src/components/schedule/CreateEncounterForm.tsx
@@ -39,7 +39,7 @@ export function CreateEncounterForm(props: CreateEncounterFormProps): JSX.Elemen
   );
 
   const handleSubmit = useCallback(async () => {
-    if (!encounterClass || !planDefinition) {
+    if (!encounterClass) {
       showNotification({
         color: 'yellow',
         icon: <IconAlertSquareRounded />,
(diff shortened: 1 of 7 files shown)


Filupnot:
@claude review this

examples/medplum-provider/src/components/schedule/CreateEncounterForm.tsx
```diff
@@ -92,12 +92,11 @@ export function CreateEncounterForm(props: CreateEncounterFormProps): JSX.Elemen
...
           label="Care template"
           onChange={setPlanDefinition}
-          required={true}
         />
 
         <PlanDefinitionSummary planDefinition={planDefinition} />
 
-        <Button fullWidth type="submit" disabled={!planDefinition || !encounterClass}>
+        <Button fullWidth type="submit" disabled={!encounterClass}>
           Apply
```
techdavidy: Since we are not applying anymore. 
Save, Continue, Next, etc  ?

Filupnot: went with "continue" - that seems to make the most sense here.

techdavidy (approved):

Issue #7448, opened 2025-09-23
Create a welcome email for newly registered projects


reshmakh:
Include server settings for email aliases and copy

PR #9871, opened 2026-07-14
Send welcome email to newly registered projects

maddyli:
Addresses https://github.com/medplum/medplum/issues/7448

Screenshot 2026-07-28 at 3 24 47 PM


packages/server/package.json
```diff
@@ -88,6 +88,7 @@
     "ipaddr.js": "2.4.0",
     "jose": "6.2.3",
     "jszip": "3.10.1",
+    "marked": "18.0.5",
```
maddyli: Added `marked` which is a docs dependency, but unsure if we really need. Alternative options could be to add a lightweight parsing function in welcomeemail.ts, or to just have plain HTML that we have to hand-edit. Open to any opinions here, with my goal of keeping the email lightweight but well-rendered and easy to update if needed

The-Alchemist: I would suggest removing the `marked` dependency and just doing plain ol' text like the rest of the system emails like `admin/invite.ts`, `auth/utils.ts`, or `fhir/operations/update-user-email.ts`.

packages/server/src/auth/welcomeemail.ts
```diff
@@ -0,0 +1,178 @@
...
+// ---------------------------------------------------------------------------
+// Rendering & delivery — you should rarely need to edit below this line.
+// ---------------------------------------------------------------------------
+
+/**
+ * Escapes the HTML-significant characters so free-form values can't inject markup.
+ * @param value - The raw string to escape.
+ * @returns The HTML-escaped string.
+ */
+function escapeHtml(value: string): string {
```
mattlong: Echoing @The-Alchemist , agree that just sticking to plain text is preferable since we don't have a repeatable email framework to ensure proper html escaping/sanitizing, etc.

mattlong (changes requested):

packages/server/src/auth/welcomeemail.ts
maddyli: Makes sense. That being said, will wait on a review of the copy/format first. My guess is that going forward, we will probably have less technical folks making changes to things like the welcome email and also want to format it more, so it could be worth the investment?

mattlong (approved):

PR #9872, opened 2026-07-14
[core]: Improve EventTarget listener exception safety 

noahsilas:
When an EventTarget listener throws an error, it should not prevent the EventTarget from invoking other listeners registered for the same event type.

In this commit we wrap event listener invocation with a `try/catch`, so that each listener is more isolated from failures in others.

Breaking change
---------------

Previously, errors in event listeners would be synchronously thrown; they will now be caught and logged to `console.error`. It is recommended that event listener authors wrap their listener code in a `try`/`catch` block and perform their own error handling. This allows the application to use its own context to take the most appropriate action in response to the exception.

### DIFF
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -4907,6 +4907,13 @@ describe('Client', () => {
       expect(fetch).toHaveBeenCalledTimes(2);
     });
   });
+
+  test('setEventListenerErrorHandler refines the type based on client-supported event types', () => {
+    const client = new MedplumClient();
+    client.setEventListenerErrorHandler((error, event) => {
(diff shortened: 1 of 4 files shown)


noahsilas:
@claude please review

codyebberson:
@claude - please re-review

1. Please document the specific breaking changes
2. Do we actually need the `setEventListenerErrorHandler()` helper?  I understand the argument that it could be considered necessary for preserving legacy behavior, but it's not clear that that is desirable or worth preserving.

ThatOneBro (approved):

codyebberson:
@claude please re-review - are all open items resolved?

PR #9873, opened 2026-07-15
Add toHaveStatus matcher

mattlong:
We should start adopting this when checking response status so that the body of the response will be shown to help track down flaky tests.

### DIFF
--- a/packages/server/src/test-matchers.ts
+++ b/packages/server/src/test-matchers.ts
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
+import type { Response } from 'supertest';
 import { expect } from 'vitest';
 
 interface CustomMatchers<R = unknown> {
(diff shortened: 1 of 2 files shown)


The-Alchemist (approved):

mattwiller (approved):

packages/server/src/test-matchers.ts
```diff
@@ -56,4 +63,24 @@ expect.extend({
...
+    }
+    let bodyStr: string;
+    try {
+      bodyStr = JSON.stringify(received.body, null, 2);
+    } catch {
+      bodyStr = received.text ?? '(empty)';
+    }
+    return {
+      pass: false,
+      message: () => `Expected status ${expected}, received ${received.status}\n\nResponse body:\n${bodyStr}`,
```
noahsilas, line 83: Nice! I had taken to writing tests in a slightly unintuitive order to get behavior similar to what you achieve:
```ts
expect(response.body).not.toHaveProperty('issue');
expect(response.status).toBe(200);
```

I like your approach much better!

noahsilas (approved):

PR #9874, opened 2026-07-15
Align patient matching docs with match operation

codyebberson:
The "Built-in matching" tip on the patient deduplication Matching (`packages/docs/docs/fhir-datastore/patient-deduplication/matching.mdx`) page described the Patient/$match operation in a way that contradicted its own technical reference (`packages/docs/docs/api/fhir/operations/patient-match.md`):

- It said matching used weighted demographics, but the operation is based on the CMS Patient Matching framework — the reference explicitly states it is not an ad-hoc weighted heuristic and the score is not a probability.
- It listed gender as a matching factor, but the reference explicitly notes gender is not used.

This updates the tip to describe the CMS-framework approach and the correct set of identity factors, so the two pages are consistent.

No behavior change — documentation only.

Before:

image

After:

image


The-Alchemist:
@claude review

The-Alchemist (approved):

PR #9877, opened 2026-07-15
Use new expect().toHaveStatus() matcher in tests

mattwiller:
Follow-up to #9873 

```
s/\.status\)\.to[a-zA-Z]+\((\d)/).toHaveStatus($1/g
```

### DIFF
--- a/packages/server/src/fhir/operations/book.test.ts
+++ b/packages/server/src/fhir/operations/book.test.ts
@@ -183,7 +183,7 @@ describe('Appointment/$book', () => {
         ],
       });
 
-    expect(response.status).toEqual(400);
+    expect(response).toHaveStatus(400);
     expect(response.body).toHaveProperty('issue', [
       {
         code: 'invalid',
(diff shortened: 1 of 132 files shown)


packages/server/src/fhir/operations/book.test.ts
```diff
@@ -242,7 +242,7 @@ describe('Appointment/$book', () => {
       });
 
     expect(response.body).not.toHaveProperty('issue');
-    expect(response.status).toEqual(201);
+    expect(response).toHaveStatus(201);
```
noahsilas: This is a pattern I used a lot to get at some of the value of the `toHaveStatus` matcher: in the "OK" scenarios, I first tested that `response.body.issue` did not exist, which had the side effect of emitting a helpful error message saying what was present in `issue` when the request wasn't successful.

Overall, I think that the better thing here is now to drop that first assertion, as the full response body (including `issue`) will be in this assertion's output.

Would you mind if I push a commit onto this branch that drops those leading `issue` assertions?

noahsilas (approved):
🎉

PR #9883, opened 2026-07-15
Gate Provider workflows on missing integration dependencies

### LINKED ISSUE #9824: Medplum Provider should surface missing shared-project dependencies upfront

Filupnot:
Fixes #9824 

## Problem
When a project is missing a dependency that a workflow needs, users hit broken flows and a wall of errors they can't act on. We shouldn't let users into a workflow they have no way to complete, and we shouldn't surface errors with no actionable fix.

## What this does
- **Blocks the Order Labs workflow upfront when the Health Gorilla bot isn't linked**, so users never enter a flow that can't complete (this also stops `useHealthGorillaLabOrder` from mounting and throwing). The message is role-aware:
  - **Admins** see which integration to link, with a docs link.
  - **End users** are told the workflow isn't set up yet and to contact their administrator.

  image
  

- **Adds an admin-only banner on the Get Started page** summarizing every missing dependency, so admins can fix them up front instead of discovering them one blocked workflow at a time.

  

- **Adds reusable plumbing** so future bot-dependent workflows can be gated the same way with a single manifest entry:
  - `dependencies.ts` - maps each gated workflow to its bot dependencies.
  - `useWorkflowAvailability` - probes each dependency via a cheap `Bot?identifier=…` search (cached per session; the bot is never executed).
  - `<WorkflowGate>` - wraps a workflow's entry point and renders the role-aware guidance above when a dependency is missing. `OrderLabsPage` is wrapped once, covering all three places it's rendered.

- **Makes the missing-profile error on resource creation actionable.** Creating a resource whose type requires a FHIR profile (e.g. a Patient requiring `us-core-patient`) previously failed with a raw `StructureDefinition profile ... not found` server error on the create page. It now shows the **same role-aware card as the workflow gate** for visual consistency: admins see which profile to import plus a link to the profiles docs; other users are directed to an administrator. `ResourceFormWithRequiredProfile` already blocked submission in this case; this only replaces the raw error with guidance, and callers that pass no message (the resource edit pages) keep the existing fallback.
image

- **Adds a link to the Stedi insurance-eligibility docs** from the existing eligibility upsell.

  

## Why only *some* things are gated
There are two kinds of integration-backed workflows, and they're handled differently on purpose:
- **Core** - baked into the app; users expect them to work (e.g. ordering labs via Health Gorilla). These are gated upfront (banner + `WorkflowGate`).
- **Isolated** - exist *only* because an integration is present (e.g. Stedi eligibility, DoseSpot, ScriptSure). A missing optional integration isn't a broken core workflow, so these are intentionally left out of the manifest and banner and keep their existing in-context handling (eligibility keeps its "Contact Support" upsell, now with the Stedi docs link; DoseSpot/ScriptSure hide their entry points).

## What about ValueSet dependencies?
Out of scope: an audit of every coded field in Provider found none with a hard ValueSet dependency (no field passes `creatable={false}`; they all accept custom values), so a missing terminology ValueSet degrades gracefully rather than breaking a flow.

### DIFF
--- a/examples/medplum-provider/src/components/UnavailableNotice.tsx
+++ b/examples/medplum-provider/src/components/UnavailableNotice.tsx
@@ -1,29 +1,39 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Center, Paper, Stack, Title } from '@mantine/core';
+import { Center, Stack, Text, Title } from '@mantine/core';
 import type { JSX, ReactNode } from 'react';
 
 export interface UnavailableNoticeProps {
   /** Decorative icon shown above the title. Callers should mark it `aria-hidden`. */
   readonly icon: ReactNode;
-  readonly title: string;
-  readonly children: ReactNode;
+  readonly title: ReactNode;
+  /** Dimmed prose shown under the title. */
+  readonly description?: ReactNode;
+  /** Anything that follows the description — what is missing, where to go next. */
+  readonly children?: ReactNode;
 }
 
-// Shared presentational card explaining, in a role-aware way, why a workflow or page is
+// Shared presentational notice explaining, in a role-aware way, why a workflow or page is
 // unavailable (a missing integration, an uninstalled profile, etc.). Centralizing the shell keeps
 // these notices visually consistent instead of relying on hand-synced copies.
+//
+// The props and slot order deliberately mirror Mantine's `EmptyState` (icon, title, description,
+// then children; no card container of its own) so that this collapses into that component when we
+// move to Mantine >= 9.4.
 export function UnavailableNotice(props: UnavailableNoticeProps): JSX.Element {
-  const { icon, title, children } = props;
+  const { icon, title, description, children } = props;
   return (
     <Center p="xl">
-      <Paper shadow="md" p="xl" radius="md" withBorder maw={480}>
-        <Stack align="center" gap="sm" ta="center">
-          {icon}
-          <Title order={3}>{title}</Title>
-          {children}
-        </Stack>
-      </Paper>
+      <Stack align="center" gap="sm" ta="center" maw={480}>
+        {icon}
+        <Title order={3}>{title}</Title>
+        {description && (
+          <Text size="sm" c="dimmed">
+            {description}
+          </Text>
+        )}
+        {children}
+      </Stack>
     </Center>
   );
 }
--- a/examples/medplum-provider/src/components/UnavailableNotice.tsx
+++ b/examples/medplum-provider/src/components/UnavailableNotice.tsx
@@ -1,0 +1,29 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import { Center, Paper, Stack, Title } from '@mantine/core';
+import type { JSX, ReactNode } from 'react';
+
+export interface UnavailableNoticeProps {
+  /** Decorative icon shown above the title. Callers should mark it `aria-hidden`. */
+  readonly icon: ReactNode;
+  readonly title: string;
+  readonly children: ReactNode;
+}
+
+// Shared presentational card explaining, in a role-aware way, why a workflow or page is
+// unavailable (a missing integration, an uninstalled profile, etc.). Centralizing the shell keeps
+// these notices visually consistent instead of relying on hand-synced copies.
+export function UnavailableNotice(props: UnavailableNoticeProps): JSX.Element {
+  const { icon, title, children } = props;
+  return (
+    <Center p="xl">
+      <Paper shadow="md" p="xl" radius="md" withBorder maw={480}>
+        <Stack align="center" gap="sm" ta="center">
+          {icon}
+          <Title order={3}>{title}</Title>
+          {children}
+        </Stack>
+      </Paper>
+    </Center>
+  );
+}
--- a/examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
+++ b/examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import {
+  Anchor,
   Badge,
   Box,
   Button,
@@ -265,9 +266,21 @@ function CoverageCard(props: CoverageCardProps): JSX.Element {
           <Box pt="md">
             {!benefitsLoading && !eligibilityResponse && (
               <Text size="sm" c="dimmed">
-                {eligibilityBot
-                  ? 'No eligibility check found. Click "Check Eligibility" to run a check.'
-                  : 'No eligibility check found. Contact support to enable eligibility checks.'}
+                {eligibilityBot ? (
+                  'No eligibility check found. Click "Check Eligibility" to run a check.'
+                ) : (
+                  <>
+                    No eligibility check found. Contact support to enable eligibility checks, or{' '}
+                    <Anchor
+                      href="https://www.medplum.com/docs/integration/stedi/insurance-eligibility/eligibility-checks"
+                      target="_blank"
+                      rel="noreferrer"
+                    >
+                      learn about insurance eligibility
+                    </Anchor>
+                    .
+                  </>
+                )}
               </Text>
             )}
             {!benefitsLoading && eligibilityResponse?.outcome === 'error' && (
--- a/examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
+++ b/examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
@@ -1,7 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
 import {
-  Anchor,
   Badge,
   Box,
   Button,
@@ -36,6 +35,7 @@ import type { JSX } from 'react';
 import { useCallback, useEffect, useState } from 'react';
 import { isSelfPayCoverage } from '../../utils/coverage';
 import { showErrorNotification } from '../../utils/notifications';
+import { DocsLink } from '../DocsLink';
 import { BenefitsTable } from '../insurance/BenefitsTable';
 
 interface EncounterCoverageEligibilityModalProps {
@@ -271,13 +271,9 @@ function CoverageCard(props: CoverageCardProps): JSX.Element {
                 ) : (
                   <>
                     No eligibility check found. Contact support to enable eligibility checks, or{' '}
-                    <Anchor
-                      href="https://www.medplum.com/docs/integration/stedi/insurance-eligibility/eligibility-checks"
-                      target="_blank"
-                      rel="noreferrer"
-                    >
+                    <DocsLink path="integration/stedi/insurance-eligibility/eligibility-checks">
                       learn about insurance eligibility
-                    </Anchor>
+                    </DocsLink>
                     .
                   </>
                 )}
--- a/examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
+++ b/examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
@@ -40,6 +40,7 @@ import orderSetBundleData from '../../data/order-set-example-bundle.json';
 import patientBundleData from '../../data/patient-david-james-williams.json';
 import visitBundleData from '../../data/simple-initial-visit-bundle.json';
 import { showErrorNotification } from '../../utils/notifications';
+import { WorkflowDependenciesPanel } from '../../workflow/WorkflowDependenciesPanel';
 import classes from './GetStartedPage.module.css';
 import { buildOrderSetImportNotification } from './orderSetImportNotification';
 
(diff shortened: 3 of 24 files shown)


Filupnot:
@claude review please

examples/medplum-provider/src/workflow/useWorkflowAvailability.ts
```diff
@@ -0,0 +1,165 @@
...
+ * the bot most likely can't complete the workflow either.
+ * @param medplum - The Medplum client used to look up bots by identifier.
+ * @param dependencies - The dependencies to probe.
+ * @returns The subset of `dependencies` confirmed missing.
+ */
+async function findMissingDependencies(
+  medplum: MedplumClient,
+  dependencies: readonly WorkflowDependency[]
+): Promise<WorkflowDependency[]> {
+  const results = await Promise.all(
```
techdavidy: I think we should use Promise.allSettled. 
With Promise.all if one fails the rest would fail.

Filupnot: Since we had the try/catch swallowing errors, I don't think the individual probes should never actually reject. But I agree `allSettled` is a bit cleaner: it lets the concurrent call be a plain one-liner and moves the "transient failure = treat as present" logic into one spot. Updated.

noahsilas:
It looks like this is built with some of the same commits as #9819, which has had additional work added and merged. There are now some conflicts, and a bunch of extra context in the diff on this PR. Could you rebase this on top of the latest `main` to make it easier to review? Thanks!

Filupnot:
@noahsilas - Thanks for calling that out. Should be cleaned up now!

examples/medplum-provider/src/components/encounter/EncounterCoverageEligibilityModal.tsx
```diff
@@ -265,9 +266,21 @@ function CoverageCard(props: CoverageCardProps): JSX.Element {
...
+                {eligibilityBot ? (
+                  'No eligibility check found. Click "Check Eligibility" to run a check.'
+                ) : (
+                  <>
+                    No eligibility check found. Contact support to enable eligibility checks, or{' '}
+                    <Anchor
+                      href="https://www.medplum.com/docs/integration/stedi/insurance-eligibility/eligibility-checks"
+                      target="_blank"
+                      rel="noreferrer"
+                    >
```
noahsilas: Consider using `<DocsLink path="integration/stedi/insurance-eligibility/eligibility-checks">` - today this has much the same effect, but I hope to eventually build out some static typing to check that docs paths in Provider stay in sync with pages in the docs package. (Imagine renaming a docs page and getting a typescript error at build time instead of a silently broken link!)

packages/react/src/UnavailableNote/UnavailableNote.tsx
```diff
@@ -4,13 +4,25 @@ import { ActionIcon, Text, Tooltip } from '@mantine/core';
...
+
+const SEVERITY_COLORS: Record<UnavailableNoteSeverity, string> = {
+  warning: 'yellow.9',
+  error: 'red',
+};
+
 export interface UnavailableNoteProps {
   readonly text: string;
-  readonly color: string;
+  readonly severity: UnavailableNoteSeverity;
```
noahsilas, line 20: This change in props is a breaking change; I think it's unlikely that anyone outside the monorepo is depending on this (particularly since it was only added 1 week ago), but since this is a published package (the current API was tagged in v5.1.27) we don't really know that for sure.

IMO it's OK to make this change (brand new component getting tweaked), but if you want to be maximally safe you could omit this change and we could schedule a future breaking change.

I wonder if we want to align this shape with `OperationOutcomeIssue.severity` - the current values are a subset, which makes it fairly easy to expand support to also include "fatal", "information", and "success" later if desired, probably fine to support only this subset for now.

packages/react-hooks/src/useValueSetAvailability/useValueSetAvailability.ts
```diff
@@ -69,6 +69,9 @@ export function useValueSetAvailabilities(urls: readonly (string | undefined)[])
     }
 
     let cancelled = false;
+    // The abort is load-bearing beyond cancelling the request: `MedplumClient` caches rejections
+    // and only evicts an entry when its request is aborted, so aborting on unmount is what lets a
+    // later mount re-probe a value set that has since been imported.
```
noahsilas: Ah, right; when the abort is triggered the cache is cleared even if the request already completed. I think this might not be fully designed/intended, but rather an outcome of the particular implementation in the client.

I think a downside in this case is that we do want a success to stay cached even after this component unmounts, which suggests to me that we need to guard the `abort` in the cleanup function, or teach MedplumClient abort handling not to clear completed results.

If the intent here is to skip reading cache, we should probably use the more explicit `cache: 'reload'` option in the `valueSetExpand()` call instead of relying on "abort" to clear the cache.

examples/medplum-provider/src/workflow/WorkflowGate.tsx
```diff
@@ -0,0 +1,79 @@
...
+      icon={<IconPlugConnectedX size={48} color="var(--mantine-color-gray-5)" aria-hidden />}
+      title={`${workflowLabel} is unavailable`}
+    >
+      <Text size="sm" c="dimmed">
+        This workflow depends on the following {plural ? 'integrations that are' : 'integration that is'} not linked to
+        your project:
+      </Text>
+      {dependencyList}
+      <Text size="sm" c="dimmed">
+        Link the required project{plural ? 's' : ''} to enable this workflow.
```
noahsilas: Total #nit - this style of "append an 's'" pluralization works for english but makes future I18n work a little less clear. If you can keep the entire word together its a little more repetitive but more amenable to future I18n tooling.

```suggestion
        Link the required {plural ? 'projects' : 'project'} to enable this workflow.
```

examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
```diff
@@ -62,6 +63,14 @@ export interface OrderLabsPageProps {
 }
 
 export function OrderLabsPage(props: OrderLabsPageProps): JSX.Element {
+  return (
+    <WorkflowGate workflow="order-labs">
```
noahsilas, line 67: It feels to me like this is hiding some details specific to this call site a few layers down the call stack (WorkflowGate -> useWorkflowAvailability -> dependencies.ts).

Providing this "workflow ID" string makes some of the useEffect work nice and tidy inside these elements, but I think it is inverting some of the data in a way that feels awkward to me.

What do you think of going the opposite way and lifting that definition from `dependencies` up to this call site?

```tsx
  return (
    <WorkflowGate
      dependencies={[
        {
          kind: 'bot',
          identifier: 'https://www.medplum.com/integrations/bot-identifier|health-gorilla-labs/send-to-health-gorilla',
          label: 'Health Gorilla lab ordering',
          docsUrl: 'https://www.medplum.com/docs/integration/health-gorilla',
        },
      }]
  >
    ...
  </WorkflowGate>
```

examples/medplum-provider/src/components/UnavailableNotice.tsx
```diff
@@ -0,0 +1,29 @@
...
+  /** Decorative icon shown above the title. Callers should mark it `aria-hidden`. */
+  readonly icon: ReactNode;
+  readonly title: string;
+  readonly children: ReactNode;
+}
+
+// Shared presentational card explaining, in a role-aware way, why a workflow or page is
+// unavailable (a missing integration, an uninstalled profile, etc.). Centralizing the shell keeps
+// these notices visually consistent instead of relying on hand-synced copies.
+export function UnavailableNotice(props: UnavailableNoticeProps): JSX.Element {
```
noahsilas, line 16: There doesn't seem to be anything here that is specific to an "Unavailable Notice" - I wonder if there is a more general name we can give this layout component so that we can use the styling in more places. For example, I can imagine some empty-states getting a similar shape with some instructions. (cc @kevinwadeshaw to weigh in on if any of that is actually desirable or if this should stay contained to this semantic use case)

examples/medplum-provider/src/pages/getstarted/GetStartedPage.tsx
```diff
@@ -157,33 +158,38 @@ export function GetStartedPage(): JSX.Element {
...
-            or{' '}
-            <Text component="a" href="mailto:support@medplum.com" c="blue.6" className={classes.link} span>
-              contact us
-            </Text>{' '}
-            to integrate these services.
-          </Text>
-        </Box>
+        {/* Header, with the admin-only missing-dependencies banner above it. The banner lives in a
+            tight group so it never inherits the 6rem section rhythm; when it renders null (non-admin
+            or nothing missing) the Stack gap simply collapses, leaving the header spacing unchanged. */}
```
noahsilas: I feel like this is just describing the behavior of `<Stack>`, can we drop this comment?

packages/react/src/UnavailableNote/UnavailableNote.tsx
Filupnot: Good callout. I decided to keep the change here with the assumption that it's new enough to be OK to tweak this early on. I made it so that it can be `'warning' | 'error'`, which is a subset of the options in `OperationOutcomeIssue.severity`, so if we decide to change that later to match then that should not be a breaking change.

examples/medplum-provider/src/components/UnavailableNotice.tsx
Filupnot: @kevinwadeshaw For context this is what the three styles of this look like. Do we think we'd want to re-use this component in other contexts outside this PR? IMO it's pretty specific to this PR. 

image

examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
Filupnot: The tricky part about your suggestion is that then we wouldn't be able to have the analyst view in the Getting Started page telling them of all the dependencies that are missing upfront- to get that we need to have a manifest of all dependencies in one place and not just at the call sites. 

An alternative that might address your concern: replace the string key with a reference to the definition itself, like `<WorkflowGate workflow={WORKFLOWS['order-labs']}>`, typing `WorkflowGateProps.workflow` as `WorkflowDefinition` instead of `WorkflowId` so that intellisense will surface it at the call site. Let me know what you think of this.

packages/react-hooks/src/useValueSetAvailability/useValueSetAvailability.ts
Filupnot: I deliberated on this a fair bit, and decided to remove `abort` entirely.

The comment described the abort as load-bearing: it's what lets a later mount re-probe a value set that was previously missing and has since been imported. But the request cache TTL already handles that (60 seconds in Provider), so we'd be re-probing after a minute regardless. Removing it also means a success now stays cached after unmount, which is the behavior you were after.

The biggest win we're getting out of this cache is for situations where a form has multiple fields bound to the same valueset. I didn't go for `reload` because that would skip the cache read in this situation.

To get what you're really describing (remembering successes and expiring failures), we'd need our own cache above the client's. The request cache can't express this because the entry gets created before the request settles, so the TTL would apply equally between successful and failed probes. For the sake of keeping this PR scoped, I decided to not do this here, but I'm happy to open a separate issue if you think this would be a good idea in the future to reduce network calls.

examples/medplum-provider/src/components/UnavailableNotice.tsx
kevinwadeshaw: Good question! At some point we should definitely use a shared component for both these/future notices and our empty states. For now though, I think we can keep as-is since there aren't many instances. Also, we'll likely upgrade to Mantine 9.0 in the next ~six months and can just use their new [EmptyState](https://mantine.dev/core/empty-state/) component then 🙃

For this PR, my only near-term design recommendation would be to remove from the container/paper element and then also remove any other page elements where possible (like the blurred UI we can see behind the notice in Spaces):

Screenshot 2026-08-11 at 10 07 55 PM 

–


These near-term changes would create a foundation for adding future visuals/illustrations/CTAs; ultimately, I think notices should be mini 'ads' for these these features/integrations and include a clear next step for accessing them (including self-serve options once we've shipped marketplace). Some Claude examples:

Screenshot 2026-08-11 at 10 53 10 PM

Screenshot 2026-08-11 at 10 50 10 PM

Filupnot: Thanks for the detailed response! Kept location as-is but updated props to closer match what the new Mantine component has, and removed the Paper. I kept the blur because it was not touched by this PR.

examples/medplum-provider/src/pages/labs/OrderLabsPage.tsx
Filupnot: I implemented my alternative - let me know what you think.

PR #8530, opened 2026-02-25
Add ConceptMap Mappings tab with grouped sections table

maddyli:
## Summary

- Adds a **Mappings** tab to the ConceptMap resource page in the Medplum app
- Displays `ConceptMap.group[]` as grouped 2D table sections, one per source→target system pair
- Each section shows source/target system URIs as a header, and a table with columns: Source Code, Source Display, Equivalence, Target Code, Target Display, Comment
- Handles 1:many mappings via `rowSpan` on source cells
- Shows `group.unmapped` fallback rules as a note below each group table
- Shows an empty state when no groups are defined

## New files

- `ConceptMapMappingsDisplay.tsx` — main display component
- `ConceptMapMappingsDisplay.module.css` — scoped styles (Mantine variables, light/dark mode)
- `MappingsPage.tsx` — route handler (mirrors `PreviewPage` pattern)

## Modified files

- `ResourcePage.tsx` — adds `'Mappings'` to `getTabs()` for `ConceptMap`
- `AppRoutes.tsx` — registers the `mappings` route

## Test plan

- [x] 23 tests across 2 test files (`ConceptMapMappingsDisplay.test.tsx`, `MappingsPage.test.tsx`)
- [x] `ConceptMapMappingsDisplay.tsx`: 100% statements, 95.45% branches, 100% functions, 100% lines
- [x] `MappingsPage.tsx`: 100% across all metrics
- [x] TypeScript type check passes (`tsc --noEmit`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9884, opened 2026-07-15
feat(react): ConceptMap builder UI

### LINKED ISSUE #7567: Create ConceptMap Build UI
Similar to how PlanDefinition has a Builder tab, create a ConceptMap Builder tab. 

This tab should have a table with the following features: 
- A left side of the table where you can view/select the source system and code
- A right side of the table where you can view/select the corresponding target systems and codes
- Ability to repeat this component to edit/create multiple mappings left-to-right
- Other data (name, description, url, etc) stored in normal details tab and not editable via the tab/component in question, much like the PlanDefinition builder

everett-williams:
## Summary

Adds a **Builder** tab for `ConceptMap`, for visually authoring `group[] / element[] / target[]`
mappings (e.g. SNOMED CT → ICD-10). Implements #7567.

Small maps get an editable row grid; maps above 10,000 source codes get a read-only mappings
table instead, reviving the design from the closed #8530.

## What's in it

**Editable builder** (`packages/react/src/ConceptMapBuilder/ConceptMapBuilder.tsx`)
- Per-group source/target `CodeSystem` pickers, prefilled from the group's stored canonical URL,
  with a raw-URL fallback for systems that aren't loaded as resources
- Bound source/target code pickers via `CodingInput`, using the resolved `CodeSystem.valueSet`
- Required, grouped equivalence `Select` (all 10 R4 values), plus explicit no-map
  (`equivalence: unmatched`), visually distinct from not-yet-mapped
- Search over source and target codes, displays, and comments; filter by mapped / unmapped /
  no-map; per-group coverage count
- 200-row display cap. The cap never touches state, so large maps round-trip losslessly.

**Read-only mappings table** (`ConceptMapMappingsTable.tsx`)
- One table per source→target system pair; separate source code / display columns; `rowSpan` so a
  1:many mapping reads as one logical row; `group.unmapped` fallback rule; 1,000-row cap per group
- Exported separately, so it can back a standalone view later

**App wiring**: `ConceptMap` Builder tab (`ResourcePage`) + `BuilderPage` case.

## Notes for reviewers

- **Comments are required for `narrower`/`inexact`.** FHIR invariant `cmd-1` mandates them and the
  server rejects the save, so the form blocks it up front. Two existing test fixtures were invalid
  under `cmd-1` and are corrected here.
- **Search exists because find-in-page is misleading.** Both views cap rendered rows, so Ctrl-F
  silently misses matches. Search runs over the full `element[]`, before the cap.
- **Maps over the edit limit are not editable via JSON either** — the server rejects bodies over
  `maxJsonSize` (1 MB default). The read-only banner points at `ConceptMap/$import` instead.
- **`$import` caveat.** Mappings loaded with `ConceptMap/$import` live outside the resource, so
  this tab can't see them, and saving any inline mapping makes the server re-derive its mapping
  table from `group[]` alone — deleting them. There's no API to detect imported mappings, so the
  builder warns when an edit is pending on a map that arrived with no inline groups. That's a
  mitigation, not a fix; the underlying behaviour needs provenance on the server's mapping rows.

## Test plan

- [x] 45 tests across `ConceptMapBuilder.test.tsx` and `ConceptMapMappingsTable.test.tsx`
- [x] `tsc --noEmit` clean for `@medplum/react` and `@medplum/app`; lint and prettier clean
- [x] Exercised against a local server with generated maps at 7, 5, 350, and 10,050 elements,
      covering the render cap, the read-only threshold, and lossless save above the cap


techdavidy:
@claude review please

packages/react/src/ConceptMapBuilder/ConceptMapBuilder.tsx
```diff
@@ -0,0 +1,1102 @@
...
+  const [selectedKey, setSelectedKey] = useState<string>();
+  const [hoverKey, setHoverKey] = useState<string>();
+  const [value, setValue] = useState<ConceptMap>();
+  const [error, setError] = useState<string>();
+
+  useEffect(() => {
+    medplum
+      .requestSchema('ConceptMap')
+      .then(() => setSchemaLoaded(true))
+      .catch(console.log);
```
techdavidy, line 87: show error notification here

techdavidy:
Screenshot 2026-07-24 at 5 24 16 PM

let's get some designer feedback. 
Maybe instead of having source and target side to side they can be stacked.

PR #9886, opened 2026-07-16
[Provider]: Add availability overlay during scheduling

noahsilas:
- Hoist the selected HealthcareService state up to `<ScheduleDetails>`, so that it can extract availability information and pass it into the Calendar widget

- Update the Calendar widget to highlight the available time (via the FullCalendar "business hours" feature)

- We stop auto-selecting a healthcare-service when there is only one, as it makes the availability highlight a bit confusing and not dismiss-able. Now you have to explicitly click in to the service to see the highlight appear.

## Implementation Notes
- A Schedule may have may Scheduling Parameter extensions, which may each have many `availability` sub-extensions, which may each have many `availableTime` sub-sub-extensions. This makes some of the data handling work here tricky.
- We support `availableStartTime` + `availableEndTime` or `allDay` options within the `availableTime` entries
- If the end time is less than the start time, we consider it to wrap to the next day. Example: "22:00:00 - 06:00:00" is an eight hour shift from 10pm - 6am the following day.

## Screenshots

The screenshots are slightly out of date - they include some footer information that showed the source of scheduling information. I've removed that to keep this PR more tightly scoped, and intend to follow up on revealing scheduling configuration more thoughtfully later.

- A schedule type with no availability info set is 24/7 available
Screenshot 2026-07-15 at 7 01 05 PM

- HealthcareService availability can be highlighted
Screenshot 2026-07-15 at 7 02 03 PM

- Per-Schedule overrides of the availability take precedence over HealthcareService availability
Screenshot 2026-07-15 at 7 03 26 PM

### DIFF
--- a/examples/medplum-provider/src/components/schedule/FindPane.test.tsx
+++ b/examples/medplum-provider/src/components/schedule/FindPane.test.tsx
@@ -268,36 +268,6 @@ describe('FindPane', () => {
     });
   });
 
-  describe('Auto-Selection with Single Service', () => {
-    test('auto-selects when there is exactly one schedulable service', async () => {
-      const schedule = createScheduleWithServices([healthcareService]);
-
-      await act(async () => {
-        setup({ schedule });
-      });
-
-      // Should immediately show the service type name, not the selection UI
-      expect(screen.getByText('Annual Checkup')).toBeInTheDocument();
-      expect(screen.queryByText('Schedule…')).not.toBeInTheDocument();
-
-      // Should fetch slots automatically
-      expect(medplum.get).toHaveBeenCalled();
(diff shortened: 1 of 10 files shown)


noahsilas:
I haven't polished and fully tested this yet, putting it up for @finnbergquist and @AndyStoneman as a reference implementation for now.

I'm not sure if the "Details" expando in the find-pane footer is the right place for exposing that information, but it made it much easier for me to spot-check while I was debugging, so I left it in for now. I have a separate branch working on a more full UI treatment of all the scheduling parameters but that one is even less prepped than this one.

mattlong (approved):

PR #9889, opened 2026-07-16
seed.test.ts improvements

mattlong:
Most notably:

* reduce calls of seedDatabase() by one
* flush cache Redis

### DIFF
--- a/packages/server/src/seed.test.ts
+++ b/packages/server/src/seed.test.ts
@@ -20,7 +20,7 @@ import {
   preparePostDeployMigrationAsyncJob,
 } from './migrations/migration-utils';
 import { getLatestPostDeployMigrationVersion, MigrationVersion } from './migrations/migration-versions';
-import { getCacheRedis } from './redis';
+import { closeRedis, getCacheRedis, initRedis } from './redis';
 import * as seedModule from './seed';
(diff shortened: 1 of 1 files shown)


packages/server/src/seed.test.ts
```diff
@@ -53,55 +55,62 @@ async function synchronouslyRunPostDeployMigration(systemRepo: SystemRepository,
...
+
+    // Scheduling features use serializable transactions that touch these
+    // tables. The `fastUpdate` feature can cause seemingly unrelated transactions
+    // to append to the same "pending list", which can cause transaction
+    // failures.
+    //
+    // Here we update the indexes on Appointment and Slot tables to disable `fastUpdate`,
+    // and then vacuum the tables to clear any existing pending list entries.
+    const actions: OutputAction[] = [];
+    const tables = ['Appointment', 'Appointment_References', 'Slot', 'Slot_References'];
```
The-Alchemist, line 94: @mattlong help me understand why just these tables?

mattlong: @noahsilas added those since some scheduling tests were often hitting serializable errors during test. @noahsilas, do you remember which tests were affected?

The-Alchemist (approved):

PR #9891, opened 2026-07-17
Server Scoped Subscriptions Server Config

finnbergquist:


packages/docs/docs/self-hosting/server-config.md
```diff
@@ -677,16 +677,24 @@ Flag to enable pre-commit subscriptions for the interceptor pattern.
 
 **Default:** `false`
 
+### serverScopedSubscriptionsEnabled
+
+Flag to enable server-scoped rest-hook Subscriptions. When enabled, the subscription worker evaluates not only the subscriptions within a resource's own project, but also "server-scoped" subscriptions that are not scoped to any project (i.e. stored in the system project). This allows a single set of subscriptions to fire for changes across every project on the server, which is useful for large self-hosted deployments that use a project-per-tenant model but need a central data-processing flow for all tenants.
```
mattlong: You're more familiar with the docs than I am, so maybe what you had is how we refer to system resources, but there technically isn't a system project; they are resources not part of any project.

```suggestion
Flag to enable server-scoped rest-hook Subscriptions. When enabled, the subscription worker evaluates not only the subscriptions within a resource's own project, but also "server-scoped" subscriptions that are not scoped to any project (i.e. a system resource created by a super admin). This allows a single set of subscriptions to fire for changes across every project on the server, which is useful for large self-hosted deployments that use a project-per-tenant model but need a central data-processing flow for all tenants.
```

mattlong (approved):

PR #9892, opened 2026-07-17
Clarify and normalize BullMQ worker config helpers

mattlong:
Stemmed primarily from the realization that the effective `concurrency` of the `batch` worker was the default `20` instead of the intended `1`. Along with the helper signature changes, `batch` currency defaults to `15`.


The-Alchemist:
@claude review.  especially validate that the pre-existing config for packages/server/src/workers/data-warehouse-sync.ts is the same, semantically

The-Alchemist (approved):

PR #9896, opened 2026-07-17
Job priority based fair queueing for the async batch queue

mattlong:
**UPDATE** - This PR needs some work; priority alone is not enough to get the desired behavior; e.g. newer jobs for a given project will be sequenced before older one still on the queue over time.

An opt-in feature that prevents projects from starving the async batch processing queue by lowering the priority of async batch requests based on how many jobs a project already has in the queue.

### DIFF
--- a/packages/server/src/config/types.ts
+++ b/packages/server/src/config/types.ts
@@ -105,9 +105,9 @@ export interface MedplumServerConfig {
   /**
    * Enables project-based fair queueing for async batch jobs, so one project's backlog does not
    * starve other projects. Default is `true`; overridable per-project via the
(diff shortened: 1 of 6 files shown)


The-Alchemist:
@claude review

mattlong:
@claude Review again

The-Alchemist (approved):

ThatOneBro (changes requested):
We talked offline about this, seems like this priority-based system might be too simple to get us the fair queuing that we need

Still noodling on a design that would be both simple and adequately address our fair queueing needs

PR #9901, opened 2026-07-18
fix(react): handle unresolvable JSON patch paths in ResourceDiffTable

### LINKED ISSUE #9900: ResourceDiffTable crashes with "Cannot read properties of undefined (reading 'type')" when array elements are reordered, inserted, or removed
### Describe the bug

The resource history diff view (`ResourceDiffTable`) crashes with:

```
Something went wrong
Cannot read properties of undefined (reading 'type')
```

whenever an edit reorders, inserts, or removes array elements (e.g. `telecom`, `extension`) in a way that produces JSON patch operations whose array indices do not resolve against the stored versions.

### Root cause

`ResourceDiffTable` computes an RFC 6902 patch between the two versions and then evaluates each operation's path (converted to FHIRPath) **statically** against the original/revised resources. But JSON patch paths are **sequential** — each operation assumes the prior operations were already applied.

Minimal example — reordering two telecom entries:

```ts
const original = {
  resourceType: 'Practitioner',
  telecom: [
    { system: 'phone', value: '555-555-1234' },
    { system: 'email', value: 'alice@example.com' },
  ],
};
const revised = {
  resourceType: 'Practitioner',
  telecom: [
    { system: 'email', value: 'alice@example.com' },
    { system: 'phone', value: '555-555-1234' },
  ],
};
createPatch(original, revised);
// [
//   { op: 'add', path: '/telecom/0', value: {...} },
//   { op: 'remove', path: '/telecom/2' }   // <-- index 2 only exists AFTER the add is applied
// ]
```

`evalFhirPathTyped('telecom[2]', typedOriginal)` returns `[]` (the original only has indices 0–1). The empty array passes the `!input` guard in `touchUpValue`, and then:

```ts
type: Array.isArray(input) ? input[0].type : input.type,
// input = [] -> input[0] is undefined -> TypeError: Cannot read properties of undefined (reading 'type')
```

https://github.com/medplum/medplum/blob/main/packages/react/src/ResourceDiffTable/ResourceDiffTable.tsx#L173

The same failure is reachable through the `.url`-suffix stripping (`/extension/5/url` → `extension[5]`) and through `mergePatchOperations` index consolidation.

### How often this happens in practice

We replayed the `ResourceDiffTable` diff computation offline against the full version history of one small production project (1,042 resources, 2,447 version pairs): **102 history rows crash**, across Composition, Patient, Practitioner, and Encounter resources. Any app that edits extension arrays in place (very common) generates these.

### To reproduce

1. Create a Practitioner with `telecom: [phone, email]`.
2. Update it to `telecom: [email, phone]` (swap the order).
3. Open the resource history in the Medplum app and view the diff for that version.

### Expected behavior

The diff row renders (even if one side is blank because the static path did not resolve) instead of crashing the whole diff for that version.

Happy to submit a PR — a minimal guard in `touchUpValue` (return `undefined` when the evaluated result is an empty array) fixes the crash and degrades gracefully to an empty Before/After cell. A deeper fix would be evaluating patch paths against sequentially-patched intermediate states so the true values could be displayed, but that changes a lot more of the diff logic — the guard may be the right first step.

ajwwong:
Fixes #9900

`ResourceDiffTable` evaluated JSON patch operation paths statically against the original/revised resources, but JSON patch paths are sequential — each operation assumes the prior operations were already applied. When an edit reorders, inserts, or removes array elements (e.g. `telecom`, `extension`), an operation like `remove /telecom/2` can reference an index that does not exist in either stored version. `evalFhirPathTyped` then returned an empty array, which passed the `!input` guard in `touchUpValue` and crashed on `input[0].type`:

```
Cannot read properties of undefined (reading 'type')
```

Per review feedback, this now applies the patch operations one at a time to an intermediate state and evaluates each operation's path against the state just before the operation (for removed/replaced values) and just after it (for added/replaced values), so every diff row shows the true value. A reorder now renders as `Add telecom[0]` / `Remove telecom[2]` with the actual moved values instead of empty cells or a crash. Consolidated array rewrites (multiple adds/removes on the same array) keep the existing behavior of showing the whole array before and after, and the empty-eval guard in `touchUpValue` remains as a defensive fallback.

The regression test reproduces the crash on `main` (reordering two telecom entries produces `add /telecom/0` + `remove /telecom/2`) and asserts that both rows now render the true values.

### DIFF
--- a/packages/react/src/ResourceDiffTable/ResourceDiffTable.test.tsx
+++ b/packages/react/src/ResourceDiffTable/ResourceDiffTable.test.tsx
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import type { MedicationRequest, Patient } from '@medplum/fhirtypes';
+import type { MedicationRequest, Patient, Practitioner } from '@medplum/fhirtypes';
 import { MockClient } from '@medplum/mock';
 import { MedplumProvider } from '@medplum/react-hooks';
 import { act, render, screen } from '../test-utils/render';
@@ -213,6 +213,39 @@ describe('ResourceDiffTable', () => {
 
     const operations = screen.getAllByText('Replace identifier');
     expect(operations).toHaveLength(1);
+  });
+
+  test('Array reorder with out-of-range patch index', async () => {
+    // Reordering array elements produces a JSON patch such as:
+    //   [{ op: 'add', path: '/telecom/0' }, { op: 'remove', path: '/telecom/2' }]
+    // JSON Patch paths are sequential, so "/telecom/2" only exists after the "add" is applied.
+    // Evaluating "telecom[2]" statically against the original resource resolves to nothing.
+    // This previously crashed with "Cannot read properties of undefined (reading 'type')".
+    const original: Practitioner = {
+      resourceType: 'Practitioner',
+      id: '123',
+      meta: { versionId: '456' },
+      telecom: [
+        { system: 'phone', value: '555-555-1234' },
+        { use: 'work', system: 'email', value: 'alice@example.com' },
+      ],
+    };
+
+    const revised: Practitioner = {
+      ...original,
+      meta: { versionId: '457' },
+      telecom: [
+        { system: 'email', value: 'alice@example.com' },
+        { system: 'phone', value: '555-555-1234' },
+      ],
+    };
+
+    await act(async () => {
+      setup({ original, revised });
+    });
+
+    expect(await screen.findByText('Add telecom[0]')).toBeInTheDocument();
+    expect(screen.getByText('Remove telecom[2]')).toBeInTheDocument();
   });
 
   test('Change attachment URL', async () => {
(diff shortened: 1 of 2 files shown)


noahsilas (changes requested):
Hi @ajwwong, thanks for the contribution!

It looks like your commit does not have a [DCO trailer](https://github.com/apps/dco) in its message. Take a look at [the DCO check](https://github.com/medplum/medplum/pull/9901/checks?check_run_id=88097714723) for some instructions on how you can update your PR to indicate your agreement.

While I appreciate how your current approach is fairly minimal, I think that the alternative strategy you propose is going to be better for viewers. Showing a row like "Removed telecom[2]" without showing the value that was removed is not very helpful in understanding that this was a reordering rather than an addition.

I'd be very interested to see the approach you described where we incrementally apply the patch operations at each step. I think that will help viewers see the impact of each update most readily.

Thanks!

ajwwong:
@noahsilas Thanks for the review! Both items addressed:

- **DCO**: commits are now signed off.
- **Sequential evaluation**: implemented the approach you asked for — patch operations are applied one at a time to an intermediate state, and each path is evaluated against the state just before the operation (removed/replaced values) and just after it (added/replaced values). A reorder now shows the actual moved values on both the add and remove rows instead of empty cells. The regression test asserts the true values render.

Ready for another look.

ajwwong:
@noahsilas Gentle ping on this one when you have a moment — the sequential-evaluation rework and DCO sign-off from your review are in place, and CI/coverage are green after the recent main merge. We still hit this crash regularly on history rows where an app reorders the extension array (e.g. a removed-and-reinserted extension produces replace/remove ops at indices that don't resolve statically against the before-resource). Happy to adjust anything further.

PR #9905, opened 2026-07-21
Add end-to-end regression tests for every supported auth strategy

### LINKED ISSUE #9879: Add regression tests covering every supported authentication strategy
We lack automated regression coverage across all supported authentication strategies (native, external/OIDC, direct JWT bearer, machine-to-machine).

## Acceptance criteria:

* Automated test suite exercises every supported authentication strategy end-to-end.
* Tests fail if a change alters the behavior of a strategy other than the one being modified.

ianplunkett:
## Summary
- Closes #9879 by adding `packages/server/src/oauth/auth-strategies.test.ts`, a coarse-grained regression suite that drives each supported authentication strategy through its real HTTP front door, obtains a token, and proves it against a live `/fhir/R4` call — the missing "did the whole flow actually work" check that existing per-strategy unit tests don't provide (they mint tokens directly or stop at the token response, never chaining into a FHIR API call).
- Covers the four strategies named in the issue — native password login, external/OIDC login, JWT bearer via RFC 8693 token exchange, and client credentials (M2M) — plus five more discovered while auditing `Login.authMethod` and the `/oauth2/token` grant-type switch: Google-specific login, the OpenID4VCI pre-authorized code grant, direct HTTP Basic auth against the FHIR API, and the mTLS and private_key_jwt variants of client credentials.
- Each strategy has its own isolated Project/ClientApplication/user fixtures and its own `describe` block, so a regression introduced in one strategy's code cannot spuriously fail another strategy's tests, and CI output attributes a failure to the exact strategy by name.
- Out of scope: the RFC 7523 `grant_type=jwt-bearer` grant used by `MedplumClient.startJwtBearerLogin()` has no server-side handler in `oauth/token.ts` at all (confirmed on current `main`) — this is a separate, real gap I've flagged for a follow-up rather than folding into this PR.

## Test plan
- [x] `npx vitest run src/oauth/auth-strategies.test.ts` — all 9 tests pass
- [x] Full `packages/server` suite (`npx vitest run --exclude src/seed.test.ts`) — 248/249 files pass; the only failure is a pre-existing, unrelated `oauth/token.test.ts` "Client assertion" issue (confirmed to fail identically with this file removed)
- [x] `tsc --noEmit` and `eslint` clean on the new file


The-Alchemist:
@claude review

ianplunkett:
Addressed the review feedback:

1. **jose mock scope** — Reworked the mock so it only bypasses signature verification for the two external-JWKS cases that genuinely can't be mocked otherwise (Google's public certs, a client's `jwksUri` for private_key_jwt). Medplum's own access-token verification (`oauth/keys.ts` `verifyJwt`) now runs the real `jwtVerify`, so every `/fhir/R4` call in this file is verifying a real signature again. Added a tampered-signature negative assertion in the native login test to prove it (confirmed locally: a mutated token now gets rejected with 401).
2. **Leaked `fetch` spy** — Your suggested fix (`vi.restoreAllMocks()` in each block's own `afterAll`) actually breaks the file: `vi.spyOn` calls are layered during collection (all `describe` bodies run before any test executes), so a later block's spy already sits on top of an earlier one by the time tests run, and restoring the earlier spy mid-run clobbers the later block's active spy out from under it. Fixed by restoring once in a single top-level `afterAll` after every describe in the file has finished, which avoids the ordering hazard.
3. **9x `initApp`/`shutdownApp`** — Left as-is. Isolation is already achieved at the fixture (Project/Client) level per your own note, and consolidating to one shared app would save some CI time but adds cross-describe app-state coupling for a suite whose whole point is per-strategy isolation — didn't seem worth the risk for a "not a blocker" item, happy to revisit if you feel otherwise.
4. **base64 vs base64url** — Fixed.

All 9 tests still pass, plus the new negative assertion; full `packages/server` suite is green except the same pre-existing, unrelated `oauth/token.test.ts` "Client assertion" failures (confirmed to fail identically on `main` without this PR — local `jose` version mismatch in my sandbox, not something this PR touches).

The-Alchemist:
@claude review

The-Alchemist (approved):

PR #9907, opened 2026-07-21
Use existing identity when conditional create retry finds resource

mattwiller:


The-Alchemist:
@claude review

The-Alchemist (approved):

packages/fhir-router/src/batch.ts
```diff
@@ -676,6 +677,26 @@ export class BatchProcessor {
...
+   * the entry should resolve to that existing resource, not the one assigned during preprocessing.
+   * @param entry - The original Bundle entry that was processed.
+   * @param result - The response entry produced for it, carrying the persisted resource.
+   */
+  private reconcileResolvedIdentity(entry: BundleEntry, result: BundleEntry): void {
+    const placeholder = entry.fullUrl;
+    const resource = result.resource;
+    // Only update placeholders that were resolved during preprocessing
+    if (placeholder && this.resolvedIdentities[placeholder] && resource?.id) {
+      this.resolvedIdentities[placeholder] = getReferenceString(resource as WithId<Resource>);
```
mattlong: `this.resolvedIdentities` is part of the `BatchInitialState` which was previously static after `preprocess`. This is changing that invariant, so there needs to be some feedback to the caller, i.e. `execBatchJob`, that the processor's `BatchInitialState` has changed and needs to be updated in storage again. "initial" state isn't an accurate name either now.

mattlong (changes requested):

packages/fhir-router/src/batch.ts
```diff
@@ -659,15 +710,15 @@ export class BatchProcessor {
...
+        for (let i = this.state.position; i < bundleInfo.ordering.length; i++) {
           this.setResult(bundleInfo.ordering[i], buildBundleResponse(err.outcome));
         }
         this.state.position = bundleInfo.ordering.length;
         return;
       }
 
       this.setResult(entryIndex, buildBundleResponse(normalizeOperationOutcome(err)));
-      this.state.position = n + 1;
+      this.state.position++;
```
mattlong, line 721: Seems like the increment could go in a `finally`?

mattlong:
@claude look again since the implementation changed meaningfully

Issue #9350, opened 2026-05-29
[v6] Richer return value of `useResource`


mattlong:
From @noahsilas in [this comment](https://github.com/medplum/medplum/pull/9280/changes#r3314704794) on #9280.

> I think this hook should return a tuple `[resource, loading, outcome]`, similar to `useSearch*`, so that components like this can easily show a `<Loader />` or similar when appropriate.

mattlong:
react-query does object-style return value while useState does array-style, which way to go is TBD.

PR #9909, opened 2026-07-21
fix(provider): show lab result details when the order ServiceRequest cannot be read

techdavidy:
Fixed LabDetailPane so a DiagnosticReport whose ServiceRequest was deleted shows the result instead of spinning forever; refactored per your requests to an explicit readReference fetch with loading state.

### DIFF
--- a/examples/medplum-provider/src/components/labs/LabDetailPane.tsx
+++ b/examples/medplum-provider/src/components/labs/LabDetailPane.tsx
@@ -4,8 +4,10 @@ import { Box, Center, Loader } from '@mantine/core';
 import type { WithId } from '@medplum/core';
 import { isReference } from '@medplum/core';
 import type { DiagnosticReport, Reference, ServiceRequest } from '@medplum/fhirtypes';
-import { useResource } from '@medplum/react';
+import { useMedplum } from '@medplum/react';
 import type { JSX } from 'react';
+import { useEffect, useState } from 'react';
+import { showErrorNotification } from '../../utils/notifications';
 import { LabOrderDetails } from './LabOrderDetails';
 import { LabResultDetails } from './LabResultDetails';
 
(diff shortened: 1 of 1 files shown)


examples/medplum-provider/src/components/labs/LabDetailPane.tsx
```diff
@@ -23,33 +25,65 @@ interface LabDetailPaneProps {
...
   const { item } = props;
+  const medplum = useMedplum();
 
-  // A ServiceRequest is its own order. A DiagnosticReport may be based on one,
-  // in which case we prefer the order view over the raw result.
   const basedOnRef =
     item.resourceType === 'DiagnosticReport'
       ? item.basedOn?.find((ref): ref is Reference<ServiceRequest> => isReference(ref, 'ServiceRequest'))
       : undefined;
-  const order = useResource(item.resourceType === 'ServiceRequest' ? item : basedOnRef);
```
noahsilas, line 33: This seems like another situation where a richer return value from `useResource` could have helped - https://github.com/medplum/medplum/issues/9350. (No action needed, just adding supporting evidence for that future work).

examples/medplum-provider/src/components/labs/LabDetailPane.tsx
```diff
@@ -23,33 +25,65 @@ interface LabDetailPaneProps {
...
+        return;
+      }
+      setLoading(true);
+      try {
+        const result = await medplum.readReference(basedOnRef);
+        if (subscribed) {
+          setOrder(result);
+        }
+      } catch (err) {
+        showErrorNotification(err);
```
noahsilas:
```suggestion
        if (subscribed) {
          showErrorNotification(err);
        }
```

examples/medplum-provider/src/components/labs/LabDetailPane.tsx
```diff
@@ -23,33 +25,65 @@ interface LabDetailPaneProps {
...
+  // fetched from its basedOn reference.
+  const [order, setOrder] = useState<WithId<ServiceRequest> | undefined>(
+    item.resourceType === 'ServiceRequest' ? item : undefined
+  );
+  const [loading, setLoading] = useState<boolean>(!!basedOnRef);
+
+  useEffect(() => {
+    let subscribed = true;
+    const fetchOrder = async (): Promise<void> => {
+      if (item.resourceType === 'ServiceRequest') {
```
noahsilas, line 45: It is starting to feel a little confusing with all these type checks, I wonder if we would be better served by breaking this into two more specialized components? Just a thought, not a blocker.

noahsilas (approved):

examples/medplum-provider/src/components/labs/LabDetailPane.tsx
techdavidy: this is a mix of both DiagnosisReport and ServiceRequest. There are separate component more detailed information from both.

PR #9910, opened 2026-07-21
Gracefully handle reading Binary with no content

mattlong:
Matches the 404 behavior of the pre-signed read URLs. Previously this would result in a 500 response in part because the status and headers were prematurely set in `sendResponseHeaders` before establishing the content read stream.

### DIFF
--- a/packages/server/src/fhir/response.ts
+++ b/packages/server/src/fhir/response.ts
@@ -1,12 +1,15 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { ContentType, concatUrls, getStatus, isCreated } from '@medplum/core';
+import { ContentType, concatUrls, getStatus, isCreated, notFound } from '@medplum/core';
 import type { FhirResponseOptions } from '@medplum/fhir-router';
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/response.ts
```diff
@@ -31,13 +34,38 @@ export function sendResponseHeaders(_req: Request, res: Response, outcome: Opera
...
+    getLogger().error('Error reading Binary content', {
+      err,
+      binary: binaryResource.id ? `Binary/${binaryResource.id}` : undefined,
+      versionId: binaryResource.meta?.versionId,
+    });
+    if (!res.headersSent) {
+      res.removeHeader('ETag');
+      res.removeHeader('Last-Modified');
+      res.removeHeader('Content-Type');
+      sendOutcome(res, notFound);
```
mattlong, line 65: I could be convinced that this should return a 204 instead of a 404, but that would require updating storage implementations to differentiate "this key doesn't exist" from other failure modes.

mattlong:
@claude take a look

ThatOneBro (approved):

packages/server/src/fhir/response.ts
ThatOneBro: Eh I think 404 is good, think 204 could be confusing. Plus this is a pretty rare edge case

PR #9911, opened 2026-07-21
Batch AuditEvent writes to Postgres

mattwiller:


packages/server/src/fhir/repo.ts
```diff
@@ -2324,20 +2421,65 @@ export class Repository extends FhirRepository implements Disposable {
...
+   * 2. the un-awaited save begins a transaction which would clash with any subsequent mainline
+   * transactions started on the current Repository and cause one of them to fail.
+   *
+   * Batching all events from a single interaction into one transaction (one multi-row INSERT into the
+   * AuditEvent table, one into AuditEvent_History, and one batched lookup-table write) avoids the
+   * per-event transaction + round-trip overhead that would otherwise scale linearly with the number
+   * of events — see {@link readReferences}.
+   *
+   * AuditEvents are trusted, internally-generated resources, so this deliberately skips the heavy
+   * permission/validation path of `updateResourceImpl`, but reuses {@link assignWriteMeta} to apply
```
mattlong, line 2444: Skipping `updateResourceImpl` is scary since previously all resources written to the DB  passed through it; there's a risk of it drifting from the transforms, resource validation, and permissions checks that happen there:

Bypassing permissions checks seems particularly problematic since an access policy could theoretically prevent someone from being able to create AuditEvent.

It would also bypass the pre-commit logic where someone could theoretically have a pre-commit subscription looking for AuditEvent.

And then just the general resource validation.

I wonder if instead `updateResourceImpl` should be broken up into or call a preflight method that does all everything short of actually writing the resource...

mattlong (changes requested):

PR #9920, opened 2026-07-22
Move MockClient control methods into `.mock` namespace

noahsilas:
When reading some tests, it's not always obvious that some methods (particularly ones like `.setProfile()`) are part of the MockClient testing API, and not part of the actual MedplumClient surface.

Moving these methods into a `.mock` namespace makes it clear that these methods are affecting the behavior of the MockClient, and not representative of a method that would be called on the actual instance.

I've deprecated the existing methods, and we should remove them in a future major release.


ThatOneBro (approved):
Nice, makes it clear what is a mock now.

I also like that you marked all the old methods as deprecated 🙏 

You might also want to create an issue to remove the old deprecated methods

Good stuff 👍

mattlong (approved):

PR #9922, opened 2026-07-22
Add resource type header to resource page

### LINKED ISSUE #9919: Resource type header on resource page
### Problem 

Image

### Recommended solution

Adham-Kiwan:
Fixes #9919.

## Summary

Adds a header to the resource list/search page (`HomePage.tsx`, e.g. `/Bot`) showing the resource type, so it's clear what you're looking at even for resource types that aren't in the primary sidebar. Follows the same header pattern already used on `CreateResourcePage`.

## Changes

- `packages/app/src/HomePage.tsx`: adds a `Text` header above `SearchControl` showing `search.resourceType`.
- `packages/app/src/HomePage.test.tsx`: adds a test asserting the header renders for a non-sidebar resource type (`Bot`).

## Test plan

- `tsc --noEmit` on `packages/app`: clean.
- `eslint` on the two changed files: 0 errors (1 pre-existing, unrelated warning on an untouched line).
- `vitest run src/HomePage.test.tsx`: 15/15 passing.
- `vitest run src/resource/ResourcePage.test.tsx`: 17/17 passing (confirms no leftover changes from the earlier revision).

### DIFF
--- a/packages/app/src/HomePage.test.tsx
+++ b/packages/app/src/HomePage.test.tsx
@@ -40,6 +40,12 @@ describe('HomePage', () => {
     expect(await screen.findByTestId('search-control')).toBeInTheDocument();
   });
 
+  test('Renders resource type header', async () => {
+    await setup('/Bot');
+    expect(await screen.findByTestId('search-control')).toBeInTheDocument();
+    expect(screen.getByText('Bot')).toBeInTheDocument();
+  });
+
   test('Renders with resourceType and fields', async () => {
     await setup('/Patient?_fields=id,_lastUpdated,name,birthDate,gender');
     expect(await screen.findByTestId('search-control')).toBeInTheDocument();
--- a/packages/app/src/HomePage.tsx
+++ b/packages/app/src/HomePage.tsx
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Paper } from '@mantine/core';
+import { Paper, Text } from '@mantine/core';
 import { showNotification } from '@mantine/notifications';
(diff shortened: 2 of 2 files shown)


noahsilas:
Hi @Adham-Kiwan, thanks for the contribution!

It looks like your commit does not have a [DCO trailer](https://github.com/apps/dco) in its message. Take a look at [the DCO check](https://github.com/medplum/medplum/pull/9922/checks?check_run_id=89017701381) for some instructions on how you can update your PR to indicate your agreement. 

It looks to me like this is updating the instance-level page, which does already show the resource type, and not the list view that #9919 has flagged is missing this context. Can you update to move this into the list view instead? Attaching a screenshot of the rendered output would also help your work get reviewed in a timely manner.

Thanks!

noahsilas (changes requested):
This update should be applied to the list view instead of the detail page view (which already shows the type in its header).

Adham-Kiwan:
Thanks for the quick review!

- DCO: fixed, the commit now has a proper sign-off.
- Good catch on the location — you're right that the detail page already shows the type in its header. I moved the change to `HomePage.tsx` so it now shows up on the list/search view instead (e.g. `/Bot`), which is what #9919 was actually about.

I don't have a way to grab a live screenshot from an authenticated deploy right now, but I added a test (`Renders resource type header` in `HomePage.test.tsx`) that asserts the header text renders on a non-sidebar resource type, and ran the full `HomePage` and `ResourcePage` suites locally to confirm nothing else regressed. Happy to add a screenshot if there's an easy way to get one from the preview deploy.

noahsilas:
Thanks for the contribution @Adham-Kiwan, I think for this issue we are likely to go with #9946 - it's very similar to your solution, but has added some styling that we can evaluate easily from the screenshots. We appreciate your work on this issue, and hope you will continue contributing!

PR #9923, opened 2026-07-22
Add resource type header on resource list page (#9919)

### LINKED ISSUE #9919: Resource type header on resource page

ahmadmustafa02:
## What
Adds a header showing the resource type name (e.g. "Bot") above the resource table in HomePage.tsx — the resource list/search view. This is distinct from PR #9922, which addresses the single-resource detail page (ResourcePage.tsx).

## Why
Closes #9919 (list view aspect)

## Changes
- Added `<Title order={3} mb="sm">{search.resourceType}</Title>` above `<SearchControl>` in HomePage.tsx
- Confirmed no existing helper maps ResourceType → display label elsewhere in the app; raw ResourceType strings are used as-is throughout the codebase for display purposes

## Testing
- Added new test `Renders resource type header` in HomePage.test.tsx, asserting the heading renders with the correct resourceType name
- Full HomePage.test.tsx suite passes (15/15)
- `tsc --noEmit` and `eslint` pass with no errors

### DIFF
--- a/packages/app/src/HomePage.test.tsx
+++ b/packages/app/src/HomePage.test.tsx
@@ -45,6 +45,12 @@ describe('HomePage', () => {
     expect(await screen.findByTestId('search-control')).toBeInTheDocument();
   });
 
+  test('Renders resource type header', async () => {
+    await setup('/Bot');
+    expect(await screen.findByTestId('search-control')).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


techdavidy:
Thanks for the input on the issues mentioned for this PR.
We already merged a fix for it https://github.com/medplum/medplum/pull/9946

PR #9924, opened 2026-07-22
Handle (conditional) FHIRPath Patch in Batch requests

### LINKED ISSUE #9917: FHIRPath Patch: rejected in batch/transaction bundles, no enforced write precondition, and broken `insert` — blockers for adopting it as our patch transport
## Background

Following up on #1263 / #9377 / #9476 (thanks for landing these!). We ran a full cutover investigation to move our EHR from JSON Patch to FHIRPath Patch as the wire format, testing live against hosted Medplum (`api.medplum.com`, `software.version: 5.1.26-eb2ba7e`).

Direct single-resource PATCH with a FHIRPath `Parameters` body works ✅. But three gaps block actual adoption — the biggest being that batch/transaction bundles (where our concurrent-write races originate, and the whole reason we want element-keyed patching) reject FHIRPath `Parameters` entirely.

All repros below use a throwaway `Basic` resource. `$TOKEN` is a project access token; `$BASE` is `https://api.medplum.com/fhir/R4`.

### Setup

```bash
curl -s -X POST "$BASE/Basic" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/fhir+json" \
  -d '{"resourceType":"Basic","code":{"text":"fhirpath-patch-repro"}}'
# => 201; export ID=<returned id>, VID=<returned meta.versionId>
```

### Baseline (works): direct PATCH with FHIRPath Parameters

```bash
curl -s -X PATCH "$BASE/Basic/$ID" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/fhir+json" \
  -d '{
    "resourceType": "Parameters",
    "parameter": [{
      "name": "operation",
      "part": [
        {"name": "type", "valueCode": "add"},
        {"name": "path", "valueString": "Basic"},
        {"name": "name", "valueString": "created"},
        {"name": "value", "valueDate": "2026-01-01"}
      ]
    }]
  }'
# => 200, `created` applied ✅
```

---

## Issue 1 — FHIRPath `Parameters` PATCH entries rejected in batch/transaction bundles

The same `Parameters` body that succeeds on direct PATCH fails as a bundle entry:

```bash
curl -s -X POST "$BASE" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/fhir+json" \
  -d '{
    "resourceType": "Bundle",
    "type": "batch",
    "entry": [{
      "request": {"method": "PATCH", "url": "Basic/'$ID'"},
      "resource": {
        "resourceType": "Parameters",
        "parameter": [{
          "name": "operation",
          "part": [
            {"name": "type", "valueCode": "add"},
            {"name": "path", "valueString": "Basic"},
            {"name": "name", "valueString": "created"},
            {"name": "value", "valueDate": "2026-01-01"}
          ]
        }]
      }
    }]
  }'
```

Entry response:

```json
{
  "response": {
    "status": "400",
    "outcome": {
      "issue": [{"severity": "error", "code": "invalid", "details": {"text": "PATCH Parameters missing op"}}]
    }
  }
}
```

A control entry in the same bundle carrying the legacy Binary JSON Patch form (`{"resourceType":"Binary","contentType":"application/json-patch+json","data":"<base64 ops>"}`) succeeds with `200`.

**Root cause:** `packages/fhir-router/src/batch.ts` → `parsePatchBody` only understands `Parameters` as JSON-Patch-encoded parts (`op`/`path`/`value`/`from`). The FHIRPath case is acknowledged in a comment — `// FHIRPatch will also use Parameters; however, it uses 'name = type'` — but the disambiguation isn't implemented, so bundle PATCH entries never reach `fhirpathPatchTypedValue`. (Conditional PATCH `PATCH /{type}?search` similarly only accepts `Operation[]` arrays.)

**Ask:** route bundle PATCH entries whose `Parameters.parameter[].part` uses `name = "type"` through the FHIRPath applier, same as `repo.patchResource` does for direct PATCH.

---

## Issue 2 — no enforced write precondition on PATCH (`If-Match` / `entry.request.ifMatch` ignored)

FHIRPath Patch has no `test` op (per spec), so the standard optimistic-concurrency idiom for JSON Patch — a leading `test` on `/meta/versionId` — has no FHIRPath equivalent. The spec'd replacement is the `If-Match` precondition, but it is not enforced:

```bash
# Deliberately wrong version — expect 412, get 200
curl -s -X PATCH "$BASE/Basic/$ID" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/fhir+json" \
  -H 'If-Match: W/"00000000-0000-0000-0000-000000000000"' \
  -d '{"resourceType":"Parameters","parameter":[{"name":"operation","part":[
    {"name":"type","valueCode":"replace"},
    {"name":"path","valueString":"Basic.created"},
    {"name":"value","valueDate":"2026-02-02"}]}]}'
# => 200, patch applied anyway ❌ (header confirmed sent via curl -v)
```

Same result with a JSON Patch body + wrong `If-Match` (so it's format-independent), and same in bundles with `entry.request.ifMatch: 'W/"bogus"'` — the entry returns `200` and applies.

Control showing the JSON Patch `test` op IS enforced:

```bash
curl -s -X PATCH "$BASE/Basic/$ID" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json-patch+json" \
  -d '[{"op":"test","path":"/meta/versionId","value":"wrong-version"},
       {"op":"replace","path":"/code/text","value":"nope"}]'
# => 400 "Test failed: <actual> != <supplied>" ✅ enforced
```

**Consequence:** the JSON Patch `test` op is currently the *only* working concurrency guard on this server. Cutting over to FHIRPath Patch today would silently drop optimistic concurrency — the opposite of why element-keyed patching is attractive.

**Ask:** enforce `If-Match` on direct PATCH and `Bundle.entry.request.ifMatch` on bundle PATCH entries (412 on mismatch), or document the intended precondition mechanism for FHIRPath Patch.

---

## Issue 3 — FHIRPath `insert` never parses its `index` (always "undefined")

With `Basic.identifier` holding one element:

```bash
curl -s -X PATCH "$BASE/Basic/$ID" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/fhir+json" \
  -d '{
    "resourceType": "Parameters",
    "parameter": [{
      "name": "operation",
      "part": [
        {"name": "type", "valueCode": "insert"},
        {"name": "path", "valueString": "Basic.identifier"},
        {"name": "index", "valueInteger": 0},
        {"name": "value", "valueIdentifier": {"system": "urn:test", "value": "inserted"}}
      ]
    }]
  }'
# => 400 "Source index out of bounds for insert operation: undefined"
```

Tried `valueInteger` at 0 and at 1 (append position), `valuePositiveInt`, and a part named `source` — every variant fails with the index read as `undefined`. Looks like the `index` parameter isn't surviving `parseParametersFromDefinitions` on its way into `fhirpathPatchTypedValue` (the utility's own unit tests exercise the typed `FhirPathPatch[]` IR directly, which would bypass the Parameters-parsing layer).

---

## Environment

- Server: hosted `api.medplum.com`, `{"name":"medplum","version":"5.1.26-eb2ba7e"}` (from `/fhir/R4/metadata` and `/healthcheck`)
- All repros against a `Basic` resource in a dev project; direct PATCH discrimination confirmed to be body-shape based (array vs `Parameters`), Content-Type-independent
- CapabilityStatement still advertises only `patchFormat: ["application/json-patch+json"]` — worth updating once the above land

## Why this matters to us

We maintain a client-side bridge that authors FHIRPath `Parameters` patches (element-keyed, `where(url=...)` targeting) but lowers them to JSON Patch before submit, purely because of the gaps above — full context in the writeup on #1263. Issues 1 + 2 are the blockers: once bundle support and an enforced precondition exist, we can delete the bridge and adopt FHIRPath Patch end to end. Happy to split this into three issues, provide raw request/response captures, or test a canary build.

mattwiller:
Resolves #9917


packages/core/src/fhirpath/patch.ts
```diff
@@ -19,6 +20,60 @@ export type FhirPathPatch =
...
+ * @param parameters - The FHIRPath Patch Parameters resource.
+ * @returns The parsed FHIRPath Patch operations.
+ */
+export function parseFhirPathPatchParameters(parameters: Parameters): FhirPathPatch[] {
+  const operations: FhirPathPatch[] = [];
+  for (const param of parameters.parameter ?? []) {
+    if (param.name !== 'operation') {
+      continue;
+    }
+    const op: Record<string, unknown> = {};
```
codyebberson:
```suggestion
    const op: Partial<FhirPathPatch> = {};
```

packages/core/src/fhirpath/patch.ts
```diff
@@ -19,6 +20,60 @@ export type FhirPathPatch =
...
+          op.value = getTypedValueFromPart(part);
+          break;
+      }
+    }
+    operations.push(op as FhirPathPatch);
+  }
+  return operations;
+}
+
+function getTypedValueFromPart(part: ParametersParameter): TypedValue | undefined {
```
codyebberson: This makes sense.

In theory, you could use `getTypedPropertyValue()`, which would be safer, but it's also slower and heavier.

It would protect against things like `"valueJibberish": "foo"` being incorrectly parsed.

packages/core/src/fhirpath/patch.ts
```diff
@@ -19,6 +20,60 @@ export type FhirPathPatch =
...
+  for (const key of Object.keys(part)) {
+    if (key.startsWith('value')) {
+      return { type: key.substring(5), value: part[key as keyof ParametersParameter] };
```
codyebberson: Is `.substring(5)` adequate here?  `valueString` -> `type: 'String'`, when it should be lowercase `s`?

codyebberson (approved):
Minor, non-blocking

codyebberson:
@claude please review

packages/core/src/fhirpath/patch.ts
mattwiller: I ended up using `getTypedPropertyValue()` for the safety 👍

packages/core/src/fhirpath/patch.ts
mattwiller: Building the object from parts in the loop below doesn't work well with `Partial<FhirPathPatch>`, since many of the fields of `FhirPathPatch` can't be `undefined` and the type is itself a union of operation shapes.  `Object.create(null)` didn't work well either, due to the behavior of vitest's `expect().toStrictEqual()` assertion with objects of different prototypes, so I just went with `{} as any` for now, in line with the `as FhirPathPatch` below 🤷

PR #9930, opened 2026-07-22
[MockClient] test cleanup

noahsilas:
While working on #9920, I noticed some weirdness when looking to enhance test coverage in here.

- In one test we were stubbing `console.log`, and not cleaning it up. This made some debugging steps working on other tests difficult.
- After clearing that stub, some extra log lines started showing up in the output. 
  - removed an explicit `console.log`
  - cleared LocalStorage between tests, preventing one test from polluting the shared state and causing Mock Clients created with `debug: true` from polluting the logs when they try to resume a session found in the shared localstorage.


The-Alchemist:
@claude review

The-Alchemist (approved):

PR #9931, opened 2026-07-22
Extract useSchedulingResources hook

noahsilas:
This encapsulates fetching the appointment/slot data, managing the state and subscribing to events needed to keep it all in sync.

Implementation notes:
- To handle future multi-schedule UI, we make this hook take an array of Schedule resources instead of just one.

- Minor functional change: we're now fetching Appointment records related to any of the Schedule's actors, instead of assuming that the Schedule resource has exactly one actor and reading `schedule.actor[0]`. Medplum still recommends using one Actor per Schedule (our scheduling API endpoints like `$book` assume that this is the case), but our UI does not need to enforce that restriction to render components like calendar widgets.

- We emit one `Appointment` and one `Slot` query per input schedule. This means that each one can be cached independently in the MedplumClient query cache, and actions like selecting or unselecting a single Schedule for display don't need to re-fetch all the other Appointment/Slot resources that we have already loaded.

- We do some slightly awkward maneuvering around exhaustive-deps; we convert our input Schedule resources into stable strings to avoid re-fetching when unstable inputs are passed in. This makes the single-schedule use case (i.e. `useSchedulingResources([schedule], range)`) nice to consume without explicit memoization of that list.

For rapid iteration, this is being developed in `/examples/medplum-provider`, with the intent of moving into `/packages/react-hooks` after it feels it has stabilized.

### DIFF
--- a/examples/medplum-provider/src/components/schedule/ScheduleDetails.test.tsx
+++ b/examples/medplum-provider/src/components/schedule/ScheduleDetails.test.tsx
@@ -3,7 +3,7 @@
 import { MantineProvider } from '@mantine/core';
 import { Notifications } from '@mantine/notifications';
 import type { WithId } from '@medplum/core';
-import type { Appointment, Schedule, Slot } from '@medplum/fhirtypes';
+import type { Schedule } from '@medplum/fhirtypes';
 import { MockClient } from '@medplum/mock';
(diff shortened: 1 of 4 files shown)


codyebberson:
@claude please review

noahsilas:
@claude please re-review

codyebberson (approved):

PR #9933, opened 2026-07-23
Support filter on code string prefix in `ValueSet/$expand`

mattwiller:
Adds a new index to the `Coding` table to support prefix matching on the `code` string, with updates to `$expand` to ensure it hits the new index when filtering.  The partial index to filter out synonyms has a significant (40–50%) impact on index size and query performance: [verified locally](https://claude.ai/code/artifact/f0b45992-1b74-46cc-be07-8e4c9b68ffc7?via=auto_preview) against the full UMLS import to ensure queries remain well-indexed and overall performance is preserved


codyebberson:
@claude please review

codyebberson (approved):

PR #9934, opened 2026-07-23
Improve performance of terminology queries

mattwiller:
A collection of three loosely-related performance fixes for common terminology queries used in `ValueSet/$expand` tested locally against the imported UMLS data set using k6 to benchmark end-to-end for multiple relevant query patterns. 

## Fixes

|    | Fix                                                                                                                                                                | File                                       | Pattern improved                                          |
| --- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------ | --------------------------------------------------------- |
| A   | Emit `"t"."target" > 0` as a literal so the planner uses the partial `Coding_Property_reverse_rel_lookup_idx` and drives the recursion by nested-loop index lookup | `terminology.ts addDescendants`            | No-filter `is-a`/`descendent-of` subtree expansion        |
| B   | Below 3 chars, match `code` only (skip `display ILIKE '%x%'`, which can't use the trigram GIN)                                                                     | `expand.ts applyExpansionFilters`          | 1–2 char typeahead over a whole CodeSystem                |
| C   | Bounded candidate count picks `ancestor` (selective) vs `descendant` materialize (broad) per include                                                               | `expand.ts chooseParentFilterStrategy`     | ≥3 char typeahead over an `is-a`/`descendent-of` include  |

---

## Benchmark comparison (median ms, 20 iters × 1 VU)

| Group          | Case                                                     | before | after |       Δ |     change |
| -------------- | -------------------------------------------------------- | -----: | ----: | ------: | ---------: |
| A no-filter    | Condition, no filter n=1000                              |  102.7 |   8.9 |   −93.8 | **−91.3%** |
| A no-filter    | Procedure, no filter n=1000                              |  105.5 |   7.5 |   −98.0 | **−92.8%** |
| A no-filter    | Encounter, no filter n=1000                              |   95.7 |   7.1 |   −88.5 | **−92.6%** |
| A no-filter    | US Core Condition, no filter n=500 (multi-system)        |   89.1 |   4.8 |   −84.3 | **−94.7%** |
| B short-filter | SNOMED, filter=di (whole-system, 2-char)                 |  404.2 |   5.6 |  −398.6 | **−98.6%** |
| B short-filter | SNOMED, filter=dia (whole-system, 3-char)                |   98.2 |  96.2 |    −2.0 |      −2.0% |
| B short-filter | Condition, filter=di (is-a guard) ‡                      |    4.4 |   2.5 |    −1.8 | **−42.3%** |
| B short-filter | Procedure, filter=ap (is-a guard) ‡                      |    3.2 |   2.0 |    −1.1 | **−35.2%** |
| C selective    | Condition, filter=diabetes                               |  108.7 | 113.2 |    +4.6 |      +4.2% |
| C selective    | Procedure, filter=appendectomy                           |   23.4 |  29.9 |    +6.5 |     +27.8% |
| C selective    | US Core Condition, filter=diabetes n=1000 (multi-system) |  373.4 | 382.4 |    +9.0 |      +2.4% |
| C selective    | US Core Condition, filter=appendicitis n=1000            |   74.8 |  92.4 |   +17.5 |     +23.4% |
| C selective    | US Core Condition, filter=melanoma n=1000                |  316.6 | 325.4 |    +8.8 |      +2.8% |
| C broad        | Condition, filter=dia                                    | 1431.6 | 460.7 |  −970.9 | **−67.8%** |
| C broad        | Condition, filter=dis                                    | 7084.4 | 701.9 | −6382.5 | **−90.1%** |
| C broad        | Condition, filter=disorder                               | 5646.3 | 732.8 | −4913.5 | **−87.0%** |

The `allSnomed` cases isolate Fix B: a whole-CodeSystem include (all of SNOMED, ~1M codes, no `is-a` filter)
runs the text predicate directly. The 2-char case shows the fix — **404 → 5.6 ms** (−98.6%), a baseline Seq Scan
vs. a code-only index probe.  The 3-char case confirms Fix B leaves the ≥3-char trigram path unchanged
(98 → 96 ms)

‡ **Caveat:** the `cond/proc filter=di/ap` (is-a guard) rows do **not** exercise Fix B — an `is-a` include with
a <3-char filter is short-circuited to an empty result (the "minimum filter length" guard
predates the baseline). Kept for coverage of that guard path.

---

### Fix A — reverse-index-driven subtree materialization

No-filter `is-a` expansion of Clinical finding (132k descendants), `LIMIT 1001`.

|                  | before (no `target > 0`)                                                                                                                           | after (`target > 0` literal)                                                                                                                                            |
| ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Recursion driver | **Hash Join**, `Index Only Scan on Coding_Property_full_idx` scanning **347,669 rows/level** (all `property=362`) then hash-probing the work table | **Nested Loop**, `Index Only Scan on Coding_Property_reverse_rel_lookup_idx`, `Index Cond: (target = t2.id AND property = 362)` — one indexed lookup per work-table row |
| Shared buffers   | **107,744**                                                                                                                                        | **4,266** (~25× fewer)                                                                                                                                                  |
| Execution time   | **232.5 ms**                                                                                                                                       | **7.8 ms** (~30×)                                                                                                                                                       |

The partial index `Coding_Property_reverse_rel_lookup_idx` has predicate `target > 0`; the planner will only use
it if it can prove every row satisfies that predicate. The predicate must be literal to allow the planner to prove the predicate and switch to a parameterized nested-loop index probe.

### Fix B — code-only predicate below 3 characters

Whole-CodeSystem typeahead, filter `di`, `LIMIT 11`.

|                | before (`code = 'di' OR display ILIKE '%di%'`)                                                  | after (`code = 'di'`)                                                                             |
| -------------- | ----------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
| Scan           | **Parallel Seq Scan** on `Coding`, ~90,586 rows/worker surviving, **625,285 removed by filter** | **Index Scan** on `Coding_system_code_display_synonymOf_idx`, `Index Cond: (system, code = 'di')` |
| Shared buffers | 30,556                                                                                          | 4                                                                                                 |
| Execution time | **508 ms**                                                                                      | **0.15 ms** (~3400×)                                                                              |

A `%di%` substring needs a full 3-char trigram to use the `pg_trgm` GIN index, so at 2 chars the `ILIKE`
degrades to a sequential scan of the whole code system. Dropping the `display` branch below 3 chars keeps the
query on the `(system, code)` index.

End-to-end this matches the `allSnomed filter=di` k6 case: **404 → 5.6 ms** median (−98.6%). The 3-char `allSnomed filter=dia` case is flat (98 → 96 ms) — Fix B does not touch the ≥3-char path.

### Fix C — cost-based ancestor/descendant branch

Broad typeahead `dis` over `is-a 404684003`, `LIMIT 11`. ~113k candidate rows match the trigram filter.

|                | before (correlated `EXISTS(ancestor walk)`)                                                               | after (materialize subtree, filter within)                                                                         |
| -------------- | --------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| Shape          | Bitmap Heap Scan → **SubPlan (recursive `cte_ancestors`) executed 126,846 times**, once per candidate row | `cte_descendants` materialized **once** (132,174 rows) via the reverse-index nested loop, then `CTE Scan` + filter |
| Shared buffers | **31,080,685**                                                                                            | **1,500,435** (~20× fewer)                                                                                         |
| Execution time | **7,298 ms**                                                                                              | **814 ms** (~9×)                                                                                                   |

The branch decision is a bounded candidate count (`COUNT(*)` over a `LIMIT 2001` subquery):
`Seq Scan … LIMIT 2001` → **5.5 ms**, 670 buffers. `dis` returns 2001 (capped) > 2000 threshold → `descendant`.
This tiny probe (added to every ≥3-char is-a typeahead) is what routes broad filters away from the pathological case of walking large result sets.  On selective filters the probe returns quickly and the query stays on the (cheaper)
ancestor path, shown as +4–18 ms overhead seen on the selective/`melanoma`/`diabetes` cases, the tradeoff for
avoiding multi-second latency in the worst cases.

---

## Summary

- **Fix A** is a clean ~90% cut on every no-filter `is-a` expansion, with 25× fewer buffers — pure plan
  improvement, no downside.
- **Fix B** cuts whole-system 2-char typeahead from **404 → 5.6 ms** end-to-end (−98.6%; ~3400× at the SQL
  level: seq scan → index), and leaves the ≥3-char trigram path unchanged (98 → 96 ms).
- **Fix C** eliminates the multi-second broad-typeahead tail (`dis` 7.1 s → 0.70 s, 31M → 1.5M buffers). Cost:
  a ~5 ms bounded-count probe per ≥3-char is-a typeahead and +4–18 ms on selective filters. Good trade.
- No pattern regressed materially. The only added cost is small, bounded probe overhead on selective typeahead.

### DIFF
--- a/packages/server/src/fhir/operations/expand.test.ts
+++ b/packages/server/src/fhir/operations/expand.test.ts
@@ -15,7 +15,9 @@ import request from 'supertest';
 import { initApp, shutdownApp } from '../../app';
 import { loadTestConfig } from '../../config/loader';
 import { createTestProject, initTestAuth, withTestContext } from '../../test.setup';
-import { addExpansionItems } from './expand';
+import { repoAccess } from '../repository/access-tracker';
+import type { PgQueryable } from '../sql';
+import { addExpansionItems, countCandidatesBounded, expansionQuery, hydrateCodeSystemProperties } from './expand';
 
 describe('Expand', () => {
   const app = express();
@@ -821,6 +823,69 @@ describe('Expand', () => {
 
       const system = codeSystem.url;
       expect(expansion.contains).toContainExactly([{ system, code: 'PET', display: 'pet' }]);
+    });
+  });
+
+  describe('Cost-based parent-filter strategy', () => {
+    const system = 'http://example.com/CodeSystem/' + randomUUID();
+    const codeSystem: CodeSystem = {
+      resourceType: 'CodeSystem',
+      status: 'active',
+      content: 'example',
+      url: system,
+      hierarchyMeaning: 'is-a',
+      concept: [
+        {
+          code: 'PAR',
+          display: 'parent alpha',
+          concept: [
+            { code: 'CHD', display: 'child alpha' },
+            { code: 'PET', display: 'pet beta' },
+          ],
+        },
+      ],
+    };
+    let stored: WithId<CodeSystem>;
+    let db: PgQueryable;
+
+    beforeAll(async () => {
+      const res = await request(app)
+        .post(`/fhir/R4/CodeSystem`)
+        .set('Authorization', 'Bearer ' + accessToken)
+        .set('Content-Type', ContentType.FHIR_JSON)
+        .send(codeSystem);
+      expect(res).toHaveStatus(201);
+      stored = res.body as WithId<CodeSystem>;
+
+      await withTestContext(async () => {
+        const { repo } = await createTestProject({ withRepo: true });
+        db = repo.getDatabaseClient(repoAccess.sqlRead('CodeSystem', { source: 'test' }));
+        await hydrateCodeSystemProperties(db, stored);
+      });
+    });
+
+    test('countCandidatesBounded returns min(actual, limit)', async () => {
+      // 'alpha' matches PAR + CHD (2). PET does not.
+      expect(await countCandidatesBounded(db, stored, 'alpha', 10)).toBe(2);
+      expect(await countCandidatesBounded(db, stored, 'alpha', 1)).toBe(1); // bounded
+      expect(await countCandidatesBounded(db, stored, 'zzz', 10)).toBe(0);
+    });
+
+    test('descendant and ancestor strategies return identical members', async () => {
+      const include = { system, filter: [{ property: 'concept', op: 'is-a' as const, value: 'PAR' }] };
+      const params = { filter: 'alpha' };
+
+      const ancestorQuery = expansionQuery(include, stored, params, 'ancestor');
+      const descendantQuery = expansionQuery(include, stored, params, 'descendant');
+      if (!ancestorQuery || !descendantQuery) {
+        throw new Error('expected both strategies to build a query');
+      }
+      const ancestorRows = await ancestorQuery.execute(db);
+      const descendantRows = await descendantQuery.execute(db);
+
+      const codes = (rows: { code: string }[]): string[] => rows.map((r) => r.code).sort();
+      expect(codes(ancestorRows)).toEqual(['CHD', 'PAR']);
+      expect(codes(descendantRows)).toEqual(codes(ancestorRows));
     });
   });
 
--- a/packages/server/src/fhir/operations/expand.test.ts
+++ b/packages/server/src/fhir/operations/expand.test.ts
@@ -1456,13 +1456,62 @@ describe('Expand', () => {
     expect(vsRes).toHaveStatus(201);
 
     const res = await request(app)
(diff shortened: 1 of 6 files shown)


codyebberson:
Nice 👍 

@claude  - please review

codyebberson (approved):
lgtm 👍

PR #9936, opened 2026-07-23
fix(server): enforce maximum password length across all endpoints

codyebberson:
Cap password length at 72 bytes (bcrypt's effective limit) consistently across every password entry point, and add a defense-in-depth check at the bcrypt hashing chokepoint so any current or future caller is protected regardless of route.

Previously only the newuser endpoint enforced a maximum; setpassword, changepassword, login, and the admin super/project/invite endpoints only checked a minimum. bcrypt silently truncates past 72 bytes, so unbounded input was both a footgun and wasted work.

Adds shared MIN_PASSWORD_LENGTH/MAX_PASSWORD_LENGTH constants so the values cannot drift apart again.

Addresses external report of unbounded password length. Note bcrypt only consumes the first 72 bytes and requests are already body-size capped, so there was no actual DoS; this is input-validation hardening.

### DIFF
--- a/packages/server/src/admin/project.ts
+++ b/packages/server/src/admin/project.ts
@@ -10,6 +10,7 @@ import { resetPassword } from '../auth/resetpassword';
 import { setPassword } from '../auth/setpassword';
 import type { MfaMethod } from '../auth/utils';
 import { getEnrolledMfaMethods } from '../auth/utils';
+import { MAX_PASSWORD_LENGTH, MIN_PASSWORD_LENGTH } from '../constants';
 import { getAuthenticatedContext } from '../context';
 import { sendEmail } from '../email/email';
 import { reconcileDefaultAccessPolicy } from '../fhir/accesspolicy';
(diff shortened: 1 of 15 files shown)


packages/server/src/admin/project.ts
```diff
@@ -31,7 +32,9 @@ projectAdminRouter.post(
   '/setpassword',
   [
     body('email').isEmail().withMessage('Valid email address is required'),
-    body('password').isLength({ min: 8 }).withMessage('Invalid password, must be at least 8 characters'),
+    body('password')
+      .isByteLength({ min: MIN_PASSWORD_LENGTH, max: MAX_PASSWORD_LENGTH })
```
noahsilas: It might be a good idea to also leave `isLength({ min: 8 })`, just to preempt future security reports saying that their password with multi-byte characters is accepted and violates the security policy. (I just _know_ some joker is eventually going to try to set their password to "🔑🔓" and then complain that it's not long enough and was accepted)

There's an argument that for maximum correctness we should draw the "characters" vs. "bytes" distinction everywhere, but IMO that's more confusing and not likely to be important in practice for anyone other than future pen-testers. 🤷

noahsilas (approved):

packages/server/src/admin/project.ts
codyebberson: Yep, good call, agreed, will fix

codyebberson:
@claude please review

PR #9941, opened 2026-07-23
Improve AIRealTimeQuestionnaireForm responsiveness 

techdavidy:
1. @medplum/react: AIRealTimeQuestionnaireForm silence debounce cut from 3s to 500ms (configurable via new silenceDebounceMs prop), so dictation flushes to the AI ~2.5s sooner on every pause.

2. ai-realtime-questionnaire bot: the LLM now returns a flat {linkId: value} delta instead of authoring FHIR — the bot builds the QuestionnaireResponse deterministically (option/$expand-verified codings with carry-over reuse, createReference lookups, form-identical Date-based date/time parsing, legacy prompt kept as fallback) — measured ~2× faster on multi-field turns and 10/10 correct on repeating groups vs 3/10 for the old bot.


techdavidy:
@claude review

codyebberson (approved):

PR #9942, opened 2026-07-23
feat(agent): add `arBehavior` channel config for application-level rejects

ThatOneBro:
## Summary

Adds a per-channel **`arBehavior`** (endpoint URL query param) and an agent-wide **`channelArBehavior`** (`Agent.setting`) that control what a durable-queue channel does when a message lands in the terminal `rejected` state — an **application reject** (MSA-1 of `AR`/`CR` from the far side, or a permanent server 4xx).

**Only takes effect when the durable queue is on** (`durableQueue=true`). Configuring it with the queue off logs a warning and has no effect.

| Value | Behavior | When to use |
| --- | --- | --- |
| `continue` (**default**) | Keep draining past the rejected message. | You **cannot control when the far side sends `AR`/`CR`** and a single reject must not stall the whole pipe. |
| `pause` | Stop draining the channel while a rejected row exists — no later message is processed past the rejected one. | Ordering matters and a later message may **depend** on the rejected one. |

`continue` is the default because it matches the pre-existing durable-queue behavior (see [Behavior change](#behavior-change) below); `pause` is opt-in.

**Precedence:** `arBehavior` URL param → `channelArBehavior` setting → `continue`.

## How pause works

The pause is enforced in the SQL claim itself (`CLAIM_NEXT_PAUSE_ON_REJECT`: `claimNext` refuses to hand out work while the channel has a `rejected` row), so:

- It **survives an agent restart** — durable, not in-memory.
- The operator **resumes by clearing the rejected row** (reclassify its `state` or delete it); no restart needed. The channel then auto-resumes on the next claim.
- The `NOT EXISTS` gate rides the existing `idx_inbound_channel_state_id` index (guarded in `query-plan.test.ts`), so it adds an index seek, never a scan.

## Changes

- `queue/types.ts` — `ArBehavior` enum, `DEFAULT_AR_BEHAVIOR` (`continue`), `isArBehavior` guard.
- `queue/queries.ts` + `queue/durable-queue.ts` — `CLAIM_NEXT_PAUSE_ON_REJECT` and a `pauseOnReject` arg on `claimNext`.
- `queue/worker.ts` — `arBehavior` option + `setArBehavior`; the loop passes `pauseOnReject`; an operator-facing log when a reject pauses the channel.
- `hl7.ts` — `parseArBehavior`, per-channel resolution (`resolveAndApplyArBehavior`), queue-off warning, `getArBehavior`.
- `app.ts` — reads the `channelArBehavior` setting + `getChannelArBehaviorDefault`.
- Docs: row in `packages/docs/docs/agent/features.md`; config tables in `DURABLE_QUEUE_ARCHITECTURE.md`.
- Tests: worker pause/continue/resume, `parseArBehavior`, and a query-plan guard.

## Behavior change

**None.** The default (`continue`) matches the pre-existing durable-queue behavior, so existing channels behave exactly as before; `pause` is strictly opt-in.

Verified against the base commit: the worker loop's `claimNext` had no pause gate and `handleFailure` simply marked the row `rejected` and returned, so the next iteration claimed the next queued row. This held in **`guaranteedDelivery` mode** too — there the only code that reaches terminal `rejected` is `UpstreamRejected` (MSA-1 `AR`/`CR`, the sole member of `GUARANTEED_TERMINAL_CODES`), and such a reject continued the pipe. Making `continue` the default preserves that.

To halt a channel on a reject, opt in with `arBehavior=pause` (per channel) or `channelArBehavior: pause` (agent-wide).

## Testing

- `packages/agent`: `worker.test.ts`, `hl7.test.ts`, `query-plan.test.ts`, `integration.test.ts`, `app.test.ts` all pass.
- `tsc --noEmit`, `eslint`, and `prettier --check` clean on all changed files.


The-Alchemist:
@claude review

PR #9943, opened 2026-07-23
Additional database configuration options

mattlong:
We've deviated from the the node-postgres names before, e.g. `maxConnections` instead of `max`. We might want to do the same for the timeout options:
* `idleTimeoutMs` instead of `idleTimeoutMillis`
* `connectionTimeoutMs` instead of `connectionTimeoutMillis`

**UPDATE:** Renamed as described above


mattlong:
@claude review

mattlong:
@claude take another look

ThatOneBro (approved):
Nice, LGTM

PR #9944, opened 2026-07-23
`$rate-limits` endpoint respect UserConfiguration override

maddyli:


packages/server/src/fhir/fhirquota.ts
```diff
@@ -26,12 +26,38 @@ export const FhirQuotaCost = {
...
+ * @returns The effective project-wide FHIR quota points.
+ */
+export function getProjectFhirQuota(project: AuthState['project'] | undefined): number {
+  const perProjectLimit = project?.systemSetting?.find((s) => s.name === 'totalFhirQuota')?.valueInteger;
+  return perProjectLimit ?? getUserFhirQuota(project) * 10;
+}
 
+export function getFhirQuotaConfig(authState: AuthState): FhirQuotaConfig {
+  const { project, userConfig } = authState;
+  const userLimit = getUserFhirQuota(project, userConfig);
```
ThatOneBro, line 58: Because we use this `userLimit` below to calculate `projectLimit` in the absence of a `perProjectLimit`, I think we need to tweak this function to calculate the fallback below by multiplying `projectPerUserLimit` * 10 since it should not be a function of the user's `UserConfiguration`

ThatOneBro (commented):
Looks good for the most part, one comment about the `getFhirQuotaConfig`

ThatOneBro (approved):

PR #9946, opened 2026-07-23
Add resource type header to resource list page

### LINKED ISSUE #9919: Resource type header on resource page

Filupnot:
Fixes #9919 

This PR displays the resource type as a page heading above the search results, so each resource list has a clear title. 

Screenshot of example with **ProjectMembership**:

image


noahsilas (approved):
Nice 👍

techdavidy:
Looks good, do we have any input from design ?

Filupnot:
Kevin did a review & pushed a quick sizing update, considering this good to merge.

kevinwadeshaw (approved):
Added small changes to make header the same size it's rendered on other subpages (batch/config/security)—otherwise lgtm!

PR #9956, opened 2026-07-24
Create `@medplum/storybook` package

noahsilas:
We're looking to split some heavy-weight scheduling components into a separate package from `@medplum/react` (we want to bundle some heavy-weight dependencies without affecting existing users).

When we do that, we will want to be able to show stories from the new package in the same Storybook instance. This can also help us start providing stories in packages like `health-gorilla-react` and embedding them into our shared storybook, or do other similar co-location into more appropriate packages.

To that end, we start by extracting a new `@medplum/storybook` package which can import stories from other packages. Turborepo even suggests a pattern for this: using relative paths in `.storybook/main` to bring in files from elsewhere in the monorepo[^1].

We add a step in `scripts/cicd-deploy` to build the storybook and push the result to an S3 bucket when there are changes in the `packages/storybook` or `packages/react` directories (more to come later). While we test this deployment path, that bucket is set up to serve at storybook2.medplum.com, leaving the existing vercel deployment in place.

[^1]: https://turborepo.dev/docs/guides/tools/storybook for details.


packages/react/package.json
```diff
@@ -91,18 +83,13 @@
     "@types/react": "19.2.17",
     "@types/react-dom": "19.2.3",
     "@vitejs/plugin-react": "6.0.4",
-    "@vueless/storybook-dark-mode": "10.0.8",
-    "chromatic": "18.1.0",
     "dotenv": "17.4.2",
     "jsqr": "1.4.0",
-    "postcss": "8.5.22",
-    "postcss-preset-mantine": "1.18.0",
```
noahsilas, line 99: `postcss` in here is sort of an interesting scenario; `vite` includes `postcss` as a transitive dependency, so this direct dependency is not actually used. It does sort of act to document the shared version that vite will use, but could end up out of sync.

Further, `vite` is used for the storybook build, but `esbuild` is used for the regular package publish. So our published package includes some surprising CSS tokens like `$mantine-breakpoint-xs` that aren't fully valid, and need to themselves be passed through a build step to be translated into real values (such as the ones in our `app` or `provider` appliations) to work most completely.

In this commit I'm removing `postcss` as a dependency in `@medplum/react`, making this current behavior a little less obscure - hopefully other readers don't fall into the same confusion I did in assuming that `postcss` was being applied during the esbuild step somehow without it listed here 😅

packages/storybook/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "@storybook/builder-vite": "10.5.4",
+    "@storybook/react": "10.5.4",
+    "@storybook/react-vite": "10.5.4",
+    "@types/react": "19.2.17",
+    "@types/react-dom": "19.2.3",
+    "@vueless/storybook-dark-mode": "10.0.8",
+    "chromatic": "18.1.0",
+    "postcss": "8.5.22",
+    "postcss-preset-mantine": "1.18.0",
+    "postcss-simple-vars": "7.0.1",
```
noahsilas, line 68: This looks like a new dependency, but it's not. We've been importing from this module directly in our `postcss.config.mjs` file, which worked due to it being a transitive dependency of `postcss-preset-mantine`. Let's make that dependency explicit and pin the version here.

Filupnot:
Do we need to add this new package to `scripts/publish.sh` (under `PACKAGES`)? Since `examples/medplum-provider/package.json` declares this new package as a dependency, I believe running `npm install` will fail.

noahsilas:
> Do we need to add this new package to `scripts/publish.sh` (under `PACKAGES`)? Since `examples/medplum-provider/package.json` declares this new package as a dependency, I believe running `npm install` will fail.

To try and make reviewing each piece a little smaller/easier, I've trimmed this PR down to focus purely on the storybook extraction, and I'll follow up with creating the `react-scheduling` package and bringing stories from it in as a separate follow up.

That makes this PR a little less obviously motivated, but should keep it more clearly contained, and hopefully the context in the commit message has enough info to be explanatory.

techdavidy:
@claude review

noahsilas:
@claude re-review

packages/storybook/.storybook/main.ts
```diff
@@ -7,7 +7,12 @@ import { mergeConfig } from 'vite';
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 const config: StorybookConfig = {
-  stories: ['../src/stories/Introduction.mdx', '../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
+  stories: [
```
codyebberson, line 10: Nice, this seems like the magical change, glad it was straightforward 👍

codyebberson (approved):
lgtm, let's go 🚀

PR #9961, opened 2026-07-25
refactor(server): replace @modelcontextprotocol/sdk with a native implementation

codyebberson:
I'm not actually sure if we should do this.  It was more code than I expected.  Putting this up for discussion.

--------

Replaces the `@modelcontextprotocol/sdk` dependency with a small native implementation of the
Model Context Protocol, covering only the subset Medplum actually exposes: Streamable HTTP
transport, JSON-RPC 2.0, and tool registration/discovery/invocation. The deprecated SSE transport
is removed.

Medplum used exactly three exported classes from the SDK (`McpServer`,
`StreamableHTTPServerTransport`, `SSEServerTransport`) to support five methods (`initialize`,
`notifications/initialized`, `ping`, `tools/list`, `tools/call`). Over the past year the SDK has
been a disproportionate source of vulnerability alerts, dependency updates, API churn, and
transitive package updates relative to that usage.

The whole protocol now lives in one file, `packages/server/src/mcp/protocol.ts` (371 lines).
Every pre-existing file shrinks: **+88 / −147** across the files this touches.

## Motivation

- Eliminate a high-churn dependency
- Reduce vulnerability exposure and transitive dependency count
- Gain complete ownership of the protocol implementation
- Simplify debugging and long-term maintenance

This is deliberately **not** an attempt to implement the entire MCP specification.

## Dependency impact

Removes 9 packages from the dependency tree (3073 → 3064):

```
@modelcontextprotocol/sdk   eventsource-parser   json-schema-typed
@hono/node-server           express-rate-limit   pkce-challenge
eventsource                 hono                 zod-to-json-schema
```

`zod` is also dropped from `packages/server/package.json`. The MCP server was its only consumer;
tool schemas are now declared as literal JSON Schema, which `tools/list` emits verbatim and which
removes the zod → JSON Schema conversion layer entirely.

## Architecture

Request-oriented rather than connection-oriented. Each HTTP request is independently
authenticated, authorized, dispatched, and completed. There is no `Mcp-Session-Id`, no
server-side session state, and no persistent transport object.

```
HTTP Request → Streamable HTTP handler → JSON-RPC dispatcher → tool registry → tool handler
```

Authentication and authorization are unchanged: `mcpRouter` still runs `authenticateRequest`, and
the MCP layer receives an already-authenticated request context. No new auth, authorization, or
validation framework is introduced.

## Details

New:

- `packages/server/src/mcp/protocol.ts` — JSON-RPC 2.0 parsing and response construction, the MCP
  initialization handshake, protocol version negotiation, the tool registry, the method dispatcher,
  and the Streamable HTTP endpoint.
- `packages/server/src/mcp/protocol.test.ts` — 54 protocol conformance and unit tests. Runs against
  a bare Express app with no `initApp()` or database, so it completes in about a second.

Modified:

- `server.ts` — unchanged apart from two import lines and three `inputSchema` literals converted
  from zod to JSON Schema. The `fhir-request` handler body, including the GHSA-fjgc-c3pj-xx2c SSRF
  guard, is untouched.
- `routes.ts` — `/mcp/stream` now delegates to `getMcpServer().handleRequest(req, res)`. The SSE
  plumbing (Redis pub/sub subscriber, heartbeat ping, dummy request/response shims) is deleted.
- `routes.test.ts` — the SDK client is swapped for a raw JSON-RPC helper. The CRUD lifecycle and
  SSRF assertions (steps 1–10) are unchanged.

The tool registry keeps the SDK's `registerTool(name, config, handler)` shape and the
`new McpServer({ name, version })` constructor deliberately, so the call sites in `server.ts` did
not have to be rewritten.

## SSE removal

`/mcp/sse` now returns `410 Gone` with a message pointing at `/mcp/stream`, rather than 404, so
that any remaining client gets actionable guidance. Server logs for the previous seven days:

| Endpoint      | Requests |
| ------------- | -------: |
| `/mcp/stream` |  ~20,000 |
| `/mcp/sse`    |        3 |

SSE was already marked deprecated upstream. Note this is a hard cutover with no shadow period —
those 3 requests will start failing on deploy.

## Error handling

| Category  | Examples                                                      | Result                                   |
| --------- | ------------------------------------------------------------- | ---------------------------------------- |
| Transport | Wrong HTTP method, bad auth, malformed JSON, wrong media type | HTTP status code                         |
| JSON-RPC  | Malformed message, unknown method, invalid params             | JSON-RPC error response                  |
| Tool      | A tool handler throws                                         | Successful response with `isError: true` |

A tool execution failure is not a protocol failure, which is why it lands in the third row. This
matches MCP semantics and the SDK's previous behavior.

Malformed JSON is rejected by the existing Medplum body-parser middleware before MCP sees the
request, so it surfaces as an HTTP 400 with a FHIR `OperationOutcome` rather than a JSON-RPC parse
error. This is unchanged from the SDK implementation.

## Three deliberate deviations from the original design doc

1. **Five protocol versions, not one.** The doc proposed pinning to `2025-11-25` and returning an
   initialization error otherwise. The SDK accepted five versions and defaulted to `2025-03-26`
   when the header was absent; most shipping clients still send `2025-06-18` or older, and the MCP
   specification says a server _should_ respond with a version it supports rather than fail. The
   implementation accepts the same five versions, echoes the client's choice back, and offers
   `LATEST_PROTOCOL_VERSION` for anything unrecognized.

2. **Initialization ordering is not enforced.** The doc's Architecture section (stateless,
   request-oriented) and Initialization section ("no additional methods may be called before
   initialization completes") are in conflict — with no session state there is nowhere to record
   that `initialize` happened. Statelessness was kept. Enforcing the ordering would require sticky
   routing or shared storage across instances for no security benefit, since authentication and
   authorization run on every request regardless. This is documented in the README.

3. **JSON-RPC batching is retained.** Batching was removed from MCP in revision `2025-06-18`, but
   arrays are still accepted so that clients pinned to an earlier revision keep working.

## Testing

```bash
npm test --workspace=@medplum/server -- src/mcp/
```

54 tests covering initialization, protocol negotiation across all five supported versions,
malformed JSON, malformed JSON-RPC, unknown methods, tool registration, duplicate registration,
tool discovery, successful invocation, tool failures, argument validation, and authorization
failures.

`src/app.test.ts` also passes unchanged, `tsc --noEmit` is clean, and the production bundle
contains no SDK references.

Manual verification with MCP Inspector:

```bash
npx @modelcontextprotocol/inspector
# Transport Type: Streamable HTTP
# URL: http://localhost:8103/mcp/stream
```

## Follow-ups (intentionally not in this PR)

Two pre-existing issues were found while porting. Both are left alone so this diff stays scoped to
replacing the SDK:

1. `proxy.delete(fhirUrl, body)` in `server.ts` passes the client-supplied body where
   `MedplumClient.delete()` expects `MedplumRequestOptions`. Harmless in practice, since FHIR
   DELETE has no body, and it only became visible because native typing removed zod's `any`.
2. Step 10 of the `routes.test.ts` CRUD sequence asserts
   `expect(protoRel.resourceType).toBe('OperationOutcome')`, which is satisfied by the helper's own
   `catch` → `normalizeOperationOutcome(err)` fallback regardless of what the server returns. It
   passes, but it is not testing what it appears to. (The real behavior: `concatUrls()` strips
   leading slashes, so `//host/path` becomes an ordinary relative path under the FHIR base URL and
   404s on-origin — it never reaches an external host.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9724,18 +9724,6 @@
         "react-dom": ">=16.8.0"
       }
     },
-    "node_modules/@hono/node-server": {
-      "version": "1.19.14",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.14.tgz",
-      "integrity": "sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw==",
-      "license": "MIT",
-      "engines": {
-        "node": ">=18.14.1"
-      },
-      "peerDependencies": {
-        "hono": "^4"
-      }
-    },
     "node_modules/@humanfs/core": {
       "version": "0.19.2",
       "resolved": "https://registry.npmjs.org/@humanfs/core/-/core-0.19.2.tgz",
@@ -11150,46 +11138,6 @@
         "ajv": "~8.18.0",
         "jju": "~1.4.0",
         "resolve": "~1.22.2"
-      }
-    },
-    "node_modules/@modelcontextprotocol/sdk": {
-      "version": "1.29.0",
-      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/sdk/-/sdk-1.29.0.tgz",
-      "integrity": "sha512-zo37mZA9hJWpULgkRpowewez1y6ML5GsXJPY8FI0tBBCd77HEvza4jDqRKOXgHNn867PVGCyTdzqpz0izu5ZjQ==",
-      "license": "MIT",
-      "dependencies": {
-        "@hono/node-server": "^1.19.9",
-        "ajv": "^8.17.1",
-        "ajv-formats": "^3.0.1",
-        "content-type": "^1.0.5",
-        "cors": "^2.8.5",
-        "cross-spawn": "^7.0.5",
-        "eventsource": "^3.0.2",
-        "eventsource-parser": "^3.0.0",
-        "express": "^5.2.1",
-        "express-rate-limit": "^8.2.1",
-        "hono": "^4.11.4",
-        "jose": "^6.1.3",
-        "json-schema-typed": "^8.0.2",
-        "pkce-challenge": "^5.0.0",
-        "raw-body": "^3.0.0",
-        "zod": "^3.25 || ^4.0",
-        "zod-to-json-schema": "^3.25.1"
-      },
-      "engines": {
-        "node": ">=18"
-      },
-      "peerDependencies": {
-        "@cfworker/json-schema": "^4.1.1",
-        "zod": "^3.25 || ^4.0"
-      },
-      "peerDependenciesMeta": {
-        "@cfworker/json-schema": {
-          "optional": true
-        },
-        "zod": {
-          "optional": false
-        }
       }
     },
     "node_modules/@module-federation/error-codes": {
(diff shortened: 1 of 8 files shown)


codyebberson:
Not worth it

PR #9962, opened 2026-07-25
Upgrade react-router to 8.3.0

codyebberson:
This was motivated by some CVEs in `react-router`.  The vulnerabilities are only in RSC (React Server Components), which we do not use, so we are unaffected, but the CVEs persist and will require ongoing documentation until they are resolved.

Our normal upgrade process did not pick up `react-router` v8 because our `engines` string was `>=22.18.0` and `react-router` requires `>=22.22.0`.

This PR is kinda borked due to Docusaurus incompatibility.  Docusaurus is planning to upgrade to `react-router` v8 in Docusaurus v4 (see https://github.com/facebook/docusaurus/issues/11719).

We could let it be, and accept multiple versions of `react-router` in `node_modeules`.  That might be the safer approach.


packages/docs/package.json
```diff
@@ -63,8 +63,9 @@
     "react-dom": "19.2.8",
     "marked": "18.0.7",
     "react-intersection-observer": "10.1.0",
-    "react-router": "7.18.1",
```
codyebberson: These were always wrong, and it's a minor miracle that this worked.

Docusaurus specifically uses react-router v5.  They plan to upgrade to v8 in Docusaurus 4: https://github.com/facebook/docusaurus/issues/11719

packages/react/src/ScrollToTop/ScrollToTop.test.tsx
```diff
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Link, MemoryRouter, Route, Routes } from 'react-router-dom';
+import { Link, MemoryRouter, Route, Routes } from 'react-router';
```
codyebberson, line 3: `react-router-dom` is deprecated, it's now just a wrapper around `react-router`.

package.json
```diff
@@ -61,6 +61,6 @@
   },
   "packageManager": "npm@10.9.8",
   "engines": {
-    "node": ">=22.18.0 <23.0.0 || >=24.2.0 <25.0.0"
+    "node": ">=22.22.0 <23.0.0 || >=24.2.0 <25.0.0"
```
codyebberson, line 65: This was necessary for react-router v8.

mattlong (approved):

PR #9966, opened 2026-07-25
feat(react): spotlight quick actions and keyboard shortcut hints

kevinwadeshaw:
This PR refactors spotlight-based Search modal for Provider app to change the default state to include quick actions—simulating a globally-accessible command pallette. The exsting shortcuts from the default state and transformed into a single-row footer for easy reference. 

This is a global change that should allow for all apps built with Medplum to be able to populate their own quick actions. For Provider app, we are populating New Patient Intake, New Message, New Task and Send a Fax—with Import SMART Health Link action included in a subsequent PR (#9967)

---

### Reference Images:

| Before      | After |
| ----------- | ----------- |
|   Screenshot 2026-08-04 at 1 02 54 PM   | 631347072-e4f2f0b4-8e54-4c80-8929-41e5105fd77e |

### DIFF
--- a/examples/medplum-provider/src/App.tsx
+++ b/examples/medplum-provider/src/App.tsx
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
+import type { SpotlightActionData } from '@mantine/spotlight';
 import { getReferenceString } from '@medplum/core';
 import { useDoseSpotNotifications } from '@medplum/dosespot-react';
 import { AppShell, Loading, Logo, useMedplum, useMedplumProfile } from '@medplum/react';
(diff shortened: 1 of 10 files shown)


examples/medplum-provider/src/App.tsx
```diff
@@ -83,11 +84,40 @@ export function App(): JSX.Element | null {
...
+      id: 'action-new-patient-intake',
+      label: 'New Patient Intake',
+      leftSection: <IconUserPlus size={16} color="var(--mantine-color-dimmed)" />,
+      onClick: () => navigate('/onboarding')?.catch(console.error),
+    },
+    {
+      id: 'action-new-message',
+      label: 'New Message',
+      leftSection: <IconMail size={16} color="var(--mantine-color-dimmed)" />,
+      onClick: () => navigate('/Communication', { state: { newMessage: Date.now() } })?.catch(console.error),
```
techdavidy, line 105: We can point this pr to main.
and /new path is supported so Communication/new should work 

same for tasks and fax

packages/react/src/chat/ThreadInbox/ThreadInbox.tsx
```diff
@@ -119,6 +122,13 @@ export function ThreadInbox(props: ThreadInboxProps): JSX.Element {
     onPatientChange?.(subjectReference ? { reference: subjectReference } : undefined);
   }, [subjectReference, onPatientChange]);
 
+  // Allow the host to programmatically open the New Message dialog (e.g. from the global search).
```
techdavidy, line 125: we can drop the changes to anything other than Spotlight changes now

PR #9970, opened 2026-07-25
Follow JWT Best Practices for Audience Validation

codyebberson:
Applies two long-standing JWT and OAuth recommendations to the server's token verification paths:

- **RFC 8725 §3.9 (Use and Validate Audience)** and **§3.12 (Use Mutually Exclusive Validation Rules for Different Kinds of JWTs)** — the FHIR API path now validates the `aud` claim, and requires that a presented bearer token be an access token rather than any other server-signed JWT.
- **RFC 6749 §6** — the refresh grant now verifies a presented client secret rather than only checking that one was supplied.

No changes to the tokens the server issues. Existing access tokens continue to work unchanged.

## Background

The server signs ID tokens, access tokens, and refresh tokens with the same key and the same issuer, and all three carry a `login_id` claim. `verifyJwt` validates the issuer, algorithm, and expiry, so a valid signature establishes that the server minted the token but not which kind of token it minted. Per RFC 8725 §3.12, each kind of JWT should have validation rules that a JWT of another kind cannot satisfy.

The claims needed to distinguish them are already present:

- `aud` is the issuer on access tokens, and the client ID on ID tokens (as OpenID Connect requires).
- `refresh_secret` is present only on refresh tokens, which share the issuer audience with access tokens.

So the API path can require both, and no minted token has to change.

### `packages/server/src/oauth/utils.ts`

`getLoginForAccessToken` now requires `aud` to equal the configured issuer and rejects any token carrying a `refresh_secret` claim. Seven lines, no new imports.

### `packages/server/src/oauth/token.ts`

The refresh grant now calls the existing `validateClientIdAndSecret` helper when a `Basic` authorization header is presented, which timing-safe compares against `client.secret` and supports `client.retiringSecret` for rotation. Previously the handler verified that the client ID matched the login and that a secret was non-empty, but did not compare the secret itself.

This is the same helper already used by the authorization_code and client_credentials grants, so all three grants now authenticate clients consistently.

The `ClientApplication` read moved above the authorization header block, unchanged, because the validation needs the client resource. Two minor consequences:

- If a request presents a `Basic` header whose client ID does not match the login *and* the login's `ClientApplication` cannot be read, the error changes from `invalid_grant` / `Incorrect client` to `invalid_request` / `Invalid client`. Both are 400s.
- A login whose client has no secret configured, presented with a non-empty secret, is now rejected. An empty secret continues to take the existing path.

### DIFF
--- a/packages/server/src/oauth/middleware.test.ts
+++ b/packages/server/src/oauth/middleware.test.ts
@@ -10,7 +10,7 @@ import { initApp, shutdownApp } from '../app';
 import { getConfig, loadTestConfig } from '../config/loader';
 import { getGlobalSystemRepo } from '../fhir/repo';
 import { createTestClient, createTestProject, withTestContext } from '../test.setup';
-import { generateAccessToken, generateSecret } from './keys';
+import { generateAccessToken, generateIdToken, generateRefreshToken, generateSecret } from './keys';
 import { PROMPT_BASIC_AUTH_PARAM } from './middleware';
 
 describe('Auth middleware', () => {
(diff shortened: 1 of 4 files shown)


reshmakh (approved):

codyebberson:
@claude please review

codyebberson:
Deploying to staging...

codyebberson:
Everything looks good, existing sessions continue to work as expected

https://api.staging.medplum.dev/healthcheck

PR #9973, opened 2026-07-25
Align chained search filtering with direct searches

codyebberson:
Chained searches (`subject:Patient.name=...`) and reverse chained searches (`_has:Observation:subject:code=...`) evaluate predicates against linked resources using an `EXISTS` subquery. Those subqueries now apply the same project and access-policy filters as equivalent direct searches of the linked resource type.

This change keeps chained search behavior consistent with the rest of the search engine.

## Implementation

* Apply `addSecurityFilters(...)` to each chained search target using the appropriate table alias.
* Thread the SQL table alias through the search builder so security filters can be applied to joined tables.
* Update `addSecurityFilters`, `addProjectFilters`, and `addAccessPolicyFilters` to accept an optional table alias.
* Apply the same filtering to canonical chains.
* Add a recursion depth limit when evaluating access-policy criteria to guard against cyclic chained expressions in legacy configurations.

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -131,7 +131,14 @@ import { buildSearchExpression, searchByReferenceImpl, searchImpl } from './sear
 import { lookupTables } from './searchparameter';
 import { GLOBAL_SHARD_ID } from './sharding';
 import type { Expression, PgQueryable } from './sql';
-import { Condition, DeleteQuery, Disjunction, InsertQuery, SelectQuery } from './sql';
+import { Column, Condition, DeleteQuery, Disjunction, InsertQuery, SelectQuery } from './sql';
+
+/**
+ * Maximum nesting depth for security filters. Depth only grows when an access policy
+ * criteria contains a chained search, whose target then has its own security filters
+ * applied. Legitimate policies nest at most one or two levels deep.
+ */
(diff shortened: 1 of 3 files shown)


codyebberson:
@claude please review

packages/server/src/fhir/repo.ts
```diff
@@ -1660,13 +1668,32 @@ export class Repository extends FhirRepository implements Disposable {
...
-    // No compartment restrictions for admins.
-    if (!this.isSuperAdmin()) {
-      this.addProjectFilters(builder, resourceType);
+  addSecurityFilters(
+    builder: SelectQuery,
+    resourceType: string,
+    interaction: AccessPolicyInteraction,
+    table: string = resourceType
+  ): void {
+    // Access policy criteria may themselves contain chained searches, and each chain link
```
mattlong, line 1681: The docs have said that access policy criteria cannot use chained search for a long time; should we start the train for actually enforcing that so we can avoid complexity like this? Deprecate in `v6`, reject in `v7`?

mattlong (approved):

packages/server/src/fhir/repo.ts
codyebberson: Yes, good call.  Maybe even more aggressive if we can start logging and migrating earlier.

PR #9975, opened 2026-07-26
Align CodeSystem and ConceptMap import permission model

codyebberson:
`CodeSystem/$import` and `ConceptMap/$import` can each address their target in two ways: by `url`/`system` parameter, or by resource ID on the instance endpoint. The parameter path already required the target to belong to the caller's own project. The instance path did not apply an equivalent check.

This change applies the same rule to the instance path, so both routes resolve to the same permission model.

## Background

Terminology content imported by these operations is written to the `Coding`, `Coding_Property`, and `CodeSystem_Property` tables, keyed by the target resource's ID. These tables are not partitioned by project, so their contents are visible to every project that can resolve the owning `CodeSystem` or `ConceptMap`.

Project-scoped repositories can read terminology resources beyond their own project:

- The synthetic base FHIR R4 project is appended to every project-scoped repository (`repo.ts:259-266`) and exports `CodeSystem` (`constants.ts:15-20`), so base R4 CodeSystems are readable from any project.
- Linked projects are similarly readable, subject to their `exportedResourceType`.

The parameter path accounts for this by passing `ownProjectOnly` to `findTerminologyResource`, with existing coverage in `codesystemimport.test.ts` ("Prevents writes by Project admin to linked Project systems"). The instance path resolved its target with `repo.readResource` alone, which establishes read access but not write access. As a result, a project administrator could import into a readable resource owned by another project.

### `packages/server/src/fhir/operations/codesystemimport.ts`, `conceptmapimport.ts`

Both handlers gain the same check on the instance branch:

```ts
// Requires extended mode: `meta.project` is stripped otherwise, so this check fails closed
codeSystem = await repo.readResource<CodeSystem>('CodeSystem', req.params.id);
if (!repo.canPerformInteraction(AccessPolicyInteraction.UPDATE, codeSystem)) {
  return [forbidden];
}
```

`canPerformInteraction` performs the same project comparison the parameter path relies on (`repo.ts:2130`), and additionally evaluates the caller's access policy. Writing a terminology resource's contents now requires permission to update the resource itself.

No new helpers, exports, or reads. The remainder of the source diff is Prettier reflowing two import statements.

## Behavior changes

Three consequences worth noting in release notes:

1. **Instance-path imports into another project's resource are rejected.** This matches what the parameter path has always done, and is the intended model rather than a new restriction.

2. **The caller's access policy is now consulted.** A project administrator whose policy grants read-only access to `CodeSystem` or `ConceptMap` can no longer import into one. Callers with no access policy are unaffected. Super administrators are subject to their own access policy if one is configured; with no policy configured, behavior is unchanged.

3. **The instance path requires extended mode.** `Repository.readResource()` clears `meta.project` unless the request carries `X-Medplum: extended` (`repo.ts:2197-2205`), so without it the check has no project to compare and denies the request. `MedplumClient` sends this header by default (`client.ts:4010`), so SDK and CLI callers are unaffected; direct HTTP callers using the instance endpoint need to add it. The parameter path is unaffected.

Requests rejected by any of the above receive `403 Forbidden`.

### DIFF
--- a/packages/server/src/fhir/operations/codesystemimport.test.ts
+++ b/packages/server/src/fhir/operations/codesystemimport.test.ts
@@ -293,6 +293,113 @@ describe('CodeSystem $import', () => {
     expect(res2).toHaveStatus(400);
   });
 
+  test('Allows Super Admin to import into base CodeSystem by ID', async () => {
+    // `accessToken` is a Super Admin; base terminology maintenance must remain possible
+    const res = await request(app)
+      .get('/fhir/R4/CodeSystem?url=http://terminology.hl7.org/CodeSystem/v3-RoleCode')
+      .set('Authorization', 'Bearer ' + accessToken)
+      .send();
+    expect(res).toHaveStatus(200);
+    const baseCodeSystem = res.body.entry?.[0]?.resource as CodeSystem;
+    expect(baseCodeSystem?.id).toBeDefined();
+
+    const code = 'ZZ' + randomUUID();
+    const res2 = await request(app)
+      .post(`/fhir/R4/CodeSystem/${baseCodeSystem.id}/$import`)
+      .set('X-Medplum', 'extended')
+      .set('Authorization', 'Bearer ' + accessToken)
+      .set('Content-Type', ContentType.FHIR_JSON)
+      .send({
+        resourceType: 'Parameters',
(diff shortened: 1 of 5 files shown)


codyebberson:
@claude please review

mattwiller (approved):

PR #9977, opened 2026-07-26
fix(core): keep UUID values whole when parsing _filter expressions

### LINKED ISSUE #9950: Flaky test: server scoped subscriptions
https://github.com/medplum/medplum/pull/9949#pullrequestreview-4775026929

https://github.com/medplum/medplum/actions/runs/30109653164/job/89536521278?pr=9949

Image

ThatOneBro:
## Problem

`src/workers/subscription.test.ts > Server-scoped subscription fires across projects when enabled` fails intermittently in CI ([example run](https://github.com/medplum/medplum/actions/runs/30189316049)) with `Job not found: no subscription job for Patient/... matching Subscription/...`, and no error in the logs.

It is a real bug, not a test problem. `getSubscriptions()` searches with:

```
_filter = "_project eq <projectId> or _project pr false"
```

`Tokenizer.consumeNumber` treats "digits followed by `-`" as a date literal, so when a project UUID's first group is **all digits** (`67836554-6431-4baa-…`), the value token ends at the first letter — `67836554-6431-4` — and `parseFilterParameter` then silently discarded the leftover tokens, including the entire `or _project pr false` branch. `buildIdSearchFilter` sees a non-UUID and substitutes the nil UUID, so the query becomes:

```sql
WHERE deleted = false AND ("projectId" = '00000000-0000-0000-0000-000000000000' AND status = 'active')
```

Nothing matches, so **no** subscription fires — not the server-scoped one, and not the project-scoped one either. 8 hex digits being all-numeric is (10/16)^8 ≈ **2.3%** of projects, which matches the observed flake rate. Any `_filter` query with such a UUID value is affected, not just this code path.

Two further problems turned up while covering nested expressions:

- The fallback for unquoted values that start with digits consumed up to the next **space**, so a UUID at the end of a nested expression swallowed the closing parenthesis. `(_id eq 6783655a-6431-4baa-9243-914efcd984da)` failed to parse at all — this one predates the change above and affects every UUID, not just all-numeric ones, so `_filter` + parentheses + a UUID never worked.
- `parseFilterParameter` returned whatever it managed to parse and dropped the rest of the token stream. That is what turned the tokenizer bug into silently-wrong search results instead of an error.

## Fix

- Take the dateTime path only when the token actually looks like `YYYY-MM[-DD][time]`.
- End an unquoted digit-led value at the first character that cannot be part of a token, per the tokenizer's own `symbolRegex` (whitespace, `)` and `]` for `_filter`), rather than at the next space.
- Reject a `_filter` expression that is not fully consumed, so a future mis-tokenization cannot silently drop conditions.

`dateTimeLiterals` is enabled only by the `_filter` tokenizer, so the date-literal change is scoped to `_filter` parsing.

## Tests

- `packages/core/src/filter/parse.test.ts`
  - Flat cases: the UUID value stays whole, and date/dateTime literals still parse as before.
  - `describe.each` over both UUID shapes — all digits before the first hyphen and not, which take different paths through the tokenizer — crossed with `test.each` of nine nesting shapes: parentheses, `not (…)`, `or` connective, two parenthesized comparisons, nested connective on either side, `not ((…) or (…))`, `Patient/<uuid>` reference values, and a UUID alongside a dateTime literal. Each asserts the full parsed expression, so a dropped branch or a truncated value fails.
  - Expressions that cannot be fully parsed (trailing token, unclosed parenthesis, dangling connective) now throw.
- `packages/server/src/fhir/search.test.ts` — end-to-end, in both the plain and parenthesized forms: `_filter=_id eq 12345678-1234-4123-8123-123456789abc or _id eq <real id>` returns the real resource.

13 of the new assertions fail with the second commit reverted, and 3 fail with the first reverted.

## Verification

- A stress harness driving the exact failing path reproduced the flake in **~2.5%** of iterations before the fix and **0 of 800** after.
- Full core suite (2525 tests, including the FHIRPath and FHIR Mapping tokenizers that share this code), fhir-router, and the full server suite; eslint, prettier and tsc clean. The server suite has 109 pre-existing local failures from a DB-grant problem between parallel test files — identical counts with and without these changes, and none in `_filter`, search, or subscription tests.

## Follow-up (not in this PR)

`getSubscriptions` could skip `_filter` entirely (`_project` accepts comma-OR, and `_project pr false` is `projectId = systemResourceProjectId`), which would be simpler and faster than building and parsing a filter expression.

### DIFF
--- a/packages/core/src/filter/parse.test.ts
+++ b/packages/core/src/filter/parse.test.ts
@@ -29,6 +29,23 @@ describe('_filter Parameter parser', () => {
       "_has:Observation:patient:_id ne ''",
       { path: '_has:Observation:patient:_id', operator: Operator.NOT_EQUALS, value: '' },
     ],
+    [
+      // A UUID whose first group is all digits looks like the start of a date literal.
+      // It must be kept whole rather than truncated at the first letter.
+      'UUID with all digits before the first hyphen',
+      '_id eq 12345678-1234-4123-8123-123456789abc',
+      { path: '_id', operator: Operator.EXACT, value: '12345678-1234-4123-8123-123456789abc' },
+    ],
+    [
+      'Date literal',
+      'birthdate ge 2014-10-10',
+      { path: 'birthdate', operator: Operator.GREATER_THAN_OR_EQUALS, value: '2014-10-10' },
+    ],
+    [
+      'DateTime literal',
+      '_lastUpdated gt 2014-10-10T12:30:45.123Z',
+      { path: '_lastUpdated', operator: Operator.GREATER_THAN, value: '2014-10-10T12:30:45.123Z' },
+    ],
   ])('%s', (_, filter, expected) => {
     const result = parseFilterParameter(filter);
     assert(result instanceof FhirFilterComparison);
@@ -86,6 +103,22 @@ describe('_filter Parameter parser', () => {
     expect(right.path).toBe('birthdate');
     expect(right.operator).toBe(Operator.GREATER_THAN_OR_EQUALS);
     expect(right.value).toBe('2014-10-10');
+  });
+
+  test('Or connective with UUID value', () => {
+    // Regression: the UUID was tokenized as a date literal and truncated, which dropped the "or"
+    // and everything after it, leaving a comparison against a value that matched nothing.
+    const result = parseFilterParameter('_project eq 12345678-1234-4123-8123-123456789abc or _project pr false');
+    expect(result).toBeInstanceOf(FhirFilterConnective);
+
+    const connective = result as FhirFilterConnective;
+    expect(connective.keyword).toBe('or');
+    expect(connective.left).toMatchObject({
+      path: '_project',
+      operator: Operator.EXACT,
+      value: '12345678-1234-4123-8123-123456789abc',
+    });
+    expect(connective.right).toMatchObject({ path: '_project', operator: Operator.PRESENT, value: 'false' });
   });
 
   test('Top level parentheses', () => {
--- a/packages/core/src/filter/parse.test.ts
+++ b/packages/core/src/filter/parse.test.ts
@@ -105,20 +105,108 @@ describe('_filter Parameter parser', () => {
     expect(right.value).toBe('2014-10-10');
   });
 
-  test('Or connective with UUID value', () => {
-    // Regression: the UUID was tokenized as a date literal and truncated, which dropped the "or"
-    // and everything after it, leaving a comparison against a value that matched nothing.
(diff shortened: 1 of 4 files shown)


packages/core/src/filter/parse.test.ts
```diff
@@ -29,6 +29,23 @@ describe('_filter Parameter parser', () => {
       "_has:Observation:patient:_id ne ''",
       { path: '_has:Observation:patient:_id', operator: Operator.NOT_EQUALS, value: '' },
     ],
+    [
+      // A UUID whose first group is all digits looks like the start of a date literal.
+      // It must be kept whole rather than truncated at the first letter.
+      'UUID with all digits before the first hyphen',
+      '_id eq 12345678-1234-4123-8123-123456789abc',
+      { path: '_id', operator: Operator.EXACT, value: '12345678-1234-4123-8123-123456789abc' },
```
codyebberson, line 37: Good catch

packages/server/src/fhir/search.test.ts
```diff
@@ -3729,6 +3729,31 @@ describe.each<Project['features']>([undefined, ['range-search']])('project-scope
...
+      // The first UUID has all digits before its first hyphen, which used to be tokenized as a date
+      // literal and truncated. That silently dropped the rest of the expression -- including the
+      // "or" branch that matches the real resource.
+      const result = await repo.search({
+        resourceType: 'Patient',
+        filters: [
+          {
+            code: '_filter',
+            operator: Operator.EQUALS,
+            value: `_id eq 12345678-1234-4123-8123-123456789abc or _id eq ${patient.id}`,
```
codyebberson: Ouch 😢

codyebberson (approved):
lgtm 👍 

@claude please review

packages/core/src/filter/parse.ts
```diff
@@ -108,5 +108,12 @@ const fhirPathParserBuilder = initFhirPathParserBuilder();
 export function parseFilterParameter(input: string): FhirFilterExpression {
   const parser = fhirPathParserBuilder.construct(tokenize(input));
   parser.removeComments();
-  return new FilterParameterParser(parser).parse();
+  const result = new FilterParameterParser(parser).parse();
+  const extra = parser.peek();
+  if (extra) {
```
ThatOneBro, line 113: This is to ensure that the entire `_filter` expression is parsed at the end... Open question if this is going to break anyone today or not, but arguably it's pretty bad if entire chunks of an expression are dropped and we still return something that doesn't fully match (for example, the second half of an `and` expression 😅

PR #9985, opened 2026-07-27
feat(server): CoverageEligibilityRequest submit operation

techdavidy:
Adds CoverageEligibilityRequest/$submit
This operation is agnostic to which service it is being used to check eligibility (Stedi or Candid). 
Relies on ELIGIBILITY_SUBMIT_OPERATION_SETTING to know what custom operation to execute.


techdavidy:
@claude review

codyebberson:
https://hl7.org/fhir/R4/coverageeligibilityrequest-operation-submit.html

codyebberson (approved):

codyebberson:
I agree with @techdavidy's choice to include the entire `Bundle`, not just the `CoverageEligibilityRequest`, because IMHO it should be up to the Bot to choose what to do with the `Bundle`, and may need access to the other supporting resources.

We made the same choice in `Claim/$submit`.

> An EligibilityRequest resource or Bundle of EligibilityRequests, either as individual EligibilityRequest resources or as Bundles each containing a single EligibilityRequest plus referenced resources.

Issue #9328, opened 2026-05-28
Provider: Add recurring availability management


noahsilas:
As a Practitioner, I should be able to manage my recurring availability for scheduling.

APIs like [`$find`](https://www.medplum.com/docs/scheduling/appointment-find) can have availability be set for all participating practitioners by setting it on the originating `HealthcareService` resource, but as an individual Practitioner I want to set more specific recurring availability.

I should have UI that allows me to define my availability on a per-service basis. These choices will be stored in the Schedule resource as a [`SchedulingParameters` extension](https://www.medplum.com/docs/scheduling/defining-availability).

PR #9987, opened 2026-07-27
[Scheduling] `AvailableTime` edit component for service & actor

### LINKED ISSUE #9929: [Scheduling] Set Recurring Availability React Component
## Overview

Need some way to enter recurring availability information that fits the SchedulingParameters data model

# 
Image

Does not need to match pixel for pixel, but here is inspiration

Drawer component for setting standing weekly availability on a `Schedule` — writes the
`availability` block of Medplum's `SchedulingParameters` extension. These hours are what
`$find` / `$book` offer as synthetic slots. No `Slot` resources are created or modified.

## Proposed Implementation

```tsx
<WeeklyHoursEditor  //haven't put too much thought into naming here
  schedule={schedule}            // Schedule resource being edited
  service={healthcareService}    // the visit type whose hours are in scope
  ...
/>
```

## Data model

The component reads/writes one `SchedulingParameters` extension entry per service on the
Schedule. Each checked day with a time range becomes an `availableTime` sub-extension:

The above example configuration should result in something like this on the Schedule
```json
"extension": [
    {
      "url": "https://medplum.com/fhir/StructureDefinition/SchedulingParameters",
      "extension": [
        {
          "url": "service",
          "valueReference": {
            "reference": "HealthcareService/d37c2129-c888-4959-b9f6-1cd6951f4ca5",
            "display": "Office Visit"
          }
        },
        {
          "url": "availability",
          "extension": [
            {
              "url": "availableTime",
              "extension": [
                { "url": "daysOfWeek", "valueCode": "mon" },
                { "url": "availableStartTime", "valueTime": "09:00:00" },
                { "url": "availableEndTime", "valueTime": "17:00:00" }
              ]
            },
            {
              "url": "availableTime",
              "extension": [
                { "url": "daysOfWeek", "valueCode": "tue" },
                { "url": "availableStartTime", "valueTime": "09:00:00" },
                { "url": "availableEndTime", "valueTime": "17:00:00" }
              ]
            },
            {
              "url": "availableTime",
              "extension": [
                { "url": "daysOfWeek", "valueCode": "wed" },
                { "url": "availableStartTime", "valueTime": "09:00:00" },
                { "url": "availableEndTime", "valueTime": "20:00:00" }
              ]
            },
            {
              "url": "availableTime",
              "extension": [
                { "url": "daysOfWeek", "valueCode": "thu" },
                { "url": "availableStartTime", "valueTime": "09:00:00" },
                { "url": "availableEndTime", "valueTime": "17:00:00" }
              ]
            },
            {
              "url": "availableTime",
              "extension": [
                { "url": "daysOfWeek", "valueCode": "fri" },
                { "url": "availableStartTime", "valueTime": "09:00:00" },
                { "url": "availableEndTime", "valueTime": "13:00:00" }
              ]
            }
          ]
        }
      ]
    }
```

Filupnot:
Fixes #9929 and #9328

image

Adds a `ScheduleAvailabilityEditor` component to `@medplum/react-scheduling` for editing weekly availability hours, and uses it on the provider app's Schedule settings page. **One component covers both levels of the availability model:** pass a Schedule to edit an actor's hours for one visit service type, or omit it to edit that service's own default hours.

A `HealthcareService` defines default hours on its `availableTime` field, and a `Schedule` inherits them unless it sets its own `availability` override, scoped to that service, in its `SchedulingParameters` extension. The override stores exactly what `availableTime` holds, which is why one form serves both: same weekly editor over the same data, differing only in where the hours are written.

The component renders form content only, so the caller supplies the container (inline, Modal, or Drawer). It holds the week it is editing as internal draft state and hands back an updated resource on save, for the caller to persist. It reads no resources itself: the time zone the hours are entered in is passed in, resolved by the caller with `getSchedulingTimezone` from `@medplum/core`, which reads the Schedule parameters, then the service parameters, then the actor's timezone extension, the way the server does.

### Actor schedules

Editing what one calendar offers for a service. A switch makes inherit versus override explicit: off inherits the service default, shown read-only, and on writes an override seeded from those same hours. `onSave` receives an updated `Schedule`.

```jsx
<ScheduleAvailabilityEditor
  schedule={schedule}
  service={service}
  timezone={getSchedulingTimezone(service, schedule, actor)}
  onSave={(updated) => medplum.updateResource(updated)}
  onCancel={close}
/>
```

### Service defaults

Editing the hours every calendar without an override follows, so changes here reach all of them. There is no default to inherit and none to reset to, so the override switch and reset link are not shown. `onSave` receives an updated `HealthcareService`.

```jsx
<ScheduleAvailabilityEditor
  service={service}
  timezone={getSchedulingTimezone(service)}
  onSave={(updated) => medplum.updateResource(updated)}
/>
```

## Validation

Mostly prevention rather than validation. Times come from a list of the hours still free that day, so blocks cannot overlap, cannot start before the previous one ends, and cannot run past midnight.

The list runs in quarter hours, which decides what is offered rather than what a block may be set to. Typing narrows the list, and typing a time in full, minutes included, adds that exact time to it, so a block can start at 3:03 PM. The bounds still hold however a time is typed. A time already stored off the interval is shown as it is rather than rounded, and listed as the current selection.

The empty week is enforced at save, in both modes, for a different reason in each. An override with no available days fails FHIR constraint `ext-1` on write, so it cannot be stored at all. A service with no `availableTime` can be stored, but scheduling reads the missing field as unrestricted rather than unavailable, so that save would quietly mean the opposite of what the emptied form shows. Either way the Save button is refused, with the reason in a tooltip on it and in a description for anyone who cannot see the tooltip. Around the clock is therefore said rather than implied, day by day; a service that should never be scheduled is deactivated instead, and one that should not be scheduled on a single calendar is switched off there in schedule settings.

## Overnight hours

Scheduling reads `availableEndTime <= availableStartTime` as a window continuing into the next day. The editor reads that form but does not author it: a stored Friday 10:00 PM to 6:00 AM window opens as Friday 10:00 PM to 12:00 AM plus Saturday 12:00 AM to 6:00 AM, which resolves to the same bookable hours, and saving rewrites it in that split form. An end equal to its start is 24 hours from that start rather than a day beginning at midnight, and is read the same way. Authoring the wrapping form means writing it through the API or the `@medplum/core` helpers.

## Also in this PR

- Adds the availability helpers to `packages/core/src/scheduling.ts`, in two layers. `getScheduleParameters` / `setScheduleParameter` / `clearScheduleParameter` read and write one named parameter a Schedule sets for a service, whatever that parameter is, so a later control over something such as `bufferBefore` writes onto the same `SchedulingParameters` extension without needing new core API. `hasScheduleAvailability` / `setScheduleAvailability` / `clearScheduleAvailability` are typed wrappers over them for the `availability` parameter, and `resolveAvailability(service, schedule?)` reads the hours actually in effect as `HealthcareServiceAvailableTime[]`, falling back to `HealthcareService.availableTime`. Precedence lives in the resolvers, per parameter, the way the server does it, which is also why `getSchedulingTimezone` now takes the service first and the Schedule optionally. The earlier move of the shared `servicetype.ts` helpers from `server` into `core` shipped on its own in #10046.
- Adds `getExtensions` to `@medplum/core` for the repeating sub-extensions the `SchedulingParameters` shape is built from, replacing a private helper here.
- Documents the component in a Storybook MDX page attached to it, with the stories embedded as live canvases, and adds the next-day convention for overnight windows plus a pointer to that page in [Defining Availability](https://www.medplum.com/docs/scheduling/defining-availability#editing-availability-in-a-react-app).
- Generalizes the provider app's `AlphaBanner` into `ReleaseStageBanner` and moves the scheduling pages to Beta. Schedule settings also calls visit service types by that name, and shows per service whether the calendar follows the service default or overrides it.
- Fixes two SDK docs build problems the new JSDoc surfaced: a type reference such as `WithId<HealthcareService>` parsed as JSX, and a Markdown autolink parsed as a tag. Also adds an `npm run dev` script for the docs site, on port 3100 so it can run alongside the app and the provider example.
- Adds Storybook stories and tests for the component, its utils, and the core helpers.


Filupnot:
@claude review this

Filupnot:
@claude re-review after the recent changes

everett-williams:
Minor, but we should probably adjust to a 'beta' flag on the settings page, now that we have exited alpha

examples/medplum-provider/src/pages/schedule/ScheduleSettingsPage.tsx
```diff
@@ -29,6 +38,11 @@ export function ScheduleSettings(props: { schedule: Schedule }): JSX.Element | n
   });
   const [dirty, setDirty] = useState(false);
   const [saving, setSaving] = useState(false);
+  // The editor drawer stays mounted so Mantine can animate open/close; only
+  // `editorOpened` toggles. `editingService` selects which service it edits and
+  // is cleared after the close animation via `onExitTransitionEnd`.
+  const [editingService, setEditingService] = useState<WithId<HealthcareService>>();
+  const [editorOpened, setEditorOpened] = useState(false);
```
noahsilas: If you would like a silghtly more intention revealing hook, you could consider `useDiscosure`:

```suggestion
  const [editorOpened, editorHandlers] = useDisclosure(false);
```

As on nice side effect, this builds in some stable function references:
```tsx
      <ScheduleAvailabilityEditor onClose={editorHandlers.close} ... />
```

examples/medplum-provider/src/pages/schedule/ScheduleSettingsPage.tsx
```diff
@@ -110,15 +130,53 @@ export function ScheduleSettings(props: { schedule: Schedule }): JSX.Element | n
...
             </Group>
           );
         })}
       </Stack>
+      <ScheduleAvailabilityEditor
+        schedule={schedule}
+        service={editingService}
+        opened={editorOpened}
+        onClose={() => setEditorOpened(false)}
+        onExitTransitionEnd={() => setEditingService(undefined)}
```
noahsilas, line 195: This is a nice touch to keep the contents consistent during the animate out 👍

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,372 @@
...
+        : clearAvailabilityOverride(schedule, service);
+      await onSave(updated);
+      onClose();
+    } finally {
+      setSaving(false);
+    }
+  }
+
+  return (
+    <Drawer
```
noahsilas: How would you feel about moving the Drawer outside of this component? You can imagine scenarios where you want to have this editor inline in a page, or open it in a Modal or other container instead of the drawer. If we separate this then the internal component is simpler (no open/close state) and the caller can choose when the Drawer is appropriate.

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+  );
+}
+
+/**
+ * Finds the SchedulingParameters extension on a Schedule for a HealthcareService.
+ * @param schedule - Schedule to inspect
+ * @param service - HealthcareService referenced by the desired parameters
+ * @returns The matching SchedulingParameters extension, if present
+ */
+export function getServiceSchedulingParameters(schedule: Schedule, service: HealthcareService): Extension | undefined {
```
noahsilas: I think that we want this to return an array of matching extensions. Today it's the case that it probably matches at most one entry, but in the future we are likely to add features like "wildcard" matching that may encourage folks to have multiple entries that matter. 

I have a variant of this same idea in review here: https://github.com/medplum/medplum/pull/9886/changes?diff=unified#diff-94c70a16e4d2b76440e46f865425ecbd81e10b6fc6689b29c2bf4eabdd39eb1fR61

I think if we are putting this into `core` then it is important to get the API shape right, so I think this one probably is something I would consider blocking.

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+  const reference = createReference(service).reference;
```
noahsilas: Slightly less work to use `getReferenceString` (which is used internally by `createReference` to generate the `.reference` property):
```suggestion
  const reference = getReferenceString(service);
```

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+  return schedule.extension?.find((extension) => matchesServiceSchedulingParameters(extension, reference));
+}
+
+/**
+ * Expands a Schedule's availability override into per-day availability.
+ * @param schedule - Schedule containing the availability override
+ * @param service - HealthcareService referenced by the desired parameters
+ * @returns Weekly availability parsed from the matching override
+ */
+export function parseWeeklyAvailability(schedule: Schedule, service: HealthcareService): WeeklyAvailability {
```
noahsilas: The direction I was leaning in #9987 was to transform from our extension shape into the type stored at `HealthcareService.availableTime`, which is `HealthcareServiceAvailableTime[]`. That reuses a native R4 FHIR type, which means that components that use it can take a HealthcareService attribute directly, or the output of this kind of extraction function.

Would an approach like that fit into your work in this PR?

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+  return typeof actorTimezone === 'string' ? actorTimezone : undefined;
+}
+
+/**
+ * Converts a HealthcareService into an R4 CodeableConcept representation of
+ * CodeableReference<HealthcareService>.
+ * @param service - HealthcareService to represent
+ * @returns CodeableConcept values containing a reference to the service
+ */
+export function toCodeableReferenceLike(service: WithId<HealthcareService>): CodeableConcept[] {
```
noahsilas: I wonder if this might be better put into a shared space like `codeable-reference.ts` - we only use it for Scheduling at the moment, but the idea of this function is to encode the Reference part into an extension we can use on an R4 `CodeableConcept`.

This is a bit of a tricky balance though; this particular version is a bit specific to HealthcareService, which has `Healthcare.type` which has array semantics, as compared to other future CodeableReference types like `MedicationRequest.medication`, which is a `CodeableReference<Medication>` that would be linked to `Medication.code` which has only a single CodeableConcept stored.

So I think I'm on board with keeping this in here for now, although I wonder if we should make the name more explicit and leave this generic name open for our future selves?

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.utils.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    for (const range of ranges) {
+      if (!range.start || !range.end) {
+        dayError = 'Start and end times are required';
+        break;
+      }
+      if (range.end <= range.start) {
+        dayError = 'End time must be after start time';
+        break;
+      }
```
noahsilas: In the backend, we decided to accept ranges where end < start and interpret that as meaning that it should span into the next day. For example, a HealthcareServiceAvailabileTime like `{ daysOfWeek: ['tue'], availableStartTime: '22:00:00', availableEndTime: '06:00:00' }` would mean "starting at 10pm tuesday and going until 6am wednesday".

I swear I had found some reference in the FHIR spec that suggested that this was the intended reading, but I can't seem to turn it up anymore, which is quite frustrating; should have made better notes when implementing! 

It might be the case that we should enforce that clients handle this split and expressly say "tue: 10pm-12am, wed: 12am-6am".

One challenge for this format: [the FHIR `time` type explicitly disallows "24:00",](https://hl7.org/fhir/R4/datatypes.html#time) which suggests that in that world we need to special case `"00:00"` as an end time to be interpreted as "end-of-day". 😪 

To make this feedback most actionable: I think we should probably make our react package match how the server operations work for now, and revisit all of this together before we exit `beta`.

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.utils.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    if (!dayError && ranges.length > 1) {
+      const sorted = [...ranges].sort((a, b) => a.start.localeCompare(b.start));
+      for (let i = 1; i < sorted.length; i++) {
+        if (sorted[i].start < sorted[i - 1].end) {
+          dayError = 'Time ranges must not overlap';
```
noahsilas: It's not _strictly_ a problem if ranges overlap (the backend should handle such a scenario fine). That said this still might be something we want to warn about in the UI when in this scenario, as it might indicate that the user has made some kind of mistake.

noahsilas (changes requested):

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import type { CodeableConcept, Extension, HealthcareService, Reference, Resource, Schedule } from '@medplum/fhirtypes';
+import type { WithId } from './utils';
+import { createReference, deepClone, getExtension, getExtensionValue, getReferenceString, isDefined } from './utils';
+
+export const SchedulingParametersURI = 'https://medplum.com/fhir/StructureDefinition/SchedulingParameters';
+export const ServiceTypeReferenceURI = 'https://medplum.com/fhir/service-type-reference';
+export const TimezoneExtensionURI = 'http://hl7.org/fhir/StructureDefinition/timezone';
+
+export type DayOfWeek = 'mon' | 'tue' | 'wed' | 'thu' | 'fri' | 'sat' | 'sun';
```
techdavidy: maybe 

export const DAYS_OF_WEEK = ['mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun'] as const;
export type DayOfWeek = typeof DAYS_OF_WEEK[number];

techdavidy: it seems we can now use this export in 

packages/server/src/fhir/operations/utils/scheduling-parameters.ts:83
packages/react/src/TimingInput/TimingInput.tsx:18

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+}
+
+export interface DayAvailability {
+  allDay: boolean;
+  ranges: TimeRange[];
+}
+
+export type WeeklyAvailability = Record<DayOfWeek, DayAvailability>;
+
+export function isDayOfWeek(value: string | undefined): value is DayOfWeek {
```
techdavidy, line 50:
```
export function isDayOfWeek(value: string | undefined): value is DayOfWeek {
  return DAYS_OF_WEEK.includes(value as DayOfWeek);
}
```

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,329 @@
...
+    value === 'tue' ||
+    value === 'wed' ||
+    value === 'thu' ||
+    value === 'fri' ||
+    value === 'sat' ||
+    value === 'sun'
+  );
+}
+
+export function emptyWeeklyAvailability(): WeeklyAvailability {
```
techdavidy: maybe 
defaultWeeklyAvailability()  or
blankWeeklyAvailability()

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,372 @@
...
+interface DayDraft {
+  allDay: boolean;
+  ranges: DraftRange[];
+}
+
+type DraftAvailability = Record<DayOfWeek, DayDraft>;
+
+export interface ScheduleAvailabilityEditorProps {
+  readonly schedule: Schedule;
+  /**
```
techdavidy, line 193: let's add jsdocs to ScheduleAvailabilityEditorProps rather than inline comments

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,372 @@
...
+          <Button onClick={handleSave} loading={saving} disabled={!canSave}>
+            Save
+          </Button>
+        </Group>
+      </Box>
+    </Drawer>
+  );
+}
+
+function emptyDraft(): DraftAvailability {
```
techdavidy: doesn't emptyWeeklyAvailability() in core do the same ?

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
Filupnot: I like this idea

packages/core/src/scheduling.ts
Filupnot: Good call, that makes sense. Updated!

packages/core/src/scheduling.ts
Filupnot: This makes sense to me. I refactored to pretty much copy the `extractAvailability` function that you had in your PR and am now using that instead to take in either a `Schedule` or `HealthcareService` and return `HealthcareServiceAvailableTime[]`. Lmk if you think this looks good!

packages/core/src/scheduling.ts
Filupnot: I opted to make this more explicit to service type: 
- `toCodeableReferenceLike` -> `toServiceTypeCodeableConcepts`

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.utils.ts
Filupnot: In hindsight I think it shouldn't be a problem if time ranges overlap since we combine them anyways, and since the validation complexity would increase a fair bit with all the "wrapping" updates. 

Removed UI warning here, but just added one for when times match (`10:00->10:00`) to point them towards instead marking `All Day` instead (same interpretation). Also added a little banner for when a wrap causes availability into a date that's marked as "no availability" to avoid confusion.

examples/medplum-provider/src/components/ReleaseStageBanner.module.css
```diff
@@ -1,9 +1,9 @@
-.alphaBanner {
+.banner {
   background-color: light-dark(var(--mantine-color-violet-1), var(--mantine-color-violet-9));
```
noahsilas: My intuition is that for a "beta" banner we probably want less bold color choices; maybe something in the blue family instead of these strong purples?

packages/docs/README.md
```diff
@@ -9,10 +9,18 @@ Medplum documentation should be installed automatically by following the instruc
...
 
 ```bash
-npm run docusaurus start
+npm run dev
 ```
 
-This command starts a local development server and opens up a browser window. Most changes are reflected live without having to restart the server.
+This command generates the SDK reference pages (`docs/sdk`, from `@medplum/core`), starts a local development server with an increased Node heap, and opens up a browser window. Most changes are reflected live without having to restart the server.
+
+The server defaults to port `3000`. If that port is already in use (for example, the Medplum app also defaults to `3000`), pass a different one:
```
noahsilas: Maybe we should default this to a port that something else in the medplum stack doesn't also use?

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.utils.ts
Filupnot: Updated to support wrapping time to go into the next day. We're not using `24:00` here but instead interpreting `00:00` as being the next day (midnight).

packages/core/src/scheduling.ts
Filupnot: done as suggested

Filupnot: both are switched over now

packages/core/src/scheduling.ts
Filupnot: Went with `blankWeeklyAvailability()`. Also moved out of core since the core availability API now usese `HealthcareServiceAvailableTime[]` instead of its own shape. `WeeklyAvailability` and the blank builder now live next to the editor in `ScheduleAvailabilityEditor.utils.ts`.

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,288 @@
...
+ * Finds the SchedulingParameters extensions on a Schedule for a HealthcareService.
+ * @param schedule - Schedule to inspect
+ * @param service - HealthcareService referenced by the desired parameters
+ * @returns Every matching SchedulingParameters extension, in document order
+ */
+export function getServiceSchedulingParameters(schedule: Schedule, service: HealthcareService): Extension[] {
+  const reference = getReferenceString(service);
+  if (!reference) {
+    return [];
+  }
```
noahsilas: If you mark the type of `service` here as `WithId<HealthcareService>`, then the `getReferenceString(service)` function is typed such that it won't return `undefined` and you can remove a guard here. It also serves as a hint to future authors that this function is intended to be called with a persisted HealthcareService.


```suggestion
export function getServiceSchedulingParameters(schedule: Schedule, service: WithId<HealthcareService>): Extension[] {
  const reference = getReferenceString(service);
```

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * Immutably applies an availability override to a Schedule for a HealthcareService.
+ * @param schedule - Schedule to update
+ * @param service - HealthcareService referenced by the parameters
+ * @param availableTime - Availability to apply
+ * @returns A cloned Schedule containing the availability override
+ */
+export function applyAvailability(
+  schedule: Schedule,
+  service: HealthcareService,
```
noahsilas: Similarly to above, it would be nice to annotate this as `WithId<HealthcareService>`, as passing a resource without an ID into `createReference` can result in some weirdness. (I'd like to make that a type error in a future major version bump; I've just created a ticket documenting this idea at #10021)

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -182,6 +182,20 @@ The `availability` sub-extension mirrors the FHIR R5+ [`Availability`](https://h
...
+
+Because `00:00:00` is read as the following midnight rather than as the start of the same day, it is the correct way to say "until end of day". Do not use `23:59:59` for this: it leaves a one-second gap that splits the window in two, which prevents booking anything that spans midnight.
+
+A window is keyed to the day it **starts** on, and only that day needs to appear in `daysOfWeek`. An entry on `fri` ending at `06:00:00` makes Saturday morning bookable without `sat` being listed anywhere.
+
+:::note[Windows that cross midnight and other FHIR systems]
+
+This reading is Medplum's convention. The FHIR specification does not define what an end time before a start time means, and imposes no invariant either way, so the data is valid FHIR but other systems may not interpret it the same way. When publishing hours for outside consumption rather than for Medplum's scheduling operations, consider writing two entries (for example `22:00:00` to `23:59:59` on Tuesday plus `00:00:00` to `06:00:00` on Wednesday), which is more widely understood. The one-second gap does not matter for display, only for booking.
```
noahsilas: I worry a bit about encouraging people to use `23:59:59` as I think that today `$find` would fail to use such a definition to allow bookings up to the end of the day or crossing that threshold.

What if we leave this off of the published docs for now, and create a follow up ticket to improve handling of this special case?

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -290,6 +304,28 @@ In this example:
...
+
+The interval decides what is listed, not what a block may be set to, so times written elsewhere need not fall on it. The editor shows such a time as it is rather than rounding a value nobody asked it to change, and lists it as the current selection. The interval itself is measured from midnight, so a block stored as 9:07 AM still offers 9:15 AM and 9:30 AM alongside it.
+
+The editor does not author [windows that cross midnight](#windows-that-cross-midnight), but it does read them. On mount, a stored window like Friday 10:00 PM to 6:00 AM is split at the day boundary into Friday 10:00 PM to 12:00 AM plus Saturday 12:00 AM to 6:00 AM, which the scheduling operations read as the same bookable hours. Overlapping entries written by the API are likewise sorted and merged into single blocks. Availability authored elsewhere therefore still opens in the editor rather than locking it, though saving rewrites it into the split form.
```
noahsilas: Thanks for including docs! This block feels very specific to this React component though, which feels like it belongs at a different level than the rest of this page which is about the API/data modeling shape.

I like the idea of calling out on this page that we have such a component to make this easy, but maybe we can push most of this documentation into a second page? Or maybe even include it inside the storybook (which would keep it more co-located with the component it is describing, hopefully helping us remember to update the docs whenever it changes). What do you think?

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,426 @@
...
+import { TimeSelect } from './TimeSelect';
+
+// Width of the day name column, so every row's hours start on the same edge.
+const DAY_COLUMN_WIDTH = 168;
+// A single row's height, used to keep a day with no hours the same height as one
+// with a block of them.
+const ROW_HEIGHT = 36;
+// Width of an action slot. Reserved whether or not it holds a button, so the
+// rows above and below stay aligned.
+const ACTION_WIDTH = 28;
```
noahsilas: This sounds like a good use for [CSS Grid](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Grid_layout).

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,426 @@
...
+            const checked = e.currentTarget.checked;
+            onChange({
+              enabled: checked,
+              ranges: checked && ranges.length === 0 ? [{ ...DEFAULT_RANGE }] : ranges,
+            });
+          }}
+          color="green.6"
+          withThumbIndicator={false}
+          disabled={disabled}
+          styles={disabled && enabled ? DISABLED_ON_SWITCH_STYLES : undefined}
```
noahsilas: The usage of both `disabled` (to mean "the entire form has been disabled") and `enabled` (to mean "this particular row has been enabled") is a little confusing, especially on this line where we are checking both.

I think this is one scenario where removing the destructuring of `value` would make this clearer by retaining scope information on one of these:

```suggestion
          styles={disabled && value.enabled ? DISABLED_ON_SWITCH_STYLES : undefined}
```

I wonder if there is anything else we can do to help clarify these distinctions - the inherent complexity of nested toggles just makes this a bit of a tough thing to grok at first.

packages/react/src/ScheduleAvailabilityEditor/TimeSelect.tsx
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Props for the TimeSelect component.
+ * @param value - The selected time, as minutes from midnight
+ * @param min - Earliest selectable time, as minutes from midnight
+ * @param max - Latest selectable time, as minutes from midnight
+ * @param label - Accessible label for the input
+ * @param onChange - Called with the newly selected time
+ * @param disabled - Whether the input is disabled
+ * @param testId - Test id applied to the input
+ * @param flashKey - Change this to flash the input, drawing the eye to a value the editor changed on the user's behalf
```
noahsilas: This feels like it could be a good use case for [`useImperativeHandle()`](https://react.dev/reference/react/useImperativeHandle) - at the time that an event is mutating `value`, it could also call `timeSelectRef.current.flash()` imperatively.

packages/react/src/ScheduleAvailabilityEditor/TimeSelect.tsx
```diff
@@ -0,0 +1,196 @@
...
+    if (!dropdownOpened) {
+      return undefined;
+    }
+    const frame = requestAnimationFrame(() => {
+      combobox.selectActiveOption();
+      activeOptionRef.current?.scrollIntoView({ block: 'center' });
+    });
+    return () => cancelAnimationFrame(frame);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dropdownOpened]);
```
noahsilas: I think you can destructure `selectActiveOption` and use it as an input to this effect here instead of suppressing the warning. (It should be a stable callback function reference and so shouldn't affect when this fires).

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,426 @@
...
+        <Text c="dimmed" h={ROW_HEIGHT} lh={`${ROW_HEIGHT}px`}>
+          Unavailable
+        </Text>
+      )}
+    </Group>
+  );
+}
+
+interface CommonProps {
+  /** The HealthcareService whose availability is being edited. */
```
noahsilas: This reads a little misleadingly; the HealthcareService itself is not necessarily being edited.

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,426 @@
...
+      )}
+    </Group>
+  );
+}
+
+interface CommonProps {
+  /** The HealthcareService whose availability is being edited. */
+  readonly service: HealthcareService;
+  /** Called when the user cancels. Omit to hide the cancel button, e.g. when the editor is inline on a page. */
+  readonly onCancel?: () => void;
```
noahsilas, line 201: Is there a way to make this a controlled component with an `onChange` method instead of exposing `onCancel` / `onSave`?

I suspect that we will eventually have some pages where editing a service's default availability is one section of the form (probably sharing space next to fields like bufferBefore/bufferAfter settings), and it would be nice for these to all compose well.

If this is a huge pain due to how much state is being managed, then we can probably keep it as-is and keep this on separate sub-pages or modals in the future.

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,426 @@
...
+              value={weekly[day]}
+              disabled={!overriding}
+              onChange={(value) => setWeekly((prev) => ({ ...prev, [day]: value }))}
+              onAnnounce={setAnnouncement}
+            />
+          ))}
+        </Stack>
+        <VisuallyHidden role="status" aria-live="polite" data-testid="schedule-availability-announcement">
+          {announcement}
+        </VisuallyHidden>
```
noahsilas, line 387: Nice a11y touch here 👍

noahsilas (commented):
What do you think about putting the util move from server -> core into its own PR? I think we can merge that one nice and fast, and it will hopefully help git track the changes a little more subtly than it does in this big patch.

I'm still reviewing some of the bits, but leaving early feedback now. Overall it's looking good - I think the big open question I would want to raise is around if the component should expose the `onSave`/`onCancel` callbacks versus being a controlled component that can be more easily composed with other form fields.

Thanks for all the attention to tests, docs, and accessibility!

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,288 @@
...
+  const reference = getReferenceString(service);
+  if (!reference) {
+    return [];
+  }
+  return schedule.extension?.filter((extension) => matchesServiceSchedulingParameters(extension, reference)) ?? [];
+}
+
+function getSubExtensions(extension: Extension | undefined, url: string): Extension[] {
+  return extension?.extension?.filter((subextension) => subextension.url === url) ?? [];
+}
```
noahsilas: I think this is OK since we aren't exporting it, and having this specific name does make the intent a little clearer inside this file where we are dealing with lots of layers.

I think that long term what we probably want is a method in `core/utils` that is similar to the current `getExtension()` but that returns an array of extensions instead of finding the first match. Such a method could be used on any resource or datatype that includes an `extension: Extension[]` attribute, which includes `Extension` itself.

In a world where we have that, then some of the calls in here might change a bit:
```
  const daysOfWeek = getSubExtensions(availableTime, 'daysOfWeek') # current
  const daysOfWeek = getExtensions(availableTime, 'daysOfWeek') #future
```

We've started exploring this a bit in server (https://github.com/medplum/medplum/blob/v5.1.27/packages/server/src/util/extension.ts), but in that particular context we also are doing some extra lifting to annotate the extensions with where they are in the original extension array (for purposes of emitting more helpful error messages when something weird happens). It's not yet obvious if that type of work would be helpful in `core` yet.

Nothing very actionable in this comment, just posting it to share some context.

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,288 @@
...
+  return updated;
+}
+
+/**
+ * Immutably clears a Schedule's availability override for a HealthcareService.
+ * @param schedule - Schedule to update
+ * @param service - HealthcareService referenced by the parameters
+ * @returns A cloned Schedule without the matching availability override
+ */
+export function clearAvailabilityOverride(schedule: Schedule, service: HealthcareService): Schedule {
```
noahsilas: This could be a little more general to be useful in more future contexts:
```ts
export function clearSchedulingOverride(schedule: Schedule, service: HealthcareService, url: string): Schedule {
  const updated = deepClone(schedule);
  for (const parameters of getServiceSchedulingParameters(updated, service)) {
    if (parameters.extension) {
      parameters.extension = parameters.extension.filter((subextension) => subextension.url !== url);
    }
  return updated;
}

// sample usages:
clearSchedulingOverride(schedule, service, 'availability');
clearSchedulingOverride(schedule, service, 'bufferBefore');
```

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/**
+ * Immutably applies an availability override to a Schedule for a HealthcareService.
+ * @param schedule - Schedule to update
+ * @param service - HealthcareService referenced by the parameters
+ * @param availableTime - Availability to apply
+ * @returns A cloned Schedule containing the availability override
+ */
+export function applyAvailability(
```
noahsilas: I think this might be a good candidate to split up - there's a more common reusable operation here that is something like `setSchedulingOverride(schedule, service, subExtension)`, which you can imagine using for future controls on parameters like `bufferBefore`. Then `applyAvailability` is a nice special case of that:
```ts
export function applyAvailability(
  schedule: Schedule,
  service: HealthcareService,
  availableTime: HealthcareServiceAvailableTime[]
): Schedule {
  return setSchedulingOverride(schedule, service, buildAvailabilityExtension(availableTime));
}
```

I think because we are exporting these functions from `core`, we do want to try to create a fairly minimal set that will serve many uses. If we end up having `applyAvailability`, `applyDuration`, `applyBufferBefore`, ..., I think it would be a headache, so this is a good place for us to do a bit of forward thinking on this API design.

Still some room for adjustment in here though; taking a subextension object is pretty opaque, whereas the type annotation on this is nicely explicit. I wonder if there's a way to get all these strengths 🤔

packages/docs/README.md
Filupnot: Good call, moved to 3100.

examples/medplum-provider/src/components/ReleaseStageBanner.module.css
Filupnot: How's this look?  Switched to blue: `blue-1` / `blue-9` for the banner and `blue-6` / `blue-2` for the pill, light and dark respectively. 
image

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
Filupnot: Done. The week is one grid now rather than seven nested flex rows, so `DAY_COLUMN_WIDTH`, `ROW_HEIGHT`, and `ACTION_WIDTH` are gone, along with the placeholder `Box` elements that were reserving the action slots. Cells are placed by explicit `grid-column`, so a day's second block simply leaves the day column empty rather than needing a spacer. Styles live in a new `ScheduleAvailabilityEditor.module.css`, matching how the rest of the package does it.

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
Filupnot: Renamed rather than un-destructured, which I think gets closer to the root of it: the form-wide prop is now `readOnly` and the per-day flag is `available`. The line reads `readOnly && available`, and `DISABLED_ON_SWITCH_STYLES` became `AVAILABLE_READ_ONLY_SWITCH_STYLES` to match.

packages/react/src/ScheduleAvailabilityEditor/TimeSelect.tsx
Filupnot: Done. `TimeSelect` is a `forwardRef` exposing a `flash()` handle, and `DayRow` keeps a ref per end input, so `setStart` calls `endInputs.current[index]?.flash()` at the point it moves the value. `forwardRef` rather than `ref` as a plain prop because the package still declares a React 18 peer.

One wrinkle worth flagging: internally the flash is a counter rather than a boolean, so a second move while the first flash is still showing restarts it instead of being swallowed by the running timer. Added a test that fails under the boolean version.

packages/core/src/scheduling.ts
Filupnot: Done, with that signature. `clearAvailabilityOverride` is now just `clearSchedulingOverride(schedule, service, 'availability')`.

packages/core/src/scheduling.ts
Filupnot: Thanks, that context is useful. Left `getSubExtensions` private here. Agreed that a `getExtensions()` in `core/utils` returning every match is the better long-term shape, and these call sites should switch over when it lands.

packages/core/src/scheduling.ts
Filupnot: Done. `setSchedulingOverride(schedule, service, subextension)` is the general form, and `applyAvailability` is exactly the wrapper you sketched.

On the tension you raised at the end, I kept both rather than choosing. `setSchedulingOverride` takes the opaque sub-extension and is what you reach for when a new parameter appears; `applyAvailability` and `clearAvailabilityOverride` stay as thin typed wrappers, so callers still get the explicit `HealthcareServiceAvailableTime[]` annotation. That does mean a wrapper per parameter if we want the same explicitness for `bufferBefore` later, but each one is a single line over a path that is always available directly, which seems like a cheap way to have both.

packages/docs/docs/scheduling/defining-availability.md
Filupnot: Moved into Storybook, which I agree keeps it closer to the thing it describes. There is now an attached `ScheduleAvailabilityEditor.mdx` next to the component, with the behavior prose broken into sections and six of the stories embedded as live canvases, and `defining-availability.md` keeps a three sentence pointer to it.

Worth flagging while I was in there: autodocs is not enabled in this Storybook, so the `--docs` link that page already carried was dead. The attached MDX creates that entry, so it resolves now.

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
Filupnot: Keeping `onSave` / `onCancel` for now. The editor derives a fair amount of state on mount: it parses `availableTime` into a week, splitting windows that cross midnight and merging overlaps, and it holds on to the hours of days that are switched off so that toggling a day back on restores them rather than starting blank. Lifting that to the caller means either exporting the internal weekly shape as the public value type, so every caller carries the parse and serialize round trip, or emitting a whole updated `Schedule` on each keystroke.

The composition case you describe is real though, and the piece that makes it work is now in place: `setSchedulingOverride` in core lets a form holding `bufferBefore` and availability side by side write both onto the same SchedulingParameters extension. I would rather add an `onChange` alongside the existing props when we build that page, with a concrete second field to design against, than guess at its shape now.

Filupnot:
Pushed 3f2092577 addressing the review. Replies are on the individual threads; the summary:

- `setSchedulingOverride` / `clearSchedulingOverride` are now the general pair in core, with `applyAvailability` / `clearAvailabilityOverride` as thin typed wrappers over them.
- The week is a single CSS grid, so the three pixel constants and the placeholder elements holding the action columns open are gone.
- `TimeSelect` exposes a `flash()` handle instead of taking a `flashKey`.
- The component prose moved into an attached Storybook MDX page with the stories embedded, leaving a pointer on `defining-availability.md`.
- Dropped the `23:59:59` end of day suggestion, and the beta banner is blue.
- Kept `onSave` / `onCancel` rather than going fully controlled, with reasoning on that thread.

On splitting out the util move: opened #10046, which is the server to core move on its own, based on `main`. It is a move plus renames with no behavior change, and deletes the copy of those helpers the provider example was keeping in parallel. This branch is not rebased onto it yet; I will do that once it merges.

One thing to call out on #10046: the move condensed `packages/server/src/util/servicetype.test.ts` from twelve cases to three. I kept it that way so it matches this branch exactly and the rebase stays clean, but happy to restore the fuller set there if you would rather not lose the coverage in the move.

Filupnot:
> What do you think about putting the util move from server -> core into its own PR? I think we can merge that one nice and fast, and it will hopefully help git track the changes a little more subtly than it does in this big patch.
> 
> I'm still reviewing some of the bits, but leaving early feedback now. Overall it's looking good - I think the big open question I would want to raise is around if the component should expose the `onSave`/`onCancel` callbacks versus being a controlled component that can be more easily composed with other form fields.
> 
> Thanks for all the attention to tests, docs, and accessibility!

I agree this would probably be a good idea - moved into #10046 !

packages/docs/docs/scheduling/defining-availability.md
Filupnot: Makes sense - created #10049

packages/core/src/utils.ts
```diff
@@ -529,6 +529,26 @@ export function getExtension(resource: any, ...urls: string[]): Extension | unde
...
+ * Returns every extension reachable by the given extension URLs.
+ *
+ * Like `getExtension`, but does not stop at the first match: extensions that repeat under one URL are
+ * all returned, at every level. Use it for extensions defined with a cardinality above one, where
+ * `getExtension` would silently read only the first.
+ * @param resource - The base resource, data type, or extension. Anything that may hold an `extension` array.
+ * @param urls - Array of extension URLs. Each entry represents descending a level in a nested extension.
+ * @returns Every matching extension, in document order. Empty if none match or no URL is given.
+ */
+export function getExtensions(resource: any, ...urls: string[]): Extension[] {
```
noahsilas: We're trying to avoid introducing new `any` values as much as possible; this type effectively opts you out of all type checking, which can hide issues and errors.

In the server utils, we define a type [`Extensible`](https://github.com/medplum/medplum/blob/v5.1.28/packages/server/src/util/extension.ts#L13-L15) that takes advantage of Typescript's structural typing - it accepts anything that can have an Extension array in the `extension` property.

This causes a type error if you try to pass in something that does not support extensions (`Binary` is a notable example).

noahsilas: How would you feel about making `urls` a single parameter instead of a variadic?

```ts
export function getExtensions(resource: any, urlOrUrls: string | string[]): Extension[] {
  const urls = arrayify(urlOrUrls);
  ...
}
```

This would let us more easily add future options to this call site; you can imagine future usage like `getExtensions(resource, SchedulingParametersURI, { withPath: true })`.

packages/core/src/scheduling.ts
```diff
@@ -52,6 +68,242 @@ export function hasSchedulingParameters(resource: Schedule | HealthcareService):
...
+): string | undefined {
+  const scheduleTimezone = (schedule ? getServiceSchedulingParameters(schedule, service) : [])
+    .flatMap((parameters) => getExtensions(parameters, 'timezone'))
+    .map((subextension) => subextension.valueCode)
+    .find(isDefined);
+  if (scheduleTimezone) {
+    return scheduleTimezone;
+  }
+
+  const serviceTimezone = getExtensions(getExtension(service, SchedulingParametersURI), 'timezone')
```
noahsilas:
```suggestion
  const serviceTimezone = getExtensions(service, SchedulingParametersURI, 'timezone')
```

packages/core/src/scheduling.ts
```diff
@@ -52,6 +68,242 @@ export function hasSchedulingParameters(resource: Schedule | HealthcareService):
...
+ * FHIR timezone extension.
+ * @param schedule - Schedule whose parameters may define a timezone. Omit to resolve the service's own timezone,
+ * as when the service default hours are being read on their own rather than through a particular calendar.
+ * @param service - HealthcareService whose parameters may define a timezone
+ * @param actor - Optional Schedule actor used as a timezone fallback
+ * @returns The resolved IANA timezone identifier, if present
+ */
+export function getSchedulingTimezone(
+  schedule: Schedule | undefined,
+  service: WithId<HealthcareService>,
```
noahsilas, line 198: This feels like it should be reversed so that `schedule` can be omitted cleanly by a caller:
```suggestion
export function getSchedulingTimezone(
  service: WithId<HealthcareService>,
  schedule?: Schedule,
```

packages/core/src/scheduling.ts
```diff
@@ -52,6 +68,242 @@ export function hasSchedulingParameters(resource: Schedule | HealthcareService):
...
+  return setSchedulingParameter(schedule, service, buildAvailabilityExtension(availableTime));
+}
+
+/**
+ * Immutably clears a Schedule's availability override for a HealthcareService.
+ * @param schedule - Schedule to update
+ * @param service - HealthcareService referenced by the parameters
+ * @returns A cloned Schedule without the matching availability override
+ */
+export function clearAvailabilityOverride(schedule: Schedule, service: WithId<HealthcareService>): Schedule {
```
noahsilas: I'm a little uncertain about how much "availability" is getting baked in to the exported core API. I know that this supports the current component you are working on in this PR most closely, but I think that in the long term this sets up a pattern for a large suite of `clear*Override` functions that we should probably avoid.

What do you think about exposing the underlying `clearSchedulingParameter` directly and dropping this method?

The `setAvailabilityOverride` I'm a little more ok with, as it is doing some pretty tricky data munging to move data from the friendlier shape to the extension shape, but even that _might_ be something we should pull out of `core`. So far we've been thinking that reusing `HealthcareServiceAvailableTime` as the more useful direct format is the best way to build our UI (largely because it makes working with both HealthcareService + Schedule easy), but maybe that's not actually the right thing here. I wonder if we should push more of that transformation into specific consumers?

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.utils.ts
```diff
@@ -0,0 +1,460 @@
...
+
+/**
+ * The order days are listed in the editor. `DAYS_OF_WEEK` is the FHIR order,
+ * which starts on Monday, but the editor reads as a calendar and so starts on
+ * Sunday.
+ */
+export const DAY_DISPLAY_ORDER: readonly DayOfWeek[] = ['sun', 'mon', 'tue', 'wed', 'thu', 'fri', 'sat'];
+
+/**
+ * A block of hours, as minutes from midnight. `end` is exclusive and may be
```
noahsilas: "hours, as minutes" is a little confusing; What do you think about this slight tweak:
```suggestion
 * A block of time, as minutes from midnight. `end` is exclusive and may be
```

packages/react/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.utils.ts
```diff
@@ -0,0 +1,460 @@
...
+    if (start === undefined || end === undefined) {
+      continue;
+    }
+
+    for (const day of days) {
+      if (end > start) {
+        collected[day].push({ start, end });
+      } else if (end === start) {
+        // Scheduling reads a window ending at its start time as a full 24 hours.
+        collected[day].push({ start: 0, end: MINUTES_PER_DAY });
```
noahsilas: This seems incorrect; I think this should probably fall through into the `else` block below. An entry like `{ dayOfWeek: ['mon'], availableStartTime: '05:00:00', availableEndTime: '05:00:00' }` is interpreted by the server as "Monday 05:00 - 24:00; Tue 00:00 - 05:00".

EDIT: Looks like you already corrected this in a later commit, nice 👍

examples/medplum-provider/src/components/ReleaseStageBanner.module.css
```diff
@@ -0,0 +1,14 @@
+.banner {
+  background-color: light-dark(var(--mantine-color-blue-1), var(--mantine-color-blue-9));
+  color: light-dark(var(--mantine-color-dark-8), var(--mantine-color-gray-3));
```
noahsilas, line 3: Thanks for updating this to make the "beta" banners look good. Not immediately impactful as I think we have now converted all the current "alpha" banners to "beta", but I think that we probably should set this up so that it keeps the more eye-catching purple hues for "alpha".

To me that makes it seem like when the release stage is "alpha" we add that as a css class, and then we can add rules like `.banner.alpha { ... }` to override the standard blue color.

This also feels like it is only tangentially related to your change here; we could move this into a separate PR if that would help keep things moving fast, or if it is easier to address it inline here that's legit too 👍

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.stories.tsx
```diff
@@ -1,20 +1,33 @@
...
 import { ScheduleAvailabilityEditor } from './ScheduleAvailabilityEditor';
 
 export default {
   title: 'Medplum/ScheduleAvailabilityEditor',
   component: ScheduleAvailabilityEditor,
 } as Meta;
 
+/*
+ * Stands in for `Document` from @medplum/react, which this package does not depend on. Mirrors what
+ * that component renders, so the stories frame the editor the way a page in an app would.
```
noahsilas: It does seem like we are likely to add a dependency on `@medplum/react` at some point (lots of useful components in there for displaying various FHIR resources and things). I wonder if we should just pull that in now 🤔

noahsilas (commented):
There are lots of verbose comments that seem to be a sort of hallmark of one of the current Claude workflows. It might be a good idea to give this a pass through the [comment-compactor skill](https://github.com/medplum/medplum/blob/main/.claude/skills/comment-compactor/SKILL.md) that was recently added to see what that tidies up.

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.stories.tsx
Filupnot: Done - I added it as a `devDependency` since it looks like only Storybook is using it today.

examples/medplum-provider/src/components/ReleaseStageBanner.module.css
Filupnot: It was pretty straightforward so just added it here 

```css 
.banner.alpha {
  background-color: light-dark(var(--mantine-color-violet-1), var(--mantine-color-violet-9));
}

.pill.alpha {
  background-color: light-dark(var(--mantine-color-violet-4), var(--mantine-color-violet-2));
}
```

packages/core/src/scheduling.ts
Filupnot: I renamed the suffix `*AvailabilityOverride` to `*ScheduleAvailability` since this feels less like an "override" and more like "defining at the more specific **schedule** level", and to avoid setting precedent for using "override" around our core functions. 

I agree with not exporting `hasScheduleAvailability` and `clearScheduleAvailability` since they were just one-line wrappers around `getScheduleParameters` and `clearScheduleParameter`. 

I do think it's worth keeping `setScheduleAvailability` and `resolveAvailability` because of the "data munging" you referred to; because `availability` is the only parameter that has a nested extension, which would make it awkward to use through the general form parameter function. 

Also unrelated but I renamed `resolveAvailability` to `getEffectiveAvailability`. There's already a core function called `resolveAvailability` on the server, and this felt more targeted to what this does. 

Happy to talk more about it.

packages/core/src/scheduling.ts
```diff
@@ -45,13 +62,313 @@ export function isDayOfWeek(value: string | undefined): value is DayOfWeek {
...
+ * service is configured rather than any particular one. Ask `getScheduleParameters` about a service.
  * @param resource - Schedule or HealthcareService to inspect
  * @returns True if the resource has a SchedulingParameters extension
  */
 export function hasSchedulingParameters(resource: Schedule | HealthcareService): boolean {
   return !!getExtension(resource, SchedulingParametersURI);
 }
 
+// Scheduling matches a `service` reference on resourceType and id, so a stored reference carrying a version
+// suffix still names the service. Match the same way: a reference the server honours but this module missed
```
noahsilas, line 67: #TIL about the `/_history/:version` suffix that is allowed on a reference.

This might be something we should think about more holistically later; I think the meaning of a service-type reference with a version specifier is not obvious, and maybe should just be an error in these positions?

The server behavior is actually a bit of an accident; it could just as easily have said ``reference.reference === `${resource.resourceType}/${resource.id}`` (and maybe it should have, since that would create a single string to compare and not allocate a `string[]` 🤔). 

For now, matching the behavior from the server operations makes sense. We should follow up later and decide if `parseReference(reference)` should be version aware and if we should use it in more places instead of `reference.reference.split('/')`, or if there is a more complete `isReferenceTo(reference, resource)` that could be version aware.

No action needed on this comment for this PR, just thinking aloud and noting for later.

packages/core/src/scheduling.ts
```diff
@@ -45,13 +62,313 @@ export function isDayOfWeek(value: string | undefined): value is DayOfWeek {
...
+    .filter(isDayOfWeek);
+
+  if (getExtensions(availableTime, 'allDay')[0]?.valueBoolean) {
+    return { daysOfWeek, allDay: true };
+  }
+
+  return {
+    daysOfWeek,
+    availableStartTime: getExtensions(availableTime, 'availableStartTime')[0]?.valueTime,
+    availableEndTime: getExtensions(availableTime, 'availableEndTime')[0]?.valueTime,
```
noahsilas: I think this should probably emit some kind of error if one of these extensions is present multiple times. Our general philosophy with the scheduling stuff has been to try and stay pretty strict in how we are validating and emit error messages early - the configuration here is so complex that it gets very confusing when things go wrong.

If we just are taking the first match though, you could use `getExtension()` which can short circuit the iteration and avoid building an intermediate array to get just one value. (This advice is not super impactful in react components, but since this is in Core it can also get used in places like `@medplum/server` in the future where we are more sensitive to things like extra array allocations).

packages/core/src/scheduling.ts
```diff
@@ -45,13 +62,313 @@ export function isDayOfWeek(value: string | undefined): value is DayOfWeek {
...
+function buildAvailableTimeExtension(entry: HealthcareServiceAvailableTime): Extension {
+  const days: Extension[] = (entry.daysOfWeek ?? []).map((day) => ({ url: 'daysOfWeek', valueCode: day }));
+
+  if (entry.allDay) {
+    return { url: 'availableTime', extension: [...days, { url: 'allDay', valueBoolean: true }] };
+  }
+
+  if (!entry.availableStartTime || !entry.availableEndTime) {
+    throw new OperationOutcomeError(
+      badRequest('availableTime must set allDay, or both availableStartTime and availableEndTime')
```
noahsilas: #nit: I think this should probably be a `validationError` rather than `badRequest`. You could even set the code to `required` if you want to be maximally descriptive

packages/core/src/utils.test.ts
```diff
@@ -746,6 +747,47 @@ describe('Core Utils', () => {
...
+        {
+          url: 'http://example.com/complex',
+          extension: [{ url: 'key1', valueCode: 'baz' }],
+        },
+      ],
+    };
+
+    // Every match at one level, where getExtension would read only the first.
+    expect(getExtensions(resource, 'http://example.com/complex')).toHaveLength(2);
+    expect(getExtension(resource, 'http://example.com/complex', 'key1')).toHaveProperty('valueCode', 'foo');
```
noahsilas: Should this be `getExtensions(resource, ['http://example.com/complex', 'key1'])`?

noahsilas (approved):
Looking good, thanks for taking the time to go through so many rounds of feedback on this one.

I would still probably advocate for moving most of the functions dealing with the `HealthcareServiceAvailableTime` type out of core and into `packages/react-scheduling`, for two main reasons:
- these still seem pretty tied to the UI use case (the server code paths introduce a separate intermediate representation that normalizes "allDay" into "00:00 – 00:00" so it can operate exclusively on an hours basis).
- Making changes to core is heavier weight: it's more likely to be relied on by other parties, and any change there redeploys practically the entire Medplum stack. So we want to generally be more conservative there.

For those reasons, I like the idea of pushing this code down into `react-scheduling` and letting it "bake" there (especially since that package's README calls out that it's  alpha and we can make big changes while we are rapidly iterating).

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
```diff
@@ -0,0 +1,432 @@
...
+      </Button>
+    </Tooltip>
+  );
+
+  return (
+    <Stack gap="lg">
+      <Text c="dimmed">
+        {editingDefault
+          ? `Set the default weekly working hours for ${serviceName}. Every calendar without hours of its own follows these.`
+          : `Customize your weekly working hours to override the general availability for ${serviceName}.`}
```
noahsilas: I think we should avoid assuming that the viewer is the schedule's actor (consider cases like a manager or a scheduler who might edit someone else's availability, or when editing the schedule for a Device or Location). Can we drop "your" from here?

packages/core/src/scheduling.ts
Filupnot: I added a validation check for this on all the single-response ones (startTime, endTime, and allDay)

packages/react-scheduling/src/ScheduleAvailabilityEditor/ScheduleAvailabilityEditor.tsx
Filupnot: Updated it to this: 
> Customize the weekly working hours on this calendar, in place of the default availability for `${serviceName}`.

PR #9992, opened 2026-07-27
Document default access policies and admin reconciliation

maddyli:
## Summary

Adds documentation for the new **default access policies** to the [Access Policies](https://www.medplum.com/docs/access/access-policies) doc, at the request of the docs owner.

The new section covers:

- **Default Access Policies** — a table of the per-profile-type defaults stored on `Project.defaultAccessPolicies` (`Practitioner`, `Admin`, `Patient`, `RelatedPerson`), including what each grants and that they are applied automatically at invite time when no explicit policy is provided.
- **Switching a user to (or from) admin** — explains that the `admin` flag does not bypass the access policy (it only adds admin resource types on top), and that Medplum automatically reconciles the policy between the `Practitioner` and `Admin` defaults when the flag is toggled in either direction. Custom (non-default) policies are never modified.
- **Custom access policy combinations** — recommends overriding the default by supplying an explicit access policy at invite time, which also opts the user out of automatic reconciliation.

This also replaces the large inline Patient Access Policy Template JSON block with a pointer to the Default Access Policies section.

## Test plan

- [ ] Docs build / preview renders the new section correctly
- [ ] Internal links resolve (verified referenced resource and `medplum` API pages exist locally)

### DIFF
--- a/packages/docs/docs/access/access-policies.md
+++ b/packages/docs/docs/access/access-policies.md
@@ -675,7 +675,7 @@ To avoid this, Medplum automatically **reconciles the default access policy when
 - Promoting a user to admin: if their current policy is exactly the project's `Practitioner` default, it is swapped to the `Admin` default.
 - Demoting an admin: if their current policy is exactly the project's `Admin` default, it is swapped back to the `Practitioner` default.
 
-This reconciliation only applies when the user is currently on a recognized role default. **Custom (non-default) access policies are never modified** — if you have assigned a user a bespoke policy, toggling the admin flag leaves that policy untouched.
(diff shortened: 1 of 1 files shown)


packages/docs/docs/access/access-policies.md
```diff
@@ -550,107 +550,7 @@ Patient Access is disabled by default. See our article on [enabling open patient
 
 :::
 
-:::danger[Binary Access]
-
-Binary resources cannot use compartment-based access controls. They require explicit `securityContext` declaration. See the [Binary Security Context](/docs/access/binary-security-context) documentation for more information.
```
codyebberson, line 555: I see this already existed below 👍

codyebberson (approved):

PR #9993, opened 2026-07-27
Share _include / _revinclude between the SQL and in-memory repositories

codyebberson:
`MemoryRepository` silently ignored `_include` and `_revinclude`. A search with those parameters returned only the matched resources, with no indication that the include directives had been dropped.

This adds support by moving the include logic out of `packages/server` and into `packages/fhir-router`, where both repositories can share it, rather than writing a second implementation.

## Background

The include implementation lived in three functions in `packages/server/src/fhir/search.ts` — `getExtraEntries`, `getSearchIncludeEntries`, and `getSearchRevIncludeEntries`. Duplicating them for `MemoryRepository` would have meant a second copy of the `:iterate` depth limiting, the seen-set deduplication, target-type filtering, and the canonical-versus-literal reference handling — all of it subtle, and all of it exactly the kind of thing that drifts between the two implementations.

Auditing their dependencies showed almost nothing was actually storage-specific:

| Dependency | Where it lived | Portable |
| --- | --- | --- |
| `flatMapFilter`, `Operator.ITERATE`, `getSearchParameter`, `evalFhirPathTyped`, `toTypedValue`, `PropertyType`, `isResource`, `getReferenceString`, `badRequest` | `@medplum/core` | yes |
| `maxSearchResults` | alias for core's `DEFAULT_MAX_SEARCH_COUNT` | yes |
| `canonicalReferenceTypes` | one-line local const over core's `PropertyType` | yes |
| `getCanonicalUrl` | `'url' in resource ? resource.url : undefined` | yes |
| `repo.readReferences(...)` | already abstract on `FhirRepository` | yes |
| `getSearchParameterImplementation(...).type` | server, but `SearchParameterImplementation extends SearchParameterDetails` and the impl is built by core's `getSearchParameterDetails` (`searchparameter.ts:121`) | yes, via substitution |
| `getSelectQueryForSearch` + `getSearchEntries` | server SQL | behind a seam |
| `getFullUrl` | server, needs `getConfig().baseUrl` | no |

So the whole body of logic reduced to two injection points.

### New: `packages/fhir-router/src/search-include.ts`

The three functions, logic unchanged, rewritten against the `FhirRepository` abstract base. Repository-specific behavior is supplied through `IncludeOptions`:

```ts
export interface IncludeOptions {
  readonly fullUrl?: (resourceType: string, id: string) => string;
  readonly maxDepth?: number;
  readonly maxResults?: number;
  readonly executeSearch?: IncludeSearchFn;
}
```

- **`fullUrl`** — the only genuinely non-portable dependency. The server passes its config-aware `getFullUrl`; `MemoryRepository` omits it, and the field is left off the entry, matching that repository's existing behavior of not emitting `fullUrl`.
- **`executeSearch`** — runs the sub-searches needed for canonical `_include` and for every `_revinclude`. Defaults to `repo.searchResources()`. The server passes its existing `getSelectQueryForSearch` + `getSearchEntries` path, so its hot path is unchanged; `Repository.search()` is never re-entered, and no bundle links are built for sub-searches.

`getSearchParameterImplementation(...).type === SearchParameterType.CANONICAL` became `getSearchParameterDetails(...).type === SearchParameterType.CANONICAL`. These return the same value — the server's implementation registry is populated from core's `getSearchParameterDetails`.

`maxDepth` (5) and `maxResults` (`DEFAULT_MAX_SEARCH_COUNT`) default to the server's previous hardcoded limits, including the existing behavior of *throwing* rather than truncating when a chain exceeds them.

### `packages/server/src/fhir/search.ts`

The three functions, the local `getCanonicalUrl`, and the `canonicalReferenceTypes` const are removed. What remains is the options factory:

```ts
function getIncludeOptions(repo: Repository): IncludeOptions {
  return {
    fullUrl: getFullUrl,
    executeSearch: async (searchRequest: SearchRequest): Promise<WithId<Resource>[]> => {
      const trackedResourceTypes = new Set<ResourceType>();
      const query = getSelectQueryForSearch(repo, searchRequest, { trackedResourceTypes });
      const result = await getSearchEntries(
        repo,
        searchRequest as SearchRequestWithCountAndOffset,
        query,
        trackedResourceTypes
      );
      return result.entry.map((entry) => entry.resource as WithId<Resource>);
    },
  };
}
```

`removeResourceFields` is applied to included entries in exactly the same places and the same number of times as before.

### `packages/fhir-router/src/repo.ts`

- `MemoryRepository.search()` calls the shared `getExtraEntries` when `include` or `revInclude` is present and the page has at least one match.
- Match entries are now tagged `search: { mode: 'match' }`. Without this there is no way to tell a match from an include in the resulting bundle.
- `total` continues to reflect only the matched resources.
- `readReferences` no longer rejects the entire batch when a single reference is unresolvable. It was `Promise.all(refs.map(r => this.readReference(r)))` despite a declared return type of `(T | Error)[]`; unresolvable references now come back as errors, which is what the include code's `.filter(isResource)` expects. This was a live bug independent of `_include`.

## Behavior changes worth noting

**Canonical `_include` no longer over-matches in `MemoryRepository`.** The canonical branch searches `url` with `Operator.EQUALS`. `url` is a `uri` parameter, which `matchesSearchRequest` routes to `matchesStringValue` — a case-insensitive **substring** test. An `_include` targeting `http://example.com/q` would therefore also pull in `http://example.com/q-extra`.

This is fixed in the shared module by post-filtering canonical results against the exact requested URL set, rather than by changing `match.ts`. Fixing `uri` matching in core would also change subscription criteria evaluation, which is a larger blast radius and deserves its own change. The post-filter is a no-op on the server, where `url` already matches exactly.

**`MemoryRepository` bundles now carry `search.mode` on match entries.** This is the change with real blast radius, since `MockClient` is used throughout `react`, `app`, and the example apps. Every consuming suite passes unchanged.

## Not addressed

Two pre-existing `MemoryRepository` fidelity gaps, both outside the scope of `_include`:

- `matchesSearchRequest` treats `uri` parameters as substring matches everywhere other than the canonical path fixed here.
- `matchesSearchRequest` returns `false` for `number`, `quantity`, and composite search parameter types.

## Testing

New: `packages/fhir-router/src/search-include.test.ts` (14 tests) covering literal `_include`, dangling references, target-type filtering, deduplication, resources already present as matches, `_revinclude`, `_include:iterate` (with a companion test asserting non-`iterate` params do *not* recurse), canonical include and revinclude, exact-URL canonical matching, invalid include parameters, and the empty-match-set short circuit.

The canonical exact-match filter was verified to be load-bearing by neutering the condition and confirming the test fails with both questionnaires included.

```
npm test --workspace=@medplum/fhir-router          # 136 passed (14 new)
npm test --workspace=@medplum/server -- search.test.ts   # 487 passed (20 include-specific)
npm test --workspace=@medplum/mock                 # 90 passed
npm test --workspace=@medplum/react                # 1244 passed
npm test --workspace=@medplum/react-hooks          # 234 passed
npm test --workspace=@medplum/app                  # 458 passed
```

Also run green, as `@medplum/mock` consumers: `cli` (192), `agent` (544), `dosespot-core` (31), `dosespot-react` (63), `health-gorilla-core` (27), `health-gorilla-react` (23), `scriptsure-react` (30).

`tsc --noEmit` and `eslint` clean on `@medplum/fhir-router` and `@medplum/server`.

### DIFF
--- a/packages/fhir-router/src/index.ts
+++ b/packages/fhir-router/src/index.ts
@@ -4,4 +4,5 @@ export { BatchProcessor, buildBatchResponseBundle, processBatch } from './batch'
 export type { BatchEvent, BatchInitialState, BundlePreprocessInfo, LogEvent } from './batch';
 export * from './fhirrouter';
 export * from './repo';
+export * from './search-include';
 export * from './urlrouter';
--- a/packages/fhir-router/src/repo.ts
+++ b/packages/fhir-router/src/repo.ts
@@ -21,7 +21,16 @@ import {
   preconditionFailed,
   stringify,
 } from '@medplum/core';
-import type { Bundle, OperationOutcome, Parameters, Reference, Resource, ResourceType } from '@medplum/fhirtypes';
+import type {
+  Bundle,
+  BundleEntry,
+  OperationOutcome,
+  Parameters,
+  Reference,
+  Resource,
+  ResourceType,
+} from '@medplum/fhirtypes';
+import { getExtraEntries } from './search-include';
 
 export type CreateResourceOptions = {
   assignedId?: boolean;
@@ -586,7 +595,15 @@ export class MemoryRepository extends FhirRepository {
   async readReferences<T extends Resource>(
     references: readonly Reference<T>[]
   ): Promise<(T | OperationOutcomeError)[]> {
-    return Promise.all(references.map((r) => this.readReference<T>(r)));
+    // Unresolvable references are returned as errors rather than rejecting the whole batch,
+    // matching the `(T | Error)[]` contract on FhirRepository.
+    return Promise.all(
+      references.map(async (r) =>
+        this.readReference<T>(r).catch((err) =>
+          err instanceof OperationOutcomeError ? err : new OperationOutcomeError(normalizeOperationOutcome(err))
+        )
+      )
+    );
   }
 
   async readHistory<T extends Resource>(resourceType: string, id: string): Promise<Bundle<T>> {
@@ -622,7 +639,10 @@ export class MemoryRepository extends FhirRepository {
         result.push(resource);
       }
     }
-    let entry = result.map((resource) => ({ resource: deepClone(resource) }));
+    let entry = result.map((resource): BundleEntry<WithId<T>> => ({
+      search: { mode: 'match' },
+      resource: deepClone(resource),
+    }));
     for (const sortRule of searchRequest.sortRules ?? EMPTY) {
       entry = entry.sort((a, b) => sortComparator(a.resource as T, b.resource as T, sortRule));
     }
@@ -641,7 +661,15 @@ export class MemoryRepository extends FhirRepository {
   }
 
   async search<T extends Resource>(searchRequest: SearchRequest<T>): Promise<Bundle<WithId<T>>> {
-    return this.searchSync(searchRequest);
+    const bundle = this.searchSync(searchRequest);
+    if ((searchRequest.include || searchRequest.revInclude) && bundle.entry?.length) {
+      // `total` intentionally continues to reflect only the matched resources.
+      const entries = bundle.entry as BundleEntry[];
+      const resources = entries.map((e) => e.resource as WithId<T>);
+      await getExtraEntries(this, searchRequest, resources, entries);
+      bundle.entry = entries as BundleEntry<WithId<T>>[];
+    }
+    return bundle;
   }
 
   async conditionalCreate<T extends Resource>(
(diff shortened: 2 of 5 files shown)


codyebberson:
@claude - please review

Main considerations:
* Changes to `server` should be 100% pure refactoring, no behavior change
* Changes to `MemoryRepository` should be completely additive
* Please look for opportunities to tighten this up, keep things surgical, keep things minimal

mattwiller (approved):

PR #9995, opened 2026-07-28
Respect Slot.serviceType when subtracting busy slots from availability

### LINKED ISSUE #10473: Respect Slot.serviceType when subtracting busy slots from availability
## Problem

The [Defining Availability docs](https://www.medplum.com/docs/scheduling/defining-availability#blocking-time-by-service-type) state, for a busy Slot:

> - **With serviceType**: Blocks only that specific service
> - **Without serviceType**: Blocks all services

But [`applyExistingSlots()`](https://github.com/medplum/medplum/blob/74576f6c837ee724f39267c7a0a242139cb6b2fc/packages/server/src/fhir/operations/utils/scheduling.ts#L325-L346) only applies the `hasMatchingServiceType()` filter to **free** Slots — every `busy`/`busy-tentative`/`busy-unavailable` Slot is subtracted from availability unfiltered. A busy Slot scoped to one service therefore blocks the entire Schedule, contradicting the documented behavior.

### Reproduction (observed on `5.1.27-505140d`)

1. Create one `Schedule` linked to `HealthcareService` A and B via `serviceType` (per the [service-level availability docs](https://www.medplum.com/docs/scheduling/defining-availability#service-level-availability)).
2. [`$find`](https://www.medplum.com/docs/scheduling/appointment-find) returns the same start for both services.
3. Create a `busy-unavailable` Slot for that start with `serviceType` referencing service A only.
4. `$find` for service **B** also loses the start.

## Solution

Apply the same `hasMatchingServiceType()` filter to the busy-family branch. The helper already treats Slots without a `serviceType` as wildcards, so:

- a busy Slot **without** `serviceType` still blocks all services (unchanged), and
- a busy Slot **with** `serviceType` now blocks only matching services (the documented behavior).

Because `validateAvailability()` flows through the same function, `$book` validation becomes consistent with `$find` for free.

## In Progress
This issue is being created to track a problem raised by #9995 and allow that PR to be resurrected after the PR Vouch system closed it.

nickcatal:
## Problem

The [Defining Availability docs](https://www.medplum.com/docs/scheduling/defining-availability#blocking-time-by-service-type) state, for a busy Slot:

> - **With serviceType**: Blocks only that specific service
> - **Without serviceType**: Blocks all services

But [`applyExistingSlots()`](https://github.com/medplum/medplum/blob/74576f6c837ee724f39267c7a0a242139cb6b2fc/packages/server/src/fhir/operations/utils/scheduling.ts#L325-L346) only applies the `hasMatchingServiceType()` filter to **free** Slots — every `busy`/`busy-tentative`/`busy-unavailable` Slot is subtracted from availability unfiltered. A busy Slot scoped to one service therefore blocks the entire Schedule, contradicting the documented behavior.

### Reproduction (observed on `5.1.27-505140d`)

1. Create one `Schedule` linked to `HealthcareService` A and B via `serviceType` (per the [service-level availability docs](https://www.medplum.com/docs/scheduling/defining-availability#service-level-availability)).
2. [`$find`](https://www.medplum.com/docs/scheduling/appointment-find) returns the same start for both services.
3. Create a `busy-unavailable` Slot for that start with `serviceType` referencing service A only.
4. `$find` for service **B** also loses the start.

## Solution

Apply the same `hasMatchingServiceType()` filter to the busy-family branch. The helper already treats Slots without a `serviceType` as wildcards, so:

- a busy Slot **without** `serviceType` still blocks all services (unchanged), and
- a busy Slot **with** `serviceType` now blocks only matching services (the documented behavior).

Because `validateAvailability()` flows through the same function, `$book` validation becomes consistent with `$find` for free.

## Testing

Added four unit tests mirroring the existing free-slot serviceType suite: wildcard busy Slots block any requested service; matching-service busy Slots are removed from availability; non-matching `busy` and `busy-unavailable` Slots no longer block. The two non-matching tests fail against the previous implementation and pass with this change; all 62 tests in `scheduling.test.ts` pass.


codyebberson:
Thanks @nickcatal !

noahsilas:
Good catch on this inconsistency. I think in the original implementation we've made a bit of a mistake that is going to make fixing this a little trickier though; in the `$find` operation, we return schedulable results with proposed Slot resources that include the `serviceType` set to the value that was requested.

This made it easy to understand the origin of the slots, but when we apply this fix then suddenly already booked appointments would not prevent appointments of other types from booking over the same times.

I'm going to spend a little time thinking about a pathway forwards here. I suspect that this may need to be a bit of a phased release for backwards compatibility - maybe we stop writing `serviceType` on new slots via $find/$book first, announce this pending change with some migration steps, and then follow up with this after folks have had a couple weeks to get their data up to date?

Thanks for flagging this and bearing with us while we work out a good path forward!

techdavidy (approved):

noahsilas (changes requested):
The change and tests are reasonable, but I'm marking as not approved purely so we don't merge too early; this is pending some community communication with a migration strategy to handle existing appointments created via `$find`/`$book` that should not allow over-booking with alternate service types. (See also #9998 which stopped writing `serviceType` on new $find/$book requests in preparation for this).

noahsilas:
Following up on timing: we're planning on merging this on Sept. 1, giving existing scheduling users time to remove any existing `Slot.serviceType` fields that were incorrectly set. We've published docs about this change at https://www.medplum.com/docs/scheduling/beta-changelog to help folks prepare.

noahsilas (approved):

noahsilas:
Thanks for the patience @nickcatal. We appreciate your contribution making the implementation match the docs more closely! Apologies for the noise on the PR as we worked through some backwards compatibility and tooling issues, I'm happy to have this merged now!

PR #9997, opened 2026-07-28
feat(server): generate an MFA secret on demand at enrollment

ThatOneBro:
## Problem

A TOTP secret is only ever created for a `User` in two places: at invite time when `mfaRequired: true` is passed, and in `GET /auth/mfa/status`. Any user who comes under an MFA requirement *later* has no secret — most notably when the project-wide `mfaRequired` setting (#9678) is enabled, which applies to every existing member of the Project.

For those users, `sendLoginResult` read `user.mfaSecret as string` without generating one, so login returned an unusable enrollment URI and `login-enroll` rejected the code:

```
POST /auth/login
→ { "mfaEnrollRequired": true,
    "enrollUri": "otpauth://totp/...?secret=undefined&period=30&..." }

POST /auth/mfa/login-enroll  { method: "totp", token: "..." }
→ 400 { "details": { "text": "Secret not found" } }
```

Because the login could not complete, the member could not reach the Security page to have a secret generated either. The only way through was email-based MFA (`allowedMfaMethods: 'totp,email'`), which does work — so enabling project-wide `mfaRequired` on a Project whose members use authenticator apps locked them out.

## Fix

Generate and persist the secret when the enrollment payload is built. `GET /auth/mfa/status` already did exactly this, so both call sites now share a `buildTotpEnrollment` helper in `auth/utils.ts`. Users who already have a secret keep it, so enrollment URIs stay stable across repeated calls.

## Tests

`login-enroll requires a secret for TOTP enrollment` asserted the old behavior; it is replaced by:

- `login-enroll generates a secret for a user who does not have one` — a user with no secret is offered a real secret at login and completes authenticator enrollment end-to-end.
- `login-enroll rejects TOTP enrollment if the secret is removed mid-flow` — keeps coverage of the remaining `Secret not found` guard, which is now only reachable if the secret disappears between login and enrollment.

`src/auth` and `src/oauth` pass (512 tests). `src/auth/newproject.test.ts > GraphQL is restricted to project` fails identically on unmodified `main` in my local environment (`permission denied for table Patient`) and is unrelated.

## Docs

Documentation for the project-wide `mfaRequired` setting is in a separate PR. It currently carries a caution about this gap, which I will drop once this lands.

### DIFF
--- a/packages/server/src/auth/mfa.test.ts
+++ b/packages/server/src/auth/mfa.test.ts
@@ -1597,7 +1597,7 @@ describe('MFA', () => {
     expect(res.body).toMatchObject(badRequest('Already enrolled'));
   });
 
-  test('login-enroll requires a secret for TOTP enrollment', async () => {
+  test('login-enroll generates a secret for a user who does not have one', async () => {
     const email = `login-enroll${randomUUID()}@example.com`;
     const password = 'password!@#';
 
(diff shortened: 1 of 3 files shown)


codyebberson (approved):
Nice, surgical 🔪

PR #10005, opened 2026-07-28
feat(server): add queue activeCount, per-queue in-flight and completed job metrics, and async batch entry throughput

ThatOneBro:
## Summary

Adds queue observability in three parts.

**1. `medplum.<queue>.activeCount`** — a gauge alongside the existing waiting and delayed counts in the OpenTelemetry heartbeat.

**2. `medplum.<queue>.inFlightJobs` and `medplum.<queue>.jobsCompleted`** — worker-side metrics for every one of the ten workers, maintained by a new `trackJobMetrics` processor wrapper. `inFlightJobs` is an UpDownCounter moved up as a job starts and down as it settles; `jobsCompleted` is a monotonic counter of jobs whose processor ran to the end without throwing.

This wraps the processor rather than listening for the worker's `active`/`completed`/`failed` events, because **those events do not pair up**:

- BullMQ's `handleFailed` returns early on `DelayedError` without emitting, so every job re-queued by `moveToDelayedAndThrow` (i.e. every batch job surviving a deploy) would leak `+1` on the in-flight counter.
- Both terminal events are suppressed entirely once `connection.closing` is set.

A `finally` block runs on every exit path, so the in-flight increment and decrement always pair. A processor that throws counts as neither completed nor in flight, which correctly excludes delayed re-queues — the attempt that eventually runs to the end is the one counted. (Side note: the existing `failedErr instanceof DelayedError` guard in `batch.ts` is dead code for the same reason — left alone here.)

The wrapper declares and forwards all three processor parameters, because BullMQ reads `processor.length >= 3` to decide whether to supply an abort signal. A two-parameter wrapper would silently disable cancellation for any processor that accepts one. There's a test pinning the arity.

**3. `medplum.batch.entriesProcessed`** — a monotonic counter of individual bundle entries, which stays batch-specific since no other queue has a sub-job unit of work to count.

Rates are deliberately not computed in the server. Exporting counters lets the observability backend derive throughput over whatever window it's asked about, and gives a series that isn't tied to a single process's lifetime. The otel README gains a note making that the standing rule.

Also adds `getUpDownCounter`/`addToUpDownCounter` helpers and documents the UpDownCounter instrument in the README.

## Naming

Per-queue metric names go through `getQueueMetricName(queueName, metric)`, keeping the queue in the metric name rather than an attribute, to match how `waitingCount` and friends have always been reported. Typing the first parameter as `WorkerName` also tightened `queueEntries` from `[string, Queue][]` to `[WorkerName, Queue][]`.

The `QueueMetric` type documents why the Redis-read gauges and the worker-side metrics **are not expected to agree**: `activeCount` counts everything Redis considers active across every instance, including jobs leased by a host that has since died and whose lock hasn't expired, while `inFlightJobs` counts only what a given process currently has on the stack.

## Behavioral notes for review

- **`jobsCompleted` is a BullMQ-level signal.** A job whose processor handled a failure internally and returned normally still counts as completed. For async batches specifically this is a change in meaning: the earlier batch-only counter incremented only after `exec.completeJob` succeeded, so a batch whose AsyncJob failed did not count. The generic semantics are the right ones for a metric shared across every queue, and the type docs say so.
- **`entriesProcessed` only covers the re-entrant batch path.** The deprecated legacy path hands the whole bundle to the router in one call, so there's no per-entry hook.
- **`getActiveCount` was added to the shared bullmq mock.** bullmq is mocked globally for every server test and `app.ts` inits the heartbeat during `initAppServices`, so without it the new gauge threw inside the heartbeat listener for any test that lets a heartbeat fire.

## Test plan

- [x] `src/otel/otel.test.ts` and `src/workers/utils.test.ts` — 43 passing. Covers in-flight pairing on resolve, reject and delayed re-queue; that a completion is counted only on a clean return; the arity guard; per-queue metric naming; and the heartbeat publishing each queue count under the right metric name.
- [x] `tsc --noEmit`, `eslint`, `prettier --check` clean.
- [ ] **DB-backed worker suites not run locally** — no container runtime available, so `batch.test.ts` and friends fail in `seedDatabase`. All ten worker constructors are touched and `entriesProcessed` is only exercised on those paths, so these need a green CI run before merge.

### DIFF
--- a/packages/server/src/otel/README.md
+++ b/packages/server/src/otel/README.md
@@ -26,9 +26,9 @@ OpenTelemetry supports publishing four primary types of metrics:
   on demand.
   - e.g. jobs currently in flight, incremented as each starts and decremented as it settles
 
-> **NOTE**: Prefer a Counter over deriving a rate in the server: the observability backend can compute rates from a
-> monotonic total, and it does so without being tied to this process's lifetime. Compute a rate here only when the
-> underlying total is not itself exported as a metric.
+> **NOTE**: Do not compute rates ("X per second") in the server. Export a Counter and let the observability backend
+> derive the rate: it can then do so over whatever window it is asked about, rather than the one hard-coded here, and
+> the series is not tied to a single process's lifetime.
 
 [otel]: https://opentelemetry.io/docs/what-is-opentelemetry/
 
--- a/packages/server/src/otel/README.md
+++ b/packages/server/src/otel/README.md
@@ -17,18 +17,17 @@ OpenTelemetry supports publishing four primary types of metrics:
 
 - **Counter**: Incrementing count, useful for recording when an event occurs; can be post-processed to give rates etc.
   - e.g. Bot executions, created resources
-- **Histogram**: Event-specific metrics with a numeric value, allows calculating stats like average and percentiles
-  - e.g. request latency (in seconds), uploaded resource size (in bytes)
-- **Gauge**: A value that has meaning at the time it's measured, not associated with an event
-  - e.g. CPU usage (in percent), heap size (in bytes)
 - **UpDownCounter**: Like a Counter, but the reported deltas may be negative, so the summed total rises and falls.
   Use this instead of a Gauge for a quantity that is only observable at the moments it changes, rather than readable
   on demand.
   - e.g. jobs currently in flight, incremented as each starts and decremented as it settles
+- **Histogram**: Event-specific metrics with a numeric value, allows calculating stats like average and percentiles
+  - e.g. request latency (in seconds), uploaded resource size (in bytes)
+- **Gauge**: A value that has meaning at the time it's measured, not associated with an event
+  - e.g. CPU usage (in percent), heap size (in bytes)
 
-> **NOTE**: Do not compute rates ("X per second") in the server. Export a Counter and let the observability backend
-> derive the rate: it can then do so over whatever window it is asked about, rather than the one hard-coded here, and
-> the series is not tied to a single process's lifetime.
+> **NOTE**: Do not compute rates ("X per second") in the server. Export a Counter
+> and let the observability backend derive the rate over whatever window is requested
 
 [otel]: https://opentelemetry.io/docs/what-is-opentelemetry/
 
--- a/packages/server/src/otel/README.md
+++ b/packages/server/src/otel/README.md
@@ -13,7 +13,7 @@ flowchart LR
     C -->|Aggregated data| O[Observability]
 ```
 
-OpenTelemetry supports publishing three primary types of metrics:
+OpenTelemetry supports publishing four primary types of metrics:
 
 - **Counter**: Incrementing count, useful for recording when an event occurs; can be post-processed to give rates etc.
   - e.g. Bot executions, created resources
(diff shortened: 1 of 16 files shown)


packages/server/src/otel/README.md
```diff
@@ -13,14 +13,22 @@ flowchart LR
...
 - **Counter**: Incrementing count, useful for recording when an event occurs; can be post-processed to give rates etc.
   - e.g. Bot executions, created resources
 - **Histogram**: Event-specific metrics with a numeric value, allows calculating stats like average and percentiles
   - e.g. request latency (in seconds), uploaded resource size (in bytes)
 - **Gauge**: A value that has meaning at the time it's measured, not associated with an event
   - e.g. CPU usage (in percent), heap size (in bytes)
+- **UpDownCounter**: Like a Counter, but the reported deltas may be negative, so the summed total rises and falls.
+  Use this instead of a Gauge for a quantity that is only observable at the moments it changes, rather than readable
+  on demand.
+  - e.g. jobs currently in flight, incremented as each starts and decremented as it settles
```
mattwiller, line 23: I'd suggest moving this directly under `Counter`

packages/server/src/otel/README.md
```diff
@@ -13,14 +13,22 @@ flowchart LR
...
+
+> **NOTE**: Do not compute rates ("X per second") in the server. Export a Counter and let the observability backend
+> derive the rate: it can then do so over whatever window it is asked about, rather than the one hard-coded here, and
+> the series is not tied to a single process's lifetime.
```
mattwiller:
```suggestion
> **NOTE**: Do not compute rates ("X per second") in the server. Export a Counter
> and let the observability backend derive the rate over whatever window is requested
```

packages/server/src/workers/post-deploy-migration.ts
```diff
@@ -66,7 +67,9 @@ export const initPostDeployMigrationWorker: WorkerInitializer = (config, options
   if (options?.workerEnabled !== false) {
     worker = new Worker<PostDeployJobData>(
       PostDeployMigrationQueueName,
-      async (job) => tryRunInRequestContext(job.data.requestId, job.data.traceId, async () => jobProcessor(job)),
+      trackInFlightJobs('post-deploy-migration', async (job) =>
+        tryRunInRequestContext(job.data.requestId, job.data.traceId, async () => jobProcessor(job))
+      ),
```
mattwiller: For workers like this one (and possibly `reindex` as well), where we're using the queue more for durability than load smoothing, I wonder if it would make sense to omit the tracking/metrics for now

packages/server/src/workers/utils.test.ts
```diff
@@ -348,6 +351,68 @@ describe('worker utils', () => {
...
+    afterEach(() => {
+      counterSpy.mockRestore();
+    });
+
+    /**
+     * Collects the signed deltas reported to a worker's in-flight metric.
+     * @param workerName - The worker whose metric to read.
+     * @returns The deltas, in the order they were reported.
+     */
+    function reportedDeltas(workerName: WorkerName): number[] {
```
mattwiller:
```suggestion
    function reportedDeltas(workerName: WorkerName): number[] {
```

packages/server/src/otel/otel.ts
```diff
@@ -13,8 +14,42 @@ import { getDownloadQueue } from '../workers/download';
...
+ *
+ * The queue is part of the metric name rather than an attribute, which is how these metrics have
+ * always been reported. Prefer this over interpolating a name by hand so every per-queue metric
+ * stays consistent and greppable.
+ * @param queueName - The queue being measured.
+ * @param metric - The quantity being measured.
+ * @returns The metric name, e.g. `medplum.batch.activeCount`.
+ */
+export function getQueueMetricName(queueName: WorkerName, metric: QueueMetric): string {
+  return `medplum.${queueName}.${metric}`;
```
mattwiller:
```suggestion
export function getQueueMetricName(queueName: WorkerName, metric: QueueMetric): string {
  return `medplum.${queueName}.${metric}`;
```

packages/server/src/otel/otel.ts
```diff
@@ -105,6 +141,34 @@ export function setGauge(name: string, value: number, options?: RecordMetricOpti
...
+}
+
+/**
+ * Adds a signed delta to an UpDownCounter.
+ *
+ * Unlike a gauge, which reports an absolute reading, the collector sums the reported deltas. That
+ * makes this the right instrument for a quantity that rises and falls and is only observable at the
+ * moments it changes -- e.g. work in flight, incremented as it starts and decremented as it settles.
+ * @param name - The metric name.
+ * @param n - The delta to add. Negative values decrement.
```
mattwiller: Suggest just renaming the parameter to `delta` to make this clearer

mattwiller (approved):

packages/server/src/workers/post-deploy-migration.ts
ThatOneBro: Done, dropped it from `reindex` too 👍

PR #10006, opened 2026-07-28
fix(provider): Add support for /new in faxes and tasks

techdavidy:
Screenshot 2026-07-28 at 4 31 03 PM


codyebberson:
@claude - please review

Any opportunities for code reuse or abstractions that could make these tighter?  IMHO, it already looks pretty good, but always worth zooming out.

techdavidy:
@claude please review again after addressing the feedback

codyebberson:
@techdavidy  - i think your branch might be in a bad place, might need to rebase?

examples/medplum-provider/src/hooks/useControllableDisclosure.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Like Mantine's `useDisclosure`, but optionally controlled by the parent.
+ * When `opened`/`onOpen`/`onClose` are provided they take precedence over the internal
+ * state, letting pages drive modal state externally (e.g. from the URL) while standalone
+ * usage keeps working without any props.
+ * @param controlled - Optional controlled state and handlers.
+ * @returns The resolved open state and open/close handlers.
+ */
+export function useControllableDisclosure(controlled: ControllableDisclosureProps): ControllableDisclosureState {
```
noahsilas, line 37: I'm having a hard time understanding the use case here; it seems like this is to be used when we are unsure if the state is controlled, and so we create internal state just in case it is not?

I see that this is getting used from TaskBoard, which itself can be mounted in different contexts (both as a standalone page at `/Task` and nested in a patient context at `/Patient/:id/Task`). Is the intent here that we are supporting one of those scenarios controlling the disclosure itself and one of them delegating to the internal management?

examples/medplum-provider/src/hooks/useNewInUrl.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @returns The current `/new` state and open/close navigation handlers.
+ */
+export function useNewInUrl(basePath: string, suffix: string): NewInUrlState {
+  const location = useLocation();
+  const navigate = useNavigate();
+  return {
+    isNew: location.pathname.endsWith('/new'),
+    openNew: (): void => {
+      navigate(`${basePath}/new${suffix}`)?.catch(console.error);
+    },
```
noahsilas: #nit: this could get wrapped in a `useCallback` hook so that callers get a more stable function reference.

noahsilas (approved):

examples/medplum-provider/src/hooks/useControllableDisclosure.ts
techdavidy: Ideally every navigation is url supported. Navigation to /new would allow to open modal directly. Although it allows to be able to handle internally if no url is required.

PR #10016, opened 2026-07-29
chore: add comment-compactor Claude skill

ThatOneBro:
Adds a repo-scoped Claude Code skill that runs a dedicated pass over the comments a diff adds, and un-ignores `.claude/skills/` so skills like it can be shared.

## Why

Comments are the only part of a diff that no tool checks — eslint, prettier, `tsc`, and the test suite all pass whether a comment is useful, bloated, or already false. They are also the part with the longest half-life: code gets refactored away, but a stale comment survives and misleads the next reader.

The failure mode this targets is comments that narrate *the session that produced them* rather than the code that shipped — "the pre-rework shape", "unlike the previous approach", "as discussed". Those read fine during review and are meaningless a year later, because the reader has no access to the thing being contrasted against.

## What the skill does

Triggers as the final step before `git push` / `gh pr create`, after the mechanical gates are green. It:

1. Extracts every comment line the diff adds, as `file:line: comment`, via `git diff --merge-base main -U0` piped through awk.
2. Judges each against four questions — insightful (explains a *why*, not a restatement), durable (readable by someone who never saw this PR), compact, and correctly placed.
3. Also checks comments the diff did **not** touch but whose code it rewrote. Those are the likeliest to be outright false, and a diff-only pass misses them.

Includes a rewrite-patterns table and guidance on the highest-value case: comments explaining why the code is *not* something else. Those prevent future regressions, but only when they name the invariant rather than the diff.

## `.gitignore` change

`.claude/` becomes `.claude/*` plus `!.claude/skills/`. The trailing-slash form has to go — git will not descend into an excluded directory, so a negation inside it can never match.

`.claude/settings.local.json` and `.claude/worktrees/` remain ignored; verified with `git check-ignore`.

## Note for reviewers

This is two decisions, not one: the skill itself, and whether the repo hosts shared Claude skills at all. Happy to drop the second and keep this as a personal skill if the answer is no.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,7 +107,8 @@ playwright/.cache/
 .cursor/
 
 # Claude
-.claude/
+.claude/*
+!.claude/skills/
 CLAUDE.md
 AGENTS.md
 
--- a/.claude/skills/comment-compactor/SKILL.md
+++ b/.claude/skills/comment-compactor/SKILL.md
@@ -1,0 +1,109 @@
+---
+name: comment-compactor
+description: Review and rewrite every comment a diff adds so it is insightful, durable, and compact. Use as the final step before pushing, submitting, or finalizing a PR — after tests and lint pass, before `git push` / `gh pr create`. Also use when asked to "clean up the comments", "check the comments", or when a review flags a comment as stale, obvious, or conversation-bound.
+user_invocable: true
+---
+
+# comment-compactor
+
+A dedicated pass over the comments a diff adds, run as its own step — not folded into writing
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -107,7 +107,8 @@ playwright/.cache/
 .cursor/
 
 # Claude
-.claude/
+.claude/*
+!.claude/skills/
```
techdavidy, line 111: 🫡

codyebberson (approved):

PR #10017, opened 2026-07-29
Use project branding in MFA emails and authenticator app entries

### LINKED ISSUE #9990: Allow per-project branding of MFA (authenticator app label and email)
**Background**
Medplum hardcodes "Medplum" branding in both MFA delivery methods, with no per-project override:

TOTP (authenticator app): the enrollment otpauth:// URL's issuer/account-name are literal constants, which authenticator apps (Google Authenticator, Authy, etc.) render directly as the entry label:

```
packages/server/src/auth/mfa.ts:99-102
packages/server/src/auth/utils.ts:201-204
```

Email MFA: the verification email's subject and body text are also hardcoded, in `packages/server/src/auth/utils.ts:80-102`

**Issue**
Projects using Medplum as an IdP with custom branding have no way to change either TOTP/email-based MFA branding.

**Impact**
Confusing/off-brand UX for white-labeled deployments. End users see an unfamiliar "Medplum" name during MFA verification.

**Proposed fix**
Similar to whitelabeled invite emails, which already interpolate Project.name into the subject/body, derive branding from the Project in both MFA code paths, falling back to "Medplum" when unset.

ThatOneBro:
## Problem

MFA content generated by the server hardcoded "Medplum", so users of a white-labeled project received "Your Medplum verification code" emails and scanned QR codes that added a `medplum.com` entry to their authenticator app.

## Change

A project can set an `appName` entry in `Project.setting` — the server-side counterpart to the app's build-time `MEDPLUM_APP_NAME`, which the server cannot read:

- **Emailed codes** — subject and body read "Your Acme Health verification code", signed "The Acme Health Team".
- **Authenticator app entries** — the enrollment QR code uses the app name as the TOTP issuer, which is what authenticator apps show as the entry title, listed under the user's email address:

  ```
  Acme Health
  alice@example.com          123 456
  ```

- **Email sender name** — when the project sends through its own SMTP relay, `appName` becomes the From display name, so the message shows as "Acme Health" in the inbox rather than a bare address. A display name in `smtpFromAddress` still wins. It is deliberately not applied to the server's own sender: a display name that disagrees with the sender domain is a signal mail clients treat as suspicious, so branding `medplum.com` as the customer would work against deliverability.

MFA code emails now also pass the project to `sendEmail`, so they go out through [project SMTP](https://www.medplum.com/docs/user-management/project-smtp) like other project-scoped emails rather than always using the server transport.

Projects without the setting keep today's behavior: emails say "Medplum" and the TOTP issuer stays `medplum.com`. The one visible change for them is that new authenticator entries are listed under the bare email address rather than `Medplum - alice@example.com`, since the issuer already carries the name.

Two inputs that would have produced a broken entry are handled: colons are stripped from the app name (the Key URI spec reserves the colon as the issuer/account separator, so `Acme: Health` would be parsed as issuer `Acme`), and users invited without an email address are identified by id rather than rendering the literal string `undefined`.

## Naming

`appName` matches `MEDPLUM_APP_NAME` / `getAppName()`, which is how the codebase already refers to the user-facing product name. The reader sits in `email/utils.ts` next to `getProjectSmtpConfig`, the other project-level email config reader; `email/utils.ts` imports nothing from `auth`, so `auth` reading the app name from it adds no import cycle. Login page branding stays where it is today, per client, in `ClientApplication.signInForm`.

## Tests

`auth/mfa.test.ts` — branded email and authenticator entry, branded QR code returned at login when the project requires MFA, unbranded defaults preserved, colon handling, and the missing-email case. `email/email.test.ts` — From display name under project SMTP, `smtpFromAddress` override, and no rebranding of the server sender. Unit coverage in `branding.test.ts` and `email/utils.test.ts`. 113 tests pass across the MFA, branding, and email suites.

## Not in this PR

- **A custom icon in the authenticator app is not achievable through the QR code.** The Key URI format has no standard image field, and Google Authenticator, Microsoft Authenticator, Apple Passwords, 1Password, and Bitwarden all derive the icon from the issuer string — either a letter avatar or a match against their own brand catalog. Only a few niche apps read a non-standard `image=` parameter, and `signInForm.logo` is served as an expiring presigned URL, so embedding it would break later and bloat the QR. Getting the issuer right is the only lever that reaches those apps. Rendering the project logo on our own enrollment screen (or into the center of the QR image) is the way to put a mark in front of the user, and belongs in the app package.
- Welcome, password reset, and invite emails still name Medplum; the docs call this out and point at Custom Emails.

## Docs

- `docs/auth/mfa.md` — new "Branding MFA emails and authenticator apps" section.
- `docs/self-hosting/branding.md` — a "Beyond the app UI" section explaining that the build-time vars are server-wide and what is branded per project or per client instead.
- `docs/user-management/project-smtp.md` — display-name support on `smtpFromAddress` and how `appName` interacts with it.
- Pointer added in `project-settings.md`.


packages/server/src/branding.ts
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
```
codyebberson: Any objections to moving `branding.ts` and `branding.test.ts` into `packages/server/src/util`?  

Trying to keep top level folders tidy

codyebberson:
Overall lgtm 👍

codyebberson (approved):

PR #10022, opened 2026-07-30
Fix chained search by `_compartment`

mattwiller:
Chained search linking through the `_compartment` search parameter (e.g. `Organization?_has:Patient:_compartment:_id=123`) was broken due to a pair of problems in sequence:

1. The synthetic `SearchParameter` stored in the `@medplum/core` parameter registry for `_compartment` was missing its `target` definition, which chained search depends on.  This field must be evaluated lazily, since resource types aren't fully populated until after the synthetic default search parameters for each one are generated
2. The chained search logic has been updated to support `_compartment` as a special case, since references for this parameter aren't currently written to the `*_References` lookup tables.  Instead, we join directly to the target resource table through the `compartments UUID[]` column and associated GIN index


AndyStoneman (approved):

packages/core/src/types.ts
```diff
@@ -198,6 +200,10 @@ function getOrInitTypeSchema(resourceType: string): TypeInfo {
         code: '_compartment',
         type: 'reference',
         expression: resourceType + '.meta.compartment',
+        get target(): string[] {
```
mattlong, line 206: Could you add a comment here explaining what's going on and why the lazy-loading is necessary?

packages/server/src/fhir/search.test.ts
```diff
@@ -1144,6 +1144,76 @@ describe.each<Project['features']>([undefined, ['range-search']])('project-scope
...
+        meta: { accounts: [createReference(organizationA), createReference(organizationB)] },
+      });
+      expect(patient.meta?.compartment).toContainEqual({ reference: getReferenceString(organizationA) });
+      expect(patient.meta?.compartment).toContainEqual({ reference: getReferenceString(organizationB) });
+
+      const searchResult = await repo.search(
+        parseSearchRequest(`Organization?_has:Patient:_compartment:_id=${patient.id}`)
+      );
+      // Both compartment Organizations returned
+      expect(searchResult.entry).toHaveLength(2);
```
mattlong:
```suggestion
```

mattlong (approved):

packages/server/src/fhir/search.ts
```diff
@@ -1745,7 +1745,10 @@ function linkCanonicalReference(selectQuery: SelectQuery, currentTable: string,
 function getCompartmentJoinCondition(currentTable: string, link: ChainedSearchLink, nextTable: string): Expression {
   const holder = link.direction === Direction.FORWARD ? currentTable : nextTable;
   const otherEnd = link.direction === Direction.FORWARD ? nextTable : currentTable;
-  return new Condition(new Column(otherEnd, 'id'), 'IN_SUBQUERY', new Column(holder, link.implementation.columnName));
+  return new Conjunction([
+    new Condition(new Column(otherEnd, 'id'), 'IN_SUBQUERY', new Column(holder, link.implementation.columnName)),
+    new Condition(new Column(nextTable, 'projectId'), '=', new Column(currentTable, 'projectId')),
```
mattlong: We don't need to consider linked projects here since its not possible for a resource to be in a different project's compartment? Can we add tests to cement the expected behavior with otherwise matching resources in 1. linked projects and 2. unrelated projects?

PR #10024, opened 2026-07-30
feat(core): add defaultSearchParams to MedplumClient + compartment filter UI

maddyli:
## Summary

Adds a general-purpose way to apply default search params to every FHIR search, and a lightweight compartment-filter UI in the app built on top of it. Motivation: while logged in (e.g. as a super admin), being able to scope all queries to one or more compartments — useful for tenant/org-focused browsing.

### Core primitive — `MedplumClient.defaultSearchParams`
- New `defaultSearchParams?: URLSearchParams` option on `MedplumClientOptions`, plus runtime `getDefaultSearchParams()` / `setDefaultSearchParams()` (the setter invalidates the cache and fires the `change` event).
- Merged in the single search chokepoint `fhirSearchUrl`, so it applies to `search`, `searchOne`, `searchResources`, and `searchResourcePages` — but **not** reads-by-id.
- Explicit query params always win over defaults. The merge snapshots the caller's keys before appending, so a default with repeated values (e.g. two `_compartment` entries) is applied in full and the re-entrant `searchOne → search` path doesn't duplicate params.
- This is a **client-side convenience only** — it narrows queries but does not enforce access control. The server still applies the user's access policy, so it can never widen what a user may read.

### App — compartment filter control
- A header popover (via `AppShell`'s existing `notifications` slot, so no `@medplum/react` changes) where any user can pick one or more resources (`ReferenceInput`, any resource type) to scope all searches to those compartments.
- Multiple compartments combine with **OR** semantics via a single comma-joined `_compartment=A,B` value (matches a resource in any selected compartment).
- Selections show as removable pills; **Apply** commits and reloads once so every query surface (including `SearchControl`, which calls `medplum.search` directly) picks it up — mirroring the project switcher's reload-on-change.
- Persisted in `localStorage` (JSON array, with legacy single-string fallback) and seeded into the client at construction.

## Files
- `packages/core/src/client.ts` — `defaultSearchParams` option, getter/setter, merge in `fhirSearchUrl`
- `packages/core/src/client.test.ts` — unit tests (constructor, merge, explicit-wins, runtime set/clear, repeated-value multi-compartment)
- `packages/app/src/compartmentFilter.ts` — localStorage-backed state + params builder
- `packages/app/src/CompartmentFilterControl.tsx` — the header control
- `packages/app/src/App.tsx`, `packages/app/src/index.tsx` — wiring

## Testing
- New core unit test passing.
- Typecheck and eslint clean on all touched files.
- UI tests for the control still to be run.

## Notes / follow-ups
- Built with OR combine semantics. AND (`_compartment=A&_compartment=B`) is a small addition if wanted later — the core merge already handles repeated values correctly.
- No app-level test for the control yet; can add one with a `MockClient` passthrough for `get/setDefaultSearchParams`.

### DIFF
--- a/packages/app/src/App.tsx
+++ b/packages/app/src/App.tsx
@@ -25,6 +25,7 @@ import type { FunctionComponent, JSX } from 'react';
 import { Suspense } from 'react';
 import { Link, useLocation, useSearchParams } from 'react-router';
 import { AppRoutes } from './AppRoutes';
+import { CompartmentFilterControl } from './CompartmentFilterControl';
 
 import './App.css';
 
@@ -47,6 +48,7 @@ export function App(): JSX.Element {
       version={MEDPLUM_VERSION}
       menus={userConfigToMenu(config)}
       displayAddBookmark={!!config?.id}
+      notifications={<CompartmentFilterControl />}
       announcements={
         project?.superAdmin
           ? [
--- a/packages/app/src/CompartmentFilterControl.tsx
+++ b/packages/app/src/CompartmentFilterControl.tsx
@@ -1,0 +1,114 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import { Button, Group, Pill, Popover, Stack, Text } from '@mantine/core';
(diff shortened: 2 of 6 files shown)


techdavidy:
@claude review

PR #10046, opened 2026-07-31
Move scheduling serviceType utils into `@medplum/core`

Filupnot:
Split out of #9987 so the move is tracked on its own, as suggested in review. No behavior change; this is a move plus renames.

## Why

The serviceType helpers that encode an R4 approximation of `CodeableReference<HealthcareService>` lived in `packages/server`, with a hand-maintained copy in the provider example (`examples/medplum-provider/src/utils/servicetype.ts`) because the server package cannot be imported from a client app. They only depend on `@medplum/core` and `@medplum/fhirtypes`, so core is where they belong, and one copy means the two cannot drift.

## What moved

Into the new `packages/core/src/scheduling.ts`, renamed to say what they do rather than how they are shaped:

| Before | After |
| --- | --- |
| `toCodeableReferenceLike` | `toServiceTypeCodeableConcepts` |
| `isCodeableReferenceLikeTo` | `serviceTypeIncludesService` |
| `extractReferencesFromCodeableReferenceLike` | `extractServiceTypeReferences` |

Moved alongside them, for the same reason:

- `SchedulingParametersURI`, `ServiceTypeReferenceURI`, and `TimezoneExtensionURI`, each previously defined (and in two cases duplicated) at its point of use.
- The `DayOfWeek` type and `isDayOfWeek`, previously private to `scheduling-parameters.ts`.
- `hasSchedulingParameters`, previously the provider example's own.

Two test files had also inlined their own `ServiceTypeReferenceURI` constant rather than importing it, one of them with a typo (`medlpum`). Both now import the real one.

`packages/server/src/util/servicetype.ts` and its copy in the provider example are deleted, and all call sites now import from `@medplum/core`.

## Test plan

- [x] `packages/core`: typechecks, builds, `scheduling.test.ts` passes
- [x] `packages/server`: typechecks; `scheduling-parameters.test.ts` passes (the `book`/`hold`/`find` changes are the mechanical rename, and need a database, so they are left to CI)
- [x] `examples/medplum-provider`: typechecks, 113 schedule tests pass
- [x] eslint clean on every changed file


packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @param serviceType - CodeableConcept values to inspect
+ * @param service - HealthcareService or reference to match
+ * @returns True if any concept references the service
+ */
+export function serviceTypeIncludesService(
+  serviceType: CodeableConcept[] | undefined,
+  service: WithId<HealthcareService> | (Reference<HealthcareService> & { reference: string })
+): boolean {
+  const reference = getReferenceString(service);
+  return extractServiceTypeReferences(serviceType).some((serviceReference) => serviceReference.reference === reference);
```
noahsilas: While it's nice and tidy to reuse `extractServiceTypeReferences` this way, I think that we would prefer to try to keep this a tight loop. The old version (`isCodeableReferenceLikeTo`) did this by pushing all the work inside of a filter: `serviceType.some((concept) => { ... })`, which means that if the first concept we look at matches we never need to inspect the second concept.

This saves an array allocation at least, and in scenarios where there are multiple concepts in play it can avoid the `getExtensionValue()` calls (which IIRC is a little more expensive than you might imagine due to the type parsing going on behind the scenes in there).

packages/core/src/scheduling.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const service: HealthcareService = {
+  resourceType: 'HealthcareService',
+  id: 'service-1',
+  name: 'Office visit',
+};
+
+describe('serviceType CodeableConcepts', () => {
+  test('converts, matches, and extracts HealthcareService references', () => {
+    const serviceWithId = { ...service, id: 'service-1' };
```
noahsilas: This looks like it exactly matches `service` above; I suspect this is here to make Typescript happy?

You could explicitly mark the literal above as a `WithId<HealthcareService>`, or it to keep its richer shape:
```ts
const service = { ... } satisfies HealthcareService
```

packages/server/src/util/servicetype.ts
```diff
@@ -1,91 +0,0 @@
...
- *         {
- *           url: "https://medplum.com/fhir/service-type-reference",
- *           valueReference: { reference: "HealthcareService/123" }
- *         }
- *       ]
- *     }
- *  ]
- *
- * This constant holds the extension URI used to hold the reference.
- */
```
noahsilas, line 31: I think there is some valuable context in this comment (verbose as it is) - what do you think about pulling more of this over to the new site of this definition?

packages/core/src/scheduling.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  test('preserves service type coding while adding a reference', () => {
+    const serviceWithType = {
+      ...service,
+      id: 'service-1',
+      type: [{ coding: [{ system: 'http://example.com/service', code: 'office' }] }],
+    };
+    const serviceType = toServiceTypeCodeableConcepts(serviceWithType);
+
+    expect(serviceType[0].coding?.[0].code).toBe('office');
+    expect(serviceTypeIncludesService(serviceType, serviceWithType)).toBe(true);
```
noahsilas, line 32: I think one missing test is that `serviceTypeIncludesService` should accept a `Reference<HealthcareService>` - I think there are some places in server that rely on that behavior and we don't want to allow a regression.

packages/core/src/scheduling.ts
Filupnot: Good call, reverted to the short-circuiting form in 7ec9c0bf3. `getExtensionValue()` now happens inside the `serviceType.some()` callback, so a match on the first concept skips the remaining ones and we avoid the intermediate array.

packages/core/src/scheduling.test.ts
Filupnot: Correct, that was purely to keep TypeScript happy. The shared fixture is now declared with `satisfies WithId<HealthcareService>`, which keeps the richer shape and let me drop all three redundant `{ ...service, id: 'service-1' }` spreads.

packages/server/src/util/servicetype.ts
Filupnot: Agreed, pulled it over onto `ServiceTypeReferenceURI` in `packages/core/src/scheduling.ts`

packages/core/src/scheduling.test.ts
Filupnot: Added a `matches a HealthcareService reference` test that passes `createReference(service)` for the positive case and a bare `{ reference: 'HealthcareService/service-2' }` for the negative one, so the `Reference<HealthcareService>` overload is pinned down against regression.

examples/medplum-provider/src/utils/scheduling.ts
```diff
@@ -1,9 +1,7 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
```
techdavidy, line 2: I think we can move the whole file to core.

Filupnot: Agreed, done in d4a2847b4. Moved the whole file into `packages/core/src/scheduling.ts` and deleted it from the example app.

noahsilas: I might push back on this a little; `SchedulingTransientIdentifier` is pretty tightly scoped. I suspect that we probably will eventually want something more generic that we can use for other scenarios where we get back results that don't have strong unique identifiers.

I'm also not sure I love the mutative API style here; it was chosen more as a matter of convenience during an early implementation run. Moving this into `core` sets us up to maintain this without going through all the design rigor.

In the meantime, this exists purely as a client side concern inside medplum-provider, so I'd advocate to keep it more local (and out of our backwards-compat guarantees in core)

noahsilas: Actually the main scenario we needed this for was related to highlighting proposed slots on the calendar, which we have since cut (there were some confusing interactions that will probably be re-visited later as part of ongoing work, but this might not be the best tool for this job).

Maybe we should remove this tool entirely for now 🤔

Filupnot: Good point, and I agree on the transient identifier.

`SchedulingTransientIdentifier` shouldn't go in core. It never reaches the server, so it's purely a client concern, and moving it there would lock us into supporting its current shape before we've designed the more general version you're describing.

I do still think `SchedulingEncounterCodingURI` and `SchedulingPlanDefinitionURI` belong in core. Those URIs get written onto `HealthcareService`, same as `SchedulingParametersURI`, and a plain string constant has no API shape we'd regret later.

On dropping the transient helper entirely: `FindPane` still uses it for its list key, so that needs a replacement first. Probably its own PR.

Let me know what you think and I'll make the change.

@noahsilas @techdavidy

noahsilas: Sounds like the right step forward, thanks! Agreed that removing that transient identifier is better done separately 👍

packages/core/src/scheduling.ts
```diff
@@ -0,0 +1,108 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
+// SPDX-License-Identifier: Apache-2.0
+import type { CodeableConcept, HealthcareService, Reference, Schedule } from '@medplum/fhirtypes';
+import type { WithId } from './utils';
+import { createReference, getExtension, getExtensionValue, getReferenceString, isDefined } from './utils';
+
+export const SchedulingParametersURI = 'https://medplum.com/fhir/StructureDefinition/SchedulingParameters';
+export const SchedulingEncounterCodingURI = 'https://medplum.com/fhir/StructureDefinition/SchedulingEncounterCoding';
+export const SchedulingPlanDefinitionURI = 'https://medplum.com/fhir/StructureDefinition/SchedulingPlanDefinition';
```
noahsilas, line 9: There's also an argument to keep these in `Provider` today - right now that's the only place these extensions are set and read, allowing the scheduling UI flow to remove some extra clicks when a HealthcareService has these values set.

That said, I don't think it's particularly troublesome to move them in here, and might enable sharing that behavior into our reusable packages, so this seems fine to me. :+1:

noahsilas (approved):
Thanks for handling this centralization, and for giving a few of these functions names that are a little more expressive of their current focus/limitations.

PR #10057, opened 2026-08-01
Rework cicd-deploy.sh to use turborepo dependency graph

noahsilas:
Turborepo offers the `--affected` filter that lets us specify we should only apply a command to packages whose inputs have been touched in the diff since the `TURBO_SCM_BASE` git commit.

This has some benefits:

- The dependency graph defined by the `package.json` files across the workspace is fully respected without being reimplemented here as file checks.

- The `build` step can happen a single time (previously, when multiple apps were being deployed, each one would use a build step with `--force`, requiring rebuilding everything). All affected apps now can build in parallel, and share the common dependency builds.

There's some nuance as well:
- This turborepo feature seems to also detect changes in your working tree. This should not be impactful for this CI deploy script, but can make testing the script surprising if you have any local changes (including untracked files).

- When doing this CI build, we need to trigger the build with `--force` because the `@medplum/core` package has an implicit dependency on the current git hash (which it appends to its `MEDPLUM_VERSION` export; see https://github.com/medplum/medplum/blob/v5.1.27/packages/core/esbuild.mjs#L14).

- The `app` build still is doing some tricky ENV var work that I don't fully understand, so I've left its extra rebuild untouched.

- The deployable packages have had some extra metadata added to their `turbo.json` files to make them depend on some common files from the root (mainly the cicd-deploy script itself). This replaces some extra conditionals that used to be in the script. This requires enabling a "futureFlag" in turborepo to respect the "input" specifications when computing "affected" packages

### DIFF
--- a/packages/server/turbo.json
+++ b/packages/server/turbo.json
@@ -13,6 +13,14 @@
       "outputs": ["coverage/**"],
       "inputs": ["src/**/*.tsx", "src/**/*.ts"],
       "env": ["REDIS_PASSWORD_DISABLED_IN_TESTS", "POSTGRES_HOST", "POSTGRES_PORT", "SERVER_TEST_CACHE_KEY"]
+    },
+    "build": {
+      "inputs": [
+        "$TURBO_DEFAULT$",
+        "../../Dockerfile",
+        "../../.github/workflows/build.yml",
+        "../../scripts/cicd-deploy.sh"
+      ]
     }
   }
 }
--- a/packages/app/turbo.json
+++ b/packages/app/turbo.json
@@ -3,8 +3,7 @@
   "extends": ["//"],
   "tasks": {
     "build": {
-      "inputs": ["$TURBO_DEFAULT$", "../../scripts/cicd-deploy.sh"],
-      "outputs": ["storybook-static/**"]
+      "inputs": ["$TURBO_DEFAULT$", "../../scripts/cicd-deploy.sh"]
     }
   }
 }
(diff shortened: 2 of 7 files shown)


noahsilas:
@claude please review

packages/server/turbo.json
```diff
@@ -13,6 +13,14 @@
       "outputs": ["coverage/**"],
       "inputs": ["src/**/*.tsx", "src/**/*.ts"],
       "env": ["REDIS_PASSWORD_DISABLED_IN_TESTS", "POSTGRES_HOST", "POSTGRES_PORT", "SERVER_TEST_CACHE_KEY"]
+    },
+    "build": {
+      "inputs": [
+        "$TURBO_DEFAULT$",
+        "../../Dockerfile",
```
ThatOneBro, line 20: So we're including Dockerfile changes, as well as the changes to the script to trigger turbo "busting the cache" for this package? Could we just add this to conditions for the respective `DEPLOY_SERVER` conditional block so that turbo stays "pure" here?

Otherwise changing any of these files would also bust build cache locally

Same applies for the new `turbo.json` files added

noahsilas: That's an option, but it gets us back into encoding a second dependency graph in the deploy script.

What about adding a `deploy` task to turborepo that could depend on these extra files + the `build` task?

ThatOneBro: That seems sensible to me

cc @codyebberson @mattlong @mattwiller 

Any thoughts here?

Issue #6590, opened 2025-05-12
Cannot read properties of undefined (reading 'type') from `@medplum/core` websocket message handler


dillonstreator:
I'm seeing the following error from the websocket subscription message handler in `@medplum/core`.

```sh
Cannot read properties of undefined (reading 'type')
```

https://github.com/medplum/medplum/blob/e6e87735659bd06bb4b01e27489e76913f8207cf/packages/core/src/subscriptions/index.ts#L145-L157

ThatOneBro:
Hey @dillonstreator, thanks for reporting this issue!

Do you mind sharing some context about the error?

How are often are you seeing this? Is there any consistent repro you have for it?

Also do you mind checking your "Socket" tab in the Network tab in Chrome Dev Tools, and opening the "subscriptions-r4" connection?

<img width="1245" alt="Image" src="https://github.com/user-attachments/assets/84a59b16-5d67-4a91-8462-9f28770f1e14" />

I'm curious if it might give us any insight into what is causing -- presumably something in the Bundle chain or the Bundle itself -- to be `undefined`

dillonstreator:
Hey @ThatOneBro I unfortunately cannot repro. Sentry is picking this up from the console. While this log has happened for most users, it is infrequent across total sessions.

PR #10062, opened 2026-08-02
Guard against WebSocket subscription messages without a SubscriptionStatus

ajwwong:
## The bug

`SubscriptionManager`'s WebSocket message handler reads the notification bundle's status resource with optional chaining, then immediately dereferences it without a guard:

```ts
const status = bundle?.entry?.[0]?.resource as SubscriptionStatus;

// Handle heartbeat
if (status.type === 'heartbeat') {   // 💥 TypeError when entry[0].resource is absent
```

Any frame that parses as JSON but doesn't carry a `SubscriptionStatus` as `entry[0].resource` throws `Cannot read properties of undefined (reading 'type')`. The `catch` below then dispatches an `error` event to the master emitter **and every criteria emitter**, so an app with N active `useSubscription` hooks surfaces the error N+1 times.

I believe this is the crash reported in #6590 (closed unreproduced — the reporter was seeing it through Sentry with no repro). We hit it in production on our Medplum-based EHR: a bot wrote messages into a thread the user had open, and the chat page (two live subscriptions) popped two `Cannot read properties of undefined (reading 'type')` error toasts. The repro is deterministic once you know the shape: deliver any parseable frame without a status entry (see the updated test).

## The fix

Early-return with a `console.warn`, consistent with how the same handler already treats other anomalous-but-parseable frames ("Received notification for criteria the SubscriptionManager is not listening for"):

```ts
const status = bundle?.entry?.[0]?.resource as SubscriptionStatus | undefined;
if (!status) {
  console.warn('Received WebSocket message without a SubscriptionStatus resource; ignoring');
  return;
}
```

Genuinely unparseable frames are unaffected — `JSON.parse` failures still flow through the `catch` and surface as `error` events.

## Test changes

The existing test `should emit error event when invalid message comes in over WebSocket` was asserting the crash as the intended behavior — it sends `'invalid_json'` (which, via the mock's `jsonProtocol`, arrives as a perfectly parseable JSON string) and expects the resulting `TypeError: Cannot read properties of undefined...` to be emitted on both emitters. That accidental TypeError is exactly the noise from #6590.

Replaced it with `should warn and ignore WebSocket messages without a SubscriptionStatus`, which covers both malformed shapes (a non-Bundle JSON value, and a Bundle with no entries), asserts no `error` events fire and `console.warn` fires once per bad frame — and then delivers a valid heartbeat to prove the manager is still alive and processing afterwards.

### DIFF
--- a/packages/core/src/subscriptions/index.test.ts
+++ b/packages/core/src/subscriptions/index.test.ts
@@ -975,6 +975,16 @@ describe('SubscriptionManager', () => {
 
       expect(receivedBundle).toStrictEqual(heartbeatBundle);
       expect(console.warn).toHaveBeenCalledTimes(2);
+      expect(console.warn).toHaveBeenNthCalledWith(
+        1,
+        'Received WebSocket message without a SubscriptionStatus resource; ignoring',
+        'invalid_json'
+      );
+      expect(console.warn).toHaveBeenNthCalledWith(
+        2,
+        'Received WebSocket message without a SubscriptionStatus resource; ignoring',
+        { resourceType: 'Bundle', type: 'history', entry: [] }
+      );
       expect(errorListener).not.toHaveBeenCalled();
       console.warn = originalWarn;
     });
--- a/packages/core/src/subscriptions/index.test.ts
+++ b/packages/core/src/subscriptions/index.test.ts
@@ -932,44 +932,51 @@ describe('SubscriptionManager', () => {
       expect(receivedBundle).toStrictEqual(sentBundle);
     });
 
-    test('should emit `error` event when invalid message comes in over WebSocket', async () => {
-      const originalError = console.error;
-      console.error = vi.fn();
+    test('should warn and ignore WebSocket messages without a SubscriptionStatus', async () => {
+      const originalWarn = console.warn;
+      console.warn = vi.fn();
 
       await wsServer.connected;
 
       const emitter = defaultManager.addCriteria('Communication');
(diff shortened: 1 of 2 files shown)


ThatOneBro:
Hi @ajwwong , thanks for flagging this issue!

I did some digging and it looks like this might be from this line specifically:

https://github.com/medplum/medplum/blob/derrick-event-category-dispatch/packages/server/src/ws/subscriptions.ts#L373

Which occurs when attempting to bind to a token that is expired

Instead of just logging only a warning about dropping the message, can we also log the parsed payload received in that warning? Just for auditability

ThatOneBro (commented):
Looks good for the most part, just one comment about logging the message received along with the warning

ajwwong:
Done — the warning now includes the parsed payload (`console.warn('Received WebSocket message without a SubscriptionStatus resource; ignoring', bundle)`), and the test asserts the payload is passed through for both malformed shapes.

And thanks for tracing the server-side origin to the expired-token bind path — that matches what we saw in production (the frames appeared around long-lived chat sessions, which fits token expiry). The client-side guard here keeps N subscriptions from surfacing N+1 error toasts either way, but glad to know where the frames come from.

ThatOneBro (approved):

ThatOneBro:
@ajwwong Thanks for updating your PR! Do you mind signing off on your commits so it passes the DCO?

ajwwong:
@ThatOneBro sorry — forgot the DCO sign-off on that last commit (again!). Amended and force-pushed, no code changes.

PR #10069, opened 2026-08-03
Apply project appName white-labeling to invite and password reset emails

mokantari:
The `appName` project setting already white-labels the email sender display name, MFA email content, and the authenticator app issuer, but the invite emails (new user and existing user) and the password reset email still hardcode "Medplum" in their subjects and signatures.

This applies the same `getProjectAppName` lookup to those three templates, falling back to "Medplum" as before, so a white-labeled project brands every email it sends. `DEFAULT_APP_NAME` moves next to `getProjectAppName` in `email/utils` and the duplicate in `auth/utils` is removed.

No behavior change for projects without the setting: existing subject-line tests pass unchanged. New tests cover the branded variants of all three emails (new-user invite, existing-user invite, and password reset for a project-scoped user).

### DIFF
--- a/packages/server/src/admin/invite.test.ts
+++ b/packages/server/src/admin/invite.test.ts
@@ -5,10 +5,10 @@ import type { WithId } from '@medplum/core';
 import { allOk, ContentType, createReference, getReferenceString, normalizeErrorString } from '@medplum/core';
 import type {
   AccessPolicy,
-  Project,
   BundleEntry,
   Patient,
   Practitioner,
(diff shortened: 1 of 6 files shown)


mokantari:
Per the policy bot's guidance: opened medplum/medplum#10338 describing this work, and rebased this branch onto v5.1.33 (it now builds on the getProjectAppName helper that landed in email/utils since). Ready to link and reopen once the issue carries the open-to-community label.

PR #10071, opened 2026-08-03
feat(server): Stream tool calls from  and extract the OpenAI provider

techdavidy:
- Allow  `$ai` to stream a response and/or request tool calls.
- Extract the OpenAI wire format into `src/ai/openai.ts` behind neutral types, so `fhir/operations/ai.ts` is FHIR orchestration only and owns the SSE framing.


techdavidy:
@claude review

techdavidy:
@claude review again

packages/server/src/ai/openai.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+}
+
+/**
+ * Builds the OpenAI request body.
+ * @param context - The request and its credentials
+ * @param stream - Whether to ask for a streamed response
+ * @returns The request body to POST
+ */
+function buildRequestBody(context: AiContext, stream: boolean): any {
```
codyebberson: We try to avoid `any` as much as possible

```suggestion
function buildRequestBody(context: AiContext, stream: boolean): Record<string, unknown> {
```

packages/server/src/ai/openai.ts
```diff
@@ -0,0 +1,243 @@
...
+  const requestBody: any = {
```
codyebberson:
```suggestion
  const requestBody: Record<string, unknown> = {
```

packages/server/src/ai/openai.ts
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/**
+ * Converts accumulated stream deltas into the shared tool call type.
+ * @param acc - Sparse array of accumulated tool calls
+ * @returns The completed tool calls
+ */
+function toAiToolCalls(acc: StreamedToolCall[]): AiToolCall[] {
+  return acc
+    .filter((call) => call)
```
codyebberson:
```suggestion
    .filter(Boolean)
```

packages/server/src/ai/openai.ts
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/**
+ * Calls OpenAI and returns the complete response.
+ * @param context - The request and its credentials
+ * @returns The model's content and any tool calls it requested
+ */
+export async function callOpenAi(context: AiContext): Promise<AiResult> {
+  const response = await postChatCompletions(context, false);
+  await throwIfNotOk(response);
```
codyebberson, line 234: This now calls `await response.json()` twice.  I'd probably inline this function

packages/server/src/ai/types.ts
```diff
@@ -0,0 +1,63 @@
...
+ * What a caller asks the model to do.
+ *
+ * `messages` and `tools` are pass-through JSON in whatever shape the caller supplied, so they
+ * stay untyped — each provider is responsible for translating them to its own wire format.
+ * @param messages - The messages to send to the model
+ * @param model - The model to use for the request
+ * @param tools - The tools to use for the request
+ * @param temperature - The temperature to use for the request  
+ */
+export interface AiRequest {
```
codyebberson: Are these essentially OpenAI types?  Is there a standard NPM package we can use?

packages/server/src/ai/openai.ts
techdavidy: the helper only touches the body when !response.ok, and it throws from there, so the response.json() in callOpenAi can't run in that case. The two calls are mutually exclusive paths.

techdavidy:
@claude review

packages/server/src/ai/types.ts
techdavidy: The return type would be a fhir provider neutral response.
I added a raw property to pass what we get from llm provider.

techdavidy:
@claude review

codyebberson (approved):

PR #10072, opened 2026-08-03
Re-implement Binary reference extraction for performance

mattwiller:
Local benchmarking indicates that the new direct traversal approach is a 120–150x speedup over `JSON.stringify()` + regex


codyebberson:
@claude - please review

Can you please estimate performance change?

Can you try to do a micro benchmark in your review?

The-Alchemist (approved):

mattwiller:
The suggestion to replace `Object.entries()` with a guarded `for .. in` loop over the object keys was worth a further 2x speedup on top, so I added that optimization as well

PR #10073, opened 2026-08-03
Document upgrade path for #9995

noahsilas:
#9995 is being held to allow users of the beta Scheduling API time to update any `Slot` resources that could result in a schedulable resource being double-booked. This PR adds a page documenting the data migration that users of the Scheduling APIs should go through before Sept. 1 in order to avoid creation of such conflicts.


codyebberson:
@claude please review

The-Alchemist (approved):

PR #10075, opened 2026-08-03
feat(server): add OTEL metrics for data warehouse sync

The-Alchemist:
- Emit per-table OpenTelemetry metrics from data warehouse sync: `rows`, `duration`, `watermarkDuration`, and `tables` (with `result` / `skipReason`)
- Attribute `table` uses the Iceberg table name to keep cardinality low
- Cover success and each skip path in unit tests


The-Alchemist:
@claude review.  make sure it matches naming conventions in the rest of the codebase

The-Alchemist:
@claude review

The-Alchemist:
@claude review

The-Alchemist:
@claude review

The-Alchemist:
@claude review

packages/server/src/data-warehouse/sync.ts
```diff
@@ -329,6 +330,52 @@ function buildSkippedTableResult(
...
+    watermarkDurationMs,
+  };
+}
+
+/**
+ * Records sync-level OTEL metrics aggregated across all table outcomes.
+ * @param tables - Per-table sync results (including partial progress when a hard failure aborts the run).
+ */
+function recordSyncMetrics(tables: SyncTableResult[]): void {
+  const rowsInserted = tables.map((t) => t.rowsInserted).reduce((a, b) => a + b, 0);
```
mattlong: Reducing memory allocations is fresh on our mind after being reminded of their impact in several of our recent incidents; it looks like this could all be done in a single `for (const t of tables) { ... }` to avoid all the array allocations from the `.map` and `.filter`

mattlong (changes requested):

packages/server/src/data-warehouse/sync.ts
The-Alchemist: I changed these to use `.values()`, which returns an `Iterator` so it doesn't allocated any arrays.

I also added a few helpers in `core/utils.ts` which will hopefully steer folks into using optimized versions.

codyebberson:
@claude please re-review - are all feedback items addressed?

The-Alchemist:
@claude review

The-Alchemist:
@claude review

codyebberson (approved):

The-Alchemist:
@claude review

mattlong (approved):

PR #10079, opened 2026-08-03
Logical shard routing

mattlong:
## Logical shard routing in `Repository`

`Repository` has always had at most one `RepositoryConnection`, tied to one Postgres database. This
lifts that restriction: a `Repository` now holds one connection **per shard** and routes each
database operation to the right one, based on the resource types the operation declares.

**Nothing changes in production behavior**. Every project resolves to the global shard, so
`resolveShardId` always returns `global`, no repository ever opens a second connection, and the new
paths are exercised only by the dedicated routing tests. This lands the routing layer so that later
increments can make a second database reachable.

### Model

Resource types are either **global** — they must be readable before the current project, and
therefore its shard, is known — or **project-scoped**, living on the shard of the project that owns
them. `globalShardResourceTypes` is currently `{Project, ProjectMembership, User}` and is known to be
incomplete; expanding it is increment 3.

### Invariants enforced

1. A statement whose resource types resolve to more than one shard throws (`Operation cannot span
   shards`) — no single database could serve it. Mixing is allowed while both halves land on the
   global shard, which is why this is not an error today.
2. A repository holding a live transaction is confined to that transaction's shard
   (`Cannot use shard while a transaction is active on a different shard`). The check consults only
   that repository's own binding and reads liveness from the bound connection, so it lifts the moment
   `COMMIT` returns — post-commit callbacks may reach any shard, which is the seam for work that has
   to span shards.
3. A *sequence* of single-shard operations outside a transaction may span shards freely.
4. Every access must name at least one resource type, since that is what routes it. `resolveShardId`
   throws on an empty set rather than defaulting: "touches no resources" identifies no database, and
   picking either the global or the context shard would silently misroute whichever callers meant the
   other. Work with no resource types to name is not repository work and takes a `PoolClient`
   directly.
5. A connection set seeded with a caller-owned client is sealed to that client's shard. We did not
   take that client, so we have no license to acquire siblings we would then have to release.
6. A transaction level Postgres accepted but that was never published as a scope is rolled back and
   released by `abandonTransactionLevel`, so a failure between `BEGIN`/`SAVEPOINT` and the scope being
   returned cannot leave an untracked open transaction.

### Notable decisions

**`RepositoryConnections` deliberately offers no set-wide view of transactions.** Several of its
connections may be in transactions simultaneously, which is legal — separate databases are separate
atomicity domains, and nothing promises atomicity across them. What must never happen is a *single*
transaction spanning shards, and that is a fact about one repository's binding, not about the set.

**`assertUsable` is asymmetric on purpose.** A repository bound to a live transaction validates only
that connection's scope; an unbound one validates every connection, which is what locks a parent
repository out for the duration of its own child transaction.

**Serializing transactions across the set was considered and rejected.** Holding a lock for a
transaction's duration would turn "reach another shard mid-transaction" from an immediate exception
into a deadlock.

**Routing errors are `exception`/500 by design.** Reaching `resolveShardId` with an operation no
single database can serve means our own code composed it. Requests a client can make unsatisfiable —
a transaction bundle mixing a `Patient` and a `Project` — need a 400-level preflight in
`BatchProcessor`, this will happen in a future increment.

### Testing

Routing tests use a deliberate two-shard fixture. Since no second database is configured yet, they
assert *which connection an operation is routed to* and *which combinations are refused* — not that
data is held apart. An actual second database will be added in a future increment

### Not in scope

Future increments will
* make additional shards reachable (`getDatabasePool(mode, shardId)` and shard→database
config) and is the one with deployment consequences.
* expanding the global type set plus a detector for the silent-misrouting case, splitting the cross-shard call sites
(`projectinit`, transaction bundles, reference validation), and per-shard seeding with `Project.link`
co-location.


The-Alchemist:
@claude review.  make sure the current production behavior has not changed by comparing data from in this PR compared to main

codyebberson:
@claude - can you please review with a specific focus on transaction atomicity?  In the past, we have had occasional bugs slip through where our JS code managed some representation of lock state, but those state transitions were not actually atomic due to an `await`.  Please ensure that we have preserved atomic transactions.

The-Alchemist (approved):

codyebberson (approved):

PR #10094, opened 2026-08-04
fix: match synonym rows in ValueSet expansion property filters

### LINKED ISSUE #10093: ValueSet/$expand returns empty expansion when displayLanguage is combined with a property filter
`ValueSet/$expand` returns an empty expansion when the `displayLanguage` parameter is combined with a ValueSet compose property filter (`=`, `in`, or `exists`), even though each works correctly on its own. The `exists=false` case is mirrored: translated rows of concepts that *should* be excluded leak into the expansion.

The same root cause affects `excludeNotForUI`: abstract concepts' translated rows are not excluded when expanding with `displayLanguage`.

### To reproduce

1. Create a CodeSystem with a concept that has a property (e.g. `mptype = DIN`) and a translated designation (e.g. `language: fr`)
```
{
  "resourceType": "CodeSystem",
  "url": "https://example.com/CodeSystem/ccdd",
  "status": "active",
  "content": "complete",
  "property": [
    { "code": "mptype", "type": "string" }
  ],
  "concept": [
    {
      "code": "MP1",
      "display": "gabapentin 100 mg oral capsule",
      "designation": [
        { "language": "fr", "value": "gabapentine 100 mg capsule orale" }
      ],
      "property": [
        { "code": "mptype", "valueString": "DIN" }
      ]
    },
    {
      "code": "TM1",
      "display": "gabapentin",
      "designation": [
        { "language": "fr", "value": "gabapentine" }
      ]
    }
  ]
}
```
2. Create a ValueSet with a compose include filter on that property:   
```json
   {
     "resourceType": "ValueSet",
     "status": "active",
     "url": "https://example.com/ValueSet/mp-din",
     "compose": {
       "include": [
         {
           "system": "https://example.com/CodeSystem/ccdd",
           "filter": [{ "property": "mptype", "op": "=", "value": "DIN" }]
         }
       ]
     }
   }
   ```
3. `GET ValueSet/$expand?url=https://example.com/ValueSet/mp-din&displayLanguage=fr`

**Expected:** matching concepts with their French displays.

**Actual:** empty expansion. (Any `displayLanguage` value triggers it, including `en`.)

### Root cause

Translated displays are stored as separate `Coding` rows (`language = 'fr'`, `synonymOf` → the primary row's id), while `Coding_Property` rows are only linked to the primary `Coding` row. The expansion query applies both of these at once:

- `applyExpansionFilters` adds `WHERE language = 'fr'`, restricting results to the synonym rows
- `addPropertyFilter` adds `EXISTS (SELECT 1 FROM "Coding_Property" cp WHERE cp.coding = "Coding".id …)`, which only ever matches the primary row

For a synonym row the `EXISTS` never matches, so the intersection is always empty. For `exists=false` filters the `NOT EXISTS` is vacuously true for synonym rows, producing the inverse leak. `addAbstractFilter` (`excludeNotForUI`) has the same correlation on its `LEFT JOIN`.

The property filter correlation needs to also match through `synonymOf`, the way `findAncestor` already does for hierarchy traversal.

Note: `is-a`/`descendent-of` filters combined with `displayLanguage` are also broken, but differently — the descendant CTE recursion walks parent-property links that only exist on primary rows, so synonym rows never enter the subtree. That needs a separate fix in the CTE construction and is not covered here.

ianplunkett:
Fixes #10093

`ValueSet/$expand` returned an empty expansion when `displayLanguage` was combined with a compose property filter, because translated displays are stored as synonym `Coding` rows while `Coding_Property` rows only link to the primary row — the property filter's `EXISTS` correlation (`cp.coding = "Coding".id`) could never match a synonym row. The `exists=false` case was mirrored: `NOT EXISTS` was vacuously true for synonym rows, leaking translated rows of concepts that should have been excluded.

### Changes

- **`addPropertyFilter`** (`utils/terminology.ts`): the `Coding_Property` correlation now matches `cp.coding` against the row's `id` **or** its `synonymOf`, using the same `Disjunction` shape `findAncestor` already uses for synonym rows. This also fixes the same latent issue in `$validate-code`, which shares this helper.
- **`addAbstractFilter`** (`expand.ts`): same correlation fix on the `excludeNotForUI` LEFT JOIN, so abstract concepts' translated rows are excluded from UI expansions.
- **`expand.test.ts`**: new `Property filter with displayLanguage` describe block with regression tests for the `=` filter (translated display returned) and the `exists=false` leak. Both fail without the fix.

### Out of scope

`is-a`/`descendent-of` filters combined with `displayLanguage` are also broken, but the cause is different (the descendant CTE recursion never includes synonym rows) — noted in #10093 as a follow-up.


codyebberson:
@claude please review.  what is the expected performance impact of the extra disjunction?

mattwiller:
Closing in favor of #10211

PR #10096, opened 2026-08-04
feat(fhircast): track each subscriber's events and confirm them

### LINKED ISSUE #5981: [FHIRcast] Only propagate event notifications to a user for events they are subscribed to

### LINKED ISSUE #6788: FHIRCast: Subscription Confirmation message is missing list of subscribed events.
Subscription Confirmation on websocket connection does not match the specification.

See Section 2.4.1.3 Subscription Confirmation at https://build.fhir.org/ig/HL7/fhircast-docs/2-4-Subscribing.html 

The following object is being sent 

{
    "hub.callback": "",
    "hub.channel": "",
    "hub.events": "",
    "hub.lease_seconds": 7200,
    "hub.mode": "subscribe",
    "hub.secret": "",
    "hub.subscriber": "",
    "hub.topic": "fdb2f928-5546-4f52-87a0-0648e9ded065"
}

The expected object is 
{
  "hub.mode": "subscribe",
  "hub.topic": "fdb2f928-5546-4f52-87a0-0648e9ded065",
  "hub.events": "Patient-open,Patient-close",
  "hub.lease_seconds": 7200
}

### LINKED ISSUE #10028: FHIRcast: Spec-compliant subscription confirmation
## Issue

**Issue:** [#6788](https://github.com/medplum/medplum/issues/6788) - FHIRCast: Subscription Confirmation message is missing list of subscribed events

## Summary

The WebSocket subscription confirmation currently returns placeholder/empty fields that are not part of the FHIRcast STU3 spec.

## Current State

```typescript
// In packages/server/src/ws/fhircast.ts
socket.send(
  JSON.stringify({
    'hub.callback': '',
    'hub.channel': '',
    'hub.events': '',
    'hub.lease_seconds': 3600,
    'hub.mode': 'subscribe',
    'hub.secret': '',
    'hub.subscriber': '',
    'hub.topic': topic,
  })
);
```

## Required

Per the [FHIRcast spec](https://build.fhir.org/ig/HL7/fhircast-docs/2-4-Subscribing.html#subscription-confirmation), the confirmation should ONLY contain:

```typescript
{
  'hub.mode': 'subscribe',
  'hub.topic': topic,
  'hub.events': 'Patient-open,ImagingStudy-open',  // Subscriber's granted events
  'hub.lease_seconds': 3600
}
```

## Tasks

- [ ] Remove non-spec fields from confirmation (hub.callback, hub.channel, hub.secret, hub.subscriber)
- [ ] Store subscriber's hub.events in Redis when subscribing
- [ ] Retrieve and include hub.events in WebSocket confirmation

## References

- [FHIRcast STU3 Spec - Subscription Confirmation](https://build.fhir.org/ig/HL7/fhircast-docs/2-4-Subscribing.html#subscription-confirmation)
- [Parent Issue #10027](/issues/10027)

### LINKED ISSUE #10030: FHIRcast: Per-subscriber event filtering
## Issue

**Issue:** [#5981](https://github.com/medplum/medplum/issues/5981) - [FHIRcast] Only propagate event notifications to a user for events they are subscribed to

## Summary

Currently, all subscribers receive ALL events broadcast to a topic, regardless of their subscription. Subscribers should only receive events they explicitly subscribed to via `hub.events`.

## Required Behavior

Implement event filtering at the WebSocket layer:
- Store per-subscriber `hub.events` list in Redis
- Filter event dispatch based on subscriber's event subscription
- Case-insensitive matching per spec

## Tasks

- [ ] Store per-subscriber events list in Redis subscriber record
- [ ] Add `subscriberWantsEvent(subscriberEvents, eventName)` function
- [ ] Filter event dispatch at WebSocket layer

## Implementation Notes

```typescript
export function subscriberWantsEvent(subscriberEvents: string, eventName: string): boolean {
  const normalizedEvent = eventName.toLowerCase();
  const subscribedEvents = subscriberEvents.split(', ').map(e => e.trim().toLowerCase());
  return subscribedEvents.includes(normalizedEvent);
}
```

## References

- [FHIRcast STU3 Spec](https://build.fhir.org/ig/HL7/fhircast-docs/)
- [Parent Issue #10027](/issues/10027)
- [PR #9353](https://github.com/medplum/medplum/pull/9353)

ThatOneBro:
## Summary

Previously the Hub kept no record of what each subscriber asked for. `hub.events` came back empty in the WebSocket connection confirmation, and every socket on a topic received every event published to it.

Each `hub.mode=subscribe` request now mints its own endpoint and stores `{projectId, topic, events, version}` under it in Redis, leased for the same hour as the access token. The socket reads that record on connect and holds the events in memory (`websocketMap`), which makes three things possible:

- **Spec-shaped confirmation.** STU3 subscribers get exactly `hub.mode`, `hub.topic`, `hub.events`, `hub.lease_seconds`. STU2 subscribers keep `hub.callback`, `hub.channel`, `hub.secret`, and `hub.subscriber`, which STU3 dropped but STU2 clients still expect. The version comes from the mount path the subscription arrived on; `/api/hub` is STU3.
- **Per-subscriber delivery.** A socket receives only the events it subscribed to, matched case-insensitively the way the Hub already lowercases a published event name to route it. `heartbeat` and `syncerror` reach every subscriber regardless.
- **Scoped unsubscribe.** A `denied` message names its target subscriber and is dropped by every other socket, so one app unsubscribing no longer disconnects the rest of the topic. The marker is stripped before forwarding, so it never reaches the wire. A request that identifies no endpoint still falls back to denying the whole topic.

Everything still flows through one Redis channel per topic; filtering happens per socket.

Fixes #10028
Fixes #6788
Fixes #10030
Fixes #5981

## Notable behavior changes

- Subscribing twice to a topic used to return the same endpoint URL; each subscription now gets its own. Endpoints issued by an older server will not resolve after deploy — those subscribers get `denied — invalid endpoint` and must re-subscribe.
- Filtering means a subscriber that omits an event from `hub.events` stops receiving it.
- `hub.events` is parsed as the comma-separated list the spec calls for, tolerating whitespace around names. Space-separated lists are not split.

Left for a follow-up: requested event names are not validated against `eventsSupported`, so a subscriber asking for an unsupported or misspelled name silently receives nothing for it rather than a 400.

## Test Plan

77 tests pass across `packages/server/src/fhircast/{routes,utils}.test.ts` and `packages/server/src/ws/fhircast.test.ts`; `tsc --noEmit`, eslint, and prettier are clean. New coverage:

- distinct endpoints per subscribe, with the stored record asserted
- STU3 four-field confirmation, STU2 legacy shape, `/api/hub` as STU3, and whitespace normalized out of the echoed list (all deep-equal, so a field creeping back fails)
- an unsubscribed event dropped while a subscribed one and `syncerror` are delivered
- case-insensitive matching in both directions
- unsubscribe denying only its own socket while another subscriber on the topic keeps receiving events, plus the endpointless fallback and a cross-project unsubscribe attempt

The targeting and case-insensitivity tests were mutation-checked: removing the marker or the `toLowerCase()` makes them fail.


mattlong:
@claude review

packages/server/src/fhircast/routes.ts
```diff
@@ -211,68 +229,73 @@ async function handleSubscriptionRequest(req: Request, res: Response): Promise<v
...
+
+  if (endpoint && ownsSubscription) {
+    await deleteEndpointSubscription(endpoint);
   }
+
+  // The spec defines no body for an unsubscribe response, so echo the endpoint back only when the
+  // request identified one
+  res
+    .status(202)
+    .json(endpoint ? { 'hub.channel.endpoint': getWebSocketUrl(getConfig().baseUrl, `/ws/fhircast/${endpoint}`) } : {});
```
mattlong: If the spec doesn't define a body for the response, why do we conditionally send one?

ThatOneBro:
Addressed 1, 2, 3 and 5 in fd6fdb961d.

**1** — dropped the whole-topic fallback. An unsubscribe naming no endpoint is now a 400, and one naming an unknown, expired, or another project's endpoint is an idempotent 202 no-op that publishes nothing. One note on the blast radius as it stood: it could not cross tenants, since the denial is published to `${ctx.project.id}:${topic}` — a foreign-project token only reached its own project's topic. Within a project it was real though, which is the case this PR exists to support. The cross-project test never opened a socket, so it now asserts the connected subscriber stays put.

**2** — 400 on a `hub.events` that parses to nothing.

**3** — `extractEndpoint` strips a query string or fragment, and `handleFhircastConnection` now uses it instead of hand-rolling the same parse (which also drops an `as string` cast on `request.url`).

**5** — `getEndpointSubscription` validates the parsed value and catches a parse failure; either reads as no subscription, which callers already deny.

**4** — leaving the STU2 placeholders as a known gap.

One more found while checking 1: the client discarded Hub control messages by a truthy `hub.topic`, but the invalid-endpoint denial in `ws/fhircast.ts` sends `'hub.topic': ''`, so it slipped past and threw on the missing `event`. Now keyed off the absence of `event`, which covers both shapes.

ThatOneBro:
@claude Can you review the PR again?

ThatOneBro:
Fixed 1 and 2 in 86373b53d9.

**1** — a denial now tears the socket down from both ends. The Hub closes the socket it denies, and the client closes on a denial instead of discarding it, so it holds whichever end acts first. Server-side the close is deferred to the `send` callback, which runs once the frame has been written to the socket, and it's `close()` rather than `terminate()` — so the subscriber still gets the reason and the closing handshake completes.

The two unsubscribe tests now assert `expectClosed()` *without* the client closing first, and the multi-subscriber test publishes an event after the denial and asserts the cancelled socket heard nothing while the surviving one got it. I checked both are load-bearing by reverting the close: they fail.

**2** — `Object.values(FhircastVersion).includes(...)`.

**3** — STU2 placeholders still deferred.

Verified: 89 server tests across `fhircast/routes`, `fhircast/utils` and `ws/fhircast`, 30 in core `fhircast`, 12 in `client-fhircast`; tsc, eslint and prettier clean in both packages.

ThatOneBro:
@claude Can you do another PR review, confirming that everything is in line with the FHIRcast STU3 spec?
See: https://build.fhir.org/ig/HL7/fhircast-docs/2-4-Subscribing.html

codyebberson (approved):

ThatOneBro:
@claude Can you do 1 final review? I ignored 2-5 for now, but want a final comprehensive review ignoring those issues

mattlong (approved):

PR #10099, opened 2026-08-04
Optimize object key enumeration

mattwiller:
`Object.entries()` et al. allocate arrays for all key/value pairs, which is often unnecessary for the task at hand.  Prefer direct iteration over the object keys (with `Object.hasOwn()` guard) for a ~2x speedup and less GC pressure, especially in hot code paths


codyebberson:
@claude please review.  please confirm this is pure performance refactoring and no change in behavior to the callers

packages/core/src/typeschema/validation.ts
```diff
@@ -752,9 +757,14 @@ function unpackPrimitiveElement(v: TypedValue): [TypedValue | undefined, TypedVa
   if (primitiveValue === v.value) {
     return [undefined, { type: 'Element', value: v.value }];
   }
-  const primitiveKeys = new Set(Object.keys(primitiveValue));
-  const extensionEntries = Object.entries(v.value).filter(([k, _]) => !primitiveKeys.has(k));
-  const extensionElement = extensionEntries.length > 0 ? Object.fromEntries(extensionEntries) : undefined;
+
+  const extensionElement = Object.create(null);
```
mattlong, line 762: Previously `extensionElement` could be `undefined` whereas now it'll be an empty object. Is that okay?

mattlong (approved):
Heh, echoing claude review, but one spotted change in behavior noted

mattwiller:
@claude Re-review the refactor, now that additional defensive checks around `Object.prototype.valueOf()` and primitive extensions have been applied

PR #10076, opened 2026-08-03
Create `@medplum/react-scheduling` package

noahsilas:
The Scheduling components come with some relatively heavyweight dependencies:

- @fullcalendar/react
- temporal-polyfill

We could introduce these in @medplum/react as optional peer dependencies, but in the interest of trying to make consuming the scheduling components easier, we are trying a different strategy: creating a new separate package that includes these dependencies.

To get started, we move the `Calendar` component from the Provider app into this new package, and add a storybook entry.

PR #10100, opened 2026-08-04
React-scheduling cleanups

noahsilas:
Some fast follows after introducing @medplum/react-scheduling:
- Update Calendar storybook to be more representative
- Update preview url generation to trigger on react-scheduling changes
- Fix CICD-deploy script; restore triggering on `@medplum/react-hooks` (regression caused by changes in #10076)
- Tighten Calendar double-click handling

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,6 +122,7 @@ jobs:
           if changed '^packages/app/'; then deploy_app=true; fi
           if changed '^packages/docs/'; then deploy_docs=true; fi
           if changed '^examples/medplum-provider/'; then deploy_provider=true; fi
+          if changed '^examples/react-scheduling/'; then deploy_provider=true; fi
 
           deploy_any=false
           if [ "$deploy_app" = true ] || [ "$deploy_docs" = true ] || [ "$deploy_provider" = true ]; then
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
           if changed '^packages/app/'; then deploy_app=true; fi
           if changed '^packages/docs/'; then deploy_docs=true; fi
           if changed '^examples/medplum-provider/'; then deploy_provider=true; fi
-          if changed '^examples/react-scheduling/'; then deploy_provider=true; fi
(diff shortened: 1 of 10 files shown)


Filupnot:
Small thing on the `build.yml` change:

```sh
if changed '^examples/react-scheduling/'; then deploy_provider=true; fi
```

There is no `examples/react-scheduling` — the package lives at `packages/react-scheduling`, so this glob can never match and the line is a no-op. It also sets `deploy_provider`, which looks like the wrong target for a component-library change.

Worth noting `scripts/cicd-deploy.sh` on `main` already covers the storybook side:

```sh
if [[ "$FILES_CHANGED" =~ packages/react-scheduling ]]; then
  DEPLOY_STORYBOOK=true
fi
```

so if the intent was previews for `react-scheduling` changes, that may already be handled and this line can just be dropped.

Unrelated heads-up: #9987 moves `ScheduleAvailabilityEditor` into `packages/react-scheduling/src/ScheduleAvailabilityEditor/`. No overlap with your files (yours are all under `src/Calendar/`), so the two should rebase past each other cleanly in either order.

noahsilas:
> Small thing on the `build.yml` change:
> 
> ```shell
> if changed '^examples/react-scheduling/'; then deploy_provider=true; fi
> ```
> 
> There is no `examples/react-scheduling` — the package lives at `packages/react-scheduling`, so this glob can never match and the line is a no-op. It also sets `deploy_provider`, which looks like the wrong target for a component-library change.

Oops, good catch that this should be `packages/` instead of `examples/`. Fixed in [0e0883e](https://github.com/medplum/medplum/pull/10100/commits/0e0883ef2a9b18308f07df827cca1be4e0aef75c). The slightly frustrating aspect in here is that what we are doing is effectively re-encoding the dependency graph - because the Provider application depends on the `@medplum/react-scheduling` package, if we change that dependency the intent is to also build a preview of the consuming Provider application.

I'm exploring ways to reuse our existing dependency graph (stored in package.json and accessible via `turbo` commands) in this draft: https://github.com/medplum/medplum/pull/10057.

techdavidy (approved):

PR #10104, opened 2026-08-05
Order meds by GCN and ROUTED MED ID (without RxNorm or NDC number present)

oleg-mp:
This enables rare use cases when some niche medications may lack RxNorm or resolving of an NDC number might not be available via eRx vendor API

### DIFF
--- a/examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
+++ b/examples/medplum-provider/src/pages/meds/OrderMedicationPage.tsx
@@ -120,6 +120,40 @@ function getRoutedMedIdFromMedication(m: Medication): number | undefined {
   return Number.isFinite(n) ? n : undefined;
 }
 
+/**
+ * The single GCN_SEQNO on a Medication, or undefined when it carries none — or
+ * **several**.
+ *
+ * A dose-level formulation has exactly one GCN. A drug-name search hit for a
+ * multi-strength product carries one identifier per available GCN, and nothing
+ * on the resource says which strength the prescriber meant. Since GCN is what
+ * the vendor resolves the dose from when there is no NDC, picking the first of
+ * several would silently prescribe an arbitrary strength — so an ambiguous
+ * Medication yields nothing and the caller falls back to requiring a
+ * formulation.
+ *
+ * @param m - Medication from drug search.
+ * @returns The GCN_SEQNO as a number, or undefined when absent or ambiguous.
+ */
+function getUnambiguousGcnSeqnoFromMedication(m: Medication): number | undefined {
+  const values = new Set<string>();
+  for (const id of m.identifier ?? []) {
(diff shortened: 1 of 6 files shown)


oleg-mp:
@claude review this

oleg-mp:
Thanks — all three plus the nits addressed in b19f853d0. Two landed differently than suggested, with reasons.

**1. `line1` contract vs payload — fixed by relaxing the contract, not by deriving a value.** You framed it as either/or and I took the docs side, because deriving it would make the payload worse: the only text either client holds is the full product label, so passing it as `line1` duplicates the name in the description the vendor renders. I hit this concretely on staging — forcing `line1` next to a resolved name produced `"Tolcylen 1 % topical solution solution"`. `line1` is now documented as optional, for callers that hold dose text *separately* from `drugName` (a hand-entered form, say).

Worth noting the mismatch was broader than the compound path: the MedicationRequest-driven path doesn't set `line1` either, so nothing we ship populates it. The docs now say that outright.

**2. Nameless GCN-keyed line — fixed, and the blast radius is narrower than it reads.** `medicationToOrderDrugInput` has exactly one caller (`submitCompound`), so this was compound orders only; the single-med path goes through `medicationRequestId` and the bot derives the line from the MR. And the vendor never actually saw a nameless line — `buildGcnOnlyDrugInfo` throws `drugName is required to order routedMedId=X without an NDC`. So it failed loudly server-side rather than silently sending one.

Still worth fixing, and rather than requiring `drugName` I gave it the same fallback chain the MR path already uses (`code.text` → routed-med coding `display` → any coding `display`) so we use the data we have. A formulation key is only emitted when a name resolves; otherwise the line falls back to needing a resolvable formulation.

**3. No tests — fixed, and moved rather than exported.** The helpers are now `src/utils/medication-gcn.ts` with a sibling `.test.ts`, matching the `utils/*.ts` + `*.test.ts` convention (`coverage.ts`, `claims.ts`, `chargeitems.ts`). Exporting them from an 1858-line page component just to reach them from a test felt like the wrong direction. 18 cases: zero/one/many keys, dedup across `identifier` **and** `code.coding`, zero-padding, non-numeric, plus the name fallback chain.

**Nits, all taken:**
- Dedup key mismatch was real and worse than "unlikely" — deduping strings meant `"8346"` + `"08346"` survived as two numbers and flipped one logical strength to *ambiguous*, which silently blocks ordering. Now dedups on the parsed number, with a test.
- `parseInt` looseness: values are validated as all-digits, so `"12abc"` is dropped rather than read as `12`.
- Definition order resolved by the extraction (callee first).

The same `parseInt` looseness exists in `parseFdbKey` on the medplum-ee side; leaving that for its own PR rather than widening this one.

Verification: 55/55 `medication-order-utils`, 1244/1244 provider-app, `tsc` clean, no new lint errors.

AndyStoneman (approved):

mattlong (approved):

PR #9963, opened 2026-07-25
feat(react): add shared modal layout components and unified modal theme

kevinwadeshaw:
This PR creates a standardized (and reusable) modal layout for use across all Medplum products (Admin and Provider apps). It adds `ModalContentLayout` and `ModalActionsFooter` to @medplum/react to create a consistent header, scrollable body section, and footer with divider and full-width buttons. This removes the need for per-modal styles and hacks at each call site.

NOTE: Please see #9964 for implementation across all existing modals. 

---

### **Reference Image**
Screenshot 2026-08-04 at 12 23 15 PM

PR #10106, opened 2026-08-05
feat(react): Include medplum modal

techdavidy:
Based from https://github.com/medplum/medplum/pull/9963 

Include new react component to keep modal experience consistent. 
Implement modal in ThreadInbox.

Screenshot 2026-08-05 at 11 16 56 AM


techdavidy:
@claude please review

packages/react/src/Modal/Modal.module.css
```diff
@@ -0,0 +1,47 @@
+
+.content.content {
```
noahsilas: Is this a tricky specificity hack? If it is, what would you think about adding a root class like `.medplumModal` so that this could be written as `.medplumModal .content`? It has the same specificity, but is much clearer to later readers in its intention.

I suppose the danger is that someone might look at it and say "You don't need that outer scope in a CSS module" and remove it without thinking about how that lowers the specificity. If you leave it as-is, maybe a quick comment so that this doesn't read as an unintentional duplicate later?

packages/react/src/chat/ThreadInbox/NewTopicDialog.tsx
```diff
@@ -70,26 +70,28 @@ export const NewTopicDialog = (props: NewTopicDialogProps): JSX.Element => {
...
+      onClose={onClose}
+      title="New Message"
+      size="md"
+      actions={
+        <Button onClick={handleSubmit} disabled={!patient || practitioners.length === 0}>
+          Next
+        </Button>
+      }
+    >
+      <Stack gap="lg">
```
noahsilas: This stack looks like it only contains one element and shouldn't do anything, but I think that what is happening here is that `<ThreadMessageForm>` renders from a fragment at its top layer and so this ends up putting spacing between elements inside that component?

Can we move this stack into the ThreadMessageForm so that component has more full control of its own styles instead of using this sort of tricky outer wrapper?

packages/react/src/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+    title="Delete Task"
+    size="md"
+    actions={
+      <Group justify="flex-end">
+        <Button variant="outline">Cancel</Button>
+        <Button color="red">Delete</Button>
+      </Group>
+    }
+  >
+    <Text>A single Group as the actions gives a right-aligned row instead of stretched buttons.</Text>
```
noahsilas: This makes it sound like `<Group>` by itself triggers this alignment, but it's really the `justify="flex-end"` within the group that makes it layout that way.

noahsilas (approved):
We talked in person about naming the component "Modal" instead of "MedplumModal" - I really like that idea; if someone is importing `Modal` from "@medplum/react", it makes sense to me that they should get our modal component and not need to spell out "MedplumModal" again.

PR #10107, opened 2026-08-05
fix(test): remove/fix vitest ESM warnings

The-Alchemist:
```

(!) Your Vite config uses features that are unsupported by `configLoader: 'native'`, which is planned to become the default in a future major version of Vite:
  - import "../../vitest.config" without a file extension (vite.config.ts:8:32). Add the file extension
  - ESM syntax in a file loaded as CommonJS (../../vitest.config.ts:3:1). Use a `.mjs` extension or set `"type": "module"` in the closest package.json
Set `VITE_CONFIG_NATIVE_IGNORE_WARNING=true` to suppress this warning.
```

### DIFF
--- a/packages/app/tsconfig.json
+++ b/packages/app/tsconfig.json
@@ -10,5 +10,5 @@
       "@medplum/react": ["../react/src"]
     }
   },
-  "include": ["src/**/*.ts", "src/**/*.tsx", "vite.config.ts", "vitest.config.ts"]
+  "include": ["src/**/*.ts", "src/**/*.tsx"]
 }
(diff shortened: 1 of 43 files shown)


The-Alchemist:
@cursoragent squash commits and rebase against main

packages/app/tsconfig.json
```diff
@@ -10,5 +10,5 @@
       "@medplum/react": ["../react/src"]
     }
   },
-  "include": ["src/**/*.ts", "src/**/*.tsx", "vite.config.ts", "vitest.config.ts"]
+  "include": ["src/**/*.ts", "src/**/*.tsx"]
```
mattlong, line 13: Curious why this change was needed. I undid it locally and am not seeing any warnings/errors on `npm run {dev,preview,build}` but I could be missing something.

mattlong (approved):

packages/app/tsconfig.json
The-Alchemist: very minor: i didn't think it was necessary for tsc to type check the `vite*.config.ts` files.  seems overkill.

PR #10109, opened 2026-08-05
refactor(agent): rework byte-stream sequence mapping interface

ThatOneBro:
## Summary

Adds link-level byte handshaking to the agent's byte-stream channel — watch the incoming stream for a byte sequence and write a reply straight back to the socket, independent of `startChar`…`endChar` framing — and reworks the interface it was originally built on.

Byte sequence mappings had been configured through a nested `Endpoint.extension` with an inline magic URL and bare relative child URLs (`pattern`, `response`, `strip`) backed by no StructureDefinition, diverging from every other agent channel option. They now live on the endpoint address as query params, alongside `startChar`/`endChar` and the HL7 channel's `encoding` / `enhanced` / `appLevelAck` / `retryMode`:

```
tcp://0.0.0.0:9001?startChar=%02&endChar=%03
  &autoRespond=%05:%06        # link-level handshake
  &stripSequence=%15          # removed from the message body
  &stripControlChars=true     # remaining C0 control bytes
  &keepControlChars=%0D%0A    # bytes exempt from that sweep
  &bodyEncoding=hex           # hex (default) | utf-8
```

`autoRespond`, `stripSequence`, and `keepControlChars` may be repeated, comma-separated, or both. Byte sequences use the same `%XX` form as `startChar`/`endChar` — and it is the same mechanism, not a lookalike: `URLSearchParams` already percent-decodes, so values are read by code point exactly as `startChar` is, replacing a hand-rolled decoder that accepted `%0x` as byte `0`.

The one bundled feature is also split into independently-requested behaviors. Previously `strip` defaulted to `true`, so adding an ENQ→ACK handshake silently rewrote message bodies, and control chars `0x00`–`0x1F` were dropped on *every* byte-stream channel whether or not mappings were configured. Now nothing is filtered unless asked for, and responding to a byte is unrelated to whether that byte survives into the body.

## Fixes carried by the restructure

Splitting immutable config (`ByteStreamConfig`, on the channel) from per-stream state (`ByteSequenceMatcher`, one per connection) resolves several bugs by construction:

- **Config edits now take effect.** Mappings were parsed only from `configureTcpServerAndConnections`, which `reloadConfig` calls only on an address change — so an extension-only edit never reached a running channel. Every option now lives in the address string, so the existing comparison covers it.
- **Connections match independently.** A single channel-wide matcher held the sliding window for every socket, so concurrent connections interleaved into it: one socket's bytes could complete another's pattern, and either socket's `reset()` discarded the other's progress.
- **Multi-byte sequences are stripped whole.** The byte-at-a-time filter dropped only the byte that *completed* a pattern, leaving its leading bytes behind — pattern `AB` kept `A`.
- **One matching pass, not two.** Response matching and strip matching ran separate windows over the same bytes with the same patterns.
- **Bodies keep their bytes.** `hex` remains the default inbound and in `sendToRemote`. The interim `'ascii'` unsets the high bit of every byte on decode, corrupting anything `>= 0x80` in both directions — data loss, not just a breaking change for bots doing `Buffer.from(body, 'hex')`.
- **Two near-duplicate match paths collapse into one.** `getStripValue(byte, checkBuffer)` and its tri-state `boolean | undefined` return are gone.

Parsing and filtering are now exported pure functions, and per-byte debug logging is replaced by one summary per message.

### `keepControlChars`

`stripControlChars` alone is all-or-nothing, which no record-oriented protocol can use: an ASTM body stripped of `CR` is one run-on line the receiver can no longer split into records. `keepControlChars` is the byte set exempt from the sweep — what the old extension expressed as `strip: false` on a pattern. Grouping is meaningless for a set, so `%0D%0A` and `%0D,%0A` are equivalent. Setting it without `stripControlChars` warns, since it can then have no effect.

## Test plan

65 tests in `packages/agent/src/bytestream.test.ts`, up from 6:

- **Table-driven unit tests** for `parseAutoRespondRules`, `parseByteSequences`, `parseKeepControlChars`, `parseBodyEncoding`, `ByteSequenceMatcher`, and `filterMessageBytes` — no sockets. The previous suite drove a full `App` + mock WebSocket + real TCP socket on hardcoded ports 58010–58015, and the `strip` behavior had no coverage at all.
- **Integration tests** over real TCP sockets for the wire behavior: handshake reply, multi-byte pattern split across chunks, comma-separated rules, body filtering, and both encodings.

### ASTM E1394 end-to-end

`ASTM E1394 session: every ACK is returned and only framing is stripped` covers a real BioRad ASTM channel, translated to address params:

```
startChar=%05&endChar=%04
  &autoRespond=%05:%06,%03:%06,%04:%06,%0A:%06
  &stripControlChars=true&keepControlChars=%0D%0A&bodyEncoding=utf-8
```

It drives `ENQ`, an `ETB`-terminated intermediate frame, an `ETX`-terminated final frame, and `EOT` as separate writes — the way an analyzer that waits for each ACK does — then asserts that all five ACKs come back and nothing else, and that the transmitted body has `ENQ`/`STX`/`ETB`/`ETX`/`EOT` removed while `CR` and `LF` survive.

### Regression tests verified to fail without the fix

Rather than only asserting the fixed behavior:

- reintroducing a channel-wide matcher makes `Concurrent connections match independently` fail, with connection B receiving A's ACK
- running the previous `filterControlCharacters` on `"xABy"` with pattern `AB` returns `"xAy"`
- dropping `keepControlChars` from the ASTM test collapses both frames into one run-on line

`tsc --noEmit`, eslint, and prettier are clean; `app.test.ts`, `channel.test.ts`, and `hl7.test.ts` pass.


codyebberson:
@claude please review.  please make notes on any changes to backwards compatibility

codyebberson (approved):

PR #10112, opened 2026-08-06
fix(server): delete stored objects when Binary resources are expunged

### LINKED ISSUE #8132: Feature: Delete S3 objects when Binary resources are expunged
## Problem

When `$expunge` is called on Binary resources, the FHIR records are deleted from the database but the corresponding S3 objects remain orphaned. Over time, this leads to significant storage accumulation.

We discovered this while self-hosting Medplum and found ~9GB of orphaned audio files in S3 after deleting the Binary FHIR records.

### Current Behavior

1. `deleteResource('Binary', id)` - Soft deletes FHIR record, S3 object remains ✓ (correct)
2. `$expunge` on Binary - Deletes FHIR record permanently, **S3 object remains** ✗

### Expected Behavior

When a Binary resource is expunged (permanent delete), the corresponding S3 object should also be deleted.

## Proposed Solution

Add `deleteFile()` method to the `BinaryStorage` interface and call it from `expungeResources()` when expunging Binary resources.

### Implementation Sketch

```typescript
// 1. Add to storage interface (packages/server/src/storage/types.ts)
deleteFile(key: string): Promise<void>;

// 2. Implement in S3Storage (packages/server/src/cloud/aws/storage.ts)
async deleteFile(key: string): Promise<void> {
  await this.client.send(new DeleteObjectCommand({
    Bucket: this.bucket,
    Key: key
  }));
}

// 3. Call from repo.ts or expunge.ts when expunging Binary resources
```

## Questions

1. Should this be configurable (opt-in via config flag)?
2. Any concerns about versioned binaries - should all versions be deleted?
3. Should this also apply to other storage backends (GCS, Azure, local filesystem)?

## Context

- Confirmed as a gap in Medplum Discord (July 2024): "This is actually a gap in our current implementation - we currently don't have an API-based way to delete objects from S3"
- Reference: `packages/server/src/storage/base.ts` - `S3Storage` class has `writeFile`, `readFile`, `copyFile`, but no `deleteFile`
- The S3 key format is `binary/{binaryId}/{versionId}` (from `getKey()` in `base.ts`)

## Workaround

For now, we've implemented direct S3 deletion in our bot code after transcription, and created a cleanup script for historical orphans. Happy to share details if helpful.

ajwwong:
Closes #8132

## The problem

`$expunge` permanently deletes `Binary` rows from the database, but the corresponding objects are left in binary storage. The bytes outlive the resource, so the data is unreachable through the API yet still present in the bucket. I originally hit this as ~9GB of orphaned audio.

Soft delete intentionally keeps the object (the resource can still be read at a prior version), so this only concerns `$expunge`, which the docs already describe as permanently removing data.

## The fix

`expungeResources` now deletes one storage object per Binary version:

- **Every version, not just the current one.** Keys are `binary/<id>/<versionId>` and each write stores its own object, so an N-version Binary has up to N objects. The versionIds come from `RETURNING` on the `<Type>_History` delete that already runs, which avoids a separate read — a read would target the reader pool and conflict with the writer client pinned by the transaction.
- **After the transaction commits**, via `postCommit`, alongside the existing cache eviction. Removing stored objects is irreversible and the expunge transaction is `serializable` and may be retried or rolled back, so deleting inside it could destroy bytes for rows that still exist. This mirrors the write side, which already defers `handleBinaryUpdate` to `postCommit`.
- **Best effort.** A storage failure must not fail the expunge, since the rows are already gone by then; failures are logged with the key so the object can be reclaimed. This matches `BinaryCheckpointStore.cleanup`. Deletes run sequentially; a project-scale expunge of many Binaries could queue thousands of keys, so if you'd prefer bounded concurrency here I'm happy to add it (there's no shared concurrency util today — `runWithConcurrency` exists only as local copies).
- **Skipped when binary storage is not configured**, so deployments without `binaryStorage` behave as before. Added `isBinaryStorageInitialized()` for that, since `getBinaryStorage()` throws.
- Extracted `getBinaryStorageKey(id, versionId)` in `storage/base.ts` so the key format has one definition; `BaseBinaryStorage.getKey` now calls it.

No new interface method was needed — `deleteFile` already exists on `BinaryStorage` and all four implementations.

## Test changes

Two tests in `repo.test.ts`:

- `Expunge Binary deletes the stored object for every version` — writes two versions, expunges, asserts `deleteFile` was called for both version keys **and** that `readFile` then rejects for both, so the bytes are really gone rather than merely requested for deletion.
- `Expunge non-Binary does not touch binary storage`.

Also verified `method-guards.test.ts`, `operations/expunge.test.ts`, `storage/filesystem.test.ts` and `cloud/aws/storage.test.ts` still pass, plus `tsc --noEmit`, eslint and prettier on the changed files.

## One open question for maintainers: versioned buckets

On a bucket with versioning enabled, `DeleteObject` without a `VersionId` only writes a delete marker — the bytes survive as a noncurrent version. `packages/cdk/src/storage.ts` provisions the storage bucket with `versioned: true` and no lifecycle rules, and nothing in the server sets or reads an S3 `VersionId`. So on a stock AWS deployment this change makes the object unreachable but does not erase it, which matters if the motivation is data erasure rather than storage cost.

I deliberately kept this PR to the minimal correct change rather than picking a direction. The options as I see them:

1. `ListObjectVersions` + `DeleteObjects` with explicit VersionIds — actually erases, but needs `s3:ListBucketVersions` and `s3:DeleteObjectVersion`, and Azure/GCS have their own soft-delete analogues, so an `{ allVersions?: boolean }` option on the interface gets awkward.
2. Add `noncurrentVersionExpiration` to the CDK bucket and document the erasure window.
3. Document the behavior only.

Happy to follow up with whichever you prefer.

Separately, and probably its own issue: `Expunger.expungeByResourceType` early-returns for `Binary`, and Binaries are otherwise discovered by matching `"url":"Binary/<id>"` in resource JSON, so Binaries referenced any other way survive a `Project/$expunge` entirely — rows and objects.


The-Alchemist:
@claude review

packages/server/src/storage/loader.ts
```diff
@@ -33,6 +33,17 @@ export function getBinaryStorage(): BinaryStorage {
...
 }
 
+/**
+ * Returns true if binary storage has been initialized.
+ *
+ * Callers that clean up storage opportunistically can use this to skip the work on deployments
+ * where `binaryStorage` is not configured, rather than handling the error from `getBinaryStorage`.
+ * @returns True if binary storage has been initialized.
+ */
+export function isBinaryStorageInitialized(): boolean {
```
The-Alchemist: hi, @ajwwong .  could you inline this one-line function?  thank you!

ajwwong: Done in 5f37151. The check reads the module-private `binaryStorage` variable, so instead of exporting a one-line boolean helper I moved the best-effort cleanup loop (`deleteBinaryStorageObjects`) into `loader.ts`, where the check is a plain inline `if (!binaryStorage)`. Net effect: the extra export is gone and `repo.ts` shrinks by the helper it was carrying. Tests/tsc/lint re-run green.

PR #10117, opened 2026-08-06
feat(react): Include goals immunizations in patient summary

techdavidy:
- Include Goals and Immunizations to PatientSummary.
- Include new modal to Patient Summary.

Screenshot 2026-08-06 at 10 57 06 AM


packages/react/src/PatientSummary/sectionConfigs.tsx
```diff
@@ -261,10 +285,12 @@ export const PharmaciesSection: PatientSummarySectionConfig = createPharmaciesSe
 export function getDefaultSections(onRequestLabs?: () => void): PatientSummarySectionConfig[] {
   return [
     DemographicsSection,
+    GoalsSection,
```
codyebberson: Perhaps minor, but I don't think "Goals" should be so close to the top of the list.

For historical context, "Goals" is a relatively recent addition to the USCDI data elements, and is perhaps least likely to be populated and/or used.  I would arguably put it closer to the bottom.

mattlong (approved):

PR #10119, opened 2026-08-06
fix(server): tolerate a stored resource with no meta on update

oleg-mp:
## Summary

A row seeded outside the normal write path can have content with no `meta` at all. `updateResourceImpl` cast `existing.meta` to `Meta` and wrote `compartment` onto it unguarded, so the update died with a bare `TypeError` ("Cannot set properties of undefined") that reached the caller as an opaque "Database error" rather than anything actionable.

Found while debugging a marketplace install against a project fixture in that state, but the fix is unrelated to the marketplace and stands on its own.

## Test plan

- [x] `packages/server` `repo.test.ts` — 97/97 passing
- [x] New regression test writes the malformed content with raw SQL and then drops the resource cache entry. Without that cache eviction the read is served from cache and the test passes with or without the fix, so the test would not be falsifiable.

### DIFF
--- a/packages/server/src/fhir/repo.ts
+++ b/packages/server/src/fhir/repo.ts
@@ -972,7 +972,11 @@ export class Repository extends FhirRepository implements Disposable {
 
     const existing = create ? undefined : await this.checkExistingResource<T>(resourceType, id);
     if (existing) {
-      (existing.meta as Meta).compartment = this.getCompartments(existing); // Update compartments with latest rules
+      // A stored resource is normally written with meta, but a row seeded outside
+      // the normal write path may not have it. Without this the assignment below
+      // throws a bare TypeError that surfaces as an opaque "Database error".
+      existing.meta ??= {};
+      existing.meta.compartment = this.getCompartments(existing); // Update compartments with latest rules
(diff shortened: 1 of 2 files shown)


packages/server/src/fhir/repo.ts
```diff
@@ -972,7 +972,11 @@ export class Repository extends FhirRepository implements Disposable {
 
     const existing = create ? undefined : await this.checkExistingResource<T>(resourceType, id);
     if (existing) {
-      (existing.meta as Meta).compartment = this.getCompartments(existing); // Update compartments with latest rules
+      // A stored resource is normally written with meta, but a row seeded outside
```
codyebberson, line 975: What's the use case for a row seeded outside of the normal write path? 

We are pretty explicit with customers and in our docs that our database is an internal implementation detail, and there is no "API" or compatibility guarantees.  The only "official" allowed write path is via `Repository`

oleg-mp: this came up during automated testing and while I was breaking the initial PR into small atomic modules. not necessary based on your clarification that we only have one official write path. I'll close it

PR #10122, opened 2026-08-06
feat(server): enforce lowercase `User.email` with a profile invariant

### LINKED ISSUE #10120: Enforce lowercase `User.email` with a profile invariant
## Problem

`User.email` is implicitly expected to be lowercase. All of the standard creation paths normalize it (`newuser`, `register`, `invite`, external auth), and `getUserByEmail` lowercases the value before searching. However, nothing enforces this at the resource level — a `User` created or updated with a mixed-case email (e.g. by a super admin via the FHIR API, or any code path that forgets to normalize) is then unreachable by login lookups, since the email-based searches only ever match lowercase values.

## Proposal

- Add a FHIRPath invariant on `User.email` in `profiles-medplum.json` (same mechanism as the existing `clapp-1` / `axp-3` invariants):
  ```
  key: user-1, severity: error
  expression: $this = $this.lower()
  ```
- Mention the requirement in the element's `definition` text so it surfaces in the reference docs and generated `@medplum/fhirtypes` JSDoc.
- Normalize the two remaining creation paths that don't lowercase today: Google sign-up (`claims.email`) and `defaultSuperAdminEmail` in seeding.

## Migration note

Any existing `User` rows with mixed-case emails (possible on older installs) will fail validation on their next update until the email is corrected.

ianplunkett:
Fixes #10120

## What

- Adds a `user-1` FHIRPath invariant on `User.email` in `profiles-medplum.json` (`$this = $this.lower()`, severity `error`), enforced at runtime by the resource validator like the existing `clapp-1` / `axp-3` invariants. A mixed-case email is now rejected with `Constraint user-1 not met: Email must be all lowercase`.
- Documents the requirement in the element's `definition` text, with the downstream regenerated artifacts: `@medplum/fhirtypes` `User.d.ts` JSDoc, `fhir.schema.json`, and the docs data file `medplumDefinitions/user.json` (which also picked up the previously missing `mfaMethod` element from a stale generation).
- Normalizes the two remaining `User` creation paths that didn't lowercase: Google sign-up (`claims.email`) and `defaultSuperAdminEmail` in seeding. All other creation paths (`newuser`, `register`, `invite`, external auth) already normalized.

## Why

`User.email` was implicitly lowercase: every standard creation path normalizes it and `getUserByEmail` lowercases before searching, but nothing enforced it at the resource level. A `User` written with a mixed-case email (e.g. by a super admin via the FHIR API) is unreachable by login lookups. See #10120.

## Testing

- New `User email constraint: user-1` test in `packages/core/src/typeschema/validation.test.ts` covering accept/reject cases.
- Full local CI replication passes: build, lint, madge, server seed test, and the complete test suite (4,190+ server tests, 43 other packages).

## Migration note

Existing `User` rows with mixed-case emails (possible on older installs) will fail validation on their next update until corrected.

### DIFF
--- a/packages/docs/static/data/medplumDefinitions/user.json
+++ b/packages/docs/static/data/medplumDefinitions/user.json
@@ -236,7 +236,7 @@
       "min": 0,
       "max": "1",
       "short": "",
-      "definition": "The email address that uniquely identifies the user.",
+      "definition": "The email address that uniquely identifies the user. Must be all lowercase.",
       "comment": "",
       "inherited": false
     },
@@ -333,6 +333,22 @@
       "max": "1",
       "short": "",
       "definition": "Whether the user has completed MFA enrollment.",
+      "comment": "",
+      "inherited": false
+    },
+    {
+      "name": "mfaMethod",
+      "depth": 1,
+      "types": [
+        {
+          "datatype": "code"
+        }
+      ],
+      "path": "User.mfaMethod",
+      "min": 0,
+      "max": "*",
+      "short": "",
+      "definition": "The MFA methods the user has enrolled in. 'totp' uses an authenticator application; 'email' sends a magic link to the user's email address on login.",
       "comment": "",
       "inherited": false
     },
--- a/packages/core/src/typeschema/validation.test.ts
+++ b/packages/core/src/typeschema/validation.test.ts
@@ -32,6 +32,7 @@ import type {
   StructureDefinitionSnapshot,
   SubstanceProtein,
   Timing,
+  User,
   ValueSet,
 } from '@medplum/fhirtypes';
 import { readFileSync } from 'fs';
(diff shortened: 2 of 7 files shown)


packages/docs/static/data/medplumDefinitions/user.json
```diff
@@ -336,6 +336,22 @@
...
+        }
+      ],
+      "path": "User.mfaMethod",
+      "min": 0,
+      "max": "*",
+      "short": "",
+      "definition": "The MFA methods the user has enrolled in. 'totp' uses an authenticator application; 'email' sends a magic link to the user's email address on login.",
+      "comment": "",
+      "inherited": false
+    },
```
The-Alchemist, line 354: is it just me, or is this change out of scope for this?

The-Alchemist:
do we want to add a database migration for this, @mattlong @mattwiller as well, in a separate PR if need be?

just checked prod, there's 87 users with Uppercase letters in their email addresses, minus a few deleted ones.

codyebberson:
> do we want to add a database migration for this, @mattlong @mattwiller as well, in a separate PR if need be?
> 
> just checked prod, there's 87 users with Uppercase letters in their email addresses, minus a few deleted ones.

@The-Alchemist  - good catch.  I'm guessing that those are not actually usable at the moment, because IIRC the login flow uses `toLower()` and selects for exact match.  We should verify before merging though.

packages/docs/static/data/medplumDefinitions/user.json
ianplunkett: I considered removing this but I kept it in because it was auto-generated. I think when `User.mfaMethod` was last updated, the documentation wasn't regenerated.

mattwiller (approved):

ianplunkett:
> > do we want to add a database migration for this, @mattlong @mattwiller as well, in a separate PR if need be?
> > just checked prod, there's 87 users with Uppercase letters in their email addresses, minus a few deleted ones.
> 
> @The-Alchemist - good catch. I'm guessing that those are not actually usable at the moment, because IIRC the login flow uses `toLower()` and selects for exact match. We should verify before merging though.

Anything on Medplum hosted should already fail for logins with uppercase characters in an email. I think we are safe to merge and if we have one-off support cases we can handle that.

PR #10124, opened 2026-08-06
[Scheduling] Share the scheduling duration units in `@medplum/core`

Filupnot:
Split out of #10044.

The server converts `SchedulingParameters` durations to minutes when it places appointments. The scheduling components in #10044 need the same conversion on the client to show how long a visit is. Rather than add a second copy, it makes sense to move it into `@medplum/core` where both can reach it.

### DIFF
--- a/packages/core/src/scheduling.ts
+++ b/packages/core/src/scheduling.ts
@@ -299,7 +299,7 @@ const MINUTES_PER_UNIT: Record<string, number | undefined> = {
  * @returns The length in minutes, or undefined when the duration has no value, a
  * negative value, or a unit scheduling does not accept.
  */
-export function durationToMinutes(duration: Duration | undefined): number | undefined {
+export function schedulingDurationToMinutes(duration: Duration | undefined): number | undefined {
   const value = duration?.value;
   if (value === undefined || value < 0) {
     return undefined;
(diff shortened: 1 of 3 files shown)


packages/core/src/scheduling.ts
```diff
@@ -106,3 +106,30 @@ export function extractServiceTypeReferences(
...
+};
+
+/**
+ * Converts a SchedulingParameters duration to minutes.
+ *
+ * @param duration - The duration to convert.
+ * @returns The length in minutes, or undefined when the duration has no value, a
+ * negative value, or a unit scheduling does not accept.
+ */
+export function durationToMinutes(duration: Duration | undefined): number | undefined {
```
noahsilas: I can imagine a couple ways that we could go with this in `core`:
- We could try to be more inclusive and handle a wide variety of Duration supported units. https://ucum.org/ucum#section-Derived-Unit-Atoms defines some interpretations for values like `mo` (month), which is defined as 1/12 of `a_j` (mean julian year = 365.25 days). The fact that our scheduling practices so far have excluded that value is more of a scheduling implementation detail than a feature of interpreting a Duration object. In the world where we are making a truly generic duration interpreting tool, it probably should also emit a value in seconds (which is the base SI time unit) or milliseconds (to match JS conventions).
- We could name this function in such a way as to make it more scheduling specific, so that other consumers don't try to add more units. Maybe we could define a `Scheduling` namespace that we can put more generic function names into? `Scheduling.durationToMinutes(duration)`? I'd be interested in feedback from the wider team about a namespacing pattern before we go too far down that pathway.

Filupnot: I agree that because of all the different ways that Durations might need to be supported across the app that having a generic `durationToMinutes` might need to be more inclusive. Rather than add scope in creating a new `Scheduling` namespace, I instead decided to rename it to `schedulingDurationToMinutes` to make it clear it's specific to scheduling. Curious what you think of this. 

I also considered removing this from Core altogether, but based on it being used on both the server and the client (as of #10044) core seems like the right place.

noahsilas (approved):

PR #10125, opened 2026-08-06
[React] Month navigation and availability marking on `CalendarDateInput`

### LINKED ISSUE #10275: [Scheduling] Find & Book / MVP

Filupnot:
Split out of #10044. 

# What 

`CalendarDateInput` lacks some functionality that we'll need in appointment-booking workflows, like: 
* Showing which days are worth clicking (e.g. has an available appointment slot)
* Selecting a day to mark as "chosen" 
* Ability for caller to page between months 

We'll add the following new props. They'll all be optional and default to their previous behavior, so existing callers will be unaffected (`CalendarInput` and `Scheduler`). 

| New prop |  |
| --- | --- |
| `month` | Makes the displayed month controlled, so a caller can keep it in step with its own search. Left out, the calendar still owns it. |
| `selected` | The day to mark as chosen. |
| `allowUnavailableDates` | Lets days that are marked as unavailable (aka not in the `availableDates` prop) to still be selectable. |
| `earliestDate` | A floor. Earlier days cannot be picked and the calendar will not page back past the month it falls in. A day that has passed is a different kind of empty from a day with nothing booked, so it stays closed even where `allowUnavailableDates` opens up the rest. |

# Screenshots 
## Selected day 

image

## Available days (preexisting)

image

## Allow unavailable days 

image

## Earliest Date 

image

## Month control (caller-owned)

image

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -30,7 +30,6 @@ export * from './auth/SignInForm';
 export * from './BackboneElementDisplay/BackboneElementDisplay';
 export * from './BackboneElementInput/BackboneElementInput';
 export * from './CalendarDateInput/CalendarDateInput';
-export * from './CalendarDateInput/CalendarDateInput.utils';
 export * from './CalendarInput/CalendarInput';
 export * from './chat/BaseChat/BaseChat';
 export * from './chat/ChatModal/ChatModal';
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -30,6 +30,7 @@ export * from './auth/SignInForm';
 export * from './BackboneElementDisplay/BackboneElementDisplay';
 export * from './BackboneElementInput/BackboneElementInput';
 export * from './CalendarDateInput/CalendarDateInput';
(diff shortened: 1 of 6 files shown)


packages/react/src/index.ts
```diff
@@ -30,6 +30,7 @@ export * from './auth/SignInForm';
 export * from './BackboneElementDisplay/BackboneElementDisplay';
 export * from './BackboneElementInput/BackboneElementInput';
 export * from './CalendarDateInput/CalendarDateInput';
+export * from './CalendarDateInput/CalendarDateInput.utils';
```
techdavidy: do we need this export ?

codyebberson:
@claude please review

codyebberson:
@Filupnot - can you please take a look at the two issues above?

1. `useMemo` for `buildGrid` never actually memoizes — `CalendarDateInput.tsx:30,43`
2. `moveMonth` sets internal state even when controlled — `CalendarDateInput.tsx:33-37`

codyebberson:
@claude - quick re-review please

codyebberson (approved):

PR #10127, opened 2026-08-06
[React] Range selection on `CalendarDateInput`

### LINKED ISSUE #10276: [Scheduling] Find & Book / Range selection
Allow user to select a (finite) date range in the time picker and view all time slots available in that window, for viewing availability across multiple days at the same time.

Filupnot:
Follow-up to #10125 

## Range support on `<CalendarDateInput />`

This component now accepts user-entered date ranges: 

image

| New Prop | Description | 
| --- | --- | 
| `range` | A stretch of days asked for, drawn as a band with both ends marked as chosen. A caller showing one day at a time wants `selected` instead. |
| `onSelectRange` | Takes the range asked for. Passing the callback is what turns the gesture on; without it a drag is just a click on the day it started from. |

A range is asked for by **dragging across the days**, or by **shift-clicking a second one**. A plain click is always a single day, whatever came before it.

## Deprecate `<CalendarInput />`

This is just a thin wrapper over `CalendarDateInput` that maps `Slot[]` to their start dates. Nothing in the repo uses it anymore (Scheduler moved to CalendarDateInput in #10125), so it's marked `@deprecated` here to point new consumers at the full component.

### DIFF
--- a/packages/react/src/CalendarDateInput/CalendarDateInput.tsx
+++ b/packages/react/src/CalendarDateInput/CalendarDateInput.tsx
@@ -11,6 +11,7 @@ import {
   isBeforeDay,
   isSameDay,
   sortEnds,
+  startOfDay,
   startOfMonth,
 } from './CalendarDateInput.utils';
 import { useDayRangeDrag } from './useDayRangeDrag';
@@ -57,7 +58,9 @@ export function CalendarDateInput(props: CalendarDateInputProps): JSX.Element {
     return (!day.available && !allowUnavailableDates) || (!!earliestDate && isBeforeDay(day.date, earliestDate));
   }
 
-  const range = drag.range ?? props.range;
+  // The grid holds local midnights, so a range whose ends carry a time of day
+  // would otherwise open a band its own start day sat outside of.
+  const range = toDays(drag.range ?? props.range);
   const selected = range ? undefined : props.selected;
 
   const grid = useMemo(() => buildGrid(month, props.availableDates), [month, props.availableDates]);
@@ -80,7 +83,7 @@ export function CalendarDateInput(props: CalendarDateInputProps): JSX.Element {
           </Button>
         </Group>
       </Group>
-      <table className={classes.table}>
+      <table className={cx(classes.table, onSelectRange && classes.selectsRanges)}>
         <thead>
           <tr>
             <th>SUN</th>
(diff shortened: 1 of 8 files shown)


techdavidy:
@claude review

techdavidy:
Good work on this one. 
Screenshot 2026-08-14 at 2 39 39 PM
looks weird on dark mode

Filupnot:
> Good work on this one. Screenshot 2026-08-14 at 2 39 39 PM looks weird on dark mode

Good catch, updated: 
image

techdavidy:
@claude please review

packages/react/src/CalendarDateInput/CalendarDateInput.tsx
```diff
@@ -81,17 +97,49 @@ export function CalendarDateInput(props: CalendarDateInputProps): JSX.Element {
...
+                      }}
+                      onPointerOver={() => drag.extend(day.date)}
+                      onClick={(event) => {
+                        // A drag that crossed days has already asked for them, and
+                        // the browser reports the release as a click as well.
+                        if (drag.consumeClick()) {
+                          return;
+                        }
+                        const anchor = props.selected ?? props.range?.start;
+                        if (event.shiftKey && anchor && onSelectRange && !isSameDay(anchor, day.date)) {
```
noahsilas: The shift-click is a nice touch!

It's a little strange that it works differently when shift-clicking after the range (which increases the range to cover up to the selected date), and shift-clicking before the range (where it change the range to be from the targeted date to the start of the range). I'd sort of expect it to exclusively enlarge the range, which is probably an intuition I have that was informed by how text selection works on MacOS.  

Not a blocker, but it would be nice to have something we can point to as explanatory for why it works in this way.

noahsilas (approved):

techdavidy:
On my side I have a question. was the drag and drop a requirement. Find it odd to have to shift click it to select a range with just clicks. Not sure how to make it work on touch devices like iPad without the drag and drop.

techdavidy:
On my side I have a question. was the drag and drop a requirement. Find it odd to have to shift click it to select a range with just clicks. Not sure how to make it work on touch devices like iPad without the drag and drop.

Filupnot:
> On my side I have a question. was the drag and drop a requirement. Find it odd to have to shift click it to select a range with just clicks. Not sure how to make it work on touch devices like iPad without the drag and drop.

It also works with drag + drop

packages/react/src/CalendarDateInput/CalendarDateInput.tsx
Filupnot: Updated so that shift+clicking when there's a range already defined will move the nearest end.

PR #10135, opened 2026-08-07
fix(server): apply a FHIRcast context update only to the version it was built on

ThatOneBro:
Stacked on #10096 — review that one first; this diff is only the last commit.

## The problem

Applying a `DiagnosticReport-update` is a read-modify-write:

1. read the topic's current context
2. reject unless `event['context.versionId']` matches its `context.versionId`
3. apply the subscriber's update bundle to the content bundle
4. write the result back under a freshly generated version

Step 2 validated against a value that had already been read, so two updates racing on one topic both passed it. The write that landed second stored a context built from the version it read, dropping the first update's changes, and published a `context.priorVersionId` naming a version that was never the one it was actually applied to. Subscribers reconstructing the content from the event stream see a version chain that never happened.

The window is small but not theoretical: it is exactly the multi-subscriber case #10096 exists to support — two apps sharing a topic, both driving updates.

## The change

The write is now conditional on the context still being at the version the update was built on. `compareAndSetTopicCurrentContext` compares and replaces inside a Lua script, which Redis runs to completion before serving another command, so nothing can land in between:

```lua
local stored = redis.call('GET', KEYS[1])
if not stored then
  return 0
end
local ok, context = pcall(cjson.decode, stored)
if not ok or context['context.versionId'] ~= ARGV[1] then
  return 0
end
redis.call('SET', KEYS[1], ARGV[2])
return 1
```

The up-front version check stays — it gives a stale client a precise 400 without doing the work. The conditional write is what catches a racing update, and the loser gets a **409** so a client can tell a conflict it should retry from a request it should not resend. Nothing is published when the write is refused.

The bundle merge stays in TypeScript. Porting `processUpdateBundle` to Lua would move FHIR reference resolution and validation into an unreviewable script for no additional safety — compare-and-set gives the same guarantee.

## Scope

Only the update path. `-open` and `-close` also read-modify-write, but they establish a context rather than extend a version chain, and a racing `-open` or `-close` now correctly fails an in-flight update's conditional write.

## Tests

- `compareAndSetTopicCurrentContext` unit tests: replaces at the expected version, leaves a context that moved on, establishes nothing for a topic with no context.
- A route-level test that drives the race deterministically — a second update commits between this one's read and its write — asserting 409 and that the winner's context is left untouched. Verified load-bearing: reverting to the unconditional write fails it.

93 tests pass across `fhircast/routes`, `fhircast/utils` and `ws/fhircast`; `tsc --noEmit`, eslint and prettier clean.

### DIFF
--- a/packages/server/src/fhircast/routes.test.ts
+++ b/packages/server/src/fhircast/routes.test.ts
@@ -28,7 +28,13 @@ import { getCacheRedis } from '../redis';
 import { createTestProject, withTestContext } from '../test.setup';
 import type { EventCategory } from './routes';
 import { getEventCategory } from './routes';
-import { extractEndpoint, getEndpointSubscription, setTopicCurrentContext } from './utils';
+import {
+  extractEndpoint,
+  getCurrentContext,
+  getEndpointSubscription,
+  getTopicCurrentContextKey,
+  setTopicCurrentContext,
+} from './utils';
 
 const STU2_BASE_ROUTE = '/fhircast/STU2';
 const STU3_BASE_ROUTE = '/fhircast/STU3';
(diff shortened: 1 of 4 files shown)


ThatOneBro:
@claude Can you review this PR and see if there are any issues with regards to the atomicity and in general the approach to using this Lua script in Redis?

Can you analyze what would happen if the user sent an event without a `priorVersionId` (let's say, a Home-open event)? What would happen in that scenario?

codyebberson (approved):
nice 👍

PR #10142, opened 2026-08-07
Widen flatMapFilter to accept Iterable

The-Alchemist:
- Widen `flatMapFilter` from `T[] | undefined` to `Iterable<T> | undefined`, iterating with `for...of` while preserving sequential indexes.
- Add coverage for `undefined`, non-array iterables (`Set`), and index ordering.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1561,14 +1561,18 @@ export function removeProfileFromResource<T extends Resource = Resource>(resourc
   return resource;
 }
 
-export function flatMapFilter<T, U>(arr: T[] | undefined, fn: (value: T, idx: number) => U | U[] | undefined): U[] {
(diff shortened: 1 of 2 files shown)


packages/core/src/utils.ts
```diff
@@ -1561,14 +1561,18 @@ export function removeProfileFromResource<T extends Resource = Resource>(resourc
...
   const result: U[] = [];
-  if (!arr) {
+  if (!items) {
     return result;
   }
 
-  for (let i = 0; i < arr.length; i++) {
-    const resultValue = fn(arr[i], i);
+  let i = 0;
+  for (const value of items) {
```
ThatOneBro, line 1574: This does de-optimize this particular function for many cases where the numeric for loop is significantly faster

I would like to see numbers on how this version compares to the original version when operating on simple arrays

Issue #10143, opened 2026-08-07
[Scheduling] Location, patient, and visit type select components


Filupnot:
Piece **A** of the Find & Book roll-out plan for #10003 — see the
[plan comment](https://github.com/medplum/medplum/issues/10003#issuecomment-5210569771)
for the full tree and why the order is what it is.

Delivered by #10144.

## What lands here

| Component | Role |
| --- | --- |
| `AppointmentLocationSelect` | Choose the site an appointment is booked at |
| `AppointmentPatientSelect` | Choose the patient, searchable by name, MRN, or date of birth |
| `AppointmentServiceSelect` | Choose the visit type, narrowed by the chosen site |

Plus the supporting pieces those two need: shared story fixtures in
`stories/scheduling.ts`, an autocomplete test helper, and medical record numbers
on the mock Simpsons patients so patient lookup by MRN is demonstrable.

## The fields do not share a primitive

The plan originally named a shared `AppointmentPickList` and a
`usePickListSearch` hook here as well. **Neither ships**, and that is deliberate.
The fields split cleanly in two:

- **Bounded — one search, filtered in the browser.** Location. Every site a
  practice has fits in one page, so Mantine's `Select` holds them all
- **Unbounded — server-side search with paging.** Patient and visit type.
  Neither has a knowable upper bound, so both are `AsyncAutocomplete`-family

A shared primitive could serve both only by flattening what each one needs. It
was written, then removed once that became clear.

**`AppointmentServiceSelect` moved here from piece G** on 2026-08-10, so that all
three fields settle their conventions in one review. G is reduced to `params`
alone, which only the wizard then needs.

This adds `@medplum/react` to `react-scheduling`'s **peer** dependencies. Worth
knowing when reviewing, since the package was created deliberately free of that
dependency — but a peer dependency, so nothing is bundled.

## Why this piece was first

**A needs no utils modules at all.** `AppointmentFinder.utils` is being
introduced module by module — each PR brings the modules it is the first to
need — and these components need none of them. That made A the only piece with
zero dependencies: it targets `main`, and neither blocks nor is blocked by B
(#10130).

A gates piece **G** (`params` + `AppointmentServiceSelect`).

## Done when

- [x] Both components are exported from `@medplum/react-scheduling`
- [x] Tests ship with the components and pass
- [x] Stories render against the shared fixtures
- [x] No dependency on any `AppointmentFinder.utils` module

PR #10144, opened 2026-08-07
[Scheduling] Service select component and react-scheduling test setup

### LINKED ISSUE #10275: [Scheduling] Find & Book / MVP

Filupnot:
Split out of #10044. Addresses #10143. Adds a field for selecting the service for an appointment, plus the test and Storybook setup that `react-scheduling` needed before it could host components built on `@medplum/react`.

This started as three fields: location, patient, and service. Per [review](https://github.com/medplum/medplum/pull/10144#discussion_r3770939438), the location and patient fields are omitted. Both were thin wrappers over `ResourceInput`, and the details they chose to display (city/state for a location, DOB and MRN for a patient) are specific enough to one use case that a shared component would get in the way more often than it helps. Callers inline `ResourceInput` directly instead.

# `AppointmentServiceSelect`

Field for choosing the service provided in an appointment.

image

- Options are subtitled with service type and length (e.g. _Consultation - 30 min_)
- Automatically filters out services that don't have a `SchedulingParameters` extension. The filter runs client side because no search parameter covers an extension.
- Exposes a `location` prop to filter results based on `HealthcareService.location`
- Based on `AsyncAutocomplete` (not `ResourceInput` because that component doesn't expose `loadOptions`)

`AppointmentOptionRow` is exported alongside it: the two line option row, name over a dimmed detail line, that the field renders. Callers inlining their own pick lists can use it so their options match.

# Other changes

- Add `@medplum/react` and `@medplum/react-hooks` as peer dependencies to `@medplum/react-scheduling`. This package could not build on `@medplum/react` components at all before.
- New test utils in `react-scheduling`: rendering against a `MockClient`, and driving a debounced `AsyncAutocomplete` under fake timers
- Storybook fixtures and a decorator to load them, plus an entry for `AppointmentServiceSelect`
- Mock support for searching `HealthcareService` by `name` and `location`
- Exclude `test-utils/` from Sonar. It already excluded `test-utils.*` files, just not a directory of that name.
- Make `ResourceInput.tsx` _not_ be deprecated (it was marked so by mistake)


Filupnot:
@claude review this please

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
```diff
@@ -0,0 +1,141 @@
...
+ * a typo.
+ *
+ * @param props - The React props.
+ * @param props.search - What was searched for, empty when nothing was typed.
+ * @returns The message.
+ */
+function ServiceEmpty(props: { readonly search: string }): JSX.Element {
+  return (
+    <Text size="sm" c="dimmed" p="xs">
+      {props.search ? 'No visit types match this search.' : 'No schedulable visit types found.'}
```
techdavidy, line 112: will this be reachable? I remember AsyncAutocomplete not showing the dropdown when there are no results.

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
```diff
@@ -0,0 +1,141 @@
...
+export interface AppointmentServiceSelectProps {
+  /**
+   * The visit type the appointment is for. Reassigning it moves the field and clearing it
+   * empties the field, so a caller holding it in state can set, restore or clear the answer.
+   * Every choice the user makes is reported through `onChange`.
+   *
+   * Clearing is what a `location` change needs: the new site may not offer the visit type on
+   * screen, and only the caller knows that it has to go.
+   */
+  readonly service: WithId<HealthcareService> | undefined;
```
techdavidy, line 29: defaultValue maybe ?

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
```diff
@@ -0,0 +1,82 @@
...
+export function AppointmentLocationSelect(props: AppointmentLocationSelectProps): JSX.Element {
+  const { location, onChange, label = 'Location', error, disabled } = props;
+
+  const handleChange = useCallback((locations: WithId<Location>[]) => onChange(locations[0]), [onChange]);
+
+  return (
+    <MultiResourceInput<WithId<Location>>
+      // `MultiResourceInput` reads its value once, on mount, so the only way to show a site
+      // assigned later is to mount a new one. Keying on the chosen id does that, and leaves
+      // a caller that never reassigns the prop with a key that never moves.
```
noahsilas, line 43: I wonder how difficult it would be to make this a controlled component when `value` is present in its props? It would be nice not to need to re-key this to force "defaultValue" to get picked up again.

Another option would be to embrace that style and change the prop that this accepts from `location` to `defaultValue` or similar, and remove the tension between controlled/uncontrolled components from this layer.

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
```diff
@@ -0,0 +1,82 @@
...
+      key={location?.id ?? 'empty'}
+      resourceType="Location"
+      name="location"
+      label={label}
+      placeholder="Search sites"
+      required
+      maxValues={1}
```
noahsilas, line 50: We still expose `ResourceInput` to make this a little tidier. It looks like that was maybe mistakenly tagged with an `@deprecated` tag, which I don't think was intentional (cc @techdavidy to confirm?).

packages/react-scheduling/src/AppointmentFinder/AppointmentPatientSelect.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const itemComponent = useCallback(
+    (option: AsyncAutocompleteOption<WithId<Patient>>): JSX.Element => (
+      <AppointmentOptionRow label={option.label} detail={formatPatientDetail(option.resource, mrnSystem)} />
+    ),
+    [mrnSystem]
+  );
+
+  return (
+    <MultiResourceInput<WithId<Patient>>
+      // `MultiResourceInput` reads its value once, on mount, so the only way to show a
```
techdavidy, line 58: let's clean up these comments

packages/react-scheduling/src/AppointmentFinder/AppointmentPatientSelect.tsx
```diff
@@ -0,0 +1,104 @@
...
+const MRN_TYPE_SYSTEM = `${HTTP_TERMINOLOGY_HL7_ORG}/CodeSystem/v2-0203`;
+const MRN_TYPE_CODE = 'MR';
+
+export interface AppointmentPatientSelectProps {
+  /**
+   * The patient the appointment is for. Reassigning it moves the field and clearing it
+   * empties the field, so a caller holding it in state can set, restore or clear the answer.
+   * Every choice the user makes is reported through `onChange`.
+   */
+  readonly patient: WithId<Patient> | undefined;
```
techdavidy, line 29: defaultValue maybe ?

packages/react-scheduling/src/AppointmentFinder/AppointmentPatientSelect.tsx
```diff
@@ -0,0 +1,104 @@
...
+  // render it from, so the MRN system has to be closed over.
+  const itemComponent = useCallback(
+    (option: AsyncAutocompleteOption<WithId<Patient>>): JSX.Element => (
+      <AppointmentOptionRow label={option.label} detail={formatPatientDetail(option.resource, mrnSystem)} />
+    ),
+    [mrnSystem]
+  );
+
+  return (
+    <MultiResourceInput<WithId<Patient>>
```
techdavidy, line 57: we have a <ResourceInput> for a single value rather than using arrays

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
```diff
@@ -0,0 +1,141 @@
...
+  return <AppointmentOptionRow label={props.label} detail={formatServiceDetail(props.resource)} />;
+}
+
+function formatServiceDetail(service: WithId<HealthcareService>): string | undefined {
+  const category = formatCodeableConcept(service.type?.[0]);
+  const duration = getServiceDurationMinutes(service);
+  // `!== undefined` rather than a truthiness check: a zero-minute service is configured,
+  // and reading as if it were not would hide a misconfiguration `$find` will accept.
+  const parts = [category, duration !== undefined ? `${duration} min` : undefined].filter(Boolean);
+  return parts.length > 0 ? parts.join(' · ') : undefined;
```
techdavidy, line 140: nit: return  parts.join(' · ') || undefined;

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * @param props - The React props.
+ * @returns The location field.
+ */
+export function AppointmentLocationSelect(props: AppointmentLocationSelectProps): JSX.Element {
+  const { location, onChange, label = 'Location', error, disabled } = props;
+
+  const handleChange = useCallback((locations: WithId<Location>[]) => onChange(locations[0]), [onChange]);
+
+  return (
+    <MultiResourceInput<WithId<Location>>
```
techdavidy, line 40: we can use <ResourceInput> here for single value

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
noahsilas, line 1: This feels like a pretty thin wrapper around MultiResourceInput. I wonder if this might be better off written inline wherever it is needed 🤔

techdavidy: I agree, only AppointmentServiceSelect seems a bit more robust.

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.stories.tsx
```diff
@@ -0,0 +1,59 @@
+// SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
```
techdavidy, line 1: I think we can drop these since it mostly showcase AsyncAutocomplete use cases.

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
techdavidy: Yeah wasn't intentional. we can remove the @deprecated tag.

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
Filupnot: It's not - good callout, removed

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
Filupnot: Talked with David about this a bit, and went for what's essentially your second option. Changed these components to remove `key` and accept `defaultValue`, so that the callers of `Appointment*Select` will be responsible for re-keying. This should make the behavior more clear.

I think the _ideal_ fix here would be to update `AsyncAutocomplete` to accept a controlled state (`value`), but considering the file is pretty old & core to many components, probably not something we want to do here.

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
Filupnot: I'd like to keep this because:
- Inlining it would mean that the caller would need to locally define search criteria (currently `LOCATION_SEARCH_CRITERIA`) and the component `LocationItem`. Having this component exist seems like a low-weight way to avoid potentially duplicative work.
- It's consistent with the other `Appointment*Select` components.

If either of you feel strongly about this then I can remove it, but first wanted to jot down my thoughts.

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.stories.tsx
Filupnot: The service one is a bit more heavyweight than the other ones, so I decided to get rid of all the stories but the following: 
- `AppointmentServiceSelect`: `Basic` and `NarrowedToASite` to showcase how the location affects the options
- `AppointmentLocationSelect`: `Basic` since the option subtitle highlights how we're differentiating locations that look the same via the town they're in.

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
noahsilas: I see what you are saying, but I think we should try to bias towards a little more minimalism. In this particular case, I think that having more flexibility is something I value. For instance, I can imagine that there may be scenarios where we want to adjust if the location is "required" (which we could add as another passthrough prop), or tweak the language on the placeholder ("sites" may not be the universal).

I think I would lean in a similar direction for `<AppointmentPatientSelect />`.

There might be value in keeping some of the display components so we can test them very directly (instead of having to jump through some hoops to get options to render from the autocomplete), although I might inline more of the helper methods into them. Something like:
```tsx
export function LocationItem(props: Readonly<AsyncAutocompleteOption<Location>>): JSX.Element {
  const location = props.resource;
  const detailParts = [location.address?.city, location.address?.state].filter(Boolean);
  const detail = detailParts.join(', ') || undefined;
  return <AppointmentOptionRow label={props.label} detail={detail} />;
}
```

That said, even this is fairly specific to a use case where the address is the relevant detail - you can imagine scenarios where `location.description` might be the more relevant details to display.

I think I would argue fairly strongly that we should omit this and `AppointmentPatientSelect` for these reasons.

packages/react-scheduling/src/AppointmentFinder/AppointmentPatientSelect.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * @param system - The system MRNs are issued under, for a project that does not type them.
+ * @returns The MRN, or undefined when the patient carries nothing that says it is one.
+ */
+function getMrn(patient: Patient, system: string | undefined): string | undefined {
+  if (system) {
+    return getIdentifier(patient, system);
+  }
+  return patient.identifier?.find((identifier) =>
+    identifier.type?.coding?.some((coding) => coding.system === MRN_TYPE_SYSTEM && coding.code === MRN_TYPE_CODE)
+  )?.value;
```
noahsilas: This does seem like a potentially useful utility for us to extract; TIL about this method for encoding that an Identifier is an MRN!

For anyone else looking to learn more, some docs on this system are here: https://terminology.hl7.org/2.1.0/CodeSystem-v2-0203.html

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * @param props - The option to render.
+ * @returns The row.
+ */
+function ServiceItem(props: Readonly<AsyncAutocompleteOption<WithId<HealthcareService>>>): JSX.Element {
+  return <AppointmentOptionRow label={props.label} detail={formatServiceDetail(props.resource)} />;
+}
+
+function formatServiceDetail(service: WithId<HealthcareService>): string | undefined {
+  const category = formatCodeableConcept(service.type?.[0]);
+  const duration = getServiceDurationMinutes(service);
```
noahsilas, line 101: It's possible (though discouraged) that an individual Schedule using this service overrides the duration. I mostly point this out as something to keep in mind during future development; we may want to add some extra annotation to a result where the appointment duration coming back from `$find` differs from this service duration.

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * How many visit types one search offers, and the order they come back in.
+ */
+const SERVICE_SEARCH_CRITERIA = { _count: '25', _sort: 'name' };
+
+export interface AppointmentServiceSelectProps {
+  readonly defaultValue?: WithId<HealthcareService>;
+  readonly onChange: (service: WithId<HealthcareService> | undefined) => void;
+  /** A chosen site, which narrows the services on offer to the ones held there. */
+  readonly location?: WithId<Location>;
```
noahsilas: I think that the more likely input value here will be a `Reference<Location>`, in case we have not already fetched the related Location object.

packages/react-scheduling/src/AppointmentFinder/AppointmentLocationSelect.tsx
Filupnot: Thanks for elaborating - I can agree with your points. Removed!

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
Filupnot: It would be tricky to show the schedule's override duration in this field because in the workflow the user would likely be filling this out before _picking_ an actor. However, once we know the actor(s) and run `$find`, _then_ we'll be able to show the slot durations in the individual time cards based on that override. So I'd advocate to not make any changes to this PR.

packages/react-scheduling/src/AppointmentFinder/AppointmentServiceSelect.tsx
Filupnot: Good call, updated to take either.

packages/react-scheduling/src/AppointmentFinder/AppointmentPatientSelect.tsx
Filupnot: I actually removed this from this PR along with the `AppointmentPatientSelect` component since it wasn't being used. But I agree with this, and will add this to Core in a separate PR.

noahsilas (approved):

PR #10149, opened 2026-08-08
Fix CORS Access-Control-Max-Age

codyebberson:
Remove the default `cache: 'no-cache'` option from MedplumClient requests.

Setting `cache: 'no-cache'` causes browsers to bypass the CORS preflight cache, resulting in an unnecessary `OPTIONS` request before every cross-origin API request. The browser's default fetch caching behavior is sufficient, and API response caching remains controlled by the server's `Cache-Control` headers.

This change leaves the `cache` option unset by default while continuing to respect it when explicitly provided by the caller.


codyebberson:
@claude please review

techdavidy (approved):

PR #10150, opened 2026-08-08
Compute DICOM Study level aggregates

codyebberson:
The initial DICOMweb PR hardcoded four aggregate attributes in the QIDO-RS response:

```ts
ModalitiesInStudy: ['DX'], // TODO: Add support for multiple modalities.
NumberOfStudyRelatedSeries: 1,
NumberOfStudyRelatedInstances: 1,
NumberOfSeriesRelatedInstances: 1,
```

The obvious fix — read `ModalitiesInStudy` (0008,0061) from the incoming instance — does not work. That attribute belongs to no composite IOD. It is a Study level query key that the archive is expected to compute (PS3.4 C.6.2.1.2), so conformant stored instances do not carry it, and a sender that does include it is describing its own view of the study rather than ours. The same is true of `NumberOfStudyRelatedSeries` (0020,1206), `NumberOfStudyRelatedInstances` (0020,1208), and `NumberOfSeriesRelatedInstances` (0020,1209).

Medplum is the archive here, so computing these is our job.

## What

Two functions in `packages/server/src/dicom/utils.ts`, both recomputing from stored resources rather than from the incoming instance.

`updateStudyAggregates(repo, studyId)`:

- `modalitiesInStudy` — the sorted, deduped, upper-cased union of `DicomSeries.modality` (0008,0060), which *is* Type 1 in the General Series module and therefore reliable. Mixed modality studies (PET/CT, CT + SR, DX + PR) are the normal case, and the union is the spec definition rather than a workaround.
- `numberOfStudyRelatedSeries` — falls out of the same paginated series scan at no extra cost.
- `numberOfStudyRelatedInstances` — an accurate count of the study's instances.

`updateSeriesAggregates(repo, seriesId)` sets `numberOfSeriesRelatedInstances` from an accurate count of that series' instances.

STOW-RS calls both after every instance in the request has been stored: once per distinct study, and once per distinct series *the request actually touched*. A 500-slice single-series CT produces one study update and one series update, not 1000.

`medplumStudyToDcmjsStudy` and `medplumSeriesToDcmjsSeries` now emit the stored values.

## Implementation notes

**Recompute, don't append.** Adding the incoming instance's modality to the existing list would be cheaper, but recomputing from the stored series makes the operation idempotent. A retry, a re-upload, or an upload following a partial failure all converge on the same answer, and studies that predate this change repair themselves on the next upload without a separate backfill.

**Optimistic locking.** Concurrent STOW requests to the same study are a genuine read-modify-write race, and losing it silently drops a modality — which would then drop the study out of any modality-filtered QIDO search. Both paths share `updateWithRetry`, which guards the write with `ifMatch` on the resource's `versionId` and retries up to three times on `412`.

The retry is what makes this correct rather than best-effort: the loser of a race re-reads *after* the winner committed, so it always recomputes from strictly fresher data. With two concurrent writers one retry converges deterministically; three attempts covers three-way contention. Errors are also one-directional — because each write is a complete snapshot and instances are only ever added during ingest, a stale write can only ever omit data, never invent it.

**Only touched series are updated.** The study pass scans every series, because a new series changes the study's modality union. The series pass deliberately does not: a series that received no instances in this request cannot have changed its own count. This keeps the common single-series STOW at two extra queries and one extra write rather than scaling with the size of the study.

The asymmetry is worth noting explicitly — a `DicomStudy` recomputes from everything stored and therefore self-heals, while a `DicomSeries` only recomputes when touched. Nothing in this PR repairs pre-existing drift at either level, and there is no launched data to repair.

**No version churn.** Updates are skipped entirely when the recomputed values match what is already stored, so steady-state uploads do not create new versions.

**Failure is non-fatal.** The instances are already stored by the time this runs. A failed update is logged and the STOW response still succeeds; the next upload recomputes.

### DIFF
--- a/packages/server/src/dicom/routes.test.ts
+++ b/packages/server/src/dicom/routes.test.ts
@@ -332,6 +332,20 @@ describe('DICOM Routes', () => {
     expect(res).toHaveStatus(200);
     expect(JSON.stringify(res.body)).toContain('1.2.840.10008.5.1.4.1.1.7');
     expect(JSON.stringify(res.body)).toContain('1.2.826.0.1.3680043.10.543.1');
+
+    // The uploaded instance carries no ModalitiesInStudy (0008,0061), so the value below can only
+    // come from reconciling the study against its stored series.
+    const studies = await request(app)
+      .get(`/dicomweb/studies`)
+      .set('Authorization', 'Bearer ' + accessToken);
+    expect(studies).toHaveStatus(200);
+    const stored = (studies.body as Record<string, { Value?: unknown[] }>[]).find(
+      (study) => study['0020000D']?.Value?.[0] === '1.2.826.0.1.3680043.10.543.2'
+    );
+    expect(stored?.['00080061'].Value).toStrictEqual(['OT']);
+    // Both counts have VR IS, which dcmjs denaturalizes to a string.
+    expect(stored?.['00201206'].Value).toStrictEqual(['1']);
+    expect(stored?.['00201208'].Value).toStrictEqual(['1']);
   });
 
   test('Direct handler validation errors', async () => {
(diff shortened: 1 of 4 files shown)


reshmakh (approved):

codyebberson:
@claude please review

PR #10151, opened 2026-08-08
Fixes #10108 - CLI login scope flag

### LINKED ISSUE #10108: CLI `medplum login` ignores `--scope`, so `offline_access` can't be requested for a refresh token
The default `authorization-code` flow in `medplum login` hardcodes `scope=openid` and never reads the `--scope` flag:

https://github.com/medplum/medplum/blob/main/packages/cli/src/auth.ts#L143-L145

This means there's no way to pass `--scope "openid offline_access"` to get a refresh token from the CLI, even though the SDK (`MedplumClient.startLogin`/`signInWithRedirect`) already supports this exact pattern.

**Expected:** `medplumAuthorizationCodeLogin` should include `profile.scope` (falling back to `openid`) in the authorize request, so `medplum login --scope "openid offline_access"` works like it does in the SDK.

codyebberson:
The `medplum login` command now:
1. ✅ Properly respects the `--scope` flag for authorization code flow
2. ✅ Defaults to `'openid offline_access'` to provide refresh tokens by default

## Problem

Previously, the scope was hardcoded to `'openid'` and the `--scope` flag was completely ignored during authorization code login. This meant users couldn't get refresh tokens even when explicitly requesting them.

## Solution

- Pass `profile` parameter to `medplumAuthorizationCodeLogin` to access the scope option
- **Changed default from `'openid'` to `'openid offline_access'`** - users now get refresh tokens by default
- Users can still override with `--scope` flag if needed
- Updated `--scope` option description with helpful example
- Added comprehensive test coverage

## Usage

**Default behavior (gets refresh token):**
```bash
medplum login  # Now defaults to 'openid offline_access'

### DIFF
--- a/packages/cli/src/auth.test.ts
+++ b/packages/cli/src/auth.test.ts
@@ -111,20 +111,14 @@ describe('CLI auth', () => {
   });
 
   test('Login with default scope includes offline_access', async () => {
-    let browserUrl: string | undefined;
-    (cp.exec as unknown as Mock).mockImplementation(
-      (cmd: string, callback: (error: Error | null, stdout: string, stderr: string) => void) => {
-        // Extract the URL from the browser open command
-        const urlMatch = cmd.match(/(https?:\/\/[^\s'"]+)/);
-        if (urlMatch) {
-          browserUrl = urlMatch[1];
-        }
-        if (callback) {
-          callback(null, '', '');
-        }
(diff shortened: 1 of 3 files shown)


codyebberson:
@claude please review

techdavidy (approved):

PR #10154, opened 2026-08-10
DICOM: de-dupe stored instances + search pagination

codyebberson:
Two independent DICOMweb fixes on this branch (kept together while DICOMweb is still early access).

## 1. STOW-RS: conditional create for DicomInstance

`handleStoreInstances` already used `conditionalCreate` for `DicomStudy` and `DicomSeries`, but wrote each `DicomInstance` with a plain `createResource`. Re-storing a SOP instance that was already stored — a normal, idempotent STOW-RS operation — therefore produced a duplicate `DicomInstance` row every time.

The instance write now uses `repo.conditionalCreate` filtered on the `sop-instance-uid` search parameter, matching the study/series pattern directly above it. Re-uploading the same SOP instance resolves the existing resource instead of creating a new one.

## 2. QIDO-RS / WADO-RS: search pagination

- **QIDO-RS study and series search** (`qido-rs.ts`) now honor `limit` and `offset` query parameters, parsed via a small `parseQueryInt` helper (ignores absent/invalid/negative values so the server default and its max-count clamp of 1000 still apply).
- **WADO-RS series metadata** (`wado-rs.ts`) now requests `count: 1000` (server max) instead of relying on the default page size of 20. A CT/MR series routinely has hundreds of instances, so the default would silently truncate the series returned to the viewer.

### DIFF
--- a/packages/server/src/dicom/qido-rs.ts
+++ b/packages/server/src/dicom/qido-rs.ts
@@ -5,7 +5,7 @@ import type { DicomSeries, DicomStudy } from '@medplum/fhirtypes';
 import dcmjs from 'dcmjs';
 import type { Request, Response } from 'express';
 import { getAuthenticatedContext } from '../context';
-import { medplumSeriesToDcmjsSeries, medplumStudyToDcmjsStudy } from './utils';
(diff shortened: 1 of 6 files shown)


codyebberson:
@claude please review

reshmakh (approved):

PR #10157, opened 2026-08-10
perf(fhir-router): optimize conditional operations to reduce 40001 conflicts

codyebberson:
Carrying over from https://github.com/medplum/medplum/pull/9916

Bringing into `medplum` organization for access to all static analysis tools

@lsub-ai's original description:

--------

Redesigns conditionalCreate, conditionalUpdate, conditionalDelete, and
conditionalPatch in FhirRepository to use an Optimistic Search pattern
(Double-Checked Locking).

Rather than forcing the entire operation (search + write) into a
SERIALIZABLE transaction, we perform an optimistic search first. If exactly
one match is found (representing the vast majority of load under updates/deletes/patches
on existing resources), we execute the write within a normal REPEATABLE READ
(serializable: false) transaction. This completely avoids coarse GIN
index page-level predicate locks and false-positive 40001 serialization
errors on the happy path.

### DIFF
--- a/packages/fhir-router/src/repo.test.ts
+++ b/packages/fhir-router/src/repo.test.ts
@@ -319,5 +319,46 @@ describe('MemoryRepository', () => {
       const results = await Promise.all([p1, p2]);
       expect(results.map((r) => r.outcome.id)).toEqual(['created', 'ok']);
     });
+
+    test('conditionalUpdate', async () => {
+      const patient: Patient = {
+        resourceType: 'Patient',
+        id: 'update-test',
+        active: true,
+      };
+
+      const c1 = await repo.conditionalUpdate(patient, parseSearchRequest('Patient?_id=update-test'), { assignedId: true });
+      expect(c1.outcome.id).toBe('created');
+
+      patient.active = false;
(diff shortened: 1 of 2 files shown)


codyebberson:
@claude - please review.  please ignore lack of test coverage and minor nits.  we are currently focused on big picture strategy and long tail implications.  primarily, is this safe, or does it jeopardize transactional guarantees

codyebberson:
@claude - how is `conditionalUpdate` still transactionally safe if `serializable: false`?

1. User A starts conditionalUpdate, runs initial optimistic search, passes
2. User A starts non-serializable transaction
3. User A runs second search, passes
4. User B updates the resource
5. User A completes the conditionalUpdate

codyebberson:
@claude - but what if another resource is created that causes an additional match to the `conditionalUpdate` search criteria?  Your logic holds for the case where the one match changes.  But another resource could also invalidate the transaction, right?

lsub-ai:
@codyebberson noticed this got closed without a note — was that a "no" on the optimistic-search approach, Is there a fatal flaw in the implementation (beyond the non-retryable 409 I flagged) ?

PR #10162, opened 2026-08-10
fix(cli): respect scope in authorization-code login

### LINKED ISSUE #10108: CLI `medplum login` ignores `--scope`, so `offline_access` can't be requested for a refresh token

adityaapraveen:
## Summary

- pass the saved CLI profile into the authorization-code login flow
- use the configured `profile.scope` with `openid` as the default
- add regression tests for both the default and custom scope behavior

## Problem

The authorization-code flow used a hardcoded `scope=openid`, so the scope provided through the CLI was ignored. As a result, users could not request `offline_access` and receive a refresh token.

## Solution

The authorization-code login flow now reads the scope from the saved CLI profile:

```ts
profile.scope ?? 'openid'
```

Fixes #10108

### DIFF
--- a/packages/cli/src/auth.test.ts
+++ b/packages/cli/src/auth.test.ts
@@ -81,6 +81,10 @@ describe('CLI auth', () => {
     // Start the login
     await main(['node', 'index.js', 'login']);
 
+    const loginCommand = (cp.exec as unknown as Mock).mock.calls[0][0] as string;
+    const loginUrl = new URL(loginCommand.match(/https?:\/\/[^'"]+/)?.[0] as string);
+    expect(loginUrl.searchParams.get('scope')).toBe('openid');
+
     // Get the handler
     const handler = (http.createServer as unknown as Mock).mock.calls[0][0];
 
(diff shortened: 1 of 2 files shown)


adityaapraveen:
@AndyStoneman @mattwiller Please review whenever you get the time :)

codyebberson:
Hi @adityaapraveen  - this was fixed in https://github.com/medplum/medplum/pull/10151 - thanks for your contributiion 🙏

Issue #10003, opened 2026-07-28
[Scheduling] Find & Book Form


Filupnot:
Tracker for the full V1 Find & Book workflow: a scheduler picks a visit type and criteria, sees only times where every resource on the visit is free, and books.

PR #10164, opened 2026-08-10
[Scheduling] Actor selection field and proposed appointments hook

### LINKED ISSUE #10275: [Scheduling] Find & Book / MVP

Filupnot:
Part of #10003. Adds the field that chooses who an appointment is held on, the schedule search behind it, and the hook that turns those choices into times to offer.

## Where a provider's schedule lives

A provider holds both a `Practitioner` and a `PractitionerRole`, and this PR splits them the way scheduling reads them:

- The **schedule is held on the `Practitioner`**, so one human has one calendar. Booking on the role would let a provider holding two roles be double-booked against themselves.
- The **`PractitionerRole` carries specialty and site**, which decide whether that human is eligible for the service and location at all.

`BOOKABLE_ACTOR_TYPES` (new in `AppointmentFinder.roles`) encodes the first half; the provider branch of the location filter is the second.

## `AppointmentActorSelect`

image
image

### Use case

The user has entered a service and (optionally) a location. From that they need to pick the actors (providers, rooms, and devices) the appointment is held on, so that the times searched for are the ones every one of them is free.

### Shape

One field per role rather than one field for all of them: `role` is a prop, so a form stacks a `provider` field over an optional `room` and `device`, and `isRoleRequired` decides which of them have to be filled. A role the service has nothing configured for still renders, empty, so the form's shape does not change with the data behind it.

Built on `AsyncAutocomplete`, so schedules are searched as the name is typed rather than loaded up front. Everything picked attends: a second pick adds an actor and narrows the times, it does not replace the first.

`onChange` hands back `ScheduleCandidate[]`, the schedule with its actor, rather than a name or an id. The schedule is what `$find` is asked for, so the caller passes them straight to `getActorCombinations`.

## `searchScheduleCandidates`

What the field runs on each keystroke, exported on its own so a host application can drive the flow without the field.

### Logic

1. Search `Schedule` per service type code, narrowed to the role's actors by name.
   - Compile a list of `service-type` tokens from `HealthcareService.type` and for each run `Schedule?service-type=<service-type-tokens[n]>&active:not=false&_count=25&_include=Schedule:actor`, plus the role's chained actor filters
   - The chained filters are what makes this a search rather than a load: `actor:Practitioner.name`, `actor:Location.name` or `actor:Device.device-name` for what was typed, and `actor:<Type>.active:not` / `status:not` so an actor no longer in service is never offered. `:not` rather than `=true` / `=active` keeps the ones that never set the flag, which is most of them.
   - `active:not=false` on the Schedule for the same reason: `_count` is the budget for the whole page, so an inactive schedule that only gets dropped after it arrives has already crowded out a bookable one
   - A service carrying no type codings falls back to searching `Schedule` unfiltered by type
   - On the client, filter results by matching `Schedule.serviceType.extension(service-type-reference)` against the given `HealthcareService`, so two services sharing a type code don't bleed into each other
   - Keep only active schedules with exactly one linked actor, of type `Practitioner`, `Location`, or `Device`
     - Notably NOT including `PractitionerRole`, per the modelling above
2. Filter by `Location` if one was given (`filterCandidatesByLocation`, also exported).
   - Booking is supported across the location's `partOf` ancestry, walked up to 4 levels
   - Each role resolves its site differently:
     - Provider: the locations on their active `PractitionerRole`s, read back in one search covering every candidate rather than one search each. Containment counts in either direction, so a role covering a whole region covers the sites within it.
     - Room: the `Location` itself.
     - Device: `Device.location`.
   - Candidates are checked concurrently, and repeated `Location` reads are left to the `MedplumClient` request cache. The ancestry walk is up to 4 sequential reads that only the provider branch ever looks at, so it runs lazily: a service booked against rooms and devices never pays for it.
   - An actor with no recorded location, unreadable ancestry, or a chain deeper than 4 levels stays on offer; hiding something bookable is worse than offering something at the wrong site
3. Return the surviving candidates, each holding its schedule plus the actor resource when `_include` returned it, sorted by display name, which is the order a field offers them in.

## `AppointmentFinder.schedules`

The pure helpers that carry the user's choices from the fields to the hook, exported alongside them so a host application can drive the flow without the fields.

- `ActorSelections`: what's chosen, as candidates per role. Candidates rather than references, since a schedule is the thing being booked and the actor is only how it is named on screen.
- `getSelectedCandidates(selections)`: everything chosen, across roles, in the order the roles are asked about.
- `getSelectionError(selections)`: why the search can't run yet, as a message to show the user. Today that is only a required role left empty, meaning no provider chosen. Rooms and devices stay optional: a service may have room or device schedules without every booking needing to hold one.
- `getActorCombinations(selections)`: turns the selections into the sets of actors to search for. **Today this returns at most one combination**, holding every chosen actor, since everything chosen attends and `$find` intersects them in a single request. `useProposedAppointments` takes a list so alternatives ("any of these three surgeons") don't need an API change.
- `getCandidateDisplay` and `getCandidateRole`: read a candidate's name and role back off the schedule it was found through, so the candidate stores neither.

## `useProposedAppointments`

(The hook #10003 lists as `useAppointmentFind`.)

### Use case

The user has a service, a set of actors (from `AppointmentActorSelect` and `getActorCombinations`), and a date range. From that we need the actual times an appointment could be held at, so we can display them to the user and they can pick one.

### Logic

1. Build one `Appointment/$find` request per actor combination: `?start=&end=&service-type-reference=<service>&schedule=<a>&schedule=<b>&_count=20`
   - `$find` intersects schedules when multiple are provided in one request, so times that come back represent when *every* actor in the set is free
   - Requires a service and both ends of the range, since `$find` refuses an open range. Anything missing means no requests at all.
   - `_count` defaults to 20 (`$find`'s own default) **per combination**, and the hook doesn't paginate, so the ceiling is 20 x combinations
2. Fire them all concurrently and tolerate partial failure (`Promise.allSettled`).
   - Every combination that succeeded contributes its times
   - `error` is set **only when every combination failed**. One set of actors erroring loses that alternative, not the whole search.
3. Union the results into one list, in the order the combinations were asked.
   - `$find` answers each request with distinct times, ascending, so nothing here re-orders or de-duplicates what comes back. Order and grouping stay the caller's, usually `groupAppointmentsByDay`, which groups by day and by actor set and sorts within each group.
4. Re-search only when the request set actually changes.
   - The joined request urls are both the effect dependency and the stamp on the stored result: identical actors/range/count don't refetch, and a response that lands after the actors moved on is ignored
   - Times already found stay on screen while a narrower search runs, and are dropped only when there's nobody left to search for

### Returns

- `appointments`: every time offered — each combination's own times ascending, the combinations in the order they were given
- `requestCount`: how many `$find` requests the current combinations take
- `loading` / `error` (error suppressed while a result is stale)

## Not in this PR

- **No form.** Nothing stacks the three fields, holds the `ActorSelections` they produce, or wires them to `useProposedAppointments`, so nothing in-repo consumes `getActorCombinations` or `getSelectionError` yet.
- **No alternatives.** Everything picked for a role attends; there is no way to say "any one of these three surgeons". `getActorCombinations` returns a list and the hook searches one request per entry, so adding it later is a change to the field, not to the hook. The union's de-duplication belongs to that work too: two alternatives can name the same actor through two different Schedules, and so offer the identical time twice.
- **No searching by what an actor does.** A provider is found by name, not by the specialty their `PractitionerRole` carries.
- **No pagination or window clamping.** `$find` rejects a range wider than `MAX_FIND_WINDOW_DAYS` (31); keeping the range inside it stays the caller's responsibility, as it already is for `AppointmentDayTimes`. The schedule search takes 25 per service type code and doesn't page either.
- **Nothing writes.** Per #10003, these only read.

## Also

- `@medplum/react` and `@medplum/react-hooks` are now peer dependencies of the package. The field is an `AsyncAutocomplete` and takes its client from `MedplumProvider`.
- `test.setup.ts` now indexes the FHIR search parameter bundles package-wide. `MockClient` evaluates search parameters against in-memory resources, so an unindexed parameter matches nothing *silently*: `Schedule?service-type=` returned an empty bundle rather than an error, making every service look like it had no schedules configured.
- `test-utils/chainedActorSearch.ts` answers the chained `actor:<Type>.<param>` filters, which `MemoryRepository` does not implement and would otherwise report as an empty bundle for the same reason. It lifts them off the query, lets the repository answer the rest, and applies them to what came back; it understands only the filters this search sends, not chained search in general. The stories use it too, so the field is driveable in Storybook.
- The shared test renderer sets Mantine `env="test"`, so an open dropdown is not `display:none` in jsdom, and gains `renderWithMedplum` for components that search the server.
- Surgical story fixtures remodelled per the section above. Their providers previously held schedules on `PractitionerRole` and now carry both a `Practitioner` and a `PractitionerRole`.
- `AppointmentOptionRow`, the name-over-detail row an appointment pick list is built from.
- 4 new export lines in `src/index.ts`: the field, the option row, the `schedules` module, and the hook.
- ~1,170 lines of tests across three files, `MockClient` driven.

## Docs

`defining-availability.md` told readers to hold availability on a `PractitionerRole` when it varies by role or site — the pattern this PR declines to book against. The guidance now points at the `Practitioner`, and gives the supported way to vary a provider's availability by location: one `HealthcareService` per location, each carrying that site's hours on `availableTime`, with a per-service override on the Schedule where a provider differs from the site default.

The leftover `PractitionerRole` mentions in the timezone fallback list and the Example 2 diagram go with it.

| Before | After |
| --- | --- |
| image | image |
| image | image |


packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,664 @@
...
+
+/** How far up a `partOf` chain of Locations to look. */
+const MAX_LOCATION_DEPTH = 4;
+
+export interface FilterRoomsOptions {
+  readonly signal?: AbortSignal;
+}
+
+/**
+ * Narrows candidates to the ones available at one clinic.
```
noahsilas: I think the usage of the word "clinic" here is probably too narrow - can we keep this more generic like "location" so that folks at other practice types don't misconstrue the usage?

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,664 @@
...
+  readonly role: SchedulingRole;
+  /** A human-readable name for the actor, for use in plain-text option lists. */
+  readonly actorDisplay: string;
+  /**
+   * Codings saying what kind of thing the actor is, for narrowing a long list of
+   * them down. Empty for actors that carry nothing to narrow by.
+   */
+  readonly qualifiers: readonly Coding[];
+  /** The actor itself, when the search was able to include it. */
+  readonly actorResource: WithId<Resource> | undefined;
```
noahsilas, line 31: It seems like there's a lot of data here that is effectively extracted from `schedule` & `actorResource` - can we make this more minimal by just holding those two attributes in the state and deriving the rest on demand?

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.tsx
```diff
@@ -0,0 +1,169 @@
...
+  readonly onChange: (requirements: ActorRequirement[]) => void;
+  /**
+   * Whether one row may name several actors, any one of which would do.
+   *
+   * Off, a row is a single actor and a second pick replaces the first, which is
+   * the only shape the form offers today. On, a row becomes a choice — the state
+   * it produces is the same either way, and `getActorCombinations` already
+   * searches every way of resolving it.
+   */
+  readonly allowAlternatives?: boolean;
```
noahsilas: In terms of trying to get the first edition merged fast, can we drop the "OR" components from this PR if we are not using them yet? They seem to add a lot of complexity that seems like we could address it when we are ready to introduce into the flow.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,664 @@
...
+      } else if (resource.resourceType === 'Schedule') {
+        schedules.set(resource.id, resource);
+      }
+    }
+  }
+
+  return [...schedules.values()]
+    .map((schedule) => toScheduleCandidate(schedule, service, actorsByReference))
+    .filter(isDefined)
+    .sort((left, right) => left.actorDisplay.localeCompare(right.actorDisplay));
```
noahsilas: I'd encourage omitting `sort` here and pushing it closer to the usage in the final UI component. That makes it more obvious that this sort is a display concern, and it can potentially operate on a smaller set of values if these are filtered down by other checks.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,664 @@
...
+  };
+}
+
+/**
+ * Reads the codings that say what kind of thing an actor is.
+ *
+ * A PractitionerRole states the role it holds and the specialties it practises,
+ * and rooms and devices state their type. A plain Practitioner states none of
+ * this in R4 — a surgeon and an anesthesiologist are both just practitioners —
+ * so actors modelled that way cannot be told apart here.
```
noahsilas: There is `Practitioner.qualification` which might be something we would want to filter on.

We should probably get together and discuss how PractitionerRole/Practitioner play nicely together in scheduling. My suspicion is that the schedulable resource is really `Practitioner` - Schedule does allow setting a `PractitionerRole` as its actor, but in a world where one Practitioner has multiple PractitionerRole entries, you don't want the single human to get multi-booked for each of their Role calendars.

In that world, we may want to query PractitionerRole to get information about which Practitioners should be selectable, but ultimately submit Schedule resources linked to the Practitioners for $find/$book?

noahsilas (commented):
Partial review; high level question is how much can we simplify to get a more minimal first edition out that we can then layer a bunch of this feature set on to later? Particularly some of the work around AND/OR, and searching by qualifications seem like good candidates to push off to future work.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,664 @@
...
+  }
+
+  const resolved = new Map<string, Location | undefined>();
+  // The sites the clinic sits inside, so a role licensed for a whole state
+  // covers the clinics within it.
+  const ancestry = await getLocationAncestry(medplum, clinicReference, resolved, options);
+  const keep: ScheduleCandidate[] = [];
+
+  for (const candidate of candidates) {
+    if (await isCandidateAtClinic(medplum, candidate, clinicReference, ancestry, resolved, options)) {
```
noahsilas: This is a bit worrisome - I think this could iterate over a list of locations and make up to MAX_LOCATION_DEPTH requests for each one? Doing that in serial this way seems like it could become troublesome.

Probably OK to ship for now (I think the folks we are currently working with have a very limited location hierarchy), but it might be worth considering how we would implement this more performantly later. We may want to do something like an expanding frontier style query where we can process all the resources as a batch. Since this is all an internal optimization though, I'd say not a blocker.

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * One role's field, holding its own selection so choosing and removing work.
+ * @param props - The React props.
+ * @param props.group - The role and the actors that can fill it.
+ * @param props.initial - Requirements chosen before the story opens.
+ * @param props.allowAlternatives - Whether a row may hold a choice of actors.
+ * @param props.error - Why the selection is not yet valid.
+ * @returns The field.
+ */
+function Field(props: {
```
techdavidy, line 49: when I am testing I get this code. 

Image

I would expect `AppointmentActorSelect` code

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,664 @@
...
+export async function searchEligibleSchedules(
+  medplum: MedplumClient,
+  service: WithId<HealthcareService>,
+  options?: SearchEligibleSchedulesOptions
+): Promise<ScheduleCandidate[]> {
+  const count = (options?.count ?? 100).toString();
+  const tokens = getServiceTypeTokens(service);
+  const searches = tokens.length > 0 ? tokens.map((token) => ({ 'service-type': token })) : [{}];
+
+  const bundles = await Promise.all(
```
techdavidy: I think we should use Promise.allSettled to keep searching if one search fails.

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
```diff
@@ -0,0 +1,106 @@
...
+ * The groups are what a booking form asks about: a service held on providers
+ * and rooms produces a provider question and a room question, discovered from
+ * the data rather than configured per service.
+ *
+ * @param service - The service being booked, or undefined to load nothing.
+ * @param clinic - The site being booked at. Actors sited elsewhere are left out.
+ * @returns The candidate schedules and their groups, plus load and error state.
+ */
+export function useEligibleSchedules(
+  service: WithId<HealthcareService> | undefined,
```
techdavidy: I think we can do service: Reference<HealthcareService> | WithId<HealthcareService> to keep consistency.

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
```diff
@@ -0,0 +1,106 @@
...
+  clinic?: Reference<Location> | WithId<Location>
```
noahsilas: I have a suspicion that this flow is going to need to be inverted later to support large sets of schedules.

As I understand it, the intent of this PR is that folks would `useEligibleSchedules` to get a list of candidates, and then the AppointmentActorSelect gets those candidates as a prop. If there are a lot of schedules, I think we probably would want the filters to use something like an `<AsyncAutocomplete>` to let the viewer start typing an entry that might not have been included in the maximum of 100 calendars returned from this hook.

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+ * passed in, so a story cannot describe a room as a provider.
+ *
+ * @param schedule - The fixture Schedule the actor is booked through.
+ * @param display - The actor's name, as the Schedule records it.
+ * @param qualifiers - Codings saying what kind of thing the actor is.
+ * @returns The candidate.
+ */
+function candidate(schedule: WithId<Schedule>, display: string, qualifiers: Coding[] = []): ScheduleCandidate {
+  const actor = schedule.actor[0];
+  const actorType = actor.reference?.split('/')[0];
```
techdavidy: I think we can do
const actorType = parseReference(actor)[0]

same with other splits

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
```diff
@@ -0,0 +1,106 @@
...
+          setLoaded({ key: asked, candidates: result.kept, excludedByClinic: result.excluded, error: undefined });
+        }
+      })
+      .catch((reason: unknown) => {
+        if (!controller.signal.aborted) {
+          setLoaded({
+            key: asked,
+            candidates: [],
+            excludedByClinic: 0,
+            error: new Error(normalizeErrorString(reason), { cause: reason }),
```
techdavidy: This seems to be round-trip that loses information. medplum.search/medplum.get reject with OperationOutcomeError, which is already an Error and carries .outcome. Flattening it to .message and re-wrapping in a plain Error throws away the outcome and the original stack.

packages/react-scheduling/src/AppointmentFinder/useProposedAppointments.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+interface FindFailure {
+  readonly reason: unknown;
+}
+
+function isFailure(result: FindResult): result is FindFailure {
+  return 'reason' in result;
+}
+
+async function request(medplum: MedplumClient, url: string, signal: AbortSignal): Promise<FindResult> {
```
techdavidy: I think we can remove this function and call await medplum.get directly

packages/react-scheduling/src/AppointmentFinder/useProposedAppointments.ts
```diff
@@ -0,0 +1,180 @@
...
+/** One combination's result: its times, or why it has none. */
+type FindResult = FindSuccess | FindFailure;
+interface FindSuccess {
+  readonly bundle: Bundle<Appointment>;
+}
+interface FindFailure {
+  readonly reason: unknown;
+}
+
+function isFailure(result: FindResult): result is FindFailure {
```
techdavidy: could we use isError(value: unknown) ?

noahsilas (commented):
Strategy concern: I think that this interface is going to work well for the smaller use cases we're currently thinking about, but is going to eventually be troublesome for large installations with > 100 schedules.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,465 @@
...
+    searches.map(async (criteria) =>
+      medplum.search(
+        'Schedule',
+        { ...criteria, _count: count, _include: 'Schedule:actor' },
```
noahsilas: It looks like below we filter out all candidates that have `active: false`, maybe we should add that to the criteria here? I think `"active:not": false` would include `active: true` and resources where `active` is missing.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,465 @@
...
+  candidates: readonly ScheduleCandidate[],
+  location: Reference<Location> | WithId<Location> | undefined,
+  options?: FilterCandidatesOptions
+): Promise<ScheduleCandidate[]> {
+  const locationReference = location && getReferenceString(location);
+  if (!locationReference) {
+    return [...candidates];
+  }
+
+  const resolved = new Map<string, Location | undefined>();
```
noahsilas: The MedplumClient has a built in resource cache to avoid querying the same object too frequently. It would be nice to rely on that instead of implementing a custom cache this way.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,465 @@
...
+}
+
+/**
+ * Walks up from a Location to the one being booked at.
+ * @param medplum - The Medplum client.
+ * @param reference - The Location to start from, or undefined to say nothing.
+ * @param locationReference - The Location being looked for.
+ * @param resolved - Locations already read, to save reading them again.
+ * @param options - Abort signal.
+ * @param included - The starting Location, when a search already returned it.
```
noahsilas: It looks like this `included` parameter is here as an optimization, but I think it's a bit confusing (especially at the tail end of 6 positional parameters). If the related resource was already fetched, it should be in `resolved` (and the MedplumClient resource cache) - I think we can simplify this function without losing any real performance by removing this parameter.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,465 @@
...
+ * @returns Whether the target is at or above the Location, or unknowable.
```
noahsilas, line 385: Should we return a richer value than a boolean? We may want to be able to exclude "unknowable" from some UI later, which suggests to me that we could consider `boolean | undefined`.  That does complicate the usage slightly for the case you care about, from `isWithinLocation(...)` to `isWithinLocation(...) !== false`.

Since we're not exporting this function directly and it's just a helper to support our current usage, I think you should ignore this comment, and we can refactor cleanly if/when we do need such behavior.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,465 @@
...
+): Promise<ScheduleCandidate[]> {
+  const locationReference = location && getReferenceString(location);
+  if (!locationReference) {
+    return [...candidates];
+  }
+
+  const resolved = new Map<string, Location | undefined>();
+  // The sites the location sits inside, so a role licensed for a whole state
+  // covers the sites within it.
+  const ancestry = await getLocationAncestry(medplum, locationReference, resolved, options);
```
noahsilas: I wonder if we can simplify this by utilizing other areas of the data model. For instance, could we test `HealthcareService.coverageArea` for an exact match against `PractitionerRole.location` instead of walking this ancestry chain?

EDIT: I think I was considering approaching this from the perspective of narrowing the results after choosing a healthcare service, but I think that the perspective here is more that the viewer has chosen a Location and we are filtering possible schedulable resources based on that choice? And this code is buying us the ability to understand where the selection sits in the hierarchy and make choices based on that?

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,465 @@
...
+  }
+
+  if (actor?.resourceType === 'Device') {
+    return isWithinLocation(medplum, actor.location?.reference, locationReference, resolved, options);
+  }
+  if (actor?.resourceType === 'PractitionerRole' && actor.location?.length) {
+    for (const roleLocation of actor.location) {
+      // A role's location counts whether the chosen location sits inside it or
+      // it sits inside the chosen location.
+      if (roleLocation.reference && ancestry.has(roleLocation.reference)) {
```
noahsilas: It looks like this in the only place that we read the ancestry chain. I still think that using `PractitionerRole` as a schedulable resource is likely to be uncommon (`Practitioner` feels like the more natural schedulable resource, and it is what Provider app currently focuses on), so I wonder if there is a way to make this ancestry lookup lazy without introducing a ton of extra complexity/mechanics?

Maybe it won't be a problem in practice if folks are mostly scheduling against resources that don't have an ancestry chain, in which case that read path short circuits entirely?

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
Filupnot: Replaced with this to preserve original OperationOutcomeError:
```ts
error: isError(reason) ? reason : new Error(normalizeErrorString(reason), { cause: reason }),
```

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: I'd lean towards keeping `Promise.all` here, because there's one question being asked ("which schedules can book this service?"), and if one search fails and the other succeeds then we'd return a partial answer without any indication that it's partial. It feels better here to just fail upfront.

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.tsx
Filupnot: Ended up dropping `AppointmentActorSelect` altogether

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: Omitted. Also moved to `groupCandidatesByRole`

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.stories.tsx
Filupnot: Hmm, that's definitely strange, good catch. 
I ended up getting rid of the components & storybooks.

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
Filupnot: Updated. Also did the same with `useProposedAppointments`.

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.stories.tsx
Filupnot: Got rid of this component

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: I dug into `coverageArea` a bit. From my understanding, this represents the locations a service is available _to_, rather than where it's performed _at_. So I'm hesitant to use it in this context. 

Your second point is right. A location has a whole hierarchy behind it, and we'd want to include any services & actors linked to those sub-locations or parent locations.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: Updated to initialize lazily instead

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: We're planning to meet to talk about this more, but for now here's what I did: 

I excluded `PractitionerRole` as a bookable actor type by dropping them in `toScheduleCandidate`, so a provider is only ever offered on their `Practitioner`. We're still looking at roles when filtering providers on location. This logic is all being done in the finder UI, and `$find`/`$book` are unaffected and still resource agnostic. 

Didn't touch `Practitioner.qualification` for now, since I feel like that might be better fit in a future PR.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: Made some changes here. We're checking ancestry lazily and concurrently (using `Promise.all`), and the client cache should be caching the results for a given location so that we don't search on one more than once.

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
Filupnot: I agree about inverting this flow later. I've dropped `AppointmentActorSelect` from this PR, so nothing consumes the list yet - I'll look at `<AsyncAutocomplete>` with a server-side search when I build the field, and change the hook then.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: Update: providers are now booked on the Practitioner and sited via PractitionerRole.location, so the ancestry walk is reached through practitioners rather than role-held schedules. Still lazy, but no longer the rare path.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,510 @@
...
+ * @returns Every schedule bookable for the service, each with its actor.
+ */
+export async function searchEligibleSchedules(
+  medplum: MedplumClient,
+  service: WithId<HealthcareService>,
+  options?: SearchEligibleSchedulesOptions
+): Promise<ScheduleCandidate[]> {
+  const count = (options?.count ?? 100).toString();
+  const tokens = getServiceTypeTokens(service);
+  const searches = tokens.length > 0 ? tokens.map((token) => ({ 'service-type': token })) : [{}];
```
noahsilas: I think that you can merge these into a single comma-separated search. That has disjunctive ("OR") semantics, and so should find Schedule resources having at least one of these tokens without running multiple searches (which we also would expect to usually all return copies of the same matches anyways, since all the service type tokens should be copied over when a service type is linked).

```ts
const criteria = tokens.length > 0 ? ({ 'service-type': tokens.join(',') }) : ({});
const schedules = await medplum.search(
  'Schedule',
  { ...criteria, 'active:not': 'false', _count: count, _include: 'Schedule:actor' },
  { signal: options?.signal }
);
```

packages/react-scheduling/src/AppointmentFinder/useEligibleSchedules.ts
Filupnot: Updated again, after we spoke yesterday. I inverted the logic so that now we have `<AppointmentActorSelect />` based on `<AsyncAutocomplete />`, which will query for schedules on user input, rather than upfront. Now there shouldn't be risk of hitting that query limit here.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,490 @@
...
+): Promise<ScheduleCandidate[]> {
+  const locationReference = location && getReferenceString(location);
+  if (!locationReference) {
+    return [...candidates];
+  }
+
+  // The location where a Practitioner practices is recorded on their PractitionerRoles.
+  let roles: Promise<ReadonlyMap<string, readonly PractitionerRole[]>> | undefined;
+  const getRoles = async (): Promise<ReadonlyMap<string, readonly PractitionerRole[]>> =>
+    (roles ??= searchRolesByPractitioner(medplum, candidates, options));
```
noahsilas: It might be nice to use core's `lazy` helper to make the intention here a little clearer. It also should facilitate inferring all the type information which makes it a little less busy:

```suggestion
  const getRoles = lazy(() => searchRolesByPractitioner(medplum, candidates, options));
```

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,490 @@
...
+
+  // The sites the location sits inside, so a role licensed for a whole state
+  // covers the sites within it (lazy init).
+  let ancestry: Promise<ReadonlySet<string>> | undefined;
+  const getAncestry = async (): Promise<ReadonlySet<string>> =>
+    (ancestry ??= getLocationAncestry(medplum, locationReference, options));
```
noahsilas: If you apply `lazy` above, same idea goes here too

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,490 @@
...
+
+  if (getCandidateRole(candidate) === 'room') {
+    return isWithinLocation(medplum, actorReference, locationReference, options);
+  }
+
+  if (actor?.resourceType === 'Device') {
+    return isWithinLocation(medplum, actor.location?.reference, locationReference, options);
+  }
+
+  if (actorReference && getActorType(actorReference) === 'Practitioner') {
```
noahsilas: As a stylistic note, it would be nice to have all of these conditions switching on the same type of criteria. One thought is that you could do something like an exhaustive switch on `getCandidateRole`:

```tsx
  const role = getCandidateRole(candidate);
  if (role === 'room') { ... }
  else if (role === 'device') { ... }
  else if (role === 'provider') { ... }
  else {
    return assertNever(role);
  }
```

This has the nice property that if we ever add an additional "role" enum member, the type checker can flag that this location needs to be updated too.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -0,0 +1,490 @@
...
+      return true;
+    }
+    const ancestry = await getAncestry();
+    for (const roleLocation of practiceLocations) {
+      // A role's location counts whether the chosen location sits inside it or
+      // it sits inside the chosen location.
+      if (roleLocation.reference && ancestry.has(roleLocation.reference)) {
+        return true;
+      }
+      if (await isWithinLocation(medplum, roleLocation.reference, locationReference, options)) {
```
noahsilas: I'm still a bit concerned about the performance implications of this - we're inside a few nested loops, and introducing another potential several serially blocking queries.

What would you think about documenting that it must be an exact match today, and leaving off "contained in" querying for a future enhancement where we can be a little more targeted+complex in how we do these queries?

packages/react-scheduling/src/test-utils/chainedActorSearch.ts
```diff
@@ -0,0 +1,94 @@
...
+ * parameter table, so `actor:Practitioner.name` matches nothing — and returns an
+ * empty bundle rather than an error, which would make every test below pass for
+ * the wrong reason. The stub lifts the chained filters off the query, lets the
+ * repository answer the rest, and applies them to what came back, so these tests
+ * still run against the fixtures. It understands only the two filters the search
+ * sends, not chained search in general.
+ *
+ * @param medplum - The client to stub.
+ */
+export function stubChainedActorSearch(medplum: MockClient): void {
```
noahsilas, line 23: This seems like a reasonable step to get this PR moving, but can we file a follow-up ticket to handle this more generally later? Seems like we should push a more complete chained search implementation into MemoryRepository so that we can support this style of query without such trickery.

packages/react-scheduling/src/AppointmentFinder/useProposedAppointments.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+  url.searchParams.set('_count', count.toString());
+  return url.toString();
+}
+
+/**
+ * Unions what the combinations found, by ascending start.
+ * @param results - What each combination returned.
+ * @returns The times offered, deduplicated and sorted.
```
noahsilas: I think that as long as the combinations are unique, the results should also be unique by the appointment key. Is this guarding against some kind of tricky edge case I'm not thinking of?

I wonder if we should defer the sorting to the consumer; if they want to sort by some other dimension than start date then this would be wasted work. (In our designs I think we often end up partitioning these by actors and sorting within those groups, although there's also the calendar widget which suggests grouping/sorting on a slightly different criteria (just the date component of the timestamp).

noahsilas (approved):
Thanks for going through all the rounds of review on this one - there's a lot happening in here but I think we have the right initial shape now. I suspect we're going to have to come back to this for some query optimization for the location/device chains later, but I think getting this first version out is a great step.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
Filupnot: Removed role ancestry lookup in favor of direct location matching. Agreed this was the biggest fan-out concern, and realistically our first consumer shouldn't have build like this anyways.

packages/react-scheduling/src/test-utils/chainedActorSearch.ts
Filupnot: Created #10258

packages/react-scheduling/src/AppointmentFinder/useProposedAppointments.ts
Filupnot: Artifact from where there was OR logic. Removed

PR #10175, opened 2026-08-10
feat(server): configurable default app name for user-facing content

mokantari:
## Problem

White-labeling reads the `appName` setting from the user's project. Content addressed to a user who has no project has nothing to read:

- Practitioner invites create **server-scoped** users by default (`User.project` is only set for Patients, externalId users, or an explicit `scope: 'project'`).
- `auth/resetpassword` looks the user up and passes `user.project`, which is undefined for exactly those users.

Those messages then name Medplum however the deployment is branded, and a self-hosted operator has no configuration that changes it. The From header is already configurable via `supportEmail`, so today a recipient can get an email from "Acme Health <no-reply@acme.com>" whose body says Medplum.

## Change

Adds an optional server config setting, `appName`, used when the project does not name one. Precedence: project setting, then server config, then the existing `'Medplum'` default, so behavior is unchanged for anyone who leaves it unset.

`getAppName(project)` in `email/utils` holds that precedence in one place; the MFA email switches to it. The authenticator issuer default (`DEFAULT_MFA_ISSUER`) is deliberately untouched, since it is a domain rather than a brand name.

## Notes

- Blank config values are treated as unset, matching how `getProjectAppName` treats a blank project setting.
- Unit tests cover all four cases: project wins over config, config used when the project has none, config used when there is no project at all, and the default when neither names one.
- Companion to #10069, which applies project white-labeling to the invite and password reset emails; that PR's call sites can move to `getAppName` once either lands. The two are independent and can merge in either order.

### DIFF
--- a/packages/server/src/auth/utils.ts
+++ b/packages/server/src/auth/utils.ts
@@ -26,7 +26,7 @@ import { toDataURL } from 'qrcode';
 import { getConfig } from '../config/loader';
 import { EMAIL_MFA_CODE_EXPIRATION_MS, MAX_PASSWORD_LENGTH } from '../constants';
 import { sendEmail } from '../email/email';
-import { getProjectAppName } from '../email/utils';
+import { getAppName, getProjectAppName } from '../email/utils';
 import { sendOutcome } from '../fhir/outcomes';
 import type { SystemRepository } from '../fhir/repo';
 import { getGlobalSystemRepo, getShardSystemRepo } from '../fhir/repo';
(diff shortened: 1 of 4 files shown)


techdavidy:
@claude review

mokantari:
Per the policy bot's guidance: opened medplum/medplum#10338 describing this work (together with #10069), and rebased this branch onto v5.1.33. Ready to link and reopen once the issue carries the open-to-community label.

PR #10176, opened 2026-08-11
feat(core): reconnect the FHIRcast client's WebSocket

### LINKED ISSUE #10134: [FHIRcast] Use `ReconnectingWebSocket` in FHIRcast client
Currently we are not using the existing `ReconnectingWebSocket` in the FHIRcast client, which would greatly help with robustness of the FHIRcast connections

ThatOneBro:
## Summary

`FhircastConnection` opened a bare `WebSocket`, so any dropped connection — a hub restart, a proxy idle timeout, a laptop waking from sleep — ended the FHIRcast session silently, and callers had to resubscribe to recover. This swaps in the `ReconnectingWebSocket` already used by `SubscriptionManager`.

A subscription endpoint stays valid for the lease the hub granted it (`FHIRCAST_LEASE_SECONDS`, and the server only drops the endpoint on `hub.unsubscribe`, not on socket close), so reconnecting to the same endpoint resubscribes without a new `hub.subscribe` request.

Closes #10134

## Changes

- `FhircastConnection` uses `ReconnectingWebSocket` instead of a bare `WebSocket`.
- Listeners bind to the `ReconnectingWebSocket` rather than being registered from inside each `open` handler, which would have stacked a duplicate set of handlers per reconnect.
- A denied subscription still disconnects for good — reconnecting to an endpoint the hub refuses would only be refused again.
- `FhircastConnection` and `MedplumClient.fhircastConnect` accept an optional options bag forwarded to the socket, for backoff tuning and for supplying a `WebSocket` implementation outside the browser.
- The FHIRcast demo's `ConnectionHandler` no longer tears down its listeners on `disconnect`, so a reconnect flips the status back to `CONNECTED`.

## Behavior notes for reviewers

- `connect` and `disconnect` can now each fire more than once over the life of a connection. `disconnect` followed by `connect` is the normal shape of a dropped connection; only `connection.disconnect()` and a hub denial end it for good.
- A connection that never opens now dispatches `disconnect` (the `close` listener is attached upfront rather than from within `open`).

## Test plan

- New tests in `packages/core/src/fhircast/index.test.ts`: the connection recovers after the hub drops the socket and delivers a subsequent message exactly once; a denial does not reconnect.
- `npx vitest run src/fhircast src/client-fhircast.test.ts src/subscriptions/index.test.ts src/websockets` in `packages/core` — passing.
- `tsc --noEmit`, `eslint`, `prettier --check` on `packages/core` and the FHIRcast demo — clean.

### DIFF
--- a/examples/medplum-fhircast-demo/src/components/ConnectionHandler.tsx
+++ b/examples/medplum-fhircast-demo/src/components/ConnectionHandler.tsx
@@ -59,12 +59,9 @@ export default function ConnectionHandler(props: WebSocketHandlerProps): null {
 
     const connectHandler = (): void => setConnectionStatus('CONNECTED');
     const messageHandler = (event: FhircastMessageEvent): void => onMessageRef.current(event.payload);
-    const disconnectHandler = (): void => {
-      setConnectionStatus('DISCONNECTED');
-      connection.removeEventListener('connect', connectHandler);
-      connection.removeEventListener('message', messageHandler);
-      connection.removeEventListener('disconnect', disconnectHandler);
-    };
+    // The connection reconnects on its own, so a disconnect is a state to show rather than the end
+    // of the session: leave the listeners attached and let the next `connect` flip the status back.
+    const disconnectHandler = (): void => setConnectionStatus('DISCONNECTED');
 
     connection.addEventListener('connect', connectHandler);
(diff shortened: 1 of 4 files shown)


codyebberson:
@claude please review.  any breaking changes?

codyebberson (approved):

PR #10178, opened 2026-08-11
fix(fhircast): accept the legacy unsubscribe endpoint field

ThatOneBro:
## The break

#10096 (first released in `v5.1.29`) renamed the field the client sends when cancelling a subscription. In `serializeFhircastSubscriptionRequest`, `endpoint` became `hub.channel.endpoint`, and the Hub's `handleUnsubscribeRequest` now reads only the latter.

A subscriber on `@medplum/core` <= 5.1.28 unsubscribing from a 5.1.29+ server gets `400 Missing endpoint`. Nothing is cancelled: the lease stays in Redis until `FHIRCAST_LEASE_SECONDS` expires, the socket is never denied and never closed, and the subscriber keeps receiving the topic's events. The subscribe side never changed -- the response has always carried `hub.channel.endpoint` -- so an old client holds a valid endpoint and only names it differently on the way back.

## The fix

`handleUnsubscribeRequest` falls back to `req.body.endpoint` when `hub.channel.endpoint` names no endpoint. `extractEndpoint` already strips a query string or fragment and takes the last path segment, so the full `wss://.../ws/fhircast/<id>` URL an old client echoes back resolves the same way the current form does. Server-side only -- no client change, and nothing else reads the legacy name.

**Precedence:** `hub.channel.endpoint` wins. It is the name the Hub issues and the only one a current client sends, so the legacy field is consulted solely when the issued one is absent -- a body carrying both cannot redirect a cancellation away from the endpoint it names.

**Logging:** debug, not warn. The request is honored, so it is not an anomaly worth a warn on every legacy unsubscribe; the line exists so an operator can see whether any subscriber still depends on the old name before the shim is dropped.

`hub.events`, which old clients also send on an unsubscribe, is already harmless: its check sits past the unsubscribe branch in `handleSubscriptionRequest`. The first test below sends a full legacy-shaped body, so that stays pinned rather than being taken on faith.

Nothing under `packages/docs/docs/fhircast/` specifies the unsubscribe request shape, so there was nothing to update.

## Test plan

Added to `packages/server/src/fhircast/routes.test.ts`:

- **`Unsubscribe with a request from a pre-5.1.29 client`** -- a full legacy body (form-encoded, bare `endpoint`, `hub.events` repeated) returns 202, publishes the targeted denial to that subscriber, closes its socket, and deletes the lease, exactly as the `hub.channel.endpoint` form does. Fails with `400 Missing endpoint` without this change.
- **`Unsubscribe naming an endpoint in both fields cancels the issued one`** -- pins the precedence: the `hub.channel.endpoint` subscription is cancelled, the one named under `endpoint` is left intact.
- **`Unsubscribe with the legacy endpoint field still has to name a subscription the caller holds`** -- the legacy name from another project, and one naming an unknown endpoint, are both rejected with `Invalid endpoint`, and the subscription survives. The fallback is not a way around the ownership check.

A request naming neither field is already covered by `Unsubscribe without an endpoint is rejected`, which still passes.

`npx vitest run src/fhircast/routes.test.ts` -- 56 passed. `eslint`, `prettier --check`, and `tsc --noEmit` clean for both files.

### DIFF
--- a/packages/server/src/fhircast/routes.ts
+++ b/packages/server/src/fhircast/routes.ts
@@ -280,10 +280,21 @@ async function handleUnsubscribeRequest(req: Request, res: Response, topic: stri
   // Subscribers echo back the `hub.channel.endpoint` they were issued, which is what identifies the
   // subscription to cancel. Several subscribers can share a topic, so an unsubscribe the Hub cannot
   // address is not actionable: honoring it would mean denying every one of them.
-  const endpoint = extractEndpoint(req.body['hub.channel.endpoint']);
+  //
+  // `@medplum/core` before 5.1.29 sent it under a bare `endpoint` instead, so that name is honored
+  // when the issued one is absent. Such a client holds a valid endpoint and only names it
+  // differently; without this it could never cancel, and its subscription would sit until the lease
+  // ran out. The endpoint is still resolved to a subscription the caller owns either way.
+  const issuedEndpoint = extractEndpoint(req.body['hub.channel.endpoint']);
+  const endpoint = issuedEndpoint ?? extractEndpoint(req.body.endpoint);
   if (!endpoint) {
     sendOutcome(res, badRequest('Missing endpoint'));
     return;
+  }
+  if (!issuedEndpoint) {
+    // Debug rather than warn: the request is honored, and the line exists only to show whether any
+    // subscriber still depends on the legacy name before it is dropped.
+    getLogger().debug('[FHIRcast]: Unsubscribe request named its endpoint under the deprecated `endpoint` field');
   }
 
   const subscription = await getEndpointSubscription(endpoint);
--- a/packages/server/src/fhircast/routes.ts
+++ b/packages/server/src/fhircast/routes.ts
@@ -292,9 +292,9 @@ async function handleUnsubscribeRequest(req: Request, res: Response, topic: stri
     return;
   }
   if (!issuedEndpoint) {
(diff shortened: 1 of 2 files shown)


packages/server/src/fhircast/routes.ts
```diff
@@ -280,11 +280,22 @@ async function handleUnsubscribeRequest(req: Request, res: Response, topic: stri
...
+  const issuedEndpoint = extractEndpoint(req.body['hub.channel.endpoint']);
+  const endpoint = issuedEndpoint ?? extractEndpoint(req.body.endpoint);
   if (!endpoint) {
     sendOutcome(res, badRequest('Missing endpoint'));
     return;
   }
+  if (!issuedEndpoint) {
+    // Debug rather than warn: the request is honored, and the line exists only to show whether any
+    // subscriber still depends on the legacy name before it is dropped.
+    getLogger().debug('[FHIRcast]: Unsubscribe request named its endpoint under the deprecated `endpoint` field');
```
mattlong: Seems like this should be info/warn level instead of debug. In practice it'll never be seen on prod if it stays debug.

mattlong (approved):

packages/server/src/fhircast/routes.ts
ThatOneBro: Fair. Originally the idea was to prevent prod from being spammed with these logs but I guess since theoretically it shouldn't be super high frequency (like on every FHIRcast message request), it should be fine to make an info log.

PR #10182, opened 2026-08-11
feat(definitions): add MedicationDispense context search parameter

### LINKED ISSUE #10074: Add MedicationDispense.context as a Search Parameter
As title describes. This Search Parameter exists on R6 (as the replacement field), making this a reasonable addition to our search params. https://build.fhir.org/medicationdispense.html

adityaapraveen:
## Summary

- add `MedicationDispense-context` to the always-loaded Medplum search-parameter bundle
- define it as a reference search parameter targeting `Encounter` and `EpisodeOfCare`
- add regression coverage for the definition

## Background

The standard R4 definitions contain this search parameter, but standard parameters can be restricted through `enabledSearchParameters`. Adding the definition to the Medplum bundle ensures it is loaded consistently alongside other Medplum-supported search parameters.

The definition searches the existing R4 field:

```text
MedicationDispense.context
```

`Fixes #10074`

### DIFF
--- a/packages/definitions/src/fhir/r4/search-parameters-medplum.json
+++ b/packages/definitions/src/fhir/r4/search-parameters-medplum.json
@@ -1456,6 +1456,24 @@
       }
     },
     {
+      "fullUrl": "http://hl7.org/fhir/SearchParameter/MedicationDispense-context",
+      "resource": {
+        "resourceType": "SearchParameter",
+        "id": "MedicationDispense-context",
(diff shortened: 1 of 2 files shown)


adityaapraveen:
@ThatOneBro please review whenever you get the time :)

codyebberson:
Hi @adityaapraveen  - I believe the `MedicationDispense-context` search parameter is already in R4.  We're checking whether issue #10074 was actually correct

adityaapraveen:
okay

PR #10183, opened 2026-08-11
[Server] Scheduling: overbooking via per-slot capacity

### LINKED ISSUE #10110: [Scheduling] Overbooking support
In the current Scheduling operations like `$find` and `$book`, we operate on the assumption that a schedulable resource is either "available" or "not available", but some practitioners operate under a model where they can take several appointments of a given type at once.

We will add a new Scheduling Parameter, `slotCapacity`, that allows configuring how many appointments may be booked into the same time slot for a given service type. This should default to a value of `1`, leaving all current scheduling behavior unaffected. A HealthcareService or Schedule can set this parameter to a value like `2` to allow double booking, or `3` to allow triple booking.

We should allow different Schedule resources to enforce their own capacity requirements separately; if `practitioner-X` has a schedule that says their capacity for ServiceA is 2, and `location-y` has a schedule that says their capacity for ServiceA is 3, a request like `GET Appointment/$find?service-type-reference=HealthcareService/ServiceA&schedule=Schedule/practitioner-X&schedule=Schedule/location-Y` should find available time when Practitioner X has less than 2 appointments already booked, and Location Y has less than 3 appointments already booked.

## Limitations:
- Only appointments of the same serviceType can overbook each other; potential future enhancement allowing configuration of other service types that may overbook each other?

## Design questions:
- How does this interplay with `bufferBefore`/`bufferAfter` slots? Should such a slot be considered overbookable time, or is it a separate type that should stay fully clear? 
  - Example motivating not overbookable: if a Location has `bufferAfter` to allow cleaning after the appointment, it probably should not allow an active Appointment to overbook that buffer slot.
- Do overbookings need to align exactly? For example, consider the following service:
  > ```
  > {
  >   url: 'https://medplum.com/fhir/StructureDefinition/SchedulingParameters',
  >   extension: [
  >     { url: 'duration', valueDuration: { value: 60, unit: 'min' } },
  >     { url: 'alignmentInterval', valueDuration: { value: 30, unit: 'min' } } 
  >     { url: 'slotCapacity', valueCount: { value: 2 } } 
  >   ]
  > }
  > ```
  These parameters tell us that there are 60 minute appointments available starting on each half-hour, and that two can be booked concurrently. If there is one appointment scheduled from 09:30:00 – 10:30:00, should an appointment at 10:00:00 – 11:00:00 be bookable? 

## Implementation questions:
- How do we implement this performantly in `$find`, which may have to scan large time windows and many slots?

noahsilas:
Add a `slotCapacity` scheduling parameter (positiveInt, default 1) on the SchedulingParameters extension, representing how many other appointments may be concurrently scheduled.

Availability treats capacity as a per-booking tolerance: each booked Slot is stamped with the `SchedulingSlotCapacity` it was created under, and `applyExistingSlots` admits a new booking at an instant only while the bookings covering it stay under both the candidate's capacity and every overlapping booking's capacity (strictest wins).

A capacity-1 booking is therefore exclusive regardless of order; an unstamped slot reads back as capacity 1, so consumers that ignore the stamp fail closed. This also keeps behavior for already booked Slot resources that do not have this stamp unchanged.

Closes: #10110
Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


noahsilas:
@claude please review

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -313,6 +315,105 @@ export function removeAvailability(availableIntervals: Interval[], blockedInterv
...
+
+  const result: Interval[] = [];
+  let blockedStart: number | undefined;
+  let i = 0;
+  while (i < events.length) {
+    // Apply every event at this instant before evaluating the segment [time, next):
+    // a booking ending here leaves before one starting here is admitted, so
+    // adjacent bookings net out at a shared boundary.
+    const time = events[i].time;
+    while (i < events.length && events[i].time === time) {
```
Filupnot, line 405: This is definitely an edge case but felt worth mentioning: 

FHIR and Medplum support leap seconds (`2016-12-31T23:59:60Z`) but V8 does not. If this time existed in a slot, then `buildCapacityEvents` -> `new Date(slot.start).valueOf()` would result in `NaN`. Since `NaN === NaN -> False` this would result in an infinite loop here.

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -567,6 +667,7 @@ async function validateAvailability(
     slots: existingSlots,
     range: interval,
     serviceType: healthcareService.type,
+    capacity: parameters.get('slotCapacity'),
```
Filupnot: Since buffer slots don't have `slotCapacity` I don't think this will correctly account for buffer slots? Perhaps this can be fixed by using `getSlotCapacity()` here instead, which returns `1` in that scenario?

packages/server/src/fhir/operations/find.ts
```diff
@@ -150,6 +151,7 @@ async function handler(params: {
       slots: scheduleSlots,
       range: effectiveRange,
       serviceType: healthcareService.type,
+      capacity: schedulingParameters.get('slotCapacity'),
```
Filupnot, line 157: Same concern here as other issue around new buffer slots not having this parameter

packages/docs/docs/scheduling/defining-availability.md
```diff
@@ -262,6 +263,54 @@ Here is an example of a [Slot](/docs/api/fhir/resources/slot) resource that bloc
...
+
+- **A `slotCapacity: 1` booking is exclusive.** Nothing may overlap it — not even a higher-capacity service booked afterward. This holds regardless of order: whichever booking is placed first, an incompatible one is rejected.
+- **Overbooking only stacks among bookings that all permit it.** Two `slotCapacity: 2` bookings can share a time; a `slotCapacity: 2` and a `slotCapacity: 3` booking cap each other at 2 (the stricter of the two) while they overlap.
+
+Which slots count:
+
+- **`busy` (booked)** and **`busy-tentative` (held via [`$hold`](/docs/scheduling/appointment-hold))** Slots each consume one unit and carry a tolerance. A hold reserves its unit until it is confirmed, booked, or released.
+- **`busy-unavailable`** Slots (explicit blocks and buffer reservations) are hard blocks — they remove the time regardless of any capacity.
```
Filupnot: I think that based on the way that `getSlotCapacity` is written, if a `busy-unavailable` slot contains a slot capacity (which is technically possible through the FHIR Slot endpoint), this is not always true. Probably a rare edge-case.

packages/examples/src/scheduling/defining-availability.ts
```diff
@@ -385,6 +385,53 @@ const scheduleOverride: Schedule =
...
+            // required: identifies which HealthcareService these parameters override
+            url: 'service',
+            valueReference: {
+              reference: 'HealthcareService/f44bbf25-bf57-4263-8f10-be060cc91672',
+              display: 'Group Visit',
+            },
+          },
+          {
+            // Up to 8 appointments may occupy the same time. The default is 1 (no
+            // overbooking). Cannot be combined with bufferBefore / bufferAfter.
```
Filupnot: > Cannot be combined with bufferBefore / bufferAfter

From what I can tell, nothing enforces this. If this is true should we add some validation around this?

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: Oof, wild that FHIR supports leap seconds but doesn't allow `24:00:00` to represent end of day.

You're right that this is an edge case, but it's a great catch that it could turn into an infinite loop and cause denial of service. I'll add some validation.

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: The intention here is to restrict the capacity based on the type of service we are trying to book. So if we are booking "Office Visit (slotCapacity=3)`", we can find any times that currently have availability for up to a third booking.

`getSlotCapacity(slot)` tells us how a single slot can interact with that, but this call needs to operate at the higher level.

packages/examples/src/scheduling/defining-availability.ts
noahsilas: I think this comment should probably be improved. It's not that it "can't" be combined, but that the behavior gets funky when you do.

In particular, if you have a service with `{ slotCapacity: 3, bufferAfter: 15min }`, you might expect that you can book 3 overlapping copies of that service, but the `bufferAfter` ends up blocking that.

If you have a second service that has overbooking but no buffer, that service _would_ be schedulable on top of appointments of that first type (since they would not collide with the buffers).

Trying to think of a better way to express this is hard 😓

packages/docs/docs/scheduling/defining-availability.md
noahsilas: Yeah, that's true; just because $find/$book doesn't set this doesn't mean that it isn't allowed. I'll clarify.

packages/server/src/fhir/operations/utils/scheduling.ts
Filupnot: Ah that makes sense, thanks for the explanation!

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -313,6 +315,126 @@ export function removeAvailability(availableIntervals: Interval[], blockedInterv
...
+  const result: Interval[] = [];
+  for (const slot of slots) {
+    if (slot.status !== 'free' && slot.status !== 'entered-in-error') {
+      const start = new Date(slot.start);
+      const end = new Date(slot.end);
+      if (Number.isNaN(start.valueOf()) || Number.isNaN(end.valueOf())) {
+        throw new Error('Got Slot with invalid start or end time');
+      }
+      if (end < start) {
+        throw new Error('Got Slot with start time before end time');
```
Filupnot:
```suggestion
        throw new OperationOutcomeError(badRequest(`Slot/${slot.id} has an end time before its start time`));

```

Also fixed error message itself, which should say "start time **after** end time"

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -313,6 +315,126 @@ export function removeAvailability(availableIntervals: Interval[], blockedInterv
...
+// Performance optimization: a faster implementation of `intervalsExceedingCapacity`
+// for the common case of "slotCapacity=1" (i.e.  no overbooking allowed).
+function intervalsExceedingCapacityOne(slots: Slot[]): Interval[] {
+  const result: Interval[] = [];
+  for (const slot of slots) {
+    if (slot.status !== 'free' && slot.status !== 'entered-in-error') {
+      const start = new Date(slot.start);
+      const end = new Date(slot.end);
+      if (Number.isNaN(start.valueOf()) || Number.isNaN(end.valueOf())) {
+        throw new Error('Got Slot with invalid start or end time');
```
Filupnot:
```suggestion
        throw new OperationOutcomeError(badRequest(`Slot/${slot.id} has an invalid start or end time`));
```

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -567,6 +688,7 @@ async function validateAvailability(
     slots: existingSlots,
     range: interval,
     serviceType: healthcareService.type,
+    capacity: parameters.get('slotCapacity'),
   });
   const hasAvailability = availability.some((avail) => avail.start <= interval.start && avail.end >= interval.end);
```
Filupnot: Take this scenario: 
- A schedule has `duration: 60`, `bufferBefore: 20`, `slotCapacity: 2`. 
- Booking A is 11-12, with its buffer slot at 10:40-11
- Booking B is 12-1, with its buffer slot at 11:40-12 (during booking A). 

Since buffer slots should always have a capacity of 1, booking B should be rejected. However, since the availability check here evaluates the entire booking as one interval at one capacity, it passes through. I validated this behavior with a test.

I believe the fix here would be to validate each proposed slot against the capacity it will actually carry, instead of one capacity across the union interval.

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -737,6 +896,7 @@ export async function createProposedAppointment(
     );
   }
 
+  stampBookingCapacity(slots, schedulingParametersGroup);
   customizer(appointment, slots);
```
Filupnot, line 963: The ordering of these two function calls is load-bearing. Might be worth a quick comment in the code calling this out, to prevent a future dev from accidentally swapping these.

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: Oops, that was sloppy of me, thanks!

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: I'm not entirely sure what you mean by this; my intuition here is that we want to fully prepare the appointment and slots before passing them into the `customizer`, so that specific endpoints can determine what they want to do with full information. But in practice, I don't think anything today customizes in such a way that the slot capacity extension being present would affect it?

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: Oh, yeah, nice catch. I had been merging the actual backing slot + any before/after buffers into the interval passed into here, which was valid when they shared all their constraints, but now they can mismatch on the capacity value, which makes this unsound.

I think we do need to follow that general strategy (validate each proposed slot), just need to figure out the best way to implement 👍

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: Ah, I think I've found your meaning; the customizer can change `Slot.status` (example: in $hold it can change from "busy" to "busy-tentative"), which can be impactful.

noahsilas:
@claude Please re-review. Pay close attention to the issue @Filupnot pointed out, that we had been treating buffer slots as though they needed to match the capacity of the main backing slot rather than having capacity fixed to `1`.

packages/server/src/fhir/operations/utils/scheduling.ts
Filupnot: Yeah, that was the sensitive piece there - you swapping `slot.status === 'busy'` to `slot.status !== 'busy-unavailable'` resolved the risk there.

packages/server/src/fhir/operations/utils/scheduling.ts
Filupnot: Staying true with this comment, should we also add `busy-tentative` slots here, so that slots created by `$hold` can be debuggable as well?

packages/server/src/fhir/operations/utils/scheduling.ts
```diff
@@ -313,6 +315,126 @@ export function removeAvailability(availableIntervals: Interval[], blockedInterv
...
+    // Apply every event at this instant before evaluating the segment [time, next):
+    // a booking ending here leaves before one starting here is admitted, so
+    // adjacent bookings net out at a shared boundary.
+    const time = events[i].time;
+    while (i < events.length && events[i].time === time) {
+      const event = events[i++];
+      if (event.entering) {
+        open.push(event.capacity);
+      } else {
+        open.splice(open.indexOf(event.capacity), 1);
```
Filupnot:
```suggestion
const idx = open.indexOf(event.capacity);
if (idx >= 0) {
  open.splice(idx, 1);
}
```

Right now if `indexOf` returns `-1` then this could remove the last element of `open`. Should be unlikely to occur, but feels worth adding a quick guard.

Filupnot (approved):

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: Addressed in b2f35639 by extracting a shared status checking utility

packages/server/src/fhir/operations/utils/scheduling.ts
noahsilas: Addressed in 6c2d02d799, where I chose to use `assert(idx >= 0)`. This is expected to always pass by invariant in construction of the events list, so this documents that we expect it to always be the case and fails loudly if it ever does not behave as anticipated.

PR #10186, opened 2026-08-11
TypedValue handles primitive extensions directly

mattwiller:
Removing the footgun of boxed primitive types in the FHIRPath core, replaced with a first-class property in `TypedValue.primitiveExtension`

### DIFF
--- a/packages/core/src/fhirpath/atoms.ts
+++ b/packages/core/src/fhirpath/atoms.ts
@@ -90,7 +90,12 @@ export class SymbolAtom implements Atom {
     if (this.name.startsWith('%')) {
       throw new Error(`Undefined variable ${this.name}`);
     }
-    return input.flatMap((e) => this.evalValue(e)).filter((e) => e?.value !== undefined) as TypedValue[];
+    return (
+      input
+        .flatMap((e) => this.evalValue(e))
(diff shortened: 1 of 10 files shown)


The-Alchemist:
@claude review

The-Alchemist:
out of scope for this PR, probably, but any thoughts on the FHIR's comment:

> When using a JavaScript JSON.parse() implementation, note that JavaScript natively supports only one numeric datatype, which is a floating point number. This can cause loss of precision for FHIR numbers. In particular, trailing 0s after a decimal point will be lost e.g. 2.00 will be converted to 2. The FHIR decimal data type is defined such that precision, including trailing zeros, is preserved for presentation purposes, and this is widely regarded as critical for correct presentation of clinical measurements. Implementations should consider using a custom parser and big number library (e.g. [https://github.com/jtobey/javascript-bignum ](https://github.com/jtobey/javascript-bignum)) to meet these requirements.

i.e., "Implementations should consider using a custom parser and big number library"

mattwiller:
@The-Alchemist Yep, the FHIRPath decimal arithemetic is supposed to be fixed-point: to support everything correctly, we would need to reimplement JSON parsing, as well as much of the FHIRPath core for handling numbers

The-Alchemist (approved):

PR #10188, opened 2026-08-12
feat(server): add Cron resource type for scheduled bots

### LINKED ISSUE #9776: [Integration Marketplace] Cron Resource Type
## Background

`Bot.cron` predates the shared-project model. It does not work with the shared model because the cron setup and auth context needs to be in the customer project, not the shared project.

 It forces cron logic and secrets into customer projects as a totally separate bot, making marketplace deployments messy.

Image

## Proposal

Introduce a new Medplum custom resource type: Cron.
- Lives in the customer project in the shared project model
- Points to a bot to execute (similar to Subscription)
- Only project admins can create it
- Backed by existing BullMQ cron infrastructure

## Open Question
- Should Cron have it's own ProjectMembership &  AccessPolicy and be treated as user type resource (Bot, User, ClientApplication) or inherit it's access from a ClientApplication that it is tied to? -> my slight preference is to inherit from ClientApplication to simplify marketplace package setup logic, but I think either will work

ThatOneBro:
## Problem

A Bot's schedule lives on the Bot itself (`Bot.cron[x]`). That ties one bot to one schedule, gives no way to pass input to a scheduled run — the bot is handed itself as input — and no way to run a bot under an identity other than its own.

## Solution

Adds a Medplum-custom `Cron` resource that owns the schedule instead:

| Element | Type | Cardinality |
| --- | --- | --- |
| `onBehalfOf` | `Reference<ProjectMembership>` | `1..1` |
| `parameters` | `Parameters` | `0..1` |
| `cronString` | `cronString` | `1..1` |
| `targetReference` | `Reference<Bot>` | `1..1` |
| `active` | `boolean` | `1..1` |
| `endTime` | `dateTime` |  `0..1` |

`onBehalfOf` is the membership the bot runs as, so the same bot can run under different identities and access policies on different schedules. Naming the membership directly avoids guessing which of a profile's memberships to use. It is required: a scheduled run has to execute as somebody, and defaulting to the target bot's own membership would pick an access policy by omission. `targetReference` is required for the same reason — together they remove any way to store a job that cannot run. Constraints `cron-1` and `cron-2` (`onBehalfOf.reference.exists()` / `targetReference.reference.exists()`) additionally reject a logical reference such as `{display: '...'}`, which names nothing the job can resolve. `parameters` becomes the bot input when present, falling back to the `Cron` itself. `cron[x]` stays a choice element so a `cronTiming` variant can be added later without renaming the field; `targetReference` is a plain element, since a single-type choice costs a `[x]` path in search expressions and buys nothing until a second type exists.

Execution is wired into `addCronJobs`, the same seam that registers `Bot.cron[x]` schedules, so no new dispatch call site was needed. Cron scheduler keys are namespaced `Cron/<id>` while Bot keys stay bare, leaving already-registered Bot schedulers untouched.

A bot is code, not authority, so the two references are scoped differently. `targetReference` may name a Bot in a **linked** project that exports `Bot` -- the shared-project model this resource exists for, where the schedule and auth context live in the customer project and the Bot lives in the shared one. `onBehalfOf` may not: it decides which access policy applies, and every customer project links the same shared project, so a membership elsewhere would let any tenant run with that project's privileges.

Both are authorized on write, by resolving them through the author's own repository rather than comparing `meta.project`. That repository already encodes the rules -- `ProjectMembership` is a project admin type and so never crosses a link, while `Bot` is not -- and it keeps a `Cron` that could only ever fail from reaching the scheduler at all. `execBot` re-checks the link, which can be revoked after the write, and unregisters the job instead of failing on every tick.

The rest of bot execution already supported this: `getBotSecrets` layers the bot project's secrets under the runAs project's, and subscriptions resolve `runAs` in the subscription's project. The execution `AuditEvent` follows the identity, so it lands in the customer project; the bot's account compartments no longer travel with it when the two projects differ.

### Also fixes a pre-existing bug

`addCronJobs` was only called when the interaction was not a delete, so **deleting a scheduled Bot left its job running forever**. It now runs on delete for both resource types. The regression test fails against the old code.

### Generated artifacts

Only the `StructureDefinition` and `SearchParameter` entries are hand-written. `Cron.d.ts`, `fhir.schema.json`, `v112.ts`, and the docs page are generator output.

The docs generator strips ONC compliance text it then restores by applying `ONC-compliance-text.patch`; that patch no longer applies cleanly (it fails on `provenance.mdx` and `questionnaireresponse.mdx`), and `git apply` is all-or-nothing, so a full regeneration would strip the text from ~40 unrelated files. Only the `Cron` output is taken here. That bit-rot is pre-existing and worth a separate fix.

## Testing

- `packages/server/src/workers/cron.test.ts` — 26 passed. New coverage: `onBehalfOf` and `targetReference` each required and each rejecting a logical reference, `parameters` passed as bot input with the `Cron` itself as fallback, scheduler key namespacing, invalid `cronString` schedules nothing, clearing `cronString` removes the job, deleting a `Cron` removes the job, deleting a `Bot` removes the job, the project `cron` feature gate, and `execBot` running the target bot as `onBehalfOf`. Cross-project coverage: a linked bot runs under the customer membership, an unlinked target and a linked-but-unexported target are both rejected on write and schedule nothing, `onBehalfOf` is rejected in both a linked and an unlinked project, and revoking the link unregisters the job on the next tick.
- `dispatch` 7/7, `migrate` 30/30, `searchparameter` 32/32, `jsonschema` 9/9, core `validation` + `schema` 82 passed.
- `tsc --noEmit`, eslint, and prettier clean.

## Notes for review

- `Bot.cron[x]` and `Cron` can now both schedule a bot; a bot scheduled both ways runs twice. Whether `Cron` should supersede `Bot.cron[x]` is left open.
- The migration generator strips a hand-added `// CAUTION: LOAD-BEARING COMMENT` before `v100` in `schema/index.ts` on every run. Restored here, but it will keep disappearing.

Closes #9776


packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -10585,5 +10585,242 @@
...
+                "severity" : "error",
+                "human" : "A cron job resolves its identity by reading onBehalfOf, so a logical reference is not enough.",
+                "expression" : "onBehalfOf.reference.exists()"
+              },
+              {
+                "key" : "cron-2",
+                "severity" : "error",
+                "human" : "A cron job resolves its target by reading targetReference, so a logical reference is not enough.",
+                "expression" : "targetReference.reference.exists()"
+              }
```
mattwiller: These constraints could live directly on the fields they target

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -10585,5 +10585,242 @@
...
+              "max" : "1"
+            },
+            "type" : [{
+              "code" : "Reference",
+              "targetProfile" : ["https://medplum.com/fhir/StructureDefinition/ProjectMembership"]
+            }]
+          },
+          {
+            "id" : "Cron.parameters",
+            "path" : "Cron.parameters",
```
mattwiller: To align with general FHIR naming conventions, this should be singular, i.e. `parameter`

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -10585,5 +10585,242 @@
...
+            "definition" : "Parameters passed to the target as input each time the scheduled job runs.",
+            "min" : 0,
+            "max" : "1",
+            "base" : {
+              "path" : "Cron.parameters",
+              "min" : 0,
+              "max" : "1"
+            },
+            "type" : [{
+              "code" : "Parameters"
```
mattwiller: Rather than embedding another resource (which is kind of a weird pattern), why not just re-use the same field definition as `Parameter.parameter` here?

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -10585,5 +10585,242 @@
...
+            }]
+          },
+          {
+            "id" : "Cron.cron[x]",
+            "path" : "Cron.cron[x]",
+            "short" : "The schedule on which the job runs",
+            "definition" : "A schedule for the job to be executed, as a cron expression.",
+            "min" : 0,
```
mattwiller, line 11237: What does it mean if this field is missing?

packages/definitions/src/fhir/r4/profiles-medplum.json
```diff
@@ -10585,5 +10585,242 @@
...
+              "max" : "1"
+            },
+            "type" : [{
+              "code" : "Reference",
+              "targetProfile" : ["https://medplum.com/fhir/StructureDefinition/Bot"]
+            }]
+          }
+        ]
+      }
+    }
```
mattwiller, line 11273: Some other ideas for fields: `status` so one can be easily disabled if need be, and potentially an `endTime` after which the cron will not run

packages/docs/static/data/medplumDefinitions/cron.json
```diff
@@ -0,0 +1,256 @@
...
+      "path": "Cron.parameters",
+      "min": 0,
+      "max": "1",
+      "short": "Parameters passed to the target",
+      "definition": "Parameters passed to the target as input each time the scheduled job runs.",
+      "comment": "",
+      "inherited": false
+    },
+    {
+      "name": "cron[x]",
```
mattwiller: This should just be `cronString` today, until we add another type

packages/server/src/workers/cron.ts
```diff
@@ -95,35 +108,72 @@ export async function addCronJobs(
...
+  return resource.resourceType === 'Cron' ? getCronStringForCron(resource) : getCronStringForBot(resource);
+}
+
+function getCronStringForCron(cron: Cron): string | undefined {
+  if (cron.cronString && isValidCron(cron.cronString)) {
+    return cron.cronString;
+  }
+
+  // Otherwise, this is not a valid cron job
+  return undefined;
```
mattwiller: Is there any way for a user to know that their cron string was invalid/didn't execute?

packages/definitions/src/fhir/r4/profiles-medplum.json
ThatOneBro: Good callout, I like the idea of `status` and `endTime`

packages/server/src/workers/cron.ts
ThatOneBro: Maybe we could write an AuditEvent ?

ThatOneBro: It also might be worth enforcing this via constraint on the cron string itself

packages/definitions/src/fhir/r4/profiles-medplum.json
ThatOneBro: Good question, I think originally I was comparing with the `Bot.cron` implementation where removing the cron string removes the cron job. But using `active` as that "lever" makes more sense

mattwiller:
```suggestion
            "min" : 1,
```

packages/server/src/fhir/accesspolicy.ts
```diff
@@ -349,6 +349,11 @@ function applyProjectAdminAccessPolicy(
         hiddenFields: ['passwordHash', 'mfaSecret'],
         readonlyFields: ['email', 'emailVerified', 'mfaEnrolled', 'project'],
       },
+      {
+        // A Cron names the membership it runs as, so authoring one confers that identity
```
mattwiller:
```suggestion
```

packages/server/src/workers/cron.ts
```diff
@@ -95,33 +108,86 @@ export async function addCronJobs(
...
-    logger.info('Removing cron job for bot', { botId: bot.id });
-    await queue.removeJobScheduler(bot.id);
+    logger.info('Removing cron job for resource', { schedulerId, ...resourceIds });
+    await queue.removeJobScheduler(schedulerId);
+  }
+}
+
+function isSchedulable(resource: Resource | undefined): resource is WithId<Bot> | WithId<Cron> {
+  return (resource?.resourceType === 'Bot' || resource?.resourceType === 'Cron') && resource.id !== undefined;
+}
```
mattwiller: The input is already `WithId<Resource>`, so this can probably be simplified
```suggestion
const schedulableTypes = ['Bot', 'Cron'];
function isSchedulable(resource: WithId<Resource> | undefined): resource is WithId<Bot> | WithId<Cron> {
  return schedulableTypes.includes(resource.resourceType);
}
```

packages/server/src/workers/dispatch.ts
```diff
@@ -156,6 +156,17 @@ export async function execDispatchJob(job: Job<DispatchJobData>): Promise<void>
     });
   }
 
+  // Runs on delete as well: a deleted resource still has a schedule to tear down.
```
mattwiller, line 159: 👍

packages/server/src/workers/cron.ts
```diff
@@ -185,21 +251,53 @@ export function convertTimingToCron(timing: Timing): string | undefined {
...
+      // that can no longer run unregisters itself on its next tick.
+      await removeBullMQJobByKey(getSchedulerId(cron));
+      return;
+    }
+
+    bot = await systemRepo.readReference<Bot>(cron.targetReference);
+
+    // `onBehalfOf` is the membership the bot runs as, so it decides which access policy applies.
+    // Both it and the bot are confined to the Cron's own project; otherwise a Cron could name a
+    // membership elsewhere and run with its privileges.
```
mattwiller:
```suggestion
```

packages/server/src/workers/cron.ts
```diff
@@ -185,21 +251,53 @@ export function convertTimingToCron(timing: Timing): string | undefined {
...
+    const projectRef = `Project/${cron.meta?.project}`;
```
mattwiller: This is only used in one place below; I'd say inline it

codyebberson (commented):
Is there a way to use the real `Parameters` rather than creating new `CronParameter`?

mattwiller (approved):

mattwiller:
A few additional things to consider:
- If the `cron` feature gets turned off, should that cancel existing jobs?
- Should this have an `identifier` field for search and upserting?
- Does this need timezone support?
- For the future: do we want a `Cron/$execute` operation for manually triggering a job

PR #10201, opened 2026-08-13
`getIdentifierByType` and MRN support

### LINKED ISSUE #10275: [Scheduling] Find & Book / MVP

Filupnot:
## `getIdentifierByType(resource, type)`

New core API. Given a resource and a `type` coding, returns the value of the first identifier whose `Identifier.type.coding` contains a coding matching both the system and the code of `type`. Returns `undefined` if there's no match, or if `type` is missing either a system or a code.

This is notably different from the existing `getIdentifier(resource, system)`, which matches on `Identifier.system` (who assigned the identifier) rather than on `Identifier.type`, what kind of identifier it is. For example, `MRN_IDENTIFIER_TYPE` finds the medical record number no matter which organization issued it.

## `HL7_V2_0203`

Constant for the URL `http://terminology.hl7.org/CodeSystem/v2-0203` - HL7 v2 Table 0203 "Identifier Type", the code system for `Identifier.type`.

## `MRN_IDENTIFIER_TYPE`

Coding for the medical record number identifier type: `{ system: HL7_V2_0203, code: 'MR' }`.


packages/core/src/utils.ts
```diff
@@ -445,6 +446,40 @@ export function getIdentifier(resource: Resource, system: string): string | unde
...
+ * @param type - The identifier type coding, such as {@link MRN_IDENTIFIER_TYPE}.
+ * @returns The identifier value if found; otherwise undefined.
+ */
+export function getIdentifierByType(resource: Resource, type: Coding): string | undefined {
+  const identifiers = (resource as any).identifier;
+  if (!identifiers || !isCoding(type) || !type.system) {
+    return undefined;
+  }
+  const array = Array.isArray(identifiers) ? identifiers : [identifiers];
+  for (const identifier of array) {
```
techdavidy: Nit: I think you can do array.first ?

techdavidy:
@claude please review

techdavidy (approved):

PR #10202, opened 2026-08-13
perf(search): fix quadratic query string parsing on duplicate params

ThatOneBro:
## Problem

`parseQueryString()` in `packages/fhir-router/src/urlrouter.ts` paired `URLSearchParams.keys()` with `getAll()`:

```ts
for (const param of raw.keys()) {
  const values = raw.getAll(param);
  queryParams[param] = values.length === 1 ? values[0] : values;
}
```

`keys()` yields one entry per name/value **pair**, not per distinct name. A query with n copies of a param therefore iterated n times, and each iteration called `getAll()`, which rescanned all n pairs and allocated a fresh n-element array — then overwrote the same key n times. Duplicate params are the worst case, which is what we were seeing block the event loop in production.

This is on the hot path for **every** FHIR request: the Express layer defers query parsing to the router (`packages/server/src/fhir/routes.ts`, `query: Object.create(null)` with "Defer query param parsing to router for consistency"). Batch pays it **three times per entry**, since `BatchProcessor` calls `router.find()` during interaction bucketing, identity resolution, and execution.

## Measurements

All-duplicate params, single call:

| n params | before | after | speedup |
|---------:|-------:|------:|--------:|
| 1,000 | 19.2 ms | 0.60 ms | 32x |
| 5,000 | 191 ms | 1.98 ms | 96x |
| 10,000 | 720 ms | 2.17 ms | 332x |
| 20,000 | 3,473 ms | 2.91 ms | 1192x |

## Changes

- **`urlrouter.ts`** — accumulate in a single pass, promoting to an array only on the second occurrence. Output is unchanged: one occurrence stays a bare string, repeats become an array in document order. Verified against single values, duplicates, interleaved duplicates, and percent-encoded values. `Object.create(null)` is retained, so a `__proto__` param still cannot pollute the prototype.
- **`core/search/search.ts`** — `parseSearchRequest` built its query array with `push(...entries)`, which applies each entry as a separate argument and threw `RangeError: Maximum call stack size exceeded` at ~200k params. Since `RangeError` is not an `OperationOutcomeError`, that surfaced as a 500 rather than a 400.
- **`server/fhir/search.ts`** — hoist the five fixed special search parameter implementations (`_id`, `_lastUpdated`, `_deleted`, `_project`, `_compartment`) so their FHIRPath expressions parse once at module load instead of per matching filter. The build functions treat `impl` as read-only, so they are safe to share.
- **`core/search/derived.ts`** — memoize `deriveIdentifierSearchParameter` on a `WeakMap`.
- **`fhir-router/search-include.ts`** — pass an LRU cache to `evalFhirPathTyped` so `_include` expressions are not re-parsed per request. (Deliberately not switching to `getSearchParameterDetails().parsedExpression`: that AST is narrowed per resource type by `getParsedExpressionForResourceType` and is not equivalent to the raw expression.)

## Tests

7 new cases in `urlrouter.test.ts` covering the value shapes above, plus a regression guard that parses 10k duplicate params and asserts it completes under 250 ms — the previous implementation took ~720 ms and would fail.

Existing suites: fhir-router 149 passed, core search 212 passed, server search/batch/searchparameter 561 passed. `eslint`, `prettier --check`, and `tsc --noEmit` clean across all three packages.

Note: `packages/server/src/fhir/repo.test.ts` has 2 pre-existing failures around reader/writer pool routing. They reproduce on a clean tree and are unrelated to this change.

## Follow-ups, not in this PR

- **Attachment rewrite allocation.** `rewriteAttachments` runs on every authenticated response (`response.ts`) and unconditionally rebuilds the entire object graph via `Object.fromEntries`, plus one `await` per property, even when a resource has zero attachments. Measured ~13 ms for a 1000-resource search Bundle. A sync pre-scan that skips the async path entirely brings that to ~3 ms — but the predicate must test for an actual Binary URL via `normalizeBinaryUrl`, since every FHIR extension has a `url` property and a naive check bails immediately.
- **Batch conditional-lookup dedup.** 500 conditional-create entries produce ~1000 `parseSearchRequest` calls and ~1000 identical SQL searches (once in preprocess, once inside `conditionalCreate`). Touches transaction semantics.
- **No query-param count limit** exists anywhere in the request path. This fix largely defuses the DoS vector, but an explicit cap may still be worth adding.

### DIFF
--- a/packages/core/src/search/search.ts
+++ b/packages/core/src/search/search.ts
@@ -175,7 +175,11 @@ export function parseSearchRequest<T extends Resource = Resource>(
   // First, we convert the URLSearchParams to an array of key-value pairs
   const queryArray: [string, string][] = [];
   if (searchParams) {
-    queryArray.push(...searchParams.entries());
+    // Not `push(...entries)`: spreading applies each entry as a separate argument,
+    // which overflows the stack on very large queries.
+    for (const entry of searchParams) {
+      queryArray.push(entry);
+    }
   }
 
   // Next, we merge in the query object
--- a/packages/core/src/search/derived.ts
+++ b/packages/core/src/search/derived.ts
@@ -3,6 +3,8 @@
 import type { SearchParameter } from '@medplum/fhirtypes';
 import { DotAtom, FhirPathAtom, SymbolAtom } from '../fhirpath/atoms';
 
+const derivedIdentifierCache = new WeakMap<SearchParameter, SearchParameter>();
+
 /**
  * Derives an "identifier" search parameter from a reference search parameter.
  *
@@ -17,13 +19,19 @@ import { DotAtom, FhirPathAtom, SymbolAtom } from '../fhirpath/atoms';
  * @returns The derived "identifier" search parameter.
  */
 export function deriveIdentifierSearchParameter(inputParam: SearchParameter): SearchParameter {
-  return {
-    resourceType: 'SearchParameter',
-    code: inputParam.code + ':identifier',
-    base: inputParam.base,
-    type: 'token',
-    expression: `(${inputParam.expression}).identifier`,
-  } as SearchParameter;
+  // Callers must treat the result as read-only: it is shared by every caller for a given input.
+  let result = derivedIdentifierCache.get(inputParam);
+  if (!result) {
+    result = {
+      resourceType: 'SearchParameter',
+      code: inputParam.code + ':identifier',
+      base: inputParam.base,
+      type: 'token',
+      expression: `(${inputParam.expression}).identifier`,
+    } as SearchParameter;
+    derivedIdentifierCache.set(inputParam, result);
+  }
+  return result;
 }
 
 export function getInnerDerivedIdentifierExpression(expression: string): string | undefined {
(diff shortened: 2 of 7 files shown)


codyebberson:
@claude - please review

Please pay extra attention to any possible changes in behavior, this is intended to be purely a performance speedup.

codyebberson (approved):

packages/core/src/search/search.ts
```diff
@@ -175,7 +175,11 @@ export function parseSearchRequest<T extends Resource = Resource>(
   // First, we convert the URLSearchParams to an array of key-value pairs
   const queryArray: [string, string][] = [];
   if (searchParams) {
-    queryArray.push(...searchParams.entries());
+    // Not `push(...entries)`: spreading applies each entry as a separate argument,
+    // which overflows the stack on very large queries.
+    for (const entry of searchParams) {
+      queryArray.push(entry);
+    }
   }
```
noahsilas, line 183: I was curious if there was perf to be had in switching this to a raw spread, which I was hoping V8 would be able to do some especially good optimizations for. I compared to this
```suggestion
const queryArray = searchParams ? [...searchParams] : [];
```

Asked claude to run a quick benchmark and it reported that the for/push style you wrote here is 1-2x faster.

Nice 👍

PR #10207, opened 2026-08-13
Dedupe search filters

mattlong:
Another layer to prevent impact from degenerate searches.


ThatOneBro (approved):
Nice 🚀 

I was thinking about this as the next step too

ThatOneBro:
@claude Please review, making sure to check that there is no way that search queries are parsed semantically differently after this change

PR #10210, opened 2026-08-13
Consistent batch telemetry

mattlong:
Previously, async batches were not being considered/contributing to the telemetry.


packages/server/src/fhir/batch-telemetry.ts
```diff
@@ -0,0 +1,44 @@
...
+    const event = e as LogEvent;
+    ctx.logger.warn(event.message, { ...event.data, project: ctx.project.id });
+  });
+
+  router.addEventListener('batch', (e: any) => {
+    const ctx = getAuthenticatedContext();
+    const projectId = ctx.project.id;
+    const { count, errors, errorCount, size, bundleType } = e as BatchEvent;
+
+    const metricOpts = { attributes: { bundleType, projectId } };
```
ThatOneBro: Project ID as an attribute might be too high cardinality? I believe we pay $0.30 per metric configuration on Cloudwatch at the moment. cc @mattwiller

The-Alchemist: had a similar issues with data warehouse metrics: i wanted to measure number of rows per `project_id`.  turns out for, Cloudwatch metrics "every label combination creates a billed metric stream.".

https://gemini.google.com/share/d/1V4IPEx4EPOsJPkn7FpplA4LfpAmQXDOp?usp=sharing

maybe one day we'll switch to AWS Managed Prometheus (AMP), which apparently is much cheaper

The-Alchemist:
@claude review.  especially check `packages/server/src/fhir/batch-telemetry.ts` , line 29 and on to make sure we don't have cardinality explosion, and thus CloudWatch metric cost explosion.

The-Alchemist:
@claude review

mattlong:
@claude look again

mattlong:
@claude regression test added

ThatOneBro (approved):

PR #10215, opened 2026-08-14
fix(provider): Use ListWithDetailPane in spaces.

techdavidy:
- Use ListWithDetailPane for consistent UX in spaces.
- Touch parent Communication every time user creates a new Communication. This helps sorting topics by latest.


codyebberson:
@claude - please review

1. How is this different than `ResourceBoard` (`packages/react/src/ResourceBoard/ResourceBoard.tsx)?
2. If it is sufficiently different, can you please propose alternate names?  Something about using `With` in the middle of the component name feels inadequate.
3. Can you please analyze whether there are any existing Mantine components that we can use to streamline this implementation and try to stick to stock Mantine as much as possible?

codyebberson:
Preference for renaming `ListWithDetailPane` to just `ListDetailPane` 🤷

techdavidy:
@codyebberson 👍🏻can do in a separate PR to rename all places where it is used.

codyebberson:
> @codyebberson 👍🏻can do in a separate PR to rename all places where it is used.

Given that this will be a react component published in `@medplum/react`, I think we should do it now to avoid any backwards compatibility / breaking changes concerns.

noahsilas:
> Preference for renaming `ListWithDetailPane` to just `ListDetailPane` 🤷

To push back on this a little bit:
- This was already published as `ListWithDetailPane` in #9484 and included in v5.1.19+
- Stylistically, `ListDetailPane` to me sounds like it is just the interior pane, instead of the combo of the List pane + Detail pane.

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
```diff
@@ -321,401 +323,434 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
...
+        </ActionIcon>
+      }
+      emptyList={
+        <Text size="sm" c="dimmed" ta="center" mt="md">
+          No conversations yet
+        </Text>
+      }
+      refresh={refreshTopics}
+      renderItem={(item) => <TopicListItem topic={item} uri={onSelectedItem(item)} />}
+      renderDetail={() => (
```
noahsilas, line 346: I think that because this doesn't take `refresh`, the `refreshTopics` prop being passed in here ends up unused, which opens up an opportunity for a little simplification (the effect can be a more obvious translation of the older version).

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
```diff
@@ -321,401 +323,434 @@ export function SpacesInbox(props: SpaceInboxProps): JSX.Element {
...
+      headerActions={
+        <ActionIcon variant="subtle" color="gray" onClick={() => setSidebarOpen(false)} aria-label="Hide conversations">
+          <IconLayoutSidebarLeftCollapse size={18} />
+        </ActionIcon>
+      }
+      emptyList={
+        <Text size="sm" c="dimmed" ta="center" mt="md">
+          No conversations yet
+        </Text>
+      }
```
noahsilas, line 343: I think that this is never used because `selected` is always some kind of truthy value. Might be better to omit this prop so that it doesn't mislead future readers.

noahsilas (commented):
It feels like there's a bit of an awkward fit here; on mounting we are using a placeholder topic that isn't created yet, so the list state doesn't necessarily match the detail state. The render function for the detail view doesn't actually receive the selected item, as this component controls everything and uses the closed over state in the internal rendering.

So, it seems like this gets us the styling consistency, which is nice, but at the expense of some slightly awkward semantics that seem more likely to regress in the future? (You can imagine a rework of the `ListDetailPane` that renders the empty state if the length of the list is `0` instead of by checking for the `selected` prop).

I wonder if we could create an actual placeholder in the topics list? One difficulty is that when you select it we try to navigate to `/:id`, which doesn't exist on this unsaved row. Maybe we could navigate to a special url like `/new` instead? The initial save could do a history replacement with the persisted ID and hopefully not trigger a remount. I wonder how often we will have this kind of pattern of a list view with a form and if it's worth building this out as an explicit pattern?

examples/medplum-provider/src/components/spaces/SpacesInbox.tsx
techdavidy: emptyList depends on !loading && items.length === 0, independently of selected. It displays “No conversations yet” in the sidebar, and an existing test verifies this.

techdavidy:
@claude review. Handled the refresh as optional. 
Also on ThreadInbox we just keep the single communication in memory it is fine if it is not synced given that it requieres the user to send a message inside the thread.

techdavidy:
@claude review

PR #10217, opened 2026-08-16
Fixes returning Paramaters from custom bot operation

codyebberson:


codyebberson:
@claude - please review 

Please analyze whether to do this check in `custom.ts` vs the "make output parameters" helper

codyebberson:
@claude - please implement "Ordering vs the !result.success guard. The passthrough (custom.ts:101-103) runs before the !result.success check (105-108). If a bot execution fails but its returnValue happens to be Parameters-shaped, this returns 200 allOk and masks the failure. On error the return value is normally an OperationOutcome (already caught at line 97), but that isn't guaranteed. Safer to only pass through on success — move the check below the !result.success guard."

codyebberson:
@claude please re-review

reshmakh (approved):

PR #10218, opened 2026-08-16
feat(provider): Encounters use ResourceBoard

techdavidy:
Screenshot 2026-08-16 at 2 32 49 PM

https://youtu.be/ElL2Z22XATE


techdavidy:
@claude please review

examples/medplum-provider/src/pages/encounter/EncountersPage.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <Text c="dimmed">Select a visit to view its chart.</Text>
+          </Box>
+        }
+        onSelectFirst={(encounter) => navigate(encounterUri(encounter.id), { replace: true })?.catch(console.error)}
+        onChange={(s) => {
+          navigate(`${location.pathname}${formatSearchQuery(s)}`)?.catch(console.error);
+        }}
+        onError={showErrorNotification}
+      />
+      <Outlet />
```
noahsilas: My initial reaction here was a bit of a "Huh?" kind of moment - the ResourceBoard is a full page experience, so having an "Outlet" beneath it felt weird since there isn't a good place to render a sub-page. But in this case, the nested page is rendered as a Modal (and you have updated the TaskDetailsModal so that it no longer renders anything outside of its Modal container), so this works as expected.

I think it might be more intuitive for future readers if this component rendered the modal itself instead of breaking it into two components. I think the route would look something like `<Route path="Encounter/:encounterId/Task?/:taskId?` element={<EncountersPage/>}>`. That has some niceties - there's a bit of awkwardness in the current implementation where the EncountersPage is responsible for the querystring part of navigation even when the modal is mounted, so the modal has to be cautious not to remove those params.

On the other hand, the current implementation makes it clear how this could compose with other modal types later if we wanted to introduce `Encounter/:encounterId/ChargeItem/:chargeItemId` or something. I think I'd probably leave this as-is for that reason, and assume that future readers will also see that the only valid subpages are similar modal experiences.

examples/medplum-provider/src/components/encounter/EncounterHeader.tsx
```diff
@@ -121,7 +121,7 @@ export const EncounterHeader = (props: EncounterHeaderProps): JSX.Element => {
         <Flex justify="space-between" align="center" p="lg">
           <Stack gap={0}>
             <Text fw={800} size="lg">
-              {encounter.basedOn?.[0]?.display || 'Visit'}
+              {encounter.type?.[0]?.text || 'Visit'}
```
noahsilas, line 124: I'm not really sure how to evaluate this change - switching from `encounter.basedOn`(a `Reference<CarePlan | DeviceRequest | MedicationRequest | ServiceRequest>[]`) to `encounter.type` (a `CodeableConcept[]`) seems fine.

I wonder if this should use `<CodeableConceptDisplay value={encounter.type?.[0]}>`? Although we might need to add an extra prop like `fallback="Visit"` or something there to support this use case.

noahsilas (approved):

examples/medplum-provider/src/components/encounter/EncounterHeader.tsx
techdavidy: We are now pass the PlanDefinition title to encounter.type, and display its text.
So if you select PlanDefinition, Healthy Daily Routine, that would be the "title" of the encounter.

techdavidy:
@claude please review

PR #10230, opened 2026-08-17
Preauthorize code: fix ClientApplication read

maddyli:
Currently, if the user attempting to generate a preauthorize code does not have access to the ClientApplication, it will run into a forbidden error. 

This fix reads the ClientApplication via systemRepo, so that the user does not need to have access to the ClientApplication resource in order to properly generate the preauthorize code.


The-Alchemist:
@claude review

ThatOneBro (approved):

PR #10234, opened 2026-08-18
feat(core): add $get-cart cart-contents operation support

oleg-mp:
## Summary

Vendor-neutral client support for reading a patient's medication cart as the **vendor** holds it, reconciled against local draft `MedicationRequest`s.

Counting drafts alone gives a lower bound on what a prescriber will see in the signing widget: it misses lines staged through another UI, cannot confirm whether a failed `removeFromCart` left the vendor line behind, and offers no drift check before checkout. `getCart` closes that gap.

- **`@medplum/core`** — `MedicationCartContentsRequest` / `MedicationCartContentsResponse` / `MedicationCartLine` / `MedicationCartLineStatus`, a type guard, and `medicationCartContentsRequestToParameters` / `parametersToMedicationCartContentsResponse`, alongside the existing checkout and cart-manage converters.
- **`@medplum/react-hooks`** — `getCart` on `useMedicationCart`, POSTing to `$get-cart`.
- **`@medplum/scriptsure-react`** — flows through `useScriptSureCart` automatically (thin re-export); doc-only change.
- **Docs** — a `getCart` section on the Medication Cart page.

`status` on each line is `in-sync`, `vendor-only` (vendor holds a line with no local draft), `not-staged` (draft not checked out yet; benign), or `missing-from-vendor` (resolves the ambiguity a failed removal leaves behind). The response also carries `vendorTotal` against `draftCount` and a `locked` flag.

As with the other cart operations, the vendor binding lives entirely server-side: a project's `OperationDefinition` points `$get-cart` at whichever bot implements it. The ScriptSure implementation is in medplum-ee (medplum/medplum-ee#993, closing medplum/medplum-ee#984). Vendors without a batch cart simply never call it.

## Test plan

- [x] 60 core tests including round-trip, unknown-status rejection, and missing-field validation
- [x] 244 react-hooks tests including the new `getCart` encode/decode and non-`Parameters` guard
- [x] `medplum-provider` typechecks and its 17 `MedicationsPage` tests pass — the added interface member does not break existing cart-hook mocks
- [x] `tsc --noEmit` and eslint clean in core, react-hooks, and scriptsure-react

### DIFF
--- a/packages/core/src/medication-order-utils.ts
+++ b/packages/core/src/medication-order-utils.ts
@@ -1026,6 +1026,189 @@ export function parametersToMedicationCartManageResponse(params: Parameters): Me
 }
 
 // ============================================================================
+// Cart contents ($get-cart)
+// ============================================================================
+
+/**
+ * Stable error when a cart-read bot response does not match
+ * {@link MedicationCartContentsResponse}.
+ */
+export const INVALID_MEDICATION_CART_CONTENTS_RESPONSE = 'Invalid response from get-cart bot';
+
+/**
+ * Reconciliation verdict for one cart line, comparing what the vendor holds
+ * against the local draft `MedicationRequest`s:
+ *
+ * - `in-sync`: the vendor line matches a draft. The normal case.
+ * - `vendor-only`: the vendor holds a line with no local draft — staged through
+ *   another UI or by a session whose draft is gone. The prescriber will still
+ *   see it in the approval widget, so a draft-derived cart count under-reports.
+ * - `not-staged`: a draft that has not been checked out yet, so the vendor
+ *   legitimately does not hold it. Benign, not drift.
+ * - `missing-from-vendor`: a draft the vendor no longer holds. Resolves the
+ *   ambiguity left by a failed `removeFromCart`: the line really is gone.
+ */
+export type MedicationCartLineStatus = 'in-sync' | 'vendor-only' | 'not-staged' | 'missing-from-vendor';
+
+/** One reconciled line in a {@link MedicationCartContentsResponse}. */
+export interface MedicationCartLine {
+  readonly status: MedicationCartLineStatus;
+  /** Local draft id; absent for `vendor-only` lines. */
+  readonly medicationRequestId?: string;
+  /** Vendor cart-item reference; absent for `not-staged` drafts. */
+  readonly vendorLineId?: string;
+  /** Display fields off the vendor line; absent for `not-staged` drafts (render those from the draft). */
+  readonly drugName?: string;
+  readonly ndc?: string;
+  readonly rxnorm?: string;
+  readonly quantity?: number;
+  /** Vendor readiness hint (e.g. `Ready`, `Incomplete`); not a guarantee the line will send. */
+  readonly validationState?: string;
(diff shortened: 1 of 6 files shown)


oleg-mp:
@claude review this

oleg-mp:
Thanks — fixed the `locked` default and the test gap in 3d1e736; declining the decoder refactor, which you'd already flagged as out of scope.

**Fixed: `locked` defaulted to `false` on a malformed response.** You're right, and it's the same mistake this feature exists to correct, one layer down. The operation was built because a cart count derived from drafts silently substitutes a comfortable number for one we don't actually have; decoding an absent `locked` as "not locked" does exactly that with the write-safety signal. It was also inconsistent with its own guard — the counts fail closed through `Number.NaN`, and `locked` didn't.

It now passes `map.locked` through unchecked so the guard's `typeof obj.locked === 'boolean'` rejects a missing or malformed flag, which is the direct boolean analog of the `NaN` sentinel rather than a second validation path.

Before making it fail closed I checked that an *unlocked* cart still decodes, since dropping `false` anywhere in the chain would have turned this into a hard failure on the common case. It survives: `buildOutputParameters` skips a value only when `isEmpty(value)`, and `isEmpty(false)` is `false` for booleans, so the server emits `valueBoolean: false`; `readParameterValue` returns the first `value*` **key** it finds rather than testing truthiness, so `false` comes back intact. The ScriptSure bot always sends the field (`locked: (cart.locks?.length ?? 0) > 0`, declared `min: 1` on the OD), so no real response is affected.

**Fixed: no decode coverage for `not-staged` / `missing-from-vendor`.** Agreed these were the two worth having, since they're the ones the docs make structural claims about. Added a case asserting `not-staged` carries a `medicationRequestId` and no vendor display fields, and `missing-from-vendor` keeps the `vendorLineId` it was stamped with — that rxId is precisely what makes a vendor-side removal verifiable, so it silently dropping would defeat the status. Also added a case for the rejected `locked` flag above. 62 core tests.

**Declining: collapsing the three `*FromPart` helpers.** Agreed with your own read that it doesn't belong here — a shared `partsToMap` would mean editing the shipped `$checkout-medications` and `$remove-cart-medication` decoders for a cosmetic win, widening the blast radius of a PR that otherwise only adds. Worth doing when a fourth one arrives and the pattern is proven, rather than generalizing from three.

packages/core/src/medication-order-utils.ts
```diff
@@ -1025,6 +1025,193 @@ export function parametersToMedicationCartManageResponse(params: Parameters): Me
...
+ * @param value - Unknown bot JSON payload.
+ * @returns True when the value matches {@link MedicationCartContentsResponse}.
+ */
+export function isMedicationCartContentsResponse(value: unknown): value is MedicationCartContentsResponse {
+  if (typeof value !== 'object' || value === null) {
+    return false;
+  }
+  const obj = value as Record<string, unknown>;
+  return (
+    typeof obj.vendorPatientId === 'number' &&
```
codyebberson, line 1107: Do we already have a `isNumber(value: unknown)` util somewhere?

codyebberson (approved):
Overall lgtm.  

I didn't see a good generic `isNumber()` validator anywhere in `core`.  Might be worth doing separately as a follow-up

PR #10235, opened 2026-08-18
test(agent): cover a Beckman AU serial record on the byte-stream channel

ThatOneBro:
The byte-stream channel's existing protocol coverage is all ASTM, which is a poor stand-in for the legacy serial protocols that also run over it. A Beckman AU record has no frame number, no checksum and no record terminators — it is space-delimited fixed-width text wrapped in STX/ETX and nothing else. The only filtering it needs is `stripControlChars`, and the body has to arrive as text rather than hex for a bot to parse the leading `D ` record type.

This adds a test driving an AU result record through the channel with `startChar=%02&endChar=%03&stripControlChars=true&bodyEncoding=utf-8`, and asserting the delivered body is exactly the record with its framing stripped.

Dropping either param fails the test: without `stripControlChars` the body keeps its STX, and without `bodyEncoding=utf-8` the body arrives hex-encoded and a bot finds `0` where it expects `D`. Both are the channel's documented defaults, so this pins the configuration a text-oriented serial device actually requires.

The record in the test is invented — field widths and column layout match a real AU frame, the identifiers and values do not.

Test only — no behavior change.

### DIFF
--- a/packages/agent/src/bytestream.test.ts
+++ b/packages/agent/src/bytestream.test.ts
@@ -33,6 +33,12 @@ const CR = 0x0d;
 const LF = 0x0a;
 const ETB = 0x17;
 
+// A Beckman AU result record as it arrives off the wire, between its STX and ETX: fixed-width
+// space-delimited fields, `r` separating the result groups, and not a control byte in sight.
+// Sample id and values are invented; only the column layout is load-bearing.
+const AU_RESULT_RECORD =
+  'D 000101 0001         100001    E0                                                                 005   140r 006  0.90r 008    72r 009   110r 011    38r 012  10.5r 096 0 0 0r ';
+
 describe('Byte Stream', () => {
   beforeAll(async () => {
     console.log = vi.fn();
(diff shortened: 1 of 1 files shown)


packages/agent/src/bytestream.test.ts
```diff
@@ -33,6 +33,12 @@ const CR = 0x0d;
 const LF = 0x0a;
 const ETB = 0x17;
 
+// A Beckman AU result record as it arrives off the wire, between its STX and ETX: fixed-width
+// space-delimited fields, `r` separating the result groups, and not a control byte in sight.
+// Sample id and values are invented; only the column layout is load-bearing.
```
mattwiller:
```suggestion
// A Beckman AU result record as it arrives off the wire, between its STX and ETX: fixed-width
// space-delimited fields with`r` separating the result groups, and no control bytes
```

packages/agent/src/bytestream.test.ts
```diff
@@ -664,6 +670,33 @@ describe('Byte Stream', () => {
...
+
+      const app = new App(medplum, agentId, LogLevel.INFO);
+      await app.start();
+
+      const [client] = await connectCollecting(agentPort);
+      client.write(Buffer.concat([Buffer.from([STX]), Buffer.from(AU_RESULT_RECORD, 'utf-8'), Buffer.from([ETX])]));
+
+      await waitFor(() => bodies.length > 0, 1000, 'transmit request');
+      // Bots dispatch on the leading record type, so a surviving STX — or a hex-encoded body —
+      // costs them the `D ` they parse on.
```
mattwiller: What does "costs them the `D ` they parse on" mean?

mattwiller (approved):

packages/agent/src/bytestream.test.ts
ThatOneBro: Good question, clarified the comment

PR #10239, opened 2026-08-18
build: override @medplum/core for @metriport/shared

ThatOneBro:
`@metriport/shared@0.37.0` declares `@medplum/core: ^2.2.10`, so npm was
resolving a nested registry copy of `@medplum/core@2.2.10` under it, alongside
the workspace build. This adds a nested override so that dependency dedupes
onto `packages/core`:

```json
"overrides": {
  "@metriport/shared": {
    "@medplum/core": "5.1.30"
  },
```

`package-lock.json` loses exactly one entry —
`node_modules/@metriport/shared/node_modules/@medplum/core` — and
`@metriport/shared` now resolves `@medplum/core` through the root
`node_modules/@medplum/core` link. `npm ci --dry-run` plans a single
`@medplum/core 5.1.30`.

### Risk

Low. `@metriport/shared`'s shipped `dist/` never imports `@medplum/core` — 26
references to `@medplum/fhirtypes`, zero to core — so the 2.x to 5.x jump has no
runtime surface in that package.

Unrelated and left alone: `@metriport/api-sdk` still pins a nested
`@medplum/fhirtypes@2.2.10`, which is the package metriport actually imports.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10987,24 +10987,6 @@
         "zod": "^3.22.1"
       }
     },
-    "node_modules/@metriport/shared/node_modules/@medplum/core": {
-      "version": "2.2.10",
-      "resolved": "https://registry.npmjs.org/@medplum/core/-/core-2.2.10.tgz",
-      "integrity": "sha512-03JhETSAtOOlMOG6U9DkgqiSWGfY2b63MlYGWwR8sxft66hblY5tIxuLt2FmW8zRwlABC93xEzyoVpU8CHeYzw==",
-      "dev": true,
-      "license": "Apache-2.0",
-      "engines": {
-        "node": ">=18.0.0"
-      },
-      "peerDependencies": {
-        "pdfmake": "^0.2.5"
-      },
-      "peerDependenciesMeta": {
-        "pdfmake": {
-          "optional": true
-        }
-      }
-    },
     "node_modules/@metriport/shared/node_modules/uuid": {
       "version": "11.1.1",
       "resolved": "https://registry.npmjs.org/uuid/-/uuid-11.1.1.tgz",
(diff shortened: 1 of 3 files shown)


codyebberson:
> scripts/set-medplum-version.mjs rewrites dependencies, devDependencies,
> peerDependencies, and optionalDependencies — not overrides. So this pin
> has to be bumped by hand alongside the monorepo version. If it is left behind,
> the pin no longer matches the workspace version and npm quietly reinstalls
> registry @medplum/core at the pinned version as a nested duplicate, undoing
> the dedupe without any error. Worth folding overrides into that script, or
> switching this entry to a range, as a follow-up.

`set-medplum-version.mjs` is only used in `build-agent.yml` though, right?  I'm not sure I understand what that script does?  Regardless, I don't think it matters in this case.

I think `scripts/prepare-release.sh` is the more important file, and I'm pretty sure that we need to add another `sed` command in there to update the override in the root `package.json`.

@claude - please review and check whether `prepare-release.sh` will handle this.

ThatOneBro:
Done in 16532a78f9 — one more `sed` over the root `package.json` after the `examples/`/`packages/` sweeps, so the override tracks the workspace version on every release.

On the lockfile caveat Claude raised: not a real concern. The pin exists only to point `@metriport/shared` at our workspace build, and now that it always equals the workspace version, npm resolves it to the local `packages/core` link rather than a published release — there is nothing for `npm install` to fetch, so the release commit does not need one.

Also checked the regex only hits the override: the `--filter=@medplum/...` entries under `scripts` do not match (the pattern needs a quote immediately before `@medplum/`), and `"version"` is still handled by the existing line above.

codyebberson (approved):

PR #10243, opened 2026-08-18
Database migrations for Terminology i18n

mattwiller:
In order to support filtering on display strings in languages with diacritic marks, the database will need to store normalized unicode strings (using NFC normalization).  This adds the necessary SQL function for stripping diacritics at comparison time (`medplum_unaccent()`) and backfills existing Coding rows to ensure that normalization is complete before we migrate queries over


packages/server/src/fhir/sql.ts
```diff
@@ -1376,6 +1376,13 @@ export const TokenArrayToTextFn: SqlFunctionDefinition = {
     AS $function$SELECT e'\x03'||array_to_string($1, e'\x03')||e'\x03'$function$`,
 };
 
+export const MedplumUnaccentFn: SqlFunctionDefinition = {
+  name: 'medplum_unaccent',
+  createQuery: `CREATE FUNCTION medplum_unaccent(text)
```
codyebberson, line 1381: Is this only used in the db migration?  is there a way we can keep it scoped to the db migration?

packages/server/src/migrations/schema/v113.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import type { PoolClient } from 'pg';
+import * as fns from '../migrate-functions';
+
+// prettier-ignore
+export async function run(client: PoolClient): Promise<void> {
+  const results: { name: string; durationMs: number }[] = []
+  // Hand-added: medplum_unaccent() below depends on this extension. Mirrors the btree_gist line in v111.ts
+  await fns.query(client, results, `CREATE EXTENSION IF NOT EXISTS unaccent`);
+  await fns.query(client, results, `CREATE FUNCTION medplum_unaccent(text)
```
codyebberson: Is this the same as the definition in `packages/server/src/fhir/sql.ts`?  do we need it twice?

packages/server/src/fhir/sql.ts
mattwiller: This is going to be used in the eventual query from `$expand`, and is referenced by the index expression

The-Alchemist:
@claude review

The-Alchemist:
@claude don't change any code, just a question: should we mark `medplum_unaccent` as PARALLEL?

The-Alchemist (approved):

PR #10245, opened 2026-08-19
fix(bots): report undeployed bots instead of "Function not found"

### LINKED ISSUE #6551: Better error handling for bot code that has not been deployed
When trying to run a bot that has not been compiled or deployed the error states that there is a missing function handler.  The error should be more clear and indicate that a redeploy or recompile is needed.

asharaali:
Executing a bot whose Lambda was never deployed surfaced AWS's raw error rather than the actual problem:

```
Function not found: arn:aws:lambda:...:function:medplum-bot-lambda-<id>
```

That names the symptom, not the cause, and gives no indication that the fix is to deploy the bot.

## What changed

`runInLambda` caught the invoke error and passed the message straight through. `ResourceNotFoundException` is now recognised and reported as:

```
Bot is not deployed (AWS Lambda function "medplum-bot-lambda-<id>" not found). Deploy the bot with Bot/$deploy and try again.
```

`runInLambdaStreaming` had the same gap, so both paths now share a single `normalizeLambdaExecutionError`. Every other error is passed through unchanged.

## Verification

- `packages/server/src/cloud/aws/execute.test.ts` — 12 passed, including two new cases: `ResourceNotFoundException` produces the deploy message, and unrelated errors are untouched.
- Full `packages/server` suite run against this branch and against `main`: 124 failing on both, 4256 passing here vs 4253 on `main` — the three added tests, no new failures. The pre-existing failures are terminology tests that need post-deploy migrations, which `test.config.json` disables.

Fixes #6551

### DIFF
--- a/packages/server/src/cloud/aws/execute.test.ts
+++ b/packages/server/src/cloud/aws/execute.test.ts
@@ -1,6 +1,11 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { InvokeCommand, LambdaClient, ListLayerVersionsCommand } from '@aws-sdk/client-lambda';
+import {
+  InvokeCommand,
+  LambdaClient,
+  ListLayerVersionsCommand,
+  ResourceNotFoundException,
+} from '@aws-sdk/client-lambda';
 import { badRequest, ContentType } from '@medplum/core';
 import type { Bot } from '@medplum/fhirtypes';
 import type { AwsClientStub } from 'aws-sdk-client-mock';
@@ -13,7 +18,7 @@ import { initApp, shutdownApp } from '../../app';
 import { getConfig, loadTestConfig } from '../../config/loader';
 import { getBinaryStorage } from '../../storage/loader';
 import { initTestAuth } from '../../test.setup';
-import { getLambdaFunctionName } from './execute';
+import { getLambdaFunctionName, normalizeLambdaExecutionError } from './execute';
 
 const app = express();
 let accessToken: string;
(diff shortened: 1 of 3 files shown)


techdavidy:
@claude review

techdavidy:
@claude using bot.name would make be better than `medplum-bot-lambda-##`

asharaali:
@techdavidy Good call — `medplum-bot-lambda-<id>` is an internal detail that means nothing to whoever is running the bot. Pushed in 532bde5.

One heads up: the bot reported that it committed and pushed this back on the 25th, but nothing actually landed — it can't push to my fork. So the PR sat unchanged. Sorry for the lag on my end.

**Before**
```
Bot is not deployed (AWS Lambda function "medplum-bot-lambda-<id>" not found). Deploy the bot with Bot/$deploy and try again.
```

**After**
```
Bot "My Bot" is not deployed. Deploy the bot with Bot/$deploy and try again.
```

`normalizeLambdaExecutionError` now takes the `Bot` rather than the resolved function name, and builds the message from `bot.name ?? bot.id` — `name` is optional in FHIR, same fallback pattern as `subscription.ts:598`. Both call sites (`runInLambda`, `runInLambdaStreaming`) pass the bot; the function name is still what goes to `FunctionName` on the invoke command, only the user-facing message changed.

Tests updated, plus a new case covering the unnamed-bot fallback.

Worth noting this also settles the review's concern about custom external functions: since the message no longer names the Lambda at all, it reads correctly whether the function is medplum-managed or customer-managed via `bot-external-function-id`.

Locally: `execute.test.ts` 13 passed, `executestreaming.test.ts` 8 passed, `tsc --noEmit` and eslint clean.

techdavidy:
Thanks. Mind testing it and posting screenshots please

asharaali:
Sure. There's no UI on this path — it's a server-side `$execute` error string — so the closest thing to a screenshot is the assertion and the suite. Both are below.

**The behaviour, end to end.** `POST /fhir/R4/Bot/{id}/$execute` against a bot whose Lambda was never created, with AWS stubbed to throw `ResourceNotFoundException`:

```
HTTP 400
issue[0].details.text = 'Bot "Test Bot" is not deployed. Deploy the bot with Bot/$deploy and try again.'
```

Before this PR the same request returned `Function not found`.

**`packages/server` — `execute.test.ts`:**

```
 ✓ Execute > Submit plain text                                                39ms
 ✓ Execute > Submit FHIR with content type                                    16ms
 ✓ Execute > Submit FHIR without content type                                 16ms
 ✓ Execute > Submit HL7                                                       19ms
 ✓ Execute > Execute without code                                             29ms
 ✓ Execute > Unsupported runtime version                                      49ms
 ✓ Execute > Get function name                                                 1ms
 ✓ Execute > Execute by identifier                                            17ms
 ✓ Execute > Returned non-OK OperationOutcome marks execution as failure      20ms
 ✓ Execute > Execute bot that has not been deployed                           15ms
 ✓ normalizeLambdaExecutionError > Missing function reports the deploy operation
 ✓ normalizeLambdaExecutionError > Missing function falls back to the bot id when unnamed
 ✓ normalizeLambdaExecutionError > Other errors are passed through

 Test Files  1 passed (1)
      Tests  13 passed (13)
```

The last four are new in this PR — one end-to-end `$execute` case and three unit cases, including the `bot.id` fallback for a bot with no `name` (it's optional in FHIR).

**`executestreaming.test.ts`** — 8 passed, covering the streaming call site, which had the identical gap and now shares the helper.

`tsc --noEmit` clean, and eslint clean on `execute.ts`, `executestreaming.ts` and `execute.test.ts`.

Happy to add a case for anything I've missed — in particular, if you'd rather the message stayed silent about `Bot/$deploy` for customer-managed functions (`bot-external-function-id`), that's a one-line change and I'll make it.

techdavidy:
@asharaali we can test this using our app from packages/app to save and execute a bot.

PR #10246, opened 2026-08-19
feat(server): create Goal resources from PlanDefinition.goal in $apply

### LINKED ISSUE #7268: Add Goal resource to PlanDefinition$apply operation
https://build.fhir.org/plandefinition-operation-apply.html
https://www.medplum.com/docs/api/fhir/resources/plandefinition

FHIR spec and PlanDefinition supports creating Goal resources from the PlanDefinition.goal array. Users can use this to set goals for patients

eastagiletracker:
This PR proposes creating `Goal` resources from `PlanDefinition.goal` when `PlanDefinition/$apply` runs, and referencing them from the resulting `CarePlan` (Fixes #7268). We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/403. You can sign in with your GitHub ID to claim ownership of the project.

## What changed and why

`planDefinitionApplyHandler` walks `PlanDefinition.action` to build a `RequestGroup` and its `Task`s, but never reads `PlanDefinition.goal`. Applying a plan that defines goals therefore returns a `CarePlan` with no `goal` element and creates no `Goal` resources for the subject, so goals authored on a `PlanDefinition` are silently dropped at apply time.

Reproduced on `main` at `76fa156` by applying a `PlanDefinition` that carries two goals (one with `category`, `priority`, `start` and a `target`; one with only the required `description`) to a `Patient`:

```
$ npx vitest run src/fhir/operations/plandefinitionapply.test.ts
 FAIL  src/fhir/operations/plandefinitionapply.test.ts > PlanDefinition apply > Creates Goal resources from PlanDefinition.goal
AssertionError: Target cannot be null or undefined.
 ❯ src/fhir/operations/plandefinitionapply.test.ts:1453:27
    1452|     const carePlan = res3.body as WithId<CarePlan>;
    1453|     expect(carePlan.goal).toHaveLength(2);
```

The change adds a `createGoal` helper that creates one `Goal` per `PlanDefinition.goal`, and sets `CarePlan.goal` to references of what it created. The mapping follows the element definitions: `description` (required on both sides) carries over directly, `category` becomes a single-entry `Goal.category` array, `priority` carries over, `start` becomes `startCodeableConcept`, and each `target` maps `measure`, `detailQuantity`, `detailRange` and `detailCodeableConcept` straight across with `due` (a `Duration`) becoming `dueDuration`. New goals get `lifecycleStatus: 'proposed'` and the applied subject. Two elements are deliberately not mapped because there is no lossless target for them: `PlanDefinition.goal.addresses` is a `CodeableConcept` describing a condition, whereas `Goal.addresses` is a `Reference` to an actual `Condition`/`Observation`, and `Goal` has no counterpart to `goal.documentation`.

The behaviour is additive. `CarePlan.goal` is left undefined when a `PlanDefinition` has no `goal` array, so existing plans apply exactly as before, and no existing element is rewritten.

Verification: two tests are added to `plandefinitionapply.test.ts`. The first fails on the unmodified tree (output above) and passes with the change, asserting every mapped field on both a fully populated goal and a description-only one. The second is a control asserting that a `PlanDefinition` without goals still produces a `CarePlan` with no `goal`. I ran the whole `src/fhir/operations` suite before and after against a seeded local Postgres and Redis: 79 files and 971 tests passing beforehand, 79 files and 973 tests passing afterwards, the two extra passes being the new tests, with no test moving from passing to failing. `npx tsc --noEmit` in `packages/server` is clean, as are `eslint` and `prettier --check` on both touched files. Commits are signed off per the DCO.

## How this was managed

This work was tracked as a story on a board imported from this repository's own issues and pull requests: [Add Goal resource to PlanDefinition$apply operation](https://eastagiletracker.com/projects/403/stories/360904) on the [Medplum board](https://eastagiletracker.com/projects/403), which holds 9,392 stories mapped from your issues, pull requests and milestones.

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/packages/server/src/fhir/operations/plandefinitionapply.test.ts
+++ b/packages/server/src/fhir/operations/plandefinitionapply.test.ts
@@ -6,6 +6,7 @@ import type {
   ActivityDefinition,
   CarePlan,
   Encounter,
+  Goal,
   OperationOutcome,
   Organization,
   Patient,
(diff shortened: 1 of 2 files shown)


The-Alchemist:
@claude review

PR #10251, opened 2026-08-19
[React-Scheduling] Style more calendar statuses

noahsilas:
There may be scenarios where users of these components want to display slots/appointments in these statuses; let's be less opinionated in this component, and make these filters the responsibility of the caller.

### DIFF
--- a/packages/react-scheduling/src/Calendar/Calendar.module.css
+++ b/packages/react-scheduling/src/Calendar/Calendar.module.css
@@ -13,6 +13,15 @@
     background-color: var(--cal-background-color);
   }
 
+  /**
+   * Appointment styles per-status
+   * - Happy-path statuses (pending, booked, arrived, fulfilled) do not get an
+   *   explicit status annotation after the title
+   * - Unusual or actionable statuses get color coding (pending, cancelled)
+   */
+  :global(.appointment.proposed) {
+    .eventTitle::after { content: ' (proposed)'; }
+  }
 
   :global(.appointment.pending) {
     --cal-color: light-dark(var(--mantine-color-blue-8), var(--mantine-color-blue-6));
@@ -22,11 +31,31 @@
     border: 1px solid var(--cal-color);
   }
 
+  :global(.appointment.waitlist) {
+    .eventTitle::after { content: ' (waitlist)'; }
+  }
+
+  :global(.appointment.checked-in) {
+    .eventTitle::after { content: ' (checked in)'; }
+  }
+
   :global(.appointment.cancelled) {
+    .eventTitle::after { content: ' (cancelled)'; }
     --cal-color: light-dark(var(--mantine-color-black), var(--mantine-color-white));
     --cal-background-color: light-dark(var(--mantine-color-red-3), var(--mantine-color-red-8));
   }
 
+  :global(.appointment.noshow) {
+    .eventTitle::after { content: ' (no show)'; }
+  }
+
+  :global(.appointment.entered-in-error) {
+    .eventTitle::after { content: ' (entered in error)'; }
+  }
+
+  /**
+   * Slot styles per-status
+   */
   :global(.slot.free) {
     --cal-color: light-dark(var(--mantine-color-dark-7), var(--mantine-color-gray-1));
     --cal-background-color: light-dark(var(--mantine-color-green-1), var(--mantine-color-green-9));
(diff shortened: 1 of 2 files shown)


The-Alchemist:
@claude review

examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
```diff
@@ -129,7 +129,18 @@ export function ScheduleDetails(props: ScheduleDetailsProps): JSX.Element | null
...
+  const finalSlots = useMemo(() => {
+    const filtered = (slots ?? []).filter((slot) => slot.status !== 'entered-in-error');
+    return mergeOverlappingSlots(filtered);
+  }, [slots]);
+
+  // Omit any "cancelled" appointments
+  const finalAppointments = useMemo(
+    () => (appointments ?? []).filter((appointment) => appointment.status !== 'cancelled'),
+    [appointments]
+  );
```
Filupnot, line 144: Issue with moving this cancelled check from `Calendar.tsx` to here is that now cancelled appointments won't hit `appointmentIndex`, which suppressed "busy" slots, causing the cancelled appointment Slots to render as bare "blocked" slots on the calendar.

Filupnot: https://github.com/user-attachments/assets/dbab5558-6c41-4e40-be3c-6d62693c7c3c

packages/react-scheduling/src/Calendar/Calendar.tsx
```diff
@@ -25,7 +25,7 @@ type ExtendedEvent = { type: 'appointment'; appointment: Appointment } | { type:
 
 function appointmentsToEvents(appointments: Appointment[]): EventInput[] {
   return appointments
-    .filter((appointment) => appointment.status !== 'cancelled' && appointment.start && appointment.end)
```
Filupnot: Since we're removing cancelled filtering here, consumers who use this component without using the upstream filtering in `ScheduleDetails.tsx` will see cancelled appointments that have no styling differentiation between non-cancelled appointments. 

Maybe worth adding a `.cancelled` class for `classes[appointment.status]` to pick up when rendering these? Although I think this is non-blocking and can also be pushed out to a future PR.

Filupnot (changes requested):

examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
noahsilas: I think this might be more correct in practice; if an appointment is canceled with `$cancel` then the Slot resources should be soft-deleted and removed from the calendar, but if you use the FHIR API to cancel the appointment without doing something about the related slot, that time would still be blocked on your calendar for operations like $find/$book.

For instance, I didn't realize that updating an Encounter in the provider UI would cancel an appointment by updating its status, which puts you in this state. This is probably something else we should improve!

packages/react-scheduling/src/Calendar/Calendar.tsx
noahsilas: Good points; the class was already named, but added some styling to it and updated the storybook to demonstrate.

Screenshot 2026-08-21 at 4 04 37 PM

examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
Filupnot: Makes sense, thanks for explanation!

Filupnot (approved):
Looks good to me - approved!

noahsilas:
One of the commits in here was merged with #10274, but the follow up didn't get pulled in to there. Rebased so that this now just has the unmerged styling/story chunks.

The-Alchemist (approved):

packages/react-scheduling/src/Calendar/Calendar.module.css
The-Alchemist: i think this "duplicate selector" warning is worth fixing, imho

The-Alchemist (changes requested):

packages/react-scheduling/src/Calendar/Calendar.module.css
noahsilas: Done in [54423cc](https://github.com/medplum/medplum/pull/10251/commits/54423cc9d6f4a5e78a4bb06c7962ae010d730fa9) 👍

The-Alchemist (approved):

PR #10252, opened 2026-08-19
feat(app): environment-aware favicons

kevinwadeshaw:
This PR swaps the browser favicon by deployment environment so you can tell **production**, **staging**, and **local** tabs apart at a glance. These changes apply to the **Medplum (admin) App** (`packages/app`), **Provider App** (`examples/medplum-provider`), Medplum website (Docusaurus) and all other examples/demos/subdomains _**except**_ for **Foo Medical** and **Storybook** (which retain their unique favicons and do not change based on environment).
 

| Environment | Detected when | Color |
|---|---|---|
| **Production** | hostname is `*.medplum.com` | Grape 8 `#9C36B5` |
| **Staging** | any other host (Vercel, Cloudflare, staging, …) | Yellow 8 `#F08C00` |
| **Local** | `localhost`, `127.0.0.1`, `*.localhost`, `*.local` | Gray 5 `#ADB5BD` |

How this appears in-browser:
Screenshot 2026-08-19 at 10 28 10 AM

## How it works

- Each app ships three `.ico` files in its served publicDir (`static/` for the app, `public/` for the provider): `favicon.ico` (production), `favicon-staging.ico`, `favicon-local.ico`
- `index.html` links the canonical `/favicon.ico` (production)
- A small `setEnvironmentFavicon()` in each entry point swaps the icon link to the staging/local file when the host isn't production. **Production is a no-op** — the default `favicon.ico` is already correct
- **Website (Docusaurus):** a `clientModule` applies the swap and re-applies it on route changes

## Notes

- Detection is hostname-based, so it's host-agnostic (Vercel, Cloudflare, self-hosted) rather than tied to any one provider's env var. A Vite SPA has no server-rewrite layer, and `app.medplum.com` is served with a CSP that blocks inline scripts, so the host check runs as bundled JS
- Removes the dead `packages/app/favicon.ico` at the package root—it was never in Vite's publicDir and apprently never actually served?
- No dependencies were added to the repo—instead we are serving these within individual directories as they are currently handled (just adding new favicons and swap code)

## Test plan

- [x] Provider verified live on `localhost`: favicon swaps to the Gray local `.ico` (`200 image/x-icon`); `/favicon.ico` serves the Grape production icon
- [x] Confirmed each `.ico` carries the correct tint (Grape / Yellow / Gray)
- [x] `tsc` clean on representative apps (foomedical [now excluded], fhircast, hello-world) and the docs clientModule/config
- [ ] Not individually live-tested in all 16 example apps (just a few)—the swap code and `.ico` assets are uniform across them


packages/app/src/index.tsx
```diff
@@ -14,7 +14,21 @@ import { App } from './App';
 import { getConfig } from './config';
 import './index.css';
 
+function setEnvironmentFavicon(): void {
+  const { hostname } = window.location;
+  if (hostname === 'medplum.com' || hostname.endsWith('.medplum.com')) {
```
Filupnot: This looks fine for our own deploys, but if someone forks one of our examples and ships it under a different domain (like emr.theirdomain.com), they'll see the STAGING icon in their production. 

Maybe we should flip this to check a `.env` flag that only our deploys set, so an unrecognized host just means "leave the icons alone" and keeps it as-was? If the goal is to also let consumers use this then we can add scaffolding via `.env.example` for parameters that help define what URL points to what deployment stage.

packages/app/src/index.tsx
```diff
@@ -14,7 +14,21 @@ import { App } from './App';
...
+    return;
+  }
+  const isLocal = /^(localhost|127\.0\.0\.1)$/.test(hostname) || /\.local(host)?$/.test(hostname);
+  const link = document.querySelector<HTMLLinkElement>('link[rel="icon"]');
+  if (link) {
+    link.href = isLocal ? '/favicon-local.ico' : '/favicon-staging.ico';
+  }
```
Filupnot: Might be better to just remove the existing `<link>` and append a fresh one here instead of replacing the existing one. Two reasons: 
- WebKit ignores in-place href mutations on an already-loaded `rel=icon` link, so on Safari the icon can stay purple for the whole session 
- This runs from a deferred module script, so there might be a flash of the purple icon before it flips

packages/app/src/index.tsx
```diff
@@ -14,7 +14,21 @@ import { App } from './App';
 import { getConfig } from './config';
 import './index.css';
 
+function setEnvironmentFavicon(): void {
+  const { hostname } = window.location;
+  if (hostname === 'medplum.com' || hostname.endsWith('.medplum.com')) {
+    return;
+  }
+  const isLocal = /^(localhost|127\.0\.0\.1)$/.test(hostname) || /\.local(host)?$/.test(hostname);
```
Filupnot: Nit: maybe we should also include these, which can also be used in local dev: 
- `0.0.0.0`
- `192.168.*`

examples/medplum-smart-on-fhir-demo/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Medplum SMART on FHIR Demo</title>
-    <link rel="icon" type="image/svg" href="/src/favicon.svg" />
```
Filupnot, line 7: Maybe we should remove the unreferenced `.svg` files that we aren't looking at anymore

Filupnot (changes requested):

packages/app/src/index.tsx
kevinwadeshaw: Done — the swap now runs only when `MEDPLUM_ENVIRONMENT_FAVICON=true`, which only our deploys set, so an unrecognized/fork host leaves the favicon untouched. Documented the flag (commented, off by default) in each `.env.defaults`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/app/src/index.tsx
kevinwadeshaw: Fixed — instead of mutating `href` in place, we now remove the existing `rel=icon` link and append a fresh one so WebKit picks it up. The docs clientModule does the same, and its MutationObserver removes react-helmet's re-added default.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/app/src/index.tsx
kevinwadeshaw: Added `0.0.0.0` and `192.168.*` to the local-host check.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

examples/medplum-smart-on-fhir-demo/index.html
kevinwadeshaw: Removed the now-unreferenced `favicon.svg` files from `smart-on-fhir-demo` and `task-demo`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/app/src/index.tsx
kevinwadeshaw: lol sorry Claude auto-added these replies 😬 I think these latest changes should address this and your other comments though. 

For this specific comment, I changed the staging favicon to use a .env flag, but this needs to be enabled one-time on each Vercel project to work properly (medplum-app, medplum-provider, medplum-www, example apps). 

Would you be able to add this? It would just be `MEDPLUM_ENVIRONMENT_FAVICON = true`

This is so that forked projects will keep the grape favicon on production (instead of yellow staging favicon). And for Medplum, the hostname picks the color on its own (production = grape, preview = yellow, local = gray).

If this feels like overkill, I can revert to what we had before and forked projects can just see the staging favicon in production—which may not be too terrible since they will likely want to change this anyway?

Filupnot: Honestly, I think what we have now is probably fine, and maybe my original comment's request was slightly overkill; I'm not sure how important this is to forkers (or if they'd even spend the time to set it up in the first place). So I'd say we can keep this as-is for now. :)

packages/docs/docusaurus.config.ts
```diff
@@ -38,7 +38,13 @@ const config: Config = {
     faster: true,
   },
 
-  clientModules: ['./src/clientModules/gtagDevShim.ts'],
+  clientModules: [
+    './src/clientModules/gtagDevShim.ts',
+    // Local dev always tints; deployed builds only when Medplum sets the flag (forks stay untouched).
+    ...(process.env.MEDPLUM_ENVIRONMENT_FAVICON === 'true' || process.env.NODE_ENV !== 'production'
```
Filupnot, line 44: I believe we need to add this to the `build.env` list in `turbo.json`, and then set it in the build step of `staging.yml`, for it to be built by the time this runs. As it is right now I don't think the staging icon will ever render.

Filupnot: LIke this: 
```json
"env": [
  "MEDPLUM_BASE_URL",
  "MEDPLUM_CLIENT_ID",
  "MEDPLUM_ENVIRONMENT_FAVICON",
  "MEDPLUM_INTROSPECTION_URL",
  "MEDPLUM_REGISTER_ENABLED",
  "GOOGLE_CLIENT_ID",
  "RECAPTCHA_SITE_KEY"
]
```

packages/docs/src/clientModules/environmentFavicon.ts
```diff
@@ -0,0 +1,43 @@
...
+
+function resolveEnvironmentFavicon(): string | null {
+  const { hostname } = window.location;
+  if (/^(localhost|127\.0\.0\.1|0\.0\.0\.0)$|^192\.168\.|\.local(host)?$/.test(hostname)) {
+    return '/favicon-local.ico';
+  }
+  if (hostname === 'medplum.com' || hostname.endsWith('.medplum.com')) {
+    return null;
+  }
+  return '/favicon-staging.ico';
```
Filupnot, line 12:
```suggestion
function resolveEnvironmentFavicon(): string | null {
  const { hostname } = window.location;
  if (/^(localhost|127\.0\.0\.1|0\.0\.0\.0)$|^192\.168\.|\.local(host)?$/.test(hostname)) {
    return '/favicon-local.ico';
  }
  return null;
}
```

As it is right now, local dev servers will have the staging icon. I'm assuming we'd rather have this & other unrecognized addresses use the normal production favicon.

Filupnot (approved):
Looks good to me, but probably also worth another dev's set of eyes.

PR #10256, opened 2026-08-19
feat(cdk): add opt-in ECS deployment circuit breaker

hunterrhansen:
Without the deployment circuit breaker, a deployment whose tasks never reach a healthy state retries indefinitely — ECS keeps replacing the failed task and the deployment never concludes, so nothing ever surfaces a failure.

The service stays up on the previous tasks, which is what makes it easy to miss: the symptom is a deploy that appears to hang, not one that fails. We hit this on a stack whose new tasks crash-looped on a config problem; the deployment simply never finished and there was no failure signal to alert on.

## Change

Adds an optional `fargateDeploymentCircuitBreaker` to `MedplumInfraConfig`, mirroring the shape of the existing `fargateAutoScaling`:

```jsonc
{
  "fargateDeploymentCircuitBreaker": {
    "enable": true,
    "rollback": false
  }
}
```

It is absent by default, so existing stacks emit no `DeploymentCircuitBreaker` and behaviour is unchanged.

## Why `rollback` defaults to false

Restoring the previous task definition is not unconditionally safe alongside boot-time schema migrations: a version that migrates the schema and then crashes would roll back to an older server running against a newer schema. Opting into failing fast should not silently opt you into reverting too, so `rollback` is only set when asked for explicitly.

## Tests

Three cases in `packages/cdk/src/index.test.ts`:

- `DeploymentCircuitBreaker` is absent when the config is not set
- enabled with `Rollback: true` when configured
- `Rollback: false` when only `enable` is set

`packages/cdk` suite passes 43/43 locally, and `packages/core` typechecks clean.

### DIFF
--- a/packages/cdk/src/backend.ts
+++ b/packages/cdk/src/backend.ts
@@ -549,6 +549,10 @@ export class BackEnd extends Construct {
       securityGroups: [this.fargateSecurityGroup],
       healthCheckGracePeriod: Duration.minutes(5),
       minHealthyPercent: 100, // 50% is the default
+      // Off unless configured, preserving the previous behaviour.
+      circuitBreaker: config.fargateDeploymentCircuitBreaker?.enable
+        ? { rollback: config.fargateDeploymentCircuitBreaker.rollback ?? false }
+        : undefined,
     });
 
     // Add autoscaling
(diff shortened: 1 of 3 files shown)


hunterrhansen:
Closing this — opened in error on my part, not something we're in a position to maintain upstream. Apologies for the noise.

PR #10260, opened 2026-08-19
Pass healthcheck on reader DB failure

mattwiller:


packages/server/src/healthcheck.ts
```diff
@@ -27,10 +27,16 @@ export async function healthcheckHandler(_req: Request, res: Response): Promise<
     attributes: { ...METRIC_IN_SECS_OPTIONS.attributes, dbInstanceType: 'writer' },
   });
 
+  let postgresReaderOk: boolean | undefined;
   if (hasSeparateReaderPool()) {
     readerConn ??= await getReservedDatabaseConnection(DatabaseMode.READER);
```
Filupnot: Maybe this needs to be in the `try/catch` too?

packages/server/src/healthcheck.ts
```diff
@@ -27,10 +27,16 @@ export async function healthcheckHandler(_req: Request, res: Response): Promise<
...
     startTime = Date.now();
-    await testPostgres(readerConn);
+    try {
+      await testPostgres(readerConn);
```
Filupnot: If this returns `false` is that something that we should care about

Filupnot (approved):

noahsilas (approved):

PR #10261, opened 2026-08-20
Align SCIM user updates with normal project auth permissions

codyebberson:
A SCIM `User` maps onto two Medplum resources: a `ProjectMembership`, which is owned by a single project, and the `User` it references, which is a global record that can be a member of many projects. `PUT` and `PATCH` on `/scim/v2/Users/{id}` verified that the membership belongs to the caller's project, then wrote fields on both resources.

Elsewhere in the server, writes to a `User` require that the `User` itself belong to the caller's project. This change applies that same rule to the SCIM write path, so SCIM resolves to the same permission model as the rest of the server.


codyebberson:
@claude please review

PR #10266, opened 2026-08-20
[Scheduling] Booking form: gather criteria and find a time

### LINKED ISSUE #10275: [Scheduling] Find & Book / MVP

Filupnot:
Addresses #10003 

The pieces from #10144 and #10164 are all merged, but there's still nothing a host application can actually mount. This assembles them into `AppointmentBookingForm`: pick the site and the visit type, name the provider, room and device the visit is held on, and pick a time from `Appointment/$find`.

This is the first of three PRs and it is not usable on its own yet. The next one adds the patient field and the `$book` call itself, and exports the component from `index.ts`. A third reduces the two integrations in `examples/medplum-provider` to an `onBooked` handler each, since the form will own the `$book` block both of them currently duplicate. Then we publish, and the host can embed it.

The form deliberately has no drawer, modal, page, or navigation of its own — the host supplies the surface it sits in, so it can decide where booking happens and how wide it gets. `onToggleTimeFinder` exists for exactly that: the times sit beside the form rather than under it, so a host that puts this in a side panel needs to know when to widen the panel. Every prop is optional, and a host launching from its own calendar can pre-fill the answers it already knows:

```tsx
// The host owns the surface; the form just fills it.
<Drawer opened={opened} onClose={close} size={timesOpen ? 'xl' : 'md'}>
  <AppointmentBookingForm
    defaultLocation={clinic}
    defaultService={visitType}
    defaultStart={dayTheUserClicked}
    onToggleTimeFinder={setTimesOpen}
    onChangeService={shadeCalendarFor}
  />
</Drawer>
```

`defaultStart` seeds the day the time search opens on rather than a chosen time, because only a time `$find` offered can be held — a pre-filled time couldn't survive its own validation. Searching today starts from now rather than local midnight: `$find` treats `start` as a hard floor, and `CalendarDateInput` hands back midnight for the day you click, so picking today would otherwise be answered with times that have already gone.

The site field excludes what declares itself a room or a bed (`physical-type:not=ro,bd`) rather than admitting only what declares itself a site, since the element is optional and a Location whose physical type nobody populated still has to be offered.

The visit type field narrows to the chosen site by asking twice, together: `location=<site>`, plus `location:missing=true` for the visit types tied to no site, which no site can hold exclusively and none can exclude. A reference search needs the element present, so one query cannot ask for both. `_filter=location eq <site> or location pr false` can, and the server implements it, but `matchesSearchRequest` has no `_filter` path at all, and that is what `MemoryRepository` searches with, so against `MockClient` every test and every story would match nothing. Dropping `location` and filtering the results client-side is the other one-request option, and it breaks at any practice with more visit types than a page: the server returns the first 25 by name across every site and the filter empties them. The two responses are disjoint by construction, so merging them is a sort by name, and which of the two searches found a visit type never reaches the list.

All three role fields always render, and each takes its `role` rather than a list of candidates derived from the visit type's schedules. Showing only the roles a visit type configures needs a cheap per-role count, which is a follow-up — deriving the fields from a fetched candidate list instead would reintroduce the `_count=100` ceiling that #10164 was written to remove.

One invariant governs what the form throws away: an answer goes as soon as the thing it depended on changes. A new visit type or site clears the named resources and the chosen time. Changing any resource clears the time alone, because a chosen time is a `$find` proposal carrying its own participants and its resources' Slots, so booking it after swapping the provider would book whoever is named inside the proposal rather than whoever is on screen. `$book` cannot catch that: the proposal is internally consistent, and that time genuinely was free for whoever is in it. The search stays open through a resource change and re-runs for the new set, so the replacement time is one click away rather than behind a reopen that also costs the host the panel width it just widened.

The time is a read-only display rather than an editable field. An earlier draft let you type any time and booked it even when `$find` never offered it, which is booking over occupied time — that's an admin-only capability we haven't built yet, so it's cut rather than gated.

Two things worth calling out for review. `getFindWindowError` refuses a range wider than `$find` will answer, but nothing can reach it through the UI yet: range selection isn't in the date input until #10127, so the range is always one day. It's wired up and unit tested so the rule is already right when the gesture arrives.

And the Storybook stories needed the server stubbed to show anything at all. The role fields search through chained `actor:Practitioner.name` filters, which `MemoryRepository` answers with an empty bundle rather than an error, and `MockClient` has no `$find`. Both are now decorators, so a story reads as "nothing configured" only when that's actually true.

image

image

The `SiteAsymmetryByRole` story below is worth a look: the Room field offers Exam Room B while the Provider field leaves out Dr. Ama Osei, though both sit on the clinic's second floor. That asymmetry is inherited from #10164 rather than decided here — a room is sited by walking `partOf`, a provider only by a `PractitionerRole` naming the site exactly. It reads as a bug, which is why it now has a story.

image

### DIFF
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
@@ -1,10 +1,10 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Alert, Button, Loader, Stack, Text, TextInput } from '@mantine/core';
+import { Alert, Button, Group, Loader, Paper, Stack, Text } from '@mantine/core';
 import type { WithId } from '@medplum/core';
-import { getSchedulingTimezone, normalizeErrorString } from '@medplum/core';
+import { getReferenceString, getSchedulingTimezone, isDefined, normalizeErrorString } from '@medplum/core';
 import type { Appointment, HealthcareService, Location } from '@medplum/fhirtypes';
-import { CalendarDateInput, ResourceInput } from '@medplum/react';
+import { CalendarDateInput, ReferenceDisplay, ResourceInput } from '@medplum/react';
 import { IconCalendarSearch } from '@tabler/icons-react';
 import type { JSX } from 'react';
 import { useCallback, useMemo, useState } from 'react';
@@ -12,11 +12,11 @@ import { AppointmentActorSelect } from './AppointmentActorSelect';
 import { AppointmentDayTimes } from './AppointmentDayTimes';
 import classes from './AppointmentFinder.module.css';
 import type { SchedulingRole } from './AppointmentFinder.roles';
-import { SCHEDULING_ROLES } from './AppointmentFinder.roles';
+import { getActorRoleLabel, SCHEDULING_ROLES } from './AppointmentFinder.roles';
 import type { ActorSelections, ScheduleCandidate } from './AppointmentFinder.schedules';
 import { getActorCombinations, getSelectedCandidates, getSelectionError } from './AppointmentFinder.schedules';
 import type { DateRange } from './AppointmentFinder.times';
-import { endOfDay, getFindWindowError, groupAppointmentsByDay } from './AppointmentFinder.times';
+import { endOfDay, getDurationMinutes, getFindWindowError, groupAppointmentsByDay } from './AppointmentFinder.times';
 import { AppointmentServiceSelect } from './AppointmentServiceSelect';
 import { isServiceKeptAtLocation } from './AppointmentServiceSelect.utils';
 import { useProposedAppointments } from './useProposedAppointments';
@@ -71,8 +71,9 @@ export interface AppointmentBookingFormProps {
  * Choosing a time is deliberately a separate step behind "Find a time". The
  * times depend on every answer above them, so offering them earlier would only
  * show times that are about to change. Only a time the search offered can be
- * chosen: the field displaying it is read-only, so nothing can be booked onto
- * time that was never checked against anybody's availability.
+ * chosen: what it produced is summarised beneath the action, and there is no
+ * field to type into, so nothing can be booked onto time that was never checked
+ * against anybody's availability.
  *
  * @param props - The React props.
  * @returns The form.
@@ -205,23 +206,13 @@ export function AppointmentBookingForm(props: AppointmentBookingFormProps): JSX.
           />
         ))}
 
-        <TextInput
-          label="Date & time"
-          placeholder="Find a time to choose one"
-          // A field rather than plain text, because the chosen time belongs among
-          // the answers — but read-only, so there is nothing to type into.
-          readOnly
-          value={chosen?.start ? formatZonedDateTime(new Date(chosen.start), timezone) : ''}
-        />
-        <Button
-          variant="outline"
-          fullWidth
-          leftSection={<IconCalendarSearch size={16} stroke={1.8} />}
+        <ChosenTime
+          appointment={chosen}
+          timezone={timezone}
+          finding={finding}
           disabled={!service}
-          onClick={toggleFinder}
-        >
-          {finding ? 'Close time finder' : 'Find a time'}
-        </Button>
+          onToggleFinder={toggleFinder}
+        />
 
         {finding && (
           <Stack gap={4}>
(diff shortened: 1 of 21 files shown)


Filupnot:
@claude review this please

noahsilas:
Looking good! Just did a first scan through the code and played with the storybook a bit, and this seems like it is the right direction. I'll do a more comprehensive review shortly, but wanted to flag some weird UX related to how some of these fields are derived from others. If you have made a selection in an attribute like "Provider" but then change "Location" or "Service type" you can end up in some funky states.

- Can we disable those later choices until "service type" has been picked? I think folks will perceive it as a bug today if they try to pick a provider first and there are no options. This hopefully also will help people understand the causal link there.

- If someone changes the top level filters (location, service-type), maybe we should clear any selections they have made that don't fit those? Or at least highlight them somehow?

Here's an example of ending up in a confusing state: we choose one service type and get an eligible provider, then change the service type to one they are not eligible for. At first we correctly see no eligible times, but then we see a card with the ineligible provider listed:

https://github.com/user-attachments/assets/e5f9790e-98b0-4962-b89a-cade13326022

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+          location={location}
+          // From state, not the prop: re-seeding from `defaultService` on a remount
+          // would put back the visit type that remount was clearing.
+          defaultValue={service}
+          onChange={chooseService}
+        />
+
+        {SCHEDULING_ROLES.map((role) => (
+          <RoleField
+            key={`${role}-${roleFieldsKey}`}
```
noahsilas, line 184: The more I build around it the more I want to make AsyncAutocomplete a controlled component 😪 

As written this is a reasonable solution for now 👍

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+          variant="outline"
+          fullWidth
+          leftSection={<IconCalendarSearch size={16} stroke={1.8} />}
+          disabled={!!blockedBy}
+          onClick={onToggleFinder}
+        >
+          {getFinderLabel(searching, !!appointment)}
+        </Button>
+        {/* On the action rather than inside the search: a disabled action cannot
+            open the region that used to carry this, so it would never be read. */}
```
noahsilas: I'm not sure this comment is helpful anymore; seems like it is useful for the AI during the build process but now references "what used to be" which isn't especially legible to me as a reader/reviewer.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+
+/**
+ * The chosen time, and under it the action that found it.
+ *
+ * Above the action, because the form is a column of labelled answers. But no field
+ * at all until there is a time: an empty white input among greyed-out ones reads as
+ * the one thing still fillable.
+ *
+ * An input rather than plain text, deliberately: for a user permitted to override,
+ * this same field in this same place becomes editable and gains a warning.
```
noahsilas: Might be worth calling out that this feature (overrides) isn't implemented yet

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+  const durationMinutes = getDurationMinutes(appointment);
+
+  return (
+    <>
+      {durationMinutes > 0 && `${durationMinutes} min visit`}
+      {actors.map((actor, index) => {
+        const roleLabel = getActorRoleLabel(actor);
+        return (
+          <Fragment key={getReferenceString(actor) ?? actor.display}>
+            {(index > 0 || durationMinutes > 0) && ' · '}
```
noahsilas, line 325: I'm starting to think about a reusable "BulletSeparatedList" component that could handle this with a CSS rule like `li + li::before { content: " · "; }`; not a blocker for this PR though. Just feels like I've seen this kind of checking for where we are in a list a few too many times lately across the codebase

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+  );
+}
+
+/**
+ * Returns the range covering one whole day.
+ * @param date - Any instant during the day.
+ * @returns The day, both ends closed, as `$find` requires.
+ */
+function oneDay(date: Date): DateRange {
+  return { start: date, end: endOfDay(date) };
```
noahsilas: This seems like it should be something like `start: startOfDay(date)` to match the description.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+export interface AppointmentBookingFormProps {
+  /** Pre-fills where the visit is, for a host that already knows. */
+  readonly defaultLocation?: WithId<Location>;
+  /** Pre-fills the visit type, for a deep link or a reschedule. */
+  readonly defaultService?: WithId<HealthcareService>;
+  /**
+   * The day the time search opens on. Defaults to today.
+   *
+   * Seeds the day rather than the time: only a time `$find` offered can be
+   * chosen, so a pre-filled time could not survive its own validation.
```
noahsilas: I think this might be something we should consider the API for; today `$find` doesn't restrict you from asking for appointments in the past. So we may want to ensure that the requested start date is in the future (or at least is set to the current instant so we don't show you slots from earlier today).

We've considered setting some kind of value for "Minimum Booking Notice" which could default to "0" (meaning you can book for "right now") but also might be something like "duration: 30min" meaning that you must book at least 30 minutes out. 

We've got a hook currently in Provider that gives a timestamp at least X minutes into the future and returns a timestamp that is stable for 1min: https://github.com/medplum/medplum/blob/main/examples/medplum-provider/src/hooks/useSchedulingStartsAt.ts

We might want to pull something like that into here too.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+    // free for whoever is named in it.
+    setChosen(undefined);
+  }, []);
+
+  function chooseService(next: WithId<HealthcareService> | undefined): void {
+    setService(next);
+    onChangeService?.(next);
+    // The actors on offer come from the visit type, so what was chosen for the
+    // last one cannot mean anything for this one — nor can a time held on them.
+    // That leaves no provider, which is what closes the search.
```
noahsilas: It's possible that some resources are schedulable for multiple service types, so it could be valid to leave them in place until an incompatible service is chosen. I think that is harder to explain, so I like the idea of starting as written here with a full clear out on service change. Just calling out that this comment might be slightly misleading.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
Filupnot: I agree 😢 

I created #10288 to track eventually doing this

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
Filupnot: Created #10289

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
Filupnot: I updated this function: 
```ts 
function oneDay(date: Date): DateRange {
  const now = new Date();
  const start = date > now ? date : now;
  return { start, end: endOfDay(start) };
}
```

So now we should be starting the `$find` search from the current instant, rather than the beginning of today. It doesn't take into account the "X minutes in the future" piece, but I think that might be something better handled on the API side rather than on the client. Either way, I think this is a piece we can punt for later.

Filupnot: Created #10290 for tracking

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
Filupnot: We don't want to use `startOfDay` because that would begin the search at the very beginning of the day, which would give results in the past. 

Updated this same function in response to your other comment.

noahsilas (approved):

PR #10274, opened 2026-08-21
[Scheduling] Multi-Schedule Calendar UI Widget

noahsilas:
When working with multiple schedulable resources at once (say, a
Practitioner, a Location, and a Device), it can be useful to overlay the
appointments and slots from each Schedule onto a single Calendar widget.

In this commit we extract a chunk of shared Calendar logic into the
internal `<CalendarBase>` component, and then add a new specialized
`<MultiCalendar>` component.

This new component lets you specify multiple "sources", which can
include a color to apply to events from the source, and a `Schedule`
resource that will be provided to the caller in callbacks; this may be
useful if an Appointment was scheduled for multiple resources in the
view and the caller is trying to interact with one in particular.

We also introduce a new Extension that may be added to the Schedule
resource to pin a specific color to a calendar:
```
{
   "url": "https://medplum.com/fhir/StructureDefinition/SchedulingColor"
   "valueString": "blue"
 }
 ```

 This color name is looked up in the curent Mantine theme, and hue 7 of
 that color is used. In the future we may support other color values
 like `#rrggbb` codes, but for now let's keep it simple.

Screenshot 2026-08-20 at 5 40 27 PM


The-Alchemist:
@claude review

The-Alchemist:
@noahsilas 

Claude says:

> 1. Same appointment in multiple sources → duplicate FullCalendar event id (MultiCalendar.tsx:75, :102)
The PR description calls out the case where "an Appointment was scheduled for multiple resources in the view." If a caller puts that appointment into two sources, appointmentsToEvents emits two EventInputs with the same id (appointment.id). FullCalendar treats events sharing an id as the same underlying event, which can lead to surprising rendering/interaction behavior. Consider namespacing the event id by source (e.g. ${source.schedule?.id ?? i}:${appointment.id}) so overlaid duplicates stay distinct, or explicitly documenting that a resource should appear in only one source.

Does this sound like a legitimate concern?

noahsilas:
> @noahsilas
> 
> Claude says:
> 
> > 1. Same appointment in multiple sources → duplicate FullCalendar event id (MultiCalendar.tsx:75, :102)
> >    The PR description calls out the case where "an Appointment was scheduled for multiple resources in the view." If a caller puts that appointment into two sources, appointmentsToEvents emits two EventInputs with the same id (appointment.id). FullCalendar treats events sharing an id as the same underlying event, which can lead to surprising rendering/interaction behavior. Consider namespacing the event id by source (e.g. ${source.schedule?.id ?? i}:${appointment.id}) so overlaid duplicates stay distinct, or explicitly documenting that a resource should appear in only one source.
> 
> Does this sound like a legitimate concern?

It's plausible, but in my testing so far this seems to be working correctly. In the "MutliCalendar > Basic" story on this branch you can see that the appointment with ID "appt-1" is associated with Dr Alice Smith's schedule in Blue, and also the schedule for Location "Room 1" in purple. You can interact with each and see that callbacks fire for the appropriate Appointment and Schedule resources.

We may later start working on drag-and-drop interactions and see that we need these to be made more distinct, but I think that we probably do want them to share some features in the interaction: if you are editing the appointment on one Schedule, it _should_ affect the other entries too.

codyebberson:
We should fix both of these:

* "4. Significant duplication between Calendar.tsx and MultiCalendar.tsx"
* "7. Duplicate test. MultiCalendar.test.tsx has two effectively identical cases "

finnbergquist:
One quick question. Can you help me understand the rationale for `MultiCalendarSource` requiring you to pass slots and appointments into the component with the schedule. Did you consider just passing sources as schedule references, and then allowing MultiCalendar to retrieve slots and appointments?

noahsilas:
> One quick question. Can you help me understand the rationale for `MultiCalendarSource` requiring you to pass slots and appointments into the component with the schedule. Did you consider just passing sources as schedule references, and then allowing MultiCalendar to retrieve slots and appointments?

I thought about it, and decided that I wanted this to be a simpler UI component and not also do the data-fetching. The intent is to offer a higher level component could use `useSchedulingResources()` to fetch the data and render this calendar widget with it. This makes tests / storybook quite clean.

After this morning's rework though, I'm more open to the adjustment, as we've pushed basically all the calendar display logic down into `CalendarBase`, making `MultiCalendar` a very thin wrapper; it might make more sense now for that component to own the full data fetching workflow instead of pushing it up a level?

noahsilas:
@claude please re-review

noahsilas:
> Note on https://github.com/medplum/medplum/pull/5 — unifying the title through CalendarBase.appointmentsToEvents (CalendarBase.tsx:44-66) means the appointment title is now just the patient name; the old Calendar behavior of appending a status suffix for non-standard statuses ( ${status}) is gone. Status is now conveyed via className/color. That's a fine consolidation — just confirm dropping the textual status suffix from the single-schedule Calendar was intended, since it's a behavior change for existing Calendar users.

The textual status is still present, it's just added via CSS `::after` instead of via JS. This makes it a little less testable (now relies on the chromatic UI testing instead of a JS unit test), but it does let us apply some slightly nicer styling if we want to tweak the typography or anything else about the label.

Example: The `status: "noshow"` appointment on this calendar (on May 4 at 9:30am):
Screenshot 2026-08-24 at 2 42 40 PM

codyebberson (approved):

PR #10279, opened 2026-08-21
[Scheduling] Booking form: attach the details and book

### LINKED ISSUE #10275: [Scheduling] Find & Book / MVP

Filupnot:
Stacked on #10266: names the patient, books the appointment, and exports the form so a host can finally mount it.

The patient field is `ResourceInput` composed inline, the same way the site field is, sorted by name then birth date, with the birth date and medical record number under each name. Name first because the list is not always narrowed by a typed name: the first click, or a short prefix, leaves one that only an alphabetical sort orders usefully. Birth date is the tie-break between people who share a name, and one story has two patients called Jordan Reyes for exactly this.

It sits below the action that finds a time, and it is the only field down there. `$find` proposes times on calendars and knows nothing about who the visit is for, so naming a patient cannot change which times are offered; putting the field above the action would place work that cannot affect the result in front of the one control that can, and in a narrow panel it pushes that control out of sight. The free text a practice records about a visit (the reason, an internal note, a patient-facing instruction) was here and is now deferred: review placed it with the customizable business-specific fields a practice persists on an appointment rather than in the MVP. The elements chosen for it are on the record for when it returns: `Appointment.description`, `Appointment.comment`, and `Appointment.patientInstruction`, kept separate because §4.3 asks for the last two separately.

Reading the medical record number is worth a look. `getIdentifierByType` with `MRN_IDENTIFIER_TYPE` answers it whoever issued the identifier, which is the case that needs no configuration. But which identifier is the medical record number is a project's own convention, and there is nothing on an untyped one to recognise it by — so `mrnSystem` names the system for a project whose identifiers carry no `type`. A patient with neither is listed by name and birth date rather than hidden, since hiding a patient from a search is worse than listing one without a number.

The write lives in a wrapper, and the fields live under it. `AppointmentProposalForm` is every field, the search and the chosen time: it builds the proposal and hands it to a required `onBook`, writing and announcing nothing. `AppointmentBookingForm` wraps it and supplies that `onBook`, which posts `Appointment/$book`, announces the written Appointment and each reserved Slot, then reports through `onBooked`. So a host embedding this needs no scheduling API code of its own, which is the point of the component, and what the third PR leans on when it deletes the `$book` block both integrations in `examples/medplum-provider` currently duplicate. `$book` is a custom operation, so `MedplumClient` cannot tell what it changed; announcing it is what refreshes a calendar mounted beside the form.

That split is a reversal on review. This was one component taking both `onBook` and `onBooked`: two props for one outcome, where passing both silently ignored the second, and "not called when `onBook` took it over" was the tell. Neither component offers the other's callback now, so the exclusivity a doc comment had to explain is in the types instead. A host doing something else with the proposal (holding it through `$hold`, or writing it inside a transaction of its own) mounts the proposal form and names an `onBook` that says the write is its own; it writes and announces nothing itself, so such a host owns invalidating its own caches. Two exports do make a host choose, and the names are what answer it: the one called `AppointmentBookingForm` is the one that writes.

A refused booking stays on screen with every answer still filled in. A refusal is usually somebody else taking the time, so the next attempt is one field away rather than a form to fill in again. That holds whichever side attempted the write, since a rejected `onBook` is shown the way the server's own refusal is. Once a booking lands the form stops offering to book until an answer changes: every answer is deliberately still there for the refusal case, and a second click would otherwise take a second slot for the same visit. Changing the time or the patient makes it a different visit, which offers it again. A failure after the write, an `onBooked` callback that throws, is reported where a host's own failures are rather than as the booking's, because the appointment exists by then and a red refusal over one that was written reads as a time still free.

Both components and their props are exported from `index.ts` here. #10266 deliberately exported nothing, so nothing it landed could break a consumer; this is the component's first public API. With the free text gone there is no field the form collects that is not a resource reference or a time, and no required ValueSet binding, so `onBooked` really is the only prop a host must supply.

Two smaller things. `stubChainedActorSearch` was rebuilding its criteria even when nothing was chained, so the patient search — which chains nothing — was going through a transform it had no business in; the original query now passes straight through. And `MockClient` has no scheduling operations, so the stories stand in for the server with a `$book` stub that writes the appointment and a Slot per schedule it is held on. It reserves nothing and checks no availability, which is enough for the flow around it to behave the way it will against a real server.

319 tests pass across the package, 62 of them the form's own and split the way the components are: the proposal form's drive `onBook` directly with no `$book` stub, and the wrapper's cover the `$book` call, the announcements, `onBooked`, and a refusal. The helpers that drive the form to a bookable state moved to `test-utils/bookingForm` rather than being copied into both files. `Basic` mounts the wrapper with `onBooked` and nothing else, which is the zero-configuration case a host should be able to copy, and `UntypedMedicalRecordNumbers` covers `mrnSystem`.

### DIFF
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
@@ -1,50 +1,13 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
-import { Alert, Button, Loader, Stack, Text, TextInput } from '@mantine/core';
 import type { WithId } from '@medplum/core';
-import {
-  createReference,
-  formatDate,
-  getIdentifier,
-  getIdentifierByType,
-  getReferenceString,
-  getSchedulingTimezone,
-  isDefined,
-  MRN_IDENTIFIER_TYPE,
-  normalizeErrorString,
-} from '@medplum/core';
-import type { Appointment, Bundle, HealthcareService, Location, Patient, Slot } from '@medplum/fhirtypes';
-import type { AsyncAutocompleteOption } from '@medplum/react';
-import { CalendarDateInput, ReferenceDisplay, ResourceInput } from '@medplum/react';
+import { isDefined } from '@medplum/core';
+import type { Appointment, Bundle, Slot } from '@medplum/fhirtypes';
 import { useMedplum } from '@medplum/react-hooks';
-import { IconCalendarSearch } from '@tabler/icons-react';
 import type { JSX } from 'react';
-import { Fragment, useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { AppointmentActorSelect } from './AppointmentActorSelect';
-import { AppointmentDayTimes } from './AppointmentDayTimes';
-import classes from './AppointmentFinder.module.css';
-import type { SchedulingRole } from './AppointmentFinder.roles';
-import { getActorRoleLabel, SCHEDULING_ROLES } from './AppointmentFinder.roles';
-import type { ActorSelections, ScheduleCandidate } from './AppointmentFinder.schedules';
-import { getActorCombinations, getSelectedCandidates, getSelectionError } from './AppointmentFinder.schedules';
-import type { DateRange } from './AppointmentFinder.times';
-import { endOfDay, getDurationMinutes, getFindWindowError, groupAppointmentsByDay } from './AppointmentFinder.times';
-import { AppointmentOptionRow } from './AppointmentOptionRow';
-import { AppointmentServiceSelect } from './AppointmentServiceSelect';
-import { isServiceKeptAtLocation } from './AppointmentServiceSelect.utils';
-import { useProposedAppointments } from './useProposedAppointments';
-
-// Excludes what a room is rather than admitting what a site is: `physicalType` is
-// optional, so `physical-type=si,bu` would hide a Location that never declared one.
-const LOCATION_SEARCH_CRITERIA = { _count: '25', _sort: 'name', 'physical-type:not': 'ro,bd' };
-
-// Alphabetical, then by birth date: a short prefix — or the first click, before
-// anything is typed — leaves a list only a name orders usefully, and the birth
-// date is what tells the people sharing one apart.
-const PATIENT_SEARCH_CRITERIA = { _count: '25', _sort: 'name,birthdate' };
-
-// No month-wide scan exists, so every day is offered and the search answers.
-const NO_MARKED_DATES: Date[] = [];
+import { useCallback } from 'react';
+import type { AppointmentProposalFormProps } from './AppointmentProposalForm';
+import { AppointmentProposalForm } from './AppointmentProposalForm';
 
 /** What a booking wrote, as `Appointment/$book` returned it. */
 export interface AppointmentBooking {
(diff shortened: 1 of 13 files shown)


packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -211,6 +331,24 @@ export function AppointmentBookingForm(props: AppointmentBookingFormProps): JSX.
...
+          name="patient"
+          label="Patient"
+          placeholder="Search patients by name"
+          required
+          searchCriteria={PATIENT_SEARCH_CRITERIA}
+          defaultValue={defaultPatient}
+          itemComponent={patientItem}
+          onChange={setPatient}
+        />
+        <TextInput
```
finnbergquist: I feel like it is a little awkward to have this above the Time Finder.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -235,6 +373,30 @@ export function AppointmentBookingForm(props: AppointmentBookingFormProps): JSX.
             {windowError && <Alert color="yellow">{windowError}</Alert>}
           </Stack>
         )}
+
+        {/* Below the booking decision: they are the longest fields on the form and
+            the least likely to change what can be booked. */}
+        <Textarea
```
finnbergquist: Can we get rid of these for now.  I feel like they fit into the same category as customizable business-specific fields that will get persisted on the appointment.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -413,6 +575,95 @@ function RoleField(props: RoleFieldProps): JSX.Element {
...
+ * no configuration. `mrnSystem` is for the project whose identifiers carry no
+ * type, where nothing but the system says which one this is.
+ *
+ * @param patient - The patient to read.
+ * @param mrnSystem - The system a project issues medical record numbers under.
+ * @returns The medical record number, or undefined for a patient with none.
+ */
+function getMedicalRecordNumber(patient: WithId<Patient>, mrnSystem: string | undefined): string | undefined {
+  return (
+    getIdentifierByType(patient, MRN_IDENTIFIER_TYPE) ?? (mrnSystem ? getIdentifier(patient, mrnSystem) : undefined)
```
finnbergquist: Thoughts on making this a generic identifier lookup without requiring the MRN system?

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
finnbergquist: We know we will need to come back to this regardless

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
Filupnot: I'm not completely sure what you mean by "generic identifier lookup," can you please elaborate? 

Right now `mrnSystem` is not required, it's just a fallback for identifiers that don't have `identifier.type = <generic MRN type>`, and matches based on `identifier.system` instead.

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
Filupnot: Had this moved to below but just forgot to push :)

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
finnbergquist: Ahh, I misunderstood. Right now, we aren't searching by MRN, just displaying inline with Patients. 

There was an ask for this patient lookup to search either name **or** mrn (stored as an identifier). To support that, we might want to do something like:
```
medplum.searchResources('Patient', {
  _filter: `(name co "${input}" or identifier co "${input}")`,
  _count: '25',
  _sort: 'birthdate',
})
```

I'm not sure what extra work that introduces to get this to work with ResourceInput tho

Filupnot: Ahh ok. This would take some work: 
* This isn't possible with `ResourceInput` today because that can only search based on one field - for Patient it's hard-coded to `name`, so on the frontend there isn't a way right now to say "search this value for name _or_ mrn"
* On the server side the partial search is only supported for a handful of specified elements (like name, email, telecom, etc). Identifier isn't on that so only exact matching would work. 

I'd say we should table searching by MRN for a future PR, after this MVP is out, unless we think it's a blocker for them.

codyebberson:
@claude please review

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
finnbergquist: Not a blocker. Good to know. Thx!

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -25,20 +38,41 @@ import { useProposedAppointments } from './useProposedAppointments';
 // optional, so `physical-type=si,bu` would hide a Location that never declared one.
 const LOCATION_SEARCH_CRITERIA = { _count: '25', _sort: 'name', 'physical-type:not': 'ro,bd' };
 
+// Birth date rather than name: the list is already narrowed by the name that was
+// typed, so what orders it usefully is the thing that tells those people apart.
+const PATIENT_SEARCH_CRITERIA = { _count: '25', _sort: 'birthdate' };
```
noahsilas: I think that we want to use `_sort: "name,birthdate"` so that the autocomplete still prefers an alphabetical sort and uses the birthdate sort as a fallback on a tie. That would cover cases like the initial click to open of the autocomplete, or when someone typed  a short prefix of a name.

finnbergquist (approved):
Looks good. 

Notes for next pass:
- Design philosophy around slideout drawer may need to be iterated on (see [this discussion](https://medplum.slack.com/archives/C07ACGVH4RE/p1787582505696079?thread_ts=1787262544.340249&cid=C07ACGVH4RE))
- Still need a solution for form customizability so that people can inject their own questions and corresponding mappings form those questions to Appointment fields
- For appointment updates and re-scheduling, I assume we will reuse this component? I see there are lots of pre-fill options here. Nice! Do we see expect this sidebar to be the primary way to perform all CRUD operations on the appointment? _ex. change the patient on appointment, cancel appointment, etc_

packages/react-scheduling/src/AppointmentFinder/AppointmentBookingForm.tsx
```diff
@@ -48,10 +82,27 @@ export interface AppointmentBookingFormProps {
...
+   * owns invalidating its own caches.
+   */
+  readonly onBook?: (proposal: Appointment) => void | Promise<void>;
+  /**
+   * Called with what the booking wrote. Not called when `onBook` took it over.
+   *
+   * The answers stay on screen and the form stops offering to book, until one of
+   * them changes: a host is free to keep this mounted without it writing twice.
+   */
+  readonly onBooked: (booking: AppointmentBooking) => void | Promise<void>;
```
noahsilas: "Not called when `onBook` took it over" suggests to me that there may be some ways to make this better.

My first thought was that we could update the types to make those properties exclusive: passing both `onBook` and `onBooked` could be a type error.

Another idea here might be to have this UI component only support `onBook`. We could then add a wrapping component that only takes `onBooked`, and handles all of the `$book` logic itself.

That decomposition might make some of the tests nicer, and it becomes a little clearer exactly what is needed if you want to create a version of the component that uses $hold (or uses admin privileges to bypass the booking rules by writing raw Appointment+Slot records).

Filupnot: I implemented your second idea - now we have `AppointmentProposalForm` which has a required `onBook` prop, and then now `AppointmentBookingForm` is the wrapper which provides the callback: 
```tsx
return <AppointmentProposalForm {...formProps} onBook={book} />;
```

noahsilas (approved):

PR #10283, opened 2026-08-21
AsyncAutocomplete autosubmit fixes

noahsilas:
While testing some of the components coming up in Scheduling work like #10164 I noticed that when I throttle my network I could get into some seemingly broken states via the "autosubmit" feature built into the `<AsyncAutocomplete>` component we are building on top of.

The main fix is that when autosubmit fires, we now show the value emitted to `onChange` in the component's pills. While in here I took the opportunity to do some cleanup as well.

This branch is best reviewed commit-by-commit. The bulk of lines added are test coverage that Claude generated so I could refactor with a test loop running; these are somewhat duplicative of other tests (such as those for `<ValueSetAutocomplete>`), but having tighter unit tests for this helped me so I've chosen to leave them committed.

### DIFF
--- a/packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
+++ b/packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
@@ -120,12 +120,20 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
           // Remove from the front
           newSelected.shift();
         }
+
+        if (newSelected.length >= maxValues) {
+          // The search input is about to be hidden now that the cap is reached; reset the dropdown
+          // state so it doesn't linger with stale options and no input left to dismiss it.
+          setSearch('');
+          setOptions([]);
+          combobox.closeDropdown();
+        }
       }
 
       onChange(newSelected.map((v) => v.resource));
       setSelected(newSelected);
     },
-    [maxValues, onChange]
+    [maxValues, onChange, combobox, setSearch, setOptions]
   );
 
   const handleTimer = useCallback((): void => {
@@ -226,11 +234,6 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
       if (disabled) {
         return;
       }
-      if (maxValues === 1) {
-        setSearch('');
-        setOptions([]);
-        combobox.closeDropdown();
-      }
       lastValueRef.current = undefined;
       if (val === '$create') {
         setSearch('');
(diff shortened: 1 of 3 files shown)


techdavidy:
@claude please review

techdavidy:
@noahsilas changes look solid. Mind adding a debounce to stop triggering a search on each keystroke please

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
```diff
@@ -91,37 +91,39 @@ export function AsyncAutocomplete<T>(props: AsyncAutocompleteProps<T>): JSX.Elem
...
+      }
 
-        // when maxValues is 0, still fire the onChange when an item is selected
-        if (maxValues === 0) {
-          onChange([item.resource]);
-          return [];
-        }
+      // when maxValues is 0, still fire the onChange when an item is selected
+      if (maxValues === 0) {
+        onChange([item.resource]);
```
techdavidy, line 111: I see several  onChange and setSelected together.
Probably move to a function.

noahsilas:
> @noahsilas changes look solid. Mind adding a debounce to stop triggering a search on each keystroke please

It's a little complicated, but there is machinery in place here to debounce by managing a timer: https://github.com/medplum/medplum/blob/b4ec4c2340a7317796a5a9ad8e14ab03519e29dd/packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx#L201

I explored replacing a bunch of that code with `const deboouncedSearch = useDebouncedValue(search, 100)`, but one interesting effect of the current implementation is that the timer doesn't fire until the user interacts with the autocomplete. This is nice for mounting lots of autocompletes without triggering all their searches at once, but it is more complex than the naive debounce + useEffect. Once you rebuild that responsiveness, you end up with a similar amount of code with more complex dependencies. 😕

packages/react/src/AsyncAutocomplete/AsyncAutocomplete.tsx
noahsilas: This scenario is the one that kept me from doing that; usually the `onChange` call is very in-sync w/ `setSelected`, but in this case where `maxValues` is 0 apparently we allow you to make a selection and call `onChange` with it, but we don't set it into the state?

Honestly seems like a bad choice, but I also can't figure out what the use case for `maxValues: 0` ever would be, so I'm just leaving it as-is for now

techdavidy: maxValues: 0 defeats the purpose of the whole exercise.

techdavidy (approved):

PR #10293, opened 2026-08-24
fix(graphql): preserve OperationOutcome in GraphQL error extensions

### LINKED ISSUE #9981: GraphQL errors discard the OperationOutcome: issue[], expression, and severity collapse into one message string
## Summary

When a resolver fails, the structured `OperationOutcome` is discarded: `issue[]`, `expression`, `severity`, and `code` all collapse into a single `errors[].message` string. This makes per-field validation feedback effectively unavailable to clients that write via GraphQL, even though the same write over REST returns the full structure.

`graphql-js` already has a first-class path for carrying this data (`GraphQLError.extensions`), and Medplum can opt into it with a two-line change to `OperationOutcomeError` — no resolver, handler, or schema changes.

## Current behavior

A profile-validation failure on a `ConditionCreate` mutation (profile requires a matching code/display pair on `clinicalStatus`):

```json
{
  "data": { "ConditionCreate": null },
  "errors": [
    {
      "message": "Constraint fluent-clinical-status-pair not met: clinicalStatus, when present, must use a code from the standard condition-clinical CodeSystem together with that code's matching display text (for example code active with display Active). ({\"fhirpath\":\"coding.where(system = 'http://terminology.hl7.org/CodeSystem/condition-clinical').exists() and ...\"}) (Condition.clinicalStatus)",
      "path": ["ConditionCreate"]
    }
  ]
}
```

The identical write over REST returns everything a client needs:

```json
{
  "issue": [
    {
      "severity": "error",
      "code": "invariant",
      "expression": ["Condition.clinicalStatus"],
      "details": { "text": "Constraint fluent-clinical-status-pair not met: ..." },
      "diagnostics": "{\"fhirpath\":\"coding.where(...)\"}"
    }
  ]
}
```

## Why this matters

The flattened string cannot be reliably parsed back into its parts:

- **Multiple issues are joined with `"; "`** (`operationOutcomeToString`). Invariant text comes from `ElementDefinition.constraint.human`, which is prose and can legitimately contain semicolons — so splitting is ambiguous.
- **`expression` survives only as a trailing ` (path)` parenthetical**, which is ambiguous against invariant prose that itself ends in parentheses (the example above does: `"(for example code active with display Active)"`).
- **`severity` and `code` are gone entirely**, so a client cannot distinguish a warning from an error, or a structural violation from a terminology one.
- **`diagnostics` is inlined into the user-visible message.** R4 defines it as *"a stack dump to help trace the issue or other troubleshooting information"* — so a FHIRPath expression belongs there, but it now surfaces in the only string a client has to show a user.

The practical consequence is that R4's stated purpose for `issue.expression` is unreachable over GraphQL:

> **`OperationOutcome.issue.expression`** — *Requirements:* Allows systems to highlight or otherwise guide users to elements implicated in issues to allow them to be fixed more easily.

Applications that want per-field form validation therefore either keep writes on REST, or string-parse an unstable format.

## Root cause

At `main` (`f25d74f51`):

- **`packages/core/src/outcomes.ts:481`** — `OperationOutcomeError extends Error` with `super(operationOutcomeToString(outcome))`. The outcome is kept only on a custom `.outcome` property, which `graphql-js` has no reason to look at.
- **`packages/fhir-router/src/graphql/graphql.ts:411`** — `resolveByCreate` lets the repository's `OperationOutcomeError` propagate unwrapped (nothing needs changing here — this is what makes the fix so small).
- **`packages/fhir-router/src/graphql/graphql.ts:159`** — `graphqlHandler` returns `[allOk, result]` with no `formatError`/error-mapping step, so whatever `graphql-js` serializes is what the client receives.

Meanwhile `graphql-js` (pinned `17.0.2` in both `packages/fhir-router` and `packages/server`) already lifts `extensions` off a thrown error:

`error/locatedError.js` wraps a resolver-thrown error with `originalError` set to it:

```js
return new GraphQLError(originalError.message, {
  nodes: originalError.nodes ?? nodes, source: originalError.source,
  positions: originalError.positions, path, originalError,
});
```

`error/GraphQLError.js` then copies that error's `extensions` onto itself, and `toJSON()` serializes them:

```js
const underlyingError = originalError ?? (cause instanceof Error ? cause : undefined);
const originalExtensions = isObjectLike(underlyingError?.extensions) ? underlyingError.extensions : undefined;
this.extensions = extensions ?? originalExtensions ?? Object.create(null);
...
if (this.extensions != null && Object.keys(this.extensions).length > 0) {
  formattedError.extensions = this.extensions;
}
```

So the outcome simply needs to live on `.extensions` for it to reach clients.

## Proposed implementation

`packages/core/src/outcomes.ts`:

```diff
 export class OperationOutcomeError extends Error {
   readonly outcome: OperationOutcome;
+  /**
+   * GraphQL error extensions. `graphql-js` copies `originalError.extensions` onto the
+   * `GraphQLError` it creates for a resolver-thrown error and serializes it in `toJSON()`,
+   * so this is what preserves the structured outcome through a GraphQL response.
+   */
+  readonly extensions: { outcome: OperationOutcome };

   constructor(outcome: OperationOutcome, options?: ErrorOptions) {
     super(operationOutcomeToString(outcome), options);
     this.name = 'OperationOutcomeError';
     this.outcome = outcome;
+    this.extensions = { outcome };
   }
 }
```

Resulting response — `message` unchanged for backward compatibility, structure now available alongside it:

```json
{
  "data": { "ConditionCreate": null },
  "errors": [
    {
      "message": "Constraint fluent-clinical-status-pair not met: ... (Condition.clinicalStatus)",
      "path": ["ConditionCreate"],
      "extensions": {
        "outcome": {
          "issue": [
            {
              "severity": "error",
              "code": "invariant",
              "expression": ["Condition.clinicalStatus"],
              "details": { "text": "..." },
              "diagnostics": "{\"fhirpath\":\"...\"}"
            }
          ]
        }
      }
    }
  ]
}
```

`errors[].path` (`["ConditionCreate"]`) plus `extensions.outcome.issue[].expression` (`["Condition.clinicalStatus"]`) then give a client exact per-field attribution, matching what REST already provides.

A test could assert this directly in `packages/fhir-router/src/graphql/graphql.test.ts` — execute a mutation that fails validation and assert `errors[0].extensions.outcome.issue[0].expression`.

## Considerations / open questions

1. **Scope.** Setting `extensions` on the class means every `OperationOutcomeError` carries it, including on REST paths where it is inert (nothing outside `graphql-js` reads `.extensions`). If you'd prefer narrower scope, the alternatives are wrapping in the resolvers (`new GraphQLError(err.message, { originalError: err, extensions: { outcome: err.outcome } })`) or adding an error-mapping step in `graphqlHandler`. The class-level change is the smallest and covers read resolvers too.
2. **Information exposure.** This surfaces the whole outcome, `diagnostics` included, in GraphQL responses — the same information REST already returns for the same request, so no new disclosure, but worth confirming that is the intent. If not, filtering `diagnostics` out of the extension copy would be easy.
3. **HTTP status is unchanged** — still `200` with `errors[]`, per GraphQL convention. Not proposing to change that here.
4. **Naming.** `extensions.outcome` seems the natural key for a FHIR `OperationOutcome`, but happy to follow whatever convention you prefer.

Glad to open a PR with the change plus a test if this direction looks right.

## Environment

- `medplum/medplum` `main` @ `f25d74f51`
- `graphql` `17.0.2` (`packages/fhir-router`, `packages/server`)
- Verified against the published `graphql@17.0.2` package source quoted above.

azrabano23:
Fixes #9981

## Problem
When a GraphQL resolver fails, the structured `OperationOutcome` is discarded: `issue[]`, `severity`, `code`, and `expression` collapse into a single `errors[].message` string. The identical write over REST returns the full outcome, so clients doing per-field validation feedback must either stay on REST or string-parse an unstable format.

## Root cause
`OperationOutcomeError` keeps the outcome only on a custom `.outcome` property. `graphql-js` copies a resolver-thrown error's `extensions` onto the `GraphQLError` it creates (and serializes them in the response), but has no reason to look at `.outcome` — so nothing structured survives.

## Fix
Expose the outcome on `OperationOutcomeError.extensions`. `graphql-js` then serializes it as `errors[].extensions.outcome`, e.g.:

```json
{
  "errors": [{
    "message": "Invalid resourceType",
    "path": ["PatientCreate"],
    "extensions": {
      "outcome": {
        "issue": [{ "severity": "error", "code": "invalid", "details": { "text": "Invalid resourceType" } }]
      }
    }
  }]
}
```

`message` is unchanged, so this is backward compatible; no resolver, handler, or schema changes.

## Tests
- `outcomes.test.ts`: asserts `extensions` carries the constructor's outcome.
- `graphql.test.ts`: a `PatientCreate` mutation with a mismatched `resourceType` now surfaces the structured outcome in `errors[0].extensions.outcome`. Both fail without the fix.

### DIFF
--- a/packages/fhir-router/src/graphql/graphql.test.ts
+++ b/packages/fhir-router/src/graphql/graphql.test.ts
@@ -1109,6 +1109,48 @@ describe('GraphQL', () => {
     ]);
   });
 
+  test('Resolver error preserves OperationOutcome in extensions', async () => {
+    const request = makeSimpleRequest('POST', '/fhir/R4/$graphql', {
+      query: `mutation {
+        PatientCreate(res: { resourceType: "ServiceRequest" }) { id }
+      }`,
+    });
+    const fhirRouter = new FhirRouter();
+    const res = await graphqlHandler(request, repo, fhirRouter);
+    expect(res[0]).toMatchObject(allOk);
+
+    const errors = (res[1] as any).errors;
+    expect(errors[0].message).toStrictEqual('Invalid resourceType');
+    expect(errors[0].extensions).toMatchObject({
+      outcome: {
+        resourceType: 'OperationOutcome',
+        issue: [
+          {
+            severity: 'error',
+            code: 'invalid',
+            details: { text: 'Invalid resourceType' },
+          },
+        ],
+      },
+    });
+  });
+
+  test('Non-OperationOutcome error does not get outcome extension', async () => {
+    const request = makeSimpleRequest('POST', '/fhir/R4/$graphql', {
+      query: `query ($id: ID!) {
+        Patient(id: $id) { id }
+      }`,
+    });
+    const fhirRouter = new FhirRouter();
+    const res = await graphqlHandler(request, repo, fhirRouter);
+    expect(res[0]).toMatchObject(allOk);
+
+    const errors = (res[1] as any).errors;
+    expect(errors).toHaveLength(1);
+    expect(errors[0].message).toContain('non-null type "ID!"');
+    expect(errors[0].extensions?.outcome).toBeUndefined();
+  });
+
   test('Updating Patient Record', async () => {
     const patient = await repo.createResource<Patient>({
       resourceType: 'Patient',
--- a/packages/fhir-router/src/graphql/graphql.ts
+++ b/packages/fhir-router/src/graphql/graphql.ts
@@ -31,6 +31,7 @@ import type {
 } from 'graphql';
 import {
   execute,
+  GraphQLError,
   GraphQLFloat,
   GraphQLID,
   GraphQLInt,
@@ -154,9 +155,38 @@ export async function graphqlHandler(
       operationName,
       variableValues: variables,
     });
+
+    if (result.errors) {
+      result.errors = result.errors.map(attachOperationOutcome);
+    }
   }
 
   return [allOk, result, { contentType: ContentType.JSON }];
+}
+
+/**
+ * Exposes the structured `OperationOutcome` of a resolver-thrown `OperationOutcomeError` in the
+ * GraphQL error extensions.
+ *
+ * By default, only the flattened message string appears in the GraphQL `errors` array. Attaching
+ * the outcome under `extensions.outcome` preserves the full structure (issue, severity, code,
+ * expression, etc.) in the response. Errors not caused by an `OperationOutcomeError` are returned
+ * unchanged.
+ * @param error - The GraphQL error from the execution result.
+ * @returns The GraphQL error with `extensions.outcome` set if caused by an `OperationOutcomeError`.
+ */
+function attachOperationOutcome(error: GraphQLError): GraphQLError {
+  if (!(error.originalError instanceof OperationOutcomeError)) {
+    return error;
+  }
+  return new GraphQLError(error.message, {
+    nodes: error.nodes,
+    source: error.source,
+    positions: error.positions,
+    path: error.path,
+    originalError: error.originalError,
+    extensions: { ...error.extensions, outcome: error.originalError.outcome },
+  });
 }
 
 /**


packages/core/src/outcomes.ts
```diff
@@ -522,10 +522,21 @@ export function assertOk<T>(outcome: OperationOutcome, resource: T | undefined):
...
+   * preserves the structured `OperationOutcome` (severity, code, expression, etc.) in the GraphQL
+   * `errors` array, rather than only the flattened message string.
+   */
+  readonly extensions: { outcome: OperationOutcome };
+
   constructor(outcome: OperationOutcome, options?: ErrorOptions) {
     super(operationOutcomeToString(outcome), options);
     this.name = 'OperationOutcomeError';
     this.outcome = outcome;
+    this.extensions = { outcome };
```
mattwiller: The outcome is already attached as `this.outcome`; is it possible to transform the error into the correct shape only for the required GraphQL code path?

azrabano23:
Good call — moved the transformation into the GraphQL code path: `graphqlHandler` in `packages/fhir-router` now maps execution-result errors whose `originalError` is an `OperationOutcomeError` to a `GraphQLError` carrying `extensions.outcome`, and `OperationOutcomeError` itself is back to untouched. Also added a negative test verifying that errors not caused by an `OperationOutcomeError` (e.g. variable coercion) don't get an `outcome` extension.

PR #10301, opened 2026-08-25
Support chained search in MemoryRepository

### LINKED ISSUE #10258: Support chained search in MemoryRepository
`MemoryRepository` can't execute chained search parameters, and it fails silently rather than loudly. The parser stores any key containing a `.` verbatim and leaves interpretation to the server (`packages/core/src/search/search.ts:234`), so when the matcher looks that code up in the flat search parameter table it misses and rejects every candidate (`packages/core/src/search/match.ts:48`). A chained query against `MockClient` comes back as an empty searchset instead of an error, which means any test built on one passes with nothing to assert against.

#10164 hit this: `searchScheduleCandidates` scopes Schedules by their actor's name and status, so the PR added `packages/react-scheduling/src/test-utils/chainedActorSearch.ts` to strip those filters off the query and reapply them in JS afterward. It works for those three or four queries and gets paging and `total` wrong, and anyone else who writes a chained query gets the silent empty bundle again. Better to implement chained search in `MemoryRepository` and delete the stub, possibly reusing the chain parsing the server already does in `packages/server/src/fhir/search.ts:1837`. Worth doing first either way: make an uninterpretable filter code raise `badRequest`, so the next occurrence fails loudly instead of going green.

noahsilas:
Chained filters like `actor:Practitioner.name` previously matched nothing in MemoryRepository, since it looked the filter's whole code up in the flat search parameter table. Resolve the reference and evaluate the chained parameter against the target resource instead, matching real server semantics (a dangling reference fails the chain even under `:not`).

This lets react-scheduling's Storybook and tests search Schedule by actor directly, so the hand-rolled stubChainedActorSearch test stub can go away. Also registers the Device.status, Location.status, and Practitioner.active search parameters the scheduling stories chain through, which MockClient's curated bundle didn't have.

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>


packages/react-scheduling/src/test-utils/chainedActorSearch.ts
```diff
@@ -1,101 +0,0 @@
...
- * `searchScheduleCandidates` sends, not chained search in general.
- *
- * @param medplum - The client to stub.
- * @returns A function restoring the client's own `search`, for a caller sharing
- *   one client across renders.
- */
-export function stubChainedActorSearch(medplum: MockClient): () => void {
-  const original = medplum.search;
-  const search = medplum.search.bind(medplum);
-  medplum.search = (async (resourceType: 'Schedule', query: QueryTypes, options?: object): Promise<Bundle> => {
```
noahsilas, line 28: As I was working on more complex components that compose these complex searches with other queries (such as finding Appointments/Slots to render onto the Calendar UI), this stub stopped being sufficient. I spent a couple minutes trying to make this bind more tightly so we were only replacing the logic on the actual chained searches, but in the end it felt like that was adding so much complexity it seemed safer/easier to just add a naive chained search implementation to the MemoryRepository.

packages/fhir-router/src/repo.ts
```diff
@@ -747,6 +769,76 @@ export class MemoryRepository extends FhirRepository {
...
+  }
+
+  const left = filter.code.slice(0, dotIndex);
+  const chainedKey = filter.code.slice(dotIndex + 1);
+  const colonIndex = left.indexOf(':');
+  const refCode = colonIndex < 0 ? left : left.slice(0, colonIndex);
+
+  const searchParam = globalSchema.types[resourceType]?.searchParams?.[refCode];
+  // With no explicit `:Type`, only an unambiguous (single-target) reference param can chain.
+  const targetType = colonIndex < 0 ? searchParam?.target?.[0] : left.slice(colonIndex + 1);
```
Filupnot, line 813: If an element is multi-targeted it looks like we just arbitrarily choose the first one. Would it make sense to instead just throw an error in that scenario, and require a defined `:Type:` to be provided instead? I can potentially see developers getting confused if they attempt to run this with something like `Schedule.actor` and they're only getting back `Practitioner` matches.

Filupnot: It looks like we do this on the server in `parseChainLink (packages/server/src/fhir/search.ts:1864)`, maybe we should mirror that here?

```ts
const searchParam = getSearchParameter(resourceType, refCode);
if (!searchParam?.expression || searchParam.type !== 'reference') {
  return undefined;
}

const modifier = colonIndex < 0 ? undefined : left.slice(colonIndex + 1);
let targetType: string;
if (searchParam.target?.length === 1) {
  targetType = searchParam.target[0];
} else if (modifier && searchParam.target?.includes(modifier as ResourceType)) {
  targetType = modifier;
} else {
  throw new OperationOutcomeError(
    badRequest(`Unable to identify next resource type for search parameter: ${resourceType}?${refCode}`)
  );
}
```

packages/fhir-router/src/repo.ts
```diff
@@ -747,6 +769,76 @@ export class MemoryRepository extends FhirRepository {
...
+  const references = evalFhirPath(chain.expression, resource) as (Reference | undefined)[];
+  for (const reference of references) {
+    const [refType, refId] = reference?.reference?.split('/') ?? [];
+    if (refType !== chain.targetType || !refId) {
+      continue;
+    }
+    const target = resources.get(refType)?.get(refId);
+    if (target && matchesSearchRequest(target, chain.chained)) {
+      return true;
+    }
```
Filupnot, line 848: If `chain.chained` itself has a chained filter (like `Observation?subject:Patient.general-practitioner:Practitioner.name=Smith`), then I believe that would end up being looked up in the flat table and return false (result in something like `{resourceType: 'Patient', filters: [{code: 'general-practitioner:Practitioner.name', ...}]}`. The server handles up to 3 links. 

One potential solution could be to extract `searchSync` into a helper function and use it here recursively? We could have a function like this: 
```ts
function matchesRequest(
  resources: Map<string, Map<string, Resource>>,
  resource: Resource,
  request: SearchRequest
): boolean {
  const plainFilters: Filter[] = [];
  const chainedFilters: ChainedFilter[] = [];
  for (const filter of request.filters ?? EMPTY) {
    const chain = parseChainedFilter(request.resourceType, filter);
    if (chain) {
      chainedFilters.push(chain);
    } else {
      plainFilters.push(filter);
    }
  }
  const baseRequest = chainedFilters.length ? { ...request, filters: plainFilters } : request;
  return (
    matchesSearchRequest(resource, baseRequest) &&
    chainedFilters.every((chain) => matchesChainedFilter(resources, resource, chain))
  );
}
```

Then replace this line with this: 
```ts
if (target && matchesRequest(resources, target, chain.chained))
```

packages/fhir-router/src/repo.ts
```diff
@@ -747,6 +769,76 @@ export class MemoryRepository extends FhirRepository {
...
+  // With no explicit `:Type`, only an unambiguous (single-target) reference param can chain.
+  const targetType = colonIndex < 0 ? searchParam?.target?.[0] : left.slice(colonIndex + 1);
+  if (!searchParam?.expression || !targetType) {
+    return undefined;
+  }
+
+  return {
+    expression: searchParam.expression,
+    targetType,
+    chained: parseSearchRequest(targetType, { [chainedKey]: filter.value }),
```
Filupnot, line 822: Should we also be including `filter.operator` here? For situations where someone may pass a `SearchRequest` built in code like this: 
```ts
{code: 'subject:Patient.name', operator: Operator.NOT_EQUALS, value: 'Smith'}
```

Potential solution:
```ts
const chained = parseSearchRequest(targetType, { [chainedKey]: filter.value });
if (filter.operator && chained.filters?.length === 1) {
  chained.filters[0].operator = filter.operator;
}
```

packages/react-scheduling/src/AppointmentFinder/AppointmentActorSelect.stories.tsx
```diff
@@ -23,18 +22,13 @@ export default {
...
- * the stories stand it up themselves and stub the chain the way the tests do.
- *
  * @returns The seeded client.
  */
 async function buildClient(): Promise<MockClient> {
   const medplum = new MockClient();
   for (const resource of SchedulingFixtures) {
     await medplum.createResource(resource);
   }
-  stubChainedActorSearch(medplum);
```
Filupnot, line 37: I'm no longer seeing results in the Storybook for this component. Maybe we need to use `WithFixtures`, or `indexSchedulingSearchParameters()` here?

packages/react-scheduling/src/stories/searchParameters.ts
```diff
@@ -170,6 +170,48 @@ const SCHEDULING_SEARCH_PARAMETERS: Bundle<SearchParameter> = {
...
+        url: 'http://hl7.org/fhir/SearchParameter/Practitioner-active',
+        name: 'active',
+        status: 'draft',
+        description: 'Whether the practitioner record is active',
+        code: 'active',
+        base: ['Practitioner'],
+        type: 'token',
+        expression: 'Practitioner.active',
+      },
+    },
```
Filupnot, line 228: We might also want to add search parameters on Practitioner and Device name? We're currently searching on all five of these in `getActorCriteria` in `AppointmentFinder.schedules.ts:103-114`. 

Specifically these two: 

```ts
{
  resource: {
    resourceType: 'SearchParameter',
    id: 'Practitioner-name',
    url: 'http://hl7.org/fhir/SearchParameter/Practitioner-name',
    name: 'name',
    status: 'draft',
    description: 'A server defined search that may match any of the string fields in HumanName',
    code: 'name',
    base: ['Practitioner'],
    type: 'string',
    expression: 'Practitioner.name',
  },
},
{
  resource: {
    resourceType: 'SearchParameter',
    id: 'Device-device-name',
    url: 'http://hl7.org/fhir/SearchParameter/Device-device-name',
    name: 'device-name',
    status: 'draft',
    description: 'A server defined search that may match any of the string fields in Device.deviceName',
    code: 'device-name',
    base: ['Device'],
    type: 'string',
    expression: 'Device.deviceName.name | Device.type.coding.display | Device.type.text',
  },
},
```

packages/fhir-router/src/repo.ts
noahsilas: Good idea 👍

packages/fhir-router/src/repo.ts
noahsilas: I was thinking that this naive single-chain approach would be a reasonable compromise between what a real FHIR server supports and keeping a simpler implementation. I'd say we should leave this for a follow up down the line, but open to feedback about this!

packages/fhir-router/src/repo.ts
noahsilas: Good call out. As part of a naive initial implementation, I'm going to throw an error when the filter is not `'eq'`.

Filupnot: I like that approach 👍

Filupnot (approved):

mattwiller (approved):

PR #10313, opened 2026-08-25
[Scheduling] Extract `useSchedulingResources` from Provider-App

noahsilas:
We want to use this in more places, let's extract it.

- Added a new `onError` option to the hook that takes a callback to be invoked when fetching resources fails. This decouples the hook from the notifications system used in the Provider example app.

- The `Range` type in Provider was renamed to `DateTimeRange` in React-Scheduling, to avoid conflicting with the native [`Range`](https://developer.mozilla.org/en-US/docs/Web/API/Range) type.

### DIFF
--- a/examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
+++ b/examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
@@ -6,13 +6,12 @@ import type { WithId } from '@medplum/core';
 import { getReferenceString, isReference, isResourceWithId } from '@medplum/core';
 import type { Appointment, HealthcareService, Practitioner, Schedule, Slot } from '@medplum/fhirtypes';
 import { useMedplum, useResourceModified } from '@medplum/react';
-import { Calendar, getEffectiveAvailability } from '@medplum/react-scheduling';
+import { Calendar, getEffectiveAvailability, useSchedulingResources } from '@medplum/react-scheduling';
 import type { JSX } from 'react';
 import { useCallback, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router';
 import { AppointmentDetails } from '../../components/schedule/AppointmentDetails';
 import { CreateVisit } from '../../components/schedule/CreateVisit';
-import { useSchedulingResources } from '../../hooks/useSchedulingResources';
 import type { Range } from '../../types/scheduling';
 import { encounterUrl } from '../../utils/encounter';
 import { showErrorNotification } from '../../utils/notifications';
@@ -39,7 +38,9 @@ export function ScheduleDetails(props: ScheduleDetailsProps): JSX.Element | null
 
   const availableTime = getEffectiveAvailability(healthcareService, schedule);
 
-  const { slots, appointments, loading } = useSchedulingResources([schedule], range);
+  const { slots, appointments, loading } = useSchedulingResources([schedule], range, {
+    onError: showErrorNotification,
+  });
 
   const practitioner = schedule.actor.find((actor) => isReference<Practitioner>(actor, 'Practitioner'));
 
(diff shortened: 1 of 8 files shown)


packages/react-scheduling/src/hooks/useSchedulingResources.ts
```diff
@@ -45,13 +44,23 @@ export interface UseSchedulingAppointmentsResult {
...
+export function useSchedulingSlots(
+  schedules: WithId<Schedule>[],
+  range: DateTimeRange | undefined,
+  options?: { onError?: (error: unknown) => void }
+): UseSchedulingSlotsResult {
   const medplum = useMedplum();
   const [slots, setSlots] = useState<WithId<Slot>[] | undefined>(undefined);
   const [loading, setLoading] = useState(false);
 
+  const handleError = useEffectEvent((error: unknown) => {
```
Filupnot: `useEffectEvent` needs React 19.2, so we might need to update our peer dependency version in our [package.json](https://github.com/medplum/medplum/blob/main/packages/react-scheduling/package.json#L100), which currently lists: 
```json
"react": "^18.0.0 || ^19.0.0"
```

packages/react-scheduling/src/hooks/useSchedulingResources.ts
Filupnot:
```suggestion
    if (scheduleRefsKey.length === 0 || !rangeStart || !rangeEnd) {
```

I believe this would prevent us from potentially running a search on an empty schedule. We're doing the same sort of check for appointments on line 230:

```ts
if (actorRefsKey.length === 0 || !rangeStart || !rangeEnd) {
```

packages/react-scheduling/src/hooks/useSchedulingResources.ts
Filupnot: Also, if `options` or `onError` are undefined, then this error would throw silently. Would it maybe be worth falling back to `console.error(error))` as a fallback?

examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
```diff
@@ -6,13 +6,12 @@ import type { WithId } from '@medplum/core';
...
 import { useMedplum, useResourceModified } from '@medplum/react';
-import { Calendar, getEffectiveAvailability } from '@medplum/react-scheduling';
+import { Calendar, getEffectiveAvailability, useSchedulingResources } from '@medplum/react-scheduling';
 import type { JSX } from 'react';
 import { useCallback, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router';
 import { AppointmentDetails } from '../../components/schedule/AppointmentDetails';
 import { CreateVisit } from '../../components/schedule/CreateVisit';
-import { useSchedulingResources } from '../../hooks/useSchedulingResources';
 import type { Range } from '../../types/scheduling';
```
Filupnot: Should we migrate this to the new `DateTimeRange` too? Maybe worth just deleting this file and replacing all its type references (from a search looks like there are others in `FindPane.tsx`, `CreateVisit.tsx`, and `CreateVisit.test.tsx`)

packages/react-scheduling/src/hooks/useSchedulingResources.ts
noahsilas: Oh, arg, for some reason I thought we had made that bump already. Good call out, I'll stick to our broader supported React features in v18+. 😢

examples/medplum-provider/src/components/schedule/ScheduleDetails.tsx
noahsilas: Yeah, it's probably worth being consistent here. 👍

packages/react-scheduling/src/hooks/useSchedulingResources.ts
noahsilas: If there are no refs, then we will map over an empty array which should not emit any actual queries, but it would be nice to make these consistent. Will do 👍

Filupnot (approved):

techdavidy (approved):

PR #10314, opened 2026-08-25
[Scheduling] Load the appointments and times a set of calendars holds

Filupnot:
Lifts the slot, appointment, and combined schedule-resource hooks out of `examples/medplum-provider` into `@medplum/react-scheduling` and exports them. They are the seam a host uses to load what a set of schedules holds and render its own calendar.

Kept as they were:

- **One search per schedule**, so each schedule's results cache independently rather than one toggle invalidating the whole set.
- **The live-update subscription and its range guard.** A resource created outside the loaded range is not appended, since a refetch would not return it.

Changed on the way out: the hooks return `error` instead of calling the example's `showErrorNotification`, which would have dragged `@mantine/notifications` into the package and made a host's error UI our decision. This matches `useProposedAppointments`. The example notifies from that error at its one call site.

## Testing

- 33 hook tests, ported with the two notification assertions rewritten against `error`, plus one covering the combined hook.
- The provider example's 9 scheduling test files — 118 tests — pass unchanged against the lifted hooks.


Filupnot:
Closed because Noah beat me to the punch: #10313

PR #10316, opened 2026-08-25
[Scheduling] Add SchedulingWorkspace top level component

### LINKED ISSUE #10328: [Scheduling] Top-level workflow component

noahsilas:
This introduces a component that we will be building a full Scheduling workflow into. This lets users mount the entire thing to get some scheduling UI deployed quickly.

This is a rough first pass, with an intention for this component to be an evolving shell for our best Scheduling experience.

Screenshot 2026-08-25 at 4 58 45 PM


Filupnot:
Some other name ideas, keeping the Appointment namespace: 
- `AppointmentScheduler`
- `AppointmentSchedulePage`
- `AppointmentScheduleWorkflow`

Filupnot:
When a room or device is selected, the slot's title text contains the provider's name rather than the room/device name. Perhaps this is intentional, but just wanted to call that out in case it wasn't

image

Filupnot:
As the list of providers/devices/rooms grows, we might want to keep only a few / none of them selected on mount. Looks like right now we have all of the resources visible.

Filupnot:
Overall I think this looks great!

Filupnot:
If it's quick and easy, it would be a quick win to allow collapsing the left-hand sidebar, like we saw in our first consumer's mockup demo. Not a blocker.

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const controller = new AbortController();
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- intentional loading flag
+    setCandidatesLoading(true);
+    Promise.all(
+      SCHEDULING_ROLES.map((role) =>
+        searchScheduleCandidates(medplum, undefined, {
+          role,
+          query: '',
+          signal: controller.signal,
+        })
```
Filupnot, line 71: The default count for this function's page size is 25, which is intentionally narrow because we'd search on user keypresses for typeahead results. For this calendar view, we might want to pass in a greater count - perhaps `count: 100`?

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -0,0 +1,171 @@
...
+          onToggleRoom={(id) => setDeselectedRoomIds((prev) => toggleId(prev, id))}
+        />
+      </div>
+      <div className={classes.calendar}>
+        {error !== undefined && (
+          <Alert color="red" className={classes.alert}>
+            {normalizeErrorString(error)}
+          </Alert>
+        )}
+        <MultiCalendar sources={sources} onRangeChange={setRange} loading={resourcesLoading} />
```
Filupnot, line 172: Eventually we'll likely need to have more fleshed-out handling of time zones, via the `timeZone` prop on this component. Right now I believe that `FullCalendar` will use the browser's timezone, which could be problematic if someone views the calendar from a different region, or if schedules have different timezones associated with them. 

I don't think this is a blocker, but perhaps we should create an issue to track this? I can see it coming up.

packages/react-scheduling/src/colors.ts
```diff
@@ -0,0 +1,37 @@
...
+  'pink',
+  'violet',
+  'blue',
+  'cyan',
+  'lime',
+  'red',
+  'yellow',
+  'grape',
+  'orange',
+] satisfies (keyof MantineThemeColors)[];
```
Filupnot, line 18: We have this same thing locally defined in https://github.com/medplum/medplum/blob/main/packages/react-scheduling/src/MultiCalendar/MultiCalendar.tsx#L33-L45. Perhaps we should just export & use that instead of duplicating it here?

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -0,0 +1,171 @@
...
+    Promise.all(
+      SCHEDULING_ROLES.map((role) =>
+        searchScheduleCandidates(medplum, undefined, {
+          role,
+          query: '',
+          signal: controller.signal,
+        })
+      )
+    )
+      .then(([provider, room, device]) => setCandidatesByRole({ provider, room, device }))
```
Filupnot: Maybe we should add `setError(undefined)` here. Right now we're setting it but never clearing it, so I believe that could cause the `<Alert>` below to never go away on a transient error.

noahsilas:
> When a room or device is selected, the slot's title text contains the provider's name rather than the room/device name. 

It's actually the name of any Patient booked into the appointment, not the provider. I think that still makes sense for rooms/devices.

packages/react-scheduling/src/colors.ts
noahsilas: The intent is to use this utility from that component too, just haven't quite wired it up fully yet 😅

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
noahsilas: Yeah, figuring out some of the timezone handling stuff is going to be a bit tricky for multi-calendar display.

The good news is that even if the calendars/events are in different timezones, they all correctly resolve to whatever real instant they are booked for, and display at that instant on this calendar.

The bad news is that this can get confusing when Schedule A says "I'm available 9a-5p PT", and Schedule B says "I'm available 9a - 5p ET" and they are rendered onto the same calendar widget.

For a first draft, I think showing in the viewer's timezone is OK, but I agree that this is going to want some extra detail attention later.

noahsilas:
> As the list of providers/devices/rooms grows, we might want to keep only a few / none of them selected on mount. Looks like right now we have all of the resources visible.

Yeah, it's definitely going to get unwieldy for large lists of schedulable resources... @kevinwadeshaw and I have already started chatting a bit about "select/deselect all" options for sections, and I think that handling large lists probably does suggest that we want to load only a subset on mounting.

The decision about which calendars to show by default is interesting though - we could arbitrarily pick the top 5 in each section, or we could have some system for "pinning" your favorite calendars, or we could remember your selections from last time in localstorage or something. In any case, I'm leaving that as a future enhancement in another PR.

noahsilas:
@claude please review

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
Filupnot: Looking more, my recommended fix was incomplete- that only runs once because it's dependency is `[medplum]`. 

Claude suggests the following fix:

> ```Clear the error where the successful fetch happens, not where the candidate search happens — e.g. have useSchedulingResources expose error as state it owns (set on reject, cleared on resolve) rather than a fire-and-forget onError callback. That also removes the onErrorRef dance in useSchedulingResources.ts:60-68.```

packages/react-scheduling/src/MultiCalendar/MultiCalendar.tsx
```diff
@@ -62,14 +50,20 @@ export function MultiCalendar(props: MultiCalendarProps): JSX.Element {
...
   const eventSources = useMemo((): FhirEventSource[] => {
     return sources.map((source, i) => {
-      const colorName = [
-        source.color,
-        getExtensionValue(source.schedule, SchedulingScheduleColorURI) as string | undefined,
-      ].find((name) => typeof name === 'string' && Object.hasOwn(theme.colors, name));
-      const color = theme.colors[colorName ?? colors[i % colors.length]][7];
+      let colorName = source.color;
+
+      if (!colorName) {
```
Filupnot, line 55: If `colorName` is defined **and** doesn't link to a real theme color, then we end up skipping the extension lookup and jump to a fallback color. Looks like this was lost from the old behavior. 

Take this scenario: 
|   | old | new |
| --- | --- | --- |
| `color: 'teal' + ext 'grape'` | `teal` | `teal` | 
| `color: undefined + ext 'grape'` | `grape` | `grape` | 
| `color: 'brandPrimary' + ext 'grape'` | `grape` | `FALLBACK_COLORS[i]` |

Suggested:
```ts
let colorName = source.color && Object.hasOwn(theme.colors, source.color) ? source.color : undefined;

if (!colorName) {
  const extColor = getExtensionValue(source.schedule, SchedulingScheduleColorURI);
  if (typeof extColor === 'string') {
    colorName = extColor;
  }
}
```

Filupnot: Also related to colors, I noticed that in the month view the colors all seem to become the same: 

Image

Image

noahsilas: Ah, yeah, that color swatch is the status indicator, not a reference to which calendar the appointment is on. I meant to add background colors on the month view but forgot to come back to this. 😓

noahsilas: Month view color updated here: [b23cfb3](https://github.com/medplum/medplum/pull/10316/commits/b23cfb359dd13c2e32cf60d12537176ccc8f9c93)

Screenshot 2026-08-26 at 4 57 31 PM

Filupnot (approved):

The-Alchemist (approved):

PR #10318, opened 2026-08-26
Use shared Super Admin project in tests

mattwiller:
Instead of creating 60+ Super Admin projects for use in tests, often just to take advantage of elevated access level, the test setup creates one globally-shared project that tests can all use


The-Alchemist:
@claude review

The-Alchemist:
@mattwiller : not sure if "🔴 Shared access token is minted once with a 1-hour lifetime" is a valid concern.  wdyt?

mattwiller:
I'm inclined to ignore that one for now: if our tests are taking an hour to run, we should probably solve that issue first

The-Alchemist (approved):

PR #10325, opened 2026-08-26
[Scheduling] Add booking form to scheduling workspace

### LINKED ISSUE #10328: [Scheduling] Top-level workflow component

Filupnot:
Stacked on #10316

Adds the booking form to the top-level calendar component. Click on the calendar to open the booking form, fill it out to schedule an appointment.

image
image

The form opens in a pane on the right, with its time search seeded to the day that was clicked rather than to today. The pane widens while that search is open, because the times render beside the form rather than under it. Nothing in the workspace puts the new appointment on the calendar: the booking form announces what `$book` wrote, and the calendar beside it is already listening, so it appears without a reload.

Clicking a different day re-opens the form on it and clears what was answered for the old one; clicking around inside the day already open leaves a part-filled form alone. `defaultStart` is read at mount only, so remounting is the only way to move the search day.

`CalendarBase` takes a `selection` prop so the clicked slot stays marked while the pane is open, and turns off FullCalendar's `unselectAuto` (a click into the booking form would otherwise clear the mark). The cost is that clearing it has to go through the calendar ref when the pane closes.

A booking reports what it wrote through `onBooked`, and how to announce it is the host's — the workspace pulls in no notification library of its own. The Storybook story is the host here: it raises a toast naming the visit, which is why the shared preview now renders `<Notifications />` for every story.

Booking is all that is wired up here. Editing, rescheduling and cancelling from the calendar are not in this PR.

### DIFF
--- a/packages/react-scheduling/src/CalendarBase/CalendarBase.module.css
+++ b/packages/react-scheduling/src/CalendarBase/CalendarBase.module.css
@@ -39,7 +39,7 @@
   background-color: light-dark(#0000000a, #ffffff14);
 }
 
-.selectedSlot {
+.selectedRange {
   background-color: light-dark(#228be62e, #4dabf73d);
 }
 
--- a/packages/react-scheduling/src/CalendarBase/CalendarBase.module.css
+++ b/packages/react-scheduling/src/CalendarBase/CalendarBase.module.css
@@ -31,6 +31,18 @@
   cursor: pointer;
 }
 
+.selectableDay {
+  cursor: pointer;
+}
+
+.selectableDay:hover {
+  background-color: light-dark(#0000000a, #ffffff14);
+}
+
+.selectedSlot {
+  background-color: light-dark(#228be62e, #4dabf73d);
+}
+
 /* Put more important title info first */
 .eventTitle { order: 1; }
 .eventTime { order: 2 }
--- a/packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.stories.tsx
+++ b/packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.stories.tsx
@@ -8,13 +8,10 @@ import { SchedulingWorkspace } from './SchedulingWorkspace';
 
 const STORY_FIXTURES = [...SchedulingFixtures, ...CalendarWeekFixtures, ...PatientFixtures];
 
-// `withFindStub` is left to the stories rather than set here: a story's own decorators
-// add to these rather than replacing them, so two stubs would both install and which one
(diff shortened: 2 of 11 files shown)


packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.stories.tsx
```diff
@@ -23,10 +26,25 @@ export default {
...
-  <div style={{ height: 700, padding: '1em' }}>
+  // The workspace fills whatever it is given, so the story hands it the rest of the
+  // viewport rather than a fixed height: the calendar and the booking pane both scroll
+  // inside it, and a short host makes each of them look cramped for reasons of its own.
+  // 72px is the package banner `withSchedulingHeader` puts above every story here.
+  <div style={{ height: 'calc(100vh - 72px)', padding: '1em', boxSizing: 'border-box' }}>
     <SchedulingWorkspace />
   </div>
 );
+Basic.decorators = [withFindStub()];
```
noahsilas: This seems likely to be something we want for all the stories of this component, can we move this to the decorators array up above?

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -135,20 +153,36 @@ export function SchedulingWorkspace(props: SchedulingWorkspaceProps): JSX.Elemen
...
+
+  const finishBooking = useCallback(
+    (booking: AppointmentBooking): void => {
+      closeBooking();
+      showNotification({
+        color: 'green',
+        icon: <IconCalendarCheck size={18} />,
+        title: 'Appointment booked',
+        message: describeBooking(booking.appointment),
+      });
```
noahsilas: Let's avoid requiring using Mantine for notifications from here. Can we emit this as an event like `onBook` or something similar and let the wrapping application decide how to render the notification?

packages/react-scheduling/src/CalendarBase/CalendarBase.module.css
```diff
@@ -31,6 +31,18 @@
...
 
+.selectableDay {
+  cursor: pointer;
+}
+
+.selectableDay:hover {
+  background-color: light-dark(#0000000a, #ffffff14);
+}
+
+.selectedSlot {
```
noahsilas: Can we avoid using "slot" here when we aren't talking about a `Slot` resource? How about `highlightClass` (matching the fullcalendar prop) or `selectedRange` (mirroring our `DateTimeRange` type)?

noahsilas (approved):

PR #10334, opened 2026-08-27
Make Publish Retryable and Manually Invokable

codyebberson:

PR #10310, opened 2026-08-25
Adding draft of wh documentation: tweaks

The-Alchemist:
Added changes to original PR at #10248

PR #10337, opened 2026-08-27
Release Version 5.1.34

medplumbot:
## What's Changed

* fix(server): report entriesProcessed from the legacy batch path and tag it with hostname (#10322)
* test(agent): step the park-backlog yield instead of racing a timer (#10291)
* documentation tweaks (#10310)
* Make Publish Retryable and Manually Invokable (#10334)

**Full Changelog**: https://github.com/medplum/medplum/compare/v5.1.33...v5.1.34


noahsilas (approved):
Note: Some v5.1.33 packages did not successfully publish to NPM, and the publish script was not idempotent so could not be retried. #10334 should allow us to retry publishing on future transient errors.

PR #10339, opened 2026-08-27
[Scheduling] Remove hover, update highlight in `SchedulingWorkspace` 

Filupnot:
Some UI improvements to `<SchedulingWorkspace />`: 
- Don't show hover on mouseover on the calendar 
- Update selected slot on date/time change in the booking form
  - _Don't_ page the calendar when clicking outside the displayed range (e.g. a different week/month) to save network cost of having to load all those appointments on each day click
 
https://github.com/user-attachments/assets/1190c542-62a4-4bdc-9edb-3cd6fcb7fe63

> [!NOTE]
> The video shows time cards in ET but the calendar shows adjusted for PT. This is being tracked in a separate issue.

### DIFF
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
@@ -176,6 +176,8 @@ export function AppointmentProposalForm(props: AppointmentProposalFormProps): JS
     }
   }, [searching, onToggleTimeFinder]);
 
+  // Same shape as above: the ref holds the last time reported, so mounting takes down
+  // no marker the host put up, and clearing a time nobody chose reports no drop.
   const reportedTime = useRef<Appointment | undefined>(undefined);
   useEffect(() => {
     if (reportedTime.current !== chosen) {
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
@@ -19,6 +19,7 @@ import { CalendarDateInput, ReferenceDisplay, ResourceInput } from '@medplum/rea
 import { IconCalendarSearch } from '@tabler/icons-react';
 import type { JSX } from 'react';
 import { Fragment, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import type { DateTimeRange } from '../types';
 import { AppointmentActorSelect } from './AppointmentActorSelect';
 import { AppointmentDayTimes } from './AppointmentDayTimes';
 import classes from './AppointmentFinder.module.css';
@@ -76,6 +77,14 @@ export interface AppointmentProposalFormProps {
   /** Called with the visit type as it changes. */
   readonly onChangeService?: (service: WithId<HealthcareService> | undefined) => void;
   /**
+   * Called with the time chosen, and with `undefined` once it is dropped.
+   *
+   * Every answer above the time drops it — a different day, a different provider —
+   * so a host pointing at the time on a calendar of its own takes the marker down
+   * rather than leaving it on a time nobody chose.
+   */
+  readonly onChangeTime?: (time: DateTimeRange | undefined) => void;
+  /**
    * Performs the booking with the proposal the form assembled.
    *
    * Resolving marks the form booked, so it stops offering to book until an answer
@@ -112,6 +121,7 @@ export function AppointmentProposalForm(props: AppointmentProposalFormProps): JS
     mrnSystem,
     onToggleTimeFinder,
     onChangeService,
+    onChangeTime,
     onBook,
   } = props;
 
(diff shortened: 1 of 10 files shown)


Filupnot:
@claude review

packages/react-scheduling/src/CalendarBase/CalendarBase.tsx
```diff
@@ -222,13 +222,27 @@ export function CalendarBase(props: CalendarBaseProps): JSX.Element {
...
-    if (selectable && !selection) {
-      calendarRef.current?.getApi().unselect();
+    const api = calendarRef.current?.getApi();
+    if (!selectable || !api) {
+      return;
     }
-  }, [selectable, selection]);
+    appliedSelection.current = selection;
+    if (selection) {
+      api.select(selection.start, selection.end);
```
noahsilas: FullCalendar can accept dates as milliseconds since epoch (https://fullcalendar.io/docs/date-parsing#millisecond-time), so I think this could use `startMs` and `endMs` directly to avoid the exhaustive-deps lint exception below.

packages/react-scheduling/src/CalendarBase/CalendarBase.tsx
```diff
@@ -222,13 +222,27 @@ export function CalendarBase(props: CalendarBaseProps): JSX.Element {
...
+  // clearing is its own case only because `unselectAuto` is false.
   const calendarRef = useRef<CalendarRef>(null);
+  const appliedSelection = useRef<DateTimeRange | undefined>(undefined);
+  const startMs = selection?.start.getTime();
+  const endMs = selection?.end.getTime();
   useEffect(() => {
-    if (selectable && !selection) {
-      calendarRef.current?.getApi().unselect();
+    const api = calendarRef.current?.getApi();
+    if (!selectable || !api) {
```
noahsilas: I think if we are receiving a `selection` prop we probably should make the calendar "selectable" so that the prop is respected? What do you think about omitting this condition, and updating the above:

```ts
const selectable = Boolean(onSelectInterval || selection);
```

packages/react-scheduling/src/CalendarBase/CalendarBase.tsx
```diff
@@ -222,13 +222,27 @@ export function CalendarBase(props: CalendarBaseProps): JSX.Element {
 
   const selectable = Boolean(onSelectInterval);
 
-  // Necessary because `unselectAuto` is false
+  // FullCalendar owns the highlight and only a pointer gesture ever sets it, so the
+  // prop has to be pushed in. Keyed on the instants rather than the object: a host
+  // reporting the interval back hands over a fresh pair of `Date`s each time, and
+  // clearing is its own case only because `unselectAuto` is false.
   const calendarRef = useRef<CalendarRef>(null);
+  const appliedSelection = useRef<DateTimeRange | undefined>(undefined);
```
noahsilas: I think I'm not fully grokking the purpose of this ref; it seems like we update it only when the `selection` prop we receive changes, and we use it to gate the `onSelectInterval` callback to avoid invoking it with the same value consecutively. Is that right?

Could we make this more targeted by only using it inside the `select` callback without this effect?

```tsx
        select={(eventInfo) => {
          if (isSameRange(previousOnSelect.current, eventInfo)) {
            return;
          }
          onSelectInterval?.({ start: eventInfo.start, end: eventInfo.end });
          previousOnSelect.current = { start: eventInfo.start, end: eventInfo.end };
        }}
```

Filupnot: That's right. 

If the booking form is open and the user clicks on a slot under a different day on the main calendar, then we clear the form, which uses the `select` callback. If we handle our form-controlled highlighting on the calendar in the same way (as you recommend here), then the host would reset the booking form when a time card is selected. That's why we have this ref here.

I ended up getting rid of the ref via another approach that Claude found: `eventInfo.jsEvent` is only true when the user clicks on the calendar, and not when it's controlled from the calendar API. So we can just check that in the callback instead.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
```diff
@@ -166,6 +176,14 @@ export function AppointmentProposalForm(props: AppointmentProposalFormProps): JS
     }
   }, [searching, onToggleTimeFinder]);
 
+  const reportedTime = useRef<Appointment | undefined>(undefined);
+  useEffect(() => {
+    if (reportedTime.current !== chosen) {
+      reportedTime.current = chosen;
+      onChangeTime?.(toRange(chosen));
+    }
+  }, [chosen, onChangeTime]);
```
noahsilas, line 187: The intent of this ref+effect is a little hard to read; is the idea to call `onChangeTime` at most once per selection, while guarding against an unstable `onChangeTime` callback being passed in as a prop? A comment here might help future readers.

noahsilas (approved):

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
noahsilas: I felt like we've been re-writing this slightly tricky ref handling in too many ways; if the main goal of the ref handling is around unstable `onChangeTime`, check out https://github.com/medplum/medplum/pull/10368 which can pull some of that into a named function that makes the intent clear.

PR #10342, opened 2026-08-27
[Scheduling] Timezone display

Filupnot:
Improves timezone support of our scheduling workspace: 
- If the time picker results are in a different timezone than the viewer's, then include the timezone in the time slot results label.
- If the calendar views one or more schedules that are in a different timezone than the viewer's, then display a notice at the top of the calendar accordingly. 

**Considered but not done**: Allow the user to view the calendar in a different timezone than the one they are currently in. In my opinion this would only lead to confusion when the calendar can change its timezones; it seems better to just keep it consistent to the user's local timezone and adjust the calendar's contents accordingly. Open to thoughts on this. 

These changes can be viewed in Storybook, with the mock data being in east coast timezone: 

image


packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -195,6 +209,7 @@ export function SchedulingWorkspace(props: SchedulingWorkspaceProps): JSX.Elemen
             {normalizeErrorString(displayError)}
           </Alert>
         )}
+        <CalendarTimezoneNotice calendars={timezoneCalendars} at={range?.start} />
```
noahsilas: I think we should put this notice below the calendar rather than above - it feels lower in the hierarchy than the actual information presented in the calendar UI.

packages/core/src/scheduling.ts
```diff
@@ -227,6 +227,27 @@ export function getSchedulingTimezone(
...
+    .flatMap((parameters) => getExtensions(parameters, 'timezone'))
+    .map((subextension) => subextension.valueCode)
+    .filter(isDefined);
+  if (scheduleTimezones.length > 0) {
+    return [...new Set(scheduleTimezones)];
+  }
+
+  const actorTimezone = actor && getExtensionValue(actor, TimezoneExtensionURI);
+  return typeof actorTimezone === 'string' ? [actorTimezone] : [];
+}
```
noahsilas: I'm a bit skeptical of the value here; I think if we don't have an explicit service named, trying to merge across all the services that a schedule shows seems like it is not going to result in the desired outcomes. That feels especially true if we aren't considering the HealthcareService resources which can themselves define the timezone in their scheduling parameters that get inherited into the Schedule parameters.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
```diff
@@ -105,6 +112,48 @@ function getTimezoneOffsetMs(instant: Date, timezone: string): number {
...
+}
+
+/**
+ * Names a timezone the short way it is written beside a time, e.g. "EST" or "GMT+2".
+ * @param at - The instant to read it at, which is what decides between EST and EDT.
+ * @param timezone - IANA timezone identifier. Defaults to the browser's.
+ * @returns The zone's short name.
+ */
+export function formatTimezoneLabel(at: Date, timezone?: string): string {
+  const parts = new Intl.DateTimeFormat(undefined, { timeZone: timezone, timeZoneName: 'short' }).formatToParts(at);
```
noahsilas: I'm personally a fan of `timeZoneName: "shortGeneric"`, which uses identifiers like "ET" instead of "EDT"/"EST".

noahsilas: I think that constructing these instances can be expensive, we may want to memoize this if it is going to be called many times on the page.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
```diff
@@ -105,6 +112,48 @@ function getTimezoneOffsetMs(instant: Date, timezone: string): number {
...
+  return parts.find((part) => part.type === 'timeZoneName')?.value ?? '';
+}
+
+/**
+ * Whether two timezones give the same clock reading.
```
noahsilas: Interesting - is the idea that a TZ like "America/Phoenix" matches "America/Los_Angeles" half the year and so we can hide the TZ info while that's true?

I think I would encourage doing something simpler and always showing the TZ annotation if it is not an exact match. I mainly recommend this because I like the idea of the behavior being easy to explain.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.test.tsx
```diff
@@ -376,10 +377,39 @@ describe('AppointmentProposalForm', () => {
           timeZone: SITE_TIMEZONE,
           hour: 'numeric',
           minute: '2-digit',
+          // The runner is not on the site's clock, so the zone is named beside the time.
```
noahsilas: "on the site's clock" is a bit confusing; I might say "The viewer under the test runner does not share the site's timezone"

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
```diff
@@ -132,12 +137,22 @@ export function AppointmentProposalForm(props: AppointmentProposalFormProps): JS
...
 
   const selectionError = getSelectionError(selections);
-  const windowError = getFindWindowError(range);
 
   // Derived, not a flag: closing is never its own rule, so losing the last provider
   // closes the search however it was lost.
   const searching = finding && !selectionError;
 
+  // The first actor's schedule answers for all of them: every actor in one search
+  // shares the scheduling parameters, or `$find` rejects the request.
```
noahsilas: This is not true for the timezone parameter - we support querying resources with separate timezone information, and apply each Schedule's parameters independently, ensuring that appointments $find returns are real instants that are free in each Schedule's availability.

I do think it's fine to choose the first schedule as an exemplar to determine the TZ info, but I don't want to mislead anyone about how $find works.

packages/react-scheduling/vitest.config.ts
```diff
@@ -5,6 +5,14 @@ import { resolve } from 'node:path';
 import { defineConfig } from 'vitest/config';
 import { medplumAliases } from '../../vitest.config';
 
+/*
+ * Pin the runner's timezone. This package renders times in one zone and reads them in another
+ * throughout, and several tests turn on the two differing, so a suite that inherits whatever zone
+ * the machine is set to is a different suite on a laptop than on CI. UTC is what CI already runs
+ * in, so this only makes a local run agree with it.
+ */
+process.env.TZ = 'UTC';
```
noahsilas: My experiences with trying to tweak this env var once we are in Node land has been a bit mixed. I think it's _probably_ going to work from here in vitest.config.ts, where it should be run early on in initialization.

I think we probably should also try to allow folks to explicitly run tests from different zones; long term I think we probably want CI for this package to run at least twice to help us find TZ related issues:
```sh
TZ=UTC npm test
TZ=ET npm test
```

What do you think about adjusting this so that an explicit TZ variable is respected?

```suggestion
process.env.TZ ??= 'UTC'
```

packages/core/src/scheduling.ts
```diff
@@ -227,6 +227,27 @@ export function getSchedulingTimezone(
...
+ * @param schedule - Schedule to inspect
+ * @param actor - Optional Schedule actor used as a timezone fallback
+ * @returns The distinct IANA timezone identifiers found, in document order
+ */
+export function getScheduleTimezones(schedule: Schedule, actor?: Resource): string[] {
+  const scheduleTimezones = (schedule.extension ?? [])
+    .filter((extension) => extension.url === SchedulingParametersURI)
+    .flatMap((parameters) => getExtensions(parameters, 'timezone'))
+    .map((subextension) => subextension.valueCode)
+    .filter(isDefined);
```
mattlong: The `.filter`, `.map`, and `.flatMap` all result in a lot of memory allocations. We prefer old school for loops and if statements to avoid that.

finnbergquist:
How do other scheduling solutions where the scheduler and the scheduled entity exist in different time zones do this? Genuinely curious... Tricky problem to make the UI easy to follow.

Should we display the browser's TZ somewhere on the screen as well to make it obvious what TZ the calendar is displaying? **Sorry I see this is already there, but I feel like you maybe just need** _Calendar shown in  [tz]_

packages/react-scheduling/vitest.config.ts
mattlong: Is this going to mask any issues with daylight savings or any other oddities that exist in non-UTC timezones?

mattlong:
@claude take a look. Timezones are tricky

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
```diff
@@ -105,6 +112,48 @@ function getTimezoneOffsetMs(instant: Date, timezone: string): number {
...
+export function hasSameClock(left: string | undefined, right: string | undefined, at: Date): boolean {
+  const browser = getBrowserTimezone();
+  return getTimezoneOffsetMs(at, left ?? browser) === getTimezoneOffsetMs(at, right ?? browser);
+}
+
+/**
+ * Whether times in a timezone read the same as on the viewer's own clock.
+ * @param timezone - IANA timezone the times are in, or undefined when it could not be resolved.
+ * @param at - The instant being shown.
+ * @param viewer - The viewer's IANA timezone. Defaults to the browser's, and exists to be passed in.
```
noahsilas:
```suggestion
 * @param viewer - The viewer's IANA timezone. Defaults to the browser's
```

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
```diff
@@ -105,6 +112,48 @@ function getTimezoneOffsetMs(instant: Date, timezone: string): number {
...
+ * @returns True when the zone need not be spelled out.
```
noahsilas: This makes it feel like this function is making a display judgement rather than reporting a fact (which the callers use to make the display judgement).

noahsilas (commented):
I think we need to work through some nuance around how we describe a calendar's TZ in the absence of a selected service type that we can use to concretely identify it; I'd probably suggest exclusively using the Actor's TZ in that case (which we can pretty easily get away with while we only support a single actor per Schedule, and which we expect to match all their service types nearly 100% of the time when it is set).

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
Filupnot: Updated. 

However, worth noting that for non-US timezones the `shortGeneric` string gets a little long. I'd expect this to be OK though for now since we're mainly targeting US customers. I'm not sure what our international policy is right now but I'd assume a better experience for the vast majority of (if not all) our customers is the way to go. 

image

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
Filupnot: image

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
Filupnot: Thanks for clarifying this. I updated to this, hopefully this is more clear: 

```ts
// `$find` applies each Schedule's own parameters, so the actors in one search need not
// agree on a timezone. The first one is taken as the exemplar for what to display:
// the times themselves are real instants either way, only their labelling is at stake.
```

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
Filupnot: Updated: `* @returns True when the times read the same on the viewer's own clock.`

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
Filupnot: I can get behind this, definitely simpler to understand. Updated.

packages/react-scheduling/vitest.config.ts
Filupnot: I actually wasn't aware that this was something that could be set in CI, nice callout. Implemented your suggestion.

Filupnot: No it shouldn't. This is only respected locally; CI's runners already default to UTC, so `??=` (added from Noah's comment below) is a no-op there and this just makes a local run match CI's default. The daylight-savings-time-sensitive tests pass their own zone explicitly, so they see DST regardless of this setting. Added a comment in `vitest.config.ts` explaining this so it's answered in place next time someone wonders the same thing.

packages/core/src/scheduling.ts
Filupnot: Valid point. I updated the logic to take the service's timezone if there's exactly one service present, otherwise falling back to the actor's timezone. This seems to be how the priority list is laid out in our [timezone resolution docs](https://www.medplum.com/docs/scheduling/defining-availability#timezone-resolution). 

I can see a _potential_ argument to just always prioritize the actor's schedule instead in this scenario for consistency, and since we should probably expect them to match most of the time anyways, but for now seems best to keep consistent with our docs.

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -148,6 +154,80 @@ export function SchedulingWorkspace(props: SchedulingWorkspaceProps): JSX.Elemen
     });
   }, [activeCandidates, slots, appointments, colorByScheduleId]);
 
+  // A Schedule naming exactly one service in its `serviceType` is unambiguous about which
+  // service's scheduling parameters apply; one naming several, or none, is not, so those
+  // fall back to the actor below instead of guessing between them.
```
noahsilas: This feels like a lot of complexity to address a pretty out-there edge case (showing a notice when an actor without a TZ extension is related to exactly one HealthcareService that has a TZ scheduling parameter that does not match the viewer's TZ). I'd say we should omit this for now and build it in if/when we need it.

I have a further suspicion that we are going to add tools that let you filter the workspace view based on things like Service Type, at which point code we can actually delve into parameter resolution for every schedule without doing an arbitrary number of queries for services.

packages/react-scheduling/src/SchedulingWorkspace/CalendarTimezoneNotice.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const elsewhere = timezones.some((timezone) => !isViewerTimezone(timezone, viewerTimezone));
+  if (!elsewhere) {
+    return null;
+  }
+
+  const viewerLabel = formatTimezoneLabel(viewerTimezone);
+
+  return (
+    <Group gap={6} wrap="nowrap" align="center" mt="xs" className={className} data-testid="calendar-timezone-notice">
+      <IconInfoCircle size={14} stroke={1.8} />
```
noahsilas: Can we color this icon to match the text it is adjacent to? To my eyes, it stands out a bit too much otherwise.

Filupnot: Updated:
image 
image

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
Filupnot: Agreed, I was thinking a similar thing [earlier](https://github.com/medplum/medplum/pull/10342#discussion_r3883835301) but opted against it because of [docs](https://www.medplum.com/docs/scheduling/defining-availability#timezone-resolution). But now I'm realizing those docs don't really apply in this calendar-view scenario. 

Took out a chunk of code in be0460c, following your suggestion.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.test.ts
```diff
@@ -244,3 +249,89 @@ describe('getFindWindowError', () => {
...
+
+  test('Covers a day that daylight saving made short or long', () => {
+    const hours = (day: Date, timezone: string): number => {
+      const range = getZonedDayRange(day, timezone);
+      return (range.end.getTime() + 1 - range.start.getTime()) / 3_600_000;
+    };
+
+    // Assuming an ordinary 24 hours would lose an hour of one and overrun the other.
+    expect(hours(new Date(2099, 2, 8), EASTERN)).toBe(23);
+    expect(hours(new Date(2099, 10, 1), EASTERN)).toBe(25);
```
noahsilas, line 357: Absolute #nit that I'm just pointing out to inform future thinking: testing timezone information in the future is a little fraught, because TZ information is a political choice. In particular, I have hopes that the US will eventually decide to follow the HI/AZ example and stop observing DST before 2099. 

So, it's safer in tests to use times in the past or near future in tests to reduce that uncertainty. 

EDIT: For a function like this that uses the current date as implicit input, that probably entails stubbing the clock, which adds complexity to the test for absolutely marginal future safety, so I'd actually encourage leaving this one as-is. Our future selves can handle any repercussions from changes to TZ rules.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
```diff
@@ -328,6 +408,39 @@ export interface DateRange {
...
+  const now = new Date();
+  const opens = startOfZonedDay(day.getFullYear(), day.getMonth() + 1, day.getDate(), timezone) ?? day;
+  // If the day is already under way, start from now rather than from the beginning to prevent
+  // getting back times from `$find` that have already passed
+  const start = opens > now ? opens : now;
+
+  const parts = getZonedParts(start, timezone);
+  const nextMidnight = startOfZonedDay(parts.year, parts.month, parts.day + 1, timezone);
+  // The instant before the next day begins
+  return { start, end: nextMidnight ? new Date(nextMidnight.getTime() - 1) : endOfDay(start) };
```
noahsilas: I think we may want to remove `-1` from here; hypothetically an appointment could be for something like "11pm ET - 12am ET", which I think a $find that didn't include that endpoint would not match.

packages/react-scheduling/src/SchedulingWorkspace/CalendarTimezoneNotice.stories.tsx
noahsilas, line 1: I think with the simplifications we've made here, we can probably drop these stories. Maybe we should add a story to the SchedulingWorkspace for "Scheduling from a different timezone" or something similar to show the broader TZ behavior?

packages/react-scheduling/src/SchedulingWorkspace/CalendarTimezoneNotice.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const elsewhere = timezones.some((timezone) => !isViewerTimezone(timezone, viewerTimezone));
+  if (!elsewhere) {
+    return null;
+  }
+
+  const viewerLabel = formatTimezoneLabel(viewerTimezone);
+
+  return (
+    <Group gap={6} wrap="nowrap" align="center" mt="xs" className={className} data-testid="calendar-timezone-notice">
```
noahsilas: As a convention, I would encourage omitting margin-top/bottom from components, and instead controlling it from the caller. An individual component generally doesn't know where it will be embedded (for instance, in this PR this had to change when we repositioned this from the top to the bottom). You can also imagine wanting to put this into something like a `<Stack gap="xs">`, in which case this margin would make the spacing between it and its predecessor too large.

One option would be to make this component take the properties of a Mantine `<Box>`, so that the component using this one could say `<CalendarTimezoneNotice mt="xs" ... />`.

packages/core/src/scheduling.ts
The-Alchemist: we have `flatMapFilter` in `packages/core` now for just this reason now

packages/react-scheduling/src/SchedulingWorkspace/CalendarTimezoneNotice.tsx
Filupnot: Moved to the `.timezoneNotice` css class

packages/react-scheduling/src/SchedulingWorkspace/CalendarTimezoneNotice.stories.tsx
Filupnot: Good call - updated accordingly
image

packages/react-scheduling/src/SchedulingWorkspace/SchedulingWorkspace.tsx
```diff
@@ -148,6 +150,23 @@ export function SchedulingWorkspace(props: SchedulingWorkspaceProps): JSX.Elemen
     });
   }, [activeCandidates, slots, appointments, colorByScheduleId]);
 
+  /*
+   * A calendar's zone is read off its actor alone. A Schedule can also carry a zone per
+   * service in its scheduling parameters, but nothing here names a service to pick between
+   * them, and while a Schedule holds a single actor the actor's own zone is the one those
+   * parameters are expected to agree with anyway.
```
noahsilas: total #nit: I might say "is likely to" instead of "is expected to". We think this is going to be the common case, but anticipate scenarios where they differ (imagine a Practitioner who works in multiple locations on different sides of a TZ boundary).

noahsilas (approved):

PR #10351, opened 2026-08-27
[Scheduling] Name actors by resource name over reference display

Filupnot:
When displaying actor names in the booking form (for example "Dr. Maya Rivera" in the below screenshot), we would prioritize displaying the reference.display, and if not present, fall back to the reference ID. This was done under the false assumption that the reference display would be reliably populated.

A more reliable approach, and what this PR does, is prioritize the actor resource's name over the reference display. This is also low-cost since we've already loaded the actor's resource anyways.

Two places name actors, and each is fixed in its own way.

The slot cards and the "Date & time" description under the chosen time render the actors of a proposed appointment. `$find` copies `Schedule.actor` onto every time it offers, so a proposal names its actors by reference and by whatever `display` the Schedule carried, which may be absent and is never revised. `AppointmentSlotGroup.actors` now carries the actors' own resources instead: the role fields already read them to offer the actors, so `getSelectedActorResources` just hands what they fetched to `groupAppointmentsByDay`, and the components name them with `<ResourceName>`, which resolves the resource where `<ReferenceDisplay>` only ever printed `Reference.display`. Reading the resources back off the client's cache is not an option: a search's cached entries are dropped when its request is aborted, which is what every keystroke in an `AsyncAutocomplete` does.

An actor the form does not have a resource for (nothing in the current selections matches the reference the proposal carries) falls back to that reference, and `<ResourceName>` reads it back rather than printing the Schedule's stale copy of it. `getActorRoleLabel` takes either shape now and reads the type off the resource where there is one.

`getCandidateDisplay` names a schedule candidate, and its fallback order is reversed the same way. It is shared, so this reaches the provider/room/device autocomplete options, the `SchedulingWorkspace` sidebar labels, the actor-combination labels, and the order candidates are offered in. It is not what the slot cards read.

image

### DIFF
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentDayTimes.tsx
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentDayTimes.tsx
@@ -15,6 +15,8 @@ export interface AppointmentDayTimesProps {
   readonly onSelectAppointment: (appointment: Appointment) => void;
   /** IANA timezone the times are read in. Defaults to the browser's. */
   readonly timezone?: string;
+  /** What to call each actor, keyed by reference. */
+  readonly actorNames?: ReadonlyMap<string, string>;
   readonly selected?: Appointment;
 }
 
@@ -28,7 +30,7 @@ export interface AppointmentDayTimesProps {
  * @returns The day's heading and its times.
  */
 export function AppointmentDayTimes(props: AppointmentDayTimesProps): JSX.Element {
-  const { date, groups, onSelectAppointment, timezone, selected } = props;
+  const { date, groups, onSelectAppointment, timezone, actorNames, selected } = props;
 
   return (
     <Stack gap="xs">
@@ -39,6 +41,7 @@ export function AppointmentDayTimes(props: AppointmentDayTimesProps): JSX.Elemen
           key={group.key}
           group={group}
           timezone={timezone}
+          actorNames={actorNames}
           selected={selected}
           onSelectAppointment={onSelectAppointment}
         />
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
@@ -58,8 +58,8 @@ export function getCandidateRole(candidate: ScheduleCandidate): SchedulingRole |
 export function getCandidateDisplay(candidate: ScheduleCandidate): string {
   const actor = getCandidateActor(candidate);
   return (
-    actor.display ??
     (candidate.actorResource && getDisplayString(candidate.actorResource)) ??
+    actor.display ??
     actor.reference ??
     `Schedule/${candidate.schedule.id}`
   );
@@ -444,6 +444,22 @@ async function readLocation(
  */
 export function getSelectedCandidates(selections: ActorSelections): ScheduleCandidate[] {
   return SCHEDULING_ROLES.flatMap((role) => selections[role] ?? []);
+}
+
+/**
+ * Names every chosen actor, keyed by the reference an appointment names them by.
+ * @param selections - What has been chosen.
+ * @returns The name to show for each chosen actor.
+ */
+export function getActorDisplayNames(selections: ActorSelections): Map<string, string> {
+  const names = new Map<string, string>();
+  for (const candidate of getSelectedCandidates(selections)) {
+    const reference = getReferenceString(getCandidateActor(candidate));
+    if (reference) {
+      names.set(reference, getCandidateDisplay(candidate));
+    }
+  }
+  return names;
 }
 
 function toScheduleReference(candidate: ScheduleCandidate): Reference<Schedule> {
--- a/packages/react-scheduling/src/stories/scheduling.ts
+++ b/packages/react-scheduling/src/stories/scheduling.ts
@@ -225,6 +225,19 @@ function buildSchedule(
   };
 }
 
+/**
+ * Strips the name a Schedule copied onto its actor.
+ *
+ * `Schedule.actor.display` is optional, and plenty of real projects never write
+ * it — which is the case where the name has to come from the actor itself.
+ *
+ * @param schedule - The schedule to strip.
+ * @returns The same schedule, naming its actor by reference alone.
+ */
+export function withoutActorDisplay(schedule: WithId<Schedule>): WithId<Schedule> {
+  return { ...schedule, actor: [{ reference: schedule.actor[0].reference }] };
+}
+
 export const DrRiveraSchedule = buildSchedule('schedule-dr-rivera', 'Practitioner/dr-rivera', 'Dr. Maya Rivera');
 export const DrOkaforSchedule = buildSchedule('schedule-dr-okafor', 'Practitioner/dr-okafor', 'Dr. Tunde Okafor');
 export const Ultrasound1Schedule = buildSchedule(
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.roles.ts
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.roles.ts
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: Copyright Orangebot, Inc. and Medplum contributors
 // SPDX-License-Identifier: Apache-2.0
+import { getReferenceString } from '@medplum/core';
 import type { Reference, Schedule } from '@medplum/fhirtypes';
 
 /** Resource types that may appear in `Schedule.actor`. */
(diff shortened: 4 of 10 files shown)


Filupnot:
@claude review please

packages/react-scheduling/src/stories/scheduling.ts
```diff
@@ -225,6 +225,19 @@ function buildSchedule(
...
+/**
+ * Strips the name a Schedule copied onto its actor.
+ *
+ * `Schedule.actor.display` is optional, and plenty of real projects never write
+ * it — which is the case where the name has to come from the actor itself.
+ *
+ * @param schedule - The schedule to strip.
+ * @returns The same schedule, naming its actor by reference alone.
+ */
+export function withoutActorDisplay(schedule: WithId<Schedule>): WithId<Schedule> {
```
codyebberson: Is this function only used in tests?  If so, it should go in tests.

codyebberson (commented):
@claude - please review

It looks like we are accumulating a lot of `useMemo` calls in these components that are not strictly necessary.  Is precomputing and storing `actorNames` the right approach?  Or is there a more straightforward model where we can pass the actor in from `selections`?

Also, why didn't `<ReferenceDisplay value={actor} link={false} />` do the right thing here?  The actors should have been in memory if loaded from `MedplumClient`, right?

Is it possible the fix is more surgical if we add support for more resource types in `getDisplayString` in `packages/core/src/utils.ts` ?

packages/react-scheduling/src/AppointmentFinder/AppointmentDayTimes.tsx
```diff
@@ -15,6 +15,8 @@ export interface AppointmentDayTimesProps {
   readonly onSelectAppointment: (appointment: Appointment) => void;
   /** IANA timezone the times are read in. Defaults to the browser's. */
   readonly timezone?: string;
+  /** What to call each actor, keyed by reference. */
+  readonly actorNames?: ReadonlyMap<string, string>;
```
noahsilas: My intuition is that we should be trying to pass down the fetched Resources so that the display components can make the ultimate determination of how to represent them, probably with something like the `<ResourceName>` component.

That's sort of the opposite strategy of what is happening in this PR, where the display logic is pushed up to the coordinating component, and then threaded down through intermediate layers via the specialized `actorNames` prop.

In the `AppointmentSlotGroup` we have an `actors` attribute which is currently populated with Reference types, I wonder if we can fill that with the actual Resource values that were fetched? That would be a larger structural change than the targeted fix that this PR is making, but I think it probably sets us up for the long term well.

packages/react-scheduling/src/stories/scheduling.ts
Filupnot: Moved to `AppointmentProposalForm.test.tsx`

packages/react-scheduling/src/AppointmentFinder/AppointmentDayTimes.tsx
Filupnot: Made some structural changes around this, but TL;DR, we're now storing both actor references and resources in `SchedulingActorValue`, and using those values directly when displaying names.

`AppointmentSlotGroup.actors` now carries each actor's own resource wherever the form already had it, falling back to the reference `$find` copied off `Schedule.actor`. The actor fields fetch those resources to offer the actors in the first place, so `getSelectedActorResources` just hands them to `groupAppointmentsByDay` without any extra reads. The `actorNames` prop is gone, and the slot cards and the "Date & time" description name actors with `<ResourceName>`, which resolves the resource where `<ReferenceDisplay>` only ever printed `Reference.display`.

Reading the resources back off the client's cache wasn't an option, for what it's worth: a search's cached entries are dropped when its request is aborted, which is what every keystroke in an `AsyncAutocomplete` does.

packages/react-scheduling/src/stories/scheduling.ts
Filupnot: To your above comment:
- Got rid of the precomputed `actorNames` and `useMemo`. Per Noah's suggestion the loaded actor resources now ride on the slot groups, so the components render the actor itself instead of just taking the name as a prop. 
- `<ReferenceDisplay>` doesn't load the resource here, it just prints `Reference.display` and falls back tot he reference string. `<ResourceName>` is the one that we want here since that actually reads the resource, so that's what the cards use now. 
- These resource types were already supported in `getDisplayString`, the issue was just around priority.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
```diff
@@ -28,7 +28,7 @@ import { getActorsKey } from './AppointmentFinder.times';
 export interface ScheduleCandidate {
   readonly schedule: WithId<Schedule>;
   /** The actor itself, when the search was able to include it. */
-  readonly actorResource: WithId<Resource> | undefined;
+  readonly actorResource: SchedulingActorResource | undefined;
```
noahsilas, line 31: Are there situations where we don't get back the actor? I would guess that this could happen if the viewer has an access policy that lets them read a Schedule resource without allowing them to read the related Actor.

My suspicion is that such a scenario is going to fail down the line when calling `$find` - it uses the caller's access to read Schedule.actor to find TZ info. Maybe we could loosen that in scenarios where the HealthcareService or Schedule define the TZ as a scheduling parameter at a higher priority than the TZ extension on the actor, so this might be good future proofing in that way?

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
noahsilas: We could potentially narrow the type here to something like `WithId<Schedule> | WithId<SchedulingActorResource> | undefined`. That would let us rewrite the check below:

```ts
if (resource.resourceType === 'Schedule') {
  schedules.push(resource);
} else {
  // type is now refined to SchedulingActorResource
  actorsByReference.set(...)
}

File: packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts (hunk 3 above)
Comment (Filupnot) [reply packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts:31]: Good catch! I didn't think about this scenario. 

It looks like this access policy 400s today when calling `$find` because it reads `Schedule.actor[0]` before looking at the timezone. The timezone resolution itself should actually be fine since in `getSchedulingTimezone` we fall back to the schedule/service when `actorResource` is undefined (that might've been introduced in #10342 after you left this comment). 

There's also a second gap: On the client we read the actor's timezone when rendering the timezone helptext in the calendar view, so if we can't read the actor then we won't show the timezone banner. I created a separate PR to fix this: #10477 . 

I didn't include either fix in this PR since they seemed unrelated (the line you called out shouldn't affect the behavior, it just narrows the type). No changes made in this PR.

Comment (Filupnot) [reply packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts:31]: Fixed in https://github.com/medplum/medplum/pull/10477

File: packages/react-scheduling/src/actors.ts [hunk 4]
```diff
@@ -11,6 +11,29 @@ import type { Schedule } from '@medplum/fhirtypes';
 export type SchedulingActor = Schedule['actor'][number];
 export type SchedulingActorType = Dereference<SchedulingActor>['resourceType'];
 
+/** A loaded actor resource. Tied to the types `Schedule.actor` admits. */
+export type SchedulingActorResource = WithId<ExtractResource<SchedulingActorType>>;
```
noahsilas: We could simplify by doing this in a slightly different order:

```suggestion
export type SchedulingActorResource = Dereference<SchedulingActor>;
export type SchedulingActorType = SchedulingActorResource['resourceType'];
```

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.schedules.ts
noahsilas: We could potentially narrow the type here to something like `WithId<Schedule> | WithId<SchedulingActorResource> | undefined`. That would let us rewrite the check below:

```ts
if (resource.resourceType === 'Schedule') {
  schedules.push(resource);
} else {
  // type is now refined to SchedulingActorResource
  actorsByReference.set(...)
}
```

Then we wouldn't need `isActorResource`, which also removes the need for a constant that is re-encoding the type information for that guard

packages/react-scheduling/src/AppointmentFinder/AppointmentDayTimes.tsx
noahsilas: Hm. I think that the intent in AsyncAutocomplete is to only abort searches that haven't resolved when the next keystroke comes in, because we are scheduling a more specific search at that moment. I'll take a look at this behavior a little later and see if we need any updates here.

noahsilas (approved):

PR #10352, opened 2026-08-27
Storybook test mode and cleanups

mattlong:
The star of the show is adding `--test` when building for tests. That along with avoiding building storybook twice should save about two minutes when running on GH


ThatOneBro (approved):

The-Alchemist:
@claude review

PR #10353, opened 2026-08-27
Add extended stats on Coding_Property

mattwiller:


mattlong:
@claude take a look

packages/server/src/migrations/schema/v116.ts
```diff
@@ -0,0 +1,27 @@
...
+export async function run(client: PoolClient): Promise<void> {
+  const results: { name: string; durationMs: number }[] = [];
+
+  // "property" and "value" are strongly correlated: a given value is only ever used by a handful of properties,
+  // so estimating the two clauses independently understates the row count of a `property = X AND value = Y`
+  // lookup by orders of magnitude (30x-2000x on the representative dataset)
+  await fns.query(
+    client,
+    results,
+    `CREATE STATISTICS IF NOT EXISTS "Coding_Property_property_value_stat" ON "property", "value" FROM "Coding_Property"`
```
mattlong: Do we need all three types of statistics or just a subset? I'm unsure how much extra space each type takes.

packages/server/src/migrations/schema/v116.ts
```diff
@@ -0,0 +1,27 @@
...
+  );
+
+  // The default target of 100 has ~30% error; increasing to 1000 captures ~450 MCV pairs for ~5% error, at the cost of a slower ANALYZE
+  await fns.query(client, results, `ALTER STATISTICS "Coding_Property_property_value_stat" SET STATISTICS 1000`);
+  await fns.analyzeTable(client, results, 'Coding_Property');
```
mattlong, line 26: analyze should go in a post-deploy; or perhaps all of this should?

mattlong (changes requested):

packages/server/src/migrations/schema/v116.ts
mattwiller: I measured it at a couple seconds even with all of UMLS loaded; which I think should be fine to run at pre-deploy

packages/server/src/migrations/schema/v116.ts
mattwiller: Capturing only `mcv` should work fine, and cuts the `ANALYZE` cost by about 500 ms

mattlong (approved):

PR #10358, opened 2026-08-28
Fixes #10354 - customize CapabilityStatement

### LINKED ISSUE #10354: Enhancement to customize CapabilityStatement
As a consumer building FHIR services on top of Medplum, the ability to restrict [CapabilityStatement](https://www.medplum.com/docs/api/fhir/resources/capabilitystatement) to only publish the FHIR Resources and Resource Interactions the Medplum implementer actually supports is a necessity. Medplum supports the entirety of the FHIR specification, but an implementer of Medplum may only choose to support a subset for several reasons. Here are a couple of examples:

- Not supporting patient-entered data, so blocking CapabilityStatement from advertising create/modify interactions on resources exposed through a consumer-based traffic endpoint.
- Only supporting USCDI-required concepts, so blocking CapabilityStatement from advertising other non-USCDI related resource like those around financial interactions, clinical trials or supply chain workflows.

Overall, the idea would be to configure what is published based on a potential mixture of criteria likely stemming from the server URL path.

codyebberson:
Fixes #10354

## Why

Medplum supports the entirety of FHIR R4, but an implementer may only support a subset. A consumer-facing
endpoint might not accept patient-entered data, or a deployment might only expose USCDI-required concepts. Today
`/fhir/R4/metadata` advertises everything unconditionally, so those deployments over-advertise their conformance.

Two approaches were considered and rejected. Serving a stored `CapabilityStatement` resource as-is is easy to
implement, but the implementer stops receiving updates when Medplum adds resource types, search parameters, or
profiles — ongoing merge conflict hell. Invoking a Bot at startup is maximally flexible, but makes a public
conformance endpoint depend on the project data plane and Lambda availability, and is neither diffable nor
testable alongside the rest of a deployment's infrastructure.

This PR keeps the generated statement as the base and lets the implementer configure deltas, so upgrades keep
flowing through.

## What

A `capabilityStatement` server config setting:

```json
{
  "capabilityStatement": {
    "overlay": { "title": "Acme FHIR API", "publisher": "Acme Health" },
    "includeResourceTypes": ["Patient", "Observation", "DocumentReference"],
    "interactions": {
      "*": ["read", "vread", "search-type"],
      "DocumentReference": ["read", "search-type", "create"]
    },
    "systemInteractions": ["batch"]
  }
}
```

- `overlay` — a partial CapabilityStatement merged over the generated one. Covers the common case of
  customizing only vanity fields such as `id`, `name`, and `title`, without a config key per field.
- `includeResourceTypes` / `excludeResourceTypes` — mutually exclusive resource type filters, mirroring the
  existing `dataWarehouse` settings.
- `interactions` — advertised interactions per resource type. The `*` key sets the default for resource types
  not listed explicitly; an empty array advertises none.
- `systemInteractions` — advertised system interactions (`transaction`, `batch`).
- `supportedProfiles` — set to `false` to omit `supportedProfile`.

Search parameters, profiles, and operations continue to be generated for whatever survives the filters, so a
restricted statement still picks up upstream changes.

## Details

- Adds `MedplumCapabilityStatementConfig` to `MedplumServerConfig`.
- Registers `capabilityStatement` in `objectKeys` (`config/utils.ts`), so the setting can be delivered as a
  single JSON string from AWS SSM Parameter Store, matching `defaultOAuthClients` and `externalAuthProviders`.
- Adds `config/capabilitystatement.ts` with the filtering helpers.
- Applies the filters in `buildResourceTypes` and `buildRest`, and the overlay in `buildCapabilityStatement`.
- Exports `buildCapabilityStatement` so tests can build a statement without the module-level cache.

Defaults are unchanged: with no `capabilityStatement` setting, the generated statement is byte-identical to
today's.

### DIFF
--- a/packages/server/src/config/types.ts
+++ b/packages/server/src/config/types.ts
@@ -251,8 +251,8 @@ export interface MedplumServerConfig {
 export interface MedplumCapabilityStatementConfig {
   /**
    * Partial CapabilityStatement merged over the server generated statement.
-   * Top level fields replace the generated values.
-   * `resourceType` and `rest` cannot be set here; use the filters below to restrict `rest`.
+   * Top level fields replace the generated values wholesale, so setting `rest` here replaces the
+   * generated `rest` entirely; prefer the filters below to restrict it.
    */
   overlay?: Record<string, unknown>;
 
--- a/packages/server/src/config/types.ts
+++ b/packages/server/src/config/types.ts
@@ -279,9 +279,13 @@ export interface MedplumCapabilityStatementConfig {
   systemInteractions?: string[];
 
   /**
-   * Whether to advertise `supportedProfile` for each resource type. Defaults to true.
+   * Controls the advertised `supportedProfile` for each resource type.
+   * - `true` or omitted: advertise the server generated profiles (US Core).
+   * - `false`: omit `supportedProfile` entirely.
+   * - object: per resource type override, keyed by resource type. Listed types replace the generated
+   *   profiles (an empty array advertises none for that type); unlisted types keep the generated defaults.
    */
-  supportedProfiles?: boolean;
+  supportedProfiles?: boolean | Record<string, string[]>;
 }
 
 export interface SubscriptionAutoDisableTrigger {
--- a/packages/server/src/config/types.ts
+++ b/packages/server/src/config/types.ts
@@ -241,6 +241,47 @@ export interface MedplumServerConfig {
    * as they are necesary for system functionality.
    */
   enabledSearchParameters?: string[];
+
+  /**
+   * Optional customizations to the server generated CapabilityStatement.
+   */
+  capabilityStatement?: MedplumCapabilityStatementConfig;
+}
+
+export interface MedplumCapabilityStatementConfig {
+  /**
+   * Partial CapabilityStatement merged over the server generated statement.
+   * Top level fields replace the generated values.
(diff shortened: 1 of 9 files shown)


codyebberson:
@claude please review

packages/server/src/config/types.ts
```diff
@@ -241,6 +241,47 @@ export interface MedplumServerConfig {
...
+
+  /**
+   * Optional advertised system level interactions, such as `transaction` and `batch`.
+   */
+  systemInteractions?: string[];
+
+  /**
+   * Whether to advertise `supportedProfile` for each resource type. Defaults to true.
+   */
+  supportedProfiles?: boolean;
```
mattwiller: Would users also want to add things to `supportedProfiles`?

mattwiller (approved):

packages/server/src/config/types.ts
codyebberson: mm, good question, let's research

codyebberson:
@claude - should `supportedProfiles` change from `boolean` to a string array?  or to some kind of map of `resourceType -> profiles`?  How invasive would that be to this PR?

codyebberson:
@claude

> Want me to implement it on this branch?

yes, please proceed

codyebberson:
@claude please re-review

PR #10362, opened 2026-08-28
[Scheduling] Find & Book / View times over a date range

### LINKED ISSUE #10276: [Scheduling] Find & Book / Range selection

Filupnot:
Clicking a day on the booking form's calendar searches that day and only that day. The days currently being searched are named in a line under the calendar, along with the gesture for asking about more of them, since dragging and shift-clicking leave nothing on screen to discover them from.

The time finder now has a **Show more days** button at the bottom. When clicked, two more days will be loaded and displayed. The column also now has its own scroll window, so viewing long results won't cause the form to be hidden.

https://github.com/user-attachments/assets/41b28b9f-d4b2-4184-98a6-77ceed9826ea

Only the days not already answered are requested, so the times already on screen stay put while the next ones load. A day that was searched and offered nothing is still listed under its own heading, so it reads as searched rather than skipped. The button has no end state: nothing in a `$find` response says how far ahead there is still something to find, so it stays offered even when every day on screen came back empty.

The booking form's calendar now supports ranged user input; the user can now drag their mouse over a date range, or shift+click a second date, to create a range and view all availability within that date range.

https://github.com/user-attachments/assets/4025c1f0-eeea-483b-b754-7596d0426165

How the calendar marks the days follows how they were asked for: a stretch picked on purpose marks both of its ends, and so does a day clicked on its own once **Show more days** widens the range around it. That needed a change to `CalendarDateInput`, whose `selected` prop previously took priority over `range` outright, so a day clicked kept only itself marked even after the range around it grew past a single day. A multi-day `range` now takes priority instead; `selected` still applies on its own when there is no such range, and is suppressed while a drag is under way.

The day search itself (which days are covered, which windows have already been answered, the `$find` each one costs, and the grouping the form renders) moved out of `AppointmentProposalForm` into a `useDaySearch` hook.

One known rough edge: `$find` has no pagination, so each window asks for a generous 50 times per day it covers. An oversized page is cheap, while an undersized one silently drops the later days in the window. Returning a truncation point from `$find` was raised in #10401 and closed as not planned.

**Out of scope:** Displaying indicators on the calendar for days that have availability, before the user clicks on them. This would definitely be a good user experience improvement, but will likely require separate server changes to `$find` to support a return shape that will not be as costly to the client.

### DIFF
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
@@ -12,6 +12,13 @@ export const MAX_FIND_WINDOW_DAYS = 31;
 
 const MS_PER_DAY = 24 * 60 * 60 * 1000;
 
+/**
+ * The most days one listing will name, however many were searched. A search that
+ * grows a couple of days at a time can outrun a single `$find` window, so the bound
+ * is a year rather than that operation's own.
+ */
+const MAX_LISTED_DAYS = 366;
+
 export type TimeOfDay = 'any' | 'morning' | 'afternoon';
 
 /** A calendar day's worth of available times, split by the actors offering them. */
@@ -165,17 +172,23 @@ export function filterByTimeOfDay(
  *
  * @param appointments - Proposed appointments from `$find`.
  * @param timezone - IANA timezone identifier. Defaults to the browser's.
+ * @param searched - Days to list whether or not they offer anything, so a day that was
+ *   asked about and came back empty says so rather than going missing. Read as calendar
+ *   days on the reader's own clock, which is the calendar a day is picked from.
  * @returns Days in ascending order, each holding its groups.
  */
-export function groupAppointmentsByDay(appointments: readonly Appointment[], timezone?: string): AppointmentDay[] {
+export function groupAppointmentsByDay(
+  appointments: readonly Appointment[],
+  timezone?: string,
+  searched?: DateRange
+): AppointmentDay[] {
   const days = new Map<string, Map<string, Appointment[]>>();
 
   for (const appointment of appointments) {
     if (!appointment.start) {
       continue;
     }
-    const { year, month, day } = getZonedParts(new Date(appointment.start), timezone);
-    const dayKey = `${year}-${pad(month)}-${pad(day)}`;
+    const dayKey = getZonedDayKey(new Date(appointment.start), timezone);
 
     let groups = days.get(dayKey);
     if (!groups) {
@@ -189,6 +202,13 @@ export function groupAppointmentsByDay(appointments: readonly Appointment[], tim
       group.push(appointment);
     } else {
       groups.set(groupKey, [appointment]);
+    }
+  }
+
+  for (const day of enumerateDateRange(searched ?? {}, MAX_LISTED_DAYS)) {
+    const key = getDayKey(day.getFullYear(), day.getMonth() + 1, day.getDate());
+    if (!days.has(key)) {
+      days.set(key, new Map());
     }
   }
 
@@ -260,6 +280,28 @@ export function getDurationMinutes(appointment: Appointment | undefined): number
  */
 export function getAppointmentKey(appointment: Appointment): string {
   return `${appointment.start}/${appointment.end}/${getActorGroupKey(appointment)}`;
+}
+
+/**
+ * Keys the calendar day an instant falls on in a timezone.
+ * @param date - The instant to read.
+ * @param timezone - IANA timezone identifier. Defaults to the browser's.
+ * @returns The day as `YYYY-MM-DD`.
+ */
+function getZonedDayKey(date: Date, timezone: string | undefined): string {
+  const { year, month, day } = getZonedParts(date, timezone);
+  return getDayKey(year, month, day);
+}
+
+/**
+ * Writes a calendar date as the key days are held under.
+ * @param year - The full year.
+ * @param month - The month, from 1.
+ * @param day - The day of the month.
+ * @returns The day as `YYYY-MM-DD`.
+ */
+function getDayKey(year: number, month: number, day: number): string {
+  return `${year}-${pad(month)}-${pad(day)}`;
 }
 
 /**
--- a/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
+++ b/packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
@@ -100,6 +100,15 @@ export function formatDayHeading(date: Date): string {
 }
 
 /**
+ * Names a calendar day without its weekday (e.g. "July 27").
+ * @param date - Local midnight of the day.
+ * @returns The formatted day.
+ */
+export function formatDayLabel(date: Date): string {
+  return new Intl.DateTimeFormat(undefined, { month: 'long', day: 'numeric' }).format(date);
+}
+
+/**
  * Returns how far a timezone is from UTC at a given instant.
  * @param instant - The instant to read the offset at.
  * @param timezone - IANA timezone identifier.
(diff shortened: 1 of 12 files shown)


Filupnot:
@claude review

noahsilas:
It looks like a click results in a 3-day range where the final day is not highlighted, but clicking and dragging to the same range highlights the final day. Is there some intentionality in that experience?

noahsilas:
It seems like we've lost the ability to actually focus in on one exact day; I can get the selected range down to two days with a drag, but can't go narrower than that. (Not necessarily a blocker, just a bit of a surprise)

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
```diff
@@ -183,6 +227,7 @@ export function AppointmentProposalForm(props: AppointmentProposalFormProps): JS
     // a resource changes it would hold whoever is named inside the proposal, and
     // `$book` cannot catch that: the proposal is internally consistent.
     setChosen(undefined);
+    setDaySearch(collapseDays);
```
noahsilas: I found it a bit difficult to understand `collapseDays` here. The idea is that we should reset the search window if someone had clicked "show more" and then changes the filter choices, right? I wonder if we can make this more obvious. It looks like `collapseDays` is only called from this position inside `setDaySearch`, so what about a tweak like this:
```ts
  const resetDaySearch = () => {
    setDaySearch((prev) => ({
      picked: prev.picked,
      range: prev.first,
      first: prev.first,
      found: [],
      extended: false,
    })
  }
```
 - Avoiding the spread operation makes it more obvious to a casual reader; now all the fields are explicitly carried over. This tradeoff is that this is slightly verbose (`{ picked: prev.picked, first: prev.first }` vs. `{ ...prev }`).
 - Naming this "resetDaySearch" feels like it describes the action better; `collapseDays` made it feel like there was some kind of merging of results into per-day buckets or removing entries without results or something.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
```diff
@@ -549,20 +615,67 @@ function getMedicalRecordNumber(patient: WithId<Patient>, mrnSystem: string | un
   );
 }
 
+/** Days a search covers, both ends closed, as `$find` requires. */
+interface SearchWindow {
+  readonly start: Date;
+  readonly end: Date;
+}
```
noahsilas: Should this reuse the `DateTimeRange` type, or is this separate because you foresee it diverging over time?

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
```diff
@@ -145,7 +161,12 @@ export function AppointmentProposalForm(props: AppointmentProposalFormProps): JS
     [searching, windowError, selections]
   );
 
-  const search = useProposedAppointments({ service, combinations, range });
+  const search = useProposedAppointments({
+    service,
+    combinations,
+    range: daySearch.range,
+    count: TIMES_PER_DAY * enumerateDateRange(daySearch.range).length,
```
noahsilas: This makes it seem like you expect results per input day in the range, but the way $find works today is to start at the beginning of the search range and move forward until a full page of results is generated.

So, if you have a 5-day search range, this might ask for 100 options from $find, but they could all end up on the first day of that range.

It sort of seems like deriving the count from the date range length is going to incentivize users to pick wide ranges starting from the day they want in order to get the largest page size of options.

We should probably give a more thorough consideration to pagination from `$find` - I think we had imagined that if a full page is returned, we could display a "find more" button and do a follow up query starting from after the last result's time. It would be better if the server could directly respond with if there are more times available and some kind of pagination token we could use to advance to the right place.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
```diff
@@ -549,20 +623,82 @@ function getMedicalRecordNumber(patient: WithId<Patient>, mrnSystem: string | un
...
+  readonly picked: Date | undefined;
+  /** The window the search opened on, which is what putting the added days away goes back to. */
+  readonly first: SearchWindow;
+  /** The days being asked about now, which is the newest window alone. */
+  readonly range: SearchWindow;
+  /** Times the earlier windows offered, kept on screen while a further one is out. */
+  readonly found: readonly Appointment[];
+  /** Whether days beyond the first window have been asked for. */
+  readonly extended: boolean;
+}
```
noahsilas: This state and the related management is pretty complex. I wonder if there's a way to encapsulate more of this into a smaller component, maybe as a custom hook or something? Intermingled with all the responsibilities of this complex component I find it quite difficult to track it all.

Filupnot:
> It looks like a click results in a 3-day range where the final day is not highlighted, but clicking and dragging to the same range highlights the final day. Is there some intentionality in that experience?

Yeah, my intention in keeping it this way was to have different UI between a range that's auto-created on the user clicking a single day, versus a range that the user explicitly dragged/shift+clicked. I figured it would be good to have different UI between these, but I can go either way - @kevinwadeshaw curious if you have thoughts? 

**EDIT**: Got thoughts. Getting rid of auto-selecting 3 days in favor of just selecting the day that's clicked on.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
Filupnot: We spoke offline. Increased heuristic `TIMES_PER_DAY` from 20 to 50, which should more than cover the returned number of times per day, and be performant enough for our first implementation. Holding off on in-card "Show more" for future.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
Filupnot: Extracted a bunch of this into `useDaySearch.ts`, should be more readable now. There's an opportunity to also split this into different components, but that refactoring would blow up the scope of this PR, so I opted to not do that here.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.tsx
Filupnot: I refactored this a bit, including your suggestion of renaming this to `resetDaySearch`. Should be more readable now.

packages/react-scheduling/src/AppointmentFinder/AppointmentProposalForm.test.tsx
```diff
@@ -358,6 +366,282 @@ describe('AppointmentProposalForm', () => {
...
+      await chooseActor(/provider/i, 'riv', 'Dr. Maya Rivera');
+      await openTimeFinder();
+
+      // Not just the day the form opens on: clicking a later day narrows the same way.
+      await chooseDay('20');
+
+      expect(await screen.findByText(/Thursday, August 20/)).toBeInTheDocument();
+      expect(screen.queryByText(/Friday, August 21/)).not.toBeInTheDocument();
+      expect(new Date(lastFindParams(get)?.get('end') as string).getDate()).toBe(20);
+    });
```
noahsilas: This feels like it could be merged with the first test in this block ("Offers the day picked, and none of the days around it")

packages/react-scheduling/src/AppointmentFinder/useDaySearch.ts
```diff
@@ -0,0 +1,187 @@
...
+      onDaysChanged?.();
+    },
+    [onDaysChanged]
+  );
+
+  // `search.appointments` is the settled result of the current window rather than an
+  // in-flight one, because the control that calls this is disabled while `loadingMoreDays`.
+  const showMoreDays = useCallback((): void => {
+    setDaySearch((previous) => ({
+      first: previous.first,
```
noahsilas: Minor naming #style #nit: what do you think about something like `original` instead of `first`? My original thought on reading was that this might be something like the first day of the range.

Not a blocker, I think that `first` as in "the _first_ range the viewer provided" is fine too.

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
```diff
@@ -290,6 +364,15 @@ export function getNativeInputType(type: 'date' | 'time'): string {
   return import.meta.env.NODE_ENV === 'test' ? 'text' : type;
 }
 
+/**
+ * Returns the first instant of a day, so that a range opens at the top of it.
+ * @param date - Any instant during the day.
+ * @returns Local midnight at the start of that day.
+ */
+export function startOfDay(date: Date): Date {
+  return new Date(date.getFullYear(), date.getMonth(), date.getDate());
```
noahsilas, line 373: Does this need to be timezone aware? As written this will give you the start of the date in the browser's timezone, which I think works for now, since so far we only show the UI in the brower's TZ, right?

noahsilas (approved):

packages/react-scheduling/src/AppointmentFinder/AppointmentFinder.times.ts
Filupnot: Correct. If we change that we can change this down the road.

PR #10367, opened 2026-08-29
Estimate btree index bloat instead of expensive exact calculation

mattlong:
Previous approach not viable for a 60s timeout nor was it totally necessary; an estimate is fine for our purposes


codyebberson:
@claude please review

codyebberson (approved):

mattlong:
@claude take another look

PR #10368, opened 2026-08-29
Add useStabilizedCallback hook

noahsilas:
Components and hooks in this libraries are intended to be safely and easily used by the consuming applications. One nicety that we provide in a number of places is the ability to provide a callback that is unstable, most typically as a result of an inline arrow function:

```tsx
  <SignatureInput onChange={(signature) => doWork(signture)} />
```

When we want to use those callbacks outside of the immediate render, we can run into rough edges - as input to dependency arrays of other hooks, this can cause those to also re-compute every render.

To that end, we often use a pattern of capturing the latest version of these callbacks into a Ref, which lets us opt them out of the dependency arrays. That works, but it is not very intention revealing, and it can be hard to understand why the Ref is being used without verbose comments. There also were a number of variants used around the codebase: render-time ref assignment, useEffect, or useLayoutEffect.

In this commit we extract that latest-ref pattern into a reusable hook. The hook updates a ref in useInsertionEffect, which runs before every layout and passive effect in the commit, so a callback fired from a descendant effect is never a render behind.

Converts useMedicationIFrame, useSubscription, useWhisper, ListWithDetailPane, QrCodeScanner, and SignatureInput to use the new hook.

### DIFF
--- a/packages/react-hooks/src/index.ts
+++ b/packages/react-hooks/src/index.ts
@@ -17,6 +17,7 @@ export * from './useResource/useResource';
 export * from './useResourceBoard/useResourceBoard';
 export * from './useResourceModified/useResourceModified';
 export * from './useSearch/useSearch';
+export * from './useStabilizedCallback/useStabilizedCallback';
 export * from './useSubscription/useSubscription';
 export * from './useSyncOrderSet/useSyncOrderSet';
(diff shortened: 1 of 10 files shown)


noahsilas:
@claude pkease review

The-Alchemist (approved):

